PR #3206, opened 2022-05-23
Upgrade in-memory and on-disk state representation with tree states

michaelsproul:
This is an umbrella PR for fundamental changes to Lighthouse's state representation.

The plan is to merge smaller PRs into this one, and then merge it to `unstable` without squashing.

## TODO List

There are few things that need fixing on this branch before it's ready for primetime:

- [x] Remove recursion from `get_state`. The recursion is liable to overflow the stack, so we should switch to an iterative algorithm.
- [x] Fix the validator monitor
- [x] Clean up the state processing optimisations
- [x] Fuzz the state processing optimisations
- [x] Add or fix state processing optimisations for phase0 so that block replay isn't too slow
- [x] Add new metrics for `jemalloc`
- [x] Resolve all remaining `FIXME(sproul)`s in the code
- [x] Re-do the optimisation that was reverted in #3658

## Issues Addressed

Closes #2806
Closes #2805
Closes #2944
Closes #2371
Closes #3208
Closes #3505

PR #5533, opened 2024-04-08
In-memory tree states

michaelsproul:
## Issue Addressed

This is the memory-only portion of `tree-states`, i.e. without any database schema changes.

- https://github.com/sigp/lighthouse/pull/3206

## Proposed Changes

- [x] Make `BeaconState` cheap to clone thanks to persistent data structures from [milhouse](https://github.com/sigp/milhouse).
- [x] Replace the snapshot cache with a new state cache inside the store which holds 32x more states for around the same memory cost :tada:
- [x] Delete the now-unnecessary `block_production_state`.
- [x] Delete `StateProcessingStrategy`. This could cause us to do more tree hashing in the case of an attester shuffling cache miss, but this is mitigated by 1) tree-states has faster tree hashing due to structural sharing, 2) the state cache is likely to hit even when the shuffling cache misses. It is less risky to remove the concept of inconsistent states altogether so that they can't end up in the new `state_cache` and poison it. We can re-optimise this codepath in future by adopting the `HotStateSummary` changes from full `tree-states`.


beacon_node/beacon_chain/src/state_advance_timer.rs
```diff
@@ -243,86 +253,12 @@ async fn state_advance_timer<T: BeaconChainTypes>(
...
-                                    <= current_slot
-                                {
-                                    drop(cache.take());
-                                }
-                            }
-                        }
-
-                        // Signal block proposal for the next slot (if it happens to be waiting).
-                        if let Some(tx) = &beacon_chain.fork_choice_signal_tx {
-                            if let Err(e) = tx.notify_fork_choice_complete(next_slot) {
```
michaelsproul, line 303: Not to self: should put this back

michaelsproul:
Some benchmarks as of `unstable` @ f5e0404fb8c5ad924b5d0f0f4e953f7a3778e327 and `tree-states-memory` @ bca33f1f5fd3f3ed373f78887820e04e604989da

| Slot (offset) | `unstable` total (slot/sigs/core/hash) | `tree-states-memory` total (slot/sigs/core/hash) |
| ------------- | -------------------------------- | ------------------------------  |
| 8867873 (1)   | 56.5ms (0ms/11.1ms/7.1ms/38.1ms) | 39.8ms (0ms/8.9ms/21.0ms/9.8ms) |
| 8867872 (0)   | 360.7ms (274.7ms/8.8ms/8.6ms/68.5ms) | 654.4 (493.2ms/11.0ms/19.1ms/130.96ms) |

This shows that we're ~30% faster processing average blocks (offsets 1-31) compared to unstable, due to the faster tree hashing.

However, compared to `unstable` with single-pass and _no tree-states_ we are slower at epoch boundaries. That's 81% slower to be precise :scream: Disregarding the slot-processing, which should happen as part of state advance, we have 86ms for unstable vs 161ms for tree-states, which is 87% slower.

This is not really a fair comparison however, due to single-pass, so I am in the process of re-running the benchmarks against `stable`.

michaelsproul:
Ok the epoch boundary case is similar on `stable` (345ms total, similar breakdown to `unstable`), while mid-epoch block processing is substantially slower (73ms).

I'm going to check there isn't something wrong with the single-pass caches.

michaelsproul:
Ok, the issue was in my benchmarking technique. For `tree-states`, the slot advance was creating a lot of pending updates, which were being applied in one big batch. A more realistic scenario is that these pending updates are applied after epoch processing as part of the state advance. By using `lcli skip-slots` to get the pre-state at slot 8867872 I got what I think is a more representative result:

```
[2024-04-16T05:21:46Z DEBUG lcli::transition_blocks] Slot processing: 310ns
[2024-04-16T05:21:46Z DEBUG lcli::transition_blocks] Build all caches (again): 1.89µs
[2024-04-16T05:21:46Z DEBUG lcli::transition_blocks] Batch verify block signatures: 10.485968ms
[2024-04-16T05:21:46Z DEBUG lcli::transition_blocks] Process block: 22.432465ms
[2024-04-16T05:21:46Z DEBUG lcli::transition_blocks] Post-block tree hash: 9.284713ms
[2024-04-16T05:21:46Z INFO  lcli::transition_blocks] Run 9: 42.259264ms
```

This is once again competitive with `unstable` (actually faster). So we can conclude that tree-states will almost always be faster, except in the case of a failed state advance. Even in the case the state is not advanced, that usually implies, there's a reorg or several skipped slots. Tree states is more likely to do well in these scenarios because it can clone the parent state faster and doesn't need to worry about snapshot cache misses.

michaelsproul:
Logging on one of our Holesky nodes running this branch shows an uptick in attestation queues becoming full, and this log:

> ERRO Failed to maintain availability cache   error: DecodeError(OffsetOutOfBounds(1157628160)), service: beacon

I need to dig into why the `DecodeError` is happening. It's definitely a bug.

michaelsproul:
Bug was actually in `unstable`, fixed here:

- https://github.com/sigp/lighthouse/pull/5598

Cargo.toml
```diff
@@ -106,6 +106,7 @@ c-kzg = "1"
 clap = "2"
 compare_fields_derive = { path = "common/compare_fields_derive" }
 criterion = "0.3"
+crossbeam-channel = "0.5.8"
```
realbigsean: this is up to `0.5.12` should we perhaps define it as `0.5`

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -1864,45 +1863,10 @@ fn load_parent<T: BeaconChainTypes, B: AsBlock<T::EthSpec>>(
         return Err(BlockError::ParentUnknown(block.into_rpc_block()));
     }
 
-    let block_delay = chain
```
realbigsean, line 1867: should this have been removed?

beacon_node/beacon_chain/src/builder.rs
```diff
@@ -520,6 +519,12 @@ where
         let (_, updated_builder) = self.set_genesis_state(genesis_state)?;
         self = updated_builder;
 
+        // Build the committee caches before storing. The database assumes that states have
+        // committee caches built before storing.
+        weak_subj_state
+            .build_all_committee_caches(&self.spec)
```
realbigsean: Do we need to call `build_all_committe_caches` here if we are calling `build_all_caches` on L464?

testing/ef_tests/src/cases/ssz_generic.rs
```diff
@@ -125,10 +125,16 @@ impl Case for SszGeneric {
...
+                if length == "0" {
+                    log_file_access(self.path.join("serialized.ssz_snappy"));
+                    return Ok(());
+                }
+
                 type_dispatch!(
                     ssz_generic_test,
                     (&self.path),
-                    FixedVector,
+                    Vector,
```
realbigsean, line 137: Should we be testing `Vector` and `FixedVector` since we're still using both?

consensus/types/src/chain_spec.rs
```diff
@@ -309,6 +309,13 @@ impl ChainSpec {
         }
     }
 
+    /// Return the name of the fork activated at `slot`, if any.
+    pub fn fork_activated_at_slot<E: EthSpec>(&self, slot: Slot) -> Option<ForkName> {
```
realbigsean: Not seeing this used anywhere

consensus/types/Cargo.toml
```diff
@@ -52,6 +52,8 @@ serde_json = { workspace = true }
 smallvec = { workspace = true }
 maplit = { workspace = true }
 strum = { workspace = true }
+milhouse = { workspace = true }
+rpds = "0.11.0"
```
realbigsean: Same comment as elsewhere about potentially moving this to a workspace dep

consensus/state_processing/src/common/initiate_validator_exit.rs
```diff
@@ -30,13 +30,14 @@ pub fn initiate_validator_exit<E: EthSpec>(
         exit_queue_epoch.safe_add_assign(1)?;
     }
 
-    let validator = state.get_validator_mut(index)?;
+    let validator = state.get_validator_cow(index)?;
```
realbigsean, line 33: Is there a reason to get a `Cow` here when it's made `into_mut` a couple lines later as opposed to using the getter that gets a mutable reference?

consensus/state_processing/Cargo.toml
```diff
@@ -28,6 +28,7 @@ arbitrary = { workspace = true }
 lighthouse_metrics = { workspace = true }
 lazy_static = { workspace = true }
 derivative = { workspace = true }
+vec_map = "0.8.2"
```
realbigsean: Should this be a workspace dep? Honestly not sure whether our policy is all deps to workspace, or only deps used in multiple crates

beacon_node/store/src/state_cache.rs
```diff
@@ -0,0 +1,301 @@
...
+pub struct SlotMap {
+    slots: BTreeMap<Slot, Hash256>,
+}
+
+#[derive(Debug)]
+pub struct StateCache<E: EthSpec> {
+    finalized_state: Option<FinalizedState<E>>,
+    states: LruCache<Hash256, BeaconState<E>>,
+    block_map: BlockMap,
+    capacity: NonZeroUsize,
```
realbigsean: If this is just tracking the `LruCache` capacity, `LruCache` has a `cap` method that could be used instead

beacon_node/store/src/state_cache.rs
```diff
@@ -0,0 +1,301 @@
...
+                state_root,
+                slot: state.slot(),
+            });
+        }
+
+        // Update the cache's idea of the max epoch.
+        self.max_epoch = std::cmp::max(state.current_epoch(), self.max_epoch);
+
+        // If the cache is full, use the custom cull routine to make room.
+        if let Some(over_capacity) = self.len().checked_sub(self.capacity.get()) {
```
realbigsean: If this is just comparing the `LruCache` length vs capacity, I don't see how this condition could be hit, so I don't know that this custom `cull` method is actually being used.

consensus/types/src/beacon_state.rs
```diff
@@ -531,6 +628,21 @@ impl<E: EthSpec> BeaconState<E> {
         Ok(self.pubkey_cache().get(pubkey))
     }
 
+    /// Immutable variant of `get_validator_index` which errors if the cache is not up to date.
+    pub fn get_validator_index_read_only(
```
realbigsean: Not seeing this used anywhere

consensus/types/src/beacon_state.rs
```diff
@@ -207,16 +219,99 @@ impl From<BeaconStateHash> for Hash256 {
...
+                groups(tree_lists)
+            )),
+            num_fields(all()),
+        )),
+        Merge(metastruct(
+            mappings(
+                map_beacon_state_bellatrix_fields(),
+                map_beacon_state_bellatrix_tree_list_fields(mutable, fallible, groups(tree_lists)),
+            ),
+            bimappings(bimap_beacon_state_merge_tree_list_fields(
```
realbigsean: The other map macros here use `bellatrix`, we should probably keep it consistent

consensus/types/src/beacon_state.rs
```diff
@@ -1834,96 +1920,51 @@ impl<E: EthSpec> BeaconState<E> {
         *self.pubkey_cache_mut() = PubkeyCache::default()
     }
 
+    pub fn has_pending_mutations(&self) -> bool {
```
realbigsean, line 1914: This seems like it'd be a really good use case for `metastruct` mappings. So we wouldn't forget to update this with new fields. I tried out making an immutable an infallibe mapping for this and it's compiling for me:

```
    pub fn has_pending_mutations(&self) -> bool {
        let mut any_pending_mutations = false;
        match &self {
            Self::Base(self_inner) => {
                map_beacon_state_base_tree_list_fields_immutable!(self_inner, |_, self_field| {
                    if self_field.has_pending_updates() {
                        any_pending_mutations = true;
                    }
                });
            }
            Self::Altair(self_inner) => {
                map_beacon_state_altair_tree_list_fields_immutable!(self_inner, |_, self_field| {
                    if self_field.has_pending_updates() {
                        any_pending_mutations = true;
                    }
                });
            }
            Self::Merge(self_inner) => {
                map_beacon_state_bellatrix_tree_list_fields_immutable!(
                    self_inner,
                    |_, self_field| {
                        if self_field.has_pending_updates() {
                            any_pending_mutations = true;
                        }
                    }
                );
            }
            Self::Capella(self_inner) => {
                map_beacon_state_capella_tree_list_fields_immutable!(
                    self_inner,
                    |_, self_field| {
                        if self_field.has_pending_updates() {
                            any_pending_mutations = true;
                        }
                    }
                );
            }
            Self::Deneb(self_inner) => {
                map_beacon_state_deneb_tree_list_fields_immutable!(self_inner, |_, self_field| {
                    if self_field.has_pending_updates() {
                        any_pending_mutations = true;
                    }
                });
            }
            Self::Electra(self_inner) => {
                map_beacon_state_electra_tree_list_fields_immutable!(
                    self_inner,
                    |_, self_field| {
                        if self_field.has_pending_updates() {
                            any_pending_mutations = true;
                        }
                    }
                );
            }
        };
        any_pending_mutations
    }
```

realbigsean (changes requested):
This is looking awesome! No snapshot cache is really nice!

beacon_node/beacon_chain/src/block_verification.rs
michaelsproul: yeah this was previously only used for deciding whether to clone from the snapshot cache, and logging associated with that

consensus/state_processing/src/common/initiate_validator_exit.rs
michaelsproul: Hmm I think it was in case the validator is already exited and we hit the `return Ok(())` case. But I don't imagine that happens very often

realbigsean: ndb either way

consensus/state_processing/Cargo.toml
michaelsproul: deleted this as it wasn't actually used explicitly

beacon_node/store/src/state_cache.rs
michaelsproul: If `len == capacity` then we'll get `over_capacity == 0`, and will cull `over_capacity + 1 == 1`, right?

Maybe we should try increasing the number of states that are culled (the +1)

consensus/types/src/beacon_state.rs
michaelsproul: done!

beacon_node/store/src/state_cache.rs
michaelsproul: nice, fixed

michaelsproul:
I pushed a commit 970f3dfc6eca93bbdbf11b89e6c93157eb693bcd which seems to have made memory usage less spikey on a restart (see 12:30 in chart). The problem previously was that the head state would not get cached in the `state_cache` on startup, creating the possibility for a cache-miss and reload from disk, which would increase memory usage due to the lack of full structural sharing.

spikeyness

The memory usage is still a bit spikier than I'd like, and this is despite a complete absence of cache misses. This may be memory usage unrelated to the state cache.

Even so, tree-states is using consistently less memory with lower spikes than `stable`:

bn_memory_comparison

The yellow line at the bottom is this branch, the rest are running variants of stable/unstable.

beacon_node/store/src/state_cache.rs
realbigsean: oh right `len == capacity` 😂 my bad

> Maybe we should try increasing the number of states that are culled (the +1)

No, I think it's fine how it is, it's sort of just customizing the lru cache's prune mechanic which makes sense.

realbigsean (approved):
🚀

realbigsean:
@mergify queue

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -4529,10 +4404,11 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
                 let block = self
                     .get_blinded_block(&parent_block_root)?
                     .ok_or(Error::MissingBeaconBlock(parent_block_root))?;
-                let state = self
-                    .get_state(&block.state_root(), Some(block.slot()))?
+                let (state_root, state) = self
```
jimmygchen, line 4407: I think the log above needs to be updated to remove snapshot cache:

https://github.com/sigp/lighthouse/blob/970f3dfc6eca93bbdbf11b89e6c93157eb693bcd/beacon_node/beacon_chain/src/beacon_chain.rs#L4398-L4403

jimmygchen:
Some minor comments:
- [The description for `--state-cache-size`](https://github.com/sigp/lighthouse/blob/6bac5ce12b94c397ee0a7d5423a0eeab814c3682/beacon_node/src/cli.rs#L623-L629) needs an update i think - replace the "snapshot cache" and also update the default to `128`.
- Remove snapshot cache related metrics [here](https://github.com/sigp/lighthouse/blob/de7ae83abc3cb657453eb730b18eaf114336d6b3/beacon_node/beacon_chain/src/metrics.rs#L35-L46)

realbigsean:
@mergify dequeue

realbigsean:
sorry @jimmygchen didn't realize you were mid-review!

jimmygchen:
All good, happy to merge this as it's gone through some testing and you've already done a round of review!
I'm just going through the changes to help me understand them, it'd probably take me a while. I haven't found anything major that's worth stopping the mege, I think we're good to merge and address further comments in a separate PR? 

I'd be keen to get this in so 
- we avoid further code conflicts
- others working on `unstable` can help discover potential issues not yet identified

realbigsean:
@jimmygchen has expressed offline he's good with merging and allowing feedback resolution in a followup PR so as to avoid conflicts

realbigsean:
@mergify queue

realbigsean:
@mergify requeue

beacon_node/beacon_chain/src/metrics.rs
```diff
@@ -1188,15 +1184,10 @@ pub fn scrape_for_metrics<T: BeaconChainTypes>(beacon_chain: &BeaconChain<T>) {
...
-        .try_write_for(SNAPSHOT_CACHE_TIMEOUT)
-    {
-        set_gauge(
-            &BLOCK_PROCESSING_SNAPSHOT_CACHE_SIZE,
-            snapshot_cache.len() as i64,
-        )
-    }
+    set_gauge_by_usize(
+        &BLOCK_PROCESSING_SNAPSHOT_CACHE_SIZE,
+        beacon_chain.store.state_cache_len(),
```
jimmygchen, line 1189: Can we rename the metrics to `BLOCK_PROCESSING_STATE_CACHE_SIZE` instead?

PR #5530, opened 2024-04-05
Block lookup tests with `SyncTester`

jimmygchen:
## Issue Addressed

This PR adds a new test utility for Lighthouse sync/lookup and a regression test that uses it. 

Currently we're lacking integration tests around sync components and we're finding it difficult to reproduce production issues (especially ones that are related to race conditions). Hopefully this test utility would help increase coverage in this area and make reproducing sync/lookup related issues easier.

## Proposed Changes

- Add a new [`SyncTester`](https://github.com/sigp/lighthouse/blob/6bfeaba2b31b9e58865c0cddb85246c0a8c01d5b/beacon_node/network/src/sync/testing/mod.rs#L60) test utility. This takes a lot of inspiration from the existing sync [`TestRig`](https://github.com/sigp/lighthouse/blob/857f94faa140eda249841f37e44a72fc632f222d/beacon_node/network/src/sync/block_lookups/tests.rs#L29), and [`DenebTester`](https://github.com/sigp/lighthouse/blob/857f94faa140eda249841f37e44a72fc632f222d/beacon_node/network/src/sync/block_lookups/tests.rs#L1178), with the intention of making it more flexible and event-driven - perhaps we can eventually consolidate them.
- Add a [regression test](https://github.com/sigp/lighthouse/blob/6bfeaba2b31b9e58865c0cddb85246c0a8c01d5b/beacon_node/network/src/sync/block_lookups/tests.rs#L2140-L2202) written using the `SyncTester`.
- I removed the [duplicate instantiation of `BeaconChainHarness` in `NetworkBeaconProcessor::null_for_testing`](https://github.com/sigp/lighthouse/pull/5530/files#diff-ddaed1f961430fe8a42862e24095790bad14352cb24509fbf6f64da0f2c0974fR633), and this shaved off ~300ms from the above tests (~50%). I assume this would speed up existing `DenebTester` tests too.
- ~~Extracted public methods of `RangeSync`, `BackFillSync` and `BlockLookups` into traits (no logic change there), so we can swap out some components out during testing.~~ (not required, thanks @dapplion!)

## Additional Info

The `SyncTester` enables sending messages to `SyncManager`, and then making assertions on outgoing messages from it.

It covers the following:
1. Sending `SyncMessage` to `SyncManager` to trigger `RangeSync`, `BackFillSync` and `BlockLookups` behaviours.
2. Making assertions on `WorkEvent`s received from sync
3. Making assertion on `NetworkMessage` received from sync (e.g. outgoing RPC requests).

[...]

PR #5534, opened 2024-04-08
Event-based block lookup tests

dapplion:
## Issue Addressed

Extends @jimmygchen refactor by
- Applying the changes to all existing block lookup tests
- Injects events directly on handle_message to avoid async machinery :)

This PR does not:
- Change the syntatic style of existing tests
- Modify or extend the tested scenarios. However it adds helper functions to make them more succint

Switching to event-based tests is very valuable to generalize tests, allow to test more diverse edge cases, and allow refactors without being too tied to block lookup internals.

This PR is an alternative to 
- https://github.com/sigp/lighthouse/pull/5530

## Proposed Changes

All tests now interface with sync through this event channels

```
                     +-----------------+
                     | BeaconProcessor |
                     +---------+-------+
                            ^  |
                            |  |
                  WorkEvent |  | SyncMsg
                            |  | (Result)
                            |  v
+--------+            +-----+-----------+             +----------------+
| Router +----------->|  SyncManager    +------------>| NetworkService |
+--------+  SyncMsg   +-----------------+ NetworkMsg  +----------------+
          (RPC resp)  |  - RangeSync    |  (RPC req)
                      +-----------------+
                      |  - BackFillSync |
                      +-----------------+
                      |  - BlockLookups |
                      +-----------------+
```

## Additional Info

RE: style we can discuss as follow-up to this PR, I see as the contentious options:

- Use declarative "state-less" API (pre-deneb tests)
- Use a chained builder-stype API for tests with internal implicit state (post-deneb)

Some notes to inform the above:

**chained API doesn't play well with backtrace**

Consider this code, and a panic in action_a

```
10: rig
11:   .action_a()
12:   .action_b()
13:   .action_a()
```

The backtrace line for `action_a` will declare line 10. You can't know which of the two action_a invocations caused the error

**internal implicit state succinctness**

Post-deneb test are shorter at the expense of clarity. There's machinery in the back that makes assumptions about what request is doing what. For straightforward cases it's fine but it can be restrictive for more complex cases. Pre-deneb tests with proper helper functions should look as succint as those with a chained API. We can remane `rig` to `r` and it almost looks like chained, which allowing conditional flow and storing variables.

**Comments**

Any sync tests are quite difficult to understand for those that have not written them. Having comments explaining the case is very important for either style. In that direction I think we should focus much more on having the test easy to understand and well documented with comments besides succintness


beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -743,14 +764,19 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
...
         let request_state = R::request_state_mut(&mut lookup);
 
-        let Ok(peer_id) = request_state.get_state().processing_peer() else {
-            return;
+        let peer_id = match request_state.get_state().processing_peer() {
+            Ok(peer_id) => peer_id,
+            Err(e) => {
+                trace!(self.log, "Attempting to process single block lookup in bad state"; "id" => target_id, "response_type" => ?R::response_type(), "error" => e);
+                return;
+            }
```
dapplion: Added this two traces to debug tests, as they were super helpful. This cases should never happen in practice and if they do, important events are ignored. I think they should be at least debug, thoughts?

realbigsean: Yes, debug sounds good to me

beacon_node/network/src/sync/manager.rs
```diff
@@ -619,6 +650,7 @@ impl<T: BeaconChainTypes> SyncManager<T> {
             SyncMessage::UnknownParentBlock(peer_id, block, block_root) => {
                 let block_slot = block.slot();
                 let parent_root = block.parent_root();
+                debug!(self.log, "received sync_message"; "message" => "UnknownParentBlock", "block_root" => %block_root, "parent_root" => %parent_root);
```
realbigsean:
```suggestion
                debug!(self.log, "Received sync message"; "message" => "UnknownParentBlock", "block_root" => %block_root, "parent_root" => %parent_root);
```

beacon_node/network/src/sync/manager.rs
```diff
@@ -638,6 +670,7 @@ impl<T: BeaconChainTypes> SyncManager<T> {
                 }
                 let mut blobs = FixedBlobSidecarList::default();
                 *blobs.index_mut(blob_index as usize) = Some(blob);
+                debug!(self.log, "received sync_message"; "message" => "UnknownParentBlob", "block_root" => %block_root, "parent_root" => %parent_root);
```
realbigsean:
```suggestion
                debug!(self.log, "Received sync message"; "message" => "UnknownParentBlob", "block_root" => %block_root, "parent_root" => %parent_root);
```

beacon_node/network/src/sync/manager.rs
```diff
@@ -646,11 +679,12 @@ impl<T: BeaconChainTypes> SyncManager<T> {
                     ChildComponents::new(block_root, None, Some(blobs)),
                 );
             }
-            SyncMessage::UnknownBlockHashFromAttestation(peer_id, block_hash) => {
+            SyncMessage::UnknownBlockHashFromAttestation(peer_id, block_root) => {
                 // If we are not synced, ignore this block.
                 if self.synced_and_connected(&peer_id) {
+                    debug!(self.log, "received sync_message"; "message" => "UnknownBlockHashFromAttestation", "block_root" => %block_root);
```
realbigsean:
```suggestion
                    debug!(self.log, "Received sync message"; "message" => "UnknownBlockHashFromAttestation", "block_root" => %block_root);
```

beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -29,15 +31,44 @@ type T = Witness<ManualSlotClock, CachingEth1Backend<E>, E, MemoryStore<E>, Memo
...
+//                             |  v
+// +--------+            +-----+-----------+             +----------------+
+// | Router +----------->|  SyncManager    +------------>| NetworkService |
+// +--------+  SyncMsg   +-----------------+ NetworkMsg  +----------------+
+//           (RPC resp)  |  - RangeSync    |  (RPC req)
+//                       +-----------------+
+//                       |  - BackFillSync |
+//                       +-----------------+
+//                       |  - BlockLookups |
+//                       +-----------------+
```
realbigsean: Since half is doc format and half is comment format this looks weird in my IDE, not sure if this is intentional: 

Screenshot 2024-04-08 at 6 49 35 PM

```suggestion
///                      | BeaconProcessor |
///                      +---------+-------+
///                             ^  |
///                             |  |
///                   WorkEvent |  | SyncMsg
///                             |  | (Result)
///                             |  v
/// +--------+            +-----+-----------+             +----------------+
/// | Router +----------->|  SyncManager    +------------>| NetworkService |
/// +--------+  SyncMsg   +-----------------+ NetworkMsg  +----------------+
///           (RPC resp)  |  - RangeSync    |  (RPC req)
///                       +-----------------+
///                       |  - BackFillSync |
///                       +-----------------+
///                       |  - BlockLookups |
///                       +-----------------+
```

beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -56,91 +87,456 @@ impl TestRig {
...
+        let r = Self::test_setup();
+        if r.after_deneb() {
+            Some(r)
+        } else {
+            None
+        }
+    }
+
+    fn after_deneb(&self) -> bool {
+        matches!(self.current_fork(), ForkName::Deneb | ForkName::Electra)
```
realbigsean: Here's a place fork order would be useful, so we don't forget to enable these in future forks

beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -188,988 +601,557 @@ impl TestRig {
...
         if i < parent_lookup::PARENT_FAIL_TOLERANCE {
-            assert_eq!(
-                bl.parent_lookups[0]
-                    .current_parent_request
-                    .block_request_state
-                    .state
-                    .failed_attempts(),
-                dbg!(i)
-            );
+            // assert_eq!(
```
realbigsean: Are you able to add this back?

beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -188,988 +601,557 @@ impl TestRig {
...
```
realbigsean: This too

realbigsean (commented):
Love the move to event based testing, thanks @jimmygchen and @dapplion !

The test do already look more straightforward and looks like there are some logging improvements in here. 

> stateless vs. builder-ish API

This PR is worth merging regardless of where we land there. 

> Any sync tests are quite difficult to understand for those that have not written them. Having comments explaining the case is very important for either style. In that direction I think we should focus much more on having the test easy to understand and well documented with comments besides succintness

Agree! My fault for lack of comments on a lot of the existing ones.

beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -56,91 +87,456 @@ impl TestRig {
...
+                .blob_ids
+                .to_vec()
+                .iter()
+                .all(|r| r.block_root == for_block) =>
+            {
+                Some(*id)
+            }
+            _ => None,
+        })
+        .expect(&format!("Expected blob parent request for {for_block:?}"))
```
realbigsean: There's a few instances of this failing the linter
```suggestion
        .unwrap_or_else(|_| panic!("Expected blob parent request for {for_block:?}"))
```

beacon_node/network/src/sync/block_lookups/tests.rs
dapplion: I don't think we should. The goal of this PR is to decouple tests from the internals of block lookups. This assertion is implicitly asserted in batch of the end of the test. What we really care about is that after N failures, the lookup is dropped, regardless of how accounting is done internally. I think I left the comment to double check with you, but I would vote to remove

beacon_node/network/src/sync/manager.rs
```diff
@@ -233,31 +233,62 @@ pub fn spawn<T: BeaconChainTypes>(
...
+                beacon_processor.clone(),
+                beacon_chain.clone(),
+                log.clone(),
+            ),
+            range_sync: RangeSync::new(beacon_chain.clone(), log.clone()),
+            backfill_sync: BackFillSync::new(beacon_chain.clone(), network_globals, log.clone()),
+            block_lookups: BlockLookups::new(
+                beacon_chain.data_availability_checker.clone(),
+                log.clone(),
+            ),
```
jimmygchen, line 272: Unrelated to this PR but since you're touching this I thought it might be worth mentioning - perhaps we could create child logger for each of these individual components so we know where the log is coming from?

`log.new(slog::o!("service" => "range_sync")`

jimmygchen: ps. i'm not suggesting a change in this PR

dapplion:
Moved the logging changes into their own PR
- https://github.com/sigp/lighthouse/pull/5535

beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -29,15 +31,44 @@ type T = Witness<ManualSlotClock, CachingEth1Backend<E>, E, MemoryStore<E>, Memo
 struct TestRig {
     beacon_processor_rx: mpsc::Receiver<WorkEvent<E>>,
     network_rx: mpsc::UnboundedReceiver<NetworkMessage<E>>,
+    network_rx_queue: Vec<NetworkMessage<E>>,
     rng: XorShiftRng,
     harness: BeaconChainHarness<T>,
+    sync_manager: SyncManager<T>,
+    beacon_processor: Arc<NetworkBeaconProcessor<T>>,
```
jimmygchen: Any reasons to have `beacon_processor` in the `TestRig`? I don't see it being used other than accessing `network_globals`.

I kept it out because the test rig should not need to directly interact with the `beacon_processor` directly in any way, otherwise it could interfere testing of the `SyncManager`.

jimmygchen: Could we have some of the field documentation back as I think it would help people writing the tests understand what each of these fields are for, without reading through the implementation.

https://github.com/sigp/lighthouse/pull/5530/files#diff-13439a1c05bab56ff7fb8b91e2d4f569076456e8a909000e1d9d3b5df8afb1b0R59-R81

dapplion: Added!

dapplion: Good point! Removed beacon_processor for network_globals, and harness for fork_name

beacon_node/network/src/sync/manager.rs
dapplion: Does this label then apply to all logs? That's a good suggestion, can add to a future PR def

beacon_node/network/src/sync/block_lookups/tests.rs
realbigsean: let's remove 🚀

realbigsean (approved):

realbigsean:
@mergifyio queue

realbigsean:
@Mergifyio refresh

realbigsean:
@mergifyio requeue

jimmygchen:
@mergify requeue

PR #5543, opened 2024-04-09
Beta compiler fixes

realbigsean:
## Issue Addressed

- duplicate imports are being flagged in some tests where we use `use super::*;`
- there's also a new lint about defining trait bounds in two places, this only affects us in one place


pawanjay176 (approved):

realbigsean:
@mergifyio queue

Issue #3651, opened 2022-10-21
Light client server support [tracking issue]


michaelsproul:
## Description

We should support the full node side of the light client protocol.

The components we need for this are:

### Groundwork

- [x] Light client types in `consensus/types`. This includes `LightClientBootstrap`, `LightClientUpdate` and all the other types defined [here](https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/sync-protocol.md#containers).
- [x] Ability to construct merkle proofs for the `state.finalized_checkpoint.root`, `state.current_sync_committee` and `state.next_sync_committee`. 
- [x] CLI flags to enable / disable the light client server (#3714)
- [x] #4926
- [x] #4896 
- [x] #4897 

### P2P (Gossip & ReqResp)

- [x] Support the new gossip topics defined in the [P2P spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/p2p-interface.md)
  - [x]  #3693 
- [x] Support the new network RPCs defined in the P2P spec
  - [x] #3886 
  - [x] #3849
  - [x] #3711 

### REST APIs

- [x] #3954 
- [x] #4895 

### Optimization

- [x] Optimise light client bootstrap and range
  - [x] #5545 
  - [x] #5915

### Production: enabling Light Client server by default

- [ ] #4940
- [x] Measure performance impact and additional resource requirement
- [ ] Review inbound request quota config
- [ ] Evaluate whether changes to peer count / inbound peer ratio is required
- [x] Enable light client server by default

## Additional Details

- **Constructing merkle proofs**: I think it will be easiest if we pursue a non-generic approach for this initially, as a generic approach would require digging into the guts of [`tree_hash_derive`](https://github.com/sigp/lighthouse/blob/stable/consensus/tree_hash_derive/src/lib.rs) to add methods for generating merkle proofs based on field indices. We probably want that eventually, but a non-generic approach can inform the generic implementation. If we implement this atop `unstable` we probably want to integrate with the [`BeaconTreeHashCache`](https://github.com/sigp/lighthouse/blob/stable/consensus/types/src/beacon_state/tree_hash_cache.rs) so that we don't need to re-hash the whole state for the intermediate nodes of the proof. Alternatively we could implement it without caching and rely on the arrival of [`tree-states`](https://github.com/sigp/lighthouse/pull/3206>) to improve performance (because caching is then built in to each field's `tree_hash_root()` method).
[...]

PR #5545, opened 2024-04-09
Persist light client updates

eserilev:
## Issue Addressed

Partially #3651

## Proposed Changes

This PR aims to persist light client updates. Once persisted, this will allow lighthouse nodes to serve them over RPC.


beacon_node/beacon_chain/src/light_client_server_cache.rs
```diff
@@ -152,6 +168,75 @@ impl<T: BeaconChainTypes> LightClientServerCache<T> {
...
+                "Beacon state not available {:?}",
+                attested_block.state_root()
+            )))?;
+
+        let mut attested_state = store
+            .get_state(&attested_block.state_root(), Some(attested_block.slot()))?
+            .ok_or(BeaconChainError::DBInconsistent(format!(
+                "Attested state not available {:?}",
+                attested_block.state_root()
+            )))?;
```
dapplion: This not acceptable w.r.t performance. You should extend the use of cached_parts such that you never have to get states from the store in the happy path. The function to generate updates in `LightClientUpdate::new` should not be used.

See this document for an implementation suggestion https://hackmd.io/@dapplion/BkxPbc6eQT

beacon_node/beacon_chain/src/light_client_server_cache.rs
```diff
@@ -152,6 +197,84 @@ impl<T: BeaconChainTypes> LightClientServerCache<T> {
         Ok(())
     }
 
+    pub fn store_light_client_update(
```
dapplion: You can inline this function above 
```rust
let should_persist_light_client_update = if let Some(prev_light_client_update) = prev_light_client_update {
    is_better_light_client_update(
        &prev_light_client_update,
        &new_light_client_update,
        chain_spec,
    )?
} else {
    true
} 
```

beacon_node/beacon_chain/src/light_client_server_cache.rs
```diff
@@ -140,6 +151,40 @@ impl<T: BeaconChainTypes> LightClientServerCache<T> {
                     signature_slot,
                     chain_spec,
                 )?);
+
+                let new_light_client_update = LightClientUpdate::new(
```
dapplion: I think you should not condition creating new `LightClientUpdate` on having a non-zero finalized_block_root. If the chain does not finalize for a full sync committee period we still need to produce them. Your `is_better_light_client_update` implementation covers the case of finalized vs non-finalized updates and will prioritize the finalized ones.

`LightClientUpdate` should be produced anytime we may be able to produce a better update

beacon_node/beacon_chain/src/light_client_server_cache.rs
```diff
@@ -249,3 +379,126 @@ fn is_latest_optimistic_update<E: EthSpec>(
         attested_slot == prev_slot && signature_slot > *prev.signature_slot()
     }
 }
+
+// Implements spec prioritization rules:
+// Full nodes SHOULD provide the best derivable LightClientUpdate for each sync committee period
+// ref: https://github.com/ethereum/consensus-specs/blob/113c58f9bf9c08867f6f5f633c4d98e0364d612a/specs/altair/light-client/full-node.md#create_light_client_update
+fn is_better_light_client_update<E: EthSpec>(
```
dapplion: Should we move this implementation to the `LightClientUpdate` type?

dapplion (commented):
Overall looks good! Two points:

We should run the specs tests for the is_better_update function, see here https://github.com/ethereum/consensus-specs/blob/dev/tests/formats/light_client/update_ranking.md

The function under which updates are computed only gets triggered for recent blocks. See
https://github.com/sigp/lighthouse/blob/ffe29c087da3ad7c9f5586083686bb9d2ba42c9c/beacon_node/beacon_chain/src/beacon_chain.rs#L3944-L3948

Now that the LC server computes recent updates AND full updates, the latter should include historical updates. Therefore we should be producing updates during forward range sync. Also for backfill sync, but that's a feature for another PR :)

consensus/types/src/light_client_update.rs
```diff
@@ -340,14 +344,37 @@ impl<E: EthSpec> LightClientUpdate<E> {
...
     slot: Slot,
     chain_spec: &ChainSpec,
 ) -> Result<Epoch, ArithError> {
-    slot.epoch(E::slots_per_epoch())
-        .safe_div(chain_spec.epochs_per_sync_committee_period)
+    println!("slot {}", slot);
+    let x = slot.epoch(E::slots_per_epoch())
+        .safe_div(chain_spec.epochs_per_sync_committee_period).unwrap();
+
+    println!("sync comm period {}", x);
```
dapplion: println

beacon_node/beacon_chain/src/light_client_server_cache.rs
eserilev: got it, we should now be persisting non finalized updates

eserilev:
This should be ready for another review! in the meantime i started on work to persist light client bootstraps based on the strategy detailed in this doc written by lion:  https://hackmd.io/@dapplion/BkxPbc6eQT

beacon_node/beacon_chain/src/light_client_server_cache.rs
```diff
@@ -119,23 +128,55 @@ impl<T: BeaconChainTypes> LightClientServerCache<T> {
             )?);
         };
 
+        let maybe_finalized_block_root =
```
dapplion: Should be named `maybe_finalized_block`

beacon_node/beacon_chain/src/light_client_server_cache.rs
```diff
@@ -119,23 +128,55 @@ impl<T: BeaconChainTypes> LightClientServerCache<T> {
...
+            if let Some(prev_light_client_update) = prev_light_client_update {
+                prev_light_client_update
+                    .is_better_light_client_update(&new_light_client_update, chain_spec)?
+            } else {
+                true
+            };
+
+        if should_persist_light_client_update {
+            self.store_light_client_update(&store, sync_period, &new_light_client_update)?;
+        }
```
dapplion: Purely organizational comment, but inserting light_client_update computation between the optimistic and finalized update seem to suggest there a dependency between light_client_update computation and finalized update computation. I don't think that's the case so can you move the light_client_update code to either before or after optimistic and finalized update code?

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -106,6 +107,264 @@ fn get_harness_generic(
...
+    // in the db.
+    let lc_updates = beacon_chain
+        .get_light_client_updates(sync_period, 100)
+        .unwrap();
+
+    assert_eq!(lc_updates.len(), 1);
+    assert_eq!(lc_updates.first().unwrap().clone(), lc_update);
+
+    // advance a bunch of slots & extend the chain
+    for _i in 0..1000 {
```
dapplion: Can you only advance the min necessary amount of slots to advance one sync committee? Compute that value from spec parameters

beacon_node/http_api/src/light_client.rs
```diff
@@ -0,0 +1,165 @@
...
+        ),
+        types::ForkName::Deneb => ChainSpec::compute_fork_digest(
+            chain.spec.deneb_fork_version,
+            chain.genesis_validators_root,
+        ),
+        types::ForkName::Electra => ChainSpec::compute_fork_digest(
+            chain.spec.electra_fork_version,
+            chain.genesis_validators_root,
+        ),
+    };
```
dapplion: This should be a method of ChainSpec if not already

beacon_node/http_api/tests/tests.rs
```diff
@@ -1709,6 +1709,36 @@ impl ApiTester {
...
+            .light_client_server_cache
+            .get_light_client_updates(
+                &self.chain.store,
+                current_sync_committee_period as u64,
+                1,
+                &self.chain.spec,
+            )
+            .unwrap();
+
+        assert_eq!(result.clone().unwrap().len(), expected.len());
```
dapplion, line 1842: What's the expected len? Could this be asserting `assert_eq!(0, 0)`?

beacon_node/store/src/lib.rs
```diff
@@ -270,6 +270,9 @@ pub enum DBColumn {
     BeaconHistoricalSummaries,
     #[strum(serialize = "olc")]
     OverflowLRUCache,
+    /// For persisting eagerly computed light client data
+    #[strum(serialize = "lcu")]
+    LightClientUpdate,
```
dapplion, line 305: @michaelsproul This bucket acts like finalized data, but it will be re-written multiple times as we find better updates through a light-client period. Is that okay? As a simplification we choose to not guarantee that stored lightclient updates are canonical. I have argued extensively that this is not necessary today: canonical-ness becomes relevant when implementing some backfilling strategies.

consensus/types/src/light_client_update.rs
```diff
@@ -259,6 +248,150 @@ impl<E: EthSpec> LightClientUpdate<E> {
 
         Ok(update)
     }
+
+    // Implements spec prioritization rules:
+    // Full nodes SHOULD provide the best derivable LightClientUpdate for each sync committee period
+    // ref: https://github.com/ethereum/consensus-specs/blob/113c58f9bf9c08867f6f5f633c4d98e0364d612a/specs/altair/light-client/full-node.md#create_light_client_update
+    pub fn is_better_light_client_update(
+        &self,
+        new: &LightClientUpdate<E>,
```
dapplion:
```suggestion
        new: &Self<E>,
```

consensus/types/src/light_client_update.rs
```diff
@@ -259,6 +248,150 @@ impl<E: EthSpec> LightClientUpdate<E> {
...
+            LightClientUpdate::Altair(prev) => prev.attested_header.beacon.slot,
+            LightClientUpdate::Capella(prev) => prev.attested_header.beacon.slot,
+            LightClientUpdate::Deneb(prev) => prev.attested_header.beacon.slot,
+        };
+
+        let prev_finalized_header_slot = match self {
+            LightClientUpdate::Altair(prev) => prev.finalized_header.beacon.slot,
+            LightClientUpdate::Capella(prev) => prev.finalized_header.beacon.slot,
+            LightClientUpdate::Deneb(prev) => prev.finalized_header.beacon.slot,
+        };
```
dapplion: These should be methods of `LightClientUpdate::attested_header_slot`, `LightClientUpdate::finalized_header_slot`

consensus/types/src/light_client_update.rs
```diff
@@ -259,6 +248,150 @@ impl<E: EthSpec> LightClientUpdate<E> {
...
+        let new_active_participants = new.sync_aggregate().sync_committee_bits.num_set_bits();
+        let prev_active_participants = self.sync_aggregate().sync_committee_bits.num_set_bits();
+
+        let new_has_super_majority =
+            new_active_participants.safe_mul(3)? >= max_active_participants.safe_mul(2)?;
+        let prev_has_super_majority =
+            prev_active_participants.safe_mul(3)? >= max_active_participants.safe_mul(2)?;
+
+        if new_has_super_majority != prev_has_super_majority {
+            return Ok(new_has_super_majority & !prev_has_super_majority);
```
dapplion: Are you using Bitwise And intentionally here? Took me a bit to verify this statement. The spec uses `new > old`, which is actually equivalent to your statement. However since we only consider `new != old` the statement can be simplified just to `new`.

| new | old | new > old | new & !old | !old |
| --- | --- | --------- | ---------- | ---- |
| 1   | 0   | 1         | 1          | 1    |
| 0   | 1   | 0         | 0          | 0    |

```suggestion
            return Ok(new_has_super_majority);
```

consensus/types/src/light_client_update.rs
```diff
@@ -259,6 +248,150 @@ impl<E: EthSpec> LightClientUpdate<E> {
...
+            compute_sync_committee_period_at_slot::<E>(*new.signature_slot(), chain_spec)?;
+
+        let prev_attested_header_sync_committee_period =
+            compute_sync_committee_period_at_slot::<E>(prev_attested_header_slot, chain_spec)?;
+
+        let prev_signature_slot_sync_committee_period =
+            compute_sync_committee_period_at_slot::<E>(*self.signature_slot(), chain_spec)?;
+
+        // Compare presence of relevant sync committee
+        let new_has_relevant_sync_committee = !new.is_next_sync_committee_branch_empty()
```
dapplion: You should add a `is_sync_committee_update` method to reflect the spec, will make it easier to review changes

consensus/types/src/light_client_update.rs
```diff
@@ -259,6 +248,150 @@ impl<E: EthSpec> LightClientUpdate<E> {
...
+            && (new_attested_header_sync_committee_period
+                == new_signature_slot_sync_committee_period);
```
dapplion: Can also condense this logic into a method `attested_header_in_signature_period`

consensus/types/src/light_client_update.rs
```diff
@@ -259,6 +248,150 @@ impl<E: EthSpec> LightClientUpdate<E> {
...
+        // Compare indication of any finality
+        let new_has_finality = !new.is_finality_branch_empty();
+        let prev_has_finality = !self.is_finality_branch_empty();
+        if new_has_finality != prev_has_finality {
+            return Ok(new_has_finality);
+        }
+
+        // Compare sync committee finality
+        if new_has_finality {
+            let new_has_sync_committee_finality =
```
dapplion: Could also add `has_sync_committee_finality` method

dapplion (commented):
Almost there! Just some minor comments

One note about this strategy to persist updates:

This PR for simplicity assumes the current light client update is canonical, and it is stored in a single bucket keyed by sync period. Therefore forks are not supported. It is possible for this PR to persist a light client update that contains a header that ends up not being canonical. This is okay for now as light client updates main purpose is to discover the latest set of sync committee pubkeys. Therefore we can rephrase the criteria to:

1. Persisted light client updates SHOULD include canonical headers
2. Persisted light client updates MUST include canonical next sync committee pubkeys

We satisfy the first statement as we may include non-canonical headers. However, we can't ensure the second statement.  For us to store a non-canonical next sync committee the following must happen:
- Have a re-org through a sync committee boundary such that we persist at least 1 light client update with an orphaned sync committee
- Not have any better light client update during a full period. This means: no finality during 27 hours, and no better participation in any slot.

If we lose finality during 27 hours, we will have bigger problems than the possibility of a bad light client update. So IMO this design is good for the vast majority of cases and much better than what we have now (nothing).

beacon_node/beacon_chain/src/light_client_server_cache.rs
```diff
@@ -119,23 +128,55 @@ impl<T: BeaconChainTypes> LightClientServerCache<T> {
...
+            maybe_finalized_block_root.as_ref(),
+            chain_spec,
+        )?;
+
+        let sync_period = block_slot
+            .epoch(T::EthSpec::slots_per_epoch())
+            .sync_committee_period(chain_spec)?;
+
+        let prev_light_client_update =
+            self.get_light_client_update(&store, sync_period, chain_spec)?;
```
dapplion: As an optimization, you can cache a single item from the store to prevent reading the same update over and over.

beacon_node/beacon_chain/src/light_client_server_cache.rs
```diff
@@ -152,6 +193,84 @@ impl<T: BeaconChainTypes> LightClientServerCache<T> {
...
 
+    pub fn store_light_client_update(
+        &self,
+        store: &BeaconStore<T>,
+        sync_committee_period: u64,
+        light_client_update: &LightClientUpdate<T::EthSpec>,
+    ) -> Result<(), BeaconChainError> {
+        let column = DBColumn::LightClientUpdate;
+
+        store.hot_db.put_bytes(
```
dapplion, line 218: Should you store to hot_db or cold_db? @michaelsproul

consensus/types/src/light_client_update.rs
eserilev: sure i'll simplify this. the bitwise and was actually forced on me by clippy!

beacon_node/http_api/tests/tests.rs
eserilev: yeah the goal was to add a check to ensure that the plumbing for the `get_beacon_light_client_updates` endpoint is correct. but yeah were essentially asserting 0 eq 0

dapplion (commented):
Thank you @eserilev ! This is good to go, after choosing if either to store updates in hot or cold. I favor cold, as we treat it like finalized data (index by time not root) and in practice won't change much. Moving to cold allows us to add a hot bucket latter (indexed by root) if we want to support canonical only data.

is_better_light_client_update reads really nice now

dapplion:
Chatted with @michaelsproul offline, and we agree to put this data on the hot db, so this PR is good to go

dapplion:
Same CI failures
```
FAIL [   4.186s] beacon_chain::beacon_chain_tests store_tests::light_client_updates_test
```
```
FAIL [   1.187s] ef_tests::tests light_client_update
```
And the basic simulator

eserilev:
Was able to fix the CI and also added a TODO

Also included changes from this PR
https://github.com/sigp/lighthouse/pull/6201

eserilev:
Since the fork epoch boundaries in `basic-sim` aren't multiples of 256, the light client periods dont align with the fork boundaries. We end up getting some `BeaconStateError::IncorrectStateVariant` when constructing light client updates

I've gotten the `basic-sim` tests to pass by moving the light client update logic to the bottom of `recompute_and_cache_updates`. This prevents the light client update stuff from affecting optimistic and finality update

I've also completely removed any checks related to light client updates in `basic-sim`. I've kept the test vectors related to optimistic and finality updates. I think we have enough light client update test coverage in other places to live with this for now...

dapplion (approved):
Pushed a commit with:
- Clean up two info statements likely there from debugging
- Remove pub where possible
- A few cosmetic nits

PR is good to go and passing tests :)

jimmygchen (approved):
LGTM too!

jimmygchen:
@mergify queue

PR #5554, opened 2024-04-10
Allow 1 count block request to return 0 blocks

dapplion:
## Issue Addressed

Noted the following bug in ReqResp logic:
- We issue a ReqResp request for blocks where the count of requested elements is 1
- The peer does not have the requested element and returns 0 elements
- We consider that a protocol violation with `RPCError::IncompleteStream`

Not having a block or blob is tolerated as per the req resp spec. It's the downstream caller that should decide if the peer should have or not that specific element.

## Proposed Changes

Rename `expected_responses` to `max_responses`, as for multiple responses that what the number really represents. The peer is allow to return less than the request count of elements, including 0.

Add a new function `expect_exactly_one_response` to differentiate those request where closing the stream before the first element is actually a protocol violation (e.g. status method)

I have reversed some of the if else cases, (from `if !condition A else B` to `if condition B else A`, but the actual logic is unchanged.

## Additional Info

### DIFF
--- a/beacon_node/lighthouse_network/src/rpc/handler.rs
+++ b/beacon_node/lighthouse_network/src/rpc/handler.rs
@@ -163,7 +163,7 @@ struct InboundInfo<E: EthSpec> {
     /// Protocol of the original request we received from the peer.
     protocol: Protocol,
     /// Responses that the peer is still expecting from us.
-    remaining_chunks: u64,
+    max_remaining_chunks: u64,
     /// Useful to timing how long each request took to process. Currently only used by
     /// BlocksByRange.
     request_start_time: Instant,
@@ -471,7 +471,7 @@ where
                         // Process one more message if one exists.
                         if let Some(message) = info.pending_items.pop_front() {
                             // If this is the last chunk, terminate the stream.
-                            let last_chunk = info.remaining_chunks <= 1;
+                            let last_chunk = info.max_remaining_chunks <= 1;
                             let fut =
                                 send_message_to_inbound_substream(substream, message, last_chunk)
                                     .boxed();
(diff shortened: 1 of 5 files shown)


beacon_node/lighthouse_network/src/rpc/outbound.rs
```diff
@@ -92,7 +92,7 @@ impl<E: EthSpec> OutboundRequest<E> {
     /* These functions are used in the handler for stream management */
 
     /// Number of responses expected for this request.
```
pawanjay176:
```suggestion
    /// Maximum number of responses expected for this request.
```

beacon_node/lighthouse_network/src/rpc/rate_limiter.rs
```diff
@@ -247,7 +247,7 @@ impl<E: EthSpec> RateLimiterItem for super::OutboundRequest<E> {
     }
 
     fn expected_responses(&self) -> u64 {
```
pawanjay176: Can change this to max_responses too maybe for consistency?

beacon_node/lighthouse_network/src/rpc/handler.rs
```diff
@@ -725,7 +725,16 @@ where
                         self.outbound_substreams_delay.remove(delay_key);
                         entry.remove_entry();
                         // notify the application error
-                        if request.expected_responses() > 1 {
+                        if request.expect_exactly_one_response() {
+                            // else we return an error, stream should not have closed early.
```
pawanjay176:
```suggestion
                            // we return an error, stream should not have closed early.
```

pawanjay176 (approved):
Nice catch. LGTM

realbigsean:
@mergifyio queue

PR #5555, opened 2024-04-10
Only load Kzg in tests if necessary and only load it once

jimmygchen:
## Issue Addressed

Loading trusted setup file and instantiating the `Kzg` instance is very expensive (>80% of test time for most tests), and we're doing this quite a lot of times in tests even when we don't need them (e.g. pre-deneb tests), and in some cases they're loaded twice when `mock_execution_layer` is used.

This PR improves this by loading `Kzg` trusted setup only once at most, if necessary. 

Some numbers when I run beacon chain tests locally:
- `unstable` branch 180s (~17m on CI)
- `unstable` + fix (current branch): 173s
- `das` branch: 990s (>55m on CI)
- `das` branch with this fix: 269s

The difference isn't significant on `unstable`, so it looks like `Kzg` initialisation time has increased quite significantly due the changes in the `das` version - there's probably more optimisation that can be done on `c-kzg` but we should probably avoid loading it multiple times in a single test anyway, as we run each of these beacon chain tests once per fork.

Test command used: 
```bash
env FORK_NAME=deneb cargo nextest run --release --features "fork_from_env,slasher/lmdb,portable" -p beacon_chain
```

### DIFF
--- a/beacon_node/beacon_chain/src/builder.rs
+++ b/beacon_node/beacon_chain/src/builder.rs
@@ -25,7 +25,7 @@ use eth1::Config as Eth1Config;
 use execution_layer::ExecutionLayer;
 use fork_choice::{ForkChoice, ResetPayloadStatuses};
 use futures::channel::mpsc::Sender;
-use kzg::{Kzg, TrustedSetup};
+use kzg::Kzg;
 use operation_pool::{OperationPool, PersistedOperationPool};
 use parking_lot::{Mutex, RwLock};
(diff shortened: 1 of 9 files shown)


jimmygchen:
This only works in the `beacon-chain-tests` crate as the client crate needs to be updated to but I'm not sure if this is a good idea. Will investigate a bit more.

pawanjay176 (approved):
LGTM. Thanks for running the numbers. I'm hoping the `das` branch in ckzg still has room for optimisations

realbigsean:
@mergifyio queue

PR #5556, opened 2024-04-10
Ensure proper ReqResp response stream termination

dapplion:
## Issue Addressed

Noted that while serving ResResp BlobsByRoot for some errors we properly terminate the stream with an error, while other don't. I believe this is a typo, and caused by the fact that every return statement from the function must include an outbound event. This requirement is prone to typos, and may not be apparent to a contributor.

## Proposed Changes

Wrap ReqResp handling functions with a `terminate_response_*` function that forces every handler call to terminate properly:
- If multi-response: with either a stream termination or an error
- If single response: with either a single item or an error

My hope is that future contributors will just extend this PR's pattern and not let streams hanging.

## Additional Info

I can't cover the `blocks_by*` methods are their termination is handled on an spawned function. @michaelsproul any way to get the return value from some variant of `executor.spawn` to be able to change the signature of `handle_blocks_by_range_request` to `Result<StreamTerminator, RPCError>`?

### DIFF
--- a/beacon_node/network/src/network_beacon_processor/rpc_methods.rs
+++ b/beacon_node/network/src/network_beacon_processor/rpc_methods.rs
@@ -224,7 +224,7 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
             peer_id,
             request_id,
             self.handle_blobs_by_root_request_inner(peer_id, request_id, request),
-            Response::BlobsByRoot
+            Response::BlobsByRoot,
         );
     }
 
--- a/beacon_node/network/src/network_beacon_processor/rpc_methods.rs
+++ b/beacon_node/network/src/network_beacon_processor/rpc_methods.rs
@@ -224,7 +224,7 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
             peer_id,
             request_id,
             self.handle_blobs_by_root_request_inner(peer_id, request_id, request),
-            Response::BlobsByRange,
+            Response::BlobsByRoot
         );
     }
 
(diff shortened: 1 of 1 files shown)


beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -212,17 +212,33 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
         self: Arc<Self>,
         peer_id: PeerId,
         request_id: PeerRequestId,
         request: BlobsByRootRequest,
     ) {
+        self.terminate_response_stream(
+            peer_id,
+            request_id,
+            self.handle_blobs_by_root_request_inner(peer_id, request_id, request),
+            Response::BlobsByRange,
```
realbigsean:
```suggestion
            Response::BlobsByRoot
```

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -362,25 +367,23 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
             peer_id,
-            Response::LightClientFinalityUpdate(Arc::new(light_client_finality_update)),
             request_id,
-        )
+            match self
+                .chain
+                .light_client_server_cache
+                .get_latest_finality_update()
+            {
+                // TODO(lc): Arc cached update inside the light_client_server_cache
```
realbigsean:
```suggestion
```

Moved this to an issue https://github.com/sigp/lighthouse/issues/5578

realbigsean (changes requested):
Big fan of the change

realbigsean (approved):

realbigsean:
@mergifyio queue

Issue #5549, opened 2024-04-10
Block lookup sync technical debt


dapplion:
Tracker issue for multiple technical debt cleaning ideas or simplifications that came up while working on DAS.

### PR queue by intended merge order

- https://github.com/sigp/lighthouse/pull/5562
- https://github.com/sigp/lighthouse/pull/5561
- https://github.com/sigp/lighthouse/pull/5563
- https://github.com/sigp/lighthouse/pull/5564
- https://github.com/sigp/lighthouse/pull/5583

### Handle streams in `SyncNetworkContext`

Today, each sync component (Range, Backfill, Lookup) handles stream progression and termination. None of the components do anything until the stream is completed. Therefore they don't need to be aware of streams at all. Their current API to inject stream events is
```rust
impl SomeSync {
  /// None = stream termination
  fn block_response(block: Option<Block>) {}
  fn block_error(error: RPCError) {}
}
```
Instead their API could simply into
```rust
impl SomeSync {
  fn blocks_response(blocks: Result<Vec<Block>, RPCError>) {}
}
```
Where `SyncNetworkContext` tracks and accumulates stream items the same way we do today with range sync blocks and blobs.

- **status**: WIP branch: https://github.com/dapplion/lighthouse/tree/handle-sync-requests
- Best to wait to merge first: https://github.com/sigp/lighthouse/pull/5561 https://github.com/sigp/lighthouse/pull/5562

### Merge `BackFillBlocks`, `BackFillBlockAndBlobs`, `RangeBlocks`, `RangeBlockAndBlobs` handling

All those requests return the same type of data, a `Vec<RpcBlock>`, which can be handled with a proper request id.

- **status**: PR up https://github.com/sigp/lighthouse/pull/5561

### Merge `SingleBlock` + `ParentLookup` request Ids

There's no need for the network router to be aware of the distinction between parent and current block lookups. Nesting the lookup type into `SingleLookupReqId` de-clutters the router and sync manager code.

- **status**: PR up https://github.com/sigp/lighthouse/pull/5562

### Strict state transitions in single lookup state `State`

All properties of `SingleLookupRequestState` are public and some are mutated in disperse parts of the codebase, making it hard to follow state changes. We should follow more strict syntax like we do in range sync batches, where the state is checked before mutating it.

- **status**: 
  - Partially addressed with https://github.com/sigp/lighthouse/pull/5563
[...]

PR #5562, opened 2024-04-11
Nest lookup type into request id SingleBlock and SingleBlob

dapplion:
## Issue Addressed

Part of:

It's unnecessary to have distinct RequestIds for parent and current block lookups. The router does not need to be concerned with this distinction, only the block lookups struct needs to know where to route the response.

## Proposed Changes

Nest lookup type into request id SingleBlock and SingleBlob.

This PR does not change any sync logic. It fixes a typo in the test `no_peer_penalty_when_rpc_response_already_known_from_gossip` which was fetching the wrong ID


realbigsean (approved):
Nice simplification ✅

realbigsean:
@mergifyio queue

PR #5563, opened 2024-04-11
Make SingleLookupRequestState fields private

dapplion:
## Issue Addressed

`SingleLookupRequestState` fields are public and mutated all over the place. This makes the code unnecessarily difficult to follow. Instead, its fields should be private and mutated through setters such that looking the the implementation of `SingleLookupRequestState` is sufficient to view and understand all possible state transitions.

Part of:

## Proposed Changes

- Make SingleLookupRequestState fields private
- Move state transition to download result from `verify_response_inner` (specific per request) to `verify_response` (generic for the trait)
- In `verify_response` use the actual response sending peer instead of the one being tracked in the state. See code comment about potential inconsistencies.

A future PR should add more checks in setters like `on_download_success` such that a state transition from `Processed` -> `Downloading` is not possible

### DIFF
--- a/beacon_node/network/src/sync/block_lookups/common.rs
+++ b/beacon_node/network/src/sync/block_lookups/common.rs
@@ -12,7 +12,6 @@ use beacon_chain::data_availability_checker::ChildComponents;
 use beacon_chain::{get_block_root, BeaconChainTypes};
 use lighthouse_network::rpc::methods::BlobsByRootRequest;
 use lighthouse_network::rpc::BlocksByRootRequest;
-use rand::prelude::IteratorRandom;
 use std::ops::IndexMut;
 use std::sync::Arc;
 use std::time::Duration;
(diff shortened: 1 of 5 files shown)


realbigsean (approved):

realbigsean:
definitely support consolidating where these state changes are made. Not tracking this led to one of our recent bugs in duplicate parent lookups

> A future PR should add more checks in setters like on_download_success such that a state transition from Processed -> Downloading is not possible

Agreed

realbigsean:
@dapplion mentioned offline the TODOs here will be fixed in https://github.com/sigp/lighthouse/pull/5583

realbigsean:
@mergifyio queue

dapplion:
@realbigsean about this todo

https://github.com/sigp/lighthouse/blob/6cc98a7a4a0ab052f1853c5108031fe1135bcf93/beacon_node/network/src/sync/block_lookups/common.rs#L169-L171

not sure if this was just overlooked in the original design. The peer returned from the network is decided by internal logic. If there are not bugs in routing the correct request_id the peer_id should always match. Then, should we even track the downloading peer?

PR #5569, opened 2024-04-11
Re-process early sampling requests

dapplion:
## Issue Addressed

Allow data availability samplers to send requests early as a network-wide optimization to complete sampling faster. A custodial node will queue the requests and resolve them once the block has been imported.

## Next steps

This is a basic (non-tested) implementation, next steps:
- Resolve requests when column arrives, not the whole block
- Separate requests for known block and unknown block, the former can be DOS protected better
- Only allow retry mechanism for recent slots
- Add more metrics

## Design

photo_2024-04-12 11 08 08

two phases
a) requests at the start of the start of the slot have a long timeout — e.g. 3s (expected time for most of the network to get samples on gossip)
b) then shift into fast timeout mode

At start of the slot, send your sample requests to peers you expect to get your needed samples in (a) mode (long timeout). If they don't respond by timeout, send further queries in (b) mode to various peers, trying to quickly get the final samples

> Forwarded from codex (credit to them) 👇 

The usage of (a) allows for samples queries to be responded to immediately when a peer gets it from gossip, instead of the node waiting until some arbitrary time at the slot that they think people will likely have gotten the sample to start the requests

The messaging primitive I proposed is a request with an explicit timeout. This simple primitive allows us to build a nice dynamic behavior, as you see in the example on the slide. It allows us to have the first period, beneficial for two reasons:
 - reduce traffic explosion on the happy path: if things go well, every sample query will only consume two messages. A request and a response. This is beneficial both in the peerDAS case, and in the later stages, where we would go for individual segments on the 2D encoding grid. 
 - reduce sampling delay close to the minimum: sending the sampling query at the start of the slot allows us to minimize the sampling delay to a single 1-hop latency on top of to diffusion delay (the delay with which the node we queried received the segment or column).

The minimum sampling latency would be what you could achive if you ask, right when you know what samples you need, all your peers that could have the sample. That would however mean that you receive lots of responses, adding considerable extra load. Here we only ask one of them as a compromise between bandwidth efficiency and latency, and our simulations currently show the latency compromise is very limited on the happy path. Obviusly, if we are not on the happy path, we should start doing something. On the slide you find a few proposals, considering that we should still avoid traffic explosion.

There a few other points there, improving dynamic behaviour:
1, anticipate the hunt for peers that could provide samples. We have a few options here:
  - a node could try to keep a few peers (but at least one) for each column. This can be done independent of slots, continously
  - a node can notice at the beginning of the slot, that it misses peers for a column selected for sampling. The node has almost 4 seconds to find a peer and send a query
  - a node can also act if a request was sent out, but a response is not coming close to the 4s mark. If there are other peers for the same sample, send another request. If not, start searching for a peer to request from.

2, to avoid traffic explosion, we also recommend controlling the number of outstanding queries, together with timeouts. This practically means that we can avoid traffic explosion. 
3, we've also proposed the extension of the sample. This is something I'm writing up in an ethresear.ch post 🙂

Regarding the number of queries is stage (a): I think 1 per sample is fine, but of course 2 or 3 can also be used. These are comprimises that can even be made locally, we can't really mandate them anyway. (actually we could, using RLN and similar spam protection, but that would require stake to send DAS queries, which is not what we want now).

Finally, there was another idea on my previous slide on messaging primitives. I would include a have list in the query response message. The reason behind this is that it gives extra information on the state of diffusion that we would not have otherwise. Given the 1D or 2D structure, a have list can be compressed very well to a small bitmap, so it is not much extra bytes, so I think it's worth to add. Eventually, to stay generic on the message primive, we can make it a flag in the request whether you want a have list in the response.

### DIFF
--- a/beacon_node/beacon_processor/src/work_reprocessing_queue.rs
+++ b/beacon_node/beacon_processor/src/work_reprocessing_queue.rs
@@ -685,7 +685,7 @@ impl<S: SlotClock> ReprocessQueue<S> {
                 // Register the delay.
                 let delay_key = self
                     .sampling_requests_delay_queue
-                    .insert(id, QUEUED_LIGHT_CLIENT_UPDATE_DELAY);
+                    .insert(id, QUEUED_SAMPLING_REQUESTS_DELAY);
 
                 self.awaiting_sampling_requests_per_block_root
                     .entry(queued_sampling_request.beacon_block_root)
--- a/beacon_node/beacon_processor/src/work_reprocessing_queue.rs
+++ b/beacon_node/beacon_processor/src/work_reprocessing_queue.rs
@@ -51,6 +51,9 @@ pub const QUEUED_LIGHT_CLIENT_UPDATE_DELAY: Duration = Duration::from_secs(12);
 /// For how long to queue rpc blocks before sending them back for reprocessing.
 pub const QUEUED_RPC_BLOCK_DELAY: Duration = Duration::from_secs(4);
 
+/// For how long to queue sampling requests for reprocessing.
+pub const QUEUED_SAMPLING_REQUESTS_DELAY: Duration = Duration::from_secs(12);
+
 /// Set an arbitrary upper-bound on the number of queued blocks to avoid DoS attacks. The fact that
 /// we signature-verify blocks before putting them in the queue *should* protect against this, but
 /// it's nice to have extra protection.
--- a/beacon_node/beacon_processor/src/work_reprocessing_queue.rs
+++ b/beacon_node/beacon_processor/src/work_reprocessing_queue.rs
@@ -62,6 +62,10 @@ const MAXIMUM_QUEUED_ATTESTATIONS: usize = 16_384;
 /// How many light client updates we keep before new ones get dropped.
 const MAXIMUM_QUEUED_LIGHT_CLIENT_UPDATES: usize = 128;
 
+/// How many sampling requests we queue before new ones get dropped.
+/// TODO(das): choose a sensible value
+const MAXIMUM_QUEUED_SAMPLING_REQUESTS: usize = 16_384;
+
 // Process backfill batch 50%, 60%, 80% through each slot.
 //
 // Note: use caution to set these fractions in a way that won't cause panic-y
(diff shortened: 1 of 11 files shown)


beacon_node/beacon_processor/src/work_reprocessing_queue.rs
```diff
@@ -630,6 +657,35 @@ impl<S: SlotClock> ReprocessQueue<S> {
...
+                    return;
+                }
+
+                let id: QueuedSamplingRequestId = self.next_sampling_request_update;
+                self.next_sampling_request_update += 1;
+
+                // Register the delay.
+                let delay_key = self
+                    .sampling_requests_delay_queue
+                    .insert(id, QUEUED_LIGHT_CLIENT_UPDATE_DELAY);
```
jimmygchen:
```suggestion
                    .insert(id, QUEUED_SAMPLING_REQUESTS_DELAY);
```

beacon_node/beacon_processor/src/work_reprocessing_queue.rs
```diff
@@ -232,22 +250,29 @@ struct ReprocessQueue<S> {
     queued_lc_updates: FnvHashMap<usize, (QueuedLightClientUpdate, DelayKey)>,
     /// Light Client Updates per parent_root.
     awaiting_lc_updates_per_parent_root: HashMap<Hash256, Vec<QueuedLightClientUpdateId>>,
+    /// Queued sampling requests per block root.
```
jimmygchen:
```suggestion
    /// Queued sampling requests.
```

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -573,6 +573,11 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
         })
     }
 
+    pub fn has_block(&self, block_root: &Hash256) -> Result<bool, Error> {
+        self.hot_db
+            .key_exists(DBColumn::BeaconBlock.into(), block_root.as_bytes())
+    }
```
jimmygchen: I think this is the same as [block_exists](https://github.com/sigp/lighthouse/blob/27d468474cdeacd2a8f58b7a8d65c1b5381412ef/beacon_node/store/src/hot_cold_store.rs#L642-L646)

jimmygchen (changes requested):
Looks good! 

@pawanjay176 mentioned that we would also need to send first byte so that the connection doesn't timeout (TTFB = 5s). I can follow this up in a separate PR.

beacon_node/store/src/hot_cold_store.rs
jimmygchen: Fixed in https://github.com/sigp/lighthouse/pull/5569/commits/4adbd29f038e9311dc57aa92de238f5e54dd2723

jimmygchen (approved):
@dapplion I've made some minor fixes and I think this PR is good to go! I'll address the TTFB issue in a separate PR. Thanks!

jimmygchen:
We probably don't need to worry too much about TTFB timeout for now - as we send sampling request after we receive the block (instead of start of slot), so in most cases 5s would be good enough. Will leave this for later if we need it.

Issue #4983, opened 2023-12-06
PeerDAS - Tracking Issue


AgeManning:
## Description

This is a tracking issue for the work that is being undertaken on building out Data Availability Sampling.

This will be updated and edited as we go.

Instructions for running a DAS devnet can be found in [this gist](https://gist.github.com/jimmygchen/54574443ea8a94118f0f748b69a5e610).

Main PR to `unstable`: 
- #5683 

## Devnet 5

- [x] #6251
- [ ] #6258
- [x] #5990 
- [x] #6268 
- [x] #5894
- [x] #6111
- [x] #6018 
- [x] #6107 
- [ ] Implement validator custody ([spec](https://github.com/ethereum/consensus-specs/pull/3779)) https://github.com/sigp/lighthouse/issues/6767
- [x] https://github.com/sigp/lighthouse/issues/6854
- [ ] Address all `TODO(das)` comments
- [ ] Batch verify KZG proofs where possible https://github.com/sigp/lighthouse/pull/6851
- [x] Don't lookup sync individual columns received from gossip. WIP  [`dapplion:dedup-columns-by-root-requests`](https://github.com/sigp/lighthouse/compare/unstable...dapplion:lighthouse:dedup-columns-by-root-requests?expand=1)

## `TODO(das)`

- Collection of small / easy TODOs https://github.com/sigp/lighthouse/pull/6862
- https://github.com/sigp/lighthouse/pull/6850

### Data format and distribution (#5050)

- [x] Add `ColumnSidecar` type 
- [x] Extend blobs horizontally with dummy data (`0x{random_uniq_bits}...{col_id}`) and distribute to column subnets
- [x] Create 32 column subnets `NUM_COLUMN_SUBNETS`
- [x] Compute subnets to subscribe to based on `node_id` (see att example [here](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#attestation-subnet-subscription))
- [x] Node to subscribe to only `CUSTODY_REQUIREMENT` column subnets

### Data Column Custody

- [x] #5196
- [x] Advertise extra custodied columns through ENR (@jacobkaufmann)
- [x] Implement `PEERDAS_ACTIVATION_EPOCH` to transition from deneb blobs to data columns (@jimmygchen)
- [x] #5570
- [x] #5684
[...]

PR #5570, opened 2024-04-12
Send custody data column to `DataAvailabilityChecker` for determining block importability

jimmygchen:
## Issue Addressed

Send custody data column to `DataAvailabilityChecker` for determining block importability.

## Proposed Changes

- [x] Add `--subscribe-all-data-column-subnets` flag, and provide `custody_column_count` to `da_checker.availability_cache`.
- [x] Make PeerDAS activate on a soft fork `PEER_DAS_EPOCH` and start separating DAS from deneb logic.
- [x] Update `PendingComponent.is_available` to check for number of expected columns.
- [x] When publishing a block, send only the custody data columns to the `DataAvailabilityChecker`.
- [x] ~~Database migration for `OverflowLRUCache` if necessary.~~ (Moved to tracking issue)

### Custody Columns from gossip
- After receiving a gossip verfieid data column, we ~~wrap it in a `CustodyDataColumn` object and~~ send it to `DAChecker`. (nodes only subscribe to subnets they custody) 
- `DAChecker` is aware of how many columns it's custodying, so once it collects enough distinct `CustodyDataColumn`s then custody requirement is fullfiled.

### Custody Columns from RPC (out of scope for this PR as RPC not yet implemented)
- If we receive an attestation with unknown block, this triggers a block and data column lookup for the root from sync (sync can compute the custody columns)
- When response for this request type (`RequestId::SingleDataColumn`) is received, sync wraps the data in a `CustodyDataColumn` and send it to DAChecker
- `DAChecker` is aware of how many columns it's custodying, so once it collects enough distinct `CustodyDataColumn`s then custody requirement is fullfiled.
- For sampling responses, I guess we would have a different `RequestId` to differentiate, e.g. RequestId::Sampling?

### Custody Columns from HTTP API
- `http_api` has access to `network_globals` and can calculate which columns need to custody, so we can wrap them the same way and send them to the beacon chain.


jimmygchen:
It looks like updating the `is_available`/`is_importable` function has opened up a can of worms and broken a bunch of tests, because the current branch allows blobs & columns to coexist, which would require updating a bunch of code that we eventually don't need. I'll open a new PR to implement activating PeerDAS on a specific `PEER_DAS_EPOCH`. (thanks to @dapplion for the suggestion)

jimmygchen:
Tested this on a local devnet, fixed a few bugs and I think this is now working! 

I was able to see the regular node persisting its custody columns during import, and supernodes (`--subscribe-all-data-column-subnet` persisting all custody columns during import.

beacon_node/beacon_chain/src/builder.rs
```diff
@@ -104,6 +104,7 @@ pub struct BeaconChainBuilder<T: BeaconChainTypes> {
     kzg: Option<Arc<Kzg>>,
     task_executor: Option<TaskExecutor>,
     validator_monitor_config: Option<ValidatorMonitorConfig>,
+    import_all_data_columns: bool,
```
dapplion, line 107: Should this be achieved with a high `custody_requirement` parameter?

EDIT: I think for now this bool flag is easier to work with 👍

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -230,15 +230,48 @@ impl<E: EthSpec> PendingComponents<E> {
...
+            }
+            BlockImportRequirement::CustodyColumns(num_expected_columns) => {
+                let num_received_data_columns = self.num_received_data_columns();
+
+                debug!(
+                    log,
+                    "Checking block and data column importability";
+                    "block_root" => %self.block_root,
+                    "num_received_data_columns" => num_received_data_columns,
+                );
```
dapplion, line 266: This debug here appears like it could become very noisy. It's not a response to a state-changing event, but just a getter

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -350,7 +381,7 @@ impl<E: EthSpec> PendingComponents<E> {
 enum OverflowKey {
     Block(Hash256),
     Blob(Hash256, u8),
-    DataColumn(Hash256, u8),
+    DataColumn(Hash256),
```
dapplion: Why not necessary to key by column_index?

beacon_node/http_api/src/publish_blocks.rs
```diff
@@ -238,6 +240,35 @@ pub async fn publish_block<T: BeaconChainTypes, B: IntoGossipVerifiedBlockConten
...
+                            log,
+                            "Invalid blob provided to HTTP API";
+                            "reason" => &msg
+                        );
+                        Err(warp_utils::reject::custom_bad_request(msg))
+                    };
+                }
+            }
+        }
+    }
```
dapplion, line 270: Should you import all the columns when producing a block? I see that there's no mechanism for others to retrieve data you created but don't declare to custody. However it frightens me a little that network liveness depends on those very few peers of the block proposer that are meant to custody specific rows. If a proposer happens to not be peered at the time of publication with a peer on each subnet there will be data loss.

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -752,23 +779,34 @@ impl<T: BeaconChainTypes> OverflowLRUCache<T> {
         }
     }
 
+    fn block_import_requirement(
+        &self,
+        pending_components: &PendingComponents<T::EthSpec>,
+    ) -> Result<BlockImportRequirement, AvailabilityCheckError> {
+        let epoch = pending_components
+            .epoch()
+            .ok_or(AvailabilityCheckError::UnableToDetermineImportRequirement)?;
```
dapplion, line 798: Is this error necessary? The import requirements are not complete until the block shows up, so you can assume that it will always eventually show up. Instead of returning an error here, you can return something variant signaling that requirements are not known yet.

beacon_node/beacon_chain/src/builder.rs
jimmygchen: I thought about this too, and I decided to go with a bool as for simplicity now. I'm not sure if there's any motivation to set a number of custody subnets between `min_custody_requirement` and all subnets. 

For 1D PeerDAS, users could set this to half of the column count and be able to reconstruct without having to query more columns from the network - I'm not sure if this is a practical use case though.

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: Yep agree that it could be become noisy, and I thought about downgrading it too - it was quite useful when I was debugging issues with block importability. I guess we could get this somehow with other logs though. Will downgrade to `trace`. (Also not opposed to just removing it)

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: Good catch, wasn't sure what I was thinking!
The index is no longer used for inserting it into `PendingComponents` but is still important for persistence. I'll revert this.

beacon_node/http_api/src/publish_blocks.rs
jimmygchen: Yeah that's a good point. I initially run into an issue when sending all columns one-by-one to the `beacon_chain` because it wasn't expecting that additional data columns, and it basically imported the block after receiving `min_custody_requirement_columns`, and return `BlockIsAlreadyKnown` error for additional columns:

https://github.com/sigp/lighthouse/blob/41d622521dd7bf3d01dd3ce6769d76a83f125d08/beacon_node/beacon_chain/src/beacon_chain.rs#L2950-L2964

Perhaps we could add an additional function dedicated for proposer - that adds all the columns before importing? e.g. `chain.process_proposer_gossip_data_columns`?

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: Good question - I thought about this and my reason was that `pending_components.epoch` should always return an epoch unless there's a bug - there should be at least one component in `pending_components` when we're checking availability, and any component (block/blob/column) would have a slot / epoch attached to it.

I think it's safer to return an error here, in case if there's a bug in `pending_components.epoch()` - if we've already received all necessary components, this would never be called again and we end up having stuck pending components.

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: Addressed in [3424dcd](https://github.com/sigp/lighthouse/pull/5570/commits/3424dcd36ee14c3ebcdf21469a83e5a698677ef3)

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: Addressed in [3424dcd](https://github.com/sigp/lighthouse/pull/5570/commits/3424dcd36ee14c3ebcdf21469a83e5a698677ef3)

jimmygchen:
@dapplion Thanks for the review! I've addressed some of your comments, and raised some questions to the remaining.

dapplion (approved):
Looks great! Happy to merge as is, and continue improving once the rest of the moving parts land

beacon_node/http_api/src/publish_blocks.rs
jimmygchen: Added this to tracking issue https://github.com/sigp/lighthouse/issues/4983

PR #5573, opened 2024-04-12
Make beacon processor queue sizes dynamic

dapplion:
## Issue Addressed

Some users have complained of their nodes rejecting attestations due to full queues. The current size of 16k was good before, but it's insufficient for the current mainnet's size. Instead of playing catch with these numbers, we can make them dynamic on the validator set size. The set size is not meant to change significantly between client restarts so sampling once at start-up should be sufficient.

## Proposed Changes

Make beacon processor queue sizes dynamic, and some of its values variable on initial active validator set size. This change allows future testing of the processor where we want to make the queue very short to test their behavior.

**NOTE**: I have not changed any numbers except for `attestation_queue`, `unknown_block_attestation_queue`. Happy to change or justify existing numbers if you have rationales.

### DIFF
--- a/beacon_node/beacon_processor/src/lib.rs
+++ b/beacon_node/beacon_processor/src/lib.rs
@@ -60,7 +60,7 @@ use std::time::Duration;
 use task_executor::TaskExecutor;
 use tokio::sync::mpsc;
 use tokio::sync::mpsc::error::TrySendError;
-use types::{Attestation, Hash256, SignedAggregateAndProof, SubnetId};
+use types::{Attestation, BeaconState, ChainSpec, Hash256, SignedAggregateAndProof, SubnetId};
 use types::{EthSpec, Slot};
 use work_reprocessing_queue::IgnoredRpcBlock;
 use work_reprocessing_queue::{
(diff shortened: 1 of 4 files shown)


beacon_node/beacon_processor/src/lib.rs
```diff
@@ -85,123 +85,87 @@ const MAX_IDLE_QUEUE_LEN: usize = 16_384;
...
 
-/// The maximum number of priority-1 (second-highest priority) messages that will be queued before
-/// they begin to be dropped.
-const MAX_API_REQUEST_P1_QUEUE_LEN: usize = 1_024;
+impl BeaconProcessorQueueLengths {
+    pub fn from_state<E: EthSpec>(
+        state: &BeaconState<E>,
+        spec: &ChainSpec,
+    ) -> Result<Self, String> {
+        let active_validator_count = state
```
paulhauner: We could have a go at the cached method and fall back to the manual method if the caches aren't built:

https://github.com/sigp/lighthouse/blob/3058b96f2560f1da04ada4f9d8ba8e5651794ff6/consensus/types/src/beacon_state.rs#L545-L558

That would save us a state iteration on startup but the value of that isn't obviously high. Maybe it'd be worth the effort when we get tree-states and iteration is slow?

paulhauner (approved):
Looks good, very simple. I was thinking of *dynamic* as adjusting the queue size as the validator set grows but doing it once at startup is a big gain for very little complexity 👍 I don't think it's clear that *fully dynamic* is really worth the complexity.

I made a minor suggestion, feel free to take it or leave it.

beacon_node/beacon_processor/src/lib.rs
dapplion: +1 on not doing duplicated work. However, at this early start-up point are the caches expected to be built? Otherwise, this function trigger cache building.

beacon_node/beacon_processor/src/lib.rs
```diff
@@ -85,123 +85,87 @@ const MAX_IDLE_QUEUE_LEN: usize = 16_384;
...
+            .len();
+        let slots_per_epoch = E::slots_per_epoch() as usize;
+
+        Ok(Self {
+            aggregate_queue: 4096,
+            unknown_block_aggregate_queue: 1024,
+            // Capacity for a full slot's worth of attestations if subscribed to all subnets
+            attestation_queue: active_validator_count / slots_per_epoch,
+            // Capacity /2 of a full slot's worth of attestations, no specific reason for 2
+            unknown_block_attestation_queue: active_validator_count / slots_per_epoch / 2,
```
dapplion: @paulhauner do you recall any motivation for the /2 factor here relative to `attestation_queue`?

beacon_node/beacon_processor/src/lib.rs
```diff
@@ -85,123 +85,87 @@ const MAX_IDLE_QUEUE_LEN: usize = 16_384;
...
+            .get_active_validator_indices(state.slot().epoch(E::slots_per_epoch()), spec)
+            .map_err(|e| format!("Error computing active indices: {:?}", e))?
+            .len();
+        let slots_per_epoch = E::slots_per_epoch() as usize;
+
+        Ok(Self {
+            aggregate_queue: 4096,
+            unknown_block_aggregate_queue: 1024,
+            // Capacity for a full slot's worth of attestations if subscribed to all subnets
+            attestation_queue: active_validator_count / slots_per_epoch,
```
dapplion, line 150: To void the need of dynamic queues we can add some overhead margin on top of the expected value. The original hardcoded value did exceed the expected value at the time. With current inbound churn limitation active validator indices count can grow by 5% monthly max. So bumping this value by say 10% gives users a suitable target for 2 months at least.

beacon_node/beacon_processor/src/lib.rs
```diff
@@ -85,123 +85,87 @@ const MAX_IDLE_QUEUE_LEN: usize = 16_384;
...
-/// The maximum number of priority-1 (second-highest priority) messages that will be queued before
-/// they begin to be dropped.
-const MAX_API_REQUEST_P1_QUEUE_LEN: usize = 1_024;
+impl BeaconProcessorQueueLengths {
+    pub fn from_state<E: EthSpec>(
+        state: &BeaconState<E>,
+        spec: &ChainSpec,
+    ) -> Result<Self, String> {
+        let active_validator_count = state
+            .get_active_validator_indices(state.slot().epoch(E::slots_per_epoch()), spec)
```
michaelsproul:
```suggestion
            .get_active_validator_indices(state.current_epoch(), spec)
```

beacon_node/beacon_processor/src/lib.rs
michaelsproul: Agree, lets over-provision slightly

beacon_node/beacon_processor/src/lib.rs
michaelsproul: I think we should make this the same as the full capacity. In the worst case our node could be geographically isolated and get all the slot's unaggregated attestations (small-ish) prior to getting the block.

michaelsproul:
Nominating this for 5.2 as we are increasingly seeing queues filling up shortly after a restart

michaelsproul (approved):

michaelsproul:
@mergifyio queue

PR #5574, opened 2024-04-12
Lenient duplicate checks on HTTP API for block publication

michaelsproul:
## Issue Addressed

Fix the much-talked-about bug in Lighthouse's HTTP API which prevents blocks with blobs from being published if the block has already been seen on gossip

See: https://gist.github.com/benhenryhunter/7b7d9c9e3218aad52f75e3647b83a6cc

## Proposed Changes

- Change the signature of `process_blob` so that it takes a `publish_fn` which can be used to publish the block in the case where it is imported during blob processing.
- Split the `BlockIsAlreadyKnown` error into two cases so we can differentiate duplicate blocks that have already been fully imported from ones that are being imported/failed import.
- Extensively re-work `publish_block` so that it publishes all unseen blocks and blobs on gossip.
- Add new tests to cover scenarios involving partial publication of block & blobs on gossip.


michaelsproul:
Ready for review, but I wouldn't recommend that relays start testing this PR until the issues with `unstable` and v5.2.0 are resolved.

beacon_node/beacon_chain/src/blob_verification.rs
```diff
@@ -334,16 +334,16 @@ impl<E: EthSpec> KzgVerifiedBlobList<E> {
...
-        let blobs = blob_list.into_iter().collect::<Vec<_>>();
-        verify_kzg_for_blob_list(blobs.iter(), kzg)?;
+        let blobs = blob_list
+            .into_iter()
+            .map(|blob| KzgVerifiedBlob {
+                blob,
+                seen_timestamp,
+            })
+            .collect::<Vec<_>>();
+        verify_kzg_for_blob_list(blobs.iter().map(|b| &b.blob), kzg)?;
```
michaelsproul: This is a small optimisation to avoid allocating two `Vec`s.

beacon_node/http_api/tests/broadcast_validation_tests.rs
```diff
@@ -1395,3 +1400,366 @@ pub async fn blinded_equivocation_full_pass() {
...
+    assert!(!tester
+        .harness
+        .chain
+        .block_is_known_to_fork_choice(&block.canonical_root()));
+
+    // Post just the block to the HTTP API.
+    let response: Result<(), eth2::Error> = tester
+        .client
+        .post_beacon_blocks_v2(
+            &PublishBlockRequest::new(block.clone(), None),
```
jimmygchen: From Deneb the request is assumed to be `PublishBlockRequest::BlockContents`:
- ssz for Deneb request is always decoded to `BlockContents`: https://github.com/sigp/lighthouse/blob/79680c886dcdd1577edbd3a93f920df221cafc8c/common/eth2/src/types.rs#L1825 
- `TryFrom` impl does not allow constructing from only signed block: https://github.com/sigp/lighthouse/blob/79680c886dcdd1577edbd3a93f920df221cafc8c/common/eth2/src/types.rs#L1909-L1911

Although I think this code path is still possible from VC sending json body.

beacon_node/http_api/src/publish_blocks.rs
```diff
@@ -119,130 +144,252 @@ pub async fn publish_block<T: BeaconChainTypes, B: IntoGossipVerifiedBlockConten
...
-                } else {
-                    error!(
-                        log,
-                        "Invalid blob provided to HTTP API";
-                        "reason" => &msg
-                    );
-                    Err(warp_utils::reject::custom_bad_request(msg))
-                };
+        Err(BlockError::DuplicateFullyImported(root)) => {
+            if published.load(Ordering::SeqCst) {
```
jimmygchen: This is always false for the `BroadcastValidation::Gossip` case, even when blobs are published I think?

jimmygchen: Never mind, I misunderstood the code, and there's [a test case](https://github.com/sigp/lighthouse/blob/4f3cf2c7292a127c76cc7c81063fcb408fc7a1d7/beacon_node/http_api/tests/broadcast_validation_tests.rs#L1407) that covers this scenario!

jimmygchen: Trying to understand the `published` logic here - in what scenario would this be `false` here?

When `BroadcastValidation::Gossip` == true, we hit this code path if
- gossip verification for block fails with `DuplicateFullyImported`
- AND none of the blob import fails
- `published` is always true in this scenario

When validation level is not gossip, `published` remains `false` if `publish_block_p2p` fails, or if if we get a `BlockError::DuplicateFullyImported` when processing blobs - in both cases we return a `custom_bad_request` rejection I think?

beacon_node/http_api/tests/broadcast_validation_tests.rs
```diff
@@ -1395,3 +1400,366 @@ pub async fn blinded_equivocation_full_pass() {
...
+    // This should result in the block being fully imported.
+    response.unwrap();
+    assert!(tester
+        .harness
+        .chain
+        .block_is_known_to_fork_choice(&block.canonical_root()));
+}
+
+/// This test checks that an HTTP POST request with the block & blobs succeeds with a 200 response
+/// even if the block has already been seen on gossip without any blobs.
```
jimmygchen: `without any blobs` -> `without all blobs` or something similar

beacon_node/http_api/tests/broadcast_validation_tests.rs
jimmygchen: I'm wonder under what scenario does this happen? or is this test to make sure we cover all cases?

beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -512,7 +512,8 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
 
         let action = match result {
             BlockProcessingResult::Ok(AvailabilityProcessingStatus::Imported(_))
-            | BlockProcessingResult::Err(BlockError::BlockIsAlreadyKnown(_)) => {
+            | BlockProcessingResult::Err(BlockError::DuplicateFullyImported(..))
+            | BlockProcessingResult::Err(BlockError::DuplicateImportStatusUnknown(..)) => {
                 // Successfully imported
```
jimmygchen, line 521: This comment may be outdated now?

jimmygchen (changes requested):
I've added a few more comments. Looks great!

I wasn't expecting that much change, but looks like we do need to cover all the scenarios  you've identified here. Nice tests you've added here - it helps me understand what we're trying to achieve here! 👍 

Looks like I will have to rework quite a bit in PeerDAS 💀

beacon_node/http_api/src/publish_blocks.rs
michaelsproul: Oh yes you're right. I think I intended to _not_ return an error early here if the blobs are duplicates:

https://github.com/sigp/lighthouse/blob/27113b28960fff9462dc682dd2b09b93025de63e/beacon_node/http_api/src/publish_blocks.rs#L275-L276

The `false` case is meant to be the one where block & blobs were fully known and already imported.

beacon_node/http_api/src/publish_blocks.rs
```diff
@@ -119,130 +144,252 @@ pub async fn publish_block<T: BeaconChainTypes, B: IntoGossipVerifiedBlockConten
...
+                    warp_utils::reject::custom_bad_request(format!("{e:?}"))
+                })
+            })
+            .collect::<Result<Vec<_>, Rejection>>()?
+    } else {
+        vec![]
+    };
 
-    let block_root = block_root.unwrap_or(gossip_verified_block.block_root);
+    // Gossip verify the block and blobs separately.
```
jimmygchen: I think we'll have to break down some of the logic here into smaller functions - this will be really long when we introduce peerdas, but I'm happy to do this later in a separate PR.

beacon_node/http_api/src/publish_blocks.rs
michaelsproul: I'll try writing a new test that fails and then tweaking the logic to make it pass

beacon_node/http_api/tests/broadcast_validation_tests.rs
michaelsproul: This is meant to test the scenario where the caller POSTs a block without blobs, but you're right that it would be better to do that with a `PublishBlockRequest::BlockContents` where the list of blobs is empty. I will update accordingly.

michaelsproul: Test updated

beacon_node/http_api/src/publish_blocks.rs
michaelsproul: Ok the test I wrote passed, and also hit this branch! :tada:

See commit 52495eeef31289cb05e776dfeb3c5f3cb183ecd3

The reason it doesn't fail at the `process_gossip_blob` step is because the duplicate blobs have already been filtered out (converted to `None`) when we construct the gossip-verified blobs

beacon_node/network/src/sync/block_lookups/mod.rs
michaelsproul: Asked for input on this from LH network gang

michaelsproul: I have a `TODO` on this to add a comment saying this is unreachable because we never send RPC blocks for gossip verification, so we can't hit `DuplicateImportStatusUnknown`

beacon_node/http_api/src/publish_blocks.rs
```diff
@@ -119,130 +144,252 @@ pub async fn publish_block<T: BeaconChainTypes, B: IntoGossipVerifiedBlockConten
...
+                    // by the other publisher.
+                    debug!(
+                        log,
+                        "Blob for publication already known";
+                        "blob_index" => blob_sidecar.index,
+                        "slot" => slot,
+                        "proposer" => proposer,
+                    );
+                    Ok(None)
+                }
```
michaelsproul, line 251: @jimmygchen I think this might handle the case where a block is being imported from gossip and HTTP, and both imports trip the `RepeatBlob` condition.

With this change, the HTTP API will continue with trying to import the block.

If the block is also seen on gossip before HTTP we will get a `DuplicateImportStatusUnknown` and still call `process_block`

jimmygchen: Yep that looks right!

jimmygchen: So i think our issue is we're re-processing the block here:
https://github.com/sigp/lighthouse/blob/4f3cf2c7292a127c76cc7c81063fcb408fc7a1d7/beacon_node/http_api/src/publish_blocks.rs#L336-L343

Which is probably already imported earlier via `process_gossip_blob`, and hence will always been rejected as `DuplicateFullyImported`
https://github.com/sigp/lighthouse/blob/4f3cf2c7292a127c76cc7c81063fcb408fc7a1d7/beacon_node/http_api/src/publish_blocks.rs#L270-L273

And because of the `BlockError` from `process_block`, the head is not recomputed in `post_block_import_logging_and_response`. We should probably just use the `AvailabilityProcessingStatus` from last `process_gossip_blob` call? and if the block is  imported it will recompute the head.

jimmygchen: Logs and some notes here:

```
# [gossip] block received
2024-07-29T05:22:23.1173355Z Jul 29 05:22:23.117 INFO New block received                      root: 0xa37555b6829209e977de1a8ed8a40651e4536e92d0c7c39bdeb81e162749dea8, slot: 43, service: node_1

# [gossip] blob 0 received
2024-07-29T05:22:23.1201914Z Jul 29 05:22:23.120 DEBG Successfully verified gossip blob       commitment: 0xa41e…f05f, index: 0, root: 0xa375…dea8, slot: 43, service: node_1
2024-07-29T05:22:23.1378924Z Jul 29 05:22:23.137 DEBG Beacon block awaiting blobs             block_slot: 43, block_root: 0xa37555b6829209e977de1a8ed8a40651e4536e92d0c7c39bdeb81e162749dea8, service: beacon, service: node_1

# [gossip] blob 1 received
2024-07-29T05:22:23.1499531Z Jul 29 05:22:23.149 DEBG Successfully verified gossip blob       commitment: 0xa41e…f05f, index: 1, root: 0xa375…dea8, slot: 43, service: node_1

# [http] block and blobs (0,1,2,3) received
2024-07-29T05:22:23.1590330Z Jul 29 05:22:23.158 DEBG Signed block received in HTTP API       slot: 43, service: node_1

# [http] blob 0 already known from gossip
2024-07-29T05:22:23.1621330Z Jul 29 05:22:23.161 DEBG Blob for publication already known      proposer: 9, slot: 43, blob_index: 0, service: node_1

# [http] blob 1 already known from gossip
2024-07-29T05:22:23.1623295Z Jul 29 05:22:23.162 DEBG Blob for publication already known      proposer: 9, slot: 43, blob_index: 1, service: node_1

# [gossip] blob 2 already known from http
2024-07-29T05:22:23.1742409Z Jul 29 05:22:23.171 DEBG Could not verify blob sidecar for gossip. Ignoring the blob sidecar, commitment: 0xa41e…f05f, index: 2, root: 0xa375…dea8, slot: 43, error: RepeatBlob { proposer: 9, slot: Slot(43), index: 2 }, service: node_1

# [http] nothing is actually published here it seems, because of the import below, and gossip block received beforehand, so `should_import` = false. We might want to remove this log?
2024-07-29T05:22:23.1783377Z Jul 29 05:22:23.178 INFO Signed block published to network via HTTP API, publish_delay_ms: 19, slot: 43, service: node_1

# [http] we're in `chain.import_block`! (possibly via `chain.process_gossip_blob` in `publish_block`)
2024-07-29T05:22:23.1816196Z Jul 29 05:22:23.181 DEBG Writing blobs to store                  count: 4, block_root: 0xa375…dea8, service: beacon, service: node_1

# [gossip] blob 3 already known from http
2024-07-29T05:22:23.1838572Z Jul 29 05:22:23.181 DEBG Could not verify blob sidecar for gossip. Ignoring the blob sidecar, commitment: 0xa41e…f05f, index: 3, root: 0xa375…dea8, slot: 43, error: RepeatBlob { proposer: 9, slot: Slot(43), index: 3 }, service: node_1

# [http] api partially publishing blobs in `chain.process_gossip_blob`
2024-07-29T05:22:23.1865775Z Jul 29 05:22:23.185 DEBG Sending pubsub messages                 topics: [BlobSidecar(2), BlobSidecar(3)], count: 2, service: network, service: node_1

# [http] block seen from gossip, import status unknown, try processing the block. 
2024-07-29T05:22:23.1902460Z Jul 29 05:22:23.186 DEBG Block previously seen                   slot: 43, block_root: 0xa37555b6829209e977de1a8ed8a40651e4536e92d0c7c39bdeb81e162749dea8, service: node_1

# [http] rejected. We've already imported the block in `chain.process_gossip_blob` and we should probably look at the last `AvailabilityProcessingStatus` from that call instead of looking at an outdated process_block result?
2024-07-29T05:22:23.1905042Z Jul 29 05:22:23.187 DEBG Beacon block rejected                   reason: DuplicateFullyImported(0xa37555b6829209e977de1a8ed8a40651e4536e92d0c7c39bdeb81e162749dea8), service: beacon, service: node_1
2024-07-29T05:22:23.1907689Z Jul 29 05:22:23.187 WARN Error processing HTTP API request       method: POST, path: /eth/v1/beacon/blocks, status: 202 Accepted, elapsed: 54.184048ms, service: node_1
2024-07-29T05:22:24.0035956Z Jul 29 05:22:24.001 DEBG Slot timer                              sync_state: Synced, current_slot: 43, head_slot: 42, head_block: 0x2c26…f8b3, finalized_epoch: 3, finalized_root: 0x83ad…d15a, peers: 5, service: slot_notifier, service: node_1

# block is not yet head. That's expected because we've imported block but haven't recomputed head.
2024-07-29T05:22:24.0037668Z Jul 29 05:22:24.001 INFO Synced                                  slot: 43, block:    …  empty, epoch: 5, finalized_epoch: 3, finalized_root: 0x83ad…d15a, exec_hash: 0x694a…b4c0 (verified), peers: 5, service: slot_notifier, service: node_1
1918, slot: 43, proposer_index: 9, block_root: 0xa37555b6829209e977de1a8ed8a40651e4536e92d0c7c39bdeb81e162749dea8, service: beacon, service: node_1
```

beacon_node/beacon_chain/src/blob_verification.rs
michaelsproul: Pulled this out into a separate PR to make this PR easier to review:

- https://github.com/sigp/lighthouse/pull/6282

beacon_node/http_api/src/publish_blocks.rs
michaelsproul: Resolved this case in d37b0806d948cb15f4317ad923a0ba70cab28ed1.

I am going to try soak testing the simulator with this branch.

beacon_node/http_api/src/publish_blocks.rs
```diff
@@ -59,28 +86,29 @@ pub async fn publish_block<T: BeaconChainTypes, B: IntoGossipVerifiedBlockConten
...
 
     /* actually publish a block */
-    let publish_block = move |block: Arc<SignedBeaconBlock<T::EthSpec>>,
-                              blobs_opt: Option<BlobSidecarList<T::EthSpec>>,
-                              data_cols_opt: Option<DataColumnSidecarList<T::EthSpec>>,
-                              sender,
-                              log,
-                              seen_timestamp| {
+    let publish_block_p2p = move |block: Arc<SignedBeaconBlock<T::EthSpec>>,
+                                  should_publish: bool,
```
jimmygchen: might be clearer to rename this to `should_publish_block`

beacon_node/http_api/src/publish_blocks.rs
```diff
@@ -150,168 +171,362 @@ pub async fn publish_block<T: BeaconChainTypes, B: IntoGossipVerifiedBlockConten
...
-                &block_clone,
-                &log_clone,
-            )?;
-            publish_block(
-                block_clone,
-                blobs_opt,
-                data_cols_opt,
-                sender_clone,
-                log_clone,
+    let published = Arc::new(AtomicBool::new(false));
```
jimmygchen: I think this is slightly confusing because publish may have happened already for the `BroadcastValidation::Gossip` case above. If the intent is to track whether we completed the `publish_fn`, maybe we could rename it to `publish_fn_completed`?

beacon_node/http_api/src/publish_blocks.rs
```diff
@@ -1,53 +1,80 @@
...
-    AvailabilityProcessingStatus, BeaconChain, BeaconChainError, BeaconChainTypes, BlockError,
-    IntoGossipVerifiedBlockContents, NotifyExecutionLayer,
+    build_blob_data_column_sidecars, AvailabilityProcessingStatus, BeaconChain, BeaconChainError,
+    BeaconChainTypes, BlockError, IntoGossipVerifiedBlock, NotifyExecutionLayer,
+};
+use eth2::types::{
+    BlobsBundle, BroadcastValidation, ErrorMessage, ExecutionPayloadAndBlobs, FullPayloadContents,
+    PublishBlockRequest, SignedBlockContents,
 };
-use eth2::types::{into_full_block_and_blobs, BroadcastValidation, ErrorMessage};
```
jimmygchen, line 9: looks like the function `eth2::types::into_full_block_and_blobs` is no longer used and can be removed.

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -187,12 +180,18 @@ pub enum BlockError {
...
     /// imported.
     NotFinalizedDescendant { block_parent_root: Hash256 },
-    /// Block is already known, no need to re-import.
+    /// Block is already known and valid, no need to re-import.
     ///
     /// ## Peer scoring
     ///
     /// The block is valid and we have already imported a block with this hash.
-    BlockIsAlreadyKnown(Hash256),
+    DuplicateFullyImported(Hash256),
```
jimmygchen, line 188: There's one more ocurrence of `BlockIsAlreadyKnown` here:

https://github.com/sigp/lighthouse/blob/c0b4f01cf37257bab8dab2b4312dcf07eac93572/beacon_node/network/src/sync/network_context.rs#L620

jimmygchen (changes requested):
Just a few more minor comments, otherwise looks good to me. Would be good to get another pair of eyes on this.

There's probably some potential refactor we can do in `publish_block`, but perhaps we can do it in a separate PR, as this PR has a few simplifications that a few other changes depend on, e.g. PeerDAS publish block optimisations made in b0e6283f. The added tests in this PR should make future refactoring easier.

beacon_node/http_api/src/publish_blocks.rs
realbigsean: why does this bool not also suppress blob publish?

realbigsean: michael explained to me offline the blobs should be published with gossip validation level so this is correct, and they *won't* be published when the block gossip validation fails for consensus level validation because this is called in `publish_fn` only after the block + blobs are available

realbigsean (approved):
looks good!

beacon_node/beacon_chain/src/block_verification.rs
michaelsproul: good catch, done in 0f7b5a14c3fd5bb648434a1d99b99c20229cc58f

beacon_node/http_api/src/publish_blocks.rs
michaelsproul: nice! deleted in 0f7b5a14c3fd5bb648434a1d99b99c20229cc58f

beacon_node/http_api/src/publish_blocks.rs
michaelsproul: yeah that's much better, 0f7b5a14c3fd5bb648434a1d99b99c20229cc58f

beacon_node/http_api/src/publish_blocks.rs
michaelsproul: yeah, renamed to `should_publish_block` as that's much clearer 0f7b5a14c3fd5bb648434a1d99b99c20229cc58f

jimmygchen (approved):
LGTM!

michaelsproul:
@mergify queue

michaelsproul:
tasty insta merge

finally

PR #5575, opened 2024-04-12
Fix stuck backfill when scheduled work queue is at capacity

### LINKED ISSUE #5504: Backfill stuck on `Failed to send scheduled backfill work`
## Description

User has reported stuck backfill which happens after this log:

```
Mar 29 07:08:06.001 ERRO Failed to send scheduled backfill work, info: sending work back to queue, service: bproc, module: beacon_processor::work_reprocessing_queue:884
```
We do re-insert the batch into `queued_backfill_batches` [here](https://github.com/sigp/lighthouse/blob/stable/beacon_node/beacon_processor/src/work_reprocessing_queue.rs#L889). Do we need to also update `next_backfill_batch_event` like we do [here](https://github.com/sigp/lighthouse/blob/stable/beacon_node/beacon_processor/src/work_reprocessing_queue.rs#L751-L753)?

jimmygchen:
## Issue Addressed

Addresses #5504.

Fix stuck backfill when scheduled work queue is at capacity and make sure the batch is rescheduled.

## Additional Info

~I tried adding a regression test but can't get it to pass. Will look into it next week.~ (fixed)


beacon_node/beacon_processor/src/work_reprocessing_queue.rs
```diff
@@ -1002,4 +1020,80 @@ mod tests {
...
+        work_reprocessing_tx
+            .try_send(ReprocessQueueMessage::BackfillSync(QueuedBackfillBatch(
+                Box::pin(async {}),
+            )))
+            .unwrap();
+        tokio::task::yield_now().await;
+
+        // Advance the time by more than 1/2 the slot to trigger a scheduled backfill batch to be sent.
+        // This should fail as the `ready_work` channel is at capacity, and it should be rescheduled.
+        let more_than_half_slot = Duration::from_secs(slot_duration / 2 + 1);
```
realbigsean: So this is to hit the timing defined by `BACKFILL_SCHEDULE_IN_SLOT`, right? Can we use that directly here so as to not break this test in the future if this schedule changes?

realbigsean (changes requested):
This style of test seems like a good way to deal with timings! Nice!

beacon_node/beacon_processor/src/work_reprocessing_queue.rs
jimmygchen: Done in https://github.com/sigp/lighthouse/pull/5575/commits/a5f76cea0bedc97bbbdfc9f9a9ba6b0c30992f65

realbigsean (approved):

realbigsean:
@mergify queue

PR #5582, opened 2024-04-13
Ensure proper ReqResp blocks_by_* response stream termination

dapplion:
## Issue Addressed

Apply
- https://github.com/sigp/lighthouse/pull/5556

To blocks_by_* routes after merging
- https://github.com/sigp/lighthouse/pull/5557

## Proposed Changes

Covers some codepaths that were not actually closing the stream properly. The difference with this PR is the server will return more details about error than before. We are currently communicating the general cause of error to the caller already so IMO it's fine.


realbigsean (approved):
Awesome

realbigsean:
@mergify queue

PR #5583, opened 2024-04-13
Handle sync lookup request streams in network context

dapplion:
## Issue Addressed

Part of

Currently the sync lookup request states deal with tracking streams, retries and their own sync logic (parent chains, etc). The complexities of streams can be abstracted up into the network context to simplify downstream code.

This abstraction would have avoided the previous month bug regarding `ExtraBlocks`

## Proposed Changes

Adopt the same accumulator pattern used today for range blocks + blobs for lookup requests. Allow to merge error handling into the same codepath and remove the concept of response validation from the RequestState trait.

In draft to wait for https://github.com/sigp/lighthouse/pull/5563

### DIFF
--- a/beacon_node/network/src/sync/block_lookups/mod.rs
+++ b/beacon_node/network/src/sync/block_lookups/mod.rs
@@ -392,7 +392,6 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
                         .get_state_mut()
                         .on_download_success()
                         .map_err(LookupRequestError::BadState)?;
-
                     self.send_block_for_processing(
                         block_root,
                         block,
@@ -403,6 +402,10 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
                 }
             }
             CachedChild::DownloadIncomplete => {
+                R::request_state_mut(lookup)
+                    .get_state_mut()
+                    .on_download_success()
+                    .map_err(LookupRequestError::BadState)?;
                 // If this was the result of a block request, we can't determine if the block peer
                 // did anything wrong. If we already had both a block and blobs response processed,
                 // we should penalize the blobs peer because they did not provide all blobs on the
--- a/beacon_node/network/src/sync/block_lookups/mod.rs
+++ b/beacon_node/network/src/sync/block_lookups/mod.rs
@@ -126,7 +126,7 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
     pub fn trigger_single_lookup(
         &mut self,
         mut single_block_lookup: SingleBlockLookup<Current, T>,
-        cx: &SyncNetworkContext<T>,
+        cx: &mut SyncNetworkContext<T>,
     ) {
         let block_root = single_block_lookup.block_root();
         match single_block_lookup.request_block_and_blobs(cx) {
@@ -312,36 +312,32 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
         &mut self,
         lookup_id: SingleLookupReqId,
         peer_id: PeerId,
-        response: Option<R::ResponseType>,
+        response: R::VerifiedResponseType,
         seen_timestamp: Duration,
-        cx: &SyncNetworkContext<T>,
+        cx: &mut SyncNetworkContext<T>,
     ) {
         let id = lookup_id.id;
         let response_type = R::response_type();
 
         let Some(lookup) = self.get_single_lookup::<R>(lookup_id) else {
-            if response.is_some() {
-                // We don't have the ability to cancel in-flight RPC requests. So this can happen
-                // if we started this RPC request, and later saw the block/blobs via gossip.
-                debug!(
-                    self.log,
-                    "Block returned for single block lookup not present";
-                        "response_type" => ?response_type,
-                );
-            }
+            // We don't have the ability to cancel in-flight RPC requests. So this can happen
+            // if we started this RPC request, and later saw the block/blobs via gossip.
+            debug!(
+                self.log,
+                "Block returned for single block lookup not present";
+                    "response_type" => ?response_type,
+            );
             return;
         };
 
         let expected_block_root = lookup.block_root();
-        if response.is_some() {
-            debug!(self.log,
-                "Peer returned response for single lookup";
-                "peer_id" => %peer_id ,
-                "id" => ?id,
-                "block_root" => ?expected_block_root,
-                "response_type" => ?response_type,
-            );
-        }
+        debug!(self.log,
+            "Peer returned response for single lookup";
+            "peer_id" => %peer_id ,
+            "id" => ?id,
+            "block_root" => ?expected_block_root,
+            "response_type" => ?response_type,
+        );
 
         match self.single_lookup_response_inner::<R>(peer_id, response, seen_timestamp, cx, lookup)
         {
@@ -368,9 +364,9 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
     fn single_lookup_response_inner<R: RequestState<Current, T>>(
         &self,
         peer_id: PeerId,
-        response: Option<R::ResponseType>,
+        response: R::VerifiedResponseType,
         seen_timestamp: Duration,
-        cx: &SyncNetworkContext<T>,
+        cx: &mut SyncNetworkContext<T>,
         mut lookup: SingleBlockLookup<Current, T>,
     ) -> Result<SingleBlockLookup<Current, T>, LookupRequestError> {
         let response_type = R::response_type();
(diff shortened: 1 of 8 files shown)


beacon_node/network/src/sync/network_context.rs
```diff
@@ -45,6 +61,9 @@ pub struct SyncNetworkContext<T: BeaconChainTypes> {
     /// A sequential ID for all RPC requests.
     request_id: Id,
 
+    blocks_by_root_requests: FnvHashMap<SingleLookupReqId, ActiveBlocksByRootRequest>,
```
realbigsean, line 92: Could you add docs here

beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -1006,20 +948,21 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
             }
             BlockProcessingResult::Ok(AvailabilityProcessingStatus::Imported(_))
             | BlockProcessingResult::Err(BlockError::BlockIsAlreadyKnown(_)) => {
+                let (chain_hash, blocks, hashes, block_request) =
+                    parent_lookup.parts_for_processing();
+
+                let blocks = self.add_child_block_to_chain(chain_hash, blocks, cx).into();
+
+                let process_id = ChainSegmentProcessId::ParentLookup(chain_hash);
```
realbigsean, line 962: Is the ordering change here meant to drop the child lookup when  the beacon processor is not available? makes sense

beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -665,8 +603,8 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
                 // This happens if the peer disconnects while the block is being
                 // processed. Drop the request without extra penalty
             }
-            RequestError::BadState(_) => {
-                // Should never happen
+            RequestError::BadState(..) => {
+                // Internal error should never happen
```
realbigsean: Maybe this one warrants a `WARN` log so we could get a user report if it happens

beacon_node/network/src/sync/network_context.rs
```diff
@@ -405,4 +421,93 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+                    request.remove();
+                    Err(e)
+                }
+            },
+            RpcEvent::StreamTermination => {
+                // Stream terminator
+                match request.remove().terminate() {
+                    // TODO: Should deal only with Vec<Arc<BlobSidecar>>
+                    Some(blobs) => to_fixed_blob_sidecar_list(blobs)
+                        // TODO: a seen_timestamp for an array of blobs doesn't make much sense
```
realbigsean: The purpose is to see how long it takes a blobs by root request to complete, so it makes sense to have it at stream termination. Maybe a misnomer.

beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -1368,6 +1319,36 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
...
+        peer_id: &PeerId,
+        error: &RPCError,
+        cx: &SyncNetworkContext<T>,
+    ) {
+        // Note: logging the report event here with the full error display. The log inside
+        // `report_peer` only includes a smaller string, like "invalid_data"
+        debug!(self.log, "reporting peer for sync lookup error"; "error" => %error);
+        if let Some(action) = match error {
+            // Protocol errors are heavily penalized
+            RPCError::SSZDecodeError(..)
```
realbigsean: Looks like we already downscore here on RPC errors: https://github.com/sigp/lighthouse/blob/unstable/beacon_node/lighthouse_network/src/service/mod.rs#L1407

So on everything apart from the new `InvalidData`s we'd double downscore.   My suggestion would be to revive the `LookupVerifyError` and add a wrapper for the two:
```
enum LookupFailure {
    LookupVerifyError(LookupVerifyError),
    RpcError(RpcError)
}

match e {
    LookupVerifyError(e) => {..} // downscore and drop,
    RpcError(e) => {..} // drop
}
```
Or maybe add an `RpcError` variant to the `LookupVerifyError`.

beacon_node/network/src/sync/network_context.rs
```diff
@@ -405,4 +421,93 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+            return None;
+        };
+
+        Some(match block {
+            RpcEvent::Response(block, seen_timestamp) => {
+                match request.get_mut().add_response(block) {
+                    Ok(block) => Ok((block, seen_timestamp)),
+                    Err(e) => {
+                        // The request must be dropped after receiving an error.
+                        // TODO: We could NOT drop the request here, and penalize the peer again if
```
realbigsean: Is your intent to add this in a follow up? If so we can move it to an issue

beacon_node/network/src/sync/network_context.rs
```diff
@@ -405,4 +421,93 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+        request_id: SingleLookupReqId,
+        blob: RpcEvent<Arc<BlobSidecar<T::EthSpec>>>,
+    ) -> RpcProcessingResult<FixedBlobSidecarList<T::EthSpec>> {
+        let Entry::Occupied(mut request) = self.blobs_by_root_requests.entry(request_id) else {
+            return None;
+        };
+
+        Some(match blob {
+            RpcEvent::Response(blob, _) => match request.get_mut().add_response(blob) {
+                // TODO: Should deal only with Vec<Arc<BlobSidecar>>
```
realbigsean: Should this be moved to an issue as well?

realbigsean (changes requested):
Already finding this much easier to follow!

beacon_node/network/src/sync/block_lookups/mod.rs
realbigsean: https://github.com/dapplion/lighthouse/pull/24/commits/507bcc6d15e71135d4d2444717281e5bd8962883

beacon_node/network/src/sync/network_context.rs
realbigsean: https://github.com/dapplion/lighthouse/pull/24/commits/50c007b44fe8e78d9e94a79e9e0bc8c074013d8e

beacon_node/network/src/sync/network_context.rs
realbigsean: removed TODO here https://github.com/dapplion/lighthouse/pull/24/commits/47ca705cf11b145054063fc3230db13a9f3e58e0

beacon_node/network/src/sync/block_lookups/mod.rs
realbigsean: Made these updates here: https://github.com/dapplion/lighthouse/pull/24/commits/7ed104f23a5fa4462343613e0b00af91c98ce576

beacon_node/network/src/sync/block_lookups/mod.rs
realbigsean: no change requested

beacon_node/network/src/sync/network_context.rs
realbigsean: removed TODO here: https://github.com/dapplion/lighthouse/pull/24/commits/c8ac5a0492a14909eb6521ddaeaef388ea5e8cd9

issue here: https://github.com/sigp/lighthouse/issues/5604

beacon_node/network/src/sync/network_context.rs
realbigsean: removed TODO here: https://github.com/dapplion/lighthouse/pull/24/commits/c8ac5a0492a14909eb6521ddaeaef388ea5e8cd9

issue here: https://github.com/sigp/lighthouse/issues/5603

beacon_node/network/src/sync/network_context.rs
```diff
@@ -405,4 +421,93 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+                    Err(e) => {
+                        // The request must be dropped after receiving an error.
+                        // TODO: We could NOT drop the request here, and penalize the peer again if
+                        // sends multiple penalizable chunks after the first invalid.
+                        request.remove();
+                        Err(e)
+                    }
+                }
+            }
+            RpcEvent::StreamTermination => match request.remove().terminate() {
```
jimmygchen, line 477: Do we need to handle the scenario where a peer doesn't send stream termination? (`ActiveBlocksByRootRequest` stored is quite small in size though)

dapplion: Current sync codebase expects the stream termination to always appear when syncing or doing blob lookups. The API contract with network's ReqResp behaviour is that you always get either an error or stream termination. There are timeouts in place in case the peer leaves the stream open and doesn't send the termination.

beacon_node/network/src/sync/block_lookups/mod.rs
dapplion: The change is motivated by the borrow checker, but as you say the change also makes sense to drop the lookup completely on that case

realbigsean (approved):
woop woop

realbigsean:
@mergify queue

PR #5007, opened 2023-12-13
Config for web3signer keep-alive

michaelsproul:
## Issue Addressed

Addresses #4439 in the context of web3signer connections.

One of the large staking pools was having issues with web3signer connections failing with `IncompleteMessage` errors.

## Proposed Changes

- Add a flag to control the connection idle timeout. Web3signer uses 30s by default so anything shorter than this should help. In practice the staking pool found success using an extremely short 10ms cutoff, but we didn't experiment with pushing it higher.
- Add a flag to control connection pooling. Setting the pool size to 0 is reported to help somewhat (see https://github.com/hyperium/hyper/issues/2136).

## Additional Info

I'm not sure if this is the best way to solve this. Requiring the node operator to fiddle with yet more CLI flags is bad from a UX PoV. We may be better off adopting automatic retry.

PR #5587, opened 2024-04-16
Set web3signer keep-alive to 20s by default

michaelsproul:
## Issue Addressed

Follow-up to:

- #5007

## Proposed Changes

Set the default keep-alive to 20s, which is shorter than web3signer's default 30s.

Several of our users are successfully making use of this flag now, so I think it makes sense to set the default to a value that works. CC @guybrush (of EthPool/beaconcha.in)


realbigsean (approved):

realbigsean:
@mergify queue

realbigsean:
@mergify requeue

michaelsproul:
@mergifyio requeue

michaelsproul:
@mergifyio queue

realbigsean:
@mergify requeue

Issue #3307, opened 2022-07-05
Remove redundancy in the ulrs in the execution layer' config


divagant-martian:
## Description

After removing the fallback for engines in the execution_layer crate, we don't need the `secret_files` and `execution_endpoint` to be able to hold more than one in each field. Those could be changed to be options or even mandatory fields.

The relevant code is [here](https://github.com/sigp/lighthouse/blob/748475be1ded2c9a29cbd8b985f6ae2980720a36/beacon_node/execution_layer/src/lib.rs#L119-L123)

AndyBarnard:
between the code example and this issue I can't tell - are you saying `execution_endpoints` and `secret_files` don't need to be vectors, and can be `Option`s, or are you saying just `builder_url` should be an `Option`?

divagant-martian:
`builder_url` is not related to this issue. Just `execution_endpoints` and `secret_files`. Happy to help if you want to give this a go

AndyBarnard:
ah my fork didn't have `builder_endpoints` as an `Option`, got a little mixed up there. I'll give it a go now!

divagant-martian:
little checkup for whoever is interested. Still relevant to this date

iamit-singh:
Hi @divagant-martian, please review my [PR](https://github.com/sigp/lighthouse/pull/5588) for this issue, thanks!

michaelsproul:
Fixed by 

- #5588

PR #5588, opened 2024-04-16
Fix execution layer redundancy

iamit-singh:
## Issue Addressed

#3307

## Proposed Changes

renamed and changed types(from Vec to Option) of below fields of Config struct:
execution_endpoints: Vec => execution_endpoint: Option
secret_files: Vec => secret_file: Option

### DIFF
--- a/beacon_node/beacon_chain/src/test_utils.rs
+++ b/beacon_node/beacon_chain/src/test_utils.rs
@@ -408,7 +408,7 @@ where
         self
     }
 
-q    pub fn execution_layer_from_url(mut self, url: &str) -> Self {
+    pub fn execution_layer_from_url(mut self, url: &str) -> Self {
         assert!(
             self.execution_layer.is_none(),
             "execution layer already defined"
--- a/beacon_node/beacon_chain/src/test_utils.rs
+++ b/beacon_node/beacon_chain/src/test_utils.rs
@@ -408,21 +408,20 @@ where
         self
     }
 
-    pub fn execution_layer_from_urls(mut self, urls: &[&str]) -> Self {
+q    pub fn execution_layer_from_url(mut self, url: &str) -> Self {
         assert!(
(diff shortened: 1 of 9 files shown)


beacon_node/beacon_chain/src/test_utils.rs
```diff
@@ -408,21 +408,20 @@ where
...
 
-        let urls: Vec<SensitiveUrl> = urls
-            .iter()
-            .map(|s| SensitiveUrl::parse(s))
-            .collect::<Result<_, _>>()
-            .unwrap();
+        let url = match SensitiveUrl::parse(url) {
+            Ok(s) => Some(s),
+            Err(_) => None,
+        };
```
realbigsean: you could add a bit of sugar here:
```suggestion
        let url = SensitiveUrl::parse(url).ok();
```

realbigsean (dismissed):
Thanks for the contribution! It's a nice cleanup, just had one small suggestion

realbigsean:
Looks like some tests need updating as well, if you want to check locally whether tests compile `make lint` will also check them

iamit-singh:
@realbigsean thanks for review. Actually I ran `make lint` in my local, strangely it doesn't give any error. Anyways, I see the errors in here, I'll work on them.

realbigsean:
> @realbigsean thanks for review. Actually I ran `make lint` in my local, strangely it doesn't give any error. Anyways, I see the errors in here, I'll work on them.

ah there must be some tests are are missed. usually i also run `make test-beacon-chain` and that seems to cover the rest

iamit-singh:
@realbigsean fixed the tests!

realbigsean (approved):
Awesome!

realbigsean:
@mergify queue

realbigsean:
@mergify requeue

michaelsproul:
@mergifyio requeue

PR #5591, opened 2024-04-16
Rename Functions to More Closely Match Spec

ethDreamer:
## Issue Addressed

The naming of the functions in lighthouse now match the spec names:
* [get_validator_churn_limit()](https://github.com/ethereum/consensus-specs/blob/v1.4.0/specs/phase0/beacon-chain.md#get_validator_churn_limit)
* [apply_deposit()](https://github.com/ethereum/consensus-specs/blob/v1.4.0/specs/phase0/beacon-chain.md#deposits)

which will make implementing Electra much cleaner.


realbigsean (approved):

realbigsean:
@mergify queue

PR #5598, opened 2024-04-18
Fix on-disk consensus context format

michaelsproul:
## Issue Addressed

Fix a regression in `unstable` that causes logs of the form:

> ERRO Failed to maintain availability cache error: DecodeError(OffsetOutOfBounds(1157628160)), service: beacon

This was due to the addition of fields to `ConsensusContext` in the single-pass epoch processing PR. `ConsensusContext` is now part of our on-disk database schema via `DietAvailabilityPendingExecutedBlock`:

https://github.com/sigp/lighthouse/blob/3058b96f2560f1da04ada4f9d8ba8e5651794ff6/beacon_node/beacon_chain/src/data_availability_checker/state_lru_cache.rs#L20-L30

## Proposed Changes

- Define a new `OnDiskConsensusContext` struct which keeps only the previous fields.
- Convert between in-memory consensus contexts and the stable on-disk format as necessary.

### DIFF
--- a/beacon_node/beacon_chain/src/data_availability_checker/state_lru_cache.rs
+++ b/beacon_node/beacon_chain/src/data_availability_checker/state_lru_cache.rs
@@ -8,8 +8,9 @@ use crate::{
 use lru::LruCache;
 use parking_lot::RwLock;
 use ssz_derive::{Decode, Encode};
-use state_processing::{BlockReplayer, ConsensusContext, StateProcessingStrategy};
+use state_processing::{BlockReplayer, StateProcessingStrategy};
 use std::sync::Arc;
+use store::OnDiskConsensusContext;
 use types::beacon_block_body::KzgCommitments;
 use types::{ssz_tagged_signed_beacon_block, ssz_tagged_signed_beacon_block_arc};
 use types::{BeaconState, BlindedPayload, ChainSpec, Epoch, EthSpec, Hash256, SignedBeaconBlock};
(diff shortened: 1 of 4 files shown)


beacon_node/beacon_chain/src/data_availability_checker/state_lru_cache.rs
```diff
@@ -26,7 +27,7 @@ pub struct DietAvailabilityPendingExecutedBlock<E: EthSpec> {
     parent_block: SignedBeaconBlock<E, BlindedPayload<E>>,
     parent_eth1_finalization_data: Eth1FinalizationData,
     confirmed_state_roots: Vec<Hash256>,
-    consensus_context: ConsensusContext<E>,
+    consensus_context: OnDiskConsensusContext,
```
realbigsean: `DietAvailabilityPendingExecutedBlock` is constructued on every block, and since `OnDiskConsensusContext` drops `indexed_attestations`, that cache would be dropped every block.

Could fix this easily by keeping the `indexed_attestations` field in `OnDiskConsensusContext` with `#[ssz(skip_serializing, skip_deserializing)]`.

realbigsean (changes requested):

beacon_node/beacon_chain/src/data_availability_checker/state_lru_cache.rs
michaelsproul: Great catch, just pushed a fix

realbigsean (approved):
LGTM

michaelsproul:
@mergifyio queue

Issue #5605, opened 2024-04-18
Implement Electra Fork


ethDreamer:
This issue will track the tasks for implementing the [Electra upgrade](https://github.com/ethereum/consensus-specs/tree/dev/specs/electra)

We propose doing things differently in this fork. Instead of working off of a dedicated branch and having one big review and merge at the end, all PRs will target `unstable` directly and we will review and merge changes in one PR at a time. This means lighthouse **must** fully pass CI for each of these changes. We will not be able to turn on the Electra spec tests until Electra is fully implemented.

- [x] [devnet-1](https://github.com/ethereum/consensus-specs/tree/v1.5.0-alpha.3/specs/electra) (https://github.com/sigp/lighthouse/pull/5937)
  - [x] Move signature field in Attestation (https://github.com/sigp/lighthouse/pull/5924)
  - [x] Rename `Constants`, `Presets`, and `Containers`(https://github.com/sigp/lighthouse/pull/5974)
  - [x] Move `Consolidations` from `BeaconBlock` to `ExecutionPayload` & rename fields (https://github.com/sigp/lighthouse/pull/5974)
  - [x] Update `compute_proposer_index()` (https://github.com/sigp/lighthouse/pull/5983)
  - [x] Modified `get_next_sync_committee_indices()` (https://github.com/sigp/lighthouse/pull/5983)
  - [x] New changes for `process_pending_balance_deposits()` (https://github.com/sigp/lighthouse/pull/6005)
  - [x] All changes in `process_operations()` (https://github.com/sigp/lighthouse/pull/5994)
  - [x] Engine API Changes (https://github.com/sigp/lighthouse/pull/6001)
  - [x] Get ef-tests passing (https://github.com/sigp/lighthouse/pull/6019)
  - [x] Stop polling `Eth1Data` from EL post-electra
- [x] [devnet-0](https://github.com/ethereum/consensus-specs/tree/v1.5.0-alpha.0/specs/electra)
  - [x] Add Electra Fork boiler plate (https://github.com/sigp/lighthouse/pull/5122)
[...]

PR #5606, opened 2024-04-18
Electra: Add Preset, Constants, & Config

ethDreamer:
## Issue Addressed

* #5605 

## Proposed Changes

This PR adds the [Constants](https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.0/specs/electra/beacon-chain.md#constants), [Preset](https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.0/specs/electra/beacon-chain.md#preset) & [Configuration](https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.0/specs/electra/beacon-chain.md#configuration) changes for Electra

### DIFF
--- a/consensus/types/src/chain_spec.rs
+++ b/consensus/types/src/chain_spec.rs
@@ -25,6 +25,7 @@ pub enum Domain {
     SyncCommittee,
     ContributionAndProof,
     SyncCommitteeSelectionProof,
+    Consolidation,
     ApplicationMask(ApplicationDomain),
 }
 
@@ -76,6 +77,7 @@ pub struct ChainSpec {
     pub genesis_fork_version: [u8; 4],
     pub bls_withdrawal_prefix_byte: u8,
     pub eth1_address_withdrawal_prefix_byte: u8,
+    pub compounding_withdrawal_prefix_byte: u8,
 
     /*
      * Time parameters
@@ -108,6 +110,7 @@ pub struct ChainSpec {
     pub(crate) domain_voluntary_exit: u32,
     pub(crate) domain_selection_proof: u32,
     pub(crate) domain_aggregate_and_proof: u32,
+    pub(crate) domain_consolidation: u32,
 
     /*
      * Fork choice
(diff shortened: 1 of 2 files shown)


consensus/types/src/chain_spec.rs
```diff
@@ -709,6 +726,30 @@ impl ChainSpec {
              */
             electra_fork_version: [0x05, 00, 00, 00],
             electra_fork_epoch: None,
+            unset_deposit_receipts_start_index: u64::MAX,
+            full_exit_request_amount: 0,
+            min_activation_balance: option_wrapper(|| {
+                u64::checked_pow(2, 5)?.checked_mul(u64::checked_pow(10, 9)?)
```
dapplion, line 732: Why not just write `32000000000` here? What value does it add to write it in a factored formula?

ethDreamer: mostly consistency with the way the other values are written.. the spec also provides them factored this way so that you can do quick calculations in your head if you want

consensus/types/src/chain_spec.rs
```diff
@@ -177,6 +180,15 @@ pub struct ChainSpec {
     pub electra_fork_version: [u8; 4],
     /// The Electra fork epoch is optional, with `None` representing "Electra never happens".
     pub electra_fork_epoch: Option<Epoch>,
+    pub unset_deposit_receipts_start_index: u64,
+    pub full_exit_request_amount: u64,
```
realbigsean, line 184: These two aren't configurable, so I'm not sure we should add them here. We aren't totally consistent with `ChainSpec`, because the signature domains and prefixes are on this struct. But it should be config + preset from the spec. We have a `consensus/types/consts.rs` file we could move them to.

realbigsean:
@mergify queue

realbigsean (approved):

realbigsean:
@mergify refresh

consensus/types/src/chain_spec.rs
realbigsean: moved to issue https://github.com/sigp/lighthouse/issues/5631

PR #5611, opened 2024-04-21
Delete repetitive execute command in local testnet scripts

chong-he:
This is a very minor bug in the local-testnet script where the `el_bootnode.sh` is executed twice, causing the outcome that when `stop_local_testnet.sh` is called, the terminal will show: 

`./kill_processes.sh: line 15: kill: (2254405) - No such process`

It doesn't really affect anything though, but it would be nice to delete this so that no error is shown when the testnet is stopped.


michaelsproul (approved):

realbigsean:
@mergify queue

PR #5489, opened 2024-03-27
Use `portable` builds by default 

antondlr:
## Issue Addressed

Heavily reduces strain on CI infra during docker builds and releases 
(docker builds: 9 -> 4* ; release builds 8 -> 4) 
See #5457

## Proposed Changes

Make `blst-portable` the default across the board and stop offering `force-adx` builds

## Additional Info

In this PR, all current docker tags and direct release links** are preserved but lead to a `portable` build instead, which should work for everyone. 

The idea is to release with these placeholders to try and assess how many downloads we still get on the to-be-deprecated release links.

Before we consider this, we should make absolutely sure that the performance hit, if present, is negligible on modern hardware. 

(* 8 lighthouse and 1 lcli build --> 2 lighthouse and 2 lcli builds)
(** excluding the link for a non-portable windows build)

### Summary of changes

- all lighthouse builds (both binaries and docker) use the portable `blst` library
- `lcli` is built through `cross`, and published as multiplatform image
- regular docker images now include `spec-minimal`

Issue #5457, opened 2024-03-21
Get rid of historic bloat on the build pipeline


antondlr:
## Description

We currently do a lot of different builds (esp. on the docker front) which slows down CI pipelines without any clear benefit.
Additionally, we publish a **lot** of different docker image tags as part of the build pipeline. 

Tangentially related is [this issue](https://github.com/sigp/lighthouse/issues/4777) about the docker pipeline being started more often than necessary, resulting in even more CI overhead. 

## Present Behaviour

Release binaries are typically built with `portable` and `modern` blst config, which has been redundant [since august 2023](https://github.com/sigp/lighthouse/pull/4624#issue-1852388232), when we introduced runtime cpu feature detection to the portable builds. 

On the docker front, another flavour in built for every spec (-`dev`), which only includes `spec-minimal`

This results on these builds being released:
```
Binaries:
- lighthouse-v5.1.1-x86_64-apple-darwin.tar.gz
- lighthouse-v5.1.1-x86_64-apple-darwin-portable.tar.gz
- lighthouse-v5.1.1-x86_64-unknown-linux-gnu.tar.gz
- lighthouse-v5.1.1-x86_64-unknown-linux-gnu-portable.tar.gz
- lighthouse-v5.1.1-aarch64-unknown-linux-gnu.tar.gz
- lighthouse-v5.1.1-aarch64-unknown-linux-gnu-portable.tar.gz
- lighthouse-v5.1.1-x86_64-windows.tar.gz
- lighthouse-v5.1.1-x86_64-windows-portable.tar.gz

Docker: 
- v5.1.1
- v5.1.1-modern
- v5.1.1-amd64
- v5.1.1-amd64-dev
- v5.1.1-amd64-modern
- v5.1.1-amd64-modern-dev
- v5.1.1-arm64
- v5.1.1-arm64-dev
- v5.1.1-arm64-modern
- v5.1.1-arm64-modern-dev
```

## Expected Behaviour

- make `portable` the default and get rid of `portable` and `modern` tags, as this should always give you (very close to?) optimal performance, with fallback in place for older systems. 
- add the `spec-minimal` profile to regular docker builds and get rid of `-dev`
- stop publishing architecture-specific docker images, only publish proper multiarch images
- (build and add an aarch64 `lcli` image to the `lcli` image manifest)

The releases would thus be cut down to:

```
Binaries:
- lighthouse-v5.1.1-x86_64-apple-darwin.tar.gz
[...]

PR #5614, opened 2024-04-22
Only `portable` builds (docker)

antondlr:
See #5489 for details, this is a cleaner version of the docker part of that PR

For backwards compatibility, we still publish all the different "specialised" (i.e., `-dev`) docker image tags (pointing to one image). This is meant to be phased out as we direct our users towards the regular image tag.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -15,8 +15,6 @@ concurrency:
 env:
     DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
     DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
-    IMAGE_NAME: ${{ github.repository_owner}}/lighthouse
-    LCLI_IMAGE_NAME: ${{ github.repository_owner }}/lcli
     # Enable self-hosted runners for the sigp repo only.
     SELF_HOSTED_RUNNERS: ${{ github.repository == 'sigp/lighthouse' }}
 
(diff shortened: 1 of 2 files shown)


antondlr:
fyi the published docker image would no longer include `slasher-mdbx` with these proposed changes, I guess that means a breaking change (or re-adding it here) ?

realbigsean (approved):

realbigsean:
> `slasher-mdbx`

is there a reason for dropping it? I have no idea if it's used, or if we want to keep experimenting with it

antondlr:
not really, my takeaway after reading [this](https://lighthouse-book.sigmaprime.io/slasher.html#database-backend) was that it was in silent deprecation but I'm probably totally wrong 
let's consult the oracle to make sure @michaelsproul

realbigsean:
> not really, my takeaway after reading [this](https://lighthouse-book.sigmaprime.io/slasher.html#database-backend) was that it was in silent deprecation but I'm probably totally wrong

Reading that makes me think it should be alright to stop publishing it with the standard image. Would you mind pulling in `unstable` to get CI passing?

If this is blocking people we can merge and follow up with a PR to re-add `slasher-mdbx` if our oracle guides us a different direction

sebastianst:
Are you gonna publish new multi-arch docker images right after this is merged? ([context](https://github.com/sigp/lighthouse/issues/5457#issuecomment-2069866796))

antondlr:
> Are you gonna publish new multi-arch docker images right after this is merged? ([context](https://github.com/sigp/lighthouse/issues/5457#issuecomment-2069866796))

a `latest-unstable` multiarch image will be automatically published after successful merge!

realbigsean:
@mergify queue

PR #5615, opened 2024-04-22
Only `portable` builds (binaries)

antondlr:
See #5489 for details, this is a cleaner version of the release/binaries part of that PR

For backwards compatibility, we still publish a copy of the binaries with the `portable`name. This is meant to be phased out as we direct our users towards the regular (now, portable) release.


.github/workflows/release.yml
```diff
@@ -166,6 +131,7 @@ jobs:
                   tar -czf lighthouse-${{ needs.extract-version.outputs.VERSION }}-${{ matrix.arch }}.tar.gz lighthouse.exe
                   gpg --passphrase "$env:GPG_PASSPHRASE" --batch --pinentry-mode loopback -ab lighthouse-${{ needs.extract-version.outputs.VERSION }}-${{ matrix.arch }}.tar.gz
                   move *tar.gz* ..
+                  # todo copy to `-portable`
```
michaelsproul: I think this comment can be deleted (already done above)

michaelsproul (approved):
approved modulo removing that comment

.github/workflows/release.yml
michaelsproul:
```suggestion
```

michaelsproul:
@mergifyio queue

realbigsean:
@mergify queue

realbigsean:
@mergify queue

realbigsean:
@mergify requeue

PR #5561, opened 2024-04-11
Sync lookup dedup range and blobs

dapplion:
## Issue Addressed

Part of 
- https://github.com/sigp/lighthouse/issues/5549

## Proposed Changes

Merges BackFillBlocks, BackFillBlockAndBlobs, RangeBlocks, RangeBlockAndBlobs handling. All those requests return the same type of data, a Vec<RpcBlock>, which can be handled with an enum request id to differentiate RangeSync from BackfillSync.

Pre-deneb requests of only blocks and post-deneb requests of blocks and blobs can be handled together by tracking if a blobs request has been sent.

This PR does not change any logic or behaviour of either sync, nor their tests. @realbigsean please look into this line change https://github.com/sigp/lighthouse/pull/5561/files#diff-3d6c8081703730f63787240a11f38fdde1d721397245cf316207b108f1df3defR962 to ensure it does not regress this fix https://github.com/sigp/lighthouse/pull/4869

PR #5616, opened 2024-04-22
DAS sampling on sync

dapplion:
## Issue Addressed

Implements the foundational networking infrastructure / wiring for sampling.

Consumers will trigger sampling sending a `SyncMessage::SampleBlock(Hash256)`.  Sampling requests are tracked independently on `SyncManager::sampling`. Each `ActiveSamplingRequest` manages multiple individual sample column requests. A `ActiveSamplingRequest` manages retries and may extend the count of sample columns following LossyDAS. When a sampling requests has enough required successes, it's complete.

SyncManager will forward the sampling result (a single bit Ok/Not) to the chain who should update the corresponding fork-choice node with trailing DA.

**Range sync**

Range sync may explicitly wait for the block to be sampled before sending it for import, or sampling could trail behind imported blocks. This PR can easily accommodate both, by adding a new enum variant `SamplingRequester::RangeSync`.


beacon_node/beacon_processor/src/lib.rs
```diff
@@ -833,6 +843,8 @@ impl<E: EthSpec> BeaconProcessor<E> {
         // Using a FIFO queue since blocks need to be imported sequentially.
         let mut rpc_block_queue = FifoQueue::new(MAX_RPC_BLOCK_QUEUE_LEN);
         let mut rpc_blob_queue = FifoQueue::new(MAX_RPC_BLOB_QUEUE_LEN);
+        let mut rpc_verify_data_column_queue = FifoQueue::new(MAX_RPC_VERIFY_DATA_COLUMN_QUEUE_LEN);
+        let mut sampling_result_queue = FifoQueue::new(MAX_SAMPLING_RESULT_QUEUE_LEN);
```
jimmygchen, line 847: Looks like we never read / `pop` from these queues. We'll need to add them under this section:
https://github.com/sigp/lighthouse/blob/9de3178c7ffe6939210c643e28ec35d154d0c679/beacon_node/beacon_processor/src/lib.rs#L992

jimmygchen: Also worth adding the queue metrics so we don't forget:
https://github.com/sigp/lighthouse/blob/9de3178c7ffe6939210c643e28ec35d154d0c679/beacon_node/beacon_processor/src/lib.rs#L1420-L1423

beacon_node/lighthouse_network/src/rpc/methods.rs
```diff
@@ -378,6 +378,14 @@ pub struct DataColumnsByRootRequest {
     pub data_column_ids: RuntimeVariableList<DataColumnIdentifier>,
 }
 
+impl DataColumnsByRootRequest {
+    pub fn new(blob_ids: Vec<DataColumnIdentifier>, spec: &ChainSpec) -> Self {
```
jimmygchen:
```suggestion
    pub fn new(data_column_ids: Vec<DataColumnIdentifier>, spec: &ChainSpec) -> Self {
```

beacon_node/network/src/router.rs
```diff
@@ -514,11 +514,11 @@ impl<T: BeaconChainTypes> Router<T> {
     ) {
         let request_id = match request_id {
             RequestId::Sync(sync_id) => match sync_id {
-                SyncId::SingleBlock { .. } | SyncId::SingleBlob { .. } => {
-                    crit!(self.log, "Block lookups do not request BBRange requests"; "peer_id" => %peer_id);
+                id @ SyncId::RangeBlockAndBlobs { .. } => id,
+                other => {
+                    crit!(self.log, "BlocksByRange response on incorrect request"; "request" => ?other);
```
jimmygchen, line 519: Maybe worth keeping the `peer_id` logging here?

beacon_node/network/src/router.rs
```diff
@@ -577,12 +577,8 @@ impl<T: BeaconChainTypes> Router<T> {
...
             RequestId::Sync(sync_id) => match sync_id {
                 id @ SyncId::SingleBlock { .. } => id,
-                SyncId::RangeBlockAndBlobs { .. } => {
-                    crit!(self.log, "Batch syncing do not request BBRoot requests"; "peer_id" => %peer_id);
-                    return;
-                }
-                SyncId::SingleBlob { .. } => {
-                    crit!(self.log, "Blob response to block by roots request"; "peer_id" => %peer_id);
+                other => {
+                    crit!(self.log, "BlocksByRoot response on incorrect request"; "request" => ?other);
```
jimmygchen, line 581: Same as above, should we keep `peer_id`?

beacon_node/network/src/sync/manager.rs
```diff
@@ -116,6 +125,13 @@ pub enum SyncMessage<E: EthSpec> {
         seen_timestamp: Duration,
     },
 
+    RpcDataColumn {
```
jimmygchen, line 129: Please add some doc here

beacon_node/network/src/sync/manager.rs
```diff
@@ -326,6 +362,9 @@ impl<T: BeaconChainTypes> SyncManager<T> {
             RequestId::SingleBlob { id } => {
                 self.on_single_blob_response(id, peer_id, RpcEvent::RPCError(error))
             }
+            RequestId::DataColumnsByRoot(id) => {
```
jimmygchen, line 366: Perhaps rename this to `SingleDataColumn` for consistency?

beacon_node/network/src/sync/manager.rs
```diff
@@ -895,6 +954,36 @@ impl<T: BeaconChainTypes> SyncManager<T> {
...
+                self.on_single_data_column_response(
+                    id,
+                    peer_id,
+                    match data_column {
+                        Some(data_column) => RpcEvent::Response(data_column, seen_timestamp),
+                        None => RpcEvent::StreamTermination,
+                    },
+                );
+            }
+            RequestId::RangeBlockAndBlobs { id } => {
```
jimmygchen, line 985: We should handle `RangeBlockAndDataColumns` here instead right?

beacon_node/network/src/sync/manager.rs
```diff
@@ -949,6 +1038,68 @@ impl<T: BeaconChainTypes> SyncManager<T> {
...
+        id: Id,
+        peer_id: PeerId,
+        data_column: RpcEvent<Arc<DataColumnSidecar<T::EthSpec>>>,
+    ) {
+        let Some((requester, resp)) = self
+            .network
+            .on_data_columns_by_root_response(id, data_column)
+        else {
+            // TOOD(das): error o log
+            return;
```
jimmygchen: Are you still thinking to add a log here? It looks like we don't log anything if block or blob requests aren't found.

beacon_node/network/src/sync/network_context.rs
```diff
@@ -142,13 +151,41 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
     }
 
+    // TODO(das): epoch argument left here in case custody rotation is implemented
+    pub fn get_custodial_peers(&self, _epoch: Epoch, column_index: ColumnIndex) -> Vec<PeerId> {
+        let mut peer_ids = vec![];
+
+        for (peer_id, peer_info) in self.network_globals().peers.read().peers() {
+            if let Some(enr) = peer_info.enr() {
+                // TODO(das): do not hardcode `custody_subnet_count`
+                let custody_subnet_count = 2;
```
jimmygchen: Should be able to use `enr.custody_subnet_count()` for this now.

https://github.com/sigp/lighthouse/blob/d5e7e73abc64312d7a6f5ce3a54b0df455e2b243/beacon_node/lighthouse_network/src/discovery/enr.rs#L67

beacon_node/network/src/sync/network_context.rs
```diff
@@ -482,19 +575,20 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
                 request.remove();
                 Err(e.into())
             }
-        })
+        };
+        Some(((), resp))
```
jimmygchen, line 583: It feels a bit confusing that the tuple order (ID, Resp) is the opposite to the generic parameters (Resp, ID)

beacon_node/network/src/sync/sampling.rs
```diff
@@ -0,0 +1,486 @@
...
+
+#[derive(Debug, Hash, PartialEq, Eq, Clone, Copy)]
+pub struct SamplingId {
+    pub id: SamplingRequester,
+    pub column_index: ColumnIndex,
+}
+
+#[derive(Debug, Hash, PartialEq, Eq, Clone, Copy)]
+pub enum SamplingRequester {
+    ImportedBlock(Hash256),
```
jimmygchen, line 25: Does this mean we're only triggering sampling after importing a block? What other variants are you thinking to add?

beacon_node/network/src/sync/sampling.rs
```diff
@@ -0,0 +1,486 @@
...
+impl<T: BeaconChainTypes> ActiveSamplingRequest<T> {
+    fn new(
+        block_root: Hash256,
+        block_slot: Slot,
+        requester_id: SamplingRequester,
+        sampling_config: &SamplingConfig,
+        log: slog::Logger,
+    ) -> Self {
+        // Select ahead of time the full list of to-sample columns
+        let mut column_shuffle = (0..64).collect::<Vec<ColumnIndex>>();
```
jimmygchen: Should we be using `EthSpec::number_of_columns()` instead? (currently `128`)

jimmygchen:
Could you merge latest `das` onto your branch? I fixed a merge conflict issue and it should now trigger the test suite again.

beacon_node/network/src/sync/sampling.rs
```diff
@@ -0,0 +1,486 @@
...
+            return Ok(Some(()));
+        }
+
+        let mut sent_requests = 0;
+
+        // First, attempt to progress sampling by requesting more columns, so that request failures
+        // are accounted for below.
+        for idx in 0..*required_successes {
+            // Re-request columns
+            let column_index = self.column_shuffle[idx];
```
jimmygchen: lint would probably complain about this, we should use `get` here to prevent index out of bounds.

beacon_node/network/src/sync/sampling.rs
```diff
@@ -0,0 +1,486 @@
...
+        requester_id: SamplingRequester,
+        sampling_config: &SamplingConfig,
+        log: slog::Logger,
+    ) -> Self {
+        // Select ahead of time the full list of to-sample columns
+        let mut column_shuffle = (0..64).collect::<Vec<ColumnIndex>>();
+        let mut rng = thread_rng();
+        column_shuffle.shuffle(&mut rng);
+
+        Self {
```
jimmygchen, line 211: is it worth adding an extra check that `required_successes` is always less than the length of `column_shuffle`?

beacon_node/network/src/sync/network_context.rs
```diff
@@ -142,13 +151,41 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
             chain,
             log,
         }
     }
 
+    // TODO(das): epoch argument left here in case custody rotation is implemented
+    pub fn get_custodial_peers(&self, _epoch: Epoch, column_index: ColumnIndex) -> Vec<PeerId> {
+        let mut peer_ids = vec![];
+
+        for (peer_id, peer_info) in self.network_globals().peers.read().peers() {
```
jimmygchen: `peers` returns all known peers in the `PeerDB`. Maybe `connected_peers` should be used here instead?

beacon_node/network/src/sync/sampling.rs
```diff
@@ -0,0 +1,486 @@
...
+
+    // TODO: When is a fork and only a subset of your peers know about a block, sampling should only
+    // be queried on the peers on that fork. Should this case be handled? How to handle it?
+    fn on_sample_downloaded(
+        &mut self,
+        _peer_id: PeerId,
+        column_index: ColumnIndex,
+        resp: Result<(DataColumnSidecarList<T::EthSpec>, Duration), LookupFailure>,
+        cx: &mut SyncNetworkContext<T>,
+    ) -> Result<Option<()>, SamplingError> {
```
jimmygchen, line 240: Can we add some docs for the return value? as `Option<()` isn't very descriptive.

beacon_node/network/src/sync/sampling.rs
```diff
@@ -0,0 +1,486 @@
...
+                    debug!(self.log, "Sampling peer claims to not have the data"; "block_root" => %self.block_root, "column_index" => column_index);
+                    request.on_sampling_error()?;
+                }
+            }
+            Err(err) => {
+                debug!(self.log, "Sample download error"; "block_root" => %self.block_root, "column_index" => column_index, "error" => %err);
+
+                // Error downloading, maybe penalize peer and retry again.
+                // TODO(das) with different peer or different peer?
+                request.on_sampling_error()?;
```
jimmygchen, line 296: Do we currently or plan to keep track of the peers that fail to return sample results somewhere and penalise if it's recurring?

beacon_node/network/src/sync/sampling.rs
```diff
@@ -0,0 +1,486 @@
...
+    }
+
+    fn handle_sampling_result(
+        &mut self,
+        result: Result<Option<()>, SamplingError>,
+        id: &SamplingRequester,
+    ) -> Option<SamplingResult> {
+        let result = result.transpose();
+        if result.is_some() {
+            debug!(self.log, "Removed sampling request"; "id" => ?id);
```
jimmygchen: Maybe some reasons on why its removed would be useful, e.g. "Remove completed sampling request" (completed means success/failed here)

jimmygchen:
Hi @dapplion 

The overall structure of looks really good and simple, I like it! 

From my understanding, this PR does not implement triggering of sampling block`SyncMessage::SampleBlock` yet. What's your thoughts about this?  We could potentially trigger this from `BeaconProcessor` as soon as we receive the first verified block component?

I've added some comments above, and additionally I think it would be useful to:
- Add some docs to the pub methods of `ActiveSamplingRequest`
- Add some metrics to track sampling, maybe:
  - sample request attempts
  - sample request success
  - sample request failures
  - active sampling requests

I don't think we need to address all of my comments in the PR, happy to merge this PR as soon as it's in a good shape for merge, and we've tracked the extra tasks in the tracking issue.

Nice work! 🔥

jimmygchen (changes requested):
Added my review above.

Some diagrams I've created to help my understanding, may helps other reviewers:


### Initiating sampling request

```mermaid
sequenceDiagram
    Note over NetworkBeaconProcessor: On slot OR component received
    NetworkBeaconProcessor->>SyncManager: `SampleBlock(Hash256, Slot)` message
    SyncManager->>Sampling: new sample request
    Note over Sampling: create `ActiveSamplingRequest` for block root 
    Note over Sampling: sends request via `SyncNetworkContext`
```

### Handling sampling RPC response

```mermaid
sequenceDiagram
    Note over SyncManager: On RPC column received
    SyncManager->>Sampling: downloaded column response
    Sampling->>BeaconProcessor: send for verification (`Work::RpcVerifyDataColumn`)
    Note over BeaconProcessor: validate data columns
    BeaconProcessor->>SyncManager: send `SampleVerified` message
    SyncManager->>Sampling: send result for sampling status re-evaluation
    Sampling-->>SyncManager: returns sampling result
    SyncManager->>BeaconProcessor: Send sampling completed (`Work::SamplingResult`)
    BeaconProcessor->>BeaconChain: Update fork choice (Unimplemented)
```

beacon_node/network/src/router.rs
dapplion: I don't think so, an error is an implementation / internal error on our end. There's nothing a peer can do to cause this error

beacon_node/network/src/sync/manager.rs
dapplion: So `SingleBlob` blob does not really mean 1 blob, it means a blob for a SingleLookup. Here data columns are not strictly part of a SingleLookup as range sync will also consume then. I can do a PR latter to align `SingleBlob` into `BlobByRoot` which is also correct, but using `SingleDataColumn` would be a missnomer.

beacon_node/network/src/sync/manager.rs
dapplion: I think we should rename `RangeBlockAndBlobs` to `RangeBlockComponents` or `RangeBlockAndData` to be represent the set of sidecars the block needs. There's no need to have a separate enum variant. See
- https://github.com/sigp/lighthouse/pull/5561

beacon_node/network/src/sync/manager.rs
dapplion: Fixed, None is an expected value and should not have a dedicated log

beacon_node/network/src/sync/sampling.rs
dapplion: We could trigger sampling as soon as we see the block over gossip for a not yet imported block. Also if we are doing same slot sampling.

beacon_node/network/src/sync/sampling.rs
dapplion: I don't think so, the list of successes is hardcoded by us. It is only dynamic to control testing which we also control. This way we don't need to make the constructor fallible.

beacon_node/network/src/sync/sampling.rs
dapplion: Scoring in sampling is a complicated topic. There's no clear attribution if the proposer is playing delay games to force donwscoring events or the peers are being actively malicious. For now I would just do nothing and figure it out latter in the interop

dapplion:
> From my understanding, this PR does not implement triggering of sampling blockSyncMessage::SampleBlock yet. What's your thoughts about this? We could potentially trigger this from BeaconProcessor as soon as we receive the first verified block component?

The easiest now is to wait for the block to make sure if has > 0 blobs.

> Add some docs to the pub methods of ActiveSamplingRequest

Done

> Add some metrics to track sampling

WIP

jimmygchen:
> The easiest now is to wait for the block to make sure if has > 0 blobs.

If we receive a `DataColumnSidecar` which would also have list of kzg commitments, would that not be sufficient? or is there any security concerns around triggering sampling before we get a block?

beacon_node/beacon_processor/src/lib.rs
```diff
@@ -988,6 +1000,11 @@ impl<E: EthSpec> BeaconProcessor<E> {
                             self.spawn_worker(item, idle_tx);
                         } else if let Some(item) = rpc_blob_queue.pop() {
                             self.spawn_worker(item, idle_tx);
+                        // TODO(das): decide proper priorization for sampling columns
+                        } else if let Some(item) = rpc_verify_data_column_queue.pop() {
+                            self.spawn_worker(item, idle_tx);
+                        } else if let Some(item) = sampling_result_queue.pop() {
+                            self.spawn_worker(item, idle_tx);
```
jimmygchen, line 1007: I think this is good since we prioritise importing rpc blocks over everything else 👍

beacon_node/beacon_processor/src/lib.rs
```diff
@@ -1371,6 +1394,14 @@ impl<E: EthSpec> BeaconProcessor<E> {
                     &metrics::BEACON_PROCESSOR_RPC_BLOB_QUEUE_TOTAL,
                     rpc_blob_queue.len() as i64,
                 );
+                metrics::set_gauge(
+                    &metrics::BEACON_PROCESSOR_RPC_VERIFY_DATA_COLUMN_QUEUE_TOTAL,
+                    rpc_blob_queue.len() as i64,
```
jimmygchen:
```suggestion
                    rpc_verify_data_column_queue.len() as i64,
```

beacon_node/beacon_processor/src/lib.rs
```diff
@@ -1371,6 +1394,14 @@ impl<E: EthSpec> BeaconProcessor<E> {
...
+                );
+                metrics::set_gauge(
+                    &metrics::BEACON_PROCESSOR_SAMPLING_RESULT_QUEUE_TOTAL,
+                    rpc_blob_queue.len() as i64,
```
jimmygchen:
```suggestion
                    sampling_result_queue.len() as i64,
```

beacon_node/beacon_processor/src/metrics.rs
```diff
@@ -86,6 +86,16 @@ lazy_static::lazy_static! {
...
         "Count of blobs from the rpc waiting to be verified."
     );
+    // Rpc verify data columns
+    pub static ref BEACON_PROCESSOR_RPC_VERIFY_DATA_COLUMN_QUEUE_TOTAL: Result<IntGauge> = try_create_int_gauge(
+        "beacon_processor_rpc_verify_data_column_queue_total",
+        "Count of data columns from the rpc waiting to be verified."
+    );
+    // Sampling result
+    pub static ref BEACON_PROCESSOR_SAMPLING_RESULT_QUEUE_TOTAL: Result<IntGauge> = try_create_int_gauge(
+        "beacon_processor_rpc_blob_queue_total",
```
jimmygchen:
```suggestion
        "beacon_processor_sampling_result_queue_total",
```

dapplion:
> If we receive a DataColumnSidecar which would also have list of kzg commitments, would that not be sufficient? or is there any security concerns around triggering sampling before we get a block?

Touche, not more complains from me

beacon_node/network/src/sync/network_context.rs
```diff
@@ -142,13 +153,43 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+    // TODO(das): epoch argument left here in case custody rotation is implemented
+    pub fn get_custodial_peers(&self, _epoch: Epoch, column_index: ColumnIndex) -> Vec<PeerId> {
+        let mut peer_ids = vec![];
+
+        for (peer_id, peer_info) in self.network_globals().peers.read().connected_peers() {
+            if let Some(enr) = peer_info.enr() {
+                // TODO(das): ignores decode errors
+                let custody_subnet_count = enr
+                    .custody_subnet_count::<T::EthSpec>()
+                    .unwrap_or(T::EthSpec::min_custody_requirement() as u64);
```
jimmygchen, line 173: The method already fallsback to min custody requirement. An error here means there's an issue decoding ENR, I'm not sure if we should fallback to min custody requirement here? I think we should at least log an error here to avoid a hard to find bug.

I think we could probably fallback to `min_custody_requirment` though if ENR fails to decode, providing that we do the same everywhere else.

dapplion:
@jimmygchen Added metrics on https://github.com/sigp/lighthouse/pull/5616/commits/f8acdac4c4dc4cb54ebaa5d3871302f0c78bd6c6 bundled with remote marge by mistake

beacon_node/network/src/sync/sampling.rs
```diff
@@ -0,0 +1,538 @@
...
+            cx: &mut SyncNetworkContext<T>,
+        ) -> Result<bool, SamplingError> {
+            match &self.status {
+                Status::NoPeers | Status::NotStarted => {} // Ok to continue
+                Status::Sampling(_) => return Ok(false),   // Already downloading
+                Status::Verified => return Ok(false),      // Already completed
+            }
+
+            // TODO: When is a fork and only a subset of your peers know about a block, sampling should only
+            // be queried on the peers on that fork. Should this case be handled? How to handle it?
```
jimmygchen, line 490: Maybe we could use status messages, similar to how we do it in range_sync - happy to leave this out of this current PR though

https://github.com/sigp/lighthouse/blob/23671531bcb2c8050888b51b80f1be2310384037/beacon_node/network/src/sync/manager.rs#L345-L351

dapplion: It's still tricky because peers can have slightly different heads on the same fork such that they _should_ be grouped together but they will appear in different forks

jimmygchen: Yep let's think about this - going to resolve this and we'll get back to this TODO later

beacon_node/network/src/sync/network_context.rs
jimmygchen: Had a chat with Lion offline and it makes sense to ignore peer ENR decoding error:

> RE ENR decode, can decoding fail due to the other party creating a bad ENR? In that case we should ignore the value

beacon_node/network/src/sync/sampling.rs
jimmygchen: Sounds good

jimmygchen:
Nice, we're close to merging this! Just the last few metrics comments and a green `network-tests` and we're good to merge!

Ignore the `execution-engine-integtration` as it's been fixed on `unstable` - will merge `unstable` into `das` after this PR is merged.

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -2855,6 +2855,15 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         ChainSegmentResult::Successful { imported_blocks }
     }
 
+    /// Updates fork-choice node into a permanent `available` state so it can become a viable head.
+    /// Only completed sampling results are received. Blocks are unavailable by default and should
+    /// be pruned on finalization, on a timeout or by a max count.
+    pub async fn process_sampling_completed(self: &Arc<Self>, block_root: Hash256) {
+        // TODO(das): update fork-choice
```
jimmygchen, line 2862: I think it would be possible that sampling is completed before block is imported into fork choice - we'll need to remember handling this scenario.

beacon_node/network/src/sync/sampling.rs
```diff
@@ -71,7 +71,12 @@ impl<T: BeaconChainTypes> Sampling<T> {
                 self.log.clone(),
             )),
             Entry::Occupied(_) => {
-                warn!(self.log, "Ignoring duplicate sampling request"; "id" => ?id);
+                // Sampling is triggered from multiple sources, duplicate sampling requests are
+                // likely (gossip block + gossip data column)
+                // TODO(das): Should track failed sampling request for some time? Otherwise there's
+                // a risk of a loop with multiple triggers creating the request, then failing,
```
jimmygchen, line 77: I think the main one we want to avoid retrying is `TooManyFailures` right? so it doesn't reset the required success to the min number? ([`LossyDAS`](https://ethresear.ch/t/lossydas-lossy-incremental-and-diagonal-sampling-for-data-availability/18963#incrementaldas-dynamically-increase-the-sample-size-10))

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -635,10 +635,21 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                // We have observed a data column sidecar with valid inclusion proof, such that
+                // `block_root` must have data. The column may or not be imported yet.
+                // TODO(das): Sampling should check that sampling is not completed already.
+                //
+                // Trigger sampling early, potentially before processing the block. At this point column
+                // custodials may not have received all their columns. Triggering sampling so early is
+                // only viable with either:
+                // - Sync delaying sampling until some latter window
+                // - Re-processing early sampling requests: https://github.com/sigp/lighthouse/pull/5569
+                self.send_sync_message(SyncMessage::SampleBlock(block_root, slot));
```
jimmygchen: Minor - for supernodes, we might get flooded with `Received SampleBlock message` if we're sending a sync message for every gossip column received
https://github.com/sigp/lighthouse/blob/f8acdac4c4dc4cb54ebaa5d3871302f0c78bd6c6/beacon_node/network/src/sync/manager.rs#L656

beacon_node/network/src/sync/network_context/requests.rs
```diff
@@ -147,3 +151,90 @@ impl<E: EthSpec> ActiveBlobsByRootRequest<E> {
...
+            resolved: false,
+        }
+    }
+
+    /// Appends a chunk to this multi-item request. If all expected chunks are received, this
+    /// method returns `Some`, resolving the request before the stream terminator.
+    /// The active request SHOULD be dropped after `add_response` returns an error
+    pub fn add_response(
+        &mut self,
+        blob: Arc<DataColumnSidecar<E>>,
```
jimmygchen: `data_column`

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
dapplion: Right, triggering on column is a timing optimization. We can drop it for simplicity

beacon_node/network/src/sync/sampling.rs
jimmygchen: Resolving this as we can address later.

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: I'll add a comment

jimmygchen:
I've pushed a commit to resolve some minor review comments, and added a check for `PEER_DAS_EPOCH` before sending a sampling request.

I've looked into the test failures (network_tests, fallback sim, and basic sim tests), and ~~they all passed locally, so I'll let it run on CI again (could be caching issue).~~

**EDIT**: found a `Duration` subtraction panic in the logs, investigating now.

I'll do a merge from `unstable` to `das` after this is merged to `das`.

jimmygchen:
I couldn't reproduce the sim tests issues locally, and it appears to be happening on [other PRs](https://github.com/sigp/lighthouse/actions/runs/8859873093) too. I'm going to look into it later tonight.

The error I'm seeing on CI is:
> 2024-04-29T07:35:16.5280075Z thread 'main' panicked at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/core/src/time.rs:954:31:
2024-04-29T07:35:16.5280267Z overflow when subtracting durations
2024-04-29T07:35:16.5280709Z note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

The execution tests should pass once we merge `unstable` into `das`, so I think we're good to merge this one now!

PR #5620, opened 2024-04-22
Drop lookup type trait for a simple arg

dapplion:
## Issue Addressed

Part of

Passing around a simple `lookup_type: LookupType` to the requester function achieves the same functionality with less machinery. Makes code simpler to extend, related to DAS.

## Proposed Changes

- Drop lookup type trait for a simple arg
- Drop leftover unit tests in `beacon_node/network/src/sync/block_lookups/single_block_lookup.rs`. Their tested scenarios are covered by the event-based tests

### DIFF
--- a/beacon_node/network/src/sync/block_lookups/common.rs
+++ b/beacon_node/network/src/sync/block_lookups/common.rs
@@ -29,34 +29,12 @@ pub enum LookupType {
     Parent,
 }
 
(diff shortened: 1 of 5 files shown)


realbigsean (approved):
Nice

realbigsean:
@mergify queue

jimmygchen:
@mergify requeue

Issue #4388, opened 2023-06-09
Consider reducing the frequency of pending validator indices queries from validator client


jimmygchen:
## Description

Currently the Validator Client (VC) polls the validator states for all inactive validators from the Beacon Node (for indices retrieval), once every slot (12 seconds on mainnet) 

https://github.com/sigp/lighthouse/blob/f16795183564a8487c32890f511a24d6abac82e4/validator_client/src/duties_service.rs#L279-L292

This is mostly not an issue until the number of _inactive_ validator validators reaches a large number  ~1000, which is probably quite rare. However we've recently seen some performance issues when the endpoint `beacon/states/{state_id}/validators/{validator_id}` is called repeatedly in a short period of time. Here's a [script](https://github.com/michaelsproul/spam/blob/main/src/main.rs) created by @michaelsproul to spam this endpoint aggressively, and it turns out this could cause an OOM on the beacon node. 

There has been some discussions on how to improve this, potentially queuing the requests, however they might take a while to implement. In the mean time, we can probably reduce the frequency of this query to once or twice per epoch to reduce the performance impact on the node, as validator activation only happens once every epoch, and it may not be necessary to query the indices so often.

paulhauner:
One way to reduce the frequency of polling would be to make an assumption that the VC only needs to know a validator index for an *active* validator.

Validators are added to the `BeaconState` via `Deposit` objects in a `BeaconBlock`. This happens on a *per slot* basis, so new validators can land in the `BeaconState` (i.e. be assigned an index) each slot. However, validators are *activated* on a *per epoch* basis (`process_epoch` > `process_registry_updates`).

So, it seems that it would be safe for us to:

1. Poll for *all* validator indices on startup.
1. When we've received a `None` response for a validator, poll each epoch.
    - I think we should consider *not* polling in the first slot of the epoch. That slot is known to be a weak point for us. Validators are activated at least `1 + MAX_SEED_LOOKAHEAD == 5` epochs after they first appear in the `BeaconState` so there's no rush to discover the validator.

There's an edge-case here where a syncing BN might return `None`, then import several epochs of the chain and then return `Some` and that validator is already active. In per-slot polling we would resolve this within a slot, but with per-epoch processing it takes an epoch to resolve. I'm not sure how I feel about this edge-case, it's hard to weigh off the costs of optimising for weird sync cases vs optimising for the case of big operators running lots of undeposited validators.

jimmygchen:
Thanks Paul! 
[...]

PR #5628, opened 2024-04-23
Reduce frequency of polling unknown validators to avoid overwhelming the Beacon Node

jimmygchen:
## Issue Addressed

Addresses #4388.

A large number of unknown validators on a VC is known to overwhelm the beacon node because the VC triggers a retrieval for each validator per slot. This PR reduces the frequency to one query per unknown validator each epoch. 

## Proposed Changes

1. Poll for all validator indices on startup (unchanged)
2. If any validator is unknown, register to poll again in the 32 slots (1 epoch) instead of the next slot.
3. Avoid polling on the first 1st slot of epoch.

For more details and rationale, see [comment here](https://github.com/sigp/lighthouse/issues/4388#issuecomment-2054233516).

### DIFF
--- a/validator_client/src/duties_service.rs
+++ b/validator_client/src/duties_service.rs
@@ -493,12 +493,7 @@ async fn poll_validator_indices<T: SlotClock + 'static, E: EthSpec>(
             let current_slot_opt = duties_service.slot_clock.now();
 
             if let Some(current_slot) = current_slot_opt {
-                let is_first_slot_of_epoch = {
-                    let current_epoch_first_slot = current_slot
-                        .epoch(E::slots_per_epoch())
-                        .start_slot(E::slots_per_epoch());
-                    current_slot == current_epoch_first_slot
-                };
+               let is_first_slot_of_epoch = current_slot % E::slots_per_epoch() == 0;
 
                 // Query an unknown validator later if it was queried within the last epoch, or if
                 // the current slot is the first slot of an epoch.
--- a/validator_client/src/duties_service.rs
+++ b/validator_client/src/duties_service.rs
@@ -493,7 +493,7 @@ async fn poll_validator_indices<T: SlotClock + 'static, E: EthSpec>(
             let current_slot_opt = duties_service.slot_clock.now();
 
             if let Some(current_slot) = current_slot_opt {
-               let is_first_slot_of_epoch = current_slot % E::slots_per_epoch() == 0;
+                let is_first_slot_of_epoch = current_slot % E::slots_per_epoch() == 0;
 
                 // Query an unknown validator later if it was queried within the last epoch, or if
                 // the current slot is the first slot of an epoch.
--- a/validator_client/src/duties_service.rs
+++ b/validator_client/src/duties_service.rs
@@ -480,18 +480,6 @@ async fn poll_validator_indices<T: SlotClock + 'static, E: EthSpec>(
         .validator_store
         .voting_pubkeys(DoppelgangerStatus::ignored);
 
-    let current_slot_opt = duties_service.slot_clock.now();
-    let next_poll_slot_opt = current_slot_opt.map(|slot| slot.saturating_add(E::slots_per_epoch()));
-
-    let is_first_slot_of_epoch = if let Some(current_slot) = current_slot_opt {
-        let current_epoch_first_slot = current_slot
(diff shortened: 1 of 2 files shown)


validator_client/src/duties_service.rs
```diff
@@ -214,6 +214,8 @@ pub struct DutiesService<T, E: EthSpec> {
     pub sync_duties: SyncDutiesMap<E>,
     /// Provides the canonical list of locally-managed validators.
     pub validator_store: Arc<ValidatorStore<T, E>>,
+    /// Maps unknown validator pubkeys to the next slot time when a poll should be conducted again.
+    pub unknown_validator_next_poll_slots: RwLock<HashMap<PublicKeyBytes, Slot>>,
```
pawanjay176, line 218: Does this need to be behind a lock? Not that big of an overhead but seems like its just the update service that uses it.

pawanjay176 (approved):
Nice, LGTM. Just a minor question

validator_client/src/duties_service.rs
jimmygchen: Not sure to be honest - I initially tried using `HashMap` directly but ran into some issues. The async task that runs the polling only gets a immutable reference to the `DutiesService` via an `Arc`, and I couldn't borrow it as a mutable directly,  so I kinda borrowed the existing pattern that's used to mutate `DutiesService`. Would be happy to get rid of this if there's a way to do this nicely.

pawanjay176: Makes sense, since we pass around a reference, I think this is the only clean way to mutate.

jimmygchen:
I'll do a bit of manual testing next week before we merge this!

jimmygchen:
@chong-he found an issue during testing and it looks like the VC re-queries the BN during the first slot of the epoch, I'll look into this.

jimmygchen:
Upon investigating the logs, I _think_ it is working as intended, and the poll in the first slot of epoch is skipped. Will wait for CK to confirm.

However, I think there's likelihood that this could happen due to the async nature of this function, in the scenario where BN returns a response really late - basically if iterating through all the validators takes close to 12 seconds, then we could be querying in a new slot, which could potentially be first slot of a new epoch.

This is because of the calculation _outside_ the for loop:
https://github.com/sigp/lighthouse/blob/ce914d191d804a765c573843c3c7448facd04ccb/validator_client/src/duties_service.rs#L483-L502

For more accuracy, I'll move the calculation into the for loop given we await on each validator query and the slot status could be inaccurate if we have a large list or slow BN.

chong-he:
> @chong-he found an issue during testing and it looks like the VC re-queries the BN during the first slot of the epoch, I'll look into this.

Apologies! I was looking at the wrong slot time. 

The VC does not re-query on the first slot in the next epoch (even though we start the VC in the first slot). It waits until the 12s has lapsed and only re-query the BN about the status (i.e., query at the second slot of an epoch). The VC also only query once per epoch, which is a great improvement. 

All good upon testing, this PR is good to go

validator_client/src/duties_service.rs
```diff
@@ -488,6 +490,29 @@ async fn poll_validator_indices<T: SlotClock + 'static, E: EthSpec>(
...
         if !is_known {
+            let current_slot_opt = duties_service.slot_clock.now();
+
+            if let Some(current_slot) = current_slot_opt {
+                let is_first_slot_of_epoch = {
+                    let current_epoch_first_slot = current_slot
+                        .epoch(E::slots_per_epoch())
+                        .start_slot(E::slots_per_epoch());
+                    current_slot == current_epoch_first_slot
+                };
```
michaelsproul: This could be a bit simpler as:

```suggestion
               let is_first_slot_of_epoch = current_slot % E::slots_per_epoch() == 0;
```

That's usually how we do it

michaelsproul (approved):
Looks good. Nice improvement

I think we should include this in 5.2

validator_client/src/duties_service.rs
jimmygchen: Nice, thanks!

michaelsproul:
sorry looks like cargo fmt is failing because I messed up the indentation in my suggestion

jimmygchen:
My bad, should've checked myself :P thanks!

michaelsproul:
@mergifyio queue

PR #5630, opened 2024-04-23
Add electra presets to beacon API

realbigsean:
## Issue Addressed

an extension of https://github.com/sigp/lighthouse/pull/5606

### DIFF
--- a/consensus/types/presets/gnosis/electra.yaml
+++ b/consensus/types/presets/gnosis/electra.yaml
@@ -1,3 +1,45 @@
-# Gnosis preset - Electra
+# Mainnet preset - Electra
 
-ELECTRA_PLACEHOLDER: 0
+# Gwei values
+# ---------------------------------------------------------------
+# 2**5 * 10**9 (= 32,000,000,000) Gwei
+MIN_ACTIVATION_BALANCE: 32000000000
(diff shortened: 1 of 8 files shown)


realbigsean:
`get_extra_fields` needs to be updated with constants (looks like deneb constants are actually missing there)

realbigsean:
> `get_extra_fields` needs to be updated with constants (looks like deneb constants are actually missing there)

Looks like there are a bunch of phase0 and altair constants missing.. not sure if there's a pattern? or these are just spec refactors we've missed?

pawanjay176 (changes requested):
Gnosis and minimal preset chainspec defaults seem to be wrong

realbigsean:
There's no gnosis preset in the spec so I wasn't sure what to add. Forgot it wouldn't have been able to read from the file. I guess I'll drop mainnet preset in there for now? https://github.com/gnosischain/specs

realbigsean:
I raised an issue to track gnosis presets https://github.com/sigp/lighthouse/issues/5638

pawanjay176 (approved):

realbigsean:
@mergify queue

realbigsean:
@mergify requeue

Issue #5572, opened 2024-04-12
Docker build failing on `x86_64_dev` since moving to the `c-kzg` from crate-io 


jimmygchen:
## Description

Our docker `x86_64_dev` cross build has been failing since moving to the crate-io version of `c-kzg` (https://github.com/sigp/lighthouse/pull/5542). We should fix this before the next release.

```
thread 'main' panicked at /cargo/registry/src/index.crates.io-6f17d22bba15001f/blst-0.3.11/build.rs:173:25:
Cannot compile with both \`portable\` and \`force-adx\` features
```

https://github.com/sigp/lighthouse/actions/runs/8639102912/job/23684787522

pawanjay176:
We can remove `force-adx` since the portable build should detect it on runtime and use it if available.
cc @michaelsproul

michaelsproul:
Yeah, agreed. We could go full steam ahead with Anton's change:

- https://github.com/sigp/lighthouse/pull/5489

I think the alternative would be to compile czkg with `no-default-features` so that it doesn't turn on `blst/portable`: https://github.com/ethereum/c-kzg-4844/blob/1ef67c17bf659547749750e6ebaaf09563e0149d/Cargo.toml#L21-L23

antondlr:
> Yeah, agreed. We could go full steam ahead with Anton's change:
> 
> * [Use `portable` builds by default  #5489](https://github.com/sigp/lighthouse/pull/5489)
> 
> I think the alternative would be to compile czkg with `no-default-features` so that it doesn't turn on `blst/portable`: https://github.com/ethereum/c-kzg-4844/blob/1ef67c17bf659547749750e6ebaaf09563e0149d/Cargo.toml#L21-L23

fyi I split out the docker and release/binaries parts into PR's #5614 and #5615 
the docker one (#5614) can thus be merged without (too much) worries of breaking the release flow.

jimmygchen:
Fixed in https://github.com/sigp/lighthouse/pull/5614 🎉

PR #5641, opened 2024-04-24
Bump jobserver and fix non-portable builds

michaelsproul:
## Issue Addressed

Fix broken builds as in: https://github.com/sigp/lighthouse/actions/runs/8809536524/job/24180492013

Closes:

- https://github.com/sigp/lighthouse/issues/5572

## Proposed Changes

- The main fix is to bump `jobserver` from 0.1.30 to 0.1.31 to remove usage of the `preadv2` syscall that was broken under Cross: https://github.com/rust-lang/jobserver-rs/commit/66782f5e64ccd5b0f2ff1b74fd8f54e4b081b421
- This PR also turns off `default-features` for `c-kzg` so that it's possible to compile Lighthouse _without_ portable blst. Even though we're switching to portable by default (https://github.com/sigp/lighthouse/pull/5615) we may as well unbreak the ability to build non-portable binaries.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1175,7 +1175,6 @@ dependencies = [
  "glob",
  "hex",
  "libc",
- "serde",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


michaelsproul:
Prior to this PR, both `make build-x86_64` and `make build-x86_64-portable` were broken on my machine. With this PR, both work.

jimmygchen (approved):
LGTM!

michaelsproul:
@mergifyio queue

jimmygchen:
@mergify requeue

PR #5647, opened 2024-04-25
Fix execution integration tests

realbigsean:
## Issue Addressed

Broken geth integration tests are due to the fact that it no longer returns "synced" until a CL queries it. This change is to rather than wait for the EL to return "synced", just wait for it to have an "online" state.


ethDreamer (approved):

realbigsean:
@mergify queue

PR #5652, opened 2024-04-25
Electra other containers

realbigsean:
## Issue Addressed
Resolves https://github.com/sigp/lighthouse/issues/5608
Part of https://github.com/sigp/lighthouse/issues/5605

## Proposed Changes

- Add new fields for `ExecutionPayload`, `ExecutionPayloadHeader`, `BeaconState`, `PartialBeaconState`
- implement `upgrade_to_electra`
  - This required adding the following helper methods
    - `get_balance_churn_limit`
    - `get_activation_exit_churn_limit`
    - `get_consolidation_churn_limit`
    - `queue_entire_balance_and_reset_validator`
    - `queue_excess_active_balance`
    - `has_compounding_withdrawal_credential`
    - `is_compounding_withdrawal_credential`

PR #5653, opened 2024-04-25
Add more electra helpers

pawanjay176:
## Issue Addressed

Partly addresses #5605 
Based on #5652 

## Proposed Changes

Implements the following functions from the electra spec:
- is_eligible_for_activation_queue
- has_execution_withdrawal_credential
- is_fully_withdrawable_validator
- is_partially_withdrawable_validator
- get_validator_max_effective_balance
- get_active_balance
- get_pending_balance_to_withdraw
- switch_to_compounding_validator
- compute_exit_epoch_and_update_churn
- compute_consolidation_epoch_and_update_churn

### DIFF
--- a/consensus/types/src/beacon_state.rs
+++ b/consensus/types/src/beacon_state.rs
@@ -2092,6 +2092,40 @@ impl<E: EthSpec> BeaconState<E> {
             .map_err(Into::into)
     }
 
+    pub fn get_active_balance(
+        &self,
+        validator_index: usize,
+        spec: &ChainSpec,
+    ) -> Result<u64, Error> {
+        let max_effective_balance = self
+            .validators()
+            .get(validator_index)
+            .map(|validator| validator.get_validator_max_effective_balance(spec))
+            .ok_or(Error::UnknownValidator(validator_index))?;
+        // TODO(pawan): this is assuming balances and validat
+        Ok(std::cmp::min(
+            *self
+                .balances()
+                .get(validator_index)
+                .ok_or(Error::UnknownValidator(validator_index))?,
+            max_effective_balance,
+        ))
+    }
+
+    pub fn get_pending_balance_to_withdraw(&self, validator_index: u64) -> Result<u64, Error> {
+        Ok(self
+            .pending_partial_withdrawals()?
+            .iter()
+            .filter_map(|withdrawal| {
+                if withdrawal.index == validator_index {
+                    Some(withdrawal.amount)
+                } else {
+                    None
+                }
+            })
+            .sum())
+    }
+
     // ******* Electra mutators *******
 
     pub fn queue_excess_active_balance(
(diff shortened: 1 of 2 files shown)


consensus/types/src/validator.rs
```diff
@@ -135,10 +135,44 @@ impl Validator {
...
+    /// Returns `true` if the validator has a 0x01 or 0x02 prefixed withdrawal credential.
+    pub fn has_execution_withdrawal_credential(&self, spec: &ChainSpec) -> bool {
+        self.has_compounding_withdrawal_credential(spec)
+            || self.has_eth1_withdrawal_credential(spec)
+    }
+
+    /// Returns `true` if the validator if fully withdrawable.
+    ///
+    /// Modified in electra.
+    pub fn is_fully_withdrawable_validator(
```
realbigsean: Looks like we already have this one named `is_fully_withdrawable_at` on L133 here (still needs a slight modification though)

consensus/types/src/beacon_state.rs
```diff
@@ -2142,6 +2174,101 @@ impl<E: EthSpec> BeaconState<E> {
...
+        &mut self,
+        validator_index: usize,
+        spec: &ChainSpec,
+    ) -> Result<(), Error> {
+        let validator = self
+            .validators_mut()
+            .get_mut(validator_index)
+            .ok_or(Error::UnknownValidator(validator_index))?;
+        if validator.has_eth1_withdrawal_credential(spec) {
+            validator.withdrawal_credentials.as_fixed_bytes_mut()[0] =
```
realbigsean, line 2188: Should we use `first_mut` and throw an error if it's `None` here instead of the slice access?

realbigsean (changes requested):

consensus/types/src/validator.rs
pawanjay176: ahh nice catch. I'll remove the new one I added and modify the existing one. 
The existing naming makes more sense

consensus/types/src/beacon_state.rs
pawanjay176: I thought of doing it that way first but I couldn't figure out what error to throw in that case. I think it should always be safe because H256 wraps a fixed 32 byte array, so we can always index 0 to 31.
We could also do something like
```rust
let mut bytes = [0; 32];
bytes.copy_from_slice(&validator.withdrawal_credentials.as_bytes());
bytes[0] = spec.compounding_withdrawal_prefix_byte;
validator.withdrawal_credentials = Hash256::from_slice(&bytes);
```
if we want it to be more explicit.

realbigsean: Yea i think it's actually fine how it is, `has_eth1_withdrawal_credential` checks the first byte exists anyways

consensus/types/src/validator.rs
pawanjay176: Fixed in [c9510ab](https://github.com/sigp/lighthouse/pull/5653/commits/c9510abe6c30c9f78b126a6a765616fd5019600f)

realbigsean (approved):
nice!

realbigsean:
@mergify queue

PR #915, opened 2020-03-16
Fix master failing test

AgeManning:
## Issue Addressed

Master currently fails if `cargo update` is applied. The mock BLS key used to test ssz encoding is not of the correct range. The new BLS throws a panic and the test fails. 

This PR specifies a valid BLS key for the test.

PR #5654, opened 2024-04-26
Deterministic block generation for tests

dapplion:
## Issue Addressed

PR 
- https://github.com/sigp/lighthouse/pull/915

broke determinism in block generation. This is very annoying when debugging sync lookup tests, as the block roots keep changing every run.

## Proposed Changes

As the blocks don't contain any BLS pubkey, make the signature generation always default to the G2 point at infinity.


jimmygchen (approved):
Nice! Looks good to me

realbigsean:
@mergify queue

PR #5655, opened 2024-04-26
Merge parent and current sync lookups

dapplion:
## Issue Addressed

Part of 

Block lookups fulfill the need to:
- Sync blocks referenced by root by other consensus objects (i.e. attestations)
- Sync and handle a chain of blocks whose parent is unknown (not rooted in our view of the chain)

Both goals interleave with each other, but in current unstable code each one is treated differently. This has been a significant source of pain and bugs, and the logic to transition between "parent lookup" and "current lookup" is gnarly.

This PR attempts to reduce the surface area for issues by:
- Reducing the state transition on these request lookup / states, and making them strict
- Splitting the complexity of handling the lookup (retries, send to processor) and the complexity of unknown parent sync itself. In a similar spirit of https://github.com/sigp/lighthouse/pull/5583 which moved the complexity of streams to somewhere else

## Proposed Changes

- Merge parent and current sync lookups

Deep dive into the rationale and changes overview (26 minutes) =D

https://drive.google.com/file/d/1vTOrvZKuYdn6iAJ3fdiNZhvqJycwkoiy/view?usp=sharing

Screenshot 2024-04-30 at 20 31 14

## Next steps

- https://github.com/sigp/lighthouse/issues/5660
- https://github.com/sigp/lighthouse/issues/5662


beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -2918,6 +2918,9 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             .fork_choice_read_lock()
             .contains_block(&block_root)
         {
+            // TODO: Should also check for:
```
realbigsean: moved to issue https://github.com/sigp/lighthouse/issues/5660

beacon_node/network/src/sync/block_lookups/common.rs
```diff
@@ -53,115 +43,77 @@ pub trait RequestState<T: BeaconChainTypes> {
...
-
-        if request_state.failed_attempts() >= lookup_type.max_attempts() {
-            let cannot_process = request_state.more_failed_processing_attempts();
-            Err(LookupRequestError::TooManyAttempts { cannot_process })
-        } else {
-            Ok(())
+        // Attempt to progress awaiting downloads
+        if self.get_state().is_awaiting_download() {
+            // Verify the current request has not exceeded the maximum number of attempts.
+            // TODO: Okay to use `SINGLE_BLOCK_LOOKUP_MAX_ATTEMPTS` for both current and parent
```
realbigsean: i think this is ok, don't understand the benefit of treating them differently, and the complexity saved by treating them the same is valuable

beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -1,756 +1,432 @@
...
-
-    processing_parent_lookups: HashMap<Hash256, (Vec<Hash256>, SingleBlockLookup<T>)>,
-
     /// A cache of failed chain lookups to prevent duplicate searches.
     failed_chains: LRUTimeCache<Hash256>,
 
-    single_block_lookups: FnvHashMap<Id, SingleBlockLookup<T>>,
-
-    pub(crate) da_checker: Arc<DataAvailabilityChecker<T>>,
+    // TODO: Why not index lookups by block_root?
```
realbigsean, line 67: Moved to issue https://github.com/sigp/lighthouse/issues/5662

beacon_node/network/src/sync/block_lookups/single_block_lookup.rs
```diff
@@ -342,36 +208,22 @@ impl BlockRequestState {
...
-    /// been performed and no kzg verification has been performed.
-    Ok(RpcBlock<E>),
-    /// All child components have not yet been received.
-    DownloadIncomplete,
-    /// Child components should not be cached, send this directly for processing.
-    NotRequired,
-    /// There was an error during consistency checks between block and blobs.
-    Err(AvailabilityCheckError),
-}
+pub type DownloadResult<T> = (T, Hash256, Duration, PeerId);
```
realbigsean: make struct

beacon_node/network/src/sync/block_lookups/common.rs
```diff
@@ -53,115 +43,77 @@ pub trait RequestState<T: BeaconChainTypes> {
...
+            // make_request returns true only if a request was made
+            if self.make_request(id, peer_id, downloaded_block_expected_blobs, cx)? {
+                self.get_state_mut().on_download_start()?;
+            }
+
+        // Otherwise, attempt to progress awaiting processing
+        // If this request is awaiting a parent lookup to be processed, do not send for processing.
+        // The request will be rejected with unknown parent error.
+        } else if !awaiting_parent &&
+            // TODO: Blob processing / import does not check for unknown parent. As a temporary fix
```
realbigsean: moved to issue https://github.com/sigp/lighthouse/issues/5660

beacon_node/network/src/sync/block_lookups/parent_chain.rs
```diff
@@ -0,0 +1,197 @@
...
+            parent_to_child
+                .entry(parent_root)
+                .or_default()
+                .push(node.block_root);
+        }
+    }
+
+    let mut parent_chains = vec![];
+
+    // Iterate blocks which no child
```
realbigsean:
```suggestion
    // Iterate blocks with no children
```

beacon_node/network/src/sync/block_lookups/parent_chain.rs
```diff
@@ -0,0 +1,197 @@
...
+    for tip in nodes {
+        let mut block_root = tip.block_root;
+        if parent_to_child.get(&block_root).is_none() {
+            let mut chain = vec![];
+
+            // Resolve chain of blocks
+            loop {
```
realbigsean: loop labels?

beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -1,756 +1,432 @@
...
     }
 
-    /// Creates a lookup for the block with the given `block_root`, while caching other block
-    /// components we've already received. The block components are cached here because we haven't
-    /// imported its parent and therefore can't fully validate it and store it in the data
-    /// availability cache.
-    ///
-    /// The request is immediately triggered.
-    pub fn search_child_block(
+    /// Seach a block that we don't known its parent root.
```
realbigsean:
```suggestion
    /// Seach a block whose parent root is unknown.
```

beacon_node/network/src/sync/block_lookups/parent_chain.rs
```diff
@@ -0,0 +1,197 @@
...
+    let parent_chain = parent_chains.get(chain_idx).ok_or("chain_idx off bounds")?;
+    // Find the first block in the target parent chain that is not in other parent chains
+    // Iterate in ascending slot order
+    for block in parent_chain.chain.iter().rev() {
+        if !other_blocks.contains(block) {
+            return Ok(*block);
+        }
+    }
+
+    // If no match means that the chain is fully contained within another chain. This should never
```
realbigsean:
```suggestion
    // No match means that the chain is fully contained within another chain. This should never
```

beacon_node/network/src/sync/block_lookups/parent_chain.rs
```diff
@@ -0,0 +1,197 @@
...
+    for (i, parent_chain) in parent_chains.iter().enumerate() {
+        if i != chain_idx {
+            for block in &parent_chain.chain {
+                other_blocks.insert(block);
+            }
+        }
+    }
+
+    // Should never happen
+    let parent_chain = parent_chains.get(chain_idx).ok_or("chain_idx off bounds")?;
```
realbigsean:
```suggestion
    let parent_chain = parent_chains.get(chain_idx).ok_or("chain_idx out of bounds")?;
```

beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -1,756 +1,432 @@
...
+                        .find(|(_, l)| l.block_root() == block_to_drop)
+                    {
+                        for &peer_id in lookup.all_used_peers() {
+                            cx.report_peer(
+                                peer_id,
+                                PeerAction::LowToleranceError,
+                                "chain_too_long",
+                            );
+                        }
+                        self.drop_lookup_and_childs(*lookup_id);
```
realbigsean:
```suggestion
                        self.drop_lookup_and_children(*lookup_id);
```

beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -851,501 +529,86 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
...
-                let Some(id) = self
-                    .single_block_lookups
-                    .iter()
-                    .find_map(|(id, req)| (req.block_root() == chain_hash).then_some(*id))
-                else {
-                    warn!(self.log, "No id found for single block lookup"; "chain_hash" => %chain_hash);
-                    return;
-                };
+    /// Drops `dropped_id` lookup and all its children recursively. Lookups awaiting a parent need
+    /// the parent to make progress to resolve, therefore we must drop them is the parent is
```
realbigsean:
```suggestion
    /// the parent to make progress to resolve, therefore we must drop them if the parent is
```

beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -759,90 +435,92 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
...
-                            lookup.handle_availability_check_failure(cx);
-                            Action::Retry
-                        }
-                    },
+                    BlockError::AvailabilityCheck(e)
+                        if e.category() == AvailabilityCheckErrorCategory::Internal =>
+                    {
+                        // There errors indicate internal problems and should not downscore the  peer
+                        warn!(self.log, "Internal availability check failure"; "block_root" => %block_root, "error" => ?e);
+                        // TODO: This lines represent an improper transition of download states,
```
realbigsean: Not following this comment

beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -759,90 +435,92 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
...
-                        if let Ok(block_peer) = lookup.block_request_state.state.processing_peer() {
-                            cx.report_peer(
-                                block_peer,
-                                PeerAction::MidToleranceError,
-                                "single_block_failure",
-                            );
+                        debug!(self.log, "Invalid lookup component"; "block_root" => %block_root, "component" => ?R::response_type(), "error" => ?other);
+                        let peer_id = request_state.on_processing_failure()?;
+                        // TODO: Why is the original code downscoring the block peer regardless of
+                        // type of request? Sending a blob for verification can result in an error
```
realbigsean: When we split up block and blob processing errors, it should make it more obvious who we should be downscoring. 

A lot of the errors here should be coming from block processing after the block + blobs have left the availability cache.  A blob might have been returned last via RPC and triggered block processing. So it doesn't really make sense to just downscore the blob peer in this case. Maybe the correct thing to do in some cases is to downscore both.

beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -1,756 +1,432 @@
...
+                .map(|lookup| lookup.into())
+                .collect::<Vec<_>>(),
+        )
     }
 
     /* Lookup requests */
 
     /// Creates a lookup for the block with the given `block_root` and immediately triggers it.
-    pub fn search_block(
+    /// Returns true if the lookup is created or already exists
```
realbigsean: Fix comment

beacon_node/network/src/sync/block_lookups/single_block_lookup.rs
```diff
@@ -397,74 +247,156 @@ impl SingleLookupRequestState {
...
+            State::AwaitingProcess(result) => Some(&result.0),
+            State::Processing(result) => Some(&result.0),
+            State::Processed { .. } => None,
+        }
+    }
+
+    /// Switch to `AwaitingProcessing` if the request is in `AwaitingDownload` state, otherwise
+    /// ignore.
+    pub fn insert_verified_response(&mut self, result: DownloadResult<T>) {
+        if let State::AwaitingDownload = &self.state {
```
realbigsean, line 294: Will this not cause us to drop results rather than merging them when we are merging duplicate requests? the state of those requests could be `Downloading` for example

```
        // Do not re-request a block that is already being requested
        if let Some((_, lookup)) = self
            .single_block_lookups
            .iter_mut()
            .find(|(_id, lookup)| lookup.is_for_block(block_root))
        {
            trace!(self.log, "Adding peer to existing single block lookup"; "block_root" => %block_root);
            lookup.add_peers(peers);
            if let Some(block_component) = block_component {
                lookup.add_child_components(block_component);
            }
            return true;
        }
```

beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -1066,36 +1211,47 @@ fn test_same_chain_race_condition() {
...
+        rig.single_block_component_processed_imported(block.canonical_root());
+    }
+    rig.expect_no_active_lookups();
+}
 
-    rig.parent_chain_processed_success(chain_hash);
-    assert_eq!(rig.active_parent_lookups_count(), 0);
+#[test]
+fn test_penalize_wrong_peer_with_cached_child() {
+    // peer A sends blob with malicious data as unknown parent
```
realbigsean: Is this a place holder?

beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -1028,8 +1171,10 @@ fn test_parent_lookup_ignored_response() {
     let peer_id = rig.new_connected_peer();
 
     // Trigger the request
-    rig.trigger_unknown_parent_block(peer_id, block.into());
+    rig.trigger_unknown_parent_block(peer_id, block.clone().into());
     let id = rig.expect_parent_request_block_and_blobs(parent_root);
+    // Note: single block lookup for current `block` does not trigger any request because it does
+    // not has blobs, and the block is already cached
```
realbigsean:
```suggestion
    // not have blobs, and the block is already cached
```

beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -1232,6 +1388,21 @@ mod deneb_only {
             })
         }
 
+        fn log(self, msg: &str) -> Self {
+            self.rig.log(msg);
+            self
+        }
+
+        // TODO: Eventually deprecate this function
+        fn set_block_id_for_import(mut self) -> Self {
```
realbigsean: when should this be deprecated? I'm going to remove the TODO because it should be identified as dead code when we will want to remove it

beacon_node/network/src/sync/block_lookups/tests.rs
realbigsean: I'm going to remove it in my PR but if you have time to add it in, by all means

beacon_node/network/src/sync/block_lookups/mod.rs
realbigsean: fixed in https://github.com/dapplion/lighthouse/pull/26

beacon_node/network/src/sync/block_lookups/mod.rs
realbigsean: fixed in https://github.com/dapplion/lighthouse/pull/26

beacon_node/network/src/sync/block_lookups/mod.rs
realbigsean: fixed in https://github.com/dapplion/lighthouse/pull/26

beacon_node/network/src/sync/block_lookups/mod.rs
realbigsean: fixed in https://github.com/dapplion/lighthouse/pull/26

beacon_node/network/src/sync/block_lookups/parent_chain.rs
realbigsean: fixed in https://github.com/dapplion/lighthouse/pull/26

beacon_node/network/src/sync/block_lookups/parent_chain.rs
realbigsean: added some here: https://github.com/dapplion/lighthouse/pull/26

they don't change functionality, i like them in nested loops for clarity though

beacon_node/network/src/sync/block_lookups/parent_chain.rs
realbigsean: fixed in https://github.com/dapplion/lighthouse/pull/26

beacon_node/network/src/sync/block_lookups/parent_chain.rs
realbigsean: fixed in https://github.com/dapplion/lighthouse/pull/26

beacon_node/network/src/sync/block_lookups/single_block_lookup.rs
realbigsean: added in https://github.com/dapplion/lighthouse/pull/26

beacon_node/network/src/sync/block_lookups/tests.rs
realbigsean: fixed in https://github.com/dapplion/lighthouse/pull/26

beacon_node/network/src/sync/block_lookups/tests.rs
realbigsean: removed in https://github.com/dapplion/lighthouse/pull/26/

beacon_node/network/src/sync/block_lookups/tests.rs
realbigsean: removed in https://github.com/dapplion/lighthouse/pull/26/

beacon_node/network/src/sync/block_lookups/common.rs
```diff
@@ -53,115 +43,77 @@ pub trait RequestState<T: BeaconChainTypes> {
...
-    }
-
-    /// Construct a new request and send it.
-    fn build_request_and_send(
+    /// Potentially makes progress on this request if it's in a progress-able state
+    fn continue_request(
         &mut self,
         id: Id,
-        lookup_type: LookupType,
+        awaiting_parent: bool,
```
realbigsean: I updated the bool args to enums in my PR just so you can't mix up the bools as you pass them into the function https://github.com/dapplion/lighthouse/pull/26/

beacon_node/network/src/sync/block_lookups/common.rs
realbigsean: removed the comment in https://github.com/dapplion/lighthouse/pull/26/

realbigsean (changes requested):

beacon_node/network/src/sync/block_lookups/parent_chain.rs
dapplion: Will commit a simpler impl that I drafted with @michaelsproul, should make the labels not necessary:

```rust
while let Some(parent_root) = child_to_parent.get(&block_root) {
    // block_root is a known block that may or may not have a parent root
    chain.push(block_root);
    if let Some(parent_root) = parent_root {
        block_root = *parent_root;
    } else {
        break;
    }
}
```

EDIT: committed in 0203843202f2c451609c0a1afa54fee23bc560ed

beacon_node/network/src/sync/block_lookups/mod.rs
dapplion: Ok will try to illustrate with an example, see you can help me get a better comment:

- Lookup created
- Block request sent
- Blob request sent
- Blob request completed, and sent for processing (blob request state = Processing)
- Block request completed, sent for processing
- Block processing returns `BlockError::AvailabilityCheck(AvailabilityCheckErrorCategory::Internal)`
- Call `blob.on_download_failure()` results in `bad state: expected Downloading got Processing`

This error can happen with different variations of this scenario as you don't know the status of the blob request which is happening in parallel. You can error the block request because you know it is in `Processing` state, but not the blob request.

If we are currently forcing the blob request to transition from `Processed` into `AwaitingDownload` it means that a `AvailabilityCheckErrorCategory::Internal` can invalidate blobs that were previously considered valid, which is very weird. I don't understand under what circumstances that would happen. If there are concerns that an `AvailabilityCheckErrorCategory::Internal` break consistency or it is not recoverable we should just drop the request.

beacon_node/network/src/sync/block_lookups/mod.rs
dapplion: Then I think the `BlockProcessingResult` is not well specified for this function, too much guessing happening here. It should be more clear to attribute faults to internal, block or blob providers.

beacon_node/network/src/sync/block_lookups/tests.rs
dapplion: The deneb test rig made some assumptions about which is the target block to download that somehow broke when modifying the test cases, I think when adding `block_missing_components` function. I could not reason why, so I added this method to get the tests to pass. Seems like a bit of a hack, could you look into it?

beacon_node/network/src/sync/block_lookups/single_block_lookup.rs
dapplion: single block lookups assume there is a single request in-flight. If the state is `Downloading` and we insert a verified response we can expect that eventually the in-flight request will finish and try to insert a download result on a request that is not in a `Downloading` state.

This is only relevant in practice if we receive a `UnknownParentBlob` first and a `UnknownParentBlock` later for the same block root. Since `UnknownParentBlob` triggers a request for its block the optimization is to "cancel" or ignore that request once you receive `UnknownParentBlock`. Note that we are spending our (and the serving peer's) bandwidth anyway. It's just a timing optimization to send the block for verification earlier potentially.

dapplion:
Bump: Added a screen recording of mine going over the rationale and top level changes (26 min) link above in the main PR body

beacon_node/network/src/sync/block_lookups/single_block_lookup.rs
dapplion: Added a log in e8fc4825afcb5baa40f48ba3200921b1471f4df6 to track and measure the times we ignore a block component

realbigsean: right gotcha

realbigsean: gotcha

beacon_node/network/src/sync/block_lookups/mod.rs
realbigsean: created an issue for this https://github.com/sigp/lighthouse/issues/5667

beacon_node/network/src/sync/block_lookups/tests.rs
realbigsean: it makes sense that we wouldn't have a block req id because the block was included in unknown parent error, so no request was ever triggered. i added a fallback to the blob req id here:

4726cbd9819dc5b0484d23d82514637baa2ab4ef

Also seems like there was a duplicate blobs response in the test, which didn't look like what it was trying to test so i removed the second

beacon_node/network/src/sync/block_lookups/mod.rs
realbigsean: Thanks for this description, i had my hand at writing a comment. does this seem accurate to you? 3544fbde94bba56a8560b5223a027803383589b2

realbigsean (approved):
This looks good to go from my POV. Awesome job lion!

realbigsean:
@mergify queue

Issue #5602, opened 2024-04-18
Scoring issue


sauliusgrigaitis:
## Description

Seems there is an issue that Lighthouse incorrectly penalizes node that sends blobs via gossip later than Lighthouse gets it from RPC.

## Version

5.1.3

## Present Behaviour
```
Apr 18 06:19:27.073 DEBG Sending BlobsByRoot Request, lookup_type: Current, peer: 16Uiu2HAm67zXpZL32xYBTEptF6jfkykuinRJWeouLj9vMhPeX4zF, blob_indices: [0], block_root: 0x7e51083501c2beda888e93f0bb0038f0658cf469dc99593f5f807f1515618783, method: BlobsByRoot, service: sync, module: network::sync::network_context:490
Apr 18 06:19:27.268 DEBG Peer returned response for single lookup, response_type: Blob, block_root: 0x7e51083501c2beda888e93f0bb0038f0658cf469dc99593f5f807f1515618783, id: 2404, peer_id: 16Uiu2HAm67zXpZL32xYBTEptF6jfkykuinRJWeouLj9vMhPeX4zF, service: sync, module: network::sync::block_lookups:312
Apr 18 06:19:27.268 DEBG RPC blobs received, commitments: [0xad8d…38d4], slot: 1459897, block_root: 0x7e51…8783, indices: [0], module: network::network_beacon_processor::sync_methods:241
Apr 18 06:19:27.293 DEBG Simulated attestation evaluated, head_hit: true, target_hit: true, source_hit: true, attestation_slot: Slot(1459880), attestation_head: 0xa835baf1692cbf23b15dc59b6d6dc44ab20891bf2ccd5da3af15f46d2947351a, attestation_target: 0x55ebc094fb13bef949fdf7686b33435cc1c9e6c8e419badd14ffca8c67e2328a, attestation_source: 0x42e5943fc31651d5c6424fc26a5aaa0dc73ee6d7602d06ec17955958ba96667b, service: val_mon, service: beacon, module: beacon_chain::validator_monitor:2050
Apr 18 06:19:27.303 DEBG Writing blobs to store, count: 3, block_root: 0x7e51…8783, service: beacon, module: beacon_chain::beacon_chain:3467
Apr 18 06:19:27.889 DEBG Successfully verified gossip blob, commitment: 0xad8d…38d4, index: 0, root: 0x7e51…8783, slot: 1459897, module: network::network_beacon_processor::gossip_methods:641
[...]

PR #5656, opened 2024-04-27
Ignore gossip blob already imported

dapplion:
`process_gossip_blob` returns an error if the blob is already imported

https://github.com/sigp/lighthouse/blob/3058b96f2560f1da04ada4f9d8ba8e5651794ff6/beacon_node/beacon_chain/src/beacon_chain.rs#L2884-L2889

The gossip blob handler treats that error as a penalizable offense.

Instead, we should ignore the blob and do nothing

Closes https://github.com/sigp/lighthouse/issues/5602

#### TODO / Note

This problem stems from re-using `BlockError` for gossip import, where there are many variants of the enum that do not apply. Maybe we should use a dedicated BlobImportError and explicitly handle each case. This patch addresses https://github.com/sigp/lighthouse/issues/5602 but there are some extra errors that can happen within the `availability_cache` that are internal errors and should not result in downscoring.

### DIFF
--- a/beacon_node/network/src/network_beacon_processor/gossip_methods.rs
+++ b/beacon_node/network/src/network_beacon_processor/gossip_methods.rs
@@ -778,23 +778,27 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                     "block_root" => %block_root,
                 );
             }
+            Err(BlockError::BlockIsAlreadyKnown(_)) => {
+                debug!(
+                    self.log,
+                    "Ignoring gossip blob already imported";
+                    "block_root" => ?block_root,
+                    "blob_index" =>  blob_index,
+                );
+            }
             Err(err) => {
                 debug!(
                     self.log,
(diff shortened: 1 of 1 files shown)


realbigsean:
> This problem stems from re-using BlockError for gossip import, where there are many variants of the enum that do not apply. Maybe we should use a dedicated BlobImportError and explicitly handle each case. 

I think @pawanjay176 mentioned he started some work on this?

realbigsean (approved):

realbigsean:
@mergify queue

pawanjay176:
Yeah I had started some work on it. It would involve changing the result type of `BeaconChain::import_block` into a `Result<Hash256, BlockComponentError<T::EthSpec>>` and then defining a 
```
enum BlockComponentError<E: EthSpec> {
    Block(BlockError<E>),
    Blob(BlobError<E>),
}
```
and then propagate the error types upwards.

PR #5657, opened 2024-04-28
EIP7549 `get_attestation_indices`

eserilev:
## Issue Addressed

#5605

## Proposed Changes

Implement the electra variant of `get_attesting_indices`. Refactored `map_attestation_committee` to run the `map_fn` over a list of committees for electra compatibility.

I've also removed attestation aggregation logic for electra attestations. computing the on chain aggregate will require some non trivial changes to the op pool. in order to be ready for dev net, this PR aims to get all tests to pass. this does mean we might end up constructing some suboptimal blocks, but for devnet purposes I think that should be ok.


consensus/types/src/eth_spec.rs
```diff
@@ -63,7 +63,7 @@ pub trait EthSpec:
      * Misc
      */
     type MaxValidatorsPerCommittee: Unsigned + Clone + Sync + Send + Debug + PartialEq + Eq;
-    type MaxValidatorsPerCommitteePerSlot: Unsigned + Clone + Sync + Send + Debug + PartialEq + Eq;
+    type MaxValidatorsPerSlot: Unsigned + Clone + Sync + Send + Debug + PartialEq + Eq;
     type MaxCommitteesPerSlot: Unsigned + Clone + Sync + Send + Debug + PartialEq + Eq;
```
ethDreamer, line 67: big fan of the name change.. had this on my list of stuff to get to

beacon_node/beacon_chain/src/observed_aggregates.rs
```diff
@@ -103,29 +100,39 @@ pub trait SubsetItem {
 }
 
 impl<'a, E: EthSpec> SubsetItem for AttestationRef<'a, E> {
-    type Item = BitList<E::MaxValidatorsPerCommittee>;
+    type Item = BitList<E::MaxValidatorsPerSlot>;
     fn is_subset(&self, other: &Self::Item) -> bool {
         match self {
```
ethDreamer, line 105: Is this supposed to calculate `subset`/`superset` between attestations across forks (i.e. an `electra` attestation and a `deneb` attestation that's been extended)? Because that won't work the way `extend_aggregation_bits()` is written

PR #5658, opened 2024-04-29
Report RPC Errors to the application on peer disconnections

AgeManning:
As we are overhaling some internal RPC infrastructure, a desired feature is to report peer disconnects on RPC requests. 

This PR should report an RPCError(Disconnected) if a connection is terminated whilst an RPC request is underway.

### DIFF
--- a/beacon_node/lighthouse_network/tests/rpc_tests.rs
+++ b/beacon_node/lighthouse_network/tests/rpc_tests.rs
@@ -996,6 +996,113 @@ fn test_tcp_blocks_by_root_chunked_rpc_terminates_correctly() {
     })
 }
 
+#[test]
+fn test_disconnect_triggers_rpc_error() {
+    // set up the logging. The level and enabled logging or not
+    let log_level = Level::Debug;
+    let enable_logging = true;
+
+    let log = common::build_log(log_level, enable_logging);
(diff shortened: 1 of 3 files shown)


dapplion (approved):
Thank you! `expect` usage is ok

beacon_node/lighthouse_network/tests/rpc_tests.rs
```diff
@@ -996,6 +996,113 @@ fn test_tcp_blocks_by_root_chunked_rpc_terminates_correctly() {
...
+                        debug!(log, "Sending RPC");
+                        sender.send_request(peer_id, 10, rpc_request.clone());
+                    }
+                    NetworkEvent::ResponseReceived {
+                        peer_id: _, id: 10, ..
+                    } => {
+                        debug!(log, "Sender received a response");
+                    }
+                    NetworkEvent::RPCFailed { id, peer_id, error } => {
+                        panic!("got RPCError");
```
pawanjay176, line 1054: We should return true from the future if we entered this branch and only pass the test in that case.

pawanjay176 (changes requested):
Nice.

dapplion:
- Closing for https://github.com/sigp/lighthouse/pull/5680

PR #5659, opened 2024-04-29
Beta compiler fix

realbigsean:
## Issue Addressed

- rename guards in the `LoggingLayer` struct so they aren't flagged as unused
- there are some errors including unnamed fields that we log out but don't use otherwise. I added `#[allow(dead_code)]` to the unnamed fields
- `BlobRequestState`'s `blob_download_queue` is no longer used since https://github.com/sigp/lighthouse/pull/5583 but is now being flagged as unused, so I removed that

### DIFF
--- a/common/logging/src/lib.rs
+++ b/common/logging/src/lib.rs
@@ -256,14 +256,14 @@ pub fn create_tracing_layer(base_tracing_log_path: PathBuf) {
         return;
     };
 
-    let (libp2p_non_blocking_writer, libp2p_guard) = NonBlocking::new(libp2p_writer);
-    let (discv5_non_blocking_writer, discv5_guard) = NonBlocking::new(discv5_writer);
+    let (libp2p_non_blocking_writer, _libp2p_guard) = NonBlocking::new(libp2p_writer);
+    let (discv5_non_blocking_writer, _discv5_guard) = NonBlocking::new(discv5_writer);
 
     let custom_layer = LoggingLayer {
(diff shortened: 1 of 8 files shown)


common/logging/src/lib.rs
```diff
@@ -256,14 +256,12 @@ pub fn create_tracing_layer(base_tracing_log_path: PathBuf) {
         return;
     };
 
-    let (libp2p_non_blocking_writer, libp2p_guard) = NonBlocking::new(libp2p_writer);
-    let (discv5_non_blocking_writer, discv5_guard) = NonBlocking::new(discv5_writer);
+    let (libp2p_non_blocking_writer, _) = NonBlocking::new(libp2p_writer);
```
pawanjay176: I think we need the guards for the logging to actually happen. From https://docs.rs/tracing-appender/latest/tracing_appender/non_blocking/index.html

> Note that the WorkerGuard returned by non_blocking must be assigned to a binding that is not _, as _ will result in the WorkerGuard being dropped immediately. Unintentional drops of WorkerGuard remove the guarantee that logs will be flushed during a program’s termination, in a panic or otherwise.

pawanjay176 (changes requested):

common/logging/src/lib.rs
realbigsean: 💯  nice catch, added back

pawanjay176 (approved):

realbigsean:
@mergify queue

PR #5661, opened 2024-04-30
Remove snapshot cache related code

jimmygchen:
## Issue Addressed

Some remaining review comments from #5533.


beacon_node/src/cli.rs
```diff
@@ -624,7 +624,7 @@ pub fn cli_app<'a, 'b>() -> App<'a, 'b> {
             Arg::with_name("state-cache-size")
                 .long("state-cache-size")
                 .value_name("STATE_CACHE_SIZE")
-                .help("Specifies the size of the snapshot cache [default: 3]")
+                .help("Specifies the size of the state cache [default: 3]")
```
```suggestion
                .help("Specifies the size of the state cache [default: 128]")
```

book/src/help_bn.md
```diff
@@ -438,7 +438,7 @@ OPTIONS:
             Specifies how often a freezer DB restore point should be stored. Cannot be changed after initialization.
             [default: 8192 (mainnet) or 64 (minimal)]
         --state-cache-size <STATE_CACHE_SIZE>
-            Specifies the size of the snapshot cache [default: 3]
+            Specifies the size of the state cache [default: 3]
```
michaelsproul:
```suggestion
            Specifies the size of the state cache [default: 128]
```

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -4397,12 +4397,6 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             if cached_head.head_block_root() == parent_block_root {
                 (Cow::Borrowed(head_state), cached_head.head_state_root())
             } else {
-                info!(
-                    self.log,
-                    "Missed snapshot cache during withdrawals calculation";
-                    "slot" => proposal_slot,
-                    "parent_block_root" => ?parent_block_root
-                );
```
michaelsproul, line 4405: I deleted this as it wasn't accurate. The state cache kicks in below when we call `get_advanced_hot_state`. If it misses, it will log a warning from the store.

beacon_node/src/cli.rs
jimmygchen: Good catch!

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Oh yeah, thanks!

michaelsproul:
Pushed some tests for the `--state-cache` flag as I realised we didn't have any.

michaelsproul:
@mergifyio queue

jimmygchen (commented):
Nice, thanks for adding the missing tests!

jimmygchen:
Oh looks like it’s missing a green tick before it can be added to the merge queue!

realbigsean (approved):

realbigsean:
@mergify queue

PR #5665, opened 2024-04-30
pin macos release runner to `macos-13`

antondlr:
## Issue Addressed

the `macOS-latest` runners have switched over to ARM based hardware, which will result in a failed release pipeline. 
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -52,7 +52,7 @@ jobs:
                         runner: ${{ github.repository == 'sigp/lighthouse' && fromJson('["self-hosted", "linux", "release", "large"]') || 'ubuntu-latest'  }}
                         profile: maxperf
                     -   arch: x86_64-apple-darwin
-                        runner: macos-latest
+                        runner: macos-13
                         profile: maxperf
                     -   arch: x86_64-apple-darwin-portable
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -52,7 +52,7 @@ jobs:
                         runner: ${{ github.repository == 'sigp/lighthouse' && fromJson('["self-hosted", "linux", "release", "large"]') || 'ubuntu-latest'  }}
                         profile: maxperf
                     -   arch: x86_64-apple-darwin
-                        runner: macos-latest
+                        runner: macos-13
                         profile: maxperf
                     -   arch: x86_64-apple-darwin-portable
                         runner: macos-latest
```
jimmygchen: Another one below:

```suggestion
                        runner: macos-13
```

antondlr: oh yeah good catch. I was on a branch that had #5615 merged.

jimmygchen (approved):
LGTM!

jimmygchen:
@mergify queue

PR #5670, opened 2024-04-30
Update Cargo.lock

jxs:
And remove no longer required patch'ed `rust-yamux`


michaelsproul:
I think we probably can include this in v5.2.0 now, now that we're a bit delayed anyway.

michaelsproul (approved):

michaelsproul:
@mergifyio queue

PR #5672, opened 2024-04-30
 delete spammy log

realbigsean:
## Issue Addressed

This log happens once per attestation so is very spammy. Don't think it's worth having for tracing either.

`Searching for block components` currently is only ever triggered by this code path, if there's an unknown parent triggering the lookup we would get `Searching for block components`, so we aren't losing info here


pawanjay176 (approved):

realbigsean:
@mergify queue

PR #5675, opened 2024-04-30
Ensure block only range requests don't fail on download

realbigsean:
## Issue Addressed

I noticed backfill getting stuck:
```
Apr 30 16:18:13.313 WARN Blocks and blobs request for range received invalid data, error: MissingBlobs, sender_id: BackfillSync { batch_id: Epoch(40840) }, peer_id: 16Uiu2HAm99wjuFBKfbnnvAE3Bcpqy29s6SCNgreVo3pYvfxnbXha, service: sync, module: network::sync::manager:933
```

I think in this PR https://github.com/sigp/lighthouse/pull/5561/files, when we starting using the same response handling for block + blob and block only range requests, we started creating `RpcBlock`s with empty lists of blobs in block only responses. If a list of blobs is passed into `RpcBlock`, it'll check its length matches the kzg commitments in a block, which would throw an error for blocks with pruned blobs. This PR makes it so `RpcBlock` will treat an empty list of blobs as if no blobs at all were present. The check for whether blobs *should* exist happens elsewhere, during the processing of `RpcBlock`.

### DIFF
--- a/beacon_node/beacon_chain/src/block_verification_types.rs
+++ b/beacon_node/beacon_chain/src/block_verification_types.rs
@@ -97,7 +97,7 @@ impl<E: EthSpec> RpcBlock<E> {
 
     /// Constructs a new `BlockAndBlobs` variant after making consistency
     /// checks between the provided blocks and blobs. This struct makes no
-    /// guaruntees about whether blobs should be present, only that they are
+    /// guarantees about whether blobs should be present, only that they are
     /// consistent with the block. An empty list passed in for `blobs` is
     /// viewed the same as `None` passed in.
     pub fn new(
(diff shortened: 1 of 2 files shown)


realbigsean:
Alternatively I'm wondering if this should work by trying to match block + blobs and just dropping blobs if there is any mismatch? and then we just fail later in processing when we check if we need blobs?

dapplion (approved):
Nice find! Fix LGTM

> I'm wondering if this should work by trying to match block + blobs and just dropping blobs if there is any mismatch?

That's a good idea to make the sync more dumb and delegate to the da_checker. Tho best to explore after 5.2.0

jimmygchen:
@mergify queue

PR #5680, opened 2024-05-01
Report RPC Errors to the application on peer disconnections

dapplion:
## Issue Addressed

Extends
- https://github.com/sigp/lighthouse/pull/5658

As we are overhaling some internal RPC infrastructure, a desired feature is to report peer disconnects on RPC requests.

This PR should report an RPCError(Disconnected) if a connection is terminated whilst an RPC request is underway.

## Proposed Changes

- Emit `RPCError::Disconnect` to any outbound streams with a disconnecting peer
- Remove sync code that fails download attempts of disconnected peers, and expect `inject_error` to handle it

### DIFF
--- a/beacon_node/lighthouse_network/src/service/mod.rs
+++ b/beacon_node/lighthouse_network/src/service/mod.rs
@@ -1379,7 +1379,7 @@ impl<AppReqId: ReqId, E: EthSpec> Network<AppReqId, E> {
         let peer_id = event.peer_id;
 
         if !self.peer_manager().is_connected(&peer_id) {
-            // Sync expectes a RPCError::Disconnected to drop associated lookups with this peer.
+            // Sync expects a RPCError::Disconnected to drop associated lookups with this peer.
             // Silencing this event breaks the API contract with RPC where every request ends with
             // - A stream termination event, or
             // - An RPCError event
--- a/beacon_node/lighthouse_network/src/service/mod.rs
+++ b/beacon_node/lighthouse_network/src/service/mod.rs
@@ -972,6 +972,12 @@ impl<AppReqId: ReqId, E: EthSpec> Network<AppReqId, E> {
             .goodbye_peer(peer_id, reason, source);
     }
 
+    /// Hard (ungraceful) disconnect for testing purposes only
+    /// Use goodbye_peer for disconnections, do not use this function.
+    pub fn __hard_disconnect_testing_only(&mut self, peer_id: PeerId) {
+        let _ = self.swarm.disconnect_peer_id(peer_id);
+    }
+
     /// Returns an iterator over all enr entries in the DHT.
     pub fn enr_entries(&self) -> Vec<Enr> {
         self.discovery().table_entries_enr()
--- a/beacon_node/lighthouse_network/src/service/mod.rs
+++ b/beacon_node/lighthouse_network/src/service/mod.rs
@@ -1379,12 +1379,18 @@ impl<AppReqId: ReqId, E: EthSpec> Network<AppReqId, E> {
         let peer_id = event.peer_id;
 
         if !self.peer_manager().is_connected(&peer_id) {
-            debug!(
-                self.log,
-                "Ignoring rpc message of disconnecting peer";
-                event
-            );
-            return None;
+            // Sync expectes a RPCError::Disconnected to drop associated lookups with this peer.
+            // Silencing this event breaks the API contract with RPC where every request ends with
+            // - A stream termination event, or
+            // - An RPCError event
+            if !matches!(event.event, HandlerEvent::Err(HandlerErr::Outbound { .. })) {
+                debug!(
+                    self.log,
+                    "Ignoring rpc message of disconnecting peer";
+                    event
+                );
+                return None;
+            }
         }
 
         let handler_id = event.conn_id;
--- a/beacon_node/lighthouse_network/tests/rpc_tests.rs
+++ b/beacon_node/lighthouse_network/tests/rpc_tests.rs
@@ -3,7 +3,7 @@
 mod common;
 
 use common::Protocol;
-use lighthouse_network::rpc::methods::*;
+use lighthouse_network::rpc::{methods::*, RPCError};
 use lighthouse_network::{rpc::max_rpc_size, NetworkEvent, ReportSource, Request, Response};
 use slog::{debug, warn, Level};
 use ssz::Encode;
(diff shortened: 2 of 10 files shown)


beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -307,49 +307,15 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
...
-    ) -> Result<(), BackFillError> {
+    pub fn peer_disconnected(&mut self, peer_id: &PeerId) -> Result<(), BackFillError> {
         if matches!(
             self.state(),
             BackFillState::Failed | BackFillState::NotRequired
         ) {
             return Ok(());
         }
 
-        if let Some(batch_ids) = self.active_requests.remove(peer_id) {
```
pawanjay176, line 322: This is a great simplification. The repeated logic was a source of many seen/unseen bugs earlier. Kudos for having the big picture and spotting that this can be removed 🙌 

My only concern with this is that in `inject_error`, we call `batch.download_failed(true)` instead of `false` which we shouldn't be doing for disconnections maybe? Repeated peer disconnections for the same batch  might end up marking the entire chain as invalid and redoing a bunch of stuff.

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -174,30 +174,8 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
...
-        &mut self,
-        peer_id: &PeerId,
-        network: &mut SyncNetworkContext<T>,
-    ) -> ProcessingResult {
-        if let Some(batch_ids) = self.peers.remove(peer_id) {
-            // fail the batches
-            for id in batch_ids {
-                if let Some(batch) = self.batches.get_mut(&id) {
-                    if let BatchOperationOutcome::Failed { blacklist } =
-                        batch.download_failed(true)?
```
pawanjay176, line 187: similar comment as above regarding marking the batch as failed/not failed.

In forward sync, this might mean potentially not retrying a valid chain because the peers on the good chain are disconnecting.

beacon_node/lighthouse_network/tests/rpc_tests.rs
```diff
@@ -996,6 +996,96 @@ fn test_tcp_blocks_by_root_chunked_rpc_terminates_correctly() {
...
+        let sender_future = async {
+            loop {
+                match sender.next_event().await {
+                    NetworkEvent::PeerConnectedOutgoing(peer_id) => {
+                        // Send a STATUS message
+                        debug!(log, "Sending RPC");
+                        sender.send_request(peer_id, 42, rpc_request.clone());
+                    }
+                    NetworkEvent::RPCFailed { error, id: 42, .. } => match error {
+                        RPCError::Disconnected => return,
```
pawanjay176, line 1037: the test should make sure we only get to this branch

pawanjay176 (commented):
This is a great catch, I just have a question

beacon_node/network/src/sync/backfill_sync/mod.rs
AgeManning: Good point. Should also be noted that these disconnects are ungraceful. 

i.e When lighthouse disconnects from a peer, it will wait to try and fulfill all its requests. It wont just drop the connection. In fact, a stream timeout will occur before a disconnection in a graceful disconnect. 

The error peer disconnect should only happen when a peer drops the connection without fulfilling a request (which lighthouse doesn't do unless there is a network error).

AgeManning (approved):
I did a quick look. This looks good to me. I like the simplification. 

The errors that we now see are from ungraceful disconnects, which probably should be punished and treated like an RPC error, imo

beacon_node/network/src/sync/backfill_sync/mod.rs
dapplion: Right, current stable assumes that peer disconnection == failed download. But this `RPCError::Disconnect` error will only fire if there is an active outgoing request that gets terminated ungracefully. 

> Repeated peer disconnections for the same batch might end up marking the entire chain as invalid and redoing a bunch of stuff.

Should only apply if:
- we initiate request to peer A
- peer A disconnects ungracefully before completing request
- we initiate a retry request to peer B
- peer B disconnects ungracefully before completing request

This should not happen frequently

beacon_node/lighthouse_network/tests/rpc_tests.rs
dapplion: Since this branch is the only way to break out of the loop, not hitting this branch will timeout the test. I considered adding something explicit but it feels redundant

dapplion:
@realbigsean noted that a lookup can get stuck if it has no available peers and is awaiting a download. This case should never happen with current code as a lookup is never left in AwaitingDownload state. However, for completeness I have added a check to drop lookups in that case in https://github.com/sigp/lighthouse/pull/5680/commits/02f1b2d53198fc1f0b1e7d31a1a7dd5fe45346eb

beacon_node/network/src/sync/backfill_sync/mod.rs
pawanjay176: fair enough

beacon_node/lighthouse_network/tests/rpc_tests.rs
pawanjay176: right.

pawanjay176 (approved):

dapplion:
The RPCError events are never received by sync due to this condition here
 https://github.com/sigp/lighthouse/blob/3058b96f2560f1da04ada4f9d8ba8e5651794ff6/beacon_node/lighthouse_network/src/service/mod.rs#L1370-L1377
 
The tests on network/sync/block_lookups and lighthouse_network pass respectively as they don't test the full integration.

@AgeManning there's a lot of code in this function that is currently not expecting events for disconnected peers. Would be best to just allow events for disconnected peers if the event type if RPCError::Disconnect?

AgeManning (approved):
The latest changes allowing the RpcError::Disconnected to propagate, looks good to me. 

To be clear, this edge case happens where:
1. We make a request
2. The peer disconnects ungracefully
3. There is a race between receiving the RpcError::Disconnected from the rpc handler and the Swarm peer-discconected message. Potentially we always lose this race, and the peer manager considers the peer disconnected before we can read the final handler message. 

The only issue I see here, is that it's going to break a previously intuitive construct we previously had, which was that the last log/message we ever see from a peer is "Peer Disconnected". 

After this change, we can see logs and messages after the peer has disconnected. 

i.e 
"Peer Disconnected"
"RPC Error::Disconnected"

I don't immediately see a solution to this, because the ordering is coming from the swarm which we don't have much control over here.

beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -1373,12 +1379,18 @@ impl<AppReqId: ReqId, E: EthSpec> Network<AppReqId, E> {
         let peer_id = event.peer_id;
 
         if !self.peer_manager().is_connected(&peer_id) {
-            debug!(
-                self.log,
-                "Ignoring rpc message of disconnecting peer";
-                event
-            );
-            return None;
+            // Sync expectes a RPCError::Disconnected to drop associated lookups with this peer.
```
AgeManning:
```suggestion
            // Sync expects a RPCError::Disconnected to drop associated lookups with this peer.
```
I can has airdrop? ;)

realbigsean:
@mergify queue

PR #5681, opened 2024-05-01
Check da_checker before doing a block lookup request

dapplion:
## Issue Addressed

Testing of v5.2.0 shows many more `rpc_block` than usual. This is due this sequence of events:

- Block is known to the da_checker but not yet imported into fork-choice
- The event `UnknownBlockHashFromAttestation` is sent at least once every block (this is likely a bug and should be fixed separately)
- A new lookup is created for the block, downloads block, and sends for processing

In current stable the single lookup checks against the da_checker if the block is known before sending a request, but after 
- https://github.com/sigp/lighthouse/pull/5655

this check was dropped by mistake

## Proposed Changes

- Add a check in `SyncNetworkContext::block_lookup_request` and skip the download if the block is already known

However, now we need to handle the case where a lookup goes directly from created -> completed. This takes a bit of refactoring, and to handle this gracefully and safely for all cases I have introduced the `LookupResult` with a `must_use` directive.

### DIFF
--- a/beacon_node/network/src/sync/network_context.rs
+++ b/beacon_node/network/src/sync/network_context.rs
@@ -298,6 +298,8 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
 
     /// Request block of `block_root` if necessary by checking:
     /// - If the da_checker has a pending block from gossip or a previous request
+    ///
+    /// Returns false if no request was made, because the block is already imported
     pub fn block_lookup_request(
         &mut self,
         lookup_id: SingleLookupId,
@@ -341,7 +343,7 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
     /// - If the da_checker has a pending block
     /// - If the da_checker has pending blobs from gossip
     ///
-    /// Returns false if no request was made, because we don't need to fetch (more) blobs.
+    /// Returns false if no request was made, because we don't need to import (more) blobs.
     pub fn blob_lookup_request(
         &mut self,
         lookup_id: SingleLookupId,
--- a/beacon_node/network/src/sync/network_context.rs
+++ b/beacon_node/network/src/sync/network_context.rs
@@ -306,7 +306,12 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
         peer_id: PeerId,
         block_root: Hash256,
     ) -> Result<bool, &'static str> {
-        if self.chain.data_availability_checker.has_block(&block_root) {
+        if self
+            .chain
+            .reqresp_pre_import_cache
+            .read()
+            .contains_key(&block_root)
+        {
             return Ok(false);
         }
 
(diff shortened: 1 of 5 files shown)


beacon_node/network/src/sync/network_context.rs
```diff
@@ -296,12 +296,20 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+    ///
+    /// Returns false if no request was made, because the block is already imported
     pub fn block_lookup_request(
         &mut self,
         lookup_id: SingleLookupId,
         peer_id: PeerId,
-        request: BlocksByRootSingleRequest,
+        block_root: Hash256,
     ) -> Result<bool, &'static str> {
+        if self.chain.data_availability_checker.has_block(&block_root) {
```
realbigsean: don't we want to use the `reqresp_pre_import_cache`? Otherwise we'll miss blocks that we're in the process of verifying execution

realbigsean: the check in `stable` is essentially this because `data_availability_checker.has_block()` used to check the processing cache

realbigsean: lion agreed offline, so I've pushed the change here: https://github.com/sigp/lighthouse/pull/5681/commits/5db3e1b5e5ed91886cad9931b6905282c40fe858

realbigsean (approved):

realbigsean:
@mergify queue

beacon_node/network/src/sync/network_context.rs
```diff
@@ -296,12 +296,25 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+        block_root: Hash256,
     ) -> Result<bool, &'static str> {
+        if self
+            .chain
+            .reqresp_pre_import_cache
+            .read()
+            .contains_key(&block_root)
+        {
+            return Ok(false);
+        }
```
michaelsproul, line 316: I think this is the cause of the recent de-sync issues.

Consider the following sequence of events:

1. Block arrives on gossip, we add it to the `reqresp_pre_import_cache`.
2. We send the block's payload to the EL for verification.
3. We receive blobs on gossip and add them to the DA cache.
4. The EL _times out_. Block processing is aborted as a result.
5. Subsequent attempts to look up the block and re-process it are blocked by the entry in the `reqresp_pre_import_cache`.
6. We stall and are unable to reprocess the block until range sync and/or a blocks by range request is made in spite of the check above.

I've confirmed that this happened for block 0xfc4a1afe39d84a410c61f213cf15cb405fdd15b96ca54f0c826d4054671b1275 at slot 1574856 on Holesky. We get the block and all the blobs, then the EL times out and everything gets stuck:

> May 04 05:31:20.803 DEBG Failed to verify execution payload      error: ExecutionPayloadError(RequestFailed(EngineError(Api { error: HttpClient(url: http://127.0.0.1:8551/, kind: timeout, detail: operation timed out) })))

michaelsproul: Actually scratch that, @jimmygchen pointed out that we remove from the `reqresp_pre_import_cache` on failure here:

https://github.com/sigp/lighthouse/blob/63fad7e330c4559474243346de7bc7389f6541d0/beacon_node/beacon_chain/src/beacon_chain.rs#L2945-L2950

PR #5682, opened 2024-05-01
Ignore unknown block root events for processing blocks

dapplion:
## Issue Addressed

From testing on v5.2.0 of Sepolia and Gnosis there is a very large volume of `UnknownBlockHashFromAttestation` and one single block lookup created every slot.

`UnknownBlockHashFromAttestation` errors are triggered if the block is not on fork-choice, but we may have the block stuck somewhere in the processing pipeline.

## Proposed Changes

This PR attempts to prevent `UnknownBlockHashFromAttestation` events for blocks that are not yet on fork-choice but we are processing.

## Caveats

If a single peer imports the block before we receive it and sends us an attestation we will still have the problem of constant lookups for blocks that arrive anyway. Not sure if this is an issue at all, but worth to think about it.

We have an ample collection of places a block may be stuck into:
- database
- fork-choice
- early attester cache
- da_checker
- pre_import_cache (former processing_cache)
- duplicate_cache

ReqResp lookups only check the da_checker (from https://github.com/sigp/lighthouse/pull/5681) while other codepaths check more places. Might be worth it to consolidate their usage a bit to have more consistency.

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -1146,6 +1146,15 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         Ok(BeaconBlockStreamer::<T>::new(self, CheckCaches::No)?.launch_stream(block_roots))
     }
 
+    /// Returns true if the given block_root is known but has not yet been imported. Checks caches
+    /// only, and does not check fork-choice nor the database.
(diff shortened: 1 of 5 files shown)


dapplion:
Does not apply anymore, fixed by the debounce of similar unknown attestation events and 

https://github.com/sigp/lighthouse/blob/2e0eb6d1b8705bbda2ba56eb195d9cc7c6575e95/beacon_node/network/src/sync/network_context.rs#L387-L403

PR #5684, opened 2024-05-01
PeerDAS custody lookup sync

dapplion:
Implements the patterns from
- https://github.com/sigp/lighthouse/pull/5616

For custody: 
- [x] lookup sync
- [ ] range sync

## Proposed changes

- Adds two new types to clearly differentiate data columns for sampling and data columns for custody `CustodyDataColumn`, and `KzgVerifiedCustodyDataColumn`. This distinction is important because the chain has no notion of which specific columns we have to custody. We don't want to mistakenly import a column from sampling.

- Add new processor event `Work::RpcCustodyColumn`, which verifies and imports data columns. `Work::RpcVerifyDataColumn` only verifies then

- Lookup requests track a `PeerGroup` now instead of just a single `PeerId`. Necessary now that custody requests are multi-peer

- Update `blob_lookup_request` to not issue requests if we are into peerdas. Currently it copied the same logic from stable where it checks against the clock epoch. This seems very dangerous around forks IMO. We might be already into peerdas fork but issue a lookup request for a block that still needs blobs. **TODO**: fix in future PR, but it would require delaying column fetching until we have downloaded the block on a lookup.

- Implements custody columns for lookups in the SyncNetworkContext, tracked in `custody_by_root_requests` as a `ActiveCustodyRequest` which can make multiple data columns by root requests.

- Add a single `custody_lookup_happy_path` that ensures custody request succeeds and no blobs are requested


dapplion:
Merging as is with green CI. @jimmygchen agreed to review in follow-up PR if we need changes

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -345,7 +331,17 @@ impl<E: EthSpec> PendingComponents<E> {
...
+            // TODO(das) Do we need a check here for number of expected custody columns?
+            // TODO(das): Update store types to prevent this conversion
+            data_columns: Some(
+                VariableList::new(
+                    verified_data_columns
+                        .into_iter()
+                        .map(|d| d.into_inner())
+                        .collect(),
+                )
+                .expect("data column list is within bounds"),
```
jimmygchen, line 343: This would be an unexpected scenario due to code error. Agree we should update the store types! 

I think we can `?` this to handle unexpected errors consistently.

dapplion: Hopefully we can update the store types today or tomorrow :)

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3074,6 +3076,33 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+        // we don't need to process its blobs again.
+        if self
+            .canonical_head
+            .fork_choice_read_lock()
+            .contains_block(&block_root)
+        {
+            return Err(BlockError::BlockIsAlreadyKnown(block_root));
+        }
+
+        // TODO(das): custody column SSE event
```
jimmygchen, line 3096: Do we have a use case for this? is it mostly for monitoring ?

beacon_node/beacon_chain/src/block_verification_types.rs
```diff
@@ -86,7 +88,10 @@ enum RpcBlockInner<E: EthSpec> {
     BlockAndBlobs(Arc<SignedBeaconBlock<E>>, BlobSidecarList<E>),
     /// This variant is used with parent lookups and by-range responses. It should have all
     /// requested data columns, all block roots matching for this block.
-    BlockAndDataColumns(Arc<SignedBeaconBlock<E>>, DataColumnSidecarList<E>),
+    BlockAndCustodyColumns(
+        Arc<SignedBeaconBlock<E>>,
+        VariableList<CustodyDataColumn<E>, <E as EthSpec>::DataColumnCount>,
```
jimmygchen, line 93: `CustodyDataColumnList<E>`

beacon_node/beacon_chain/src/block_verification_types.rs
```diff
@@ -144,6 +149,35 @@ impl<E: EthSpec> RpcBlock<E> {
...
+        let block_root = block_root.unwrap_or_else(|| get_block_root(&block));
+
+        if let Ok(block_commitments) = block.message().body().blob_kzg_commitments() {
+            // The number of required custody columns is out of scope here.
+            if !block_commitments.is_empty() && custody_columns.is_empty() {
+                return Err(AvailabilityCheckError::MissingCustodyColumns);
+            }
+        }
+        // Treat empty blob lists as if they are missing.
+        let inner = if custody_columns.is_empty() {
```
jimmygchen, line 166: This should be `!custody_columns.is_empty()` but it looks like we can just remove this method as it doesn't seem to be used anywhere?

beacon_node/network/src/network_beacon_processor/mod.rs
```diff
@@ -478,8 +480,32 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
     }
 
+    /// Create a new `Work` event for some custody columns. `process_rpc_custody_columns` reports
+    /// the result back to sync.
+    pub fn send_rpc_custody_columns(
+        self: &Arc<Self>,
+        block_root: Hash256,
+        custody_columns: Vec<CustodyDataColumn<T::EthSpec>>,
+        seen_timestamp: Duration,
+        process_type: BlockProcessType,
```
jimmygchen, line 490: Passing the process type here seems a bit redundant since it's always going to be `SingleDataColumn`, I guess this is remain from parent lookup?

beacon_node/network/src/sync/block_lookups/common.rs
```diff
@@ -171,3 +175,52 @@ impl<T: BeaconChainTypes> RequestState<T> for BlobRequestState<T::EthSpec> {
...
 }
+
+impl<T: BeaconChainTypes> RequestState<T> for CustodyRequestState<T::EthSpec> {
+    type VerifiedResponseType = Vec<CustodyDataColumn<T::EthSpec>>;
+
+    fn make_request(
+        &self,
+        id: Id,
+        // TODO(das): consider selecting peers that have custody but are in this set
+        _peer_id: PeerId,
```
jimmygchen, line 186: Is it worth trying to retrieve from this peer at all? It could send an attestation with unknown block but may not have be custodying the same columns as our node.

beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -306,28 +305,21 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
...
+        response: Result<(R::VerifiedResponseType, PeerGroup, Duration), LookupFailure>,
         cx: &mut SyncNetworkContext<T>,
     ) -> Result<LookupResult, LookupRequestError> {
-        // Downscore peer even if lookup is not known
-        // Only downscore lookup verify errors. RPC errors are downscored in the network handler.
-        if let Err(LookupFailure::LookupVerifyError(e)) = &response {
-            // Note: the error is displayed in full debug form on the match below
-            cx.report_peer(peer_id, PeerAction::LowToleranceError, e.into());
-        }
+        // Note: do not downscore peers here for requests errors, SyncNetworkContext does it.
```
jimmygchen, line 322: We should add some tests to verify the penalties.

beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -341,12 +333,12 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
...
 
         match response {
-            Ok((response, seen_timestamp)) => {
+            Ok((response, peer_group, seen_timestamp)) => {
                 debug!(self.log,
                     "Received lookup download success";
                     "block_root" => ?block_root,
                     "id" => id,
-                    "peer_id" => %peer_id,
+                    "peer_group" => ?peer_group,
```
jimmygchen, line 341: Debug log for peer_group is quite verbose, we could probably implement `Display`:

```
// %peer_id
1AnFkF14ZcbnBiT1UwXGbwhCqenCAhpr88EcdkYuYkFXkC

// ?peer_group
PeerGroup { peers: [PeerId("1AnFkF14ZcbnBiT1UwXGbwhCqenCAhpr88EcdkYuYkFXkC")] }

```

beacon_node/network/src/sync/network_context.rs
```diff
@@ -496,6 +527,100 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
         Ok(())
     }
 
+    pub fn custody_lookup_request(
+        &mut self,
+        lookup_id: SingleLookupId,
+        block_root: Hash256,
+        downloaded_block_expected_data: Option<usize>,
+    ) -> Result<bool, &'static str> {
```
jimmygchen, line 535: Please add some docs for the return values

beacon_node/network/src/sync/network_context.rs
```diff
@@ -496,6 +527,100 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+        }
+
+        let custody_indexes_imported = self
+            .chain
+            .data_availability_checker
+            .imported_custody_column_indexes(&block_root)
+            .unwrap_or_default();
+
+        // TODO(das): figure out how to pass block.slot if we end up doing rotation
+        let block_epoch = Epoch::new(0);
```
jimmygchen, line 574: Feels like we should just remove the epoch from the `custody_columns` function until we need it - adding this dummy value could be a source of confusion / bugs in the future, in case things change or if we log this epoch.

beacon_node/beacon_chain/src/beacon_chain.rs
dapplion: Just a note for future implementors I imagine this is something we want to implement eventually. No usecase tho at the moment, columns are not immediatelly useful for data consumers but can be useful for timing calculations i.e. ef devops

beacon_node/beacon_chain/src/block_verification_types.rs
jimmygchen: Addressed in #5748

beacon_node/beacon_chain/src/block_verification_types.rs
jimmygchen: Method removed in Addressed in #5748

PR #5685, opened 2024-05-01
EIP7549 `compute_on_chain_aggregate`

eserilev:
## Issue Addressed

https://github.com/sigp/lighthouse/issues/5605

## Proposed Changes

Electra Attestations that share equal `AttestationData` will now be aggregated before being submitted on chain. See [spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/validator.md#attestations) for implementation details

## Additional Info

Running through several test cases, I found situations where we had duplicate attestations being included in a beacon block. Aggregating duplicate attestations will cause InvalidSignature errors, so I added an additional check to skip duplicate attestations during the aggregation process.


michaelsproul:
I think this needs to consider the committee_bits:

https://github.com/sigp/lighthouse/blob/3a41e137d126acf95e036cfe9153f303d7f1639d/consensus/types/src/attestation.rs#L225-L230

I think the logic for disjointness should be:

- `committee_bits` are disjoint (fast check), OR
- `committee_bits` overlap, but at all overlapping committees, the `aggregation_bits` for those committees are disjoint

Similarly this `aggregate` function should build the new list of committee bits and append all of the `aggregation_bits` in order:

https://github.com/sigp/lighthouse/blob/cd0f08b7b847d2b61f636223f6064c8169b96c8a/consensus/types/src/attestation.rs#L247-L252

PR #5279, opened 2024-02-23
Single-pass epoch processing and optimised block processing

michaelsproul:
## Proposed Changes

In preparation for merging `tree-states` down, this PR includes all of the consensus-related changes for epoch and block processing.

The epoch processing changes are covered by a (paper) proof here:

- https://ethresear.ch/t/formally-verified-optimised-epoch-processing/17359

We (@onomatic and I) are in the process of formalising this proof in Isabelle/HOL.

The block processing changes are not covered by a proof, and should be reviewed carefully.

## Notes for reviewers

- `ProgressiveBalancesMode` is deprecated because we use the progressive balances cache everywhere, including for calculating justification and finalization. It is no longer optional.
- This PR is a breaking change because the `previous_epoch_active_gwei` field is removed from the `/lighthouse/validator_inclusion/{epoch}/global` endpoint. It never really should have been in there because justification and finalization only cares about the _current_ epoch total active balance.

## Additional Info

Thanks to @dapplion for extracting this from the main `tree-states` branch, solving `FIXME`s and getting it ready for review :tada:

Main tree-states PR:

- https://github.com/sigp/lighthouse/pull/3206

PR #5688, opened 2024-05-02
Add metric for current epoch total balance

michaelsproul:
## Proposed Changes

Add a new metric `beacon_participation_current_epoch_active_gwei_total` to replace the metric `beacon_participation_prev_epoch_active_gwei_total` which was deleted in:

- https://github.com/sigp/lighthouse/pull/5279

The new metric is better for calculating the participation rate because for justification and finalization the total active balance for the current epoch is always used.


jimmygchen (approved):
LGTM!

michaelsproul:
@mergifyio queue

Issue #5693, opened 2024-05-02
Too many `Searching for block components`


dapplion:
**TLDR**: Not actively harmful, just wasted CPU cycles while block and blobs are in the da_checker and we keep getting attestations

- Introduced in https://github.com/sigp/lighthouse/pull/5583

17,901 hits on `ho-h-bn-suit` during 1 hour after `2024 May 2, @ 18:26:30.000` all of them for block `0xf26fcf6a6985185172b27782fda904a25c83a637e7268f0fd878fb9b55e12b7e`

```
May 02 09:26:38.901 DEBG Received unknown block hash message     block_root: 0xf26f…2b7e, service: sync
May 02 09:26:38.902 DEBG Dropping completed lookup               block: 0xf26f…2b7e, service: sync
May 02 09:26:38.902 DEBG Searching for block components          block: 0xf26fcf6a6985185172b27782fda904a25c83a637e7268f0fd878fb9b55e12b7e, peer_ids: [PeerId("16Uiu2HAmUyPcQeJ123zf7RzhqxRd9Berjj9m7x4bNecmQMj2sUQ1")], service: sync
```

The condition to complete the lookup early is hit and it is dropped immediately. Then another unknown block event is processed and repeat

Block and blobs received in time, all 6 blobs. However, execution verification was very slow 4.7 sec.

```
May 02 09:26:37.701 DEBG Successfully verified gossip blob       commitment: 0xb1ce…5bab, index: 0, root: 0xf26f…2b7e, slot: 1561633
May 02 09:26:37.939 INFO New block received                      root: 0xf26fcf6a6985185172b27782fda904a25c83a637e7268f0fd878fb9b55e12b7e, slot: 1561633
May 02 09:26:42.980 DEBG Delayed head block                      set_as_head_time_ms: 32, imported_time_ms: 289, attestable_delay_ms: 6779, available_delay_ms: 6658, execution_time_ms: 4726, blob_delay_ms: 1838, observed_delay_ms: 1931, total_delay_ms: 6979, slot: 1561633, proposer_index: 788787, block_root: 0xf26fcf6a6985185172b27782fda904a25c83a637e7268f0fd878fb9b55e12b7e, service: beacon
May 02 09:26:42.948 DEBG Beacon block imported                   block_slot: 1561633, block_root: 0xf26fcf6a6985185172b27782fda904a25c83a637e7268f0fd878fb9b55e12b7e, service: beacon
```

[...]

PR #5696, opened 2024-05-02
Component search loop

realbigsean:
## Issue Addressed

Addresses: https://github.com/sigp/lighthouse/issues/5693

- Suggestion by @dapplion: Adds a `completed_lookups` time cache, similar to the `failed_chains` cache as a simple way to track and suppress lookups that have already been completed but are still being processed somewhere

- I also realized this PR https://github.com/sigp/lighthouse/pull/5672 would keep us from being able to distinguish current lookups triggered by unknown parents vs attestations easily in the logs, so I added a new enum `LookupTrigger` solely for the purpose of distinguishing the two

### DIFF
--- a/beacon_node/network/src/sync/block_lookups/mod.rs
+++ b/beacon_node/network/src/sync/block_lookups/mod.rs
@@ -237,7 +237,7 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
 
     /// Searches for a single block hash. If the blocks parent is unknown, a chain of blocks is
     /// constructed.
-    /// Returns true if the lookup is created or already exists
+    /// Returns true if the lookup is created, already exists, or was already completed.
     fn new_current_lookup(
         &mut self,
         block_root: Hash256,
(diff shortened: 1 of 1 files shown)


dapplion (approved):

beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -204,20 +225,33 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
...
         block_root: Hash256,
         block_component: Option<BlockComponent<T::EthSpec>>,
         awaiting_parent: Option<Hash256>,
+        lookup_trigger: LookupTrigger,
         peers: &[PeerId],
         cx: &mut SyncNetworkContext<T>,
     ) -> bool {
+        // If the lookup is complete, don't create a new one.
+        if self.completed_lookups.contains(&block_root) {
+            return true;
```
michaelsproul, line 252: It might be nice to have a `debug!` log here to know when this is firing

dapplion:
I would prefer not to go with this solution, instead go for:
- https://github.com/sigp/lighthouse/pull/5722

realbigsean:
closing for https://github.com/sigp/lighthouse/pull/5722

PR #5698, opened 2024-05-03
Make modified helpers in electra fork aware

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

In https://github.com/sigp/lighthouse/pull/5653/ , we modified a bunch of functions for electra without making them fork aware.
This PR aims to make the functions that the spec marks as "Modified in electra" fork aware.

## Additional info
I intend to create pre and post electra versions for existing functions that are modified in the electra spec even if the logic doesn't change to make sure we don't shoot ourself in the foot in cases like these in the future.


jimmygchen (approved):
LGTM. Confirmed that existing pre-electra logic unchanged.

michaelsproul (approved):

jimmygchen:
@mergify queue

PR #5700, opened 2024-05-03
Consistent logging of full root in lookup debug logs

dapplion:
## Issue Addressed

From offline discussions:

I think we should aim for consistency either way. sean proposed:

- full root on debug logs
- shorten root on info logs

## Proposed Changes

- Consistent logging of full root in lookup debug logs
- Fix clippy warning
- Add missing warn log on unexpected cas

### DIFF
--- a/beacon_node/network/src/sync/block_lookups/single_block_lookup.rs
+++ b/beacon_node/network/src/sync/block_lookups/single_block_lookup.rs
@@ -172,17 +172,11 @@ impl<T: BeaconChainTypes> SingleBlockLookup<T> {
         )
     }
 
-    /// Add all given peers to both block and blob request states.
-    pub fn add_peer(&mut self, peer_id: PeerId) {
-        self.block_request_state.state.add_peer(&peer_id);
-        self.blob_request_state.state.add_peer(&peer_id);
-    }
-
-    /// Add all given peers to both block and blob request states.
-    pub fn add_peers(&mut self, peers: &[PeerId]) {
-        for peer in peers {
-            self.add_peer(*peer);
-        }
+    /// Add peer to all request states. The peer must be able to serve this request.
+    /// Returns true if the peer was newly inserted into some request state.
+    pub fn add_peer(&mut self, peer_id: PeerId) -> bool {
+        self.block_request_state.state.add_peer(&peer_id)
+            || self.blob_request_state.state.add_peer(&peer_id)
     }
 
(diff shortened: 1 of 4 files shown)


realbigsean (approved):
Looks good!

beacon_node/network/src/sync/block_lookups/single_block_lookup.rs
```diff
@@ -172,17 +172,11 @@ impl<T: BeaconChainTypes> SingleBlockLookup<T> {
...
-    /// Add all given peers to both block and blob request states.
-    pub fn add_peers(&mut self, peers: &[PeerId]) {
-        for peer in peers {
-            self.add_peer(*peer);
-        }
+    /// Add peer to all request states. The peer must be able to serve this request.
+    /// Returns true if the peer was newly inserted into some request state.
+    pub fn add_peer(&mut self, peer_id: PeerId) -> bool {
+        self.block_request_state.state.add_peer(&peer_id)
+            || self.blob_request_state.state.add_peer(&peer_id)
```
jimmygchen: Is it intentional to short circuit this? i.e. if `self.block_request_state.state.add_peer(&peer_id)` returns true then it won't add peer to `blob_request_state`. 
If it is, could we add some reasoning here as docs?

jimmygchen (changes requested):
Nice improvements! I've just got a question above.

beacon_node/network/src/sync/block_lookups/single_block_lookup.rs
dapplion: Nice pick! Definitely not intended behaviour

jimmygchen (approved):
Looks good! 👍

realbigsean:
@mergify queue

dapplion:
```
May 07 17:26:28.005 DEBG Received lookup processing result       result: Err(ParentUnknown(RpcBlock(0x3a35cce1b7ce7090e1c698e27887b468a635ea3d808c9fd3653f1b11319eb486))),
id: 2046, block_root: 0x3a35cce1b7ce7090e1c698e27887b468a635ea3d808c9fd3653f1b11319eb486, 
component: Block, service: lookup_sync, service: sync
```

Declaring the service two times does not overwrite the label, but log it twice. Not sure if this is the intended behavior. Elastic search can deal with it fine, as `service = [lookup_sync, sync]` and a filter by `service is "lookup_sync"` works.

PR #5701, opened 2024-05-03
Add data column kzg verification and update `c-kzg`.

jimmygchen:
## Issue Addressed

Add data column kzg verification and update `c-kzg`.


dapplion (approved):
Looks good!

PR #5704, opened 2024-05-03
Move sync lookup trait function to its caller

dapplion:
## Issue Addressed

Noticed that `RequestState` trait implements a function that can be moved to its caller `SingleBlockLookup`. Because `RequestState::continue_request` needed state from the lookup it had to pass some arguments around.

## Proposed Changes

- Move sync lookup trait function to its caller
- Logic is identical just copy pasted code from a file to the other


realbigsean (approved):
Nice cleanup

realbigsean:
@mergify queue

PR #5705, opened 2024-05-03
Bound lookup parent chain length with tip extension

dapplion:
## Issue Addressed

Sync lookup bounds the maximum length of parent chains. Current code covered the case where a new parent lookup extended the ancestor of a parent chain beyond the max length. However, there's a second case. Consider these events:
- New lookup for unknown block root
- Receive block root which happens to be the child of a parent's chain tip
- Parent chain exceeds max length

## Proposed Changes

Bound lookup parent chain length with tip extension. Extending the sequence of events above:

- New lookup for unknown block root
- Receive block root which happens to be the child of a parent's chain tip
- Send block for processing
- Receive `ParentUnknown`
- Call `search_parent_of_child(parent, block)`

This PR introduces the max length bound in this final step


realbigsean (approved):
LGTM

realbigsean:
@mergify queue

PR #5706, opened 2024-05-03
Debounce UnknownBlockHashFromAttestation events

dapplion:
## Issue Addressed

debounce duplicated `UnknownBlockHashFromAttestation` for the same root peer tuple. A peer may forward us thousands of attestations, each triggering an individual event. Only one event is useful, the rest generating log noise and wasted cycles

PR
- https://github.com/sigp/lighthouse/pull/5672

Noticed the issue first but there's another log that gets spammy if the peer is disconnecting here

https://github.com/sigp/lighthouse/blob/ee974db0baba2aa0ef04b320901dcfb9d65f6f67/beacon_node/network/src/sync/manager.rs#L694

I think this fix is more complete as it also reduces CPU cycles downstream, as for every event it wants to find a lookup and insert the peer to it

## Proposed Changes

Debounce UnknownBlockHashFromAttestation events

- **Q**: Should we revert https://github.com/sigp/lighthouse/pull/5672 now that the log is not spammy?

### DIFF
--- a/beacon_node/network/src/sync/manager.rs
+++ b/beacon_node/network/src/sync/manager.rs
@@ -56,6 +56,7 @@ use lighthouse_network::rpc::RPCError;
 use lighthouse_network::types::{NetworkGlobals, SyncState};
 use lighthouse_network::SyncInfo;
 use lighthouse_network::{PeerAction, PeerId};
+use lru_cache::LRUTimeCache;
 use slog::{crit, debug, error, info, trace, warn, Logger};
 use std::ops::Sub;
 use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


realbigsean:
> Q: Should we revert https://github.com/sigp/lighthouse/pull/5672 now that the log is not spammy?

Yes! down for that

AgeManning (approved):
Looks good to me. 

The reason we are using the (peer_id, root) tuple is because we want distinct peers to send us the same root to add them to current lookups right?

dapplion:
> The reason we are using the (peer_id, root) tuple is because we want distinct peers to send us the same root to add them to current lookups right?

Yes, if a second peer claims to know about a root of an existing lookup it will be added to the pool of "peers that claim to know the block". If the initial request of the lookup fails, that second peer may be used as retry. It's also important for scoring. If the block ends up being invalid we want to penalize everyone that claims to have imported it

realbigsean:
@mergify queue

PR #5710, opened 2024-05-03
Pass vec to range sync batch

dapplion:
## Issue Addressed

Non-urgent clean-up possible after
- https://github.com/sigp/lighthouse/pull/5583

This can wait for after v5.2.0

Currently we are:
- Fetching an entire Vec of blocks in the network context
- Constructing a stream with None terminator in sync manager
- Batch handles each block individually

## Proposed Changes

Instead, we can just pass the vec from the network context to the batch

All logic in unchanged, most of the diff is a change in indentation


pawanjay176 (approved):
Nice find

AgeManning (approved):
Yeah, looks like a clear improvement. We drop a single log field in blocks.len() in the downloading state, but this is much cleaner

realbigsean:
@mergify queue

PR #5712, opened 2024-05-03
Electra: Remaining Consensus Data Structures

ethDreamer:
## Issue Addressed

* #5605 

This PR implements the remaining consensus data structure changes in `Electra`. It can't be merged until we cut a release for `5.2.0`


testing/ef_tests/src/cases/operations.rs
```diff
@@ -122,8 +122,15 @@ impl<E: EthSpec> Operation<E> for AttesterSlashing<E> {
...
+    fn decode(path: &Path, fork_name: ForkName, _spec: &ChainSpec) -> Result<Self, Error> {
+        Ok(match fork_name {
+            ForkName::Base
+            | ForkName::Altair
+            | ForkName::Bellatrix
+            | ForkName::Capella
+            | ForkName::Deneb => Self::Base(ssz_decode_file(path)?),
+            ForkName::Electra => Self::Electra(ssz_decode_file(path)?),
+        })
     }
```
ethDreamer, line 138: something is strange here.. the EF `attester_slashing` tests pass even when I make this always decode `Electra`

testing/ef_tests/src/cases/fork_choice.rs
```diff
@@ -183,10 +184,24 @@ impl<E: EthSpec> LoadCase for ForkChoiceTest<E> {
...
-                Step::AttesterSlashing { attester_slashing } => {
-                    ssz_decode_file(&path.join(format!("{}.ssz_snappy", attester_slashing)))
-                        .map(|attester_slashing| Step::AttesterSlashing { attester_slashing })
-                }
+                Step::AttesterSlashing { attester_slashing } => match fork_name {
+                    ForkName::Base
+                    | ForkName::Altair
+                    | ForkName::Bellatrix
+                    | ForkName::Capella
+                    | ForkName::Deneb => {
```
ethDreamer: here too.. the EF tests pass even if i make this always decode `electra`..

beacon_node/beacon_chain/src/attestation_verification.rs
```diff
@@ -246,9 +265,30 @@ pub enum Error {
     BeaconChainError(BeaconChainError),
 }
 
+// TODO(electra) the error conversion changes here are to get a test case to pass
```
realbigsean: I removed this conversion, haven't figure out which tests this was referring to  https://github.com/sigp/lighthouse/pull/5972/commits/3063066b96dc158f87ed2bd5ce1addaf2e09bd6a

beacon_node/beacon_chain/src/attestation_verification.rs
```diff
@@ -1239,42 +1335,106 @@ pub fn verify_signed_aggregate_signatures<T: BeaconChainTypes>(
...
 ///
-/// This function exists in this odd "map" pattern because efficiently obtaining the committee for
-/// an attestation can be complex. It might involve reading straight from the
+/// This function exists in this odd "map" pattern because efficiently obtaining the committees for
+/// an attestations slot can be complex. It might involve reading straight from the
 /// `beacon_chain.shuffling_cache` or it might involve reading it from a state from the DB. Due to
 /// the complexities of `RwLock`s on the shuffling cache, a simple `Cow` isn't suitable here.
 ///
-/// If the committee for `attestation` isn't found in the `shuffling_cache`, we will read a state
+/// If the committees for an `attestation`'s slot isn't found in the `shuffling_cache`, we will read a state
```
realbigsean:
```suggestion
/// If the committees for an `attestation`'s slot aren't found in the `shuffling_cache`, we will read a state
```

beacon_node/beacon_chain/src/attestation_verification.rs
```diff
@@ -1239,42 +1335,106 @@ pub fn verify_signed_aggregate_signatures<T: BeaconChainTypes>(
...
+        }
     })
 }
 
 /// Runs the `map_fn` with the committee and committee count per slot for the given `attestation`.
 ///
-/// This function exists in this odd "map" pattern because efficiently obtaining the committee for
-/// an attestation can be complex. It might involve reading straight from the
+/// This function exists in this odd "map" pattern because efficiently obtaining the committees for
+/// an attestations slot can be complex. It might involve reading straight from the
```
realbigsean:
```suggestion
/// an attestation's slot can be complex. It might involve reading straight from the
```

testing/ef_tests/src/cases/operations.rs
dapplion: The only difference is on the merkleization right, operation tests only SSZ deserialize and test functionality.

validator_client/src/http_api/tests/keystores.rs
```diff
@@ -1236,7 +1236,7 @@ async fn delete_nonexistent_keystores() {
 }
 
 fn make_attestation(source_epoch: u64, target_epoch: u64) -> Attestation<E> {
-    Attestation {
+    Attestation::Base(AttestationBase {
```
realbigsean, line 1239: Should we extend these tests to work with electra attestations? I think these test slashing protection db migration but no slashing protection logic has changed

testing/web3signer_tests/src/lib.rs
```diff
@@ -542,7 +542,7 @@ mod tests {
 
     /// Get a generic, arbitrary attestation for signing.
     fn get_attestation() -> Attestation<E> {
-        Attestation {
+        Attestation::Base(AttestationBase {
```
realbigsean, line 545: We should add web3 signer tests once it's supported

consensus/types/src/signed_aggregate_and_proof.rs
```diff
@@ -66,9 +83,28 @@ impl<E: EthSpec> SignedAggregateAndProof<E> {
...
+            AggregateAndProof::Electra(message) => {
+                SignedAggregateAndProof::Electra(SignedAggregateAndProofElectra {
+                    message,
+                    signature: secret_key.sign(signing_message),
+                })
+            }
+        }
+    }
+
+    pub fn message(&self) -> AggregateAndProofRef<E> {
```
realbigsean: used superstruct map macros here https://github.com/sigp/lighthouse/pull/5972/commits/beddbd984f8bab621b9ec43630a810efdaff79d3

consensus/types/src/aggregate_and_proof.rs
```diff
@@ -1,49 +1,87 @@
...
+    #[superstruct(flatten)]
     pub aggregate: Attestation<E>,
     /// A proof provided by the validator that permits them to publish on the
     /// `beacon_aggregate_and_proof` gossipsub topic.
     pub selection_proof: Signature,
 }
 
+impl<'a, E: EthSpec> AggregateAndProofRef<'a, E> {
+    /// Returns `true` if `validator_pubkey` signed over `self.aggregate.data.slot`.
+    pub fn aggregate(self) -> AttestationRef<'a, E> {
```
realbigsean, line 62: used map macros here https://github.com/sigp/lighthouse/pull/5972/commits/beddbd984f8bab621b9ec43630a810efdaff79d3

consensus/state_processing/src/verify_operation.rs
```diff
@@ -13,29 +13,133 @@ use ssz::{Decode, Encode};
...
     Eq,
-    Hash(bound = "T: Encode + Decode + std::hash::Hash, E: EthSpec")
+    Hash(bound = "T: TransformPersist + std::hash::Hash, E: EthSpec")
 )]
-pub struct SigVerifiedOp<T: Encode + Decode, E: EthSpec> {
+pub struct SigVerifiedOp<T: TransformPersist, E: EthSpec> {
     op: T,
     verified_against: VerifiedAgainst,
-    #[ssz(skip_serializing, skip_deserializing)]
+    //#[ssz(skip_serializing, skip_deserializing)]
```
realbigsean: deleted here: https://github.com/sigp/lighthouse/pull/5972/commits/c33907c229637165af626ea90eedde9da3b12354

consensus/state_processing/src/per_block_processing/signature_sets.rs
```diff
@@ -312,21 +312,22 @@ where
...
     let domain = spec.get_domain(
-        indexed_attestation.data.target.epoch,
+        indexed_attestation.data().target.epoch,
         Domain::BeaconAttester,
         fork,
         genesis_validators_root,
     );
 
-    let message = indexed_attestation.data.signing_root(domain);
+    // TODO(electra), signing root isnt unique in the case of electra
```
dapplion: Why is that relevant? One can include identical attestations in a Deneb block today, so having non-unique signing roots in the same block

consensus/state_processing/src/consensus_context.rs
```diff
@@ -148,26 +147,32 @@ impl<E: EthSpec> ConsensusContext<E> {
...
-            Entry::Vacant(vacant) => {
-                let committee =
-                    state.get_beacon_committee(attestation.data.slot, attestation.data.index)?;
-                let indexed_attestation =
-                    get_indexed_attestation(committee.committee, attestation)?;
-                Ok(vacant.insert(indexed_attestation))
-            }
+        attestation: AttestationRef<'a, E>,
+    ) -> Result<IndexedAttestationRef<E>, BlockOperationError<AttestationInvalid>> {
+        let key = attestation.tree_hash_root();
```
dapplion, line 155: Why not key by signature here? If the attestation is valid there's a single possible valid signature value for each attestation

beacon_node/operation_pool/src/attestation_storage.rs
```diff
@@ -125,13 +156,43 @@ impl CheckpointKey {
...
+    pub fn should_aggregate(&self, other: &Self) -> bool {
+        match (self, other) {
+            (CompactIndexedAttestation::Base(this), CompactIndexedAttestation::Base(other)) => {
+                this.should_aggregate(other)
+            }
+            (
+                CompactIndexedAttestation::Electra(this),
+                CompactIndexedAttestation::Electra(other),
+            ) => this.should_aggregate(other),
+            // TODO(electra) is a mix of electra and base compact indexed attestations an edge case we need to deal with?
```
realbigsean: should always be aggregating over the same slot so i think false here is fine, removed the todo: https://github.com/sigp/lighthouse/pull/5972/commits/2296cfe99ea545dc92884ec638f4158253180c25

beacon_node/operation_pool/src/attestation_storage.rs
```diff
@@ -125,13 +156,43 @@ impl CheckpointKey {
...
+    pub fn aggregate(&mut self, other: &Self) -> Option<()> {
+        match (self, other) {
+            (CompactIndexedAttestation::Base(this), CompactIndexedAttestation::Base(other)) => {
+                this.aggregate(other)
+            }
+            (
+                CompactIndexedAttestation::Electra(this),
+                CompactIndexedAttestation::Electra(other),
+            ) => this.aggregate_same_committee(other),
+            // TODO(electra) is a mix of electra and base compact indexed attestations an edge case we need to deal with?
```
realbigsean: same as above: https://github.com/sigp/lighthouse/pull/5972/commits/2296cfe99ea545dc92884ec638f4158253180c25

beacon_node/beacon_chain/src/attestation_verification.rs
```diff
@@ -1239,42 +1335,106 @@ pub fn verify_signed_aggregate_signatures<T: BeaconChainTypes>(
...
+                    .map_err(|e| {
+                        let index = att.committee_index().unwrap_or(0);
+                        if e == BlockOperationError::BeaconStateError(NoCommitteeFound(index)) {
+                            Error::NoCommitteeForSlotAndIndex {
+                                slot: att.data.slot,
+                                index,
+                            }
+                        } else {
+                            Error::Invalid(e)
+                        }
```
dapplion: Why is this error re-mapping necessary? The `att.committee_index().unwrap_or(0)` is ugly. Are you sure this attestation will have at most one committee? Otherwise you are mapping on the first committee. If the consumer of this function really depends on this error we can do
```rust
                        if let BlockOperationError::BeaconStateError(NoCommitteeFound(index)) = e {
                            Error::NoCommitteeForSlotAndIndex {
                                slot: att.data.slot,
                                index,
                            }
                        } else {
                            Error::Invalid(e)
                        }

```

beacon_node/beacon_chain/src/attestation_verification.rs
```diff
@@ -1239,42 +1335,106 @@ pub fn verify_signed_aggregate_signatures<T: BeaconChainTypes>(
...
-    map_attestation_committee(chain, attestation, |(committee, committees_per_slot)| {
-        get_indexed_attestation(committee.committee, attestation)
-            .map(|attestation| (attestation, committees_per_slot))
-            .map_err(Error::Invalid)
+    map_attestation_committees(chain, attestation, |(committees, committees_per_slot)| {
+        match attestation {
+            AttestationRef::Base(att) => {
+                let committee = committees
+                    .iter()
+                    .filter(|&committee| committee.index == att.data.index)
```
dapplion, line 1356: Why are committees not sorted so that we can just index here?

beacon_node/beacon_chain/src/naive_aggregation_pool.rs
```diff
@@ -121,48 +238,57 @@ impl<E: EthSpec> AggregateMap for AggregatedAttestationMap<E> {
...
+                .map_err(|iter| Error::MoreThanOneAggregationBitSet(iter.count()))?
+                .ok_or(Error::NoAggregationBitsSet)?,
+            AttestationRef::Electra(att) => att
+                .aggregation_bits
+                .iter()
+                .enumerate()
+                .filter_map(|(i, bit)| if bit { Some(i) } else { None })
+                .at_most_one()
+                .map_err(|iter| Error::MoreThanOneAggregationBitSet(iter.count()))?
+                .ok_or(Error::NoAggregationBitsSet)?,
```
dapplion: Should move this logic to the attestation type?

beacon_node/beacon_chain/src/observed_aggregates.rs
```diff
@@ -95,35 +106,60 @@ pub trait SubsetItem {
...
 
-impl<E: EthSpec> SubsetItem for Attestation<E> {
-    type Item = BitList<E::MaxValidatorsPerCommittee>;
+impl<'a, E: EthSpec> SubsetItem for AttestationRef<'a, E> {
+    type Item = BitList<E::MaxValidatorsPerSlot>;
     fn is_subset(&self, other: &Self::Item) -> bool {
-        self.aggregation_bits.is_subset(other)
+        match self {
+            Self::Base(att) => {
+                if let Ok(extended_aggregation_bits) = att.extend_aggregation_bits() {
```
dapplion, line 121: This seems expensive to have to extend the bits for every single attestation we receive

beacon_node/beacon_chain/src/observed_aggregates.rs
```diff
@@ -95,35 +106,60 @@ pub trait SubsetItem {
...
+                    return extended_aggregation_bits.is_subset(other);
+                }
+                false
```
dapplion, line 124: When could this false happen? For `extend_aggregation_bits` to return error it would need `aggregation_bits` to be longer than `MaxValidatorsPerSlot`

beacon_node/beacon_chain/tests/attestation_verification.rs
```diff
@@ -213,14 +227,20 @@ fn get_valid_aggregated_attestation<T: BeaconChainTypes>(
...
 fn get_non_aggregator<T: BeaconChainTypes>(
     chain: &BeaconChain<T>,
-    aggregate: &Attestation<T::EthSpec>,
+    aggregate: AttestationRef<T::EthSpec>,
 ) -> (usize, SecretKey) {
     let head = chain.head_snapshot();
     let state = &head.beacon_state;
     let current_slot = chain.slot().expect("should get slot");
 
+    // TODO(electra) make fork-agnostic
```
dapplion: Why exactly is this not fork-agnostic?

beacon_node/operation_pool/src/lib.rs
```diff
@@ -284,6 +283,16 @@ impl<E: EthSpec> OperationPool<E> {
         let mut num_prev_valid = 0_i64;
         let mut num_curr_valid = 0_i64;
 
+        // TODO(electra): Work out how to do this more elegantly. This is a bit of a hack.
+        let mut all_attestations = self.attestations.write();
+
+        if fork_name >= ForkName::Electra {
+            all_attestations.aggregate_across_committees(prev_epoch_key);
+            all_attestations.aggregate_across_committees(curr_epoch_key);
```
dapplion, line 291: Why do you need to persist the result of aggregating the attestations across committees in the pool? You can just retrieve the best attestation on each committee, merge them and return the result for block production

beacon_node/store/src/consensus_context.rs
```diff
@@ -21,8 +21,7 @@ pub struct OnDiskConsensusContext<E: EthSpec> {
     ///
     /// They are not part of the on-disk format.
     #[ssz(skip_serializing, skip_deserializing)]
-    indexed_attestations:
-        HashMap<(AttestationData, BitList<E::MaxValidatorsPerCommittee>), IndexedAttestation<E>>,
+    indexed_attestations: HashMap<Hash256, IndexedAttestation<E>>,
```
dapplion, line 24: Should this root be wrapped into a struct to clarify its meaning?

dapplion (commented):
Almost done with the review, I'm missing the op_pool changes and get_attesting_indices. Sent some review comments that prefer to discuss than fix directly

I have pushed ~10 more commits to 
- https://github.com/sigp/lighthouse/pull/5940

consensus/state_processing/src/common/get_attesting_indices.rs
```diff
@@ -1,32 +1,279 @@
...
+        get_attesting_indices::<E>(&committees, &att.aggregation_bits, &att.committee_bits)
+    }
+
+    /// Returns validator indices which participated in the attestation, sorted by increasing index.
+    pub fn get_attesting_indices<E: EthSpec>(
+        committees: &[BeaconCommittee],
+        aggregation_bits: &BitList<E::MaxValidatorsPerSlot>,
+        committee_bits: &BitVector<E::MaxCommitteesPerSlot>,
+    ) -> Result<Vec<u64>, BeaconStateError> {
+        let mut output: HashSet<u64> = HashSet::new();
```
dapplion: Why is this a HashSet? All committees are exclusive so it is not possible for a validator to appear multiple times in an aggregation bits

beacon_node/operation_pool/src/attestation_storage.rs
```diff
@@ -172,19 +331,101 @@ impl<E: EthSpec> AttestationMap<E> {
         }
     }
 
+    /// Aggregate Electra attestations for the same attestation data signed by different
+    /// committees.
+    ///
+    /// Non-Electra attestations are left as-is.
+    pub fn aggregate_across_committees(&mut self, checkpoint_key: CheckpointKey) {
```
dapplion, line 335: @michaelsproul why is it necessary to pre-aggregate across committees before running max-cover? All committees are exclusive so you can run max-cover on each committee, pick the collection of solutions and merge them. Is there any efficiency reason? Seems simpler to merge the end instead of mutating the pool every run.

beacon_node/beacon_chain/src/observed_aggregates.rs
eserilev: right, thats the only error case

beacon_node/beacon_chain/src/attestation_verification.rs
eserilev: yeah we require the attestation here to have at most one committee. we should only see attestations aggregated across committees at the block packing phase. Agreed that its ugly and should be cleaned up

beacon_node/beacon_chain/tests/attestation_verification.rs
eserilev: it is now, will remove the TODO

consensus/state_processing/src/common/get_attesting_indices.rs
eserilev: yes agreed, it should just be a Vec

beacon_node/beacon_chain/src/observed_aggregates.rs
realbigsean: copied from offline chat let me know what you think: 

> We should be able to do this conversion only once but I think we need it at least once unless we want multiple observed caches. 
> is_subset and is_superset shouldn't need this conversion because these should always be called on attestations of the same slot, which should have the same bitvector lengths, right? So I think we just need to do this conversion prior to storing
> During the conversion, the new bitvector we're making is backed by a smallvec so it should stay on the stack. I guess it's a question of the iteration speed is quick, but they're of length 2048 so each is small. And this conversion will stop happening on the fork. so I'm not sure it's worth optimizing further than just making sure we do this only once per attesatation.

realbigsean:
couple fixes: 
- wrong persisted op pool deserialization from disk https://github.com/sigp/lighthouse/pull/5712/commits/afb9122cc1c9b243542dddaaac611493c01f7af6
- json serialization issue with `AggregateAndProofRef` (used in the web3signer) https://github.com/sigp/lighthouse/pull/5712/commits/381bbaba94df0ba23a267f92038f4637942fd2d3

consensus/state_processing/src/per_block_processing/signature_sets.rs
eserilev: yes, you're right, this comment is irrelevant

beacon_node/operation_pool/src/lib.rs
eserilev: yes the TODO above mentions we should find a way to do this more elegantly. I agree that persisting the final aggregation in the pool is unnecessary. I'll open up a separate issue for this

consensus/state_processing/src/consensus_context.rs
eserilev: sure, I dont see why not, seems like a nice optimization to make

beacon_node/operation_pool/src/attestation_storage.rs
dapplion: Response: The final selection of attestations can only include 8 distinct attestation data. Running max-cover first on each committee index may not yield a set of results that have overlapping attestation data. Aggregating across committees makes the max-cover algo aware of this limitation and produces better results.

beacon_node/beacon_chain/src/attestation_verification.rs
```diff
@@ -1067,14 +1169,13 @@ pub fn verify_propagation_slot_range<S: SlotClock, E: EthSpec>(
 
     let current_fork =
         spec.fork_name_at_slot::<E>(slot_clock.now().ok_or(BeaconChainError::UnableToReadSlot)?);
-    let earliest_permissible_slot = match current_fork {
-        ForkName::Base | ForkName::Altair | ForkName::Bellatrix | ForkName::Capella => {
-            one_epoch_prior
-        }
-        // EIP-7045
-        ForkName::Deneb | ForkName::Electra => one_epoch_prior
+    let earliest_permissible_slot = if current_fork < ForkName::Deneb {
```
realbigsean: Used `deneb_enabled` method here: https://github.com/sigp/lighthouse/pull/5972/commits/c1c0c6f85bbe50714ed9e646282e69edfc556b25

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -5115,6 +5158,28 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             bls_to_execution_changes,
         } = partial_beacon_block;
 
+        let (attester_slashings_base, attester_slashings_electra) =
```
realbigsean, line 5161: maybe time to make `PartialBeaconBlock` a superstruct? (out of scope for this PR though)

beacon_node/operation_pool/src/lib.rs
```diff
@@ -307,14 +316,19 @@ impl<E: EthSpec> OperationPool<E> {
             )
             .inspect(|_| num_curr_valid += 1);
 
+        let curr_epoch_limit = if fork_name < ForkName::Electra {
```
realbigsean: used `electra_enabled` method here: https://github.com/sigp/lighthouse/pull/5972/commits/c1c0c6f85bbe50714ed9e646282e69edfc556b25

consensus/types/src/attestation.rs
```diff
@@ -18,47 +21,356 @@ pub enum Error {
...
+    pub fn empty_for_signing(
+        committee_index: u64,
+        committee_length: usize,
+        slot: Slot,
+        beacon_block_root: Hash256,
+        source: Checkpoint,
+        target: Checkpoint,
+        spec: &ChainSpec,
+    ) -> Result<Self, Error> {
+        if spec.fork_name_at_slot::<E>(slot) >= ForkName::Electra {
```
realbigsean: used `electra_enabled` method here: https://github.com/sigp/lighthouse/pull/5972/commits/c1c0c6f85bbe50714ed9e646282e69edfc556b25

consensus/types/src/attestation.rs
```diff
@@ -18,47 +21,356 @@ pub enum Error {
...
+                .filter(|(_i, bit)| *bit)
+                .map(|(i, _bit)| i)
+                .collect::<Vec<_>>(),
+        }
+    }
+}
+
+impl<E: EthSpec> AttestationElectra<E> {
     /// Are the aggregation bitfields of these attestations disjoint?
+    // TODO(electra): check whether the definition from CompactIndexedAttestation::should_aggregate
```
realbigsean: I'm only seeing this used in debug asserts so I removed this method. does that seem ok to people? https://github.com/sigp/lighthouse/pull/5972/commits/1567c36f0ec66f740f04de88c6571c9f994ab3a4

beacon_node/operation_pool/src/attestation.rs
```diff
@@ -170,12 +177,16 @@ impl<'a, E: EthSpec> MaxCover for AttMaxCover<'a, E> {
...
-    attestation: &AttestationRef<E>,
+    attestation: &CompactAttestationRef<E>,
     state: &BeaconState<E>,
     base_state: &BeaconStateBase<E>,
 ) -> BitList<E::MaxValidatorsPerCommittee> {
     // Bitfield of validators whose attestations are new/fresh.
-    let mut new_validators = attestation.indexed.aggregation_bits.clone();
+    let mut new_validators = match attestation.indexed {
+        CompactIndexedAttestation::Base(indexed_att) => indexed_att.aggregation_bits.clone(),
+        // TODO(electra) per the comments above, this code path is obsolete post altair fork, so maybe we should just return an empty bitlist here?
```
realbigsean: instantiating with a zero size bitlist and unwrap because we do the same a few lines later https://github.com/sigp/lighthouse/pull/5972/commits/fe8ab233a2f4074dc606c8abce16223c1c272233

beacon_node/operation_pool/src/attestation_storage.rs
```diff
@@ -158,8 +315,9 @@ impl<E: EthSpec> AttestationMap<E> {
         // NOTE: this is sub-optimal and in future we will remove this in favour of max-clique
         // aggregation.
         let mut aggregated = false;
+
         for existing_attestation in attestations.iter_mut() {
-            if existing_attestation.signers_disjoint_from(&indexed) {
+            if existing_attestation.should_aggregate(&indexed) {
                 existing_attestation.aggregate(&indexed);
```
realbigsean: `aggregate` here returns an `Option` denoting whether it was actually aggregated. The option isn't used though, so I updated this here: https://github.com/sigp/lighthouse/pull/5972/commits/ab819dcdb07ba57c308b5198eac5ad95e303e824

realbigsean (changes requested):

consensus/types/src/attestation.rs
michaelsproul: I support its deletion. It's a footgun

validator_client/src/http_api/tests/keystores.rs
realbigsean: I'm thinking we leave it as is

beacon_node/beacon_chain/src/beacon_chain.rs
realbigsean: raised an issue: https://github.com/sigp/lighthouse/issues/5981

testing/web3signer_tests/src/lib.rs
realbigsean: issue: https://github.com/sigp/lighthouse/issues/5982

michaelsproul (approved):
I'm ok with merging this to `unstable`. There are no blockers that I'm aware of.

realbigsean (approved):
me too

realbigsean:
@mergify queue

PR #5722, opened 2024-05-06
Notify lookup sync of gossip processing results

dapplion:
## General problem

Lookup sync and gossip interact in very racy ways. They interact at all because we want to prevent downloading things via ReqResp that we are already processing. Note that this optimization carries complexity, which we may decide if it's worth it at all.

Screenshot 2024-05-06 at 23 11 48

Consider the image above, we can receive an attestation referencing a block not yet imported into fork-choice at these moments in time:

- `A` before processing: Easy, download and process the block
- `B` during processing: Block is already processing, should skip download. However, we must consider:
  - What to do if processing fails?
  - Should we create a lookup at all? If not, what to do if we receive a second attestation referencing the child of this block?
- `C` while waiting blobs: Block is fully processed but not yet in fork-choice. In this case we don't have to worry about processing failures, but the child lookup issue remains.
- `D` after import: This trigger should never happen. If it does we will download all block components and get a `BlockAreadyKnown` error.

## Issues with current unstable da9d38698d9ea25202511ed924f8d7baddbf1f27

- https://github.com/sigp/lighthouse/pull/5681

^ introduces the optimization to prevent downloads for triggers `B` and `C`. However, it fails to explicitly handle both pitfalls listed above. If processing fails, a lookup may get stuck.

Also, it creates a spammy situation where a lookup get dropped immediately if the block is still processing, once per attestation

- https://github.com/sigp/lighthouse/pull/5696

^ Fixed the spammy loop, but now it make the case of processing failure worse. Completed chains remain in a hashset for 60 seconds, so a failing lookup will be blocked from being retried for that period.

## Proposed Changes

Sync lookup must be aware of gossip processing results. I think there's no way around that.

The least invasive way to achieve that IMO is:
- If block is the processing cache, mark lookup request as processing
- When gossip processing completes, send sync message `BlockComponentProcessed`

With the above we tackle:
1. Immediately retry lookups for blocks that fail processing, don't wait 60 seconds
2. Immediately continue or drop child lookup of blocks from gossip
3. No spammy loop, lookups are not marked completed until imported into fork-choice

## Pitfalls

This PR relies on these assumptions:
- If `availability_checker.has_block(block)` returns true, the block is permanently valid and is missing blobs
- If `reqresp_pre_import_cache.contains_key(block)` returns true, a `BlockComponentProcessed` event MUST be emitted some time in the future

We don't have e2e to assert those conditions at the moment. I think code comments should be good for now but it would be great to codify them in the future somehow

## Todo

- [ ] Update lookup tests, behaviour has slightly changed

Closes https://github.com/sigp/lighthouse/issues/5693

### DIFF
--- a/beacon_node/network/src/network_beacon_processor/sync_methods.rs
+++ b/beacon_node/network/src/network_beacon_processor/sync_methods.rs
@@ -1,5 +1,6 @@
 use crate::metrics;
 use crate::network_beacon_processor::{NetworkBeaconProcessor, FUTURE_SLOT_TOLERANCE};
+use crate::sync::manager::BlockProcessSource;
 use crate::sync::BatchProcessResult;
 use crate::sync::{
     manager::{BlockProcessType, SyncMessage},
@@ -53,7 +54,7 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
         block_root: Hash256,
         block: RpcBlock<T::EthSpec>,
         seen_timestamp: Duration,
-        process_type: BlockProcessType,
+        source: BlockProcessSource,
     ) -> AsyncFn {
         let process_fn = async move {
             let reprocess_tx = self.reprocess_tx.clone();
@@ -62,7 +63,7 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                 block_root,
                 block,
                 seen_timestamp,
-                process_type,
+                source,
                 reprocess_tx,
                 duplicate_cache,
             )
@@ -77,20 +78,21 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
         block_root: Hash256,
         block: RpcBlock<T::EthSpec>,
         seen_timestamp: Duration,
-        process_type: BlockProcessType,
+        source: BlockProcessSource,
     ) -> (AsyncFn, BlockingFn) {
         // An async closure which will import the block.
         let process_fn = self.clone().generate_rpc_beacon_block_process_fn(
             block_root,
             block,
             seen_timestamp,
-            process_type.clone(),
+            source.clone(),
         );
         // A closure which will ignore the block.
         let ignore_fn = move || {
             // Sync handles these results
             self.send_sync_message(SyncMessage::BlockComponentProcessed {
-                process_type,
+                process_type: BlockProcessType::SingleBlock,
+                source,
                 result: crate::sync::manager::BlockProcessingResult::Ignored,
             });
         };
@@ -104,7 +106,7 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
         block_root: Hash256,
         block: RpcBlock<T::EthSpec>,
         seen_timestamp: Duration,
-        process_type: BlockProcessType,
+        source: BlockProcessSource,
         reprocess_tx: mpsc::Sender<ReprocessQueueMessage>,
         duplicate_cache: DuplicateCache,
     ) {
@@ -115,7 +117,7 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                 "Gossip block is being processed";
                 "action" => "sending rpc block to reprocessing queue",
                 "block_root" => %block_root,
-                "process_type" => ?process_type,
+                "source" => ?source,
             );
 
             // Send message to work reprocess queue to retry the block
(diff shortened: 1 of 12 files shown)


AgeManning:
I was thinking about this more. 

In order to avoid sync having to know anything about block processing results (explicitly) can't we do the following:

- If a block request comes in and we are processing, then ignore the request
- If a request for a child comes in and the parent is processing, ignore (or as an optimization, cache). 

Eventually the block will either fail, or be processed. We will eventually get more requests for the chain of blocks and eventually it will succeed because the block will either be failed by the beacon processor, or it will be successful. 

The issue is around just restarting the chains. To prevent loops, like we wait a slot or something. 

Just thinking this would be away to avoid explicit communication from the beacon processor.

dapplion:
> If a block request comes in and we are processing, then ignore the request
> If a request for a child comes in and the parent is processing, ignore (or as an optimization, cache).

For child lookups it a bit tricky. Assume you are bit behind, such that given this chain of blocks:

```
Block slot 10 <- peers attesting here
Block slot 9
Block slot 8
Block slot 7
Block slot 6 <- you are here
```

You have to create a lookup for slot 10, fetch block, attempt process, get unknown parent, fetch 9, etc until 6. If 6 is processing lookup 7 needs some event to make progress. So the beacon processor must tell sync that block 6 has been processed. Otherwise dropping all lookups child of 6 while 6 is processing sounds silly. You must leave child lookups of 6 paused somehow awaiting for their ancestor to process.

Sync lookup technically doesn't need to know about the processing result, it would be enough to have a signal to continue childs of 6. But then we have to add a new SyncMessage and have a dedicated handler in block lookups. While it reduces how much information the processor leaks, I don't see how it simplifies things. We already have a `BlockComponentProcessed` message and its handler, why not re-use it? 

Note: The moment that sync lookup checks the chain's processing cache or da_checker we break the separation of concerns between sync lookup and gossip processing. If we are crossing that line, let's implement a proper informational complete fix.

Notice that this ignores will happen thousands of times, one per each attestation. Unless we have something like https://github.com/sigp/lighthouse/pull/5706

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -170,21 +172,20 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
             if reprocess_tx.try_send(reprocess_msg).is_err() {
                 error!(self.log, "Failed to inform block import"; "source" => "rpc", "block_root" => %hash)
             };
-            if matches!(process_type, BlockProcessType::SingleBlock { .. }) {
```
AgeManning, line 173: Why are we removing this match? 

It it redundant because we now only process single blocks here?

dapplion: Yes, blobs are processed in `process_rpc_blobs`

AgeManning (approved):

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -112,10 +112,9 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
         let Some(handle) = duplicate_cache.check_and_insert(block_root) else {
             debug!(
                 self.log,
-                "Gossip block is being processed";
+                "Rpc block is being processed";
```
realbigsean:
```suggestion
                "Gossip block is being processed";
```

beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -561,6 +565,30 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
...
+            return;
+        };
+
+        let lookup_result = if imported {
+            Ok(LookupResult::Completed)
+        } else {
+            lookup.continue_requests(cx)
+        };
+        let id = *id;
+        self.on_lookup_result(id, lookup_result, "external_processing_result", cx);
```
realbigsean, line 589:
```suggestion
        self.on_lookup_result(*id, lookup_result, "external_processing_result", cx);
```

realbigsean (approved):

realbigsean:
@mergify queue

dapplion:
@mergify queue

realbigsean:
@mergify refresh

realbigsean:
@mergify queue

PR #5724, opened 2024-05-06
Do not request current child lookup peers

### LINKED ISSUE #5707: Sync lookup requests blocks from peers that may not have it
### Sync lookup `NoResponseReturned` downscoring issue

Creating a child lookup with a peer from UnknownParentBlock is not okay since that peer may not know about the block. Consider the following sequence of events:
- Peer receives blob of block before block
- Peer fowards blob of block to us
- We don't know the parent of block and trigger UnknownParentBlob(parent, block, peer)
- Create a lookup for parent
- Create a lookup for block
- Lookup for block requests block to peer
- Peer respons with empty
- We penalize peer with NoResponseReturned 

Instead we should:
1. Wait for a gossip attestation and request that peer for the block, or an unknown parent event for its child block
2. Don't create a current lookup. Cache child components somewhere else, or not at all
3. Allow peers to return empty. This will cause "zombie" lookups where no peer has the block but we keep the lookup around somehow

### Solution 1

Create a child lookup with 0 peers and cache the child components there. Scenarios:
- Block has no blobs, so after the parent finishes processing the child block is imported immediately
- Block has blobs, but lookup still has 0 peers. What to do? The lookup will fail and the cached block will be dropped

Example A:

- UnknownParentBlock(parent, block)
- Create lookup for block
- Resolve parent
- Lookup for block dropped because NoPeers

Example B:

- UnknownParentBlock(parent, block)
- Create lookup for block
- UnknownBlockFromAttestation(block)
- Resolve parent
- Lookup for block continues with new peer

### Solution 2

Don't create current block lookup on UnknownParentBlock events, instead cache child components somewhere else. They can be cached in the sync network context. Then when the lookup wants to request blocks, immediately reply with the cached blocks.

This is a bit ugly because it forces the network context into sending sync events. It introduces a cache of blocks that should be pruned with some timeout

We can also choose to not cache child components at all

----

Thoughts? Any better solution?

dapplion:
## Issue Addressed

Description of the problem ⬇️ 
- https://github.com/sigp/lighthouse/issues/5707

## Proposed Changes

Implement solution 1 from https://github.com/sigp/lighthouse/issues/5707

- Only add peers to a lookup that MUST be able to serve the lookup's components
- Allow lookups to have zero peers ONLY IF they are awaiting a parent (some future event will make progress)

## Todo

- [ ] Update lookup tests, behaviour has slightly changed

Closes https://github.com/sigp/lighthouse/issues/5707

### DIFF
--- a/beacon_node/network/src/sync/block_lookups/tests.rs
+++ b/beacon_node/network/src/sync/block_lookups/tests.rs
@@ -67,6 +67,7 @@ type T = Witness<ManualSlotClock, CachingEth1Backend<E>, E, MemoryStore<E>, Memo
 struct TestRig {
     /// Receiver for `BeaconProcessor` events (e.g. block processing results).
     beacon_processor_rx: mpsc::Receiver<WorkEvent<E>>,
+    beacon_processor_rx_queue: Vec<WorkEvent<E>>,
     /// Receiver for `NetworkMessage` (e.g. outgoing RPC requests from sync)
     network_rx: mpsc::UnboundedReceiver<NetworkMessage<E>>,
     /// Stores all `NetworkMessage`s received from `network_recv`. (e.g. outgoing RPC requests)
@@ -127,6 +128,7 @@ impl TestRig {
         let rng = XorShiftRng::from_seed([42; 16]);
         TestRig {
             beacon_processor_rx,
+            beacon_processor_rx_queue: vec![],
             network_rx,
             network_rx_queue: vec![],
             rng,
@@ -283,15 +285,6 @@ impl TestRig {
 
     fn expect_no_active_lookups(&self) {
         self.expect_no_active_single_lookups();
-    }
-
-    fn expect_lookups(&self, expected_block_roots: &[Hash256]) {
-        let block_roots = self
-            .active_single_lookups()
-            .iter()
-            .map(|(_, b, _)| *b)
-            .collect::<Vec<_>>();
-        assert_eq!(&block_roots, expected_block_roots);
     }
 
     fn new_connected_peer(&mut self) -> PeerId {
@@ -544,6 +537,12 @@ impl TestRig {
         }
     }
 
+    fn drain_processor_rx(&mut self) {
+        while let Ok(event) = self.beacon_processor_rx.try_recv() {
+            self.beacon_processor_rx_queue.push(event);
+        }
+    }
+
     fn pop_received_network_event<T, F: Fn(&NetworkMessage<E>) -> Option<T>>(
         &mut self,
         predicate_transform: F,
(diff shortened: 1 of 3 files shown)


AgeManning (approved):
LGTM

beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -2329,106 +2302,37 @@ mod deneb_only {
...
-            .parent_block_imported()
-            .block_response()
-            .blobs_response()
-            .expect_parent_chain_process()
-            .block_imported()
-            .expect_no_active_lookups();
-    }
-
-    #[test]
-    fn parent_block_and_blob_lookup_child_returned_first_blob_trigger() {
```
dapplion, line 2351: Not possible now, child is not requested until parent is imported

beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -2203,99 +2235,44 @@ mod deneb_only {
...
-            .parent_block_response()
-            .parent_blob_response()
-            .expect_block_process()
-            .parent_block_imported()
-            .blobs_response()
-            .expect_parent_chain_process();
-    }
-
-    #[test]
-    fn parent_block_and_blob_lookup_child_returned_first() {
```
dapplion, line 2224: Not possible now, child is not requested until parent is imported

beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -1287,19 +1317,6 @@ fn test_skip_creating_failed_parent_lookup() {
     rig.expect_no_active_lookups();
 }
 
-#[test]
-fn test_skip_creating_failed_current_lookup() {
```
dapplion, line 1291: Not applicable anymore. A peer makes no claim that it has imported the child block, so it should not be downscored

beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -1049,40 +1102,15 @@ fn test_parent_lookup_wrong_response() {
...
+    rig.trigger_unknown_block_from_attestation(block_root, peer_id);
     // Processing succeeds, now the rest of the chain should be sent for processing.
     rig.parent_block_processed_imported(block_root);
     rig.expect_parent_chain_process();
     rig.parent_chain_processed_success(block_root, &[]);
     rig.expect_no_active_lookups();
 }
 
-#[test]
-fn test_parent_lookup_empty_response() {
```
dapplion, line 1060: Duplicated test, child and parent logic are the same now

realbigsean (approved):

realbigsean:
great simplification !

realbigsean:
@mergify queue

PR #5734, opened 2024-05-07
Add EIP-7594 boilerplate 

jimmygchen:
## Issue Addressed

Add EIP-7594 boilerplate. 

This fork is built on top of Deneb, and its types currently have the exact same structure as Deneb types.

This PR adds the boiler plate only, will do a separate PR for:
- Activating PeerDAS at EIP-7594 fork.
- Removing the `PEER_DAS_EPOCH` soft fork config.


dapplion (changes requested):
From offline interop chat discussion the fork boilerplate won't be necessary. 
There is no data structure change in EIP7594.

Testnets will use either of these two configurations:
- `EIP7594_FORK_VERSION = DENEB_FORK_VERSION`, `EIP7594_FORK_EPOCH = 0`: no need for fork triggers, same fork version as deneb
- `EIP7594_FORK_VERSION = DENEB_FORK_VERSION`, `EIP7594_FORK_EPOCH > 0`: same fork version as deneb, having a PEERDAS_ACTIVATION_EPOCH = EIP7594_FORK_EPOCH will result is correct behaviour

Unless PeerDAS requires data structure changes we can keep a `peerdas_activation_epoch` config variable internally and set it to the value of config parameter `EIP7594_FORK_EPOCH`

jimmygchen:
Nice, thanks!
Closing this for now. We already have `PEER_DAS_EPOCH` implemented (happy to rename) and should be sufficient for interop.

PR #5735, opened 2024-05-07
Drop block data from BlockError and BlobError

dapplion:
## Issue Addressed

`BlockError::ParentUnknown` has a full RpcBlock are argument. This arrangement is:
- **unnecessary**: whenever `BlockError::ParentUnknown` is handled a block is already available
- **strange**: requires implementing `into_rpc_block` which is a strange downgrade from types with strong meaning like `MaybeAvailableBlock::Available`

## Proposed Changes

- Drop block data from BlockError and BlobError
- BlockError and a bunch of similar types are no longer generic
- Drop `into_rpc_block`

### DIFF
--- a/beacon_node/beacon_chain/src/blob_verification.rs
+++ b/beacon_node/beacon_chain/src/blob_verification.rs
@@ -21,7 +21,7 @@ use types::{
 };
 
 /// An error occurred while validating a gossip blob.
-#[derive(Debug, strum::Display)]
+#[derive(Debug)]
 pub enum GossipBlobError {
     /// The blob sidecar is from a slot that is later than the current slot (with respect to the
     /// gossip clock disparity).
@@ -143,6 +143,12 @@ pub enum GossipBlobError {
     /// It's unclear if this block is valid, but it conflicts with finality and shouldn't be
     /// imported.
     NotFinalizedDescendant { block_parent_root: Hash256 },
+}
+
+impl std::fmt::Display for GossipBlobError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self)
+    }
 }
 
 impl From<BeaconChainError> for GossipBlobError {
--- a/beacon_node/beacon_chain/src/blob_verification.rs
+++ b/beacon_node/beacon_chain/src/blob_verification.rs
@@ -21,8 +21,8 @@ use types::{
 };
 
 /// An error occurred while validating a gossip blob.
-#[derive(Debug)]
-pub enum GossipBlobError<E: EthSpec> {
+#[derive(Debug, strum::Display)]
+pub enum GossipBlobError {
     /// The blob sidecar is from a slot that is later than the current slot (with respect to the
     /// gossip clock disparity).
     ///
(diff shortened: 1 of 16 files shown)


beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -414,30 +414,30 @@ impl From<execution_layer::Error> for ExecutionPayloadError {
...
     }
 }
 
-impl<E: EthSpec> std::fmt::Display for BlockError<E> {
+impl std::fmt::Display for BlockError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            BlockError::ParentUnknown(block) => {
-                write!(f, "ParentUnknown(parent_root:{})", block.parent_root())
+            BlockError::ParentUnknown { parent_root, .. } => {
```
realbigsean: This special case handling shouldn't be necessary anymore

realbigsean (approved):
Looks like a great simplification, glad to get rid of all these generics

beacon_node/beacon_chain/src/blob_verification.rs
```diff
@@ -143,28 +143,19 @@ pub enum GossipBlobError<E: EthSpec> {
     NotFinalizedDescendant { block_parent_root: Hash256 },
 }
 
-impl<E: EthSpec> std::fmt::Display for GossipBlobError<E> {
+impl std::fmt::Display for GossipBlobError {
```
pawanjay176, line 148: Can just derive strum::Display maybe?

beacon_node/beacon_chain/src/block_verification.rs
pawanjay176: Same as above, can derive display now

pawanjay176 (approved):
Nice

michaelsproul:
CI failing a lot

jimmygchen:
Just need a passing CI before we can merge this

beacon_node/beacon_chain/src/blob_verification.rs
```diff
@@ -145,28 +145,19 @@ pub enum GossipBlobError<E: EthSpec> {
...
-            GossipBlobError::BlobParentUnknown(blob_sidecar) => {
-                write!(
-                    f,
-                    "BlobParentUnknown(parent_root:{})",
-                    blob_sidecar.block_parent_root()
-                )
-            }
-            other => write!(f, "{:?}", other),
-        }
+        write!(f, "{:?}", self)
```
dapplion, line 150: Kept this implementation of Display as some release tests assert the full error description. I'm not sure of which consumers expect the full error details in Display formatting. If we switch to `strum::Display` we may lose information in some error logs throughout the codebase. Thoughts?

dapplion:
CI gud

common/lighthouse_metrics/src/lib.rs
michaelsproul, line 1: this file got reverted to an old version during a merge commit

michaelsproul (approved):

common/lighthouse_metrics/src/lib.rs
michaelsproul: fixed in https://github.com/sigp/lighthouse/pull/5735/commits/bacd93d577e47ad3ad252982062c3903211f0f48

michaelsproul:
@mergify queue

PR #5738, opened 2024-05-08
Log block import source

### LINKED ISSUE #5721: Track block source on import log / metric
## Description

When debugging it's useful to know if an imported block was sourced from gossip / lookup / range sync. This information can be ultimately deduced by inspecting logs, but it's time-consuming. It would be just simpler to append a `source` property in the import log.

eserilev:
## Issue Addressed

Closes #5721

## Proposed Changes

Add a source tag to block import logs to indicate where the imported block came from. Options are Lookup, Gossip, RangeSync and HttpApi

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -3055,7 +3055,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
                     "Beacon block imported";
                     "block_root" => ?block_root,
                     "block_slot" => block_slot,
-                    "source" => block_source.to_string(),
+                    "source" => %block_source,
                 );
 
                 // Increment the Prometheus counter for block processing successes.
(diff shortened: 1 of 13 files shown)


eserilev:
image
example logs

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3048,6 +3055,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
                     "Beacon block imported";
                     "block_root" => ?block_root,
                     "block_slot" => block_slot,
+                    "source" => block_source.to_string(),
```
dapplion: You should able to do this if you implement display
```
"source" => %block_source,
```

dapplion (approved):
Thank you for implementing this!

jmcph4:
> image example logs

Love this!

realbigsean:
@mergify queue

PR #5742, opened 2024-05-08
fix get attesting indices

realbigsean:
## Issue Addressed

This check comes from `process_attestation` in the spec and has been updated:
`aggregation_bits.len() != beacon_committee.committee.len()`

This is the new spec logic: 

```
    committee_indices = get_committee_indices(attestation.committee_bits)
    participants_count = 0
    for index in committee_indices:
        assert index < get_committee_count_per_slot(state, data.target.epoch)
        committee = get_beacon_committee(state, data.slot, index)
        participants_count += len(committee)

    assert len(attestation.aggregation_bits) == participants_count
```

### DIFF
--- a/consensus/state_processing/src/common/get_attesting_indices.rs
+++ b/consensus/state_processing/src/common/get_attesting_indices.rs
@@ -119,7 +119,7 @@ pub mod attesting_indices_electra {
             if let Some(&beacon_committee) = committees_map.get(&index) {
                 // This check is new to the spec's `process_attestation` in Electra.
                 if index >= committee_count_per_slot {
-                    return Err(BeaconStateError::InvalidBitfield);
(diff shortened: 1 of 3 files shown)


consensus/state_processing/src/common/get_attesting_indices.rs
```diff
@@ -138,8 +142,11 @@ pub mod attesting_indices_electra {
             } else {
                 return Err(Error::NoCommitteeFound);
             }
+        }
 
-            // TODO(electra) what should we do when theres no committee found for a given index?
```
realbigsean, line 142: Also i removed this todo because it looks like we error in this case already (L139) which looks right to me

eserilev:
this looks good to me!

PR #5743, opened 2024-05-08
Electra engine api

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Implements electra engine api. 

## Additional Info

The spec does not handle the `engine_getPayloadBodies` for the new electra fields. We need to add it to the spec and implement it.
This PR currently implements it in the manner that I think it will be implemented (i.e. by extending `ExecutionPayloadBodyV1` to include `deposit_receipts` and `withdrawal_requests`. 
Working on getting a spec out for this.


beacon_node/execution_layer/src/engine_api.rs
```diff
@@ -613,35 +635,38 @@ impl<E: EthSpec> ExecutionPayloadBodyV1<E> {
...
-                    Err(format!(
-                        "block {} is post-capella but payload body doesn't have withdrawals",
+                let (Some(withdrawals), Some(deposit_receipts), Some(withdrawal_requests)) = (
+                    self.withdrawals,
+                    self.deposit_receipts,
+                    self.withdrawal_requests,
+                ) else {
+                    return Err(format!(
+                        "block {} is post-electra but payload body doesn't have withdrawals/deposit_receipts/withdrawal_requests \
+                        Check that ELs are returning receipts and withdrawal_requests in getPayloadBody requests",
```
dapplion: Making this error per property will make debugging easier by knowing which property is missing

beacon_node/execution_layer/src/engine_api.rs
```diff
@@ -63,6 +67,8 @@ pub enum Error {
     TransitionConfigurationMismatch,
     SszError(ssz_types::Error),
```
dapplion, line 67: Could rename this one to `DeserializeTransactions` to match the others

beacon_node/execution_layer/src/engine_api/http.rs
```diff
@@ -926,19 +930,43 @@ impl HttpJsonRpc {
...
                 Ok(JsonGetPayloadResponse::V3(response).into())
             }
+            ForkName::Base
+            | ForkName::Altair
+            | ForkName::Bellatrix
+            | ForkName::Capella
+            | ForkName::Electra => Err(Error::UnsupportedForkVariant(format!(
+                "called get_payload_v3 with {}",
+                fork_name
+            ))),
```
dapplion, line 940: `if let ForkName::Deneb = fork_name else { err }` for less future fork boilerplate :)

beacon_node/execution_layer/src/engine_api/json_structures.rs
```diff
@@ -101,6 +104,12 @@ pub struct JsonExecutionPayload<E: EthSpec> {
     #[superstruct(only(V3, V4))]
     #[serde(with = "serde_utils::u64_hex_be")]
     pub excess_blob_gas: u64,
+    #[superstruct(only(V4))]
+    // TODO(electra): Field name should be changed post devnet-0. See https://github.com/ethereum/execution-apis/pull/544
+    pub deposit_requests: VariableList<JsonDepositRequest, E::MaxDepositReceiptsPerPayload>,
```
dapplion: Current spec is still `deposit_requests` https://github.com/ethereum/execution-apis/blob/3ae3d29fc9900e5c48924c238dff7643fdc3680e/src/engine/prague.md

beacon_node/execution_layer/src/engine_api.rs
```diff
@@ -197,6 +203,10 @@ pub struct ExecutionBlockWithTransactions<E: EthSpec> {
     #[superstruct(only(Deneb, Electra))]
     #[serde(with = "serde_utils::u64_hex_be")]
     pub excess_blob_gas: u64,
+    #[superstruct(only(Electra))]
+    pub deposit_receipts: Vec<JsonDepositRequest>,
```
dapplion: This can just be named `DepositRequest` as the consensus data structure is `DepositReceipt`. Or if we need to namespace call it `EngineDepositRequest` or similar

dapplion (approved):
These changes are quite straightforward, I see them correct according to current spec

Is it intentional to merge epoch processing changes into this branch?

realbigsean:
> Is it intentional to merge epoch processing changes into this branch?

Yes since we were planning to merge it into unstable after epoch processing, re-targeted the branch to fix the diff

realbigsean:
Something that's missing is the payload bodies by range changes: https://github.com/ethereum/execution-apis/pull/545

This PR implements an extension of the existing V1 endpoint, what's currently spec'd is to actually add a V2 endpoint: https://github.com/ethereum/execution-apis/pull/545#issuecomment-2115215133

There may be further discussion about whether to just extend v1 but for now we won't be able to serve blocks whose payloads we've pruned until we implement this as a v2

We could use the same `ExecutionPayloadBody` type for both v1 and v2 because v2 is fully backwards compatible but there will be annoying fork-switching logic we we want to query the v2 endpoint after the electra fork so we're sure the EL supports it

beacon_node/execution_layer/src/engine_api.rs
pawanjay176: `TransitionConfigurationMismatch` doesn't seem to be used anywhere anymore so just removed it

beacon_node/execution_layer/src/engine_api/http.rs
pawanjay176: We usually match all fork names so that we don't miss potential changes to functions on addition of new forks.
For e.g. if ForkName::F* had some changes that affect this function, then we might potentially miss it in a refactor if we do.the if let pattern. 
It won't affect this specific function, but I think its good to be consistent.

pawanjay176:
Strong agree on extending V1. We can implement the V2 method but its unnecessarily complex and gains us nothing imo.

beacon_node/execution_layer/src/engine_api/http.rs
ethDreamer: I think we've kinda shifted policy here towards being less verbose, or at least considering it on a case by case basis. In this case it makes sense that any fork post-electra shouldn't be calling this.

beacon_node/beacon_chain/src/electra_readiness.rs
```diff
@@ -22,7 +20,7 @@ pub enum ElectraReadiness {
     /// The execution engine is electra-enabled (as far as we can tell)
     Ready,
     /// We are connected to an execution engine which doesn't support the V3 engine api methods
```
realbigsean:
```suggestion
    /// We are connected to an execution engine which doesn't support the V4 engine api methods
```

consensus/types/src/chain_spec.rs
```diff
@@ -180,7 +180,8 @@ pub struct ChainSpec {
     pub electra_fork_version: [u8; 4],
     /// The Electra fork epoch is optional, with `None` representing "Electra never happens".
     pub electra_fork_epoch: Option<Epoch>,
-    pub unset_deposit_receipts_start_index: u64,
+    // TODO(pawan): this shouldn't be a config parameter?
```
realbigsean:
```suggestion
```
Will be resolved in 
https://github.com/sigp/lighthouse/pull/6028

realbigsean (approved):
woop!

beacon_node/execution_layer/src/engine_api/http.rs
realbigsean: we can leave this one since this is in line with all the other similar methods in this file. But we can come back and change them all to reduce boiler plate

realbigsean:
@mergify queue

PR #4575, opened 2023-08-07
Assume Content-Type is json for endpoints that require json

eserilev:
## Issue Addressed

#4568

## Proposed Changes

Create a custom filter that parses json in the request body by default without checking the content-type header. Apply the filter to endpoints that expect JSON in the request body.

## Additional Info

The built in warp json filter `warp::body::json` defaults to `Content-Type application/json` if a Content-Type isn't provided. If a `Content-Type` is provided and it is not `application/json` the request will fail.

I ran several tests on my own, and was unable to recreate a failed endpoint request when `Content-Type` was omitted. I was however able to recreate a failed endpoint request when `Content-Type` was included but wasn't `application/json`.

-----

In out test suite note that `builder.json()` appends `Content-Type application/json` to the header by default.

I created a new function `post_generic_json_without_content_type_header` which creates a post request with serialized json in the body without including `Content-Type application/json` in the header.

Locally I replaced all post requests in our test suite to NOT include the `Content-Type` header and all the tests passed. In this PR I only updated a single test to send a POST request without the `Content-Type` header.

I plan on also spinning up the client locally and making some curl requests myself to make sure I haven't missed anything.

PR #5744, opened 2024-05-09
Beacon api + validator electra

realbigsean:
## Issue Addressed

- implements this block production change https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/validator.md#deposits

### Actual endpoint changes
- [x] `GET /eth/v2/validator/aggregate_attestation` - this now requires `committee_index` as a query param

### Just the types change
We may need to filter for the `consensus_version_header` in requests, add it to responses, and wrap responses in `ForkVersionedData`. I don't think any of these are required to a LH  BN<>VC connection so I don't think these need to be implemented until they are spec'd.
- [x] `GET /eth/v1/beacon/pool/attestations`
- [x] `POST /eth/v1/beacon/pool/attestations`
- [x] `GET /eth/v1/beacon/pool/attester_slashings`
- [x] `POST /eth/v1/beacon/pool/attester_slashings/`
- [x] `GET /eth/v1/beacon/blocks/{block_id}/attestations`
- [x] `POST /eth/v1/validator/aggregate_and_proofs`


realbigsean:
spec pr https://github.com/ethereum/beacon-APIs/pull/447

dapplion (approved):
Straightforward changes, they match the route as spec'ed here https://github.com/ethereum/beacon-APIs/pull/447

realbigsean:
I need to make the fork versioned updates that have been spec'd now: https://github.com/ethereum/beacon-APIs/pull/448

beacon_node/http_api/src/lib.rs
```diff
@@ -1807,20 +1834,40 @@ pub fn serve<T: BeaconChainTypes>(
...
+                        .or_else(|| {
+                            attestations
+                                .first()
+                                .map(|att| att.data().slot)
+                                .or_else(|| chain.slot_clock.now())
+                        })
+                        .ok_or(warp_utils::reject::custom_server_error(
+                            "unable to read slot clock".to_string(),
+                        ))?;
+                    let fork_name = chain.spec.fork_name_at_slot::<T::EthSpec>(slot);
```
dapplion: What if someone queries the pool at the fork boundary? There may be attestations from both forks

beacon_node/http_api/src/lib.rs
```diff
@@ -1857,15 +1904,29 @@ pub fn serve<T: BeaconChainTypes>(
...
+                task_spawner.blocking_response_task(Priority::P1, move || {
+                    let slashings = chain.op_pool.get_all_attester_slashings();
+                    let slot = slashings
+                        .first()
+                        .map(|slashing| slashing.attestation_1().data().slot)
+                        .or_else(|| chain.slot_clock.now())
+                        .ok_or(warp_utils::reject::custom_server_error(
+                            "unable to read slot clock".to_string(),
+                        ))?;
+                    let fork_name = chain.spec.fork_name_at_slot::<T::EthSpec>(slot);
```
dapplion: Not a reliable way to get the attester slashing fork. One can have an attester slashing electra with an attestation data from deneb. Instead we should implement `fork_name_unchecked` in `slashing`

beacon_node/http_api/src/lib.rs
```diff
@@ -3184,29 +3245,45 @@ pub fn serve<T: BeaconChainTypes>(
...
+            |endpoint_version: EndpointVersion,
+             query: api_types::ValidatorAggregateAttestationQuery,
              not_synced_filter: Result<(), Rejection>,
              task_spawner: TaskSpawner<T::EthSpec>,
              chain: Arc<BeaconChain<T>>| {
                 task_spawner.blocking_json_task(Priority::P0, move || {
                     not_synced_filter?;
-                    chain
-                        .get_pre_electra_aggregated_attestation_by_slot_and_root(
+                    let res = if endpoint_version == V2 {
```
dapplion, line 3276: should `match`?

beacon_node/http_api/src/lib.rs
pawanjay176: Why aren't we always calculating the fork based on `slot_clock.now()` here?

consensus/types/src/attestation.rs
```diff
@@ -487,6 +487,46 @@ impl<'a, E: EthSpec> From<AttestationRefOnDisk<'a, E>> for AttestationRef<'a, E>
...
+        } else {
+            let attestation: AttestationBase<E> =
+                serde_json::from_value(value).map_err(serde::de::Error::custom)?;
+            Ok(Attestation::Base(attestation))
+        }
+    }
+}
+
+impl<E: EthSpec> ForkVersionDeserialize for Vec<Attestation<E>> {
+    fn deserialize_by_fork<'de, D: serde::Deserializer<'de>>(
```
pawanjay176, line 514: Can we not have a `Vec<Attestation>` with both variants?

common/eth2/src/lib.rs
```diff
@@ -395,6 +408,34 @@ impl BeaconNodeHttpClient {
...
+        body: &T,
+        timeout: Option<Duration>,
+        fork: ForkName,
+    ) -> Result<Response, Error> {
+        let mut builder = self.client.post(url);
+        if let Some(timeout) = timeout {
+            builder = builder.timeout(timeout);
+        }
+
+        let serialized_body = serde_json::to_vec(body).map_err(Error::InvalidJson)?;
```
pawanjay176: I didn't get why we are doing it this way for attester slashings. Why not just post the json with the content header?

beacon_node/http_api/src/lib.rs
realbigsean: that's a great point.. looks like we'll have to convert to the new format for all of these attestations

Seems to mean Pawan's idea is the best way to approach it: https://github.com/sigp/lighthouse/pull/5744#discussion_r1672185892

beacon_node/http_api/src/lib.rs
realbigsean: Good catch

> Why aren't we always calculating the fork based on slot_clock.now() here?

I think this makes sense considering return types could be mixed for a given `Eth-Consensus-Version` as pointed out here  https://github.com/sigp/lighthouse/pull/5744#discussion_r1671810990

consensus/types/src/attestation.rs
realbigsean: I think with these updates no: 386aacda2ab6032ed84fbea7b7d847b0cb5a86d6

This traits for when JSON wraps a data structure with a fork name like this:
```
{
  "version": "phase0",
  "finalized": false,
  "data": [
    {
      "aggregation_bits": "0x01",
      "data": {
        "slot": "1",
        "index": "1",
        "beacon_block_root": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2",
        "source": {
          "epoch": "1",
          "root": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2"
        },
        "target": {
          "epoch": "1",
          "root": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2"
        }
      },
      "signature": "0x1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505cc411d61252fb6cb3fa0017b679f8bb2305b26a285fa2737f175668d0dff91cc1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505"
    }
  ]
}
```

The fork name's meant to be used to deserialize the struct, so i think a consistent fork through the list is implied. More specifically, this is used for when we are displaying attestations from a specific block, which should be uniform, or from the op pool, which should be uniform with the changes listed

common/eth2/src/lib.rs
realbigsean: i sort of just went with what was there already. I actually think this was just included for testing this PR and should have been reverted https://github.com/sigp/lighthouse/pull/4575

beacon_node/http_api/src/lib.rs
realbigsean: raised a spec issue as well https://github.com/ethereum/beacon-APIs/issues/458

common/eth2/src/lib.rs
realbigsean: updated 71a2eadc467040a4a2996177e557673e5dc104da

beacon_node/http_api/src/lib.rs
realbigsean: `V2` is just a struct wrapper around a `u64` not an enum

```
pub const V2: EndpointVersion = EndpointVersion(2);

#[derive(Debug, Clone, Copy, PartialEq)]
pub struct EndpointVersion(pub u64);
```

realbigsean:
Thanks for the reviews! Updates have been made

pawanjay176 (approved):

realbigsean:
In the issue I made on the spec repo, sounds like other teams just drop messages from the op pool that are from the old fork. How do we feel about going that route instead? Might be a little more straightforward for an end user because those message won't actually make it into a block.

realbigsean:
@mergify queue

PR #5751, opened 2024-05-09
P2p electra

realbigsean:
## Issue Addressed

adds the following new gossip conditions for both unaggregated and aggregated attestations:

```
    [REJECT] len(committee_indices) == 1, where committee_indices = get_committee_indices(aggregate).
    [REJECT] aggregate.data.index == 0
```


realbigsean:
these changes ended up being needed upstream so, I'm going to remove this from the conga line 🫡

PR #5752, opened 2024-05-09
Compute on chain aggregate impl

eserilev:
Added logic to compute the on chain aggregate attestation for electra

`aggregate_across_committees`

iterates through a list of attestations with a shared `Checkpoint`. We then aggregate these attestations by committee. If they are electra attestations we further aggregate via `compute_on_chain_aggregate`


michaelsproul (approved):
I'm going to merge this PR into my branch and then make a few small tweaks. Thanks!

PR #5772, opened 2024-05-13
PeerDAS spec tests

jimmygchen:
## Issue Addressed

Part of #4983.

- Add PeerDAS spec tests
- Fix overflow and ordering in `get_custody_columns` function
- Load KZG only once to speed up tests


dapplion (approved):
Nice debugging!

PR #5774, opened 2024-05-13
Don't request block components until having block

dapplion:
## Issue Addressed

Current stable when creating a new block lookup for `block_root` will:
- Request block 
- Request all possible blobs for `block_root` before knowing if the block has blobs and how many

This optimistic / blind request forces us to handle the case where we have processed blobs + blocks and turns out that the blob peer did not return enough blobs.

By delaying fetching blobs until we get the block we can be sure to:
- request blobs only when block has data
- know exactly how many blobs we expect and penalize the peer immediately if it returns less blobs than expected

## Proposed Changes

- Update `ActiveBlobsByRootRequest` to expect peer to return exactly the number of requested blobs. Else error with `NotEnoughResponsesReturned`
- If `AvailabilityProcessingStatus::MissingComponents && both_components_processed == true` hard error as an internal error. The condition above should ensure that this never happens
- To cover the case of `single_block_response_then_too_many_blobs_response_attestation` the sync network context can produce multiple return values for a single request. In the test case it produces: (1) `Some(blobs)`, (2) `Err(TooManyResponses)`. However block lookup mod expects a single result, so we track if the request has been resolved (= a result value has been returned). If resolved, downscore but do not forward the result

### DIFF
--- a/beacon_node/network/src/sync/block_lookups/tests.rs
+++ b/beacon_node/network/src/sync/block_lookups/tests.rs
@@ -287,6 +287,11 @@ impl TestRig {
         self.expect_no_active_single_lookups();
     }
 
+    fn expect_no_active_lookups_empty_network(&mut self) {
+        self.expect_no_active_lookups();
+        self.expect_empty_network();
+    }
+
     fn new_connected_peer(&mut self) -> PeerId {
         let peer_id = PeerId::random();
         self.network_globals
@@ -461,14 +466,16 @@ impl TestRig {
         );
     }
 
-    fn complete_single_lookup_block_valid(&mut self, block: SignedBeaconBlock<E>, import: bool) {
+    fn complete_lookup_block_download(&mut self, block: SignedBeaconBlock<E>) {
         let block_root = block.canonical_root();
-        let block_slot = block.slot();
         let id = self.expect_block_lookup_request(block_root);
         self.expect_empty_network();
         let peer_id = self.new_connected_peer();
         self.single_lookup_block_response(id, peer_id, Some(block.into()));
         self.single_lookup_block_response(id, peer_id, None);
+    }
+
+    fn complete_lookup_block_import_valid(&mut self, block_root: Hash256, import: bool) {
         self.expect_block_process(ResponseType::Block);
         let id = self.find_single_lookup_for(block_root);
         self.single_block_component_processed(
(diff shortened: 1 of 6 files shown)


beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -2044,77 +2020,27 @@ mod deneb_only {
...
-        let Some(tester) = DenebTester::new(RequestTrigger::AttestationUnknownBlock) else {
-            return;
-        };
-        tester
-            .blobs_response() // hold blobs for processing
-            .block_response_triggering_process()
-            .block_missing_components() // blobs not yet imported
-            .blobs_response_was_valid()
-            .blob_imported(); // now blobs resolve as imported
-    }
```
dapplion, line 2065: Blobs can't be returned first

beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -2044,77 +2020,27 @@ mod deneb_only {
...
-
-        tester
-            .blobs_response()
-            .expect_no_penalty_and_no_requests()
-            // blobs not sent for processing until the block is processed
-            .empty_block_response()
-            .expect_penalty("NoResponseReturned")
-            .expect_block_request()
-            .expect_no_blobs_request();
-    }
```
dapplion, line 2116: Scenario already covered by `empty_block_is_retried`

beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -2044,77 +2020,27 @@ mod deneb_only {
...
-            .expect_penalty("NoResponseReturned")
-            .expect_block_request()
-            .expect_no_blobs_request()
-            .empty_blobs_response()
-            .expect_empty_beacon_processor()
-            .expect_no_penalty()
-            .expect_no_block_request()
-            .expect_no_blobs_request()
-            .block_response_triggering_process()
-            .missing_components_from_block_request();
```
dapplion, line 2083: Scenario already covered by `empty_block_is_retried`

beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -2171,37 +2096,12 @@ mod deneb_only {
...
-    fn too_few_blobs_response_then_block_response_attestation() {
-        let Some(tester) = DenebTester::new(RequestTrigger::AttestationUnknownBlock) else {
-            return;
-        };
-        tester
-            .invalidate_blobs_too_few()
-            .blobs_response() // blobs are not sent until the block is processed
-            .expect_no_penalty_and_no_requests()
-            .block_response_triggering_process();
-    }
```
dapplion, line 2189: Can't respond blobs first, same for `too_many_blobs_response_then_block_response_attestation`

YoutacRandS-VA (approved):

realbigsean (approved):

realbigsean:
@mergify queue

PR #5775, opened 2024-05-13
Make sampling tests deterministic

dapplion:
## Issue Addressed

Sampling tests fail sometimes because the peer's ENR is randomized, so it's possible that we get a subnet with zero peers

## Proposed Changes

Add a supernode peer so that all columns have at least one peer


jimmygchen (approved):
Looks good!

PR #5776, opened 2024-05-13
Trigger sampling on sync events

dapplion:
## Issue Addressed

Currently we only trigger sampling on gossip blocks.

## Proposed Changes

Trigger sampling for blocks:
- Imported on range sync
- Imported on lookup sync

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -6830,10 +6830,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
     /// TODO(das): check if the block is still within the da_window
     pub fn should_sample_slot(&self, slot: Slot) -> bool {
         self.spec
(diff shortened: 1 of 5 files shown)


beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -6823,6 +6826,16 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         self.data_availability_checker.data_availability_boundary()
     }
 
+    /// Returns true if we should issue a sampling request for this block
+    /// TODO(das): check if the block is still within the da_window
+    pub fn should_sample_slot(&self, slot: Slot) -> bool {
+        self.spec
```
jimmygchen, line 6832: nit:

```rust
self.spec.is_peer_das_enabled_for_epoch(slot.epoch(T::EthSpec::slots_per_epoch()))
```

YoutacRandS-VA (approved):

PR #5777, opened 2024-05-13
Enforce sync lookup receives a single result

dapplion:
## Issue Addressed

Current stable lookup requests include an internal request counter to differentiate requests. Because each lookup manages retries and streams it must handle interleaved packets from different requests.

PR removed the need for this internal counter, where network context handles streams and uniquely tags each request
- https://github.com/sigp/lighthouse/pull/5583

However, a bug may break the invariant of 1 request at a time per lookup. If this happens, it will result in undefined behavior.

## Proposed Changes

- Each lookup keeps a unique ID of the current inflight requests.
- If it gets a download result from any other ID, consider it an internal bug and hard error

This change is quite strict but ensures that in the case of duplicate requests, we don't mix errors and results from other requests. The hard error is easy to pick up from metrics (in `sync_lookups_dropped_total{reason = "BadState"}`) and logs.


realbigsean (approved):

realbigsean:
@mergify queue

PR #5778, opened 2024-05-14
Log stuck lookups

dapplion:
## Issue Addressed

Debugging sync lookup issues that involve a lookup being stuck are very time consuming. To locate which lookup is actually stuck one needs to correlate events from a log indexer and deduce which one has not been dropped.

Then, there's no clear visibility on the state of the lookup. One has to reconstruct the state from a sequence of events, which is possible but very time consuming.

Debugging could be much faster if we knew two things:
- Which lookup is stuck
- What is its internal state

We can expose that data over HTTP API but it's a bit ugly since sync should regularly post this data on a global state. Using metrics is problematic as labeling by block_root would increase the cardinality of metrics too much. So dumping to debug logs is easy to implement, and easy to consume.

## Proposed Changes

- Dump a short summary of all block lookups if they still exist 60 seconds after creation.
- Add metric to track lookups that are older than 60 seconds

### DIFF
--- a/beacon_node/network/src/sync/block_lookups/mod.rs
+++ b/beacon_node/network/src/sync/block_lookups/mod.rs
@@ -678,6 +678,7 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
     }
 
     pub fn log_stuck_lookups(&self) {
+        let mut stuck_count = 0;
         for lookup in self.single_block_lookups.values() {
             if lookup.elapsed_since_created() > Duration::from_secs(LOOKUP_MAX_DURATION_SECS) {
                 debug!(self.log, "Lookup maybe stuck";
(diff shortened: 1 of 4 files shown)


beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -675,4 +676,21 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
...
         );
     }
+
+    pub fn log_stuck_lookups(&self) {
+        let mut stuck_count = 0;
+        for lookup in self.single_block_lookups.values() {
+            if lookup.elapsed_since_created() > Duration::from_secs(LOOKUP_MAX_DURATION_SECS) {
+                debug!(self.log, "Lookup maybe stuck";
+                    // Fields id and block_root are also part of the summary. However, logging them
+                    // here allows log parsers o index them and have better search
```
realbigsean, line 676:
```suggestion
                    // here allows log parsers to index them and have better search
```

realbigsean (approved):

realbigsean:
@mergify queue

realbigsean:
@mergify queue

dapplion:
@mergifyio refresh

realbigsean:
@mergify queue

PR #5779, opened 2024-05-14
PeerDAS parameter changes for devnet-0

jimmygchen:
## Issue Addressed

- Update min custody requirement to `4`
- Update data column req/resp limits and quota


jimmygchen:
Hold off merging this now - our config names are outdated from the spec, I'll update them now

dapplion (approved):

PR #5781, opened 2024-05-14
Implement unconditional reconstruction for supernodes

dapplion:
Part of

After inserting a valid custody data column into the da_checker, if the following conditions are met:
- Custody requirements == all columns
- Count of verified columns >= 50% of columns

Reconstruct all columns, import the block components, and gossip publish the new columns.

This flow is triggered by columns received from gossip and lookup sync.


jimmygchen (approved):
Lets go!

PR #5782, opened 2024-05-14
Subscribe to the correct subnets for electra attestations

eserilev:
Attestation subnet subscription logic was using `AttestationData.index` for electra attestations. This PR fixes this by looking at the first set bit in the `committee_bits` field to calculate the correct subnet for electra attestations.

### DIFF
--- a/beacon_node/beacon_chain/src/attestation_verification.rs
+++ b/beacon_node/beacon_chain/src/attestation_verification.rs
@@ -816,8 +816,7 @@ impl<'a, T: BeaconChainTypes> IndexedUnaggregatedAttestation<'a, T> {
         committees_per_slot: u64,
         subnet_id: Option<SubnetId>,
         chain: &BeaconChain<T>,
-    ) -> Result<(u64, SubnetId), Error> {
-        // TODO(electra) if we want
+    ) -> Result<(u64, SubnetId), Error> { 
         let expected_subnet_id = SubnetId::compute_subnet_for_attestation::<T::EthSpec>(
             &attestation,
             committees_per_slot,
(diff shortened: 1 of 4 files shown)


beacon_node/beacon_chain/src/attestation_verification.rs
```diff
@@ -817,8 +817,9 @@ impl<'a, T: BeaconChainTypes> IndexedUnaggregatedAttestation<'a, T> {
         subnet_id: Option<SubnetId>,
         chain: &BeaconChain<T>,
     ) -> Result<(u64, SubnetId), Error> {
-        let expected_subnet_id = SubnetId::compute_subnet_for_attestation_data::<T::EthSpec>(
-            indexed_attestation.data(),
+        // TODO(electra) if we want
+        let expected_subnet_id = SubnetId::compute_subnet_for_attestation::<T::EthSpec>(
+            &attestation,
```
eserilev, line 822: the indexed_attestaton doesnt have committee_bits as a field, so we use the provided `AttestationRef` instead. Is that ok?

beacon_node/beacon_chain/src/attestation_verification.rs
```diff
@@ -816,9 +816,9 @@ impl<'a, T: BeaconChainTypes> IndexedUnaggregatedAttestation<'a, T> {
         committees_per_slot: u64,
         subnet_id: Option<SubnetId>,
         chain: &BeaconChain<T>,
-    ) -> Result<(u64, SubnetId), Error> {
-        let expected_subnet_id = SubnetId::compute_subnet_for_attestation_data::<T::EthSpec>(
-            indexed_attestation.data(),
+    ) -> Result<(u64, SubnetId), Error> { 
+        let expected_subnet_id = SubnetId::compute_subnet_for_attestation::<T::EthSpec>(
+            &attestation,
             committees_per_slot,
```
eserilev, line 822: using the provided attestation, instead of the indexed_attestation here. is that ok?

realbigsean (approved):
nice!

beacon_node/beacon_chain/src/attestation_verification.rs
realbigsean: Can't think of a reason to use indexed_attestation

PR #5783, opened 2024-05-14
Add column gossip verification and handle unknown parent block

jimmygchen:
## Issue Addressed

Part of #4983.

- Add data column gossip verification.
- Handle missing parent block for columns.

## TODO

- [ ] Verify Inclusion proof


beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -660,11 +660,81 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                            "parent_root" => %column.block_parent_root(),
+                        );
+                        self.send_sync_message(SyncMessage::UnknownParentDataColumn(
+                            peer_id, column,
+                        ));
+                    }
+                    GossipDataColumnError::KzgNotInitialized
+                    | GossipDataColumnError::PubkeyCacheTimeout
+                    | GossipDataColumnError::BeaconChainError(_) => {
+                        crit!(
```
dapplion, line 702: Add `TODO(das)` on each log to downgrade them latter

jimmygchen: Why downgrade? These are internal errors and I think it would be useful to catch them in testing.

jimmygchen: ah I see what you mean. You want a non crit in prod. These ar also crit for blocks and blobs though.

dapplion: Makes sense

dapplion (approved):

PR #5787, opened 2024-05-15
Verify inclusion proof should not be fallible

dapplion:
## Issue Addressed

Adding the verify inclusion proof function for data columns I noticed that `verify_blob_sidecar_inclusion_proof` is fallible due to the need to do math with the EthSpec variables.

## Proposed Changes

Since the EthSpec values are non-configurable and properly chosen, it's not necessary to make `verify_blob_sidecar_inclusion_proof` fallible. A unit test in EthSpec ensures that the values are correctly chosen and then we can safely expect.


realbigsean (approved):
Makes sense to me

jimmygchen (approved):
Looks good! I think this is a low risk change if tested (as things would just blow up if this is doesn't work) and should be good to merge.

I wrote some tests here and they pass on your branch:
https://github.com/dapplion/lighthouse/pull/33

realbigsean:
@mergify queue

PR #5797, opened 2024-05-16
Update subscription filter.

jimmygchen:
## Issue Addressed

I observed that one of the nodes were unable to subscribe to all column subnets after bumping the column subnet count. 

This is due to an outdated subscription filter, and using an incorrect configuration. 

This should fix the constant occurence of "unknown parent for column".


dapplion (approved):
Send it

PR #5808, opened 2024-05-17
Type sync network context send errors

dapplion:
## Issue Addressed

SyncNetworkContext has a bunch of functions that send messages to network channel and processor channel with signature

```
) -> Result<LookupRequestResult, &'static str> {
```

So far this is okay-ish, but in das we introduce more complex failure modes that can't be coerced to a static string. See:

https://github.com/sigp/lighthouse/blob/8059c3aa39ed989ac6a8d57bd2b7f46abaf024a2/beacon_node/network/src/sync/network_context.rs#L681

## Proposed Changes

Type sync network context send errors in unstable, and backport to das when merged

### DIFF
--- a/beacon_node/network/src/sync/network_context.rs
+++ b/beacon_node/network/src/sync/network_context.rs
@@ -52,7 +52,7 @@ pub enum RpcEvent<T> {
     RPCError(RPCError),
 }
 
-pub type RpcRequestResult<T> = Result<(T, Duration), RpcResponseError>;
+pub type RpcResponseResult<T> = Result<(T, Duration), RpcResponseError>;
 
 pub enum RpcResponseError {
     RpcError(RPCError),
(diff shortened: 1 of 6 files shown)


jimmygchen (approved):

beacon_node/network/src/sync/network_context.rs
```diff
@@ -52,31 +52,43 @@ pub enum RpcEvent<T> {
     RPCError(RPCError),
 }
 
-pub type RpcProcessingResult<T> = Result<(T, Duration), LookupFailure>;
+pub type RpcRequestResult<T> = Result<(T, Duration), RpcResponseError>;
```
jimmygchen: `RpcResponseResult`

jimmygchen:
@mergify queue

PR #5810, opened 2024-05-17
Fix failing attestation tests and misc electra attestation cleanup

eserilev:
NOTE: this PR includes changes from currently open attestation related PR's. In terms of the "conga line" this should probably be last 

### committee_index

We added the function `committee_index` to both `Attestation` and `AttestationRef` to fetch the committee index across both attestation variants (base & electra). Previously the function was returning a u64. Its been updates to return a `Option<u64>` since the electra variant can technically return `None` if no committee_bits are set. If `None` is returned for an attestations committee_index received over gossip, this violates the P2P rules and will result in the peer being down scored. The relevant logic for that has been added by returning a `NotExactlyOneCommitteeBitSet` error in `verify_early_checks`.

### get_indexed_attestation_from_signed_aggregate

When fetching an indexed attestation via a `signed_aggregate` for the base variant we make the following checks
- selection_proof.is_aggregator returns true
- committee contains the signed aggregates aggregator index

These checks didn't exist for the electra variant, which caused inconsistencies in our testing infra. This PR adds a new function `get_indexed_attestation_from_signed_aggregate` for electra variant signed aggregates and adds the checks i've described above. Theres some additional cleanup that needs to happen here, i've added additional info about that in some newly included TODOs

### Aggregation pool

Several unwraps were removed in favor of `Option`

### IndexedAggregatedAttestation

Now expectes a `observed_attestation_key_root` instead of `attestation_data_root`. The new key is generated by hashing the following struct 

``` rust
struct ObservedAttestationKey {
    committee_index: u64,
    attestation_data: AttestationData
}
```

This fixes an issue where the `observed_attestation` map was keyed off of `attestation_data_root` and therefore not effectively tracking unique attestations for electra. Weirdly enough, even without this change the local devnets durings interop seemed to be working fine, but I still believe this change is probably needed.

### Rename `AttestationRef`

In order to reduce confusion between the superstruct `AttestationRef` and the op pool `AttestationRef`, I've renamed the op pool struct to `CompactAttestationRef`

### Test fixes

- Fix an issue where beacon chain tests werent compiling
- All beacon chain tests except one are passing. The one that fails doesnt fail on a different branch of mine that also includes the ef test changes


realbigsean:
We should probably make these changes directly to `electra_attestation_changes` branch so we can keep all the attesation changes together. (we can pull the changes through the conga line after merging this to that branch)

eserilev:
i'll work on rebasing to `electra_attestation_changes`

beacon_node/beacon_chain/tests/payload_invalidation.rs
```diff
@@ -1222,13 +1221,13 @@ async fn attesting_to_optimistic_head() {
     let get_aggregated = || {
         rig.harness
             .chain
-            .get_aggregated_attestation(attestation.data())
+            .get_aggregated_attestation(&attestation.to_ref())
     };
```
ethDreamer, line 1227: Things that accept `&ObjectRef` should be changed to accept `ObjectRef` since `ObjectRef` is already `Copy`. We should make a lint for this at some point..

beacon_node/operation_pool/src/attestation.rs
```diff
@@ -119,14 +119,14 @@ impl<'a, E: EthSpec> AttMaxCover<'a, E> {
 
 impl<'a, E: EthSpec> MaxCover for AttMaxCover<'a, E> {
     type Object = Attestation<E>;
-    type Intermediate = AttestationRef<'a, E>;
+    type Intermediate = CompactAttestationRef<'a, E>;
     type Set = HashMap<u64, u64>;
 
-    fn intermediate(&self) -> &AttestationRef<'a, E> {
+    fn intermediate(&self) -> &CompactAttestationRef<'a, E> {
         &self.att
```
ethDreamer, line 126: This should return `CompactAttestationRef` rather than `&CompactAttestationRef` (assuming the `CompactAttestationRef` type is made by `superstruct`

ethDreamer: nevermind.. looks like it's not

ethDreamer (approved):
Nice work! Just small things and we should probably wait for `electra_attestation_changes` to be cleaned up (some of the PRs during interop were merged to the wrong branches) before merging this.

beacon_node/beacon_chain/tests/payload_invalidation.rs
eserilev: fixed in [7328924](https://github.com/sigp/lighthouse/pull/5810/commits/7328924c5d8eaba5fd72313bb0c2b9447e7893e2)

PR #5812, opened 2024-05-19
Remove unused dependencies

ackintosh:
## Proposed Changes

Removed the unused dependencies identified by [cargo-machete](https://github.com/bnjbvr/cargo-machete).

## Additional Info

We are running cargo-udeps in [CI](https://github.com/sigp/lighthouse/blob/stable/.github/workflows/test-suite.yml#L383-L404), but it seems imperfect as some unused dependencies were found by cargo-machete. Should we add cargo-machete to the CI job?


dapplion (approved):

michaelsproul (approved):

michaelsproul:
I don't mind whether we add machete to CI. We already have quite good coverage for unused depdendencies via `cargo-udeps`, which I think is why machete didn't find any totally unused crates that could be removed from `Cargo.lock`.

michaelsproul:
@mergifyio queue

PR #5813, opened 2024-05-19
Use JSON header by default for `/eth/v1/beacon/deposit_snapshot`

chong-he:
Most of current Lighthouse beacon API uses JSON by default when the header is not specified in querying the beacon API. The `/eth/v1/beacon/deposit_snapshot` uses SSZ when the header is not specified. This PR changes it to use JSON to be consistent with others.

Thanks @michaelsproul for the guidance.


michaelsproul (approved):
Looks great, thanks CK!

michaelsproul:
Marked this as backwards-incompat so we remember to mention it in the release notes. Some software that is sending the `*/*` accept header may be surprised to receive JSON rather than SSZ.

michaelsproul:
@mergifyio queue

PR #5815, opened 2024-05-20
Fix lookup disconnect peer

dapplion:
## Issue Addressed

Looking at holesky nodes for `Lookup maybe stuck` logs, found the following state:

```
May 19 09:00:04.308 DEBG Lookup maybe stuck                      summary: SingleBlockLookup { id: 6956, block_request_state: BlockRequestState { state: SingleLookupRequestState { state: AwaitingDownload, available_peers: 0, failed_processing: 0, failed_downloading: 0 } }, blob_request_state: BlobRequestState { state: SingleLookupRequestState { state: AwaitingDownload, available_peers: 1, failed_processing: 0, failed_downloading: 0 } }, block_root: 0x77921f4f47635e2a95efb4753625523aaba24a4636b7dae8544b45e894c04661, awaiting_parent: None, created: Instant { tv_sec: 20313759, tv_nsec: 73955749 } }, block_root: 0x77921f4f47635e2a95efb4753625523aaba24a4636b7dae8544b45e894c04661, id: 6956, service: lookup_sync, service: sync
```

This state is problematic because the block request has 0 peers and the blob request has 1 peer. The peer sets of all request should be identical, and the cause if this return early here

https://github.com/sigp/lighthouse/blob/b5de925d8f3868502cc1813ab66d565b7865aedb/beacon_node/network/src/sync/block_lookups/single_block_lookup.rs#L254-L257

The tests did not caught the bug, because the covered test case always returned RPCError for all active requests. In the case a lookup only sends a block request (not a blob request) and the peer disconnects, the lookup may get stuck.

## Proposed Changes

Duplicating the list of peers (= peers that claim to have imported the set of block components) between block and blob requests in not necessary. This PR hoists the peer set out of the request state into the lookup struct; indirectly fixing the return early bug.

Also add test to cover the case a lookup loses all peers but does not receive a RPCError.

### DIFF
--- a/beacon_node/network/src/sync/block_lookups/mod.rs
+++ b/beacon_node/network/src/sync/block_lookups/mod.rs
@@ -191,7 +191,7 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
                         .iter()
                         .find(|(_, l)| l.block_root() == block_to_drop)
                     {
-                        for &peer_id in lookup.all_used_peers() {
+                        for &peer_id in lookup.all_peers() {
                             cx.report_peer(
                                 peer_id,
                                 PeerAction::LowToleranceError,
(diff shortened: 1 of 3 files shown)


realbigsean (approved):
Nice this makes a lot of sense considering the other recent changes

realbigsean:
@mergify queue

PR #5817, opened 2024-05-20
Fix/remove some Electra attestation todos

realbigsean:
## Issue Addressed

- Removed logging about deneb attestations at the fork, current plan across clients is to just ignore them, and the logging here causes confusion
- run a few tests for both electra and base attestation types
- Remove a few todo's about verifying logic is correct. All of these reside in `consensus/types/src/beacon_block.rs` and look like they're correct. Would appreciate additional confirmation from reviewers here.

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -5120,50 +5120,6 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             },
         );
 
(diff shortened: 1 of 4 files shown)


eserilev:
not the best person to comment on the beacon_chain changes, but the changes to `attestation`, `attester_slashing` and the TODO removals in `beacon_block` look fine to me

ethDreamer:
image

ethDreamer (approved):

PR #5822, opened 2024-05-22
Add randomization in sync retry batch peer selection

dapplion:
## Issue Addressed

When retrying a batch, peers are prioritized by:
1. They have some failures
2. Count of active requets
3. Break tie with NodeID value

While the prioritization is okay, breaking ties by NodeID is not fair and prioritizes peers that just happen to have a peer ID if a smaller value or those that have farmed peer IDs with smaller values.

## Proposed Changes

Add a random u32 before sorting batch retry peers


beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -900,13 +900,14 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
...
                         self.active_requests.get(peer).map(|v| v.len()).unwrap_or(0),
+                        rand::thread_rng().gen::<u32>(),
                         *peer,
                     )
                 })
                 .collect::<Vec<_>>();
             // Sort peers prioritizing unrelated peers with less active requests.
             priorized_peers.sort_unstable();
-            priorized_peers.first().map(|&(_, _, peer)| peer)
+            priorized_peers.first().map(|&(_, _, _, peer)| peer)
```
jxs: not related with this PR, but do we need to iterate, collect, and then sort? To get the lowest peer by that `Ord` criteria we can just iterate once and get the lowest with :
```rust
let new_peer = self
            .network_globals
            .peers
            .read()
            .synced_peers()
            .map(|peer| {
                (
                    failed_peers.contains(peer),
                    self.active_requests.get(peer).map(|v| v.len()).unwrap_or(0),
                    rand::thread_rng().gen::<u32>(),
                    *peer,
                )
            })
            .min()
            .map(|(_, _, _, peer)| peer);
```

same applies to the other change on this PR

jxs (commented):
the rationale for the change makes sense to me! Left a comment wrt to the implementation only

beacon_node/network/src/sync/backfill_sync/mod.rs
dapplion: Nice suggestion! Pushed

beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -19,7 +19,7 @@ use beacon_chain::block_verification_types::RpcBlock;
 use beacon_chain::{BeaconChain, BeaconChainTypes};
 use lighthouse_network::types::{BackFillState, NetworkGlobals};
 use lighthouse_network::{PeerAction, PeerId};
-use rand::seq::SliceRandom;
+use rand::{seq::SliceRandom, Rng};
```
jxs:
```suggestion
use rand::seq::SliceRandom;
```

beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -890,24 +890,22 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
...
+        let new_peer = self
+            .network_globals
+            .peers
+            .read()
+            .synced_peers()
+            .map(|peer| {
+                (
+                    failed_peers.contains(peer),
+                    self.active_requests.get(peer).map(|v| v.len()).unwrap_or(0),
+                    rand::thread_rng().gen::<u32>(),
```
jxs: nit:
```suggestion
                    rand::random::<u32>(),
```
and we can avoid importing `Rng` above.

jxs (approved):
LGTM lion! Left one final suggestion which can just be applied

jxs:
@mergifyio queue

jxs:
@mergify refresh

AgeManning (approved):

jxs:
@mergifyio queue

PR #5824, opened 2024-05-22
Drop stuck lookups

dapplion:
## Issue Addressed

PR
- https://github.com/sigp/lighthouse/pull/5778

Makes us aware of the extent of the problem of sync lookups randomly getting stuck.

While useful I believe we should:
- Take it one step further and actively drop lookups that get stuck
- Increase the log level of "lookup stuck alarm"

While
- Fix underlying bugs that cause lookups to get stuck, both find by us and reported by users

The trade-off is quite good:
- A healthy lookup needing to exist for > 16 minutes is highly unlikely. Even if it's the case, range sync can rescue the node eventually
- So far, bugs that cause lookups to get stuck are sporadic, usually dependent of a specific sequence of events that may not repeat in a while.

Therefore dropping stuck lookups should be a net positive

## Proposed Changes

- If a lookup lasts more than LOOKUP_MAX_DURATION_SECS this function will find its oldest ancestor and then drop it and all its children.
- Increase the "lookup stuck" log from debug to warn, as now it will log only once per lookup id

### DIFF
--- a/beacon_node/network/src/metrics.rs
+++ b/beacon_node/network/src/metrics.rs
@@ -257,9 +257,9 @@ lazy_static! {
         "sync_lookups_completed_total",
         "Total count of sync lookups completed",
     );
-    pub static ref SYNC_LOOKUPS_STUCK: Result<IntGauge> = try_create_int_gauge(
-        "sync_lookups_stuck",
-        "Current count of sync lookups that may be stuck",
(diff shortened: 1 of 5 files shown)


AgeManning:
This seems reasonable to me.

AgeManning (approved):

jimmygchen (approved):
Looks good to me!

jimmygchen:
@mergify queue

PR #5828, opened 2024-05-23
Move PeerDAS type-level config to configurable `ChainSpec`

jimmygchen:
## Issue Addressed

Move PeerDAS type-level config to configurable `ChainSpec`.

- Use `ChainSpec` config values instead of `EthSpec` compile time type constants.
- Use `RuntimeVariableList` instead of `VariableList` for data column sidecar lists, as sizes are no longer known at compile time.
- Remove unused `child_components` file


dapplion (approved):
Looks great!

PR #5834, opened 2024-05-24
don't block mev boost till genesis finalization

realbigsean:
## Issue Addressed

We added a check not to use mev boost until the merge finalizes. This was done by checking the finalized block hash was not the zero hash. This check is no longer necessary on mainnet, and now just causes issues in testnets (no mev boost usage until two epochs past genesis) which is annoying. This just updates the check to ensure the head block hash isn't zero, so has a similar effect of disabling mev boost pre merge but doesn't wait an extra couple epochs.


michaelsproul (approved):
Nice. LGTM.

michaelsproul:
@mergifyio queue

Issue #5833, opened 2024-05-24
Stuck lookup reports


dapplion:
Collection of investigations of `Notify the devs, a sync lookup is stuck` events:

### ho-h-bn-toga - 0x9bf61d9f31473fc8a9a91891b9601f9c6a04450ebe55bbde0a083c60a0b36997

```
May 24 01:57:15.980 WARN Notify the devs, a sync lookup is stuck, ancestor_lookup: SingleBlockLookup { 
    id: 56, 
    block_request_state: BlockRequestState { 
        state: SingleLookupRequestState { 
            state: AwaitingDownload, failed_processing: 0, failed_downloading: 0 
            } 
        }, 
    blob_request_state: BlobRequestState { 
        state: SingleLookupRequestState { 
            state: AwaitingDownload, failed_processing: 0, failed_downloading: 0
        } 
    }, 
    peers: 7, 
```

Lookup `0x9bf61d9f31473fc8a9a91891b9601f9c6a04450ebe55bbde0a083c60a0b36997`

^ this block has 1 blob

- We had a child lookup with cached block (id = 55), and zero peers.
- When parent is processed, child continues
- The child block is sent for processing
- The blob request fails because it has no peers, and the lookup is dropped
- When the block processing result returns, it triggers and unknown lookup error
- Immediately then we get "unknown block hash message" for same root, and create a new lookup with (id = 56)

```
May 24 01:41:01.108 DEBG Unknown parent for gossip block         root: 0x9bf61d9f31473fc8a9a91891b9601f9c6a04450ebe55bbde0a083c60a0b36997
May 24 01:41:01.108 DEBG Searching for components of a block with unknown parent, id: 55, block_root: 0x9bf61d9f31473fc8a9a91891b9601f9c6a04450ebe55bbde0a083c60a0b36997, peer_ids: [], service: lookup_sync, service: sync
..
May 24 01:41:01.518 DEBG Received unknown block hash message     peer: PeerId("16Uiu2HAmKdxFEg3UKJRQ1BUGQaFjU2TBpE7qcz3YibWqaqX1T6RH"), block_root: 0x9bf61d9f31473fc8a9a91891b9601f9c6a04450ebe55bbde0a083c60a0b36997, service: sync
[...]

PR #5836, opened 2024-05-24
Report sync of block import via blob import

dapplion:
## Issue Addressed

Fix lookup stuck incident reported in 
- https://github.com/sigp/lighthouse/issues/5833#issuecomment-2129566873

Sync should be notified when a block abandons the da_checker to continue its child lookups. Block import can be completed both via block gossip import and blob import.

I am not 100% sure this fix is strictly necessary, but it's a good to have that can unstuck lookups if the logic to de-duplicate blob requests fails.

## Proposed Changes

Emit `GossipBlockProcessResult` also from blob import if the block gets imported


realbigsean (approved):
nice!

realbigsean:
@mergify queue

PR #5837, opened 2024-05-24
Do not double log uknown lookup

dapplion:
## Issue Addressed

Minor purely log fix. If sync attempts to process an event for an unknown lookup it will log "Dropping child lookup" which is not true. If the lookup is unknown it can't be dropped.

Noticed in https://github.com/sigp/lighthouse/issues/5833

## Proposed Changes

Do not double log uknown lookup

### DIFF
--- a/beacon_node/network/src/sync/block_lookups/mod.rs
+++ b/beacon_node/network/src/sync/block_lookups/mod.rs
@@ -625,7 +625,11 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
     /// dropped.
     pub fn drop_lookup_and_children(&mut self, dropped_id: SingleLookupId) {
         if let Some(dropped_lookup) = self.single_block_lookups.remove(&dropped_id) {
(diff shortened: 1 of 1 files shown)


beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -625,7 +625,11 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
     /// dropped.
     pub fn drop_lookup_and_children(&mut self, dropped_id: SingleLookupId) {
         if let Some(dropped_lookup) = self.single_block_lookups.remove(&dropped_id) {
-            debug!(self.log, "Dropping child lookup"; "id" => ?dropped_id, "block_root" => ?dropped_lookup.block_root());
+            debug!(self.log, "Dropping lookup";
```
dapplion, line 628: Instead of labeling this as "child", log if it's awaiting_parent

realbigsean (approved):

realbigsean:
@mergify queue

PR #5838, opened 2024-05-24
Add more info about why lookup is in AwaitingDownload

dapplion:
## Issue Addressed

When debugging lookup stuck incidents in https://github.com/sigp/lighthouse/issues/5833, it's very helpful to know if the specific reason why a lookup is at the `AwaitingDownload` state.

## Proposed Changes

As purely informational, append to `AwaitingDownload` enum variant the last reason why the lookup is at that state.


beacon_node/network/src/sync/block_lookups/single_block_lookup.rs
```diff
@@ -397,7 +411,7 @@ impl<T: Clone> SingleLookupRequestState<T> {
                     });
                 }
                 self.failed_downloading = self.failed_downloading.saturating_add(1);
-                self.state = State::AwaitingDownload;
+                self.state = State::AwaitingDownload("retry");
```
realbigsean: Might be good to differentiate "retry after download failure" from "retry after processing failure"

beacon_node/network/src/sync/block_lookups/single_block_lookup.rs
```diff
@@ -364,18 +369,27 @@ impl<T: Clone> SingleLookupRequestState<T> {
...
             false
         }
     }
 
+    /// Append metadata on why this request is in AwaitingDownload status. Very helpful to debug
+    /// stuck lookups. Not fallible as it's purely informational.
+    pub fn update_awaiting_download_status(&mut self, new_status: &'static str) {
+        match &mut self.state {
+            State::AwaitingDownload(status) => *status = new_status,
+            _ => {}
```
realbigsean: looks like a lint is failing here

realbigsean (changes requested):
I like the idea!

beacon_node/network/src/sync/block_lookups/single_block_lookup.rs
dapplion: Changed to "not started" since it will never remain in that state

realbigsean (approved):

realbigsean:
@mergify queue

michaelsproul:
needs an update

jimmygchen:
@mergify queue

PR #5840, opened 2024-05-24
Ensure lookup sync checks caches correctly

dapplion:
## Issue Addressed

Sync lookup needs to have a view of what's currently being processed to prevent double downloads and not get lookups stuck.

Part of an issue in https://github.com/sigp/lighthouse/issues/5833 was caused because blob requests don't check into the processing cache (`reqresp_pre_import_cache`). If the block is already processing, blob requests should continue.

To prevent futures issues like this I think we should encapsulate the details of the beacon_chain caches to there. And then expose only the minimum information sync needs to know with a simple enum `BlockProcessStatus`.

## Proposed Changes

- Add `get_block_process_status` to not leak cache details into sync lookup
- When attempting a `blob_lookup_request`, consider blocks in `NotValidated` state as download and continue the download of blobs

### DIFF
--- a/beacon_node/network/src/sync/network_context.rs
+++ b/beacon_node/network/src/sync/network_context.rs
@@ -397,14 +397,8 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
             // can assert that this is the correct value of `blob_kzg_commitments_count`.
             match self.chain.get_block_process_status(&block_root) {
                 BlockProcessStatus::Unknown => None,
-                BlockProcessStatus::NotValidated {
-                    blob_kzg_commitments_count,
-                    ..
-                }
-                | BlockProcessStatus::ExecutionValidated {
-                    blob_kzg_commitments_count,
-                    ..
-                } => Some(blob_kzg_commitments_count),
+                BlockProcessStatus::NotValidated(block)
+                | BlockProcessStatus::ExecutionValidated(block) => Some(block.num_expected_blobs()),
             }
         }) else {
             // Wait to download the block before downloading blobs. Then we can be sure that the
--- a/beacon_node/network/src/sync/network_context.rs
+++ b/beacon_node/network/src/sync/network_context.rs
@@ -12,7 +12,7 @@ use crate::status::ToStatusMessage;
 use crate::sync::block_lookups::SingleLookupId;
 use crate::sync::manager::{BlockProcessType, SingleLookupReqId};
 use beacon_chain::block_verification_types::RpcBlock;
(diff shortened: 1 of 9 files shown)


beacon_node/network/src/sync/network_context.rs
```diff
@@ -401,9 +392,20 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
         downloaded_block_expected_blobs: Option<usize>,
     ) -> Result<LookupRequestResult, RpcRequestSendError> {
         let Some(expected_blobs) = downloaded_block_expected_blobs.or_else(|| {
-            self.chain
-                .data_availability_checker
-                .num_expected_blobs(&block_root)
```
realbigsean, line 406: We can delete `num_blobs_expected`

realbigsean:
a couple lookup tests are broken, but i think the tests are wrong? the scenarios looks like they're testing that we don't trigger a block or blob lookup if we have a block in the processing cache, when really we just want to suppress the block lookup

realbigsean:
fixed the pr comments here: https://github.com/sigp/lighthouse/pull/5840/commits/2b550177bb73ccea9c36dd79bea1c8454f35918e

realbigsean:
Something I feel like I'm missing - is this addressing a bug that actually causes a lookup to get stuck? it looks like it's just an optimization where we are requesting blobs sooner

In the extreme case -> all attestations arrive while the block is in the processing cache, so none arrive to trigger a lookup when it hits the DA cache. When the block eventually hits the DA cache, we should get a missing components that causes this lookup to progress, right? or is there a separate bug preventing this?

dapplion:
The bug this PR is fixing is the blob requests not checking the block processing cache to figure out if the block is already downloaded.

Consider a case where:
- The block request did not request the block because it's in the processing cache, so it's in AwaitingDownload state
- The block is not in the da_checker

The blob request won't continue, because it's waiting for the block to download. However, if the block is not actually processing the request may be stuck.

I think the root issue is that during import the block is removed from the da_checker but not from the processing cache until later. 
- This PR eliminates this "status regression" by removing the block from the da_checker only when import_block completes https://github.com/sigp/lighthouse/pull/5845

However, I think blob_request should consider blocks that are both on the da_checker and processing_cache to be asymmetric with block_request logic. Otherwise, I think this can be a source of inconsistency if you sandwich other bugs.

realbigsean:
Thanks for the explanation :)

realbigsean:
@mergify queue

realbigsean (approved):

realbigsean:
@mergify refresh

PR #5846, opened 2024-05-25
fix: serde rename camel case for execution payload body

mattsse:
## Issue Addressed

I'm not exactly sure if this is the right branch to target, or if this is even an issue at all, but I noticed the get payload responses don't use camelCase

## Proposed Changes

rename camelCase 

## Additional Info


pawanjay176 (approved):
Thanks for this.

PR #5850, opened 2024-05-27
Rename deploy_block in network config

pawanjay176:
## Issue Addressed

https://github.com/eth-clients/holesky/pull/104

## Proposed Changes

Rename deploy_block.txt which was a Lighthouse specific config file to the agreed upon standard.

## Additional Info

This shouldn't cause any backward incompatibility issues with built in network configs.
This would cause backwards incompatibility issues with older versions of lighthouse trying to load networks with just `deploy_block.txt` in the testnet-dir which isn't a big deal imo.


jimmygchen (approved):
LGTM

jimmygchen:
@michaelsproul could use another review here before we merge just to be sure. I think this may break some testing tools, but that's probably it?

michaelsproul:
I reckon good to go

jimmygchen:
Looks like `lcli` need a rebuild on the runners:

> Unable to open testnet dir at "/home/runner/.lighthouse/local-testnet/testnet": Unable to open deposit_contract_block.txt: Os { code: 2, kind: NotFound, message: "No such file or directory" }

I'm thinking it might be worth to just build `lcli` all the time now it [takes less than 3 minutes](https://github.com/sigp/lighthouse/actions/runs/9264856460/job/25485751293) on the self hosted runners?

jimmygchen:
PR to build `lcli`: https://github.com/sigp/lighthouse/pull/5860
This PR can be queued after that one.

realbigsean:
@mergify queue

PR #5854, opened 2024-05-28
Fix BitVectors TestRandom implementation

eserilev:
## Issue Addressed

For electra attestations a new field `committee_bits: BitVector<E::MaxCommitteesPerSlot>` is introduced

Theres some assumptions in `BitVector`s `TestRandom` impl that `N` is divisible by 8. For the minimal spec `MaxCommitteesPerSlot` = 4. In this case, when generating a random bit vector, a `raw_bytes` vec of length 1 is created that is filled with a randomly generated value. In some cases the randomly generated value has bits set that are higher than `raw_bytes.len()`. This PR introduces some bit masking logic to the `TestRandom` impl for `BitVector` to ensure that we zero out bits > `raw_bytes.len()` in the case where `N` isn't divisible by 8.

This change should fix some failing electra networking tests. This PR along with other electra attestation PR's should get us passing CI for electra attestation changes. Thanks Michael for the assistance!


michaelsproul:
we could even target this at unstable?

dapplion (approved):

michaelsproul:
@mergifyio queue

PR #5856, opened 2024-05-28
Electra attestation changes rm decode impl

ethDreamer:
## Issue Addressed

Got rid of inefficient `Decode` implementation for `Attestation` & finished implementing schema migration.

## Additional Info

I got rid of a ton of legacy code for older versions of the persisted operation pool. This needs the closest look to validate that it was done right.

### DIFF
--- a/beacon_node/beacon_chain/src/schema_change/migration_schema_v20.rs
+++ b/beacon_node/beacon_chain/src/schema_change/migration_schema_v20.rs
@@ -11,7 +11,7 @@ pub fn upgrade_to_v20<T: BeaconChainTypes>(
     db: Arc<HotColdDB<T::EthSpec, T::HotStore, T::ColdStore>>,
     log: Logger,
 ) -> Result<Vec<KeyValueStoreOp>, Error> {
-    // Load a V14 op pool and transform it to V15.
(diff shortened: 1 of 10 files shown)


beacon_node/beacon_chain/src/schema_change/migration_schema_v20.rs
```diff
@@ -0,0 +1,103 @@
...
+use slog::{debug, info, Logger};
+use std::sync::Arc;
+use store::{Error, HotColdDB, KeyValueStoreOp, StoreItem};
+use types::Attestation;
+
+pub fn upgrade_to_v20<T: BeaconChainTypes>(
+    db: Arc<HotColdDB<T::EthSpec, T::HotStore, T::ColdStore>>,
+    log: Logger,
+) -> Result<Vec<KeyValueStoreOp>, Error> {
+    // Load a V14 op pool and transform it to V15.
```
michaelsproul:
```suggestion
    // Load a V15 op pool and transform it to V20.
```

michaelsproul (commented):
LGTM, nice!

michaelsproul (approved):

Issue #5708, opened 2024-05-03
Add peers to lookup ancestors


dapplion:
## Description

If a peer claims to have imported block A, and we have a parent chain of blocks with block A as tip, then the peer must know about all those ancestor blocks.

We should insert the peer into all those lookups, not just the lookup of block A.

## Version

`unstable`

1010adigupta:
would like to work on this

dapplion:
We are currently debugging the logic that this issue depends on. If you don't mind @1010adigupta I'll take this one

1010adigupta:
okay , I will try to learn from your pr

PR #5858, opened 2024-05-28
Add peers to parent lookups

dapplion:
## Issue Addressed

Previous PR
- https://github.com/sigp/lighthouse/pull/5655

Dropped the following feature: If a peers claims to have imported a block, we assume that it must have imported all its parents and will be considered to fetch data from any parent lookup

Having this feature is important to ensure that parent lookups don't "thin-out". Also, malicious peers can strategically send unknown data first to create lookups with a single peer.

## Proposed Changes

Add peers that claim to have imported a block to all lookups that are an ancestor of that block

Closes https://github.com/sigp/lighthouse/issues/5708

### DIFF
--- a/beacon_node/network/src/sync/block_lookups/mod.rs
+++ b/beacon_node/network/src/sync/block_lookups/mod.rs
@@ -244,17 +244,11 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
         }
 
         // Do not re-request a block that is already being requested
-        if let Some((_, lookup)) = self
+        if let Some((&lookup_id, lookup)) = self
             .single_block_lookups
             .iter_mut()
             .find(|(_id, lookup)| lookup.is_for_block(block_root))
         {
-            for peer in peers {
-                if lookup.add_peer(*peer) {
-                    debug!(self.log, "Adding peer to existing single block lookup"; "block_root" => ?block_root, "peer" => ?peer);
-                }
-            }
-
             if let Some(block_component) = block_component {
                 let component_type = block_component.get_type();
                 let imported = lookup.add_child_components(block_component);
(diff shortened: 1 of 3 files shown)


beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -85,6 +85,9 @@ pub struct BlockLookups<T: BeaconChainTypes> {
     log: Logger,
 }
 
+#[cfg(test)]
+pub(crate) type BlockLookupSummary = (Id, Hash256, Option<Hash256>, Vec<PeerId>);
```
jimmygchen, line 91:
```suggestion
// Tuple of `SingleLookupId`, requested block root, awaiting parent block root (if any), and list of peers that claim to have imported this set of block components.
pub(crate) type BlockLookupSummary = (Id, Hash256, Option<Hash256>, Vec<PeerId>);
```

beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -812,7 +824,55 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
...
+                    "block_root" => ?lookup.block_root(),
+                    "peer" => ?peer
+                );
+            }
+        }
+
+        let id = lookup.id;
+        let awaiting_parent = lookup.awaiting_parent();
+        // After adding new peers a lookup may be able to make progress.
+        let result = lookup.continue_requests(cx);
```
jimmygchen: It's not clear to me why we try to progress to the lookup here, before we progress the parent. Wouldn't it make more sense to do it in reverse? Also this action feels quite out of place in this function although I'm not sure what's a better place for this.

jimmygchen (commented):
Just a minor question & comment above. 
I did a round of quick testing locally to trigger parent lookups and it works fine 👍

beacon_node/network/src/sync/block_lookups/mod.rs
dapplion: After https://github.com/sigp/lighthouse/pull/5839 lookups may have zero peers. I agree it won't be a common case that the ancestor lookup in a parent chain ends up with zero peers, pauses and then needs continuation from the code of this PR. I would be okay with dropping this new `continue_requests` statement, it's really an optimization and not strictly required

realbigsean (approved):

realbigsean:
@mergify queue

PR #4946, opened 2023-11-21
Add Capella & Deneb light client support 

eserilev:
## Issue Addressed

#4896

## Proposed Changes

conform to capella/deneb light client specs

- [Capella light client spec](https://github.com/ethereum/consensus-specs/tree/dev/specs/capella/light-client)
- [Deneb light client spec](https://github.com/ethereum/consensus-specs/tree/dev/specs/deneb/light-client)

## Additional Info

https://github.com/sigp/lighthouse/pull/3954 is a blocker

PR #5859, opened 2024-05-29
Log warning on startup if chain spec contains misaligned fork epochs

jimmygchen:
## Issue Addressed

Addresses this comment: https://github.com/sigp/lighthouse/pull/4946#discussion_r1475547414

If fork epochs are not multiples of 256, sync committee period may span across forks - this is mostly fine for non-production and  testing code unrelated to light client. Adding a check would be useful for identifying issues with future forks / light client testing.

## Proposed Changes

Check fork epochs in the `ChainSpec` are aligned with the start of sync committee period, otherwise log a warning durinog startup.


dapplion (commented):
I am thinking who is the person that needs to see this warn. Anyone consuming lightclient data could be affected by an improperly aligned fork, which is a bigger group than the node operator. However, in practice no real network will adopt a bad fork epoch. This should apply to local devnets or experiments.

jimmygchen:
The intent was mainly for setting up expectation and giving a heads up on the configuration during testing (local devnet / testnets).

So yes, this only applies to local devnet and experiments, and we should never see this on a production network.

dapplion (approved):
LGTM!

jimmygchen:
@mergify queue

PR #5860, opened 2024-05-29
Always build `lcli` on CI

jimmygchen:
## Issue Addressed

The PR removes the "Install lcli" condition and always build lcli on CI, given it now only [takes less than 3 minutes](https://github.com/sigp/lighthouse/actions/runs/9264856460/job/25485751293) on the self-hosted runners. 

So far we've had to comment out / uncomment the condition every time there's a breaking change.

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -324,7 +324,6 @@ jobs:
       run: |
           make
     - name: Install lcli
-      if: env.SELF_HOSTED_RUNNERS == 'false'
       run: make install-lcli
     - name: Run the doppelganger protection failure test script
       run: |


macladson (approved):

antondlr:
`takes less than 3 minutes` -- on (I assume) low contention runs, best we keep an eye on this during busier periods!

right now, we have an `lcli` binary included in the runners image which requires manual updates - how often do these breaking changes occur? if we had a weekly rebuild-trigger with manual option to fire it off, would that work?

adding 2-3 minutes to CI runs hurts me a little I guess :-) but not against it per sé. 
just want to clarify that there's probably other means to get there.

pawanjay176:
> how often do these breaking changes occur? if we had a weekly rebuild-trigger with manual option to fire it off, would that work?

Breaking changes are not very often imo. We would have breaking changes when adding new forks or instances like this where we change the local testnet structure.
If we could retrigger at times when we noticed a breaking change, that also is good enough for me.

macladson:
It's unfortunately not enough to just update the binary more frequently, since in the cases where we introduce a breaking change to `lcli` in a PR, that PR will never pass CI unless CI compiles it, since we require that exact version of `lcli`. I don't believe keeping it up to date with `unstable` is enough.

We would need some kind of logic which checks if there any changes to `lcli` in the PR, and in those instances compiles it from source, otherwise it's free to use the cached version.

realbigsean:
> We would need some kind of logic which checks if there any changes to lcli in the PR, and in those instances compiles it from source, otherwise it's free to use the cached version.

Yes, I agree this is the way. I'm going to merge this one for now to unblock https://github.com/sigp/lighthouse/pull/5850

realbigsean:
@mergify queue

PR #4806, opened 2023-10-05
[Merged by Bors] - Reduce attestation subscription spam from VC

michaelsproul:
## Proposed Changes

Instead of sending every attestation subscription every slot to every BN:

- Send subscriptions 32, 16, 8, 7, 6, 5, 4, 3 slots before they occur.
- Track whether each subscription is sent successfully and retry it in subsequent slots if necessary.

## Additional Info

- [x] Add unit tests for `SubscriptionSlots`.
- [x] Test on Holesky.
- [x] Based on #4774 for testing.

PR #5863, opened 2024-05-29
Fix attestations not getting added to the aggregation pool

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

On receiving unaggregated attestations from the network, lighthouse sets the `should_process` variable on attestations depending on if we have a connected validator having aggregation duties for the same slot .
For `should_process == false`, we only validate and forward the attestation, otherwise, we add it to the naive_aggregation_pool so that we can aggregate them.

The `should_process` calculation happens in the subnet service where it checks if the duty slot and subnet exists in a delay map.
https://github.com/sigp/lighthouse/blob/3058b96f2560f1da04ada4f9d8ba8e5651794ff6/beacon_node/network/src/subnet_service/attestation_subnets.rs#L390-L393

The bug is that during insertion into the delay map, we use the default expiration delay of a single slot duration(12s). 

Since we implemented https://github.com/sigp/lighthouse/pull/4806 to reduce subscription spam, we don't send subscriptions every slot before the duty, so we get a situation where an entry into the delay map gets evicted before it has been read. 

Hence, the `should_process_attestation` function almost always returns false unless `import-all-attestations` flag is turned on, leading to Lighthouse producing bad quality aggregate attestations.

### DIFF
--- a/beacon_node/network/src/subnet_service/attestation_subnets.rs
+++ b/beacon_node/network/src/subnet_service/attestation_subnets.rs
@@ -477,8 +477,16 @@ impl<T: BeaconChainTypes> AttestationService<T> {
             time_to_subscription_slot.saturating_sub(advance_subscription_duration)
         };
 
+        let mut time_to_subscription_slot = self
+            .beacon_chain
+            .slot_clock
+            .duration_to_slot(slot)
+            .unwrap_or_default(); // If this is a past slot we will just get a 0 duration.
+
+        time_to_subscription_slot += slot_duration * 2;
+
         if let Some(tracked_vals) = self.aggregate_validators_on_subnet.as_mut() {
-            tracked_vals.insert(ExactSubnet { subnet_id, slot });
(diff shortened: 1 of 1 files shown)


michaelsproul (approved):
Fix looks good to me. I will add this to v5.2.0 for testing!

pawanjay176:
@michaelsproul do you think it would be worth it to run a few testnet nodes without `--import-all-attestations` to observe the `beacon_processor_unaggregated_attestation_verified_total ` metric?  
I checked a holesky node(with validators attached) running without `import-all-attestations` and the metric was flat at 0 even when gossip_attestation work events were hovering around 500.

image

michaelsproul:
Yeah I'm down to try running some nodes with `--import-all-attestations`.

AgeManning (approved):
Yep, nice find. Looks good to me. 

The duration is likely to be updated multiple times, but it should do what we expect here.

michaelsproul:
@mergifyio queue

PR #5864, opened 2024-05-30
Add `block_gossip` Beacon API events

chong-he:
Add `block_gossip` topic to the beacon API `eth1/v1/events`

Thank you @michaelsproul for the guidance and help


beacon_node/http_api/tests/tests.rs
```diff
@@ -5318,6 +5354,22 @@ impl ApiTester {
...
+
+        let block_gossip_events =
+            poll_events(&mut events_future, 1, Duration::from_millis(10000)).await;
+        assert_eq!(
+            block_gossip_events.as_slice(),
+            &[EventKind::BlockGossip(Box::new(BlockGossip {
+                slot: next_slot,
+                block: block_root,
+            }))]
+        );
```
michaelsproul: I think instead of polling again, you can put the `EventKind::BlockGossip` as the first entry of the list above. In the assert on line 5352 (the one that is currently failing). The test is failing at the moment because the `BlockGossip` event is being emitted for the first block (`self.next_block`), but the test is not expecting it.

If we take this approach, then we don't need `self.block_gossip`, because the gossip event is tested as part of testing `self.next_block`.

common/eth2/src/types.rs
```diff
@@ -954,6 +954,12 @@ pub struct SseHead {
     pub execution_optimistic: bool,
 }
 
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+
```
michaelsproul:
```suggestion
```

generally good to avoid blank lines between attributes (`#[...]`) and the thing they are applied to

beacon_node/http_api/tests/tests.rs
chong-he: Fix with your help, thanks Michael!

michaelsproul:
I'm going to drop this from 5.2.1 as it's not an essential part of the hotfix

michaelsproul (approved):
Tests look good, and I tested manually to ensure we were using quotes on the `slot` (we are). Good to go for 5.3

michaelsproul:
@mergify queue

PR #5865, opened 2024-05-30
Replace local testnet script with Kurtosis

jimmygchen:
## Issue Addressed

The local testnet script has become quite difficult to maintain over time and using Kurtosis for local testing is now the prefer approach. 

This PR replaces the scripts with a kurtosis config file and runs the same test on CI.

## TODO

- [x] Migrate doppelganger tests to kurtosis
- [x] Remove lcli commands that are no longer needed (e.g. `new-testnet`, `insecure-validators` etc)
- [ ] Update Lighthouse book and README

## Future considerations

- Add assertoor tests and utilise [assertoor github action](https://github.com/ethpandaops/kurtosis-assertoor-github-action) for CI.


dapplion (approved):

realbigsean:
legend

scripts/local_testnet/start_local_testnet.sh
```diff
@@ -1,147 +1,83 @@
...
+while getopts "e:b:n:phc" flag; do
   case "${flag}" in
-    v) VC_COUNT=${OPTARG};;
-    d) DEBUG_LEVEL=${OPTARG};;
-    p) BUILDER_PROPOSALS="-p";;
+    e) ENCLAVE_NAME=${OPTARG};;
+    b) BUILD_IMAGE=${OPTARG};;
+    n) NETWORK_PARAMS_FILE=${OPTARG};;
+    p) BUILDER_PROPOSALS=true;;
+    c) CI=true;;
```
jimmygchen, line 22: I didn't include `DEBUG_LEVEL` and `VC_COUNT`, as they are already in `network_parameters.yaml` and is easy to modify if needed rather than adding a bunch of flags.

Some reasoning for the default values:
- `BUILD_IMAGE=true`: when testing locally, we usually want to test the code in working directory, so this is on by default. If anyone wants to use latest image, can also set this to false and update the image in `network_parameters.yaml` file
- `CI=false`: when testing locally, a block explorer and metrics explorer is usually handy but not useful on CI.

scripts/tests/doppelganger_protection.sh
```diff
@@ -126,30 +152,20 @@ if [[ "$BEHAVIOR" == "success" ]]; then
     sleep $(( $SECONDS_PER_SLOT * 32 * 2 ))
     for val in 0x*; do
         [[ -e $val ]] || continue
-        curl -s localhost:9100/lighthouse/validator_inclusion/5/$val | jq | grep -q '"is_previous_epoch_target_attester": true'
-        IS_ATTESTER=$?
-        if [[ $IS_ATTESTER -eq 0 ]]; then
+        resp=$(curl -s $bn2_2_host_url/lighthouse/validator_inclusion/5/$val)
+        # TODO: error handling
```
jimmygchen: I realised this is silently failing, hence we don't see any errors on CI sometimes. I'll add some error handing code.

pawanjay176:
The issue with the doppelganger tests was that the doppelganger VC was starting before epoch 1 and we currently don't support that prior to epoch 1 . See https://github.com/sigp/lighthouse/blob/3058b96f2560f1da04ada4f9d8ba8e5651794ff6/validator_client/src/doppelganger_service.rs#L407-L416

Fixed that to start the VC exactly on epoch 1 and its working fine. Although we still need to capture the exit code correctly i think.

scripts/local_testnet/stop_local_testnet.sh
```diff
@@ -1,10 +1,15 @@
...
+# Delete existing logs directory and make sure parent directory exists.
+rm -rf $LOGS_SUBDIR && mkdir -p $LOGS_PATH
+kurtosis enclave dump $ENCLAVE_NAME $LOGS_SUBDIR
+echo "Local testnet logs stored to $LOGS_SUBDIR."
 
-PID_FILE=$TESTNET_DIR/PIDS.pid
-./kill_processes.sh $PID_FILE
-rm -f $PID_FILE
+kurtosis enclave rm -f $ENCLAVE_NAME
+echo "Local testnet stopped."
```
pawanjay176, line 15: Probably worth adding a 
`kurtosis clean` here

scripts/local_testnet/network_params.yaml
```diff
@@ -0,0 +1,16 @@
+# Full configuration reference [here](https://github.com/kurtosis-tech/ethereum-package?tab=readme-ov-file#configuration).
+participants:
+  - el_type: geth
+    el_image: ethereum/client-go:latest
+    cl_type: lighthouse
+    cl_image: lighthouse:local
+    cl_extra_params:
+      - --target-peers=3
+    vc_type: lighthouse
```
pawanjay176: do we need a separate vc and bn image here?

scripts/local_testnet/start_local_testnet.sh
```diff
@@ -1,147 +1,83 @@
...
+    echo "Not rebuilding Lighthouse Docker image."
+fi
 
-for (( bn=1; bn<=$BN_COUNT; bn++ )); do
-    secret=$DATADIR/geth_datadir$bn/geth/jwtsecret
-    echo $secret
-    execute_command_add_PID beacon_node_$bn.log ./beacon_node.sh $SAS -d $DEBUG_LEVEL $DATADIR/node_$bn $((BN_udp_tcp_base + $bn)) $((BN_udp_tcp_base + $bn + 100)) $((BN_http_port_base + $bn)) http://localhost:$((EL_base_auth_http + $bn)) $secret
-done
+# Stop local testnet
+kurtosis enclave rm -f $ENCLAVE_NAME 2>/dev/null || true
```
pawanjay176, line 79: `kurtosis clean` here too maybe?

scripts/local_testnet/start_local_testnet.sh
pawanjay176: Wondering if its easier from the end user perspective if `BUILD_IMAGE` is the actual docker image that the user wants to use, and if nothing is specified, then we build the local image and use it?

e.g.
```
# Uses the remote docker image
./start_local_testnet -b ethpandaops/lighthouse:unstable
# Builds local image and uses that
./start_local_testnet
```

pawanjay176 (approved):
Looks good to me overall. Just a few minor nits.
I'm happy with merging this as is and addressing them in the subsequent docs PR.

scripts/local_testnet/start_local_testnet.sh
jimmygchen: I think the boolean flag may still be useful in the case of local testing, because if a user already have a local image and want to re-use it, then they could run it without rebuilding, and _sometimes_ they would want to rebuild.

My thoughts around designing the params is to prefer the use the `network_parameters.yaml` file where possible - this essentially replaces the `vars.env` - and avoid having config in two places for simplicity and not having to worry about priority. 

Same reason why I removed `DEBUG_LEVEL` and `VC_COUNT` from the scripts. I'm not a big fan of maintaining large shell scripts and would love to keep functionalities as small as possible.

scripts/local_testnet/stop_local_testnet.sh
jimmygchen: I think `kurtosis enclave rm` already removes the enclave?

`kurtosis clean` removes all stopped enclaves, which might kill other enclaves unintentionally - if users are running multiple enclaves (similar to the `killall` command concern we had before)

scripts/local_testnet/network_params.yaml
jimmygchen: We don't, I'll remove this. Thanks!

scripts/local_testnet/stop_local_testnet.sh
jimmygchen: I was also running two instances of doppelganger (success & failure) tests at the same time, so would prefer to only destroy the specified enclave.

scripts/local_testnet/start_local_testnet.sh
jimmygchen: Please see comment above

jimmygchen:
@mergify queue

jimmygchen:
Thanks for the review @pawanjay176 , happy to discuss / address the comments above further.

scripts/local_testnet/stop_local_testnet.sh
pawanjay176: Ohh my bad, i thought `kurtosis enclave rm` just stops the enclave

PR #5867, opened 2024-05-30
Errors for all RPC Requests

AgeManning:
Sync now requires all RPC messages to either return a success or a failure. A disconnection doesn't count as a failure. 

Prior to this PR, there were two known cases where an RPC request can get "lost", i.e no error response. These are:
- A peer disconnects just after sync sends an RPC request message - These messages get dropped silently
- A peer disconnects whilst we are self rate limiting our own RPC requests

This PR should remedy these cases. In the first case, we instantly send an error response when the peer has disconnected. In the second case, we now watch for disconnection messages and then report an error for all messages that we are currently holding due to self rate limiting. 

In the process of doing this PR, I thought I needed an updated version of tokio-util in order to make the code cleaner. It turned out that wasn't the case, but I had already done the work to upgrade it. So I have left these changes in, which imo improve the code by removing a match case in most cases.


beacon_node/lighthouse_network/src/rpc/self_limiter.rs
```diff
@@ -158,13 +158,39 @@ impl<Id: ReqId, E: EthSpec> SelfRateLimiter<Id, E> {
...
+    /// Informs the limiter that a peer has disconnected. This removes any pending requests and
+    /// returns their IDs.
+    pub fn peer_disconnected(&mut self, peer_id: PeerId) -> Vec<(Id, Protocol)> {
+        // It's not ideal to iterate this map, but the key is (PeerId, Protocol) and this map
+        // should never really be large. So we iterate for simplicity
+        let mut failed_requests = Vec::new();
+        self.delayed_requests
+            .retain(|(map_peer_id, protocol), queue| {
+                if map_peer_id == &peer_id {
+                    // NOTE: Currently cannot remove entries from the DelayQueue, we will just let
```
jxs, line 174: I was looking and I think we don't need `ready_requests` as when `next_peer_request` is ready we can just return it and keep returning as long as it's ready `SelfRateLimiter` will keep being polled. I can submit a subsequent PR removing it if you agree.

jxs (approved):
LGTM age

beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -917,12 +917,24 @@ impl<AppReqId: ReqId, E: EthSpec> Network<AppReqId, E> {
...
-            request.into(),
-        )
+    pub fn send_request(
+        &mut self,
+        peer_id: PeerId,
+        request_id: AppReqId,
+        request: Request,
+    ) -> Result<(), (AppReqId, RPCError)> {
+        // Check if the peer is connected before sending an RPC request
+        if self.swarm.is_connected(&peer_id) {
```
jxs: nit, wdyt of:
```suggestion
        if !self.swarm.is_connected(&peer_id) {
            return Err((request_id, RPCError::Disconnected));
         }
```
I'd argue it's easier to read as we know one line below what happens if the peer is not connected and then we return to the main context of the function instead.

beacon_node/lighthouse_network/src/rpc/self_limiter.rs
AgeManning: I had an initial version of this that removed the delalyQueue and used a HashMapDelay. The problem I ran into was that typically there is a queue or backlog per (peer_id, protocol). When we insert them, the delay time is for the first in the queue. When I was using a HashMapDelay, they all expired really quickly which meant we had to keep requeuing them. 

The current version just expires the first element in the queue then we check for the next element and queue that, rather than all elements in the queue, if this makes sense.

dapplion (approved):
LGTM

ackintosh (approved):

michaelsproul:
@mergifyio queue

PR #5875, opened 2024-06-01
Re-export "useful" crates from `lighthouse`

jmcph4:
## Issue Addressed

N/A

## Proposed Changes

 - Add a library entry for the `lighthouse` crate
 - Re-export the following crates from this new library entry point so that they're accessible via `use lighthouse::foo::bar`
   - `beacon_node`
   - `environment`
   - `types` (as `eth2_types`)
   - `eth2_network_config`
   - `task_executor`
   - `boot_node`
   - `account_manager`
   - `validator_client`
   - `database_manager`
   - `slasher`
   - `validator_manager`
   - `lighthouse_version`
   - `lighthouse_network`

## Additional Info

Consider the problem of starting Lighthouse programmatically; i.e., an external program that wishes to run code in-process with Lighthouse. Currently, the imports for this look like:

```rust
use beacon_node::{ClientConfig, ProductionBeaconNode};
use environment::{
    Environment, EnvironmentBuilder, LoggerConfig, RuntimeContext,
};
use eth2::types::eth_spec::MainnetEthSpec;
use eth2_network_config::{Eth2NetworkConfig, DEFAULT_HARDCODED_NETWORK};
use task_executor::ShutdownReason;
```

(For a full demonstration, consult [this gist](https://gist.github.com/jmcph4/80b61b4c0bfb5da0376e2e85558670a9)).

With this PR, these imports would change to:

```rust
use lighthouse::{
    beacon_node::{ClientConfig, ProductionBeaconNode},
    environment::{
        Environment, EnvironmentBuilder, LoggerConfig, RuntimeContext,
    },
    eth2_network_config::{Eth2NetworkConfig, DEFAULT_HARDCODED_NETWORK},
    eth2_types::eth_spec::MainnetEthSpec,
    task_executor::ShutdownReason,
};
```

Additionally, specifying dependencies becomes vastly simpler. Compare:

```toml
lighthouse = { git = "https://github.com/sigp/lighthouse" }
beacon_node = { git = "https://github.com/sigp/lighthouse" }
task_executor = { git = "https://github.com/sigp/lighthouse" }
eth2 = { git = "https://github.com/sigp/lighthouse" }
environment = { git = "https://github.com/sigp/lighthouse" }
eth2_network_config = { git = "https://github.com/sigp/lighthouse" }
```

to:

```toml
lighthouse = { git = "https://github.com/sigp/lighthouse" }
```

### DIFF
--- a/lighthouse/src/lib.rs
+++ b/lighthouse/src/lib.rs
@@ -1,14 +1,13 @@
-pub use beacon_node;
-pub use environment;
-pub use types as eth2_types;
-pub use eth2_network_config;
-pub use task_executor;
-pub use boot_node;
 pub use account_manager;
-pub use validator_client;
+pub use beacon_node;
+pub use boot_node;
 pub use database_manager;
-pub use slasher;
-pub use validator_manager;
-pub use lighthouse_version;
+pub use environment;
+pub use eth2_network_config;
 pub use lighthouse_network;
-
+pub use lighthouse_version;
+pub use slasher;
+pub use task_executor;
+pub use types as eth2_types;
+pub use validator_client;
+pub use validator_manager;
--- a/lighthouse/src/lib.rs
+++ b/lighthouse/src/lib.rs
@@ -8,6 +8,6 @@ pub use lighthouse_network;
 pub use lighthouse_version;
 pub use slasher;
 pub use task_executor;
-pub use types as eth2_types;
+pub use types;
 pub use validator_client;
 pub use validator_manager;
--- a/lighthouse/src/lib.rs
+++ b/lighthouse/src/lib.rs
@@ -1,0 +1,16 @@
+mod metrics;
+
+pub use beacon_node;
+pub use environment;
+pub use types as eth2_types;
+pub use eth2_network_config;
+pub use task_executor;
+pub use boot_node;
+pub use account_manager;
+pub use validator_client;
+pub use database_manager;
+pub use slasher;
+pub use validator_manager;
+pub use lighthouse_version;
+pub use lighthouse_network;
+
--- a/lighthouse/src/lib.rs
+++ b/lighthouse/src/lib.rs
@@ -1,5 +1,3 @@
-mod metrics;
-
 pub use beacon_node;
 pub use environment;
 pub use types as eth2_types;
--- a/lighthouse/src/lib.rs
+++ b/lighthouse/src/lib.rs
@@ -1,12 +1,15 @@
 pub use account_manager;
 pub use beacon_node;
 pub use boot_node;
-pub use database_manager;
 pub use environment;
+pub use eth2 as http;
 pub use eth2_network_config;
+pub use execution_layer;
 pub use lighthouse_network;
 pub use lighthouse_version;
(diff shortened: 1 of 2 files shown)


lighthouse/src/lib.rs
```diff
@@ -0,0 +1,13 @@
...
+pub use beacon_node;
+pub use boot_node;
+pub use database_manager;
+pub use environment;
+pub use eth2_network_config;
+pub use lighthouse_network;
+pub use lighthouse_version;
+pub use slasher;
+pub use task_executor;
+pub use types as eth2_types;
```
michaelsproul: eth2 is deprecated, we could probably just call this `types` and still avoid confusion

lighthouse/src/lib.rs
```diff
@@ -0,0 +1,13 @@
...
+pub use validator_client;
+pub use validator_manager;
```
michaelsproul, line 16: IMO other useful additions would be:

- `eth2` crate (maybe renamed to `http`, it's our HTTP client for the beacon-API)
- `store` crate instead of `database_manager`. The DB manager is just the CLI subcommand, `store` is where all the good stuff happens.
- `slot_clock` for time management
- `execution_layer` for EL interop types & APIs

These are based on the things I commonly use externally, e.g. in `eleel`:

https://github.com/sigp/eleel/blob/cfc8ea2fe73d6a1835b6eb7625685635d01ff37e/Cargo.toml#L7-L12

And in `blockdreamer`:

https://github.com/blockprint-collective/blockdreamer/blob/030475d95cc46cab8c36b776bfa1738217b1ba25/Cargo.toml#L23-L27

michaelsproul (changes requested):

lighthouse/src/lib.rs
jmcph4: Addressed in [`fee55d3`](https://github.com/sigp/lighthouse/pull/5875/commits/fee55d35f60a1108d4e90a7dd5ea68383351a062).

lighthouse/src/lib.rs
jmcph4: Addressed in [`1a4e548`](https://github.com/sigp/lighthouse/pull/5875/commits/1a4e548090d355443eede81f55c8f8502ccfbbf6).

michaelsproul:
CI is failing because some of those crates aren't in the `Cargo.toml`

dapplion:
What API stability are we giving for these internal crates? I assume we may not want to respect semver, so might have to specify somewhere that they may break without notice?

michaelsproul:
Yeah that's a good idea. We should say that absolutely no stability guarantee is provided for these

People should pin a version, and update when there's a hard fork, or they feel like solving random dependency issues

jmcph4:
> Yeah that's a good idea. We should say that absolutely no stability guarantee is provided for these
> 
> People should pin a version, and update when there's a hard fork, or they feel like solving random dependency issues

I agree.

michaelsproul:
Maybe we could add a para to https://lighthouse-book.sigmaprime.io/contributing.html?

PR #5876, opened 2024-06-02
Allow all RPC messages on disconnect

AgeManning:
This PR makes it more permissive for outbound RPC messages to reach the application layer. 

Previously, RPC requests were not getting stream terminations in race conditions between a peer disconnecting and the RPC handler propagating the message up. 

This PR now allows messages to pass up to the application layer in the events that we start disconnecting from a peer. This will even allow slower messages to be sent to the application layer after peers have been disconnected. 

This should help resolve further sync lookup bugs.

### DIFF
--- a/beacon_node/lighthouse_network/src/service/mod.rs
+++ b/beacon_node/lighthouse_network/src/service/mod.rs
@@ -1378,26 +1378,17 @@ impl<AppReqId: ReqId, E: EthSpec> Network<AppReqId, E> {
     ) -> Option<NetworkEvent<AppReqId, E>> {
         let peer_id = event.peer_id;
 
-        if !self.peer_manager().is_connected(&peer_id) {
-            // Sync expects a RPCError::Disconnected to drop associated lookups with this peer.
-            // Silencing this event breaks the API contract with RPC where every request ends with
-            // - A stream termination event, or
-            // - An RPCError event
-            return if let HandlerEvent::Err(HandlerErr::Outbound {
-                id: RequestId::Application(id),
-                error,
-                ..
-            }) = event.event
-            {
-                Some(NetworkEvent::RPCFailed { peer_id, id, error })
-            } else {
-                debug!(
-                    self.log,
-                    "Ignoring rpc message of disconnecting peer";
-                    event
-                );
-                None
-            };
+        if !self.peer_manager().is_connected(&peer_id)
+            // Do not permit Inbound events from peers that are being disconnected
+            && matches!(event.event, HandlerEvent::Err(HandlerErr::Inbound { .. }))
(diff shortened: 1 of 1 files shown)


dapplion (approved):

beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -1378,26 +1378,17 @@ impl<AppReqId: ReqId, E: EthSpec> Network<AppReqId, E> {
...
-                debug!(
-                    self.log,
-                    "Ignoring rpc message of disconnecting peer";
-                    event
-                );
-                None
-            };
+        if !self.peer_manager().is_connected(&peer_id)
+            // Do not permit Inbound events from peers that are being disconnected
+            && matches!(event.event, HandlerEvent::Err(HandlerErr::Inbound { .. }))
```
dapplion: Why do you want to suppress Inbound errors? Is there any known scoring issue that can happen here? We should document why if that's the case

dapplion:
@AgeManning we can move the is_connected check to the main match. I am not convinced of having to ignore inbound errors. I've this commit moving the check to the match + only ignoring inbound requests https://github.com/sigp/lighthouse/commit/32c8fcec191f5f7b74319f2dcd2447cf016b1441

AgeManning:
Yeah fair point. I don't recall off the top of my head, I imagine there are few sneaky edge cases, like peers instantly dropping connections or something when we send a goodbye. 

Your commit does look cleaner, but I would test it before merging it in here. You can follow this guide: https://github.com/sigp/lighthouse/tree/stable/testing/network_testing

To run a fresh node on mainnet without syncing or anything and see if it has any weird errors. If not, lets do it.

beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -1378,26 +1378,17 @@ impl<AppReqId: ReqId, E: EthSpec> Network<AppReqId, E> {
...
+            && matches!(event.event, HandlerEvent::Ok(RPCReceived::Request(..)))
```
realbigsean: looks like this condition only hits if `event.event` is both an `Err` and `Ok` at the same time

realbigsean: I'm guessing the intent was `&&` should be `||`

michaelsproul: I just pushed this change

michaelsproul:
I pushed a commit here and then force pushed it out of existence, because I misread some logs on our infra thinking that the `Ignoring rpc message of disconnecting peer` was being logged for _outbound_ messages. This was only the case on 3 hosts that weren't actually running this branch 🙃 So all is well

michaelsproul:
@mergifyio queue

PR #5878, opened 2024-06-03
Add #! in `./el_bootnode.sh` script

Sushants-Git:
## Proposed Changes
Adding `#!/usr/bin/env bash` in the `./el_bootnode.sh` script


jimmygchen:
Hi,

We are replacing `local-testnet` scripts with `kurtosis` so this script will no longer exist in the next release.

Thanks for contributing regardless!

PR #5006, opened 2023-12-13
Investigate `UnknownTargetRoot` slasher errors

armaganyildirak:
## Issue Addressed
https://github.com/sigp/lighthouse/issues/4972

## Proposed Changes
This PR addresses the issue where attestations failing with `UnknownHeadBlock` are incorrectly processed through the slasher, leading to subsequent failures in signature verification (`SignatureNotChecked`) The subsequent failure occurs when attempting to load the state for the target block. This is expected when the target is identical to the head, leading to an unknown state. The proposed solution involves requeuing attestations without involving the slasher when encountering `UnknownHeadBlock`.

Some `UnknownTargetRoot` errors could still occur particularly during startup. It happens when the `beacon_block_root` slot is later than `target_root` slot. However they are harmless because the requeueing will usually take care of processing them later.

PR #5882, opened 2024-06-04
Increase slog buffer size

michaelsproul:
## Proposed Changes

Increase the `slog` buffer size to 16K to reduce the occurrence of:

> ERRO slog-async: logger dropped messages due to channel overflow, count: 4

These errors sometimes occur on our Holesky BNs which are receiving attestations from 100K validators each. They also occur occasionally on other BNs supporting large numbers of validators.

Examples of messages that overflow the channel include:

- `DEBG Attempted to publish duplicate message`
- `INFO Started monitoring validator`
- `DEBG Unable to obtain indexed form of attestation for slasher` (reduced by #5006)

Arguably some of these should be debounced, but for others it's not obvious that they should be. Increasing the buffer size temporarily at least gives us a full view of the logs until some of these spammier logs are toned down. It's frustrating to miss important ERRO/CRIT logs due to overflows.

### DIFF
--- a/lighthouse/environment/src/lib.rs
+++ b/lighthouse/environment/src/lib.rs
@@ -36,7 +36,7 @@ use {futures::channel::oneshot, std::cell::RefCell};
 
 pub use task_executor::test_utils::null_logger;
 
-const LOG_CHANNEL_SIZE: usize = 2048;
+const LOG_CHANNEL_SIZE: usize = 16384;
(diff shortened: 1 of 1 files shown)


jimmygchen (approved):
LGTM!

jimmygchen:
@mergify queue

dapplion:
Why 16k and not more? What's the avg size of a log record?

PR #5884, opened 2024-06-04
Fix getClientVersionV1 for Geth 1.13

michaelsproul:
## Issue Addressed

Fix an incompatibility with Geth 1.13 which does not follow the spec for `engine_getClientVersionV1`:

- https://github.com/ethereum/go-ethereum/pull/29351

## Proposed Changes

If Lighthouse is used with this buggy version of Geth we set the `name` to `""`. Lighthouse doesn't use the name for anything at present, so IMO there's no point adding complexity to read the `clientName` field.


jimmygchen (approved):
LGTM!

jimmygchen:
@mergify queue

Issue #5819, opened 2024-05-21
Update Criterion


eserilev:
## Description

Our current version of criterion depends on clap 2.x, the new criterion version (0.5) depends on clap 4.x. In #5273 were upgrading clap to 4.x. We should upgrade Criterion to 0.5 so we can completely remove the clap 2.x dependency from our codebase

michaelsproul:
I've started on this

michaelsproul:
PR here:

- https://github.com/sigp/lighthouse/pull/5886

eserilev:
Closed in #5886

PR #5886, opened 2024-06-04
Update criterion

michaelsproul:
## Issue Addressed

Closes:

- https://github.com/sigp/lighthouse/issues/5819

## Proposed Changes

Update criterion and remove our transitive dependency on `clap` v2 as a result :tada:


realbigsean (approved):
Awesome

realbigsean:
@mergify queue

PR #5890, opened 2024-06-05
Store data columns individually in store and caches

dapplion:
## Issue Addressed

Currently data columns are stored in the same bucket on the `BeaconDataColumn` DB column as a runtime list. In the case of a supernode, that's 128 items in that one key. However, a peer may only request a single-column index from us via data columns by range or by root. We then have to do x128 more i/o than necessary.

Note that blobs are currently stored in the same concatenated form. Since blob access patterns are more uniform (everyone is likely to request everything always) it makes sense.

## Proposed Changes

Change the `BeaconDataColumn` key from just `block_root` to `block_root | column_index`. Retrieve columns one by one on columns by range / by root requests.

Note that pruning is not yet implemented, so there a bit of legwork todo still to delete on columns on the same block root efficiently.

### DIFF
--- a/beacon_node/store/src/hot_cold_store.rs
+++ b/beacon_node/store/src/hot_cold_store.rs
@@ -20,8 +20,8 @@ use crate::metadata::{
 use crate::metrics;
 use crate::state_cache::{PutStateOutcome, StateCache};
 use crate::{
-    get_key_for_col, ChunkWriter, DBColumn, DatabaseBlock, Error, ItemStore, KeyValueStoreOp,
-    PartialBeaconState, StoreItem, StoreOp,
+    get_data_column_key, get_key_for_col, ChunkWriter, DBColumn, DatabaseBlock, Error, ItemStore,
+    KeyValueStoreOp, PartialBeaconState, StoreItem, StoreOp,
 };
 use itertools::process_results;
 use leveldb::iterator::LevelDBIterator;
@@ -36,6 +36,7 @@ use state_processing::{
     SlotProcessingError,
 };
 use std::cmp::min;
+use std::collections::HashMap;
 use std::marker::PhantomData;
 use std::num::NonZeroUsize;
 use std::path::Path;
@@ -89,7 +90,7 @@ pub struct HotColdDB<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> {
 struct BlockCache<E: EthSpec> {
     block_cache: LruCache<Hash256, SignedBeaconBlock<E>>,
     blob_cache: LruCache<Hash256, BlobSidecarList<E>>,
-    data_column_cache: LruCache<Hash256, DataColumnSidecarList<E>>,
+    data_column_cache: LruCache<Hash256, HashMap<ColumnIndex, Arc<DataColumnSidecar<E>>>>,
 }
 
 impl<E: EthSpec> BlockCache<E> {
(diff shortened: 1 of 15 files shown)


beacon_node/store/src/hot_cold_store.rs
```diff
@@ -1653,30 +1667,49 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
     pub fn get_data_columns(
         &self,
         block_root: &Hash256,
-    ) -> Result<Option<DataColumnSidecarList<E>>, Error> {
+    ) -> Result<Vec<Arc<DataColumnSidecar<E>>>, Error> {
+        let mut items = vec![];
+        for entry_result in self.blobs_db.iter_column_from::<Vec<u8>>(
+            DBColumn::BeaconDataColumn,
+            &get_data_column_key(block_root, &0),
+            &get_data_column_key(block_root, &u64::max_value()),
```
jimmygchen: why `u64::max_value` here rather than `self.spec.number_of_columns`?

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -1054,9 +1065,12 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
                     }
                     true
                 }
-                StoreOp::DeleteDataColumns(block_root) => {
+                StoreOp::DeleteDataColumns(block_root, _column_indices) => {
```
jimmygchen: why does using `column_indices` depend on pruning?

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -1653,30 +1667,49 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
 
-    /// Fetch data_columns for a given block from the store.
+    /// Fetch all data_columns for a given block from the store.
     pub fn get_data_columns(
         &self,
         block_root: &Hash256,
-    ) -> Result<Option<DataColumnSidecarList<E>>, Error> {
+    ) -> Result<Vec<Arc<DataColumnSidecar<E>>>, Error> {
+        let mut items = vec![];
+        for entry_result in self.blobs_db.iter_column_from::<Vec<u8>>(
```
jimmygchen: Should we load from cache first? similar to `get_data_column`?

jimmygchen (changes requested):
Looks good! 
I just have a couple of questions above.

beacon_node/store/src/hot_cold_store.rs
dapplion: Both should be equal. Using u64 max value is more expressive to requesting the entire column key space for the prefix `block_root`. This code is asking: give me all entries that start with `block_root` regardless of their value.

beacon_node/store/src/hot_cold_store.rs
dapplion: Removed in https://github.com/sigp/lighthouse/pull/5890/commits/7d72c364eb7142eef1ca20b3f5fe4d261bf67d2e

beacon_node/store/src/hot_cold_store.rs
dapplion: Good question. I am not able to justify this log. I went ahead an implemented pruning to see how it would look like. Please see https://github.com/sigp/lighthouse/pull/5890/commits/7d72c364eb7142eef1ca20b3f5fe4d261bf67d2e

beacon_node/store/src/hot_cold_store.rs
dapplion: Removed in https://github.com/sigp/lighthouse/pull/5890/commits/7d72c364eb7142eef1ca20b3f5fe4d261bf67d2e

jimmygchen (approved):
Nice, changes look good!  👍

PR #5891, opened 2024-06-05
Drop overflow cache

dapplion:
## Issue Addressed

Overflow cache is designed to never forget un-finalized not-yet-available blocks.

I believe this is not necessary, and we can drop the complexity to handle that case. Rationale:

> Usually a block becomes available on its slot within a second of receiving its first component
> over gossip. However, a block may never become available if a malicious proposer does not
> publish its data, or there are network issues that prevent us from receiving it. If the block
> does not become available after some time we can safely forget about it. Consider this two
> cases:
>
> - Global unavailability: If nobody has received the block components it's likely that the
>   proposer never made the block available. So we can safely forget about the block as it will
>   never become available.
> - Local unavailability: Some of our peers will attest to a descendant of the dropped block and
>   lookup sync will eventually fetch its components
>
> Even in periods of non-finality, the proposer is expected to publish the block's data
> immediately. Because this cache only holds fully valid data, its capacity is bounded to 1 block
> per slot and fork: before inserting into this cache we check proposer signature and correct
> proposer. Having a capacity > 1 is an optimization to prevent sync lookup from having re-fetch
> data during moments of unstable network conditions.

## Proposed Changes

Allow `DataAvailabilityChecker` internal LRUCache to drop the least used entry if it exceeds capacity.

This PR does **not** change the `DataAvailabilityChecker` state cache. I think its approach is good as is.

### DIFF
--- a/beacon_node/beacon_chain/src/data_availability_checker.rs
+++ b/beacon_node/beacon_chain/src/data_availability_checker.rs
@@ -33,7 +33,7 @@ pub const OVERFLOW_LRU_CAPACITY: NonZeroUsize = new_non_zero_usize(1024);
 pub const STATE_LRU_CAPACITY_NON_ZERO: NonZeroUsize = new_non_zero_usize(2);
 pub const STATE_LRU_CAPACITY: usize = STATE_LRU_CAPACITY_NON_ZERO.get();
 
-/// Cache to hold fully valid data that can't be imported to fork-choice yet. After dencun hard-fork
+/// Cache to hold fully valid data that can't be imported to fork-choice yet. After Dencun hard-fork
 /// blocks have a sidecar of data that is received separately from the network. We call the concept
 /// of a block "becoming available" when all of its import dependencies are inserted into this
 /// cache.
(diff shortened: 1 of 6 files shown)


beacon_node/store/src/lib.rs
```diff
@@ -268,8 +268,6 @@ pub enum DBColumn {
     OptimisticTransitionBlock,
     #[strum(serialize = "bhs")]
     BeaconHistoricalSummaries,
-    #[strum(serialize = "olc")]
-    OverflowLRUCache,
```
dapplion: What's the policy to deprecate DBColumns?

michaelsproul: We've never done it before

We could keep it around in `DBColumn` so it can be used with `lighthouse db inspect`, and then delete it once the release that removes it is buried under a hard fork

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -33,12 +33,30 @@ pub const OVERFLOW_LRU_CAPACITY: NonZeroUsize = new_non_zero_usize(1024);
...
+///   proposer never made the block available. So we can safely forget about the block as it will
+///   never become available.
+/// - Local unavailability: Some of our peers will attest to a descendant of the dropped block and
+///   lookup sync will eventually fetch its components
+///
+/// Even in periods of non-finality, the proposer is expected to publish the block's data
+/// immediately. Because this cache only holds fully valid data, its capacity is bounded to 1 block
+/// per slot and fork: before inserting into this cache we check proposer signature and correct
+/// proposer. Having a capacity > 1 is an optimization to prevent sync lookup from having re-fetch
+/// data during moments of unstable network conditions.
```
realbigsean, line 59: great docs!

beacon_node/store/src/lib.rs
```diff
@@ -268,8 +268,6 @@ pub enum DBColumn {
     OptimisticTransitionBlock,
     #[strum(serialize = "bhs")]
     BeaconHistoricalSummaries,
```
realbigsean:
```suggestion
    BeaconHistoricalSummaries,
    #[strum(serialize = "olc")]
    OverflowLRUCache,
```

beacon_node/store/src/lib.rs
realbigsean: added it back here https://github.com/sigp/lighthouse/pull/5891/commits/8366504e7e03fb66b30923a9582b0de388a1577d

realbigsean (approved):

jimmygchen (approved):
LGTM!

jimmygchen:
@mergify queue

PR #5893, opened 2024-06-05
Check known parent on rpc blob process

dapplion:
## Issue Addressed

We currently do not check that a blob from RPC is a descendant of a known block. Since we don't check the proposer signature we can insert data into the da_checker that is completely fabricated or descendant of invalid data.

I don't see any good reason on why to not check that parent is known before inserting into the da_checker. We may also want to check the proposer signature, and correct proposer.

Lookup sync today gets around this issues by:
- do not fetch blobs for some root until having downloaded that block. This forces an attacker into sending us _some_ block and get down-scored if it's not valid
- do not send blobs for processing until having processed the block. This patches the issue of not checking for unknown parent on blobs, but it slows down a bit lookup sync.

## Proposed Changes

Resolve this TODO

https://github.com/sigp/lighthouse/blob/4e675cf5da22990aecf75a4a008dd9f336d50bf4/beacon_node/network/src/sync/block_lookups/single_block_lookup.rs#L245-L246

Check known parent on RPC blob process

Depends on this PR to make the `ParentUnknown` error not expect an RpcBlock

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -3073,6 +3073,23 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             return Err(BlockError::BlockIsAlreadyKnown(block_root));
         }
 
+        // Reject RPC blobs referencing unknown parents. Otherwise we allow potentially invalid data
+        // into the da_checker, where invalid = descendant of invalid blocks.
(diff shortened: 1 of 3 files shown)


realbigsean:
Looks like the test is failing because there's no parent for the generated blob

realbigsean (approved):

realbigsean:
@mergify queue

Issue #5892, opened 2024-06-05
Potential leak of pending components in overflow cache


dapplion:
## Description

~60% of our fleet holds more than 10 items in the overflow cache

<img width="1517" alt="Screenshot 2024-06-05 at 12 46 20" src="https://github.com/sigp/lighthouse/assets/35266934/e0563187-ff48-4d01-9678-d79e0be90b30">

Some nodes reach the hundreds

<img width="1517" alt="Screenshot 2024-06-05 at 12 45 20" src="https://github.com/sigp/lighthouse/assets/35266934/99b92f9c-7a24-41d8-8953-77763b7b840c">

Pending components should be held at maximum until finality. Since mainnet is finalizing this is indicative of some leak or pruning issue

## Version

5.1.3

jimmygchen:
Closed by #5895

PR #5895, opened 2024-06-05
Iterate expired components from the da_checker

dapplion:
## Issue Addressed

Closes https://github.com/sigp/lighthouse/issues/5892

## Proposed Changes

On the pruning routine, iterate expired entries both when threshold is reached and when it's not

### DIFF
--- a/beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
+++ b/beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
@@ -359,6 +359,22 @@ impl<T: BeaconChainTypes> DataAvailabilityCheckerInner<T> {
     pub fn do_maintenance(&self, cutoff_epoch: Epoch) -> Result<(), AvailabilityCheckError> {
         // clean up any lingering states in the state cache
         self.state_cache.do_maintenance(cutoff_epoch);
+
+        // Collect keys of pending blocks from a previous epoch to cutoff
+        let mut write_lock = self.critical.write();
+        let mut keys_to_remove = vec![];
+        for (key, value) in write_lock.in_memory.iter() {
+            if let Some(epoch) = value.epoch() {
+                if epoch < cutoff_epoch {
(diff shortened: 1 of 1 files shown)


michaelsproul:
Thoughts on including this in 5.2 if we are about to re-test over the weekend?

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -776,6 +776,21 @@ impl<T: BeaconChainTypes> OverflowLRUCache<T> {
...
+        for (key, value) in write_lock.in_memory.iter() {
+            if let Some(epoch) = value.epoch() {
+                if epoch < cutoff_epoch {
+                    keys_to_remove.push(*key);
+                }
+            }
+        }
+        // Now remove keys
+        for key in keys_to_remove {
+            write_lock.in_memory.pop(&key);
```
michaelsproul: this just stops the cache from building up but won't clear old junk that's already on disk, will it?

michaelsproul:
On 2nd thoughts I don't think this is worth including in 5.2. Let's get 5.2 done and then we can consider a follow-up release

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
dapplion: There's a second maintenance routine that prunes keys in disk that are not in the memory map

michaelsproul:
Do we still need this now that the overflow cache is gone?

- https://github.com/sigp/lighthouse/pull/5891

dapplion:
> Do we still need this now that the overflow cache is gone?

Yes, the in-memory map has a leak that is independent to the disk portion that just got dropped.

realbigsean (approved):
LGTM

ethDreamer (approved):
looks good :)

realbigsean:
@mergify queue

Issue #5894, opened 2024-06-05
Subscribe to column subnets from PeerDAS (Fulu) epoch


dapplion:
The logic to subscribe / unsubscribe topics at a fork is the following:

`network/src/service.rs` holds three timer futures:
- `next_fork_subscriptions`:  Calls `subscribe_new_fork_topics` and subscribes to any existing topics but with the new fork-digest. Also, subscribes to new topics returned from `fork_core_topics(new_fork)`
- `next_unsubscribe`: Calls `unsubscribe_from_fork_topics_except` to unsubscribe all topics expect those with the current fork-digest
- `next_fork_update`: Sets new futures for `next_fork_subscriptions`, `next_unsubscribe` and `next_fork_update`

Outside of a fork-boundary the network service may receive a `SubscribeCoreTopics` message and subscribe to all topics returned by `core_topics_to_subscribe`

**With peerdas**

Now we have a new topic transition point where:
- we need to unsubscribe to a topic in all fork-digests
- do not rotate topics on a new fork-digest

Note that data_columns topics are not core topics, so we may not have to subscribe ahead of time. If we can accept a few slots of instability while we find peers we can avoid the complexity.

jimmygchen:
We've moved PeerDAS activation to the Fulu fork, however this is still relevant, see this comment:
https://github.com/sigp/lighthouse/pull/6795/files#r1912688432

jimmygchen:
Completed in #6849

PR #5899, opened 2024-06-06
Soft fork gossip topic activation and deactivation

dapplion:
## Issue Addressed

PeerDAS requires new features not available in current gossip topic machinery:

- Unsubscribe from a topic kind at a fork boundary (so far we only added)
- Change core topics at an epoch that does not align with a ForkName fork
- Change core topics without changing fork-digest

Most of the changes from this PR can be back-ported to unstable but I want to make sure they align with what we need for PeerDAS

## Proposed Changes

- Move logic to compute what to subscribe to the function `core_topics_to_subscribe(epoch, opts)`. It consolidates logic that is fork dependant with other user options like `subscribe-all-subnets`. The previous approach was very rigid expecting a set of topics for every ForkName. Now the set of topics change depending on peer das epoch
- Define a list of **"fork transitions"** which include your beloved altair, capella, etc and now a non-fork transition "peerdas".
- Add two functions `new_topics_to_subscribe_at_epoch` and `old_topics_to_unsubscribe_at_epoch` which diff the output of `core_topics_to_subscribe` across a "fork transition". Then, at the fork transition epoch subscribe the new topics, and forget the old ones

---

Closes https://github.com/sigp/lighthouse/issues/5894

### DIFF
--- a/beacon_node/lighthouse_network/src/types/topics.rs
+++ b/beacon_node/lighthouse_network/src/types/topics.rs
@@ -58,7 +58,7 @@ pub fn core_topics_to_subscribe<E: EthSpec>(
         }
     }
 
-    if fork_name >= ForkName::Altair {
+    if fork_name.altair_enabled() {
         topics.push(GossipKind::SignedContributionAndProof);
 
         if opts.subscribe_all_subnets {
@@ -73,11 +73,11 @@ pub fn core_topics_to_subscribe<E: EthSpec>(
         }
     }
 
-    if fork_name >= ForkName::Capella {
+    if fork_name.capella_enabled() {
         topics.push(GossipKind::BlsToExecutionChange);
     }
 
-    if fork_name >= ForkName::Deneb && !spec.is_peer_das_enabled_for_epoch(epoch) {
+    if fork_name.deneb_enabled() && !spec.is_peer_das_enabled_for_epoch(epoch) {
         // All of deneb blob topics are core topics
         for i in 0..spec.blob_sidecar_subnet_count {
             topics.push(GossipKind::BlobSidecar(i));
@@ -104,14 +104,20 @@ pub fn new_topics_to_subscribe_at_epoch<E: EthSpec>(
     next_epoch: Epoch,
     opts: &TopicSubscriptionOpts,
 ) -> Vec<GossipKind> {
-    let prev_epoch = next_epoch.saturating_sub(Epoch::new(1));
-    let prev_set = core_topics_to_subscribe::<E>(spec, prev_epoch, opts)
-        .into_iter()
-        .collect::<HashSet<_>>();
-    let next_set = core_topics_to_subscribe::<E>(spec, next_epoch, opts)
-        .into_iter()
-        .collect::<HashSet<_>>();
-    next_set.difference(&prev_set).cloned().collect()
+    if next_epoch == Epoch::new(0) {
+        // Handle the genesis case explicitly. Otherwise prev_set == next_set and the node will not
+        // subscribe to anything
+        core_topics_to_subscribe::<E>(spec, next_epoch, opts)
+    } else {
+        let prev_epoch = next_epoch.saturating_sub(Epoch::new(1));
+        let prev_set = core_topics_to_subscribe::<E>(spec, prev_epoch, opts)
+            .into_iter()
+            .collect::<HashSet<_>>();
+        let next_set = core_topics_to_subscribe::<E>(spec, next_epoch, opts)
+            .into_iter()
+            .collect::<HashSet<_>>();
+        next_set.difference(&prev_set).cloned().collect()
+    }
 }
 
 pub fn old_topics_to_unsubscribe_at_epoch<E: EthSpec>(
@@ -119,6 +125,8 @@ pub fn old_topics_to_unsubscribe_at_epoch<E: EthSpec>(
     next_epoch: Epoch,
     opts: &TopicSubscriptionOpts,
 ) -> Vec<GossipKind> {
+    // No need to handle genesis case explicitly, as we never need to unsubscribe anything at
+    // genesis. Because prev_set == next_set at genesis, this function will return `vec![]`
     let prev_epoch = next_epoch.saturating_sub(Epoch::new(1));
     let prev_set = core_topics_to_subscribe::<E>(spec, prev_epoch, opts)
         .into_iter()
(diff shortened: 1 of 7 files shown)


jimmygchen:
My understanding from latest discussions on ACDC & breakout call is that the plan is to activate PeerDAS at a hard fork (due to consensus changes and ease of coordination), thus `PEERDAS_ACTIVATION_EPOCH` would be the same as either `ELECTRA_FORK_EPOCH` or `FSTAR_FORK_EPOCH`. Therefore topic change could just occur as part of the hard fork transition?

dapplion:
What consensus changes ?

jimmygchen:
Rules to import into fork choice are different (custody columns instead of blobs) and incompatible; and additionally trailing fork choice changes. Are these not consensus changes?

New nodes won't be sharing the same view of the chain as old nodes, isn't that technically a hard fork?

jimmygchen:
> - Unsubscribe from a topic kind at a fork boundary (so far we only added)
> - Change core topics at an epoch that does not align with a ForkName fork
> - Change core topics without changing fork-digest

I think these features added could still be quite useful though, regardless of whether we use it for PeerDAS - and maybe we _could_ use it for PeerDAS - I'm just raising what I understood from various discussions.

dapplion:
fork choice changes are not a consensus change, because you can still interop and only matters in your local view. Fork-choice changes SHOULD be coordinated tho, otherwise weird things can happen, but that's why using an activation epoch is enough

beacon_node/lighthouse_network/src/types/topics.rs
```diff
@@ -24,70 +28,138 @@ pub const BLS_TO_EXECUTION_CHANGE_TOPIC: &str = "bls_to_execution_change";
...
+            topics.push(GossipKind::LightClientFinalityUpdate);
+            topics.push(GossipKind::LightClientOptimisticUpdate);
+        }
+    }
+
+    if fork_name >= ForkName::Capella {
+        topics.push(GossipKind::BlsToExecutionChange);
     }
+
+    if fork_name >= ForkName::Deneb && !spec.is_peer_das_enabled_for_epoch(epoch) {
```
realbigsean:
```suggestion
    if fork_name.deneb_enabled() && !spec.is_peer_das_enabled_for_epoch(epoch) {
```

beacon_node/lighthouse_network/src/types/topics.rs
```diff
@@ -24,70 +28,138 @@ pub const BLS_TO_EXECUTION_CHANGE_TOPIC: &str = "bls_to_execution_change";
...
         }
-        ForkName::Electra => vec![],
+
+        if opts.enable_light_client_server {
+            topics.push(GossipKind::LightClientFinalityUpdate);
+            topics.push(GossipKind::LightClientOptimisticUpdate);
+        }
+    }
+
+    if fork_name >= ForkName::Capella {
```
realbigsean:
```suggestion
    if fork_name.capella_enabled() {
```

beacon_node/lighthouse_network/src/types/topics.rs
```diff
@@ -24,70 +28,138 @@ pub const BLS_TO_EXECUTION_CHANGE_TOPIC: &str = "bls_to_execution_change";
...
+        GossipKind::AttesterSlashing,
+    ];
+
+    if opts.subscribe_all_subnets {
+        for index in 0..spec.attestation_subnet_count {
+            topics.push(GossipKind::Attestation(index.into()));
+        }
+    }
+
+    if fork_name >= ForkName::Altair {
```
realbigsean:
```suggestion
    if fork_name.altair_enabled() {
```

beacon_node/lighthouse_network/src/types/topics.rs
```diff
@@ -24,70 +28,138 @@ pub const BLS_TO_EXECUTION_CHANGE_TOPIC: &str = "bls_to_execution_change";
...
-            (0..E::SyncCommitteeSubnetCount::to_usize()).map(|sync_committee_id| {
-                GossipKind::SyncCommitteeMessage(SyncSubnetId::new(sync_committee_id as u64))
-            }),
-        )
+pub fn new_topics_to_subscribe_at_epoch<E: EthSpec>(
+    spec: &ChainSpec,
+    next_epoch: Epoch,
+    opts: &TopicSubscriptionOpts,
+) -> Vec<GossipKind> {
+    let prev_epoch = next_epoch.saturating_sub(Epoch::new(1));
```
realbigsean: I'm not sure if this method is called at genesis, but if it is, I don't think it works since you'll never have a different `prev_set` and `next_set`

beacon_node/lighthouse_network/src/types/topics.rs
```diff
@@ -24,70 +28,138 @@ pub const BLS_TO_EXECUTION_CHANGE_TOPIC: &str = "bls_to_execution_change";
...
+    opts: &TopicSubscriptionOpts,
 ) -> Vec<GossipKind> {
-    let mut topics = fork_core_topics::<E>(&current_fork, spec);
-    while let Some(previous_fork) = current_fork.previous_fork() {
-        let previous_fork_topics = fork_core_topics::<E>(&previous_fork, spec);
-        topics.extend(previous_fork_topics);
-        current_fork = previous_fork;
-    }
-    topics
+    let prev_epoch = next_epoch.saturating_sub(Epoch::new(1));
```
realbigsean, line 130: Same as above

consensus/types/src/fork_context.rs
```diff
@@ -124,4 +132,10 @@ impl ForkContext {
     pub fn all_fork_digests(&self) -> Vec<[u8; 4]> {
         self.digest_to_fork.keys().cloned().collect()
     }
+
+    /// Returns true if peerdas fork is schduled for some epoch != FAR_FUTURE_EPOCH
+    pub fn peerdas_fork_scheduled(&self) -> bool {
+        self.spec
+            .is_peer_das_enabled_for_epoch(Epoch::max_value().saturating_sub(Epoch::new(1)))
```
realbigsean:
```suggestion
        Some(self.spec.far_future_epoch) != self.spec.eip7594_fork_epoch
```

beacon_node/lighthouse_network/src/types/topics.rs
dapplion: Nice pick! Fixed in https://github.com/sigp/lighthouse/pull/5899/commits/6fa81e819d46542a6ac047eef365c39246fffca9

michaelsproul:

dapplion:
Obsolete, PeerDAS is part of Fulu

PR #5273, opened 2024-02-20
upgrade clap to v4.5

eserilev:
## Issue Addressed

#5087

## Proposed Changes

This PR upgrades clap from version 2.x to 4.5.1. There are some breaking changes between the two versions, v4.x is a bit stricter when it comes to flag definitions and has a slightly different api. Once we upgrade clap to v4 we can begin working on migrating to clap derive, which gives us the following benefits

- easier to read cli flag definitions
- a bit more type safety/compile time guarantees
- less verbose initialization logic
-  an easy interface to enable reading cli flags from a config file

We'll migrate to clap derive in separate PR's on a crate by crate basis, to make the PRs a bit more manageable

### List of breaking changes:

- `latency-measurement-service` flag is now deprecated in favor of `disable-latency-measurement-service`. By default the latency measurement service is enabled, and can be disabled by providing the new `disable-latency-measurement-service` flag.
- The `self-limiter` flag no longer accepts arguments. Providing the `self-limiter` flag enables the outbound rate limiter service. A new flag `self-limiter-protocols` is introduced to allow for configuring the outbound rate limiter service per protocol. In order to use the new `self-limiter-protocols` flag, `self-limiter` must also be provided. 
-  The `inbound-rate-limiter` flag is now deprecated in favor of `disable-inbound-rate-limiter`.  By default the inbound rate limiter service is enabled and can be disabled by providing the new `disable-inbound-rate-limiter` flag. Also, an additional flag `inbound-rate-limiter-protocols` is introduced to allow for configuring the inbound rate limiter service per protocol. Note that the `disable-inbound-rate-limiter` flag conflicts with the `inbound-rate-limiter-protocols` flag, they cannot be used in tandem.
- `slasher-broadcast` now requires either a true or false value. The flag cannot be used without providing a value.
- The above should cover the big breaking changes introduced by this PR. However its important to note that cli flag definitions are a bit stricter now. Anything listed under Options in the help screen must have an argument provided. Anything listed under Flags in the help screen cannot have an argument provided. If those conditions are violated the program will fail and an error message will be returned specifying what cli rules were violated.

PR #5902, opened 2024-06-07
Undo some CLI flag breakages

michaelsproul:
## Issue Addressed

The clap upgrade introduced a few hard breaking changes without deprecation warnings:

- https://github.com/sigp/lighthouse/pull/5273

## Proposed Changes

We've had users be caught out by these sorts of breakages in the past, so I've removed the hard breaking changes for two of the flags more likely to be in use: `--slasher-broadcast` and `--latency-measurement-service`:

- `--slasher-broadcast` without args is no longer deprecated. It works using [`default_missing_value`](https://docs.rs/clap/latest/clap/struct.Arg.html#method.default_missing_value), even if we ignore `clap`'s advice about using `.require_equals(true)`. We may need to change this or deprecate this in future. I couldn't work out at a glance how to tell whether the flag was provided with no value in order to deprecate _just_ that usage. So I figure we may as well keep it for now with no deprecation.
- `--latency-measurement-service` is readded but is deprecated (and hidden). Lighthouse logs a warning if it is provided, rather than failing to start.

### DIFF
--- a/beacon_node/src/cli.rs
+++ b/beacon_node/src/cli.rs
@@ -1101,6 +1101,8 @@ pub fn cli_app() -> Command {
                        [Enabled by default].")
                 .action(ArgAction::Set)
                 .default_value("true")
+                .num_args(0..=1)
+                .default_missing_value("true")
                 .display_order(0)
         )
         .arg(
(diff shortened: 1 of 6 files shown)


eserilev:
looks good to me!

michaelsproul:
@mergifyio queue

jxs (approved):

PR #5903, opened 2024-06-07
Fix SelfRateLimiter breaks the sequence of delayed_requests.

ackintosh:
## Issue Addressed

I noticed that `SelfRateLimiter::next_peer_request_ready()` breaks the sequence of the requests in its `delayed_requests`. 

For example, if there are RequestIds in `delayed_requests` in the sequence of `2, 3, 4, 5`. After running `next_peer_request_ready()`, the sequence will be `4, 5, 3` although we expect `3, 4, 5`. This happens because `push_back` is used to requeue.

## Proposed Changes

Use `push_front` instead of `push_back` to maintain the correct sequence of requests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4964,6 +4964,7 @@ dependencies = [
  "libp2p-mplex",
  "lighthouse_metrics",
  "lighthouse_version",
+ "logging",
  "lru",
  "lru_cache",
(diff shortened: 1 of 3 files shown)


AgeManning (approved):
Oh yeah. Nice catch!

AgeManning:
@mergifyio queue

PR #5084, opened 2024-01-18
Hold HeadTracker lock until persisting to disk

dapplion:
## Issue Addressed

Closes https://github.com/sigp/lighthouse/issues/4773
Implements the fix from https://github.com/sigp/lighthouse/pull/5073

## Proposed Changes

When persisting the head tracker in the context of the BeaconChain, hold its read lock until it has been written to disk. Otherwise there's a race condition with the pruning thread that can cause it to persist stale data.

## Additional Info

Linking another line of concern for this issue, pruning routine dropping the lock before writing the serialised head_tracker data to disk.

https://github.com/sigp/lighthouse/blob/f22e5b0f85ac0a667b1fab712abb8112d1dbc465/beacon_node/beacon_chain/src/migrate.rs#L695-L705

As long as pruning does not run twice concurrently the only other possible mutation is the BeaconNode importing a block. We have observed that happening during a unwind sequence, so it's definitely possible. In that case we can end-up with a block that exists in the DB but not in the HeadTracker. This can slightly pollute the DB but not break pruning.

Issue #4773, opened 2023-09-22
Lighthouse space usage - Increasing dramastically from 87GB to 430GB in 7 days


luarx:
## Description

We are running several Lighthouse clients and I have noticed that in 10% of them there are some ocassions were the volume usage is increased 5x in 7 days!

My temporal solution is about deleting the volume date and resync, after that the Lighthouse client is stable again. But some weeks later a different Lighthouse client repeats this weird behaviour...

![image](https://github.com/sigp/lighthouse/assets/22997139/6f3e2156-8498-44f1-a1ec-c67f563ce2b3)

## Version
Lighthouse v4.1.0 (docker image)

## Present Behaviour
The volume usage increases 5x in 7 days

Params:
```
          args: [
              "--debug-level=info",
              "--datadir=/beacondata",
              "--network=mainnet",
              "beacon_node",
              "--disable-enr-auto-update",
              "--enr-address=127.0.0.1",
              "--enr-tcp-port=9000",
              "--enr-udp-port=9000",
              "--port=9000" ,
              "--discovery-port=9000",
              "--eth1",
              "--http",
              "--http-address=0.0.0.0",
              "--http-port=5052",
              "--metrics",
              "--metrics-address=0.0.0.0",
              "--metrics-port=5054",
              "--listen-address=0.0.0.0",
              "--target-peers=100",
              "--http-allow-sync-stalled",
              "--disable-packet-filter",
              "--execution-endpoint=http://localhost:9545",
              "--jwt-secrets=/tmp/jwtsecret",
              "--disable-deposit-contract-sync",
              "--checkpoint-sync-url=https://beaconstate-mainnet.chainsafe.io"
              ]
```

## Expected Behaviour
Do not increment so heavy the volume usage in a few days, at least some stable values

## Steps to resolve
Do not increment so heavy the volume usage in a few days

luarx:
Another screenshot of a different Ligthouse client that is using the same version and same params.
Look how the space was stable again after deleting the volume data and resyncing

![image](https://github.com/sigp/lighthouse/assets/22997139/07a3d818-0271-4a09-a8bd-c4e5fe51e647)

michaelsproul:
If you still have any oversized nodes (or if it happens again) you can try:

Running `du -sh freezer_db chain_db` in the `$datadir/beacon`. I suspect it's the chain_db that's bloated.

[...]

PR #1132, opened 2020-05-11
Race condition fix + Reliability improvements around forks pruning

adaszko:
Fixes https://github.com/sigp/lighthouse/issues/1109

PR #5905, opened 2024-06-10
Store linear blocks in freezer db

dapplion:
## Issue Addressed

Part of tree-states Disk Edition ™️  

Currently Lighthouse stores blocks by root in the same bucket as unfinalized blocks. Storing finalized blocks in a separate bucket allows faster reads of sequential slot ranges. It also simplifies pruning logic and can allow us to drop the head-tracker which has been a source of bugs in the past:
- https://github.com/sigp/lighthouse/pull/1132
- https://github.com/sigp/lighthouse/issues/4773
- https://github.com/sigp/lighthouse/pull/5084

## Proposed Changes

- [x] Given a block ID read from either hot or cold DB based on a slot hint
- [x] On finalization move finalized blocks from hot DB to the freezer on a new column `BeaconBlockFrozen` and compress them
- [ ] Migrate existing blocks to freezer

### DIFF
--- a/beacon_node/beacon_chain/src/schema_change/migration_schema_v20.rs
+++ b/beacon_node/beacon_chain/src/schema_change/migration_schema_v20.rs
@@ -1,0 +1,108 @@
+use crate::BeaconChainTypes;
+use slog::{info, Logger};
+use std::sync::Arc;
+use store::{get_key_for_col, DBColumn, Error, HotColdDB, KeyValueStore, KeyValueStoreOp};
+use types::{Hash256, Slot};
+
+/// Chunk size for freezer block roots in the old database schema.
+const OLD_SCHEMA_CHUNK_SIZE: u64 = 128;
+
+fn old_schema_chunk_key(cindex: u64) -> [u8; 8] {
+    (cindex + 1).to_be_bytes()
+}
+
+pub fn upgrade_to_v20<T: BeaconChainTypes>(
+    db: Arc<HotColdDB<T::EthSpec, T::HotStore, T::ColdStore>>,
+    log: Logger,
+) -> Result<Vec<KeyValueStoreOp>, Error> {
+    info!(log, "Upgrading freezer database schema");
+    upgrade_freezer_database::<T>(&db, &log)?;
+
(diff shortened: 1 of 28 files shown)


beacon_node/store/src/hot_cold_store.rs
```diff
@@ -2591,6 +2707,14 @@ pub fn migrate_database<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>>(
     }) {
         let (block_root, state_root, slot) = maybe_tuple?;
 
+        // Delete block from hot DB
+        hot_db_ops.push(StoreOp::DeleteBlock(block_root));
+        // Write block to cold DB
+        let block = store
+            .get_hot_blinded_block(&block_root)?
+            .ok_or(HotColdDBError::MissingBlockToFreeze(block_root))?;
+        store.blinded_block_as_cold_kv_store_ops(&block_root, &block, &mut cold_db_ops)?;
```
dapplion, line 2716: Does the state_root_iter yield the same block_root more than once for skipped slots? If yes we are doing extra work here

beacon_node/beacon_chain/src/schema_change/migration_schema_v20.rs
```diff
@@ -0,0 +1,108 @@
...
+
+pub fn upgrade_to_v20<T: BeaconChainTypes>(
+    db: Arc<HotColdDB<T::EthSpec, T::HotStore, T::ColdStore>>,
+    log: Logger,
+) -> Result<Vec<KeyValueStoreOp>, Error> {
+    info!(log, "Upgrading freezer database schema");
+    upgrade_freezer_database::<T>(&db, &log)?;
+
+    // No hot DB changes
+    return Ok(vec![]);
```
dapplion, line 22: @michaelsproul thoughts on migrating all finalized blocks? If we just delete that's going to be a lot of bandwidth on the entire network to backfill. However, migrating all blocks will take a while. Are DB migrations blocking the start routine of Lighthouse?

dapplion:
Closing for stale, we can revisit later

PR #5906, opened 2024-06-10
Fix slasher tests

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Fixes ordering issue in `AttesterSlashingStatus::into_slashing` . Also adds electra attestations to the attester slashings test cases


michaelsproul (approved):
Looks good, nice work!

I wonder if we may still need to convert an `AttesterSlashingBase` to an `AttesterSlashingElectra` if we are slashing two base attestations after the fork, when only `AttesterSlashingElectra` slashings can be included on chain (and verified on gossip).

PR #5908, opened 2024-06-10
Fix `CommandLineTest` port conflicts on CI

jimmygchen:
## Issue Addressed

Fixes recurring port conflicts in beacon node command line tests. This PR changes the `CommandLineTest` to not start a `ProductionBeaconNode` (and HTTP server) by default - which isn't strictly necessary as we're just verifying command line configurations - so that it doesn't fail due to port conflicts. 

https://github.com/sigp/lighthouse/actions/runs/9446884349/job/26017488730


michaelsproul (approved):
Nice, this should save us some hassle with these tests in future too

realbigsean:
@mergify queue

PR #5915, opened 2024-06-11
Persist light client bootstrap

eserilev:
## Issue Addressed

Partially https://github.com/sigp/lighthouse/issues/3651

## Proposed Changes

This doc details how a `LightClientBootstrap` object can be stored in chunks so that it can then be served later on without any expensive computations/db queries
https://hackmd.io/@dapplion/BkxPbc6eQT

To summarize

- `LightClientHeader` can be constructed by fetching the full block. Blocks are already stored by root and are available forever. Fetching a single block by root isn't that expensive of a db query
- We'll start storing `SyncCommitteeBranch` in the db, keyed off the block root
- `SyncCommittee` will also be stored in the db, keyed off the sync committee period.
   -  If were fetching a `LightClientBootstrap` from the current finalized period, we can construct the `SyncCommittee` from the current state, since fetching the current beacon state is cheap
   - else we fetch the `SyncCommittee` from the db. 

## Additional Info

This PR builds off of #5545. 
#5545 should be merged before this PR


beacon_node/beacon_chain/src/light_client_server_cache.rs
```diff
@@ -195,6 +209,50 @@ impl<T: BeaconChainTypes> LightClientServerCache<T> {
...
+    fn store_sync_committee(
+        &self,
+        store: &BeaconStore<T>,
+        cached_parts: &LightClientCachedData<T::EthSpec>,
+        sync_committee_period: u64,
+        finalized_period: u64,
+    ) -> Result<(), BeaconChainError> {
+        let column = DBColumn::SyncCommittee;
+
+        // TODO should i always write to db, even if a record already exists?
```
dapplion: This implementation is not handling re-orgs. So there can exist a record that became orphaned. You can keep a hash of the latest persisted sync committee and only persist if the hash changes

beacon_node/beacon_chain/src/light_client_server_cache.rs
```diff
@@ -312,6 +370,96 @@ impl<T: BeaconChainTypes> LightClientServerCache<T> {
...
+            let sync_committee_branch: FixedVector<Hash256, CurrentSyncCommitteeProofLen> =
+                FixedVector::from_ssz_bytes(&bytes).unwrap();
+
+            return Ok(Some(sync_committee_branch));
+        }
+
+        Ok(None)
+    }
+
+    pub fn get_sync_committee(
```
dapplion: These methods should be moved to hot_cold_store

beacon_node/beacon_chain/src/migrate.rs
```diff
@@ -676,6 +676,7 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> BackgroundMigrator<E, Ho
                     StoreOp::DeleteBlock(block_root),
                     StoreOp::DeleteExecutionPayload(block_root),
                     StoreOp::DeleteBlobs(block_root),
```
dapplion, line 678: This prunes all non-canonical blocks. You should also prune finalized canonical non-checkpoint sync committee branches

beacon_node/beacon_chain/src/light_client_server_cache.rs
```diff
@@ -312,6 +370,96 @@ impl<T: BeaconChainTypes> LightClientServerCache<T> {
...
+
+    #[allow(clippy::type_complexity)]
+    pub fn get_light_client_bootstrap(
+        &self,
+        store: &BeaconStore<T>,
+        block_root: &Hash256,
+        finalized_period: u64,
+        chain_spec: &ChainSpec,
+    ) -> Result<Option<(LightClientBootstrap<T::EthSpec>, ForkName)>, BeaconChainError> {
+        if let Some(block) = store.get_full_block(block_root)? {
```
dapplion:
```suggestion
        let Some(block) = store.get_full_block(block_root)? else { return None };
```
to un-indent everything

beacon_node/beacon_chain/src/light_client_server_cache.rs
```diff
@@ -312,6 +370,96 @@ impl<T: BeaconChainTypes> LightClientServerCache<T> {
...
+                // TODO unwrap
+                Arc::new(
+                    self.get_sync_committee(store, sync_committee_period)?
+                        .unwrap(),
+                )
+            } else {
+                // were at the current finalized period, so this data should be easily retrievable from the currently cached state
+                // TODO unwrap
+                let state = store.get_state(&state_root, Some(slot))?.unwrap();
+                state.current_sync_committee()?.clone()
```
dapplion: Even tho the state may be in the store it's not guaranteed. If we hit this branch we are potentially doing a lot of work: state load, apply HDiffs, block replay. I would delete this path and ensure that we always have a sync_committee cached ready for any slot.

EDIT: get bootstrap is meant to only support finalized block roots. So no need to support a case where 
`sync_committee_period > finalized_period`

consensus/types/src/light_client_bootstrap.rs
```diff
@@ -111,6 +111,37 @@ impl<E: EthSpec> LightClientBootstrap<E> {
...
+                header: LightClientHeaderAltair::block_to_light_client_header(block)?,
+                current_sync_committee,
+                current_sync_committee_branch,
+            }),
+            ForkName::Capella => Self::Capella(LightClientBootstrapCapella {
+                header: LightClientHeaderCapella::block_to_light_client_header(block)?,
+                current_sync_committee,
+                current_sync_committee_branch,
+            }),
+            ForkName::Deneb | ForkName::Electra => Self::Deneb(LightClientBootstrapDeneb {
```
dapplion: TODO(electra): Use LightClientHeaderElectra

beacon_node/beacon_chain/src/light_client_server_cache.rs
```diff
@@ -158,12 +145,20 @@ impl<T: BeaconChainTypes> LightClientServerCache<T> {
...
-            self.get_light_client_update(&store, sync_period, chain_spec)?;
+        let sync_period = block_slot
+            .epoch(T::EthSpec::slots_per_epoch())
+            .sync_committee_period(chain_spec)?;
+
+        // Spec: Full nodes SHOULD provide the best derivable LightClientUpdate (according to is_better_update)
+        // for each sync committee period
+        let prev_light_client_update = match &self.latest_light_client_update.read().clone() {
+            Some(prev_light_client_update) => Some(prev_light_client_update.clone()),
+            None => self.get_light_client_update(&store, sync_period, chain_spec)?,
```
dapplion: I would add the logic to read from `latest_light_client_update` into `get_light_client_update` to ensure consistency. Plus if you read from store once you should write to the `latest_light_client_update` cache

beacon_node/beacon_chain/src/light_client_server_cache.rs
```diff
@@ -250,6 +260,9 @@ impl<T: BeaconChainTypes> LightClientServerCache<T> {
             )?;
         }
 
+        *self.latest_current_sync_committee.write() =
```
dapplion: The goal of this variable is to prevent repeated writes, so it should be named `latest_written_current_sync_committee` or `latest_stored_current_sync_committee`, and move the new code inside the `if finalized_period >= sync_committee_period - 1` condition. Otherwise, you may write into `latest_current_sync_committee` indicating it got persisted when not.

beacon_node/beacon_chain/src/light_client_server_cache.rs
```diff
@@ -419,46 +434,42 @@ impl<T: BeaconChainTypes> LightClientServerCache<T> {
...
+            return Ok(None);
+        };
 
-            return Ok(Some((light_client_bootstrap, fork_name)));
+        if sync_committee_period > finalized_period {
+            return Ok(None);
         }
 
-        Ok(None)
+        let Some(current_sync_committee) = self.get_sync_committee(store, sync_committee_period)?
```
dapplion: In this function we read three pieces of data: 
- `get_full_block`
- `get_sync_committee_branch`
- `get_sync_committee`

If we have the block but not the branch, what does that mean? Someone may request a block root that does not exist but I feel we should be stricter about our DB consistency. Can we assure that if a block exist, there must exist its `sync_committee_branch` and `sync_committee`?

eserilev: it should be the case that a `sync_committee_branch` and `sync_committee` exist for a finalized block root as long as that block root was processed by the node after initial sync.

so if start my node up, it finishes syncing, and immediately after syncing my head is at block 100, `sync_committee_branch` and `sync_committee` will begin persisting for blocks after block 100. if we have the block but not the branch + sync committee, its probably because that block was persisted during the nodes initial sync process OR it was persisted before the node enabled light client functionality. 

I think we'd need a backfill mechanism to ensure that a finalized block root always has an associated `sync_committee_branch` and `sync_committee`

eserilev: also if the provided block root isnt at a checkpoint, we wouldnt have the `sync_committee_branch` for it

dapplion: Can you add comments in the code reasoning about each case, with examples of when things may be missing? Also for better debuggability we should not return the same `None` for everything. The API route [`getLightClientBootstrap`](https://github.com/ethereum/beacon-APIs/blob/515216384db1f17f2b5f3ab8a0fc39340db8ee9d/apis/beacon/light_client/bootstrap.yaml#L55) supports a message, so we can differentiate 
- 404: block unknown
- 404: no sync committee branch
- 404: no sync committee

Same for the p2p route, we can attach messages on a resource unavailable response

dapplion:
Thanks for fixing the conflicts! nicer diff now, my unresolved review comments from ~June still apply to current code

beacon_node/beacon_chain/src/light_client_server_cache.rs
eserilev: this should now be resolved: 
8c1f980

beacon_node/beacon_chain/src/light_client_server_cache.rs
eserilev: added better comments, and returned helpful error messages here: 
e75b84f

beacon_node/beacon_chain/src/light_client_server_cache.rs
eserilev: [8ef297b](https://github.com/sigp/lighthouse/pull/5915/commits/8ef297b0752031fc8c345decc0d69fe060c88e6e)

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -351,6 +351,24 @@ async fn light_client_updates_test() {
...
+            BlockStrategy::OnCanonicalHead,
+            AttestationStrategy::AllValidators,
+        )
+        .await;
+    
+    let lc_bootstrap = beacon_chain
+        .get_light_client_bootstrap(&block_root)
+        .unwrap();
+
+    println!("{:?}", lc_bootstrap);
```
dapplion: leftover print?

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -2822,6 +2832,8 @@ pub fn migrate_database<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>>(
 
     let mut hot_db_ops = vec![];
     let mut cold_db_ops = vec![];
+    let mut epoch_boundary_blocks = vec![];
+    let mut non_epoch_boundary_blocks = vec![];
```
dapplion: Invoking @michaelsproul to review this pruning code

eserilev:
This PR is in a good spot for another review. The pruning logic has been simplified, the bootstrap tests are passing and I've done a bit of tweaking to the test infra to make light client testing a bit easier in general. I am planning on writing some specific tests related to the pruning logic. I'm also planning on running a mainnet node with light client enabled and compare I/O + disk usage vs a node w/ light client disabled. If its a negligible difference it'd be cool to just enable light clients by default

beacon_node/beacon_chain/src/light_client_server_cache.rs
eserilev: i've moved the relvant methods to hot_cold_store

beacon_node/beacon_chain/src/light_client_server_cache.rs
eserilev: im now only persisting if the current_sync_committee has changed

beacon_node/beacon_chain/src/migrate.rs
eserilev: I am now pruning canonical non-checkpoint sync committee branches

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -2906,95 +2916,14 @@ pub fn migrate_database<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>>(
...
-    } else {
-        non_epoch_boundary_blocks
-            .iter()
-            .for_each(|(_, block_root)| {
-                hot_db_ops.push(StoreOp::DeleteSyncCommitteeBranch(*block_root));
-            });
-    }
+    // Prune sync committee branch data for all non checkpoint block roots.
+    // Note that `non_checkpoint_block_roots` should only contain non checkpoint block roots
+    // as long as `finalized_state.slot()` is at an epoch boundary.
```
dapplion: What happens otherwise?

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -2861,10 +2861,20 @@ pub fn migrate_database<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>>(
...
+        // `non_checkpoint_block_roots` hash set.
+        if slot % E::slots_per_epoch() == 0 {
+            epoch_boundary_blocks.insert(block_root);
         } else {
-            epoch_boundary_blocks.push((slot, block_root))
+            non_checkpoint_block_roots.insert(block_root);
+        }
+
+        if epoch_boundary_blocks.contains(&block_root) {
+            non_checkpoint_block_roots.remove(&block_root);
```
dapplion, line 3018: Much cleaner <3 

The intended logic is:

_no skipped slot_
- Slot 2/1  root C  : Add C to non_checkpoint_block_roots
- Slot 2/0  root B  : Add B to epoch_boundary_blocks
- Slot 1/31 root A  : Add A to non_checkpoint_block_roots

_slot 31 skipped_
- Slot 2/1  root C  : Add C to non_checkpoint_block_roots
- Slot 2/0  root A  : Add A to epoch_boundary_blocks
- Slot 1/31 root A  : Add A to non_checkpoint_block_roots, remove A from non_checkpoint_block_roots

_slot 0 skipped_
- Slot 2/1  root B  : Add B to non_checkpoint_block_roots
- Slot 2/0  root B  : Add B to epoch_boundary_blocks, remove B from non_checkpoint_block_roots
- Slot 1/31 root A  : Add A to non_checkpoint_block_roots

beacon_node/store/src/hot_cold_store.rs
eserilev: take the following example

current split slot: Epoch A slot 0
finalized_state.slot(): Epoch C slot 31

as we go through the pruning logic we'll end up pruning Epoch C slot 31. If Epoch D slot 0 happens to be a skipped slot we will have deleted the sync committee branch for a checkpoint slot.

This should never happen because we make this check before pruning

``` rust
if finalized_state.slot() % E::slots_per_epoch() != 0 {
   return Err(HotColdDBError::FreezeSlotUnaligned(finalized_state.slot()).into());
}
```

So as long as this invariant doesn't change, we should be ok

dapplion: Makes sense, please document in the code comment

eserilev: a202eac

dapplion (approved):
Good to go! Thank you @eserilev for driving this 🙏 

_Approve PR now,_
_Sync and store with no delay,_
_Light shines on the chain._

beacon_node/beacon_chain/src/light_client_server_cache.rs
```diff
@@ -178,57 +198,57 @@ impl<T: BeaconChainTypes> LightClientServerCache<T> {
...
+        if let Some(latest_sync_committee) =
+            self.latest_written_current_sync_committee.read().clone()
+        {
+            if latest_sync_committee == cached_parts.current_sync_committee {
+                return Ok(());
+            }
+        };
 
-        *self.latest_light_client_update.write() = Some(light_client_update.clone());
+        if finalized_period >= sync_committee_period - 1 {
```
michaelsproul: doesn't this overflow for `sync_committee_period == 0`?

could switch it to:

```
finalized_period + 1 >= sync_committee_period
```

beacon_node/http_api/tests/tests.rs
```diff
@@ -1933,19 +1933,24 @@ impl ApiTester {
...
             .await
         {
-            Ok(result) => result.unwrap().data,
+            Ok(result) => result,
             Err(e) => panic!("query failed incorrectly: {e:?}"),
         };
 
-        let expected = block.slot();
-        assert_eq!(result.get_slot(), expected);
+        assert!(result.is_none());
```
michaelsproul: why do we expect this test to fail now? because the server isn't enabled in these tests?

eserilev: yeah exactly. it might be beneficial to enable the light client server in http tests to get full test coverage for the `light_client_bootstrap` and `light_client_updates` endpoint

beacon_node/beacon_chain/src/light_client_server_cache.rs
eserilev: [82a57b5](https://github.com/sigp/lighthouse/pull/5915/commits/82a57b591be27569b0380e8ef931e23b48ca204c)

beacon_node/http_api/tests/tests.rs
eserilev: it should be fairly trivial to enable the light client server for these tests. I'll just go ahead and do that

eserilev: e141c80

michaelsproul (approved):

michaelsproul:
@mergify queue

Issue #5904, opened 2024-06-09
Support SSZ StableContainer 


dapplion:
Implement
- [EIP-7495: SSZ StableContainer](https://eips.ethereum.org/EIPS/eip-7495)
- [EIP-7688: Forward compatible consensus data structures](https://eips.ethereum.org/EIPS/eip-7688)

Mostly update underlying lib https://github.com/sigp/tree_hash to support StableContainer merkleization and have some nice DX way to specify property ordering

michaelsproul:
In progress here:

- https://github.com/sigp/ethereum_ssz/pull/16
- https://github.com/sigp/tree_hash/pull/15

PR #5920, opened 2024-06-13
Implement EIP-7688: Forward compatible consensus data structures [WIP]

macladson:
## Issue Addressed

https://github.com/sigp/lighthouse/issues/5904

## Proposed Changes

Converts certain consensus types into forward compatible versions (Profiles) following [EIP-7688](https://eips.ethereum.org/EIPS/eip-7688), using the `StableContainer` and `Profile` implementations in the `tree_hash` and `ethereum_ssz` crates. 

## Additional Info
 
Currently based on `unstable`, but will likely be needed to be rebased onto an `electra` branch in order to join the testnet.

Relies on the following PRs:
https://github.com/sigp/ethereum_ssz/pull/16
https://github.com/sigp/tree_hash/pull/15
https://github.com/sigp/ssz_types/pull/23


consensus/types/src/execution_payload.rs
```diff
@@ -50,52 +54,71 @@ pub type WithdrawalRequests<E> =
 #[serde(bound = "E: EthSpec", untagged)]
 #[arbitrary(bound = "E: EthSpec")]
 #[ssz(enum_behaviour = "transparent")]
-#[tree_hash(enum_behaviour = "transparent")]
+#[tree_hash(enum_behaviour = "transparent_stable")]
```
dapplion, line 59: Could you specify a behaviour that takes the existing order by default so we don't have to specify `#[tree_hash(stable_index = i)]` in every single data structure? Only in those that have experience deprecation of fields

PR #5922, opened 2024-06-13
Remove `extern crate`

michaelsproul:
## Proposed Changes

As of Rust 2018 (which we use), `extern crate` is unnecessary, and kinda janky.

This PR removes all uses of `extern crate` in favour of regular `use`s.


realbigsean (approved):
LGTM

realbigsean:
@mergify queue

realbigsean:
@mergify requeue

Issue #5785, opened 2024-05-15
Delayed RPC Send Using Tokens


AgeManning:
## Description

Currently we use the leaky-bucket algorithm to generate tokens for RPC requests. We use this to enforce rate limiting. If a peer runs out of tokens we send a rate limit error on the RPC, with a time at which they can start requesting again. 

The current problem with this is that the rate limit errors are not spec'd and not really used. We practically just downscore peers that hit our rate limits and the errors are somewhat meaningless. 

The specification is shifting more toward the behaviour where if a peer hits our rate limit, we hold the stream open but slowly send responses as the rate limit tokens get regenerated. This practically slows the response down to match our limits. The requesting peer may decide to drop the request if the responses are too slow. 

## Implementation Details and Rough Guide

The rate limit logic is found here: https://github.com/sigp/lighthouse/blob/stable/beacon_node/lighthouse_network/src/rpc/rate_limiter.rs

I don't think we need to change any code in here, but if you are interested in seeing how the tokens get re-generated this is where the logic is.

When an RPC request comes in, we check that it is within our rate limits here:
https://github.com/sigp/lighthouse/blob/stable/beacon_node/lighthouse_network/src/rpc/mod.rs#L301

If it hits our limit we simply return an error:
https://github.com/sigp/lighthouse/blob/stable/beacon_node/lighthouse_network/src/rpc/mod.rs#L327

There are probably a few things to note here:
1. In the error case, if the request is too large (i.e outside of the spec), we probably still want to error. 
2. If the request is requesting something that would make us wait longer than the spec TIMEOUT time (currently 10 seconds), then we should just error and close the stream rather than wait the 10 seconds and reach the timeout. 
3. For batch requests, like blocks_by_range. It could be that the first request uses most of the tokens, so we process straight away, then a second request doesn't have enough tokens to fulfill the entire request, but we will be able to fulfill the request in the 10 second mark, then we probably want to send the entire request to the block processor to read from the db in a batch, then trickle the responses back to the peer as tokens become available. 

I think it might be easier to implement the cases where we process the request then trickle the responses. 

In order to trickle the responses, I think (off the top of my head) there are maybe two or three ways to do it. There may be more, but these are some suggestions to get started. 

1. Once we have processed the request and send the response to the handler, i.e via: https://github.com/sigp/lighthouse/blob/stable/beacon_node/lighthouse_network/src/rpc/mod.rs#L169
[...]

PR #5923, opened 2024-06-13
Delayed RPC Send Using Tokens

ackintosh:
## Issue Addressed

closes https://github.com/sigp/lighthouse/issues/5785

## Proposed Changes

The diagram below shows the differences in how the receiver (responder) behaves before and after this PR. The following sentences will detail the changes.

```mermaid
flowchart TD

subgraph "*** After ***"
    Start2([START]) --> AA[Receive request]
    AA --> COND1{Is there already an active request <br> with the same protocol?}
    COND1 --> |Yes| CC[Send error response]
    CC --> End2([END])
    %% COND1 --> |No| COND2{Request is too large?}
    %% COND2 --> |Yes| CC
    COND1 --> |No| DD[Process request]
    DD --> EE{Rate limit reached?}
    EE --> |Yes| FF[Wait until tokens are regenerated]
    FF --> EE
    EE --> |No| GG[Send response]
    GG --> End2
end

subgraph "*** Before ***"
    Start([START]) --> A[Receive request]
    A --> B{Rate limit reached <br> or <br> request is too large?}
    B -->|Yes| C[Send error response]
    C --> End([END])
    B -->|No| E[Process request]
    E --> F[Send response]
    F --> End
end
```

### `Is there already an active request with the same protocol?`

This check is not performed in `Before`. This is taken from the PR in the consensus-spec, which proposes updates regarding rate limiting and response timeout.
https://github.com/ethereum/consensus-specs/pull/3767/files
> The requester MUST NOT make more than two concurrent requests with the same ID.

The PR mentions the requester side. In this PR, I introduced the `ActiveRequestsLimiter` for the `responder` side to restrict more than two requests from running simultaneously on the same protocol per peer. If the limiter disallows a request, the responder sends a rate-limited error and penalizes the requester.

### `Rate limit reached?` and `Wait until tokens are regenerated`

UPDATE: I moved the limiter logic to the behaviour side. https://github.com/sigp/lighthouse/pull/5923#issuecomment-2379535927

~~The rate limiter is shared between the behaviour and the handler.  (`Arc<Mutex<RateLimiter>>>`) The handler checks the rate limit and queues the response if the limit is reached. The behaviour handles pruning.~~

~~I considered not sharing the rate limiter between the behaviour and the handler, and performing all of these either within the behaviour or handler. However, I decided against this for the following reasons:~~

- ~~Regarding performing everything within the behaviour: The behaviour is unable to recognize the response protocol when `RPC::send_response()` is called, especially when the response is `RPCCodedResponse::Error`. Therefore, the behaviour can't rate limit responses based on the response protocol.~~
- ~~Regarding performing everything within the handler: When multiple connections are established with a peer, there could be multiple handlers interacting with that peer. Thus, we cannot enforce rate limiting per peer solely within the handler. (Any ideas? 🤔 )~~

#### Naming

I have renamed the fields of the behaviour to make them more intuitive: https://github.com/sigp/lighthouse/pull/5923#issuecomment-2719236781

#### Testing

 I have run beacon node with this changes for 24hours, it looks work fine.

The rate-limited error has not occurred anymore while running this branch.

image

#### Metrics

These metrics have been added in this PR.

- `self_limiter_request_idling_seconds`: The time our own request remained idle in the self-limiter.
- `response_limiter_idling_seconds`: The time our response remained idle in the response limiter.


beacon_node/lighthouse_network/src/rpc/active_requests_limiter.rs
```diff
@@ -0,0 +1,178 @@
...
+use libp2p::PeerId;
+use std::collections::hash_map::Entry;
+use std::collections::HashMap;
+use std::time::Duration;
+use tokio::time::Instant;
+
+/// Restricts more than two requests from running simultaneously on the same protocol per peer.
+pub(super) struct ActiveRequestsLimiter {
+    resp_timeout: Duration,
+    requests: HashMap<PeerId, Vec<(Protocol, ConnectionId, SubstreamId, Instant)>>,
```
pawanjay176: I'm not convinced why we need `ConnectionId` and `SubstreamId` here. 
They would be relevant in the case where a peer has multiple connections but even in that case, I don't think we want a peer to split multiple requests over multiple connection ids.

Even for `Instant`, I'm not fully convinced. We wouldn't respond within 15 secs only when we are heavily overloaded so we are probably not capable of serving a new request and in this case. Are there any other cases where replacing the request would be relevant?

Adding `Instant` makes sense but doesn't seem to useful to me in a realistic scenario.

beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -348,57 +376,60 @@ where
...
+                            "Rate limited. There is an active request with the same protocol"
+                                .into(),
+                        ),
+                    );
+                    return;
+                }
+
+                if let Some(limiter) = self.inbound_request_size_limiter.as_ref() {
+                    // Check if the request is conformant to the quota
+                    if limiter.allows(req) {
```
pawanjay176: Shouldn't the limiter here be peer specific here? Do we want rate limiting here if multiple peers collectively cross the limit?
For e.g. for a ping request which is 2 requests every 10 secs, if we get > 2 peers sending us ping requests in a 10 sec window, the third peer will get rate limited here even though it is its first ping request

beacon_node/lighthouse_network/src/rpc/active_requests_limiter.rs
```diff
@@ -0,0 +1,178 @@
+use crate::rpc::{Protocol, SubstreamId};
+use libp2p::swarm::ConnectionId;
+use libp2p::PeerId;
+use std::collections::hash_map::Entry;
+use std::collections::HashMap;
+use std::time::Duration;
+use tokio::time::Instant;
+
+/// Restricts more than two requests from running simultaneously on the same protocol per peer.
```
jxs:
```suggestion
/// Restricts more than two inbound requests from running simultaneously on the same protocol per peer.
```

beacon_node/lighthouse_network/src/rpc/active_requests_limiter.rs
jxs: > I'm not convinced why we need ConnectionId and SubstreamId here.
They would be relevant in the case where a peer has multiple connections but even in that case, I don't think we want a peer to split multiple requests over multiple connection ids.

agree with Pawan here, I think Akihito used the combination of `ConnectionId` and `SubstreamId` for the same reason he mentioned in the PR:
>Regarding performing everything within the behaviour: The behaviour is unable to recognize the response protocol when RPC::send_response() is called, especially when the response is RPCCodedResponse::Error. 

Therefore, the behaviour can't then remove the Request from the `ActiveRequestsLimiter`
as we currently don't have `Protocol` on the `RPCCodedResponse`, but imho it's cleaner to just have `HashSet<(PeerId, Protocol)>` and align on a single `Protocol` enum in the repo  (for now we can have `From<&RPCCodedResponse> for Protocol`)

>Even for Instant, I'm not fully convinced. We wouldn't respond within 15 secs only when we are heavily overloaded so we are probably not capable of serving a new request and in this case. Are there any other cases where replacing the request would be relevant?

>Adding Instant makes sense but doesn't seem to useful to me in a realistic scenario.

as per the [spec update PR](https://github.com/ethereum/consensus-specs/pull/3767/files#diff-fba209728a99ea3daf786665da792e9bdcfcbe52c937f3cb38874aa5b934d981R629):
> The responder MAY penalise peers that concurrently open more than two streams for the same request type, for the protocol IDs defined in this specification.

Without mention of a timeout, and [later it also mentions](https://github.com/ethereum/consensus-specs/pull/3767/files#diff-fba209728a99ea3daf786665da792e9bdcfcbe52c937f3cb38874aa5b934d981R1468):

> A typical approach for the requesting side is to implement a timeout on the request that depends on the nature of the request and on connectivity parameters in general - for example when requesting blocks, a peer might choose to send a request to a second peer if the first peer does not respond within a reasonable time, and to reset the request to the first peer if the second peer responds faster. Clients may use past response performance to reward fast peers when implementing peer scoring.

So Rpc should just penalize if a requesting peer sends the same protocol request twice, it should be the requesting side implementing a timeout and requesting it to another peer it seems.
With that in mind Akihito, what do you think of merging this part of the limiter into the main limiter? Also because this one it not optional whereas the main one is, btw @age, with the spec update PR in mind,  is there a situation where rate limiting may be optional?

beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -348,57 +376,60 @@ where
...
+                            protocol,
+                            Protocol::BlocksByRange
+                                | Protocol::BlobsByRange
+                                | Protocol::BlocksByRoot
+                                | Protocol::BlobsByRoot
+                        ) {
+                            debug!(self.log, "Request too large to process"; "request" => %req, "protocol" => %protocol);
+                        } else {
+                            // Other protocols shouldn't be sending large messages, we should flag the peer kind
+                            crit!(self.log, "Request size too large to ever be processed"; "protocol" => %protocol);
```
jxs:
```suggestion
                            crit!(self.log, "Request size too large to even be processed"; "protocol" => %protocol);
```

beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -117,10 +121,15 @@ pub struct NetworkParams {
...
 pub struct RPC<Id: ReqId, E: EthSpec> {
-    /// Rate limiter
-    limiter: Option<RateLimiter>,
+    /// Rate limiter for our responses. This is shared with RPCHandlers.
+    response_limiter: Option<Arc<Mutex<RateLimiter>>>,
     /// Rate limiter for our own requests.
-    self_limiter: Option<SelfRateLimiter<Id, E>>,
+    outbound_request_limiter: Option<SelfRateLimiter<Id, E>>,
+    /// Limiter for inbound requests, which checks the request size.
+    inbound_request_size_limiter: Option<RequestSizeLimiter>,
```
jxs: Ideally I think this should also belong inside `RateLimiter` so that `Behaviour` only knows about the inbound and the outbound request limiters?

jxs (commented):
Hi Akihito, thanks for starting and driving this, left some comments but the main one is that I think we can simplify this a lot by aligning the `rpc::protocol::Protocol` with the `RPCResponse` wdyt?

AgeManning:
This is a significant change to our RPC behaviour. As we are already putting quite a lot of significant upgrades into 5.3, i suggest we hold off till 5.3.1 for this PR

beacon_node/lighthouse_network/src/rpc/active_requests_limiter.rs
ackintosh: > Therefore, the behaviour can't then remove the Request from the ActiveRequestsLimiter
as we currently don't have Protocol on the RPCCodedResponse, 

Yes, that's why I think we need the combination of `ConnectionId` and `SubstreamId`.

> but imho it's cleaner to just have HashSet<(PeerId, Protocol)> and align on a single Protocol enum in the repo (for now we can have From<&RPCCodedResponse> for Protocol)

That would be cleaner, but I'm not sure if we can do that, because `RPCCodedResponse::Error` doesn’t contain a protocol. 🤔

ackintosh: > Even for Instant, I'm not fully convinced. We wouldn't respond within 15 secs only when we are heavily overloaded so we are probably not capable of serving a new request and in this case. Are there any other cases where replacing the request would be relevant?

The intention of adding `Instant` was to ensure this limiter allows new requests even if there are some bugs that cause requests to remain in this limiter despite the response having been sent.

I thought again about this, and yeah, it might not be useful. I’m happy to remove it. 🙂

beacon_node/lighthouse_network/src/rpc/mod.rs
ackintosh: `self.inbound_request_size_limiter` (`RequestSizeLimiter`)  looks similar to the `Limiter`, but it just checks if the request is too large; it doesn't perform rate-limiting. So, this doesn't need to be peer-specific.


For reference, here is the reason why I introduced `RequestSizeLimiter`, quoted from [the PR comment](https://github.com/sigp/lighthouse/pull/5923#issue-2351213755):

> This request size check is also performed in `Before` using the `Limiter`, but in this PR, I introduced `RequestSizeLimiter` to handle this. Unlike the `Limiter`, `RequestSizeLimiter` is dedicated to perform this check.
>
> The reasons why I introduced `RequestSizeLimiter` are:
>
> - In `After`, the rate limiter is shared between the behaviour and the handler (`Arc<Mutex<RateLimiter>>>`). (This is detailed at the next sentence)
> - The request size check does not heavily depend on the rate-limiting logic, so it can be separated with minimal code duplication.
> - The request size check is performed on the behaviour side. By separating the request size check from the rate limiter, we can reduce the locking of the rate limiter.

beacon_node/lighthouse_network/src/rpc/mod.rs
ackintosh: That might be better in terms of cleaner code. The reason why I introduced `RequestSizeLimiter` is that `RateLimiter` is shared between behaviour and connection handler as `Arc<Mutex<RateLimiter>>`. Locking time will be increased if we move `RequestSizeLimiter` inside `RateLimiter` and call the (shared) limiter in `on_connection_handler_event`. 

For reference, here is the reason why I introduced `RequestSizeLimiter`, quoted from [the PR comment](https://github.com/sigp/lighthouse/pull/5923#issue-2351213755):

> This request size check is also performed in `Before` using the `Limiter`, but in this PR, I introduced `RequestSizeLimiter` to handle this. Unlike the `Limiter`, `RequestSizeLimiter` is dedicated to perform this check.
>
> The reasons why I introduced `RequestSizeLimiter` are:
>
> - In `After`, the rate limiter is shared between the behaviour and the handler (`Arc<Mutex<RateLimiter>>>`). (This is detailed at the next sentence)
> - The request size check does not heavily depend on the rate-limiting logic, so it can be separated with minimal code duplication.
> - The request size check is performed on the behaviour side. By separating the request size check from the rate limiter, we can reduce the locking of the rate limiter.

beacon_node/lighthouse_network/src/rpc/mod.rs
pawanjay176: Ohh thanks for explaining. 

In that case, can't we vastly simplify the entire check by just doing
```rust
match request.protocol() {
            Protocol::Ping
            | Protocol::Status
            | Protocol::Goodbye
            | Protocol::MetaData
            | Protocol::LightClientBootstrap
            | Protocol::LightClientOptimisticUpdate
            | Protocol::LightClientFinalityUpdate
            | Protocol::BlocksByRoot
            | Protocol::BlobsByRoot => true,
            Protocol::BlocksByRange => request.max_responses() > spec.max_request_blocks(),
            Protocol::BlobsByRange => request.max_responses() > spec.max_blobs(),
        }
```

For the roots requests, the RuntimeVariable ssz list ensures that we don't get more requests than the max specified in the config. The other non block/blob protocols will always return true in the limiter anyway

ackintosh: That makes sense!

I’m working on the simplification you suggested, but I couldn’t find `spec.max_blobs()`. Could you provide any advice on this? 🙏

pawanjay176: Sorry i meant `max_request_blob_sidecars`

beacon_node/lighthouse_network/src/rpc/mod.rs
ackintosh: Suggestion from pawan about `RequestSizeLimiter`: https://github.com/sigp/lighthouse/pull/5923#discussion_r1689948084

I removed the `RequestSizeLimiter` and added `is_request_size_too_large()` instead, which checks if the count of requested blocks/blobs is within the number defined in the spec. That looks much simpler. 🙂  https://github.com/sigp/lighthouse/pull/5923/commits/5a9237fdc6a5ad11281ec7c36b2b7bdfdeba7ed4

beacon_node/lighthouse_network/src/rpc/mod.rs
ackintosh: https://github.com/sigp/lighthouse/pull/5923/commits/5a9237fdc6a5ad11281ec7c36b2b7bdfdeba7ed4

I removed the `RequestSizeLimiter` and added `is_request_size_too_large()` instead, which does the check. That looks much simpler. ✨ Thanks for your suggestion.

jimmygchen (commented):
I've only looked at the diagram and have some questions to the new behaviour:

- Does this change mean we no longer send rate limited response to requester?
- If we delay the response by 10 seconds, the response could be less useful to the requester, but still end up consuming the rate limit bucket and this could result in a snowball effect?  and the requester would not know the response came back late due to rate limiting.

AgeManning:
@jimmygchen - Yep, we essentially remove the rate limit response.

This has been a point of contention from the beginning of consensus clients. Some clients will not spec a rate-limit response and feel its the wrong strategy. 

If we can't spec a rate-limit response, then sending one isn't very useful. Even at this moment, Lighthouse doesn't really acknowledge the rate limit response from another Lighthouse node. It just treats it as a general error. We dont have logic to parse the timeout and then re-request. Adding this kind of logic (especially in sync as you know) will complicate things quite a bit. 

After quite a few discussions with other client teams, it seems the lesser of all evils is to take this approach. Where we naturally rate limit them in time. If we are to rate limit someone over the 10 second mark, we should just not send the message at all. At the moment, this is equivalent (from a peer scoring perspective) as sending a rate-limit message. Both are down scored. 
The down-side of this approach is that the receiver has to wait 10 seconds to find out that the message isn't coming. I guess in this case, we should send a rate limit message to not make them wait 10 seconds.

AgeManning:
I'm keen to get this in. @ackintosh - Let me know when comments are resolved and this is ready for a final review

beacon_node/lighthouse_network/src/rpc/active_requests_limiter.rs
ackintosh: Removed `Instant` in  https://github.com/sigp/lighthouse/pull/5923/commits/a325438d8ca141b71501af44ac25e54f83f8f182

ackintosh:
@AgeManning @pawanjay176 @jxs This PR is now ready for review again. I've addressed the previous feedback and merged the latest unstable branch.

beacon_node/lighthouse_network/src/rpc/active_requests_limiter.rs
```diff
@@ -0,0 +1,121 @@
+use crate::rpc::{Protocol, SubstreamId};
+use libp2p::swarm::ConnectionId;
+use libp2p::PeerId;
+use std::collections::hash_map::Entry;
+use std::collections::HashMap;
+
+/// Restricts more than two inbound requests from running simultaneously on the same protocol per peer.
```
pawanjay176: This seems to be restricting more than one inbound request, not two.

I'm not completely sure about the intuition for allowing 2 concurrent streams, allowing a single stream per protocol makes more sense to me, I have asked in the spec PR https://github.com/ethereum/consensus-specs/pull/3767#issuecomment-2349955453

beacon_node/lighthouse_network/src/rpc/active_requests_limiter.rs
```diff
@@ -0,0 +1,121 @@
...
+            }
+            Entry::Vacant(entry) => {
+                // No active requests for the peer.
+                entry.insert(vec![(protocol, *connection_id, *substream_id)]);
+                true
+            }
+        }
+    }
+
+    /// Removes the request with the given SubstreamId.
```
pawanjay176:
```suggestion
    /// Removes the request with the given SubstreamId and ConnectionId.
```

pawanjay176 (commented):
Few more comments. 
Also noting that we currently reject requests on the server side that exceed max concurrent streams per protocol of 1, but we don't restrict it on the client side.
We should probably add an `ActiveRequestLimiter` type object in self limiter as well?

ackintosh:
~~I received advice from Pawan, and I am now trying to add the protocol info to RPC errors so that we can handle all the rate-limiting logic within the behaviour.~~

https://github.com/sigp/lighthouse/pull/6444 has been merged, so we can remove the need to share a `Mutex`.

ackintosh:
Based on #6444, I've fixed the `Mutex` issue. Now we have the entire response limiting logic in the behaviour instead of shareing between the behaviour and handler.

Also I'm digging into an issue where a node is immediately banned with the following logs:

```
3117:Oct 08 12:37:45.665 DEBG RPC Error                               direction: Incoming, score: inf, peer_id: 16Uiu2HAkw8cyrjC4UpnQKxdDGHsCQwK32PBKMmdc4xBdsnC8iDYg, client: Lighthouse: version: v5.3.0-156565c+, os_version: aarch64-macos, err: RPC response was an error: Rate limited with reason: Rate limited. There is an active request with the same protocol, protocol: data_column_sidecars_by_root, service: libp2p
3598:Oct 08 12:37:57.658 DEBG RPC Error                               direction: Incoming, score: inf, peer_id: 16Uiu2HAkw8cyrjC4UpnQKxdDGHsCQwK32PBKMmdc4xBdsnC8iDYg, client: Lighthouse: version: v5.3.0-156565c+, os_version: aarch64-macos, err: RPC response was an error: Rate limited with reason: Rate limited. There is an active request with the same protocol, protocol: data_column_sidecars_by_root, service: libp2p
3645:Oct 08 12:37:58.599 DEBG RPC Error                               direction: Incoming, score: inf, peer_id: 16Uiu2HAkw8cyrjC4UpnQKxdDGHsCQwK32PBKMmdc4xBdsnC8iDYg, client: Lighthouse: version: v5.3.0-156565c+, os_version: aarch64-macos, err: RPC response was an error: Rate limited with reason: Rate limited. There is an active request with the same protocol, protocol: data_column_sidecars_by_root, service: libp2p
```

ackintosh:
@pawanjay176 @AgeManning @jxs 
As Pawan mentioned [here](https://github.com/sigp/lighthouse/pull/5923#pullrequestreview-2303887056), we should add a limiter logic similar to `ActiveRequestsLimiter` in self limiter as well, to limit max concurrent streams per protocol on the client side. From my testing on a local testnet, multiple `data_column_sidecars_by_root` requests are being sent to the same peer at the same time. It seems that this is more likely to occur when the network is smaller in scale.

I think, unlike the server side limiter, the client side limiter we'll add would just delay the request that exceeds max concurrent streams per protocol, instead of rejecting it. So, [the limiter](https://github.com/sigp/lighthouse/blob/9008d3e3b09999b74296bbf1aac02cdf713dfb99/beacon_node/lighthouse_network/src/rpc/self_limiter.rs#L35) in `SelfRateLimiter` should be replaced with `ActiveRequestsLimiter`, so that requests are limited based on the number of concurrent requests rather than tokens.

What do you think? I just want to make sure this is the right direction.

beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -151,10 +154,15 @@ pub struct NetworkParams {
...
+    /// Rate limiter for our responses.
+    response_limiter: Option<ResponseLimiter<E>>,
     /// Rate limiter for our own requests.
-    self_limiter: Option<SelfRateLimiter<Id, E>>,
+    outbound_request_limiter: Option<SelfRateLimiter<Id, E>>,
+    /// Limiter for inbound requests, which restricts more than two requests from running
+    /// simultaneously on the same protocol per peer.
+    active_inbound_requests_limiter: ActiveRequestsLimiter,
+    /// Active inbound requests that are awaiting a response.
+    active_inbound_requests: HashMap<RequestId, (ConnectionId, Request<E>)>,
```
jxs: if we require `ConnectionId` here let's rather move it into the `Request` struct wdyt Akihito?

beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -207,22 +216,54 @@ impl<Id: ReqId, E: EthSpec> RPC<Id, E> {
...
-        id: (ConnectionId, SubstreamId),
-        _request_id: RequestId,
+        _id: (ConnectionId, SubstreamId),
+        request_id: RequestId,
         event: RpcResponse<E>,
     ) {
+        let Some((connection_id, request)) = self.active_inbound_requests.remove(&request_id)
+        else {
+            error!(self.log, "Request not found in active_inbound_requests. Response not sent"; "peer_id" => %peer_id, "request_id" => ?request_id, "response" => %event);
+            return;
```
jxs, line 221: yeah I thought this could happen, I suggest we change the signature of the function to `send_response) -> Result<>` and let whoever calls it to deal with an invalid `Request`

beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -207,22 +216,54 @@ impl<Id: ReqId, E: EthSpec> RPC<Id, E> {
...
+
+        if let Some(response_limiter) = self.response_limiter.as_mut() {
+            if !response_limiter.allows(
+                peer_id,
+                request.r#type.protocol(),
+                connection_id,
+                request.substream_id,
+                event.clone(),
+            ) {
+                return;
```
jxs, line 252: same here, we can return an `Error` for rate limited requests so that the caller can decide what to do if the request was not allowed, instead of silent ignoring it

beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -207,22 +216,54 @@ impl<Id: ReqId, E: EthSpec> RPC<Id, E> {
...
+        request_id: RequestId,
         event: RpcResponse<E>,
     ) {
+        let Some((connection_id, request)) = self.active_inbound_requests.remove(&request_id)
+        else {
+            error!(self.log, "Request not found in active_inbound_requests. Response not sent"; "peer_id" => %peer_id, "request_id" => ?request_id, "response" => %event);
+            return;
+        };
+
+        // Add the request back to active requests if the response is not a stream termination.
```
jxs: if we do this at the end of the scope we don't need to clone the request here right?

beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -207,22 +216,54 @@ impl<Id: ReqId, E: EthSpec> RPC<Id, E> {
     pub fn send_response(
         &mut self,
         peer_id: PeerId,
-        id: (ConnectionId, SubstreamId),
-        _request_id: RequestId,
+        _id: (ConnectionId, SubstreamId),
```
jxs: I like this, after this we can deprecate `(ConnectionId, SubstreamId)` as the request identifier

beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -405,82 +471,69 @@ where
...
-                        // No rate limiting, continue.
-                        Ok(()) => {}
-                    }
-                }
+                let request = Request {
+                    id,
+                    substream_id,
+                    r#type,
+                };
+                self.active_inbound_requests
```
jxs: shouldn't we only insert the request if it is allowed?

beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -405,82 +471,69 @@ where
...
-                        substream_id,
-                        r#type,
-                    })),
-                }));
+                        RpcResponse::Error(
+                            RpcErrorResponse::InvalidRequest,
+                            "The request requires responses greater than the number defined in the spec.".into(),
+                        ),
+                    );
+                } else {
```
jxs: nit: instead of of an `if else` block where the `else` code ends way below I'd argue it's more legible to return early if the `if` condition matches, so that the subsequent block is at function scope level, in this case, if the request is too large return instead of having an else.

beacon_node/lighthouse_network/src/rpc/active_requests_limiter.rs
```diff
@@ -0,0 +1,121 @@
...
+                    .push((protocol, *connection_id, *substream_id));
+                true
+            }
+            Entry::Vacant(entry) => {
+                // No active requests for the peer.
+                entry.insert(vec![(protocol, *connection_id, *substream_id)]);
+                true
+            }
+        }
+    }
```
jxs: since this function is now just this, wdyt of having this function at the `Behaviour` level, removing `ActiveRequestsLimiteR` and therefore eliminating  the need for a duplicated `HashMap` with requests?

jxs (commented):
Thanks for continuing driving this Akihito!

>@pawanjay176 @AgeManning @jxs
As Pawan mentioned https://github.com/sigp/lighthouse/pull/5923#pullrequestreview-2303887056, we should add a limiter logic similar to ActiveRequestsLimiter in self limiter as well, to limit max concurrent streams per protocol on the client side. From my testing on a local testnet, multiple data_column_sidecars_by_root requests are being sent to the same peer at the same time. It seems that this is more likely to occur when the network is smaller in scale.

>I think, unlike the server side limiter, the client side limiter we'll add would just delay the request that exceeds max concurrent streams per protocol, instead of rejecting it. So, [the limiter](https://github.com/sigp/lighthouse/blob/9008d3e3b09999b74296bbf1aac02cdf713dfb99/beacon_node/lighthouse_network/src/rpc/self_limiter.rs#L35) in SelfRateLimiter should be replaced with ActiveRequestsLimiter, so that requests are limited based on the number of >concurrent requests rather than tokens.
>
>What do you think? I just want to make sure this is the right direction.

Can we leave that for a subsequent PR and in this one close the rework of the inbound request rate limiter?

AgeManning:
> @pawanjay176 @AgeManning @jxs As Pawan mentioned [here](https://github.com/sigp/lighthouse/pull/5923#pullrequestreview-2303887056), we should add a limiter logic similar to `ActiveRequestsLimiter` in self limiter as well, to limit max concurrent streams per protocol on the client side. From my testing on a local testnet, multiple `data_column_sidecars_by_root` requests are being sent to the same peer at the same time. It seems that this is more likely to occur when the network is smaller in scale.
> 
> I think, unlike the server side limiter, the client side limiter we'll add would just delay the request that exceeds max concurrent streams per protocol, instead of rejecting it. So, [the limiter](https://github.com/sigp/lighthouse/blob/9008d3e3b09999b74296bbf1aac02cdf713dfb99/beacon_node/lighthouse_network/src/rpc/self_limiter.rs#L35) in `SelfRateLimiter` should be replaced with `ActiveRequestsLimiter`, so that requests are limited based on the number of concurrent requests rather than tokens.
> 
> What do you think? I just want to make sure this is the right direction.

@pawanjay176 @jxs 

I'm not fully across this. But I recall the self-limiter is optional and it is used so that we don't try and send more messages than would break the rate limit. If we set the limiter to only be based on number of concurrent requests, we could still make a concurrent request that breaks the rate limit? Or is that not possible? 

Just curious as to whether we need to limit tokens and concurrent requests or is just concurrent requests fine?

jxs:
> > @pawanjay176 @AgeManning @jxs As Pawan mentioned [here](https://github.com/sigp/lighthouse/pull/5923#pullrequestreview-2303887056), we should add a limiter logic similar to `ActiveRequestsLimiter` in self limiter as well, to limit max concurrent streams per protocol on the client side. From my testing on a local testnet, multiple `data_column_sidecars_by_root` requests are being sent to the same peer at the same time. It seems that this is more likely to occur when the network is smaller in scale.
> > I think, unlike the server side limiter, the client side limiter we'll add would just delay the request that exceeds max concurrent streams per protocol, instead of rejecting it. So, [the limiter](https://github.com/sigp/lighthouse/blob/9008d3e3b09999b74296bbf1aac02cdf713dfb99/beacon_node/lighthouse_network/src/rpc/self_limiter.rs#L35) in `SelfRateLimiter` should be replaced with `ActiveRequestsLimiter`, so that requests are limited based on the number of concurrent requests rather than tokens.
> > What do you think? I just want to make sure this is the right direction.
> 
> @pawanjay176 @jxs
> 
> I'm not fully across this. But I recall the self-limiter is optional and it is used so that we don't try and send more messages than would break the rate limit. If we set the limiter to only be based on number of concurrent requests, we could still make a concurrent request that breaks the rate limit? Or is that not possible?
> 
> Just curious as to whether we need to limit tokens and concurrent requests or is just concurrent requests fine?

if and after the [spec PR](https://github.com/ethereum/consensus-specs/pull/3767/files) gets merged  the self limiter (or some reworked form of self limiter) should not be optional, as per spec update:

> The requester MUST NOT make more than `MAX_CONCURRENT_REQUESTS` concurrent requests with the same protocol ID.

beacon_node/lighthouse_network/src/rpc/mod.rs
ackintosh: Yeah, I thought about that. The handler [creates a `Request`](https://github.com/sigp/lighthouse/blob/5dbac58af4f260d919e7d0a97e7b3812cb9c0f40/beacon_node/lighthouse_network/src/rpc/handler.rs#L898-L903), but it doesn’t know the `ConnectionId`, so I didn’t proceed with that approach. Do you have any ideas?

beacon_node/lighthouse_network/src/rpc/mod.rs
ackintosh: In this case, where the `response_limiter` doesn’t allow the response, [it is queued inside the limiter](https://github.com/sigp/lighthouse/blob/cb87af07f3bc2c6a02f08f6408891539fc675285/beacon_node/lighthouse_network/src/rpc/response_limiter.rs#L78-L88) and [will eventually be sent after a delay](https://github.com/sigp/lighthouse/blob/cb87af07f3bc2c6a02f08f6408891539fc675285/beacon_node/lighthouse_network/src/rpc/mod.rs#L563-L579). So I don’t think it’s necessary to return an error.

beacon_node/lighthouse_network/src/rpc/mod.rs
jxs: you can pass it to it when creating the handler, `handle_established` functions have the `ConnectionId`

beacon_node/lighthouse_network/src/rpc/mod.rs
jxs: yup, thanks for explaining Akihito

beacon_node/lighthouse_network/src/rpc/mod.rs
ackintosh: Ah, that makes sense thanks João!

beacon_node/lighthouse_network/src/rpc/mod.rs
ackintosh: We are returning early [here](https://github.com/sigp/lighthouse/blob/023c54234a9b2064fbc4429a4936599afd7b0717/beacon_node/lighthouse_network/src/rpc/mod.rs#L245), so I didn't put it at the end.

beacon_node/lighthouse_network/src/rpc/mod.rs
ackintosh: We need to insert the request regardless of whether it is allowed, since we send an error response (RateLimited) if it is not allowed. This might not be intuitive, so I'll add a comment to explain it. 🙏

beacon_node/lighthouse_network/src/rpc/mod.rs
ackintosh: It definitely improves readability. Thanks for the suggestion!

beacon_node/lighthouse_network/src/rpc/active_requests_limiter.rs
ackintosh: That makes sense. 💡 Thanks!

ackintosh: Removed `ActiveRequestsLimiter` in https://github.com/sigp/lighthouse/pull/5923/commits/3c058b3bacdc378721d1c6faf2a4d65c5b77725e

beacon_node/lighthouse_network/src/rpc/mod.rs
ackintosh: Added a comment in https://github.com/sigp/lighthouse/pull/5923/commits/14ffeec4cc344bb9e48df847bd3c10e443f920a5

beacon_node/lighthouse_network/src/rpc/active_requests_limiter.rs
ackintosh: Fixed in  https://github.com/sigp/lighthouse/pull/5923/commits/3c058b3bacdc378721d1c6faf2a4d65c5b77725e#diff-7269dc8a543f95b719952593e35056791882dbe4c8019e387530ed7d0429bd7cR54-R55

ackintosh:
> Just curious as to whether we need to limit tokens and concurrent requests or is just concurrent requests fine?

I tested this with a simple script. Even if we set the limiter to be based only on the number of concurrent requests, we still make a concurrent request that breaks the rate limit. So, it seems we need to limit both tokens and concurrent requests, although it adds more complexity. Limiting tokens could be optional, but limiting concurrent requests should not be, according to the spec.

pawanjay176:
>  Even if we set the limiter to be based only on the number of concurrent requests, we still make a concurrent request that breaks the rate limit.

I did not understand this. If we get a request that breaks the concurrent limit, why can't we just wait to send that until the existing streams have concluded?

Right now, if I run this branch on a kurtosis peerdas network, the lighthouse server rightly responds with an error because the lighthouse client is breaking the concurrency limit
```
Oct 30 02:23:38.044 DEBG RPC Error                               direction: Outgoing, score: 0, peer_id: 16Uiu2HAmAcK84B5BJz3EiXzLK7SzQsEi7JWNsqL1YcdiAGZDj4aA, client: Lighthouse: version: v5.3.0-450326c, os_version: aarch64-linux, err: RPC response was an error: Rate limited with reason: Rate limited. There is an active request with the same protocol, protocol: data_column_sidecars_by_range, service: libp2p
```

ackintosh:
[This](https://github.com/ackintosh/lighthouse/blob/delayed-rpc-response-debug-rate-limiting/beacon_node/lighthouse_network/tests/rpc_tests.rs#L1393-L1512) is the test I used (created in another branch forked from this PR). In the test, the sender maintains two (or fewer) active requests at a time, but self rate-limiting is triggered. The test is quite simple, but I think it shows that we need to limit both concurrent requests and (optionally) tokens. What do you think?

image

ackintosh:
The [spec PR](https://github.com/ethereum/consensus-specs/pull/3767) has been merged.

jxs:
> The [spec PR](https://github.com/ethereum/consensus-specs/pull/3767) has been merged.

nice! Is this ready for another review then @ackintosh?

ackintosh:
Let me organize the remaining tasks of this PR :pray: :

1. Need to limit concurrent requests on client (outbound requests) side, to comply with the spec update.
1. Whether we also need to limit tokens is under duscussion.

Note: Currently, if we run this branch (on a Kurtosis local network), the lighthouse server responds with an error as the client side is exceeding the concurrency limit, resulting in a ban.

I think task `1.` needs to be implemented in this PR because of the banning mentioned above. What do you think, @jxs? (Asking because of [this comment](https://github.com/sigp/lighthouse/pull/5923#pullrequestreview-2366031333))

jxs:
> Let me organize the remaining tasks of this PR 🙏 :
> 
>     1. Need to limit concurrent requests on client (outbound requests) side, to comply with the spec update.
> 
>     2. Whether we also need to limit tokens is under duscussion.
> 
> 
> Note: Currently, if we run this branch (on a Kurtosis local network), the lighthouse server responds with an error as the client side is exceeding the concurrency limit, resulting in a ban.
> 
> I think task `1.` needs to be implemented in this PR because of the banning mentioned above. What do you think, @jxs? (Asking because of [this comment](https://github.com/sigp/lighthouse/pull/5923#pullrequestreview-2366031333))

yeah makes sense Akihito, go for it :rocket:

ackintosh:
https://github.com/sigp/lighthouse/actions/runs/12109786536/job/33759312718?pr=5923

The `network-tests` failure was not reproduced locally. It might be related to https://github.com/sigp/lighthouse/pull/6646.

ackintosh:
@jxs @pawanjay176 @AgeManning 

This is ready for another review. 🙏 

I have added a concurrency limit on the self-lmiter. Now, the self-limiter limits outbound requests based on both the number of concurrent requests and tokens (optional). Whether we also need to limit tokens in the self-limiter is still under duscussion. Let me know if you have any ideas.

---

(FYI)

I also ran lighthouse (this branch) on the testnet for ~24hours. During this time, the LH node responded with 21 RateLimited errors due to the number of active requests. These errors appear in the logs like the example below. Note that this is about inbound requests, not the self-limiter (outbound requests).

```
Dec 09 13:38:56.806 DEBG There is an active request with the same protocol, protocol: beacon_blocks_by_range, request: Blocks by range: Start Slot: 2738468, Count: 64, Step: 1, peer_id: 16Uiu2HAmERvtCC321A2Nu1pH6QmhXgvqALxgCnrp4qxr2qeVWr2P, service: libp2p_rpc, service: libp2p, module: lighthouse_network::rpc:491
```

AgeManning:
@pawanjay176 @jxs @dapplion @jimmygchen  - If anyone has any spare time, I think this is a good one to get in.

ackintosh:
I removed `RPC::is_request_size_too_large` in https://github.com/sigp/lighthouse/pull/5923/commits/cfea9d2c04802af9c88c4aff681b29f573b0ac24 because such invalid requests are now handled in the Handler, as implemented in the following PRs:

- https://github.com/sigp/lighthouse/pull/6847
- https://github.com/sigp/lighthouse/pull/6986

beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -28,28 +27,33 @@ pub(crate) use methods::{
...
 mod outbound;
 mod protocol;
 mod rate_limiter;
+mod response_limiter;
 mod self_limiter;
 
 static NEXT_REQUEST_ID: AtomicUsize = AtomicUsize::new(1);
 
+// Maximum number of concurrent requests per protocol ID that a client may issue.
+const MAX_CONCURRENT_REQUESTS: usize = 2;
```
dapplion, line 52: This means we can have at most two `blocks_by_range` ReqResp requests per peer?

beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -403,65 +441,52 @@ where
...
+                if is_concurrent_request_limit_exceeded {
+                    // There is already an active request with the same protocol. Send an error code to the peer.
+                    debug!(self.log, "There is an active request with the same protocol"; "peer_id" => peer_id.to_string(), "request" => %request.r#type, "protocol" => %request.r#type.versioned_protocol().protocol());
+                    self.send_response(
+                        peer_id,
+                        (conn_id, substream_id),
+                        id,
+                        RpcResponse::Error(
+                            RpcErrorResponse::RateLimited,
+                            "Rate limited. There is an active request with the same protocol"
```
dapplion:
```suggestion
                            format!("Rate limited. There are already {MAX_CONCURRENT_REQUESTS} active requests with the same protocol")
```

beacon_node/lighthouse_network/src/rpc/response_limiter.rs
```diff
@@ -0,0 +1,177 @@
...
+            &self.log,
+        ) {
+            self.delayed_responses
+                .entry((peer_id, protocol))
+                .or_default()
+                .push_back(QueuedResponse {
+                    peer_id,
+                    connection_id,
+                    substream_id,
+                    response,
```
dapplion, line 86: We consider the memory cost of storing responses too low to worry? The worst case is buffering 2 x `data_columns_by_range` for all possible indices for all peers. For 9 blobs per block, that's `131072*2 * 9 * 100 / 1e6 = 235 MB`

dapplion: Is the point of this PR to not have an additional global rate limiter? It would help to reduce the worst case scenario

beacon_node/lighthouse_network/src/rpc/self_limiter.rs
```diff
@@ -139,8 +178,14 @@ impl<Id: ReqId, E: EthSpec> SelfRateLimiter<Id, E> {
         if let Entry::Occupied(mut entry) = self.delayed_requests.entry((peer_id, protocol)) {
             let queued_requests = entry.get_mut();
             while let Some(QueuedRequest { req, request_id }) = queued_requests.pop_front() {
```
dapplion: If you remove a `QueuedRequest ` item from `queued_requests` and the limiter doesn't allow to send, is this request lost?

dapplion (commented):
Overall architecture looks good! Just some comments.

Could you add a metric to track the time our own requests are idling in the self-rate limiter? It will help inform is sync performance is hindered by this new rate limit policy. We should also track the time outbound responses are idling in the rate limiter.

beacon_node/lighthouse_network/src/rpc/mod.rs
ackintosh: Yes, that's correct. Each peer can handle up to two blocks_by_range ReqResp requests at most.

beacon_node/lighthouse_network/src/rpc/self_limiter.rs
ackintosh: The `QueuedRequest` that the limiter did not allow is requeued in `queued_requests` here:
https://github.com/sigp/lighthouse/blob/1f521d69cbcc447451155439f88d32a30bccf3ef/beacon_node/lighthouse_network/src/rpc/self_limiter.rs#L189-L195

beacon_node/lighthouse_network/src/rpc/response_limiter.rs
ackintosh: Thanks for your feedback! I hadn't considered a global rate limiter, and your point about its potential to mitigate worst-case memory usage is well taken. 💡 I'll look into what configuration values would be optimal.

beacon_node/lighthouse_network/src/rpc/handler.rs
```diff
@@ -314,9 +326,10 @@ where
         if matches!(self.state, HandlerState::Deactivated) {
             // we no longer send responses after the handler is deactivated
             debug!(self.log, "Response not sent. Deactivated handler";
-                "response" => %response, "id" => inbound_id);
+                        "response" => %response, "id" => inbound_id);
```
jxs: nit: this indentation is not correct right?

beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -151,10 +157,12 @@ pub struct NetworkParams {
 /// Implements the libp2p `NetworkBehaviour` trait and therefore manages network-level
 /// logic.
 pub struct RPC<Id: ReqId, E: EthSpec> {
-    /// Rate limiter
-    limiter: Option<RateLimiter>,
+    /// Rate limiter for our responses.
+    response_limiter: Option<ResponseLimiter<E>>,
```
jxs, line 151:
```suggestion
    inbound_rate_limiter: Option<ResponseLimiter<E>>,
```
I would keep the previous naming, to also still follow the naming of the configurations. But mostly what we are rate limiting are external requests right? Not our responses

beacon_node/lighthouse_network/src/rpc/methods.rs
```diff
@@ -606,6 +606,20 @@ pub enum ResponseTermination {
     LightClientUpdatesByRange,
 }
 
+impl ResponseTermination {
+    pub fn protocol(&self) -> Protocol {
```
jxs: following the [naming conventions](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv) this could be `as_protocol`.
```suggestion
    pub fn as_protocol(&self) -> Protocol {
```

beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -151,10 +157,12 @@ pub struct NetworkParams {
...
     /// Rate limiter for our own requests.
-    self_limiter: Option<SelfRateLimiter<Id, E>>,
+    outbound_request_limiter: SelfRateLimiter<Id, E>,
```
jxs, line 153:
```suggestion
    outbound_rate_limiter: SelfRateLimiter<Id, E>,
```
Here I would follow the same naming, so that it's clear that both outbound and inbound we are rate limiting requests right?

beacon_node/lighthouse_network/src/rpc/mod.rs
jxs: can we then leave a comment explain that Akihito, as we have on `send_request`? 
Thanks!

beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -403,65 +445,52 @@ where
...
+                    .filter(|(_request_id, active_request)| {
+                        active_request.peer_id == peer_id
+                            && active_request.r#type.protocol() == request.r#type.protocol()
+                    })
+                    .count()
+                    >= MAX_CONCURRENT_REQUESTS;
+
+                // We need to insert the request regardless of whether it is allowed by the limiter,
+                // since we send an error response (RateLimited) if it is not allowed.
+                self.active_inbound_requests.insert(id, request.clone());
```
jxs: Hum, this is kinda fishy, may expose us to DOS attacks right? even if we are rate limiting the requester client we are still incrementing `active_inbound_requests` unboundly.
Can we rework `send_response` for an internal version that accepts the `request_id`? So that we only keep the requests that are not rate limited on `active_inbound_requests`

beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -476,21 +505,40 @@ where
...
                     conn_id,
                     message: Ok(rpc),
                 }));
             }
             HandlerEvent::Err(err) => {
+                // Inform the limiter that the request has ended with an error.
+                let protocol = match err {
+                    HandlerErr::Inbound { proto, .. } => proto,
+                    HandlerErr::Outbound { proto, .. } => proto,
+                };
```
jxs: nit:
```suggestion
                let protocol = match err {
                    HandlerErr::Inbound { proto, .. } | HandlerErr::Outbound { proto, .. } => proto,
                };
```

beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -476,21 +505,40 @@ where
...
-                        substream_id,
-                        r#type,
-                    })),
+                    message: Ok(RPCReceived::Request(request)),
                 }));
             }
             HandlerEvent::Ok(rpc) => {
+                // Inform the limiter that a response has been received.
+                match &rpc {
+                    RPCReceived::Request(_) => unreachable!(),
```
jxs: nit:
can we move this sub `match` arms to the main `match` to avoid the `unreachable`? 
I feel it's more elegant to repeat the `self.events.push` code than to have an `unreachable` here, wdyt Akihito?

beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -208,38 +219,59 @@ impl<Id: ReqId, E: EthSpec> RPC<Id, E> {
...
     pub fn send_request(&mut self, peer_id: PeerId, request_id: Id, req: RequestType<E>) {
-        let event = if let Some(self_limiter) = self.self_limiter.as_mut() {
-            match self_limiter.allows(peer_id, request_id, req) {
-                Ok(event) => event,
-                Err(_e) => {
-                    // Request is logged and queued internally in the self rate limiter.
-                    return;
-                }
+        match self
+            .outbound_request_limiter
```
jxs, line 263: ok now I have understood the naming.
Honestly I don't think it makes sense for us to rate limit our responses, from what I read in the code we weren't previously right?
I read the [spec update](https://github.com/ethereum/consensus-specs/pull/3767/files) and what it mentions is:
>The responder MAY rate-limit chunks by withholding each chunk until capacity is available. The responder MUST NOT respond with an error or close the stream when rate limiting.

but that's not what we are doing here from what I read, we are applying the previous `rate_limiter` to our responses right? 
If we want back pressure and gauge the outbound throughput I'd rather do it the way [we did in gosipsub](https://github.com/libp2p/rust-libp2p/pull/5595)

jxs (commented):
Hi, Akihito, thanks for following this. left some comments, specially regarding limiting our responses. 
I think we should just rate limit incoming requests.

beacon_node/lighthouse_network/src/rpc/methods.rs
ackintosh: I didn't know the naming conventions. 💡 Thanks!

beacon_node/lighthouse_network/src/rpc/mod.rs
ackintosh: Yeah, that's much simpler. Thanks!

ackintosh:
@jxs @dapplion @pawanjay176 @jimmygchen @AgeManning 

In this PR, the rate limiters have been updated in terms of naming and functionality according to the [spec update](https://github.com/ethereum/consensus-specs/pull/3767/files). Here is a table summarizing the changes to the limiters. The current naming might not be ideal, so I hope this will be helpful for your feedback. 🙏 

| Previous Naming                             | Current Naming                               | What This Does |
|---------------------------------------------|----------------------------------------------|-------------------------------------------------------------------|
| `limiter: Option<RateLimiter>`              | `response_limiter: Option<ResponseLimiter<E>>` | Previously, this rate-limited inbound requests. <br /> Currently, it rate-limits **our responses** to inbound requests. |
| `self_limiter: Option<SelfRateLimiter<Id, E>>` | `outbound_request_limiter: SelfRateLimiter<Id, E>` | Rate-limits **our outbound requests**.  <br /> Previously, it was optional.|

Additionally, a global rate limiter for inbound requests has been proposed to reduce memory cost in worst-case senarios.
( https://github.com/sigp/lighthouse/pull/5923#discussion_r1970446793 )

beacon_node/lighthouse_network/src/rpc/mod.rs
ackintosh: I pushed a [commit](https://github.com/sigp/lighthouse/pull/5923/commits/2a682ff1c555b6e90e3d2b9780cb96267913f286) to address the issue that you pointed out. Please let me know if you have any suggestions. :pray:

AgeManning:
Ok, in the interest of not getting this thing stale and requiring us to maintain it, how do we all feel about a merge?
@jxs @pawanjay176 @dapplion ?

dapplion:
I would like to have this in unstable soon, and work PeerDAS around this rate limit paradigm

pawanjay176:
I'm in favour of getting this in as well. Will test this for a bit and report back

beacon_node/lighthouse_network/src/metrics.rs
```diff
@@ -206,6 +206,20 @@ pub static REPORT_PEER_MSGS: LazyLock<Result<IntCounterVec>> = LazyLock::new(||
     )
 });
 
+pub static SELF_LIMITER_REQUEST_IDLING: LazyLock<Result<Histogram>> = LazyLock::new(|| {
```
pawanjay176: Just noting that if we are changing the naming as mentioned in https://github.com/sigp/lighthouse/pull/5923#issuecomment-2719236781 , then we should probably change the metric names here too for consistency

beacon_node/lighthouse_network/src/metrics.rs
```diff
@@ -206,6 +206,20 @@ pub static REPORT_PEER_MSGS: LazyLock<Result<IntCounterVec>> = LazyLock::new(||
...
+    try_create_histogram(
+        "self_limiter_request_idling_seconds",
+        "The time our own request remained idle in the self-limiter",
+    )
+});
+
+pub static RESPONSE_LIMITER_RESPONSE_IDLING: LazyLock<Result<Histogram>> = LazyLock::new(|| {
+    try_create_histogram(
```
pawanjay176, line 217: same as above

beacon_node/lighthouse_network/src/rpc/mod.rs
jxs: Hi Akihito! Yeah looks safe now, thanks :)

dapplion (approved):
Approving minus testing by @pawanjay176

jxs (approved):
Thanks for the patience Ahikito, this looks reasonably good to me to be landed and tested!

beacon_node/lighthouse_network/src/rpc/mod.rs
ackintosh: As summarized [here](https://github.com/sigp/lighthouse/pull/5923#issuecomment-2719236781), in this PR this limiter is limiting our responses, so I think `response_limiter` is a more intuitive name than one based on `inbound` to maintain consistency with what this limiter does. 

> to also still follow the naming of the configurations

Oh, you're right, there is a naming conflict between the limiter and the configuration... 🤔 : 

https://github.com/sigp/lighthouse/blob/753e8457610009e631f413582352085e07bc7f98/beacon_node/lighthouse_network/src/rpc/mod.rs#L181-L185

pawanjay176 (approved):
This looks good to me as well. Just minor nits.
Been running this for sometime on mainnet with hammering a synced node for requests and it holds up as expected.
Great work on this one, sorry it took so long for review and testing.

beacon_node/lighthouse_network/src/metrics.rs
ackintosh: Thanks for the feedback! I have pushed a [commit](https://github.com/ackintosh/lighthouse/commit/11468d8e0e9739e512a4ab63ba34eb49b7a581b5) to update the metric names, but the commit is not showing up in this PR. I'm not sure why; maybe waiting a bit will resolve this.

ackintosh:
@pawanjay176  Sorry for the delay.I've addressed your feedback.

pawanjay176 (approved):
Awesome, let's get this in!

beacon_node/lighthouse_network/src/rpc/self_limiter.rs
```diff
@@ -13,30 +20,31 @@ use tokio_util::time::DelayQueue;
...
+
+#[allow(clippy::type_complexity)]
 pub(crate) struct SelfRateLimiter<Id: ReqId, E: EthSpec> {
-    /// Requests queued for sending per peer. This requests are stored when the self rate
+    /// Active requests that are awaiting a response.
+    active_requests: HashMap<PeerId, HashMap<Protocol, usize>>,
+    /// Requests queued for sending per peer. These requests are stored when the self rate
     /// limiter rejects them. Rate limiting is based on a Peer and Protocol basis, therefore
     /// are stored in the same way.
     delayed_requests: HashMap<(PeerId, Protocol), VecDeque<QueuedRequest<Id, E>>>,
```
dapplion, line 41: Would be great to have these additional metrics to debug PeerDAS

- Counter: Total count of queue requests labeled by `Protocol` string
  - Same for rate_limiter delayed_responses
- Histogram:  `VecDeque` length at the moment of queuing a request, labeled by `Protocol` string = observe length before pushing
  - Same for rate_limiter delayed_responses
- Gauge: Current sum of `VecDeque` lengths for all items in `delayed_requests`

PR #5926, opened 2024-06-13
update kurtosis repo location

realbigsean:
## Issue Addressed

EF devops is taking over the kurtosis ethereum package: https://github.com/ethpandaops/ethereum-package

updating the references i found

PR #5927, opened 2024-06-13
Rust 1.79 lints

realbigsean:
## Issue Addressed

fixing 1.79 lints including:
- using `u64::MAX` instead of `u64::max_value()` or `std::u64::MAX`
- some errors trigger a dead code lint when the field isn't read, but errors as fields are useful when printing via debug, so I just added `#[allow(dead_code)]` to these errors
- removed `spec: WatchSpec<E>` from `UpdateHandler` as this was unused - wasn't sure if this field was intentionally there and should be kept?

### DIFF
--- a/watch/src/blockprint/database.rs
+++ b/watch/src/blockprint/database.rs
@@ -33,6 +33,7 @@ pub struct WatchBlockprint {
 }
 
 #[derive(Debug, QueryableByName, diesel::FromSqlRow)]
+#[allow(dead_code)]
 pub struct WatchValidatorBlockprint {
     #[diesel(sql_type = Integer)]
     pub proposer_index: i32,
--- a/watch/src/updater/handler.rs
+++ b/watch/src/updater/handler.rs
@@ -9,6 +9,7 @@ use eth2::{
 };
 use log::{debug, error, info, warn};
 use std::collections::HashSet;
+use std::marker::PhantomData;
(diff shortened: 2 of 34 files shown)


watch/src/updater/handler.rs
```diff
@@ -47,7 +48,7 @@ pub struct UpdateHandler<E: EthSpec> {
     pub blockprint: Option<WatchBlockprintClient>,
     pub config: Config,
     pub slots_per_epoch: u64,
-    pub spec: WatchSpec<E>,
+    pub _phantom: PhantomData<E>,
```
realbigsean, line 51: Should this generic just be removed instead?

watch/src/blockprint/database.rs
```diff
@@ -33,6 +33,7 @@ pub struct WatchBlockprint {
 }
 
 #[derive(Debug, QueryableByName, diesel::FromSqlRow)]
+#[allow(dead_code)]
 pub struct WatchValidatorBlockprint {
```
realbigsean, line 37: should this be deleted? i don't see it used but not sure if it's a useful export if anything depends on beacon watch?

realbigsean:
Also I need to cherry pick https://github.com/sigp/lighthouse/pull/5926 in to get CI to pass, so I'm going to close that PR

michaelsproul (approved):
LGTM

Let's merge for now and we can cleanup beaconwatch later if we like

michaelsproul:
@mergifyio queue

PR #5928, opened 2024-06-14
Remove some easy Electra TODOs

michaelsproul:
## Proposed Changes

Remove 3 TODOs related to Electra:

- Delete legacy payload reconstruction for Electra. We don't even have _proper_ payload reconstruction specced yet for Electra (see https://github.com/sigp/lighthouse/pull/5743), and I'm doubtful we will need or want the legacy reconstruction. I'm actually tempted to delete legacy payload reconstruction entirely, but it _is_ a little bit useful for sanity-checking the EL genesis block and preventing mistakes similar to Holesky v1.0 (see `bellatrix_readiness.rs` which does genesis block checks).
- Confirm that non-optional fields are OK in `PartialBeaconState`. This is going away soon with tree-states anyway.
- Remove comment from `ExecutionPayloadHeader::ssz_max_var_len_for_fork`. The extra data is still the only variable-length field in `ExecutionPayloadHeader`, so this function is still correct. If the spec changes, the light client tests should catch any discrepancy here too.

### DIFF
--- a/beacon_node/execution_layer/src/lib.rs
+++ b/beacon_node/execution_layer/src/lib.rs
@@ -1985,39 +1985,10 @@ impl<E: EthSpec> ExecutionLayer<E> {
                     excess_blob_gas: deneb_block.excess_blob_gas,
                 })
             }
-            ExecutionBlockWithTransactions::Electra(electra_block) => {
-                let withdrawals = VariableList::new(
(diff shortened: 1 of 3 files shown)


dapplion (approved):

realbigsean:
@mergify queue

PR #5929, opened 2024-06-14
Clean up Electra observed aggregates

michaelsproul:
## Proposed Changes

Fix some bugs and TODOs related to the observed aggregates cache.

1. Ensure that attestation roots are always calculated using the committee index, i.e. `ObservedAttestationKey`. Previously we were ok because we were using the key consistently in `attestation_verification.rs`, but if the code were refactored to depend on `SubsetItem::root`, it would break due to the `root` method ignoring the committee index. This would lead to false positives for subset aggregates, as aggregates would be compared for different committees.
2. Remove `expect`s and `unwrap`s by adding a few new error variants. These should be hard to reach, but not necessarily impossible, so I think an error is better than an `expect`, or an attempt to make the methods infallible.

### DIFF
--- a/beacon_node/beacon_chain/src/attestation_verification.rs
+++ b/beacon_node/beacon_chain/src/attestation_verification.rs
@@ -35,8 +35,10 @@
 mod batch;
 
 use crate::{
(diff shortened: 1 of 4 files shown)


eserilev (approved):
Looks good to me!

realbigsean (approved):

PR #5930, opened 2024-06-14
cleanup RPC  self limiter

jxs:


beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -395,6 +395,13 @@ where
                     connection: CloseConnection::All,
                 });
             }
+            HandlerEvent::Err(err) => {
```
pawanjay176, line 398: But the event being generated in the case of us disconnecting from the peer from a `prune_peers` routine is 
`HandlerEvent::Close(RPCError::Disconnected)` right?

jxs: when `RPCHandler.shutdown()` is called it bubbles up to the `RPC` the pending requests in:
https://github.com/sigp/lighthouse/blob/f1d88ba4b1256e084d3a2a547e0ce574b896246c/beacon_node/lighthouse_network/src/rpc/handler.rs#L255-L262

which is then received by `RPC` but not bubbled up to the `Swarm`. (which is what this PR addresses)

It is only when `RPCHandler` has finishing shutting down reports the `HandlerEvent::Close(RPCError::Disconnected`

PR #5933, opened 2024-06-17
Update CI nethermind and LLVM versions

jimmygchen:
## Issue Addressed

Update nethermind and LLVM versions.

- Rust 1.79 [updates the minimum external LLVM to 17](https://github.com/rust-lang/rust/blob/1.79.0/RELEASES.md#compatibility-notes). Note we no longer use GitHub hosted runners for release and I think our windows runners use latest version. (FYI @antondlr in case if we run into any issues)
- Update Nethermind version to latest release.


jimmygchen:
Requires installing dotnet8 on the self-hosted runners.

michaelsproul (approved):
LGTM!

jimmygchen:
@mergify queue

jimmygchen:
@mergify requeue

PR #5935, opened 2024-06-17
Lower tolerance of stale blobs on gossip

michaelsproul:
## Proposed Changes

Some peers are spamming old blobs from prior to finalization on gossip. We don't ban them quickly enough due to the high-tolerance peer penalty. This PR changes the penalty to a low-tolerance one.

From the peers that I spot-checked, all the spam seemed to be coming from Erigon/Caplin peers. I've communicated this to the Caplin devs, so new releases of Caplin shouldn't have this problem and won't suffer from by being banned by Lighthouse.

I've kept the gossipsub behaviour as `Ignore` rather than combining with the `Reject` block above. We are required to `Ignore` these blobs by the spec: https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/p2p-interface.md#blob-subnets

### DIFF
--- a/beacon_node/network/src/network_beacon_processor/gossip_methods.rs
+++ b/beacon_node/network/src/network_beacon_processor/gossip_methods.rs
@@ -713,9 +713,7 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                             MessageAcceptance::Reject,
                         );
                     }
-                    GossipBlobError::FutureSlot { .. }
-                    | GossipBlobError::RepeatBlob { .. }
-                    | GossipBlobError::PastFinalizedSlot { .. } => {
+                    GossipBlobError::FutureSlot { .. } | GossipBlobError::RepeatBlob { .. } => {
                         warn!(
                             self.log,
                             "Could not verify blob sidecar for gossip. Ignoring the blob sidecar";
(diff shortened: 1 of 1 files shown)


AgeManning (approved):
lgtm

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -737,6 +735,28 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                        warn!(
+                            self.log,
+                            "Could not verify blob sidecar for gossip. Ignoring the blob sidecar";
+                            "error" => ?err,
+                            "slot" => %slot,
+                            "root" => %root,
+                            "index" => %index,
+                            "commitment" => %commitment,
+                        );
+                        // Prevent recurring behaviour by penalizing the peer slightly.
```
dapplion: not a slight penalty

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -737,6 +735,28 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                             MessageAcceptance::Ignore,
                         );
                     }
+                    GossipBlobError::PastFinalizedSlot { .. } => {
+                        warn!(
```
dapplion: This log is not actionable by the user, I should not be a warn.

michaelsproul: This applies to all the warns in this file

michaelsproul: We could change them to debug?

I think "user-actionable" is a high bar though. Not even all our ERROs and CRITs are user actionable

michaelsproul: Kept the warning for the more serious/rare blob verification errors. Like warning for invalid blocks, these are potentially interesting as a canary for network issues

michaelsproul: If we want to change that behaviour, we can do it in a separate PR. AFAIK the warning that's left has never (?) tripped on mainnet

dapplion (approved):

jimmygchen:
@mergify queue

jimmygchen:
@mergify requeue

jimmygchen:
@mergify requeue

jimmygchen:
@mergify unqueue

jimmygchen:
@mergify requeue

PR #5938, opened 2024-06-17
chore: add `recover_cells_and_compute_proofs` method

kevaundray:
## Issue Addressed

This PR makes the KZG API stricter. Instead of exposing recover_cells, and cells_to_blob, we have a method which recovers the cells and computes the proofs.

## Proposed Changes

Please list or describe the changes introduced by this PR.

## Additional Info

Please provide any additional information. For example, future considerations
or information useful for reviewers.

PR #5941, opened 2024-06-17
chore: add rust PeerdasKZG crypto library for peerdas functionality and rollback c-kzg dependency to 4844 version

kevaundray:
## Issue Addressed

This relies on #5938 

Below was copied from Dapplion:

Description: Replace c-kzg peerdas crypto with new lib peerdas-kzg
Rationale: Do not touch mainnet paths, and use this lib only for das codepaths
Justification: Eventually this lib should become canonical, and we can add a switch later when ckzg peerdas merges to main

## Proposed Changes

## Additional Info


beacon_node/beacon_chain/src/kzg_utils.rs
```diff
@@ -11,8 +11,11 @@ fn ssz_blob_to_crypto_blob<E: EthSpec>(blob: &Blob<E>) -> Result<KzgBlob, KzgErr
 
 /// Converts a cell ssz List object to an array to be used with the kzg
 /// crypto library.
-fn ssz_cell_to_crypto_cell<E: EthSpec>(cell: &Cell<E>) -> Result<KzgCell, KzgError> {
-    KzgCell::from_bytes(cell.as_ref()).map_err(Into::into)
+fn ssz_cell_to_crypto_cell<E: EthSpec>(cell: &Cell<E>) -> Result<KzgCellRef, KzgError> {
```
kevaundray, line 14: Noting that this method is duplicated in data_column_sidecar.rs

crypto/kzg/src/lib.rs
```diff
@@ -156,84 +166,92 @@ impl Kzg {
...
+            Err(e) => Err(Error::VerifierKZG(e)),
+        }
     }
 
-    pub fn recover_all_cells(
+    pub fn recover_cells_and_compute_kzg_proofs<'a>(
         &self,
         cell_ids: &[u64],
-        cells: &[Cell],
-    ) -> Result<Box<[Cell; c_kzg::CELLS_PER_EXT_BLOB]>, Error> {
```
kevaundray: I don't believe the way Box was being used before is guaranteed to always allocate on the heap, there is a tracking issue here for it: https://github.com/rust-lang/rust/issues/27779#issuecomment-378416911

Its unclear to me in what cases the compiler will be smart enough to allocate on the heap directly, so this seems like it may end up with sporadic stackoverflows

kevaundray: The crypto lib tries to mitigate this by taking Cell as Vec<u8>, then Boxing it, and turning the Boxed Vec<u8> into a Boxed [u8;N]

This looks to be reliable, though I can see an argument for returning Vec<u8> -- you just lose type safety

testing/ef_tests/src/cases/kzg_verify_blob_kzg_proof.rs
```diff
@@ -40,8 +40,9 @@ pub fn parse_cell(cell: &str) -> Result<Cell, Error> {
     hex::decode(strip_0x(cell)?)
         .map_err(|e| Error::FailedToParseTest(format!("Failed to parse cell: {:?}", e)))
         .and_then(|bytes| {
-            Cell::from_bytes(bytes.as_ref())
-                .map_err(|e| Error::FailedToParseTest(format!("Failed to parse proof: {:?}", e)))
+            bytes
+                .try_into()
+                .map_err(|e| Error::FailedToParseTest(format!("Failed to parse cell: {:?}", e)))
```
kevaundray, line 45: small typo

crypto/kzg/src/lib.rs
```diff
@@ -156,84 +166,92 @@ impl Kzg {
...
-            cell_ids,
-            cells,
-            &self.trusted_setup,
-        )?)
+        cells: &[CellRef<'a>],
+    ) -> Result<([Cell; CELLS_PER_EXT_BLOB], [KzgProof; CELLS_PER_EXT_BLOB]), Error> {
+        let (cells, proofs) = self
+            .context
+            .prover_ctx()
+            .recover_cells_and_proofs(cell_ids.to_vec(), cells.to_vec(), vec![])
```
kevaundray: We can improve on the crypto API here, and take &[Cell] instead of Vec<Cell>, same for other parameters.

proofs is left as an empty parameter because the underlying library does not use it, it just recomputes every proof each time

crypto/kzg/src/lib.rs
```diff
@@ -50,6 +59,7 @@ impl Kzg {
                 // Ref: https://notes.ethereum.org/@jtraglia/windowed_multiplications
                 8,
             )?,
+            context: PeerDASContext::default(),
```
kevaundray: This is doing quite a few things, which we could also pull out (roughly):

```rust
let trusted_setup = TrustedSetup::default()
let prover_context = ProverContext::default()
let verifier_context = VerifierContext::default()
let num_threads = 16
PeerDASContext{trusted_setup, num_threads, prover_context, verifier_context} 
```
You can also initialize a trusted setup from the json string in the consensus-specs by doing `TrustedSetup::from_json(&str)` -- This is what `TrustedSetup::default()` is doing

kevaundray: Closing as I've added the more explicit API

crypto/kzg/src/lib.rs
```diff
@@ -36,20 +44,32 @@ impl From<c_kzg::Error> for Error {
...
+
+        // Initialize the trusted setup using default parameters
+        // 
+        // Note: One can also use `from_json` to initialize it from the consensus-specs
+        // json string.
+        let peerdas_trusted_setup = PeerDASTrustedSetup::default();
+        // Set the number of threads to be used
+        //
+        // we set it to 1 to match the c-kzg performance
+        const NUM_THREADS : usize = 1;
```
kevaundray: This can be set to something sensible in another PR

crypto/kzg/src/lib.rs
```diff
@@ -36,20 +44,32 @@ impl From<c_kzg::Error> for Error {
...
+
+        let context = PeerDASContext::with_threads(&peerdas_trusted_setup, NUM_THREADS);
+
         Ok(Self {
             trusted_setup: KzgSettings::load_trusted_setup(
                 &trusted_setup.g1_points(),
                 &trusted_setup.g2_points(),
-                // Enable precomputed table for 8 bits, with 96MB of memory overhead per process
-                // Ref: https://notes.ethereum.org/@jtraglia/windowed_multiplications
-                8,
```
kevaundray, line 56: This parameter does not exist on the c-kzg 4844 auditted branch -- since I've reverted c-kzg back to v1, it has been removed

crypto/kzg/src/lib.rs
```diff
@@ -13,13 +13,21 @@ pub use c_kzg::{
     Blob, Bytes32, Bytes48, KzgSettings, BYTES_PER_BLOB, BYTES_PER_COMMITMENT,
     BYTES_PER_FIELD_ELEMENT, BYTES_PER_PROOF, FIELD_ELEMENTS_PER_BLOB,
 };
-pub use c_kzg::{Cell, CELLS_PER_EXT_BLOB};
```
kevaundray, line 16: Note: the only type which I've changed/removed is the Cell type -- The Blob types etc that are needed for 4844 are left unchanged along with their codepaths. 

We convert from c-kzg Blob type to PeerDASKZG blob type inside of this file.

This is just so that we can be sure that the old c-kzg code has been left untouched.

Cargo.toml
```diff
@@ -92,16 +92,14 @@ bytes = "1"
...
-# feature ourselves when desired.
-c-kzg = { git = "https://github.com/ethereum/c-kzg-4844", rev = "114fa0382990e9b74b1f90f3b0dc5f97c2f8a7ad" }
+c-kzg = { version = "1", default-features =  false }
 compare_fields_derive = { path = "common/compare_fields_derive" }
 criterion = "0.3"
 delay_map = "0.3"
 derivative = "2"
 dirs = "3"
 either = "1.9"
+eip7594 = { git = "https://github.com/crate-crypto/peerdas-kzg", rev = "5a44c5995ade5de0ac06909bcbacab8437847ba4" }
```
dapplion: Why not name the lib peerdas-kzg?

crypto/kzg/src/lib.rs
```diff
@@ -13,13 +13,22 @@ pub use c_kzg::{
...
 use mockall::automock;
 
+pub use eip7594::{
+    constants::{BYTES_PER_CELL, CELLS_PER_EXT_BLOB},
+    Cell, CellID, CellRef, TrustedSetup as PeerDASTrustedSetup,
```
dapplion, line 20: Will we have to bake 2 trusted setups for now? AFAIR ckzg peerdas needed a different representation of its points

crypto/kzg/src/lib.rs
```diff
@@ -36,20 +45,31 @@ impl From<c_kzg::Error> for Error {
...
     pub fn new_from_trusted_setup(trusted_setup: TrustedSetup) -> Result<Self, Error> {
+        // Initialize the trusted setup using default parameters
+        //
+        // Note: One can also use `from_json` to initialize it from the consensus-specs
+        // json string.
+        let peerdas_trusted_setup = PeerDASTrustedSetup::default();
+        // Set the number of threads to be used
+        //
+        // we set it to 1 to match the c-kzg performance
+        const NUM_THREADS: usize = 1;
```
dapplion, line 64: What threading does this constant activate? We are already parallelizing on a per-blob basis, do we need more?

Cargo.toml
kevaundray: The name is going to change in the future, it is called eip7594 for historical reasons -- the "finalized" name will include the word "eth", likely "rust-eth-kzg"

crypto/kzg/src/lib.rs
kevaundray: Yep thats correct, peerdasKZG will accept the official trusted setup json file from the consensus-specs

crypto/kzg/src/lib.rs
kevaundray: This is threading internal to the library, ie there are parts that you can multithread inside of the library when computing proofs for example. You can leave this at 1, if you think its not beneficial

jimmygchen:
Thanks @kevaundray! Are there any benchmarks for the `peerdas-kzg` crate?

Cargo.toml
jimmygchen: Perhaps we could rename it here for clarity within the Lighthouse repo?

```suggestion
peerdas-kzg = { git = "https://github.com/crate-crypto/peerdas-kzg", rev = "5a44c5995ade5de0ac06909bcbacab8437847ba4", package = "eip7594" }
```

kevaundray:
> Thanks @kevaundray! Are there any benchmarks for the `peerdas-kzg` crate?

Yep, I added some a few hours ago -- you can `cd eip7594` and `cargo bench` -- For a single thread, it should have around the same performance as c-kzg -- for computing cells and proofs, I get ~2x when I set NUM_THREADS to 8. I expect this number to be driven down further when more components are parallelized

crypto/kzg/src/lib.rs
```diff
@@ -41,20 +46,31 @@ impl From<c_kzg::Error> for Error {
...
 
 #[automock]
 impl Kzg {
     /// Load the kzg trusted setup parameters from a vec of G1 and G2 points.
     pub fn new_from_trusted_setup(trusted_setup: TrustedSetup) -> Result<Self, Error> {
+        // Initialize the trusted setup using default parameters
+        //
+        // Note: One can also use `from_json` to initialize it from the consensus-specs
+        // json string.
+        let peerdas_trusted_setup = PeerDASTrustedSetup::default();
```
dapplion, line 60: Note to self: `default` loads the JSON setup from the consensus-specs. What's the binary size increase of adding this JSON file? The file in JSON form is 861 KB. Aesthetics note: @kevaundray maybe you should use a more descriptive name?

kevaundray: It gets compressed using libflate with a compression ratio of about 2, so its around 400KB (closer to 490KB using zlib) on disk. We can put it under a feature flag if its an issue.

The `default` method name is because it implements the `default` trait. We could use a user defined method for clarity, though not a blocker I think since Default::default is pretty common

dapplion (approved):
LGTM!

jimmygchen:
Nice! Thanks a lot for this @kevaundray 

There are a few things we need to do before merging this, as `das` is the active branch currently being tested on `peerdas-devnet-1`:
- Investigate and fix CI failures
- I'd love to see some local devnet testing before we merge this in (I'm using the config from #5968)
- It'd be nice to make this either runtime configurable or behind a feature gate, so we can easily switch between libs without blocking devnet testing in case we run into bugs / perf issues.

If you're busy I'd be happy to keep this moving in the next week, also happy to do some testing.

kevaundray:
> Investigate and fix CI failures

Seems my change locally didn't get pushed wrt to renaming the library in the workspace, sorry about that. Fixed the other linter complaints.

> It'd be nice to make this either runtime configurable or behind a feature gate, so we can easily switch between libs without blocking devnet testing in case we run into bugs / perf issues.

Note that this means that we would need to point c-kzg back to the das branch, whereas this PR reverts it to its 4844 status so that no 4844 code is affected. Is there a way to check if there are performance issues before merging? 

Re bugs, the cryptography libraries have their own set of test vectors, so I'd be suprised if a bug is encountered and the test vectors still pass unless the test vectors are lacking (which is also possible)

kevaundray:
> I'd love to see some local devnet testing before we merge this in (I'm using the config from https://github.com/sigp/lighthouse/pull/5968)

I ran the local devnet script in that PR you linked. I looked at the logs and see no errors except for some sporadic handshake failures (not familiar with the p2p layer, so not sure if this is normal, though it shouldn't be related to this PR I think)

Also see: `[el-1-geth-lighthouse] ERROR[...] Nil finalized block cannot evict old blobs`

Were you looking for something in particular wrt to testing it locally?

kevaundray:
Although we don't use the proofs that get passed to `recover_cells_and_kzg_proofs`, we need to pass it to the library as a parameter and perform redundant checks (This should eventually be removed, the consensus-specs PR for it is https://github.com/ethereum/consensus-specs/pull/3819)

kevaundray:
Unclear to me why this is passing on the DAS branch since we also don't pass proofs into there and its using the old interface where you recover_cells, then covert cells to blob, then compute proofs.

In order to make it work, we would need `reconstruct` to also pass along the proof for each cell to recover_cells_and_proofs.


EDIT: It passes on the DAS branch because it is not using the most updated version of c-kzg which has all of the extra checks on the input proof

kevaundray:
I've pulled out the API change for `recover_cells_and_proofs` to #5998

kevaundray:
@jimmygchen I've ran the ef-tests locally and `make lint` -- will leave local devnet testing and other blockers in your hand

jimmygchen:
Sure, I'll do the testing, thanks a lot for this :relaxed: 

Sorry about the delay in responding! and just to answer your questions:

> Is there a way to check if there are performance issues before merging?

Yep we do have some kzg metrics exposed via the `{host}:{metrics_port}/metrics` endpoint, and may be useful to compare them:
```
> curl -s $(kurtosis port print local-testnet cl-2-lighthouse-geth metrics)/metrics | grep -E "HELP.+column"
# HELP beacon_blobs_column_sidecar_gossip_verification_seconds Full runtime of data column sidecars gossip verification
# HELP beacon_blobs_column_sidecar_processing_requests_total Count of all data column sidecars submitted for processing
# HELP beacon_blobs_column_sidecar_processing_successes_total Number of data column sidecars verified for gossip
# HELP beacon_data_column_gossip_propagation_verification_delay_time Duration between when the data column sidecar is received over gossip and when it is verified for propagation.
# HELP beacon_data_column_gossip_slot_start_delay_time Duration between when the data column sidecar is received over gossip and the start of the slot it belongs to.
# HELP beacon_data_column_last_delay Keeps track of the last data column sidecar's delay from the start of the slot
# HELP beacon_processor_gossip_data_column_queue_total Count of data column sidecars from gossip waiting to be verified.
# HELP beacon_processor_gossip_data_column_verified_total Total number of gossip data column sidecar verified for propagation.
# HELP beacon_processor_rpc_custody_column_queue_total Count of custody columns from the rpc waiting to be imported.
# HELP beacon_processor_rpc_verify_data_column_queue_total Count of data columns from the rpc waiting to be verified.
# HELP data_availability_reconstructed_columns_total Total count of reconstructed columns
# HELP data_availability_reconstruction_time_seconds Time taken to reconstruct columns
# HELP data_column_sidecar_computation_seconds Time taken to compute data column sidecar, including cells, proofs and inclusion proof
# HELP data_column_sidecar_inclusion_proof_verification_seconds Time taken to verify data_column sidecar inclusion proof
# HELP kzg_verification_data_column_single_seconds Runtime of single data column kzg verification
```

> Were you looking for something in particular wrt to testing it locally?

I think the main thing to look for when testing peerdas is making sure the network continue to produce blocks and nodes are in sync with each other - this means columns are being produced and propagated and likely works end-to-end. And if we want to make sure it interops with c-kzg, it might be worth running the `network-params-das-interop.yaml` one (with prysm and teku).

kevaundray:
> Sure, I'll do the testing, thanks a lot for this :relaxed: 
> 
> 
> 
> Sorry about the delay in responding! and just to answer your questions:
> 
> 
> 
> > Is there a way to check if there are performance issues before merging?
> 
> 
> 
> Yep we do have some kzg metrics exposed via the `{host}:{metrics_port}/metrics` endpoint, and may be useful to compare them:
> 
> ```
> 
> > curl -s $(kurtosis port print local-testnet cl-2-lighthouse-geth metrics)/metrics | grep -E "HELP.+column"
> 
> # HELP beacon_blobs_column_sidecar_gossip_verification_seconds Full runtime of data column sidecars gossip verification
> 
> # HELP beacon_blobs_column_sidecar_processing_requests_total Count of all data column sidecars submitted for processing
> 
> # HELP beacon_blobs_column_sidecar_processing_successes_total Number of data column sidecars verified for gossip
> 
> # HELP beacon_data_column_gossip_propagation_verification_delay_time Duration between when the data column sidecar is received over gossip and when it is verified for propagation.
> 
> # HELP beacon_data_column_gossip_slot_start_delay_time Duration between when the data column sidecar is received over gossip and the start of the slot it belongs to.
> 
> # HELP beacon_data_column_last_delay Keeps track of the last data column sidecar's delay from the start of the slot
> 
> # HELP beacon_processor_gossip_data_column_queue_total Count of data column sidecars from gossip waiting to be verified.
> 
> # HELP beacon_processor_gossip_data_column_verified_total Total number of gossip data column sidecar verified for propagation.
> 
> # HELP beacon_processor_rpc_custody_column_queue_total Count of custody columns from the rpc waiting to be imported.
> 
> # HELP beacon_processor_rpc_verify_data_column_queue_total Count of data columns from the rpc waiting to be verified.
> 
> # HELP data_availability_reconstructed_columns_total Total count of reconstructed columns
> 
> # HELP data_availability_reconstruction_time_seconds Time taken to reconstruct columns
> 
> # HELP data_column_sidecar_computation_seconds Time taken to compute data column sidecar, including cells, proofs and inclusion proof
> 
> # HELP data_column_sidecar_inclusion_proof_verification_seconds Time taken to verify data_column sidecar inclusion proof
> 
> # HELP kzg_verification_data_column_single_seconds Runtime of single data column kzg verification
> 
> ```
> 
> 
> 
> > Were you looking for something in particular wrt to testing it locally?
> 
> 
> 
> I think the main thing to look for when testing peerdas is making sure the network continue to produce blocks and nodes are in sync with each other - this means columns are being produced and propagated and likely works end-to-end. And if we want to make sure it interops with c-kzg, it might be worth running the `network-params-das-interop.yaml` one (with prysm and teku).
> 
> 
> 
> 

No worries! I can see you have your hands full :) 

I'm working on integrations for a few other repos, so I'll probably have to circle back to collecting information next week on local devnet. This information is useful in general so if you get to it before me, feel free to post it here!

beacon_node/beacon_chain/src/kzg_utils.rs
jimmygchen: Cool, we can probably remove this one

beacon_node/beacon_chain/src/kzg_utils.rs
```diff
@@ -11,8 +11,11 @@ fn ssz_blob_to_crypto_blob<E: EthSpec>(blob: &Blob<E>) -> Result<KzgBlob, KzgErr
...
+    let cell_bytes: &[u8] = cell.as_ref();
+    Ok(cell_bytes
+        .try_into()
+        .expect("expected cell to have size {BYTES_PER_CELL}. This should be guaranteed by the `FixedVector type"))
```
jimmygchen, line 18:
```suggestion
        .expect("expected cell to have size {BYTES_PER_CELL}. This should be guaranteed by the `FixedVector` type"))
```

jimmygchen:
Hi @kevaundray, 

Apologies on the delay getting to this. I was out the last couple of days.

I did a quick round of testing locally and it seems to be fine with a Lighthouse only network, but when I ran it with Prysm and Teku, the network forked pretty quickly.

I'll look into it again tomorrow, maybe I'm missing something in my network config.

kevaundray:
> Hi @kevaundray, 
> 
> 
> 
> Apologies on the delay getting to this. I was out the last couple of days.
> 
> 
> 
> I did a quick round of testing locally and it seems to be fine with a Lighthouse only network, but when I ran it with Prysm and Teku, the network forked pretty quickly.
> 
> 
> 
> I'll look into it again tomorrow, maybe I'm missing something in my network config.
> 
> 

No worries :) 

I'm guessing the DAS branch without this change is okay?

jimmygchen:
I’m not entirely sure tbh, didn’t get too much time on this today. I didn’t see any KZG errors so I suspect it may be related to mismatch das network defaults for different clients and we may have to override them in the Kurtosis config, I’ll try again tomorrow.

jimmygchen:
I've tested again, looking pretty good after I added network config overrides, so I guess clients currently have different default values:

```
participants:
  - cl_type: prysm
    cl_image: ethpandaops/prysm-beacon-chain:peerDAS

  - cl_type: lighthouse
    cl_extra_params: [
      --subscribe-all-data-column-subnets,
    ]
    cl_image: lighthouse:local

  - cl_type: lighthouse
    cl_image: lighthouse:local

  - cl_type: teku
    cl_image: ethpandaops/teku:nashatyrev-das

#  - cl_type: nimbus
#    cl_image: ethpandaops/nimbus-eth2:kzgpeerdas
#
#  - cl_type: grandine
#    cl_image: ethpandaops/grandine:das
#
#   - cl_type: lodestar
#     cl_image: ethpandaops/lodestar:peerDAS
network_params:
  eip7594_fork_epoch: 0
  eip7594_fork_version: "0x50000038"
  data_column_sidecar_subnset_count: 64
  samples_per_slot: 16
  custody_requirement: 4
  target_number_of_peers: 70
snooper_enabled: false
global_log_level: debug
additional_services:
  - dora
  - goomy_blob
```

I'll post some metrics comparison soon

jimmygchen:
Metrics running this branch, looks pretty good and comparable to `das` branch. 

This branch (over 1 h):
image

`das` branch (over 5 m):
image

jimmygchen:
@dapplion I've tested the branch and it works pretty well! 

I've attached some metrics in the above comment. I've ran out of time to review the new changes, but feel free to merge if you're happy with it, otherwise I'll re-review it when I have a chance.

Thanks a lot for this @kevaundray ! 🙏

PR #5946, opened 2024-06-18
chore: simplify method to generate a random valid blob

kevaundray:
## Issue Addressed

The previous method was indexing the blob_bytes on `i * BYTES_PER_FIELD_ELEMENT` -- this method iterates the blob_bytes directly and skips to every `BYTES_PER_FIELD_ELEMENT` so we can avoid the failure modes:

- Index out of bounds is no longer possible since the iterator will just stop iterating at that point
- Checked mul can also be avoided 

## Proposed Changes

## Additional Info


dapplion (approved):

jimmygchen:
@mergify queue

jimmygchen:
@mergify requeue

jimmygchen:
@mergify requeue

jimmygchen:
@mergify unqueue

jimmygchen:
@mergify requeue

PR #5940, opened 2024-06-17
Electra attestations #5712 review

dapplion:
## Issue Addressed

Review fixes to
- https://github.com/sigp/lighthouse/pull/5712

45% done looking at the code but got enough changes for a PR.

A misc collection of unrelated changes, I have made all commits atomics. Let me know if you disagree with some of the changes I can drop them

## Proposed Changes

- [De-dup attestation constructor logic](https://github.com/sigp/lighthouse/commit/eaed2929c164ad6bea5c079592cf107378621a9b)
- [Remove unwraps in Attestation construction](https://github.com/sigp/lighthouse/commit/8473e3a7772fb043b46aa7e13d15da6b2f0c8ee7)
- [Dedup match_attestation_data](https://github.com/sigp/lighthouse/commit/38bc5d5a5f9b73c9a8e7c4bfd6603056784f3a58)
- [Remove outdated TODO](https://github.com/sigp/lighthouse/commit/a375a6271d0758557319a3d46f71fcd149ba04bc)
- [Use ForkName Ord in fork-choice tests](https://github.com/sigp/lighthouse/commit/31057ebd95f851c7cf574fee43a674562cdb6aaa)
- [Use ForkName Ord in BeaconBlockBody](https://github.com/sigp/lighthouse/commit/2d11fc452e5bdc0babd7ab53101c757063f9bc7d)
- [Make to_electra not fallible](https://github.com/sigp/lighthouse/commit/67ea6c679462e77ac0afbe950d33745dac5666e2)
- [Remove TestRandom impl for IndexedAttestation](https://github.com/sigp/lighthouse/commit/b123b41b9d068e500b2667f3a5c0c36eb218fa1f)
- [Remove IndexedAttestation faulty Decode impl](https://github.com/sigp/lighthouse/commit/2e6712590f6ab997bf85f87e0dc43a99f650839f)
- [Drop TestRandom impl](https://github.com/sigp/lighthouse/commit/ed65a6a875e0b3001aa635edc654efff2533d59a)
- [Add PendingAttestationInElectra](https://github.com/sigp/lighthouse/commit/ae15ff3de288eb64dee54cd2cc1de54004602bb0)

## Additional Info

Please provide any additional information. For example, future considerations
or information useful for reviewers.

PR #5951, opened 2024-06-18
Add {fork_name}_enabled functions

eserilev:
Add  `{fork_name}_enabled` functions to the `ForkName` impl and use them in place of `fork_name >= ForkName::{fork_name_variant}` checks and in places where we can remove `match` boilerplate code. 

There was some discussion [here](https://github.com/sigp/lighthouse/pull/5940#discussion_r1644693373) about implementing something like this to help reduce chances of developer error.


beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -2574,10 +2574,11 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
         let current_fork = self.spec.fork_name_at_slot::<T::EthSpec>(self.slot()?);
-        if let ForkName::Base | ForkName::Altair | ForkName::Bellatrix = current_fork {
-            Ok(false)
-        } else {
+
+        if current_fork.capella_enabled() {
             Ok(true)
+        } else {
+            Ok(false)
         }
```
dapplion: Just return `Ok(current_fork.capella_enabled())`

dapplion (commented):
Should we name if `{fork_name}_enabled` or `after_{fork_name}`?

eserilev:
im open to either `{fork_name}_enabled` or `after_{fork_name}`, whatever makes the most sense

beacon_node/beacon_chain/src/beacon_chain.rs
eserilev: 3ba2f78

dapplion:
Let's do `{fork_name}_enabled` we already committed to it

realbigsean:
> Let's do `{fork_name}_enabled` we already committed to it

We get autocomplete for the right fork faster this way :)

dapplion (approved):
All logic is equivalent 👍 

This is such a nice cosmetic improvement, no need to sit on this one for longer.

@eserilev please fix conflicts, and good to go

eserilev:
conflicts are resolved. Note that I've added an additional electra related TODO in an ef test here: 
6f94f62

dapplion (approved):
LGTM

jimmygchen:
@eserilev would you mind fixing the conflicts again, and lets get this merged!

jimmygchen (approved):

jimmygchen:
@mergify queue

Issue #5950, opened 2024-06-18
Update custody subnet count format in ENR


jimmygchen:
## Description

See https://github.com/ethereum/consensus-specs/pull/3772

jimmygchen:
Completed in #5966

PR #5966, opened 2024-06-19
Update `csc` format in ENR and spec tests for devnet-1

jimmygchen:
## Issue Addressed

#5950 

- [x] Update `csc` format in ENR.
- [x] Add check for `csc` value in ENR, and fallback to `custody_requirement ` if invalid.
- [x] Update spec tests to v1.5.0-alpha.3. Add test for `recover_cells_and_kzg_proofs`.


jimmygchen:
Merging this as devnet-1 has been launched and das branch is not spec compliant.

PR #5967, opened 2024-06-19
Resend timed out requests in sync

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Resend requests that haven't received a response after a timeout.

PR #5969, opened 2024-06-20
Remove all batches related to a peer on disconnect

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

In https://github.com/sigp/lighthouse/pull/5680 , we decoupled a peer disconnect from an rpc error to simplify the logic in sync a bit. 
The peer disconnect message from the network was responsible for removing the peer from all the sync data structures.
The rpc error was responsible for re-requesting the failed batch/lookup.

We assumed that an rpc error would always go together with a peer disconnect, however we learned the hard way that there are many edge cases where the rpc error never gets generated when the peer disconnects. e.g. the peer manager disconnects a peer that has its rpc response mid-flight, peer disconnects before a rpc request reaches the network etc. There are many other cases we may not know about.
This leads to multiple bugs in sync where sync stalls because a batch/lookup never gets re-requested on peer disconnects.

https://github.com/sigp/lighthouse/pull/5967 was an attempt at adding timeouts to the rpc requests sent from sync that trigger re-requests when the timer elapses. However, as @AgeManning noted, this would lead to many more edge cases.

This PR triggers re-request of batches/lookups on peer disconnects partially reverting #5680 . This isn't the cleanest but considering that we cannot always rely on `lighthouse_network` to generate rpc errors to go with peer disconnects, this seems the best option.

I intend to test this extensively on mainnet before merging

### DIFF
--- a/beacon_node/network/src/sync/backfill_sync/mod.rs
+++ b/beacon_node/network/src/sync/backfill_sync/mod.rs
@@ -320,7 +320,7 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
         }
 
         if let Some(batch_ids) = self.active_requests.remove(peer_id) {
-            // fail the batches
+            // fail the batches.
             for id in batch_ids {
                 if let Some(batch) = self.batches.get_mut(&id) {
                     match batch.download_failed(false) {
@@ -335,7 +335,7 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
                     // If we have run out of peers in which to retry this batch, the backfill state
                     // transitions to a paused state.
                     // We still need to reset the state for all the affected batches, so we should not
-                    // short circuit early
+                    // short circuit early.
                     if self.retry_batch_download(network, id).is_err() {
                         debug!(
                             self.log,
--- a/beacon_node/network/src/sync/backfill_sync/mod.rs
+++ b/beacon_node/network/src/sync/backfill_sync/mod.rs
@@ -307,7 +307,11 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
     /// A peer has disconnected.
     /// If the peer has active batches, those are considered failed and re-requested.
     #[must_use = "A failure here indicates the backfill sync has failed and the global sync state should be updated"]
-    pub fn peer_disconnected(&mut self, peer_id: &PeerId) -> Result<(), BackFillError> {
+    pub fn peer_disconnected(
+        &mut self,
+        peer_id: &PeerId,
+        network: &mut SyncNetworkContext<T>,
+    ) -> Result<(), BackFillError> {
         if matches!(
             self.state(),
             BackFillState::Failed | BackFillState::NotRequired
@@ -315,7 +319,37 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
             return Ok(());
         }
 
-        self.active_requests.remove(peer_id);
+        if let Some(batch_ids) = self.active_requests.remove(peer_id) {
+            // fail the batches
+            for id in batch_ids {
+                if let Some(batch) = self.batches.get_mut(&id) {
+                    match batch.download_failed(false) {
+                        Ok(BatchOperationOutcome::Failed { blacklist: _ }) => {
+                            self.fail_sync(BackFillError::BatchDownloadFailed(id))?;
+                        }
+                        Ok(BatchOperationOutcome::Continue) => {}
+                        Err(e) => {
+                            self.fail_sync(BackFillError::BatchInvalidState(id, e.0))?;
+                        }
+                    }
+                    // If we have run out of peers in which to retry this batch, the backfill state
+                    // transitions to a paused state.
+                    // We still need to reset the state for all the affected batches, so we should not
+                    // short circuit early
+                    if self.retry_batch_download(network, id).is_err() {
+                        debug!(
+                            self.log,
+                            "Batch could not be retried";
+                            "batch_id" => id,
+                            "error" => "no synced peers"
+                        );
+                    }
+                } else {
+                    debug!(self.log, "Batch not found while removing peer";
+                        "peer" => %peer_id, "batch" => id)
+                }
+            }
+        }
 
         // Remove the peer from the participation list
         self.participating_peers.remove(peer_id);
--- a/beacon_node/network/src/sync/block_lookups/mod.rs
+++ b/beacon_node/network/src/sync/block_lookups/mod.rs
@@ -2,7 +2,7 @@ use self::parent_chain::{compute_parent_chains, NodeChain};
 pub use self::single_block_lookup::DownloadResult;
 use self::single_block_lookup::{LookupRequestError, LookupResult, SingleBlockLookup};
 use super::manager::{BlockProcessType, BlockProcessingResult, SLOT_IMPORT_TOLERANCE};
-use super::network_context::{RpcResponseResult, SyncNetworkContext};
+use super::network_context::{RpcResponseError, SyncNetworkContext};
 use crate::metrics;
 use crate::sync::block_lookups::common::ResponseType;
 use crate::sync::block_lookups::parent_chain::find_oldest_fork_ancestor;
@@ -337,7 +337,7 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
         &mut self,
         id: SingleLookupReqId,
         peer_id: PeerId,
-        response: RpcResponseResult<R::VerifiedResponseType>,
+        response: Result<(R::VerifiedResponseType, Duration), RpcResponseError>,
         cx: &mut SyncNetworkContext<T>,
     ) {
         let result = self.on_download_response_inner::<R>(id, peer_id, response, cx);
@@ -349,7 +349,7 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
         &mut self,
         id: SingleLookupReqId,
         peer_id: PeerId,
-        response: RpcResponseResult<R::VerifiedResponseType>,
+        response: Result<(R::VerifiedResponseType, Duration), RpcResponseError>,
         cx: &mut SyncNetworkContext<T>,
     ) -> Result<LookupResult, LookupRequestError> {
         // Note: no need to downscore peers here, already downscored on network context
--- a/beacon_node/network/src/sync/block_lookups/mod.rs
+++ b/beacon_node/network/src/sync/block_lookups/mod.rs
@@ -779,12 +779,12 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
             };
 
             if stuck_lookup.id == ancestor_stuck_lookup.id {
-                warn!(self.log, "Notify the devs, a sync lookup is stuck";
+                warn!(self.log, "Notify the devs a sync lookup is stuck";
                     "block_root" => ?stuck_lookup.block_root(),
                     "lookup" => ?stuck_lookup,
                 );
             } else {
-                warn!(self.log, "Notify the devs, a sync lookup is stuck";
+                warn!(self.log, "Notify the devs a sync lookup is stuck";
                     "block_root" => ?stuck_lookup.block_root(),
                     "lookup" => ?stuck_lookup,
                     "ancestor_block_root" => ?ancestor_stuck_lookup.block_root(),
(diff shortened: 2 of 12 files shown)


dapplion:
For block lookups we should iterate the active requests here and fail them https://github.com/sigp/lighthouse/blob/f1d88ba4b1256e084d3a2a547e0ce574b896246c/beacon_node/network/src/sync/network_context.rs#L120-L123

Plus add tests to ensure that lookups can handle getting late errors for a requests that has already been failed for disconnection. Current lookup code is quite strict in that it expects a single event per network request.

dapplion:
Note that sync doesn't need to check if the peer is connected before sending a request. Consider these two cases:

Case 1: (Receive disconnect after request)
- Network: peer A disconnects
- Sync: sends request X to peer A
- Sync: receives disconnect event for peer A -> Fails request X

Case 2: (Receive disconnect before request)
- Network: peer A disconnects
- Sync: receives disconnect event for peer A -> Removes peer A from pool of available peers
- Sync: sends request X to peer some other peer

beacon_node/network/src/sync/network_context.rs
```diff
@@ -616,20 +636,20 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
                         .map(|blobs| (blobs, seen_timestamp))
                         .map_err(|e| (e.into(), request.resolve())),
-                    Ok(None) => return None,
+                    Ok(None) => return RpcResponseResult::NoOp,
                     Err(e) => Err((e.into(), request.resolve())),
                 }
             }
             RpcEvent::StreamTermination => match request.remove().terminate() {
-                Ok(_) => return None,
+                Ok(_) => return RpcResponseResult::StreamTermination,
```
dapplion: Why is it useful to differentiate between Noop and StreamTermination if both result in noop on consuming code? Kind of breaks the point of this abstraction, leaking stream terminology. IMO `Option<Result<T>>` is straightforward enough and easy to consume downstream

beacon_node/network/src/sync/network_context.rs
```diff
@@ -177,6 +182,16 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
 
+    /// Remove all active requests associated with the peer.
+    pub fn peer_disconnected(&mut self, peer_id: PeerId) {
+        self.blocks_by_root_requests
+            .retain(|_, request| request.peer_id != peer_id);
+        self.blobs_by_root_requests
+            .retain(|_, request| request.peer_id != peer_id);
+        self.range_blocks_and_blobs_requests
+            .retain(|_, request| request.1.peer_id != peer_id);
+    }
```
dapplion, line 218: This will cause lookups to get stuck upon disconnection. Each active lookup in this hashmaps must result in a Result being propagated to downstream code (block_lookups). When a peer disconnects, every single active request must result in `.on_download_response()` being called with an Err

michaelsproul:
I've re-targeted this at `release-v5.2.1`. See:

- https://github.com/sigp/lighthouse/pull/5984

AgeManning:
I think we can probably also now remove the RPC Error logic we added. I think its not that necessary if we are now relying on the PeerDisconnection message. i.e I think we can remove this entire function: https://github.com/sigp/lighthouse/blob/stable/beacon_node/lighthouse_network/src/rpc/handler.rs#L356

AgeManning:
I can also remove this extra code in a future PR if we prefer to test this standalone?

pawanjay176:
@AgeManning removed in [cd17f9f](https://github.com/sigp/lighthouse/pull/5969/commits/cd17f9f1baee47e114047a1feb36400d5d97ef42)
I have tested this PR on mainnet for forward sync, backfill and lookups. I think its in a good state now.

@dapplion I changed lookups peer_disconnected behaviour such that it drops a lookup only if there are no peers to drive the request forward or if there are no existing downloaded components. 
This would mean that there could be parent requests in lookups that have no associated peer. I think this is okay, lmk if you disagree.

AgeManning (approved):
Nice!!

beacon_node/network/src/sync/network_context.rs
pawanjay176: I was using it for logging the other case, but its not useful. Removed it

beacon_node/network/src/sync/network_context.rs
pawanjay176: Nice catch, I think it should be fixed by [3e1c41a](https://github.com/sigp/lighthouse/pull/5969/commits/3e1c41a18a2212f5f6be191cc80e62fb1dd71e38)

beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -410,21 +410,13 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
...
-                    "block_root" => ?lookup.block_root()
-                );
-                false
-            } else {
-                true
-            }
-        });
+            // Keep the lookup if it was some peers that can drive progress
+            // or if it has some downloaded components that can be processed
+            !lookup.has_no_peers() || lookup.can_progress_without_peer()
```
dapplion: Do we really need to drop lookups here? If a lookup has no peers and needs to make progress it will be dropped. I can not convince myself in what situation dropping a lookup in the `peer_disconnected` function is necessary. Less code and less complexity

beacon_node/network/src/sync/network_context.rs
dapplion: Looks good! That should do it

beacon_node/network/src/sync/network_context.rs
```diff
@@ -177,6 +177,46 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
         }
     }
 
+    /// Returns all ids that were requested to the given peer_id
```
jxs: nit:
```suggestion
    /// Returns the ids of all the requests made to the given peer_id.
```

beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -410,21 +410,13 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
...
-                    "Dropping single lookup after peer disconnection";
-                    "block_root" => ?lookup.block_root()
-                );
-                false
-            } else {
-                true
-            }
-        });
+            // Keep the lookup if it was some peers that can drive progress
+            // or if it has some downloaded components that can be processed
```
jxs:
```suggestion
            // or if it has some downloaded components that can be processed.
```

beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -307,15 +307,49 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
...
+                        }
+                        Ok(BatchOperationOutcome::Continue) => {}
+                        Err(e) => {
+                            self.fail_sync(BackFillError::BatchInvalidState(id, e.0))?;
+                        }
+                    }
+                    // If we have run out of peers in which to retry this batch, the backfill state
+                    // transitions to a paused state.
+                    // We still need to reset the state for all the affected batches, so we should not
+                    // short circuit early
```
jxs:
```suggestion
                    // short circuit early.
```

beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -307,15 +307,49 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
...
         if matches!(
             self.state(),
             BackFillState::Failed | BackFillState::NotRequired
         ) {
             return Ok(());
         }
 
-        self.active_requests.remove(peer_id);
+        if let Some(batch_ids) = self.active_requests.remove(peer_id) {
+            // fail the batches
```
jxs:
```suggestion
            // fail the batches.
```

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -174,8 +174,30 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
...
     /// If the peer has active batches, those are considered failed and re-requested.
-    pub fn remove_peer(&mut self, peer_id: &PeerId) -> ProcessingResult {
-        self.peers.remove(peer_id);
+    pub fn remove_peer(
+        &mut self,
+        peer_id: &PeerId,
+        network: &mut SyncNetworkContext<T>,
+    ) -> ProcessingResult {
+        if let Some(batch_ids) = self.peers.remove(peer_id) {
+            // fail the batches
```
jxs:
```suggestion
            // fail the batches.
```

beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -410,21 +410,13 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
     /* Error responses */
 
     pub fn peer_disconnected(&mut self, peer_id: &PeerId) {
+        /* Check disconnection for single lookups */
```
jxs:
```suggestion
        // Check disconnection for single lookups.
```
or was there a reason for the block comments?

beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -1026,6 +1026,28 @@ fn test_single_block_lookup_failure() {
...
 }
 
+#[test]
+fn test_single_block_lookup_peer_disconnected_then_rpc_error() {
+    let mut rig = TestRig::test_setup();
+
+    let block_hash = Hash256::random();
+    let peer_id = rig.new_connected_peer();
+
+    // Trigger the request
```
jxs:
```suggestion
    // Trigger the request.
```

beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -1026,6 +1026,28 @@ fn test_single_block_lookup_failure() {
...
+    rig.trigger_unknown_block_from_attestation(block_hash, peer_id);
+    let id = rig.expect_block_lookup_request(block_hash);
+
+    // The peer disconnect event reaches sync before the rpc error
```
jxs:
```suggestion
    // The peer disconnect event reaches sync before the rpc error.
```

beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -1026,6 +1026,28 @@ fn test_single_block_lookup_failure() {
...
+    rig.peer_disconnected(peer_id);
+    // Only peer and no progress on lookup, so lookup should be removed
```
jxs:
```suggestion
    // Only peer and no progress on lookup, so lookup should be removed.
```

beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -1026,6 +1026,28 @@ fn test_single_block_lookup_failure() {
...
+    rig.expect_no_active_lookups();
+    // The request fails
```
jxs:
```suggestion
    // The request fails.
```

beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -1026,6 +1026,28 @@ fn test_single_block_lookup_failure() {
...
+    rig.single_lookup_failed(id, peer_id, RPCError::Disconnected);
+    rig.expect_block_lookup_request(block_hash);
+    // The request should be removed from the network context on disconnection
```
jxs:
```suggestion
    // The request should be removed from the network context on disconnection.
```

jxs (approved):
Thanks for driving this Pawan!

dapplion:
@pawanjay176 Added some comments relevant to the API expectations of lookup sync. I think they are helpful to understand bug sources and overall behaviour
- https://github.com/pawanjay176/lighthouse/pull/10

beacon_node/network/src/sync/block_lookups/mod.rs
pawanjay176: Agree, removed in [caee7c8](https://github.com/sigp/lighthouse/pull/5969/commits/caee7c8a853eacdf334fde38bb73d34cb2d92584)

jimmygchen:
@mergify queue

PR #5758, opened 2024-05-09
Ef tests electra

realbigsean:
## Issue Addressed

ef test updates

PR #5972, opened 2024-06-20
Electra attestation changes sean review

realbigsean:


testing/ef_tests/tests/tests.rs
```diff
@@ -232,7 +232,6 @@ mod ssz_static {
     ssz_static_test!(fork, Fork);
     ssz_static_test!(fork_data, ForkData);
     ssz_static_test!(historical_batch, HistoricalBatch<_>);
-    ssz_static_test!(indexed_attestation, IndexedAttestation<_>);
```
realbigsean, line 235: we can't generate these tests with superstructed objects, but the full test implementation is included in https://github.com/sigp/lighthouse/pull/5758

consensus/state_processing/src/verify_operation.rs
```diff
@@ -48,7 +48,6 @@ pub trait TransformPersist {
 pub struct SigVerifiedOp<T: TransformPersist, E: EthSpec> {
     op: T,
     verified_against: VerifiedAgainst,
-    //#[ssz(skip_serializing, skip_deserializing)]
```
michaelsproul, line 51: I am surprised this works.

michaelsproul (approved):

PR #5975, opened 2024-06-21
Avoid changing slasher schema for Electra

michaelsproul:
## Proposed Changes

Remove the need for a slasher schema migration at Electra by storing indexed attestations in a fork-neutral format on disk.

This is preferable to adding a tag byte to `IndexedAttestation` because the slasher DB could contain hundreds of gigabytes of indexed attestations, and rewriting them would take considerable time (I'd guess 30 mins+). There's also no real benefit to adding the tag bytes, as the `Vec<u64>` for the attesting indices is quite easy to understand & maintain.

## Additional Info

The slasher didn't previously know anything about `ChainSpec`, so I've plumbed that in.

### DIFF
--- a/slasher/src/database.rs
+++ b/slasher/src/database.rs
@@ -865,7 +865,7 @@ mod test {
         let decoded_on_disk = IndexedAttestationOnDisk::from_ssz_bytes(&encoded).unwrap();
         assert_eq!(decoded_on_disk, on_disk);
 
-        let decoded = on_disk.into_indexed_attestation(&spec).unwrap();
+        let decoded = on_disk.into_indexed_attestation(spec).unwrap();
         assert_eq!(decoded, fork_attestation);
     }
 
--- a/slasher/src/database.rs
+++ b/slasher/src/database.rs
@@ -258,7 +258,7 @@ impl IndexedAttestationOnDisk {
         spec: &ChainSpec,
     ) -> Result<IndexedAttestation<E>, Error> {
         let fork_at_target_epoch = spec.fork_name_at_epoch(self.data.target.epoch);
-        if fork_at_target_epoch >= ForkName::Electra {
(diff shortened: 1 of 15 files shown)


slasher/src/database.rs
```diff
@@ -775,3 +815,84 @@ impl<E: EthSpec> SlasherDB<E> {
...
+                epoch: Epoch::new(0),
+                root: Hash256::repeat_byte(0xbb),
+            },
+            target: Checkpoint {
+                epoch: Epoch::new(31),
+                root: Hash256::repeat_byte(0xcc),
+            },
+        };
+
+        let attesting_indices = vec![1, 14, 160, 812737];
```
dapplion: Would be best to test with a Vec of length of MaxValidatorsPerCommittee and MaxValidatorsPerSlot

michaelsproul: I've gotta run off, do you mind pushing this change?

dapplion (approved):
LGTM! Agree on not requiring the migration

slasher/src/database.rs
```diff
@@ -236,6 +239,43 @@ impl AsRef<[u8]> for IndexedAttestationId {
...
+    signature: AggregateSignature,
+}
+
+impl IndexedAttestationOnDisk {
+    fn into_indexed_attestation<E: EthSpec>(
+        self,
+        spec: &ChainSpec,
+    ) -> Result<IndexedAttestation<E>, Error> {
+        let fork_at_target_epoch = spec.fork_name_at_epoch(self.data.target.epoch);
+        if fork_at_target_epoch >= ForkName::Electra {
```
realbigsean:
```suggestion
        if fork_at_target_epoch.electra_enabled() {
```

realbigsean (approved):
Awsome, thanks Michael!

PR #5976, opened 2024-06-21
Broadcast validator registration to all synced beacon nodes

jimmygchen:
## Issue Addressed

The validator client currently publish validator registrations to its first beacon node, and it doesn't require the beacon node to be synced. This could potentially result in registration not being sent to all builders beacon nodes are connected to, because:
- Beacon nodes may be connected to different relays, and only sending it to the first one means registrations may not be sent to all relays.
- Beacon nodes read validator data from the chain, and only send registration for active validators. However if the node is out of sync, it could either be missing the validator or have an outdated status of the validator, resulting in registration for the active validator to be missed.

https://github.com/sigp/lighthouse/blob/b4556a3d621c39cb28ef10df9daec4ee935909b6/validator_client/src/preparation_service.rs#L478-L487

## Proposed Changes

To ensure registrations for all active validators are sent to all relays, we publish validator registration to **all synced** beacon nodes.

### DIFF
--- a/validator_client/src/preparation_service.rs
+++ b/validator_client/src/preparation_service.rs
@@ -478,7 +478,7 @@ impl<T: SlotClock + 'static, E: EthSpec> PreparationService<T, E> {
                 match self
                     .beacon_nodes
                     .broadcast(
-                        RequireSynced::Yes,
+                        RequireSynced::No,
                         OfflineOnFailure::No,
                         |beacon_node| async move {
                             beacon_node.post_validator_register_validator(batch).await
--- a/validator_client/src/preparation_service.rs
+++ b/validator_client/src/preparation_service.rs
@@ -477,8 +477,8 @@ impl<T: SlotClock + 'static, E: EthSpec> PreparationService<T, E> {
             for batch in signed.chunks(self.validator_registration_batch_size) {
                 match self
                     .beacon_nodes
-                    .first_success(
-                        RequireSynced::No,
+                    .broadcast(
+                        RequireSynced::Yes,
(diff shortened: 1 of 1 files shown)


validator_client/src/preparation_service.rs
```diff
@@ -477,8 +477,8 @@ impl<T: SlotClock + 'static, E: EthSpec> PreparationService<T, E> {
             for batch in signed.chunks(self.validator_registration_batch_size) {
                 match self
                     .beacon_nodes
-                    .first_success(
-                        RequireSynced::No,
+                    .broadcast(
+                        RequireSynced::Yes,
```
dapplion: > Beacon nodes read validator data from the chain, and only send registration for active validators. However if the node is out of sync, it could either be missing the validator or have an outdated status of the validator, resulting in registration for the active validator to be missed.

Why is it best to require synced status? Eagerly sending registrations before synced status should not be harmful. If there any downside to it? Registrations will be sent every epoch regardless

jimmygchen: Yeah you're right - it only needs to require synced status if we're using `first_success` (as the node may exclude validators not active in local view; and in fact `RequireSynced::No` may miss registrations for some BNs connected to different relays, so this should be `Yes` here.

I'll fix this. Thanks!

jimmygchen: Fixed!

dapplion (approved):

michaelsproul:
@mergifyio queue

jimmygchen:
@mergify requeue

PR #5978, opened 2024-06-21
Hierarchical state diffs

michaelsproul:
## Proposed Changes

This PR contains the freezer DB changes from the main `tree-states` branch, bringing the small & efficient archive node database to stable Lighthouse :tada:

## Additional Info

(From @dapplion): Some resources to get up to speed with HDiff concepts:
- https://github.com/dapplion/tree-states-review-guide/blob/main/persisted_hdiff.md
- https://hackmd.io/G82DNSdvR5Osw2kg565lBA


dapplion:
I get a compilation error from xdelta3 trying to compile this branch:

EDIT: Fixed (cropped previous text because it's long)

michaelsproul:
looks like it might be a Mac thing

I had to fork xdelta3 already, we should move it under sigp and patch it

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -1247,48 +1268,130 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
+        self.store_cold_state_summary(state_root, state.slot(), ops)?;
 
-        if state.slot() % self.config.slots_per_restore_point != 0 {
-            return Ok(());
+        let slot = state.slot();
+        match self.hierarchy.storage_strategy(slot)? {
+            StorageStrategy::ReplayFrom(from) => {
+                debug!(
+                    self.log,
+                    "Storing cold state";
```
dapplion, line 1634: Not really, @michaelsproul do you want the same log message for all variants? Here it's just storing a summary

dapplion:
## TODOs

EDIT: _completed_

dapplion:
TODO ~ Some metrics that we should add:

- https://github.com/sigp/lighthouse/pull/6041
  - byte size of `diff_buffer_cache_size`
  - elements in `diff_buffer_cache_size`
  - counter for replayed blocks
  - counter for diff_buffer_cache hits / misses
  - timer for storage time
  - timer for load times
- https://github.com/sigp/lighthouse/pull/6041  
  - counter for diffs loaded / stored
  - counter for snapshots loaded / stored

beacon_node/store/src/hot_cold_store.rs
michaelsproul: I quite like the uniformity of the logs because it helps with grepping things like "was there anything stored at this slot?". The `strategy: replay` makes it clear enough that a full state isn't being stored, at least for a debug log.

beacon_node/src/cli.rs
```diff
@@ -980,6 +994,15 @@ pub fn cli_app() -> Command {
                 .default_value("0")
                 .display_order(0)
         )
+        .arg(
+            Arg::new("allow-tree-states-migration")
+                .long("allow-tree-states-migration")
```
michaelsproul: This should probably be an flag without an argument rather than taking a bool

michaelsproul:
After the migration, we've got this error trying to fetch a historic block:

```json
{
  "code": 500,
  "message": "UNHANDLED_ERROR: DBError(InvalidBytes)",
  "stacktraces": []
}
```

Need to look into it.

michaelsproul:
Hmm we also have:

> ERRO Error during iteration over blocks      error: DBError(ForwardsIterGap(BeaconBlockRoots, Slot(2396192), Slot(1069824))), peer: 16Uiu2HAmE3J6yaG8iooyxwRRn9zd2pkc1wv7D7q29ys3ESjaALuv, request: V2(BlocksByRangeRequestV2 { start_slot: 1069824, count: 32 })

xrchz:
So cannot wait for this

beacon_node/src/cli.rs
michaelsproul: Deleted

michaelsproul:
Ready for final review and merge. This is confirmed working on test infra, and all FIXMEs have been resolved.

beacon_node/src/cli.rs
```diff
@@ -789,11 +803,23 @@ pub fn cli_app() -> Command {
             Arg::new("historic-state-cache-size")
                 .long("historic-state-cache-size")
                 .value_name("SIZE")
-                .help("Specifies how many states from the freezer database should cache in memory")
+                .help("This cache is currently inactive. Please use hdiff-buffer-cache-size instead.")
```
dapplion: Should prepend `DEPRECATED`

beacon_node/store/src/hdiff.rs
```diff
@@ -0,0 +1,453 @@
...
+            Ok(slot)
+        } else {
+            Ok((slot / last + 1) * last)
+        }
+    }
+
+    /// Return `true` if the database ops for this slot should be committed immediately.
+    ///
+    /// This is the case for all diffs in the 2nd lowest layer and above, which are required by diffs
+    /// in the 1st layer.
```
dapplion: Why is this required?

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -2855,30 +2804,25 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
-        info!(
-            self.log,
-            "Deleting historic states";
-            "num_kv" => cold_ops.len(),
-        );
-        self.cold_db.do_atomically(std::mem::take(&mut cold_ops))?;
-
-        // If we just deleted the the genesis state, re-store it using the *current* schema, which
-        // may be different from the schema of the genesis state we just deleted.
+        // If we just deleted the the genesis state, re-store it using the current* schema.
```
dapplion:
```suggestion
        // If we just deleted the genesis state, re-store it using the current* schema.
```

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -2972,28 +2916,22 @@ pub fn migrate_database<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>>(
...
-                && anchor_slot.map_or(true, |anchor_slot| slot >= &anchor_slot)
-        }
-        Err(_) => true,
-    }) {
-        let (block_root, state_root, slot) = maybe_tuple?;
+    let state_roots = RootsIterator::new(&store, finalized_state)
+        .take_while(|result| match result {
+            Ok((_, _, slot)) => {
+                slot >= &current_split_slot
+                    && anchor_slot.map_or(true, |anchor_slot| slot >= &anchor_slot)
```
dapplion: Why do you need to check for both `current_split_slot` and `anchor_slot`?

michaelsproul: Yeah we don't, this was just preserving what was there. Will fix

beacon_node/store/src/hdiff.rs
michaelsproul: can't diff a leaf against the previous layer if the previous layer doesn't exist

michaelsproul: will add a comment

beacon_node/src/cli.rs
michaelsproul: I kind of don't want to, because we might want to bring it back.

beacon_node/store/src/hot_cold_store.rs
michaelsproul: https://github.com/sigp/lighthouse/pull/5978/commits/3d90ac68239a0dfee853f2fc2d4ef28a22a701b2

beacon_node/store/src/hdiff.rs
michaelsproul: https://github.com/sigp/lighthouse/pull/5978/commits/3d90ac68239a0dfee853f2fc2d4ef28a22a701b2

beacon_node/src/cli.rs
```diff
@@ -789,11 +803,23 @@ pub fn cli_app() -> Command {
...
                 .display_order(0)
         )
+        .arg(
+            Arg::new("hdiff-buffer-cache-size")
+                .long("hdiff-buffer-cache-size")
+                .value_name("SIZE")
+                .help("Number of hierarchical diff (hdiff) buffers to cache in memory. Each buffer \
+                       is around the size of a BeaconState so you should be cautious about setting \
+                       this value too high. This flag is irrelevant for most nodes, which run with \
+                       state pruning enabled.")
```
jimmygchen, line 816: Is it worth adding a `.requires` on archive node args?

beacon_node/src/cli.rs
```diff
@@ -759,9 +759,23 @@ pub fn cli_app() -> Command {
             Arg::new("slots-per-restore-point")
                 .long("slots-per-restore-point")
                 .value_name("SLOT_COUNT")
-                .help("Specifies how often a freezer DB restore point should be stored. \
-                       Cannot be changed after initialization. \
-                       [default: 8192 (mainnet) or 64 (minimal)]")
+                .help("DEPRECATED. This flag has no effect.")
```
jimmygchen, line 759: There are a few more references of this args that needs to be removed, e.g.:

https://github.com/sigp/lighthouse/blob/3070cb7c3982f8321bbeb33413009f8f928bdf8a/book/src/advanced_database.md#L16

https://github.com/sigp/lighthouse/blob/7b1a0d4bba4d6ad8c85273745e273fd8cdf74cc6/watch/README.md#L42

beacon_node/src/cli.rs
michaelsproul: Update: we brought it back

beacon_node/src/cli.rs
michaelsproul: A node can be an archive node even if it isn't running with `--reconstruct-historic-states` (if it ran with that flag in the past), so I think we can't really do this

beacon_node/src/cli.rs
michaelsproul: updated the docs in this commit: https://github.com/sigp/lighthouse/pull/5978/commits/e87a618a29d1cce7060dee8607e5eb3b3c29eb13

beacon_node/src/cli.rs
jimmygchen: Got it, make sense

book/src/advanced_database.md
```diff
@@ -72,17 +83,19 @@ The historical state cache size can be specified with the flag `--historic-state
...
+- _Cold DB_: see _Freezer DB_.
+- _HDiff_: hierarchical state diff.
+- _Hierarchy Exponents_: configuration for hierarchical state diffs, which determines the density
+  of stored diffs and snapshots in the freezer DB.
+- _Hot DB_: part of the database storing recent states, all blocks, and other runtime data. Full
   states are stored every epoch.
-* _Restore Point_: a full `BeaconState` stored periodically in the freezer DB.
-* _Slots Per Restore Point (SPRP)_: the number of slots between restore points in the freezer DB.
-* _Split Slot_: the slot at which states are divided between the hot and the cold DBs. All states
+- _Snapshot_: a full `BeaconState` stored periodically in the freezer DB. Yearly by default.
```
jimmygchen: should be every 291 days?

beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -253,35 +250,21 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
...
-                            "Backfill sync not required whilst in failed status"
-                        );
-                        self.set_state(BackFillState::NotRequired);
-                        return Err(BackFillError::InvalidSyncState(String::from(
-                            "backfill not required",
-                        )));
-                    }
-                    Ok(_) => {}
+                if let Err(e) = self.reset_start_epoch() {
+                    let ResetEpochError::SyncCompleted = e;
```
jimmygchen, line 253: This line seems redundant?

beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -163,21 +163,18 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
...
-                        BackFillState::Paused,
-                        anchor_info
-                            .oldest_block_slot
-                            .epoch(T::EthSpec::slots_per_epoch()),
-                    )
-                }
-            }
-            None => (BackFillState::NotRequired, Epoch::new(0)),
-        };
+        let anchor_info = beacon_chain.store.get_anchor_info();
```
jimmygchen, line 163: Part of the comment above can be removed as it's no longer relevant.

beacon_node/http_api/src/lib.rs
```diff
@@ -2693,24 +2693,37 @@ pub fn serve<T: BeaconChainTypes>(
...
                             .map_err(inconsistent_fork_rejection)?;
+                        let timer = metrics::start_timer(&metrics::HTTP_API_STATE_SSZ_ENCODE_TIMES);
+                        let response_bytes = state.as_ssz_bytes();
+                        drop(timer);
+                        debug!(
+                            log,
+                            "HTTP state load";
+                            "load_time_ms" => t.elapsed().as_millis(),
+                            "target_slot" => state.slot()
+                        );
```
jimmygchen, line 2722: Is it intentional to include the encoding time in `load_time_ms`?

common/eth2_network_config/src/lib.rs
```diff
@@ -154,6 +154,26 @@ impl Eth2NetworkConfig {
         }
     }
 
+    pub fn genesis_state_root<E: EthSpec>(&self) -> Result<Option<Hash256>, String> {
```
jimmygchen, line 162: Do we need the `Option` here? 
Might be able to simplify quite a few functions if we remove this `Option`

beacon_node/store/src/config.rs
```diff
@@ -74,22 +114,88 @@ impl Default for StoreConfig {
...
+            Err(StoreConfigError::ZeroEpochsPerBlobPrune)
+        }
+    }
+
+    /// Estimate the size of `len` bytes after compression at the current compression level.
+    pub fn estimate_compressed_size(&self, len: usize) -> usize {
+        if self.compression_level == 0 {
+            len
+        } else {
+            len / EST_COMPRESSION_FACTOR
```
jimmygchen, line 179: Does the compression factor vary depending on the `compression_level`?

michaelsproul: No, weirdly the compression level doesn't seem to have any effect :rofl:

beacon_node/store/src/errors.rs
```diff
@@ -38,27 +39,43 @@ pub enum Error {
...
     ///
     /// This should never happen (it's a logic error).
-    StateReconstructionDidNotComplete,
+    StateReconstructionLogicError,
     StateReconstructionRootMismatch {
         slot: Slot,
         expected: Hash256,
         computed: Hash256,
     },
+    MissingStateRoot(Slot),
```
jimmygchen: Unused, is this for a future PR?

beacon_node/store/src/errors.rs
```diff
@@ -38,27 +39,43 @@ pub enum Error {
...
+    MissingState(Hash256),
+    MissingGenesisState,
+    MissingSnapshot(Slot),
+    NoBaseStateFound(Hash256),
```
jimmygchen: Unused as well

beacon_node/store/src/errors.rs
```diff
@@ -38,27 +39,43 @@ pub enum Error {
...
     BlockReplayError(BlockReplayError),
+    MilhouseError(milhouse::Error),
+    Compression(std::io::Error),
+    MissingPersistedBeaconChain,
```
jimmygchen: A few of these are unused, can we remove them for now and add it when needed?

michaelsproul: Yes let's get rid of them! I think they were just leftover from code that got deleted. It would be good if we could catch these automatically, but I don't think we can because the type is `pub`

beacon_node/http_api/src/lib.rs
michaelsproul: Yeah. Maybe we could rename it to `total_time_ms`?

beacon_node/network/src/sync/backfill_sync/mod.rs
michaelsproul: This is to make sure we handle all variants if the `ResetEpochError` gets refactored.

If we wrote:

```rust
if let Err(ResetEpochError::SyncCompleted) = self.reset_start_epoch() {
   ...
}
```

Then we could accidentally _not_ run this code for a newly added variant. Similarly if we just use `e` and run this code for all variants then that may not be correct. So we're forcing a breakage if any variants get added in future, similar to if we used `match`, but with less nesting:

```rust
match self.reset_start_epoch() {
    Err(ResetEpochError::SyncCompleted) => { ... },
    Ok(_) => {},
}
```

I'll add a comment.

beacon_node/store/src/errors.rs
michaelsproul: Updated in https://github.com/sigp/lighthouse/pull/5978/commits/c4d6ef3544b529bbbc8a919916165e6e976eb1fb

beacon_node/store/src/errors.rs
michaelsproul: Middle two are used, but removed the first and last in https://github.com/sigp/lighthouse/pull/5978/commits/c4d6ef3544b529bbbc8a919916165e6e976eb1fb

beacon_node/store/src/errors.rs
michaelsproul: Deleted in https://github.com/sigp/lighthouse/pull/5978/commits/c4d6ef3544b529bbbc8a919916165e6e976eb1fb

beacon_node/store/src/config.rs
michaelsproul: Added a comment https://github.com/sigp/lighthouse/pull/5978/commits/c4d6ef3544b529bbbc8a919916165e6e976eb1fb

beacon_node/http_api/src/lib.rs
michaelsproul: Renamed in https://github.com/sigp/lighthouse/pull/5978/commits/c4d6ef3544b529bbbc8a919916165e6e976eb1fb

beacon_node/network/src/sync/backfill_sync/mod.rs
michaelsproul: Removed in https://github.com/sigp/lighthouse/pull/5978/commits/c4d6ef3544b529bbbc8a919916165e6e976eb1fb

beacon_node/network/src/sync/backfill_sync/mod.rs
michaelsproul: Comment added https://github.com/sigp/lighthouse/pull/5978/commits/c4d6ef3544b529bbbc8a919916165e6e976eb1fb

book/src/advanced_database.md
michaelsproul: Good catch! https://github.com/sigp/lighthouse/pull/5978/commits/c4d6ef3544b529bbbc8a919916165e6e976eb1fb

common/eth2_network_config/src/lib.rs
michaelsproul: Hmm, I think we need the `None` case for networks with unknown genesis. I'll refactor this so that it returns `None` rather than an error in that case.

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -1530,48 +1600,142 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
+            let _timer = metrics::start_timer(&metrics::STORE_BEACON_HDIFF_BUFFER_COMPUTE_TIME);
+            HDiff::compute(&base_buffer, &target_buffer, &self.config)?
+        };
+        let diff_bytes = diff.as_ssz_bytes();
 
+        let key = get_key_for_col(
+            DBColumn::BeaconStateDiff.into(),
+            &state.slot().as_u64().to_be_bytes(),
+        );
+        ops.push(KeyValueStoreOp::PutKeyValue(key, diff_bytes));
```
jimmygchen, line 1738: is it worth compressing the diffs too?

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -1580,157 +1744,226 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
+
+                self.historic_state_cache
+                    .lock()
+                    .put_both(slot, state.clone(), buffer);
+                Ok(state)
+            }
+            StorageStrategy::ReplayFrom(from) => {
+                let base_state = if let Some(state) = cached_state {
+                    // Found a prior state in the historic state cache.
+                    state
```
jimmygchen: `cached_state` can't be `Some` here, as it would have returned above.

beacon_node/store/src/hot_cold_store.rs
michaelsproul: Nah I tested this and it doesn't save any space (and takes a lot longer). The diff algorithm already effectively compresses

beacon_node/store/src/hot_cold_store.rs
michaelsproul: nice catch!

michaelsproul: fixed in https://github.com/sigp/lighthouse/pull/5978/commits/f6412f32b9df557dbe73d8aa430fc5b80da1c22a

common/eth2_network_config/src/lib.rs
michaelsproul: Fixed and documented here: https://github.com/sigp/lighthouse/pull/5978/commits/6c327330845a5a6b82d2ceb4a752a63a65a75de8

beacon_node/store/src/hdiff.rs
```diff
@@ -0,0 +1,489 @@
...
+    /// - Full state stored every 2^21 slots (2097152 slots or 291 days)
+    /// - First diff layer stored every 2^13 slots (8192 slots or 2.3 hours)
+    /// - Second diff layer stored every 2^5 slots (32 slots or 1 epoch)
+    ///
+    /// To reconstruct a state at slot 3,000,003 we load each closest layer
+    /// - Layer 0: 3000003 - (3000003 mod 2^21) = 2097152
+    /// - Layer 1: 3000003 - (3000003 mod 2^13) = 2998272
+    /// - Layer 2: 3000003 - (3000003 mod 2^5)  = 3000000
+    ///
+    /// Layer 0 is full state snaphost, apply layer 1 diff, then apply layer 2 diff and then replay
```
jimmygchen:
```suggestion
    /// Layer 0 is full state snapshot, apply layer 1 diff, then apply layer 2 diff and then replay
```

beacon_node/store/src/hdiff.rs
```diff
@@ -0,0 +1,489 @@
...
+}
+
+impl HDiffBuffer {
+    pub fn from_state<E: EthSpec>(mut beacon_state: BeaconState<E>) -> Self {
+        let _t = metrics::start_timer(&metrics::STORE_BEACON_HDIFF_BUFFER_FROM_STATE_TIME);
+        // Set state.balances to empty list, and then serialize state as ssz
+        let balances_list = std::mem::take(beacon_state.balances_mut());
+
+        let state = beacon_state.as_ssz_bytes();
+        let balances = balances_list.to_vec();
```
jimmygchen, line 194: We no longer need `balance_list`, I'm wondering if it's more efficient to convert it using `into_Iter()` rather than cloning? although I don't know how much differences this make given that overhead of copying a `u64` is minimal - although this list is really large.

michaelsproul: Oh great catch! I had assumed this was not cloning, but it is :facepalm:

I'll change it to `Vec::from(balances_list)` which will just move the vec out of the `VariableList`: https://docs.rs/ssz_types/0.8.0/src/ssz_types/variable_list.rs.html#133-137

michaelsproul: Actually nvm, I got confused :facepalm:

This is a `milhouse::List` (i.e. a tree) so we can't make it into a vec any more cheaply than this. And I think we need to keep the vec so that the diff computation remains fast

jimmygchen: Got it, thanks for looking into this

jimmygchen (approved):
Looks really good, thanks for addressing the comments! 🎉

I don't fully understand all the database details, but I've reviewed all the changes and I don't see any obvious issues.

I believe this is good to go, and happy to help optimise further if we need to.

dapplion (dismissed):
Testing this we noted much worse performance of the binary patch library than we expected originally. The BeaconState has grown significantly since Michael did the original research of what diff algorithm to use. Despite Electra activating in some months there's no guarantee that the network will consolidate, so rampant state growth is a concern.

Currently it takes 1.3 sec to compute a diff on average on a healthy holesky node. 


  |   | State size (MB) |   |  
-- | -- | -- | -- | --
% Supply staked | Validators (M) | 0% bloat | 100% bloat | 200% bloat
100 | 3.75 | 543.75 | 1087.5 | 1631.25
80 | 3 | 435 | 870 | 1305
60 | 2.25 | 326.25 | 652.5 | 978.75
40 | 1.5 | 217.5 | 435 | 652.5
30 | 1.125 | 163.125 | 326.25 | 489.375
20 | 0.75 | 108.75 | 217.5 | 326.25

There's a possibility of the BeaconState growing close to 1GB. I run some benchmarks on our diff library, and the time to compute a diff is (random mutations of total size diff bytes), (tested on an M2 chip)

| data size bytes | diff bytes | compute diff time (ms)
| - | - | - |
| 2e8 (today's size) | 1e3 | 584.03
| 2e8 | 1e4 | 635.24
| 2e8 | 1e5 | 778.79
| 2e8 | 1e6 | 1507.0
| 1e9 (future size) | 1e3 | 4895.6
| 1e9 | 1e4 | 3885.0
| 1e9 | 1e5 | 6301.1
| 1e9 | 1e6 | 15858.0

For 1GB input data those are pretty terrifying numbers. I'm hitting **Request changes** to:
- Make this branch future proof to allow us to improve the diff algorithm in the future. For example, version the diffs and use more advance strategies latter
- Come up with a better diff algorithm now

michaelsproul:
Versioning the diff algorithm sounds like a great idea. I also think we could probably include https://github.com/sigp/lighthouse/pull/6556

michaelsproul:
I'll merge this PR first and then resolve merge conflicts:

- https://github.com/sigp/lighthouse/pull/6268

michaelsproul (commented):
Approved :heavy_check_mark: 

obama

michaelsproul:
@mergify queue

PR #5980, opened 2024-06-21
Add electra lightclient types

dapplion:
## Issue Addressed

Extends LightClient types with Electra variant. Nothing explicitly new, but the execution header has new fields.

## Proposed Changes

Lots of boilerplate for Electra variants -.- Hopefully we can throw macros here when we have time.

Plus fixes a bug in `ssz_max_len_for_fork` where it was always computing the fixed size of altair variant.


realbigsean (approved):
Nice, looks good

PR #5983, opened 2024-06-21
Fix Calculations Relying on MAX_EB

ethDreamer:
## Issue Addressed

* #5605 

this addresses the changes to both `compute_proposer_index()` and `get_next_sync_committee_indices()`

### DIFF
--- a/consensus/types/src/beacon_state.rs
+++ b/consensus/types/src/beacon_state.rs
@@ -892,11 +892,7 @@ impl<E: EthSpec> BeaconState<E> {
             return Err(Error::InsufficientValidators);
         }
 
-        let max_effective_balance = if self.fork_name_unchecked() >= ForkName::Electra {
-            spec.max_effective_balance_electra
-        } else {
-            spec.max_effective_balance
-        };
+        let max_effective_balance = spec.max_effective_balance_for_fork(self.fork_name_unchecked());
 
         let mut i = 0;
         loop {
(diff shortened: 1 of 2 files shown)


consensus/types/src/beacon_state.rs
```diff
@@ -1091,6 +1095,12 @@ impl<E: EthSpec> BeaconState<E> {
 
         let seed = self.get_seed(epoch, Domain::SyncCommittee, spec)?;
 
+        let max_effective_balance = if self.fork_name_unchecked() >= ForkName::Electra {
+            spec.max_effective_balance_electra
+        } else {
+            spec.max_effective_balance
```
dapplion: Should we make `max_effective_balance` a function to prevent these typos?

ethDreamer: i think that would only do us any good if we made `max_effective_balance` private so it's only accessible via the function. Would you wanna do that?

dapplion: How many places is `max_effective_balance` used in a context that does not require to be fork aware?

ethDreamer: Okay I've used a function for this that already existed on the `spec` object. I haven't yet made these fields private because the code in `validator.rs` is super convoluted and needs to be refactored to support that. Just don't have time to do it in this PR as we need to get `devnet-1` out and we only have 1 day and i gotta get working on the ef-tests

Issue #5640, opened 2024-04-24
Will lighthouse return an empty list when fetching an expired blob?


zhiqiangxu:
We're talking about how should op stack deal with expired blobs [here](https://github.com/ethereum-optimism/optimism/pull/10269). Currently op stack was expecting an error to be returned for expired blobs. `Prysm` will return `empty list/status=200` for expired blobs. What about lighthouse?

I've read the code for `beacon/blob_sidecars/{block_id}`, it seems lighthouse will also return an empty list when fetching an expired blob.

Just wanna confirm with the official team:)

michaelsproul:
Yeah at the moment Lighthouse will also return an empty list.

I don't really like this behaviour though, because it combines the cases for (1) no blobs ever existed at this slot and (2) there were blobs at this slot but we've deleted them.

dapplion:
Worth it to raise the issue in the beacon APIs specs. There has been similar discussions for the case of missed slots. Would having different HTTP status code serve your use case?

zhiqiangxu:
I think so, [here](https://github.com/ethereum-optimism/optimism/blob/54eb4161dc0d2dad7e6840b788ef9704a0f5c721/op-service/sources/l1_beacon_client.go#L69-L93) is how op stack is handling the response.

michaelsproul:
I think we should just return a 404 here. I'll raise a PR.

michaelsproul:
PR here:

- https://github.com/sigp/lighthouse/pull/6331

michaelsproul:
The fix will be in the next release - v6.0.0

PR #5987, opened 2024-06-24
add missing fields to get blob sidecars request

### LINKED ISSUE #5977: Missing fields in `getBlobSidecars` response
## Version

v5.2.0

## Present Behaviour

Missing `version`, `execution_optimistic`, `finalized` fields in `getBlobSidecars` JSON response.

This isn't an issue at the moment as there's only one version (Deneb) of blob sidecars, but it would be ideal to be spec compliant.

## Expected Behaviour

Include the fields mentioned above in JSON response.

Beacon API spec:
https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Beacon/getBlobSidecars

eserilev:
## Issue Addressed

Closes #5977 

## Proposed Changes

Adds missing fields to get blob sidecars request


beacon_node/http_api/src/lib.rs
```diff
@@ -1715,6 +1715,12 @@ pub fn serve<T: BeaconChainTypes>(
                     let indices = indices_res?;
                     let blob_sidecar_list_filtered =
                         block_id.blob_sidecar_list_filtered(indices, &chain)?;
+                    let (block, execution_optimistic, finalized) =
+                        block_id.blinded_block(&chain)?;
```
michaelsproul: This is a bit inefficient because we end up computing the block root twice, once when loading the blobs and then again when loading the block. I think this is probably an acceptable performance tradeoff for now, but maybe we could add a comment here noting that this could be optimised further if desired

eserilev: updated to only compute the block root once [a7c60a4](https://github.com/sigp/lighthouse/pull/5987/commits/a7c60a448a60aff6d502824af4f64f28834d6f1f)

jimmygchen: Would it be more efficient if we get the root first with 

```
let root = self.root(chain)?.0;
```

and then use this root for both `chain.get_blinded_block` and `chain.get_blobs`, so we don't have to do `block.message.tree_hash_root()`?

eserilev: added in 969e4c7fa76b4bb90afa714e90530638bd492d08

beacon_node/http_api/src/block_id.rs
```diff
@@ -252,23 +255,39 @@ impl BlockId {
...
+    > {
+        let (root, execution_optimistic, finalized) = self.root(chain)?;
+        let block = match &self.0 {
+            CoreBlockId::Head => {
+                let (cached_head, _) = chain
+                    .canonical_head
+                    .head_and_execution_status()
+                    .map_err(warp_utils::reject::beacon_chain_error)?;
+                cached_head.snapshot.beacon_block.clone_as_blinded()
+            }
```
michaelsproul: Although it's nice to fetch the head block without having to go to the DB, there's a race condition here because the canonical head could change between the call to get the `root` and the call to load the `head_and_execution_status`. This race could result in occasionally returning the wrong data, like the blobs from slot `N` and the block from slot `N + 1`.

I think the simplest fix might be to just remove the special casing for the head here. Loading from the DB usually isn't too slow, because the DB does its own caching

michaelsproul (changes requested):
Looks good on the whole, just one small issue that I think we should fix before merging

chong-he:
Tested and can confirm that the fields `version`, `execution_optimistic` and `finalized` are now in the beacon API query of endpoint `eth/v1/beacon/blob_sidecars`

dapplion (approved):
http crate is not my expertise but can't find any issues with the current approach :)

michaelsproul (approved):

beacon_node/http_api/src/block_id.rs
```diff
@@ -252,23 +255,30 @@ impl BlockId {
...
+        ),
+        warp::Rejection,
+    > {
+        let (root, execution_optimistic, finalized) = self.root(chain)?;
+        let block = BlockId::blinded_block_by_root(&root, chain)?.ok_or_else(|| {
+            warp_utils::reject::custom_not_found(format!("beacon block with root {}", root))
+        })?;
+
+        // Return the `BlobSidecarList` identified by `self`.
+        let blob_sidecar_list = chain
```
michaelsproul, line 278: Now that we're loading the block and blobs together, it could be good to check that all the blobs are loaded, and return a 404 in case they aren't

Related to this issue:

- https://github.com/sigp/lighthouse/issues/5640

I added (and then removed) the `backwards-incompat` tag because I thought this change may already accomplish this, but we don't have the check (yet), so it doesn't.

michaelsproul:
@mergify queue

Issue #2, opened 2018-07-09
Implement new serialization format


paulhauner:
As detailed here: https://ethresear.ch/t/blob-serialisation/1705

paulhauner:
Closing this as it is not immediately required.

PR #5990, opened 2024-06-25
Reconstruct data columns without blocking processing and import

jimmygchen:
## Issue Addressed

Part of #4983. This is a work in progress.

The current sequence of reconstruction:
1. Receive blocks & data columns
2. As soon as the supernode receives 50% of columns, start reconstruction (while holding the availability cache write lock)
3. After reconstruction completes, the node imports block and publishes the remaining 50% of columns
4. Now it may also receives the remaining 50%, which gets verified and then ignored

## Proposed Changes

- Changes to data column reconstruction:
  - Attempt reconstruction without holding availability cache lock, so we can process other gossip / rpc data columns simultaneously
  - Check availability cache again before publishing reconstructed columns to avoid publishing excess duplicates
- Remove some unnecessary `RuntimeVariableList` conversion for data columns

## Additional Info

Below are logs from local testing, notice that the reconstruction blocks processing of incoming data columns (due to holding the availability cache write lock), and they gets ignored as duplicates:

```
Jun 24 08:01:15.781 DEBG Successfully verified gossip data column sidecar, index: 63, block_root: 0x690d…0e3e, slot: 40556
Jun 24 08:01:16.280 DEBG Reconstructed columns                   count: 64, service: availability_cache, service: beacon
Jun 24 08:01:16.281 DEBG Writing data_columns to store           count: 128, block_root: 0x690d…0e3e, service: beacon
Jun 24 08:01:16.284 INFO Gossipsub data column processed, imported fully available block, block_root: 0x690d…0e3e
Jun 24 08:01:16.284 DEBG Sending pubsub messages                 topics: [DataColumnSidecar(DataColumnSubnetId(0)), DataColumnSidecar(DataColumnSubnetId(1)), DataColumnSidecar(DataColumnSubnetId(2)) .. DataColumnSidecar(DataColumnSubnetId(31))], count: 64, service: network
Jun 24 08:01:16.314 DEBG Successfully verified gossip data column sidecar, index: 64, block_root: 0x690d…0e3e, slot: 40556
Jun 24 08:01:16.314 DEBG Ignoring gossip column already imported, data_column_index: 64, block_root: 0x690d4885fe843d085464b4f36b774597525d2dbddcd26ceca95fd4fdf4930e3e
Jun 24 08:01:16.323 DEBG Successfully verified gossip data column sidecar, index: 65, block_root: 0x690d…0e3e, slot: 40556
Jun 24 08:01:16.323 DEBG Ignoring gossip column already imported, data_column_index: 65, block_root: 0x690d4885fe843d085464b4f36b774597525d2dbddcd26ceca95fd4fdf4930e3e
Jun 24 08:01:16.332 DEBG Successfully verified gossip data column sidecar, index: 66, block_root: 0x690d…0e3e, slot: 40556
Jun 24 08:01:16.332 DEBG Ignoring gossip column already imported, data_column_index: 66, block_root: 0x690d4885fe843d085464b4f36b774597525d2dbddcd26ceca95fd4fdf4930e3e
Jun 24 08:01:16.341 DEBG Successfully verified gossip data column sidecar, index: 67, block_root: 0x690d…0e3e, slot: 40556
Jun 24 08:01:16.341 DEBG Ignoring gossip column already imported, data_column_index: 67, block_root: 0x690d4885fe843d085464b4f36b774597525d2dbddcd26ceca95fd4fdf4930e3e
Jun 24 08:01:16.349 DEBG Successfully verified gossip data column sidecar, index: 68, block_root: 0x690d…0e3e, slot: 40556
Jun 24 08:01:16.349 DEBG Ignoring gossip column already imported, data_column_index: 68, block_root: 0x690d4885fe843d085464b4f36b774597525d2dbddcd26ceca95fd4fdf4930e3e
Jun 24 08:01:16.358 DEBG Successfully verified gossip data column sidecar, index: 69, block_root: 0x690d…0e3e, slot: 40556
Jun 24 08:01:16.358 DEBG Ignoring gossip column already imported, data_column_index: 69, block_root: 0x690d4885fe843d085464b4f36b774597525d2dbddcd26ceca95fd4fdf4930e3e
Jun 24 08:01:16.367 DEBG Successfully verified gossip data column sidecar, index: 70, block_root: 0x690d…0e3e, slot: 40556
```


beacon_node/beacon_chain/src/block_verification_types.rs
```diff
@@ -605,18 +600,10 @@ impl<E: EthSpec> AsBlock<E> for AvailableBlock<E> {
             (Some(blobs), _) => RpcBlockInner::BlockAndBlobs(block, blobs),
             (_, Some(data_columns)) => RpcBlockInner::BlockAndCustodyColumns(
                 block,
-                RuntimeVariableList::new(
-                    data_columns
-                        .into_iter()
-                        // TODO(das): This is an ugly hack that should be removed. After updating
-                        // store types to handle custody data columns this should not be required.
-                        // It's okay-ish because available blocks must have all the required custody
-                        // columns.
```
dapplion, line 614: This comment still applies, the ugly hack here is about calling `from_asserted_custody`

dapplion: This code should go away after merging

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -854,16 +848,96 @@ impl<T: BeaconChainTypes> OverflowLRUCache<T> {
...
+                    pending_components_opt.map(|pending_components| {
+                        pending_components
+                            .verified_data_columns
+                            .iter()
+                            .map(|d| d.index())
+                            .collect::<HashSet<_>>()
+                    })
+                })
+            else {
+                // If block is already imported (no longer in cache), abort publishing data columns
```
dapplion, line 924: This assumes custody requirement = all columns. Should add a comment about this assumption with a TODO

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -854,16 +848,96 @@ impl<T: BeaconChainTypes> OverflowLRUCache<T> {
...
+            // Block may have been imported as it does not exist in availability cache.
+            return Ok(None);
+        };
+
+        let should_reconstruct = self
+            .block_import_requirement(&pending_components)
+            .map(|r| self.should_reconstruct(&r, &pending_components))?;
+
+        if should_reconstruct {
+            *self.reconstructing_block_root.lock() = Some(block_root);
```
dapplion: Need to set `reconstructing_block_root` to None at the end of this code block. If there are errors the lock will remain as Some<T> forever. Not sure if this is the best pattern to achieve this 1 item queue.

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -964,7 +988,8 @@ impl<T: BeaconChainTypes> OverflowLRUCache<T> {
         let has_missing_columns = pending_components.verified_data_columns.len() < num_of_columns;
 
         has_missing_columns
-            && !pending_components.reconstruction_started
+            // for simplicity now, we only reconstruct columns for one block at a time.
+            && self.reconstructing_block_root.lock().is_none()
```
dapplion: This silent skip is dangerous. We want to be intentional on when we NOT reconstruct and why. If the network experiences two data incomplete blocks close together, all supernodes could simultaneously hit this condition and not reconstruct.

If we really want to implement this "only 1 reconstruction at once" the right place to impose work limitation is the beacon processor. Here it should always go ahead (my preferred), wait (I won't recommend) or throw an error (also bad).

beacon_node/network/src/network_beacon_processor/mod.rs
```diff
@@ -775,6 +777,69 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                 "error" => %e)
         });
     }
+
+    /// Re-seed the network with reconstructed data columns.
+    fn handle_data_columns_to_publish(
```
dapplion: Maybe just inline below now that this function is quite simple

dapplion (changes requested):

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: Yeah good point - I didn't realized that when I made the change. Maybe we can stick that field back to the `PendingComponents`

beacon_node/network/src/network_beacon_processor/mod.rs
jimmygchen: Done

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: Done

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: Reverted `reconstructing_block_root` as suggested in the other comment

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: Done

beacon_node/beacon_chain/src/block_verification_types.rs
jimmygchen: Oops nice catch, added back

jimmygchen:
Thanks for the review! I've addressed the comments and will start local testing.

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -334,28 +334,27 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
-            }
+                AvailabilityProcessingStatus::MissingComponents(_, _) => {
+                    debug!(
+                        self.log,
+                        "Missing components over rpc";
+                        "block_hash" => %block_root,
+                    );
+                    // Attempt reconstruction here before notifying sync, to avoid sending out more requests
+                    // that we may no longer need.
+                    self.attempt_data_column_reconstruction(block_root).await;
```
dapplion: If reconstruction succeeds, you should send Imported to sync, not MissingComponents

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -1019,9 +994,7 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
             .process_gossip_data_columns(vec![verified_data_column])
             .await
         {
-            Ok((availability, data_columns_to_publish)) => {
-                self.handle_data_columns_to_publish(data_columns_to_publish);
-
+            Ok(availability) => {
```
dapplion, line 997: You could add a match here to re-assign result
```rust
let result = match result {
  MissingComponents => self.attempt_data_column_reconstruction(),
  other => other,
};
```
to re-use the handling of successful conditions. Same could apply to sync

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -362,6 +361,7 @@ impl<E: EthSpec> PendingComponents<E> {
         )))
     }
 
+    /// Mark reconstruction as started for this `PendingComponent`.
```
dapplion, line 367: With this approach reconstruction will only be attempted once. I think that's okay, but we should carefully track in networks with metrics or crit logs if reconstruction errors. Worth it to add a comment explaining this

dapplion (changes requested):

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
jimmygchen: Not quite sure what you mean, as we don't use the `availability` result after this, are you meaning to comment on sync? or do you want to extract the handling of successful condition to a method?

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: Yep this was the intention as the reconstruction code would currently only fail due to code errors or kzg errors, which probably shouldn't be retried.

Good point adding this to the comment! I'll add an error log and metrics to track the failures. Not sure if it should be CRIT here though. (something that would trigger an alert to people monitoring infrastructure)

beacon_node/network/src/network_beacon_processor/sync_methods.rs
jimmygchen: Nice catch! Fixed

jimmygchen:
Thanks for the review! Addressed your comments in https://github.com/sigp/lighthouse/pull/5990/commits/7d2d826a892f46dd965484a1ecc49401db57ba54.

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -1044,7 +1017,7 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                             "block_root" => %block_root,
                         );
 
-                        // Potentially trigger reconstruction
+                        self.attempt_data_column_reconstruction(block_root).await;
```
jimmygchen, line 1020: Looks like this doesn't work, issues:

1. Still seeing reconstruction blocking gossip data column processing
2. This change should be made to the cached block instead of the clone: https://github.com/sigp/lighthouse/blob/7d2d826a892f46dd965484a1ecc49401db57ba54/beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs#L885
3. Because of 1., after the block is imported, I see a bunch of:
```
Jul 01 06:27:21.130 DEBG Attempted to publish duplicate message  kind: data_column_sidecar_31, service: libp2p
```

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
dapplion: Right, for `process_gossip_verified_data_column` it's not that useful and the function is simple. The way you mutate `result` now should be fine

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
jimmygchen: Fixed #2.

I can't seem to figure out why reconstruction is blocking the processing of other incoming data columns. I don't see any lock being held during reconstruction, perhaps it is just waiting for a beacon processor worker to become available.

We could send it to the back of the beacon processor queue, but I'll do some more testing to make sure this is the case.

dapplion:
Generally looks good to me! Lots of conflicts, tho

Do you want to port this to unstable?

jimmygchen:
Thanks, I've just tried to resolve conflicts, but unfortunately I see this is going to conflict a lot with #6268, and it might be easier to implement it once that one is merged, as it has quite a bit of common logic and changes. Since this one is very outdated, I think it make sense to just create a new PR, using the branch from #6268 as base branch.

jimmygchen:
Leaving this PR open so we don't forget this.

michaelsproul:

jimmygchen:
Closing in favour of #6403

PR #5992, opened 2024-06-25
Avoid rayon in lighthouse block verification

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

The blst library's `verify_multiple_aggregate_signatures` function [already](https://github.com/supranational/blst/blob/afd60c5f689ef11863852353dd394265a0c37863/bindings/rust/src/lib.rs#L1234-L1239) spreads the verification across multiple cores, we are doing redundant parallelization in lighthouse. 
This PR just calls the blst function directly without doing any additional parralelization. 

I'm guessing we did this when we were also using milagro bls as it presumably didn't parallelize, but now that we exclusively use blst, this is not needed anymore.

## Additional info

This PR need not necessarily be in 5.2.1, we can get some additional long term testing for this to check it doesn't result in performance degradation somehow.


dapplion (approved):

michaelsproul (approved):
Rayon getting deleted. I love to see it.

Agree we should just merge this to unstable rather than fitting it into 5.2.1

michaelsproul:
@mergifyio queue

michaelsproul:
@mergifyio queue

jimmygchen:
@mergify requeue

PR #5989, opened 2024-06-25
Release v5.2.1 (testing branch)

michaelsproul:
## Proposed Changes

Testing PR for v5.2.1

This will be eventually merged into `release-v5.2.1`, which will then be merged to `stable`.

This PR is blocked on:

- https://github.com/sigp/lighthouse/pull/5969
- #5993

PR #5993, opened 2024-06-25
Attempt to continue lookups after adding peers

dapplion:
## Issue Addressed

Testing v5.2.1 a node hit a corner case of lookup sync causing a lookup to get stuck:

The lookup was originally created from an unknown parent block event with a block as child components. Therefore the original state of each request was:
- block: AwaitingProcess
- blobs: AwaitingDownload

When the parent imported successfully, it got called lookup `continue_request`. For each request:
- block: request is in AwaitingProcess so the block was sent to processing
- blobs: request is AwaitingDownload but the lookup has 0 peers so it doesn't make progress

https://github.com/sigp/lighthouse/blob/f1d88ba4b1256e084d3a2a547e0ce574b896246c/beacon_node/network/src/sync/block_lookups/single_block_lookup.rs#L202

When the block completes processing and it gets the MissingComponents, `continue_request` is called again, but the lookup still has 0 peers. So no progress is made.

Then immediately after that attempt at `continue_request` the lookup gets peers. When the `drop_lookups_without_peers` function runs, the lookup has peers but nothing is trying to make progress on it.

It appears the comment below is too optmistic. Yes this situation is a rare case but it results in lookups getting stuck. We should attempt to make progress on lookups when adding peers

https://github.com/sigp/lighthouse/blob/f1d88ba4b1256e084d3a2a547e0ce574b896246c/beacon_node/network/src/sync/block_lookups/mod.rs#L852-L856

## Proposed Changes

Attempt to make progress on existing lookups that are not awaiting parent and got new peers

### DIFF
--- a/beacon_node/network/src/sync/block_lookups/mod.rs
+++ b/beacon_node/network/src/sync/block_lookups/mod.rs
@@ -269,7 +269,7 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
                 }
             }
 
-            if let Err(e) = self.add_peers_to_lookup_and_ancestors(lookup_id, peers) {
+            if let Err(e) = self.add_peers_to_lookup_and_ancestors(lookup_id, peers, cx) {
                 warn!(self.log, "Error adding peers to ancestor lookup"; "error" => ?e);
             }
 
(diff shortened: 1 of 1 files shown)


realbigsean (approved):
LGTM

pawanjay176 (approved):
Nice

jimmygchen (approved):
Looks good to me too! 
It's indeed a rare case (wasn't easy for me to reproduce) and seem to happen on only few overwhelmed node mostly.

jimmygchen:
I'm merging this into the [testing branch](https://github.com/sigp/lighthouse/pull/5989). Please do not merge to release branch yet.

jimmygchen:
@mergify queue

Issue #5970, opened 2024-06-20
v5.2.0 built from source runs out of memory


alecalve:
## Description

We build a Docker image from the Lighthouse v5.2.0 `Dockerfile` with very minor changes:

- adding `curl` inside the image directly
- some changes relating to how the binary is run

The binary itself is built the same way as in this repo's Dockerfile, using `make` and the default values for `FEATURES`, `PROFILE`, etc..

```
diff --git a/Dockerfile b/Dockerfile
index e0c48699b..da033a4bf 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,21 @@ FROM ubuntu:22.04
 RUN apt-get update && apt-get -y upgrade && apt-get install -y --no-install-recommends \
   libssl-dev \
   ca-certificates \
-  && apt-get clean \
-  && rm -rf /var/lib/apt/lists/*
+  curl && \
+  apt-get clean && \
+  rm -rf /var/lib/apt/lists/*
+
 COPY --from=builder /usr/local/cargo/bin/lighthouse /usr/local/bin/lighthouse
+RUN useradd -m -u 1000 -s /bin/bash lighthouse
+RUN chown -R lighthouse:lighthouse /opt
+USER lighthouse
+RUN mkdir /opt/data
+WORKDIR /opt/lighthouse
+
+# Firewall configurations
+# P2P TCP, UDP
+EXPOSE 9001/TCP
+EXPOSE 9001/UDP
+# RPC
+EXPOSE 5053
+ENTRYPOINT [ "lighthouse" ]
```

The node is run using these arguments:

```
      --datadir=/opt/data
      --http
      --listen-address=0.0.0.0
      --listen-address=::
      --http-address=::
      --http-allow-origin=*
      --http-port=5052
      --target-peers=50
      --slots-per-restore-point=32
      --prune-blobs=false
      --execution-jwt=/secrets/jwtsecret
      --execution-endpoint=http://....:8551
      --port=9000
      --port6=9000
      --network=mainnet
      --disable-deposit-contract-sync
      --state-cache-size=2
      --metrics
      --metrics-address=::
      --metrics-port=8080
      --metrics-allow-origin=*
```

## Version

The Docker image is built from `v5.2.0`

## Present Behaviour

Once started, the application runs then finally runs out of memory after hitting the limit of 64GB we assigned to it.

This did not happen with the same modifications applied to previous versions of Lighthouse. The last one we had tested this with was `v5.0.0`.

[...]

PR #5995, opened 2024-06-26
Enable jemalloc by default on non windows targets

jimmygchen:
## Issue Addressed

A few discord users that build Lighthouse from source has reported high RAM usage since v5.2.0. @michaelsproul and @chong-he [have found](https://github.com/sigp/lighthouse/issues/5970#issuecomment-2190216812) that this is due to `jemalloc` feature being disabled. This is now necessary (on linux) with the introduction of `tree-states` and should always be enabled.

The Dockerfile doesn't have a default for `FEATURES`:
https://github.com/sigp/lighthouse/blob/f1d88ba4b1256e084d3a2a547e0ce574b896246c/Dockerfile#L4-L7

From @michaelsproul 
> it sets it as an _argument_, which I think means if you don't do `--arg FEATURES=jemalloc` it will build with `FEATURES=""`

This PR removes the feature flag and enables jemalloc by default on non windows targets, even when it's not specified via `FEATURES` environment variable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,14 +14,6 @@ BUILD_PATH_AARCH64 = "target/$(AARCH64_TAG)/release"
 PINNED_NIGHTLY ?= nightly
 CLIPPY_PINNED_NIGHTLY=nightly-2022-05-19
 
-# List of features to use when building natively. Can be overridden via the environment.
-# No jemalloc on Windows
-ifeq ($(OS),Windows_NT)
-    FEATURES?=
-else
-    FEATURES?=jemalloc
-endif
-
 # List of features to use when cross-compiling. Can be overridden via the environment.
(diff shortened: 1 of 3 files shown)


jimmygchen:
Oops looks like this doesn't work in windows, I'll look into it.

michaelsproul:
With the `jemalloc` feature deprecated, we should update the `allocator_name()` function used to show the allocator in `lighthouse --version` to use the same logic (display "system" on Windows and "jemalloc" on everything else):

https://github.com/sigp/lighthouse/blob/f1d88ba4b1256e084d3a2a547e0ce574b896246c/lighthouse/src/main.rs#L53-L59

jimmygchen:
Nice catch, updated!

michaelsproul (approved):

michaelsproul:
@mergifyio queue

PR #5998, opened 2024-06-26
chore: add `proof` parameter to `recover_cells_and_kzg_proofs`

kevaundray:
## Issue Addressed

To recover the cells and proofs, we do not need the proofs that correspond to the cells we already have. As of now the DAS branch recovers_cells_and_proofs using three methods `recover_cells`, `cells_to_blob`, `compute_cells_and_proofs`. 

The consensus-specs has introduced a new method which encapsulates all three of these methods, called `recover_cells_and_proofs`. This method takes in a proof parameter and there are additional checks that libraries that implement that function will perform. This means that the `proofs` parameter cannot be missing or empty. As an example, libraries will check that the number of proofs is equal to the number of cells.

We will eventually remove the proof parameter here: https://github.com/ethereum/consensus-specs/pull/3819 but until then it is needed.

## Proposed Changes

In reconstruct, we also fetch the proof for the given cell. 

The specs do it by forming an ExtendedMatrix, however most codebases seem to just have PartialDataColumnSideCars instead, so was not able to corroborate with the specs

## Additional Info

### DIFF
--- a/consensus/types/src/data_column_sidecar.rs
+++ b/consensus/types/src/data_column_sidecar.rs
@@ -215,6 +215,7 @@ impl<E: EthSpec> DataColumnSidecar<E> {
             .map(|row_index| {
                 let mut cells: Vec<KzgCell> = vec![];
                 let mut cell_ids: Vec<u64> = vec![];
+                let mut proofs = vec![];
                 for data_column in data_columns {
                     let cell = data_column.column.get(row_index).ok_or(
                         KzgError::InconsistentArrayLength(format!(
@@ -224,8 +225,13 @@ impl<E: EthSpec> DataColumnSidecar<E> {
 
                     cells.push(ssz_cell_to_crypto_cell::<E>(cell)?);
                     cell_ids.push(data_column.index);
+                    proofs.push(*data_column.kzg_proofs.get(row_index).ok_or(
+                    KzgError::InconsistentArrayLength(format!(
+                        "Missing data column proof at index {row_index}"
+                    )),
+                    )?);
                 }
-                kzg.recover_cells_and_compute_kzg_proofs(&cell_ids, &cells)
+                kzg.recover_cells_and_compute_kzg_proofs(&cell_ids, &cells, proofs.as_slice())
             })
             .collect::<Result<Vec<_>, KzgError>>()?;
 
--- a/consensus/types/src/data_column_sidecar.rs
+++ b/consensus/types/src/data_column_sidecar.rs
@@ -226,9 +226,9 @@ impl<E: EthSpec> DataColumnSidecar<E> {
                     cells.push(ssz_cell_to_crypto_cell::<E>(cell)?);
                     cell_ids.push(data_column.index);
                     proofs.push(*data_column.kzg_proofs.get(row_index).ok_or(
-                    KzgError::InconsistentArrayLength(format!(
-                        "Missing data column proof at index {row_index}"
-                    )),
+                        KzgError::InconsistentArrayLength(format!(
+                            "Missing data column proof at index {row_index}"
+                        )),
                     )?);
                 }
                 kzg.recover_cells_and_compute_kzg_proofs(&cell_ids, &cells, proofs.as_slice())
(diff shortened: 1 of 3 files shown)


kevaundray:
Updating c-kzg should also surface the errors in the ef-tests and on devnet. In particular, it should refuse to make proofs and unconditionally return an error.

kevaundray:
Check here for the `recover_cells_and_kzg_proofs` method in consensus-specs to corroborate that it now takes a proof parameter: https://github.com/ethereum/consensus-specs/blob/dev/specs/_features/eip7594/polynomial-commitments-sampling.md#recover_cells_and_kzg_proofs

kevaundray:
Note: I have not updated c-kzg in this PR because there are other breaking changes that were added that we would need to modify the codebase for, which would be out of scope, this is just to align the API with the consensus-spec-tests while ensuring nothing breaks.

consensus/types/src/data_column_sidecar.rs
```diff
@@ -224,8 +225,13 @@ impl<E: EthSpec> DataColumnSidecar<E> {
 
                     cells.push(ssz_cell_to_crypto_cell::<E>(cell)?);
                     cell_ids.push(data_column.index);
+                    proofs.push(*data_column.kzg_proofs.get(row_index).ok_or(
+                        KzgError::InconsistentArrayLength(format!(
+                            "Missing data column proof at index {row_index}"
+                        )),
+                    )?);
```
kevaundray, line 232: This is sort of the important part of this PR -- ie ensuring that this proof is the proof for the given cell

testing/ef_tests/src/cases/kzg_recover_cells_and_kzg_proofs.rs
```diff
@@ -62,6 +62,7 @@ impl<E: EthSpec> Case for KZGRecoverCellsAndKZGProofs<E> {
                     .recover_cells_and_compute_kzg_proofs(
                         cell_indices.as_slice(),
                         input_cells.as_slice(),
+                        input_proofs.as_slice(),
```
jimmygchen, line 65: FyI the spec tests are currently passing even without this because we do a workaround to compare the output with the inputs, to satisfy the current spec tests - should be able to remove that altogether once the spec PR is merged.

https://github.com/sigp/lighthouse/pull/5998/files#diff-910a7a6a592415f79d81b45a2f8ab500b09ce31bb2e69c85af2bbb1c125a321dR73-R83

kevaundray: Would need to wait for a spec release too I think?

An alternative is to tentatively modify the cryptography code and remove all of test vectors that specifically involve invalid proofs -- would you prefer this?

jimmygchen: I don't really know if we need these changes - If I understand your explanation correctly, the PR is to make the `recover_cells_and_compute_kzg_proofs` api match the current consensus spec _temporarily_ - given that the proofs parameter is likely going to be removed in the next spec release - and we don't actually use the proofs array at all. So it seems like we can avoid this change so we don't have to worry about removing it later, right? It doesn't actually change what we're testing in the spec tests.  Or am I missing something?

kevaundray: That's correct -- the main issue is that the cryptography implementations are following the specs so they will fail if not given the proofs

jimmygchen: I see. The `proofs` is not required in the current version of c-kzg used on `das` branch. Is this change to prepare support for newer version of c-kzg / or the rust peerdasKzg library?

kevaundray: That branch uses an older version of c-kzg, before this spec change was added

kevaundray: It got removed yesterday from c-kzg following that spec test PR, so spec tests will be failing in c-kzg until we merge that consensus-spec PR: 

https://github.com/ethereum/c-kzg-4844/commit/9af09715877f74e6966210283153aa73bd1ba41f

I can do the same in peerdas-kzg

kevaundray:
Closing and making the cryptography library ignore the test vectors involving proofs instead

PR #6000, opened 2024-06-26
Fix data column by range rpc request not being sent due to incorrect limits set

jimmygchen:
## Issue Addressed

Fix data column RPC request not being sent / accepted due to incorrect RPC limits set.


dapplion (approved):
Nice find!

PR #6004, opened 2024-06-27
Various fixes to custody range sync 

jimmygchen:
## Issue Addressed

The current behaviour is the syncing node will start sending batches on startup, and if there isn't any peers on custody column subnets, the `DataColumnByRange` requests will return an error immediately, triggering another retry. And this may cause the chain to be removed and peers to disconnect shortly.

This PR addresses the above and a few other sync related issues: 
- Only request batches when there are good peers across all custody subnets. 
- Use subscribed peers on subnets before [MetaDataV3](https://github.com/ethereum/consensus-specs/pull/3821) is implemented as it's difficult to get custody peers on small devnets without identifying supernodes. 
- Remove `peer_id` matching when injecting error because multiple peers are used for range requests. This prevents sync from getting stuck.
- Use randomized custodial peer to avoid repeatedly sending requests to failing peers. 
- Batch by range request where possible


jimmygchen:
Looks like there are still a few more issues - most likely the issue is that we're assuming enr is always available, will look into that tomorrow.

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -1018,6 +1030,30 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
...
+                .network_globals()
+                .custody_subnets(&network.chain.spec)
+                .all(|subnet_id| {
+                    let peer_count = network
+                        .network_globals()
+                        .peers
+                        .read()
+                        .good_peers_on_subnet(Subnet::DataColumn(subnet_id))
+                        .count();
+                    debug!(self.log, "Peers found on custody subnet"; "subnet_id" => ?subnet_id, "peer_count" => peer_count);
```
dapplion: This log is gonna be very noisy, instead we should have a metric: gauge `peers_per_column_subnet` with label "subnet_id"

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -396,6 +396,9 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
                     self.request_batches(network)?;
                 }
             }
+        } else if !self.good_peers_on_custody_subnets(self.processing_target, network) {
+            // If there's no good custody peers for this epoch, batch won't be created
+            return Ok(KeepChain);
```
dapplion, line 425: Why is this condition necessary?

dapplion (commented):
> The redundant rotation epoch parameter is also removed (sorry probably should've kept this separate, wasn't expecting it to change many files).

Can you split in a separate PR? The main change is very critical and it can brick sync so best to review on its own

beacon_node/network/src/sync/range_sync/chain.rs
jimmygchen: This is the key change of this PR.

The current behaviour on startup:
- `send_batch` sends both range requests
- Column range request may fail immediately with `NoCustodyPeers`
- It will then immediately retry with `send_batch`, spamming `BlocksByRange`requests to peers
- At some point it probably either reached its attempt limit or get rate limited by peer.

This change prevent sending out range requests until we have at least one peer across all custody subnets.

This is now working on a local devnet and we're able to range sync from a Lighthouse supernode 🎉

jimmygchen: There are a few things to fix though:
- We're sampling too fast during sync - we _could_ bump up the limit for now, ~~but we have no throttling at all~~ (we have self limiter), and I just see logs scrolling through so we may be DOSing the peer. I'll have to check the self rate limiter config to fix this.
- `good_peers_on_custody_subnets` implementation can be improved to include fallbcaks and using MetaData, but I think the current version would be fine to roll out.

I think once we fix the sampling issue, it would be great to deploy to devnet, and we can continue to iterate from there.

jimmygchen: Oh sorry I thought this was the other KeepChain!

This is to handle the case where no batches was sent for the current processing target, which ends up in the error below.

This is now a valid state when we have no custody peers, but we still want to keep the existing condition to return an error if no batch was sent for some other reasons.

jimmygchen: Updated code comment in ede964ee

beacon_node/network/src/sync/range_sync/chain.rs
jimmygchen: Updated in ede964ee

beacon_node/network/src/sync/network_context.rs
```diff
@@ -406,6 +381,38 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+    fn make_columns_by_range_requests(
+        &self,
+        epoch: Epoch,
+        request: BlocksByRangeRequest,
+        custody_indexes: &Vec<ColumnIndex>,
+    ) -> Result<HashMap<PeerId, DataColumnsByRangeRequest>, RpcRequestSendError> {
+        let mut peer_id_to_request_map = HashMap::new();
+
+        for column_index in custody_indexes {
+            let Some(custody_peer) = self.get_random_custodial_peer(epoch, *column_index) else {
```
jimmygchen, line 443: There's potentially an issue with the peer selection approach here - as peer selection was previously done one layer up:
- [`request_batches`](https://github.com/sigp/lighthouse/blob/022b848cfb86e0e949aaddb0429a2b6206b1c9ea/beacon_node/network/src/sync/range_sync/chain.rs#L987-L999): here we use a list of "idle peers" and randomly select one
- [`retry_batch_download`](https://github.com/sigp/lighthouse/blob/022b848cfb86e0e949aaddb0429a2b6206b1c9ea/beacon_node/network/src/sync/range_sync/chain.rs#L864-L875): here we sort peers prioritizing unrelated peers with less active requests

Are these selection rules important? I've ran into some issues with concurrent requests to the same peer with the approach on this branch, and need to investigate further.

jimmygchen: Checked with Age, and it should be fine to send requests to peers that aren't "idle".

jimmygchen: The peer selection logic here isn't ideal - we should probably avoid retrying peers that failed to return a response. Perhaps we could pass the sorted peer list into this method?

jimmygchen: Added a comment for this, probably requires some thoughts and refactoring as we don't currently track `data_column_by_range` peers in `BatchState`.

jimmygchen: Note to self: we might also need to apply some of these changes to backfill sync.

dapplion (approved):

PR #6005, opened 2024-06-27
Update pending balance deposit processing

michaelsproul:
## Proposed Changes

Update the pending deposit processing for the changes in spec `v1.5.0-alpha.3`. These changes likely come with a bit of a performance impact, but I'm not too worried about optimising this code path yet, as it is likely to change again:

- https://github.com/ethereum/consensus-specs/pull/3776#issuecomment-2186038176

### DIFF
--- a/consensus/state_processing/src/per_epoch_processing/single_pass.rs
+++ b/consensus/state_processing/src/per_epoch_processing/single_pass.rs
@@ -831,6 +831,9 @@ impl PendingBalanceDepositsContext {
             // `true` both for the actual value & the default placeholder value (`FAR_FUTURE_EPOCH`).
             let validator = state.get_validator(deposit.index as usize)?;
             let already_exited = validator.exit_epoch < spec.far_future_epoch;
+            // In the spec process_registry_updates is called before process_pending_balance_deposits
+            // so we must account for process_registry_updates ejecting the validator for low balance
+            // and setting the exit_epoch to < far_future_epoch
             let will_be_exited = validator.is_active_at(current_epoch)
                 && validator.effective_balance <= spec.ejection_balance;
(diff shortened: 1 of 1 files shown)


consensus/state_processing/src/per_epoch_processing/single_pass.rs
```diff
@@ -805,22 +810,54 @@ impl PendingBalanceDepositsContext {
...
+            // *after* changes applied by `process_registry_updates`, which in our implementation
+            // does not happen until after this (but in the spec happens before). However it's not
+            // hard to work out: we don't need to know exactly what value the `exit_epoch` will
+            // take, just whether it is non-default. Nor do we need to know the value of
+            // `withdrawable_epoch`, because `current_epoch <= withdrawable_epoch` will evaluate to
+            // `true` both for the actual value & the default placeholder value (`FAR_FUTURE_EPOCH`).
+            let validator = state.get_validator(deposit.index as usize)?;
+            let already_exited = validator.exit_epoch < spec.far_future_epoch;
+            let will_be_exited = validator.is_active_at(current_epoch)
+                && validator.effective_balance <= spec.ejection_balance;
```
dapplion, line 838: Add a comment, like
```
            // In the spec process_registry_updates is called before process_pending_balance_deposits
            // so we must account for the process_registry_updates ejecting the validator for low balance
            // and setting the exit_epoch to < far_future_epoch
```

dapplion (approved):
Change looks correct to me 👍  we should process very few deposits in practice, so no performance concerns, 16 ops vs 1,400,000 indexes

ethDreamer:
Can we get this merged soon? Then we can have a branch ready for `devnet-1`.

dapplion (approved):

PR #6008, opened 2024-06-27
Prevent connections from peers with a banned ip history

AgeManning:
This synchronizes the logic we have between what peers we allow to connect to us and what think are connecting to us. 

Specifically, peers that have a history of connecting to us via an IP address that is currently banned (i.e has a number of peers currently banned associated with this IP) we prevent their connection, even if they are connecting on a new IP address. 

This should resolve some potential error logs also.


jimmygchen (approved):
Nice simple fix!

jimmygchen:
@mergify queue

PR #4515, opened 2023-07-18
Use histograms for `BeaconProcessor` queue lengths

paulhauner:
## Issue Addressed

NA

## Proposed Changes

Uses a Prometheus [historgram](https://prometheus.io/docs/practices/histograms/) to track the lengths of queues in the beacon processor.

We're currently using a gauge, which ends up reporting the queue length at exactly the time of metrics scraping. This is not so useful since it doesn't capture peaks/troughs between scrapes. A histogram will give us more of an average over time (but in a discrete, bucket-y sorta way).

## Additional Info

I've removed the `BEACON_PROCESSOR_<OBJECT>_QUEUE_TOTAL` and `BEACON_PROCESSOR_REPROCESSING_QUEUE_TOTAL` metrics. I was in two minds about removing/renaming the old metrics or just changing their underlying data type. I settled on removing/renaming since it should play nicer with setups that have a mix of old/new versions of Lighthouse.

Blocked on #4462.

PR #6012, opened 2024-06-27
Register processor queue length as histogram

dapplion:
## Issue Addressed

Second attempt at
- https://github.com/sigp/lighthouse/pull/4515

It would be very useful to know how often queues reach zero or very high capacity values. Today we only have sparse snapshots of the queue length, which can change very rapidly in miliseconds

## Proposed Changes

Register the queue length as histogram instead of a gauge.

This PR registers the length metric for a single queue instead of for all queues. Reasons why:
- We don't want to register the same value over and over. It would make the metric less useful and be quite wasteful
- Using an enum ensures that we register metrics for all queues. With the current approach it's easy to forget to add metrics for new queues: this has happened when adding lightclient messages and with peerdas.

## Metrics breaking changes

All metrics previously used to track queue lengths change. Also the type label for all metrics in the processor changes capitalization from `gossip_block` to `GossipBlock`. Is that okay? Does the `IntoStaticStr` macro support a different capitalization?


beacon_node/beacon_processor/src/lib.rs
```diff
@@ -615,47 +588,94 @@ pub enum Work<E: EthSpec> {
 
 impl<E: EthSpec> fmt::Debug for Work<E> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{}", self.str_id())
+        write!(f, "{}", Into::<&'static str>::into(self.to_type()))
     }
 }
 
+#[derive(IntoStaticStr, PartialEq, Eq, Debug)]
```
michaelsproul, line 624: You can use `#[strum(serialize_all = "snake_case")]` to keep the snake case identifiers

https://docs.rs/strum/latest/strum/additional_attributes/index.html

beacon_node/beacon_processor/src/lib.rs
```diff
@@ -987,46 +1008,48 @@ impl<E: EthSpec> BeaconProcessor<E> {
                     .map_or(false, |event| event.drop_during_sync);
 
                 let idle_tx = idle_tx.clone();
-                match work_event {
+                let modified_queue_id = match work_event {
```
michaelsproul, line 1011: why is this variable called _modified_ queue ID? it never gets modified

michaelsproul (approved):
LGTM modulo my query about variable naming

I would like to merge this ASAP so it doesn't conflict with:

- https://github.com/sigp/lighthouse/issues/6291

beacon_node/beacon_processor/src/lib.rs
dapplion: This refers to what queue has been mutated by a pop operation. Tells downstream code "the queue associated with this queue ID has been mutated, so register its length"

michaelsproul: oh yeah I get you, makes sense

let's do it

michaelsproul:
@mergify queue

jimmygchen:
@mergify requeue

PR #618, opened 2019-11-21
Validator client refactor

paulhauner:
## Issue Addressed

- Resolves #604
- Resolves #342
- Resolves #336

## Proposed Changes

- Modifies `account_manager` to suit the deposit flow in #604.
- Refactors the validator client to be more use http and be more async.
- Removes gRPC project-wide.
- Adds a maximum skip slots of 128. Without this, it causes lots of problems with high validator counts and skip slots. HTTP requests start to pile up.
- Adds a snazzy new github actions CI.

PR #6014, opened 2024-06-27
Drop skip too large condition for liveness safety

dapplion:
## Issue Addressed

Chating with @michaelsproul he noted that Lighthouse still includes a condition that could jeopardize the liveness of the chain under extreme circumstances.

If some codepath (for example block production) needs to advance a state multiple epochs such that the total runtime of that advance is larger than `seconds_per_slot`, it will just error. If this condition happens the node will be stuck, unable to propose blocks on the chain and making the problem worse. If enough nodes follow this rule, the network will never progress and stall. The motivation for this error is to protect the node against very expensive state advance requests, but these are necessary for liveness.

For context, the condition was introduced in this PR
- https://github.com/sigp/lighthouse/pull/618

I believe we should drop the condition now as is. In the future, we may choose to harden the node against corner cases where the network is in a problematic state, such as having no head in the last N epochs.

## Proposed Changes

Drop skip too large condition for liveness safety

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -114,7 +114,7 @@ use std::collections::HashSet;
 use std::io::prelude::*;
 use std::marker::PhantomData;
 use std::sync::Arc;
-use std::time::{Duration, Instant};
+use std::time::Duration;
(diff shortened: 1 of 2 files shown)


michaelsproul (approved):
Agree that this condition needs to go

michaelsproul:
@mergify queue

PR #6027, opened 2024-07-01
Delete duplicated serde code

michaelsproul:
## Proposed Changes

Found some serde code for `IndexedAttestation` that is just copy-pasta of stuff we already have in `ethereum_serde_utils`. I think it actually predates that module though, and the duplication just never got cleaned up (until now).

The new code also supports omitting quotes during deserialisation, so there is no breaking change here: https://docs.rs/ssz_types/latest/ssz_types/serde_utils/quoted_u64_var_list/index.html


realbigsean (approved):
nice!

realbigsean:
@mergify queue

PR #6028, opened 2024-07-01
Consolidate non-configurable constants from chainspec

### LINKED ISSUE #5631: Constant handling in `ChainSpec`
## Description

`ChainSpec` is made up of the "Configs" and "Presets" from the consensus spec, with some, but not all "Constants". Should we consolidate constants? I'd propose moving them to `consensus/types/consts.rs`. We could potentially avoid passing `ChainSpec` around places we shouldn't need it.

pawanjay176:
## Issue Addressed

Resolves #5631 

## Proposed Changes

We currently have a lot of spec constants listed under lighthouse's ChainSpec struct. Values in chainspec are meant to be configurable from a given config file. Hence, there is no point in having spec defined constants under chainspec.

This PR moves all values defined as constant in the spec to types/consts.rs. 
This is more consistent imo and also reduces the need to pass a `ChainSpec` in many places across the code.
Most of the diff is renames from `spec.x` to `X` and importing the required constants.

## Additional Info

This specific test is a confusing case. https://github.com/sigp/lighthouse/blob/70bcba1e6b7a7123b2f48e1c77b9afb9bbb11c26/consensus/types/src/beacon_state/iter.rs#L118

We are setting a custom value for genesis_slot which is defined as a constant in the [spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#misc) 
I don't think this is required behaviour from lighthouse, not completely sure though.

Happy to delay this until electra and peerdas branches are merged.


dapplion:
Conflicts 🙌

consensus/types/src/beacon_state/iter.rs
```diff
@@ -108,46 +109,47 @@ mod test {
...
-        );
-
-        *state.slot_mut() = Slot::new(6);
-        assert_eq!(
-            all_roots(&state, &spec),
-            vec![root_slot(5), root_slot(4)],
-            "second slot after genesis has two slot history"
-        );
-    }
+    // TODO(do we need this test?)
```
realbigsean, line 118: I think you can revive this test if instead of using `all_roots` you directly contruct an iter: `BlockRootsIter::new(self, GENESIS_SLOT)`

consensus/types/src/consts.rs
```diff
@@ -1,3 +1,40 @@
+pub mod phase0 {
```
realbigsean, line 1: Looks like `INTERVALS_PER_SLOT` is actually phase0/

Also `TARGET_AGGREGATORS_PER_COMMITTEE` looks like it was missed in the migration here

consensus/types/src/config_and_preset.rs
```diff
@@ -99,34 +102,34 @@ pub fn get_extra_fields(spec: &ChainSpec) -> HashMap<String, Value> {
...
+        "bls_withdrawal_prefix".to_uppercase() => u8_hex(BLS_WITHDRAWAL_PREFIX),
+        "eth1_address_withdrawal_prefix".to_uppercase() => u8_hex(ETH1_ADDRESS_WITHDRAWAL_PREFIX),
+        "domain_beacon_proposer".to_uppercase() => u32_hex(DOMAIN_BEACON_PROPOSER),
+        "domain_beacon_attester".to_uppercase() => u32_hex(DOMAIN_BEACON_ATTESTER),
+        "domain_randao".to_uppercase()=> u32_hex(DOMAIN_RANDAO),
+        "domain_deposit".to_uppercase()=> u32_hex(DOMAIN_DEPOSIT),
+        "domain_voluntary_exit".to_uppercase() => u32_hex(DOMAIN_VOLUNTARY_EXIT),
+        "domain_selection_proof".to_uppercase() => u32_hex(DOMAIN_SELECTION_PROOF),
+        "domain_aggregate_and_proof".to_uppercase() => u32_hex(DOMAIN_AGGREGATE_AND_PROOF),
+        "domain_application_mask".to_uppercase()=> u32_hex(APPLICATION_DOMAIN_BUILDER),
```
realbigsean, line 114: I think we should use 2 separate constants for the mask vs the builder domain even though `DOMAIN_APPLICATION_MASK == APPLICATION_DOMAIN_BUILDER`. probably better to make a specific `DOMAIN_APPLICATION_MASK` constant available

realbigsean (changes requested):
Love how much `ChainSpec` passing we can get rid of

consensus/types/src/consts.rs
pawanjay176: Seems like we play with the value of `target_aggregators_per_committee` for easier testing.
https://github.com/sigp/lighthouse/blob/unstable/beacon_node/beacon_chain/src/test_utils.rs#L173

I'm a little hesitant to do this PR if we are not moving all constants. Practically, its just a minor annoyance so I'm fine either way. 
What do you think? cc @michaelsproul

michaelsproul: Hmm, target aggregators per committee probably needs to stay tweak-able

pawanjay176:
Closing this as the advantages don't seem that great and we'll loose some nice testing setups

PR #6029, opened 2024-07-01
Check that `KzgCommitments` are consistent across all sidecars in batch during verification

jimmygchen:
## Issue Addressed

Range data column responses are failing due to `KzgVerificationFailed`, looks like there's a bug somewhere in batch verifying Kzg. 

I've added a test and fixed a logic error when converting `kzg_commitments` - we **MUST** only send the one set of `kzg_commitments` to CKZG instead of a flatmap of all `kzg_commitments` - they are required to be the same.

**NOTE**: I can't reproduce the error anymore, but I think this change is necessary. See [comment here](https://github.com/sigp/lighthouse/pull/6029#issuecomment-2205216913).


kevaundray:
I don't understand the issue, though I'll explain the interface for bacth verification:

The commitments vector can be seen as a deduplicated list of KZGCommitments. (It shouldn't need to be)

The rowIndices can be seen as a pointer into the commitments vector.

This means that if you want to verify 100 cells/samples for the same blob. You would have 1 commitment because there is only one blob, you would have 100 row indices, all being 0 because they are pointing to the first element in the commitments vector. 100 cells, and 100 proofs.

The main thing is that rowIndices, is a pointer into the commitments vector. So with the example above, you could have 100 commitments, and rowIndices would just go from 0 too 99, or it could be all zeroes since the commitments vector contains all of the same elements.

kevaundray:
Last week a bug arose because of this, so I have been thinking about updating the specs, so that its one commitment, one proof, one cell. This would also mean we get rid of rowIndices.

jimmygchen:
The explanation is useful, thanks Kev!

I was reading through the code yesterday and I _think_ this confirms what I have understood so far.

So I don't think I've fixed the problem here, but the `kzg_commitments` uniqueness check seems like an important check to have, so we don't accidentally pass verification for invalid kzg commitments proivided within a batch.

Consider the following batch for a block containing a single blob, the current `das` branch supplies the following to the kzg lib:

```
cells: [cell_1, cell_2, cell_3]
proofs: [proof_1, proof_2, proof_3]
row_indices: [0, 0, 0]
column_indices: [0, 1, 1]
commitments: [kzg_commitment_1, kzg_commitment_1, kzg_commitment_1]
```

kzg lib only uses the first commitment in the vector for verficiaton, and ignores the rest. 

This PR changes the arguments provided to the kzg lib to the following after verifying that all `kzg_commitments` vector are the same within the batch, otherwise returns a `InconsistentKzgCommitments` error:

```
cells: [cell_1, cell_2, cell_3]
proofs: [proof_1, proof_2, proof_3]
row_indices: [0, 0, 0]
column_indices: [0, 1, 1]
commitments: [kzg_commitment_1]
```

jimmygchen:
I can't reproduce this anymore, although I think it's worth keeping this check on the Lighthouse side until the interface is updated, as the library ignores the duplicated kzg commitments that could potentially be invalid.

asn-d6:
Interesting. I guess our API design was:
```
# All commitments should be unique, and they should be as many as the unique row indices
assert len(commitments) == len(set(commitments)) == len(set(row_indices)) 
for row_index in row_indices:
  commitment = commitments[row_index]
```
whereas you thought it would be:
```
# There should be as many commitments as cells
assert len(commitments) == len(cells)
for i in range(cells):
  commitment = commitments[i]
```

I think both approaches make sense. I agree that we should document the API better, and consider adding a duplication check to guard against invalid uses of the API (even though it's gonna be annoying to code in C).

It would also be nice to get some feedback from client devs, who know better how the function is used. The goal is to make it as ergonomic as possible and minimize footguns.

(/cc @jtraglia )

jtraglia:
> Last week a bug arose because of this, so I have been thinking about updating the specs, so that its one commitment, one proof, one cell. This would also mean we get rid of rowIndices.

Hmm I understand the argument but I'm against this. It would be quite inefficient to do it this way. It would force clients to create a new list of commitments (with a bunch of duplicates) instead of using the set of commitments they already have. Also, regarding the bug... this is explains why I was so confused. I didn't realize you were doing it this way.

> I think both approaches make sense. I agree that we should document the API better, and consider adding a duplication check to guard against invalid uses of the API (even though it's gonna be annoying to code in C).

Yeah definitely. This wasn't advertised very well & the misunderstanding is justified. But we shouldn't add length checks for the list of commitments like you maybe suggested. It would be more usable if clients could *always* provide the `BeaconBlockBody.blob_kzg_commitments` list without creating a new filtered list with only those being used.

kevaundray:
> > Last week a bug arose because of this, so I have been thinking about updating the specs, so that its one commitment, one proof, one cell. This would also mean we get rid of rowIndices.
> 
> Hmm I understand the argument but I'm against this. It would be quite inefficient to do it this way. It would force clients to create a new list of commitments (with a bunch of duplicates) instead of using the set of commitments they already have. Also, regarding the bug... this is explains why I was so confused. I didn't realize you were doing it this way.
> 
> > I think both approaches make sense. I agree that we should document the API better, and consider adding a duplication check to guard against invalid uses of the API (even though it's gonna be annoying to code in C).
> 
> Yeah definitely. This wasn't advertised very well & the misunderstanding is justified. But we shouldn't add length checks for the list of commitments like you maybe suggested. It would be more usable if clients could _always_ provide the `BeaconBlockBody.blob_kzg_commitments` list without creating a new filtered list with only those being used.

I think duplicating commitments (48 bytes) is not guaranteed to be inefficient (allocations are on the order of nano to microseconds for Megabytes) -- we would need to deduplicate them internally in the batch verify method to have the same behaviour we have now, ie the row indices will be created internally though.

I agree with George in that if this is the most ergonomic way to do it with the least amount of footguns, then it makes sense to do that and check if there are any performance concerns with respects to allocations.

> Also, regarding the bug... this is explains why I was so confused. I didn't realize you were doing it this way.

It was not being done this way, since that would make the API incompatible with the specs.

> It would be more usable if clients could always provide the BeaconBlockBody.blob_kzg_commitments list without creating a new filtered list with only those being used.

Yeah generally not a fan of giving more degrees of freedom than is needed, though I think in this case, I think it would be less footgunny since the rowIndices would need to be offsetted for filtered commitments. 

In this case, I think to prevent invalid commitments being inputted but not used, it makes sense to validate that all of the inputs are valid (which would come at a cost). Its not clear to me whether this will cause security concerns up the stack if we don't do it, but imo it is better to be conservative and loosen up on the security assumption when it can be shown that its okay to not be lax, or leave it in if it is seen to be insignificant performance wise in the worse case.

My general opinion is that the batch verify method's API should look like the single verify method, except for the fact that it takes a list of items and not a single item, and then internally it can deduplicate etc because the cryptography code is aware that duplicate commitments can be optimized away. Benchmarks could be made using worse case numbers to check if allocations are an issue here in terms of performance.

jimmygchen:
Hey @asn-d6 @kevaundray @jtraglia thanks for the inputs! Definitely helpful to understand the library from your angles and the design rationale.

An example for batch verification usage in Lighthouse:
1. The BN receives a batch (e.g. size 128) of `DataColumnSidecar`s, with _each_ of them containing `n` cells, `n` proofs and `n` commitments for the same block with `n` blobs.
2. The BN then combine the 128 sidecars into arrays containing `128n` cells, `128n` proofs and `n` kzg commitments (note that kzg commitments are de-duplicated in this PR, on `das` branch it's currently `128n`), and then pass it to the kzg library for batch verification

> It would be more usable if clients could always provide the BeaconBlockBody.blob_kzg_commitments list without creating a new filtered list with only those being used.

Note that the `kzg_commitments` used for batch verification are from the column sidecars, not the blocks (although they are the same) - so we aren't creating a new filtered list, and the full list of kzg commitments is _always_ provided to the library.

The reason for this is Lighthouse performs batch verficaition on the column sidecars first, and then checks them against the block when all components are available. This means we'd have many copies of the kzg commitments already (from each sidecar received), and with the current api, we _should_ perform de-duplcation before passing it to the library.

Without de-duplication, it still worked and it seems like the library just ignores the redundant values, however it poses a risk where not all kzg commitments from all sidecars are verified - the verification will pass as long as the first `n` kzg commitments are valid. I think this is potentially a major security risk if neither the client nor library perform this check, and the library would ideally de-duplicate, check for consistency or validate the length to avoid this attack factor. 

This is a bug on our side, although I think the original behaviour of "passing everything we've got" is quite an easy assumption to make, and would be ideal if the inputs are validated.

### Example

Here's [a passing test](https://github.com/jimmygchen/lighthouse/blob/f0e8c907c196b2da23de0eebb139bb917206a612/beacon_node/beacon_chain/src/kzg_utils.rs#L174-L183) that should've failed if we didn't do de-duplication and consistency check. 

The commitments array looks like this, which `c` representing a commitment (`c1` is commitment for first blob):
```
[ c1, c2, c3, c4, c5, c6, *c1, c2, .. , c5, c6 ]
```

where [I intentionally modified](https://github.com/jimmygchen/lighthouse/blob/kzg-batch-verfication-example/beacon_node/beacon_chain/src/kzg_utils.rs#L64-L74) all the `c1`s except for the first one - invalid commitment is denoted by `*c1` - this should've failed but it successfully verified.

asn-d6:
Hello @jimmygchen! Thanks for all the feedback.
We have a PR ready for consensus-specs that should help address your above concerns: https://github.com/ethereum/consensus-specs/pull/3834
Do you have time to check it out?

jimmygchen:
I think this PR is no longer needed as we have the above spec change merged. Thanks all!

PR #5897, opened 2024-06-06
Store pubkey cache decompressed on disk

michaelsproul:
## Issue Addressed

Closes:

- https://github.com/sigp/lighthouse/issues/3505

## Proposed Changes

Resolve a long-standing performance pitfall involving the decompression of pubkeys on startup. This PR improves Lighthouse's startup time _dramastically_.

## Additional Info

I propose we merge this PR as the first DB schema change adopted from `tree-states`, after the Electra PR which implements v20 is merged:

- https://github.com/sigp/lighthouse/pull/5712 

Blocked on a fix to the v20 schema:

- https://github.com/sigp/lighthouse/pull/6035

PR #6035, opened 2024-07-02
Fix SigVerifiedOp SSZ implementation

michaelsproul:
## Issue Addressed

The schema v20 migration was liable to fail and brick the database due to a discrepency in the `Encode` and `Decode` implementations for `SigVerifiedOp`. This was identified while testing the v21 schema migration: https://github.com/sigp/lighthouse/pull/5897#issuecomment-2196266245.

## Proposed Changes

Rather than trying to tweak the hand-written implementation I've replaced it with a more obviously correct implementation that translates to and from `SigVerifiedOp{Encode,Decode}` types. IMO this is much easier to check the correctness of, because the `Decode` and `Encode` implementations are just mappings.

I've also added some roundtrip tests which pass on the new impl and fail on the old.

## Additional Info

Manual testing of the migration didn't catch this because the error only occurs sometimes: when there are `SigVerifiedOp`s in the op pool.

### DIFF
--- a/consensus/state_processing/src/verify_operation.rs
+++ b/consensus/state_processing/src/verify_operation.rs
@@ -7,17 +7,17 @@ use crate::per_block_processing::{
     verify_proposer_slashing,
 };
 use crate::VerifySignatures;
+use arbitrary::Arbitrary;
 use derivative::Derivative;
 use smallvec::{smallvec, SmallVec};
 use ssz::{Decode, Encode};
 use ssz_derive::{Decode, Encode};
 use std::marker::PhantomData;
+use test_random_derive::TestRandom;
 use types::{
-    AttesterSlashing, AttesterSlashingBase, AttesterSlashingOnDisk, AttesterSlashingRefOnDisk,
-};
-use types::{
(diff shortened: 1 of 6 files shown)


consensus/state_processing/src/verify_operation.rs
```diff
@@ -39,12 +39,13 @@ pub trait TransformPersist {
 ///
 /// The inner `op` field is private, meaning instances of this type can only be constructed
 /// by calling `validate`.
-#[derive(Derivative, Debug, Clone)]
+#[derive(Derivative, Debug, Clone, Arbitrary)]
```
michaelsproul, line 42: I added this arbitrary implementation because my first reproduction used `cargo-fuzz`. It was quicker than wiring up `TestRandom`.

We don't usually check in `cargo fuzz` targets, but I figured there's no harm in leaving this in case we want to wire it up again in future.

ethDreamer (approved):
Yeah this is way simpler. Thanks for finding and fixing this!

michaelsproul:
@mergify queue

PR #6039, opened 2024-07-03
Show `blst` hardware support in `lighthouse --version`

### LINKED ISSUE #5918: Show `blst` hardware support in `lighthouse --version`
## Description

Now that we're using `portable` `blst` everywhere, we could show the hardware acceleration support _dynamically_ in `lighthouse --version`, like we do for SHA256.

Something like:

```
Lighthouse v5.2.0-f1d88ba
BLS library: blst-portable
SHA256 hardware acceleration: false
Allocator: jemalloc
Profile: maxperf
Specs: mainnet (true), minimal (true), gnosis (true)
```

scafe3:
Show `blst` hardware support in `lighthouse --version`:

```
$ ./target/debug/lighthouse --version
Lighthouse v5.2.1
BLS library: blst-portable
BLS hardware acceleration: false
Allocator: system
Profile: debug
```

Will close #5918

### DIFF
--- a/lighthouse/src/main.rs
+++ b/lighthouse/src/main.rs
@@ -53,15 +53,12 @@ fn bls_library_name() -> &'static str {
 }
 
 #[inline(always)]
-#[allow(unreachable_code)]
 fn bls_hardware_acceleration() -> bool {
-    #[cfg(feature = "portable")]
-    return false;
-
     #[cfg(target_arch = "x86_64")]
     return std::is_x86_feature_detected!("adx");
 
-    false
+    #[cfg(target_arch = "aarch64")]
+    return std::arch::is_aarch64_feature_detected!("neon");
 }
 
 fn allocator_name() -> &'static str {
--- a/lighthouse/src/main.rs
+++ b/lighthouse/src/main.rs
@@ -52,8 +52,14 @@ fn bls_library_name() -> &'static str {
     }
 }
 
+#[inline(always)]
+#[allow(unreachable_code)]
 fn bls_hardware_acceleration() -> bool {
-    !cfg!(feature = "portable")
+    #[cfg(feature = "portable")]
+    return false;
+
+    #[cfg(target_arch = "x86_64")]
+    return std::is_x86_feature_detected!("adx");
 }
 
 fn allocator_name() -> &'static str {
--- a/lighthouse/src/main.rs
+++ b/lighthouse/src/main.rs
@@ -26,12 +26,14 @@ lazy_static! {
     pub static ref LONG_VERSION: String = format!(
         "{}\n\
          BLS library: {}\n\
+         BLS hardware acceleration: {}\n\
          SHA256 hardware acceleration: {}\n\
          Allocator: {}\n\
          Profile: {}\n\
          Specs: mainnet (true), minimal ({}), gnosis ({})",
         SHORT_VERSION.as_str(),
         bls_library_name(),
+        bls_hardware_acceleration(),
         have_sha_extensions(),
         allocator_name(),
         build_profile_name(),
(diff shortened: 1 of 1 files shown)


lighthouse/src/main.rs
```diff
@@ -50,6 +52,10 @@ fn bls_library_name() -> &'static str {
     }
 }
 
+fn bls_hardware_acceleration() -> bool {
+    !cfg!(feature = "portable")
```
michaelsproul: This just shows whether the portable feature flag is enabled, which is a compile-time property. We want to check whether the user's CPU supports the instructions _at runtime_.

To do that, we can do something similar to the SHA-extension detection in `ethereum_hashing`:

https://github.com/sigp/ethereum_hashing/blob/b6b5ff2d8db972ee6453a710e82aee3c638d5d82/src/lib.rs#L127-L140

The CPU feature we are interested in on x86 is ADX. On ARM I think there's a different extension we should check for, we would have to check the `blst` docs/code to determine what it is

michaelsproul (changes requested):

lighthouse/src/main.rs
scafe3: Oh got it. Will have a fix!

scafe3: @michaelsproul updated, please review.

On AMD it's also ADX: https://github.com/supranational/blst/issues/10#issue-658121495

lighthouse/src/main.rs
```diff
@@ -50,6 +52,18 @@ fn bls_library_name() -> &'static str {
     }
 }
 
+#[inline(always)]
+#[allow(unreachable_code)]
+fn bls_hardware_acceleration() -> bool {
+    #[cfg(feature = "portable")]
+    return false;
```
michaelsproul:
```suggestion
```

The `portable` feature is irrelevant. We want to do runtime detection regardless of whether it is enabled.

lighthouse/src/main.rs
```diff
@@ -50,6 +52,18 @@ fn bls_library_name() -> &'static str {
...
+
+    #[cfg(target_arch = "x86_64")]
+    return std::is_x86_feature_detected!("adx");
```
michaelsproul, line 58: This is good for x86_64, but for ARM i.e. aarch64 we need to use [`is_aarch64_feature_detected!`](https://doc.rust-lang.org/std/arch/macro.is_aarch64_feature_detected.html).

Based on a quick read of some BLST issues, I think the feature we want on aarch64 is `neon`.

https://github.com/supranational/blst/issues/213

lighthouse/src/main.rs
```diff
@@ -50,6 +52,18 @@ fn bls_library_name() -> &'static str {
     }
 }
 
+#[inline(always)]
+#[allow(unreachable_code)]
```
michaelsproul:
```suggestion
```

If you write the body of the function with mutually exclusive `cfg`s so that there's only 1 return statement, then this won't be necessary. I think removing the `cfg(feature = "portable")` will get most of the way there.

michaelsproul (changes requested):

lighthouse/src/main.rs
michaelsproul: I was referring to ARM, i.e. aarch64, not AMD (which uses the x86_64 ISA)

See https://github.com/sigp/lighthouse/pull/6039#discussion_r1664981977 for a suggested fix

lighthouse/src/main.rs
scafe3: Do you mean it can utilize the hardware acceleration even if it's portable binary?

michaelsproul: Yes

michaelsproul (approved):
Awesome, thanks!

Tested on my M1 Pro Mac and it shows `true`!

michaelsproul:
@mergify queue

scafe3:
@michaelsproul Thanks for your guidance!

PR #6040, opened 2024-07-03
Tree states archive metrics

dapplion:
## Issue Addressed

Add metrics for

## Proposed Changes

Adds a bunch of useful metrics to debug / understand the performance of Tree states archive. Some of this metrics could be ported to unstable now.

This branch may not compile as I have a local unresolved issue with xdelta3 compilation of macos arm chips

### DIFF
--- a/beacon_node/store/src/hot_cold_store.rs
+++ b/beacon_node/store/src/hot_cold_store.rs
@@ -1615,6 +1615,8 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold> 
         state_root_iter: Option<impl Iterator<Item = Result<(Hash256, Slot), Error>>>,
         pre_slot_hook: Option<PreSlotHook<E, Error>>,
     ) -> Result<BeaconState<E>, Error> {
+        metrics::inc_counter_by(&metrics::BEACON_STORE_REPLAYED_BLOCKS, blocks.len());
+
         let mut block_replayer = BlockReplayer::new(state, &self.spec)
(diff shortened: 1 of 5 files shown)


beacon_node/store/src/hot_cold_store.rs
```diff
@@ -1442,20 +1484,31 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
                 "Hit diff buffer cache";
                 "slot" => slot
             );
+            metrics::inc_counter(&metrics::STORE_BEACON_DIFF_BUFFER_CACHE_HIT);
             return Ok((slot, buffer.clone()));
+        } else {
+            metrics::inc_counter(&metrics::STORE_BEACON_DIFF_BUFFER_CACHE_MISS);
         }
 
+        // Do not time recursive calls into load_hdiff_buffer_for_slot to not double count
```
michaelsproul, line 1505: I'm not convinced how useful this `recursion` param is, because this will cause us just to time the top-level hdiff buffer request, which could be highly variable due to the number of recursive calls.

michaelsproul: Maybe we could time the whole call and then divide by `recursion` at the end? :thinking:

dapplion: I'm not super happy about the recursion param either. Counting the same time multiple times feels like it defeats the purpose of the metric tho. I would like to metric also how many recursions are necessary per hdiff load. Maybe we should move this metrics to the caller to avoid recursion?

PR #6041, opened 2024-07-03
Track store metrics by DB column

dapplion:
## Issue Addressed

When adding metrics for
- https://github.com/sigp/lighthouse/pull/5978#issuecomment-2182439210

Noted that the DB metrics are already there but we don't track them by column. I would like to be able to tell:
- How many reads / writes happen on the state diff buffer column?
- How many bytes do we load / store on the state diff buffer column?

## Proposed Changes

- Switch store metrics from `IntCounter` to `IntCounterVec` and label by column enum 3 letter string. 
- Track metrics in `do_atomically` calls, optimistically assuming success

### DIFF
--- a/beacon_node/store/src/leveldb_store.rs
+++ b/beacon_node/store/src/leveldb_store.rs
@@ -59,16 +59,13 @@ impl<E: EthSpec> LevelDB<E> {
     ) -> Result<(), Error> {
         let column_key = get_key_for_col(col, key);
 
-        metrics::inc_counter(&metrics::DISK_DB_WRITE_COUNT);
-        metrics::inc_counter_by(&metrics::DISK_DB_WRITE_BYTES, val.len() as u64);
-        let timer = metrics::start_timer(&metrics::DISK_DB_WRITE_TIMES);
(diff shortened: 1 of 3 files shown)


beacon_node/store/src/leveldb_store.rs
```diff
@@ -137,14 +138,28 @@ impl<E: EthSpec> KeyValueStore<E> for LevelDB<E> {
...
+                        &metrics::DISK_DB_WRITE_BYTES,
+                        &[&col],
+                        value.len() as u64,
+                    );
+
                     leveldb_batch.put(BytesKey::from_vec(key), &value);
                 }
 
                 KeyValueStoreOp::DeleteKey(key) => {
+                    let col = get_col_from_key(&key).unwrap_or("unknown".to_owned());
```
dapplion, line 153: `get_col_from_key` should never error as we define the DBColumn enum variants. The use of col is purely informational on metrics, so unwrapping to unknown to have something

michaelsproul (approved):
LGTM, thanks!

jimmygchen:
@mergify queue

PR #6042, opened 2024-07-03
Drop unused store metrics

dapplion:
## Issue Addressed

Is there a way for the linter to caught these unused static refs? I changed them to pub(crate) and still did not raise a warn

## Proposed Changes

Drop unused store metrics


michaelsproul:
> Is there a way for the linter to caught these unused static refs?

I haven't found one. There are probably a bunch of other unused metrics hanging around as well. Maybe a script to `git grep` the whole repo is the easiest option

realbigsean (approved):

realbigsean:
@mergify queue

PR #6044, opened 2024-07-04
Add KZG commitment consistency check

jimmygchen:
## Issue Addressed

Check data column commitments against block commitments when adding to `DataAvailabilityChecker` and constructing an `RpcBlock`.

### DIFF
--- a/beacon_node/beacon_chain/src/block_verification_types.rs
+++ b/beacon_node/beacon_chain/src/block_verification_types.rs
@@ -8,6 +8,7 @@ use crate::data_column_verification::{
 use crate::eth1_finalization_cache::Eth1FinalizationData;
 use crate::{get_block_root, GossipVerifiedBlock, PayloadVerificationOutcome};
 use derivative::Derivative;
+use itertools::Itertools;
 use ssz_types::VariableList;
 use state_processing::ConsensusContext;
 use std::fmt::{Debug, Formatter};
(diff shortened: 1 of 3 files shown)


beacon_node/beacon_chain/src/block_verification_types.rs
```diff
@@ -162,10 +163,32 @@ impl<E: EthSpec> RpcBlock<E> {
...
+            column_commitments,
+            block.message().body().blob_kzg_commitments(),
+        ) {
+            if column_commitments.len() != block_commitments.len() {
+                return Err(AvailabilityCheckError::KzgCommitmentListMismatch);
+            }
+            for (&column_commitment, &block_commitment) in
+                column_commitments.iter().zip(block_commitments.iter())
+            {
+                if column_commitment != block_commitment {
```
dapplion, line 183: I actually want to remove this check for blob sidecars since the inclusion proof should ensure that this never happens. Have you observed this inconsistency issue on devnets?

jimmygchen: Yeah I actually ran into a few KZG errors during sync. 

I think you raised a good point and this PR shouldn't be required. The proper fix is to ensure that we verify the inclusion proof - I think we're missing this in sampling and by range rpc. I'll raise an issue.

jimmygchen: Raised issue #6111. Closing this one

PR #6046, opened 2024-07-04
Record `BEACON_BLOCK_DELAY_GOSSIP` metric only after a bock is verified

chong-he:
@michaelsproul found that the `BEACON_BLOCK_DELAY_GOSSIP` metric is recorded before the block is verified:

https://github.com/sigp/lighthouse/blob/d84e3e391e336a5819b8734a3cf285f8821204a7/beacon_node/network/src/network_beacon_processor/gossip_methods.rs#L910-L919

This causes the metric could record a time in the range minutes-hours if an old block is received, see https://discord.com/channels/605577013327167508/746918775919738893/1255734712971038810

This PR fixes it so that the metric is only recorded after the block is verified.

Thanks @michaelsproul !


michaelsproul (approved):
LGTM, thanks

michaelsproul:
@mergify queue

PR #6048, opened 2024-07-04
Remove timeout locks

dapplion:
## Issue Addressed

Chatting with @michaelsproul he noted that locks with timeouts can kill liveness, and do more harm than good. If an operation takes more than the timeout to acquire the lock, erroring causes that operation to be retried and fail again. 

I recall that timeouts helped in the past with deadlocks. Looks like this thread is well mitigated now?

## Proposed Changes

Replace these three timeout rw locks with rw locks

```rust
pub const ATTESTATION_CACHE_LOCK_TIMEOUT: Duration = Duration::from_secs(1);
pub const VALIDATOR_PUBKEY_CACHE_LOCK_TIMEOUT: Duration = Duration::from_secs(1);
pub const ETH1_FINALIZATION_CACHE_LOCK_TIMEOUT: Duration = Duration::from_millis(200);
```


michaelsproul:
We should also hide and deprecate the `disable-lock-timeouts` flag when removing these

I was going to block the removal of the timeouts on adding deadlock detection, but it seems like getting `lockbud` to run could be quite involved:

https://github.com/BurtonQin/lockbud

It depends on Rust Nightly, so it will need to be continually updated as we increase our MSRV, and might break in surprising ways (my experience with writing lints pinned to nightly)

beacon_node/src/cli.rs
```diff
@@ -1238,16 +1238,6 @@ pub fn cli_app() -> Command {
                 .action(ArgAction::Set)
                 .display_order(0)
         )
-        .arg(
-            Arg::new("disable-lock-timeouts")
```
michaelsproul, line 1247: You gotta leave this part for a soft-deprecation. We should accept the flag, but log a warning. We _should not_ refuse to start if the flag is supplied until a few releases are done

lighthouse/tests/beacon_node.rs
```diff
@@ -159,21 +159,6 @@ fn max_skip_slots_flag() {
...
-fn enable_lock_timeouts_default() {
-    CommandLineTest::new()
-        .run_with_zero_port()
-        .with_config(|config| assert!(config.chain.enable_lock_timeouts));
-}
-
-#[test]
-fn disable_lock_timeouts_flag() {
-    CommandLineTest::new()
-        .flag("disable-lock-timeouts", None)
```
michaelsproul, line 172: this test should stay, but we should just assert that it doesn't panic

michaelsproul (approved):

michaelsproul:
@dapplion a few CI failures to fix

michaelsproul:
@mergify queue

PR #4277, opened 2023-05-09
Harden against finalized block spam

michaelsproul:
## Proposed Changes

- Remove a double block root calculation for blocks that are prior to finalization based on their slot. We just plumb the error through from `block_verification.rs`. This has the potential to halve the time spent rejecting these messages (from 10-20ms to 5-10ms).
- Increase the peer penalty for blocks that are prior to finalization. They are still `Ignore`d though rather than rejected, as that's what the spec requires.
- Increase the peer penalty for blocks that conflict with finalization (side chains that are obsoleted by finalization). They are now `Reject`ed, which also aligns with the spec.

## Additional Info

- Remove `check_block_against_anchor_slot`. I added it in 2021 when I implemented checkpoint sync, but it's superfluous because the anchor slot is always <= the finalized slot, so any block caught by the anchor check will be caught by the finalized slot check.
- There are a few other places we could plumb through the block root: `BlockIsAlreadyKnown`, `RepeatProposal`, `NotFinalizedDescendant`, `BlockIsNotLaterThanParent`, `TooManySkippedSlots` are all _prior_ to the proposer signature check.

PR #6050, opened 2024-07-05
Increase penalty for old block gossip spam

michaelsproul:
## Issue Addressed

Simplified and streamlined version of:

- https://github.com/sigp/lighthouse/pull/4277

## Proposed Changes

We're still seeing a lot of stale pre-finalization blocks being gossiped by Erigon/Caplin. To encourage users to update to a non-buggy version, this PR increases the peer penalty which will result in more frequent bans.

As far as I can tell there are no cases where honest peers should be gossiping blocks prior to finalization. If a peer is syncing, they should not be broadcasting the blocks they're syncing on gossip. In case of a consensus fault and chain split, it's somewhat irrelevant whether the two sides of the split remain connected to each other, as the damage has already been done by finalizing and nodes will need to be manually resynced to the correct chain. E.g. if Lighthouse-Geth finalizes an invalid checkpoint, then it will view the (valid) minority chain as conflicting with its view of finalization. There is no way it will continue importing blocks from the valid chain until its DB is blown away to revert the faulty finalization. Most attached validators (ones that attested to the invalid finalization) will also be lost at this point (leaked out to 0).

## Additional Info

Related to:

- https://github.com/sigp/lighthouse/pull/6046

### DIFF
--- a/beacon_node/network/src/network_beacon_processor/gossip_methods.rs
+++ b/beacon_node/network/src/network_beacon_processor/gossip_methods.rs
@@ -1010,16 +1010,36 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                 self.propagate_validation_result(message_id, peer_id, MessageAcceptance::Ignore);
                 return None;
             }
-            Err(e @ BlockError::FutureSlot { .. })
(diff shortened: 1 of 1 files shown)


AgeManning (approved):
All for banning bad behaviour

michaelsproul:
@mergify queue

PR #6061, opened 2024-07-08
Add 7636 support and update discv5 to 0.6.0

SirSpudlington:
This PR:

- adds EIP-7636 Support
- updates the discv5 library to 0.6.0
- Updates how the persisted DHT is stored to be compatible with the newer library


SirSpudlington:
I am leaving this in a draft state until the EIP reaches a final state.

PR #6069, opened 2024-07-09
Update PeerDAS interop testnet config

jimmygchen:
## Issue Addressed

- Add overrides on network config, so that client configs match
- Enable ethereum metrics exporter and prometheus_grafana


jimmygchen:
Merging this for ease of testing.

Issue #6072, opened 2024-07-11
Progressively merge `das` to `unstable`


jimmygchen:
## Description

Progressively merge `das` branch into `unstable`. Full `das` branch PR for reference 👇 
- https://github.com/sigp/lighthouse/pull/5683

The following proposed by @dapplion :

### PeerDAS to unstable (topological sort)

- Add DataColumn type - https://github.com/sigp/lighthouse/pull/5913
- Store changes to persist data columns - https://github.com/sigp/lighthouse/pull/6073
- da_checker is_available update - https://github.com/sigp/lighthouse/pull/6076
- Add gossip data_column topic - https://github.com/sigp/lighthouse/pull/6147
- Add gossip validation - https://github.com/sigp/lighthouse/pull/6181
- Subscribe to data_column topic on peerdas epoch
- Import gossip data_column to da_checker https://github.com/sigp/lighthouse/pull/6197
- Add KZG lib integration #6212 
- Add supernode functionality #6216

### _Conga line_ from here
 1. #6224
 2. #6238
 3. #6237 
4. #6255 
- Block production + publishing updates        <<< Here e2e test a supernode only network
- Reconstruction
- Lookup sync (custody) updates      <<< Here e2e test a N regular node + 1 supernode network
- Range sync (custody) updates        <<< Here e2e test adding new node and sync to head

(here have parity with das branch w.r.t. custody)

- Re-process early sampling requests
- Checkpoint Sync data columns
- Respond with first byte to extend timeout for early sampling requests
- Batch verify KZG proofs where possible

(now tackle sampling, which should be purely additive)

- Add sampling
- Network service update to maintain peers on all columns
- Forkchoice updates w.r.t. sampling result

... plus improvements, metrics, etc

jimmygchen:
Completed in #5683 🎉

PR #5050, opened 2024-01-09
1D PeerDAS prototype: Data format and Distribution

jimmygchen:
## Issue Addressed

This PR contains the initial groundwork of the 1D PeerDAS prototype based on this research post: [From 4844 to Danksharding: a path to scaling Ethereum DA](https://ethresear.ch/t/from-4844-to-danksharding-a-path-to-scaling-ethereum-da/18046). It mostly covers changes mentioned under the [Data format](https://ethresear.ch/t/from-4844-to-danksharding-a-path-to-scaling-ethereum-da/18046/1#data-format-3) and [Distribution](https://ethresear.ch/t/from-4844-to-danksharding-a-path-to-scaling-ethereum-da/18046/1#distribution-4) sections. 

**EDIT**: The PR has now been updated with changes from [consensus-spec#3574](https://github.com/ethereum/consensus-specs/pull/3574).

This initial prototype does not currently build column data from erasure coded blobs, but generates random column data based on the number of blobs per block, which means the size of the columns would be the same as if they were built from erasure coded blobs. This will allow us to experiment with distributing the `DataColumnSidecar` objects in a network and gather some initial measurements on bandwidth, feasibility and data availability. 

## Changes

- Add `DataColumnSidecar` type. This is the used as the sampling unit for columns: 
  > Blobs are individually extended horizontally, stacked vertically and subdivided in `NUM_COLUMNS` (128) columns, which are used as the sampling unit.
- Produce `DataColumnSidecar` objects, [filled with dummy data](https://github.com/jimmygchen/lighthouse/blob/d6d6a64893d0620271933572fe2a9397727d135e/consensus/types/src/blob_column_sidecar.rs#L45), with its size derived from the number of blobs associated with the block, and publish them to `NUM_COLUMN_SUBNETS ` (32) column subnets over gossip.
- By default, each node subscribes to 1 data column subnets (`CUSTODY_REQUIREMENT =1`), with the subnet id determined using the `NodeId` (see the [compute_subnets_for_data_column](https://github.com/jimmygchen/lighthouse/blob/b9a970bf09669c1c49006c0cc5250ceabc061728/consensus/types/src/data_column_subnet_id.rs#L49) function ).

## Running a Network using Kurtosis

I've got a little devnet running locally using Kurtosis, and it shows blob columns getting published and received by peers. Instructions can be found in [this gist](https://gist.github.com/jimmygchen/54574443ea8a94118f0f748b69a5e610).

[...]

PR #6073, opened 2024-07-11
Store changes to persist data columns

jimmygchen:
## Issue Addressed

Part of #6072.

This PR contains only the Store changes to persist data columns.

DataColumns are stored individually by index. Sampling peers will request single columns or arbitrary slots. Storing bundled like blobs would be inefficient. 

Upstream PRs:
- https://github.com/sigp/lighthouse/pull/5050
- https://github.com/sigp/lighthouse/pull/5890


beacon_node/store/src/lib.rs
```diff
@@ -143,6 +142,28 @@ pub fn get_key_for_col(column: &str, key: &[u8]) -> Vec<u8> {
...
 }
 
+pub fn get_data_column_key(block_root: &Hash256, column_index: &ColumnIndex) -> Vec<u8> {
+    let mut result = block_root.as_bytes().to_vec();
+    result.extend_from_slice(&column_index.to_le_bytes());
+    result
+}
+
+pub fn parse_data_column_key(data: Vec<u8>) -> Result<(Hash256, ColumnIndex), Error> {
+    if data.len() != 32 + 8 {
```
realbigsean: we can use `DBColumn::BeaconDataColumn.key_size()` here

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -1748,6 +1914,24 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
         self.blob_info.read_recursive().clone()
     }
 
+    /// Initialize the `DataColumnInfo` when starting from genesis or a checkpoint.
+    pub fn init_data_column_info(&self, anchor_slot: Slot) -> Result<KeyValueStoreOp, Error> {
+        let oldest_data_column_slot = self.spec.deneb_fork_epoch.map(|fork_epoch| {
```
realbigsean: should this be pectra fork epoch

jimmygchen: Good point - currently we're testing peerdas on top of deneb, so we've been using deneb in a few places, but this should be `eip7594_fork_epoch` and we can change it to electra later once we rebase PeerDAS on `electra`.

jimmygchen: Fixed in https://github.com/sigp/lighthouse/pull/6073/commits/9b2016875a3922331e5390e9ed7f115df87479b3

beacon_node/store/src/lib.rs
jimmygchen: ah nice, thanks! Fixed in https://github.com/sigp/lighthouse/pull/6073/commits/9b2016875a3922331e5390e9ed7f115df87479b3

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -294,11 +317,39 @@ impl<E: EthSpec> HotColdDB<E, LevelDB<E>, LevelDB<E>> {
...
+                let oldest_data_column_slot = data_column_info
+                    .oldest_data_column_slot
+                    .or(eip7594_fork_slot);
+                DataColumnInfo {
+                    oldest_data_column_slot,
+                }
+            }
+            // First start.
+            None => DataColumnInfo {
+                // Set the oldest data column slot to the Deneb fork slot if it is not yet set.
```
```suggestion
                // Set the oldest data column slot to the fork slot if it is not yet set.
```

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -294,11 +317,39 @@ impl<E: EthSpec> HotColdDB<E, LevelDB<E>, LevelDB<E>> {
...
         db.compare_and_set_blob_info_with_write(<_>::default(), new_blob_info.clone())?;
+
+        let data_column_info = db.load_data_column_info()?;
+        let eip7594_fork_slot = db
+            .spec
+            .eip7594_fork_epoch
+            .map(|epoch| epoch.start_slot(E::slots_per_epoch()));
+        let new_data_column_info = match &data_column_info {
+            Some(data_column_info) => {
+                // Set the oldest data column slot to the Deneb fork slot if it is not yet set.
```
```suggestion
                // Set the oldest data column slot to the fork slot if it is not yet set.
```

michaelsproul (approved):
LGTM!

Although this introduces new columns on disk I think it is OK to merge this without a schema version increment as nothing will be stored in those columns until the fork epoch is set.

Ideally we would want to block downgrading to a column-oblivious schema once peer DAS is activated, but it's kind of a moot point because nodes won't be able to follow the chain without these changes (and the rest of the DAS logic).

We will also be bumping the schema version (and disallowing downgrades) with this PR:

michaelsproul:
@mergify queue

michaelsproul:
@mergify refresh

michaelsproul:
@mergify queue

PR #6076, opened 2024-07-11
Update `is_available` check to support PeerDAS.

jimmygchen:
## Issue Addressed

Part of #6072. See the issue to see the current breakdown of das changes.

This PR contains only changes to update `is_available` check in DA checker and does not impact current mainnet code path.


beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -94,8 +94,17 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
         log: &Logger,
         spec: ChainSpec,
     ) -> Result<Self, AvailabilityCheckError> {
-        let overflow_cache =
-            DataAvailabilityCheckerInner::new(OVERFLOW_LRU_CAPACITY, store, spec.clone())?;
+        // TODO(das): support supernode or custom custody requirement
```
pawanjay176, line 97: Not sure I understand the custom custody requirement in this comment.

I had assumed that `spec.custody_requirement` is what all nodes need to satisfy to pass `is_data_available` condition. Any node that increases its custody requirement would simply download everything but not wait for the entire data download to complete for any validity conditions.

For supernodes, are you saying that a node would return `is_data_available == false` until it has downloaded all the data?

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -212,6 +240,35 @@ impl<E: EthSpec> PendingComponents<E> {
             AvailableExecutedBlock::new(available_block, import_data, payload_verification_outcome),
         )))
     }
+
+    /// Returns the epoch of the block if it is cached, otherwise returns the epoch of the first blob.
+    pub fn epoch(&self) -> Option<Epoch> {
```
pawanjay176: Maybe we just pass the epoch to the da_checker functions instead of trying to calculate it every time?

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -134,15 +146,29 @@ impl<E: EthSpec> PendingComponents<E> {
...
-        if let Some(num_expected_blobs) = self.num_expected_blobs() {
-            num_expected_blobs == self.num_received_blobs()
-        } else {
-            false
+    /// Returns `true` if both the block exists and the number of received blobs / custody columns
+    /// matches the number of expected blobs / custody columns.
+    pub fn is_available(&self, block_import_requirement: &BlockImportRequirement) -> bool {
+        match block_import_requirement {
+            BlockImportRequirement::AllBlobs => {
+                if let Some(num_expected_blobs) = self.num_expected_blobs() {
```
realbigsean: you can use `map_or(false, ...)` here

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -134,15 +146,29 @@ impl<E: EthSpec> PendingComponents<E> {
...
+                    num_expected_blobs == self.num_received_blobs()
+                } else {
+                    false
+                }
+            }
+            BlockImportRequirement::CustodyColumns(num_expected_columns) => {
+                let num_received_data_columns = self.num_received_data_columns();
+                if let Some(num_expected_blobs) = self.num_expected_blobs() {
```
realbigsean: you can use `map_or(false, ...)` here

beacon_node/beacon_chain/src/data_availability_checker.rs
jimmygchen: This PR doesn't introduce supernode yet - which is currently enabled by the `--import-all-data-columns` cli flag on the `das` branch - hence the comment to remind us this is incomplete. Here's how the `custody_subnet_count` is currently derived (either a supernode or a full node with minimal custody):

https://github.com/sigp/lighthouse/blob/04d9eef8a74ae23eee7f473cbafd48f64f0ffa01/beacon_node/beacon_chain/src/data_availability_checker.rs#L88-L92

`spec.custody_requirement` is the _minimum_ custody requirement for full nodes. If they custody more columns they would need to collect all its custody columns before they can consider data available for the current slot - for supernodes this would mean all columns would need to be downloaded and verified before the block can be imported and attested. Nodes do this not just to store and serve more data, but it also provides them a stronger level of security to their attestations and block proposals (if they have to verify a higher number custody columns before importing). In an [upcoming proposal](https://github.com/ethereum/consensus-specs/pull/3779), `validator custody` is introduced so that nodes with validators attached would need to custody more columns for this reason - reduced risk from sampling.

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: Not sure if I fully understand, we would still have to calculate the epoch from the slot of the block / blob right? or do you mean storing the `epoch` in `PendingComponents` whenever the first component is added so we don't re-calculate?

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: Nice, updated in 9a83a7c4f8e59f8c5ba51755e32f0683cb695900

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: Updated in 9a83a7c4f8e59f8c5ba51755e32f0683cb695900

beacon_node/beacon_chain/src/data_availability_checker.rs
pawanjay176: I had misunderstood the requirement here then. Clear now, thanks for explaining :)

pawanjay176 (approved):

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
pawanjay176: Added in [379f40a](https://github.com/sigp/lighthouse/pull/6076/commits/379f40a345a4276f836f9bc24ef8e75b68b54056)

jimmygchen:
Thanks for the review! 🙏

jimmygchen:
@mergify queue

PR #6077, opened 2024-07-11
Fix builds with `slasher-redb`

michaelsproul:
## Issue Addressed

Currently `unstable` fails to build with the `slasher-redb` feature due to a last-minute tweak I made to the `redb` impl prior to merging.

## Proposed Changes

- Correct the compilation error
- Run the tests for the `redb` slasher on CI (would have caught this issue)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,8 +174,9 @@ test-network-%:
 # Run the tests in the `slasher` crate for all supported database backends.
 test-slasher:
 	cargo nextest run --release -p slasher --features "lmdb,$(TEST_FEATURES)"
+	cargo nextest run --release -p slasher --no-default-features --features "redb,$(TEST_FEATURES)"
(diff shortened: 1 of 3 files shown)


jimmygchen (approved):
Looks good to me!

jimmygchen:
Oops, there's one failing slasher test: `failed_override_with_existing_mdbx_db`

michaelsproul:
Deleted the failing test. It's a duplicate of another test with the same name `failed_override_with_existing_mdbx_db` which tests the case where you already have an MDBX database and start without MDBX enabled

michaelsproul:
Ok third time lucky. Disabled all those slasher override tests when redb is involved.

michaelsproul:
@mergify queue

jimmygchen:
@mergify requeue

Issue #6022, opened 2024-07-01
Switch Lighthouse and all deps off `ethereum-types`


michaelsproul:
## Description

Presently Lighthouse is using `ethereum-types` for many of our core type definitions like `Hash256`. This is an old Parity crate that is not actively maintained any more, and there is some desire to move Lighthouse to Alloy so that it synergises better with the _new_ Rust Ethereum ecosystem around Reth, forge, etc.

This task would involve:

1. Updating all our external crates like `ethereum_hashing`, `tree_hash`, `ethereum_ssz` so that they do not depend on `ethereum-types`.
2. Updating Lighthouse to depend on the updated versions of the external crates prior to releasing the crates on `crates.io`. We don't want to commit to merging alloy into our external crates until we're sure it's going to work & not be too much hassle. We should use `patch` to point at the feature branches for each crate so that the crate overrides apply globally (e.g. so that we are also patching transitive dependencies on `ethereum_hashing` via `ethereum_ssz` to use the `git` dep).
3. Update API usage within Lighthouse
4. Merge the changes to the external crates and release new minor versions signalling the breaking change.
5. Update the Lighthouse PR to use the released versions of the crates, and then merge that to unstable.

Part 1 is already underway thanks to @KolbyML:

- https://github.com/sigp/tree_hash/pull/14
- https://github.com/sigp/ethereum_ssz/pull/15
- https://github.com/sigp/ssz_types/pull/22
- https://github.com/sigp/ethereum_serde_utils/pull/7

I have merged these branches down into an `alloy` branch on their respective repositories and created new PRs:

- https://github.com/sigp/tree_hash/pull/16
- https://github.com/sigp/ethereum_ssz/pull/21
- https://github.com/sigp/ssz_types/pull/25
- https://github.com/sigp/ethereum_serde_utils/pull/8

## Ideas to minimise the size of the diff

- We should alias `B256` to `Hash256` to avoid a huge amount of unnecessary rename churn. Even if we want to use the "proper" type name eventually, IMO it is better to separate the semantic change of using alloy, from a bunch of hard-to-review rename churn.

jimmygchen:
I don't see direct reference of `ethereum-types` anymore in Lighthouse, has this been completed? @michaelsproul

michaelsproul:
@jimmygchen No, unfortunately we still depend on two versions of `ethereum-types` transitively:

[...]

PR #6078, opened 2024-07-11
Migrate from `ethereum-types` to `alloy-primitives`

eserilev:
## Issue Addressed

partially #6022 

## Proposed Changes

Migrate LH from `ethereum-types` to `alloy-primitives`

## Additional Info

Some of the dependencies are referencing branches from my repo. I've opened PR's that need to be merged into those repos so we can revert back to referencing the sigp repos.

I've attempted to patch in some dependencies, but quickly faced some cyclic dep issues that I was unable to resolve


michaelsproul:
I tried investigating the circular dependency issue that crops up when using `[patch]`.

I think it is down to how Cargo handles optional dependencies, and possibly the `ethereum_ssz?/arbitrary` dependency here:

https://github.com/alloy-rs/core/blob/6540523385ea10c9adce31beae7106cfa7da770f/crates/primitives/Cargo.toml#L96-L105

I made a minimal test case here, and it is very broken, even without the `?` syntax:

https://github.com/michaelsproul/circular-test

I think our two options are:

1. Tolerate the existence of _two_ `ethereum_ssz` crates in the build tree until `alloy-primitives` is updated to not depend on `ethereum_ssz` _at all_. If we bump the `ethereum_ssz` version to `0.6` on the alloy branch, and depend on that version in Lighthouse, then the `ethereum_ssz` dep of `alloy-primitives` will not get patched.
2. Raise a PR to `alloy-primitives` to remove the `ethereum_ssz` dep, and make this part of the `patch` section. This probably needs to be the long-term solution, so maybe we should start working on this even if we take option (1) for now.

michaelsproul:
Related Cargo issue:

- https://github.com/rust-lang/cargo/issues/6743

eserilev:
Thanks for taking a look at this @michaelsproul 

For option (2), removing the `ethereum_ssz` dep, I think this means we'll need to impl ssz Encode & Decode directly in the `alloy_primitives` crate. Does that sound correct?

michaelsproul:
I think we should move the SSZ impls that are currently in alloy-primitives into ethereum_ssz itself.

E.g. implementing Decode/Encode for H256 should be in ethereum_ssz

eserilev:
I went ahead and made some progress on option (2)

- PR to `alloy` https://github.com/alloy-rs/core/pull/701
- PR to `ethereum_ssz` https://github.com/sigp/ethereum_ssz/pull/26

I've also patched in the relevant deps into Lighthouse. No more circular dependency issues!

michaelsproul:
I've released preliminary versions of all our crates which we can use in this PR. I think we could merge it to `unstable` and then come back for the coordinated release to remove `ethereum_ssz 0.5.4` later.

eserilev:
sounds good, will resolve the conflicts and get those dependencies pointed to the right versions

eserilev:
Updated to point to the correct dependencies and resolved all conflicts. Since we're not patching any deps in anymore, we're actually ok to use the existing alloy-primitives crate.

beacon_node/network/src/service.rs
```diff
@@ -809,7 +810,7 @@ impl<T: BeaconChainTypes> NetworkService<T> {
             }
         } else {
             for column_subnet in DataColumnSubnetId::compute_custody_subnets::<T::EthSpec>(
-                self.network_globals.local_enr().node_id().raw().into(),
+                U256::from_le_slice(self.network_globals.local_enr().node_id().raw().as_slice()),
```
eserilev: this can panic if the value is larger than what can fit in u256. that should never happen here, but flagging it just in case. theres a few other places in this PR where we've done things similarly.

michaelsproul:
I've just released new versions of all the packages depending on `alloy-primitives 0.8`, so we should be able to cut out the duplicate version of `ethereum_ssz` now!

Can you bump every dep by +0.1.0?

eserilev:
we cant bump alloy-primitives to 0.8 until this gets merged: https://github.com/alloy-rs/alloy/pull/1167

`alloy-consensus` still depends on `alloy-primitives` 0.7.7 and theres some diff between `alloy_primitives::TxKind` on 0.7 vs 0.8 that causes issues

maybe theres a way to work around this issue? not 100% sure. We can either wait for the new `alloy` release or figure out a workaround

EDIT: in 
11d8a26 I've commented out the problematic code and added a TODO to reintroduce it once `alloy` cuts a new release

consensus/types/src/attestation.rs
```diff
@@ -412,13 +412,7 @@ impl<E: EthSpec> AttestationBase<E> {
...
         &self,
     ) -> Result<BitList<E::MaxValidatorsPerSlot>, ssz_types::Error> {
-        let mut extended_aggregation_bits: BitList<E::MaxValidatorsPerSlot> =
-            BitList::with_capacity(self.aggregation_bits.len())?;
-
-        for (i, bit) in self.aggregation_bits.iter().enumerate() {
-            extended_aggregation_bits.set(i, bit)?;
-        }
-        Ok(extended_aggregation_bits)
+        self.aggregation_bits.resize::<E::MaxValidatorsPerSlot>()
```
eserilev, line 415: Snuck in this small change since the new `ssz_types` release has this function available

mattsse:
fyi new alloy-rs/alloy crate release coming later today, tmrw at the latest

eserilev:
alloy cut a release so were no longer blocked by that

beacon_node/beacon_chain/src/metrics.rs
```diff
@@ -2133,7 +2134,7 @@ fn set_gauge_by_epoch(gauge: &Result<IntGauge>, value: Epoch) {
 }
 
 fn set_gauge_by_hash(gauge: &Result<IntGauge>, value: Hash256) {
-    set_gauge(gauge, value.to_low_u64_le() as i64);
+    set_gauge(gauge, value.as_slice().get_i64());
```
michaelsproul: I'm a little bit hesitant about `get_i64` and similar functions because they can panic.

For use cases like metrics I would prefer something that might truncate or return junk if we make a mistake.

Maybe we could adds something like `to_low_u64_le()` to the mixin trait?

beacon_node/execution_layer/src/lib.rs
```diff
@@ -2197,14 +2197,12 @@ fn verify_builder_bid<E: EthSpec>(
     let header = &bid.data.message.header();
 
     // Avoid logging values that we can't represent with our Prometheus library.
-    let payload_value_gwei = bid.data.message.value() / 1_000_000_000;
-    if payload_value_gwei <= Uint256::from(i64::MAX) {
```
michaelsproul, line 2201: I think we probably want to keep this condition, or incorporate something similar into the safe mixin method for converting H256/U256 to i64

beacon_node/lighthouse_network/src/discovery/subnet_predicate.rs
```diff
@@ -41,7 +42,7 @@ where
                 .map_or(false, |b| b.get(*s.deref() as usize).unwrap_or(false)),
             Subnet::DataColumn(s) => {
                 let mut subnets = DataColumnSubnetId::compute_custody_subnets::<E>(
-                    enr.node_id().raw().into(),
+                    U256::from_le_slice(enr.node_id().raw().as_slice()),
```
michaelsproul: We might want big-endian here, yeah?

beacon_node/lighthouse_network/src/peer_manager/peerdb/peer_info.rs
```diff
@@ -374,7 +375,7 @@ impl<E: EthSpec> PeerInfo<E> {
         if self.custody_subnets.is_empty() {
             if let Ok(custody_subnet_count) = meta_data.custody_subnet_count() {
                 let custody_subnets = DataColumnSubnetId::compute_custody_subnets::<E>(
-                    node_id.raw().into(),
+                    U256::from_le_slice(&node_id.raw()),
```
michaelsproul: Big-endian again?

I'm actually wondering if a better API would be to pass `[u8; 32]` to these compute functions, and do the endianness conversion to `U256` inside.

beacon_node/network/src/service.rs
michaelsproul: this is another LE

beacon_node/network/src/service/tests.rs
```diff
@@ -176,7 +176,7 @@ mod tests {
         // Make sure the service is subscribed to the topics.
         let (old_topic1, old_topic2) = {
             let mut subnets = SubnetId::compute_subnets_for_epoch::<MinimalEthSpec>(
-                network_globals.local_enr().node_id().raw().into(),
+                Uint256::from_le_slice(&network_globals.local_enr().node_id().raw()),
```
michaelsproul: another LE (I think lets do the `[u8; 32]` API and that will force a refactor of all of these)

common/eth2/src/types.rs
```diff
@@ -1790,12 +1793,12 @@ impl TryFrom<&HeaderMap> for ProduceBlockV3Metadata {
             })?;
         let execution_payload_value =
             parse_required_header(headers, EXECUTION_PAYLOAD_VALUE_HEADER, |s| {
-                Uint256::from_dec_str(s)
+                Uint256::from_str(s)
```
michaelsproul: To maintain the same behaviour I think we should use `from_str_radix(s, 10)` here. The `from_str` function accepts hex, octal and binary inputs which we should probably reject:

https://docs.rs/ruint/1.12.3/src/ruint/string.rs.html#108-121

common/eth2/src/types.rs
```diff
@@ -1790,12 +1793,12 @@ impl TryFrom<&HeaderMap> for ProduceBlockV3Metadata {
...
                     .map_err(|e| format!("invalid {EXECUTION_PAYLOAD_VALUE_HEADER}: {e:?}"))
             })?;
         let consensus_block_value =
             parse_required_header(headers, CONSENSUS_BLOCK_VALUE_HEADER, |s| {
-                Uint256::from_dec_str(s)
+                Uint256::from_str(s)
```
michaelsproul: As above.

consensus/types/src/execution_transaction.rs
michaelsproul, line 1: This file can be deleted

crypto/bls/src/lib.rs
```diff
@@ -33,18 +33,70 @@ mod zeroize_hash;
...
 
-pub type Hash256 = ethereum_types::H256;
+pub type Hash256 = alloy_primitives::B256;
+pub trait FixedBytesExtended {
+    fn from_low_u64_be(value: u64) -> Self;
+    fn from_low_u64_le(value: u64) -> Self;
+    fn zero() -> Self;
+}
+
+impl<const N: usize> FixedBytesExtended for FixedBytes<N> {
```
michaelsproul: Rather than creating 3 copies of this trait, maybe we should pull it out into its own crate in `consensus` which is pulled in by `types`, `merkle_proof` and `bls`?

We could call the crate `fixed_bytes`

testing/ef_tests/src/cases/epoch_processing.rs
```diff
@@ -201,6 +201,7 @@ impl<E: EthSpec> EpochTransition<E> for PendingBalanceDeposits {
     }
 }
 
+// TODO(alloy) test fails here
```
michaelsproul: This comment can be removed as the tests pass now

testing/ef_tests/src/cases/fork_choice.rs
```diff
@@ -33,6 +33,7 @@ use types::{
 #[serde(deny_unknown_fields)]
 pub struct PowBlock {
     pub block_hash: ExecutionBlockHash,
+
```
michaelsproul: May as well avoid this newline

```suggestion
```

michaelsproul (changes requested):
Looks great on the whole!

I just have a couple of suggestions and then we can merge!

beacon_node/execution_layer/src/lib.rs
eserilev: Introduced a safe mixin here: [e555728](https://github.com/sigp/lighthouse/pull/6078/commits/e555728d25bb960595f4eef6f64ea074317921ad)

beacon_node/beacon_chain/src/metrics.rs
eserilev: Introduced `to_low_u64_le()` here: [4cf9de8](https://github.com/sigp/lighthouse/pull/6078/commits/4cf9de815d4d24e802d07677318c388f72adc0a2)

common/eth2/src/types.rs
eserilev: a7876e7

common/eth2/src/types.rs
eserilev: a7876e7

crypto/bls/src/lib.rs
eserilev: Implemented in [5162316](https://github.com/sigp/lighthouse/pull/6078/commits/5162316d8927b69451837468c211c69576a1fbf3)

testing/ef_tests/src/cases/epoch_processing.rs
eserilev: deleted TODO

testing/ef_tests/src/cases/fork_choice.rs
eserilev: deleted newline

consensus/types/src/execution_transaction.rs
eserilev: deleted

beacon_node/network/src/service/tests.rs
eserilev: updated: 
bbeb7de

beacon_node/lighthouse_network/src/peer_manager/peerdb/peer_info.rs
eserilev: updated api here:  bbeb7de

beacon_node/lighthouse_network/src/discovery/subnet_predicate.rs
eserilev: bbeb7de

eserilev:
this should be ready for another review!

Cargo.toml
```diff
@@ -205,6 +204,7 @@ eth2_keystore = { path = "crypto/eth2_keystore" }
 eth2_network_config = { path = "common/eth2_network_config" }
 eth2_wallet = { path = "crypto/eth2_wallet" }
 execution_layer = { path = "beacon_node/execution_layer" }
+fixed_bytes = { path = "consensus/fixed_bytes" }
```
michaelsproul, line 208: I think this also needs to be added to `[workspace.members]` above, but I'm not 100% on that

michaelsproul (approved):

michaelsproul:
@mergify queue

PR #6079, opened 2024-07-12
Add requires `suggested-fee-recipient` flag when `always-prepare-payload` is set

chong-he:
Currently, when the flag `--always-prepare-payload` is set and a fee recipient is not set in the beacon node, the beacon node will print the log:

`CRIT Fee recipient unknown                   proposer_index: 368409, msg: the suggested_fee_recipient was unknown during block production. a junk address was used, rewards were lost! check the --suggested-fee-recipient flag and VC configuration., service: exec`

which can be noisy and not desirable. This PR changes it so that the flag ` --suggested-fee-recipient` must always be set when `--always-prepare-payload` is set. The documentation is also updated to reflect this change.

Thanks to the discussion raised by @just_in_time and @michaelsproul in Discord

### DIFF
--- a/beacon_node/src/cli.rs
+++ b/beacon_node/src/cli.rs
@@ -1333,6 +1333,7 @@ pub fn cli_app() -> Command {
                 .action(ArgAction::SetTrue)
                 .help_heading(FLAG_HEADER)
                 .display_order(0)
+                .requires("suggested-fee-recipient")
         )
         .arg(
             Arg::new("fork-choice-before-proposal-timeout")
--- a/book/src/api-bn.md
+++ b/book/src/api-bn.md
@@ -128,7 +128,7 @@ You can replace `1` in the above command with the validator index that you would
 
 ### Events API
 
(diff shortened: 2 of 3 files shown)


michaelsproul:
Based on the test failure it seems like `suggested-fee-recipient` requires `execution-endpoint`, so you need to set that flag as well

chong-he:
> Based on the test failure it seems like `suggested-fee-recipient` requires `execution-endpoint`, so you need to set that flag as well

Thank you!
Added in [5ed5de4](https://github.com/sigp/lighthouse/pull/6079/commits/5ed5de40af243db0d00009af3aeef9705afe786d) and [1375779](https://github.com/sigp/lighthouse/pull/6079/commits/1375779b04f98bd4b990532d5583f5afeddb681e) and looks good now

realbigsean (approved):

realbigsean:
@mergify queue

michaelsproul:
@mergify requeue

PR #6083, opened 2024-07-12
Unimplement `TreeHash` for `BeaconState`

michaelsproul:
## Issue Addressed 

Since we moved to `tree-states` the `tree_hash_root` method on `BeaconState` became potentially dangerous. It would **panic** if any pending updates were waiting to be applied to the tree. We've done a good job of vetting that this doesn't happen on production code paths, but it's easy to do accidentally in tests, and I'm worried about regressions.

More background here:

- https://github.com/sigp/milhouse/issues/43

## Proposed Changes

- Remove the trait implementation of `TreeHash` for `BeaconState` as a safeguard against accidental misuse.
- Change the `canonical_root` method to call the safe mutable method `update_tree_hash_cache`.
- Update all usages (mostly tests). I've used `unwrap` liberally in tests as hashing a valid state should never fail.

### DIFF
--- a/beacon_node/beacon_chain/src/block_verification.rs
+++ b/beacon_node/beacon_chain/src/block_verification.rs
@@ -93,7 +93,6 @@ use std::io::Write;
 use std::sync::Arc;
 use store::{Error as DBError, HotStateSummary, KeyValueStore, StoreOp};
 use task_executor::JoinHandle;
-use tree_hash::TreeHash;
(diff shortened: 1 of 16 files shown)


dapplion (approved):
Quite painless change to remove this nasty footgun. Future mantainers will appreciate :)

michaelsproul:
@mergify queue

PR #6087, opened 2024-07-12
Update Dockerfile to work on latest version of Docker

jimmygchen:
## Issue Addressed

After updating my Docker version to latest (27.0.3), I get the following error when trying to build a Lighthouse image. Looks like the legacy format is no longer supported. This PR updates the Dockerfile to use the `ENV key=value` format instead of `ENV key value`.

```
 => CANCELED [builder 4/4] RUN cd lighthouse && make                                                                           10.7s

 2 warnings found (use --debug to expand):
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 7)
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 8)
ERROR: failed to solve: Canceled: context canceled
```


michaelsproul (approved):
confirmed working on Docker 25 as well (the version on my Ubuntu machine)

michaelsproul:
@mergify queue

PR #5510, opened 2024-04-02
Add timing for block availability

michaelsproul:
## Issue Addressed

Presently the block import metrics that measure block observation delay and import delay (time to import a block) are unaware of blob-processing and related delays.

This can lead to the `import_delay` seeming artificially high because it also factors in the time spent waiting for blobs to arrive.

## Proposed Changes

- [x] Add an `available` timestamp to the block delay cache which records when a block became fully available.
- [x] Add a similar `attestable` timestamp that records when a block became capable of being attested to (this allows us to ignore the non-critical disk writes and similar that are present in the `import_delay`). Revived from an old PR: https://github.com/sigp/lighthouse/pull/3232.
- [x] Log these delays _every slot_ (not just for late blocks), in milliseconds.
- [x] Add a `debug` log with the exact time taken by each `newPayload` call to the EL. This helps per-block investigations and is clearer and more accurate than the metric histogram. We can also use the exact `ms` value combined with log timestamp and the `observed_delay` to calculate the time taken doing other stuff prior to calling the EL (`log_timestamp - slot_start - time_taken_ms - observed_delay`).
- [x] Add matching prometheus metrics.

## Additional Info

The `available` delay currently reported by this PR is much higher than I expected, and sometimes seems to include things like snapshot cache misses. I need to do some more investigating to make sure I've put the observation of availability at the earliest possible point.

PR #6089, opened 2024-07-15
Measure consensus verification time

michaelsproul:
## Issue Addressed

We added a metric for "execution verification time" in a previous PR, but it actually captured consensus verification time as well:

- https://github.com/sigp/lighthouse/pull/5510

## Proposed Changes

- Add a new metric to the block timing cache, the `consensus_verification_time` which is the delay between the block being observed and it being fully validated from Lighthouse's PoV (parent state loaded & state transition complete).
- Fix the execution time metric so it measures the time between the request to the EL starting and finishing.

## Additional Info

We should ideally update the missing attestations blog post:

- https://lighthouse-blog.sigmaprime.io/attestation-analysis.html

As well as the dashboard in `lighthouse-metrics`:

- https://github.com/sigp/lighthouse-metrics/blob/master/dashboards/MissedAttestations.json

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -3088,14 +3088,21 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
                 notify_execution_layer,
             )?;
             publish_fn()?;
+
+            // Record the time it took to complete consensus verification.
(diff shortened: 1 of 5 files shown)


jimmygchen (approved):
The changes make sense to me 👍

michaelsproul:
@mergify queue

Issue #6075, opened 2024-07-11
`lighthouse account validator` subcommands are not displaying CLI help text


chong-he:
## Description

Since 5.2.0, `lighthouse account validator` subcommands are not displaying the help text. For example, `lighthouse account validator exit --help` returned:

`error: unexpected argument '--help' found`

Similar for `lighthouse account validator import --help` and other subcommands

## Version

5.2.0 and 5.2.1

I will look into fixing this

eserilev:
nice catch! if theres anything I can do to help here, please let me know

chong-he:
Closed by #6091

PR #6091, opened 2024-07-15
Fix `lighthouse account validator` subcommands help text

chong-he:
## Issue Addressed

* #6075  

Tested and it's working, though not sure if there is an easier way of doing it


michaelsproul:
CK and I discussed a simpler approach offline (using `.global(true)`), so putting this on pause until we try it

chong-he:
> CK and I discussed a simpler approach offline (using `.global(true)`), so putting this on pause until we try it

Thank you so much! Brings the diff to only 2 lines, tested and it's working

michaelsproul (approved):

michaelsproul:
Nice!

michaelsproul:
@mergify queue

AgeManning:
@mergifyio refresh

jimmygchen:
@mergify requeue

PR #6092, opened 2024-07-15
Bump default `logfile-max-number` to 10

jimmygchen:
## Issue Addressed

This PR bumps the default `logfile-max-number` from `5` to `10`.

When the beacon node has a large number of validators attached to it, the log file gets filled up and sometimes gets rotated very quickly. This gets a bit worse if the node is overwhelmed / behind. It would be better if the debug logs can stay on the machine for at least 2~3 days, so that we have enough time to respond when an issue occurs. There's been multiple instances where we request logs from user and the old logs just got rotated out.

The current default limits the total logs stored on disk to 1Gb, this change would increase it to 2Gb, which isn't too bad.

### DIFF
--- a/book/src/help_bn.md
+++ b/book/src/help_bn.md
@@ -241,7 +241,7 @@ Options:
           [possible values: DEFAULT, JSON]
       --logfile-max-number <COUNT>
           The maximum number of log files that will be stored. If set to 0,
-          background file logging is disabled. [default: 5]
+          background file logging is disabled. [default: 10]
       --logfile-max-size <SIZE>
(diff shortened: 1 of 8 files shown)


michaelsproul (approved):

michaelsproul:
Flagging this as `backwards-incompat` even though it isn't, just so we remember to mention it in release notes

michaelsproul:
@mergify queue

realbigsean:
@mergify queue

jimmygchen:
@mergify requeue

PR #6093, opened 2024-07-15
Enable the outbound rate limiter by default, and update blobs method quotas

jimmygchen:
## Issue Addressed

Enable outbound rate limiter by default to prevent sending execessive amount of requests the peers in case of bugs.

There are **no known issues** on the `stable` branch that would require this change. However, such issues have occurred on some feature branches, and it makes sense to enable the rate limiter by default. This is also important in preparation for PeerDAS, to moderate the speed at which nodes perform sampling during sync.


AgeManning (approved):

realbigsean:
@mergify queue

PR #6095, opened 2024-07-15
Add range sync metrics to track efficiency

dapplion:
## Issue Addressed

To benchmark some potential changes to range sync I want to answer a few questions:
- How often do chains rotate into new chains? = rate of syncing chain creation and removal
- When chains rotate, how many blocks are "wasted" = dropped?
- How long do buffered batches stay in AwaitingProcessing state? Are we buffering too much?
- How many downloaded blocks are ignored by the processor? 

This PR does not intend to change any logic in range sync itself. All changes are purely view to carry expose the data necessary for metrics

### DIFF
--- a/beacon_node/network/src/metrics.rs
+++ b/beacon_node/network/src/metrics.rs
@@ -252,6 +252,16 @@ lazy_static! {
         "Total count of dropped blocks when removing a syncing chain per range type",
         &["range_type"]
     );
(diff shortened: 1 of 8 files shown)


realbigsean (approved):
LGTM

realbigsean:
@mergify queue

PR #6103, opened 2024-07-16
Remove trace logging

AgeManning:
In practice I do not use trace logging and I don't know if anyone else does. 

The way slog works is it passes these logs into its own task and then filters based on the log level. We are currently compiling with trace logging, which I would imagine has a non-negligible performance hit by producing all the trace logging then filtering them out. 

Even when we are debugging we rarely use trace-level logging (to my knowledge). Even if we wanted to use them, we would have to restart the client as most default configurations favour debug-level. 

For those really wanting trace-level logging, I propose they re-compile lighthouse to enable it, to spare everyone else not using tracing from generating and filtering the logs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -154,7 +154,7 @@ serde_json = "1"
 serde_repr = "0.1"
 serde_yaml = "0.9"
 sha2 = "0.9"
-slog = { version = "2", features = ["max_level_trace", "release_max_level_trace", "nested-values"] }
+slog = { version = "2", features = ["max_level_debug", "release_max_level_debug", "nested-values"] }
 slog-async = "2"
 slog-term = "2"
 sloggers = { version = "2", features = ["json"] }


michaelsproul (approved):
Yeah I support this. I would even be on board with deleting all uses of `trace!` eventually, due to how useless they are.

Usually if we want to debug something, we'll add a `debug` log.

michaelsproul:
I think this is a breaking change because users who are running with `--debug-level trace` will get an error.

AgeManning:
> I think this is a breaking change because users who are running with `--debug-level trace` will get an error.

Will they get an error, or will it just not output anything?

michaelsproul:
I'll compile and test it

michaelsproul:
Ok yeah it's fine, it doesn't error, it just drops the trace logs.

michaelsproul:
@mergify queue

dapplion:
The instance of useful trace logs I have seen is in debugging sync tests, where the log is valuable on a single run of sync 1 block but would be to noisy if enabled in prod. We could replace the trace log with
```rust
#[cfg(test)]
debug!()
```

PR #6115, opened 2024-07-16
Disable slasher default features

SamWilsn:
## Issue Addressed

Because slasher had lmdb as a default feature, it was impossible to have a pure mdbx build of lighthouse.

## Proposed Changes

This commit sets `default-features = false` for slasher so lighthouse can actually choose one or the other.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -224,7 +224,7 @@ pretty_reqwest_error = { path = "common/pretty_reqwest_error" }
 proto_array = { path = "consensus/proto_array" }
 safe_arith = { path = "consensus/safe_arith" }
 sensitive_url = { path = "common/sensitive_url" }
-slasher = { path = "slasher" }
+slasher = { path = "slasher", default-features = false }
 slashing_protection = { path = "validator_client/slashing_protection" }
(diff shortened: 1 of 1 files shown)


michaelsproul:
The use case here is that you want to build Lighthouse with `cargo build --release --no-default-features` and have that exclude LMDB, right? You don't mind building with the extra `--no-default-features` flag?

SamWilsn:
> The use case here is that you want to build Lighthouse with `cargo build --release --no-default-features` and have that exclude LMDB, right?

No particular use case in mind. Just seemed odd that the lighthouse crate had a `slasher-lmdb` feature that, when disabled, didn't actually disable the `lmdb` feature in the slasher crate.

> You don't mind building with the extra `--no-default-features` flag?

This is part of my "packaging Lighthouse for Gentoo" saga. In Gentoo, the default for Rust crates is to compile with `--no-default-features` and have [USE flag](https://wiki.gentoo.org/wiki/USE_flag)s that map to the features of the crate.

michaelsproul (approved):

michaelsproul:
@mergify queue

Issue #6107, opened 2024-07-16
Add option to select KZG library for PeerDAS


jimmygchen:
## Description

We've recently switched from the `c-kzg` library to the Rust `peerdas-kzg` library on the `das` branch (#5941). The purpose for the switch was to:
1. Try out a new alternative to c-kzg; 
2. Allow us to merge `das` branch to `unstable` without having to update `c-kzg` to a development version

However we've now lost the option to run the `das` branch with `ckzg`, which has been quite stable for a while. We've recently run into a few issues with `peerdas-kzg`, including slow startup time and stack overflow during block production (#6105). Therefore I think it would be great to have the option to use `ckzg` library, especially for public denvets, where a small number of reliable supernodes is required.

Ideally we want to be able to:
1. maintain the ckzg library version for the mainnet code path
2. optionally include the ckzg peerdas library when it's enabled via a feature flag, so both ckzg library will co-exist when this `ckzg-peerdas` feature is enabled, however this library will only be referenced by the PeerDAS portion of the code.
3. allow selecting the ckzg-peerdas library via a hidden runtime flag

jimmygchen:
We currently use c-kzg for Deneb, and rust-eth-kzg for PeerDAS. To fully support this we'd need:
- Re-integrate PeerDAS with c-kzg, so we can have the option to load c-kzg only
- Use rust-eth-kzg to perform deneb KZG operations, so we can have the option to load rust-eth-kzg only
- If we complete both changes, then we can have option to choose which KZG library to use, or feature gate them (#6107)

rust-eth-kzg seems to perform better in PeerDAS - I'm not sure if this has changed.

Alternatively, an option is to maintain the status quo for now, and gradually phase out `c-kzg` over time? We can't completely switch over to rust KZG yet, AFAIK it hasn't been audited, but I'd assume there are plans to do this.

jimmygchen:
@pawanjay176 also expressed his preference to maintain the status quo now and eventually get rid of `c-kzg`. It's unclear whether both libraries will be maintained in the longer term.

PR #6117, opened 2024-07-16
chore: update the trusted setup to match the consensus-specs version

kevaundray:
## Issue Addressed

This adds the exact trusted file that is in the consensus-specs. The previous(current) file was a truncated version that did not include the g1_monomial points. If these are not included, then the KZG libraries will need to recompute them since peerDAS requires them.

## Proposed Changes

This PR copies the file from the consensus-specs repo and overwrites the previous one. The g1_lagrange and g2_monomial points are the same. Since I wanted to match the checksum in the consensus-specs repository, this does change the formatting. 

## Additional Info

Please provide any additional information. For example, future considerations
or information useful for reviewers.

PR #6119, opened 2024-07-16
chore: Add c-kzg library behind a runtime flag

kevaundray:
## Issue Addressed

This adds back the c-kzg library (updated to the latest commit).

This PR is reliant on https://github.com/sigp/lighthouse/pull/6117 since the new c-kzg API requires g1_monomial. Once that gets merged, then the lines modified should reduce.

## Proposed Changes

## Additional Info


kevaundray:
This closes #6107 though note that we cannot have two c-kzg libraries in this repository because the c-kzg library links to a c static lib, so the mainnet path will always be "affected" -- Another solution would be to maintain a branch that gets merged into stable which has the rust library and the c-kzg-4844 code. Then another branch which is used for peerdas which has both c-kzg-peerdas and the rust lib. This mainly just imposes a cost to maintaining both branches.

crypto/kzg/src/lib.rs
```diff
@@ -174,20 +192,30 @@ impl Kzg {
...
+            Ok((cells, proofs))
+        } else {
+            let blob_bytes: &[u8; BYTES_PER_BLOB] = blob
+                .as_ref()
+                .try_into()
+                .expect("Expected blob to have size {BYTES_PER_BLOB}");
+
+            let (cells, proofs) = self
+                .context
+                .prover_ctx()
```
kevaundray, line 211: Ideally two new files are created for these two codepaths

crypto/kzg/src/lib.rs
```diff
@@ -67,13 +72,26 @@ impl Kzg {
 
         Ok(Self {
             trusted_setup: KzgSettings::load_trusted_setup(
-                &trusted_setup.g1_points(),
-                &trusted_setup.g2_points(),
+                &trusted_setup.g1_monomial_points(),
+                &trusted_setup.g1_lagrange_points(),
+                &trusted_setup.g2_monomial_points(),
```
jtraglia, line 77: Since we changed this from a list of points to a list of flat bytes, I think we should update the names.
```suggestion
                &trusted_setup.g1_monomial_bytes(),
                &trusted_setup.g1_lagrange_bytes(),
                &trusted_setup.g2_monomial_bytes(),
```

crypto/kzg/src/trusted_setup.rs
```diff
@@ -21,117 +11,36 @@ struct G2Point([u8; BYTES_PER_G2_POINT]);
...
+        self.g1_monomial_points.iter().flat_map(|p| {
+            let stripped = strip_prefix(p);
+            hex::decode(stripped).expect("expected g1 monomial points to be well formed hex strings")
+        }).collect()
+    }
+    pub fn g2_monomial_points(&self) -> Vec<u8> {
+        self.g2_monomial_points.iter().flat_map(|p|{
+            let stripped = strip_prefix(p);
+            hex::decode(stripped).expect("expected g2 monomial points to be well formed hex strings")
+        }).collect()
```
jtraglia, line 39: Small nit, I've been trying to use consistent order for these: g1_monomial, g1_lagrange, g2_monomial.

Can we move the middle function to the top?

crypto/kzg/src/lib.rs
```diff
@@ -174,20 +192,30 @@ impl Kzg {
...
+
+            let (cells, proofs) = self
+                .context
+                .prover_ctx()
+                .compute_cells_and_kzg_proofs(blob_bytes)
+                .map_err(Error::ProverKZG)?;
+
+            let c_kzg_proof = proofs.map(KzgProof);
+
+            Ok((cells, c_kzg_proof))
```
jtraglia, line 217: This should be plural. Plus minor `c_kzg` vs `ckzg` nit. Going with `ckzg`, like `use_ckzg`.

```suggestion
            let ckzg_proofs = proofs.map(KzgProof);

            Ok((cells, ckzg_proofs))
```

jimmygchen:
@kevaundray I had a chat with @michaelsproul and he suggested that it might not be a good idea to have two versions of `c-kzg` - we had some issues with it before and had to rename a bunch of things to avoid symbol collisions. 

So it might be better to hold off on this one and add this switch once the `c-kzg` main version supports PeerDAS. Right now we're going to focus on getting our `das` stuff onto `unstable` (main dev branch).

michaelsproul:
Now that we've merged `das` to `unstable` please update to point at `unstable` by either:

Issue #6099, opened 2024-07-15
Allow to sync a long unknown fork


dapplion:
## Description

Currently, Lighthouse cannot sync a long unknown fork its current head is progressing. The issue is rooted in how range sync considers peers "Advanced". It will trigger head range sync if the peer's head is sufficiently head, at least 1 epoch ahead.

If Lighthouse is currently at the purple branch, knows nothing about the blue branch, and receives `N` it will trigger lookup sync. However, lookup sync will error if the chain is longer than 2 epochs and downscore all peers. This protects the node against peers sending us malicious garbage chains that we can't validate.

![image](https://github.com/user-attachments/assets/eb2730e9-eba5-4146-a7d0-7bb42ff97a07)

## Potential solution

We can consider `Status` messages as another source to trigger unknown block lookups. If range sync receives a `Status` message that considers `FullySynced` but includes an unknown root, relay that to lookup sync.

If lookup sync reaches this max depth it could tell range sync to force a head sync against that target, despite the head having a similar epoch. For example, it can send a message to sync manager 

```rust
    AddPeersForceRangeSync {
        peers: Vec<PeerId>,
        head_root: Hash256,
        head_slot: Option<Slot>,
    },
```

which triggers the creation of a head syncing chain skipping the logic of `remote_sync_type` forcing that set of peers to be considered `Advanced`.

However, there's an issue with attributability. Range sync will ask for the peers for the same set of blocks ***again***. So if malicious peers send us garbage first, they can send the correct data later. CC @AgeManning

PR #6122, opened 2024-07-17
Transition block lookup sync to range sync

dapplion:
## Issue Addressed

For context of the issue 👇 

- https://github.com/sigp/lighthouse/issues/6099

Closes https://github.com/sigp/lighthouse/issues/6099

## Proposed Changes

This solution links range sync and lookup sync by:
- Trigger lookup sync with status messages that include unknown root. Currently range sync "swallows" peer statuses that are too close to head but on a potentially unknown fork. Range sync is not suitable to handle short forks, so instead deffer to lookup sync to retrieve that unknown root.
- If lookup sync reaches the max chain length, trigger range sync on that set of peers. The idea is that _most_ peers in that group should have this chain as canonical so we can `blocks_by_range` from them. In the future we may use https://github.com/ethereum/consensus-specs/pull/3845 to be sure what fork are we syncing from.

I believe this change is the minimal solution to cover the corner case of #6099. It's not maximally efficient as it will download a section of blocks twice. However this situation is rare, and we just need _some_ way to recover even if it's at the expense of some extra bandwidth.

## Testing

There are plans to with EF devops to test nasty long non-finality situations, but not now. For now I can try to write representative unit sync tests with some sample scenarios. This change _should_ not impact good network codepaths, so should have no downsides. Range sync will only be triggered when we need to discover a long unknown fork, which has only happened in LH in cooked testnets of lookup sync bugs.


beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -248,22 +252,48 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
...
+                {
+                    // If a lookup chain is too long we can't distinguish a valid chain between a
+                    // malicious chain. We must attempt to sync this chain to not lose liveness. If
+                    // the chain grows too long we stop lookup sync and transition this head to
+                    // forwards range sync. We need to tell range sync what head to sync too and
+                    // from which peers. The lookup of the very tip of this chain may contain zero
+                    // peers if it's the parent child lookup. So we do a bit a trick here:
+                    // - Tell range sync to sync to the tip's root (if available, else its ancestor)
+                    // - But use all peers in the ancestor lookup, which should have at least one
+                    //   peer and its peer set is a strict superset of the tip's lookup.
```
kevaundray:
```suggestion
                    // If a lookup chain is too long, we can't distinguish a valid chain from a 
                    // malicious one. We must attempt to sync this chain to not lose liveness. If 
                    // the chain grows too long, we stop lookup sync and transition this head to 
                    // forward range sync. We need to tell range sync which head to sync to, and 
                    // from which peers. The lookup of the very tip of this chain may contain zero 
                    // peers if it's the parent-child lookup. So we do a bit of a trick here:
                    // - Tell range sync to sync to the tip's root (if available, else its ancestor)
                    // - But use all peers in the ancestor lookup, which should have at least one 
                    //   peer, and its peer set is a strict superset of the tip's lookup.
```

Minor typo nit that reads clearer for me

dapplion:
Updated tests to assert a syncing chain is created. Plus removed down-scoring the peers for a long chain

https://github.com/sigp/lighthouse/blob/4ec5594f4bf507b54542cf9a6b2943b322c90460/beacon_node/network/src/sync/block_lookups/mod.rs#L294-L299

As a final item we could check that for a syncing chain if it reaches the end, assert that the block root matches that syncing root

beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -252,22 +256,52 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
...
-                                peer_id,
-                                PeerAction::LowToleranceError,
-                                "chain_too_long",
-                            );
+                        Some((_, tip_lookup)) => {
+                            (parent_chain_tip, tip_lookup.peek_downloaded_block_slot())
                         }
-                        self.drop_lookup_and_children(*lookup_id);
-                    }
+                        None => (lookup.block_root(), lookup.peek_downloaded_block_slot()),
```
AgeManning: I ran out of time looking through to find out the scenario when this occurs. 

Its when we don't have any single block lookups that match the current parent tip. In this case we start a range sync based on the parents ancestor, rather than the parent_chain_tip? 

I didn't fully parse this. 

Overall I like the idea!

dapplion: This should never happen. I can log an error and return, but used this fallback value for "simplicity". Happy to change if this confuses more than helps

AgeManning: Oh yeah. I think that might be helpful. This way we can catch any wild states rather than doing a range sync. Also (at least for me, will make it easier to follow).

AgeManning (approved):
Yeah I think we should test this out. Logic seems reasonable to me.

realbigsean (approved):
LGTM

realbigsean:
@mergify queue

realbigsean:
@mergify refresh

realbigsean:
@mergify queue

Issue #6105, opened 2024-07-16
Investigate PeerDAS KZG library stack overflow during block production


jimmygchen:
## Description

We've recently switched from the `c-kzg` library to the Rust `peerdas-kzg` library on the `das` branch (#5941). The purpose for the switch was to:
1. Try out a new alternative to c-kzg; 
2. Allow us to merge `das` branch to `unstable` without having to update `c-kzg` to a development version

However we've recently ran into some issues during interop testing after the change, and we're getting stack overflow panics, likely during block production:

```
Jul 15 10:03:55.007 DEBG Fork choice successfully updated before block production, fork_choice_slot: 60, slot: 60, service: beacon
Jul 15 10:03:55.007 DEBG Not attempting re-org                   reason: head not late, service: beacon
Jul 15 10:03:55.008 ERRO No valid eth1_data votes, `votes_to_consider` empty, outcome: casting `state.eth1_data` as eth1 vote, genesis_time: 1721037115, earliest_block_timestamp: 1721037095, lowest_block_number: 0, service: deposit_contract_rpc
Jul 15 10:03:55.008 DEBG Produced vote for eth1 chain            block_hash: 0xcc93bda48e1227e94e768ee757160974267c72c872f4c89213fbee552bb2ecf2, deposit_count: 0, deposit_root: 0xd70a234731285c6804c2a4f56711ddb8c82c99740f207854891028af34e27e5e, service: deposit_contract_rpc
Jul 15 10:03:55.008 DEBG Issuing engine_getPayload               parent_hash: 0xd32b8265b9e5694d48a42d14175d2692fdb918d305b91d336f2fd2c3167a8df5, timestamp: 1721037835, prev_randao: 0x4c74c85276c06a77410fb641d54224b09045bd03cf19c131ac158b534cd31ad6, suggested_fee_recipient: 0x8943545177806ed17b9f23f0a21ee5948ecaa776, service: exec
Jul 15 10:03:55.011 DEBG Processed HTTP API request              method: GET, path: /eth/v1/node/version, status: 200 OK, elapsed: 86.916µs
Jul 15 10:03:55.103 DEBG Produced block on state                 block_size: 2141, service: beacon
Jul 15 10:03:55.115 DEBG Processed HTTP API request              method: GET, path: /eth/v2/validator/blocks/60, status: 200 OK, elapsed: 108.519167ms
[...]

PR #6125, opened 2024-07-18
Fix `proposer-nodes` cli flag name

jimmygchen:
## Issue Addressed

Getting this error while trying to run validator client locally (with a custom built docker image) and it shuts down immediately with:
```
Unable to initialize validator config: Unable to parse proposer_nodes: Unknown argument or group id.  Make sure you are using the argument id and not the short or long flags
```

This PR fixes the `--proposer-node` flag parsing to use the correct arg name.

https://github.com/sigp/lighthouse/blob/947e2e8db1646ce56b5fdf9e04cd2b80d5b564ff/validator_client/src/cli.rs#L33-L35

### DIFF
--- a/lighthouse/tests/validator_client.rs
+++ b/lighthouse/tests/validator_client.rs
@@ -2,6 +2,7 @@ use validator_client::{config::DEFAULT_WEB3SIGNER_KEEP_ALIVE, ApiTopic, Config};
 
 use crate::exec::CommandLineTestExec;
 use bls::{Keypair, PublicKeyBytes};
+use sensitive_url::SensitiveUrl;
 use std::fs::File;
 use std::io::Write;
 use std::net::IpAddr;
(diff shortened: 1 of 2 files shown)


jimmygchen:
Note this was tested on a custom built image. I can't reproduce the crash using an image built using our Dockerfile, so it could be that the image wasn't built with `--locked`, and potentially using a different clap version.

michaelsproul:
I can't reproduce the crash, even building without the lockfile (`cargo install --path lighthouse`). Were you using the `--proposer-nodes` flag?

Regardless, I think we should fix it

jimmygchen:
Nope I wasn’t using the flag. Hmm really weird!

michaelsproul:
When did you build the image? I notice there were two clap versions released really close together: https://crates.io/crates/clap/versions

michaelsproul (approved):

michaelsproul:
Can we add a test in `lighthouse/tests/validator_client.rs` so this doesn't regress?

jimmygchen:
I used this image: `kevaundray/lighthouse-eth:das-dbg-04d9eef8a`

built by kev for this issue: 
https://github.com/sigp/lighthouse/issues/6105#issuecomment-2234281264

I’ve reach out to him and asked whether this was produced with our local testnet scripts.

jimmygchen:
Good idea, I’ll add a test!

michaelsproul (approved):
Nice

michaelsproul:
@mergify queue

PR #6127, opened 2024-07-18
Add PeerDAS presets and configs.

jimmygchen:
## Issue Addressed

Part of #6072

## Proposed Changes

This PR pulls all presets and configs related changes from the `das` branch.


consensus/types/src/chain_spec.rs
```diff
@@ -1426,6 +1477,10 @@ const fn default_max_request_blob_sidecars() -> u64 {
     768
 }
 
+const fn default_max_request_data_column_sidecars() -> u64 {
+    512
```
pawanjay176: Spec says 16384 🤔 
https://github.com/ethereum/consensus-specs/blob/dev/configs/mainnet.yaml#L163

pawanjay176 (changes requested):
Looks good except a config value

realbigsean (approved):
looks good other than pawan's comment

consensus/types/src/chain_spec.rs
jimmygchen: Good catch! Fixed

pawanjay176 (approved):

jimmygchen:
@mergify queue

jimmygchen:
@mergify requeue

jimmygchen:
@mergify requeue

jimmygchen:
@mergify requeue

PR #6131, opened 2024-07-18
Delete `--builder-profit-threshold` flag

chong-he:
There is a bug when using the deprecated flag `--builder-profit-threshold` in Lighthouse 5.2.0 and above, reported by @CouzDelCouzzz on Discord: https://discord.com/channels/605577013327167508/605577013331361793/1262698540358963261

To reproduce, run Lighthouse with the flag, e.g.:

`./lighthouse bn --checkpoint-sync-url https://mainnet.checkpoint.sigp.io --builder-profit-threshold 100 --execution-endpoint http://localhost:8551 --execution-jwt /home/hi/.ethereum/geth/jwtsecret`

@michaelsproul mentioned that this is due to the `parse_flag` should only be used when the flag does not take an argument (i.e., the flag is a boolean, such as the `--metrics` flag): https://github.com/sigp/lighthouse/blob/9e12c21f268c80a3f002ae0ca27477f9f512eb6f/beacon_node/src/config.rs#L1379-L1381

This is not the case for `--builder-profit-threshold` which takes an argument:
https://github.com/sigp/lighthouse/blob/9e12c21f268c80a3f002ae0ca27477f9f512eb6f/beacon_node/src/config.rs#L338-L344

This PR deletes the `--builder-profit-threshold` flag so that the panic is not possible to manifest.

### DIFF
--- a/beacon_node/src/cli.rs
+++ b/beacon_node/src/cli.rs
@@ -1408,14 +1408,6 @@ pub fn cli_app() -> Command {
                 .display_order(0)
         )
         .arg(
-            Arg::new("builder-profit-threshold")
-                .long("builder-profit-threshold")
-                .value_name("WEI_VALUE")
-                .help("This flag is deprecated and has no effect.")
-                .action(ArgAction::Set)
-                .display_order(0)
-        )
-        .arg(
             Arg::new("builder-user-agent")
                 .long("builder-user-agent")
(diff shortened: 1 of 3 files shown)


michaelsproul (approved):
Nice. Let's do it

michaelsproul:
@mergify queue

PR #6134, opened 2024-07-19
Attempt to fix flaky basic sim test

jimmygchen:
## Issue Addressed

Fix `target_peer` config for basic_sim, so a node doesn't disconnected by peers because of `TooManyPeers`.

I think the test fails when enough nodes disconnect one of the nodes and messages from that node don't get propagated to the network correctly.

### DIFF
--- a/testing/simulator/src/basic_sim.rs
+++ b/testing/simulator/src/basic_sim.rs
@@ -37,6 +37,8 @@ pub fn run_basic_sim(matches: &ArgMatches) -> Result<(), String> {
         .unwrap_or(&String::from("0"))
         .parse::<usize>()
         .unwrap_or(0);
+    // extra beacon node added with delay
+    let extra_nodes: usize = 1;
     println!("PROPOSER-NODES: {}", proposer_nodes);
     let validators_per_node = matches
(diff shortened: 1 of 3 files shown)


jimmygchen:
The test failed again, but I think the fix is still valid.

The `target_peer` issue was the reason this run failed: https://github.com/sigp/lighthouse/actions/runs/9986823946/job/27600016474

The [current run](https://github.com/sigp/lighthouse/actions/runs/10001418617/job/27644989879?pr=6134) is failing due to ~~the new node (node_6) having a non-optimistic head block. Investigating~~.

UPDATE: the above is not correct actually, node_6 has no validators attached so it wouldn't contribute to sync message issues.

The error is 
```
2024-07-19T02:32:12.8166404Z called `Result::unwrap()` on an `Err` value: "Sync aggregate at slot 17 was not full, got: 20, expected: 32"
```

so it looks like the sync contributions were not published for all signers, this is something we've experienced before and usually happens around fork boundaries.

michaelsproul (approved):

michaelsproul:
@mergify queue

jimmygchen:
@mergify requeue

PR #6140, opened 2024-07-21
Fix unexpected `Marking peer disconnected in DHT`

ackintosh:
## Issue Addressed

We get the log below when LH fails to dial a peer:

```
2024-07-19T02:30:14.4401703Z Jul 19 02:30:14.439 DEBG Marking peer disconnected in DHT        error: Connection denied: ConnectionDenied { inner: Exceeded { limit: 1, kind: EstablishedPerPeer } }, peer_id: 
```

I noticed that the disconnect happens unexpectedly in the following case:

image

Lighthouse dials to a peer twice using TCP and QUIC (if QUIC is not disabled). Usually, one establishes a connection, and the other fails because the peer allows only one connection per peer. I think we shouldn't `Marking peer disconnected in DHT` since a TCP (in this case) connection has been established between the nodes.

## Proposed Changes

I added a check to see if there’s an active connection before the disconnect.


beacon_node/lighthouse_network/src/discovery/mod.rs
```diff
@@ -1156,8 +1156,19 @@ impl<E: EthSpec> Discovery<E> {
...
+                    if self.network_globals.peers.read().is_connected(&peer_id) {
+                        // There's an active connection, so we don’t disconnect the peer.
+                        // Lighthouse dials to a peer twice using TCP and QUIC (if QUIC is not
+                        // disabled). Usually, one establishes a connection, and the other fails
+                        // because the peer allows only one connection per peer.
+                        return;
+                    }
+                    // set peer as disconnected in discovery DHT
+                    debug!(self.log, "Marking peer disconnected in DHT"; "peer_id" => %peer_id, "error" => %ClearDialError(error));
+                    self.disconnect_peer(&peer_id);
```
jxs, line 1169:
```suggestion
```
this seems redundant right? if the peer is not connected, we shouldn't need to disconnected it

jxs (commented):
LGTM Akihito, thanks! Left a comment

beacon_node/lighthouse_network/src/discovery/mod.rs
jxs: @age do you know why are we disconnecting from discovery on a dial failure instead of when a connection has been closed?

jxs (commented):
after yesterday's call where @jimmygchen and @AgeManning mentioned that this was occurring on a testnet with only Ligthouse clients I went to research this further.

It turns out that The [connection limits `NetworkBehaviour`](https://github.com/sigp/lighthouse/blob/stable/beacon_node/lighthouse_network/src/service/behaviour.rs#L25) currently [forbids](https://github.com/sigp/lighthouse/blob/stable/beacon_node/lighthouse_network/src/service/mod.rs#L380) us to have [more than 1](https://github.com/sigp/lighthouse/blob/stable/beacon_node/lighthouse_network/src/service/utils.rs#L25) connection per peer.
The error returned from the `Swarm` when a `NetworkBehaviour` rejects an outbound connection is `DialError::Denied`, so this PR addresses the issue it aims to.

beacon_node/lighthouse_network/src/discovery/mod.rs
AgeManning: In this case we are connected via discovery on UDP. However we cannot connect via libp2p. This could mean that the ENR they are advertising is not correct, so we want to remove them from our DHT so we don't advertise to others, because if we can't dial them, no one will be able to.

AgeManning (approved):
This looks good to me, in that is masks the error. 

However the deeper problem is that we are trying to connect to a peer twice. 

It seems to me the easiest solution is that mask this error and force sequential connections when dialing. So this PR and a future one together is the easiest solution I think

AgeManning:
@ackintosh  - I think you just need to merge lastest unstable to fix CI

beacon_node/lighthouse_network/src/discovery/mod.rs
jxs: yeah makes sense age, thanks!

jxs:
@mergifyio queue

PR #6146, opened 2024-07-22
Simple Subnet Management

AgeManning:
## Overview

This is an implementation of https://github.com/ethereum/consensus-specs/pull/3735

Essentially, rather than having Lighthouse rotate every 24 hours around long-lived subnets, it will stay fixed on subnets for the duration of its run (and it's lifetime until a user manually deletes the network.key in the lighthouse directory). 

The rotations have a dubious practical importance and adds quite a bit of complexity in our code base. By removing the rotations I was able to simplify the code. This also helps with debugging and keeping tracking of subnets and ENRs (our ENRs will also be stable, as we no longer need to rotate the subnet ids there). 

While I was in there, I took the liberty to combine the SyncCommitteeSubnets with the AttestationSubnets into a single `SubnetService` which I think is a bit cleaner and easier to reason about. 

Because this PR changes a bit of code and if there are bugs will immediately impact the performance of validators, we should take caution in merging this before significant testing.


consensus/types/src/subnet_id.rs
```diff
@@ -78,45 +77,15 @@ impl SubnetId {
...
+    pub fn compute_attestation_subnets<E: EthSpec>(
         node_id: ethereum_types::U256,
-        epoch: Epoch,
         spec: &ChainSpec,
-    ) -> Result<(impl Iterator<Item = SubnetId>, Epoch), &'static str> {
-        // simplify variable naming
-        let subscription_duration = spec.epochs_per_subnet_subscription;
+    ) -> impl Iterator<Item = SubnetId> {
+        // The bits of the node-id we are using to define the subnets.
         let prefix_bits = spec.attestation_subnet_prefix_bits as u64;
```
jimmygchen, line 88: I think your spec PR suggests replacing this with `int(ceillog2(ATTESTATION_SUBNET_COUNT)`?

beacon_node/network/src/service.rs
```diff
@@ -170,9 +167,7 @@ pub struct NetworkService<T: BeaconChainTypes> {
     /// The underlying libp2p service that drives all the network interactions.
     libp2p: Network<RequestId, T::EthSpec>,
     /// An attestation and subnet manager service.
```
jimmygchen:
```suggestion
    /// An attestation and sync committee subnet manager service.
```

beacon_node/network/src/subnet_service/tests/mod.rs
```diff
@@ -114,31 +117,20 @@ lazy_static! {
     static ref CHAIN: TestBeaconChain = TestBeaconChain::new_with_system_clock();
 }
 
-fn get_attestation_service(
-    log_level: Option<slog::Level>,
-) -> AttestationService<TestBeaconChainType> {
-    let log = get_logger(log_level);
+fn get_subnet_service() -> SubnetService<TestBeaconChainType> {
+    let log = get_logger(TEST_LOG_LEVEL);
```
jimmygchen: can we replace this with the `test_logger`?

beacon_node/network/src/subnet_service/mod.rs
```diff
@@ -17,12 +32,645 @@ pub enum SubnetServiceMessage {
...
+    /* Public functions */
+
+    /// Establish the service based on the passed configuration.
+    pub fn new(
+        beacon_chain: Arc<BeaconChain<T>>,
+        node_id: NodeId,
+        config: &NetworkConfig,
+        log: &slog::Logger,
+    ) -> Self {
+        let log = log.new(o!("service" => "attestation_service"));
```
jimmygchen:
```suggestion
        let log = log.new(o!("service" => "subnet_service"));
```

beacon_node/network/src/subnet_service/mod.rs
```diff
@@ -17,12 +32,645 @@ pub enum SubnetServiceMessage {
...
+        if config.subscribe_all_subnets {
+            slog::info!(log, "Subscribing to all subnets");
+        }
+
+        // Build the list of known permanent subscriptions, so that we know not to subscribe or
+        // discover them.
+        let mut permanent_attestation_subscriptions = HashSet::default();
+        if config.subscribe_all_subnets {
+            // We are subscribed to all subnets, set all the bits to true.
+            for index in 0..<T::EthSpec as EthSpec>::SubnetBitfieldLength::to_u64() {
```
jimmygchen: We should probably be using `ATTESTATION_SUBNET_COUNT` from config instead even though this is unlikely to change.

beacon_node/network/src/subnet_service/mod.rs
```diff
@@ -17,12 +32,645 @@ pub enum SubnetServiceMessage {
...
+            beacon_chain.slot_clock.slot_duration().as_secs() * T::EthSpec::slots_per_epoch();
+        let default_sync_committee_duration = Duration::from_secs(
+            epoch_duration_secs.saturating_mul(spec.epochs_per_sync_committee_period.as_u64()),
+        );
+
+        let track_validators = !config.import_all_attestations;
+        let aggregate_validators_on_subnet =
+            track_validators.then(|| HashSetDelay::new(slot_duration));
+
+        let mut events = VecDeque::with_capacity(10);
```
jimmygchen, line 164: do we need to allocate more capacity for events now the two services are merged? or are there less events because long live subscriptions are now permanant?

beacon_node/network/src/subnet_service/mod.rs
```diff
@@ -17,12 +32,645 @@ pub enum SubnetServiceMessage {
...
+
+        // Queue discovery queries for the permanent attestation subnets
+        events.push_back(SubnetServiceMessage::DiscoverPeers(
```
jimmygchen: Do we query disocver when `self.discovery_disabled` is true?

beacon_node/network/src/subnet_service/mod.rs
```diff
@@ -17,12 +32,645 @@ pub enum SubnetServiceMessage {
...
+        subnet: Subnet,
+        duration_to_unsubscribe: Duration,
+        slot_required_until: Slot,
+    ) {
+        // Return if we have subscribed to all subnets
+        if self.subscribe_all_subnets {
+            return;
+        }
+
+        // Return if we already have a subscription for the subnet and its closer or
```
jimmygchen: I think this comment may be outdated? Maybe something like:

```rust
// Update the unsubscription duration if we already have a subscription for the subnet
```

beacon_node/network/src/subnet_service/mod.rs
```diff
@@ -17,12 +32,645 @@ pub enum SubnetServiceMessage {
...
+            .unwrap_or_default();
+
+        // First check this is worth doing.
+        if time_to_subscription_end.is_zero() {
+            return Err("Time when subscription would end has already passed.");
+        }
+
+        // We need to check and add a subscription for the right kind, regardless of the presence
+        // of the subnet as a subscription of the other kind. This is mainly since long lived
+        // subscriptions can be removed at any time when a validator goes offline.
```
jimmygchen: Is this still possible?

beacon_node/network/src/subnet_service/mod.rs
```diff
@@ -17,12 +32,645 @@ pub enum SubnetServiceMessage {
...
+                );
+                self.queue_event(SubnetServiceMessage::Subscribe(subnet));
+            }
+        }
+        Ok(())
+    }
+
+    // Unsubscribes from a subnet that was removed if it does not continue to exist as a
+    // subscription of the other kind. For long lived subscriptions, it also removes the
+    // advertisement from our ENR.
```
jimmygchen: I think the "long lived subscriptions" comment is no longer applicable

beacon_node/network/src/subnet_service/mod.rs
```diff
@@ -17,12 +32,645 @@ pub enum SubnetServiceMessage {
...
+pub struct ExactSubnet {
+    /// The `SubnetId` associated with this subnet.
+    pub subnet: Subnet,
+    /// For Attestations, this slot represents the start time at which we need to subscribe to the
+    /// slot. For SyncCommittee subnet id's this represents the end slot at which we no longer need
+    /// to subscribe to the subnet.
+    // NOTE: There was different logic between the two subscriptions and having a different
+    // interpretation of this variable seemed like the best way to group the logic, even though it
+    // may be counter-intuitive (apologies to future readers).
+    pub slot: Slot,
```
jimmygchen, line 64: I don't see where this field is used for `SyncCommittee`?

consensus/types/src/subnet_id.rs
AgeManning: Yeah. I've left it as a config value. The default in the config is 6, which is the value for 64 subnets. 

If someone changes the ATTESTATION_SUBNET_COUNT in the config, they SHOULD then update the attestation_subnet_prefix to match this. 

The reason I left it as a config, is that then its stored on the spec and we don't have to comptuer the int(ceillog2()) each time, i.e it becomes static. 

Also, its not the end of the world if someone wants to configure this out of spec.

beacon_node/network/src/subnet_service/tests/mod.rs
AgeManning: done

beacon_node/network/src/subnet_service/mod.rs
AgeManning: done

beacon_node/network/src/subnet_service/mod.rs
AgeManning: yep

beacon_node/network/src/subnet_service/mod.rs
AgeManning: Its kind of an arbitrary number. There's usually 3 or 4 at any given time (discoveries, subscriptions etc). But if the node is slow these could queue a bit. Figured 10 was reasonable. These messages are tiny in memory tho, so 10 is still very small.

beacon_node/network/src/subnet_service/mod.rs
AgeManning: Nope, good catch! Thanks

beacon_node/network/src/subnet_service/mod.rs
AgeManning: Yeah good catch. I did multiple passes of this, to try and make the logic a bit easier to understand and ended up removing this type from sync committee logic. 

Have updated the comment.

AgeManning:
Thanks for the review @jimmygchen - I think i've addressed all of these

beacon_node/network/src/service/tests.rs
```diff
@@ -1,4 +1,4 @@
-#[cfg(not(debug_assertions))]
+// #[cfg(not(debug_assertions))]
```
jimmygchen: Might want to uncomment this?

beacon_node/network/src/subnet_service/tests/mod.rs
```diff
@@ -213,71 +206,79 @@ mod attestation_service {
         // subscription config
         let committee_index = 1;
         // Keep a low subscription slot so that there are no additional subnet discovery events.
```
jimmygchen: Is this comment still relevant?

beacon_node/network/src/subnet_service/tests/mod.rs
```diff
@@ -213,71 +206,79 @@ mod attestation_service {
...
-        )
-        .unwrap();
         let expected = [
-            SubnetServiceMessage::Subscribe(Subnet::Attestation(subnet_id)),
-            SubnetServiceMessage::Unsubscribe(Subnet::Attestation(subnet_id)),
+            SubnetServiceMessage::Subscribe(subnet),
+            SubnetServiceMessage::Unsubscribe(subnet),
         ];
 
         // Wait for 1 slot duration to get the unsubscription event
```
jimmygchen, line 267: Looks like we're waiting for 1 epoch here?

beacon_node/network/src/subnet_service/tests/mod.rs
```diff
@@ -338,110 +339,83 @@ mod attestation_service {
...
-                attestation_service.subscription_count(),
-                subnets_per_node + 1
-            );
+        if subnet_service.is_subscribed(&Subnet::Attestation(subnet_id1)) {
+            // If we are permanently subscribed to this subnet, we won't see a subscribe message
+            let _ = get_events(&mut subnet_service, None, 1).await;
         } else {
-            assert!(attestation_service.subscription_count() == subnets_per_node);
+            let subscription = get_events(&mut subnet_service, None, 1).await;
+            assert_eq!(subscription, [expected]);
```
jimmygchen, line 352: This test is currently failing here
https://github.com/sigp/lighthouse/actions/runs/10933145564/job/30351144513?pr=6146

beacon_node/network/src/subnet_service/tests/mod.rs
```diff
@@ -338,110 +339,83 @@ mod attestation_service {
...
-                    subnet_id1
-                ))],
-                unsubscribe_event[..]
-            );
+        let expected = SubnetServiceMessage::Unsubscribe(Subnet::Attestation(subnet_id1));
+        if !subnet_service.is_subscribed(&Subnet::Attestation(subnet_id1)) {
+            assert_eq!([expected], unsubscribe_event[..]);
         }
 
         // Should be subscribed 2 long lived subnet after unsubscription.
```
jimmygchen: I think this is checking:

```suggestion
        // Should no longer be subscribed to any short lived subnet after unsubscription.
```

beacon_node/network/src/subnet_service/tests/mod.rs
jimmygchen: This is now passing in the 2nd run and on my local env 🤔 
Is there a potential timing issue here?

beacon_node/network/src/subnet_service/tests/mod.rs
```diff
@@ -296,8 +296,9 @@ mod attestation_service {
         let com2 = 0;
 
         // create the attestation service and subscriptions
```
jimmygchen: nit:

```suggestion
        // create the subnet service and subscriptions
```

beacon_node/network/src/subnet_service/tests/mod.rs
```diff
@@ -338,110 +339,83 @@ mod attestation_service {
...
-        ) {
-            // The index is 3*subnets_per_node (because we subscribe + discover + enr per long lived
-            // subnet) + 1
-            let index = 3 * subnets_per_node;
-            assert_eq!(expected, events[index]);
-            assert_eq!(
-                attestation_service.subscription_count(),
-                subnets_per_node + 1
-            );
+        if subnet_service.is_subscribed(&Subnet::Attestation(subnet_id1)) {
```
jimmygchen, line 347: Looks like this is checking `short lived subscriptions`?

jimmygchen: Not sure how it's possible to hit the `else` block given that the subscription should have happened. is this a timing thing?

consensus/types/src/subnet_id.rs
```diff
@@ -77,48 +76,16 @@ impl SubnetId {
     /// Computes the set of subnets the node should be subscribed to during the current epoch,
     /// along with the first epoch in which these subscriptions are no longer valid.
     #[allow(clippy::arithmetic_side_effects)]
-    pub fn compute_subnets_for_epoch<E: EthSpec>(
+    pub fn compute_attestation_subnets(
```
jimmygchen, line 83: The comment above needs to be updated as we no longer pass epoch to this function.

consensus/types/src/subnet_id.rs
```diff
@@ -77,48 +76,16 @@ impl SubnetId {
...
         raw_node_id: [u8; 32],
-        epoch: Epoch,
         spec: &ChainSpec,
-    ) -> Result<(impl Iterator<Item = SubnetId>, Epoch), &'static str> {
-        // simplify variable naming
-        let subscription_duration = spec.epochs_per_subnet_subscription;
+    ) -> impl Iterator<Item = SubnetId> {
```
jimmygchen, line 86: Just to confirm my understanding - this change doesn't impact discovery because `compute_attestation_subnets` is only used by the node to compute its own permanent subscriptions, and the node then advertises its persistent attestation subnet via ENR attestation bitfield.

Peers then use attestation bitfield advertised by the node for discovery - which means different nodes could have different logic for computing attestation subnets?

consensus/types/src/subnet_id.rs
```diff
@@ -77,48 +76,16 @@ impl SubnetId {
...
+        // The bits of the node-id we are using to define the subnets.
         let prefix_bits = spec.attestation_subnet_prefix_bits as u64;
-        let shuffling_prefix_bits = spec.attestation_subnet_shuffling_prefix_bits as u64;
-        let node_id = U256::from_be_slice(&raw_node_id);
 
+        let node_id = U256::from_be_slice(&raw_node_id);
         // calculate the prefixes used to compute the subnet and shuffling
         let node_id_prefix = (node_id >> (256 - prefix_bits)).as_le_slice().get_u64_le();
```
jimmygchen: Maybe worth creating a constant for `NODE_ID_BITS = 256` for readability

beacon_node/network/src/subnet_service/mod.rs
```diff
@@ -17,12 +32,640 @@ pub enum SubnetServiceMessage {
...
+    /// - Search for peers for required subnets.
+    /// - Request subscriptions for subnets on specific slots when required.
+    /// - Build the timeouts for each of these events.
+    ///
+    /// This returns a result simply for the ergonomics of using ?. The result can be
+    /// safely dropped.
+    pub fn validator_subscriptions(
+        &mut self,
+        subscriptions: impl Iterator<Item = Subscription>,
+    ) -> Result<(), String> {
```
jimmygchen: I don't see any Error returned from this function, maybe this doesn't need to return a `Result?`

beacon_node/network/src/service.rs
```diff
@@ -824,23 +813,17 @@ impl<T: BeaconChainTypes> NetworkService<T> {
 
     /// Handle a message sent to the network service.
     async fn on_validator_subscription_msg(&mut self, msg: ValidatorSubscriptionMessage) {
-        match msg {
+        if let Err(e) = match msg {
```
jimmygchen: See earlier comments - i dont see errors returned from the `validator_subscriptions`. 

If we do return an error, it might be worth also logging the subscription message type

consensus/types/src/subnet_id.rs
jimmygchen: I see, make sense!

beacon_node/network/src/subnet_service/mod.rs
```diff
@@ -17,12 +32,640 @@ pub enum SubnetServiceMessage {
     Unsubscribe(Subnet),
     /// Add the `SubnetId` to the ENR bitfield.
     EnrAdd(Subnet),
-    /// Remove the `SubnetId` from the ENR bitfield.
-    EnrRemove(Subnet),
```
jimmygchen, line 21: don't we need to remove sync subnet id from ENR when the subscription expires??

beacon_node/network/src/subnet_service/mod.rs
```diff
@@ -17,12 +32,640 @@ pub enum SubnetServiceMessage {
...
+    /// For Attestations, this slot represents the start time at which we need to subscribe to the
+    /// slot.
+    pub slot: Slot,
+}
+
+/// The enum used to group all kinds of validator subscriptions
+#[derive(Debug, Clone, PartialEq)]
+pub enum Subscription {
+    Attestation(ValidatorSubscription),
+    SyncCommittee(SyncCommitteeSubscription),
```
jimmygchen, line 71: Nice, I think this refactor will make PeerDAS validator subscription a lot easier to implement.

beacon_node/network/src/subnet_service/mod.rs
```diff
@@ -17,12 +32,640 @@ pub enum SubnetServiceMessage {
...
+        // discover them.
+        let mut permanent_attestation_subscriptions = HashSet::default();
+        if config.subscribe_all_subnets {
+            // We are subscribed to all subnets, set all the bits to true.
+            for index in 0..beacon_chain.spec.attestation_subnet_count {
+                permanent_attestation_subscriptions
+                    .insert(Subnet::Attestation(SubnetId::from(index)));
+            }
+        } else {
+            // Not subscribed to all subnets, so just calculate the required subnets from the
```
jimmygchen:
```suggestion
            // Not subscribed to all subnets, so just calculate the required subnets from the node id.
```

beacon_node/network/src/subnet_service/mod.rs
```diff
@@ -17,12 +32,640 @@ pub enum SubnetServiceMessage {
...
+            aggregate_validators_on_subnet,
+            waker: None,
+            discovery_disabled: config.disable_discovery,
+            subscribe_all_subnets: config.subscribe_all_subnets,
+            proposer_only: config.proposer_only,
+            log,
+        }
+    }
+
+    /// Return count of all currently subscribed subnets (long-lived **and** short-lived).
```
jimmygchen: short-lived subscriptions only right?

```suggestion
    /// Return count of all currently subscribed subnets (short-lived).
```

beacon_node/network/src/subnet_service/mod.rs
```diff
@@ -17,12 +32,640 @@ pub enum SubnetServiceMessage {
...
+                    }
+                }
+            }
+        }
+
+        // If the discovery mechanism isn't disabled, attempt to set up a peer discovery for the
+        // required subnets.
+        if !self.discovery_disabled {
+            if let Err(e) = self.discover_peers_request(subnets_to_discover.into_iter()) {
+                warn!(self.log, "Discovery lookup request error"; "error" => e);
```
jimmygchen, line 355: Is it worth logging `subnets_to_discover`?

beacon_node/network/src/subnet_service/mod.rs
```diff
@@ -17,12 +32,640 @@ pub enum SubnetServiceMessage {
...
+        if let Some(waker) = &self.waker {
+            waker.wake_by_ref()
+        }
+    }
+    /// Checks if there are currently queued discovery requests and the time required to make the
+    /// request.
+    ///
+    /// If there is sufficient time, queues a peer discovery request for all the required subnets.
+    /// `subnets_to_discover` takes a (subnet_id, Option<Slot>), where if the slot is not set, we
+    /// send a discovery request immediately.
```
jimmygchen: This function no longer takes an `Option<Slot>`

beacon_node/network/src/subnet_service/mod.rs
```diff
@@ -17,12 +32,640 @@ pub enum SubnetServiceMessage {
...
+        } else {
+            // This is a future slot, schedule subscribing.
+            self.scheduled_subscriptions
+                .insert_at(subnet, time_to_subscription_start);
+        }
+
+        Ok(())
+    }
+
+    /// Adds a subscription event and an associated unsubscription event if required.
```
jimmygchen: This fn no longer adds unsubscription event:

```suggestion
    /// Adds a sync subscription event with an expiry specified by `duration_to_unsubscribe`.
    /// Once the duration elapses, the subscription will be unsubscribed automatically.
```
```

File: beacon_node/network/src/subnet_service/mod.rs [hunk 31]
```diff
@@ -17,12 +32,640 @@ pub enum SubnetServiceMessage {
...
+    fn subscribe_to_sync_subnet(
+        &mut self,
+        subnet: Subnet,
+        duration_to_unsubscribe: Duration,
+        slot_required_until: Slot,
```
jimmygchen, line 498: This doesn't seem to be used other than logging, and conflicts with `duration_to_unsubscribe` I think?

beacon_node/network/src/subnet_service/mod.rs
```diff
@@ -17,12 +32,640 @@ pub enum SubnetServiceMessage {
...
+            if duration_to_unsubscribe > current_duration_to_unsubscribe {
+                self.subscriptions
+                    .update_timeout(&subnet, duration_to_unsubscribe);
+            }
+        } else {
+            // We have not subscribed before, so subscribe
+            self.subscriptions
+                .insert_at(subnet, duration_to_unsubscribe);
+            // We are not currently subscribed and have no waiting subscription, create one
+            debug!(self.log, "Subscribing to subnet"; "subnet" => ?subnet, "until" => ?slot_required_until);
```
jimmygchen, line 523: `until` no longer relevant, we could log `duration_to_unsubscribe`?

jimmygchen (changes requested):
I've reviewed the PR and it looks much cleaner! I've added a few comments / questions.
I think this will simplify implementing PeerDAS validator custody (subscriptions) too.

beacon_node/network/src/subnet_service/tests/mod.rs
AgeManning: I dont think so, removed

beacon_node/network/src/subnet_service/tests/mod.rs
AgeManning: The third parameter here is the timeout. It gets the next two events, but if we dont get 2 events in an epoch it times out. 

I think we are waiting for the next events here, which should be the next slot

beacon_node/network/src/subnet_service/tests/mod.rs
AgeManning: Yeah, there is a timeout on the `get_events`. Ideally I wanted to move this to a deterministic thing rather than times, but I couldn't find a good way to do it. 

This is why we're using timeouts in the get_events() and making them fairly large. I think this test could be improved to not rely on a single slot tho. 

I haven't done it in this PR. If the test keeps failing, we'll look into it I rekon

beacon_node/network/src/subnet_service/tests/mod.rs
AgeManning: Agreed

beacon_node/network/src/subnet_service/tests/mod.rs
AgeManning: I think here we're worried about the long-lived subscriptions. 

If the subnet chosen, happens to be the same one we are permanently subscribed to, then we don't create a subscription or unsubscription and is_subscribed() i think only returns short-lived subscriptions

consensus/types/src/subnet_id.rs
AgeManning: Yep, done

consensus/types/src/subnet_id.rs
AgeManning: Exactly.

consensus/types/src/subnet_id.rs
AgeManning: Done

beacon_node/network/src/subnet_service/mod.rs
AgeManning: huh, probably not. Nice

beacon_node/network/src/subnet_service/mod.rs
AgeManning: Yeah. i recall this was annoying. I think the formatting of the duration was annoying and it was different between sync committees and normal slots. Adding this annoying param was the nicest workaround to print pretty logs to follow. 

Happy to try and re-engineer, but I recall it being the simplest of all the options, i don't remember now tho

beacon_node/network/src/subnet_service/mod.rs
AgeManning: Didn't follow why its not relevant here? I think it is.

beacon_node/network/src/subnet_service/mod.rs
AgeManning: Yeah we do. I've still got to add this!

beacon_node/network/src/subnet_service/mod.rs
jimmygchen: ah I didn't read the code that was calling this - `duration_to_unsubscribe` was derived from `slot_required_until`.

I initially thought we're not using `slot_required_until` here because it's only used for logging in this function, and we use `duration_to_unsubscribe` when inserting into the `HashSetDelay`.

All good!

beacon_node/network/src/subnet_service/tests/mod.rs
jimmygchen: Sounds good

jimmygchen (approved):
Nice, looks good to me now! 👍

michaelsproul:
@mergify queue

consensus/types/src/chain_spec.rs
```diff
@@ -1701,9 +1669,6 @@ impl Config {
             maximum_gossip_clock_disparity_millis: spec.maximum_gossip_clock_disparity_millis,
             message_domain_invalid_snappy: spec.message_domain_invalid_snappy,
             message_domain_valid_snappy: spec.message_domain_valid_snappy,
-            attestation_subnet_extra_bits: spec.attestation_subnet_extra_bits,
```
michaelsproul, line 1704: The removal of these fields may break some third-party HTTP API consumers. I will flag this PR as backwards-incompat so that this is mentioned in the release notes.

PR #6147, opened 2024-07-22
Add `DataColumnSidecar` gossip topic and message handling

jimmygchen:
## Issue Addressed

Part of #6072. 

This PR contains:
- Add gossip data column topic and message handling (excludes gossip validation)

Upstream PRs:
- #5050


beacon_node/lighthouse_network/src/types/pubsub.rs
```diff
@@ -270,6 +275,41 @@ impl<E: EthSpec> PubsubMessage<E> {
...
                     }
+                    GossipKind::DataColumnSidecar(subnet_id) => {
+                        match fork_context.from_context_bytes(gossip_topic.fork_digest) {
+                            Some(ForkName::Deneb | ForkName::Electra) => {
+                                let col_sidecar = Arc::new(
+                                    DataColumnSidecar::from_ssz_bytes(data)
+                                        .map_err(|e| format!("{:?}", e))?,
+                                );
+                                let peer_das_enabled =
+                                    fork_context.spec.is_peer_das_enabled_for_epoch(
```
pawanjay176, line 287: This could be dangerous as the slot value can be set to anything which could lead to a column sidecar bubbling up the stack and hitting one of our `unimplemented!`.
This wouldn't happen because we wouldn't subscribe to the required topic but better to be safe at every point in the networking stack specially given that we want this code in stable.

pawanjay176: Also, curious why we have allow for both electra + deneb. Is it just to provide flexibility for devnets on top of deneb/electra or something else?

beacon_node/network/src/sync/manager.rs
```diff
@@ -646,6 +649,10 @@ impl<T: BeaconChainTypes> SyncManager<T> {
                     }),
                 );
             }
+            SyncMessage::UnknownParentDataColumn(_peer_id, _data_column) => {
+                // TODO(das): data column parent lookup to be implemented
+                unimplemented!("data column parent lookup to be implemented")
```
pawanjay176: can we not have a panic macro here and just the comment? Just trying to be extra careful as this will go to stable eventually.
A change in gossip_methods in a future PR without the corresponding change here might go unnoticed

consensus/types/src/data_column_subnet_id.rs
```diff
@@ -0,0 +1,198 @@
...
+pub struct DataColumnSubnetId(#[serde(with = "serde_utils::quoted_u64")] u64);
+
+impl DataColumnSubnetId {
+    pub fn new(id: u64) -> Self {
+        id.into()
+    }
+
+    pub fn from_column_index<E: EthSpec>(column_index: usize, spec: &ChainSpec) -> Self {
+        (column_index
+            .safe_rem(spec.data_column_sidecar_subnet_count as usize)
```
pawanjay176, line 23: Curious why we don't enforce that `column_index < data_column_sidecar_subnet_count` instead of taking the remainder

consensus/types/src/data_column_subnet_id.rs
```diff
@@ -0,0 +1,198 @@
...
+        // value, but here we assume it is valid.
+
+        let mut subnets: HashSet<u64> = HashSet::new();
+        let mut current_id = node_id;
+        while (subnets.len() as u64) < custody_subnet_count {
+            let mut node_id_bytes = [0u8; 32];
+            current_id.to_little_endian(&mut node_id_bytes);
+            let hash = ethereum_hashing::hash_fixed(&node_id_bytes);
+            let hash_prefix = [
+                hash[0], hash[1], hash[2], hash[3], hash[4], hash[5], hash[6], hash[7],
```
pawanjay176: Maybe do something like 
```rust
let hash_prefix: [u8; 8] = hash[0..8].try_into().expect("hash_fixed produces a 32 byte array");
```

consensus/types/src/data_column_subnet_id.rs
```diff
@@ -0,0 +1,198 @@
...
+    use crate::MainnetEthSpec;
+
+    type E = MainnetEthSpec;
+
+    #[test]
+    fn test_compute_subnets_for_data_column() {
+        let spec = E::default_spec();
+        let node_ids = [
+            "0",
+            "88752428858350697756262172400162263450541348766581994718383409852729519486397",
```
pawanjay176, line 145: might be good to have a link to the test vectors here

pawanjay176 (changes requested):
Looks mostly good. The das logic seems sane, but I have mainly focused on ensuring that we don't trigger any of the das related code if this is run on mainnet.

Just to be extra safe, I would prefer to remove all instances of panic macros and just let the `TODO(das)` comments guide us to incomplete code.

Also confirmed that we are not subscribing to any data column subnet topics in this PR so any of the upstream code wouldn't get triggered

consensus/types/src/data_column_subnet_id.rs
jimmygchen: Not sure if I understand. This function calculates the `DataColumnSubnetId` from a `ColumnIndex`. `column_index` isn't necessary smaller than `data_column_sidecar_subnet_count`. The current spec value for `data_column_sidecar_subnet_count` is `32`, and we've been doing interop testing  with `64`. Latest proposal is to bump this value further up to `128`. Essentially we need to be able to support different combinations as both `NUMBER_OF_COLUMNS` and `DATA_COLUMN_SUBNET_COUNT` are configurable values.

consensus/types/src/data_column_subnet_id.rs
jimmygchen: These tests were written before the `consensus-spec` tests were added and the `node_ids` here were taken from the `subnet_id.rs` tests:
https://github.com/sigp/lighthouse/blob/c52c598f6922059900d60723bb541add6865a87e/consensus/types/src/subnet_id.rs#L181

I can remove these tests now they're covered by the spec tests? (will require updating to `v1.5.0-alpha.3` though)

jimmygchen: A bit of a trade off here - it's kinda nice to have some basic unit tests next to the actual code, these are our own unit tests, whereas the spec tests are to make sure we interop with others - though having overlapping tests adds maintenance effort.

beacon_node/network/src/sync/manager.rs
jimmygchen: Yeah agree! I'll remove it. 

I originally added it cause I thought that was the preference when we were working on Deneb, but I think that was also before merging to `unstable`.

beacon_node/lighthouse_network/src/types/pubsub.rs
jimmygchen: I've removed the `unimplemented!` calls, if there anything that you think we should change here?

I actually added this extra check (not currently on `das` branch) to make sure we don't break Deneb and Electra, but you're right that the slot can be set to anything. Is it possible though given that `EIP7594_FORK_EPOCH` is configured to `u64::max` on mainnet? As you said we've got some protection as we'll only subscribe to data column subnets on `EIP7594_FORK_EPOCH`.

jimmygchen: > Also, curious why we have allow for both electra + deneb. Is it just to provide flexibility for devnets on top of deneb/electra or something else?

Yes, we're currently testing on top of Deneb but will switch over to Electra once it's stable enough for us to rebase. I'll add a TODO comment here to make sure we don't forget changing this.

consensus/types/src/data_column_subnet_id.rs
jimmygchen: Updated in https://github.com/sigp/lighthouse/pull/6147/commits/29aab7a11afd862b09291947698c017e088bc872

beacon_node/network/src/sync/manager.rs
jimmygchen: Removed in https://github.com/sigp/lighthouse/pull/6147/commits/29aab7a11afd862b09291947698c017e088bc872

consensus/types/src/data_column_subnet_id.rs
pawanjay176: Ohh nice, didn't realise these are from our own tests.
That's neat.

beacon_node/lighthouse_network/src/types/pubsub.rs
pawanjay176: Yeah went through the gossipsub code and can confirm that a gossipsub message will not reach the application (lighthouse) unless it is subscribed to it. So we are good here.

consensus/types/src/data_column_subnet_id.rs
pawanjay176: Ohh I miunderstood this one, looks good then

pawanjay176 (approved):

beacon_node/lighthouse_network/src/types/pubsub.rs
```diff
@@ -270,6 +275,42 @@ impl<E: EthSpec> PubsubMessage<E> {
                             )),
                         }
                     }
+                    GossipKind::DataColumnSidecar(subnet_id) => {
+                        match fork_context.from_context_bytes(gossip_topic.fork_digest) {
```
dapplion, line 279: Use deneb_activated()

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -599,6 +601,67 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+    ) {
+        let slot = column_sidecar.slot();
+        let block_root = column_sidecar.block_root();
+        let index = column_sidecar.index;
+        let delay = get_slot_delay_ms(seen_duration, slot, &self.chain.slot_clock);
+        // Log metrics to track delay from other nodes on the network.
+        metrics::set_gauge(
+            &metrics::BEACON_DATA_COLUMN_DELAY_GOSSIP,
+            delay.as_millis() as i64,
+        );
```
dapplion, line 621: This should be an observation of a histogram not a gauge, same for `BEACON_DATA_COLUMN_DELAY_GOSSIP_VERIFICATION`

beacon_node/network/src/sync/manager.rs
```diff
@@ -107,6 +107,9 @@ pub enum SyncMessage<E: EthSpec> {
     /// A blob with an unknown parent has been received.
     UnknownParentBlob(PeerId, Arc<BlobSidecar<E>>),
 
+    /// A data column with an unknown parent has been received.
+    UnknownParentDataColumn(PeerId, Arc<DataColumnSidecar<E>>),
```
dapplion, line 111: It's actually useless to have to forward the payload in this error :)

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
jimmygchen: This was changed for block and blobs in #5510 and I made the same change here for consistency.

beacon_node/network/src/sync/manager.rs
jimmygchen: I think you're referring to `GossipDataColumnError::ParentUnknown` instead of `SyncMessage` here? we do need to keep the data column in the `BlockComponent` when doing a parent lookup.

beacon_node/lighthouse_network/src/types/pubsub.rs
jimmygchen: Updated to use `deneb_enabled` in https://github.com/sigp/lighthouse/pull/6147/commits/f4cd1fa47c750a73b32760109e7a0ec91a1e0704

beacon_node/network/src/sync/manager.rs
jimmygchen: I've removed the unused error types in `GossipDataColumnError` and leaving this line unchanged. Let me know if you still want the payload removed for this error?

beacon_node/lighthouse_network/src/discovery/subnet_predicate.rs
```diff
@@ -1,15 +1,22 @@
...
+pub fn subnet_predicate<E>(
+    subnets: Vec<Subnet>,
+    log: &slog::Logger,
+    spec: &ChainSpec,
+) -> impl Fn(&Enr) -> bool + Send
 where
     E: EthSpec,
 {
     let log_clone = log.clone();
+    let spec_clone = spec.clone();
```
michaelsproul: We can do this later, but we should probably avoid cloning the `ChainSpec` so frequently by `Arc`ing it

michaelsproul (approved):
Approved modulo the `ChainSpec` cloning in `subnet_predicate` which could represent a performance regression on mainnet

beacon_node/lighthouse_network/src/discovery/subnet_predicate.rs
jimmygchen: To do it properly we should probably `Arc` it everywhere, but we could probably do it in a separate PR as it will be  quite a big change. I've `Arc`'d the spec only in `Discovery` and `NetworkService` to avoid the performance regression for now.

michaelsproul (approved):

michaelsproul:
@mergify queue

PR #6148, opened 2024-07-22
level down libp2p quic accept and close errors

jxs:
## Issue Addressed

After https://github.com/quinn-rs/quinn/pull/1804/ `quinn` now returns this accept error which is also bubbled by `libp2p` as a `ListenerError`, this PR silences `libp2p-quic` accept errors in general as they are mostly connection refusals. 
`libp2p::quic::Error::Connection` also covers connection close errors which are also redundant on our case.

### DIFF
--- a/beacon_node/lighthouse_network/src/service/mod.rs
+++ b/beacon_node/lighthouse_network/src/service/mod.rs
@@ -1840,13 +1840,17 @@ impl<E: EthSpec> Network<E> {
                     }
                 }
                 SwarmEvent::ListenerError { error, .. } => {
-                    // this is non fatal, but we still check
-                    warn!(self.log, "Listener error"; "error" => ?error);
-                    if Swarm::listeners(&self.swarm).count() == 0 {
-                        Some(NetworkEvent::ZeroListeners)
+                    // Ignore quic accept and close errors.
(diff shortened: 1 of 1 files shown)


beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -1840,13 +1840,16 @@ impl<E: EthSpec> Network<E> {
...
+                    // Ignore quic accept and close errors.
+                    if error
+                        .get_ref()
+                        .and_then(|err| err.downcast_ref::<libp2p::quic::Error>())
+                        .filter(|err| matches!(err, libp2p::quic::Error::Connection(_)))
+                        .is_none()
+                    {
+                        warn!(self.log, "Listener error"; "error" => ?error);
                     }
+                    None
```
jxs, line 1853: previously we were checking the number of listeners, but it's redundant here as they will only decrease if gracefully closed, `ListenerError`s are non fatal and therefore shouldn't close the listener

ackintosh (approved):
Looks good to me!

AgeManning (approved):

AgeManning:
@mergifyio queue

PR #6153, opened 2024-07-23
Show correct total validator count during validator recovery

dknopik:
## Issue Addressed

A small one I found just now: The progress indicator displayed during `lighthouse account validator recover` would show something like `0/4294967276 Index: 40` when `count < first_index`.

## Proposed Changes

Simply display the count there instead of subtracting the first index.

## Additional Info

None.

### DIFF
--- a/account_manager/src/validator/recover.rs
+++ b/account_manager/src/validator/recover.rs
@@ -147,7 +147,7 @@ pub fn cli_run(matches: &ArgMatches, validator_dir: PathBuf) -> Result<(), Strin
 
         println!(
             "{}/{}\tIndex: {}\t0x{}",
-            index - first_index,
+            index - first_index + 1,
             count,
             index,
             voting_pubkey
--- a/account_manager/src/validator/recover.rs
+++ b/account_manager/src/validator/recover.rs
@@ -148,7 +148,7 @@ pub fn cli_run(matches: &ArgMatches, validator_dir: PathBuf) -> Result<(), Strin
         println!(
             "{}/{}\tIndex: {}\t0x{}",
             index - first_index,
-            count - first_index,
+            count,
             index,
             voting_pubkey
         );


account_manager/src/validator/recover.rs
```diff
@@ -148,7 +148,7 @@ pub fn cli_run(matches: &ArgMatches, validator_dir: PathBuf) -> Result<(), Strin
         println!(
             "{}/{}\tIndex: {}\t0x{}",
             index - first_index,
-            count - first_index,
+            count,
```
chong-he, line 151: This is option 1:
If count = 2, the UX would be:
```
0/1
1/1
```

```suggestion
            count - 1,
```

chong-he (commented):
Thanks for the contribution, nice catch! 

Just a nitpick, thought it would be better to have the count as `count - 1`, since the count displayed starts from 0. So, when we do:

`lighthouse account validator recover --first-index 4 --count 2`, it will show:
```
0/1
1/1 
```
instead of:

```
0/2 
1/2
```

Edit: other than that, tested and confirmed the UX shown as intended and it's good

dknopik:
Hey @chong-he, good point. But then actually, I would argue for
```
1/2 
2/2
```
which might be more intuitive to users, as people usually index starting at 1 in non-code contexts. But that is a matter of taste I guess.

chong-he:
Good point, that would be:
```
 index - first_index + 1,
 count,
```

I have commented above as well. 

I think this is good to go

account_manager/src/validator/recover.rs
chong-he: For the devs, this is option 1: 

If `count = 2`, the UX would be:
```
0/1
1/1 
```

account_manager/src/validator/recover.rs
```diff
@@ -148,7 +148,7 @@ pub fn cli_run(matches: &ArgMatches, validator_dir: PathBuf) -> Result<(), Strin
         println!(
             "{}/{}\tIndex: {}\t0x{}",
             index - first_index,
```
chong-he, line 150: This is option 2:
If count = 2, the UX would be:
```
1/2
2/2 
```
This would be more intuitive from the user perspective. I would opt for this one too

chong-he: This is option 2:
If count = 2, the UX would be:
```
1/2
2/2 
```
This would be more intuitive from the user perspective. I would opt for this one too
```suggestion
            index - first_index + 1,
```

chong-he (commented):
Adding the option 2 which looks better

dapplion (approved):

jimmygchen:
@mergify queue

PR #5088, opened 2024-01-19
Add `purge-db-safe` flag which requires manual confirmation

macladson:
## Issue Addressed

https://github.com/sigp/lighthouse/issues/5028

## Proposed Changes

Add a new flag called `--purge-db-safe` which functions like `--purge-db` but requires manual confirmation. As such it is not suitable for environments that are designed to be automated.

## Additional Info

I moved the `stdin-inputs` flag to be a global flag under the `lighthouse` command, as I felt this is a cleaner solution than currently, even if not every subcommand will use it. Happy to change it back if someone has a different opinion.

Issue #5028, opened 2023-12-26
[Feature Request] Add confirmation prompt for `--purge-db`


SeaMonkey82:
I've accidentally restarted lighthouse with `--purge-db` in my command line  on my testing machine enough times that I need a safety net to prevent me from being dumb.

macladson:
I can work on this. I think it will need to be a separate flag so we don't break compatibility with automated scripts.
Something like `--purge-db-safe`.

macladson:
Implemented in #6154

PR #6154, opened 2024-07-23
Require manual confirmation to purge database with `purge-db`

macladson:
## Issue Addressed

https://github.com/sigp/lighthouse/issues/5028

## Proposed Changes

The `purge-db` flag now requires manual confirmation in order to purge the database.
A new flag called `purge-db-force` has been added for those who still wish to purge with no intervention.

## Additional Info

This is a pretty spicy breaking change as I suspect `--purge-db` is widely used.

See the [alternative (although outdated) PR here](https://github.com/sigp/lighthouse/pull/5088) which adds a `purge-db-safe` flag, which means the safety would effectively become opt-in. In my opinion this is unnatural, although more convenient as it doesn't require a breaking change.

### DIFF
--- a/beacon_node/src/config.rs
+++ b/beacon_node/src/config.rs
@@ -90,8 +90,7 @@ pub fn get_config<E: EthSpec>(
                     log,
                     "The `--purge-db` flag was passed, but Lighthouse is not running \
                     interactively. The database was not purged. Use `--purge-db-force` \
-                    or `lighthouse db purge` to purge the database without requiring \
-                    confirmation."
+                    to purge the database without requiring confirmation."
                 );
             }
         }
--- a/beacon_node/src/config.rs
+++ b/beacon_node/src/config.rs
@@ -22,6 +22,7 @@ use slog::{info, warn, Logger};
 use std::cmp::max;
 use std::fmt::Debug;
 use std::fs;
+use std::io::IsTerminal;
 use std::net::Ipv6Addr;
 use std::net::{IpAddr, Ipv4Addr, ToSocketAddrs};
 use std::num::NonZeroU16;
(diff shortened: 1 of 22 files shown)


beacon_node/src/config.rs
```diff
@@ -50,26 +54,44 @@ pub fn get_config<E: EthSpec>(
...
-            fs::remove_dir_all(blobs_db)
-                .map_err(|err| format!("Failed to remove blobs_db: {}", err))?;
+    if client_config.data_dir().exists() {
+        if cli_args.get_flag("purge-db-force") {
+            let chain_db = client_config.get_db_path();
+            let freezer_db = client_config.get_freezer_db_path();
+            let blobs_db = client_config.get_blobs_db_path();
+            purge_db(chain_db, freezer_db, blobs_db)?;
+        } else if cli_args.get_flag("purge-db") {
+            if std::io::stdin().is_terminal() {
```
macladson: `--purge-db` will now check to see if it running interactively before purging. If it isn't it will not prompt for confirmation, but instead log a warning and ignore the purge. This should prevent users running under systemd from accidentally stalling Lighthouse

beacon_node/src/config.rs
```diff
@@ -50,26 +54,44 @@ pub fn get_config<E: EthSpec>(
...
+                } else {
+                    eprintln!("Database was not deleted. Lighthouse will now close.");
+                    std::process::exit(1);
+                }
+            } else {
+                warn!(
+                    log,
+                    "The `--purge-db` flag was passed, but Lighthouse is not running \
+                    interactively. The database was not purged. Use `--purge-db-force` \
+                    or `lighthouse db purge` to purge the database without requiring \
```
michaelsproul: we don't actually have `lighthouse db purge` implemented (yet)

I wouldn't be opposed to adding it to this PR if you have time

beacon_node/src/config.rs
michaelsproul: I think this check should maybe be `is_terminal() || stdin_inputs`. If we are reading from `stdin` then it doesn't matter if we are running interactively. E.g. someone could run `lighthouse bn --purge-db < confirm` in a script or daemon

beacon_node/src/config.rs
```diff
@@ -50,26 +54,44 @@ pub fn get_config<E: EthSpec>(
...
+            let blobs_db = client_config.get_blobs_db_path();
+            purge_db(chain_db, freezer_db, blobs_db)?;
+        } else if cli_args.get_flag("purge-db") {
+            if std::io::stdin().is_terminal() {
+                let stdin_inputs = cfg!(windows) || cli_args.get_flag(STDIN_INPUTS_FLAG);
+                eprintln!(
+                    "You are about to delete the chain database. This is irreversable \
+                    and you will need to resync the chain."
+                );
+                eprintln!(
```
michaelsproul: I wonder if we should log these messages at `info!` level using slog? It might be nice for consistency

macladson: Fixed in [206252c](https://github.com/sigp/lighthouse/pull/6154/commits/206252cd055560a132bfd648be3503c0f6a157cf)

beacon_node/src/config.rs
macladson: Added in [c2f1146](https://github.com/sigp/lighthouse/pull/6154/commits/c2f1146d5b17292a3da88c21b303a990e16398c6)

This now allows you to do `lighthouse bn --purge-db --stdin-inputs <<< confirm`

Note that this will mean that if Lighthouse is running under Windows, it will never ignore the flag, and will always block until it gets a response from stdin

beacon_node/src/config.rs
michaelsproul: @macladson can merge this soon if you amend this text

michaelsproul (approved):

michaelsproul:
@mergify queue

wjmelements:
I don't use interactive mode. For me this was a stupid and annoying breaking change. I am curious if anyone was grateful about it. #5088 would have been better because it would not have been a breaking change. Those wanting confirmation in interactive mode would still have it.

wjmelements:
The most annoying aspect is that it fails to purge the db quietly. My node was broken for an excessive amount of time because the db was not purged as expected.

wjmelements:
It would have been less annoying even if it had caused an early exit, because then I would have noticed weeks earlier.

michaelsproul:
@wjmelements We have a lot of users who used to `purge-db` excessively due to either forgetting to remove the flag from their config, or not understanding that they should remove it. This change was designed to help them, and I think so far it has been effective.

> It would have been less annoying even if it had caused an early exit, because then I would have noticed weeks earlier.

We try to avoid breaking changes like this that temporarily brick the node, because they represent a headache for node admins. Having a node suddenly fail to start is more annoying for many people than having it start with a warning and a different behaviour to what was expected. Especially in this case, where the different behaviour is somewhat benign (keeping the DB).

> The most annoying aspect is that it fails to purge the db quietly. My node was broken for an excessive amount of time because the db was not purged as expected.

I don't understand how `purge-db` was both an essential part of your node operation, and also how your node was out of sync for _weeks_ without you noticing. I suspect you're using Lighthouse in a somewhat different way to the majority of users, where you don't run it all the time, and just checkpoint sync it whenever you _do_ want it. But then I still don't understand how you didn't notice that it wasn't synced? You mustn't have actually been relying on it for anything?

Anyway, the ship has sailed on this. The breaking change was clearly described in the release notes. We're not going to implement _another_ breaking change to put it back to how it was.

wjmelements:
> We have a lot of users who used to purge-db excessively due to either forgetting to remove the flag from their config, or not understanding that they should remove it

Unless these users are disproportionately using interactive mode, it's going to be the same problem but with `purge-db-force`.

> We're not going to implement another breaking change

Until you decide to implement `purge-db-force-force`

> I don't understand how purge-db was both an essential part of your node operation, and also how your node was out of sync for weeks without you noticing. I suspect you're using Lighthouse in a somewhat different way to the majority of users, where you don't run it all the time, and just checkpoint sync it whenever you do want it. But then I still don't understand how you didn't notice that it wasn't synced? You mustn't have actually been relying on it for anything?

I run between 5 and 20 nodes. I don't check the broken ones constantly. There are scripts to unbreak them. You broke those scripts. The logs indicating why the purge didn't work got cycled. Pardon my frustration.

PR #6156, opened 2024-07-23
Remove duplicated connection limits checks

jxs:
## Issue Addressed

Following https://github.com/sigp/lighthouse/pull/6140#pullrequestreview-2195206263 found duplicated code and took the time to improve the wording with regards to dial logic.

Reviewing by commit is suggested as I left detailed notes on the commit messages.

@AgeManning if you feel it contains too many code changes we should probably wait for after `5.3` release.


beacon_node/lighthouse_network/src/peer_manager/network_behaviour.rs
```diff
@@ -251,19 +244,6 @@ impl<E: EthSpec> PeerManager<E> {
 
         // Count dialing peers in the limit if the peer dialed us.
         let count_dialing = endpoint.is_listener();
-        // Check the connection limits
-        if self.peer_limit_reached(count_dialing)
```
AgeManning, line 253: Removing this changes the functioning logic and so we should be careful about what are the consequences. 

Probably we should document how this works a bit more, so I'll try and go through the differences. 

There are two connection limits. 
- The ConnectionLimit behaviour
- The peer manager

The connection limit behaviour sets the maximum allowed for all connections but cannot take into account things like whether we need a peer for a subnet or not. IIRC it's limits are set higher than the peer manager, so the peer manager can be the discriminating entity as to whether we accept a peer or not. The ConnectionLImit behaviour creates caps on the allowed connections so can close connections early. We probably want to group these into one to avoid this confusion (i.e remove the connection limit behaviour). 

Anyway, lets walk through with real numbers the difference we are talking about here. 

The connection limit behaviour has the following max_established limit:
```rust
                    (config.target_peers as f32 * (1.0 + PEER_EXCESS_FACTOR + PRIORITY_PEER_EXCESS))
                        .ceil() as u32,

```
which for target_peers = 100 should be: 100(1 + 0.1 +0.2) = 130

The peer manager has two values for incoming or outgoing. For incoming its:
```rust
self.network_globals.connected_or_dialing_peers() >= self.max_peers()
```
Where max_peers is target_peers ( 1 + 0.1) = 110
and for outgoing its:
```rust
            self.network_globals.connected_peers() >= self.max_outbound_dialing_peers()
```
which is target_peers ( 1 +  0.1 + 0.2/2) = 120

So we can see the peer manager is stricter than the connection limit behaviour. The reason being is that we reserve a small portion (in this case 10-20) peer slots for peers that connect to us that are useful for subnets (this is the has_future_duty() function) or that are outbound peers. 

By removing this, we are taking away this functionality. We probably want to explicitly discuss why we are doing this and if its necessary as the change here will have an impact on the functionality of lighthouse.

beacon_node/lighthouse_network/src/peer_manager/mod.rs
```diff
@@ -340,15 +339,15 @@ impl<E: EthSpec> PeerManager<E> {
...
                     self.network_globals
                         .peers
                         .write()
-                        .update_min_ttl(&enr.peer_id(), min_ttl);
+                        .update_min_ttl(&peer_id, min_ttl);
                 }
-                let peer_id = enr.peer_id();
                 if self.dial_peer(enr) {
-                    debug!(self.log, "Dialing discovered peer"; "peer_id" => %peer_id);
+                    debug!(self.log, "Added discovered ENR peer to dial queue"; "peer_id" => %peer_id);
```
AgeManning, line 349: We are now emitting a log for every dial, we can prob remove this one now

jxs: yeah but this is different, these are the new discovered, the others are the cached ones, do you still want to remove this?

beacon_node/lighthouse_network/src/peer_manager/network_behaviour.rs
jxs: thanks for explaining age, wdyt of then moving this then to `handle_established_{inbound,outbound}_connection` methods (to in the future be easier merging `connection_limits` behaviour with `peer_manager` )? I did it in the last commit ptal

beacon_node/lighthouse_network/src/peer_manager/mod.rs
AgeManning: Yeah, you are right

AgeManning (approved):
I think this looks good. We should test it a bit to make sure it does what we expect.

AgeManning:
bumping this

jxs:
@mergifyio queue

PR #6162, opened 2024-07-24
Prevent continuous searches for low-peer networks

AgeManning:
We have logic to try and maintain a healthy amount of outbound connections. 

For testnets and low-networks it could be the case that we have connected to every node on the network and there are no more nodes to establish outbound connections to. In these cases, we do repetitive loops of discovery requests. 

I didn't want to add yet another flag to the CLI, so I added heuristic, that removes the logic to try and find more outbound peers if we set the target peers to <= 10. 

It should at the very least help with the kurtosis local testnet scripts.


jimmygchen (approved):
Looks good, this would be helpful for small testnets, and we can eventually merge this into `unstable` as well as it shouldn't be invasive to mainnet.

Issue #6161, opened 2024-07-24
CI jobs reaching Docker Hub image request limit


jimmygchen:
## Description

This job failed due to rate limit from DockerHub:
https://github.com/sigp/lighthouse/actions/runs/10070215403/job/27838668899?pr=6159

We could authenticate to increase the pull rate limit and/or consider solutions to reduce the amount of image requests 
https://www.docker.com/increase-rate-limits/

PR #6163, opened 2024-07-24
Revert --image-download always flag to avoid CI getting rate limited

jimmygchen:
## Issue Addressed

Partially #6161. 

This PR reverts `--image-download=always` flag to avoid CI getting rate limited by Docker Hub. It looks like the Kurtosis tests are now pulling all images every job, whereas it wasn't doing this before the flag was added.


antondlr (approved):
lgtm

jimmygchen:
@mergify queue

PR #6164, opened 2024-07-24
Remove default target peer for proposer-only mode for PeerDAS.

jimmygchen:
## Issue Addressed

The default target peer for proposer node (=15) will need to be increased for peerdas. A node would need peers across all data column subnets to be able to publish the data columns successfully. 

If we can have ["decentralized blob building"](https://hackmd.io/@dapplion/blob_fetch) working _reliably_ on PeerDAS in the future, then proposers won't necessary need to publish blobs themselves and this number can potentially be reduced again.

### DIFF
--- a/beacon_node/src/config.rs
+++ b/beacon_node/src/config.rs
@@ -1435,15 +1435,12 @@ pub fn set_network_config(
     }
 
     // Proposer-only mode overrides a number of previous configuration parameters.
-    // Specifically, we avoid subscribing to long-lived subnets and wish to maintain a minimal set
-    // of peers.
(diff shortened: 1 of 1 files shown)


michaelsproul:

jimmygchen:
This feature makes the `proposer-only` mode less useful than it was. I'm inclined to see whether decentralized blob building #6268 formally make it into the spec before deciding to merge this one. 

IMO I think we should ship peerdas with #6268, and we'd be able to keep the `proposer-only` feature.

jimmygchen:
This PR was closed because the source branch was delete. With `getBlobsV1` being included in the spec, I think we can close this anyway.

PR #6168, opened 2024-07-24
Remove backwards compatibility for `el_offline` and `is_optimstic`

macladson:
## Proposed Changes

Remove the `Option` around the `is_optimistic` and `el_offline` fields in the Beacon API responses.

## Additional Info

This was previously added to ensure that older nodes remained backwards compatible with the VC, however these fields are [now mandatory in the spec](https://github.com/ethereum/beacon-APIs/blob/master/apis/node/syncing.yaml#L20), so this `Option` should be removed.


jimmygchen (approved):
Looks good!

jimmygchen:
@mergify queue

PR #6179, opened 2024-07-25
chore: Use reference to an array representing a blob instead of an owned KzgBlob (only for peerdas functions)

kevaundray:
## Issue Addressed

This changes the blob type in the kzg inteface to take a reference to a Blob instead of taking ownership of a Blob. We are still allocating the blob on the stack inside of the library.

## Proposed Changes

## Additional Info


beacon_node/beacon_chain/src/kzg_utils.rs
```diff
@@ -1,12 +1,13 @@
...
 use types::{Blob, DataColumnSidecar, EthSpec, Hash256, KzgCommitment, KzgProof};
 
 /// Converts a blob ssz List object to an array to be used with the kzg
 /// crypto library.
-fn ssz_blob_to_crypto_blob<E: EthSpec>(blob: &Blob<E>) -> Result<KzgBlob, KzgError> {
-    KzgBlob::from_bytes(blob.as_ref()).map_err(Into::into)
+fn ssz_blob_to_crypto_blob<E: EthSpec>(blob: &Blob<E>) -> Result<KzgBlobRef, KzgError> {
+    let blob_bytes: &[u8] = blob.as_ref();
+    Ok(blob_bytes.try_into().expect("expected cell to have size {BYTES_PER_BLOB}. This should be guaranteed by the `FixedVector type"))
 }
```
kevaundray: The main change is here, we push the allocation of the blob into the kzg library

crypto/kzg/src/lib.rs
```diff
@@ -173,7 +189,11 @@ impl Kzg {
     }
 
     /// Computes the cells and associated proofs for a given `blob` at index `index`.
-    pub fn compute_cells_and_proofs(&self, blob: &Blob) -> Result<CellsAndKzgProofs, Error> {
+    #[allow(clippy::needless_lifetimes)]
+    pub fn compute_cells_and_proofs<'a>(
+        &self,
+        blob: KzgBlobRef<'a>,
+    ) -> Result<CellsAndKzgProofs, Error> {
         let blob_bytes: &[u8; BYTES_PER_BLOB] = blob
```
jimmygchen: I think this conversion can now be removed as `KzgBlobRef` is the exact same type?

consensus/types/src/data_column_sidecar.rs
```diff
@@ -128,8 +128,11 @@ impl<E: EthSpec> DataColumnSidecar<E> {
         let blob_cells_and_proofs_vec = blobs
             .into_par_iter()
             .map(|blob| {
-                let blob = KzgBlob::from_bytes(blob).map_err(KzgError::from)?;
-                kzg.compute_cells_and_proofs(&blob)
+                let blob = blob
+                    .as_ref()
+                    .try_into()
+                    .expect("blob should have a guaranteed size due to FixedVector");
+                kzg.compute_cells_and_proofs(blob)
```
jimmygchen, line 135: I think this is the main change that fixes the stack overflow on `das` testing right?

The other changes change mainnet code path - it's probably not something we can merge to the `unstable` branch now, although I think we do want them eventually when `c-kzg` adopts it or when `rust-eth-kzg` is production-ready for mainnet.

I think we can merge this branch to `das` and I'll just cherry-pick the changes relevant to `compute_cells_and_proofs` onto `unstable`.

beacon_node/beacon_chain/src/kzg_utils.rs
jimmygchen: Cool, could we adopt the same change for `c-kzg-4844`? (to prepare it for increased blob count)

jimmygchen (changes requested):
Looks good thanks @kevaundray , please see my comments above!
We can probably merge this to `das` if you agree, and I'll create a PR to merge to `unstable` soon.

consensus/types/src/data_column_sidecar.rs
kevaundray: Yep this is correct

crypto/kzg/src/lib.rs
kevaundray: Oh yep that's true, this conversion is infallible

beacon_node/beacon_chain/src/kzg_utils.rs
kevaundray: I think c-kzg would need to allocate under the hood -- I'll open up a PR when I get back on my laptop to track arguments for and against

kevaundray:
> Looks good thanks @kevaundray , please see my comments above!
> 
> We can probably merge this to `das` if you agree, and I'll create a PR to merge to `unstable` soon.

Yep agreed to all comments -- I'm away from keyboard at the moment, so feel free to cherry-pick changes without me

jimmygchen (approved):
I've update the PR. Thanks again for this @kevaundray !

PR #6184, opened 2024-07-25
limit dial concurrency on Swarm

jxs:
## Issue Addressed

Addresses https://github.com/sigp/lighthouse/pull/6140#pullrequestreview-2198134169 and sets the max concurrent dial attempts to 1 so we don't dial quic and tcp multiaddresses concurrently, therefore being certain that we privilege quic connections


AgeManning (approved):

AgeManning:
@mergifyio queue

PR #6190, opened 2024-07-26
Use upgradable read lock for pubkey cache

michaelsproul:
## Proposed Changes

Small optimisation to avoid taking a write lock on the `pubkey_cache` during block processing in the common case where no new validators are deposited.

An upgradable read lock does not exclude other readers, so this reduces contention on the lock and may make certain ops faster. An upgradable read _does_ exclude other upgradable reads, but seeing as we expect to only be importing ~1 block at a time, this should rarely occur.

This seemed to show a small improvement in attestation verification times when I was benchmarking with @kevaundray & @jimmygchen yesterday, although we didn't quite control all our variables properly so I'll be repeating that experiment shortly.

## Additional Info

We will likely refactor & consolidate the pubkey cache at some point in the near future, but until then I think this is a nice improvement, albeit minor.


jimmygchen (approved):
Looks good!

jimmygchen:
@mergify queue

PR #6192, opened 2024-07-26
Add assertoor tests to CI

eserilev:
## Add assertoor tests to CI

This PR adds a new step to the local_testnet CI. This new step spins up a kurtosis testnet (defined in `assertoor_network_params.yml`) and runs assertoor tests as a github action via `kurtosis-assertoor-github-action`

### Issues w/ assertoor github action

Note that the repository ppa:rmescandon/yq is no longer maintained and does not contain the most up to date version of `yq`

 https://github.com/mikefarah/yq/issues/1308
 
 Installation recommendations for `yq` can be found here:
https://github.com/mikefarah/yq?tab=readme-ov-file#install

I've updated the `local_testnet.yml` script to install `yq` via `snap`

### Choosing which tests to run

The list of assertoor tests can be found here: https://github.com/ethpandaops/assertoor-test/tree/master/assertoor-tests

Theres a few tests that are broken at the momemnt (mev-block-proposal as an example) and a few that take 48 hours to complete. I've picked the most relevant tests that finish running at a reasonable time.

Theres some electra tests that we may want to add in the future


realbigsean:
I don't think we need `opcodes` tests, those sound like they'd be EVM specific

`big-calldata-tx-test` might be useful for us as it effects blocks size

`stability-check` and `synchronized-check` also sound interesting but not sure what they are

eserilev:
the `mev-block-proposal-check` test fails for both lighthouse and prysm across multiple EL combos. not sure whats up with that, I've left that test out for now

dapplion:
Can you upload the logs to the CI run artifacts? To debug failures

eserilev:
logs should now be uploaded with a 3 day retention window

.github/workflows/local-testnet.yml
```diff
@@ -69,6 +69,14 @@ jobs:
         run: ./stop_local_testnet.sh local-blinded
         working-directory: scripts/local_testnet
 
+      - name: Start a local testnet and run assertoor tests
+        uses: ethpandaops/kurtosis-assertoor-github-action@v1
+        with:
+          ethereum_package_args: scripts/tests/assertoor_network_params.yaml
+          path: |
+            scripts/local_testnet/logs
+          retention-days: 3
```
jimmygchen, line 78: Looks like these keys are invalid for this github action. (see warnings on the [run](https://github.com/sigp/lighthouse/actions/runs/11433456088))
The step below defines the artifact path and retention, I think this can be removed?

jimmygchen: We already upload the local testnet logs, did you mean to upload the assertoor logs? They could be included in the enclave logs too but i haven't checked

scripts/tests/assertoor_network_params.yaml
```diff
@@ -0,0 +1,23 @@
...
+global_log_level: debug
+snooper_enabled: false
+additional_services: 
+  - assertoor
+assertoor_params:
+  run_stability_check: true
+  run_block_proposal_check: true
+  run_blob_transaction_test: true
+  tests:
+    - https://raw.githubusercontent.com/ethpandaops/assertoor-test/master/assertoor-tests/big-calldata-tx-test.yaml
```
jimmygchen, line 23: Looks like the `big-calldata-tx-test.yaml` is the one taking the most time (>20m) and doesn't run concurrently with the other 3 tests. Maybe we can exclude this one from the CI job and have it as a recurring test on GitHub action?

PR #5067, opened 2024-01-15
Tree states database upgrade 🏗️

michaelsproul:
## Issue Addressed

Closes #4482
Closes #4484

## Proposed Changes

Add a database upgrade to v24 for `tree-states`. This allows existing Lighthouse nodes to upgrade to tree-states — one of the final pieces required for us to stabilise the feature.

## Additional Info

Implementation is still a little rough, so this is WIP pending a bit more polish and testing.

PR #6193, opened 2024-07-28
Implement DB schema upgrade for hierarchical state diffs

dapplion:
Implements the DB migration required for

## Proposed Changes

This schema upgrade does the following:

- Maintain archived blocks, and migrate BeaconBlockRoots to the new non-chunked format
- Destructively delete all archived states, and expect a new reconstruction run

Directly migrating archived states to the new format will be complex. It's simpler to reason about to re-trigger reconstruction. Because of the destructive nature of the operation users must pass the explicit flag `--allow-tree-states-migration` only if they have some archival data in the previous format.


dapplion:
check-code is failing with unrelated issue

```
error: doc list item missing indentation
  --> consensus/swap_or_not_shuffle/src/lib.rs:10:5
   |
10 | //! Runs in less time than it takes to run `shuffle_list`.
   |     ^
   |
   = help: if this is supposed to be its own paragraph, add a blank line
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#doc_lazy_continuation
   = note: `-D clippy::doc-lazy-continuation` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(clippy::doc_lazy_continuation)]`
```

michaelsproul:
We probably need to merge `unstable` into `tree-states-archive` to get the latest lint fixes

michaelsproul:
Done

beacon_node/beacon_chain/src/schema_change/migration_schema_v22.rs
```diff
@@ -0,0 +1,86 @@
...
+        return Err(Error::DestructiveFreezerUpgrade);
+    }
+
+    let mut ops = vec![];
+
+    rewrite_block_roots::<T>(&db, anchor.oldest_block_slot, split_slot, &mut ops, &log)?;
+
+    let mut genesis_state = db
+        .load_cold_state_by_slot(Slot::new(0))?
+        .ok_or(Error::MissingGenesisState)?;
```
michaelsproul: This doesn't work. This will try to load the genesis state as a snapshot, but it hasn't been stored as a snapshot yet.

I'll try to fix it shortly.

michaelsproul (approved):

beacon_node/beacon_chain/src/schema_change/migration_schema_v22.rs
michaelsproul: Fixed

michaelsproul:
The approach to loading the genesis state was copied from this PR:

- https://github.com/sigp/lighthouse/pull/5067

PR #6201, opened 2024-07-30
Use blinded blocks for light client proofs

michaelsproul:
## Issue Addressed

A user on Discord reported the following error when running with the light client server enabled:

> Jul 29 23:30:22.614 ERRO error computing light_client updates DBError(HotColdDBError(MissingFullBlockExecutionPayloadPruned(0x86d1262c063c06cfe49a36bcfda8f606e33aca2821a10b33f9332e98ef761507, Slot(2199359)))), service: lc_update

This is due to the light client logic trying to load a full block from the database when the execution payload has been pruned (which is the default if the slot is finalized).

## Proposed Changes

Use blinded blocks in all light client logic. There's no need to have the full execution payload at present, and I don't foresee us needing it in future.

This also enabled a few simplifications.


eserilev:
really nice simplifications, this looks good to me!

eserilev (approved):
LGTM

dapplion (approved):

michaelsproul:
@mergify queue

PR #6202, opened 2024-07-30
Remove double-locking in `eth/v1/node/syncing`

michaelsproul:
## Issue Addressed

While investigating a slow Holesky node, I noticed that `/eth/v1/node/syncing` responses were taking 4-10s on occasion! Even with the beacon processor bypassed.

## Proposed Changes

This PR removes double-locking of the `canonical_head` rw-lock. Previously we would lock it once to load the head slot, and then again to load the optimistic status. Now we just lock it once, grab both pieces of data, and then drop it.


jimmygchen (approved):
Looks good!

dapplion (approved):

jimmygchen:
@mergify queue

PR #6203, opened 2024-07-30
Swap finalized chains based on processed batches

AgeManning:
## Issue Addressed

Previously we decided whether to swap finalized chains based on validated patches. It is a more robust approach to make this decision based on processed batches. 

This PR shifts the switch to use processed batches rather than validated.


jimmygchen (approved):
Simple fix, looks good!

michaelsproul:
@mergify queue

dapplion:
👍 Fix makes sense!

PR #6204, opened 2024-07-30
Add time metrics to aquire fork-choice lock

dapplion:
## Issue Addressed

Better visibility into lock contention issues regarding the fork-choice

## Proposed Changes

Add basic time metrics to aquire fork-choice lock, without labels

## Possible improvements

- Label the histogram extending `fn fork_choice_write_lock(label: &str)` so that each caller tags the time measurement by its location.
- Wrap the returned type into `(RwLockWriteGuard<BeaconForkChoice<T>>, HistogramTimer)` to always time how long the lock is acquired for. Is there a cleaner way to achieve the same goal generically? Can we mess with `RwLockWriteGuard`?


beacon_node/beacon_chain/src/metrics.rs
```diff
@@ -582,6 +582,20 @@ pub static FORK_CHOICE_PROCESS_ATTESTATION_TIMES: LazyLock<Result<Histogram>> =
             "Time taken to add an attestation to fork choice",
         )
     });
+pub static FORK_CHOICE_READ_LOCK_AQUIRE_TIMES: LazyLock<Result<Histogram>> = LazyLock::new(|| {
+    try_create_histogram_with_buckets(
+        "beacon_fork_choice_read_lock_aquire_seconds",
+        "Time taken to aquire the fork-choice read lock",
+        exponential_buckets(1e-3, 2.0, 10),
```
michaelsproul: I think we could start the lower end at 0.1ms, as I think on healthy nodes the lock should take a lot less than 1ms to acquire? We could also set the exponential factor to 4x so that it ramps up quicker?

```suggestion
        exponential_buckets(1e-4, 4.0, 7),
```

beacon_node/beacon_chain/src/metrics.rs
```diff
@@ -582,6 +582,20 @@ pub static FORK_CHOICE_PROCESS_ATTESTATION_TIMES: LazyLock<Result<Histogram>> =
...
+    )
+});
+pub static FORK_CHOICE_WRITE_LOCK_AQUIRE_TIMES: LazyLock<Result<Histogram>> = LazyLock::new(|| {
+    try_create_histogram_with_buckets(
+        "beacon_fork_choice_write_lock_aquire_seconds",
+        "Time taken to aquire the fork-choice write lock",
+        exponential_buckets(1e-3, 2.0, 10),
```
michaelsproul:
```suggestion
        exponential_buckets(1e-3, 4.0, 7),
```

michaelsproul (approved):

michaelsproul:
@mergify queue

PR #6205, opened 2024-07-30
Add metrics inside fork-choice crate

dapplion:
## Issue Addressed

Debugging an unrelated issue, I noted that the metric `beacon_fork_choice_process_attestation_seconds` is imprecise because it measures two distinct spans:
- in gossip processing = lock acquire time + attestation process time
- in block processing = attestation process time

Instead, we can move the timer calls inside the fork-choice crate and always measure the same thing.

## Proposed Changes

- Consistent timing measures for fork-choice calls
- Add metrics for fork-choice data structures

## Notes

I have maintained the same metrics names as they were in the beacon_chain crate. 
- Is it okay to break the metric names for consistency?
- Should the metric names be prefixed by `beacon_fork_choice` or just `fork_choice`?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3099,6 +3099,7 @@ dependencies = [
  "beacon_chain",
  "ethereum_ssz",
  "ethereum_ssz_derive",
+ "lighthouse_metrics",
  "proto_array",
  "slog",
  "state_processing",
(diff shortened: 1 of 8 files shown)


michaelsproul (approved):
LGTM

michaelsproul:
@mergify queue

PR #6207, opened 2024-07-30
Skip recursive discovery query if no useful ENRs

dapplion:
## Issue Addressed

Lighthouse's basic simulator logs are completely un-usable because of the spam from the sequence of logs:
```
2024-07-30T08:50:55.2102984Z Jul 30 08:50:55.160 DEBG Starting a new peer discovery query     wanted: 6, outbound: 1, target: 6, connected: 1, service: libp2p, service: node_1
2024-07-30T08:50:55.2104783Z Jul 30 08:50:55.160 DEBG Starting a peer discovery request       target_peers: 6, service: libp2p, service: node_1
2024-07-30T08:50:55.2106299Z Jul 30 08:50:55.160 DEBG Discovery query completed               peers_found: 1, service: libp2p, service: node_1
2024-07-30T08:50:55.2108037Z Jul 30 08:50:55.160 DEBG Starting a new peer discovery query     wanted: 6, outbound: 1, target: 6, connected: 1, service: libp2p, service: node_1
2024-07-30T08:50:55.2109828Z Jul 30 08:50:55.160 DEBG Starting a peer discovery request       target_peers: 6, service: libp2p, service: node_1
2024-07-30T08:50:55.2111755Z Jul 30 08:50:55.160 DEBG Discovery query completed               peers_found: 1, service: libp2p, service: node_1
```

The root cause is that there are not enough peers in the network ever to satisfy the needs of the peer manager.

The peer manager enters an infinite loop of discovery queries that never yield new useful peers.

@AgeManning Tried to fix the issue with this PR but it doesn't work. The first condition of current_count < target is never met, so the spam still happens even with https://github.com/sigp/lighthouse/pull/6162
- https://github.com/sigp/lighthouse/pull/6162

Note that this infinite loop can theoretically happen in Mainnet too, if somehow the pool of ENRs shrinks to less than our target, or someone tries to run lighthouse with `--target-peers 10000`. 

## Proposed Changes

IMO the recursiveness is an optimization that we can safely skip under certain circumstances to prevent the infinite loop. This PR implements the heuristic that if:
- A query includes > 0 results
- None of those results is diable

Then we have likely exhausted the discv5 table and should stop the recursive query. Note that if there are false positives the heartbeat will trigger queries latter anyway.

I have tested this PR locally with `cargo run --release --bin simulator basic-sim` and there is no more spam of the logs above and the node still reaches the expected 5 target peers quickly.

### DIFF
--- a/beacon_node/lighthouse_network/src/peer_manager/mod.rs
+++ b/beacon_node/lighthouse_network/src/peer_manager/mod.rs
@@ -323,6 +323,7 @@ impl<E: EthSpec> PeerManager<E> {
     #[allow(clippy::mutable_key_type)]
     pub fn peers_discovered(&mut self, results: HashMap<Enr, Option<Instant>>) {
         let mut to_dial_peers = 0;
+        let results_count = results.len();
         let connected_or_dialing = self.network_globals.connected_or_dialing_peers();
         for (enr, min_ttl) in results {
             // There are two conditions in deciding whether to dial this peer.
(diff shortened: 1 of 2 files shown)


beacon_node/lighthouse_network/src/peer_manager/mod.rs
```diff
@@ -323,6 +323,7 @@ impl<E: EthSpec> PeerManager<E> {
     #[allow(clippy::mutable_key_type)]
     pub fn peers_discovered(&mut self, results: HashMap<Enr, Option<Instant>>) {
         let mut to_dial_peers = 0;
+        let results_count = results.len();
```
jxs, line 326: nit: we are just using `results_count`  for the `if` conditional and the `debug!` log, I'd say it's faster to understand the code if we just use `results.len()` as there is one less layer of indirection.

jxs (approved):
LGTM! Lion

beacon_node/lighthouse_network/src/peer_manager/mod.rs
dapplion: result is consumed at line 328, I think it's best to store the len in a variable than clone all ENRs

jxs: right, thanks for explaining Lion :)

AgeManning (approved):
Yeah agreed. I think this is a good soln. Nice!

AgeManning:
My changes to the DAS branch are superseded by this I think. So we should revert those when merging down.

jimmygchen:
@mergify queue

Issue #6113, opened 2024-07-16
Investigate stuck / slow range sync on PeerDAS networks


jimmygchen:
## Description

#6004 fixes a bunch of known issues and Lighthouse is now able to slowly sync with peers, provided that peers are able to respond to data column by range requests. However it's not really reliable and gets "stuck" quite easily, so I suspect there are some other issues out there that haven't been discovered.

I don't have the logs with me now, however it's quite easy to reproduce:
1. Start a local testnet with the `network_params_das_local.yaml` config
2. Stop one Lighthouse node, and wait for 2-3 epochs to make sure it triggers range sync
3. Start the Lighthouse node, notice that sync gets stuck pretty quickly

This is now a bit harder to test due to #6108 and it might make sense to get to the bottom of that one first, or alternatively we could try to run a testnet with Prysm / Teku, whichever is able to serve the requests.

jimmygchen:
Fixed in #6276

Issue #6108, opened 2024-07-16
Lighthouse unable to reliably serve `DataColumnsByRange`


jimmygchen:
## Description

When testing sync locally, a syncing Lighthouse node isn't able to download data columns from it's Lighthouse peer reliably. The network seems to be functioning fine with 100% participation and all peers in sync. However the peer returns 0 columns most of the time. See logs below.

To reproduce:
1. Start a local testnet with the `network_params_das_local.yaml` config
2. Stop one Lighthouse node, and wait for 2-3 epochs to make sure it triggers range sync
3. Start the Lighthouse node, notice that sync gets stuck pretty quickly and peers don't return columns

## Version

`das` branch

## Present Behaviour

Logs from syncing node:

```
Jul 15 07:27:09.430 DEBG Sending DataColumnsByRange requests, peer: 16Uiu2HAm4HH1Bioy2Qp75Rymx2X6TZwTnWeGwmjLKMHAZ5evVmcc, columns: [16, 24, 50, 64, 80, 88], epoch: 0, count: 32, method: DataColumnsByRange, service: sync, module: network::sync::network_context:398
Jul 15 07:27:09.430 DEBG Sending DataColumnsByRange requests, peer: 16Uiu2HAmKxsGh6atb4WQDVz3gb37XaJbi3MZNW8ZRsSwkBu7URYT, columns: [0, 114], epoch: 0, count: 32, method: DataColumnsByRange, service: sync, module: network::sync::network_context:398
Jul 15 07:27:09.430 DEBG Sending DataColumnsByRange requests, peer: 16Uiu2HAm4HH1Bioy2Qp75Rymx2X6TZwTnWeGwmjLKMHAZ5evVmcc, columns: [0, 16, 24, 50, 64, 80, 88, 114], epoch: 1, count: 32, method: DataColumnsByRange, service: sync, module: network::sync::network_context:398
Jul 15 07:27:09.431 DEBG Sending DataColumnsByRange requests, peer: 16Uiu2HAm4HH1Bioy2Qp75Rymx2X6TZwTnWeGwmjLKMHAZ5evVmcc, columns: [0, 16, 24, 80, 88], epoch: 2, count: 32, method: DataColumnsByRange, service: sync, module: network::sync::network_context:398
[...]

Issue #6106, opened 2024-07-16
Lighthouse node get instantly banned by peers on restart due to excessive concurrent requests


jimmygchen:
## Description

It looks like the restarting node may have sent out too many requests and exceeded the number of `inbound_substreams` and getting insta-bannded by lighthouse peers (Seeing `HandlerRejected` in logs):

https://github.com/sigp/lighthouse/blob/bf4cbd3b0a11fdb810a9e0bc1c3c3ed0f8873689/beacon_node/lighthouse_network/src/rpc/handler.rs#L881-L886

After a restart, the beacon node would usually have to perform some block lookups to sync to the latest head, it seems like we've exceed the max number of concurrent requests allowed per peer (32 [defined here](https://github.com/sigp/lighthouse/blob/bf4cbd3b0a11fdb810a9e0bc1c3c3ed0f8873689/beacon_node/lighthouse_network/src/rpc/handler.rs#L38))

logs:

```
Jul 15 07:04:45.903 DEBG Received DataColumnsByRoot Request, returned: 6, request: [(0xd9a57b85fd527477dbd8188a16c1949d0849f5dba5cb364194c8183a0a82696e, [81, 18, 34, 82, 46, 110])], peer: 16Uiu2HAmHMG5brx3krT1rbsFFPYGkNKTRxoRbmSqQHQbbiPqNqwq, module: network::network_beacon_processor::rpc_methods:425
Jul 15 07:04:46.376 DEBG RPC Error, direction: Incoming, score: 0.000008898132403756511, peer_id: 16Uiu2HAmHMG5brx3krT1rbsFFPYGkNKTRxoRbmSqQHQbbiPqNqwq, client: Lighthouse: version: v5.2.1-84fdf0f+, os_version: x86_64-linux, err: Handler rejected the request, protocol: data_column_sidecars_by_root, service: libp2p, module: lighthouse_network::peer_manager:489
Jul 15 07:04:46.376 DEBG Peer has been banned, score: -100.00, peer_id: 16Uiu2HAmHMG5brx3krT1rbsFFPYGkNKTRxoRbmSqQHQbbiPqNqwq, service: libp2p, module: lighthouse_network::peer_manager::peerdb:1097
[...]

PR #6209, opened 2024-07-31
Add cli flag to enable sampling and disable by default

jimmygchen:
## Issue Addressed

Add a `--enable-sampling` flag to enable data column sampling, and disable sampling by default. 

Sampling is currently not on the critical path (and it might be the case that we ship PeerDAS without sampling) but is currently causing some issues on local testnets (e.g. #6106).

Disabling it now so we can focus on other more blocking issues, such as sync and serving data columns by range. (#6113, #6108)


dapplion (approved):
Agree, was thinking of doing the same

Issue #6206, opened 2024-07-30
Expensive fork-choice queued attestation mutation


dapplion:
## Description

Debugging a Holesky node we noted that the split_off here was moving ~1 GB of data every time we process an attestation
https://github.com/sigp/lighthouse/blob/c52c598f6922059900d60723bb541add6865a87e/consensus/fork_choice/src/fork_choice.rs#L258

Questions:
1. Why did the queued attestations Vec grow so large?
2. Why is it being mutated every call?
3. Do we have to use a Vec for this?

Tackling 2 and 3, we could switch to a `HashMap<Slot, Vec<QueuedAttestation>>` as time strictly advances forward, so each slot Vec is strictly append-only.

WIP of this approach: https://github.com/sigp/lighthouse/compare/stable...dapplion:lighthouse:fork-choice-queued-attestations?expand=1

However we should answer 1 too as there may be some other lingering issue

michaelsproul:
Nice detective work @jimmygchen !

jimmygchen:
I've done some further investigation and noticed an invalid attestation (bad `data.slot`) in fork choice, causing this strange behaviour. However I think it's not possible for this to happen as the validation [here](https://github.com/sigp/lighthouse/blob/c52c598f6922059900d60723bb541add6865a87e/consensus/fork_choice/src/fork_choice.rs#L970) prevents this invalid attestation to be queued at all.

I think it might be worth running a memtest on the machine.

jimmygchen:
Further investigation shows that this cannot happen (see the link above) and @michaelsproul found that the bad attestation's slot actually has 1 bit flipped compared to its correct slot, so it could very well be a memory / disk issue.

I couldn't get anything from memtests on that machine. I think we could deploy v5.3.0 to more testnet nodes and see if it happens elsewhere (including slasher), if it doesn't we could probably conclude this bug to be likely a hardware issue, and move this issue out of v5.3.0.

michaelsproul:
If this happens more often with nodes that _have been_ running slashers, maybe it is some UB in the slasher code, potentially the C database code!

We could try running a slasher node under valgrind

jimmygchen:
This undefined behaviour might have surfaced since the recent refactor in #4529, which revealed a bug in the LMDB bindings. See #6211 for a workaround fix.

Under normal circumstances the queued attestations vec should be quite small as there are validations to prevent attestations for future slots to be queued. It would be useful to add a metric for this though.

[...]

PR #4529, opened 2023-07-21
Redb slasher backend impl

eserilev:
## Issue Addressed

#4424 

## Proposed Changes

add redb implementation to the slasher backend

PR #6211, opened 2024-08-01
Work around UB in LMDB bindings

michaelsproul:
## Issue Addressed

Fix a crazy memory corruption bug in the LMDB slasher impl.

## Proposed Changes

Copy the contents of the cursor's `value` prior to returning it, so that LMDB can't mutate the data being pointed to when `.delete_current()` is called.

## Additional Info

This possibly explains the strange behaviour seen on some slasher nodes:

The bug had existed latently for a long time, but only now became executable with the refactor in:

- https://github.com/sigp/lighthouse/pull/4529

### DIFF
--- a/slasher/src/database/lmdb_impl.rs
+++ b/slasher/src/database/lmdb_impl.rs
@@ -165,8 +165,12 @@ impl<'env> Cursor<'env> {
     }
 
     pub fn get_current(&mut self) -> Result<Option<(Key<'env>, Value<'env>)>, Error> {
+        // FIXME: lmdb has an extremely broken API which can mutate the SHARED REFERENCE
+        // `value` after `get_current` is called. We need to convert it to a Vec here in order
+        // to avoid `value` changing after another cursor operation. I think this represents a bug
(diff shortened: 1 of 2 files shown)


michaelsproul:
Thanks @jimmygchen for helping debug this.

I'll have another follow-up PR that fixes the file descriptor leak in the iterated tests. I want to keep it separate from this PR because it involves `unsafe`.

jimmygchen (approved):
🎉

jimmygchen:
Nice, let's deploy this fix to our testnet slasher.

jimmygchen:
@mergify queue

michaelsproul:
For reference, I think the exact type of UB triggered was:

> Mutating immutable bytes. All bytes inside a [const](https://doc.rust-lang.org/reference/items/constant-items.html) item are immutable. The bytes owned by an immutable binding are immutable, unless those bytes are part of an [UnsafeCell<U>](https://doc.rust-lang.org/std/cell/struct.UnsafeCell.html).

From https://doc.rust-lang.org/reference/behavior-considered-undefined.html

PR #6118, opened 2024-07-16
chore: update peerdas-kzg library

kevaundray:
## Issue Addressed

This adds the following changes from upstream:

- Removed the need to call `prover_context` and `verifier_context` when accessing methods
- [Replaced rust-embed with include_str!](https://github.com/crate-crypto/peerdas-kzg/pull/77) due to an issue in devnet
- Removed subgroup checks that were being done on the trusted setup. Its safer to have them, though you really only need to do it once, if your input is constant.

I've added a benchmark for initializing the context for peerdas-kzg and c-kzg (I can remove the benchmark, was just incase you wanted it) -- though note that it benchmarks c-kzg (4844) which does a lot less things than the peerdas variant. 

## Proposed Changes

Please list or describe the changes introduced by this PR.

## Additional Info

Please provide any additional information. For example, future considerations
or information useful for reviewers.

PR #6212, opened 2024-08-01
Add PeerDAS KZG lib integration (construction & KZG verification)

jimmygchen:
## Issue Addressed

Add PeerDAS KZG library and use it for data column construction and KZG verification.

## Proposed Changes

Upstream PRs:
- #5701
- #5941
- #6118 
- #6179

Thanks @kevaundray for the PRs!

## Additional Info

I've also moved `build_data_column_sidecars` and `reconstruct_data_columns` functions from `consensus/types` to `kzg_utils` in the `beacon_chain` crates, where all other KZG related logic are located.

This PR doesn't include the following, in order to keep the PR size moderate and solely focused on KZG lib integration:
- Constructing the data column sidecars when proposing a block
- PeerDAS KZG spec tests


beacon_node/beacon_chain/src/metrics.rs
```diff
@@ -1645,6 +1645,12 @@ pub static BLOB_SIDECAR_INCLUSION_PROOF_COMPUTATION: LazyLock<Result<Histogram>>
             "Time taken to compute blob sidecar inclusion proof",
         )
     });
+pub static DATA_COLUMN_SIDECAR_COMPUTATION: LazyLock<Result<Histogram>> = LazyLock::new(|| {
+    try_create_histogram(
+        "data_column_sidecar_computation_seconds",
+        "Time taken to compute data column sidecar, including cells, proofs and inclusion proof",
+    )
```
dapplion, line 1653: Are the default bucket values good enough?

dapplion (approved):
LGTM, ok to move the code to kzg_utils

beacon_node/beacon_chain/src/metrics.rs
jimmygchen: Definitely not! (we probably overuse the default bucket too much)
I've udpated in https://github.com/sigp/lighthouse/pull/6212/commits/2fdcc356731d76c4fe3faf083e1b2347f54575a based on some local benchmarks. Let me know what you think.

jimmygchen:
I'm looking into the windows CI failure.

beacon_node/beacon_chain/src/metrics.rs
dapplion: Looks good!

crypto/kzg/src/lib.rs
```diff
@@ -167,52 +173,68 @@ impl Kzg {
...
+        Ok((cells, c_kzg_proof))
     }
 
     /// Verifies a batch of cell-proof-commitment triplets.
     ///
     /// Here, `coordinates` correspond to the (row, col) coordinate of the cell in the extended
     /// blob "matrix". In the 1D extension, row corresponds to the blob index, and col corresponds
     /// to the data column index.
-    pub fn verify_cell_proof_batch(
+    #[allow(clippy::needless_lifetimes)]
```
realbigsean: I was curious why this lifetime isn't necessary. Looks like you can get rid of this lint, and the lifetime int he method if you use an anonymous lifetime in the param `CellRef<'_>`

crypto/kzg/src/lib.rs
```diff
@@ -167,52 +173,68 @@ impl Kzg {
     }
 
     /// Computes the cells and associated proofs for a given `blob` at index `index`.
-    #[allow(clippy::type_complexity)]
-    pub fn compute_cells_and_proofs(
+    #[allow(clippy::needless_lifetimes)]
```
realbigsean: same comment re: lifetime

crypto/kzg/src/lib.rs
```diff
@@ -167,52 +173,68 @@ impl Kzg {
...
+        // Modify the result so it matches roughly what the previous method was doing.
+        match verification_result {
+            Ok(_) => Ok(()),
+            Err(e) if e.invalid_proof() => Err(Error::KzgVerificationFailed),
+            Err(e) => Err(Error::PeerDASKZG(e)),
+        }
     }
 
-    pub fn recover_all_cells(
+    #[allow(clippy::needless_lifetimes)]
```
realbigsean: same comment re: lifetime

crypto/kzg/src/lib.rs
```diff
@@ -29,42 +43,34 @@ impl From<c_kzg::Error> for Error {
...
 }
 
 impl Kzg {
     /// Load the kzg trusted setup parameters from a vec of G1 and G2 points.
     pub fn new_from_trusted_setup(trusted_setup: TrustedSetup) -> Result<Self, Error> {
+        // Initialize the trusted setup using default parameters
+        //
+        // Note: One can also use `from_json` to initialize it from the consensus-specs
+        // json string.
+        let peerdas_trusted_setup = PeerDASTrustedSetup::default();
```
realbigsean: Looks like there's an embedded trusted setup here, so we're loading the same setup twice right? so maybe we can at least add an equivalence check here if there's no way to consolidate the two yet

beacon_node/beacon_chain/src/kzg_utils.rs
```diff
@@ -19,6 +34,50 @@ pub fn validate_blob<E: EthSpec>(
...
+                .iter()
+                .map(|&proof| Bytes48::from(proof))
+        })
+        .collect::<Vec<_>>();
+
+    let column_indices = data_column_iter
+        .clone()
+        .flat_map(|data_column| {
+            let col_index = data_column.index;
+            (0..data_column.column.len()).map(move |_| col_index)
```
realbigsean:
```suggestion
            data_column.column.iter().map(move |_| col_index)
```

beacon_node/beacon_chain/src/kzg_utils.rs
```diff
@@ -76,3 +135,299 @@ pub fn verify_kzg_proof<E: EthSpec>(
...
+
+    Ok(sidecars)
+}
+
+/// Reconstruct all data columns from a subset of data column sidecars (requires at least 50%).
+pub fn reconstruct_data_columns<E: EthSpec>(
+    kzg: &Kzg,
+    data_columns: &[Arc<DataColumnSidecar<E>>],
+    spec: &ChainSpec,
+) -> Result<Vec<Arc<DataColumnSidecar<E>>>, KzgError> {
```
realbigsean:
```suggestion
) -> Result<DataColumnSidecarList<E>, KzgError> {
```

beacon_node/beacon_chain/src/kzg_utils.rs
```diff
@@ -76,3 +135,299 @@ pub fn verify_kzg_proof<E: EthSpec>(
...
+        .map(|blob| {
+            let blob = blob
+                .as_ref()
+                .try_into()
+                .expect("blob should have a guaranteed size due to FixedVector");
+            kzg.compute_cells_and_proofs(blob)
+        })
+        .collect::<Result<Vec<_>, KzgError>>()?;
+
+    for (blob_cells, blob_cell_proofs) in blob_cells_and_proofs_vec {
```
realbigsean, line 195: From here to the end of the method looks like the same logic exists in `reconstruct_data_columns`, maybe we can make a common method

realbigsean (changes requested):

crypto/kzg/src/lib.rs
jimmygchen: Yep that's right. Yeah I think equivalence check is a good idea, or potentially cloning the values from existing trusted setup. There are a few more things I've done on another WIP branch, but I'm still thinking about whether it's a good idea:
- Should we only load this if PeerDAS is scheduled (`None` otherwise), so we don't impact mainnet code path now (I have a branch [here](https://github.com/sigp/lighthouse/blob/89fd62a3c08c0e099099a0883c703e01943d55dc/crypto/kzg/src/lib.rs#L61-L73) that does this)
- Load PeerDAS setup from existing `trusted_setup` using clone ([code](https://github.com/sigp/lighthouse/blob/89fd62a3c08c0e099099a0883c703e01943d55dc/crypto/kzg/src/lib.rs#L66)), so we avoid reading file system and also avoid breaking `--trusted-setup` flag - although we don't really use this anymore

beacon_node/beacon_chain/src/kzg_utils.rs
jimmygchen: Refactored in https://github.com/sigp/lighthouse/pull/6212/commits/9b69f1db20e36a8cad27f0bee8583cb84305f82e. Thanks!

crypto/kzg/src/lib.rs
jimmygchen: Nice, fixed in https://github.com/sigp/lighthouse/pull/6212/commits/9b69f1db20e36a8cad27f0bee8583cb84305f82e

crypto/kzg/src/lib.rs
jimmygchen: Fixed in https://github.com/sigp/lighthouse/pull/6212/commits/9b69f1db20e36a8cad27f0bee8583cb84305f82e

beacon_node/beacon_chain/src/kzg_utils.rs
jimmygchen: Fixed in https://github.com/sigp/lighthouse/pull/6212/commits/9b69f1db20e36a8cad27f0bee8583cb84305f82e

crypto/kzg/src/lib.rs
jimmygchen: I'm going to cherry-pick both changes into this branch.

With (2), I  just did a `bench` on both approaches (loading from json / clone and encode existing trusted setup) - cloning is actually _slightly_ faster than re-loading from json (-20ms), plus we get consistency, and we won't need to perform equivalent check.

jimmygchen: I haven't done #1, it actually involves [touching a lot of files](https://github.com/jimmygchen/lighthouse/commit/89fd62a3c08c0e099099a0883c703e01943d55dc) and I'm not sure if it's worth it, as loading the PeerDAS trusted setup only adds ~550ms on startup and ~100mb of memory for storing the precomputation. Existing Deneb code path still uses c-kzg so should be unaffected.

Once both libraries are feature complete and stable for deneb and PeerDAS, we could do this properly and only load one library with a cli flag.

What do you think @realbigsean ?

jimmygchen: Replaced loading embedded trusted setup with cloning the existing trusted setup values.
https://github.com/sigp/lighthouse/pull/6212/commits/11c53a81b915eef3b0a73cec657c5bc177f6db6c

crypto/kzg/src/lib.rs
```diff
@@ -29,42 +45,34 @@ impl From<c_kzg::Error> for Error {
...
-    pub fn into_inner(self) -> [u8; 2048usize] {
-        self.bytes
-    }
-}
-
 /// A wrapper over a kzg library that holds the trusted setup parameters.
 #[derive(Debug)]
 pub struct Kzg {
     trusted_setup: KzgSettings,
+    context: DASContext,
```
dapplion: Is it possible to conditionally load the DAS context for peerdas networks only? It's kindoff a bad penalty to force mainnet users to pay the extra 110MB in memory for an unused feature

crypto/kzg/src/lib.rs
jimmygchen: Added conditionally load KZG on startup only if PeerDAS is enabled: https://github.com/sigp/lighthouse/pull/6212/commits/688198155d9c2fa030710a0d9b5e1f029b84f8f5

crypto/kzg/src/lib.rs
jimmygchen: Done in https://github.com/sigp/lighthouse/pull/6212/commits/688198155d9c2fa030710a0d9b5e1f029b84f8f5

dapplion (approved):
LGTM

crypto/kzg/src/lib.rs
realbigsean: > Once both libraries are feature complete and stable for deneb and PeerDAS, we could do this properly and only load one library with a cli flag.

That sounds good!

realbigsean (approved):

jimmygchen:
@mergify queue

PR #5301, opened 2024-02-26
Remove use of ethers core

dospore:
## Issue Addressed

Working from #5207 to begin conversion from ethers_core to alloy

## Proposed Changes

Uses alloy RlpEncodable/RlpDecodable types instead of ethers_core rlp stream

## Additional Info

There are many more uses of ethers_core in this repository so thought I would reduce the number of changes and potentially put more pr's in the future.

PR #6213, opened 2024-08-02
Delete legacy payload reconstruction

michaelsproul:
## Issue Addressed

I noticed that even after recent changes (see linked PR) to remove `ethers`, we still have it in the `Cargo.lock`.

- https://github.com/sigp/lighthouse/pull/5301

## Proposed Changes

`ethers` was still being used for two things:

1. Legacy payload reconstruction, which dates to the time before the existence of the `getPayloadBodies` methods.
2. Some extra tests in `execution-integration-tests`

This PR addresses just the first of these uses. We can update the `execution-integration-tests` in a follow-up PR.

I've opted just to delete the legacy payload reconstruction, as the only place where it was marginally useful was in `bellatrix_readiness` for checking genesis validity. Even with it gone, there are still some checks on Bellatrix genesis that should prevent mistakes.

IMO the amount of work required to rewrite this logic into `alloy` is not worth it at the moment.

## Additional Info

`ethers` depends on `ethereum-types`, so this PR also helps towards:

- https://github.com/sigp/lighthouse/issues/6022


beacon_node/beacon_chain/src/beacon_block_streamer.rs
```diff
@@ -864,147 +862,4 @@ mod tests {
             }
         }
     }
-
-    #[tokio::test]
-    async fn check_fallback_altair_to_electra() {
```
michaelsproul, line 869: This test was testing the legacy payload reconstruction, so it's no longer relevant.

beacon_node/execution_layer/src/lib.rs
```diff
@@ -1823,7 +1823,9 @@ impl<E: EthSpec> ExecutionLayer<E> {
 
         // Use efficient payload bodies by range method if supported.
         let capabilities = self.get_engine_capabilities(None).await?;
-        if capabilities.get_payload_bodies_by_range_v1 {
+        if capabilities.get_payload_bodies_by_range_v1
+            || capabilities.get_payload_bodies_by_range_v2
```
michaelsproul, line 1827: This is a pre-emptive bug fix. If ELs remove support for v1 in future we still want to use v2 by calling `get_payload_bodies_by_range` below.

jimmygchen (approved):
> I've opted just to delete the legacy payload reconstruction, as the only place where it was marginally useful was in bellatrix_readiness for checking genesis validity. Even with it gone, there are still some checks on Bellatrix genesis that should prevent mistakes.
> IMO the amount of work required to rewrite this logic into alloy is not worth it at the moment.

Agree with this and the change looks great! -600 less lines of code 🔪

jimmygchen:
I was hoping we can remove `ethers-core` from `execution_layer` but looks like some tests still use it. Can address this in a seprate PR.

jimmygchen:
@mergify queue

michaelsproul:
Thanks Jimmy!

Yeah next PR in the line to remove ethers is this one:

- https://github.com/sigp/lighthouse/pull/6366

Then probably another one to do the execution integration tests

PR #5409, opened 2024-03-13
feat: add peerdas custody field to ENR

jacobkaufmann:
## Issue Addressed

#4983

## Proposed Changes

* add `custody_subnet_count` field to ENR as defined in EIP-7594 (i.e. peerdas)
* add ENR subnet predicate for data column subnet
* add functions to `DataColumnSubnetId` to compute custody obligations
* add minimum custody requirement to `EthSpec`

## Additional Info

target `das` branch

PR #6216, opened 2024-08-02
Add plumbing for PeerDAS supernodes (#5050, #5409, #5570, #5966)

jimmygchen:
## Issue Addressed

Part of #6072. 

Add plumbing for peerdas supernodes (#5050, #5409, #5570, #5966)
- add cli option `--subscribe-to-all-data-columns`
- add custody subnet count to ENR, only if PeerDAS is scheduled
- subscribe to data column topics, only if PeerDAS is scheduled

Upstream PRs:
- #5050
- #5409 (Thanks @jacobkaufmann ❤️ )
- #5570
- #5966


book/src/help_bn.md
```diff
@@ -585,6 +585,10 @@ Flags:
           server on localhost:5052 and import deposit logs from the execution
           node. This is equivalent to `--http` on merge-ready networks, or
           `--http --eth1` pre-merge
+      --subscribe-all-data-column-subnets
```
dapplion: If we are going to commit to this flag, is this the best name? Another angle is to have a `--supernode` flag. I don't have a strong opinion but now is the time to think it twice

realbigsean: I like `--supernode` better

beacon_node/network/src/service.rs
```diff
@@ -733,6 +737,17 @@ impl<T: BeaconChainTypes> NetworkService<T> {
                     }
                 }
 
+                // TODO(das): This is added here for the purpose of testing, *without* having to
+                // activate Electra. This should happen as part of the Electra upgrade and we should
+                // move the subscription logic once it's ready to rebase PeerDAS on Electra, or if
+                // we decide to activate via the soft fork route:
+                // https://github.com/sigp/lighthouse/pull/5899
+                let peer_das_scheduled = self.fork_context.spec.eip7594_fork_epoch
```
realbigsean: In `ChainSpec` we interpret a `None` on a fork epoch as equivalent to `FAR_FUTURE_EPOCH` (fork unset). So this could incorrectly return true

beacon_node/lighthouse_network/src/discovery/enr.rs
```diff
@@ -221,6 +238,16 @@ pub fn build_enr<E: EthSpec>(
 
     builder.add_value(SYNC_COMMITTEE_BITFIELD_ENR_KEY, &bitfield.as_ssz_bytes());
 
+    // only set `csc` if PeerDAS fork epoch has been scheduled
+    if spec.eip7594_fork_epoch != Some(spec.far_future_epoch) {
```
realbigsean: Same comment here

realbigsean (dismissed):

book/src/help_bn.md
jimmygchen: Just dropping some thoughts here - I thought about a few alternatives when originally picking this name
- `--custody-subnet-count 12`: number of column subnets to custody, this is flexible but we don't really have a use case, and isn't really compatible with the upcoming validator custody and custody group changes.
- `--subscribe-all-data-column-subnets`: this name is consistent with the attestation subnet flag, but thinking about this now, it's actually a bit different, because it means subscribe all subnets + import all data columns + reconstruction + distributed blob building etc. so `--supernode` does feel a bit more suitable.
- `--min-data-column-custody <number|all>`: with the upcoming validator custody change, the number of custody columns may change during runtime, depending on the number of validators registered. Note I dropped the `subnet` here because of the upcoming change to decouple subnets from custody.

We've used the `supernode` term a lot between implementers, but feels a bit vague for people without the PeerDAS context, e.g. would a regular user think it subscribes to all attestation subnets and import all attestations too, or would it have a higher `--network-load` value? 

The downside is that it is a bit rigid and we may need to add more flags, if one wants to run a lower bandwidth and lower storage supernode, they _may_ want to subscribe to just 50% of the subnets - just to be able to locally perform reconstruction without having to query peers when having to serve `blob_sidecars` over HTTP.

Btw all other clients have adopted the same thing so far:

```yaml
participants:
# Supernodes
  - cl_type: prysm
    cl_image: ethpandaops/prysm-beacon-chain:peerdas-initial-sync
    cl_extra_params: [--subscribe-all-subnets]
  - cl_type: lighthouse
    cl_image: ethpandaops/lighthouse:das-devnet-2
    cl_extra_params: [--subscribe-all-data-column-subnets]
  - cl_type: teku
    cl_image: ethpandaops/teku:nashatyrev-das
    cl_extra_params: [--p2p-subscribe-all-custody-subnets-enabled]
  - cl_type: nimbus
    cl_image: ethpandaops/nimbus-eth2:peerdas-p2p
    cl_extra_params: [--subscribe-all-subnets=true]
```

beacon_node/lighthouse_network/src/discovery/enr.rs
jimmygchen: Fixed in https://github.com/sigp/lighthouse/pull/6216/commits/94f1d3dd93622b358e52b1ee79febbed67292c84

beacon_node/network/src/service.rs
jimmygchen: Thanks, I've added a `spec.is_peer_das_scheduled` function that checks for both `None` and `FAR_FUTURE_EPOCH` values. 
https://github.com/sigp/lighthouse/commit/94f1d3dd93622b358e52b1ee79febbed67292c84

book/src/help_bn.md
jimmygchen: we should probably hide all peerdas flags at this stage?

jimmygchen: btw I also don’t have a strong opinion, happy to switch to supernode, and I think we can still change our mind before go live

dapplion (approved):
Ok to keep the name, but we should either hide it or note it that it is experimental and may break

book/src/help_bn.md
jimmygchen: Updated help and made this a hidden flag.

dapplion (approved):
LGTM!

jimmygchen:
@mergify queue

PR #6217, opened 2024-08-02
patch quick-protobuf

jxs:
`quick-protobuf` has become unmaintained, for now let's use our fork whilst a more definitive solution is considered.


michaelsproul:
I rebased on `release-v5.3.0` so we can include it. Also reverted some of the general `cargo update` stuff to keep the diff minimal. Added the `rev` to the `Cargo.toml` too so that it's constrained and won't shift without us meaning to shift it

michaelsproul:
@mergify queue

michaelsproul (approved):

michaelsproul:
@mergify refresh

PR #6218, opened 2024-08-03
Store computed custody subnets in PeerDB and fix custody lookup test

jimmygchen:
## Issue Addressed

This PR stores the computed custody subnets in `PeerDB`, so that we don't have to compute it every time we do a data column query.

Additionally it contains a fix to the failing custody lookup test - the `peer_id` was previously randomly generated and didn't match the ENR generated in [`__add_connected_peer_testing_only`](https://github.com/sigp/lighthouse/blob/abeb795cb16d2e0de0813539a1b1cff60afb67d6/beacon_node/lighthouse_network/src/peer_manager/peerdb.rs#L707), and we were therefore not finding peers in the `PeerDB`.


dapplion (approved):
LGTM!

PR #6219, opened 2024-08-04
Batch attestation slashibility checking

### LINKED ISSUE #1914: Batch attestation slashability checking
## Description

With large numbers of validators and heavy disk congestion, we've observed the slashing protection database timing out with errors like this:

```
: Nov 16 00:50:31.056 INFO Successfully published attestation      type: unaggregated, slot: 39850, committee_index: 2, head_block: 0x3bd75ebc65de718b36911eaab7dad3a9ef7ca44c7ba03d32c7092195c43bcf43, service: attestation
: Nov 16 00:50:32.838 CRIT Not signing slashable block             error: SQLPoolError("Error(None)")
: Nov 16 00:50:32.838 CRIT Error whilst producing block            message: Unable to sign block, service: block
```

The default timeout is 5 seconds.

Presently, we sign attestations one at a time and broadcast them, which means that each attester requires a new SQLite database transaction.

https://github.com/sigp/lighthouse/blob/master/validator_client/src/attestation_service.rs#L332-L404

To alleviate the congestion slightly, we could switch to an algorithm like:

1. Begin an SQLite transaction `txn`
2. Check and sign all attestations as part of `txn`
3. Commit `txn`
4. Broadcast attestations

That way we preserve the property that an attestation is only broadcast if its signature has been persisted to disk (which is crash-safe). Broadcasting after each check but _before_ the transaction commits would violate this property.

## Version

Lighthouse v0.3.4

eserilev:
## Issue Addressed

Closes  #1914 

## Proposed Changes

The attestation service spawns a thread for each validator duty for a given slot. Within that thread the service checks if the attestation is safe, and if safe signs and publishes it. The attestation safety check requires reading and writing to a Sqlite db instance. Sqlite only allows for one open write transaction at any given time, and that transaction cannot be shared across threads. Opening a write transaction on one thread will block all other threads. This PR aims to refactor the attestation service so that database read/writes are not causing performance bottlenecks. These performance issues only become evident when large numbers of validators are involved.

Threads are still spawned per validator duty. However, within each thread we immediately sign the attestation. We collect the signed attestations and then perform the attestation safety checks. The db related tasks all happen in a single thread. The attestations that are deemed safe are all broadcasted at once and then the service continues with its normal flow.

I've also introduced an attestation data serivce. Its a simple service that downloads attestations by slot/committee index from the BN and caches them for later. It also prevents us from trying to download duplicate attestation data and helps abstract away the differences between how we handle attestation data pre and post electra.

Additionally, i've also parallelized producing and publishing attestation aggregates.


validator_client/src/attestation_data_service.rs
```diff
@@ -0,0 +1,84 @@
...
+
+use crate::{
+    beacon_node_fallback::{BeaconNodeFallback, OfflineOnFailure, RequireSynced},
+    http_metrics::metrics,
+};
+
+/// The AttestationDataService is responsible for downloading and caching attestation data at a given slot
+/// for a range of committee indexes. It also helps prevent us from re-downloading identical attestation data.
+pub struct AttestationDataService<T: SlotClock, E: EthSpec> {
+    attestation_data_by_committee: HashMap<CommitteeIndex, AttestationData>,
```
dapplion: Note that all attestations votes are the same on each committee, the only change is in the `index` field which equals the key of this hashmap `CommitteeIndex`. If you are refactoring the flow you can consider an optimization to only get a single `AttestationData` for all committees.

However, some DVT solution relies on a call for each `CommitteeIndex` being made. I'm not sure if this is the case but it was ~1.5 years ago.

eserilev:
This one should be ready for a review. I still need to add some more granular metrics though.

michaelsproul:
Closing in favour of:

I tried to fix conflicts, but there were too many changes to the underlying VC code, and it was a bit of a mess

PR #6221, opened 2024-08-05
Don't expect DAS config in HTTP spec response

michaelsproul:
## Issue Addressed

Lighthouse v5.3.0 RC validator clients unintentionally broke backwards compatibility with prior BNs due to some peer DAS fields that were considered mandatory in the `/eth/v1/config/spec` response:

> ERRO Unable to read spec from beacon node endpoint: https://localhost:5052/, error: HttpClient(, kind: decode, detail: missing field `CUSTODY_REQUIREMENT` at line 1 column 4281)

## Proposed Changes

- Add `default` annotations for the peer DAS fields of `Config` so the VC doesn't fail to decode responses that lack these fields. This is not necessarily a great approach, but it's what we've done for every fork so far. It might be nice to clean this up in future.

## Additional Info

In future we could also remove the `default`s for past forks.

### DIFF
--- a/consensus/types/src/chain_spec.rs
+++ b/consensus/types/src/chain_spec.rs
@@ -1365,10 +1365,13 @@ pub struct Config {
     #[serde(with = "serde_utils::quoted_u64")]
     max_per_epoch_activation_exit_churn_limit: u64,
 
+    #[serde(default = "default_custody_requirement")]
     #[serde(with = "serde_utils::quoted_u64")]
     custody_requirement: u64,
+    #[serde(default = "default_data_column_sidecar_subnet_count")]
     #[serde(with = "serde_utils::quoted_u64")]
     data_column_sidecar_subnet_count: u64,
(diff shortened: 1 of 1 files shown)


dapplion (approved):

realbigsean (approved):

michaelsproul:
@mergify queue

jimmygchen:
@mergify requeue

michaelsproul:
@mergify requeue

michaelsproul:
Rebasing this so we can re-run CI here on the PR until it succeeds, and then merge instantly

michaelsproul:
@mergify queue

michaelsproul:
@mergify dequeue

michaelsproul:
@mergify queue

michaelsproul:
@mergify refresh

michaelsproul:
@mergify queue

michaelsproul:
@mergify requeue

PR #6222, opened 2024-08-05
Deprecate mplex

AgeManning:
Mplex is deprecated and is becoming a liability. 

We currently support two better alternatives, yamux (for TCP) and QUIC (which has its own multiplexer) over UDP. 

The issue is that other clients may still depend solely on mplex. This PR should not be merged until it has been brought to the attention of all the client teams.

Relevant spec change: https://github.com/ethereum/consensus-specs/pull/3866

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4671,25 +4671,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "libp2p-mplex"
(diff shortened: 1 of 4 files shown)


dapplion:
Sorry @AgeManning we want the PR number down, we can reopen when mplex is actually dead

Issue #5232, opened 2024-02-14
VC fallback `broadcast` runs in series


paulhauner:
## Description

The [`BeaconNodeFallback::broadcast`](https://github.com/sigp/lighthouse/blob/c7e5dd1098a145c0d2174dc65bd23faeb5074249/validator_client/src/beacon_node_fallback.rs#L584-L589) function broadcasts in *series* when I think it would be more effective in *parallel*.

The issue with a series broadcast is that the last node in the BN list doesn't receive the message until waiting for the full round-trip of all other BNs.

This presents two immediate issues:

1. It's likely that time-sensitive messages (like sync committees) will frequently arrive late to the last node in the BN list, creating spammy logs.
2. If an BN early in the list times out then we're very likely to be too late for all other BNs. This is the case for blocks.

## Steps to resolve

Instead of this sort of pattern:

```rust
for node in nodes {
  node.publish().await
}
```

We should adopt something like:

```rust
use tokio::task::JoinSet;

let mut set = JoinSet::new();

// Start executing the futures in parallel.
for node in nodes {
  set.spawn(node.publish())
}

// Process the results as they come in.
while let Some(response) = set.join_next().await {
  // handle `response`
}
```

michaelsproul:
Released in v5.3.0 via:

- #6223

PR #6223, opened 2024-08-05
Broadcast VC requests in parallel and fix subscription error

michaelsproul:
## Issue Addressed

Closes:

- https://github.com/sigp/lighthouse/issues/5232

## Proposed Changes

Use `join_all` to run futures in parallel.

This seemed a bit cleaner than using `JoinSet`, and is similarly efficient. Each loop iteration should be quick because `.status().await` just checks an rwlock, it doesn't actually make any requests to the BN.

This PR also fixes a bug whereby we wouldn't mark subscriptions as sent if _any_ node failed. This meant that in the case of 1 offline node, subscriptions would continue getting spammed at the BN right up to the duty slot, which would result in the infamous warning:

> WARN Not enough time for a discovery search

This is fixed in 2 ways:

- We check if we are too close to the duty slot in `should_send_subscription_at`
- We mark subscriptions as sent as long as one node receives them (we also just log a warning in this case rather than an error).

Along the way I also added a 500ms timeout for subscription requests, which previously had _no timeout_! I'm open to tweaking this timeout or putting it in another PR.

## Additional Info

This improvement should help the performance of clusters of nodes when one of them goes AWOL due to an issue like:

### DIFF
--- a/validator_client/src/beacon_node_fallback.rs
+++ b/validator_client/src/beacon_node_fallback.rs
@@ -190,11 +190,6 @@ impl<E: EthSpec> CandidateBeaconNode<E> {
         spec: &ChainSpec,
         log: &Logger,
     ) -> Result<(), CandidateError> {
-        debug!(
-            log,
-            "Refresh started";
-            "endpoint" => %self.beacon_node,
-        );
         let previous_status = self.status(RequireSynced::Yes).await;
         let was_offline = matches!(previous_status, Err(CandidateError::Offline));
 
@@ -213,12 +208,6 @@ impl<E: EthSpec> CandidateBeaconNode<E> {
         // status. I deem this edge-case acceptable in return for the concurrency benefits of not
         // holding a write-lock whilst we check the online status of the node.
         *self.status.write().await = new_status;
-
-        debug!(
-            log,
-            "Refresh complete";
-            "endpoint" => %self.beacon_node,
-        );
 
         new_status
     }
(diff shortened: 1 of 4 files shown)


realbigsean (approved):
LGTM

validator_client/src/beacon_node_fallback.rs
```diff
@@ -599,46 +599,44 @@ impl<T: SlotClock, E: EthSpec> BeaconNodeFallback<T, E> {
         F: Fn(&'a BeaconNodeHttpClient) -> R,
         R: Future<Output = Result<O, Err>>,
     {
-        let mut results = vec![];
         let mut to_retry = vec![];
         let mut retry_unsynced = vec![];
 
         // Run `func` using a `candidate`, returning the value or capturing errors.
         //
         // We use a macro instead of a closure here since it is not trivial to move `func` into a
```
jimmygchen: The macro comment can be removed now

michaelsproul:
This isn't working as well as expected (we still seem to have requests getting massively delayed), so I'm trying some improvements

michaelsproul:
I think I've fixed the bug in https://github.com/sigp/lighthouse/pull/6223/commits/112679535f4f0e1e9c98aefa44fa1cff5ad82f22

Rolling out now

jimmygchen:
Random fact: 
> The probability of getting the first 9 characters of a Git commit hash as digits (0-9) is approximately 1.455%

michaelsproul:
@mergify queue

PR #6234, opened 2024-08-06
add missing use std::sync::Lazylock to malloc_utils::glibc

jxs:


jimmygchen (approved):
Thanks for fixing this @jxs

jimmygchen:
@mergify queue

PR #5196, opened 2024-02-05
Add `DataColumnSidecarsByRoot ` req/resp protocol

jimmygchen:
## Issue Addressed

Part of #4983.

## Proposed Changes

Add implementation for `DataColumnSidecarsByRoot`.

Currently blobs are served from three sources, in the following order
1. data availability checker (before we have all the blocks and blobs)
2. early attester cache (before we import the block)
3. blobs db (after import)

I've implemented the equivalent for data columns, and it currently reuses the blobs db but with a different `DbColumn` to separate data columns from blobs. 

This PR does not cover the following:
1. Sending outbound requests
2. Handling of responses from peers

PR #6237, opened 2024-08-07
Implement PeerDAS RPC handlers

dapplion:
## Issue Addressed

Continue porting network changes of `das` branch. Extends https://github.com/sigp/lighthouse/pull/6238

Part of 

_all branches are on the sipg/lighthouse repo to start another PR conga :)_

## Proposed Changes

- Implement serving RPC handlers for `data_columns_by_root` and `data_columns_by_range`. 
- This PR **excludes** the feature to reprocess early sampling requests.

Upstream PRs:

- https://github.com/sigp/lighthouse/pull/5196

### DIFF
--- a/beacon_node/network/src/network_beacon_processor/rpc_methods.rs
+++ b/beacon_node/network/src/network_beacon_processor/rpc_methods.rs
@@ -320,14 +320,53 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
     pub fn handle_data_columns_by_root_request(
         self: Arc<Self>,
         peer_id: PeerId,
-        _request_id: PeerRequestId,
+        request_id: PeerRequestId,
         request: DataColumnsByRootRequest,
     ) {
-        // TODO(das): implement handler
-        debug!(self.log, "Received DataColumnsByRoot Request";
-            "peer_id" => %peer_id,
-            "count" => request.data_column_ids.len()
+        let mut send_data_column_count = 0;
+
+        for data_column_id in request.data_column_ids.as_slice() {
+            match self.chain.get_data_column_checking_all_caches(
+                data_column_id.block_root,
+                data_column_id.index,
+            ) {
+                Ok(Some(data_column)) => {
+                    send_data_column_count += 1;
+                    self.send_response(
+                        peer_id,
+                        Response::DataColumnsByRoot(Some(data_column)),
+                        request_id,
+                    );
+                }
+                Ok(None) => {} // no-op
+                Err(e) => {
+                    self.send_error_response(
+                        peer_id,
+                        RPCResponseErrorCode::ServerError,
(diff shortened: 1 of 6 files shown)


jimmygchen:
Oops I thought this was the first PR in the conga line and incorrectly set the base branch to `unstable`. I've reverted my merge commit and it's now back to what it was before.

jimmygchen:
I've merged latest base branch into this branch so i can extend the conga line.

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -320,14 +320,53 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
     pub fn handle_data_columns_by_root_request(
```
realbigsean, line 320: should this use `terminate_response_stream` like we use in other similar methods?

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -320,14 +320,53 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                    return;
+                }
+            }
+        }
+
+        debug!(
+            self.log,
+            "Received DataColumnsByRoot Request";
+            "peer" => %peer_id,
+            "request" => ?request.group_by_ordered_block_root(),
```
realbigsean, line 376: what is the benefit of ordering by block roots in this log?

realbigsean (approved):
one small suggestion and one question but looks good

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
dapplion: I introduced `group_by_ordered_block_root` originally to handle the reprocessing of requests. I left it here as it produces a more succint output. The raw request repeats the root for every column. Usually peers requests multiple indexes for the same root, so the output of this is nicer.

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
dapplion: Added in aa0b9037c689de617a9092cdfc0bc9fa5283a4f2

dapplion:
@jimmygchen I have forced push a rebase of this branch against unstable, as the commit from https://github.com/sigp/lighthouse/pull/6238 had leaked into this PR

realbigsean:
@mergify queue

realbigsean:
@mergify refresh

realbigsean:
@mergify queue

realbigsean:
@mergify dequeue

realbigsean:
@mergify queue

realbigsean:
@mergify requeue

PR #6224, opened 2024-08-05
Implement data columns by network boilerplate

dapplion:
## Issue Addressed

Start porting network changes of `das` branch. This PR introduces mindless boilerplate around the new RPC routes data columns by range and by root. Both will be used by PeerDAS custody sync on both range sync and lookup sync.

Part of 
- https://github.com/sigp/lighthouse/issues/4983

## Proposed Changes

Upstream PRs:

- https://github.com/sigp/lighthouse/pull/5196
- https://github.com/sigp/lighthouse/pull/5353

PR #6238, opened 2024-08-07
Add PeerDAS RPC import boilerplate

dapplion:
## Issue Addressed

Continue porting network changes of `das` branch. Extends https://github.com/sigp/lighthouse/pull/6224

Part of 

## Proposed Changes

- Add a new `Work:: RpcCustodyColumn` event
- Wire process function with importing columns into da_checker
- Introduce new type `DataColumnsSameBlock` as downstream code assumes that all submitted columns are in the same block. The implementation in `das` just assumes this to be the case, but this PR makes this assumption explicit via a new type. I changed the gossip import method from accepting N columns to exactly 1, so we can assert that the single column is always part of the same block.

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -23,9 +23,7 @@ use crate::chain_config::ChainConfig;
 use crate::data_availability_checker::{
     Availability, AvailabilityCheckError, AvailableBlock, DataAvailabilityChecker,
 };
-use crate::data_column_verification::{
-    DataColumnsSameBlock, GossipDataColumnError, GossipVerifiedDataColumn,
-};
+use crate::data_column_verification::{GossipDataColumnError, GossipVerifiedDataColumn};
 use crate::early_attester_cache::EarlyAttesterCache;
 use crate::errors::{BeaconChainError as Error, BlockProductionError};
 use crate::eth1_chain::{Eth1Chain, Eth1ChainBackend};
@@ -2970,13 +2968,23 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
 
     /// Cache the data columns in the processing cache, process it, then evict it from the cache if it was
     /// imported or errors.
-    pub async fn process_gossip_data_column(
+    pub async fn process_gossip_data_columns(
         self: &Arc<Self>,
-        data_column: GossipVerifiedDataColumn<T>,
+        data_columns: Vec<GossipVerifiedDataColumn<T>>,
     ) -> Result<AvailabilityProcessingStatus, BlockError<T::EthSpec>> {
+        let Ok((slot, block_root)) = data_columns
+            .iter()
+            .map(|c| (c.slot(), c.block_root()))
+            .unique()
+            .exactly_one()
+        else {
+            return Err(BlockError::InternalError(
+                "Columns should be from the same block".to_string(),
+            ));
+        };
+
         // If this block has already been imported to forkchoice it must have been available, so
         // we don't need to process its samples again.
-        let block_root = data_column.block_root();
         if self
             .canonical_head
             .fork_choice_read_lock()
@@ -2986,7 +2994,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         }
 
         let r = self
-            .check_gossip_data_columns_availability_and_import(data_column)
+            .check_gossip_data_columns_availability_and_import(slot, block_root, data_columns)
             .await;
         self.remove_notified_custody_columns(&block_root, r)
     }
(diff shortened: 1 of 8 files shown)


beacon_node/network/src/network_beacon_processor/mod.rs
```diff
@@ -476,6 +477,24 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
         })
     }
 
+    /// Create a new `Work` event for some custody columns. `process_rpc_custody_columns` reports
+    /// the result back to sync.
+    pub fn send_rpc_custody_columns(
```
dapplion, line 481: Future PR touching sync will consume this function

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -192,20 +212,18 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
         )
     }
 
-    pub fn put_gossip_data_columns(
+    pub fn put_gossip_data_column(
         &self,
-        slot: Slot,
-        block_root: Hash256,
-        gossip_data_columns: Vec<GossipVerifiedDataColumn<T>>,
+        gossip_data_column: GossipVerifiedDataColumn<T>,
```
jimmygchen: I changed this in #5806 to avoid reconstruction for proposer nodes. Is this scenario handled?

dapplion: Right, thanks for pointing that out I wasn't aware. `publish_block` also ensures that all columns are from the same block. Do you think we need the `*SameBlock` type here too? Switching to a single item was just a way to get around having to create a `GossipVerifiedDataColumnsSameBlock` or similar.

jimmygchen: I think it was clear that these are from the same block given the `block_root` and `slot` parameters. Is it necessary to introduce a new type? as we always either pass a single data column around, or pass a list of data columns for the same block. 

Could we maybe delay these kind of refactor in separate future PRs. Feels like it makes conga line PRs difficult to manage (we'll also have to re-work all subsequent PRs)

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -2968,23 +2970,13 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
 
     /// Cache the data columns in the processing cache, process it, then evict it from the cache if it was
```
realbigsean:
```suggestion
    /// Cache the data column in the processing cache, process it, then evict it from the cache if it was
```

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -197,20 +217,18 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
...
-        let epoch = slot.epoch(T::EthSpec::slots_per_epoch());
-        let custody_columns = gossip_data_columns
-            .into_iter()
-            .map(|c| KzgVerifiedCustodyDataColumn::from_asserted_custody(c.into_inner()))
-            .collect::<Vec<_>>();
+        let custody_column = gossip_data_column.into_inner();
+
+        // Will never error as there's exactly one column
+        let custody_columns_same_block = KzgVerifiedDataColumnsSameBlock::new(vec![custody_column])
+            .expect("exactly one column is always in same block");
```
realbigsean: I think it'd be better to make a new infallible constructor in `KzgVerifiedDataColumnsSameBlock` that takes a single column as a param

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -221,34 +222,91 @@ impl<E: EthSpec> KzgVerifiedDataColumn<E> {
...
-    pub fn from_asserted_custody(kzg_verified: KzgVerifiedDataColumn<E>) -> Self {
-        Self {
-            data: kzg_verified.to_data_column(),
+impl<E: EthSpec> DataColumnsSameBlock<E> {
+    pub fn new(columns: DataColumnSidecarList<E>) -> Result<Self, &'static str> {
+        let first_column = columns.first().ok_or("empty columns")?;
+        let signed_block_header = first_column.signed_block_header.clone();
+        for column in columns.iter().skip(1) {
+            if column.signed_block_header != signed_block_header {
+                return Err("no same block");
```
realbigsean: `not same block` or maybe more detailed`mismatched data column headers` or something

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -221,34 +222,91 @@ impl<E: EthSpec> KzgVerifiedDataColumn<E> {
...
+    columns: Vec<KzgVerifiedDataColumn<E>>,
+}
+
+impl<E: EthSpec> KzgVerifiedDataColumnsSameBlock<E> {
+    pub fn new(columns: Vec<KzgVerifiedDataColumn<E>>) -> Result<Self, &'static str> {
+        let first_column = columns.first().ok_or("empty columns")?;
+        let signed_block_header = first_column.as_data_column().signed_block_header.clone();
+        for column in columns.iter().skip(1) {
+            if column.as_data_column().signed_block_header != signed_block_header {
+                return Err("no same block");
```
realbigsean: same as above

realbigsean (commented):
Just some small feedback but looks good!

dapplion:
@jimmygchen I have reverted the refactor of the *SameBlock types. This PR just extends same style as blobs

Also rebased on unstable to clean the commits :)

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -181,6 +183,38 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
             .put_kzg_verified_blobs(block_root, epoch, verified_blobs)
     }
 
+    /// Put a list of custody columns received via RPC into the availability cache. This performs KZG
+    /// verification on the blobs in the list.
+    #[allow(clippy::type_complexity)]
```
realbigsean:
```suggestion
```

i think this is not necessary

realbigsean (approved):
Looks like my feedback was related to the refactor, just had one nit!

realbigsean:
@mergify queue

Issue #6232, opened 2024-08-06
Fallback node out of sync briefly during block production due to race condition


jimmygchen:
## Description

Our basic simulator tests have been flaky and I think it may be caused by a race condition.

i think whats happening now is, we have validator 0 proposing a block, broadcasting to node 0 and 1:
- node 0 publishes the block
- node 1 receives it via gossip, and also receives it via http (fallback node). However this block doesn't get imported via either path, and the node would get out of sync for one slot until a block lookup is triggerd.

The race condition prevents the block import and return `RepeatBlob` in both code paths when gossip verifying the blobs.

In the simulator test setup (1 validator with 2 beacon nodes), it impacts its sync committee performance briefly and causes the simulator test to fail, as the fallback node is out of sync briefly.

This will likely be addressed in @michaelsproul's PR here, see the discussion thread here for more details and the investigation
https://github.com/sigp/lighthouse/pull/5574#discussion_r1704944723

jimmygchen:
I believe this has been addressed in #5574 and I haven't seen sim tests failing with this error lately.

PR #6239, opened 2024-08-07
Naive attempt at 0 ports in simulator

michaelsproul:
## Proposed Changes

Work-in-progress PR to try to make the simulator tests more reliable by using zero ports.

The aim is to prevent an `AddrInUse` crash that I got on my machine by running the simulator in a loop:

```fish
while [ $status -eq 0 ]
    cargo run --release --bin simulator fallback-sim
end
```

## Additional Info

Related to:

- https://github.com/sigp/lighthouse/issues/6232

### DIFF
--- a/testing/simulator/src/local_network.rs
+++ b/testing/simulator/src/local_network.rs
@@ -16,11 +16,6 @@ use std::{
 };
 use types::{ChainSpec, Epoch, EthSpec};
 
-const BOOTNODE_PORT: u16 = 42424;
-const QUIC_PORT: u16 = 43424;
-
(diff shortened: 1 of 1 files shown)


AgeManning:
I think we need to specify a port for the bootnodes, so all the nodes can discover each other. 

Its annoying that this cannot be static, which is the stable solution. 

We can attempt our previous solution which is akin to the OS finding an open port for us by using: 
https://github.com/sigp/lighthouse/blob/stable/common/unused_port/src/lib.rs#L75

Then using the resulting port to bind to and to set our bootnodes ENR to.

PR #6242, opened 2024-08-08
Enables TCP port reuse

AgeManning:
This is a temporary PR as this flag has been deprecated in the latest libp2p release. 

We need to ugprade to the latest libp2p version and change this flag to be used in the DialOpts when dialing. However this may be useful for the time being.


michaelsproul:
Closing this now that this PR is merged:

- https://github.com/sigp/lighthouse/pull/6249

PR #6189, opened 2024-07-26
Replace `lazy_static!` with `LazyLock`

jimmygchen:
## Issue Addressed

Replace `lazy_static!` with `LazyLock`, stable from Rust 1.80.

This removes the usage of `lazy_static` crate and we no longer have to worry about macro recursion limit of `lazy_static!`.

PR #6243, opened 2024-08-08
Fix compilation of LCLI

AgeManning:
I think https://github.com/sigp/lighthouse/pull/5922 broke the compilation of LCLI. 

Seems like no one has compiled it in a few months. 

I think this PR fixes it.


jimmygchen:
Oh I actually broke this recently in #6189, @jxs made the same fix in #6234. Going to merge it now.

jimmygchen:
Thanks for the fix! I'm closing this one as a duplicate

PR #6244, opened 2024-08-08
Correct gossipsub mesh and connected peer inconsistencies

AgeManning:
## Issue Addressed

We are seeing errors in gossipsub related to an inconsistent state between mesh peers and connected peers. 

There appears to be a few problems. This PR will be updated as we find all the bugs. 

The first I have noticed is that the IDONTWANT messages also look for in-flight gossipsub promises. When a peer disconnects we dont remove the tracking of in-flight gossipsub promises, because it is inefficient search (it's not index'd by peer-id). Therefore we can have disconnected peers in this mapping and if we go to send an IDONTWANT message, we will not have this peer-id in the connected mapping. 

Am currently still searching for the others....


AgeManning:
Ok, found the other bug. This should fix the libp2p issues I think

beacon_node/lighthouse_network/gossipsub/src/behaviour.rs
```diff
@@ -2043,8 +2043,11 @@ where
             }
         }
 
-        // remove unsubscribed peers from the mesh if it exists
+        // remove unsubscribed peers from the mesh and fanout if they exist there
```
jxs:
```suggestion
        // remove unsubscribed peers from the mesh and fanout if they exist there.
```

beacon_node/lighthouse_network/gossipsub/src/behaviour.rs
```diff
@@ -2706,8 +2712,8 @@ where
 
         for peer_id in recipient_peers {
             let Some(peer) = self.connected_peers.get_mut(peer_id) else {
-                tracing::error!(peer = %peer_id,
```
jxs, line 2716: instead of removing this error, can we instead filter above for the gossip promises? I.e.
```rust
// Gossip promises are kept for disconnected peers.
let iwant_peers = self
            .gossip_promises
            .peers_for_message(msg_id)
            .iter()
            .filter(|peer_id| self.connected_peers.contains_key(peer_id));
```

AgeManning: I was thinking about this. I don't mind. 
The logic that went through my head was:
Adding this check adds a hashmap lookup for each peer, but a few lines down the page we do the exact same check.

I figured we are already doing the check, by just removing the error, we have no extra hash lookup, however the cost is we don't log an error in case a mesh peer isn't in the connected mapping (but we have these errors elsewhere). 

Its such a tiny difference, so happy to do whatever you think is best, just explaining why I removed the error.

jxs: thanks for elaborating age! Do you know where else we check? It def doesn't make sense to double check this adding another iteration as you refer, I just think we shouldn't stop logging for this event

jxs (commented):
btw age, can we also add an entry to the changelog?

beacon_node/lighthouse_network/gossipsub/src/behaviour.rs
AgeManning: So yeah happy to do what you prefer. 

I can't think of a way to keep the error without adding an extra hashmap lookup. Its a trivial lookup, and maybe the error is worth it. So I'm happy to go with the soln you have suggested here if you want the error.

AgeManning:
I've added the changelog :). Feel free to add the error back if you like.

AgeManning:
@jxs - I didnt realise we haven't merged this. We probably want this in

AgeManning:
@mergifyio queue

jxs (approved):
LGTM!

AgeManning:
@mergifyio queue

AgeManning:
@mergifyio refresh

AgeManning:
https://github.com/Mergifyio refresh

AgeManning:
@mergifyio queue

PR #6245, opened 2024-08-08
Implement hierarchical state diffs config migration

dapplion:
## Issue Addressed

Extends the migration PR to include handling the changing DB config
- https://github.com/sigp/lighthouse/pull/6193

## Proposed Changes

Superstruct the on disk DB config and start prefixing the serialization starting from V1.

**NOTE!**: When downgrading, the previous lighthouse version will be unable to deserialize the on disk config as it expects an SZZ `u64`. However, it should be fine as the on_disk config is only there to ensure compatibility.

@michaelsproul please confirm that this is ok, LH will just ignore a badly encoded on disk config

### DIFF
--- a/beacon_node/store/src/config.rs
+++ b/beacon_node/store/src/config.rs
@@ -59,16 +59,27 @@ pub struct StoreConfig {
 
 /// Variant of `StoreConfig` that gets written to disk. Contains immutable configuration params.
 #[superstruct(
-    variants(V0, V1),
+    variants(V1, V22),
     variant_attributes(derive(Debug, Clone, PartialEq, Eq, Encode, Decode))
 )]
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct OnDiskStoreConfig {
-    #[superstruct(only(V0))]
-    pub slots_per_restore_point: u64,
-
     #[superstruct(only(V1))]
+    pub slots_per_restore_point: u64,
+    /// Prefix byte to future-proof versions of the `OnDiskStoreConfig` post V1
+    #[superstruct(only(V22))]
+    version_byte: u8,
+    #[superstruct(only(V22))]
     pub hierarchy_config: HierarchyConfig,
+}
+
+impl OnDiskStoreConfigV22 {
+    fn new(hierarchy_config: HierarchyConfig) -> Self {
+        Self {
+            version_byte: 22,
+            hierarchy_config,
+        }
+    }
 }
 
 #[derive(Debug, Clone)]
@@ -115,9 +126,7 @@ impl Default for StoreConfig {
 
 impl StoreConfig {
     pub fn as_disk_config(&self) -> OnDiskStoreConfig {
-        OnDiskStoreConfig::V1(OnDiskStoreConfigV1 {
-            hierarchy_config: self.hierarchy_config.clone(),
-        })
+        OnDiskStoreConfig::V22(OnDiskStoreConfigV22::new(self.hierarchy_config.clone()))
     }
 
     pub fn check_compatibility(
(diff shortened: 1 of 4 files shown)


beacon_node/store/src/config.rs
```diff
@@ -210,28 +225,67 @@ impl StoreItem for OnDiskStoreConfig {
...
             hierarchy_config: store_config.hierarchy_config.clone(),
+        });
+        let split = Split::default();
+        assert!(store_config
+            .check_compatibility(&on_disk_config, &split, None)
+            .is_ok());
+    }
+
+    #[test]
+    fn check_compatibility_after_migration() {
```
dapplion, line 277: This test shows the case of migrating a V0 config an not raising any compatibility issues

beacon_node/store/src/config.rs
```diff
@@ -57,9 +58,16 @@ pub struct StoreConfig {
 }
 
 /// Variant of `StoreConfig` that gets written to disk. Contains immutable configuration params.
-#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode)]
-// FIXME(sproul): schema migration
+#[superstruct(
+    variants(V0, V1),
```
michaelsproul: Usually the naming convention we use for versioned structs in the database follows the schema version: https://github.com/sigp/lighthouse/tree/stable/beacon_node/beacon_chain/src/schema_change#naming-conventions

```suggestion
    variants(V1, V22),
```

beacon_node/store/src/config.rs
```diff
@@ -210,28 +225,70 @@ impl StoreItem for OnDiskStoreConfig {
...
 
     fn as_store_bytes(&self) -> Vec<u8> {
-        self.as_ssz_bytes()
+        match self {
+            // V0 is not prefixed explicitly, but unless the user chose a crazy configuration the
+            // serialized value will have a leading 0 byte.
+            OnDiskStoreConfig::V0(value) => value.as_ssz_bytes(),
+            OnDiskStoreConfig::V1(value) => {
+                let mut out = value.as_ssz_bytes();
+                out.insert(0, 1);
```
michaelsproul: Rather than doing this, why don't we do it in a slightly more compositional way by adding a `u8` field to the `OnDiskStoreConfig` for `V22` and later? Like:

```rust
struct OnDiskStoreConfig {
    #[only(V1)]
    slots_per_restore_point: u64,
    #[only(V22)]
    version_byte: u8,
    #[only(V22)]
    hierarchy_config: HierarchyConfig,
}
```

Then we can slightly reduce the amount of manual byte manipulation we need to do.

beacon_node/store/src/config.rs
```diff
@@ -210,28 +225,70 @@ impl StoreItem for OnDiskStoreConfig {
...
+            .is_ok());
+    }
+
+    #[test]
+    fn check_compatibility_after_migration() {
+        let store_config = StoreConfig {
+            ..Default::default()
         };
+        let on_disk_config = OnDiskStoreConfig::V0(OnDiskStoreConfigV0 {
+            slots_per_restore_point: 0,
```
michaelsproul:
```suggestion
            slots_per_restore_point: 8192,
```

I think we should use a valid config here, `slots_per_restore_point == 0` was never allowed.

michaelsproul (changes requested):
Looks good! Just a few minor requests

beacon_node/store/src/config.rs
dapplion: Fixed in e5661720685236e73c6fee55998f15ba44a9ed88

beacon_node/store/src/config.rs
dapplion: Fixed in e5661720685236e73c6fee55998f15ba44a9ed88

beacon_node/store/src/config.rs
dapplion: Fixed in e5661720685236e73c6fee55998f15ba44a9ed88

michaelsproul (approved):

PR #6247, opened 2024-08-09
Make `portable` the default in Lighthouse book and Makefile, update workflows

chong-he:
Since `portable` is now the default build, this PR:
- removes some sections in Lighthouse book that are no longer relevant and make `portable` the default where necessary
- updates Makefile to use `portable` as the default build
- removes the `portable` tag in `release.yml` 
- removes the `modern` tag in `docker.yml`


Makefile
```diff
@@ -96,12 +90,8 @@ build-release-tarballs:
 	[ -d $(BIN_DIR) ] || mkdir -p $(BIN_DIR)
 	$(MAKE) build-x86_64
 	$(call tarball_release_binary,$(BUILD_PATH_X86_64),$(X86_64_TAG),"")
-	$(MAKE) build-x86_64-portable
-	$(call tarball_release_binary,$(BUILD_PATH_X86_64),$(X86_64_TAG),"-portable")
```
chong-he, line 100: Not sure if I can delete this and the below `portable` one

jimmygchen: Looks ok to me!

book/src/docker.md
```diff
@@ -77,8 +72,7 @@ The `stability` is:
 
 The `modernity` is:
 
-* `-modern` for optimized builds
-* empty for a `portable` unoptimized build
+* `-portable` for the default optimized builds
```
jimmygchen: modernity is gone from the image tag, so we can remove this, and `modernity` from line 54

jimmygchen (changes requested):
Looks good to me, just added a small comment.
Be good to get a second review from @michaelsproul or @antondlr .

book/src/docker.md
chong-he: Thank you, removed in [9339a49](https://github.com/sigp/lighthouse/pull/6247/commits/9339a497a4a5616a8fa4a119aede6ae48a339ed1)

antondlr (approved):
nice, looking good!
now that we're at it I would prefer to remove the `-dev` tag, too. 
no more confusing double tags for any image!

michaelsproul (approved):
LGTM

Agree we should also get rid of `-dev`

jimmygchen (approved):
LGTM

michaelsproul:
@mergify queue

PR #6249, opened 2024-08-09
update libp2p to version 0.54

jxs:
## Issue Addressed

superseeds #6242 as this version deprecates `port_reuse()` at the `Swarm` level, @michaelsproul do you know where do we need to dial with port reuse?


beacon_node/lighthouse_network/gossipsub/src/behaviour.rs
```diff
@@ -3161,6 +3165,7 @@ where
         peer_id: PeerId,
         _: &Multiaddr,
         _: Endpoint,
+        _: PortUse,
```
michaelsproul, line 3168: Is this related to `SO_REUSEADDR`? From the docs it seems more like a port-reuse option within libp2p itself, but I'm not sure.

The test I was trying to fix was the simulator, which currently crashes if you run it in a loop because the ports aren't fully freed by the time it runs again, and port-reuse is not enabled:

```fish
while [ $status -eq 0 ]
    cargo run --release --bin simulator fallback-sim
end
```

If we can get this test to run indefinitely without crashing, I'll be happy.

See also:

- https://github.com/sigp/lighthouse/pull/6239

beacon_node/lighthouse_network/gossipsub/src/behaviour/tests.rs
```diff
@@ -284,6 +285,7 @@ where
         let fake_endpoint = ConnectedPoint::Dialer {
             address: Multiaddr::empty(),
             role_override: Endpoint::Dialer,
+            port_use: PortUse::New,
```
michaelsproul: If this is `SO_REUSEADDR` I think we'll want `PortUse::Reuse` everywhere. I don't see any harm in always reusing.

jxs: yup you are right Michael, updated!

AgeManning (approved):
lgtm

jxs:
@mergifyio queue

PR #6254, opened 2024-08-13
Prevent fd leak in random slasher tests

michaelsproul:
## Issue Addressed

Presently the random slasher tests crash after a few thousand iterations with an error about file descriptor exhaustion. The reason for this is that we use `Box::leak` to give the slasher database a `'static` lifetime. As a result, if more than 1 `SlasherDB` is ever created, its destructor will not run and the open files will never be closed. Even though the `tempdir` will delete the DB directory containing the files, this does not free the

## Proposed Changes

Reuse the same database for each test rather than creating a new one. The contents of the database are deleted completely between iterations. Some refactoring to allow changing the `config` between iterations was required.

## Additional Info

You can reproduce the crash after a few minutes running this command on `unstable`:

```
cargo test -p slasher --release --test random -- --ignored no_crash --nocapture
```

After this PR, it can run indefinitely.

### DIFF
--- a/slasher/src/database.rs
+++ b/slasher/src/database.rs
@@ -848,7 +848,7 @@ impl<E: EthSpec> SlasherDB<E> {
     }
 
     fn reset_db(&self, txn: &mut RwTransaction<'_>, db: &Database<'static>) -> Result<(), Error> {
-        let mut cursor = txn.cursor(&db)?;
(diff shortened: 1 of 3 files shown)


dapplion (approved):

michaelsproul:
@mergify queue

PR #6256, opened 2024-08-14
Batch sampling requests by peer

ackintosh:
## Issue Addressed

Closes https://github.com/sigp/lighthouse/issues/6106

We're not batching sampling requests to the same peer, so the requests might cause a `Handler rejected` error as the number of substreams exceeds `MAX_INBOUND_SUBSTREAMS`, especially when the node can only request from a single supernode, such as in a local-testnet.

## Proposed Changes

Batching sampling requests.

####  SamplingId

I changed `SamplingId` as below:

```diff
#[derive(Debug, Hash, PartialEq, Eq, Clone, Copy)]
pub struct SamplingId {
    pub id: SamplingRequester,
-   pub column_index: ColumnIndex,
+   pub sampling_request_id: SamplingRequestId,
}
```

First I tried to change the `column_index` field from single to multiple, such as `Vec<ColumnIndex>` or something fixed-length arrays, but these ideas ware bad because it can't implement the `Copy` trait or requires generic parameter to fix its length according to the spec.

So I introduced `SamplingRequestId` instead:

```diff
+ /// Identifier of sampling requests.
+ #[derive(Debug, Hash, PartialEq, Eq, Clone, Copy)]
+ pub struct SamplingRequestId(pub usize);

pub struct ActiveSamplingRequest<T: BeaconChainTypes> {
    block_root: Hash256,
    block_slot: Slot,
    requester_id: SamplingRequester,
    column_requests: FnvHashMap<ColumnIndex, ActiveColumnSampleRequest>,
+   /// Mapping of column indexes for a sampling request.
+   column_indexes_by_sampling_request: FnvHashMap<SamplingRequestId, Vec<ColumnIndex>>,
+   /// Sequential ID for sampling requests.
+   current_sampling_request_id: SamplingRequestId,
    column_shuffle: Vec<ColumnIndex>,
    required_successes: Vec<usize>,
    /// Logger for the `SyncNetworkContext`.
    pub log: slog::Logger,
    _phantom: PhantomData<T>,
}
```

`SamplingRequestId` is a sequential ID for (batched) sampling requests. `ActiveSamplingRequest` manages the mapping between `SamplingRequestId` and batched column indexes.

#### Testing

I have confirmed that sampling requests are batched on the local-testnet.

```
[cl-2-lighthouse-geth] Aug 17 13:14:49.477 DEBG Sending DataColumnsByRoot Request       req_id: 14, requester: Sampling(SamplingId { id: ImportedBlock(0x4e62c1db108f1b26b9879c419acbd9e5413ebcc22133eaf85f0c01fcada0d11c), sampling_request_id: SamplingRequestId(0) }), peer: 16Uiu2HAmRe4gC4J7pt5uiZLsqDEjNosxeg4TpdVN2xjpqC3gMgjp, indices: [48, 105, 106, 21, 63, 53, 40, 114, 44, 3, 98, 93, 34, 2, 14, 61], block_root: 0x4e62c1db108f1b26b9879c419acbd9e5413ebcc22133eaf85f0c01fcada0d11c, method: DataColumnsByRoot, service: sync
```


beacon_node/network/src/sync/sampling.rs
```diff
@@ -311,34 +367,52 @@ impl<T: BeaconChainTypes> ActiveSamplingRequest<T> {
         // Progress requests
         // If request fails retry or expand search
         // If all good return
-        let Some(request) = self.column_requests.get_mut(&column_index) else {
-            warn!(
-                self.log,
-                "Received sampling response for unrequested column index"
-            );
+        let Some(column_indexes) = self
```
jimmygchen, line 365: The comment above is probably outdated now and can be removed

beacon_node/network/src/sync/sampling.rs
```diff
@@ -235,57 +245,103 @@ impl<T: BeaconChainTypes> ActiveSamplingRequest<T> {
...
+                // Handle the downloaded data columns.
+                if data_columns.is_empty() {
+                    debug!(self.log,"Received empty response"; "block_root" => %self.block_root);
+                    self.column_indexes_by_sampling_request
+                        .remove(&sampling_request_id);
+                } else {
+                    // Overwrite `column_indexes` with the column indexes received in the response.
+                    let column_indexes = data_columns.iter().map(|d| d.index).collect::<Vec<_>>();
+                    self.column_indexes_by_sampling_request
+                        .insert(sampling_request_id, column_indexes.clone());
```
jimmygchen, line 306: What's the reason for this insert? 
In this case, the sampling request id should already be in the mapping right?

jimmygchen, line 306: Do we also need to update the `self.column_requests` mapping, when only a subset of columns is returned? e.g. in the case where the response is missing a column, we need to update the `ActiveColumnSampleRequest` status.

jimmygchen (changes requested):
@ackintosh Nice, just a comment above on handling responses containing only a subset of columns. I think we might have to update the column request status and also update `peers_dont_have`?

beacon_node/network/src/sync/sampling.rs
ackintosh: Yeah, in the case where the peer doesn't have the requested column, the `ActiveColumnSampleRequest` status is updated to `NotStarted` here via `request.on_sampling_error()`:

https://github.com/sigp/lighthouse/blob/acedd7a0a9b5610a85bad71f1aed052f2441998b/beacon_node/network/src/sync/sampling.rs#L272-L281

Additionally `peers_dont_have` is updated as well. 🙂

beacon_node/network/src/sync/sampling.rs
ackintosh: Removed in https://github.com/sigp/lighthouse/pull/6256/commits/acedd7a0a9b5610a85bad71f1aed052f2441998b

beacon_node/network/src/sync/sampling.rs
jimmygchen: oh cool, i missed that!  niiice

jimmygchen:
Looks good to me, Thanks ! 🎉

PR #5347, opened 2024-03-04
Validator manager commands for standard key-manager APIs

pahor167:
## Issue Addressed

https://github.com/sigp/lighthouse/issues/4854

## Proposed Changes

Added ability to 

- import validators from standard key-manager
- list validators in VC
- remove validator from VC

## Additional Info

This is my first PR, please let me know if anything is missing.

Issue #4854, opened 2023-10-17
Validator manager commands for standard key-manager APIs


michaelsproul:
## Description

The Lighthouse validator manager `import` command takes a [`Vec<ValidatorSpecification>`](https://github.com/sigp/lighthouse/blob/441fc1691b69f9edc4bbdc6665f3efab16265c9b/validator_manager/src/common.rs#L42) as input, which makes it unsuitable for use with keystores from non-Lighthouse sources.

It would be great to have a command that can import standard keystore files, with options for passwords, etc. Basically an analogue of the old _account-manager_ import command that works with the keymanager API.

For completeness it would also be good to have basic commands for the other key-manager APIs like list, delete, etc, which don't use any Lighthouse-specific extensions. This would allow interop with non-Lighthouse VCs, and provide a nicer front-end to the KM APIs that doesn't involve crafting JSON by hand.

## Version

Lighthouse v4.5.0

## Steps to resolve

Add new commands to `lighthouse validator-manager` which _just_ use the standard APIs.

pahor167:
I created following https://github.com/sigp/lighthouse/pull/5347 - this is my first PR, please let me know if anything is missing

jimmygchen:
@chong-he I think we can now close this one?

chong-he:
Closed in #6261

PR #6261, opened 2024-08-15
Validator manager commands for the Keymanager APIs

chong-he:
## Issue Addressed

* #4854 

## Proposed Changes

Add commonds to `lighthouse vm` to support the Keymanager API to manage local keystores, i.e., `list`, `import` and `delete`

## Additional Info

Building on top of @pahor167 PR: https://github.com/sigp/lighthouse/pull/5347

Some other modifications:
- modify the `import` logging where only successful import will display the log (previously it will display the log once the HTTP API is called even though the import is not successful, e.g., incorrect password)
- added logging for successful `delete`


eserilev:
If it's not too much of a hassle it'd be cool to define the new cli flags here w/ clap derive. Alternatively, I can make a separate PR to migrate to derive once this is merged 

EDIT: I think we can just migrate to clap derive later after this PR is merged. thanks for taking a look CK!

validator_manager/src/list_validators.rs
```diff
@@ -0,0 +1,206 @@
...
+    }
+}
+
+pub async fn run<'a>(config: ListConfig) -> Result<String, String> {
+    let ListConfig {
+        vc_url,
+        vc_token_path,
+    } = config;
+
+    let (http_client, _keystores) = vc_http_client(vc_url.clone(), &vc_token_path).await?;
```
macladson: I think you can actually use the keystores directly here without needing to make a second call to the VC. Something like this:

```suggestion
    let (_, validators) = vc_http_client(vc_url.clone(), &vc_token_path).await?;
```

validator_manager/src/list_validators.rs
```diff
@@ -0,0 +1,206 @@
...
+
+    let validators = http_client.get_keystores().await.unwrap().data;
```
macladson: Then you can remove this line
```suggestion
```

validator_manager/src/delete_validator.rs
```diff
@@ -0,0 +1,266 @@
...
+}
+
+pub async fn run<'a>(config: DeleteConfig) -> Result<(), String> {
+    let DeleteConfig {
+        vc_url,
+        vc_token_path,
+        validator_to_delete,
+    } = config;
+
+    let (http_client, _keystores) = vc_http_client(vc_url.clone(), &vc_token_path).await?;
```
macladson: See my comments in `list_validators`, they should apply here too.

validator_manager/src/delete_validator.rs
```diff
@@ -0,0 +1,266 @@
...
+                .action(ArgAction::Set)
+                .display_order(0),
+        )
+}
+
+#[derive(Clone, PartialEq, Debug, Serialize, Deserialize)]
+pub struct DeleteConfig {
+    pub vc_url: SensitiveUrl,
+    pub vc_token_path: PathBuf,
+    pub validator_to_delete: PublicKeyBytes,
```
macladson: I wonder if it is worth supporting being able to delete vals in bulk. Maybe just starting with supporting comma separation in the CLI for now. Later maybe we could support passing in a file with them in a list or something.

validator_manager/src/import_validator.rs
macladson, line 1: There's a decent amount of code duplication between this and the `import_validators.rs` file.

It would be nice if would could merge the two into the same command, with maybe a `--use-standard-format` flag or something in order to specify which format we are using (next level would be detecting which format it is automagically. I believe we can also make certain flags mutually exclusive so we can disable the ones that are not applicable for each format.

This _may_ not be worth the complexity, but it's just quite unintuitive (imo) to have an `import` subcommand and then also an `import-standard` subcommand.

macladson (commented):
This looks very clean!
I do have a few suggestions for you which range from easy fixes to large rewrites :grin:

validator_manager/src/list_validators.rs
macladson: This also has the added benefit of removing the `unwrap`

validator_manager/src/delete_validator.rs
chong-he: Simplified in [a874551](https://github.com/sigp/lighthouse/pull/6261/commits/a874551ffc9519cd3dd4cdff1df1fa471c82e84a)

validator_manager/src/delete_validator.rs
chong-he: Added the support for comma separated to delete more than 1 validator in [4726501](https://github.com/sigp/lighthouse/pull/6261/commits/472650108dc599ee47a3aa612d5430059a980f50) with a lot of help from @michaelsproul, thank you!

lighthouse/tests/validator_manager.rs
```diff
@@ -203,6 +203,14 @@ pub fn validator_import_defaults() {
                 vc_url: SensitiveUrl::parse("http://localhost:5062").unwrap(),
                 vc_token_path: PathBuf::from("./token.json"),
                 ignore_duplicates: false,
+                password: None,
+                fee_recipient: None,
+                builder_boost_factor: None,
+                gas_limit: None,
+                builder_proposals: Some(false),
```
chong-he: I have to let this to be `Some(false)` for the comparison to be equal. Not sure if this is ok?
The same goes for `prefer_builder_proposals` field

validator_manager/src/import_validators.rs
```diff
@@ -127,13 +227,31 @@ async fn run<'a>(config: ImportConfig) -> Result<(), String> {
...
-    let validators: Vec<ValidatorSpecification> = serde_json::from_reader(&validators_file)
-        .map_err(|e| {
+
+    let validators: Vec<ValidatorSpecification> = if standard_format {
+        vec![ValidatorSpecification {
+            voting_keystore: KeystoreJsonStr(
+                Keystore::from_json_file(&validators_file_path).map_err(|e| format!("{e:?}"))?,
+            ),
+            voting_keystore_password: password.ok_or_else(|| {
+                "The --password flag is required to supply the keystore password".to_string()
```
chong-he: I have to put `ok_or_else` here to handle the `None` case (when `--password` flag) is not provided. We already have [requires here](https://github.com/sigp/lighthouse/pull/6261/files#diff-e54fe894fb02bdd0c2175fbb4d7cc6d04fa7c8b460b30e9512edb814f569f97eR154), so not sure about this here,  or under what condition will this manifest?

validator_manager/src/import_validator.rs
chong-he: Merged the 2 files in [0e37ad8](https://github.com/sigp/lighthouse/pull/6261/commits/0e37ad8bc86fa820b5ec4a14e8645a577efc7d5e)

chong-he (commented):
Thank you for your review and help @macladson and @michaelsproul. I added some comments to some parts that I am not sure. The `import` command now supports importing keystore JSON files created using:

- `lighthouse vm create` by using the command `lighthouse vm import`
-  staking-deposit-cli or ethstaker-deposit-cli using `lighthouse vm import --standard-format` 

@macladson  mentioned about supporting multiple validators with the standard format, I will look into this; in the mean time, this is ready for a re-review.

Thanks!

lighthouse/tests/validator_manager.rs
macladson: The reason is because of [here](https://github.com/sigp/lighthouse/pull/6261/files#diff-e54fe894fb02bdd0c2175fbb4d7cc6d04fa7c8b460b30e9512edb814f569f97eR186).

We set it in the config directly as `Some(get_flag)` so in the case the flag is missing it will be set to `Some(false)`.
Unless I'm missing something, I think this should probably just be a `parse_optional` like the other fields in the config.

validator_manager/src/import_validators.rs
```diff
@@ -80,14 +83,94 @@ pub fn cli_app() -> Command {
                 )
                 .display_order(0),
         )
+        .arg(
+            Arg::new(PASSWORD)
```
macladson, line 100: Is the `--password` flag only valid when using `--standard-format`? If that's the case, we should add a `.requires(STANDARD_FORMAT)` to its cli definition.  This is kind of circular but I think it should be fine. Without this, a user would be able to use the `--password` flag when using the Lighthouse format which would be silently ignored.

validator_manager/src/import_validators.rs
macladson: Because we want to support both formats, we have to have `password` be an `Option`. Then since we want to build the `ImportConfig` before we know which format we are using, we are stuck with password being an `Option`.

The `or_else` should never trigger since the `require` will trigger before we even get here, but this pattern is fine as is imo. We have redundant patterns like this in plenty of places.

chong-he: Thank you for the explanation

validator_manager/src/import_validators.rs
chong-he: Added in [7d9346a](https://github.com/sigp/lighthouse/pull/6261/commits/7d9346ab2d28d2dc1d999e3e4a905a44105493aa)

lighthouse/tests/validator_manager.rs
chong-he: Thank you so much for your help, fix in [b71d0e4](https://github.com/sigp/lighthouse/pull/6261/commits/b71d0e4db7f681ecf9f5433c87ec24b39ecee9bb)

book/src/validator-manager-api.md
```diff
@@ -0,0 +1,39 @@
+# Managing Validators
+
+The `lighthouse validator-manager` supports the [Keymanager API](https://ethereum.github.io/keymanager-APIs/#/) to list, import and delete keystores via the HTTP API. This requires the validator client running with the flag `--http`.
```
michaelsproul:
```suggestion
The `lighthouse validator-manager` uses the [Keymanager API](https://ethereum.github.io/keymanager-APIs/#/) to list, import and delete keystores via the HTTP API. This requires the validator client running with the flag `--http`.
```

book/src/validator-manager-api.md
```diff
@@ -0,0 +1,39 @@
...
+The `delete` command deletes one or more validators from the validator client. It will also modify the `validator_definitions.yml` file automatically so there is no manual action required from the user after the delete. To `delete`:
+
+```bash
+lighthouse vm delete --vc-token <API-TOKEN-PATH> --validators pubkey1,pubkey2
+```
+
+Example:
+
+```bash
+lighthouse vm remove --vc-token ~/.lighthouse/mainnet/validators/api-token.txt --validators 0x8885c29b8f88ee9b9a37b480fd4384fed74bda33d85bc8171a904847e65688b6c9bb4362d6597fd30109fb2def6c3ae4,0xa262dae3dcd2b2e280af534effa16bedb27c06f2959e114d53bd2a248ca324a018dc73179899a066149471a94a1bc92f
```
michaelsproul:
```suggestion
lighthouse vm delete --vc-token ~/.lighthouse/mainnet/validators/api-token.txt --validators 0x8885c29b8f88ee9b9a37b480fd4384fed74bda33d85bc8171a904847e65688b6c9bb4362d6597fd30109fb2def6c3ae4,0xa262dae3dcd2b2e280af534effa16bedb27c06f2959e114d53bd2a248ca324a018dc73179899a066149471a94a1bc92f
```

book/src/validator-manager-api.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+## Import
+
+The `import` command imports validator keystores generated by the staking-deposit-cli/ethstaker-deposit-cli. To import validator keystore:
```
michaelsproul:
```suggestion
The `import` command imports validator keystores generated by the staking-deposit-cli/ethstaker-deposit-cli. To import a validator keystore:
```

book/src/validator-manager-api.md
```diff
@@ -0,0 +1,39 @@
...
+
+```bash
+lighthouse vm import --standard-format --vc-token <API-TOKEN-PATH> --validators-file /path/to/json --password keystore_password
+```
+
+Example:
+
+```
+lighthouse vm import --standard-format --vc-token ~/.lighthouse/mainnet/validators/api-token.txt --validator-file keystore.json --password keystore_password
```
michaelsproul:
```suggestion
lighthouse vm import --standard-format --vc-token ~/.lighthouse/mainnet/validators/api-token.txt --validators-file keystore.json --password keystore_password
```

book/src/validator-manager-create.md
```diff
@@ -69,6 +69,8 @@ lighthouse \
 > Be sure to remove `./validators.json` after the import is successful since it
 > contains unencrypted validator keystores.
 
+> Note: To import validators with validator-manager using keystore files created using the staking deposit cli, refer to [Managing Validators](./validator-manager-api.md#import).
```
michaelsproul:
```suggestion
> Note: To import validators with validator-manager using keystore files created using the staking deposit CLI, refer to [Managing Validators](./validator-manager-api.md#import).
```

validator_manager/src/import_validators.rs
```diff
@@ -127,13 +232,31 @@ async fn run<'a>(config: ImportConfig) -> Result<(), String> {
         .create(false)
         .open(&validators_file_path)
         .map_err(|e| format!("Unable to open {:?}: {:?}", validators_file_path, e))?;
-    let validators: Vec<ValidatorSpecification> = serde_json::from_reader(&validators_file)
-        .map_err(|e| {
+
+    let validators: Vec<ValidatorSpecification> = if standard_format {
```
michaelsproul: Rather than using an extra `--standard-format` flag, I wonder if we should use a different flag name like `--keystore-file` for keystores in the standard format.

The reason being that the `--validators-file` name isn't really accurate for a single keystore, and the `validators.json` file generated by Lighthouse is a very different format to the standard format.

We could have `--keystore-file` and `--validators-file` _conflict_ so that the user is only able to set 1.

validator_manager/src/list_validators.rs
```diff
@@ -0,0 +1,203 @@
...
+    Command::new(CMD)
+        .about("Lists all validators in a validator client using the HTTP API.")
+        .arg(
+            Arg::new(VC_URL_FLAG)
+                .long(VC_URL_FLAG)
+                .value_name("HTTP_ADDRESS")
+                .help(
+                    "A HTTP(S) address of a validator client using the keymanager-API. \
+            If this value is not supplied then a 'dry run' will be conducted where \
+            no changes are made to the validator client.",
```
michaelsproul: Is this doc accurate? It looks like it was copy-pasted from the `--confirm` flag

validator_manager/src/list_validators.rs
```diff
@@ -0,0 +1,203 @@
...
+    let mut result = String::new();
+
+    result.push_str(format!("List of validators ({}):", validators.len()).as_str());
+    result.push('\n');
+    for validator in validators {
+        result.push_str(format!("{}", validator.validating_pubkey).as_str());
+        result.push('\n');
+    }
+
+    Ok(result)
```
michaelsproul: It's a bit strange that we return a string here, but never do anything with it other than `string.contains(pubkey)`. We could return the `validators` which is a `Vec<SingleKeystoreResponse>` instead, which is cleaner and simpler.

michaelsproul (changes requested):
Looking good!

Just had a few suggestions

validator_manager/src/list_validators.rs
chong-he: I removed this description in [e5a618c](https://github.com/sigp/lighthouse/pull/6261/commits/e5a618c24f0dc9b5f93f6d0d54cb0004197cb012) for all `import`, `list`, and `delete`. Because from testing, without supplying `vc-url` flag, it still able to make changes to the VC, using the default VC HTTP url.

validator_manager/src/list_validators.rs
chong-he: This is revised in [c652724](https://github.com/sigp/lighthouse/pull/6261/commits/c652724410021ed638620f6e2cadda3fc9263028), thank you for your help! 

I think I understand what you mean, now that we return Vec<SingleKeystoreResponse> from the `run`  function, which we can obtain the `validating_pubkey` from one of the field in the struct, which is more natural. We can then use that to compare with local public keys. 

Feel free to point out anything that's still weird or incorrect

validator_manager/src/import_validators.rs
chong-he: Revised in [aa110d5](https://github.com/sigp/lighthouse/pull/6261/commits/aa110d5dfcc16fa2d9b87eb1d5d82c4b1ade2c8c), that are some questions that I am a little unsure which I will highlight them in a "self-review" later

validator_manager/src/import_validators.rs
```diff
@@ -217,43 +220,57 @@ async fn run<'a>(config: ImportConfig) -> Result<(), String> {
...
-    } else {
-        serde_json::from_reader(&validators_file).map_err(|e| {
-            format!(
-                "Unable to parse JSON in {:?}: {:?}",
-                validators_file_path, e
-            )
-        })?
-    };
+    let validators: Vec<ValidatorSpecification> =
+        if let Some(validators_format_path) = &validators_file_path {
```
chong-he, line 228: As `validators-file` and `keystore-file` are now only either one that will be used, these flags become optional now. So I use the `Some()` to "extract" the file path so that I can use the value inside the `Option<PathBuf>`. 

Feel free to point out if this is not the correct/standard way to do this.

The same goes for `keystore_file_path` below.

lighthouse/tests/validator_manager.rs
```diff
@@ -233,9 +249,11 @@ pub fn validator_import_missing_token() {
 }
 
 #[test]
-pub fn validator_import_missing_validators_file() {
+pub fn validator_import_using_both_file_flags() {
     CommandLineTest::validators_import()
         .flag("--vc-token", Some("./token.json"))
+        .flag("--validators-file", Some("./vals.json"))
+        .flag("--keystore-file", Some("./keystore.json"))
```
chong-he, line 270: This test is failing: https://github.com/sigp/lighthouse/actions/runs/11404357702/job/31733366200

I found that this is because the `.required(true)` for the `validators-file` flag is removed: https://github.com/sigp/lighthouse/pull/6261/files#diff-e54fe894fb02bdd0c2175fbb4d7cc6d04fa7c8b460b30e9512edb814f569f97eL41

(as it now requires either `--validators-file` or `--keystore-file`)

The test caught it because now that the hard requirement for `--validators-file` is dropped, so missing the validators file flag will not return a `false` from the test. I change this test to testing that if both flags are used, then the test fails. 

Not sure if this is good?

book/src/validator-manager-api.md
```diff
@@ -0,0 +1,39 @@
...
+The `import` command imports validator keystores generated by the staking-deposit-cli/ethstaker-deposit-cli. To import a validator keystore:
+
+```bash
+lighthouse vm import --vc-token <API-TOKEN-PATH> --keystore-file /path/to/json --password keystore_password
+```
+
+Example:
+
+```
+lighthouse vm import --standard-format --vc-token ~/.lighthouse/mainnet/validators/api-token.txt --validators-file keystore.json --password keystore_password
```
michaelsproul:
```suggestion
lighthouse vm import --vc-token ~/.lighthouse/mainnet/validators/api-token.txt --keystyore-file keystore.json --password keystore_password
```

michaelsproul (approved):

lighthouse/tests/validator_manager.rs
michaelsproul: As discussed offline, this is good now!

validator_manager/src/import_validators.rs
michaelsproul: Yep, that's a good approach

book/src/validator-manager-api.md
chong-he: ~~This is strange, I thought I changed this here:
https://github.com/sigp/lighthouse/pull/6261/commits/aa110d5dfcc16fa2d9b87eb1d5d82c4b1ade2c8c#diff-43a661250c5b435ce4306e19cd5fc643c93124887869ad73d79ae7f6a02a1557~~

Edit: oh I didn't change the example

michaelsproul:
@mergify queue

PR #6266, opened 2024-08-15
Fix rolling file appender bug

### LINKED ISSUE #6246: "Failed to initialize libp2p rolling file" when `--logfile` is specified
## Description

A discord user reported getting `Failed to initialize libp2p rolling file appender` on startup on his beacon nodes for several versions now.

The user uses the `--logfile` flag, which is not only used as the lighthouse log _file_ path, but also used as a _directory_ path for libp2p logs.

This causes the libp2p `RollingFileAppender` to fail to build, because the file already exist and directory cannot be created.

The code below assumes that `logfile` is a directory, and is used as the directory path for libp2p logging:

https://github.com/sigp/lighthouse/blob/6856134ded14d8986c4ddd40abd3b26f87f09924/lighthouse/src/main.rs#L618-L630

This was a regression introduced in #4988 (v4.6.0) and impacts users running beacon node with the `--logfile` flag.

## Version

v5.3.0

## Present Behaviour

Error log `Failed to initialize libp2p rolling file appender` and no libp2p and discv5 created.

The only workaround now is to avoid using `--logfile` and have the logs stored under the data directory.

## Expected Behaviour

No error and libp2p & discv5 logs present in the specified log location.
How _should_ the application behave?

## Steps to resolve

Please describe the steps required to resolve this issue, if known.

eserilev:
## Issue Addressed

Closes #6246

## Proposed Changes

Ensures that the file path passed to the rolling file appender only contains directories. This PR iterates through each piece of `--logfile` and if its a dir, appends it to the file path we use for the rolling file appender. If all or parts of `--logfile` dont exist yet, we just use the full `--logfile` path.


common/logging/src/lib.rs
```diff
@@ -220,6 +221,25 @@ impl TimeLatch {
...
+                }
+            }
+            // An error here just means that part of `base_tracing_log_path` doesn't exist,
+            // we can safely continue.
+            Err(_) => {
+                tracing_log_path = base_tracing_log_path;
+                break;
+            }
+        }
+    }
```
ackintosh: I noticed that the `for` loop unexpectedly breaks in the `Err(_)` statement when running Lighthouse with the `--logfile /Users/akihito/logs/beacon.log` parameter. This happens because the strings passed to `match metadata(p)` are [`/`, `Users`, `akihito`, `logs`, `beacon.log`], and the second string, `Users` (not `/Users`), doesn't exist in my working directory.

I think we can fix this by making the following change:

```diff
    for p in base_tracing_log_path.iter() {
+       tracing_log_path = tracing_log_path.join(p);
+       match tracing_log_path.metadata() {
            Ok(metadata) => {
+               if !metadata.is_dir() {
+                   tracing_log_path.pop();
+                   break;
                }
            }
```

ackintosh (commented):
Thanks for addressing the issue!

common/logging/src/lib.rs
eserilev: thanks for catching that issue! i've pushed up a fix here: [f97e0d4](https://github.com/sigp/lighthouse/pull/6266/commits/f97e0d442ff7647586a6705cd8ab698b5feeb602)

havent had a chance to test it yet, will get to that soon

ackintosh:
@eserilev I have filed a PR to your branch, I would appreciate it if you could have a look. 🙏  https://github.com/eserilev/lighthouse/pull/11

eserilev:
thanks @ackintosh ive merged your changes

ackintosh (approved):
I’ve tested it again, and everything was fine.

michaelsproul (approved):
Looks good

I'm a bit worried that the silent popping of path components could cause strange behaviour in future, but this seems like a practical solution for now.

I'm also slightly concerned that if the log file initialisation were reordered we could regress here. E.g. if the logfile doesn't exist before we check its metadata, then we won't pop any path fragments, and then we'll fail in the same way as before.

jimmygchen:
@mergify queue

PR #6267, opened 2024-08-16
Make `BeaconChain::kzg` field mandatory

### LINKED ISSUE #6260: Make `BeaconChain::kzg` mandatory
## Description

The `kzg` field of the `BeaconChain` is unnecessarily wrapped in an `Option`.

https://github.com/sigp/lighthouse/blob/d6ba8c397557f5c977b70f0d822a9228e98ca214/beacon_node/beacon_chain/src/beacon_chain.rs#L493-L494

Removing this would simplify some error handling because we don't need to handle the case where the `kzg` setup is not initialized.

## Version

Lighthouse v5.3.0

eserilev:
## Issue Addressed

Closes #6260

## Proposed Changes

Make the kzg field in `BeaconChain` and `BeaconChainBuilder` mandatory


beacon_node/client/src/builder.rs
```diff
@@ -195,7 +195,22 @@ where
...
+            if spec.is_peer_das_scheduled() {
+                Arc::new(
+                    Kzg::new_from_trusted_setup_das_enabled(trusted_setup).map_err(kzg_err_msg)?,
+                )
+            } else {
+                Arc::new(Kzg::new_from_trusted_setup(trusted_setup).map_err(kzg_err_msg)?)
+            }
+        } else {
+            // TODO(kzg) this causes us failed tests i.e. http_server_genesis_state
+            panic!("Failed to load KZG");
```
eserilev: this panic causes some of our tests to fail i.e. `http_server_genesis_state`. I'm assuming a trusted setup isnt available pre-deneb. So for tests that run a pre-deneb network, is there some default KZG value we want to pass in here?

michaelsproul: I think we should try loading the KZG setup regardless of whether the network is pre-Deneb (if possible).

dapplion: Wouldn't that slow things significantly? Post PeerDAS loading KZG is slow given the precomputation optimization

michaelsproul: Oh. Hmm. Maybe we need to find a way to only load it once in tests, as otherwise we'll be loading it 1000s of times, even if we just load it for peer DAS tests?

kevaundray: A new feature in 0.5.0 allows a user to specify whether they want precomputation or not: https://github.com/crate-crypto/rust-eth-kzg/pull/161

Its under breaking changes: `Allow downstream users to avoid the precomputations that are needed for the FixedBaseMSM algorithm in FK20` since it adds a new parameter that was not there before.

I can push another release today/tomorrow if that would help -- Noting that if you want the same computing and verifying speeds without precomputation, you can increase the number of threads you initialize the library with to something like 4 instead of 1

eserilev: thanks Kev! That new feature sounds like exactly what we need here. I'll try it out and report back if I'm having issues

kevaundray: 0.5 includes a few API changes, so it might be better if I made the PR so as to not waste your time, if this blocking for this PR :)

kevaundray: I've opened #6309 and added you as a co-author to the commits!

dapplion:
This PR adds a new trusted setup, is that expected?

eserilev:
removed the extraneous trusted setup. It looks like loading KZG has added like 5 min to a few CI test groups. I'm going to dig in there and make sure were not doing any KZG precomputation when possible. Kevs recent PR makes this easy to do

eserilev:
There no longer seems to be any noticeable difference in CI times between this branch and unstable

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -5682,10 +5682,8 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
 
                 let kzg_proofs = Vec::from(proofs);
 
-                let kzg = self
-                    .kzg
-                    .as_ref()
-                    .ok_or(BlockProductionError::TrustedSetupNotInitialized)?;
```
michaelsproul, line 5688: We can delete this error

beacon_node/beacon_chain/benches/benches.rs
```diff
@@ -35,10 +35,11 @@ fn all_benches(c: &mut Criterion) {
...
 
-    let trusted_setup: TrustedSetup = serde_json::from_reader(TRUSTED_SETUP_BYTES)
-        .map_err(|e| format!("Unable to read trusted setup file: {}", e))
-        .expect("should have trusted setup");
-    let kzg = Arc::new(Kzg::new_from_trusted_setup(trusted_setup).expect("should create kzg"));
+    let kzg = if spec.deneb_fork_epoch.is_some() {
+        KZG.clone()
+    } else {
+        KZG_NO_PRECOMP.clone()
+    };
```
michaelsproul: maybe rather than duplicating this logic, which will need to be tweaked when we enable Peer DAS in more tests, we could define a function in `test_utils` that takes the `ChainSpec` and returns an `Arc<Kzg>` based on the values of `deneb_fork_epoch` and `eip7594_fork_epoch`?

beacon_node/beacon_chain/src/builder.rs
```diff
@@ -1204,7 +1199,13 @@ mod test {
         let (shutdown_tx, _) = futures::channel::mpsc::channel(1);
         let runtime = TestRuntime::default();
 
-        let chain = Builder::new(MinimalEthSpec)
+        let kzg = if spec.deneb_fork_epoch.is_some() {
+            KZG.clone()
+        } else {
+            KZG_NO_PRECOMP.clone()
+        };
```
michaelsproul: can use the same new method here

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -382,11 +376,7 @@ pub fn validate_data_column_sidecar_for_gossip<T: BeaconChainTypes>(
     let parent_block = verify_parent_block_and_finalized_descendant(data_column.clone(), chain)?;
     verify_slot_higher_than_parent(&parent_block, column_slot)?;
     verify_proposer_and_signature(&data_column, &parent_block, chain)?;
-
-    let kzg = chain
-        .kzg
-        .clone()
-        .ok_or(GossipDataColumnError::KzgNotInitialized)?;
+    let kzg = chain.kzg.clone();
```
michaelsproul: Not really an issue, but I don't think this clone is necessary

beacon_node/beacon_chain/src/test_utils.rs
```diff
@@ -522,12 +530,17 @@ where
         let validator_keypairs = self
             .validator_keypairs
             .expect("cannot build without validator keypairs");
-        let kzg = spec.deneb_fork_epoch.map(|_| KZG.clone());
+
+        let kzg = if spec.deneb_fork_epoch.is_some() {
+            KZG.clone()
+        } else {
+            KZG_NO_PRECOMP.clone()
+        };
```
michaelsproul: can use the same func here

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -299,7 +299,7 @@ async fn light_client_updates_test() {
         .unwrap()
         .unwrap();
 
-    let kzg = spec.deneb_fork_epoch.map(|_| KZG.clone());
+    let kzg = KZG_NO_PRECOMP.clone();
```
michaelsproul: same func here

not sure if using pre-comp here will have a perf impact, was this intentional?

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -2680,7 +2679,8 @@ async fn weak_subjectivity_sync_test(slots: Vec<Slot>, checkpoint_slot: Slot) {
     let store = get_store(&temp2);
     let spec = test_spec::<E>();
     let seconds_per_slot = spec.seconds_per_slot;
-    let kzg = spec.deneb_fork_epoch.map(|_| KZG.clone());
+
+    let kzg = KZG_NO_PRECOMP.clone();
```
michaelsproul: same as above

beacon_node/network/src/subnet_service/tests/mod.rs
```diff
@@ -45,12 +46,18 @@ impl TestBeaconChain {
         let store =
             HotColdDB::open_ephemeral(StoreConfig::default(), spec.clone(), log.clone()).unwrap();
 
+        let kzg = if spec.deneb_fork_epoch.is_some() {
+            KZG.clone()
+        } else {
+            KZG_NO_PRECOMP.clone()
+        };
```
michaelsproul: func

crypto/kzg/src/lib.rs
```diff
@@ -51,18 +51,41 @@ impl From<c_kzg::Error> for Error {
 #[derive(Debug)]
 pub struct Kzg {
     trusted_setup: KzgSettings,
-    context: Option<DASContext>,
+    context: DASContext,
```
michaelsproul, line 54: I guess we could keep this DASContext optional for the pre-peerDAS case, but then again if there's no perf impact maybe it's OK to keep it initialized

crypto/kzg/src/trusted_setup.rs
```diff
@@ -30,6 +36,20 @@ pub struct TrustedSetup {
...
+impl Default for TrustedSetup {
+    fn default() -> Self {
+        if let Ok(trusted_setup) = serde_json::from_slice(get_trusted_setup().as_slice()) {
+            trusted_setup
+        } else {
+            Self {
+                g1_monomial_points: Default::default(),
+                g1_points: Default::default(),
+                g2_points: Default::default(),
+            }
```
michaelsproul: This default impl feels a little sketchy, as I imagine the inner `default` impls here use some junk values?

I think we could remove the need for this by making `client::Config::trusted_setup` either:

- a `Vec<u8>` which is parsed on node startup, or
- an `Option<TrustedSetup>`, which we error out on at startup if it is `None`

michaelsproul (changes requested):

crypto/kzg/src/trusted_setup.rs
eserilev: The only pace the default `TrustedSetup` is used is in the default `client::Config` impl. I've moved the trusted setup parsing into the `Config` default impl, which will error out if the trusted_setup file isnt found. You think that should be ok? We could alternatively change  `client::Config::trusted_setup` to optional or a Vec<u8> as you said

crypto/kzg/src/lib.rs
eserilev: there was no noticeable performance impact that I noticed (test-wise at least)

michaelsproul (approved):

crypto/kzg/src/trusted_setup.rs
michaelsproul: yeah that seems good, I would rather have the expect

michaelsproul:
@mergify queue

beacon_node/beacon_chain/src/test_utils.rs
```diff
@@ -75,22 +75,40 @@ pub const FORK_NAME_ENV_VAR: &str = "FORK_NAME";
...
+});
+
+pub fn get_kzg(spec: &ChainSpec) -> Arc<Kzg> {
+    if spec.eip7594_fork_epoch.is_some() {
+        KZG_PEERDAS.clone()
+    } else if spec.deneb_fork_epoch.is_some() {
+        KZG.clone()
+    } else {
+        KZG_NO_PRECOMP.clone()
+    }
```
jimmygchen, line 109: The `KZG` precomputation is only used for PeerDAS.
I think we can remove this branch  and default to `KZG` if PeerDAS is not enabled?

beacon_node/client/src/builder.rs
```diff
@@ -195,7 +195,17 @@ where
...
 
-        let builder = BeaconChainBuilder::new(eth_spec_instance)
+        let kzg_err_msg = |e| format!("Failed to load trusted setup: {:?}", e);
+        let trusted_setup = config.trusted_setup.clone();
+        let kzg = if spec.is_peer_das_scheduled() {
+            Kzg::new_from_trusted_setup_das_enabled(trusted_setup).map_err(kzg_err_msg)?
+        } else if spec.deneb_fork_epoch.is_some() {
+            Kzg::new_from_trusted_setup(trusted_setup).map_err(kzg_err_msg)?
+        } else {
+            Kzg::new_from_trusted_setup_no_precomp(trusted_setup).map_err(kzg_err_msg)?
```
jimmygchen, line 205: I think we can get rid of this `new_from_trusted_setup_no_precomp` branch too?

crypto/kzg/src/lib.rs
jimmygchen: This was changed to optional because it adds 550ms to startup time and 110 mb extra memory usage for an unused feature on mainnet:
https://github.com/sigp/lighthouse/pull/6212#discussion_r1712525802

It may also impact our CI time as I imagine Kzg is loaded many times.

A profile run on our basic-sim test show that 88% of CPU time is spent on loading `DASContext` during startup 
image

I think we should make this optional until PeerDAS is scheduled.

crypto/kzg/src/lib.rs
```diff
@@ -51,18 +51,41 @@ impl From<c_kzg::Error> for Error {
...
+                &trusted_setup.g1_points(),
+                &trusted_setup.g2_points(),
+            )?,
+            context,
+        })
+    }
+
     /// Load the kzg trusted setup parameters from a vec of G1 and G2 points.
     pub fn new_from_trusted_setup(trusted_setup: TrustedSetup) -> Result<Self, Error> {
+        let peerdas_trusted_setup = PeerDASTrustedSetup::from(&trusted_setup);
```
jimmygchen, line 74: This function is now identical to `new_from_trusted_setup_das_enabled`.
I think maybe what we want is
- if PeerDAS scheduled, load it with DASContext
- Else load with no DASContext OR no precomputation (although I prefer the former as initilisation of `PeerDASTrustedSetup` still require cloning trusted setup bytes.

PR #6268, opened 2024-08-16
Improving blob propagation post-PeerDAS with Decentralized Blob Building

jimmygchen:
## Issue Addressed

Built on top of #5829, an optimization came up by @michaelsproul, to fetch blobs from the EL to reduce the delay to block import.

This PR goes further to publish the blobs to the network, which helps improve the resiliency of the network, by having nodes with more resources contribute to blob propagation. This experimental solution is an attempt to solve the self building proposer bandwidth issue discussed on [R&D Discord](https://discord.com/channels/595666850260713488/1247156108263555185/1247156110146928640) and described in @dankrad's [post here](https://notes.ethereum.org/@dankrad/self-builder-strategies).

The benefits of this proposals are:
- **Reduces block import latency**: nodes can retrieve blobs from EL without waiting for them from gossip, hence making blocks attestable earlier.
- **Improves blob propagation and network resiliency**: blob propagation work from is spread out from 1 node to the entire network, which reduces the likelihood of missed block due to delays in propagtion.
- **Allows scaling without sacrificing decentralization**: nodes with more resources will participate in blob building and propagation, allowing nodes with limited bandwidth to continue to produce block post-PeerDAS.

## Proposed Changes

- **Deneb**: `fetch_blobs_and_publish` is triggered after a node has processed a gossip / rpc block and is still missing blob components. Once the node fetches the blob from EL, it then publishes the remaining blobs that hasn't seen on gossip to the network.
- **PeerDAS**: Same trigger as above, however only supernodes will publish data columns that are unseen on gossip to the network. 

## Next steps:

- To maintain low bandwidth for smaller stakers (single validator BN), we could allow some optimisation on block publish behaviour for these nodes only. There are [some strategies](https://discord.com/channels/595666850260713488/1247156108263555185/1247529154249756817) proposed by @cskiraly to bring the outbound bandwidth requirements for a 32 blobs block to the same level as Deneb (6 blobs). However this wouldn't be recommended for nodes with enough bandwidth.
- Collect some realistic metrics for a network with 32 blobs per block.

## Challenges:

- Current KZG libraries (`c-kzg-4844` and `rust-eth-kzg`) _may_ struggle with constructing large number of cells and proofs at once due to the current memory allocation approach.
- Even if we are able reduce the bandwidth usage on the CL side, the bandwidth challenge remains on the EL side, as the node still need to pull the blob transactions into its mempool, to a lesser extent though, because:
  - it's dealing with raw blobs (4096kb for 32 blobs) rather than erasure coded blobs
  - it's pulled-based ([`eth/68`](https://eips.ethereum.org/EIPS/eip-5793)) hence doesn't incur the same gossip amplification cost (8x) on the CL.

## TODO before merging

- [x] Wait for spec to be agreed on by EL clients
  - https://github.com/ethereum/consensus-specs/pull/3864
  - https://github.com/ethereum/execution-apis/pull/559

## Reference:

- @michaelsproul's original PR targeting `unstable` https://github.com/sigp/lighthouse/pull/5829
- R&D Discord [discussion thread](https://discord.com/channels/595666850260713488/1247156108263555185/1247156110146928640) 
- [How to help self builders with blobs ](https://notes.ethereum.org/@dankrad/self-builder-strategies) by @dankrad
- [Fetch blobs from EL pool](https://hackmd.io/@dapplion/blob_fetch) by @dapplion


beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3131,6 +3132,39 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+            .fork_choice_read_lock()
+            .contains_block(&block_root)
+        {
+            return Err(BlockError::BlockIsAlreadyKnown(block_root));
+        }
+
+        if let Some(event_handler) = self.event_handler.as_ref() {
+            if event_handler.has_blob_sidecar_subscribers() {
+                for blob in blobs.iter().filter_map(|maybe_blob| maybe_blob.as_ref()) {
+                    event_handler.register(EventKind::BlobSidecar(
```
dapplion, line 3124: This may double-emit an event for a column that has previously been imported from gossip?

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3833,6 +3890,10 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         // end up with blocks in fork choice that are missing from disk.
         // See https://github.com/sigp/lighthouse/issues/2028
         let (_, signed_block, blobs, data_columns) = signed_block.deconstruct();
+        let custody_columns_count = self.data_availability_checker.get_custody_columns_count();
+        let data_columns = data_columns
+            .filter(|a| a.len() >= custody_columns_count)
+            .or_else(|| data_column_recv?.blocking_recv());
```
dapplion, line 3896: Add a nice comment here explaining the mechanism of `data_column_recv` and why trigger it here under these conditions

dapplion: Add a nice comment here explaining the mechanism of `data_column_recv` and why trigger it here under these conditions

dapplion: We should be more explicit here if the builder routine failed. If `blocking_recv` returns None because the sender was dropped we should log an error and track metrics here also. Otherwise it's mixing the signal of data_columns = None because before PeerDAS with data_columns = None on reconstruction error.

Also add a timer histogram around `blocking_recv` to check how much we are blocking this routine.

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -0,0 +1,215 @@
+use crate::observed_data_sidecars::ObservableDataSidecar;
```
dapplion: Add module docs about this strategy or link to an external resource explaining it

consensus/types/src/blob_sidecar.rs
```diff
@@ -149,6 +149,37 @@ impl<E: EthSpec> BlobSidecar<E> {
         })
     }
 
+    pub fn new_efficiently(
```
dapplion: Maybe a better name `new_with_existing_proof`

jimmygchen:
Some early testing results:
- Proposers withholding all blobs propose blocks with blobs with 100% success rate
- No outbound bandwidth spike for the full nodes with limited upload

Bandwidth-limited fullnode (`cl-01`) vs supernode (`cl-02`): 

image
(Thanks to @KatyaRyazantseva for the dashboard above☺️ )

This shows EL inbound traffic (fetch blos from peers) isn't too bad for MAX 6 blobs
The outbound traffic for EL is less relevant here because it includes sending blobs to CL.

image

Next steps:
- Add more metrics
    - Blocks made available via EL blobs
    - Number of blobs / data columns from EL blobs published
    - EL blob fetch timing
    - Compute cells and proof time
- Make `MAX_BLOBS_PER_BLOCK` configurable
- Try 32 blobs per block
    - EL gas constant update
    - Potential update on derived configs
    - Potential batching of KZG computation to avoid overflow
- Add `--limit-blob-publish` (single validator only), which allows for lower mesh peers for data columns topics and withholding certain amount of data columns

jimmygchen:
This was originally intended experimental but it's been pretty stable, 800 epochs on devnet and 100% participation, and I think we can merge this.
I'll address the review comments, thanks @dapplion for the review 🙏

consensus/types/src/blob_sidecar.rs
jimmygchen: Updated in https://github.com/sigp/lighthouse/pull/6268/commits/dbf4bb8bfdbb3ca37d82b948958ef48670aa0882

beacon_node/beacon_chain/src/fetch_blobs.rs
jimmygchen: Updated in https://github.com/sigp/lighthouse/pull/6268/commits/dbf4bb8bfdbb3ca37d82b948958ef48670aa0882

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Good catch, fixed in https://github.com/sigp/lighthouse/pull/6268/commits/dbf4bb8bfdbb3ca37d82b948958ef48670aa0882

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Good catch, If the computation fails for some reasons, we probably want to handle it the same way as "database write failed" - restore fork choice and return `Error`, instead of letting the database commit to an inconsistent state. @michaelsproul thoughts?

jimmygchen: > Also add a timer histogram around blocking_recv to check how much we are blocking this routine

Metrics added to https://github.com/sigp/lighthouse/pull/6268/commits/278a0b90a97a6443adb4ec173adadac1ac4ec635

jimmygchen: Addressed error handling in https://github.com/sigp/lighthouse/pull/6268/commits/d51ad5f21cb77be2fafbcf4ed7203a8ae4367906.

dapplion:
> Add --limit-blob-publish (single validator only), which allows for lower mesh peers for data columns topics and withholding certain amount of data columns

Bring the attack to prod

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -0,0 +1,226 @@
...
+                        "i" => i
+                    );
+                }
+            }
+            Err(e) => {
+                warn!(
+                    chain.log,
+                    "Blobs from EL - error";
+                    "error" => ?e
+                );
```
dapplion: Why not error out in these two statements? Should not happen and make the code below harder to reason about

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -0,0 +1,226 @@
...
+    if num_fetched_blobs == 0 {
+        debug!(chain.log, "Blobs from EL - response with none");
+        return Ok(());
+    } else if num_fetched_blobs < num_blobs {
+        debug!(
+            chain.log,
+            "Blobs from EL - response with some";
+            "fetched" => num_fetched_blobs,
+            "total" => num_blobs,
+        );
```
dapplion: I feel this log should be moved below on the if

```
if peer_das_enabled && all_blobs_fetched {
  .. cache hit
} else if peer_das_enabled && !all_blobs_fetched {
  .. cache miss
```

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -0,0 +1,226 @@
...
+                            return;
+                        }
+                    };
+
+                    // Check indices from cache before sending the columns, to make sure we don't
+                    // publish components already seen on gossip.
+                    let all_data_columns_iter = all_data_columns.clone().into_iter();
+                    let data_columns_to_publish = match chain_cloned.data_availability_checker.imported_custody_column_indexes(&block_root) {
+                        None => Either::Left(all_data_columns_iter),
+                        Some(imported_columns_indices) => Either::Right(all_data_columns_iter.filter(move |d| !imported_columns_indices.contains(&d.index()))),
```
dapplion: cargo fmt not working here?

beacon_node/beacon_chain/src/kzg_utils.rs
```diff
@@ -147,15 +147,16 @@ mod test {
...
         let spec = E::default_spec();
-        let (signed_block, blob_sidecars) = create_test_block_and_blobs::<E>(num_of_blobs, &spec);
+        let (signed_block, blobs) = create_test_block_and_blobs::<E>(num_of_blobs, &spec);
 
         let trusted_setup: TrustedSetup = serde_json::from_reader(TRUSTED_SETUP_BYTES)
             .map_err(|e| format!("Unable to read trusted setup file: {}", e))
             .expect("should have trusted setup");
         let kzg = Kzg::new_from_trusted_setup(trusted_setup).expect("should create kzg");
 
+        let blob_refs = blobs.iter().collect::<Vec<_>>();
```
kevaundray, line 365: I guess it might make sense for kzg libraries to take in a `impl Iterator<Item = Blob>` to possibly avoid these allocations

kevaundray:
@jimmygchen Any issues with the kzg libraries?

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -0,0 +1,226 @@
...
+                    let is_supernode =
+                        chain_cloned.data_availability_checker.get_custody_columns_count() == spec.number_of_columns;
+                    if is_supernode && !data_columns_to_publish.is_empty() {
+                        publish_fn(BlobsOrDataColumns::DataColumns(data_columns_to_publish));
+                    }
+                },
+                "compute_data_columns",
+            )
+            .ok_or(BeaconChainError::RuntimeShutdown)?;
+    } else {
```
dapplion: This else will attempt to publish blobs on peerdas

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3825,15 +3881,50 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+                        "msg" => "Restoring fork choice from disk",
+                        "error" => err_msg,
+                    );
+                    return Err(self
+                        .handle_import_block_db_write_error(fork_choice)
+                        .err()
+                        .unwrap_or(BlockError::InternalError(err_msg.to_string())));
+                }
+            }
+            // Non data columns present and compute data columns task was not spawned.
```
dapplion:
```suggestion
            // No data columns present and compute data columns task was not spawned.
```

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3825,15 +3881,50 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+                // return to avoid inconsistency in database.
+                if let Some(columns) = data_column_recv.blocking_recv() {
+                    Some(columns)
+                } else {
+                    let err_msg = "Did not receive data columns from sender";
+                    error!(
+                        self.log,
+                        "Failed to store data columns into the database";
+                        "msg" => "Restoring fork choice from disk",
+                        "error" => err_msg,
```
dapplion, line 3923: Just inline err_msg here

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3825,15 +3881,50 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+                    );
+                    return Err(self
+                        .handle_import_block_db_write_error(fork_choice)
+                        .err()
+                        .unwrap_or(BlockError::InternalError(err_msg.to_string())));
```
dapplion, line 3928: The err string here should be "Error restoring fork choice from disk after not receive data columns from sender"

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -0,0 +1,226 @@
...
+                        .filter_map(|b| b.as_ref().map(|b| &b.blob))
+                        .collect::<Vec<_>>();
+                    let data_columns_result = DataColumnSidecar::build_sidecars(
+                        &blob_refs,
+                        &block_cloned,
+                        &kzg,
+                        &spec,
+                    );
+
+                    let all_data_columns = match data_columns_result {
```
dapplion: Could inline data_columns_result

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -0,0 +1,226 @@
...
+        "Blobs from EL - start processing";
+        "num_blobs" => num_blobs,
+    );
+
+    chain
+        .process_engine_blobs(
+            block.slot(),
+            block_root,
+            fixed_blob_sidecar_list.clone(),
+            peer_das_enabled.then_some(data_columns_receiver),
```
dapplion: There's a mismatch here, `data_columns_sender` is only used if `peer_das_enabled && all_blobs_fetched` not only `peer_das_enabled`

jimmygchen:
> Bring the attack to prod

🙈  Trying to not mention the flag I use for testing. Luckily `git blame` on that line shows someone else's name haha

Just realized we need the `execution-apis` PR is [merged](https://github.com/ethereum/execution-apis/pull/559) before merge this - although there's probably no harm getting this one merged as soon as the inclusion is confirmed, as the `fetch_blobs` function handles this gracefully.

beacon_node/beacon_chain/src/fetch_blobs.rs
dapplion: You can move `let (data_columns_sender, data_columns_receiver) = tokio::sync::mpsc::channel(1);` to inside the if block and return the receiver from the if else

beacon_node/beacon_chain/src/fetch_blobs.rs
dapplion: Should add in this PR a metric to track cache miss / hit from EL when retrieving blobs

beacon_node/network/src/network_beacon_processor/mod.rs
```diff
@@ -775,6 +776,56 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+        let self_cloned = self.clone();
+        let log_cloned = self.log.clone();
+        let publish_fn = move |blobs_or_data_column| {
+            match &blobs_or_data_column {
+                BlobsOrDataColumns::Blobs(b) => {
+                    debug!(log_cloned, "Publishing blobs from EL"; "count" => b.len());
+                }
+                BlobsOrDataColumns::DataColumns(d) => {
+                    debug!(log_cloned, "Publishing data columns built from EL blobs"; "count" => d.len());
+                }
```
dapplion: You can move this log statements inside `publish_blobs_or_data_column` where `self.log` is already available

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -167,26 +168,33 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                    seen_timestamp,
+                    None,
+                    None,
+                );
 
-            self.chain.recompute_head_at_current_slot().await;
+                self.chain.recompute_head_at_current_slot().await;
+            }
+            Ok(AvailabilityProcessingStatus::MissingComponents(..)) => {
+                self.fetch_engine_blobs_and_publish(signed_beacon_block, block_root)
```
dapplion: Please add a detailed comment on why `fetch_engine_blobs_and_publish` is triggered here and only here

consensus/types/src/beacon_block_body.rs
```diff
@@ -178,57 +178,71 @@ impl<'a, E: EthSpec, Payload: AbstractExecPayload<E>> BeaconBlockBodyRef<'a, E,
         leaves
     }
 
-    /// Produces the proof of inclusion for a `KzgCommitment` in `self.blob_kzg_commitments`
-    /// at `index`.
+    /// Calculate a KZG commitment merkle proof slowly.
```
dapplion: I think slowly is misleading here, the other method just skips some of the computation and this method does the full computation

jimmygchen:
> @jimmygchen Any issues with the kzg libraries?

Nope all good so far! I was just flagging the potential challenges that I could imagine when we increase blob count, but I haven't actually run into any issues yet. Will keep you updated, thanks ☺️

jimmygchen:
`engine_getBlobsV1` now implemented in 
- Nethermind : https://github.com/NethermindEth/nethermind/pull/7322 (merged)
- Reth: https://github.com/paradigmxyz/reth/pull/9723 (open)

consensus/types/src/beacon_block_body.rs
michaelsproul: this was my flippant remark, I'll fix it

beacon_node/beacon_chain/src/fetch_blobs.rs
jimmygchen: Done

beacon_node/beacon_chain/src/fetch_blobs.rs
jimmygchen: Fixed

beacon_node/beacon_chain/src/fetch_blobs.rs
jimmygchen: Moved logs and added metrics below

beacon_node/beacon_chain/src/fetch_blobs.rs
jimmygchen: Fixed, the above error! macro formatting caused it to fail

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Done

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: This is for the case where fork choice is successfully restored, i.e. we get `Ok` from `handle_import`. We still want to fail the block import because we never computed the data columns.

If fork choice restore failed, we return that error instead.

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: See below

beacon_node/beacon_chain/src/fetch_blobs.rs
jimmygchen: Good catch, fixed

beacon_node/network/src/network_beacon_processor/mod.rs
jimmygchen: Done

beacon_node/network/src/network_beacon_processor/sync_methods.rs
jimmygchen: It's also triggered gossip as well, I'll add a comment
https://github.com/sigp/lighthouse/blob/662d6cf931eb2dca8df90f47924a13c2cc4a60c0/beacon_node/network/src/network_beacon_processor/gossip_methods.rs#L1519

jimmygchen: Done

beacon_node/beacon_chain/src/kzg_utils.rs
jimmygchen: @michaelsproul and i made an attempt today but it touches too much existing code, we might refactor this in a separate PR.

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -0,0 +1,282 @@
...
+    };
+
+    debug!(
+        chain.log,
+        "Blobs from EL - start processing";
+        "num_fetched_blobs" => num_fetched_blobs,
+    );
+
+    chain
+        .process_engine_blobs(
```
jimmygchen, line 189: just to make sure we don't forget - we should process the returned `AvailabilityProcessingStatus` and recompute head

jimmygchen: Fixed in https://github.com/sigp/lighthouse/pull/6268/commits/7977999b4a74db5fd9942d0463bcc642550feed2

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3551,11 +3604,12 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         self: &Arc<Self>,
         slot: Slot,
         availability: Availability<T::EthSpec>,
+        recv: Option<Receiver<DataColumnSidecarList<T::EthSpec>>>,
```
dapplion, line 3630: Can you add docs for this parameter?

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3807,15 +3863,50 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
-        let db_write_timer = metrics::start_timer(&metrics::BLOCK_PROCESSING_DB_WRITE);
-
         // Store the block and its state, and execute the confirmation batch for the intermediate
         // states, which will delete their temporary flags.
         // If the write fails, revert fork choice to the version from disk, else we can
         // end up with blocks in fork choice that are missing from disk.
         // See https://github.com/sigp/lighthouse/issues/2028
         let (_, signed_block, blobs, data_columns) = signed_block.deconstruct();
+        let custody_columns_count = self.data_availability_checker.get_custody_columns_count();
+        let data_columns = data_columns.filter(|columns| columns.len() >= custody_columns_count);
```
dapplion: Why is this filter necessary? Why would we have too many columns? If that is the case why are the first columns in the vec more important than the latter?

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -203,23 +198,14 @@ impl<E: EthSpec> PendingComponents<E> {
...
-                // No data columns when there are 0 blobs
-                self.num_expected_blobs()
-                    .map_or(false, |num_expected_blobs| {
-                        num_expected_blobs == 0
-                            || *num_expected_columns == num_received_data_columns
-                    })
-            }
-        }
+    pub fn is_available(&self, custody_column_count: usize) -> bool {
+        self.num_expected_blobs()
```
dapplion: Pendantic, but should rename to `block_commitments_count`? We technically don't expect blobs in peerdas

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -203,23 +198,14 @@ impl<E: EthSpec> PendingComponents<E> {
...
+            .map_or(false, |num_expected_blobs| {
+                let all_blobs_received = num_expected_blobs == self.num_received_blobs();
+                let all_columns_received = num_expected_blobs == 0
+                    || custody_column_count == self.num_received_data_columns();
```
dapplion: beacon_chain code seems to imply that we can have too many columns, should this be
```rust
self.num_received_data_columns() >= custody_column_count
```

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -434,46 +418,24 @@ impl<T: BeaconChainTypes> DataAvailabilityCheckerInner<T> {
...
         let num_of_columns = self.spec.number_of_columns;
         let has_missing_columns = pending_components.verified_data_columns.len() < num_of_columns;
-
-        has_missing_columns
+        let is_super_node = self.custody_column_count == num_of_columns;
+        let is_reconstruction_possible =
+            pending_components.verified_data_columns.len() >= num_of_columns / 2;
+        is_super_node
+            && has_missing_columns
+            && is_reconstruction_possible
```
dapplion: Conflicts with 
- https://github.com/sigp/lighthouse/pull/6403

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -0,0 +1,305 @@
...
+use types::{
+    BeaconStateError, BlobSidecar, DataColumnSidecarList, EthSpec, FullPayload, Hash256,
+    SignedBeaconBlock, SignedBeaconBlockHeader,
+};
+
+/// Number of batches that supernodes split data columns into during publishing.
+pub const SUPERNODE_DATA_COLUMN_PUBLICATION_BATCHES: usize = 4;
+
+/// The delay applied by supernodes between the sending of each data column batch.
+pub const SUPERNODE_DATA_COLUMN_PUBLICATION_BATCH_INTERVAL: Duration = Duration::from_millis(200);
```
dapplion: These (batch count + interval) are key parameters that we are just yolo guessing, should this be a flag?

beacon_node/beacon_chain/src/metrics.rs
```diff
@@ -111,6 +111,12 @@ pub static BLOCK_PROCESSING_POST_EXEC_PROCESSING: LazyLock<Result<Histogram>> =
             linear_buckets(5e-3, 5e-3, 10),
         )
     });
+pub static BLOCK_PROCESSING_DATA_COLUMNS_WAIT: LazyLock<Result<Histogram>> = LazyLock::new(|| {
+    try_create_histogram(
+        "beacon_block_processing_data_columns_wait_seconds",
+        "Time spent waiting for data columns to be computed before starting database write",
```
dapplion, line 117: Should add buckets?

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -0,0 +1,305 @@
...
+                    // publish them in batches. Our hope is that some columns arrive from
+                    // other supernodes in the meantime, obviating the need for us to publish
+                    // them. If no other publisher exists for a column, it will eventually get
+                    // published here.
+                    // FIXME(das): deduplicate this wrt to gossip/sync methods
+                    all_data_columns.shuffle(&mut rand::thread_rng());
+
+                    let batch_size =
+                        all_data_columns.len() / SUPERNODE_DATA_COLUMN_PUBLICATION_BATCHES;
+                    for batch in all_data_columns.chunks(batch_size) {
```
dapplion: Consider a case where the columns to publish are batched into `[[1,2,3,4],[5,6,7,8]]`. At the moment to publish the first batch, `[1,2,3,4]` are already seen, but `[5,6,7,8]`. With this design the node will sit idle for `SUPERNODE_DATA_COLUMN_PUBLICATION_BATCH_INTERVAL` and do nothing. 

Is that expected? Do we want max the rate limited bandwidth of the node, or distribute publishing even if some supernodes don't publish anything for an interval?

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -0,0 +1,305 @@
...
+
+        Some(data_columns_receiver)
+    } else {
+        if num_fetched_blobs == 0 {
+            debug!(
+                chain.log,
+                "No blobs fetched from the EL";
+                "num_expected_blobs" => num_expected_blobs,
+            );
+            inc_counter(&metrics::BLOBS_FROM_EL_MISS_TOTAL);
```
dapplion: The metric `BLOBS_FROM_EL_MISS_TOTAL` is being used with different meaning on each if branch. Here it means 0 fetched blobs, in peerdas it means != total. 

Also, we could add another histogram to observe the count of missing blobs from the expected total.

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -0,0 +1,305 @@
...
+            return Ok(None);
+        }
+
+        inc_counter(&metrics::BLOBS_FROM_EL_HIT_TOTAL);
```
dapplion: Again, very different meaning to the metric in peerdas

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -172,23 +179,64 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                };
+
+                // If this node is a super node, permute the columns and split them into batches.
+                // The hope is that we won't need to publish some columns because we will receive them
+                // on gossip from other supernodes.
+                data_columns_to_publish.shuffle(&mut rand::thread_rng());
+                let batch_size =
+                    data_columns_to_publish.len() / SUPERNODE_DATA_COLUMN_PUBLICATION_BATCHES;
+
+                for batch in data_columns_to_publish.chunks(batch_size) {
```
dapplion: Should de-duplicate this from `beacon_node/beacon_chain/src/fetch_blobs.rs` if we are following the same batch strategy

beacon_node/network/src/network_beacon_processor/mod.rs
```diff
@@ -767,6 +770,78 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                AvailabilityProcessingStatus::Imported(_) => {
+                    debug!(
+                        self.log,
+                        "Block components retrieved from EL";
+                        "result" => "imported block and custody columns",
+                        "block_hash" => %block_root,
+                    );
+                    self.chain.recompute_head_at_current_slot().await;
+                }
+                AvailabilityProcessingStatus::MissingComponents(_, _) => {
```
dapplion, line 929: Yes I think it's possible before peerdas if the EL returns more than 0 blobs but not all of them

beacon_node/network/src/network_beacon_processor/mod.rs
```diff
@@ -767,6 +770,78 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                    error!(
+                        self.log,
+                        "MissingComponents is not expected after engine blobs processed successfully";
+                        "block_hash" => %block_root,
+                    );
+                }
+            },
+            Ok(None) => { /* No blobs fetched from the EL. Reasons logged separately. */ }
```
dapplion: Is it worth it to still log a debug here?

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: yeah these two branches conflicts a lot and I'm planning to resolve them once one of them is merged. 

Another conflict is at the trigger point. They're both triggered when we have missing components - and we'd need to decide whether to fetch from EL or perform reconstruction in the same place. The benchmark for both are very similar but I think we should prioritise reconstruction if we have 50%, so we don't have to try fetch from EL, and fallback to reconstruction.

beacon_node/beacon_chain/src/fetch_blobs.rs
michaelsproul: yeh

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
michaelsproul: that's this FIXME

https://github.com/sigp/lighthouse/blob/4639146dfe16107aac74f2d42115c3e64e67fb37/beacon_node/beacon_chain/src/fetch_blobs.rs#L185

jimmygchen:
I think we should review blob publishing before merging this to avoid adding excessive bandwidth usage for Deneb.

beacon_node/beacon_chain/src/fetch_blobs.rs
michaelsproul: It's random, so the chance of having a first batch that's already fully known is quite unlikely.

jimmygchen:
Be great to have https://github.com/sigp/lighthouse/pull/6403 merged first, as there will be some more conflicts that needs to be resolved, and a bit easier in this order. We also need to prioritise block publishing, could do it in this PR.

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: How is this possible? only custody & sample columns should get in here, if we have other random columns getting into da checker, this check won't work.

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: Yep agree! I also updated this on the reconstruction PR

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: This is to cover the scenario where block is made available via blobs, and we don't have all the custody columns.

So if `columns.len` is less than `custody_columns_count`, we just drop them, and wait for the columns to be computed and sent via `data_column_recv`.

The comment is in the below section, but I think it's not very obvious, I'll add another comment here.

```rust
            // If the block was made available via custody columns received from gossip / rpc, use them
            // since we already have them.
            (Some(columns), _) => Some(columns),
            // Otherwise, it means blobs were likely available via fetching from EL, in this case we
            // wait for the data columns to be computed (blocking).
            (None, Some(mut data_column_recv)) => {
```

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: Resolved conflicts in https://github.com/sigp/lighthouse/pull/6268/commits/d8eedf30c54c1a2cc0294d120717329b4371ebc7

beacon_node/network/src/network_beacon_processor/mod.rs
jimmygchen: Yep, I've downgraded the log to `debug`

jimmygchen:
Remaining TODOs
- [x] prioritise block publishing (https://github.com/sigp/lighthouse/pull/6268/commits/0a2c6f76c790e2117ed4a225ef312795eb26f240)
- [x] send `IDONTWANT` on publish 
  - #6513

beacon_node/beacon_chain/src/fetch_blobs.rs
jimmygchen: Yep we should review the metrics here and add the ones you suggested.

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
dapplion: https://github.com/sigp/lighthouse/pull/6268/files#r1764455920

jimmygchen: ah ok what see what you mean, we can change that check to `==`

beacon_node/network/src/network_beacon_processor/mod.rs
jimmygchen: Addressed in 0a9d5a0f61b4114dff93fe88c46ca4b97dde75f2

beacon_node/beacon_chain/src/fetch_blobs.rs
jimmygchen: Addressed in 0a9d5a0f61b4114dff93fe88c46ca4b97dde75f2

beacon_node/beacon_chain/src/fetch_blobs.rs
jimmygchen: Addressed in 0a9d5a0f61b4114dff93fe88c46ca4b97dde75f2

beacon_node/beacon_chain/src/metrics.rs
jimmygchen: Addressed in 0a9d5a0f61b4114dff93fe88c46ca4b97dde75f2

beacon_node/beacon_chain/src/fetch_blobs.rs
jimmygchen: Addressed in 0a9d5a0f61b4114dff93fe88c46ca4b97dde75f2

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: Addressed in 0a9d5a0f61b4114dff93fe88c46ca4b97dde75f2

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Addressed in 0a9d5a0f61b4114dff93fe88c46ca4b97dde75f2

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3500,35 +3559,28 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
             .data_availability_checker
-            .put_rpc_blobs(block_root, epoch, blobs)?;
+            .put_rpc_blobs(block_root, blobs)?;
 
-        self.process_availability(slot, availability, || Ok(()))
+        self.process_availability(slot, availability, None, || Ok(()))
+            .await
+    }
+
+    async fn check_engine_blob_availability_and_import(
```
dapplion, line 3566: Should inline this function into process_engine_blobs?

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3840,15 +3893,52 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
 
-        let db_write_timer = metrics::start_timer(&metrics::BLOCK_PROCESSING_DB_WRITE);
-
         // Store the block and its state, and execute the confirmation batch for the intermediate
         // states, which will delete their temporary flags.
         // If the write fails, revert fork choice to the version from disk, else we can
         // end up with blocks in fork choice that are missing from disk.
         // See https://github.com/sigp/lighthouse/issues/2028
         let (_, signed_block, blobs, data_columns) = signed_block.deconstruct();
+        // TODO(das) we currently store all subnet sampled columns. Tracking issue to exclude non custody columns: https://github.com/sigp/lighthouse/issues/6465
```
dapplion: Exceeds max fmt width

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -248,12 +243,32 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
...
-            epoch,
             verified_custody_columns,
             &self.log,
         )
     }
 
+    /// Put a list of blobs received from the EL pool into the availability cache.
+    ///
+    /// This DOES NOT perform KZG verification because the KZG proofs should have been constructed
+    /// immediately prior to calling this function so they are assumed to be valid.
```
dapplion, line 258: Why not change this function and `process_engine_blobs` to take `KzgVerifiedBlobList` and make it correct by construction?

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -199,63 +194,49 @@ impl<E: EthSpec> PendingComponents<E> {
...
-                debug!(log,
-                    "Component(s) added to data availability checker";
-                    "block_root" => ?self.block_root,
-                    "received_block" => block_kzg_commitments_count_opt.is_some(),
-                    "received_columns" => num_received_columns,
-                    "expected_columns" => expected_columns_msg,
-                );
+        let all_blobs_received = block_kzg_commitments_count_opt
+            .map_or(false, |num_expected_blobs| {
+                num_expected_blobs == num_received_blobs
```
dapplion, line 232: Is it worth switching to `>=` in case there's some future bug?
```suggestion
                num_received_blobs >= num_expected_blobs
```

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -304,26 +284,25 @@ impl<E: EthSpec> PendingComponents<E> {
...
+                .into_iter()
+                .map(|d| d.into_inner())
+                .collect::<Vec<_>>();
+            (None, Some(data_columns))
+        } else {
+            let num_blobs_expected = diet_executed_block.num_blobs_expected();
+            let Some(verified_blobs) = verified_blobs
+                .into_iter()
+                .map(|b| b.map(|b| b.to_blob()))
+                .take(num_blobs_expected)
```
dapplion, line 299: Why is it necessary to take here? Do we consider the case where we have too many blobs?

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -0,0 +1,333 @@
...
+        num_expected_blobs as u64,
+    );
+    inc_counter_by(
+        &metrics::BLOBS_FROM_EL_RECEIVED_TOTAL,
+        num_fetched_blobs as u64,
+    );
+
+    if num_fetched_blobs == 0 {
+        debug!(
+            chain.log,
```
dapplion:
```suggestion
           log,
```

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -0,0 +1,333 @@
...
+        response,
+        signed_block_header,
+        &kzg_commitments_proof,
+    )?;
+
+    let peer_das_enabled = chain.spec.is_peer_das_enabled_for_epoch(block.epoch());
+
+    let data_columns_receiver_opt = if peer_das_enabled {
+        // Partial blobs response isn't useful for PeerDAS, so we don't bother building and publishing data columns.
+        if !num_fetched_blobs == num_expected_blobs {
```
dapplion: Again do we work on the assumption that EL could be buggy and return more blobs than expected? I feel it's more correct to do
```suggestion
        if num_fetched_blobs < num_expected_blobs {
```

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -0,0 +1,333 @@
...
+            if let Err(e) = data_columns_sender.try_send(all_data_columns.clone()) {
+                error!(log, "Failed to send computed data columns"; "error" => ?e);
+            };
+
+            // Check indices from cache before sending the columns, to make sure we don't
+            // publish components already seen on gossip.
+            let is_supernode = chain_cloned
+                .data_availability_checker
+                .get_sampling_column_count()
+                == chain_cloned.spec.number_of_columns;
```
dapplion: Could add a function to `DataAvailabilityChecker` to not spread the notion of supernode-ness in multiple places
```rust
fn is_supernode(&self) -> bool
```

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -0,0 +1,333 @@
...
+                    .iter()
+                    .filter(|col| !already_seen.contains(&col.index()))
+                    .cloned()
+                    .collect::<Vec<_>>();
+                if !publishable.is_empty() {
+                    debug!(
+                        log,
+                        "Publishing data columns from EL";
+                        "count" => publishable.len()
+                    );
```
dapplion: Here we can compute the efficiency of this delay publish mechanism: track which columns were originally scheduled for publishing and became seen in this for loop. We can track that as a counter of "already_seen_before_publish" or something related. And maybe print a summary after the for loop summarizing how many we published in total, and how many we skipped publishing or how many we intended to publish at the beginning.

Also, do we log somewhere which specific column index we publish? Otherwise, it might be helpful to log in this line a vec of indices of which columns we are publishing.

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -0,0 +1,333 @@
...
+            signed_block_header.clone(),
+            kzg_commitments_proof,
+            blob_and_proof.proof,
+        ) {
+            Ok(blob) => {
+                if let Some(blob_mut) = fixed_blob_sidecar_list.get_mut(i) {
+                    *blob_mut = Some(Arc::new(blob));
+                } else {
+                    return Err(FetchEngineBlobError::InternalError(
+                        "Unreachable: Blobs from EL - out of bounds".to_string(),
```
dapplion: This is reachable, the EL can send malformed blobs with a random index

beacon_node/network/src/network_beacon_processor/mod.rs
```diff
@@ -946,6 +1014,71 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                            "Publishing data column batch";
+                            "count" => publishable.len()
+                        );
+                        publish_fn(publishable);
+                    }
+
+                    tokio::time::sleep(supernode_data_column_publication_batch_interval).await;
+                }
+            },
+            "handle_data_columns_publish",
```
dapplion: Not a blocker to merging, but since reviews are slow maybe there's time to de-dup this bit

beacon_node/src/cli.rs
```diff
@@ -86,6 +86,22 @@ pub fn cli_app() -> Command {
...
+                .display_order(0)
+        )
+        .arg(
+            Arg::new("supernode-data-column-publication-batch-interval")
+                .long("supernode-data-column-publication-batch-interval")
+                .action(ArgAction::Set)
+                .help_heading(FLAG_HEADER)
+                .help("The delay in milliseconds applied by supernodes between the sending of each data column batch. For PeerDAS only.")
+                .display_order(0)
+        )
```
dapplion: Should we hide these?

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Addressed in bf19769e0d7bf425c3b9850216ef10e7a611491e.

beacon_node/src/cli.rs
jimmygchen: Addressed in bf19769e0d7bf425c3b9850216ef10e7a611491e.

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: I have a weak preference of not inlining it for consistency with other similar functions and keeping `process_rpc_blobs` small.

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: If we received more valid blobs than we should, then we probably want it to fail rather than covering it and importing them into the database?

Not sure how that's possible though, as blob sidecar fixed list is constructed here, and will fail if it does not have a corresponding kzg commitment:
https://github.com/sigp/lighthouse/blob/bf19769e0d7bf425c3b9850216ef10e7a611491e/beacon_node/beacon_chain/src/fetch_blobs.rs#L298-L304

beacon_node/beacon_chain/src/fetch_blobs.rs
jimmygchen: what about making `let num_fetch_blobs = fixed_blob_sidecar_list.filter(|b| b.is_some()).count()` so we always checked against the verified and de-deuplicated list?

beacon_node/beacon_chain/src/fetch_blobs.rs
jimmygchen: If EL sends a blob with invalid index, `BlobSidecar::new_with_existing_proof` returns `MissingKzgCommitment`

jimmygchen: I've updated the log message to include the index. 6e44763b26fa9fa3e5960a1d19ab738097abe24d

beacon_node/beacon_chain/src/fetch_blobs.rs
jimmygchen: Good idea, addressed in 6e44763b26fa9fa3e5960a1d19ab738097abe24d

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: This is an existing check from Deneb to make sure we have exactly the same number of blobs as the kzg commitments, otherwise it returns `AvailabilityCheckError::Unexpected`

beacon_node/beacon_chain/src/fetch_blobs.rs
michaelsproul: This condition is using unary negation! i.e. it is calculating `!num_fetched_blobs` and then comparing that for equality with `num_expected_blobs`. I think what was intended was `!(num_fetched_blobs == num_expected_blobs)` i.e. `num_fetched_blobs != num_expected_blobs`. With the unary negation I don't think the body of this `if` would ever have been executed, because negating all the bits of a low number creates some insanely high value that is mostly 1s.

beacon_node/network/src/network_beacon_processor/mod.rs
jimmygchen: Addressed in https://github.com/sigp/lighthouse/pull/6268/commits/1872ae53b80d654f0764f17595f79ee1bc45d634

beacon_node/beacon_chain/src/fetch_blobs.rs
jimmygchen: Added more logging in 81493f5e67da03aabbffbbf3c0d11f754e27ec21

> Also, do we log somewhere which specific column index we publish? Otherwise, it might be helpful to log in this line a vec of indices of which columns we are publishing

Yes this is logged in `network_service`:
https://github.com/sigp/lighthouse/blob/9aefb5539baff637d68deb3dd386ff45312f3573/beacon_node/network/src/service.rs#L666-L671

beacon_node/beacon_chain/src/beacon_chain.rs
michaelsproul: agree, let's keep it not-inlined

beacon_node/beacon_chain/src/data_availability_checker.rs
michaelsproul: I think it just puts more complexity at the call site. Not necessarily an improvement

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
michaelsproul: Resolving as beyond the scope of this PR

beacon_node/beacon_chain/src/fetch_blobs.rs
jimmygchen: Fixed in https://github.com/sigp/lighthouse/pull/6268/commits/756230e9589b6f15ef89495d7070c25adf3f97d3

jimmygchen:
I'm observing a few things from testing:

The node sometimes end up publishing all blobs, if the pending component is not found in availability cache after the block has been made available (`cached_blob_indexes` returns `None` below):
https://github.com/sigp/lighthouse/blob/0f32b73ebc1505b3495d3719e158ac4630d7020b/beacon_node/beacon_chain/src/fetch_blobs.rs#L153-L165
The evidence is that we also see a blob processing error `DuplicateFullyImported` when trying to process the already imported blobs:
```
Oct 30 10:06:02.780 DEBG Fetching blobs from the EL, num_expected_blobs: 6, block_root: 0x8ae0291f630ab24311fa859a2e86c6fb8995fb2df420516ecbbe3a8eaa339ebf, service: fetch_engine_blobs, service: beacon, module: beacon_chain::fetch_blobs:78
Oct 30 10:06:02.968 DEBG Processing engine blobs, num_fetched_blobs: 6, block_root: 0x8ae0291f630ab24311fa859a2e86c6fb8995fb2df420516ecbbe3a8eaa339ebf, service: fetch_engine_blobs, service: beacon, module: beacon_chain::fetch_blobs:170
Oct 30 10:06:02.968 ERRO Error fetching or processing blobs from EL, block_hash: 0x8ae0291f630ab24311fa859a2e86c6fb8995fb2df420516ecbbe3a8eaa339ebf, error: BlobProcessingError(DuplicateFullyImported(0x8ae0291f630ab24311fa859a2e86c6fb8995fb2df420516ecbbe3a8eaa339ebf)), module: network::network_beacon_processor:937
Oct 30 10:06:03.871 DEBG Batch blob publication complete, block_root: 0x8ae0291f630ab24311fa859a2e86c6fb8995fb2df420516ecbbe3a8eaa339ebf, published_count: 6, blob_count: 6, batch_interval: 300, service: beacon, module: network::network_beacon_processor:1066
```

Possible solution:
1. ~~Check fork choice again before publishing and processing engine blobs~~
2. Downgrade error `BlockError::DuplicateFullyImported` to a debug log, as this race can sometimes happens

**UPDATE:** I think a better solution would be gossip verifying the blobs prior to publishing - this is consistent with our blob publishing logic and if the blob has already been observed, then we would get `GossipBlobError::RepeatBlob` and not publish and process the duplicate blob.

jimmygchen:
@dapplion @michaelsproul 
I've made a few more changes and have tested this locally. This is ready for review now:
- Fix invalid inclusion proof for EL blobs
- Only publish blobs triggered from gossip block and not RPC blocks

I'll deploy this to some testnet nodes.

jimmygchen:
Some findings from mainnet testing from *a small sample size*:
- EL `getBlobs` only gets triggered once roughly every ~5-10 mins. This is likely because relays publish blobs before blocks. This may change in the future though when blob count increases, or if nodes prioritise block publishing.
- Interestingly, the pattern is very inconsistent (sometimes `getBlobs` could get triggered for 5-6 consecutive blocks), and it seems to be happen more on `ultrasound` relay and non builder blocks. (perhaps they don't prioritise sending blobs as much as other relays?)
- A large percentage of them are blocks with 5 or 6 blobs - which is what we expect, as proposer would have to disseminate more blob data to its peers and more likely to have delays.
- With the gradual publication optimisation, we only publish about 1 blob roughly every 10 mins, so it *should* have minimal impact on outbound bandwidth.

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -0,0 +1,308 @@
+//! This module implements an optimisation to fetch blobs via JSON-RPC from the EL.
+//! If a blob has already been seen in the public mempool, then it is often unnecessary to wait for
+//! it to arrive on P2P gossip. This PR uses a new JSON-RPC method (`engine_getBlobsV1`) which
+//! allows the CL to load the blobs quickly from the EL's blob pool.
+//!
+//! Once the node fetches the blobs from EL, it then publishes the remaining blobs that hasn't seen
```
michaelsproul:
```suggestion
//! Once the node fetches the blobs from EL, it then publishes the remaining blobs that it hasn't seen
```

consensus/types/src/signed_beacon_block.rs
```diff
@@ -239,6 +240,45 @@ impl<E: EthSpec, Payload: AbstractExecPayload<E>> SignedBeaconBlock<E, Payload>
...
+        // Create the block body merkle tree
+        let body_leaves = self.message().body().body_merkle_leaves();
+        let beacon_block_body_depth = body_leaves.len().next_power_of_two().ilog2() as usize;
+        let body_merkle_tree = MerkleTree::create(&body_leaves, beacon_block_body_depth);
+
+        // Compute the KZG commitments inclusion proof
+        let (_, proof) = body_merkle_tree
+            .generate_proof(BLOB_KZG_COMMITMENTS_INDEX, beacon_block_body_depth)
+            .map_err(Error::MerkleTreeError)?;
+        let kzg_commitments_inclusion_proof = FixedVector::new(proof)?;
```
michaelsproul, line 264: I notice that this was duplicated from `kzg_commitments_merkle_proof` in the process of fixing the invalid proofs: https://github.com/sigp/lighthouse/pull/6268/commits/2f11f3cef33809124137bc5067a7fbdc957129fc

Is this because the last element of the proof _is not_ the body root, as this code had previously assumed? I see you're using `body_merkle_tree.hash()` to calculate it instead?

I don't quite understand what's going on here, as the comments on `MerkleTree::generate_proof`, suggest that the root of the tree _is_ the last element:

```rust
    /// Return the leaf at `index` and a Merkle proof of its inclusion.
    ///
    /// The Merkle proof is in "bottom-up" order, starting with a leaf node
    /// and moving up the tree. Its length will be exactly equal to `depth`.
```

What am I missing?

michaelsproul: Ok, I see the issue now. The merkle proof is not expected to contain the body root, as it is verified against the body root which is passed in as an additional argument in the [spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/p2p-interface.md#verify_blob_sidecar_inclusion_proof):

```python
def verify_blob_sidecar_inclusion_proof(blob_sidecar: BlobSidecar) -> bool:
    gindex = get_subtree_index(get_generalized_index(BeaconBlockBody, 'blob_kzg_commitments', blob_sidecar.index))
    return is_valid_merkle_branch(
        leaf=blob_sidecar.kzg_commitment.hash_tree_root(),
        branch=blob_sidecar.kzg_commitment_inclusion_proof,
        depth=KZG_COMMITMENT_INCLUSION_PROOF_DEPTH,
        index=gindex,
        root=blob_sidecar.signed_block_header.message.body_root,
    )
```

I think the code duplication is OK for now.

michaelsproul (approved):

michaelsproul:
@mergify queue

consensus/types/src/signed_beacon_block.rs
jimmygchen: > Is this because the last element of the proof is not the body root, as this code had previously assumed?

Yes that's right. I also assumed the same and noticed that the body root was not returned from the `kzg_commitments_merkle_proof` function.

Here's how it looks like:

```
tree: Node(0xf260349e03f826c46f34efa8e943e0968294769a64b604d2504a80086561a2e0, Node(0x486d1c1a4f184afcdacc79f6db4d597d253b18cedc6650e5a0e476c0c6f77291, Node(0x6b2c3de18ef5ab5eab5b51436deaab9e030150aec6048801339ab285b45ae53f, Node(0x983ec82b9ba2e98e30cef7b0d0dcb14cb22acb45265996126bd53310a6e58a08, Leaf(0x8e742f5552fe005d58a4676d4c9709ac5cba3fc36cb99e387dcb2414edf53252), Leaf(0x0e07987a0f3b73e27a96321bedd723418f904225a74c471807203578b33e7217)), Node(0xea3336446bec447a262daa36ad1af5c3399d0ce139b6c923e687f5312d3a0800, Leaf(0x25748e4d40e0ebe492e451acf202dc39390cc685e22fa9a033341fff1bdb70a1), Leaf(0x8adfe28d95e5addbf1da4bff767f2693772471b83480cb6e4675b0983784a882))), Node(0xab16b427891c1851c565890e64dc70460d74166626f04a20a9088ea6ab232d93, Node(0x194b7b8ca7e899490d46bcdf67064131ce01db44e790f2f8a5c7c67dec4da2b4, Leaf(0x7a0501f5957bdf9cb3a8ff4966f02265f968658b7a9c62642cba1165e86642f5), Leaf(0x722f01799979caa2a3b3756f66e454c7e40e612cf6946e2f47b8012859798710)), Node(0x65740c64ece7e25b3d64491b394c082bef66cf91bd15fb0c135bca21c0025444, Leaf(0x792930bbd5baac43bcc798ee49aa8185ef76bb3b44ba62b91d86ae569e4bb535), Leaf(0x4a88fad606c5a81c01b68bfefbd93faba7e258046748da3a61975a3bb6c05212)))), Node(0x4ddafcd836f8a7ae8bb4c41159b4822054f7af664666882657dae73df482b33c, Node(0x3ffbaab784b1ede7fbbd558e6381cf6280bcc09f7b52cfa6321acfc240f3ec26, Node(0x5629b68a929f42c061c05145089d83b237783a93146326b25c171b97602d80e9, Leaf(0x08af26106a6b4c81ed49ee5080d465e42c6666e6163426116701922648c505cf), Leaf(0xdd63871f5155c221864d394560d89884201c2753ab67d833450a81a9d007aca7)), Node(0x877d453667ef92b0da105b398e24a2c59c104620075ae7b8b7bb11b85cbd0d93, Leaf(0x916a3b3c9be953a8db6d4887fee1dd2040ec57510fd8a747a1833fb0b806fd1b), Leaf(0xc52dde5d8ddefe2dc56a2a42484d964dbd146a4f8af2ee63aaecb240c2b4f79f))), Zero(2)))
proof: [0x916a3b3c9be953a8db6d4887fee1dd2040ec57510fd8a747a1833fb0b806fd1b, 0x5629b68a929f42c061c05145089d83b237783a93146326b25c171b97602d80e9, 0xdb56114e00fdd4c1f85c892bf35ac9a89289aaecb1ebd0a96cde606a748b5d71, 0x486d1c1a4f184afcdacc79f6db4d597d253b18cedc6650e5a0e476c0c6f77291]
body root: 0xf260349e03f826c46f34efa8e943e0968294769a64b604d2504a80086561a2e0
```

The function calculates the proof and discard the merkle tree, so the change here is just taking the `kzg_commitments_merkle_proof` logic here, and get the body root from the merkle tree, I believe there's no extra calculation.

I didn't change the existing function because it's used in multiple places.

jimmygchen:
One more thing I was thinking about - we may not want to _always_  fetch blobs on rpc response, as it only makes sense / works if we're close to the head slot, but I guess `process_rpc_block` only happens if we're close to the head (block lookup and head sync), so I don't think there's anything we need to address here, but just wanted to keep a note.
https://github.com/sigp/lighthouse/blob/8060e86275f169b26e16b41e54a7e2e50fd33774/beacon_node/network/src/network_beacon_processor/sync_methods.rs#L192-L199

PR #6271, opened 2024-08-16
Expose ENR custody columns

dapplion:
## Issue Addressed

Part of 
- https://github.com/sigp/lighthouse/issues/6072

## Proposed Changes

- Add `custody_columns` method to `NetworkGlobals`

### DIFF
--- a/beacon_node/lighthouse_network/src/types/globals.rs
+++ b/beacon_node/lighthouse_network/src/types/globals.rs
@@ -1,4 +1,5 @@
 //! A collection of variables that are accessible outside of the network thread itself.
+use crate::discovery::enr::Eth2Enr;
 use crate::peer_manager::peerdb::PeerDB;
 use crate::rpc::{MetaData, MetaDataV2};
 use crate::types::{BackFillState, SyncState};
@@ -7,7 +8,7 @@ use crate::EnrExt;
 use crate::{Enr, GossipTopic, Multiaddr, PeerId};
 use parking_lot::RwLock;
 use std::collections::HashSet;
-use types::{ChainSpec, ColumnIndex, EthSpec};
+use types::{ChainSpec, ColumnIndex, DataColumnSubnetId, EthSpec};
 
 pub struct NetworkGlobals<E: EthSpec> {
     /// The current local ENR.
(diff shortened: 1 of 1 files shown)


jimmygchen (approved):

beacon_node/lighthouse_network/src/types/globals.rs
```diff
@@ -111,10 +112,12 @@ impl<E: EthSpec> NetworkGlobals<E> {
...
 
     /// Compute custody data columns the node is assigned to custody.
-    pub fn custody_columns(&self, _spec: &ChainSpec) -> Vec<ColumnIndex> {
-        let _enr = self.local_enr();
-        //TODO(das): implement ENR changes
-        vec![]
+    pub fn custody_columns(&self, spec: &ChainSpec) -> Vec<ColumnIndex> {
+        let enr = self.local_enr();
+        let node_id = enr.node_id().raw().into();
+        let custody_subnet_count = enr.custody_subnet_count::<E>(spec);
```
pawanjay176, line 118: Just a small suggestion wrt `Enr::custody_subnet_count`, I think we should make that simply return an error on invalid/non existing enr instead of returning the default value. Basically change the function signature to
```rust
fn custody_subnet_count<E: EthSpec>(&self, spec: &ChainSpec) -> Result<u64, &'static str>;
```
Swallowing the error at the accessor function and returning the default might lead to confusing errors.

jimmygchen: Good that you raised this again - we actually thought about this when this was initially implemented, and the original reason to swallow it is wanting to have the fallback logic in one place, and don't want to do any error handling due to invalid input from peers, but maybe there's better approach and may even be useful to log an error?

pawanjay176: We could just do the fallback logic at the calling site. The accessor function on that trait doesn't seem like the right place to do fallback logic.
Consider a situation where a client has a bug in their encoding for the enr custody subnet field, in that case, we would always swallow the error which isn't what we want to do.

jimmygchen:
The main `das` branch has been merged and this change has already been included.

PR #6275, opened 2024-08-16
Add custody by root request to sync

dapplion:
## Issue Addressed

Part of 
- https://github.com/sigp/lighthouse/issues/6072

Build on / depends on
- https://github.com/sigp/lighthouse/pull/6274

## Proposed Changes

Adds a new higher order request to fetch all custody columns of a given block root.

It's a stateful request that may trigger one or more data_columns_by_root internally with features:
- per-column retry
- peer load balancing / prioritization of peer
- don't re-request peers that return failures for some time


jimmygchen:
This was automatically closed by mergify as the upstream branch was merged. I'll reopen this PR against `unstable`.

jimmygchen:
The refactoring on these PRs are making merging back to `das` really difficult, can we please avoid refactoring on these PRs, and instead do it against `das` branch? 

Some of the new function signatures require reworking parts of the existing code on `das` branch, so it would be easier if the refactor happens directly on `das` branch, otherwise we'd have to do the rework as part of the merge back to `das` which is really painful.

beacon_node/lighthouse_network/src/types/globals.rs
```diff
@@ -111,10 +114,32 @@ impl<E: EthSpec> NetworkGlobals<E> {
...
+        column_index: ColumnIndex,
+        spec: &ChainSpec,
+    ) -> Vec<PeerId> {
+        self.peers
+            .read()
+            .good_peers_on_subnet(Subnet::DataColumn(
+                DataColumnSubnetId::from_column_index::<E>(column_index as usize, spec),
+            ))
+            .cloned()
+            .collect::<Vec<_>>()
```
jimmygchen, line 142: Why not take the latest change from `das`branch and just use `good_custody_subnet_peer`?
This undos most the changes from #6218 - i understand selecting peers from entire peers is not the right solution, but even this doesn't fix it and just revert back to the previous version.

beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
```diff
@@ -673,17 +676,62 @@ impl<E: EthSpec> PeerDB<E> {
...
+            // - is_good_gossipsub_peer: score >= 0, which equals default score
+            // peer.set_meta_data(crate::rpc::MetaData::V2(crate::rpc::MetaDataV2 {
+            //     seq_number: 0,
+            //     attnets: <_>::default(),
+            //     syncnets: <_>::default(),
+            // }));
+            peer.insert_subnet(Subnet::DataColumn(subnet.into()));
+        }
+
+        peer_id
```
jimmygchen, line 734: Is all this change required? Why not just take the latest change from `das` branch? The change in #6218 fixes the custody lookup tests.

beacon_node/lighthouse_network/src/types/globals.rs
jimmygchen: Also happy if you'd like to change how it works on the `das` branch but i think it's better to target rework on the `das` branch before we finish the merge to `unstable`.

beacon_node/network/src/sync/network_context/custody_by_root.rs
```diff
@@ -0,0 +1,417 @@
...
+use types::EthSpec;
+use types::{data_column_sidecar::ColumnIndex, DataColumnSidecar, Hash256};
+
+use super::{
+    DataColumnsByRootSingleBlockRequest, LookupRequestResult, RpcResponseResult, SyncNetworkContext,
+};
+
+const FAILED_PEERS_CACHE_EXPIRY_SECONDS: u64 = 5;
+
+type DataColumnSidecarVec<E> = Vec<Arc<DataColumnSidecar<E>>>;
```
jimmygchen, line 19: For consistency:

```suggestion
type DataColumnSidecarList<E> = Vec<Arc<DataColumnSidecar<E>>>;
```

Also updated on the `das` branch `custody.rs`

beacon_node/network/src/sync/network_context/custody_by_root.rs
```diff
@@ -0,0 +1,417 @@
...
+
+pub struct ActiveCustodyRequest<T: BeaconChainTypes> {
+    block_root: Hash256,
+    requester: SingleLookupReqId,
```
jimmygchen, line 23: What's the reason to change this from `CustodyId`? (on `das` branch)

beacon_node/network/src/sync/network_context/custody_by_root.rs
```diff
@@ -0,0 +1,417 @@
...
+        req_id: DataColumnsByRootRequestId,
+    },
+}
+
+struct ActiveBatchColumnsRequest {
+    peer_id: PeerId,
+    indices: Vec<ColumnIndex>,
+}
+
+type CustodyRequestResult<E> = Result<Option<DataColumnSidecarVec<E>>, Error>;
```
jimmygchen, line 57: Are we dumping the peer group?

`das` branch: 

```
type CustodyRequestResult<E> = Result<Option<(DataColumnSidecarList<E>, PeerGroup)>, Error>;
```

beacon_node/network/src/sync/network_context/custody_by_root.rs
```diff
@@ -0,0 +1,417 @@
...
+                    let column_request = self
+                        .column_requests
+                        .get_mut(column_index)
+                        .ok_or(Error::BadState("unknown column_index".to_owned()))?;
+
+                    if let Some(data_column) = data_columns.remove(column_index) {
+                        column_request.on_download_success(
+                            req_id,
+                            peer_id,
+                            // Safe to cast, self.column_requests only contains indexes for columns we must custody
```
jimmygchen, line 138: This comment can be removed as we're no longer casting, although I'm not sure why we removed the `CustodyDataColumn` type

beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -311,12 +357,34 @@ impl TestRig {
...
+        // One supernode peer to ensure all columns have at least one peer
+        self.new_connected_supernode_peer();
+
+        // ensure we have peers in each subnet
+        for i in 0..self.harness.spec.data_column_sidecar_subnet_count {
+            let peers = self
+                .network_globals
+                .custody_peers_for_column(i, &self.harness.spec);
+            println!("peers column {} {:?}", i, peers);
+        }
```
jimmygchen, line 387: Is this just for debugging? it doesn't actually assert anything but just prints

beacon_node/network/src/sync/manager.rs
```diff
@@ -965,15 +966,31 @@ impl<T: BeaconChainTypes> SyncManager<T> {
     fn on_data_columns_by_root_response(
         &mut self,
         req_id: DataColumnsByRootRequestId,
-        _requester: SingleLookupReqId,
+        requester: SingleLookupReqId,
```
jimmygchen, line 969: This parameter type isn't compatible with the `das` branch, and if we want to change this type i think we should change it there instead, as it would affect other part of the code (e.g. sampling) and make merging back and conga line difficult

beacon_node/network/src/sync/manager.rs
```diff
@@ -965,15 +966,31 @@ impl<T: BeaconChainTypes> SyncManager<T> {
...
+            {
+                // TODO(das): get proper timestamp
+                let seen_timestamp = timestamp_now();
+                self.block_lookups
+                    .on_download_response::<CustodyRequestState<T::EthSpec>>(
+                        requester,
+                        // TODO(das): this is not the correct peer to attribute to a response. It's
+                        // just the last peer of multiple potential requests. We should switch to a
+                        // concept of "peer group" to attribute fault to the correct peer.
+                        peer_id,
```
jimmygchen, line 989: We already have this on `das` branch, is there a reason why you don't want it included in this PR? (this would diverge with `das` branch)

jimmygchen (changes requested):
Nice, another 900+ lines down 🎉 
I've went through the code and I don't see anything impacting mainnet code path, so we should be good to go once comments are addressed and CI passes.

jimmygchen:
@dapplion Is there any refactor in this PR that you'd like to merge to `unstable`? or shall we close this one?

dapplion:
I ported a change to
- https://github.com/sigp/lighthouse/pull/6317

else nothing relevant so I'll close

Issue #6258, opened 2024-08-14
Improve range sync with PeerDAS


dapplion:
## Roadmap

1. Downscore peers for invalid data https://github.com/sigp/lighthouse/pull/7352
  - Track who send what in batches. We need change `PeerId` for `PeerGroup` which allows to track who send each column
    - Partially implement in https://github.com/sigp/lighthouse/pull/6922 which removes the tracking a `PeerId -> request` inside the batches
  - Make beacon processor errors more descriptive such that sync can know which column caused the RpcBlock to be invalid
2. Downscore peers for custody failures
  - Send requests to peers that are part of the SyncingChain peerset, instead of the global pool. This will cause SyncingChains to error frequently with `NoPeer` errors
    - Modify syncing batches to allow them to stay in the `AwaitingDownload` state when they have no peers
    - Remove `good_peers_on_sampling_subnets`
    - Extras:
      - Consider adding a fallback mechanism where we fetch from the global peer set, but only for those that are synced up to the requested batch. And in that case don't penalize custody failures
      - Assume all finalized peers are in the same chain and improve SyncingChain grouping
      - Implement StatusV2
  - Change by_range sync download algorithm to fetch blocks first, then columns. Use the blocks are source of truth to match against columns and penalize custody failures. 
    - V1: Assume block peer is honest, and believe the blocks are canonical
    - V2: Send blocks to the processor to verify correct proposer index and proposer signature (significantly increases the cost of an attack). Note: this will require way more complex logic, similar to lookup sync. Note^2: batch syncing will no longer download and process in parallel. Because we need to processing is sequential and we need to process blocks before completing a batch download.
3. Request invidual columns when a peer fails to serve the columns
  - To retry each by_range request individually we need to be able to trigger them individually https://github.com/sigp/lighthouse/pull/6497
  - Add a custody_by_range meta-request similar to [beacon_node/network/src/sync/network_context/custody.rs](https://github.com/sigp/lighthouse/blob/unstable/beacon_node/network/src/sync/network_context/custody.rs)
  - Deprecate current `block_sidecar_coupling` logic for a meta-request similar to lookup sync or custody by root.
4. Reconstruct if we can't download all columns that we need but we have >= 50%
5. Improve peer selection logic: what peer to select next for column requests? i.e. if a peer has a custody failure, should we never request from it, prevent requests to it for some time?

### Extras

6. Refactor `verify_kzg_for_rpc_blocks` outside of the da_checker, it does not use the cache
[...]

PR #6276, opened 2024-08-19
Fix PeerDAS stuck range sync when multiple columns are batched in a single request

jimmygchen:
Currently the PeerDAS range sync never evaluate range requests as finished, causing it to get stuck.

In #6004 we finally got range sync working, but one of the later commit in the PR (batching multiple columns in range sync requests) broke it again. 

It turned out that range sync assumes that each data column request only contains 1 column index, hence expecting `custody_column_indices.len()` number of stream terminations, and in this case we only have 1 stream termination from a single request (multiple columns batched), so the request was never completed. 

I've tested this locally on a devnet and it's able to sync to head.

For details of the investigation:
https://hackmd.io/@jimmygchen/H1TDSKPqR

Note that this is a short term fix to make range sync work, and we're expecting an overhaul of the custody range sync logic, see #6258.


dapplion (approved):
LGTM

PR #6282, opened 2024-08-20
Avoid redundant allocations in blob verification

michaelsproul:
## Proposed Changes

Micro-optimisations to allocate 1 vec instead of 2 during KZG blob verification, and removal of an unnecessary `clone`.

## Additional Info

I haven't benchmarked this optimisation but I think it's low-risk and pretty much a no-brainer


kevaundray (approved):

realbigsean (approved):

realbigsean:
@mergify queue

realbigsean:
@mergify requeue

PR #6287, opened 2024-08-21
Merge `unstable` into `das` 20240821

jimmygchen:
## Issue Addressed

Merging `unstable` into `das`.

There are a few incompatible changes, and requiring additional changes to existing code  on `das` branch.
- I had to change `ActiveDataColumnsByRootRequest.requester` type to `DataColumnsByRootRequester` to satisfy the `das` branch.
- `network_beacon_processor.handle_data_columns_by_root_request` on `unstable` is incompatible with `das` branch. Given this is reviewed, I've accepted the change onto `das` branch and this kills the columns by root request re-processing feature.
- [commented out](https://github.com/sigp/lighthouse/pull/6287/commits/9e5140a7b625722c51506f39a703576156e35877) the penalty for data column lookup failure. This seems like a bug

If the above workaround isn't satisfactory, please make a PR to the `das` branch.

We should avoid re-writes / refactor on `das` PRs to `unstable`, and only do refactor against the `das` branch for completeness of refactoring.

Running a local kurtosis test now and will merge this once the test and CI pass.


jimmygchen:
Manually confirmed lookup sync, range sync are working.

jimmygchen:
Everything is passing now except for the `ef_tests` which is ok because our electra code base doesn't support alpha-3 yet.

PR #5529, opened 2024-04-05
Remove VC response signing and fix HTTP error handling

eserilev:
## Issue Addressed

#4597 and #5423

## Proposed Changes

- Removed VC response signing and related code paths
- prevent warp backtracking issues by changing our usage of `and_then` to `then`

Closes #4597
Closes #5423

PR #6288, opened 2024-08-21
Remove VC response signing and update api-token in Lighthouse Book 

chong-he:
The VC signature header is removed in #5529 , hence the page https://lighthouse-book.sigmaprime.io/api-vc-sig-header.html in Lighthouse book is no longer relevant (thanks @michaelsproul for the ping). 

Also update the `api-token` to the new form


book/src/api-vc-auth-header.md
```diff
@@ -6,13 +6,13 @@ The validator client HTTP server requires that all requests have the following
 HTTP header:
 
 - Name: `Authorization`
-- Value: `Bearer <api-token>`
+- Value: `Bearer <string>`
```
michaelsproul:
```suggestion
- Value: `Bearer <api-token>`
```

I think this should match the name used below on line 11

book/src/api-vc-auth-header.md
```diff
@@ -6,13 +6,13 @@ The validator client HTTP server requires that all requests have the following
...
 
-Where `<api-token>` is a string that can be obtained from the validator client
-host. Here is an example `Authorization` header:
+Where `<api-token>` is a string of 33 characters that can be obtained from the validator client
```
michaelsproul:
```suggestion
Where `<api-token>` is a string that can be obtained from the validator client
```

The token may not be 33 characters if it is initialised by the user, or still using the old format (we can still load the old format). I don't think we need to mention this

book/src/api-vc-endpoints.md
```diff
@@ -55,10 +55,10 @@ Example Response Body:
 
 > Note: The command provided in this documentation links to the API token file. In this documentation, it is assumed that the API token file is located in `/var/lib/lighthouse/validators/API-token.txt`. If your database is saved in another directory, modify the `DATADIR` accordingly. If you are having permission issue with accessing the API token file, you can modify the header to become `-H "Authorization: Bearer $(sudo cat ${DATADIR}/validators/api-token.txt)"`.
```
michaelsproul: We could fix this while we're here:

```suggestion
> Note: The command provided in this documentation links to the API token file. In this documentation, it is assumed that the API token file is located in `/var/lib/lighthouse/validators/api-token.txt`. If your database is saved in another directory, modify the `DATADIR` accordingly. If you are having permission issue with accessing the API token file, you can modify the header to become `-H "Authorization: Bearer $(sudo cat ${DATADIR}/validators/api-token.txt)"`.
```

book/src/api-vc-endpoints.md
```diff
@@ -55,10 +55,10 @@ Example Response Body:
...
 
-> As an alternative, you can also provide the API token directly, for example, `-H "Authorization: Bearer api-token-0x02dc2a13115cc8c83baf170f597f22b1eb2930542941ab902df3daadebcb8f8176`. In this case, you obtain the token from the file `API token.txt` and the command becomes:
+> As an alternative, you can also provide the API token directly, for example, `-H "Authorization: Bearer hGut6B8uEujufDXSmZsT0thnxvdvKFBvh`. In this case, you obtain the token from the file `API token.txt` and the command becomes:
```
michaelsproul:
```suggestion
> As an alternative, you can also provide the API token directly, for example, `-H "Authorization: Bearer hGut6B8uEujufDXSmZsT0thnxvdvKFBvh`. In this case, you obtain the token from the file `api-token.txt` and the command becomes:
```

book/src/api-vc-auth-header.md
chong-he: May I know how to still load the old format of api-token?

michaelsproul: The token is just whatever is in the `api-token.txt` file. So if it is the old-style token like `api-token-$blah` then we will just use those bytes as the token

michaelsproul (approved):

michaelsproul:
@mergify queue

book/src/api-vc-auth-header.md
chong-he: ok, so if the `validators` folder are deleted and users are using 5.3.0 and above, it will load the new api-token? (the string, not the `api-token-longstring`

I thought there is a command to load the old type

Issue #6291, opened 2024-08-22
Improve beacon processor scheduling


michaelsproul:
## Description

Currently Lighthouse uses a priority-based scheduling algorithm to assign work to threads in its `BeaconProcessor`: https://github.com/sigp/lighthouse/blob/stable/beacon_node/beacon_processor/src/lib.rs

We implemented the `BeaconProcessor` primarily as a way to avoid overloading the machine with too many jobs running in parallel. Its priority system mostly works fine to ensure that high-priority work gets completed ahead of low-priority work, it sometimes struggles to _ever_ complete low priority work (_starvation_). If high-prio work keeps coming in, it can indefinitely postpone the processing of low-priority work.

Another issue is that the priorities are currently determined per task and follow a strict ordering. E.g. processing blocks is higher prio than processing attestations. In some cases it's unclear exactly what the task priority ordering should be and we end up having to choose somewhat arbitrarily (e.g. when testing https://github.com/sigp/lighthouse/pull/5481 I obtained some benefit from prioritising `status` messages higher in the ordering). For API requests this is a particular issue because we have two priorities for requests: `P0` which are processed ahead of almost everything, and `P1` which are basically bottom of the barrel and won't be run until the processor has some spare cycles.

## Version

Lighthouse v5.3.0

## Steps to resolve

I would like to abstract the actual scheduling logic out so we can experiment with different algorithms. This will require untangling all of the concerns that are currently mixed into the beacon processor, e.g. backfill rate limiting which gets its own special treatment here:

https://github.com/sigp/lighthouse/blob/d6ba8c397557f5c977b70f0d822a9228e98ca214/beacon_node/beacon_processor/src/lib.rs#L863-L903

It might be prudent to start by just refactoring the current scheduling algorithm into a cleaner and more modular form, and then we can go about tweaking it.

I think the properties we desire from a scheduling algorithm are:

- Prioritisation based on task type (retain some of what we have now).
- Prioritisation based on how long a task has been waiting to run, and/or how close it is to its _deadline_ (see https://en.wikipedia.org/wiki/Earliest_deadline_first_scheduling).
- High throughput? Ideally we should be able to max out the CPU if we are running N CPU-bound tasks.
- Improved fairness.

Some of these desires are slightly in conflict, so having the ability to try multiple different backends is advantageous.

[...]

PR #6292, opened 2024-08-22
Increase priority for validator HTTP requests

michaelsproul:
## Issue Addressed

Address an issue reported by Stakely during Obol testing (for Lido) whereby [`charon`](https://github.com/ObolNetwork/charon) gets stuck in a timeout loop trying to request the details of inactive validators from Lighthouse.

The error from Charon is:

> 14:51:30.489 ERRO vapi       Validator api 5xx response: fetching non-cached validators from BN: beacon api validators: http request timeout: context deadline exceeded {"status_code": 500, "message": "Internal server error", "duration": "2.001067392s", "label": "validators", "url": "http://beaconnode:5051/eth/v1/beacon/states/head/validators?id=XXX", "method": "Get", "vapi_endpoint": "get_validator"}
        app/eth2wrap/eth2wrap.go:206 .wrapError
        app/eth2wrap/eth2wrap_gen.go:648 .Validators
        core/validatorapi/validatorapi.go:1021 .Validators
       core/validatorapi/router.go:1372 .getValidatorsByID
       core/validatorapi/router.go:388 .func7
       core/validatorapi/router.go:311 .func1
        app/app.go:954 .func2

## Proposed Changes

This PR increases the priority for validator info requests from P1 to P0, until we can do the hard work of overhauling the priority system:

- https://github.com/sigp/lighthouse/issues/6291

This is a low-risk change, but will require a new LH release before it can be used in production by Obol users. I will try to get Stakely to run it on testnets.

## Additional Info

There is a workaround that achieves a similar effect: `--http-enable-beacon-processor false`. However this comes with the downside of opening Lighthouse's HTTP API up to accidental DoS. With no limit to the number of concurrent requests, it is easy to overwhelm Lighthouse with requests and cause OOMs and other slowness.

### DIFF
--- a/beacon_node/http_api/src/lib.rs
+++ b/beacon_node/http_api/src/lib.rs
@@ -714,7 +714,14 @@ pub fn serve<T: BeaconChainTypes>(
              task_spawner: TaskSpawner<T::EthSpec>,
              chain: Arc<BeaconChain<T>>,
              query_res: Result<api_types::ValidatorsQuery, warp::Rejection>| {
-                task_spawner.blocking_json_task(Priority::P1, move || {
+                // Prioritise requests for validators at the head. These should be fast to service
+                // and could be required by the validator client.
+                let priority = if let StateId(eth2::types::StateId::Head) = state_id {
+                    Priority::P0
+                } else {
+                    Priority::P1
+                };
+                task_spawner.blocking_json_task(priority, move || {
                     let query = query_res?;
                     crate::validators::get_beacon_state_validators(
                         state_id,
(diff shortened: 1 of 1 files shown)


jimmygchen (approved):
Agree this is a low risk change. LGTM.

How often does charon query the BN for validators? My concern is if there is a large number of inactive validators, it could overwhelm the BN and making it `P0` could make things worse.

[We've had performance issue with this endpoint before](https://github.com/sigp/lighthouse/issues/4388) - and changed our poll frequency to once per epoch. Our VC also avoid polling on the first slot of the epoch (change made in #5628)

michaelsproul:
I don't think Charon sets the poll frequency because it is just middleware between the VC and the BN. So Lighthouse VC is in control of the polling schedule and should only be making requests once per epoch as of v5.2.0 (when https://github.com/sigp/lighthouse/pull/5628 was included). I've asked Stakely to confirm which version of Lighthouse VC they were running. If they were running v5.1.3 or earlier then it will have been spamming every slot and making the problem worse.

I also don't think the API performance was too bad here. We quite regularly hit 2s+ response times for P1 requests on Holesky, and this is usually fine for non-critical requests. The problem is that Charon times out after 2s and spews error logs. Because the validators are not active, Lighthouse VC (through Charon) will keep requesting, and even if some of these requests succeed, the error logs can come back the next time a request times out. So my original description of a "timeout loop" is not quite accurate: charon is likely just timing out repeatedly.

Stakely reported that the issue disappeared completely when running with `--http-enable-beacon-processor false`, which to me also points to a scheduling issue around `P1` more than a general performance issue. I think this also implies that the BN can handle these requests at `P0` which is essentially what `--http-enable-beacon-processor false` does (makes _every_ request `P0` with no rate-limiting).

michaelsproul:
@mergify queue

PR #3849, opened 2023-01-04
Support `LightClientFinalityUpdate` and `LightClientOptimisticUpdate` rpcs

GeemoCandama:
## Issue Addressed

partially addresses #3651.

## Proposed Changes

- Support `LightClientFinalityUpdate` https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/p2p-interface.md#getlightclientfinalityupdate
- Support `LightClientOptimisticUpdate` https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/p2p-interface.md#getlightclientoptimisticupdate

PR #6299, opened 2024-08-22
Light Client Bug Fix

ethDreamer:
Came across this while implementing `eip-7732` and it looks like a bug to me. Was [introduced here](https://github.com/sigp/lighthouse/pull/3849) but I'm not familiar with this contributor.

### DIFF
--- a/consensus/types/src/light_client_header.rs
+++ b/consensus/types/src/light_client_header.rs
@@ -130,8 +130,8 @@ impl<E: EthSpec> LightClientHeader<E> {
 
     pub fn ssz_max_var_len_for_fork(fork_name: ForkName) -> usize {
         match fork_name {
-            ForkName::Base | ForkName::Altair | ForkName::Bellatrix => 0,
-            ForkName::Capella | ForkName::Deneb | ForkName::Electra => {
+            ForkName::Base | ForkName::Altair => 0,
+            ForkName::Bellatrix | ForkName::Capella | ForkName::Deneb | ForkName::Electra => {
                 ExecutionPayloadHeader::<E>::ssz_max_var_len_for_fork(fork_name)
             }
         }


consensus/types/src/light_client_header.rs
```diff
@@ -130,8 +130,8 @@ impl<E: EthSpec> LightClientHeader<E> {
 
     pub fn ssz_max_var_len_for_fork(fork_name: ForkName) -> usize {
         match fork_name {
-            ForkName::Base | ForkName::Altair | ForkName::Bellatrix => 0,
-            ForkName::Capella | ForkName::Deneb | ForkName::Electra => {
+            ForkName::Base | ForkName::Altair => 0,
+            ForkName::Bellatrix | ForkName::Capella | ForkName::Deneb | ForkName::Electra => {
```
dapplion, line 134: Switch to bellatrix_enabled()

ethDreamer: In this particular case, I *think* this may be a bad idea. Presumably the fork after `Electra` might include [`EIP-7732`](https://eips.ethereum.org/EIPS/eip-7732). In this EIP (at least at the moment), the `ExecutionPayloadHeader` doesn't exist. Note that what it calls the `ExecutionPayloadHeader` is actually a bid rather than a payload header (working on renaming it). Still working on understanding how light clients work in this model, but either way, our attention needs to be drawn here when we do implement it.

dapplion: Spec tests will fail if we don't implement the right logic for future forks. The point of `bellatrix_enabled` is to prevent boilerplate churn when others develop new fork features not only us

realbigsean: aren't these just used in RPC limit calculations? I don't think those are covered in spec tests. this arm of the match statement currently assumes no other variable length fields are added to a light client header which doesn't seem like a safe assumption, so I think mark's right we should consider it every fork

michaelsproul (approved):

michaelsproul:
@mergify queue

Issue #5900, opened 2024-06-06
Implement Clap derive


eserilev:
This issue will track the tasks for migrating Lighthouse from clap builder to derive.

Clap derive offers some nice features including
- type casted flag definitions, no need to do any post processing to the flag values i.e. converting strings to other types
- easier to read flag definitions

Clap derive is a bit less flexible than the builder, but  its possible to combine the builder with derive if needed. All of our flag definitions today are fairly "standard", so derive shouldn't cause any issues. 

Another nice bonus with clap derive is it should make ingesting a config at startup easy. Its fairly trivial to map a json,yaml,whatever file to a struct. And the clap derive flag definitions are structs. An example might make things a bit clearer.

Take this function that parses cli arguments and returns some config that Lighthouse then uses to do _stuff_

```
fn parse_client_config<E: EthSpec>(
    cli_args: &ArgMatches,
    _env: &Environment<E>,
) -> Result<ClientConfig, String>
```
With clap derive, instead of passing in the cli args, we pass in the struct that defines those cli args

```
fn parse_client_config<E: EthSpec>(
    database_manager_config: &DatabaseManager,
    _env: &Environment<E>,
) -> Result<ClientConfig, String>
```

It then becomes straight foward to write some code that maps a provided yaml,json,etc file to an instance of `DatabaseManager` and feed that into the same `parse_client_config` function. Were now re-using the same code paths that ingest cli params from the command line to ingest cli params from a file.

### Tasks

- [x] allow for mixing clap derive subcommands within a builder parent command #5898
- [x] database manager #5898
- [ ] beacon node
- [x] validator client #6300
- [ ] account manager #6493
- [ ] validator manager @macladson 
- [ ] migrate lighthouse `main` to derive
- [ ] Ingest config file at start up

michaelsproul:
Adding @macladson as he is keen to work on this to unblock:

- https://github.com/sigp/lighthouse/issues/7135

I've assigned him to `validator-manager`

eserilev:
sounds good

@macladson if its helpful I have this branch here
https://github.com/eserilev/lighthouse/tree/upgrade-clap

its pretty old but it fully migrates lighthouse to clap derive.  What I usually do is copy the clap changes from that branch, run `make cli-local` to generate the book changes and then use the book diff to make whatever additional changes. just need to be careful with the hidden flags

PR #6300, opened 2024-08-23
Migrate validator client to clap derive

eserilev:
## Issue Addressed

Partially #5900

## Proposed Changes

Migrate the validator client cli to clap derive


validator_client/src/config.rs
```diff
@@ -261,32 +267,27 @@ impl Config {
...
-                Some(Duration::from_millis(
-                    s.parse().map_err(|_| "invalid timeout value".to_string())?,
-                ))
-            }
+        if validator_client_config.web3_signer_keep_alive_timeout == 0 {
+            config.web3_signer_keep_alive_timeout = None
+        } else {
+            config.web3_signer_keep_alive_timeout = Some(Duration::from_millis(
+                validator_client_config.web3_signer_keep_alive_timeout,
+            ));
```
eserilev: Note the breaking change here. Since I've defined `web3_signer_keep_alive_timeout` as a `u64`, `null` is no longer a valid option. Users will need to use `0` instead. 

If this breaking change is too intrusive I can define `web3_signer_keep_alive_timeout` as a string and we can keep functionality as is.

validator_client/src/cli.rs
```diff
@@ -1,509 +1,498 @@
...
+
+    #[clap(
+        long,
+        value_name = "ADDRESS",
+        default_value_if("metrics", ArgPredicate::IsPresent, "127.0.0.1"),
+        help = "Set the listen address for the Prometheus metrics HTTP server.",
+        display_order = 0
+
+    )]
+    pub metrics_address: IpAddr,
```
macladson: For some reason, the CLI thinks that `--metrics-address` is a required flag, even when `--metrics` is provided.

validator_client/src/cli.rs
```diff
@@ -1,509 +1,498 @@
...
-            Arg::new("metrics-allow-origin")
-                .long("metrics-allow-origin")
-                .requires("metrics")
-                .value_name("ORIGIN")
-                .help("Set the value of the Access-Control-Allow-Origin response HTTP header. \
+    styles = get_color_style(),
+    next_line_help = true,
+    term_width = 80,
+    disable_help_flag = true,
+    disable_help_subcommand = true,
```
macladson, line 20: I understand we are matching the syntax from the `DatabaseManager` but just curious what the reason is for disabling the in-built help flag?

Also if we do disable it, we'll need our own version like in the `DatabaseManager`, see `database_manager/src/cli.rs:66`

macladson (commented):
Pretty keen to get the clap derive migrations merged!

Will give it a more thorough review once the conflicts are fixed.

validator_client/src/cli.rs
eserilev: we're disabling the built in help flag because we want it to be displayed under the Flags section via 
  `.help_heading(FLAG_HEADER)`. By default the help flag isdisplayed under Options, which is where we list cli params that take an argument. Also we need `display_order(0)` so that its displayed in alphabetical order in the Flags section

eserilev: i've redefined the help flag as global in the parent lighthouse cli flag definition and was able to get rid of it here

lighthouse/tests/validator_client.rs
```diff
@@ -430,7 +437,7 @@ fn no_doppelganger_protection_flag() {
 fn no_gas_limit_flag() {
     CommandLineTest::new()
         .run()
-        .with_config(|config| assert!(config.gas_limit.is_none()));
+        .with_config(|config| assert!(config.gas_limit == Some(30_000_000)));
```
macladson: Is there a functional difference between `None` and `30_000_000` here? Might be worth now renaming this test to `gas_limit_flag_default` to better match the naming scheme.

validator_client/src/config.rs
```diff
@@ -377,55 +346,33 @@ impl Config {
...
-        }
-
-        if cli_args.get_flag("builder-proposals") {
-            config.builder_proposals = true;
-        }
+        config.enable_doppelganger_protection =
+            validator_client_config.enable_doppelganger_protection;
+        config.builder_proposals = validator_client_config.builder_proposals;
+        config.prefer_builder_proposals = validator_client_config.prefer_builder_proposals;
+        config.gas_limit = Some(validator_client_config.gas_limit);
```
macladson: If gas limit now defaults to `Some`, can we remove the `Option` altogether?

macladson (commented):
Looks like some asserts in the tests need to be updated to match.

validator_client/src/config.rs
jimmygchen: I think this is probably ok, not sure if anyone actually remove the web3signer timeout? (cc @michaelsproul)
I've added `backwards-incompat` label.

validator_client/src/cli.rs
```diff
@@ -1,478 +1,466 @@
...
+    )]
+    pub proposer_nodes: Option<Vec<String>>,
+
+    #[clap(
+        long,
+        value_name = "API_TOPICS",
+        value_delimiter = ',',
+        help = "Comma-separated list of beacon API topics to broadcast to all beacon nodes. \
+                Possible values are: none, attestations, blocks, subscriptions, \
+                sync-committee. Default (when flag is omitted) is to broadcast \
```
jimmygchen: Possible values are printed twice in help, this can now be removed:

```suggestion
        help = "Comma-separated list of beacon API topics to broadcast to all beacon nodes. \
                Default (when flag is omitted) is to broadcast \
```

lighthouse/src/main.rs
```diff
@@ -395,10 +395,10 @@ fn main() {
             .action(ArgAction::HelpLong)
             .display_order(0)
             .help_heading(FLAG_HEADER)
+            .global(true)
```
jimmygchen, line 402: Does this mean we can remove the "help" arg definition in the subcommands?

validator_client/src/cli.rs
```diff
@@ -1,478 +1,466 @@
...
+        long,
+        help = "Enable the Prometheus metrics HTTP server. Disabled by default.",
+        display_order = 0,
+        help_heading = FLAG_HEADER
+    )]
+    pub metrics: bool,
+
+    #[clap(
+        long,
+        value_name = "ADDRESS",
```
jimmygchen, line 262: we still need `requires = "metrics` here right?

validator_client/src/cli.rs
```diff
@@ -1,478 +1,466 @@
...
+        default_value_if("metrics", ArgPredicate::IsPresent, "127.0.0.1"),
+        help = "Set the listen address for the Prometheus metrics HTTP server. [default: 127.0.0.1]",
+        display_order = 0
+    )]
+    pub metrics_address: Option<String>,
+
+    #[clap(
+        long,
+        value_name = "PORT",
```
jimmygchen, line 272: same here, `requires = "metrics"`

validator_client/src/cli.rs
```diff
@@ -1,478 +1,466 @@
...
+        default_value_t = 5064,
+        help = "Set the listen TCP port for the Prometheus metrics HTTP server.",
+        display_order = 0
+    )]
+    pub metrics_port: u16,
+
+    #[clap(
+        long,
+        value_name = "ORIGIN",
```
jimmygchen, line 282: same here, requires = "metrics"

validator_client/src/cli.rs
```diff
@@ -1,478 +1,466 @@
...
+
+    #[clap(
+        long,
+        value_name = "INTEGER",
+        default_value_t = 30_000_000,
+        help = "The gas limit to be used in all builder proposals for all validators managed \
+                by this validator client. Note this will not necessarily be used if the gas limit \
+                set here moves too far from the previous block's gas limit.",
+        display_order = 0
+    )]
```
jimmygchen, line 374: `.requires("builder-proposals")` is no longer here, is it intentional?

validator_client/src/cli.rs
```diff
@@ -1,478 +1,466 @@
...
+        default_value_t = 60,
+        help = "Defines how many seconds to wait between each message sent to \
+                the monitoring-endpoint.",
+        display_order = 0
+    )]
+    pub monitoring_endpoint_period: u64,
+
+    #[clap(
+        long,
+        value_name = "ENABLE_DOPPELGANGER_PROTECTION",
```
jimmygchen: Should we go with 
`value_name = "BOOLEAN"` for consistency with some of the other flags?

validator_client/src/cli.rs
```diff
@@ -1,478 +1,466 @@
...
+    pub validator_registration_batch_size: usize,
+
+    #[clap(
+        long,
+        value_name = "UINT64",
+        help = "Defines the boost factor, \
+                a percentage multiplier to apply to the builder's payload value \
+                when choosing between a builder payload header and payload from \
+                the local execution node.",
+        display_order = 0
```
jimmygchen, line 405: Do we still need `.conflicts_with("prefer-builder-proposals")`?

jimmygchen (changes requested):
Looks good overall! I've added some comments, let me know what you think.

validator_client/src/cli.rs
eserilev: yes! i've re-addd that, thanks

validator_client/src/cli.rs
eserilev: makes sense to me

validator_client/src/cli.rs
eserilev: not intentional, ive re-added it

lighthouse/src/main.rs
eserilev: yes good point, i've gone ahead removed the "help" arg definitions in all the subcommands

eserilev:
thanks for the review Jimmy, I've pushed up changes based on your feedback 
02e305a

jimmygchen (approved):
Thanks @eserilev, looks great!

michaelsproul:
Maybe we should merge this PR first, as it will conflict:

- https://github.com/sigp/lighthouse/pull/6453

validator_client/src/config.rs
michaelsproul: Yeah this is probably obscure enough that we would be OK to change it. It's a bit dodgy because it flips the meaning of `0` from "timeout immediately" to "never timeout", but I doubt anyone is using 0

michaelsproul:
Age's PR is merged now. Some conflicts to resolve and then we can merge this.

eserilev:
merge conflicts have been resolved

macladson (approved):
LGTM!

dapplion (approved):
All flags match to current unstable

PR #6307, opened 2024-08-26
Add test for ActiveSamplingRequest

ackintosh:
## Issue Addressed

I've implemented batching sampling requests in #6256, but there are no unit tests for it yet.

## Proposed Changes

Add a test for ActiveSamplingRequest to ensure that the requests are batched, and check the status of the requests.


michaelsproul:

ackintosh:
Thanks for the comment Michael! I've rebased this on `unstable`.

beacon_node/network/src/sync/sampling.rs
```diff
@@ -626,3 +635,257 @@ mod request {
...
+                    signature: Signature::empty(),
+                },
+                kzg_commitments_inclusion_proof: FixedVector::default(),
+            }));
+        }
+        (list, Duration::default())
+    }
+
+    #[test]
+    fn test_active_sampling_request() {
```
dapplion: Can you write tests in this format and file?  https://github.com/sigp/lighthouse/blob/5a966874da03cd987fef1629f732898a2859ce87/beacon_node/network/src/sync/block_lookups/tests.rs#L2011

We recently avoided adding unit tests too close to impl for event-based tests

ackintosh:
Thanks @dapplion! I've written tests in the tests.rs file.

jimmygchen (approved):
Nice test, very tidy 🎉

jimmygchen:
@mergify queue

PR #6308, opened 2024-08-27
Improve `get_custody_columns` validation, caching and error handling

jimmygchen:
## Issue Addressed

This PR addresses two issues:
- Add validation to `get_custody_columns` to prevent infinite loop when supplied custody subnet count is invalid (> total number of data column subnets). 
- Refactor error handling of `network_globals.custody_columns` to the caller site (see @pawanjay176's comment [here](https://github.com/sigp/lighthouse/pull/6271#discussion_r1723676913))
- Store the node's custody subnets and columns on `NetworkGlobals` initialization rather than computing it every time.

### Error handling 

For node's `custody_subnet_count` from local metadata:
- if custody subnet count is not set or invalid after PeerDAS is enabled, we just panic on startup - this indicates a bug in the code, and we should detect this immediately on testing

For peer's `custody_subnet_count` from ENR & metadata:
- if `custody_subnet_count` is invalid, fall back to `custody_requirement`
- I think it's fine to not penalise peer, if they can still serve the columns from `custody_requirement`, they are still useful to us


beacon_node/lighthouse_network/src/discovery/subnet_predicate.rs
```diff
@@ -30,23 +28,23 @@ where
...
+            // If we're unable to get custody subnet count from ENR, fallback to
+            // `custody_requirement`.
+            // If we're unable to compute custody subnets given a custody subnet count, return `false`.
+            Subnet::DataColumn(s) => DataColumnSubnetId::compute_custody_subnets::<E>(
+                enr.node_id().raw().into(),
+                enr.custody_subnet_count::<E>()
+                    .unwrap_or_else(|_| spec.custody_requirement),
+                &spec,
+            )
+            .map_or(false, |mut subnets| subnets.contains(s)),
```
jimmygchen: I think it's fine to default to `false` here - it's consistent behaviour with the attestation and sync committee bitfield, and it filters out potential bad peers during discovery.

jimmygchen:
I don't really like the complexity in error handling here, will work on simplifying it tomorrow.

beacon_node/lighthouse_network/src/discovery/subnet_predicate.rs
jimmygchen: outdated

jimmygchen:
@pawanjay176 I've re-written the PR and I think it's in a better place now and ready for review.

beacon_node/lighthouse_network/src/discovery/subnet_predicate.rs
```diff
@@ -40,12 +37,20 @@ where
                 .as_ref()
                 .map_or(false, |b| b.get(*s.deref() as usize).unwrap_or(false)),
             Subnet::DataColumn(s) => {
-                let mut subnets = DataColumnSubnetId::compute_custody_subnets::<E>(
+                // If ENR does not contain a valid csc, we fall back to `custody_requirement` here.
```
dapplion: I'm divided on this. Let's consider mainnet post-peerdas activation and there's a widespread ENR encoding bug:
- A (if we error): nodes won't be able to connect to each other, we would notice immediately, but risks liveness as you can't peer up
- B (if we fallback): all nodes will appear as regular nodes (no super-nodes). It is not immediately obvious as the only metric would be a decrease in the count of supernodes (**TODO**: add a metric for this).

The non-discoverability of B seems worrying to me. Due to the current multi-client architecture of Ethereum the liveness risks of A don't appear so severe. Issues in ENR encoding should become apparent quickly in any devnet run.

I would prefer to go for A if I'm not missing something

beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
```diff
@@ -1166,6 +1160,47 @@ impl<E: EthSpec> PeerDB<E> {
...
+        // This handles the scenario where `csc` is not found or not decodable.
+        // TODO(das): Should we penalise peer?
+        debug!(
+            log,
+            "No valid csc found in peer ENR";
+            "info" => "Falling back to default custody requirement",
+            "peer_id" => %peer_id,
+            "enr" => enr.to_base64(),
+            "error" => ?e
+        );
```
dapplion: Both of this debug statements should also increase a metric

beacon_node/lighthouse_network/src/peer_manager/peerdb/peer_info.rs
```diff
@@ -504,6 +501,50 @@ impl<E: EthSpec> PeerInfo<E> {
...
+        metadata: &MetaData<E>,
+        spec: &ChainSpec,
+        log: &Logger,
+    ) -> HashSet<DataColumnSubnetId> {
+        let node_id = node_id.raw().into();
+
+        // fallback to `custody_requirement` if `csc` is invalid
+        let custody_subnet_count = match metadata.custody_subnet_count() {
+            Ok(csc) => *csc,
+            Err(_) => {
```
dapplion: This Err condition will only occur if MetaData is not a V3 variant. Invalid values will trigger an error on the ReqResp SSZ decoding step. Can we ensure that when calling this function for a peer we have already issued a metadata/v3 request? I think here we can default to `custody_requirement` without any log as it's our fault to not issue a `metadata/v3`. Or maybe signal this fact? i.e. `Expected MetaDataV3 for peer`

beacon_node/lighthouse_network/src/peer_manager/peerdb/peer_info.rs
```diff
@@ -504,6 +501,50 @@ impl<E: EthSpec> PeerInfo<E> {
...
+                spec.custody_requirement
+            }
+        };
+
+        DataColumnSubnetId::compute_custody_subnets::<E>(node_id, custody_subnet_count, spec)
+            .map(|subnets| subnets.collect())
+            .unwrap_or_else(|e| {
+                // This handles the scenario where a peer supplies an invalid `csc` value,
+                // greater than `data_column_sidecar_subnet_count`.
+                // TODO(das): Should we penalise peer?
```
dapplion: I would say so, there's no reason why a peer would set a `csc > data_column_sidecar_subnet_count`. Also add a metric as in the ENR path

consensus/types/src/data_column_subnet_id.rs
```diff
@@ -66,17 +69,37 @@ impl DataColumnSubnetId {
...
+            .map(Either::Left)
+            .unwrap_or_else(|e| {
+                error!(
+                    log,
+                    "Failed to compute default custody subnets for node";
+                    "node_id" => %node_id,
+                    "info" => "This is a bug (unreachable case), please notify the devs",
+                    "error" => ?e
+                );
+                Either::Right(std::iter::empty())
```
dapplion: Why not panic here? You can add a unit test ensuring that `custody_requirement < data_column_sidecar_subnet_count` in all our known specs. If someone tries to run a network with broken config the panic should be ok.

beacon_node/lighthouse_network/src/discovery/subnet_predicate.rs
jimmygchen: We could add a debug log here - however we can't return an error here without changing discv5 code I think, this predicate function expects a `bool` return and  is used in discv5 predicate query [here](https://github.com/sigp/discv5/blob/682b51e317216b1e2ec3d75b541cdff66d27cf32/src/discv5.rs#L691-L699).

We could return `false` to ignore the peer, but IMO if they could serve column requests they're assigned to custody, they're still useful.

> Issues in ENR encoding should become apparent quickly in any devnet run.

There could still be client changes after devnets / testnets - and adding errors / warnings due to peer issues would be confusing for end users as there's nothing they can do to resolve this.

consensus/types/src/data_column_subnet_id.rs
jimmygchen: We probably could. My thinking was because `compute_custody_subnets` return `Error` type in the error case, it could return other variants from the API perspective - although the current implementation could only fail due to `InvalidCustodySubnetCount`, which _could_ change. I usually prefer to avoid assumptions on implementation but in this case this is an implementation on the same struct, so I guess it's fine.

beacon_node/lighthouse_network/src/peer_manager/peerdb/peer_info.rs
jimmygchen: If there's a bug in their csc encoding, but they can still serve data we want, why wouldn't we want to maintain the connection?

consensus/types/src/data_column_subnet_id.rs
jimmygchen: I'm slightly concerned if this is somehow not caught on startup - e.g. all devnet peers provide their `csc` override and we never use the default. I'll add an unit test to this function.

jimmygchen: Added test here. https://github.com/sigp/lighthouse/pull/6308/commits/763ea039b04a5b47918adcd28634edae9165392c

beacon_node/lighthouse_network/src/peer_manager/peerdb/peer_info.rs
jimmygchen: Good catch. I've added a check for `is_peer_das_scheduled` before calling this function, and add `Expected MetaDataV3` to the log. 

I think it should be an error log if it's a bug though.

beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
jimmygchen: Is this really necessary?

jimmygchen: so we get a count of how many peers have invalid / missing csc in their ENR? Does this metrics help us in anyway other than trying to catch a bug? 

I feel like a log is suffice if we just want to catch a bug.

beacon_node/lighthouse_network/src/peer_manager/peerdb/peer_info.rs
jimmygchen: Updated in https://github.com/sigp/lighthouse/pull/6308/commits/e8b9d769ac92d03565f8b1ec03fbfd68c2f987f8

beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
jimmygchen: I've deleted the computation from ENR code in peerdb after a chat with @AgeManning - it's best to avoid doing error handling / peer reporting logic in `PeerDB::update_connection_state`, as the function focuses on handling various connection state transitions, and is not the right place for this.

I've updated the subnet predicate to not perform any fallback and just return `false` in the error case, so in theory we shouldn't be connecting to these peers with bad `csc` values via discovery. 

We can just rely on metadataV3 for `csc` as we always get them from peers and this should simplify things.

beacon_node/lighthouse_network/src/peer_manager/peerdb/peer_info.rs
jimmygchen: I'm reworking this function to add kicking peer on failure.

We already have metrics on number of peers by subnet. What metric are you after? Is it for debugging purpose, and logs aren't sufficient for this purpose?

Just wanted to be mindful when adding new metrics, because we'd have to think about maintaining compatibility once end users start using them, so we should only add them if we think they're going to stay and be useful going forward.

dapplion: RE metric: how many supernodes are we connected to

beacon_node/lighthouse_network/src/peer_manager/peerdb/peer_info.rs
```diff
@@ -362,25 +363,20 @@ impl<E: EthSpec> PeerInfo<E> {
...
-            if let Ok(custody_subnet_count) = meta_data.custody_subnet_count() {
-                let custody_subnets = DataColumnSubnetId::compute_custody_subnets::<E>(
-                    node_id.raw(),
-                    std::cmp::min(*custody_subnet_count, spec.data_column_sidecar_subnet_count),
-                    spec,
-                )
-                .collect::<HashSet<_>>();
-                self.set_custody_subnets(custody_subnets);
-            }
+        if spec.is_peer_das_scheduled() {
```
dapplion: Even if peerdas is scheduled, our peer may not support metadata/v3, so on the reqresp protocol negotiation we will issue a metadata/v2. Can this lead to a lot of error logs when users progressively update to a peerdas version?

We should gracefully ignore metadata/v2 peers in this function. Potentially downscore after peerdas to prioritize peerdas peers.

We can track how many of our peers are "peerdas ready" too. Expand the peer count metrics to:
- count of peers that indicate some custody count

or 

- histogram that tracks the current count of peers bucketed by custody count (sum all labels and get the metric above)

beacon_node/lighthouse_network/src/discovery/subnet_predicate.rs
jimmygchen: I actually ended up with A, but returning `false` instead of an error, if their ENR is bad we just don't connect to them and find other peers instead. 

> Issues in ENR encoding should become apparent quickly in any devnet run.

I think what you said make sense, if we're discovering and not finding peers, it'd be a bug on our side and should be apparent. If we start ignoring peers with invalid ENRs, it's a issue on the client version they're using and it's fine for us to not connect to them at all.

beacon_node/lighthouse_network/src/peer_manager/peerdb/peer_info.rs
jimmygchen: That sounds like an useful metrics, I'll add that.

beacon_node/lighthouse_network/src/peer_manager/peerdb/peer_info.rs
jimmygchen: Good point, and nice idea about the metrics - I think the histogram will be useful - and we can also use it to identify number of supernodes vs full nodes.

jimmygchen: Add int gauge here https://github.com/sigp/lighthouse/pull/6308/commits/d82c806accd87025672bb7c7664491828702088e

jimmygchen: The `peers_per_custody_subnet_count` int gauge has been added, and can be used to track peer custody subnet counts, number of supernodes, and number of peerdas ready nodes

jimmygchen: Add handling of metadata/v2 responses (gracefully ignore) in https://github.com/sigp/lighthouse/pull/6308/commits/067307a5852f8d8fd3db4716e31bb54533fa0822

beacon_node/lighthouse_network/src/peer_manager/peerdb/peer_info.rs
jimmygchen: Moved the computation and reporting of peers to `PeerManager` and disconnect peer if metadata is invalid.
https://github.com/sigp/lighthouse/pull/6308/commits/067307a5852f8d8fd3db4716e31bb54533fa0822

beacon_node/lighthouse_network/src/metrics.rs
```diff
@@ -91,6 +91,15 @@ pub static PEERS_PER_CLIENT: LazyLock<Result<IntGaugeVec>> = LazyLock::new(|| {
         &["Client"],
     )
 });
+
+pub static PEERS_PER_CUSTODY_SUBNET_COUNT: LazyLock<Result<IntGaugeVec>> = LazyLock::new(|| {
+    try_create_int_gauge_vec(
+        "peers_per_custody_subnet_count",
+        "The current count of peers by custody subnet count",
+        &["custody_subnet_count"],
```
dapplion, line 99: I would consider doing a histogram here as we can have buckets instead of exact values. Say we have a peer with csc 128 and another with 127, can grafana bucket these two labels into the same range?

beacon_node/lighthouse_network/src/peer_manager/mod.rs
```diff
@@ -1348,6 +1393,45 @@ impl<E: EthSpec> PeerManager<E> {
...
+
+        let custody_subnets = DataColumnSubnetId::compute_custody_subnets::<E>(
+            node_id.raw(),
+            custody_subnet_count,
+            spec,
+        )
+        .map(|subnets| subnets.collect())
+        .unwrap_or_else(|e| {
+            // This is an unreachable scenario unless there's a bug, as we've validated the csc
+            // just above.
```
dapplion, line 1420: If it's unreachable, why not expect?

beacon_node/lighthouse_network/src/metrics.rs
jimmygchen: Yeah I think buckets are useful, but historgrams _accumulate_ observersations over time and it doesn't have the ability to track the **current count** for each bucket at a specific moment. I feel like gauge is more suitable but if you have a way to represent the metric in histogram I'm interested to hear.

beacon_node/lighthouse_network/src/peer_manager/mod.rs
jimmygchen: This may be unreachable currently but there's no guarantee that it will never be.

I'd prefer to avoid `.expect` and `.unwrap` unless it's something we can  to detect on startup immediately (e.g. usage in `spec` is fine) and is not something we can recover from.

This is not a completely unrecoverable situation and we should always try to continue gracefully. We capture this event as an error anyway, so it'd still be useful for debugging without unnecessarily killing the client.

dapplion (approved):
LGTM!

beacon_node/lighthouse_network/src/metrics.rs
dapplion: In Lodestar we reset the histogram buckets before every scrape. It feels weird but renders nicely on Grafana.

jimmygchen: I think it's a good idea. I've tried it and unfortunately couldn't find a way to set it with the `prometheus` crate though.

Closest thing that I could find was `histogram.metric().clear_histogram();` but it doesn't do anything to the actual metric.

I'm going to merge this now, if we find any better way to do it, happy to do it in a follow up PR.

jimmygchen:
@mergify queue

PR #6309, opened 2024-08-27
chore!: Update rust_eth_kzg to v0.5.1 (formally known as peerdas-kzg)

kevaundray:
## Issue Addressed

v0.5.0 was published with the following changes: https://github.com/crate-crypto/rust-eth-kzg/pull/161

From the list of breaking changes in that link, the ones relevant to downstream rust users are:

- Rename recover_cells_and_proofs -> recover_cells_and_kzg_proofs
- Add a ThreadCount enum for specifying the number of threads to use 
- Allow downstream users to avoid the precomputations
- Add compile time flags to disable rayon

## Proposed Changes

- The library by default runs on a single thread, so one no longer specifies the threadcount.
- update `recover_cells_and_kzg_proofs`

## Additional Info

Rayon's global threadpool and local threadpool seems to have issues -- if we were to enable multithreading it would be using the global threadpool. 

In a future PR, I think we would want to re-enable multithreading as we are experimenting with a branch which allows us to parallelize across blobs. As a corollary, it reduces the precomputation by 50x and is faster on a single thread (the speedup numbers are still being looked into, currently it seems to be ~20%)

### DIFF
--- a/crypto/kzg/benches/benchmark.rs
+++ b/crypto/kzg/benches/benchmark.rs
@@ -11,9 +11,13 @@ pub fn bench_init_context(c: &mut Criterion) {
 
     c.bench_function(&format!("Initialize context rust_eth_kzg"), |b| {
         b.iter(|| {
-            const NUM_THREADS: usize = 1;
             let trusted_setup = PeerDASTrustedSetup::from(&trusted_setup);
-            DASContext::with_threads(&trusted_setup, NUM_THREADS)
+            DASContext::new(
+                &trusted_setup,
+                rust_eth_kzg::UsePrecomp::Yes {
+                    width: rust_eth_kzg::constants::RECOMMENDED_PRECOMP_WIDTH,
+                },
+            )
         })
     });
     c.bench_function(&format!("Initialize context c-kzg (4844)"), |b| {
--- a/crypto/kzg/src/lib.rs
+++ b/crypto/kzg/src/lib.rs
@@ -73,6 +73,9 @@ impl Kzg {
         // json string.
         let peerdas_trusted_setup = PeerDASTrustedSetup::from(&trusted_setup);
 
+        // It's not recommended to change the config parameter for precomputation as storage
+        // grows exponentially, but the speedup is exponential - after a while the speedup
+        // starts to become sublinear.
         let context = DASContext::new(
             &peerdas_trusted_setup,
             rust_eth_kzg::UsePrecomp::Yes {
--- a/crypto/kzg/src/lib.rs
+++ b/crypto/kzg/src/lib.rs
@@ -72,12 +72,13 @@ impl Kzg {
         // Note: One can also use `from_json` to initialize it from the consensus-specs
         // json string.
         let peerdas_trusted_setup = PeerDASTrustedSetup::from(&trusted_setup);
-        // Set the number of threads to be used
-        //
-        // we set it to 1 to match the c-kzg performance
-        const NUM_THREADS: usize = 1;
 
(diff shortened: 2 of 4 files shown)


crypto/kzg/benches/benchmark.rs
```diff
@@ -11,9 +11,13 @@ pub fn bench_init_context(c: &mut Criterion) {
 
     c.bench_function(&format!("Initialize context rust_eth_kzg"), |b| {
         b.iter(|| {
-            const NUM_THREADS: usize = 1;
             let trusted_setup = PeerDASTrustedSetup::from(&trusted_setup);
-            DASContext::with_threads(&trusted_setup, NUM_THREADS)
+            DASContext::new(
+                &trusted_setup,
+                rust_eth_kzg::UsePrecomp::Yes {
+                    width: rust_eth_kzg::constants::RECOMMENDED_PRECOMP_WIDTH,
```
kevaundray, line 18: I don't recommend changing the config parameter for precomputation as storage grows expoentially, but the speedup is not exponential -- after a while the speedup starts to become sublinear

crypto/kzg/src/lib.rs
```diff
@@ -72,12 +72,13 @@ impl Kzg {
         // Note: One can also use `from_json` to initialize it from the consensus-specs
         // json string.
         let peerdas_trusted_setup = PeerDASTrustedSetup::from(&trusted_setup);
-        // Set the number of threads to be used
-        //
-        // we set it to 1 to match the c-kzg performance
-        const NUM_THREADS: usize = 1;
```
kevaundray, line 78: By default rayon is disabled, so its implicitly single threaded

crypto/kzg/src/lib.rs
```diff
@@ -244,7 +245,7 @@ impl Kzg {
     ) -> Result<CellsAndKzgProofs, Error> {
         let (cells, proofs) = self
             .context()?
-            .recover_cells_and_proofs(cell_ids.to_vec(), cells.to_vec())
+            .recover_cells_and_kzg_proofs(cell_ids.to_vec(), cells.to_vec())
```
kevaundray, line 251: Changed the name to match the specs more -- no functionality has been changed here

crypto/kzg/benches/benchmark.rs
jimmygchen: Useful to know - I'll add this as code comment.

crypto/kzg/benches/benchmark.rs
```diff
@@ -11,9 +11,13 @@ pub fn bench_init_context(c: &mut Criterion) {
 
     c.bench_function(&format!("Initialize context rust_eth_kzg"), |b| {
         b.iter(|| {
-            const NUM_THREADS: usize = 1;
             let trusted_setup = PeerDASTrustedSetup::from(&trusted_setup);
-            DASContext::with_threads(&trusted_setup, NUM_THREADS)
+            DASContext::new(
```
jimmygchen, line 15:
```suggestion
            // It's not recommended to change the config parameter for precomputation as storage 
            // grows exponentially, but the speedup is exponential - after a while the speedup 
            // starts to become sublinear.
            DASContext::new(
```

jimmygchen: I'll move this to the `new_from_trusted_setup_das_enabled` function

jimmygchen (approved):
Looks good, thanks @kevaundray ❤️

jimmygchen:
@mergify queue

PR #6311, opened 2024-08-27
chore: Lazy initialize the KZG struct when running tests

kevaundray:
## Issue Addressed

get_kzg creates a new KZG struct each time its called, which can slow down tests since each test case will call get_kzg. 

## Proposed Changes

This PR uses LazyLock to instantiate it once using LazyLock and then other tests will just take a thread-safe reference to it. 

## Additional Info

There is another get_kzg that has been defined in kzg_utils -- That only calls get_kzg once, so its not an issue. Longer term, it might make sense to put a get_kzg method with LazyLock in crypto/get_kzg.rs, so that tests all over the workspace can access it. The main issue I see with this, is that the trusted_setup file is not in crypto/kzg.rs, so the solution won't be as clean as what it looks like in the tests modules.

### DIFF
--- a/testing/ef_tests/src/cases/kzg_verify_blob_kzg_proof.rs
+++ b/testing/ef_tests/src/cases/kzg_verify_blob_kzg_proof.rs
@@ -17,8 +17,8 @@ static KZG: LazyLock<Arc<Kzg>> = LazyLock::new(|| {
     Arc::new(kzg)
 });
 
-pub fn get_kzg() -> Result<Arc<Kzg>, Error> {
-    Ok(Arc::clone(&KZG))
+pub fn get_kzg() -> Arc<Kzg> {
+    Arc::clone(&KZG)
 }
 
 pub fn parse_cells_and_proofs(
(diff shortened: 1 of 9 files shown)


testing/ef_tests/src/cases/kzg_verify_blob_kzg_proof.rs
```diff
@@ -5,14 +5,20 @@ use eth2_network_config::TRUSTED_SETUP_BYTES;
...
+use std::sync::Arc;
+use std::sync::LazyLock;
 use types::Blob;
 
-pub fn get_kzg() -> Result<Kzg, Error> {
-    let trusted_setup: TrustedSetup = serde_json::from_reader(TRUSTED_SETUP_BYTES)
-        .map_err(|e| Error::InternalError(format!("Failed to initialize kzg: {:?}", e)))?;
+static KZG: LazyLock<Arc<Kzg>> = LazyLock::new(|| {
+    let trusted_setup: TrustedSetup =
+        serde_json::from_reader(TRUSTED_SETUP_BYTES).expect("failed to initialize kzg");
```
kevaundray: Noting: the other get_kzg method in kzg_utils also does not return Result

jimmygchen (approved):
Looks good to me! 🎉

jimmygchen:
@mergify queue

PR #6313, opened 2024-08-28
Encode DataColumn responses with Electra context bytes

jimmygchen:
## Issue Addressed

Encode DataColumn responses with Electra context bytes and remove some outdated `das` TODOs.

The current response returns Deneb context bytes as we've been testing PeerDAS with Deneb. From the `peerdas-devnet-3` onwards, we'll be testing it with Electra, so it would be good to get this change in to start testing. 

Once we're comfortable with testing on Electra, the Deneb fork conditions can be removed.

### DIFF
--- a/beacon_node/lighthouse_network/src/rpc/codec/ssz_snappy.rs
+++ b/beacon_node/lighthouse_network/src/rpc/codec/ssz_snappy.rs
@@ -422,8 +422,7 @@ fn context_bytes<E: EthSpec>(
                 RPCResponse::BlobsByRange(_) | RPCResponse::BlobsByRoot(_) => {
                     return fork_context.to_context_bytes(ForkName::Deneb);
                 }
-                RPCResponse::DataColumnsByRoot(data_column)
-                | RPCResponse::DataColumnsByRange(data_column) => {
+                RPCResponse::DataColumnsByRoot(_) | RPCResponse::DataColumnsByRange(_) => {
                     // TODO(das): Remove deneb fork after `peerdas-devnet-2`.
                     return if fork_context.spec.eip7594_fork_epoch
                         == fork_context.spec.deneb_fork_epoch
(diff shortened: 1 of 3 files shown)


dapplion (approved):
lgtm

jimmygchen:
@mergify queue

beacon_node/lighthouse_network/src/rpc/codec/ssz_snappy.rs
```diff
@@ -419,14 +419,19 @@ fn context_bytes<E: EthSpec>(
...
-                | RPCResponse::DataColumnsByRange(_) => {
-                    // TODO(das): If DataColumnSidecar is defined as an Electra type, update the
-                    // context bytes to point to ForkName::Electra
+                RPCResponse::BlobsByRange(_) | RPCResponse::BlobsByRoot(_) => {
                     return fork_context.to_context_bytes(ForkName::Deneb);
                 }
+                RPCResponse::DataColumnsByRoot(_) | RPCResponse::DataColumnsByRange(_) => {
+                    // TODO(das): Remove deneb fork after `peerdas-devnet-2`.
+                    return if fork_context.spec.eip7594_fork_epoch
+                        == fork_context.spec.deneb_fork_epoch
```
pawanjay176, line 428: Just confirming if this condition is specifically for devnet testing.
If we start a network with deneb_fork_epoch == electra_fork_epoch == peerdas_fork_epoch = 0, then this will give deneb context bytes even though we probably want electra.

jimmygchen: Good catch - this condition is indeed for testing only, but like you said this would return Deneb even if we have electra fork epoch set to 0. I'll push a fix.

jimmygchen: Created https://github.com/sigp/lighthouse/pull/6325

PR #6316, opened 2024-08-28
Fix docker & release workflows

jimmygchen:
## Issue Addressed

Failing docker build: https://github.com/sigp/lighthouse/actions/runs/10591577042/job/29349249752

Remove portable builds which no longer exists and is no longer published since #6247.


antondlr (approved):
lgtm!

jimmygchen:
@mergify queue

PR #6317, opened 2024-08-28
Add reason of NoRequestNeeded to lookup sync

dapplion:
## Issue Addressed

Porting improvement from 
- https://github.com/sigp/lighthouse/pull/6275

## Proposed Changes

Now there are more reasons why a lookup request will be `NoRequestNeeded`. To debug stuck sync lookup it's important to know why sync would consider the request unnecessary.


jimmygchen (approved):
Nice, I think this will help with debugging  👍

jimmygchen:
@mergify queue

jimmygchen:
@mergify queue

jimmygchen:
@mergify requeue

PR #6318, opened 2024-08-28
Remove duplicate should_fetch_blobs check in sync lookup

dapplion:
## Issue Addressed

The function `blob_lookup_request` now includes two checks if a block is in the da window:
- First: check if clock slot is in da_window
- Second: check if block's slot is in da_window

https://github.com/sigp/lighthouse/blob/5a966874da03cd987fef1629f732898a2859ce87/beacon_node/network/src/sync/network_context.rs#L641-L643

The first check has been there since pre-deneb, and is incorrect da_window-inclusiveness should be computed on the block  slot not the clock.

## Proposed Changes

Remove the first check


pawanjay176 (approved):
Nice catch! No idea why we had this condition in the first place

jimmygchen:
@mergify queue

PR #6321, opened 2024-08-28
Stricter match of BlockError in lookup sync

dapplion:
## Issue Addressed

Adds a strict match without fallbacks to handle all processing errors in lookup sync explicitly. With this we can:
- Be explicit on what errors must result in penalties
- Do not retry errors that are deterministic on the block root

**Example problem 1**

Today, if a block has incorrect state root we will download and process it 5 times. This is wasteful, we should discard the block immediately

**Example problem 2**

Assume we introduce a new `BlockError` variant in a future fork or network change. Lookup sync is very sensitive code, and not handling this error variant properly may result in sync getting stuck

## Proposed Changes

In the network processor convert `BlockError` to `LookupSyncProcessingResult`.

Then lookup sync logic is quite simple match on `LookupSyncProcessingResult`, without fallbacks.


beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -567,48 +562,50 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
...
+                                }
+                                recoverable
+                            }
+                            ErrorCategory::Malicious { recoverable, index } => {
+                                debug!(self.log, "Invalid lookup component"; "block_root" => ?block_root, "component" => ?R::response_type(), "error" => ?other);
+                                let peer_group = request_state.on_processing_failure()?;
+                                for peer in peer_group.of_index(index) {
+                                    cx.report_peer(
+                                        *peer,
+                                        PeerAction::MidToleranceError,
```
dapplion: I think we should up this to LowToleranceError

beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -567,48 +562,50 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
...
-                                    ResponseType::Block => "lookup_block_processing_failure",
-                                    ResponseType::Blob => "lookup_blobs_processing_failure",
-                                    ResponseType::CustodyColumn => {
-                                        "lookup_custody_column_processing_failure"
-                                    }
-                                },
-                            );
+                        let recoverable = match other.rpc_scoring() {
+                            ErrorCategory::Internal { recoverable } => {
+                                if matches!(other, BlockError::ExecutionPayloadError(_)) {
```
dapplion: I kept this extra condition from the original match as I imagine that when EL goes offline the node would spam non-actionable errors

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -428,6 +472,30 @@ impl ExecutionPayloadError {
...
+        // This match statement should never have a default case
+        match self {
+            // The peer has nothing to do with this error, do not penalize them.
+            ExecutionPayloadError::NoExecutionConnection => {
+                ErrorCategory::internal_non_recoverable()
+            }
+            // The peer has nothing to do with this error, do not penalize them.
+            ExecutionPayloadError::RequestFailed(_) => ErrorCategory::internal_recoverable(),
+            // Execution payload is invalid
+            ExecutionPayloadError::RejectedByExecutionEngine { .. }
```
dapplion: If EL says the payload is invalid we MUST downscore this peer if fetched from RPC right?

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -428,6 +472,30 @@ impl ExecutionPayloadError {
             ExecutionPayloadError::UnverifiedNonOptimisticCandidate => false,
         }
     }
+
+    pub fn penalize_rpc_peer(&self) -> ErrorCategory {
```
dapplion: Should this match statements be in beacon_chain on in sync / network?

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -778,7 +778,7 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                 })
             }
             ref err @ BlockError::ExecutionPayloadError(ref epe) => {
-                if !epe.penalize_peer() {
+                if !epe.penalize_gossip_peer() {
```
pawanjay176:
```suggestion
                if !epe.penalize_rpc_peer() {
```

beacon_node/beacon_chain/src/data_availability_checker/error.rs
```diff
@@ -27,32 +28,76 @@ pub enum Error {
...
-            Error::KzgNotInitialized
-            | Error::SszTypes(_)
-            | Error::MissingBlobs
-            | Error::MissingCustodyColumns
-            | Error::StoreError(_)
+            // KZG is only initialized once
+            Error::KzgNotInitialized => ErrorCategory::internal_non_recoverable(),
+            Error::SszTypes(_) => ErrorCategory::internal_non_recoverable(),
+            // A ChainSegment RpcBlock does not include the expected blobs or columns
+            Error::MissingBlobs | Error::MissingCustodyColumns => {
```
pawanjay176: Wouldn't these be malicious? A peer sent us a response to a rpc request which has missing blobs for the same block they sent

beacon_node/beacon_chain/src/data_availability_checker/error.rs
```diff
@@ -27,32 +28,76 @@ pub enum Error {
...
+                ErrorCategory::internal_recoverable()
+            }
+            // Assume these errors to be recoverable
+            Error::StoreError(_)
             | Error::DecodeError(_)
```
pawanjay176: Maybe decode errors are malicious too? I couldn't track down the exact place where we would return this error but seems like a peer issue

beacon_node/beacon_chain/src/data_availability_checker/error.rs
```diff
@@ -27,32 +28,76 @@ pub enum Error {
...
             | Error::Unexpected
             | Error::ParentStateMissing(_)
             | Error::BlockReplayError(_)
             | Error::UnableToDetermineImportRequirement
```
pawanjay176: Seems like this variant isn't used anywhere, can delete

beacon_node/network/src/sync/network_context.rs
```diff
@@ -137,6 +137,15 @@ impl PeerGroup {
     pub fn all(&self) -> impl Iterator<Item = &PeerId> + '_ {
         self.peers.keys()
     }
+    pub fn of_index(&self, index: usize) -> impl Iterator<Item = &PeerId> + '_ {
```
pawanjay176: Not completely sure why we need an iterator here. Would there be a scenario where you get > 1 peer responsible for a single index here?

beacon_node/beacon_chain/src/data_availability_checker/error.rs
```diff
@@ -27,32 +28,76 @@ pub enum Error {
...
 pub enum ErrorCategory {
-    /// Internal Errors (not caused by peers)
-    Internal,
-    /// Errors caused by faulty / malicious peers
-    Malicious,
+    /// Internal Errors (not caused by peers).
+    ///
+    /// An internal non-recoverable error is permanent and block processing should not be
+    /// re-attempted.
+    Internal { recoverable: bool },
```
pawanjay176:
```suggestion
    Internal { retry: bool },
```
In reviewing this, I found that for a lot of errors, it was really ambiguous what "recoverable" meant in the context.

Instead, we could just indicate the action we want sync to take given the error which would reduce the ambiguity imo.

beacon_node/beacon_chain/src/data_availability_checker/error.rs
```diff
@@ -27,32 +28,76 @@ pub enum Error {
...
+    /// Errors caused by faulty / malicious peers.
+    ///
+    /// Non-recoverable errors are deterministic against the block's root. Re-downloading data
+    /// key-ed by block root MUST result in the same non-recoverable error (i.e. invalid parent,
+    /// invalid state root, etc).
+    ///
+    /// The error also indicates which block component index is malicious if applicable.
+    Malicious { recoverable: bool, index: usize },
```
pawanjay176: same as above, maybe we use retry to indicate the action we want sync to take

beacon_node/beacon_chain/src/block_verification.rs
pawanjay176: Considering that we are using this specifically to recover from errors in block_lookups, I think it should be in block lookups, but can be convinced either way

beacon_node/beacon_chain/src/data_availability_checker/error.rs
pawanjay176: Also, thinking more about it, internal + recoverable/retry doesn't make much sense to me as a variant.
Currently, the error variants that have this category are
```
ParentUnknown/ReconstructColumnsError -> needs a special case to handle

FutureSlot -> i think this should be a malicious error and not internal assuming our clock is correct

BeaconChainError/SlotClockError -> cannot act on it so not really recoverable or worth retrying

ExecutionPayloadError::RequestFailed -> can't act on it until EL is back online

ExecutionPayloadError::UnverifiedNonOptimisticCandidate -> won't happen after the merge

MissingBlobs/Custody -> this is a malicious error

KzgCommitmentMismatch -> this never happens because we check the inclusion proof
```

beacon_node/beacon_chain/src/data_availability_checker/error.rs
```diff
@@ -27,32 +28,76 @@ pub enum Error {
...
-            | Error::BlobIndexInvalid(_)
-            | Error::DataColumnIndexInvalid(_)
-            | Error::KzgCommitmentMismatch { .. }
-            | Error::KzgVerificationFailed => ErrorCategory::Malicious,
+            | Error::SlotClockError => ErrorCategory::internal_recoverable(),
+            // BlobIndexInvalid only happens when serving RPC requests for out of bounds blob indices
+            Error::BlobIndexInvalid(_) | Error::DataColumnIndexInvalid(_) => {
+                ErrorCategory::malicious_non_recoverable()
+            }
+            // TODO: Is this error recoverable?
```
pawanjay176: Might be an internal kzg library error here that we cannot do anything about here.
The columns have already been verified before reconstruction is attempted, so i think this would be an internal error

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -328,6 +330,48 @@ pub enum BlockError<E: EthSpec> {
     InternalError(String),
 }
 
+impl<E: EthSpec> BlockError<E> {
+    pub fn rpc_scoring(&self) -> ErrorCategory {
+        match self {
+            Self::ParentUnknown { .. } => ErrorCategory::internal_recoverable(),
+            Self::FutureSlot { .. } => ErrorCategory::internal_recoverable(),
```
pawanjay176: Shouldn't happen in rpc but I think this should be a malicious error if it happens

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -328,6 +330,48 @@ pub enum BlockError<E: EthSpec> {
...
+            Self::StateRootMismatch { .. } => ErrorCategory::malicious_non_recoverable(),
+            Self::GenesisBlock => ErrorCategory::internal_non_recoverable(),
+            Self::WouldRevertFinalizedSlot { .. } => ErrorCategory::internal_non_recoverable(),
```
pawanjay176: should also be malicious i think

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -328,6 +330,48 @@ pub enum BlockError<E: EthSpec> {
...
+            Self::NotFinalizedDescendant { .. } => ErrorCategory::internal_non_recoverable(),
```
pawanjay176: same as above

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -328,6 +330,48 @@ pub enum BlockError<E: EthSpec> {
...
+            Self::BlockIsAlreadyKnown { .. } => ErrorCategory::internal_non_recoverable(),
+            Self::BlockSlotLimitReached => ErrorCategory::malicious_non_recoverable(),
+            Self::IncorrectBlockProposer { .. } => ErrorCategory::malicious_non_recoverable(),
+            Self::ProposalSignatureInvalid => ErrorCategory::malicious_recoverable(),
```
pawanjay176: not sure why this is recoverable/worth retrying if we are getting by root. We always check that the block root matches the requested root, so if it decoded correctly, that means that the block root has an invalid proposer signature so not worth re-requesting

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -328,6 +330,48 @@ pub enum BlockError<E: EthSpec> {
...
+            Self::UnknownValidator { .. } => ErrorCategory::malicious_non_recoverable(),
+            // TODO: Labeling as recoverable as it may be the proposer signature. We should check
+            // Which one is it and label as non recoverable if the proposer signature is correct.
```
pawanjay176: if the above comment is correct, this should be malicious non recoverable

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -328,6 +330,48 @@ pub enum BlockError<E: EthSpec> {
...
+            Self::InvalidSignature => ErrorCategory::malicious_recoverable(),
+            Self::BlockIsNotLaterThanParent { .. } => ErrorCategory::malicious_non_recoverable(),
+            // TODO: Not clear what to set here
+            Self::NonLinearParentRoots => ErrorCategory::malicious_recoverable(),
```
pawanjay176: this would be malicious non recoverable since we are checking the block root matches what we requested

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -328,6 +330,48 @@ pub enum BlockError<E: EthSpec> {
...
+            Self::NonLinearSlots => ErrorCategory::malicious_recoverable(),
```
pawanjay176: same as above

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -328,6 +330,48 @@ pub enum BlockError<E: EthSpec> {
...
+            // TODO: Ensure the proposer signature is not part of this error set
+            Self::PerBlockProcessingError(_) => ErrorCategory::malicious_non_recoverable(),
+            // TODO: Are we sure all BeaconChainError are internal?
```
pawanjay176: I think that is a fair assumption

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -328,6 +330,48 @@ pub enum BlockError<E: EthSpec> {
...
+            Self::BeaconChainError(_e) => ErrorCategory::internal_recoverable(),
+            Self::WeakSubjectivityConflict => ErrorCategory::malicious_non_recoverable(),
+            Self::InconsistentFork(_) => ErrorCategory::malicious_non_recoverable(),
+            Self::ExecutionPayloadError(e) => e.penalize_rpc_peer(),
+            Self::ParentExecutionPayloadInvalid { .. } => {
+                ErrorCategory::malicious_non_recoverable()
+            }
+            // TODO: Check if correct
+            Self::Slashable => ErrorCategory::malicious_non_recoverable(),
```
pawanjay176: This is a proposer making a slashable block, neither the peer's fault nor is it our fault. Maybe needs a special case

beacon_node/beacon_chain/src/block_verification.rs
pawanjay176: Yes, the peer has already imported the block into fork choice if we are fetching it over rpc

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -199,9 +199,11 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
             .now_duration()
             .ok_or(AvailabilityCheckError::SlotClockError)?;
 
+        // Note: currenlty not reporting which specific blob is invalid because we fetch all blobs
```
pawanjay176:
```suggestion
        // Note: currently not reporting which specific blob is invalid because we fetch all blobs
```

beacon_node/beacon_chain/src/block_verification.rs
dapplion: The proposer signature is in the outer `SignedBeaconBlock` container so not included in the block root `= hash_tree_root(BeaconBlock)`

beacon_node/network/src/sync/network_context.rs
dapplion: Harder to model with the type here, should we just return the first match? Or enforce uniqueness on the reverse index -> peer mapping?

pawanjay176: First match seems good to me

beacon_node/beacon_chain/src/block_verification.rs
pawanjay176: Ohh that's right. Good call, we can handle the recovery logic in a separate PR if we don't want to clutter this too much

dapplion: It's already handled, malicious_recoverable means that the serving peer will be penalized, and we will retry with a different peer

beacon_node/beacon_chain/src/block_verification.rs
dapplion: Depends on the block type. For RPC blocks we check all signatures at once so the validity of the signature set is not deterministic on the block root

beacon_node/beacon_chain/src/data_availability_checker/error.rs
dapplion: Moved MissingBlobs/Custody to malicious. 

Agree, but will look at dropping KzgCommitmentMismatch in a future PR

dapplion:
@pawanjay176 I have moved the big match into the network beacon processor. This is analogous to how we handle the result of processing chain segments and gossip objects.

I have applied all your comments to the current version.

I think the PR makes much more sense now, ready for a re-review

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -332,53 +353,159 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
                 );
+                LookupSyncProcessingResult::ImportedMissingComponents
+            }
+            Err(err) => {
+                debug!(
+                    self.log,
+                    "Error processing block component";
+                    "block_root" => %block_root,
+                    "err" => ?err,
+                );
```
dapplion, line 413: Do we want to log internal errors at a higher level than debug? If yes I can move the log statement below and match the type of ErrorCategory

beacon_node/network/src/network_beacon_processor/mod.rs
```diff
@@ -983,7 +985,7 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                             "result" => "imported block and custody columns",
                             "block_hash" => %hash,
                         );
-                        self.chain.recompute_head_at_current_slot().await;
+                        // Head will be recomputed `handle_lookup_sync_processing_result`
```
jimmygchen, line 988: This doesn't cover the gossip case right?

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -44,6 +45,60 @@ struct ChainSegmentFailed {
...
+    /// re-attempted.
+    Internal { retry: bool },
+    /// Errors caused by faulty / malicious peers.
+    ///
+    /// No retry errors are deterministic against the block's root. Re-downloading data
+    /// key-ed by block root MUST result in the same no-retry error (i.e. invalid parent,
+    /// invalid state root, etc).
+    ///
+    /// The error also indicates which block component index is malicious if applicable.
+    Malicious { retry: bool, index: usize },
```
jimmygchen, line 71: I found `index` slightly confusing - what do you think about calling this `sidecar_index` or `data_sidecar_index`, and potentially make this an `Option`, given we're using this for blocks as well?

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -214,7 +269,9 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
         // Sync handles these results
         self.send_sync_message(SyncMessage::BlockComponentProcessed {
             process_type,
-            result: result.into(),
+            result: self
+                .handle_lookup_sync_processing_result(block_root, result)
```
jimmygchen, line 273: Looks like we're potentially recomputing head twice here - one above at https://github.com/sigp/lighthouse/blob/2c8073831aefb7ea7627d28ea14dfef6b1f70e2c/beacon_node/network/src/network_beacon_processor/sync_methods.rs#L245 and one in `self.handle_lookup_sync_processing_result`.

jimmygchen (commented):
I still haven't finished reviewing this but I'll try to get back to this later today or Monday.

beacon_node/network/src/network_beacon_processor/mod.rs
jimmygchen: Feels like it might be easier to move the `recompute_head logic` back here, otherwise we'll have to either do it in two places or having to pass a flag here and conditionally recompute head.

beacon_node/network/src/network_beacon_processor/sync_methods.rs
jimmygchen: Does it make sense to call this `Faulty` instead? we don't know if the peer is malicious - i think it might be more consistent with the terminology used in sync, although it doesn't really matter, a faulty peer and malicious peer both send invalid data.

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -44,6 +45,60 @@ struct ChainSegmentFailed {
...
+    pub fn internal_no_retry() -> Self {
+        Self::Internal { retry: false }
+    }
+    pub fn internal_retry() -> Self {
+        Self::Internal { retry: true }
+    }
+    pub fn malicious_no_retry() -> Self {
+        Self::Malicious {
+            retry: false,
+            index: 0,
```
jimmygchen, line 91: I think it make sense to make this `Option`, otherwise we can't distinguish between `0` index or `None`

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -344,53 +365,159 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                    // slashable).
+                    BlockError::Slashable => ErrorCategory::internal_no_retry().into(),
+                    // TODO: BeaconChainError should be retried?
+                    BlockError::BeaconChainError(_) | BlockError::InternalError(_) => {
+                        ErrorCategory::internal_no_retry().into()
+                    }
+                    // unreachable, this error is only part of gossip
+                    BlockError::BlobNotRequired(_) => ErrorCategory::malicious_retry().into(),
+                    // Unreachable: This variants never happen in lookup sync, only in range sync.
+                    // Does not matter what we set here, just setting `internal_recoverable` to
```
jimmygchen, line 512:
```suggestion
                    // Does not matter what we set here, just setting `internal_retry` to
```

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -344,53 +365,159 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                    // put something.
+                    BlockError::NonLinearParentRoots | BlockError::NonLinearSlots => {
```
jimmygchen, line 514: For code path that's expected to be unreachable, is it worth logging a warning here?

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -344,53 +365,159 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                        } // Do not use a fallback match, handle all errors explicitly
+                    }
+                    .into(),
+                    // The proposer making a slashable block is not the peer's fault nor ours. Mark
+                    // as internal (don't penalize peer), and no retry (the block will forever be
+                    // slashable).
+                    BlockError::Slashable => ErrorCategory::internal_no_retry().into(),
+                    // TODO: BeaconChainError should be retried?
+                    BlockError::BeaconChainError(_) | BlockError::InternalError(_) => {
+                        ErrorCategory::internal_no_retry().into()
```
jimmygchen, line 507: Both of these variants cover quite a few scenarios, I'm not sure if it's safe to say no retry to them?

dapplion:
- Closing for stale for https://github.com/sigp/lighthouse/issues/7129

PR #6322, opened 2024-08-28
Run network tests with test_logger

dapplion:
## Issue Addressed

When debugging network test errors in CI I miss having access to the logs if the test fail. There should be no significant downside to enabling the logger, capturing the logs and dropping them on failure.

## Proposed Changes

Enable test logger by default on network tests CI run

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,7 +155,7 @@ test-network: $(patsubst %,test-network-%,$(FORKS))
 
 test-network-%:
 	env FORK_NAME=$* cargo nextest run --release \
-		--features "fork_from_env,$(TEST_FEATURES)" \
+		--features "fork_from_env,test_logger,$(TEST_FEATURES)" \
 		-p network
 
 # Run the tests in the `slasher` crate for all supported database backends.


michaelsproul:
There's the problem of log interleaving isn't there? I only ever use test_logger when running one test at a time. If you run lots of tests concurrently they just interleave their output in a disgusting way

dapplion:
> There's the problem of log interleaving isn't there? I only ever use test_logger when running one test at a time. If you run lots of tests concurrently they just interleave their output in a disgusting way

Yeah right.. need to figure a way for cargo to capture the output of slog so they don't get interleaved

ackintosh:
I think we can output the logs into files for each test (e.g. `sync::block_lookups::tests::sampling_with_retries.log`), and upload the log files as [Actions Artifacts](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/storing-and-sharing-data-from-a-workflow), so that we can download and check the logs of failed tests. I can tackle this if it sounds good.

michaelsproul:
Closing in favour of:

- https://github.com/sigp/lighthouse/pull/6355

PR #6323, opened 2024-08-28
Switch custody_subnet_count from u64 to u8

pawanjay176:
## Issue Addressed

https://github.com/ethereum/consensus-specs/pull/3897/

## Proposed Changes

Change metadata and enr fields to take a u8 instead of a u64. 

I have chosen not to change the config and chainspec params to u8 as well since they permeate throughout the codebase and we'll have to do a bunch of unnecessary `u8 as u64` sort of conversions. Instead, I made apply_to_chainspec fallible in case the parameters in the config exceed u8::MAX.

### DIFF
--- a/consensus/types/src/chain_spec.rs
+++ b/consensus/types/src/chain_spec.rs
@@ -1807,6 +1807,14 @@ impl Config {
             return None;
         }
 
+        let u8_max = u8::MAX as u64;
+        if custody_requirement > u8_max
+            || data_column_sidecar_subnet_count > u8_max
+            || number_of_columns > u8_max
+        {
+            return None;
+        }
+
         Some(ChainSpec {
             config_name: config_name.clone(),
             min_genesis_active_validator_count,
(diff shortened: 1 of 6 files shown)


consensus/types/src/chain_spec.rs
```diff
@@ -1807,6 +1807,14 @@ impl Config {
             return None;
         }
 
+        let u8_max = u8::MAX as u64;
+        if custody_requirement > u8_max
+            || data_column_sidecar_subnet_count > u8_max
+            || number_of_columns > u8_max
```
jimmygchen, line 1813: `number_of_columns` is still `uint64` in the spec

```
| `NUMBER_OF_COLUMNS` | `uint64(CELLS_PER_EXT_BLOB)` (= 128) | Number of columns in the extended data matrix |
```

jimmygchen (changes requested):
I just realized the change isn't strictly necessary because RLP encoding doesn't actually fill any 0 bytes, so encoding 128 is the same for `u8` and `u64`. This change just put a upper bound to the number 🤔  

I'm not sure if the spec change was necessary. it just limit us to go above 255 subnets and has no impact on the encoded value. Will ask on R&D.

pawanjay176:
This isn't required after https://github.com/ethereum/consensus-specs/pull/3908

PR #6325, opened 2024-08-29
Fix context bytes logic for DataColumn RPC responses

jimmygchen:
## Issue Addressed

The data column context bytes logic implemented in #6313 won't work if we start Electra and PeerDAS at genesis, see @pawanjay176 's comment here:
https://github.com/sigp/lighthouse/pull/6313#discussion_r1735112366

This PR fixes the issue, and eventually we should delete this logic altogether once Deneb testing is over.


pawanjay176 (approved):

jimmygchen:
@mergify queue

Issue #6319, opened 2024-08-28
Investigate flaky sampling tests


jimmygchen:
## Description

The test `network sync::block_lookups::tests::sampling_with_retries` is failing on CI intermittently, seems like it could be a race condition somewhere.

https://github.com/sigp/lighthouse/actions/runs/10592162826/job/29350969832?pr=6312

dapplion:
I recall the flakiness of this test to increase with https://github.com/sigp/lighthouse/pull/6256 as a pointer to start investigating

ackintosh:
When I reproduced the test failure,  `cx.get_custodial_peers(self.column_index)` returned only one peer (a supernode), but the test expects two or more peers to retry sampling.

A solution that came to mind is to add another supernode (making a total of two) to ensure that the node can retry sampling. What do you think?

https://github.com/sigp/lighthouse/blob/f75a2cf65b4bc4724bab6c6edd5754e5150eca6d/beacon_node/network/src/sync/sampling.rs#L570-L586

jimmygchen:
Oh I think it may also be related to #6303 - the `good_custody_subnet_peer` function now checks whether a peer is assigned to a subnet:

https://github.com/sigp/lighthouse/blob/c0b4f01cf37257bab8dab2b4312dcf07eac93572/beacon_node/lighthouse_network/src/peer_manager/peerdb.rs#L250-L264

and with #6308 we no longer use ENR for choosing peers, instead we use metadata to populate `peer_info.custody_subets`:

https://github.com/sigp/lighthouse/blob/c0b4f01cf37257bab8dab2b4312dcf07eac93572/beacon_node/lighthouse_network/src/peer_manager/mod.rs#L726-L733

In the test rig, only the supernode has `peer_info.custody_subnets` populated:

https://github.com/sigp/lighthouse/blob/c0b4f01cf37257bab8dab2b4312dcf07eac93572/beacon_node/lighthouse_network/src/peer_manager/peerdb.rs#L720-L726

So this means the remaining 100 peers we added here are not useful, because custody_subnets are empty for them:

https://github.com/sigp/lighthouse/blob/ed7cd3bf47e9d407e4f175263f7e1cbf6dd8d989/beacon_node/network/src/sync/block_lookups/tests.rs#L401-L408

I think we can either
- add an additional supernode as you suggested, and get rid of adding 100 peers; OR
[...]

PR #6326, opened 2024-08-29
Ignore flaky sampling test

jimmygchen:
## Issue Addressed

The test `network sync::block_lookups::tests::sampling_with_retries` has been failing quite regularly on CI,.

This PR ignores the test, as it is slowing down our checks and merges. Sampling is currently disabled by default anyway and is probably not worth slowing down other higher priority work.

Issue raised for this: https://github.com/sigp/lighthouse/issues/6319


pawanjay176 (approved):
Yeah i saw this on some of my branches too, can come back to this as we have the issue

jimmygchen:
@mergify queue

PR #6328, opened 2024-08-29
Fix tree states upgrade migration

michaelsproul:
## Proposed Changes

Fix bugs in the tree-states upgrade migration.


dapplion (approved):
LGTM thanks for the nice comments.

For context the bug this PR is fixing was that the migration was applying all ops to the hot DB, so it was writing the block roots with the new format into the hot DB and never deleting the chunked ones.

PR #6329, opened 2024-08-29
Make max_blobs_per_block a config parameter

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Change `max_blobs_per_block` from a preset value to a config value. This affects a lot of the codebase where we have to use runtime variants of ssz List and Vector.


beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -1162,26 +1169,28 @@ mod pending_components_tests {
...
+        usize,
     );
 
     pub fn pre_setup() -> Setup<E> {
         let mut rng = StdRng::seed_from_u64(0xDEADBEEF0BAD5EEDu64);
+        let spec = test_spec::<E>();
         let (block, blobs_vec) =
             generate_rand_block_and_blobs::<E>(ForkName::Deneb, NumBlobs::Random, &mut rng);
-        let mut blobs: FixedVector<_, <E as EthSpec>::MaxBlobsPerBlock> = FixedVector::default();
+        let max_len = spec.max_blobs_per_block(block.epoch()) as usize;
```
jimmygchen, line 1177: The epoch here will be random right?

consensus/types/src/runtime_var_list.rs
```diff
@@ -214,6 +262,85 @@ where
     }
 }
 
+/// Emulates a SSZ `Vector`.
+#[derive(Clone, Debug)]
+pub struct RuntimeFixedList<T> {
```
jimmygchen: oh we should probably move this to a separate file

beacon_node/execution_layer/src/test_utils/execution_block_generator.rs
```diff
@@ -676,7 +676,8 @@ impl<E: EthSpec> ExecutionBlockGenerator<E> {
             ForkName::Deneb | ForkName::Electra => {
                 // get random number between 0 and Max Blobs
                 let mut rng = self.rng.lock();
-                let num_blobs = rng.gen::<usize>() % (E::max_blobs_per_block() + 1);
+                // TODO(pawan): thread the chainspec value here somehow
+                let num_blobs = rng.gen::<usize>() % 6;
```
jimmygchen:
```suggestion
                let num_blobs = rng.gen::<usize>() % (6 + 1);
```

consensus/types/src/chain_spec.rs
```diff
@@ -607,6 +608,16 @@ impl ChainSpec {
         }
     }
 
+    /// Returns the deneb preset value if peerdas epoch hasn't hit.
+    /// Otherwise, returns the value obtained from the config.yaml.
+    pub fn max_blobs_per_block(&self, epoch: Epoch) -> u64 {
+        if self.is_peer_das_enabled_for_epoch(epoch) {
+            self.max_blobs_per_block
+        } else {
+            default_max_blobs_per_block()
```
jimmygchen: Is this necessary? If we're changing this on devnets, then all clients would be using  `MAX_BLOBS_PER_BLOCK` from deneb, and we won't be able to interop with other clients. This config value is already configurable for some clients for Deneb, so if we're overriding this value, other clients would be using this value even before PeerDAS.

For PeerDAS purpose I guess we'd be using a new config value when the time comes?

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -223,10 +216,11 @@ impl<E: EthSpec> PendingComponents<E> {
     }
 
     /// Returns an empty `PendingComponents` object with the given block root.
-    pub fn empty(block_root: Hash256) -> Self {
+    pub fn empty(block_root: Hash256, max_len: usize) -> Self {
         Self {
             block_root,
-            verified_blobs: FixedVector::default(),
+            // TODO(pawan): just make this a vec potentially
```
jimmygchen: I think we can - this was previous an SSZ type because of overflow LRU cache and persisting to disk, but that's been dropped.

consensus/types/src/runtime_var_list.rs
```diff
@@ -214,6 +262,85 @@ where
     }
 }
 
+/// Emulates a SSZ `Vector`.
```
jimmygchen: This doesn't really cover SSZ encoding and decoding

jimmygchen:
I've just added the `6.0.0` label to this - I think it would be great if we can release this with the `getBlobsV1` changes, as we're keen to get some data on blob count increase with the optimisation. Ideally this get merged relatively soon given the amount of files touched, however this is a nice to have, so we can drop this in favour of higher priority issues.

consensus/types/src/runtime_var_list.rs
```diff
@@ -62,21 +82,46 @@ impl<T> RuntimeVariableList<T> {
...
+    /// without first setting `max_len`.
+    pub fn empty_uninitialized() -> Self {
+        Self {
+            vec: vec![],
+            max_len: None,
+        }
+    }
+
+    pub fn set_max_len(&mut self, max_len: usize) {
+        self.max_len = Some(max_len);
```
jimmygchen: probably should try to catch unexpected mutations after the `max_len` is already set, or only allow it to be set once (would `OnceCell` be an overkill for this?)

consensus/types/src/runtime_var_list.rs
```diff
@@ -62,21 +82,46 @@ impl<T> RuntimeVariableList<T> {
...
+            return None;
+        };
+        Some(self.vec.as_mut_slice())
+    }
+
+    /// Returns an instance of `Self` with max_len = None.
+    ///
+    /// No mutating operation can be performed on an uninitialized instance
+    /// without first setting `max_len`.
+    pub fn empty_uninitialized() -> Self {
```
jimmygchen: feels like this function is for convenience? the little downside is having to make `max_len` optional when it's known most of the time. But I guess the convenience is worth it if we don't need to mutate it all.

consensus/types/src/runtime_var_list.rs
jimmygchen: Actually this function is not used at all - maybe we could drop it? it feels like it could be a potential footgun

beacon_node/http_api/src/block_id.rs
```diff
@@ -296,17 +296,24 @@ impl BlockId {
                     ))
                 })?
         } else {
-            BlobSidecarList::default()
+            BlobSidecarList::empty_uninitialized()
```
jimmygchen: `max_len` is not specified here - i think it's not an issue since it's not encoded when `BlobSidecarList` is used as a standalone object

beacon_node/http_api/src/block_id.rs
```diff
@@ -296,17 +296,24 @@ impl BlockId {
...
-                let list = blob_sidecar_list
+                let list: Vec<_> = blob_sidecar_list
                     .into_iter()
                     .filter(|blob_sidecar| vec.contains(&blob_sidecar.index))
                     .collect();
-                BlobSidecarList::new(list)
-                    .map_err(|e| warp_utils::reject::custom_server_error(format!("{:?}", e)))?
+                if let Some(max_len) = list
+                    .first()
+                    .map(|sidecar| chain.spec.max_blobs_per_block(sidecar.epoch()))
```
jimmygchen: we could use the `block` slot so we could avoid using `first`?

beacon_node/lighthouse_network/src/rpc/methods.rs
```diff
@@ -326,8 +333,13 @@ pub struct BlobsByRangeRequest {
...
 
 impl BlobsByRangeRequest {
+    /// This function provides an upper bound on number of blobs expected in
+    /// a certain slot range.
+    ///
+    /// Note: **must not** use for anything consensus critical, only for
+    /// bounds checking and rate limiting.
     pub fn max_blobs_requested<E: EthSpec>(&self) -> u64 {
-        self.count.saturating_mul(E::max_blobs_per_block() as u64)
+        self.count.saturating_mul(MAX_BLOBS_PER_BLOCK_CEILING)
```
jimmygchen: I think this may cause the node to hit the rate limit quicker because each range blob request will now consume  `max_blobs_requested` tokens from the quota

beacon_node/lighthouse_network/src/rpc/methods.rs
```diff
@@ -24,6 +24,13 @@ use types::{
 pub type MaxErrorLen = U256;
 pub const MAX_ERROR_LEN: u64 = 256;
 
+/// The max number of blobs we expect in the configs to set for compile time params.
+/// Note: This value is an estimate that we should use only for rate limiting,
+/// bounds checking and other non-consensus critical operations.
+///
+/// For exact value, we should always check the chainspec.
+pub const MAX_BLOBS_PER_BLOCK_CEILING: u64 = 16;
```
jimmygchen: I think we could even go with a higher number if it's for non critical checks and preventing DOS, so we don't run into issues in upcoming upgrades and we don't have to frequently update this?

consensus/types/src/chain_spec.rs
jimmygchen: I think I see what you're trying to do here - this function returns the `max_blobs_per_block` for the epoch, so in the future it could return `MAX_BLOBS_PER_BLOCK` for deneb, and `MAX_BLOBS_PER_BLOCK_ELECTRA` if we're in Electra?

The issue i see with returning the `default` is that if we start a devnet on `Deneb` with a different `MAX_BLOBS_PER_BLOCK`, we won't be able to interop with other clients (which is the main purpose of making this field configurable)

beacon_node/network/src/sync/network_context.rs
```diff
@@ -969,9 +981,16 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
                 let request = request.get_mut();
                 match request.add_response(blob) {
-                    Ok(Some(blobs)) => to_fixed_blob_sidecar_list(blobs)
-                        .map(|blobs| (blobs, seen_timestamp))
-                        .map_err(|e| (e.into(), request.resolve())),
+                    Ok(Some(blobs)) => {
+                        let max_len = if let Some(blob) = blobs.first() {
+                            self.chain.spec.max_blobs_per_block(blob.epoch()) as usize
+                        } else {
+                            6
```
jimmygchen: `add_response` only returns `Some` if all blob sidecars have been received, I think this may be unreachable? if we reach here it means `add_response` is buggy and might be worth doing some error handling?

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -1669,7 +1669,20 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
+                // The encoding of a VariableList is same as a regular vec.
+                let blobs: Vec<Arc<BlobSidecar<E>>> = Vec::<_>::from_ssz_bytes(blobs_bytes)?;
+                let blobs = if let Some(max_blobs_per_block) = blobs
+                    .first()
+                    .map(|blob| self.spec.max_blobs_per_block(blob.epoch()))
+                {
+                    BlobSidecarList::from_vec(blobs, max_blobs_per_block as usize)
+                } else {
+                    // This always implies that there were no blobs for this block_root
+                    BlobSidecarList::empty_uninitialized()
```
jimmygchen: This is fine for now I think, but note it may break if we end up encoding this within another object like `CacheItem` or `AvailableBlock` (I don't see any use of encoding these rn)

jimmygchen:
We also need to move `MAX_BLOBS_PER_BLOCK` from presets to config YAML files.

jimmygchen (changes requested):
I've added some comments, let me know what you think, thanks!

beacon_node/lighthouse_network/src/rpc/methods.rs
jxs: I thought about this for https://github.com/sigp/lighthouse/pull/6462, but as far as I understood this value may change per fork right?

jxs (commented):
Hi Pawan, this is awesome, I wasn't aware of this PR but this is what I had in mind with https://github.com/sigp/lighthouse/pull/6462, left a comment

beacon_node/lighthouse_network/src/rpc/methods.rs
pawanjay176: Yeah it can change, but like i mentioned in the comment, it is to be used only for rate limiting and bounds checking of the incoming rpc request, so we can bump the number up in preparation for a fork

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
pawanjay176: We are using a fixed seed, so it'll be consistent across runs

consensus/types/src/chain_spec.rs
pawanjay176: I see your point. However, I wanted to be conservative with making changes that could allow configurability with `max_blobs_per_block` in deneb since that might screw with mainnet.

From what I have seen in devnets, the electra/peerdas activation happens on the first epoch so we can get the configurability from the first epoch onwards. I didn't want to overengineer for a usecase that might not exist on devnets specially because it could lead to consensus issues on mainnet

jimmygchen: Yeah I agree being conservative on this is a good idea - although I think people testing increased blob count (for both peerdas / without peerdsa) would assume this works by just modifying `max_blobs_per_block` - but this wouldn't work without enabling peerdas. On the other hand, I also think once we test this on mainnet, it should be fine and after that we shouldn't randomly commit a modification to mainnet config.

However, I don't feel strongly about this, just wanted to bring this up and I do really see a good reason to keep it as it is! 👍

michaelsproul:
I think we need this for Electra spec v1.5.0-alpha.10, which introduces `MAX_BLOBS_PER_BLOCK_ELECTRA` as a config parameter for Electra.

michaelsproul:
I'm going to resolve conflicts on this to see if we can merge it to unstable before the alpha10 spec PRs.

consensus/types/src/chain_spec.rs
michaelsproul: This function is useful for implementing the Electra change, so lets keep it.

consensus/types/src/runtime_var_list.rs
michaelsproul: It's used in the doc tests, so I deleted it and then reinstated it.

beacon_node/http_api/src/block_id.rs
michaelsproul: done in https://github.com/sigp/lighthouse/pull/6329/commits/1095d60a40be20dd3c229b759fc3c228b51e51e3

consensus/types/src/runtime_var_list.rs
michaelsproul: I would like to get rid of this too, but it seems like it's a bit hard because there are cases where we want to construct a blob list but don't know the length, e.g. fetching from the database.

I think this approach works OK for now.

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
michaelsproul: I didn't see any reason to change this, so I've just deleted this TODO.

consensus/types/src/runtime_var_list.rs
michaelsproul: I did this in 440e85419940d4daba406d910e7908dd1fe78668

beacon_node/execution_layer/src/test_utils/execution_block_generator.rs
michaelsproul: I fixed this and plumbed through the chainspec

beacon_node/lighthouse_network/src/rpc/methods.rs
michaelsproul: Not sure what to do about this one.

beacon_node/lighthouse_network/src/rpc/methods.rs
michaelsproul: I've added a test for this in this commit https://github.com/sigp/lighthouse/pull/6329/commits/04b3743ec1e0b650269dd8e58b540c02430d1c0d

michaelsproul:
I think this is ready for re-review. I got all the tests passing.

consensus/types/src/runtime_var_list.rs
michaelsproul: Should probably return a result or something and error if it is already set

pawanjay176:
@michaelsproul Made a shot at removing `empty_uninitialized` in 0cd263f.

The only place where we required `empty_uninitialized` was when we were fetching blobs from the db by root. In the cases where there were no blobs for a given root, we did not know what max_len to provide to the runtime list. 

Got around this by defining a new enum that encapsulates that case. PTAL.

Fixing the tests now.

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -43,6 +43,63 @@ use types::data_column_sidecar::{ColumnIndex, DataColumnSidecar, DataColumnSidec
...
+impl<E: EthSpec> From<BlobSidecarList<E>> for BlobsSidecarListFromRoot<E> {
+    fn from(value: BlobSidecarList<E>) -> Self {
+        Self::Blobs(value)
+    }
+}
+
+impl<E: EthSpec> Encode for BlobsSidecarListFromRoot<E> {
+    fn as_ssz_bytes(&self) -> Vec<u8> {
+        match self {
+            Self::NoBlobs | Self::NoRoot => vec![],
```
michaelsproul: I think this is right, but can we add a test that asserts `vec![].as_ssz_bytes()` and `BlobSidecarListFromRoot::NoBlobs.as_ssz_bytes()` are equal? And maybe similarly for a vec of vecs, and a vec of `BlobSidecarListFromRoot` so we exercise the `ssz_append` impl as well

michaelsproul:
I ended up deleting the SSZ implementation, because we don't actually need it! I thought we'd use it for HTTP responses, but we know the max_len in this case and can construct an empty `BlobSidecarList`. So I went ahead and removed the places where we were storing `BlobSidecarListFromRoot` and just restored them to `BlobSidecarList`.

I'm not too worried about the stuff in `do_atomically_with_block_and_blobs`, because we will delete that as part of:

- https://github.com/sigp/lighthouse/issues/6559

pawanjay176:
Nice, this looks much cleaner.

beacon_node/lighthouse_network/src/rpc/methods.rs
jimmygchen: Maybe we could also update the quota accordingly?

https://github.com/sigp/lighthouse/blob/f8e31f62726de375cca7087b6d2ef6b9283a749f/beacon_node/lighthouse_network/src/rpc/config.rs#L112-L114

michaelsproul: or maybe we should solve this issue immediately:

- https://github.com/sigp/lighthouse/issues/6773

I might try plumbing ChainSpec in and see how bad it is

beacon_node/http_api/src/block_id.rs
```diff
@@ -304,30 +306,27 @@ impl BlockId {
...
             })?;
 
         let blob_sidecar_list_filtered = match indices {
             Some(vec) => {
                 let list: Vec<_> = blob_sidecar_list
-                    .into_iter()
+                    .iter()
                     .filter(|blob_sidecar| vec.contains(&blob_sidecar.index))
+                    .cloned()
                     .collect();
```
jimmygchen, line 325: Do we need to clone these? 

```rust
                let list: Vec<_> = blob_sidecar_list
                    .into_iter()
                    .filter(|blob_sidecar| vec.contains(&blob_sidecar.index))
                    .collect();
```

consensus/types/src/runtime_fixed_vector.rs
```diff
@@ -0,0 +1,78 @@
+/// Emulates a SSZ `Vector`.
```
jimmygchen: I think we can delete this line now as it no longer encode/decode?

beacon_node/network/src/sync/network_context.rs
jimmygchen: Should we remove this hard coded value?

jimmygchen (changes requested):
I've done another round of review. I think the main ones to address are these, and we can merge once these are addressed:
- https://github.com/sigp/lighthouse/pull/6329#discussion_r1908147990 
- https://github.com/sigp/lighthouse/pull/6329#discussion_r1908135225

beacon_node/lighthouse_network/src/rpc/methods.rs
michaelsproul: The ChainSpec thing seems too invasive for what I'm comfortable doing rn, so I've just bumped the quota

- https://github.com/sigp/lighthouse/issues/6773#issuecomment-2579221166

beacon_node/network/src/sync/network_context.rs
michaelsproul: @jimmygchen I'm not very familiar with this code, do you think it's OK to return an `InternalError` here if `blobs` is empty? I'm guessing we shouldn't be requesting single blobs for blocks that don't have them?

beacon_node/http_api/src/block_id.rs
michaelsproul: removed

beacon_node/network/src/sync/network_context.rs
jimmygchen: I had a deeper look and I think this code path is unreachable.

- This `on_single_blob_response` function can accept one of the following `rpc_event`:  a single blob sidecar, a stream termination or an RPC error. 
- The only scenario where blobs is empty here - if we get a stream termination before we get any blobs and if `expect_max_responses == false` then  blobs could be empty.  However `expect_max_responses` is always `true` for blob requests, so this is not possible - we'll always get `LookupVerifyError::NotEnoughResponsesReturned` if we get a stream termination before we get any blobs.

I think an `InternalError` makes sense. The existing behaviour (if we hit this code path somehow) is sending the empty blob list for processing, which doesn't help syncing. An error here is better because it would trigger error handling (e.g. retry)

beacon_node/lighthouse_network/src/rpc/methods.rs
jimmygchen: Thanks for raising the issue, I think we can track this in the issue.

beacon_node/network/src/sync/network_context.rs
pawanjay176: I looked into this and agree with Jimmy's assessment. The blob list being empty here would imply that we requested blobs for a block which had an empty kzg_commitments list which would be an internal error.

beacon_node/lighthouse_network/src/rpc/methods.rs
pawanjay176: Should we also change the CEILING value to 9 for the time being as that's the current value?

jimmygchen (approved):
The fix looks good, thanks! Lets' goooo 🚀

michaelsproul (approved):

beacon_node/lighthouse_network/src/rpc/methods.rs
jimmygchen: Oh I think we might still have an issue here due to this validation - it won't allow `max_blobs_requested` to be greater than `128 * 6`, I think we need to update the validation here:
https://github.com/sigp/lighthouse/blob/38388979db0bd672103984b017468b76087ce122/beacon_node/network/src/network_beacon_processor/rpc_methods.rs#L893-L894

jimmygchen:
We just found a bug with this:
https://github.com/sigp/lighthouse/blob/04b3743ec1e0b650269dd8e58b540c02430d1c0d/beacon_node/lighthouse_network/src/rpc/methods.rs#L336-L345

This function is used in a validation here:
https://github.com/sigp/lighthouse/blob/38388979db0bd672103984b017468b76087ce122/beacon_node/network/src/network_beacon_processor/rpc_methods.rs#L893-L894

and this will always fail if someone sends a range request with more than `max_request_blob_sidecars ` sidecars or `768 / 16 = 48` blocks.

We need to update the validation here to use the spec `max_blob_per_block` config instead of using a ceiling value.

pawanjay176:
@jimmygchen Nice catch. I moved the check from the rpc_methods to the decoding of the request. 
Also threaded the chainspec and current_fork to the `max_responses` function. Its not ideal but it isn't too bad imo because  it does make sense that the max_response size depends on the current_fork.

@jxs had to add the `ForkContext` into the rate limiter in 70917f7 .I did not like this initially but I feel that's the best way to avoid the ugliness with keeping different max values in sync. Having the fork context in the rate limiter doesn't seem ugly because rate limiting is also fork dependent. Happy to explore a different solution if you feel otherwise.

beacon_node/lighthouse_network/src/rpc/codec.rs
```diff
@@ -583,9 +585,21 @@ fn handle_rpc_request<E: EthSpec>(
                 )?,
             }),
         ))),
-        SupportedProtocol::BlobsByRangeV1 => Ok(Some(RequestType::BlobsByRange(
-            BlobsByRangeRequest::from_ssz_bytes(decoded_buffer)?,
-        ))),
+        SupportedProtocol::BlobsByRangeV1 => {
+            let req = BlobsByRangeRequest::from_ssz_bytes(decoded_buffer)?;
+            // TODO(pawan): change this to max_blobs_per_rpc_request in the alpha10 PR
+            let max_blobs = spec.max_blobs_per_block_by_fork(current_fork);
```
jimmygchen: Should we multiply this by `req.count`? Otherwise most range requests might be rejected?

pawanjay176: Ohh yeah good catch, I have a `max_blobs_requests_by_fork ` function in the alpha 10 specs PR that i was planning to use for this, but this will kill us until that gets merged

michaelsproul: It still looks wrong to me. Shouldn't it be:

```rust
if req.count * max_blobs_per_block > max_request_blob_sidecars {
   // error
}
```

jimmygchen (approved):
Looks great now \o/

jimmygchen:
@mergify queue

PR #6331, opened 2024-08-30
Return HTTP 404 for pruned blob requests

michaelsproul:
## Issue Addressed

Closes:

- https://github.com/sigp/lighthouse/issues/5640

## Proposed Changes

- Return a 404 if blob sidecars are requested and they do not exist in the database.
- Return a 400 if blob sidecars are requested for a pre-Deneb block.

This resolves both cases of ambiguity where we currently return an empty list of blobs. Now, the empty blobs response is reserved exclusively for post-Deneb blocks that actually do not have any blobs.

This is motivated in part by the above issue, but also by the growing desire to allow importing/exporting blobs. In this case, some extra checks on the responses from the database are prudent in case a partial set of blobs is imported (something which may be quite desirable).

This is a breaking change to Lighthouse's behaviour, but IMO is still in keeping with the spec, which allows 404s for blocks that are not found.

## Additional Info

I've added 2 new tests to confirm the new behaviour.

One oddity is that we will return `[]` for blocks without blobs _even if_ the blobs in that range have been pruned. E.g. if we make a request at `slot == 12` and `oldest_blob_slot == 32` and the block from slot 12 is post-Deneb with 0 blobs, then we will return `[]`. As long as we have the block stored we are capable of responding correctly, so we may as well. The alternative would be to artificially limit empty blob responses to _just_ slots `>= oldest_blob_slot`.

### DIFF
--- a/beacon_node/http_api/src/block_id.rs
+++ b/beacon_node/http_api/src/block_id.rs
@@ -274,10 +274,27 @@ impl BlockId {
             warp_utils::reject::custom_not_found(format!("beacon block with root {}", root))
         })?;
 
+        // Error if the block is pre-Deneb and lacks blobs.
+        let blob_kzg_commitments = block.message().body().blob_kzg_commitments().map_err(|_| {
+            warp_utils::reject::custom_bad_request(
+                "block is pre-Deneb and has no blobs".to_string(),
+            )
+        })?;
+
         // Return the `BlobSidecarList` identified by `self`.
-        let blob_sidecar_list = chain
-            .get_blobs(&root)
-            .map_err(warp_utils::reject::beacon_chain_error)?;
+        let blob_sidecar_list = if !blob_kzg_commitments.is_empty() {
+            chain
+                .store
+                .get_blobs(&root)
+                .map_err(|e| warp_utils::reject::beacon_chain_error(e.into()))?
+                .ok_or_else(|| {
+                    warp_utils::reject::custom_not_found(format!(
+                        "no blobs stored for block {root}"
+                    ))
+                })?
+        } else {
+            BlobSidecarList::default()
+        };
 
         let blob_sidecar_list_filtered = match indices.indices {
             Some(vec) => {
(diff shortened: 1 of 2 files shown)


beacon_node/http_api/src/block_id.rs
```diff
@@ -274,10 +274,27 @@ impl BlockId {
...
-        let blob_sidecar_list = chain
-            .get_blobs(&root)
-            .map_err(warp_utils::reject::beacon_chain_error)?;
+        let blob_sidecar_list = if !blob_kzg_commitments.is_empty() {
+            chain
+                .store
+                .get_blobs(&root)
+                .map_err(|e| warp_utils::reject::beacon_chain_error(e.into()))?
+                .ok_or_else(|| {
+                    warp_utils::reject::custom_not_found(format!(
```
jimmygchen, line 291: This looks like a good change to me. 

I'm not sure if we want to introduce new status codes, but there's also [`410 Gone`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/410) which would allow us to differentiate between block not found and expired blobs, without having to parse response message - but I'm also happy with 404.

In PeerDAS we'd also need a error response for nodes that are not capable of serving blobs data (custody columns < 64), and I _think_ we might be returning 404 too? 😅

jimmygchen (approved):
Minor comment but looks good to me! Feel free to merge

beacon_node/http_api/src/block_id.rs
michaelsproul: 410 sounds good, but we would need to add it to the spec: https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Beacon/getBlobSidecars.

Maybe we can advocate for that, while returning 404 in the meantime?

jimmygchen: Sounds good!

michaelsproul:
@mergify queue

PR #6336, opened 2024-08-30
Metadata request ordering

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Send metadata request only after we establish that we aren't disconnecting a peer for crossing the peer limit. We disconenct a lot of peers for reaching the peer limit and this would save a metadata stream which would probably add up over time.

Not sure if there was a specific reason why we would send the metadata request first.


AgeManning (approved):
Yeah. Dont think this was intentional. This code had moved around a bit, this ordering looks right to me

dapplion (approved):
Seems reasonable to me

jimmygchen:
@mergify queue

michaelsproul:
@mergify queue

michaelsproul:
@mergify queue

PR #6338, opened 2024-09-02
Update manual checkpoint sync content in Lighthouse book

chong-he:
## Issue Addressed

N/A

## Proposed Changes

- Update the content in manual checkpoint sync
- Update FAQ part about churn limit which has a fixed churn limit after Deneb. Updated it in a way so that it's easier to maintain


michaelsproul (approved):
Nice, thanks CK!

michaelsproul:
@mergify queue

PR #6340, opened 2024-09-02
Add blob count label to `DATA_COLUMN_SIDECAR_COMPUTATION` metric

jimmygchen:
## Issue Addressed

The `DATA_COLUMN_SIDECAR_COMPUTATION` metric currently lacks granularity, as it doesn't differentiate by the number of blobs computed.

## Proposed Changes

This PR introduces a blob count label to the `DATA_COLUMN_SIDECAR_COMPUTATION` metric. Additionally, it moves the metric timer into the compute function, ensuring that only successful computations are recorded, thereby improving the accuracy of the metrics.

Additionally the bucket is updated, so it captures times above 1s up to 10s.


beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -802,9 +802,7 @@ fn build_gossip_verified_data_columns<T: BeaconChainTypes>(
                     GossipDataColumnError::KzgNotInitialized,
                 ))?;
 
-            let timer = metrics::start_timer(&metrics::DATA_COLUMN_SIDECAR_COMPUTATION);
             let sidecars = blobs_to_data_column_sidecars(&blobs, block, kzg, &chain.spec)?;
```
dapplion: Why not call `discard_timer_on_break` here once?

jimmygchen: Oh yeah, that's a lot more tidy!

The reason I added the timer into that function is because we call `blobs_to_data_column_sidecars` from `fetch_blobs` in the other PR as well, so I wanted to make sure we don't miss the metric - but we could add the metrics to both places too, I think having `discard_timer_on_break` on the caller is indeed a lot more cleaner.

jimmygchen: Done in https://github.com/sigp/lighthouse/pull/6340/commits/7186d2817f31b7dacc2087df22caf6cd31695bfb

michaelsproul (approved):

dapplion (approved):

jimmygchen:
@mergify queue

michaelsproul:
@mergify queue

michaelsproul:
@mergify queue

PR #6342, opened 2024-09-03
Clarify validator monitor block log

michaelsproul:
## Proposed Changes

I got tripped up by the "Block from API" phrasing of this log, which is logged even for gossip blocks 🤨. Eventually I spotted the `src => gossip`, but I figure it's worth changing to something less misleading.


dapplion (approved):

pawanjay176 (approved):

jimmygchen:
@mergify queue

michaelsproul:
@mergify queue

michaelsproul:
@mergify queue

PR #6343, opened 2024-09-03
Enable `large_stack_frames` lint

michaelsproul:
## Proposed Changes

Enable the `large_stack_frames` lint and fix some existing failures while allowing others.

Using `Box::new` in the KZG utils prevents some stack growth, because even though the blob is (probably) still copied on the stack when it is passed as an argument to `Box::new`, by avoiding _returning_ `Blob` or assigning it to a stack variable, we avoid it contributing to the size of the stack frame for the calling function.

For other places that the lint tripped, like in tests and for the CLI app, I've opted to just ignore the lint for now. These failures are low impact and harder to fix because they involve lots of nested futures. I couldn't figure out why the `watch` tests were blowing out the stack so much.

## Additional Info

The clap failure might be resolved by moving the BN over to clap derive:

- https://github.com/sigp/lighthouse/issues/5900

### DIFF
--- a/beacon_node/beacon_chain/src/kzg_utils.rs
+++ b/beacon_node/beacon_chain/src/kzg_utils.rs
@@ -17,6 +17,10 @@ fn ssz_blob_to_crypto_blob<E: EthSpec>(blob: &Blob<E>) -> Result<KzgBlob, KzgErr
     KzgBlob::from_bytes(blob.as_ref()).map_err(Into::into)
 }
 
+fn ssz_blob_to_crypto_blob_boxed<E: EthSpec>(blob: &Blob<E>) -> Result<Box<KzgBlob>, KzgError> {
+    ssz_blob_to_crypto_blob::<E>(blob).map(Box::new)
+}
+
 /// Converts a cell ssz List object to an array to be used with the kzg
 /// crypto library.
 fn ssz_cell_to_crypto_cell<E: EthSpec>(cell: &Cell<E>) -> Result<KzgCellRef, KzgError> {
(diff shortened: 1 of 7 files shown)


beacon_node/beacon_chain/src/kzg_utils.rs
```diff
@@ -104,7 +108,7 @@ pub fn compute_blob_kzg_proof<E: EthSpec>(
     blob: &Blob<E>,
```
jimmygchen, line 108: There's one on line 99 in `validate_blobs` as well

michaelsproul: I left that one for 2 reasons:

- `verify_blob_kzg_proof_batch` expects a slice of blobs: `&[Blob]`
- The allocations are being stored in a vec, so they're already on the heap

dapplion:
What a super helpful lint, nice find!

beacon_node/beacon_chain/src/kzg_utils.rs
jimmygchen: ah got it thanks

jimmygchen: btw the stack overflow in debug mode suggested that they're on the stack until the `collect` happens (where I'm getting stack overflow).

image

Not related to this PR though and the change is still good!

jimmygchen: I'm not sure if we can fix this without changing the kzg lib API though

jimmygchen (approved):
LGTM! 🎉

beacon_node/beacon_chain/src/kzg_utils.rs
michaelsproul: 1 blob at a time will definitely be on the stack, but not all of them

michaelsproul: I guess 1 blob could be enough to stack overflow if there's a bunch of other stuff already on the stack

michaelsproul:
@mergify queue

PR #6344, opened 2024-09-03
Remove beta tag from gossipsub 1.2

AgeManning:
I was watching the network metrics to see how many nodes on mainnet had upgraded and are using gossipsub 1.2. 

I noticed the metrics still identify as 1.2-beta. This is not the actual protocol id just a naming. This PR removes the beta naming, to clean up the metrics.


jxs (approved):
LGTM!

pawanjay176 (commented):
Looks good. 
Tests are failing, need to change the variant names in there too

jxs:
@mergifyio queue

PR #6355, opened 2024-09-04
Output network-test logs into files in CI

ackintosh:
## Issue Addressed

https://github.com/sigp/lighthouse/pull/6322#issuecomment-2326834731

We need the logs when debugging network test errors in CI. However, there's a problem of log interleaving if we run tests concurrently.

## Proposed Changes

Added `ci_logger` feature flag, which enables outputting logs to files for each test. Also added an `Upload logs` step to the `network-tests` job, where the log files are uploaded as [GitHub Artifacts](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/storing-and-sharing-data-from-a-workflow).

The uploaded log files (Artifacts) can be downloaded from [Summary](https://github.com/sigp/lighthouse/actions/runs/10721221086?pr=6355) page of the CI run.

image


jimmygchen (approved):
Very nice, thanks @ackintosh !

jimmygchen:
@mergify queue

Issue #6356, opened 2024-09-05
Remove unreachable void patterns once MSRV >= 1.82


michaelsproul:
## Description

In Rust 1.82 the compiler has gained the ability to recognise (and warn) for unreachable patterns due to uninhabited types like `void::Void`, `!` and `std::convert::Infallible`.

In order to keep Lighthouse compiling on 1.82 beta _and_ earlier versions, we've added `#[allow(unreachable_patterns)]` in a few places. Once the MSRV is 1.82 or greater we can remove these `allow`s and the match arms that they apply to.

## Present Behaviour

Describe the present behaviour of the application, with regards to this
issue.

## Steps to resolve

Once Rust 1.82 is stable and we're OK with increasing the MSRV:

1. Increase the MSRV to 1.82
2. Remove all instances of `void::unreachable` and `allow(unreachable_patterns)` found using `grep`.

PR #6357, opened 2024-09-05
Ignore Rust 1.82 warnings about void patterns

michaelsproul:
## Proposed Changes

Fix compilation with Rust 1.82 beta, which now runs on CI.

We cannot remove these patterns entirely yet, because we still need to be able to compile with older versions of Rust. I've opened an issue to track their eventual removal:

- https://github.com/sigp/lighthouse/issues/6356


jimmygchen (approved):
LGTM, lets merge after CI passes

jimmygchen:
Beta compiler passed, ~~so I'm adding this to the merge queue.~~ I'll let it finish so we can get this insta-merged.

jimmygchen:
@mergify queue

Issue #4022, opened 2023-02-22
Fix `merkle_proof_validity` consensus-spec-test


paulhauner:
## Description

In #4021 we skipped the `merke_proof_validity` test. It is a `light_client` test and AIUI is failing due to some Capella changes we need to catch up with.

The failure is:

```
---- merkle_proof_validity stdout ----
Passed 3 tests in /Users/paul/development/lighthouse/testing/ef_tests/consensus-spec-tests/tests/mainnet/altair/light_client/single_merkle_proof
Passed 3 tests in /Users/paul/development/lighthouse/testing/ef_tests/consensus-spec-tests/tests/mainnet/bellatrix/light_client/single_merkle_proof
thread 'merkle_proof_validity' panicked at 'test should load: FailedToParseTest("Unable to parse SSZ at /Users/paul/development/lighthouse/testing/ef_tests/consensus-spec-tests/tests/mainnet/capella/light_client/single_merkle_proof/BeaconBlockBody/execution_merkle_proof/object.ssz_snappy: InvalidByteLength { len: 900, expected: 262320 }")', /Users/paul/development/lighthouse/testing/ef_tests/src/handler.rs:68:72
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

PR #6361, opened 2024-09-05
Fix electra light client types

### LINKED ISSUE #6002: Fix electra light client types
## Description

Noticed the EF SSZ tests were failing for some light client types. Looks like it's due to `FINALIZED_ROOT_GINDEX`, `CURRENT_SYNC_COMMITTEE_GINDEX` and `NEXT_SYNC_COMMITTEE_GINDEX` being updated, resulting in the lengths of fields in some types being updated

- `LightClientUpdate`
- `LightClientFinalityUpdate`
- `LightClientBootstrap`

This spec PR obviates the problem and makes more electra related updates: 

https://github.com/ethereum/consensus-specs/pull/3811/files#diff-3c5f5d62f6d1d14d3324c9c37b025560ee9ccb335770a28f553ed374bc56e385R31

eserilev:
## Issue Addressed

Resolves #6002 

Resolves https://github.com/sigp/lighthouse/issues/4022

## Proposed Changes

Update light client types to be compatible with electra

Related spec PR:

https://github.com/ethereum/consensus-specs/pull/3811

Blocked by #5915


eserilev:
The merkle proof calculations are a bit hairy now. I think stable containers could be a big help here. We could also try to do some generalizations here in the interim. Open to suggestions!

beacon_node/beacon_chain/src/light_client_server_cache.rs
```diff
@@ -432,15 +423,13 @@ impl<E: EthSpec> LightClientCachedData<E> {
     fn from_state(state: &mut BeaconState<E>) -> Result<Self, BeaconChainError> {
         Ok(Self {
             finalized_checkpoint: state.finalized_checkpoint(),
-            finality_branch: state.compute_merkle_proof(FINALIZED_ROOT_INDEX)?.into(),
+            finality_branch: state.compute_merkle_proof(FINALIZED_ROOT_INDEX_ELECTRA)?,
```
dapplion: Why hardcode to the electra Gindex here?

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -641,15 +640,14 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
     pub fn get_sync_committee_branch(
         &self,
         block_root: &Hash256,
-    ) -> Result<Option<FixedVector<Hash256, CurrentSyncCommitteeProofLen>>, Error> {
+    ) -> Result<Option<Vec<Hash256>>, Error> {
```
dapplion: Worth to define an alias `Proof = Vec<Hash256>`

consensus/types/src/light_client_update.rs
```diff
@@ -387,7 +416,24 @@ impl<E: EthSpec> LightClientUpdate<E> {
...
+            LightClientUpdate::Capella(update) => {
+                self.is_empty_branch(update.next_sync_committee_branch.as_ref())
+            }
+            LightClientUpdate::Altair(update) => {
+                self.is_empty_branch(update.next_sync_committee_branch.as_ref())
+            }
+        }
+    }
+
+    fn is_empty_branch(&self, branch: &[Hash256]) -> bool {
```
dapplion: Un-used self you can move the method outside the impl

consensus/types/src/light_client_update.rs
```diff
@@ -419,13 +473,25 @@ mod tests {
     use crate::MainnetEthSpec;
     use ssz_types::typenum::Unsigned;
 
+    // TODO(light-client) should run for every fork?
```
dapplion: If there's no downside, yes?

consensus/types/src/beacon_state.rs
```diff
@@ -2540,8 +2545,14 @@ impl<E: EthSpec> BeaconState<E> {
         };
 
         // 3. Make deposit tree.
-        // Use the depth of the `BeaconState` fields (i.e. `log2(32) = 5`).
-        let depth = light_client_update::CURRENT_SYNC_COMMITTEE_PROOF_LEN;
+        let depth = if let BeaconState::Electra(_) = self {
```
dapplion: Why not do `log2(self.num_fields_pow2)`?

consensus/types/src/beacon_state.rs
```diff
@@ -2540,8 +2545,14 @@ impl<E: EthSpec> BeaconState<E> {
         };
 
         // 3. Make deposit tree.
```
dapplion: Not a deposit tree

consensus/types/src/beacon_state.rs
```diff
@@ -2478,9 +2478,12 @@ impl<E: EthSpec> BeaconState<E> {
     }
 
     pub fn compute_merkle_proof(&self, generalized_index: usize) -> Result<Vec<Hash256>, Error> {
+        println!("generalized index {}", generalized_index);
```
dapplion: Warning, println

consensus/types/src/beacon_state.rs
```diff
@@ -2478,9 +2478,12 @@ impl<E: EthSpec> BeaconState<E> {
     }
 
     pub fn compute_merkle_proof(&self, generalized_index: usize) -> Result<Vec<Hash256>, Error> {
```
dapplion: The API of this function is quite strange, as it appears like a generic function but only allows to generate proofs of a very specific and small set of `generalized_index`. Maybe instead we should make this one private and expose
```
compute_current_sync_committee_proof()
compute_next_sync_committee_proof()
compute_finalized_root_proof()
```

beacon_node/beacon_chain/src/light_client_server_cache.rs
eserilev: fixed here: [2338ec3](https://github.com/sigp/lighthouse/pull/6361/commits/2338ec33edfe2dfc4498ba14b52c59dbd02c6ee5)

beacon_node/store/src/hot_cold_store.rs
eserilev: added type alias [aa08704](https://github.com/sigp/lighthouse/pull/6361/commits/aa08704bb940e1996a5600e2b5e5da5e530df948)

consensus/types/src/light_client_update.rs
eserilev: moved out of impl [b23642e](https://github.com/sigp/lighthouse/pull/6361/commits/b23642e7f989c92e92bcaf6e901d22888479fcf3)

consensus/types/src/light_client_update.rs
eserilev: add ssz tests to all forks for all lc types [9341d87](https://github.com/sigp/lighthouse/pull/6361/commits/9341d87127e1d99ea841e1321a0ed1e5faa5e76a)

note that the macro can only defined once per namespace

consensus/types/src/beacon_state.rs
eserilev: [9341d87](https://github.com/sigp/lighthouse/pull/6361/commits/9341d87127e1d99ea841e1321a0ed1e5faa5e76a)

consensus/types/src/beacon_state.rs
eserilev: updated api [9341d87](https://github.com/sigp/lighthouse/pull/6361/commits/9341d87127e1d99ea841e1321a0ed1e5faa5e76a)

I've introduced two public methods

```
compute_sync_committee_proof
compute_finalized_root_proof
```

since current and next sync committee proof logic is nearly identical (minus the generalized indices) I combined them into one method.

consensus/types/src/beacon_state.rs
```diff
@@ -2493,6 +2495,22 @@ impl<E: EthSpec> BeaconState<E> {
...
+        // 3. generate proof
+        self.generate_proof(field_index, &leaves)
+    }
+
+    pub fn compute_finalized_root_proof(
+        &self,
+        generalized_index: usize,
+    ) -> Result<Vec<Hash256>, Error> {
+        // 1. Convert generalized index to field index.
+        let field_index = match generalized_index {
```
dapplion: Just match on the self.fork_unchecked() and pick the right gindex. No need to pass the generalized_index as argument. I guess now that the function is repeated we don't need to list the 1,2,3,4 steps on each function. Same for compute_sync_committee_proof if you split the function into `compute_next_sync_committee_proof` and `compute_current_sync_committee_proof` they are short functions that are no longer fallible. It simplifies the logic on the caller side too

dapplion (approved):
LGTM! Until we have generalized proof generation this version looks like the most sound to me

consensus/types/src/light_client_update.rs
```diff
@@ -24,20 +24,36 @@ pub const CURRENT_SYNC_COMMITTEE_INDEX: usize = 54;
...
 pub const FINALIZED_ROOT_PROOF_LEN: usize = 6;
 pub const CURRENT_SYNC_COMMITTEE_PROOF_LEN: usize = 5;
 pub const NEXT_SYNC_COMMITTEE_PROOF_LEN: usize = 5;
 pub const EXECUTION_PAYLOAD_PROOF_LEN: usize = 4;
 
+pub const FINALIZED_ROOT_PROOF_LEN_ELECTRA: usize = 7;
+pub const NEXT_SYNC_COMMITTEE_PROOF_LEN_ELECTRA: usize = 6;
+pub const CURRENT_SYNC_COMMITTEE_PROOF_LEN_ELECTRA: usize = 6;
+
+pub type Proof = Vec<Hash256>;
```
michaelsproul: I think maybe we should use a more specific name for this type, as there are many kinds of proofs floating around, especially with peer DAS and EIP4844.

```suggestion
pub type MerkleProof = Vec<Hash256>;
```

michaelsproul (dismissed):
Looks good. I just pushed a small change to reduce new fork burden: 7d67d0d1c1cd061a90c12a78c30408c21715428d

One small naming nit, and then we can merge

dapplion (approved):
LGTM

michaelsproul (approved):

michaelsproul:
@mergify queue

PR #6371, opened 2024-09-09
Add more granularity to data column proof computation metric

jimmygchen:
## Issue Addressed

Add more granularity to data column proof computation metric to capture more variations between 0.25 and 1 second.

We currently get a lot of data point at 500ms and some at 1s, would be good to get more granularity there on computation time.


dapplion (approved):

jimmygchen:
@mergify queue

jimmygchen:
@mergify dequeue

jimmygchen:
@mergify queue

PR #6377, opened 2024-09-09
Attribute invalid column proof error to correct peer

dapplion:
## Issue Addressed

Before PeerDAS we send to the beacon processor a "block component" which is a set of things to verify, crucially all coming from the same peer. If the processor says "invalid" we just penalize the unique provider peer.

This (block component) -> (1 peer) link breaks for PeerDAS. Now the set of custody columns may come from a set of more than 1 peer. We need a way for the beacon processor to attribute fault to the right peer.

## Proposed Changes

- When verifying a batch of columns, if there's a failure re-verify individually to attribute fault. This adds CPU time but a failure will cause the offending peer to be quickly banned so it's not a DoS vector.
- Extend Kzg verification errors to include the index of the invalid column
- Update lookup sync handling of processing results to locate the provider of the invalid column and penalize it

## Caveats

The re-verification of each individual column could be done more efficiently. However, we rarely get invalid columns so it's not critical to optimize this path. Offending peers will be quickly banned.

To upload the statement above I think we should increase the penalty to invalid processing results from mid-tolerance to high tolerance.


beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -199,9 +199,11 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
             .now_duration()
             .ok_or(AvailabilityCheckError::SlotClockError)?;
 
+        // Note: currenlty not reporting which specific blob is invalid because we fetch all blobs
```
pawanjay176:
```suggestion
        // Note: currently not reporting which specific blob is invalid because we fetch all blobs
```

pawanjay176 (approved):

beacon_node/beacon_chain/src/data_availability_checker.rs
dapplion: Oh sorry missed your suggestion when porting the changes 🙏

realbigsean (approved):
nice

realbigsean:
@mergify queue

PR #6382, opened 2024-09-10
Set custody subnets on tests

ackintosh:
## Issue Addressed

Closes https://github.com/sigp/lighthouse/issues/6319

## Proposed Changes

1. Update the `__add_connected_peer_testing_only` to set custody subnets
2. Add an additional supernode in the `sampling_with_retries` test to ensure that the node can retry

### DIFF
--- a/beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
+++ b/beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
@@ -1,5 +1,5 @@
 use crate::discovery::enr::PEERDAS_CUSTODY_SUBNET_COUNT_ENR_KEY;
-use crate::discovery::CombinedKey;
+use crate::discovery::{peer_id_to_node_id, CombinedKey};
 use crate::{metrics, multiaddr::Multiaddr, types::Subnet, Enr, EnrExt, Gossipsub, PeerId};
 use peer_info::{ConnectionDirection, PeerConnectionStatus, PeerInfo};
 use rand::seq::SliceRandom;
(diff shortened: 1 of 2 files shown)


beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
```diff
@@ -723,6 +723,17 @@ impl<E: EthSpec> PeerDB<E> {
                 .map(|csc| csc.into())
                 .collect();
             peer_info.set_custody_subnets(all_subnets);
+        } else {
+            let peer_info = self.peers.get_mut(&peer_id).expect("peer exists");
+            let node_id = peer_id_to_node_id(&peer_id).expect("convert peer_id to node_id");
+            let subnets = DataColumnSubnetId::compute_custody_subnets::<E>(
+                node_id.raw(),
+                spec.custody_requirement + 4,
```
ackintosh: `spec.custody_requirement` did not fix the flakiness; the failure where `cx.get_custodial_peers(self.column_index)` returns only a supernode still occurs. Based on my testing, `spec.custody_requirement + 4` (or more) seems to fix it.

dapplion: Why not have two supernodes in the retry test + `set_custody_subnets` in all tests for completeness?

ackintosh: That makes sense to me💡 Thanks!

jimmygchen: Yeah the random generated peer ids doesn’t guarantee that peers are distributed across all subnets evenly. we need to either depend on more super nodes or have deterministic node ids in the test. In practice, LH will need to continuously maintain peers across all subnets.

jimmygchen (approved):
Looks good to me 👍  Thanks!

jimmygchen:
@mergify queue

PR #6383, opened 2024-09-10
Light client updates by range RPC

eserilev:
## Issue Addressed

Partially #3651

## Proposed Changes

Enable light client updates request over RPC


beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -380,6 +381,94 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+    ) -> Result<(), (RPCResponseErrorCode, &'static str)> {
+        debug!(self.log, "Received LightClientUpdatesByRange Request";
+            "peer_id" => %peer_id,
+            "count" => req.count,
+            "start_period" => req.start_period,
+        );
+
+        // TODO(lc-update) use correct val here
+        // Should not send more than max light client updates
+        let max_request_size: u64 = 1;
```
dapplion: 1 is definitely too low

consensus/types/src/light_client_update.rs
```diff
@@ -403,6 +404,31 @@ impl<E: EthSpec> LightClientUpdate<E> {
         }
         true
     }
+
+    // TODO(lc-update) does this math look correct?
```
dapplion: Yes looks good, you can link the spec and that mention that an update has two headers

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -380,6 +381,94 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+        });
+
+        let lc_updates_sent = lc_updates.len();
+
+        if lc_updates_sent < req.count as usize {
+            debug!(
+                self.log,
+                "LightClientUpdatesByRange outgoing response processed";
+                "peer" => %peer_id,
+                "msg" => "Failed to return all requested light client updates",
```
dapplion: Could be that the peer requested updates ahead of what's currently available, not our fault.

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -380,6 +381,94 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                error!(self.log, "Unable to obtain light client updates";
+                    "request" => ?req,
+                    "peer" => %peer_id,
+                    "error" => ?e
+                );
+                return Err((RPCResponseErrorCode::ServerError, "Database error"));
+            }
+        };
+
+        lc_updates.iter().for_each(|lc_update| {
```
dapplion: Why not for loop?

beacon_node/lighthouse_network/src/rpc/protocol.rs
```diff
@@ -542,8 +574,13 @@ impl ProtocolId {
                 <LightClientBootstrapRequest as Encode>::ssz_fixed_len(),
                 <LightClientBootstrapRequest as Encode>::ssz_fixed_len(),
             ),
+            // TODO(lc-update) make sure rpc limits are set correctly for lc data
```
dapplion: Both `LightClientOptimisticUpdate` and `LightClientFinalityUpdate` have empty requests, so the limits look ok. And `LightClientUpdatesByRangeRequest` has fixed size and your min / max functions are correct.

beacon_node/lighthouse_network/src/peer_manager/mod.rs
```diff
@@ -567,11 +567,13 @@ impl<E: EthSpec> PeerManager<E> {
                     Protocol::BlocksByRange => PeerAction::MidToleranceError,
                     Protocol::BlocksByRoot => PeerAction::MidToleranceError,
                     Protocol::BlobsByRange => PeerAction::MidToleranceError,
+                    // TODO(lc-update) review this comment and logic
```
dapplion: Looks good to me, we may implement a light-client "client" in the future, and add the boilerplate then

beacon_node/lighthouse_network/src/rpc/methods.rs
```diff
@@ -477,6 +478,41 @@ impl DataColumnsByRootRequest {
...
+pub struct LightClientUpdatesByRangeRequest {
+    /// The starting period to request light client updates.
+    pub start_period: u64,
+    /// The number of periods from `start_period`.
+    pub count: u64,
+}
+
+impl LightClientUpdatesByRangeRequest {
+    pub fn max_requested<E: EthSpec>(&self) -> u64 {
+        // TODO(lc-update) check spec for max request limits
```
dapplion: `MAX_REQUEST_LIGHT_CLIENT_UPDATES = 128`

https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/p2p-interface.md#configuration

beacon_node/lighthouse_network/src/rpc/methods.rs
```diff
@@ -477,6 +478,41 @@ impl DataColumnsByRootRequest {
...
+    pub fn ssz_min_len() -> usize {
+        LightClientUpdatesByRangeRequest {
+            start_period: 0,
+            count: 0,
+        }
+        .as_ssz_bytes()
+        .len()
+    }
+
+    pub fn ssz_max_len(_: &ChainSpec) -> usize {
```
dapplion: No need to pass spec, it's fixed size

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
dapplion: Use `MAX_REQUEST_LIGHT_CLIENT_UPDATES` defined as a constant here. It's not in the config or spec so we have to hardcode here.

dapplion (approved):

dapplion (approved):

beacon_node/lighthouse_network/src/rpc/methods.rs
```diff
@@ -477,6 +482,39 @@ impl DataColumnsByRootRequest {
...
+#[derive(Encode, Decode, Clone, Debug, PartialEq)]
+pub struct LightClientUpdatesByRangeRequest {
+    /// The starting period to request light client updates.
+    pub start_period: u64,
+    /// The number of periods from `start_period`.
+    pub count: u64,
+}
+
+impl LightClientUpdatesByRangeRequest {
+    pub fn max_requested<E: EthSpec>(&self) -> u64 {
```
jxs: we don't need the generic parametrization here right?

jxs (commented):
Thanks for adapting to the changes I introduced Eitan! Left a comment

beacon_node/lighthouse_network/src/rpc/methods.rs
eserilev: thanks, i had meant to move this value to `EthSpec`

[5928632](https://github.com/sigp/lighthouse/pull/6383/commits/59286323226c11154bebcc011545015d23c60d5e)

jxs: no worries Eitan, but can we remove entirely the `EthSpec` dependency from this structure and not have this function here? See https://github.com/sigp/lighthouse/pull/6462#discussion_r1792315770 for the why.

eserilev: ah ok makes sense. Ive removed the ethspec dependency

0798629

dapplion (approved):
Re-approved for the 3rd time, maybe this one is the charm? :)

jxs (approved):
Thanks Eitan, looks good to me now!

beacon_node/lighthouse_network/src/rpc/methods.rs
```diff
@@ -477,6 +478,39 @@ impl DataColumnsByRootRequest {
...
+        .len()
+    }
+
+    pub fn ssz_max_len() -> usize {
+        LightClientUpdatesByRangeRequest {
+            start_period: 0,
+            count: 0,
+        }
+        .as_ssz_bytes()
+        .len()
```
michaelsproul:
```suggestion
        Self::ssz_min_len()
```

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -428,6 +429,105 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+            ));
+        }
+
+        let lc_updates = match self
+            .chain
+            .get_light_client_updates(req.start_period, req.count)
+        {
+            Ok(lc_updates) => lc_updates,
+            Err(e) => {
+                error!(self.log, "Unable to obtain light client updates";
```
michaelsproul, line 487: This error can happen if we just don't have the old updates available, right? Maybe this log should be debug level

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -428,6 +429,105 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+        }
+
+        let lc_updates_sent = lc_updates.len();
+
+        if lc_updates_sent < req.count as usize {
+            debug!(
+                self.log,
+                "LightClientUpdatesByRange outgoing response processed";
+                "peer" => %peer_id,
+                "msg" => "Failed to return all requested light client updates. The peer may have requested data ahead of whats currently available.",
```
michaelsproul: I think we're still trying to avoid `msg` tags, as they collide with the main log message when logged in ELK.

```suggestion
                "info" => "Failed to return all requested light client updates. The peer may have requested data ahead of whats currently available",
```

michaelsproul (approved):
Looks good on the whole, just a few nits

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
dapplion: Calls to `iter_column_from` can error if there's no data on the range? Or based on the split position?

eserilev: Calls to `iter_column_from` shouldn't error if there's no data. i believe it'd just return an empty iterator in that case.  the error cases here are
- ssz decoding failures
- `remove_column_variable` error

so i think an error log in this case makes sense as it means theres something seriously wrong with the light client data stored in the db

jimmygchen (approved):

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
michaelsproul: ah right, didn't look under the hood at the `iter_column_from`

this is fine then

michaelsproul:
@mergify queue

PR #6392, opened 2024-09-13
Include compiler version in `lighthouse --version`

michaelsproul:
## Proposed Changes

In aid of making builds easier to reproduce, include the compiler's version information in `lighthouse --version`.

Initial testing for reproducible builds shows that Lighthouse can't (yet) be built bit-identically by `make build-x86_64` for the same commit hash & compiler version :(

See:

### DIFF
--- a/lighthouse/src/main.rs
+++ b/lighthouse/src/main.rs
@@ -35,6 +35,7 @@ pub static LONG_VERSION: LazyLock<String> = LazyLock::new(|| {
          SHA256 hardware acceleration: {}\n\
          Allocator: {}\n\
          Profile: {}\n\
+         Compiler: {}\n\
          Specs: mainnet (true), minimal ({}), gnosis ({})",
         SHORT_VERSION.as_str(),
         bls_library_name(),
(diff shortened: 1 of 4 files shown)


michaelsproul:
Related:

lighthouse/src/main.rs
```diff
@@ -85,6 +87,29 @@ fn build_profile_name() -> String {
         .to_string()
 }
 
+fn compiler_version() -> String {
+    let version = rustc_version::version_meta().expect("should be able to fetch compiler version");
```
dapplion, line 91: This will error if the command `rustc -vV` fails to run.

https://docs.rs/rustc_version/0.4.1/src/rustc_version/lib.rs.html#201
https://docs.rs/rustc_version/0.4.1/src/rustc_version/lib.rs.html#180

Is this run on compilation of on every run of LH?

michaelsproul: oh shit I think you're right, this is running at runtime 🤦‍♂️

michaelsproul:
I don't know how to fix this easily, and it's low prio so I'm not going to spend any time on it.

Issue #6393, opened 2024-09-13
Reproducible builds


michaelsproul:
## Description

There is some desire to build Lighthouse in a [_reproducible_ way](https://reproducible-builds.org/). Making Lighthouse reproducible means that two source builds of Lighthouse at the same commit and with the same toolchain would produce bit-identical binaries.

This issue is an umbrella issue to track progress towards this goal, and document cases where Lighthouse can already be built reproducibly.

## Working Reproducible Builds

- On my M1 Pro Mac, building Lighthouse twice in a row with `make reproducible` results in identical binaries.

## Failing Reproducible Builds

- On my x86_64 Linux desktop, building Lighthouse twice in a row under Cross (`make build-x86_64`) doesn't produce an identical binary.
- On my x86_64 Linux desktop, building Lighthouse twice in a row on the host (`make reproducible`) doesn't produce an identical binary.
- Building with Cross on our release builder also doesn't produce the same binary as one built locally with Cross.

## Steps to resolve

Unclear.

I have no idea why the Cross builds are more variable than the macOS builds, when one would naively expect the Docker image to provide a stable toolchain/etc. Perhaps the default Linux linker is more random by default than the macOS linker?

jmcph4:
Unlikely to completely solve our issues but this might explain some of the Mac vs Linux differences:

 > `packed` - This is the default for Windows MSVC and macOS. The term "packed" here means that all the debug information is packed into a separate file from the main executable. On Windows MSVC this is a `*.pdb` file, on macOS this is a `*.dSYM` folder, and on other platforms this is a `*.dwp` file.
>
 > `unpacked` - This means that debug information will be found in separate files for each compilation unit (object file). This is not supported on Windows MSVC. On macOS this means the original object files will contain debug information. On other Unix platforms this means that `*.dwo` files will contain debug information.

 -- https://doc.rust-lang.org/rustc/codegen-options/index.html#split-debuginfo

Furthering this, is that Cargo's default `release` profile (which I note your PR uses) has a platform-specific value for this flag.

jmcph4:
Okay, the current state of things for me locally:

```
$ cargo clean
$ make reproducible
$ cp target/reproducible/lighthouse prog1
$ cargo clean
$ cp target/reproducible/lighthouse prog2
$ xxd prog1 > prog1.txt
$ xxd prog2 > prog2.txt
$ diff prog1.txt prog2.txt
59,60c59,60
[...]

PR #6394, opened 2024-09-13
Attempt at reproducible builds

michaelsproul:
## Issue Addressed

Closes:

## Proposed Changes

- Add `make reproducible`. This is tested and working on Linux x86_64 and macOS aarch64.
- Update the `cross` builds to build in a reproducible way by default.
- Add the compiler version to `lighthouse --version`.

The key insight is setting `SOURCE_DATE_EPOCH` so OpenSSL doesn't include the build time in the binary. Thanks @jmcph4 for helping work this out.

## Limitations

There is no reproducibility between builds outside `cross` and builds within `cross`. The binary diff is large and I haven't worked out what it could be (yet). It might be due to different `libc` versions.


michaelsproul:
Dropping this from v6.0.0. No bandwidth to work on it at the moment.

michaelsproul:
Closing in favour of a more comprehensive PR from Flashbots:

- https://github.com/sigp/lighthouse/pull/6799

PR #6396, opened 2024-09-16
Improve single-node testnet support and Arc NetworkConfig/ChainSpec

michaelsproul:
## Proposed Changes

Got a bit carried away with a yak shave.

- `Arc` the `ChainSpec` to make it immutable after initialization and to remove duplication.
- Add an `Arc`'d `NetworkConfig` to the `NetworkGlobals`
- Return `is_syncing: false` from `/eth/v1/node/syncing` when sync is stalled and `target-peers` is 0 (this is the change I set out to make 😅).


ethDreamer (approved):
LGTM!

michaelsproul:
@mergify queue

PR #6397, opened 2024-09-16
Simplify database anchor

michaelsproul:
## Proposed Changes

Based on a suggestion from @dapplion, simplify the database's anchor by making it non-optional.

Rippling this change through the codebase also allowed for some other simplifications including deleting the `BackFillState::NotRequired` enum variant, which was redundant. Previously there were two checks for a backfill sync having already completed: checking if the anchor was `None` (the `BackFillState::NotRequired` case) and checking if all desired blocks had been downloaded based on the anchor's `oldest_block_slot` (the `BackFillState::Complete` case). I see no need to handle these cases differently, so I've deleted the `NotRequired` variant in the process of deleting the code that handled `anchor == None` which is no longer a possible state.

### DIFF
--- a/beacon_node/beacon_chain/src/block_verification.rs
+++ b/beacon_node/beacon_chain/src/block_verification.rs
@@ -905,9 +905,6 @@ impl<T: BeaconChainTypes> GossipVerifiedBlock<T> {
 
         let block_root = get_block_header_root(block_header);
 
-        // Disallow blocks that conflict with the anchor (weak subjectivity checkpoint), if any.
-        check_block_against_anchor_slot(block.message(), chain)?;
-
         // Do not gossip a block from a finalized slot.
         check_block_against_finalized_slot(block.message(), block_root, chain)?;
 
@@ -1137,9 +1134,6 @@ impl<T: BeaconChainTypes> SignatureVerifiedBlock<T> {
             .as_block()
             .fork_name(&chain.spec)
             .map_err(BlockError::InconsistentFork)?;
-
-        // Check the anchor slot before loading the parent, to avoid spurious lookups.
-        check_block_against_anchor_slot(block.message(), chain)?;
 
         let (mut parent, block) = load_parent(block, chain)?;
 
@@ -1773,19 +1767,6 @@ impl<T: BeaconChainTypes> ExecutionPendingBlock<T> {
             payload_verification_handle,
         })
     }
-}
-
-/// Returns `Ok(())` if the block's slot is greater than the anchor block's slot (if any).
-fn check_block_against_anchor_slot<T: BeaconChainTypes>(
-    block: BeaconBlockRef<'_, T::EthSpec>,
-    chain: &BeaconChain<T>,
-) -> Result<(), BlockError> {
-    if let Some(anchor_slot) = chain.store.get_anchor_slot() {
-        if block.slot() <= anchor_slot {
-            return Err(BlockError::WeakSubjectivityConflict);
-        }
-    }
-    Ok(())
 }
 
 /// Returns `Ok(())` if the block is later than the finalized slot on `chain`.
--- a/beacon_node/beacon_chain/src/historical_blocks.rs
+++ b/beacon_node/beacon_chain/src/historical_blocks.rs
@@ -33,8 +33,6 @@ pub enum HistoricalBlockError {
     InvalidSignature,
     /// Transitory error, caller should retry with the same blocks.
     ValidatorPubkeyCacheTimeout,
-    /// No historical sync needed.
-    NoAnchorInfo,
     /// Logic error: should never occur.
     IndexOutOfBounds,
 }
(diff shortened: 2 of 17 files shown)


beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -905,9 +905,6 @@ impl<T: BeaconChainTypes> GossipVerifiedBlock<T> {
 
         let block_root = get_block_header_root(block_header);
 
-        // Disallow blocks that conflict with the anchor (weak subjectivity checkpoint), if any.
-        check_block_against_anchor_slot(block.message(), chain)?;
```
michaelsproul, line 909: These checks were unnecessary because the anchor slot is always older than finalization, so any block older than the anchor is also older than finalization and will be caught by `check_block_against_finalized_slot`

beacon_node/beacon_chain/src/historical_blocks.rs
```diff
@@ -62,10 +60,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         &self,
         mut blocks: Vec<AvailableBlock<T::EthSpec>>,
     ) -> Result<usize, Error> {
-        let anchor_info = self
-            .store
-            .get_anchor_info()
-            .ok_or(HistoricalBlockError::NoAnchorInfo)?;
+        let anchor_info = self.store.get_anchor_info();
```
michaelsproul, line 63: The case where backfill is already complete is handled by the early return below for an empty batch.

I could add a sanity check error here instead if desired like:

```rust
if anchor.oldest_block_slot == 0 {
   return Err(HistoricalBlockError::AlreadyComplete);
}
```

?

beacon_node/lighthouse_network/src/types/globals.rs
```diff
@@ -72,7 +72,7 @@ impl<E: EthSpec> NetworkGlobals<E> {
             peers: RwLock::new(PeerDB::new(trusted_peers, disable_peer_scoring, log)),
             gossipsub_subscriptions: RwLock::new(HashSet::new()),
             sync_state: RwLock::new(SyncState::Stalled),
-            backfill_state: RwLock::new(BackFillState::NotRequired),
+            backfill_state: RwLock::new(BackFillState::Paused),
```
michaelsproul, line 75: I don't think this state should ever be read, so it doesn't matter what it is

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -2116,39 +2117,27 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
         let kv_store_op = self.compare_and_set_anchor_info(prev_value, new_value)?;
         self.hot_db.do_atomically(vec![kv_store_op])
     }
 
-    /// Load the anchor info from disk, but do not set `self.anchor_info`.
-    fn load_anchor_info(&self) -> Result<Option<AnchorInfo>, Error> {
-        self.hot_db.get(&ANCHOR_INFO_KEY)
+    /// Load the anchor info from disk.
+    fn load_anchor_info(hot_db: &Hot) -> Result<AnchorInfo, Error> {
+        // FIXME(sproul): add migration
```
michaelsproul: I still need to add this to the v22 DB migration:

- Set the anchor to `ANCHOR_FOR_ARCHIVE_NODE` if it is currently absent (`None`).

beacon_node/store/src/reconstruct.rs
```diff
@@ -21,10 +22,12 @@ where
         self: &Arc<Self>,
         num_blocks: Option<usize>,
     ) -> Result<(), Error> {
-        let Some(mut anchor) = self.get_anchor_info() else {
-            // Nothing to do, history is complete.
+        let mut anchor = self.get_anchor_info();
+
+        // Nothing to do, history is complete.
+        if anchor.state_upper_limit == anchor.state_lower_limit {
```
michaelsproul:
```suggestion
        if anchor.all_historic_states_stored() {
```

beacon_node/beacon_chain/src/block_verification.rs
pawanjay176: What if checkpoint sync to a non-finalized checkpoint though? In that case, the checkpoint slot could be ahead of finalized slot.

beacon_node/beacon_chain/src/historical_blocks.rs
pawanjay176: Don't see why we need the additional condition. Sync should be able to handle backfill already being complete

beacon_node/lighthouse_network/src/types/globals.rs
pawanjay176: Yeah don't see any issues with this

pawanjay176 (approved):
LGTM. Just a question.

beacon_node/beacon_chain/src/block_verification.rs
dapplion: In that case, we force-initialize the fork-choice as if that checkpoint was finalized.

PR #6398, opened 2024-09-16
Generalize sync ActiveRequests

dapplion:
Part of

To address PeerDAS sync issues we need to make individual by_range requests within a batch retriable. We should adopt the same pattern for lookup sync where each request (block/blobs/columns) is tracked individually within a "meta" request that group them all and handles retry logic.

The first step is to add `ActiveBlocksByRangeRequest`, `ActiveBlobsByRangeRequest` and `ActiveDataColumnsByRangeRequest`. I'm in the progress of doing that in this branch https://github.com/sigp/lighthouse/compare/unstable...dapplion:lighthouse:sync-active-requests?expand=1 but I noted a lot of sensitive code repetition.

This PR is pre-step to add `ActiveRequest` to our existing `ActiveRequest*ByRoot` (the ones for by_root). Next PR will add the `ActiveRequest*ByRange` goodies :)


beacon_node/lighthouse_network/src/service/api_types.rs
```diff
@@ -42,11 +37,19 @@ pub enum SyncRequestId {
     /// Request searching for a set of blobs given a hash.
     SingleBlob { id: SingleLookupReqId },
     /// Request searching for a set of data columns given a hash and list of column indices.
-    DataColumnsByRoot(DataColumnsByRootRequestId, DataColumnsByRootRequester),
+    DataColumnsByRoot(DataColumnsByRootRequestId),
```
dapplion, line 33: Having the requester inside the ID just simplifies downstream code and makes it symmetrical with the other variants

beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -2643,21 +2640,6 @@ mod deneb_only {
...
-        };
-        tester
-            .block_response_triggering_process()
-            .invalidate_blobs_too_many()
-            .blobs_response()
-            .expect_penalty("TooManyResponses")
-            // Network context returns "download success" because the request has enough blobs + it
-            // downscores the peer for returning too many.
-            .expect_no_block_request();
-    }
```
dapplion, line 2830: This test needs a scenario where sync receives more blob chunks than originally requested. This can never happen in practice so I choose to not handle this scenario to reduce complexity

https://github.com/sigp/lighthouse/blob/9446569a8adff0e44e305a8e5128d825b45f4f6d/beacon_node/network/src/sync/network_context/requests.rs#L114-L119

beacon_node/lighthouse_network/src/service/api_types.rs
```diff
@@ -42,11 +37,19 @@ pub enum SyncRequestId {
...
     /// Range request that is composed by both a block range request and a blob range request.
     RangeBlockAndBlobs { id: Id },
 }
 
+/// Request ID for data_columns_by_root requests. Block lookup do not issue this requests directly.
+/// Wrapping this particular req_id, ensures not mixing this requests with a custody req_id.
```
realbigsean:
```suggestion
/// Request ID for data_columns_by_root requests. Block lookups do not issue this request directly.
/// Wrapping this particular req_id, ensures not mixing this request with a custody req_id.
```

beacon_node/lighthouse_network/src/service/api_types.rs
```diff
@@ -247,6 +250,6 @@ impl slog::Value for RequestId {
 
 impl std::fmt::Display for DataColumnsByRootRequestId {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.0)
+        write!(f, "{}", self.id)
```
realbigsean: should this also display the requester?

beacon_node/network/src/sync/network_context.rs
```diff
@@ -223,9 +224,17 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
             network_send,
             execution_engine_state: EngineState::Online, // always assume `Online` at the start
             request_id: 1,
-            blocks_by_root_requests: <_>::default(),
-            blobs_by_root_requests: <_>::default(),
-            data_columns_by_root_requests: <_>::default(),
+            // true = enfore max_requests as returned for blocks_by_root. We always request a single
```
realbigsean:
```suggestion
            // true = enforce max_requests as returned for blocks_by_root. We always request a single
```

beacon_node/network/src/sync/network_context.rs
```diff
@@ -223,9 +224,17 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+            // block and the peer must have it.
+            blocks_by_root_requests: ActiveRequests::new(true, "blocks_by_root"),
+            // true = enfore max_requests are returned for blobs_by_root. We only issue requests for
```
realbigsean:
```suggestion
            // true = enforce max_requests are returned for blobs_by_root. We only issue requests for
```

beacon_node/network/src/sync/network_context.rs
realbigsean: I'm confused - when do we not want to enforce max requests returned? it looks like this is always set to true. Is there a follow up PR where this could be false?

beacon_node/network/src/sync/network_context.rs
```diff
@@ -917,142 +915,77 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
-        let resp = match rpc_event {
-            RpcEvent::Response(block, seen_timestamp) => {
-                match request.get_mut().add_response(block) {
-                    Ok(block) => Ok((block, seen_timestamp)),
-                    Err(e) => {
-                        // The request must be dropped after receiving an error.
-                        request.remove();
-                        Err(e.into())
-                    }
+        let r = self.blocks_by_root_requests.on_response(id, rpc_event);
```
realbigsean: i think you could simplify this a bit
```
    let result = response.and_then(|res| {
        match res {
            Ok((mut blocks, seen_timestamp)) => {
                blocks.pop().map(|block| Ok((block, seen_timestamp)))
                    .or(Some(Err(LookupVerifyError::NotEnoughResponsesReturned { actual: 0 }.into())))
            },
            Err(e) => Some(Err(e)),
        }
    });
```

beacon_node/network/src/sync/network_context.rs
```diff
@@ -917,142 +915,77 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
-                        .map_err(|e| (e.into(), request.resolve())),
-                    Ok(None) => return None,
-                    Err(e) => Err((e.into(), request.resolve())),
-                }
-            }
-            RpcEvent::StreamTermination => match request.remove().terminate() {
-                Ok(_) => return None,
-                // (err, false = not resolved) because terminate returns Ok() if resolved
-                Err(e) => Err((e.into(), false)),
+        let r = self.blobs_by_root_requests.on_response(id, rpc_event);
```
realbigsean: same as above

realbigsean (approved):
Just had some nits, but looks good, overall like the direction 👌

beacon_node/network/src/sync/network_context.rs
dapplion: Yeah, sorry should have added a note. For the all the `*_by_range` requests

dapplion:
@realbigsean sir I fix the test sir

realbigsean:
looks like there's another test broken, sorry @dapplion !

realbigsean:
@mergify queue

michaelsproul:
@mergify dequeue

michaelsproul:
@mergify queue

realbigsean:
@mergify queue

realbigsean:
looks like CI here is hung on [b9cd5ad](https://github.com/sigp/lighthouse/pull/6398/commits/b9cd5ad88272503a8043c3fa89baac1928479eab) ?
Screenshot 2024-10-17 at 10 08 27 AM

realbigsean:
@mergify cancel

realbigsean:
@mergify queue

PR #6400, opened 2024-09-16
Improve rpc logic

jxs:
## Issue Addressed

simplifications at the RPC `Behaviour`:
- do not expose `HandlerEvent` outside of `Behaviour`, it's not required and simplifies logic as we avoid matching on `HandlerEvent::Close`
- reply to Ping's at the `Handler` level to encapsulate Rpc behaviour and make it more easily dependable,

per commit review is suggested

### DIFF
--- a/beacon_node/lighthouse_network/src/rpc/handler.rs
+++ b/beacon_node/lighthouse_network/src/rpc/handler.rs
@@ -4,7 +4,7 @@
 use super::methods::{GoodbyeReason, RPCCodedResponse, RPCResponseErrorCode};
 use super::outbound::OutboundRequestContainer;
 use super::protocol::{InboundOutput, InboundRequest, Protocol, RPCError, RPCProtocol};
-use super::{RPCReceived, RPCSend, ReqId};
+use super::{RPCReceived, RPCResponse, RPCSend, ReqId};
 use crate::rpc::outbound::{OutboundFramed, OutboundRequest};
 use crate::rpc::protocol::InboundFramed;
 use fnv::FnvHashMap;
@@ -14,7 +14,8 @@ use libp2p::swarm::handler::{
     ConnectionEvent, ConnectionHandler, ConnectionHandlerEvent, DialUpgradeError,
     FullyNegotiatedInbound, FullyNegotiatedOutbound, StreamUpgradeError, SubstreamProtocol,
 };
-use libp2p::swarm::Stream;
+use libp2p::swarm::{ConnectionId, Stream};
+use libp2p::PeerId;
 use slog::{crit, debug, trace};
 use smallvec::SmallVec;
 use std::{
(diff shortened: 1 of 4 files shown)


AgeManning (approved):
Yeah nice. Much cleaner!

AgeManning:
@mergifyio queue

beacon_node/lighthouse_network/src/rpc/handler.rs
```diff
@@ -892,6 +903,15 @@ where
             self.shutdown(None);
         }
 
+        // If we received a Ping, we queue a Pong response.
+        if let InboundRequest::Ping(ping) = req {
+            trace!(self.log, "Received Ping, queueing Pong";"connection_id" => %self.id, "peer_id" => %self.peer_id);
+            self.send_response(
+                self.current_inbound_substream_id,
+                RPCCodedResponse::Success(RPCResponse::Pong(ping)),
```
ackintosh, line 911: Here, the request (the `ping` variable) is set to the response, but we should set our sequence number as `Network::pong()` was doing before this change, right?

https://github.com/sigp/lighthouse/blob/672dcbd868198ec6430c90f9d41373cde740dc3d/beacon_node/lighthouse_network/src/service/mod.rs#L1131-L1139

beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -1406,10 +1400,7 @@ impl<E: EthSpec> Network<E> {
         let peer_id = event.peer_id;
 
         // Do not permit Inbound events from peers that are being disconnected, or RPC requests.
-        if !self.peer_manager().is_connected(&peer_id)
-            && (matches!(event.event, HandlerEvent::Err(HandlerErr::Inbound { .. }))
```
pawanjay176, line 1410: Why do we not need these conditions anymore?

jxs: cause `HandlerEvent` is no longer exported in `RpcMessage::event` it's now `Result<RPCReceived<Id, E>, HandlerErr<Id>>`

beacon_node/lighthouse_network/src/rpc/handler.rs
jxs: Thanks Akihito, I definitely overlooked this, will submit a follow up PR bringing the sequence number to the `Behaviour` and probably we can reply the pings there. CC @AgeManning

beacon_node/lighthouse_network/src/service/mod.rs
pawanjay176: But the reason we had the extra condition iirc is that we wanted to generate a RpcFailed event for `HandlerErr::Outbound` to send to sync and allow sync to remove the peer from all mappings

beacon_node/lighthouse_network/src/rpc/handler.rs
jxs: submitted https://github.com/sigp/lighthouse/pull/6408

beacon_node/lighthouse_network/src/service/mod.rs
AgeManning: Yeah, nice catch

jxs: ah right! Thanks Pawan, that makes sense yeah, submmited https://github.com/sigp/lighthouse/pull/6408/commits/5fa8c80fa2060ab4162c7cd662dc9e888102efd7

PR #6385, opened 2024-09-11
Upgrade to discv5 0.7.0

AgeManning:
This updates to the latest version of discv5. 

The version upgrade brings a number of dependency upgrades, of important note is the use of alloy-rlp in the ENR library. 

We are phasing out the old rlp library.

PR #6402, opened 2024-09-17
Correct ENR decoding on extension trait

AgeManning:
In #6385 we upgraded to a new version of ENR. This version uses alloy-rlp. We have now explicit typing in decoding values to and from an ENR. 

The previous update assumed `Vec<u8>` would be decoded as bytes, but the alloy-rlp implements deocde for `Vec<T>` as an RLP list. So it was trying to decode a list of u8's rather than raw bytes. As the eth2 and attestation fields are not RLP lists this decoding failed and we marked the ENRs as invalid. 

This should correct this issue by force the type to be `Bytes` allowing us to decode the raw bytes in the ENR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1061,6 +1061,7 @@ name = "boot_node"
 version = "5.3.0"
 dependencies = [
  "beacon_node",
+ "bytes",
  "clap",
  "clap_utils",
  "eth2_network_config",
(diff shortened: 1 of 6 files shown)


pawanjay176 (approved):

pawanjay176:
Confirmed that discovery searches are working now. Added a commit to fix other occurrences of `EnrBuilder::add_value`

dapplion (approved):
Nice find!

jimmygchen:
@mergify queue

michaelsproul:
@mergify requeue

PR #6403, opened 2024-09-17
Refactor data column reconstruction and avoid blocking processing

jimmygchen:
## Issue Addressed

Continuation of #5990.

I've taken the changes from #5990 with some cleanups. This should simplify the code a bit and reduce supernode bandwidth and performance (reconstruction no longer blocks DA processing for the block).

## Proposed Changes

- Move reconstruction logic out of `overflow_lru_cache`, this simplifies the code and avoids having to pass `DataColumnsToPublish` around.
- Changes to data column reconstruction:
  - Attempt reconstruction without holding availability cache lock, so we can process other gossip / rpc data columns simultaneously
  - Check availability cache again before publishing reconstructed columns to avoid publishing excess duplicates


beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3166,15 +3192,13 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
 
     /// Remove any block components from the *processing cache* if we no longer require them. If the
     /// block was imported full or erred, we no longer require them.
-    fn remove_notified_custody_columns<P>(
+    fn remove_notified_custody_columns(
```
dapplion: If the return value does not have P, we can probably remove the function in favor of `remove_notified`

beacon_node/network/src/network_beacon_processor/mod.rs
```diff
@@ -767,6 +769,73 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+            Ok(Some((availability_processing_status, data_columns_to_publish))) => {
+                match &availability_processing_status {
+                    AvailabilityProcessingStatus::Imported(hash) => {
+                        debug!(
+                            self.log,
+                            "Block components available via reconstruction";
+                            "result" => "imported block and custody columns",
+                            "block_hash" => %hash,
+                        );
+                        self.chain.recompute_head_at_current_slot().await;
```
dapplion, line 887: Should we compute the head before publishing the columns? I don't have metrics but scheduling the columns for publishing should not block this thread for too long, and it's timing sensitive for the network to do so as soon as possible.

jimmygchen: Good catch! Recompute head should definitely happen after publish - publish is non blocking anyway.

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Nice, didn't realize they're identical!

beacon_node/network/src/network_beacon_processor/mod.rs
jimmygchen: Fixed in https://github.com/sigp/lighthouse/pull/6403/commits/2e641908a83732dcf7279aaa35fd9fd3806e0d17

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Fixed in https://github.com/sigp/lighthouse/pull/6403/commits/2e641908a83732dcf7279aaa35fd9fd3806e0d17

michaelsproul (approved):
LGTM.

Just needs some conflicts fixed

dapplion (approved):
good to go modulo conflicts

jimmygchen:
Thanks @michaelsproul and @dapplion ! I've resolved conflicts.
Also noticed that I was doing reconstruction in an async method, so I've moved that into a blocking task here: https://github.com/sigp/lighthouse/pull/6403/commits/d3c84e8573eb66b4682ae0a129c595daa2ceaa08

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -511,6 +512,114 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
...
+            // Block may have been imported as it does not exist in availability cache.
+            return Ok(None);
+        };
+
+        if !self.should_reconstruct(&pending_components) {
+            return Ok(None);
+        }
+
+        self.availability_cache
+            .set_reconstruction_started(block_root);
```
dapplion: There's a potential race condition here. We clone an old view of the pending components, check if `reconstruction_started`, then acquire the lock again in `set_reconstruction_started` and set `reconstruction_started` to true. We could have N parallel reconstruction attempts happening:

- thread 1: peek_pending_components + clone + drop lock
- thread 2: peek_pending_components + clone + drop lock
- thread 1:  `should_reconstruct` returns true
- thread 2:  `should_reconstruct` returns true

So either hold the lock until you set_reconstruction_started or restructure the check into being atomic.

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -511,6 +512,114 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
...
+        );
+
+        debug!(self.log, "Reconstructed columns";
+            "count" => data_columns_to_publish.len(),
+            "block_root" => ?block_root,
+            "slot" => slot,
+        );
+
+        self.availability_cache
+            .put_kzg_verified_data_columns(
```
dapplion, line 598: We have the same issue here, we take a snapshot of an old view of the shared state in `imported_custody_column_indexes` and then later acquire the write lock again in `put_kzg_verified_data_columns`. However, if the bug above is fixed and reconstruction strictly happens once we should be fine? Also worst case we publish the same thing twice. However, now that the entry into the da_checker is not dropped until after import this can still be problematic and lead to double imports?

beacon_node/network/src/network_beacon_processor/mod.rs
```diff
@@ -847,6 +849,73 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+
+                Some(availability_processing_status)
+            }
+            Ok(None) => {
+                trace!(
+                    self.log,
+                    "Reconstruction not required for block";
+                    "block_hash" => %block_root,
+                );
+                None
```
dapplion, line 908: Consider we run this feature in production and reconstruction never happens. The next question would be why. The only evidence will be the absence of logs and all metrics being 0. We may want to log and track WHY we are not attempting reconstruction. How often is `attempt_data_column_reconstruction` called? We could add a metric  here with the reason why reconstruction is not done: i.e. not enough columns, not supernode etc. Could be noisy but I worry we are missing information if reconstruction just doesn't happen.

dapplion: Also we could have a debug log when importing a data column into the da_checker that prints the current count of imported valid data columns for that block root, and the expected custody requirement. That way we can track if we ever reached the reconstruction threshold or not.

beacon_node/beacon_chain/src/data_availability_checker.rs
jimmygchen: Good catch!

beacon_node/beacon_chain/src/data_availability_checker.rs
jimmygchen: > However, if the bug above is fixed and reconstruction strictly happens once we should be fine

Yep I think so. The double import issue has been raised here, I think it might be best to address it separately:
https://github.com/sigp/lighthouse/issues/6439

beacon_node/beacon_chain/src/data_availability_checker.rs
jimmygchen: I think this is fixed in https://github.com/sigp/lighthouse/pull/6403/commits/b0c337946de2a33c08eab8515d9612a27ca9c452.

I've also added error handling to reconstruction failure, so we can recover from potentially bad columns that somehow made it through to DA checker.

beacon_node/network/src/network_beacon_processor/mod.rs
jimmygchen: I've added metric `KZG_DATA_COLUMN_RECONSTRUCTION_INCOMPLETE_TOTAL` with a `reason` label for this. I think logging would be very noisy, especially after the getBlobsV1 optimisation.

jimmygchen: Added reconstruction metric and da checker debug log in https://github.com/sigp/lighthouse/pull/6403/commits/0e6eaa227890194983cd523d5e6be498907dfdad

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -533,65 +552,67 @@ impl<T: BeaconChainTypes> DataAvailabilityCheckerInner<T> {
...
+        ReconstructColumnsDecision::Yes(pending_components.clone())
+    }
+
+    /// This could mean some invalid data columns made it through to the `DataAvailabilityChecker`.
+    /// In this case, we remove all data columns in `PendingComponents`, reset reconstruction
+    /// status so that we can attempt to retrieve columns from peers again.
+    pub fn handle_reconstruction_failure(&self, block_root: &Hash256) {
+        if let Some(pending_components_mut) = self.critical.write().get_mut(block_root) {
+            pending_components_mut.verified_data_columns = vec![];
+            pending_components_mut.reconstruction_started = false;
```
dapplion, line 615: Nice! Will sleep better with this fallback :)

dapplion (approved):
LGTM! Thanks for the iterations looks solid now

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -511,6 +525,92 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
...
+        &self,
+        block_root: &Hash256,
+    ) -> Result<DataColumnReconstructionResult<T::EthSpec>, AvailabilityCheckError> {
+        let pending_components = match self
+            .availability_cache
+            .check_and_set_reconstruction_started(block_root)
+        {
+            ReconstructColumnsDecision::Yes(pending_components) => pending_components,
+            ReconstructColumnsDecision::No(reason) => {
+                return Ok(DataColumnReconstructionResult::NotRequired(reason));
```
ethDreamer: I wanted to point out that the Error variant here is called `NotRequired` but `ReconstructionDecision` includes "not enough columns" which (unlike every other variant) doesn't actually mean it's not required, only that it can't be done yet. I thought you might instead change things to:
```
#[derive(Debug)]
pub enum RejectionCriteria {
    BlockAlreadyImported,
    AlreadyStarted,
    NotRequiredForFullNode,
    AllColumnsReceived,
    NotEnoughColumns,
}

pub enum DataColumnReconstructionResult<E: EthSpec> {
    Success(AvailabilityAndReconstructedColumns<E>),
    NotRequired(RejectionCriteria),
    Pending(RejectionCriteria),
}

pub(crate) enum ReconstructColumnsDecision<E: EthSpec> {
    Yes(PendingComponents<E>),
    No(RejectionCriteria),
}

impl<E: EthSpec> From<RejectionCriteria> for DataColumnReconstructionResult<E> {
    fn from(criteria: RejectionCriteria) -> Self {
        match criteria {
            RejectionCriteria::NotEnoughColumns => DataColumnReconstructionResult::Pending(criteria),
            _ => DataColumnReconstructionResult::NotRequired(criteria),
        }
    }
}
```

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -511,6 +525,92 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
...
+            );
+            self.availability_cache
+                .handle_reconstruction_failure(block_root);
+            metrics::inc_counter(&KZG_DATA_COLUMN_RECONSTRUCTION_FAILURES);
+            AvailabilityCheckError::ReconstructColumnsError(e)
+        })?;
+
+        // Check indices from cache again to make sure we don't publish components we've already received.
+        let Some(existing_column_indices) = self.cached_data_column_indexes(block_root) else {
+            return Ok(DataColumnReconstructionResult::RecoveredColumnsNotImported(
```
ethDreamer, line 566: Just checking my understanding here.. but shouldn't this never happen due to the `reconstruction_started` gate?

ethDreamer (commented):
I checked out all the DA checker / overflow_lru_cache changes. It looks good to me although I can't really comment on race conditions with block import until I have a conversation with you guys about it.

pawanjay176 (approved):

beacon_node/beacon_chain/src/data_availability_checker.rs
jimmygchen: Yeah good question - this scenario can happen if we receive the columns via gossip, and we no longer need to import/publish them.

beacon_node/beacon_chain/src/data_availability_checker.rs
jimmygchen: Ah yeah I thought about having an enum for all the reasons, and I decided to go with string instead because we don't really need to handle each variant differently, the main usage of the `reason` is for metric label, and it requires a string, hence I went with it for simplicity and efficiency:

https://github.com/sigp/lighthouse/blob/0e6eaa227890194983cd523d5e6be498907dfdad/beacon_node/beacon_chain/src/beacon_chain.rs#L3199-L3202

> the Error variant here is called NotRequired but ReconstructionDecision includes "not enough columns" which (unlike every other variant) doesn't actually mean it's not required, only that it can't be done yet.

Yes that's right, if this is confusing I could rename `DataColumnReconstructionResult::NotRequired` to something else?

jimmygchen: I've renamed `NotRequired` to `NotStarted`.

michaelsproul (approved):

michaelsproul:
@mergify queue

michaelsproul:
@mergify requeue

michaelsproul:
@mergify refresh

Issue #6405, opened 2024-09-17
Test failure on Rust v1.81


eserilev:
## Description

The test `prune_excess_peers` fails when running against Rust 1.81

```
 panicked at /Users/eitan/.cargo/registry/src/index.crates.io-6f17d22bba15001f/quickcheck-1.0.3/src/tester.rs:165:28:

Error: "user-provided comparison function does not correctly implement a total order"

```

I believe its related to the new sort impls in 1.81 that panic if they're unable to meaningfully sort a result (instead of just returning randomly sorted data) https://blog.rust-lang.org/2024/09/05/Rust-1.81.0.html#new-sort-implementations

cc @AgeManning @jxs

michaelsproul:
I think we need to prevent `NaN` values from creeping in as scores here:

https://github.com/sigp/lighthouse/blob/2e0eb6d1b8705bbda2ba56eb195d9cc7c6575e95/beacon_node/lighthouse_network/src/peer_manager/mod.rs#L2266-L2268

Our `PartialOrd` and `Ord` implementations for `Score` are broken for `NaN`:

https://github.com/sigp/lighthouse/blob/2e0eb6d1b8705bbda2ba56eb195d9cc7c6575e95/beacon_node/lighthouse_network/src/peer_manager/peerdb/score.rs#L340-L342

It returns `Some(Equal)` for incomparable values (any `NaN`s), which violates the first rule of [`PartialOrd`](https://doc.rust-lang.org/stable/std/cmp/trait.PartialOrd.html):

> 1. `a == b` if and only if `partial_cmp(a, b) == Some(Equal)`

We have:

```
partial_cmp(NaN, x) == Some(Equal)
```

which would require `NaN == x` (false).

eserilev:
Closed by #6407

PR #6407, opened 2024-09-17
Fix test failure on Rust v1.81

eserilev:
## Issue Addressed

#6405

## Proposed Changes

prevent `NaN` f64 by looping through `f64::arbitrary` calls until we get a valid number

### DIFF
--- a/beacon_node/lighthouse_network/src/peer_manager/mod.rs
+++ b/beacon_node/lighthouse_network/src/peer_manager/mod.rs
@@ -2330,7 +2330,6 @@ mod tests {
         use libp2p::PeerId;
         use quickcheck::{Arbitrary, Gen, TestResult};
         use quickcheck_macros::quickcheck;
-        use rand::Rng;
         use tokio::runtime::Runtime;
         use types::Unsigned;
         use types::{EthSpec, MainnetEthSpec as E};
@@ -2346,9 +2345,14 @@ mod tests {
             gossipsub_score: f64,
         }
 
-        fn rand_f64() -> f64 {
-            let mut rng = rand::thread_rng();
-            rng.gen()
+        // generate an arbitrary f64 while preventing NaN values
+        fn arbitrary_f64(g: &mut Gen) -> f64 {
+            loop {
+                let val = f64::arbitrary(g);
+                if !val.is_nan() {
+                    return val;
+                }
+            }
         }
 
         impl Arbitrary for PeerCondition {
@@ -2376,9 +2380,9 @@ mod tests {
                     outgoing: bool::arbitrary(g),
                     attestation_net_bitfield,
                     sync_committee_net_bitfield,
-                    score: rand_f64(),
+                    score: arbitrary_f64(g),
                     trusted: bool::arbitrary(g),
-                    gossipsub_score: rand_f64(),
+                    gossipsub_score: arbitrary_f64(g),
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


beacon_node/lighthouse_network/src/peer_manager/mod.rs
```diff
@@ -2370,9 +2376,9 @@ mod tests {
                     outgoing: bool::arbitrary(g),
                     attestation_net_bitfield,
                     sync_committee_net_bitfield,
-                    score: f64::arbitrary(g),
+                    score: rand_f64(),
```
michaelsproul: Quickcheck (and proptest) can do a clever thing called _shrinking_ where they simplify (shrink) failing test cases to be as small as possible. If we mix in `rand` we are taking control from quickcheck which will make the test non-deterministic based on quickcheck's seed _and_ will disable shrinking.

I think it might be better to use `f64::arbitrary` in a loop until it returns a value other than `NaN`. An even cleaner option might be generating floats in a range, which would also avoid other weird values like infinity. From a quick skim of the docs I'm not sure if quickcheck's arbitrary has a function for this though

michaelsproul: We could do the `NaN` loop for now and then come back to this if/when we rewrite our quickcheck tests using proptest. Proptest is a bit better maintained

eserilev: thanks for the additional context!

eserilev: aa9014a

jxs (commented):
can instead we do something like 
```patch
diff --git a/beacon_node/lighthouse_network/src/peer_manager/peerdb/score.rs b/beacon_node/lighthouse_network/src/peer_manager/peerdb/score.rs
index c8425fc10..b36bccfed 100644
--- a/beacon_node/lighthouse_network/src/peer_manager/peerdb/score.rs
+++ b/beacon_node/lighthouse_network/src/peer_manager/peerdb/score.rs
@@ -335,9 +335,14 @@ impl PartialOrd for Score {
 
 impl Ord for Score {
     fn cmp(&self, other: &Score) -> std::cmp::Ordering {
-        self.score()
-            .partial_cmp(&other.score())
-            .unwrap_or(std::cmp::Ordering::Equal)
+        match (self.score(), other.score()) {
+            (s, o) if o.is_nan() || s.is_nan() => std::cmp::Ordering::Equal,
+            (_s, o) if o.is_nan() => std::cmp::Ordering::Greater,
+            (s, _o) if s.is_nan() => std::cmp::Ordering::Less,
+            (s, o) => s
+                .partial_cmp(&o)
+                .expect("all possible panic cases are covered"),
+        }
     }
 }
 ```
and address the broken `Ord` impl?

eserilev:
the custom ordering implementation you suggested doesn't seem to resolve the issue. Not 100% sure why, I'll have some time this evening to dig into this a bit further

michaelsproul:
@eserilev @jxs That `Ord` impl doesn't work because `NaN == NaN` is `false`.

There's also a typo in the first `match` arm. I'm guessing you meant `o.is_nan() && s.is_nan()`. Even with that change, it won't work because we'll have `partial_cmp(s, o) = Some(Equal)` but `s != o` because they are both `NaN`.

I'm going to merge this PR as-is because CI is now blocked on it since the runners were updated to v1.81.

michaelsproul (approved):

michaelsproul:
@mergify queue

michaelsproul:
@mergify refresh

michaelsproul:
@mergify queue

PR #6408, opened 2024-09-18
fix Rpc Ping sequence number

jxs:
## Issue Addressed

Send Rpc Pong with the correct sequence number, address issue reported by @ackintosh [here](https://github.com/sigp/lighthouse/pull/6400/files#r1764095680):

### DIFF
--- a/beacon_node/lighthouse_network/src/rpc/mod.rs
+++ b/beacon_node/lighthouse_network/src/rpc/mod.rs
@@ -14,7 +14,7 @@ use libp2p::swarm::{
 use libp2p::swarm::{ConnectionClosed, FromSwarm, SubstreamProtocol, THandlerInEvent};
 use libp2p::PeerId;
 use rate_limiter::{RPCRateLimiter as RateLimiter, RateLimitedErr};
-use slog::{crit, debug, o};
+use slog::{crit, debug, o, trace};
 use std::marker::PhantomData;
 use std::sync::Arc;
 use std::task::{Context, Poll};
@@ -132,6 +132,8 @@ pub struct RPC<Id: ReqId, E: EthSpec> {
     log: slog::Logger,
     /// Networking constant values
     network_params: NetworkParams,
+    /// A sequential counter indicating when data gets modified.
+    seq_number: u64,
 }
 
 impl<Id: ReqId, E: EthSpec> RPC<Id, E> {
@@ -142,6 +144,7 @@ impl<Id: ReqId, E: EthSpec> RPC<Id, E> {
         outbound_rate_limiter_config: Option<OutboundRateLimiterConfig>,
         log: slog::Logger,
         network_params: NetworkParams,
+        seq_number: u64,
     ) -> Self {
         let log = log.new(o!("service" => "libp2p_rpc"));
 
@@ -163,6 +166,7 @@ impl<Id: ReqId, E: EthSpec> RPC<Id, E> {
             enable_light_client_server,
             log,
             network_params,
+            seq_number,
         }
     }
 
@@ -214,6 +218,10 @@ impl<Id: ReqId, E: EthSpec> RPC<Id, E> {
             event: RPCSend::Shutdown(id, reason),
         });
     }
+
+    pub fn update_seq_number(&mut self, seq_number: u64) {
+        self.seq_number = seq_number
+    }
 }
 
 impl<Id, E> NetworkBehaviour for RPC<Id, E>
@@ -245,8 +253,6 @@ where
             .log
             .new(slog::o!("peer_id" => peer_id.to_string(), "connection_id" => connection_id.to_string()));
         let handler = RPCHandler::new(
-            connection_id,
-            peer_id,
             protocol,
             self.fork_context.clone(),
             &log,
@@ -280,8 +286,6 @@ where
             .new(slog::o!("peer_id" => peer_id.to_string(), "connection_id" => connection_id.to_string()));
 
         let handler = RPCHandler::new(
-            connection_id,
-            peer_id,
             protocol,
             self.fork_context.clone(),
             &log,
@@ -359,14 +363,6 @@ where
                 if let Some(limiter) = self.limiter.as_mut() {
                     // check if the request is conformant to the quota
                     match limiter.allows(&peer_id, &req) {
-                        Ok(()) => {
-                            // send the event to the user
-                            self.events.push(ToSwarm::GenerateEvent(RPCMessage {
-                                peer_id,
-                                conn_id,
-                                message: Ok(RPCReceived::Request(id, req)),
-                            }))
-                        }
                         Err(RateLimitedErr::TooLarge) => {
                             // we set the batch sizes, so this is a coding/config err for most protocols
                             let protocol = req.versioned_protocol().protocol();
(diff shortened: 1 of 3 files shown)


beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -159,37 +159,36 @@ impl<E: EthSpec> Network<E> {
             .collect();
 
         // set up a collection of variables accessible outside of the network crate
-        let network_globals = {
```
jxs, line 162: simplification, we can do everything in the main scope of the function

beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -1103,22 +1103,21 @@ impl<E: EthSpec> Network<E> {
             .sync_committee_bitfield::<E>()
             .expect("Local discovery must have sync committee bitfield");
 
-        {
```
jxs, line 1107: no need to open another scope, we can just `drop` the write lock ahead and cache `local_metadata` to save another `read` call

beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -408,16 +405,29 @@ where
...
-                        message: Ok(RPCReceived::Request(id, req)),
-                    }))
+                        (conn_id, id),
+                        RPCCodedResponse::Success(RPCResponse::Pong(Ping {
+                            data: self.seq_number,
+                        })),
+                    );
                 }
+
+                self.events.push(ToSwarm::GenerateEvent(RPCMessage {
```
AgeManning, line 435: If we are handling the Ping response here, do we still need to send the PING request to the swarm?

AgeManning (commented):
This looks good to me. My bad for missing the bugs in the previous PR

beacon_node/lighthouse_network/src/rpc/mod.rs
jxs: yeah, I'd say in general yes, to allow the other behaviours or even the swarm to react to `Ping`s

beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -408,16 +405,29 @@ where
...
                         }
+                        // No rate limiting, continue.
+                        Ok(_) => {}
                     }
-                } else {
-                    // No rate limiting, send the event to the user
-                    self.events.push(ToSwarm::GenerateEvent(RPCMessage {
+                }
+                // If we received a Ping, we queue a Pong response.
+                if let InboundRequest::Ping(_) = req {
```
pawanjay176, line 424: I don't quite get the need to handle Ping differently in the rpc behaviour. What was the issue with the way we handled it before from lighthouse_network service?

jxs: wdym differently? The issue before is that Replying to `Ping`s with `Pong`s is part of the [spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#ping) so the Rpc behaviour should be able to reply to `Ping`s automatically, It doesn't make sense to handle Pong responses at the `Swarm` level instead, both in terms of [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns) and latency.  This is also how the [`Ping`](https://docs.rs/libp2p/latest/libp2p/ping/index.html) behaviour operates.

ackintosh (commented):
I think [`Network::ping()`](https://github.com/sigp/lighthouse/blob/5fa8c80fa2060ab4162c7cd662dc9e888102efd7/beacon_node/lighthouse_network/src/service/mod.rs#L1123-L1132) should simply call the RPC instead of making a read access to `network_globals.local_metadata` as shown below, since the RPC can create a Ping request with its seq_number:

```rust
    fn ping(&mut self, peer_id: PeerId) {
        trace!(self.log, "Sending Ping"; "peer_id" => %peer_id);
        let id = RequestId::Internal;
        self.eth2_rpc_mut()
            .ping(peer_id, id); // `RPC::ping()` also need to be implemented.
    }
```

This would be good for reducing read locks and ensuring coding consistency.

jxs (commented):
> I think [`Network::ping()`](https://github.com/sigp/lighthouse/blob/5fa8c80fa2060ab4162c7cd662dc9e888102efd7/beacon_node/lighthouse_network/src/service/mod.rs#L1123-L1132) should simply call the RPC instead of making a read access to `network_globals.local_metadata` as shown below, since the RPC can create a Ping request with its seq_number:
> 
> ```rust
>     fn ping(&mut self, peer_id: PeerId) {
>         trace!(self.log, "Sending Ping"; "peer_id" => %peer_id);
>         let id = RequestId::Internal;
>         self.eth2_rpc_mut()
>             .ping(peer_id, id); // `RPC::ping()` also need to be implemented.
>     }
> ```
> 
> This would be good for reducing read locks and ensuring coding consistency.

thanks @ackintosh makes total sense, updated!

beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -1401,7 +1393,10 @@ impl<E: EthSpec> Network<E> {
         let peer_id = event.peer_id;
 
         // Do not permit Inbound events from peers that are being disconnected, or RPC requests.
-        if !self.peer_manager().is_connected(&peer_id) {
+        if !self.peer_manager().is_connected(&peer_id)
```
ackintosh, line 1397: How about adding a comment here based on Pawan’s feedback, to clarify why we need the extra conditions?

https://github.com/sigp/lighthouse/pull/6400#discussion_r1764200888
> the reason we had the extra condition iirc is that we wanted to generate a RpcFailed event for HandlerErr::Outbound to send to sync and allow sync to remove the peer from all mappings

ackintosh (approved):
Thanks for the update! I've left a small comment, but this looks good to me.

beacon_node/lighthouse_network/src/rpc/mod.rs
pawanjay176: Yeah we need to let the peer manager know about the ping so that it can send metadata requests if the seq number is updated.

beacon_node/lighthouse_network/src/rpc/mod.rs
pawanjay176: My point is that we are separating a specific concern (replying to the ping message). But we are still handling other requirements of the spec at the swarm level (sending a metadata request if the seq number has changed, setting a timer for the next ping request). 

But I do see your point that replying to Pings fits somewhat better in the behaviour.

pawanjay176 (approved):

beacon_node/lighthouse_network/src/rpc/mod.rs
jxs: thanks for expanding Pawan, yeah that is right but to that we cannot do anything right? as it escapes `Rpc`'s scope what updates the metada's sequence number.

jxs (commented):
> How about adding a comment here based on Pawan’s feedback, to clarify why we need the extra conditions?
> 
> [#6400 (comment)](https://github.com/sigp/lighthouse/pull/6400#discussion_r1764200888)
> 
> > the reason we had the extra condition iirc is that we wanted to generate a RpcFailed event for HandlerErr::Outbound to send to sync and allow sync to remove the peer from all mappings

yup, makes total sense! Thanks Akihito and Pawan for the review

jxs:
@mergifyio queue

Issue #6409, opened 2024-09-18
Implement PeerDAS Subnet Sampling


jimmygchen:
## Description

https://github.com/ethereum/consensus-specs/pull/3870

This consensus PR covers subnet sampling and is targeted for `peerdas-devnet-2`. It was an oversight on my part to think that the PR only removes peer sampling, hence it has not yet been implemented in Lighthouse.

This should be a simple change to subscribe to 4 additional column subnets and require all samples from those topics to be validated before making a block available. The additional column subnets are for the purpose of subnet sampling only:
- the extra sampling subnets does  not need to be advertised.
- the samples from this subnet does not need to be stored, as peers should never query for these columns.

jimmygchen:
Completed in #6410.
Raised #6465 to potentially exclude storing sampling columns, in case we end up going live with subnet sampling.

Issue #3870, opened 2023-01-12
Compute RLP withdrawals roots when computing `blockHash`


michaelsproul:
## Description

In https://github.com/sigp/lighthouse/pull/3794 we added verification of the execution `blockHash`. On the `capella` branch this is currently disabled because we don't verify the withdrawals hash. There was some discussion of using an SSZ root for the withdrawals on the execution side, but it seems that _will not_ come to pass, so we need to verify the RLP withdrawals hash.

The implementation should be straightforward and similar to the implementation for transactions. This issue is just a reminder that we need to do it (I'll assign myself for now, but won't do it immediately).

PR #6410, opened 2024-09-18
Implement Subnet Sampling for PeerDAS

jimmygchen:
## Issue Addressed

Addresses #6409

## Proposed Changes

- Add `SAMPLES_PER_SLOT: 8 ` config.
- Rename `sampling` module to `peer_sampling`.
- Implement subnet sampling.

## Additional Info

This PR is basically a bunch of renames, because we basically treat sampling the same as custody:
- we need all samples for making block available
- we require all samples during sync

With the exception that sampling subnets are not advertised in metadata and ENR. The additional sampling data _may_ not need to be stored, however this is not currently defined in spec and will be clarified later.


beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3836,6 +3836,8 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         }
 
         if let Some(data_columns) = data_columns {
+            // TODO(das): `available_block includes all sampled columns, but we only need to store
+            // custody columns. To be clarified in spec.
```
dapplion, line 3855: Can you link a PR?

jimmygchen: It was discussed in the last breakout call, but there hasn't been any PRs created yet. 
This shoudn't block interop though, because peers should never query for non custody columns.

dapplion:
I'm missing where in the code we filter out the sampled columns and only store the custody requirement.

Plus the additional sampled columns can be not random? We are always fetching the same subnets every slot and then throwing them away.

jimmygchen:
> I'm missing where in the code we filter out the sampled columns and only store the custody requirement.

This hasn't been implemented yet, as it requires more changes and still needs to be clarified in spec ([comment earlier](https://github.com/sigp/lighthouse/pull/6410#discussion_r1776531074))

> Plus the additional sampled columns can be not random? We are always fetching the same subnets every slot and then throwing them away.

Yes that's right - this is the latest version of subnet sampling ([#3870](https://github.com/ethereum/consensus-specs/pull/3870)). I asked a similar question before about randomness, and Francesco doesn't think it matters for the network as a whole (see his explanation on [discord here](https://discord.com/channels/595666850260713488/988798065894907934/1268112023602991205)). And In the future clients can switch over to peer sampling whenever we're ready.

dapplion (approved):

jimmygchen:
@mergify queue

PR #6417, opened 2024-09-19
Allow custody by root requests to have no peers

dapplion:
## Issue Addressed

Currently, a post-PeerDAS lookup will immediately fail if we have no custody peers on a necessary column. This should be a common scenario if the peer manager can't find a well-distributed set of peers to cover all columns. Instantly failing the lookup is sub-optimal as the block will be added to fail chains and delays recovery.

## Proposed Changes

This PR allows lookups so sit in a "stalled" state where one or more individual column requests are sitting idle unable to progress. But:
- When added a new peer to sync, attempt to make progress on all lookups. If the new peer is a custodial of the missing column -> great we progress
- After a certain duration if the lookup has not made progress, it will fail.

### DIFF
--- a/beacon_node/network/src/sync/manager.rs
+++ b/beacon_node/network/src/sync/manager.rs
@@ -35,7 +35,9 @@
 
 use super::backfill_sync::{BackFillSync, ProcessResult, SyncStart};
 use super::block_lookups::BlockLookups;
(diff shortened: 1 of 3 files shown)


realbigsean (approved):
Nice, makes sense to me

realbigsean:
@mergify queue

PR #6420, opened 2024-09-20
Remove Score Ord, PartialOrd, Eq and PartialEq impls 

jxs:
## Issue Addressed

following the conversation on https://github.com/sigp/lighthouse/pull/6407#discussion_r1764222760 this PR follows Rust `std` convention and removes the `Ord` `PartialOrd` `Eq` and `PartialEq` for `Score` which depends on `f64`. Instead as `std` impls `total_cmp` which handles `NaN` cases.
This also reverts #6407 
cc @michaelsproul @eserilev

### DIFF
--- a/beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
+++ b/beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
@@ -1,8 +1,8 @@
 use crate::discovery::enr::PEERDAS_CUSTODY_SUBNET_COUNT_ENR_KEY;
 use crate::discovery::{peer_id_to_node_id, CombinedKey};
 use crate::{metrics, multiaddr::Multiaddr, types::Subnet, Enr, EnrExt, Gossipsub, PeerId};
+use itertools::Itertools;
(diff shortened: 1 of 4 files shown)


beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
```diff
@@ -290,15 +290,11 @@ impl<E: EthSpec> PeerDB<E> {
...
     /// Ties get broken randomly.
     pub fn worst_connected_peers(&self) -> Vec<(&PeerId, &PeerInfo<E>)> {
-        let mut connected = self
-            .peers
+        self.peers
             .iter()
             .filter(|(_, info)| info.is_connected())
-            .collect::<Vec<_>>();
-
-        connected.shuffle(&mut rand::thread_rng());
```
michaelsproul, line 299: Are we intentionally removing the shuffle because the random hashmap ordering is sufficient to break ties?

michaelsproul (approved):
Yeah I think this is nicer and should prevent misuse of a dodgy `PartialEq`/`PartialOrd` implementation.

jxs:
@mergifyio queue

Issue #6332, opened 2024-08-30
Optimize Redb temporary state cleanup


eserilev:
## Description

#4718 Introduces Redb as another database implementation for the beacon node. According to most metrics we've seen so far, its performance is on-par with leveldb in most cases. One case in which it is drastically underperforming is at temp state cleanup during a node restart. The latest restart on holesky took roughly 1.5 hours!

The issue can be found in `do_atomically`

https://github.com/sigp/lighthouse/blob/06490d4c274bd7c7f52e70f0c302513072e5355c/beacon_node/store/src/database/redb_impl.rs#L166-L203

The way this function is structured, we are constantly opening and closing write transactions during each iteration. Since Redb only allows for one open write transaction at a time, and write transactions can only be opened against individual tables, we will need to refactor our garbage collection logic to conform to Redb functionality.

My best idea so far is to make the garbage collection logic atomic on a per table basis. Instead of passing in an `ops` vec that contains transactions across multiple tables, we create a vec for each table. We can then pass each vec into a new function that keeps a single write transaction open across the full vec of `ops`. As long as we're ok with garbage collection only being atomic across individual tables, this should help us get to the performance were looking for.

I think Michael also mentioned that tree-states will help us reduce the amount of temp states being stored in general.

eserilev:
Been mucking around with different optimizations to temp state cleanup. I did get it down from 2hrs to like 30min by doing the delete ops on a per table basis. Using Redb's `remove` method is still pretty slow though. I dont think this method is great for bulk operations.

Theres another method `retain` which applies a predicate to each key value pair in the table. All entries for which predicate are evaluated to false are removed.
(https://docs.rs/redb/latest/redb/struct.Table.html#method.retain)

I have a garbage collection implementation that takes all the temp state roots & pushes them into a HashSet. The predicate in `retain` checks if a key entry exists in the hashset. If the entry exists in the set it returns false, removing that record from the table. Results so far are promising, garbage collection took something on the order of a few seconds (I've so far only tested this on a small batch of temp state roots)

I'm going to let the gown BN run until monday and try redeploying. That should give us enough temp state roots to really stress test these changes. 

Another further optimization is to simply drop/recreate the `BeaconStateTemp` table after deleting the temp state roots from the other relevant tables.

michaelsproul:
I've made a change here that I think will help:

- https://github.com/sigp/lighthouse/pull/6422

eserilev:
[...]

PR #6422, opened 2024-09-23
Clean up temporary state flags while running

michaelsproul:
## Issue Addressed

It has been observed that Lighthouse is pruning thousands of temp state flags on startup:

> Sep 22 05:48:25.130 DEBG Garbage collecting 223069 temporary states, service: freezer_db

The reason for this is that these temp flags are not being cleaned up aside from on startup. They are being created every time the state advance runs and creates an advanced pre-state for a slot. If a block arrives at the slot then the pre-state does _not_ become part of the canonical chain and remains on disk until it is cleaned up during the finalization migration (when `prune_old_hot_states` is called). After the finalization migration, the state is gone from disk but its temporary flag persists until the next startup when `delete_temp_states` runs and deletes them.

## Proposed Changes

When executing a `StoreOp::DeleteState`, always delete any associated temporary flag associated with the state. If we are deleting the state there is no need to retain its temporary flag.

## Additional Info

This PR may help with the Redb pruning issues described here, by reducing the number of items that can build up in the temp state table:

- https://github.com/sigp/lighthouse/issues/6332

### DIFF
--- a/beacon_node/beacon_chain/tests/store_tests.rs
+++ b/beacon_node/beacon_chain/tests/store_tests.rs
@@ -2516,7 +2516,7 @@ async fn pruning_test(
 }
 
 #[tokio::test]
-async fn garbage_collect_temp_states_from_failed_block() {
+async fn garbage_collect_temp_states_from_failed_block_on_startup() {
     let db_path = tempdir().unwrap();
 
     // Wrap these functions to ensure the variables are dropped before we try to open another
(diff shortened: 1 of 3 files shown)


eserilev:
small nit: since `StoreOp::DeleteState` now deletes the temp state flag I think you can also delete `ops.push(StoreOp::DeleteStateTemporaryFlag(state_root));` in `delete_temp_states` (`garbage_collection.rs` line 24)

https://github.com/sigp/lighthouse/blob/5520b3b19d4429f991b81c542cf5c6efca3e743f/beacon_node/store/src/garbage_collection.rs#L18-L26

eserilev:
LGTM

eserilev (approved):

jimmygchen (approved):
LGTM too!

michaelsproul:
@mergify queue

PR #6424, opened 2024-09-23
Modularize tracing executor and metrics rename

AgeManning:
Starting to move common crates to be more friendly to export. 

This PR adds a tracing feature to the task executor, which removes the slog dependencies and logging. 

This also renames the `lighthouse_metrics` crate to just `metrics` (as it doesn't contain lighthouse-specific metric code).


macladson (approved):
This seems good to me!

What happens once we've removed `slog` entirely? Would we remove the `cfg`'s or leave them in in case other projects don't use `tracing`?

jxs (approved):

AgeManning:
I think we remove them entirely

AgeManning:
@mergifyio queue

PR #6427, opened 2024-09-23
Separate type for unaggregated network attestations 

### LINKED ISSUE #6380: Separate type for unaggregated network attestations
## Description

Introduces a `SingleAttestation` type that helps differentiate between unaggregated and aggregated attestations during network processing>

Spec: [https://github.com/ethereum/consensus-specs/pull/3900](https://github.com/ethereum/consensus-specs/pull/3900)

We'll probably want to implement this change in two or three stages:

- [X] Add new `SingleAttestation` gossip topic. When receiving `SingleAttestation` objects over gossip, immediately convert to `Attestation` to reduce implementation complexity. This needs to be included before Electra. [https://github.com/sigp/lighthouse/pull/6488](https://github.com/sigp/lighthouse/pull/6488)
- [ ] Full implementation of `SingleAttestsation`. Don't convert it to a `Attestation` so that we reap the full benefits of the `SingleAttestation` changes. This can optionally be included before Electra. [https://github.com/sigp/lighthouse/pull/6616](https://github.com/sigp/lighthouse/pull/6616)
- [ ] Unify `GossipAttestation` and `GossipSingleAttestation` flows. This entails immediately converting `Attestation`s received over gossip into `SingleAttestations`. This allows us to delete a bunch of code. This should probably only be included post Electra.

eserilev:
## Issue Addressed

Closes #6380

## Proposed Changes

Spec: https://github.com/ethereum/consensus-specs/pull/3900

The invasive spec changes are being worked on here. This is going to include some larger refactors


eserilev:
For now i've added the relevant GossipSub boilerplate and am simply converting `SingleAttestation` to `Attestataion` as soon as its received over gossip. This is roughly the most minimal change we need to make to conform to the relevant spec changes.

If the spec changes are finalized I think I'm going to convert `Attestation` (pre-electra) to `SingleAttestation` when received over gossip and update the relevant downstream code to expect a `SingleAttestation` in all cases

Theres also some changes in the validator client and beacon api that will need to be made

consensus/types/src/attestation.rs
```diff
@@ -350,6 +360,57 @@ impl<E: EthSpec> AttestationElectra<E> {
             Ok(())
         }
     }
+
+    pub fn to_single_attestation(&self) -> Result<SingleAttestation, Error> {
```
eserilev: can convert single attestation into indexed attestation w/o computing the committee index

dapplion:
image

Issue #6018, opened 2024-06-28
Add additional PeerDAS metrics


jimmygchen:
## Description

Add additional PeerDAS metrics to provide us more visibility on the node.

Suggested list from Andrew:

```
just checked out lighthouse das branch specific metrics;
- Full runtime of data column sidecars gossip verification (counter)
- Number of data column sidecars verified for gossip (counter)
- Total count of reconstructed columns (counter)
- Time taken to reconstruct columns (histogram)
- Time taken to compute data column sidecar, including cells, proofs and inclusion proof (histogram)
- Time taken to verify data_column sidecar inclusion proof (histogram)
- Runtime of batched data column kzg verification (histogram)
- Runtime of single data column kzg verification (histogram)

Counts on custody could be useful. `column_index` as labels?
- Total count of columns in custody (counter)

Gossipsub domain metrics will be very handy imo. Should contain topic `data_column_sidecar_{subnet_id}` as labels;
- Number of gossip messages sent to each topic (counter)
- Number of bytes sent to each topic (counter)
- Number of gossip messages received from each topic (including duplicates) (counter)
- Number of bytes received from each topic (including duplicates) (counter)
- Number of gossip messages received from each topic (deduplicated) (counter)
- Number of bytes received from each topic (deduplicated) (counter)

Req/Resp domain metrics would be nice to have. Should contain protocol ID `/eth2/beacon_chain/req/data_column_sidecars_by_root/1/` and `/eth2/beacon_chain/req/data_column_sidecars_by_range/1/` as labels;
- Number of requests sent (counter)
- Number of requests received (counter)
- Number of responses sent (counter)
- Number of responses bytes sent (counter)
- Number of responses received (counter)
- Number of responses bytes received (counter)
```

jimmygchen:
I think we have most of those gossip metrics 
- `gossipsub_topic_msg_recv_bytes_total`:  number of bytes received through gossip
- `gossipsub_topic_msg_sent_bytes_total`: number of bytes sent through gossip
- `gossipsub_topic_msg_recv_counts_unfiltered_total`: number of unfiltered gossip topics received
- `gossipsub_topic_msg_recv_counts_total`: number of gossip msg received
- number of duplicates can be calculated from the above two metrics

We're lacking metrics on bytes sent / received over rpc, I've raised an issue for this (https://github.com/sigp/lighthouse/issues/6114)

[...]

PR #6431, opened 2024-09-24
Add metric 'Number of bytes received from each topic (deduplicated)'

KatyaRyazantseva:
## Issue Addressed

This PR addresses issue #6018, the list of the additional PeerDAS metrics.

## Proposed Changes

- Changed `topic_msg_recv_bytes` metric into `topic_msg_recv_bytes_unfiltered`. This can be a critical change, but needed to put metrics in order.
- Implemented `topic_msg_recv_bytes` as deduplicated.


beacon_node/lighthouse_network/gossipsub/src/types.rs
```diff
@@ -237,6 +238,21 @@ impl fmt::Debug for Message {
     }
 }
 
+/// The byte size of a message
+impl Message {
+    pub(crate) fn get_size(&self) -> usize {
+        self.source
+            .as_ref()
+            .map_or(0, |m| 1 + sizeof_len(m.to_bytes().len()))
```
dapplion, line 246: The PeerID has a fixed length by definition, we could just precompute it

beacon_node/lighthouse_network/gossipsub/src/behaviour.rs
```diff
@@ -1837,7 +1837,7 @@ where
 
         // Record the received message with the metrics
         if let Some(metrics) = self.metrics.as_mut() {
-            metrics.msg_recvd(&message.topic);
+            metrics.msg_recvd(&message.topic, message.get_size());
```
dapplion, line 1848: Here you use `get_size()` and above we use `raw_protobuf_len` what motivates the difference?

ackintosh: I think we should use `raw_protobuf_len` to obtain the byte length of received message. `Message` is transformed data from `RawMessage` and is missing some fields compared to `RawMessage`, so `message.get_size()` returns a smaller byte size than the actually received bytes.

beacon_node/lighthouse_network/gossipsub/src/topic.rs
```diff
@@ -82,6 +82,10 @@ impl TopicHash {
     pub fn as_str(&self) -> &str {
         &self.hash
     }
+
+    pub fn hash_byte_len(&self) -> usize {
+        self.hash.as_bytes().len()
+    }
```
jxs, line 88: nit: we are only using this once  and it's a one liner, no need for a function to abstract the one liner

beacon_node/lighthouse_network/gossipsub/src/behaviour.rs
jxs: yeah the `get_size` function is unrequired we can just use `raw_protobuf_len`

jimmygchen:
Closing this PR due to long period of inactivity. Feel free to re-open this if you'd like to continue work on this.

PR #6436, opened 2024-09-25
Added required `--force-bls-withdrawal-credentials` description to `--disable-deposits` usage

### LINKED ISSUE #6391: `--disable-deposits` flag of validator manager `create` command requires BLS withdrawal credentials with unclear failure message
## Description

Conflicting options with unclear docs on how to resolve.

## Version

v5.3.0

## Present Behaviour

This command fails:

`lighthouse vm create --disable-deposits --count 1 --output-path $OUTPUT`

with output
> `--eth1-withdrawal-address is required.`

Upon closer inspection, it looks like the issue is that the tool is trying to make `0x01` withdrawal credentials. But even if you supply `--eth1-withdrawal-address`, we get a conflicting message about use of `--disable-deposits`.

## Expected Behaviour

I should be able to use `--disable-deposits`.

## Steps to resolve

The resolution is to use the `--force-bls-withdrawal-credentials` but this isn't clear from the docs and was only apparent after reading the source code. Could simply add some context to the help message for the `--disable-deposits` flag.

hopinheimer:
fixes #6391


validator_manager/src/create_validators.rs
```diff
@@ -112,7 +112,9 @@ pub fn cli_app() -> Command {
                     "When provided don't generate the deposits JSON file that is \
                     commonly used for submitting validator deposits via a web UI. \
                     Using this flag will save several seconds per validator if the \
-                    user has an alternate strategy for submitting deposits.",
+                    user has an alternate strategy for submitting deposits.\
+                    If used, the --force-bls-withdrawal-credentials is also required\
```
michaelsproul:
```suggestion
                    If used, the --force-bls-withdrawal-credentials is also required \
```

michaelsproul (changes requested):
You'll also need to rebuild the CLI docs in the book using `./scripts/cli.sh`

book/src/help_bn.md
```diff
@@ -603,6 +603,4 @@ Flags:
   -z, --zero-ports
           Sets all listening TCP/UDP ports to 0, allowing the OS to choose some
           arbitrary free ports.
-```
-
-<style> .content main {max-width:88%;} </style>
+```n<style> .content main {max-width:88%;} </style>
```
michaelsproul: this looks wrong, did the script do this?

michaelsproul: which OS did you run it on?

hopinheimer: ran a `make cli-local`

hopinheimer: mac os

hopinheimer: yeah I thought something was off but the scripts/cli.sh had this

Screenshot 2024-09-26 at 12 10 11 AM

michaelsproul: ah yeah, macOS sed is a bit different

I'll run it on Linux and fix this

hopinheimer: do you think I should add this 

```
if [[ "$OSTYPE" == "darwin"* ]]; then
    sed -i '' -e '$a\'$'\n''<style> .content main {max-width:88%;} </style>' "$file"
else
    sed -i -e '$a\'$'\n''<style> .content main {max-width:88%;} </style>' "$file"
fi
```

to make our lives easier in the future?

michaelsproul: that would be great actually

michaelsproul: thank you

michaelsproul: that regexp might need a tweak? it looks the same as the Linux one?

hopinheimer: well today I learned I'm still a noob programmer, but yeah I was defeated by a linux command.  
I think using a `printf` just like the command to add the code block quote right above this line just solves the problem in an easy way and sed is just a little ambiguous for different distros. 

there's another issue I'd like to point out 

Screenshot 2024-09-26 at 9 39 40 AM

when I run `make cli` which uses docker this happens, if that would have worked I would have not come across the current issue. do you want to create a separate tracking issue for that?

michaelsproul: Hmm, I don't think that image is ever going to be public. Maybe that's just a doc issue

michaelsproul (approved):

book/src/help_bn.md
michaelsproul: Oh, it is public but it doesn't work on ARM.

michaelsproul: If you have Rosetta you can run x86_64 images on macOS using emulation. It's slower but it works.

scripts/cli.sh
```diff
@@ -16,7 +16,7 @@ write_to_file() {
     printf "# %s\n\n\`\`\`\n%s\n\`\`\`" "$program" "$cmd" > "$file"
 
     # Adjust the width of the help text and append to the end of file
-    sed -i -e '$a\'$'\n''\n''<style> .content main {max-width:88%;} </style>' "$file"
+    printf '%s\n\n' '' '<style> .content main {max-width:88%;} </style>' >> "$file"
```
michaelsproul:
```suggestion
    printf '%s\n' '' '<style> .content main {max-width:88%;} </style>' >> "$file"
```

I think we might want to remove the 2nd newline. The markdown linter isn't happy

michaelsproul (changes requested):
Markdown linter needs appeasing

scripts/cli.sh
chong-he: For consistency, may I suggest we use double quote "" (as in the code above)? We can also remove the `''` in the middle of the line for simplicity.
The following should do it:
```suggestion
    printf "\n\n%s\n" "<style> .content main {max-width:88%;} </style>" >> "$file"
```
The `\n\n` is to create an empty line after the code block

hopinheimer: I've made the requested changes @chong-he

chong-he:
This looks good now, all checks passed

hopinheimer:
can this be merged? I'm using the changes in a different PR and yes I will shift to ubuntu soon :)

jxs (approved):
LGTM! Thanks!

jxs:
@mergifyio queue

AgeManning (approved):

AgeManning:
@mergifyio queue

AgeManning:
@michaelsproul - I think you might need to unblock this

michaelsproul (approved):

jxs:
@mergifyio queue

PR #6438, opened 2024-09-26
estimate the total inbound bandwidth of IDONTWANT messages in bytes

jxs:
## Issue Addressed

While trying to understand a report from a user which its inbound bandwidth increased considerably I noticed that in general we don't have a way to estimate what is the length in bytes of each IDONTWANT message as it depends on the message id.
This PR addresses that


AgeManning (approved):

AgeManning:
@mergifyio queue

PR #6443, opened 2024-09-27
Prevent sync lookups from reverting to awaiting block

dapplion:
## Issue Addressed

Sync lookups have an "oracle" into the `da_checker` state. Blob or custody column requests are only progressed when the block is already downloaded. This allows deterministic requests after we are sure the block has data.

However, the current implementation has an interesting flaw. If the "oracle" is buggy, the sync lookup can consider that:
- t_0: block is not downloaded, halt custody column downloads
- t_1: block is downloaded, continue column downloads
- t_2: block is not downloaded, halt custody column downloads (again)

This sequence of states should never happen but the type models allow it to happen. 

## Proposed Changes

Instead, we can model `SingleBlockLookup` such that once the block is seen, it can never be unseen. This new enum:

```rust
enum ComponentRequests<E: EthSpec> {
    WaitingForBlock,
    ActiveBlobRequest(BlobRequestState<E>, usize /* expected blob count */),
    ActiveCustodyRequest(CustodyRequestState<E>),
    NotNeeded(&'static str),
}
```

and moving the "oracle" logic from the network context to the lookup logic achieve this effect.

I'm leaving the PR in draft for those familiar with the topic to thumbs up this design. The overall goal is to reduce the chance of lookups getting stuck if we change the da_checker logic.

### DIFF
--- a/beacon_node/network/src/sync/block_lookups/single_block_lookup.rs
+++ b/beacon_node/network/src/sync/block_lookups/single_block_lookup.rs
@@ -221,8 +221,6 @@ impl<T: BeaconChainTypes> SingleBlockLookup<T> {
                 // Wait to download the block before downloading blobs. Then we can be sure that the
                 // block has data, so there's no need to do "blind" requests for all possible blobs and
                 // latter handle the case where if the peer sent no blobs, penalize.
-                // - if `downloaded_block_expected_blobs` is Some = block is downloading or processing.
-                // - if `num_expected_blobs` returns Some = block is processed.
                 //
                 // Lookup sync event safety: Reaching this code means that a block is not in any pre-import
                 // cache nor in the request state of this lookup. Therefore, the block must either: (1) not
--- a/beacon_node/network/src/sync/block_lookups/single_block_lookup.rs
+++ b/beacon_node/network/src/sync/block_lookups/single_block_lookup.rs
@@ -4,7 +4,7 @@ use crate::sync::network_context::{
     LookupRequestResult, PeerGroup, ReqId, RpcRequestSendError, SendErrorProcessor,
     SyncNetworkContext,
 };
-use beacon_chain::BeaconChainTypes;
+use beacon_chain::{BeaconChainTypes, BlockProcessStatus};
 use derivative::Derivative;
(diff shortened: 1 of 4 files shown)


beacon_node/network/src/sync/block_lookups/single_block_lookup.rs
```diff
@@ -161,9 +181,68 @@ impl<T: BeaconChainTypes> SingleBlockLookup<T> {
...
+                    );
+                } else {
+                    self.component_requests = ComponentRequests::NotNeeded("outside da window");
+                }
+            } else {
+                // Wait to download the block before downloading blobs. Then we can be sure that the
+                // block has data, so there's no need to do "blind" requests for all possible blobs and
+                // latter handle the case where if the peer sent no blobs, penalize.
+                // - if `downloaded_block_expected_blobs` is Some = block is downloading or processing.
+                // - if `num_expected_blobs` returns Some = block is processed.
```
jimmygchen: These two lines seems outdated after moving here

beacon_node/network/src/sync/block_lookups/single_block_lookup.rs
```diff
@@ -161,9 +181,68 @@ impl<T: BeaconChainTypes> SingleBlockLookup<T> {
...
+                //
+                // Lookup sync event safety: Reaching this code means that a block is not in any pre-import
+                // cache nor in the request state of this lookup. Therefore, the block must either: (1) not
+                // be downloaded yet or (2) the block is already imported into the fork-choice.
+                // In case (1) the lookup must either successfully download the block or get dropped.
+                // In case (2) the block will be downloaded, processed, reach `DuplicateFullyImported`
+                // and get dropped as completed.
```
jimmygchen, line 230: Ah right, this might have explained why we're seeing request being sent after importing - and seems like we keep retrying if peers can't respond to us with the blocks. Not sure if there's a smarter way to drop this though.

jimmygchen (approved):
This looks great to me! I think moving the logic up to `SingleBlockLookup` is a lot cleaner, and avoids the possibility of the scenario you described.

jimmygchen:
@mergify queue

PR #6444, opened 2024-09-28
add a unique integer id to Rpc requests

jxs:
## Issue Addressed

This PR aims to replace the `(ConnectionId, SubstreamId)` identification method for the Rpc to a simple integer value that is unique for each request. This solves two main issues:
- avoids passing the `(ConnectionId, SubstreamId)` combination into the consensus land 
- allows the Rpc `Behaviour` to match a response to a request and potentially simplify a lot https://github.com/sigp/lighthouse/pull/5923 so that the `Handler` and `Behaviour` don't need to share a `Mutex`

review by commit is suggested

### DIFF
--- a/beacon_node/lighthouse_network/src/rpc/mod.rs
+++ b/beacon_node/lighthouse_network/src/rpc/mod.rs
@@ -26,14 +26,13 @@ pub(crate) use handler::{HandlerErr, HandlerEvent};
 pub(crate) use methods::{
     MetaData, MetaDataV1, MetaDataV2, MetaDataV3, Ping, RpcResponse, RpcSuccessResponse,
 };
-pub(crate) use protocol::InboundRequest;
+pub use protocol::RequestType;
 
 pub use handler::SubstreamId;
 pub use methods::{
     BlocksByRangeRequest, BlocksByRootRequest, GoodbyeReason, LightClientBootstrapRequest,
     ResponseTermination, RpcErrorResponse, StatusMessage,
 };
-pub(crate) use outbound::OutboundRequest;
 pub use protocol::{max_rpc_size, Protocol, RPCError};
 
 use self::config::{InboundRateLimiterConfig, OutboundRateLimiterConfig};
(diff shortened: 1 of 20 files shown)


beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -368,13 +400,29 @@ where
...
+                substream_id,
+                r#type,
+            })) => {
                 if let Some(limiter) = self.limiter.as_mut() {
                     // check if the request is conformant to the quota
-                    match limiter.allows(&peer_id, &req) {
+                    match limiter.allows(&peer_id, &r#type) {
+                        Ok(()) => {
+                            // send the event to the user
+                            self.events.push(ToSwarm::GenerateEvent(RPCMessage {
```
AgeManning: Aren't we now sending two events? One here and the other one on line 493?

beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -403,29 +451,39 @@ where
...
                                 ),
                             );
                             return;
                         }
-                        // No rate limiting, continue.
-                        Ok(_) => {}
                     }
+                } else {
+                    // No rate limiting, send the event to the user
+                    self.events.push(ToSwarm::GenerateEvent(RPCMessage {
```
AgeManning: Similarly here, two events being emitted

beacon_node/lighthouse_network/src/rpc/mod.rs
jxs: Thanks age Addressed!

beacon_node/lighthouse_network/src/rpc/mod.rs
jxs: yup, hazards of a merge conflict, thanks!

AgeManning (approved):

AgeManning:
@mergifyio queue

PR #6435, opened 2024-09-25
Move EL requests to BeaconBlockBody

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Move all EL requests from the `ExecutionPayload` to `BeaconBlockBody` according to https://github.com/ethereum/consensus-specs/pull/3875 . Modify engine api to support the changes. 

Engine api changes are not yet merged. ~~The direction seems to be removing `ExecutionPayloadBodyV2` and moving requests from `JsonExecutionPayloadV4` to `JsonGetPayloadResponse` (https://github.com/ethereum/execution-apis/pull/587/) and changing requests to an opaque ssz bytes container (https://github.com/ethereum/execution-apis/pull/577/) in the json object~~.

The engine api changes are here https://github.com/ethereum/execution-apis/pull/591

PR #6445, opened 2024-09-28
Electra updates for v1.5.0-alpha.6

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

1. https://github.com/ethereum/consensus-specs/pull/3882 . I modified the existing `process_single_slashing` function instead of creating a new function for electra as I think both the pre-electra and electra functions have the same logic. I'm not completely sure if we need to make a separate >= Electra function and keep the pre-electra logic as is.
Edit: Confirmed that we need different behaviour for pre and post electra. Modified in 
65bac11
3. https://github.com/ethereum/consensus-specs/pull/3915 : just remove unused Consolidations domain
4. Rename `get_validator_max_effective_balance` https://github.com/ethereum/consensus-specs/pull/3909
5. Move execution_requests to beacon block body https://github.com/ethereum/consensus-specs/pull/3875

Remaining [spec updates](https://github.com/ethereum/consensus-specs/releases/tag/v1.5.0-alpha.6) for electra are cosmetic

~~Blocked on https://github.com/sigp/lighthouse/pull/6435 for EF tests to pass.~~ Merged #6435 into this PR to get the tests passing.


beacon_node/execution_layer/src/engine_api/new_payload_request.rs
```diff
@@ -43,6 +44,8 @@ pub struct NewPayloadRequest<'block, E: EthSpec> {
     pub versioned_hashes: Vec<VersionedHash>,
     #[superstruct(only(Deneb, Electra))]
     pub parent_beacon_block_root: Hash256,
+    #[superstruct(only(Electra))]
+    pub execution_requests_list: Vec<Bytes>,
```
ethDreamer: This is kinda rough because it does processing and holds the data rather than a reference. I remember when I created the `NewPayloadRequest`, I got rid of copies so it was lightweight because (if I remember correctly) there's points in the code where `NewPayloadRequest` is created, then some validations are done and it gets dropped. And then later another request is made.

It would be cool to continue this and rather than holding a `Vec<Bytes>` here just hold a `&'block ExecutionRequests<E>` just call `get_execution_requests_list()` on that reference when inside`new_payload_v4_electra()`. 

Then this object would remain lightweight and the processing would only be done when it actually needs to be.

ethDreamer (changes requested):
Looks good! Almost done reviewing, just gotta do the single pass changes.

beacon_node/execution_layer/src/engine_api/new_payload_request.rs
pawanjay176: good call, will make the changes

pawanjay176: Added in ebc7265

ethDreamer (approved):
lgtm!

beacon_node/execution_layer/src/engine_api/json_structures.rs
```diff
@@ -381,6 +339,71 @@ impl<E: EthSpec> From<JsonExecutionPayload<E>> for ExecutionPayload<E> {
...
+            1 => Some(Self::Withdrawal),
+            2 => Some(Self::Consolidation),
+            _ => None,
+        }
+    }
+}
+
+/// Format of `ExecutionRequests` received over the engine api.
+///
+/// Array of ssz-encoded requests list` encoded as hex bytes.
```
michaelsproul: There's a stray backtick here, not sure what was intended:

```suggestion
/// Array of ssz-encoded requests list encoded as hex bytes.
```

beacon_node/execution_layer/src/lib.rs
```diff
@@ -194,6 +198,8 @@ pub enum BlockProposalContents<E: EthSpec, Payload: AbstractExecPayload<E>> {
         kzg_commitments: KzgCommitments<E>,
         /// `None` for blinded `PayloadAndBlobs`.
         blobs_and_proofs: Option<(BlobsList<E>, KzgProofs<E>)>,
+        // TODO(pawan): this should probably be a separate variant/superstruct
```
michaelsproul: Maybe we change this to `TODO(electra)` before merging? Or do you want to fix it first?

```suggestion
        // TODO(electra): this should probably be a separate variant/superstruct
```

beacon_node/execution_layer/src/test_utils/handle_rpc.rs
```diff
@@ -373,6 +373,8 @@ pub async fn handle_rpc<E: EthSpec>(
                                 ))?
                                 .into(),
                             should_override_builder: false,
+                            // TODO(pawan): add EL requests in mock el
```
michaelsproul:
```suggestion
                            // TODO(electra): add EL requests in mock el
```

michaelsproul (approved):
LGTM!

Just a few minor comments

ethDreamer:
@Mergifyio queue

Issue #6440, opened 2024-09-27
Invalid data columns with 0 blobs are not rejected by Lighthouse


jimmygchen:
## Description

On `peer-das-devnet-2`, it has been observed that Lighthouse does not reject data columns from a block with no blobs.

The check here passes when there are 0 KZG commitments in the block:

https://github.com/sigp/lighthouse/blob/279270533128ef8069765d04c280db993db12905/consensus/types/src/data_column_sidecar.rs#L84-L94

We should check for empty KZG commitments before passing gossip verification here.

Thanks to Prysm team for sharing this finding!

## More info

I wrote a failing test for this

```
    #[test]
    fn test_verify_data_column_inclusion_proof() {
        let (block, blobs) = generate_rand_block_and_blobs::<E>(
            ForkName::Deneb,
            NumBlobs::Number(0),
            &mut thread_rng(),
        );
        let blobs = blobs.into_iter().map(|b| b.blob).collect::<Vec<_>>();
        let columns =
            blobs_to_data_column_sidecars(&blobs.into(), &block, &KZG_PEERDAS, &E::default_spec())
                .unwrap();
        for column in columns {
            assert!(!column.verify_inclusion_proof());
        }
    }
```

jimmygchen:
This may have caused our nodes to store these invalid data columns and serve them to the network. This explains why our range sync is breaking with this error:
```
Blocks and blobs request for range received invalid data, error: Not all columns consumed
```

jimmygchen:
Completed in #6454

PR #6454, opened 2024-10-02
Fix invalid data column sidecars getting accepted

jimmygchen:
## Issue Addressed

Addresses #6440. 

Data columns with empty cells were accepted because it's possible to generate a valid proof for an empty `kzg_commiments` list, therefore the invalid data columns was accepted and broke denvet-2 as clients react differently to this.

This PR adds an additional validation for kzg commitment size.

### DIFF
--- a/beacon_node/beacon_chain/src/data_column_verification.rs
+++ b/beacon_node/beacon_chain/src/data_column_verification.rs
@@ -116,12 +116,6 @@ pub enum GossipDataColumnError {
     ///
     /// The column sidecar is invalid and the peer is faulty
     InvalidInclusionProof,
-    /// Data column not expected for a block with empty kzg commitments.
-    ///
-    /// ## Peer scoring
-    ///
-    /// The column sidecar is invalid and the peer is faulty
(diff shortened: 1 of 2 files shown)


pawanjay176 (approved):
LGTM.
Should we update the spec to explicitly not allow for sidecars with empty elements?

Edit: There's already a spec PR https://github.com/ethereum/consensus-specs/pull/3953 . Maybe we duplicate the spec function in our code too to be consistent with the rest of the checks @jimmygchen ?

jimmygchen:
> LGTM. Should we update the spec to explicitly not allow for sidecars with empty elements?
> 
> Edit: There's already a spec PR [ethereum/consensus-specs#3953](https://github.com/ethereum/consensus-specs/pull/3953) . Maybe we duplicate the spec function in our code too to be consistent with the rest of the checks @jimmygchen ?

Yep good idea, done!

pawanjay176 (approved):
Nice

jimmygchen:
@mergify queue

Issue #6437, opened 2024-09-25
Possible upgrade to lighthouse's IDONTWANT GossipSub implementation


cortze:
## Description
After last Thursday's discussion on the bandwidth savings that IDONTWANT messages could bring, we've been (the [ProbeLab](https://probelab.io/) team) monitoring the behaviour of GossipSub after the addition of these into [`go-libp2p`](https://github.com/libp2p/go-libp2p-pubsub/pull/553).

We've used our light-network-client [`Hermes`](https://github.com/probe-lab/hermes) to trace down interactions at the libp2p level, and they have shown that lighthouse's libp2p gossipsub implementation could be more efficient.

## Version
I could narrow down the behaviour to `v5.3.0-d6ba8c3` from a remote peer's `AgentVersion` (`Lighthouse/v5.3.0-d6ba8c3/x86_64-linux`). However, the code seems to be present at the `stable` branch as well.

## Present Behaviour
The current logic to process an incoming message will ALWAYS send an `IDONTWANT` message when a message is received. 
As far as I could see in [the following code snippet](https://github.com/jxs/lighthouse/blob/053a272952507e95e8f82ae054933e8674fbfced/beacon_node/lighthouse_network/gossipsub/src/behaviour.rs#L1812), there is no filter that would prevent the host from sending an `IDONTWANT` after the arrival of any message (even if it's duplicated).
```rust
fn handle_received_message(
        ...
        // Broadcast IDONTWANT messages.
        self.send_idontwant(&raw_message, &msg_id, propagation_source);

        // Check the validity of the message
        // Peers get penalized if this message is invalid. We don't add it to the duplicate cache
        // and instead continually penalize peers that repeatedly send this message.
        if !self.message_is_valid(&msg_id, &mut raw_message, propagation_source) {
            return;
        }

        if !self.duplicate_cache.insert(msg_id.clone()) {
            tracing::debug!(message=%msg_id, "Message already received, ignoring");
            if let Some((peer_score, ..)) = &mut self.peer_score {
                peer_score.duplicated_message(propagation_source, &msg_id, &message.topic);
            }
            self.mcache.observe_duplicate(&msg_id, propagation_source);
            return;
        }
        tracing::debug!(
            message=%msg_id,
            "Put message in duplicate_cache and resolve promises"
        );
        ...
``` 

## Expected Behaviour
[...]

PR #6456, opened 2024-10-02
IDONTWANT message optimisation to cutoff for smaller messages

hopinheimer:
addresses #6437


beacon_node/lighthouse_network/gossipsub/src/behaviour.rs
```diff
@@ -1828,8 +1825,12 @@ where
                 peer_score.duplicated_message(propagation_source, &msg_id, &message.topic);
             }
             self.mcache.observe_duplicate(&msg_id, propagation_source);
+
+            // Broadcast IDONTWANT messages.
+            self.send_idontwant(&raw_message, &msg_id, propagation_source);
```
AgeManning, line 1833: I think we want this after the return, after the if statement. 

At the moment, this will broadcast an IDONTWANT only if the message is a duplicate. 

We want the logic to broadcast only if the message isn't a duplicate.

beacon_node/lighthouse_network/gossipsub/src/config.rs
```diff
@@ -370,6 +371,16 @@ impl Config {
     pub fn forward_queue_duration(&self) -> Duration {
         self.connection_handler_forward_duration
     }
+
+    // Controls the message size threshold for which IDONTWANT messages are sent.
+    // Sending IDONTWANT messages for small messages can have a negative effect to the overall
+    // traffic and CPU load. This acts as a lower bound cutoff for the message size to which 
+    // IDONTWANT won't be sent to peers. Only works if the peers support Gossipsub1.3 
+    // (see https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.2.md#idontwant-message)
+    // default is 1kb
```
AgeManning: Might be worth mentioning that the unit for the usize value is bytes.

AgeManning (commented):
Nice, thanks for the help.

beacon_node/lighthouse_network/gossipsub/src/behaviour.rs
hopinheimer: oh yeah makes sense!

beacon_node/lighthouse_network/gossipsub/src/config.rs
hopinheimer: I always mess this up. resolved!

beacon_node/lighthouse_network/gossipsub/src/behaviour.rs
```diff
@@ -2699,6 +2701,10 @@ where
             return;
         };
 
+        if message.raw_protobuf_len() < self.config.idontwant_message_size_threshold(){
```
jxs: can we move this check to line `1833`? So that it's explicit that we are only calling `send_idontwant` if the message size is inferior to the threshold instead of having the function decide that.

beacon_node/lighthouse_network/gossipsub/src/config.rs
```diff
@@ -370,6 +371,16 @@ impl Config {
     pub fn forward_queue_duration(&self) -> Duration {
         self.connection_handler_forward_duration
     }
+
+    // Controls the message size threshold for which IDONTWANT messages are sent.
+    // Sending IDONTWANT messages for small messages can have a negative effect to the overall
+    // traffic and CPU load. This acts as a lower bound cutoff for the message size to which 
+    // IDONTWANT won't be sent to peers. Only works if the peers support Gossipsub1.3 
+    // (see https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.2.md#idontwant-message)
+    // default is 1kB
```
jxs:
```suggestion
    // The size threshold for which IDONTWANT messages are sent.
    // Sending IDONTWANT messages for small messages can have a negative effect to the overall
    // traffic and CPU load. This acts as a lower bound cutoff for the message size to which 
    // IDONTWANT won't be sent to peers. Only works if the peers support Gossipsub1.2 
    // (see https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.2.md#idontwant-message)
    // default is 1kB
```

beacon_node/lighthouse_network/gossipsub/src/config.rs
```diff
@@ -825,6 +837,17 @@ impl ConfigBuilder {
         self
     }
 
+    // Controls the message size threshold for which IDONTWANT messages are sent.
+    // Sending IDONTWANT messages for small messages can have a negative effect to the overall
+    // traffic and CPU load. This acts as a lower bound cutoff for the message size to which 
+    // IDONTWANT won't be sent to peers. Only works if the peers support Gossipsub1.3 
+    // (see https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.2.md#idontwant-message)
+    // default is 1kB
```
jxs:
```suggestion
    // The size threshold for which IDONTWANT messages are sent.
    // Sending IDONTWANT messages for small messages can have a negative effect to the overall
    // traffic and CPU load. This acts as a lower bound cutoff for the message size to which 
    // IDONTWANT won't be sent to peers. Only works if the peers support Gossipsub1.2 
    // (see https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.2.md#idontwant-message)
    // default is 1kB
```

jxs (commented):
Thanks! You need to `cargo fmt` to pass the `check code` CI job. The tests also need to be updated, and we should probably add another test for this case, i.e. to not send IDONTWANT when the size is inferior to the limit

beacon_node/lighthouse_network/gossipsub/src/config.rs
```diff
@@ -370,6 +371,16 @@ impl Config {
...
         self.connection_handler_forward_duration
     }
+
+    // Controls the message size threshold for which IDONTWANT messages are sent.
+    // Sending IDONTWANT messages for small messages can have a negative effect to the overall
+    // traffic and CPU load. This acts as a lower bound cutoff for the message size to which 
+    // IDONTWANT won't be sent to peers. Only works if the peers support Gossipsub1.3 
+    // (see https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.2.md#idontwant-message)
+    // default is 1kB
+    pub fn idontwant_message_size_threshold(&self) -> usize{
```
pawanjay176: We would need to upstream this builder config to the lighthouse config as well https://github.com/sigp/lighthouse/blob/unstable/beacon_node/lighthouse_network/src/config.rs#L480-L502

I would also suggest adding a lighthouse cli flag to control the limit so we can test it for other feature devnets (e.g. peerdas)

hopinheimer: understood will make the requested changes and add it to a cli flag

beacon_node/lighthouse_network/gossipsub/src/behaviour.rs
hopinheimer: also the the logic was a little off. I have changed it and written passing test for the same

hopinheimer:
waiting on #6436 to build the cli docs. I use a mac

beacon_node/lighthouse_network/src/config.rs
```diff
@@ -352,6 +356,7 @@ impl Default for Config {
             outbound_rate_limiter_config: None,
             invalid_block_storage: None,
             inbound_rate_limiter_config: None,
+            idontwant_message_size_threshold: 1000,
```
pawanjay176: It'd be good to extract the default value into a `IDONTWANT_MESSAGE_SIZE_THRESHOLD` const value.

hopinheimer: sounds reasonable. made the requested changes

hopinheimer: `target_peers` few line above also seems to have inline declaration should I extract it too?

jxs (approved):
LGTM! cc @AgeManning if you wanna take a final look

AgeManning (approved):

AgeManning:
aha, i was trying to fix the CI, but getting merge conflicts. You beat me to it. :). Thanks

hopinheimer:
haha, Imma ss this!

AgeManning:
@mergifyio queue

jxs:
@mergifyio queue

michaelsproul:
@mergify dequeue

michaelsproul:
some stuff got stuck, so just removing this from the queue for now

michaelsproul:
@mergify queue

michaelsproul:
@mergify requeue

PR #6462, opened 2024-10-03
Remove generic E from RequestId

jxs:
## Issue Addressed
Continue cleaning up rpc code by removing unrequired generic `E` from request id, and passing `MAX_BLOBS_PER_BLOCK` when constructing a `BlobsByRange` request

### DIFF
--- a/beacon_node/lighthouse_network/src/rpc/methods.rs
+++ b/beacon_node/lighthouse_network/src/rpc/methods.rs
@@ -8,7 +8,6 @@ use ssz_derive::{Decode, Encode};
 use ssz_types::{typenum::U256, VariableList};
 use std::collections::BTreeMap;
 use std::fmt::Display;
-use std::marker::PhantomData;
 use std::ops::Deref;
 use std::sync::Arc;
 use strum::IntoStaticStr;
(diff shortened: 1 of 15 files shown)


beacon_node/lighthouse_network/src/rpc/methods.rs
```diff
@@ -323,11 +314,14 @@ pub struct BlobsByRangeRequest {
 
     /// The number of slots from the start slot.
     pub count: u64,
+
+    /// maximum number of blobs in a single block.
+    pub max_blobs_per_block: usize,
```
dapplion, line 319: This value is the same on all requests, does it need to be appended to every request? Or can it be part of the handler struct

jxs: Hi Lion, good question, the thing is:
https://github.com/sigp/lighthouse/blob/48dd3f385cd3ecde1baeb9135815f376d79b048a/beacon_node/network/src/sync/network_context.rs#L406-L415
the request is created here for example, it's outside of the `Rpc` `NetworkBehaviour`.
Also `BlobsByRangeRequest::max_blobs_requested(&self)` requires it, how would you envision storing `max_blobs_per_block`? 
My objective is for `Rpc` to be completely unaware of `E: EthSpec` so that we can uncouple it from the rest of lighthouse and all its config values are passed dynamically.

dapplion: Actually, I take that back, if we increase the blob count we will have different requests with different limits at runtime. So okay with me to keep it on the request

AgeManning (approved):

AgeManning:
@mergifyio queue

beacon_node/lighthouse_network/src/rpc/methods.rs
michaelsproul: I don't think we should have added this field to `BlobsByRangeRequest`. It breaks the SSZ representation.

PR #6466, opened 2024-10-05
Bound min size of dynamic processor queues

### LINKED ISSUE #6463: Attestation queues have max size 0 if num validators is < slots_per_epoch
## Description

This code will set the maximum size of the attestation queue to 0 if `active_validator_count` is less than `slots_per_epoch`. We should probably add a minimum queue size for especially small networks. Maybe 16-32.

https://github.com/sigp/lighthouse/blob/35ac9f4602a7c68d258f039f93828fe604f37ab3/beacon_node/beacon_processor/src/lib.rs#L149-L152

## Version

v5.2.0 onwards.

dapplion:
## Issue Addressed

A discord user notes that in networks of < SLOTS_PER_EPOCH the beacon node did not process attestations. 

- #6463

It's due to integer division here

https://github.com/sigp/lighthouse/blob/35ac9f4602a7c68d258f039f93828fe604f37ab3/beacon_node/beacon_processor/src/lib.rs#L149-L152

## Proposed Changes

Min against a const

Closes #6463

### DIFF
--- a/beacon_node/beacon_processor/src/lib.rs
+++ b/beacon_node/beacon_processor/src/lib.rs
@@ -1697,3 +1697,21 @@ impl Drop for SendOnDrop {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use types::{BeaconState, ChainSpec, Eth1Data, ForkName, MainnetEthSpec};
+
+    #[test]
+    fn min_queue_len() {
+        // State with no validators.
+        let spec = ForkName::latest().make_genesis_spec(ChainSpec::mainnet());
+        let genesis_time = 0;
+        let state = BeaconState::<MainnetEthSpec>::new(genesis_time, Eth1Data::default(), &spec);
+        assert_eq!(state.validators().len(), 0);
+        let queue_lengths = BeaconProcessorQueueLengths::from_state(&state, &spec).unwrap();
+        assert_eq!(queue_lengths.attestation_queue, MIN_QUEUE_LEN);
+        assert_eq!(queue_lengths.unknown_block_attestation_queue, MIN_QUEUE_LEN);
+    }
+}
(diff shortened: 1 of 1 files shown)


michaelsproul (commented):
I feel like this could easily regress with the beacon processor refactors. Wanna add a test?

beacon_node/beacon_processor/src/lib.rs
```diff
@@ -155,9 +160,15 @@ impl BeaconProcessorQueueLengths {
             aggregate_queue: 4096,
             unknown_block_aggregate_queue: 1024,
             // Capacity for a full slot's worth of attestations if subscribed to all subnets
-            attestation_queue: active_validator_count / slots_per_epoch,
+            attestation_queue: std::cmp::min(
```
chong-he: I may be wrong here, I don't understand why it is `std::cmp::min`. From my understanding, it takes the minimum value of the two in comparison. I thought it should be `std::cmp::max`. 

Because if it is `min`, then when `active_validator_count / slots_per_epoch` is, say 16, then the queue would be still 0? But if we use `max`, then for small validator count, it will return `MIN_QUEUE_LEN` which is 128.

beacon_node/beacon_processor/src/lib.rs
```diff
@@ -155,9 +160,15 @@ impl BeaconProcessorQueueLengths {
...
+                active_validator_count / slots_per_epoch,
+                MIN_QUEUE_LEN,
+            ),
             // Capacity for a full slot's worth of attestations if subscribed to all subnets
-            unknown_block_attestation_queue: active_validator_count / slots_per_epoch,
+            unknown_block_attestation_queue: std::cmp::min(
```
chong-he: Same as above

chong-he (commented):
Not sure if we should use `std::cmp::max` here? Happy to stand corrected.

beacon_node/beacon_processor/src/lib.rs
michaelsproul: you're correct!

dapplion: ups, good catch

dapplion:
> I feel like this could easily regress with the beacon processor refactors. Wanna add a test?

To not conflict with the refactor, what API should the test use? i.e. 

```rust
let manager = spawn_manager(event_rx, work_journal_tx, BeaconProcessorQueueLengths::empty());
event_rx.send(Work::GossipAttestation {});
// assert work_journal_tx recv GossipAttestation work
```

michaelsproul:
I think it doesn't matter too much if the test conflicts, as long as it exists and can be translated over.

I would even just test initialising the Lengths struct in isolation so that it's minimally entangled with the rest of the beacon processor

michaelsproul (approved):
I pushed a test. This is ready to go.

michaelsproul:
@mergify queue

PR #6468, opened 2024-10-06
Use snap install yq in local_testnet

haxxpop:
snap is better than apt since it's recommended in yq doc

## Issue Addressed

I installed yq using apt and my version was v0.0.0. I don't know why.

## Proposed Changes

Use snap to install yq instead of apt since it's a recommended method in yq doc. See https://github.com/mikefarah/yq/?tab=readme-ov-file#linux-via-snap

## Additional Info

### DIFF
--- a/scripts/local_testnet/README.md
+++ b/scripts/local_testnet/README.md
@@ -9,7 +9,7 @@ This setup can be useful for testing and development.
 
 1. Install [Kurtosis](https://docs.kurtosis.com/install/). Verify that Kurtosis has been successfully installed by running `kurtosis version` which should display the version.
 
-1. Install [yq](https://github.com/mikefarah/yq). If you are on Ubuntu, you can install `yq` by running `sudo apt install yq -y`.
+1. Install [yq](https://github.com/mikefarah/yq). If you are on Ubuntu, you can install `yq` by running `snap install yq`.
 
 ## Starting the testnet
 


chong-he:
Looks good!

Which Linux distros are you on? Maybe `apt` only works for Ubuntu/Debian distros 

When I wrote the doc for local testnet I somehow ended up using `apt` as I was using Ubuntu, but `snap` is better as you mention per the official repo doc. Thanks!

chong-he:
@ppopth Could you switch the target branch to unstable? You also want to update your own branch to the latest unstable

haxxpop:
> Maybe `apt` only works for Ubuntu/Debian distros

I'm on ubuntu 24.04

scripts/local_testnet/README.md
```diff
@@ -9,7 +9,7 @@ This setup can be useful for testing and development.
 
 1. Install [Kurtosis](https://docs.kurtosis.com/install/). Verify that Kurtosis has been successfully installed by running `kurtosis version` which should display the version.
 
-1. Install [yq](https://github.com/mikefarah/yq). If you are on Ubuntu, you can install `yq` by running `sudo apt install yq -y`.
+1. Install [yq](https://github.com/mikefarah/yq). If you are on Ubuntu, you can install `yq` by running `snap install yq`.
```
dapplion, line 12: Why not suggest to use `snap` as fallback?

haxxpop: Because there is no reason to use `apt` at all and `snap` is the recommended one in the yq official doc

michaelsproul:
yeah I get a weird error on Ubuntu with the version of `yq` installed from apt:

```
$ ./start_local_testnet.sh
usage: yq [-h] [--yaml-output] [--yaml-roundtrip]
          [--yaml-output-grammar-version {1.1,1.2}] [--width WIDTH]
          [--indentless-lists] [--in-place] [--version]
          [jq_filter] [files ...]
yq: error: argument files: can't open '.participants[0].cl_image': [Errno 2] No such file or directory: '.participants[0].cl_image'
```

michaelsproul (approved):
Confirmed that using `snap` works better for me on Ubuntu. Merging

michaelsproul:
@mergify queue

PR #6469, opened 2024-10-06
Match block and blobs after validating execution

dapplion:
## Issue Addressed

After PeerDAS we request block and data to different peers. Consider the following scenario when doing a by range sync:
- Block at slot N contains no data
- Peer serving the block at slot N forges the block adding KZG commitments
- The peers serving data columns do not return any data columns for slot N

Current unstable will attempt to match block and columns and consider the entire RpcBlock invalid. A similar scenario can occur where the block peer is honest and the column peer is dishonest.

We need to make the code more aware of **who** is at fault, and specifically do not attempt to match columns before validating the block.

Part of

**Note: this PR touches code on the mainnet path, not only PeerDAS**

## Proposed Changes

Relax RpcBlock construction with `new_unchecked` and perform matching in `verify_kzg_for_rpc_block` after the block is execution validated.

This PR does not include a fix for peer scoring. This will be tackled latter as part of https://github.com/sigp/lighthouse/issues/6258

### DIFF
--- a/beacon_node/beacon_chain/src/block_verification_types.rs
+++ b/beacon_node/beacon_chain/src/block_verification_types.rs
@@ -1,6 +1,6 @@
 use crate::data_availability_checker::AvailabilityCheckError;
 pub use crate::data_availability_checker::{AvailableBlock, MaybeAvailableBlock};
-use crate::data_column_verification::{CustodyDataColumn, CustodyDataColumnList};
+use crate::data_column_verification::CustodyDataColumnList;
 use crate::eth1_finalization_cache::Eth1FinalizationData;
 use crate::{get_block_root, PayloadVerificationOutcome};
 use derivative::Derivative;
@@ -10,8 +10,8 @@ use std::fmt::{Debug, Formatter};
 use std::sync::Arc;
 use types::blob_sidecar::{BlobIdentifier, FixedBlobSidecarList};
 use types::{
-    BeaconBlockRef, BeaconState, BlindedPayload, BlobSidecarList, ChainSpec, Epoch, EthSpec,
-    Hash256, RuntimeVariableList, SignedBeaconBlock, SignedBeaconBlockHeader, Slot,
+    BeaconBlockRef, BeaconState, BlindedPayload, BlobSidecarList, ColumnIndex, Epoch, EthSpec,
+    Hash256, SignedBeaconBlock, SignedBeaconBlockHeader, Slot,
 };
 
 /// A block that has been received over RPC. It has 2 internal variants:
@@ -49,7 +49,7 @@ impl<E: EthSpec> RpcBlock<E> {
         match &self.block {
             RpcBlockInner::Block(block) => block,
             RpcBlockInner::BlockAndBlobs(block, _) => block,
-            RpcBlockInner::BlockAndCustodyColumns(block, _) => block,
+            RpcBlockInner::BlockAndCustodyColumns(block, _, _) => block,
         }
     }
 
@@ -57,7 +57,7 @@ impl<E: EthSpec> RpcBlock<E> {
         match &self.block {
             RpcBlockInner::Block(block) => block.clone(),
             RpcBlockInner::BlockAndBlobs(block, _) => block.clone(),
-            RpcBlockInner::BlockAndCustodyColumns(block, _) => block.clone(),
+            RpcBlockInner::BlockAndCustodyColumns(block, _, _) => block.clone(),
         }
     }
 
@@ -65,7 +65,7 @@ impl<E: EthSpec> RpcBlock<E> {
         match &self.block {
             RpcBlockInner::Block(_) => None,
             RpcBlockInner::BlockAndBlobs(_, blobs) => Some(blobs),
-            RpcBlockInner::BlockAndCustodyColumns(_, _) => None,
+            RpcBlockInner::BlockAndCustodyColumns(_, _, _) => None,
         }
     }
 
@@ -73,7 +73,7 @@ impl<E: EthSpec> RpcBlock<E> {
         match &self.block {
             RpcBlockInner::Block(_) => None,
             RpcBlockInner::BlockAndBlobs(_, _) => None,
-            RpcBlockInner::BlockAndCustodyColumns(_, data_columns) => Some(data_columns),
+            RpcBlockInner::BlockAndCustodyColumns(_, data_columns, _) => Some(data_columns),
         }
     }
 }
@@ -91,7 +91,11 @@ enum RpcBlockInner<E: EthSpec> {
     BlockAndBlobs(Arc<SignedBeaconBlock<E>>, BlobSidecarList<E>),
     /// This variant is used with parent lookups and by-range responses. It should have all
     /// requested data columns, all block roots matching for this block.
-    BlockAndCustodyColumns(Arc<SignedBeaconBlock<E>>, CustodyDataColumnList<E>),
+    BlockAndCustodyColumns(
+        Arc<SignedBeaconBlock<E>>,
+        CustodyDataColumnList<E>,
+        Vec<ColumnIndex>,
+    ),
 }
 
 impl<E: EthSpec> RpcBlock<E> {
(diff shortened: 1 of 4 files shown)


beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -85,60 +86,41 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
...
 
-        Ok(responses)
+        rpc_blocks
     }
 
     fn into_responses_with_custody_columns(
         self,
-        expects_custody_columns: Vec<ColumnIndex>,
         spec: &ChainSpec,
+        expeced_custody_columns: Vec<ColumnIndex>,
```
pawanjay176, line 122:
```suggestion
        expected_custody_columns: Vec<ColumnIndex>,
```

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -85,60 +86,41 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
     }
 
     pub fn into_responses(self, spec: &ChainSpec) -> Result<Vec<RpcBlock<E>>, String> {
-        if let Some(expects_custody_columns) = self.expects_custody_columns.clone() {
-            self.into_responses_with_custody_columns(expects_custody_columns, spec)
+        if let Some(expected_column_indices) = self.expects_custody_columns.as_ref() {
```
pawanjay176: This field should probably be renamed to `expected_column_indices` to better reflect its type. Doesn't need to be part of this PR though

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -85,60 +86,41 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
...
-            responses.push(RpcBlock::new(None, block, Some(blobs)).map_err(|e| format!("{e:?}"))?)
+        for blob in blobs {
+            let block_root = blob.block_root();
+            blobs_by_block.entry(block_root).or_default().push(blob);
         }
 
-        // if accumulated sidecars is not empty, throw an error.
-        if blob_iter.next().is_some() {
-            return Err("Received sidecars that don't pair well".to_string());
+        // Now iterate all blocks ensuring that the block roots of each block and data column match,
```
pawanjay176:
```suggestion
        // Now iterate all blocks ensuring that the block roots of each block and blob match,
```

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -85,60 +86,41 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
...
+        // plus we have columns for our custody requirements
```
pawanjay176: this comment might be in the wrong place?

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -147,71 +129,46 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
...
                 .or_default()
-                .insert(index, column)
-                .is_some()
-            {
-                return Err(format!(
-                    "Repeated column block_root {block_root:?} index {index}"
-                ));
-            }
+                // Safe to convert to `CustodyDataColumn`: we have asserted that the index of
+                // this column is in the set of `expects_custody_columns` and with the expected
```
pawanjay176: Where have we done this assertion?

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -85,60 +86,41 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
...
-                };
-                if blob_opt.is_some() {
-                    return Err("Repeat blob index".to_string());
-                } else {
-                    *blob_opt = Some(blob);
-                }
-            }
-            let blobs = VariableList::from(blobs_buffer.into_iter().flatten().collect::<Vec<_>>());
-            responses.push(RpcBlock::new(None, block, Some(blobs)).map_err(|e| format!("{e:?}"))?)
+        for blob in blobs {
```
pawanjay176, line 102: Neat simplificaiton in the logic 👍

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -147,71 +129,46 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
...
-
-        // Assert that there are no columns left for other blocks
-        if !data_columns_by_block.is_empty() {
-            let remaining_roots = data_columns_by_block.keys().collect::<Vec<_>>();
-            return Err(format!("Not all columns consumed: {remaining_roots:?}"));
+            let columns = match data_columns_by_block.remove(&block_root) {
+                Some(columns) => Some((
+                    RuntimeVariableList::new(columns, spec.number_of_columns)
+                        .map_err(|e| format!("{:?}", e))?,
+                    expeced_custody_columns.clone(),
```
pawanjay176, line 166:
```suggestion
                    expected_custody_columns.clone(),
```

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -311,8 +311,13 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
         let (block_root, block, blobs, data_columns) = block.deconstruct();
         if self.blobs_required_for_block(&block) {
             return if let Some(blob_list) = blobs.as_ref() {
+                if block.num_expected_blobs() != blob_list.len() {
```
pawanjay176, line 329: Should we also check if the block and blob commitments match here? 
Same check in `RpcBlock::new`

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -85,60 +86,41 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
...
-        if blob_iter.next().is_some() {
-            return Err("Received sidecars that don't pair well".to_string());
+        // Now iterate all blocks ensuring that the block roots of each block and data column match,
+        // plus we have columns for our custody requirements
+        let mut rpc_blocks = Vec::with_capacity(blocks.len());
+
+        for block in blocks {
+            let block_root = get_block_root(&block);
+            let blobs = blobs_by_block.remove(&block_root).map(VariableList::from);
+            rpc_blocks.push(RpcBlock::new_unchecked(block_root, block, blobs, None));
```
pawanjay176, line 113: I think using the unchecked variant here might result in an inconsistent state.

Consider the example you used, the peer/peers have sent us a valid block for a block_root `x` which contains 1 `kzg_commitment`. The peer sends us 2 blobs for this block root which do not match with the `kzg_commitment` in the block. The malicious blobs pass the kzg verification in isolation, they just doesn't match with the block commitment.

Now the unchecked `RpcBlock` is sent for processing.  The block goes through the validation pipeline
I couldn't find anywhere we check that the block and blob commitments match up. Correct me if i'm wrong here though.

We should potentially do this check when we do`RpcBlock::deconstruct`

beacon_node/beacon_chain/src/data_availability_checker.rs
pawanjay176: We could also do this in `RpcBlock::deconstruct`

beacon_node/beacon_chain/src/block_verification_types.rs
```diff
@@ -193,34 +170,60 @@ impl<E: EthSpec> RpcBlock<E> {
         Self::new(Some(block_root), block, blobs)
     }
 
+    /// Create an RpcBlock from an untrusted block, without checking that the blobs or columns are
```
pawanjay176, line 173: I think the docs for RpcBlock and RpcBlockInner need to be updated to reflect the fact that they could be constructed without checking for consistency

pawanjay176 (changes requested):

beacon_node/network/src/sync/block_sidecar_coupling.rs
dapplion: In sync we only request custody columns

beacon_node/beacon_chain/src/data_availability_checker.rs
dapplion: if we check the inclusion proof, we check that the commitments match the block. I've added a check in the da_checker. It's unnecessary in most paths but would be best to skip it only as an optimization by adding a property to `RpcBlock::valid_inclusion_proof`, and only if true, skip the check

beacon_node/network/src/sync/block_sidecar_coupling.rs
dapplion: > I couldn't find anywhere we check that the block and blob commitments match up

In lookup sync when doing a blobs request we check the inclusion proof, here

https://github.com/sigp/lighthouse/blob/40d3423193e6f97f44c42eb3673af607627c8f7f/beacon_node/network/src/sync/network_context/requests/blobs_by_root.rs#L53

However we don't for range sync. I've added the inclusion proof check in the da_checker. We can remove it after https://github.com/sigp/lighthouse/pull/6497

jimmygchen:
@dapplion I believe this PR is now outdated and superceded by some of the changes in #7352. Feel free to reopen it if you feel otherwise.

PR #6473, opened 2024-10-08
Better assert message in lookup sampling test

dapplion:
## Issue Addressed

Debugging a sampling failed test in https://github.com/sigp/lighthouse/pull/6398 it is hard to identify which specific column and request had the wrong state since there's an `assert!` statement without more metadata.

## Proposed Changes

Log which specific sampling request and column has an unexpected state

### DIFF
--- a/beacon_node/network/src/sync/block_lookups/tests.rs
+++ b/beacon_node/network/src/sync/block_lookups/tests.rs
@@ -1303,14 +1303,30 @@ impl TestRig {
         });
     }
 
-    fn assert_sampling_request_status(
-        &self,
-        block_root: Hash256,
-        ongoing: &Vec<ColumnIndex>,
-        no_peers: &Vec<ColumnIndex>,
-    ) {
-        self.sync_manager
-            .assert_sampling_request_status(block_root, ongoing, no_peers)
+    fn assert_sampling_request_ongoing(&self, block_root: Hash256, indices: &Vec<ColumnIndex>) {
+        for index in indices {
+            let status = self
+                .sync_manager
+                .get_sampling_request_status(block_root, index)
+                .unwrap_or_else(|| panic!("No request state for {index}"));
+            assert_eq!(
+                status, "Sampling",
+                "expected {block_root} {index} request to be ongoing"
+            );
+        }
+    }
+
+    fn assert_sampling_request_nopeers(&self, block_root: Hash256, indices: &Vec<ColumnIndex>) {
+        for index in indices {
+            let status = self
+                .sync_manager
+                .get_sampling_request_status(block_root, index)
+                .unwrap_or_else(|| panic!("No request state for {index}"));
+            assert_eq!(
+                status, "NoPeers",
+                "expected {block_root} {index} request to be nopeers"
+            );
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


beacon_node/network/src/sync/block_lookups/tests.rs
```diff
@@ -1303,14 +1303,30 @@ impl TestRig {
...
-        self.sync_manager
-            .assert_sampling_request_status(block_root, ongoing, no_peers)
+    fn assert_sampling_request_ongoing(&self, block_root: Hash256, indices: &Vec<ColumnIndex>) {
+        for index in indices {
+            let status = self
+                .sync_manager
+                .get_sampling_request_status(block_root, index)
+                .unwrap_or_else(|| panic!("No request state for {index}"));
+            assert_eq!(
+                status, "Sampling",
```
jimmygchen: Minor - i think it would be a bit more maintainable just matching against enum rather than comparing against string values?

jimmygchen (changes requested):
Looks good, just a minor comment above!

beacon_node/network/src/sync/block_lookups/tests.rs
dapplion: Yeah, but I'll have to leak this internal enum all the way up to the sync manager. On the other hand if you change the sampling status enum source, the test will break. This not un-usual behaviour when developing: you change source, test breaks, fix test, done. So yes it's not the prettiest but doesn't sound too bad IMO

jimmygchen: Ah ok I see why you do it this way. Usually this _should_ not happen because integration tests aren't suppose to make assertions against internals, and assertions against internals are usually done at unit test level. Modern IDEs  pick up changes on enum variants easily and make refactoring very straightforward - using a string means these needs to be searched and fixed manually. I have no problem doing this _sometimes_, but I worry If this approach get used more widely in future tests, then maintenance would be much harder. 

But yeah, I agree this is not _too_ bad if it's used sparingly, but this shouldn't be our _usual_ approach. Overall I think this PR is an improvement so I'm happy to merge it.

jimmygchen: I think perhaps exposing the enum only in tests would be a good middle ground?
```
#[cfg(test)]
pub use crate::sync::peer_sampling::request::ActiveColumnSampleRequest;
```

dapplion: Thanks for the suggestion! Added in 08f5efbcb3f6d860285f192201ba0bc64387afb6

beacon_node/network/src/sync/peer_sampling.rs
```diff
@@ -584,8 +577,9 @@ mod request {
         peers_dont_have: HashSet<PeerId>,
     }
 
-    #[derive(Debug, Clone)]
-    enum Status {
+    // Exposed only for testing assertions in lookup tests
+    #[derive(Debug, Clone, IntoStaticStr)]
```
jimmygchen: `IntoStaticStr` can be removed now I think

jimmygchen (approved):
LGTM!

michaelsproul:
@mergify queue

PR #6475, opened 2024-10-09
New historic state cache

michaelsproul:
## Proposed Changes

Attempt to unify the new HDiffBuffer cache with the old "historic state cache" by creating a cache that holds either or both kinds of states, and lazily converts between the formats as required.

## Additional Info

Performance testing on mainnet with `--hierarchy-exponents 5,7,11` shows much improved performance for sequential queries. Most queries after the first are served in 400-700ms.

### DIFF
--- a/beacon_node/beacon_chain/src/metrics.rs
+++ b/beacon_node/beacon_chain/src/metrics.rs
@@ -1992,6 +1992,8 @@ pub fn scrape_for_metrics<T: BeaconChainTypes>(beacon_chain: &BeaconChain<T>) {
         .canonical_head
         .fork_choice_read_lock()
         .scrape_for_metrics();
+
+    beacon_chain.store.register_metrics();
 }
 
 /// Scrape the given `state` assuming it's the head state, updating the `DEFAULT_REGISTRY`.
--- a/beacon_node/store/src/hdiff.rs
+++ b/beacon_node/store/src/hdiff.rs
@@ -78,8 +78,8 @@ pub enum StorageStrategy {
 /// Hierarchical diff output and working buffer.
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct HDiffBuffer {
-    state: Vec<u8>,
-    balances: Vec<u64>,
+    pub state: Vec<u8>,
+    pub balances: Vec<u64>,
 }
 
 /// Hierarchical state diff.
@@ -124,12 +124,12 @@ impl HDiffBuffer {
         HDiffBuffer { state, balances }
     }
 
-    pub fn into_state<E: EthSpec>(self, spec: &ChainSpec) -> Result<BeaconState<E>, Error> {
+    pub fn as_state<E: EthSpec>(&self, spec: &ChainSpec) -> Result<BeaconState<E>, Error> {
         let _t = metrics::start_timer(&metrics::STORE_BEACON_HDIFF_BUFFER_INTO_STATE_TIME);
         let mut state =
             BeaconState::from_ssz_bytes(&self.state, spec).map_err(Error::InvalidSszState)?;
-        *state.balances_mut() =
-            List::new(self.balances).map_err(|_| Error::InvalidBalancesLength)?;
+        *state.balances_mut() = List::try_from_iter(self.balances.iter().copied())
+            .map_err(|_| Error::InvalidBalancesLength)?;
         Ok(state)
     }
 
--- a/beacon_node/store/src/hdiff.rs
+++ b/beacon_node/store/src/hdiff.rs
@@ -79,8 +79,8 @@ pub enum StorageStrategy {
 /// Hierarchical diff output and working buffer.
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct HDiffBuffer {
-    pub state: Vec<u8>,
-    pub balances: Vec<u64>,
+    state: Vec<u8>,
+    balances: Vec<u64>,
 }
 
 /// Hierarchical state diff.
--- a/beacon_node/store/src/hdiff.rs
+++ b/beacon_node/store/src/hdiff.rs
@@ -5,6 +5,7 @@ use serde::{Deserialize, Serialize};
 use ssz::{Decode, Encode};
 use ssz_derive::{Decode, Encode};
 use std::io::{Read, Write};
+use std::ops::RangeInclusive;
 use std::str::FromStr;
 use types::{BeaconState, ChainSpec, EthSpec, List, Slot};
 use zstd::{Decoder, Encoder};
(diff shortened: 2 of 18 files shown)


michaelsproul:
This is working quite well for the sequential query case. The tricky part is managing when to build caches. Previously we were re-building them for every state that required replay, which was obviously sub-optimal. Now we ensure that any state in the historic state cache has all caches built, which makes replay faster at the expense of slowing down the loading of snapshot/diff states.

On the test machine the loads for snapshot/diff states are now around 6-7s (ouch), but each subsequent state is more like 0.5-1s (on average 700ms).

I thought about moving the cache build from when the snapshot/diff state is loaded to the next slot, which would keep the total amortised time the same, but reduce the worst-case time for the diff state by several seconds (maybe from 6-7s down to 3s). It seems cache building only takes 1.8-2.5s in the worst case, so there's still some time going missing for the snapshot state.

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -1741,37 +1756,123 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
+        }
+
+        metrics::inc_counter(&metrics::STORE_BEACON_HISTORIC_STATE_CACHE_MISS);
+
+        // Load using the diff hierarchy. For states that require replay we recurse into this
+        // function so that we can try to get their pre-state *as a state* rather than an hdiff
+        // buffer.
+        match self.hierarchy.storage_strategy(slot)? {
+            StorageStrategy::Snapshot | StorageStrategy::DiffFrom(_) => {
+                let _t = metrics::start_timer(&metrics::STORE_BEACON_HDIFF_BUFFER_LOAD_TIME);
```
michaelsproul: We should drop this after loading the hdiff buffer rather than at the end of the function. At the moment this is distorting the metrics because we're including the cache build time in the buffer load metric.

michaelsproul:
Block replay is quick now with built caches, although it's a bit slower at epoch boundaries:

> Oct 11 06:27:33.922 DEBG Replayed blocks, replay_time_ms: 10, target_slot: 8999423, service: freezer_db, module: store::hot_cold_store:1865
Oct 11 06:27:35.536 DEBG Replayed blocks, replay_time_ms: 836, target_slot: 8999424, service: freezer_db, module: store::hot_cold_store:1865

In this commit (https://github.com/sigp/lighthouse/pull/6475/commits/52a9066361c8362994aa0d8b5742d351bb5db5a3) I made it so that we replay 1 block in preference to applying a diff for states that would normally be constructed by diffing. This may be the wrong choice, as applying a diff should be ~150ms, and converting HDiffBuffer -> state should be only ~400ms or so, for 550ms total. Although this doesn't mesh with the benchmark results from prior to this commit, where diff states were taking around 4-6s! Maybe cloning the several hundred MB of hdiff buffer is taking more time than we think (could be good to measure this). Another advantage of favouring replay where possible (i.e. for sequential access) is that it minimises memory fragmentation.

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -1794,34 +1895,46 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
+                    "Built caches for state";
+                    "target_slot" => slot,
+                    "build_time_ms" => t.elapsed().as_millis()
+                );
+
+                self.historic_state_cache
+                    .lock()
+                    .put_both(slot, state, buffer.clone());
+
+                let load_time_ms = t.elapsed().as_millis();
```
michaelsproul, line 1909: This time is wrong currently because it doesn't include the hdiff buffer load time

beacon_node/beacon_chain/src/metrics.rs
```diff
@@ -1992,6 +1992,8 @@ pub fn scrape_for_metrics<T: BeaconChainTypes>(beacon_chain: &BeaconChain<T>) {
         .canonical_head
         .fork_choice_read_lock()
         .scrape_for_metrics();
+
+    beacon_chain.store.register_metrics();
```
michaelsproul, line 1996: This was a regression from a looong time ago.

michaelsproul:
I think this is ready for review.

Some further optimisations are described here, but I think current performance is now very good.

- https://github.com/sigp/lighthouse/issues/6485

beacon_node/store/src/historic_state_cache.rs
```diff
@@ -0,0 +1,85 @@
...
+    pub fn get_state(
+        &mut self,
+        slot: Slot,
+        spec: &ChainSpec,
+    ) -> Result<Option<BeaconState<E>>, Error> {
+        if let Some(state) = self.states.get(&slot) {
+            Ok(Some(state.clone()))
+        } else if let Some(buffer) = self.hdiff_buffers.get(&slot) {
+            let state = buffer.as_state(spec)?;
+            self.states.put(slot, state.clone());
```
dapplion, line 60: We need to metric time this conversion

beacon_node/store/src/historic_state_cache.rs
```diff
@@ -0,0 +1,85 @@
...
+            states: LruCache::new(state_cache_size),
+        }
+    }
+
+    pub fn get_hdiff_buffer(&mut self, slot: Slot) -> Option<HDiffBuffer> {
+        if let Some(buffer_ref) = self.hdiff_buffers.get(&slot) {
+            let _timer = metrics::start_timer(&metrics::BEACON_HDIFF_BUFFER_CLONE_TIMES);
+            Some(buffer_ref.clone())
+        } else if let Some(state) = self.states.get(&slot) {
+            let buffer = HDiffBuffer::from_state(state.clone());
```
dapplion, line 40: metric time this

beacon_node/store/src/hdiff.rs
```diff
@@ -78,8 +79,8 @@ pub enum StorageStrategy {
 /// Hierarchical diff output and working buffer.
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct HDiffBuffer {
-    state: Vec<u8>,
-    balances: Vec<u64>,
+    pub state: Vec<u8>,
+    pub balances: Vec<u64>,
```
dapplion: not necessary
```suggestion
    state: Vec<u8>,
    balances: Vec<u64>,
```

beacon_node/store/src/historic_state_cache.rs
```diff
@@ -0,0 +1,85 @@
+use crate::hdiff::{Error, HDiffBuffer};
+use crate::metrics;
+use lru::LruCache;
+use std::num::NonZeroUsize;
+use types::{BeaconState, ChainSpec, EthSpec, Slot};
+
+#[derive(Debug)]
```
dapplion, line 14:
```suggestion
/// Holds a combination of finalized states in two formats:
/// - `hdiff_buffers`: Format close to an SSZ serialized state for rapid application of diffs on top
///   of it
/// - `states`: Deserialized states for direct use or for rapid application of blocks (replay)
///
/// An example use: when requesting state data for consecutive slots, this cache allows the node to
/// apply diffs once on the first request, and latter just apply blocks one at a time.
#[derive(Debug)]
```

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -1741,37 +1746,118 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
+            if cached_state.slot() == slot {
+                metrics::inc_counter(&metrics::STORE_BEACON_HISTORIC_STATE_CACHE_HIT);
+                return Ok(cached_state);
+            }
+            metrics::inc_counter(&metrics::STORE_BEACON_HISTORIC_STATE_CACHE_MISS);
+
+            return self.load_cold_state_by_slot_using_replay(cached_state, slot);
+        }
+
+        metrics::inc_counter(&metrics::STORE_BEACON_HISTORIC_STATE_CACHE_MISS);
```
dapplion, line 1782: These two misses are different. Technically we got a hit as the cache is allowing us to skip _some_ work. Maybe we can have a metric `STORE_BEACON_HISTORIC_STATE_CACHE_RESULT` with a label:
- `hit_exact`
- `hit_replay`
- `miss`

beacon_node/store/src/historic_state_cache.rs
michaelsproul: We already have, there's a metric inside `.as_state`:

https://github.com/sigp/lighthouse/blob/3b4fa0b8c8a50eeaf4db8bd720b5a4137712511c/beacon_node/store/src/hdiff.rs#L128-L135

beacon_node/store/src/historic_state_cache.rs
michaelsproul: as above

beacon_node/store/src/hot_cold_store.rs
michaelsproul: I need to think about this some more. Will merge as-is for now

PR #6481, opened 2024-10-11
Optimization to Redb slasher

eserilev:
Upgraded to redb 2.1.4 which introduces optimizations to `first()` and `last()`

https://github.com/cberner/redb/releases/tag/v2.1.4

Updated the redb slasher to use those methods instead of using `iter().next()` and `iter().next_back()`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6797,9 +6797,9 @@ dependencies = [
 
 [[package]]
 name = "redb"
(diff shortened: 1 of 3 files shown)


jimmygchen (approved):
Straight forward change, looks good!

> almost 2x faster

👀

michaelsproul:
@mergify queue

PR #6483, opened 2024-10-13
Simplify hashing in shuffling

michaelsproul:
## Proposed Changes

Remove some bounds checks from `swap_or_not_shuffle` by using the infallible method `Hash256::from` rather than `Hash256::from_slice` which does runtime checks and can panic.

This didn't seem to make a difference to the benchmarks, but I think it's a nice cleanup regardless.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -95,7 +95,7 @@ resolver = "2"
 edition = "2021"
 
 [workspace.dependencies]
-alloy-primitives = "0.8"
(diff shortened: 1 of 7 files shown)


michaelsproul:
The `swap_or_not_shuffle` benchmarks didn't actually compile, so I've fixed the dep issues and made it so that `make lint` also checks benchmarks.

jimmygchen (approved):
Nice! LGTM

michaelsproul:
@mergify queue

PR #6486, opened 2024-10-14
Update to latest discovery

AgeManning:
This update brings a few changes. 

Specifically: 
- More accessibility for NAT'd peers to find and discover peers
- Improved enr-auto-updates for dual-stack nodes


pawanjay176 (approved):
This looks good, but I can't test it as I don't have an ipv6 connection at home.
Have you tested auto-enr with ipv6?

jxs (approved):

AgeManning:
Yeah. I've been testing with ipv6 using discv5-cli. 

There are probably some other changes we might need to introduce for ipv6 however. I've not quite got to them yet.

AgeManning:
I've made a new commit. Moves to a whole new discovery. 

I think this current version should be fine to support ipv6 and potentially dual-stack by default

beacon_node/lighthouse_network/src/config.rs
```diff
@@ -305,12 +305,12 @@ impl Default for Config {
...
             .enable_packet_filter()
             .session_cache_capacity(5000)
-            .request_timeout(Duration::from_secs(1))
+            .request_timeout(Duration::from_secs(2))
             .query_peer_timeout(Duration::from_secs(2))
             .query_timeout(Duration::from_secs(30))
             .request_retries(1)
             .enr_peer_update_min(10)
-            .query_parallelism(5)
+            .query_parallelism(8)
```
pawanjay176, line 313: Why is this bumped?

michaelsproul:
I've pushed a commit to minimise the update to `Cargo.lock`, seeing as there's something nasty lurking in the dep tree that we haven't found yet. See:

- https://github.com/sigp/lighthouse/pull/6554

beacon_node/lighthouse_network/src/config.rs
AgeManning: Some simulations or experiments were saying discovery can run faster just by increasing the parallelism. 

I was looking through the configs, and while I was there, figured I'd increase. Happy to leave it as is, if its a concern. 

Its just that I've been meaning to mess with it for a while, so figure in the update, might as well do it here.

michaelsproul:
@mergify queue

PR #6488, opened 2024-10-15
`SingleAttestation` implementation

eserilev:
## Issue Addressed

Closes https://github.com/sigp/lighthouse/issues/6380

## Proposed Changes

Spec: ethereum/consensus-specs#3900

Convert single attestation as early as possible to reduce code diff

Conga-lined off:

- https://github.com/sigp/lighthouse/pull/6731


eserilev:
bell curve
@dapplion

eserilev:
I've ran some kurtosis tests interoping against prysm and we are attesting correctly

EDIT: we are NOT interoping correctly at the moment. I am working on debugging the issue

eserilev:
Most of the added diff is tests I included for single attestations

michaelsproul:
I've just re-targeted this at the `electra-alpha10` branch so we can get a conga-line going.

beacon_node/beacon_chain/src/test_utils.rs
```diff
@@ -1024,6 +1030,79 @@ where
...
+        let single_attestation = attestation.to_single_attestation(Some(committee.clone()))?;
+
+        let attestation: Attestation<E> = single_attestation.to_attestation(Some(committee))?;
+
+        assert_eq!(
+            single_attestation.committee_index,
+            attestation.committee_index().unwrap() as usize
+        );
+        assert_eq!(single_attestation.attester_index, validator_index);
+        // assert_eq!(single_attestation.attester_index, attestation.attester_index());
```
michaelsproul: Assuming we should just delete this assert

```suggestion
```

beacon_node/http_api/src/publish_attestations.rs
```diff
@@ -129,6 +161,51 @@ fn verify_and_publish_attestation<T: BeaconChainTypes>(
...
+
+                let attestation = single_attestation.to_attestation::<T::EthSpec>(committee)?;
+
+                Ok(attestation)
+            },
+        );
+
+        if let Ok(attestation) = attestation {
+            attestations.push(attestation);
+        }
```
michaelsproul: We should probably at least log a warn here, and maybe even propagate the error up into the `failures` Vec so that it makes it into a non-200 response.

Perhaps `publish_attestations` could take a vec of `Either<Attestation<E>, PublishAttestationResult>` as input and the `PublishAttestationResult`s could just get plumbed through into the output without further processing. This way we could inject these early failures in order

validator_client/validator_services/src/attestation_service.rs
```diff
@@ -457,8 +457,16 @@ impl<T: SlotClock + 'static, E: EthSpec> AttestationService<T, E> {
...
                 );
                 if fork_name.electra_enabled() {
+                    let single_attestations = attestations
+                        .iter()
+                        .zip(validator_indices)
+                        .filter_map(|(a, i)| {
+                            a.to_single_attestation_with_attester_index(*i as usize)
+                                .ok()
+                        })
+                        .collect::<Vec<_>>();
```
michaelsproul: Could log an error if this fails. We essentially never expect it to fail, right?

michaelsproul (commented):
Did a quick re-review of about half of this. Will come back to it ASAP tomorrow

validator_client/validator_services/src/attestation_service.rs
michaelsproul: This comment is on more lines than I intended, I mean we should log if `to_single_attestation_with_attester_index` fails

beacon_node/http_api/src/publish_attestations.rs
eserilev: I've updated `publish-attestations` to take a `Vec<Either<Attestation<E>, PublishAttestationResult>>` as input here [2d61b65](https://github.com/sigp/lighthouse/pull/6488/commits/2d61b65ebdbde0433dae03b62852848211486d4e)

In an error case the `Err` should get plumbed through and logged in a similar manner to other `PublishAttestation` errors

validator_client/validator_services/src/attestation_service.rs
eserilev: Yeah I cant think of a situation where this would fail since these are guaranteed to be unaggregated attestations

I've added some logging and additional comments here
[b6d80eb](https://github.com/sigp/lighthouse/pull/6488/commits/b6d80eb9085ee9aa6a7d50769728408118d5a4fc)

consensus/types/src/subnet_id.rs
```diff
@@ -57,6 +58,21 @@ impl SubnetId {
         )
     }
 
+    /// Compute the subnet for an attestation where each slot in the
+    /// attestation epoch contains `committee_count_per_slot` committees.
+    pub fn compute_subnet_for_single_attestation<E: EthSpec>(
+        attestation: SingleAttestation,
```
pawanjay176: Can pass a reference to the SingleAttestation here I think

validator_client/validator_services/src/attestation_service.rs
```diff
@@ -457,8 +457,31 @@ impl<T: SlotClock + 'static, E: EthSpec> AttestationService<T, E> {
...
                 );
                 if fork_name.electra_enabled() {
+                    let single_attestations = attestations
+                        .iter()
+                        .zip(validator_indices)
+                        .filter_map(|(a, i)| {
+                            match a.to_single_attestation_with_attester_index(*i as usize) {
+                                Ok(a) => Some(a),
+                                Err(e) => {
+                                    // I can't think of a reason why we'd reach this error case. The
```
pawanjay176: Another potential reason for the error would be an `IncorrectStateVariant` where the BN and VC are out of sync. BN is sending Base attestations when VC is expecting electra attestations

beacon_node/lighthouse_network/src/types/pubsub.rs
```diff
@@ -30,6 +29,8 @@ pub enum PubsubMessage<E: EthSpec> {
     AggregateAndProofAttestation(Box<SignedAggregateAndProof<E>>),
     /// Gossipsub message providing notification of a raw un-aggregated attestation with its shard id.
     Attestation(Box<(SubnetId, Attestation<E>)>),
+    /// Gossipsub message providing notification of a `SingleAttestation`` with its shard id.
```
pawanjay176: Can maybe change the docs here and above to say subnet id instead?

beacon_node/http_api/tests/fork_tests.rs
```diff
@@ -155,10 +155,6 @@ async fn attestations_across_fork_with_skip_slots() {
         .post_beacon_pool_attestations_v1(&unaggregated_attestations)
         .await
         .unwrap();
-    client
```
pawanjay176, line 158: Is there a reason why we don't need this anymore?

beacon_node/http_api/src/publish_attestations.rs
```diff
@@ -73,24 +75,57 @@ enum PublishAttestationResult {
...
+                .send(NetworkMessage::Publish {
+                    messages: vec![PubsubMessage::Attestation(Box::new((
+                        attestation.subnet_id(),
+                        attestation.attestation().clone_as_attestation(),
+                    )))],
+                })
+                .map_err(|_| Error::Publication)?;
+        }
+        types::AttestationRef::Electra(attn) => {
+            chain
```
pawanjay176: Can't we keep the `SingleAttestation` around prior to `convert_to_attestation` to prevent computing it all over again here?

beacon_node/network/src/network_beacon_processor/mod.rs
```diff
@@ -84,6 +84,50 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+    pub fn send_single_attestation(
+        self: &Arc<Self>,
+        message_id: MessageId,
+        peer_id: PeerId,
+        single_attestation: SingleAttestation,
+        subnet_id: SubnetId,
+        should_import: bool,
+        seen_timestamp: Duration,
+    ) -> Result<(), Error<T::EthSpec>> {
+        let result = self.chain.with_committee_cache(
```
pawanjay176, line 97: Just confirming that this doesn't give us full benefits of the spec change. From the spec PR

> placing the validator index directly in the attestation allows verifying the signature without computing a shuffling - this closes a loophole where clients either must drop attestations or risk being overwhelmed by shuffling computations during attestation verification

We are still computing the shufflings here so we aren't getting the DoS resistance from the PR. 
Is the plan to have a simpler refactor with fewer changes first to be devnet 5 ready and then work on the more annoying refactor?

beacon_node/http_api/tests/fork_tests.rs
eserilev: I'm testing the v2 endpoint in other places so I thought it'd be okay to remove here. If we were to include it here, it would require us to convert to single attestation before posting to the v2 endpoint

beacon_node/network/src/network_beacon_processor/mod.rs
eserilev: Right, I'm working on a separate PR to make full use of the `SingleAttestation` benefits. It's a deeper change with a much larger diff, so we've decided to punt that to later, potentially after the Electra fork

beacon_node/http_api/tests/fork_tests.rs
michaelsproul: I think that's fine

beacon_node/network/src/network_beacon_processor/mod.rs
pawanjay176: Makes sense, thanks!

beacon_node/http_api/src/publish_attestations.rs
michaelsproul: Yeah, agree. I think it's as easy as calling `attestation.single_attestation()` on the `VerifiedUnaggregatedAttestation`.

michaelsproul: Or yeah, just keeping the original `Either` and pattern-matching

beacon_node/network/src/network_beacon_processor/mod.rs
michaelsproul: This is a little bit of a regression even, because we are doing the possibly expensive `with_committee_cache` call outside of the beacon processor, where it can DoS us even more than if it happens _inside_ the beacon processor.

michaelsproul: Maybe rather than calling `send_unaggregated_attestation`, we could copy its logic and make sure that the `with_committee_cache` call happens inside `process_individual`/`process_batch`. To keep things simple the `process_batch` impl for `SingleAttestation` could just loop over the single attestations and convert

michaelsproul: Opened an issue for this:

- https://github.com/sigp/lighthouse/issues/6805

validator_client/validator_services/src/attestation_service.rs
michaelsproul: I've updated the error message in:

https://github.com/sigp/lighthouse/pull/6488/commits/e900bb850109ff81b449a1a34732f95b4f7380d6

beacon_node/http_api/src/publish_attestations.rs
michaelsproul: Optimised this here:

https://github.com/sigp/lighthouse/pull/6488/commits/8cc43c1a103db9959536b8b8806010f8be425668

consensus/types/src/attestation.rs
```diff
@@ -527,6 +605,70 @@ impl<E: EthSpec> ForkVersionDeserialize for Vec<Attestation<E>> {
...
+    pub signature: AggregateSignature,
+}
+
+impl SingleAttestation {
+    pub fn to_attestation<E: EthSpec>(
+        &self,
+        committee: Option<BeaconCommittee>,
+    ) -> Result<Attestation<E>, Error> {
+        let Some(committee) = committee else {
+            return Err(Error::InvalidCommittee);
```
dapplion:
```suggestion
            return Err(Error::MissingCommittee);
```

consensus/types/src/attestation.rs
```diff
@@ -527,6 +605,70 @@ impl<E: EthSpec> ForkVersionDeserialize for Vec<Attestation<E>> {
...
+        };
+
+        if committee.index != self.committee_index as u64 {
+            return Err(Error::InvalidCommittee);
```
dapplion:
```suggestion
            return Err(Error::UnexpectedCommittee(committee.index, self.committee_index));
```

consensus/types/src/attestation.rs
```diff
@@ -350,6 +370,64 @@ impl<E: EthSpec> AttestationElectra<E> {
...
+            committee_index: committee_index as usize,
+            attester_index,
+            data: self.data.clone(),
+            signature: self.signature.clone(),
+        })
+    }
+
+    pub fn to_single_attestation(
+        &self,
+        committee: Option<BeaconCommittee>,
```
dapplion: This function appears to be used once and committee is always Some
```suggestion
        committee: BeaconCommittee,
```

consensus/types/src/attestation.rs
dapplion: Weird API tho, is it cumbersome to make this function take `committee: BeaconCommittee`?

consensus/types/src/attestation.rs
```diff
@@ -527,6 +605,70 @@ impl<E: EthSpec> ForkVersionDeserialize for Vec<Attestation<E>> {
...
+            .committee
+            .iter()
+            .enumerate()
+            .find_map(|(i, &validator_index)| {
+                if self.attester_index == validator_index {
+                    return Some(i);
+                }
+                None
+            })
+            .ok_or(Error::InvalidAggregationBit)?;
```
dapplion:
```suggestion
            .ok_or(Error::AttesterNotInCommittee(self.attester_index))?;
```

michaelsproul (approved):
Approved modulo Lion's small tweaks

beacon_node/beacon_chain/src/attestation_verification.rs
```diff
@@ -317,12 +317,25 @@ pub struct VerifiedUnaggregatedAttestation<'a, T: BeaconChainTypes> {
...
     pub fn into_indexed_attestation(self) -> IndexedAttestation<T::EthSpec> {
         self.indexed_attestation
     }
+
+    pub fn single_attestation(&self) -> Option<SingleAttestation> {
+        let Some(committee_index) = self.attestation.committee_index() else {
+            return None;
+        };
+        Some(SingleAttestation {
+            committee_index: committee_index as usize,
```
dapplion, line 333:
```suggestion
            committee_index: self.attestation.committee_index()? as usize,
```

beacon_node/beacon_chain/src/attestation_verification.rs
```diff
@@ -317,12 +317,25 @@ pub struct VerifiedUnaggregatedAttestation<'a, T: BeaconChainTypes> {
...
     indexed_attestation: IndexedAttestation<T::EthSpec>,
     subnet_id: SubnetId,
+    validator_index: usize,
 }
 
 impl<T: BeaconChainTypes> VerifiedUnaggregatedAttestation<'_, T> {
     pub fn into_indexed_attestation(self) -> IndexedAttestation<T::EthSpec> {
         self.indexed_attestation
     }
+
```
dapplion, line 327:
```suggestion

        /// Returns Some if `self.attestation` has at least one committee bit set
```

beacon_node/beacon_chain/src/events.rs
```diff
@@ -90,6 +90,10 @@ impl<E: EthSpec> ServerSentEventHandler<E> {
                 .attestation_tx
                 .send(kind)
                 .map(|count| log_count("attestation", count)),
+            EventKind::SingleAttestation(_) => self
+                .attestation_tx
```
dapplion: This is not correct I believe. There's a new event name `single_attestation`, so you should have a new (tx,rx) and broadcast there. https://github.com/ethereum/beacon-APIs/blob/aa1be259e2d8a64f451dd304913e0310779f6f80/apis/eventstream/index.yaml#L70

dapplion: Yeah otherwise calls into the event stream with error with unknown topic

eserilev: I believe the `attestation_tx` is just for SSE which is for our own internal infra? We still emit a SingleAttestation event with the new `single_attestation` topic name, which I think is spec compliant. I could be wrong though

eserilev: ok maybe I am wrong, ill double check, thanks

dapplion: https://github.com/sigp/lighthouse/blob/0d90135047519f4c2ee586d50e560f7bb2ff9b10/common/eth2/src/types.rs#L1267 this enum must include a SingleAttestation variant

eserilev: Yep you were right, heres the fix 
54844b3

consensus/types/src/attestation.rs
eserilev: I've deleted this method as it was only used in one of the tests and as you said the API is a bit weird

consensus/types/src/attestation.rs
eserilev: I've updated the args for this method. Instead of passing in the full `BeaconCommittee` obj, I'm passing in the committee slice. Upstream from this method call I make a check for the existence of the specific committee for the `slot` and `committee_index` and return either a `Error::NoCommitteeForSlotAndIndex` or log a `warn` message if the committee isn't found

consensus/types/src/attestation.rs
eserilev: I've deleted this as its only used in one of the tests

consensus/types/src/attestation.rs
eserilev: I've added the new error variant

michaelsproul:
Looks like one of the SSE tests is failing (`get_events`)

eserilev:
I believe i've fixed the test failure but I think I need to add some test coverage for the new single_attestation event topic

eserilev:
i've added test coverage for the new `single_attestation` event topic. This PR should hopefully be good to go

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -2035,10 +2035,30 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             |v| {
                 // This method is called for API and gossip attestations, so this covers all unaggregated attestation events
                 if let Some(event_handler) = self.event_handler.as_ref() {
+                    if event_handler.has_single_attestation_subscribers() {
+                        let current_fork = self
+                            .spec
+                            .fork_name_at_slot::<T::EthSpec>(v.attestation().data().slot);
+                        if current_fork.electra_enabled() {
```
dapplion, line 2042: Is it necessary to check the fork now? Can we have single attestations before electra?

dapplion (approved):
PR is good to go!

pawanjay176 (approved):
LGTM as well

pawanjay176:
@mergify queue

PR #6489, opened 2024-10-15
Remove TTD flags and `safe-slots-to-import-*`

michaelsproul:
## Proposed Changes

Start deleting some merge logic and flags. We don't need to carry this stuff around forever.

- [x] Delete all traces of `SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY`. This concept was removed from the spec, so this was just dead code.
- [x] Delete CLI flags related to terminal total difficulty.

Ripping out the TTD logic more thoroughly proved to be more involved. A lot of our tests still depend on doing the merge transition, which is annoying. I have another branch where I started that work, but will submit it separately.


lighthouse/tests/beacon_node.rs
```diff
@@ -749,16 +749,15 @@ fn jwt_optional_flags() {
...
     CommandLineTest::new()
         .flag("terminal-total-difficulty-override", Some("1337424242"))
-        .run_with_zero_port()
-        .with_spec::<MainnetEthSpec, _>(|spec| {
-            assert_eq!(spec.terminal_total_difficulty, Uint256::from(1337424242))
-        });
+        .run_with_zero_port();
 }
+// DEPRECATED but should not crash.
+// DEPRECATED. This flag is deprecated but should not cause a crash.
```
macladson:
```suggestion
// DEPRECATED. This flag is deprecated but should not cause a crash.
```

consensus/types/src/chain_spec.rs
macladson, line 1: May as well also remove the corresponding line in `test_defaults()` (line 2095)

macladson (changes requested):
I can't seem to comment on the files directly but there's a few more references to `safe_slots` in the `consensus/types/presets` files and also `book/src/api-vc-endpoints` that I think we can remove.

consensus/types/src/chain_spec.rs
michaelsproul: done

lighthouse/tests/beacon_node.rs
michaelsproul: done, thanks!

michaelsproul:
Updated thanks

macladson (approved):
LGTM!

michaelsproul:
@mergify queue

PR #6490, opened 2024-10-15
Remove deprecated flags in prep for v6.0.0

michaelsproul:
## Proposed Changes

Delete the following flags from the BN:

- `--self-limiter`
- `--http-spec-fork`
- `--http-allow-sync-stalled`
- `--disable-lock-timeouts`
- `--always-prefer-builder-payload`
- `--progressive-balances`
- `--disable-duplicate-warn-logs`
- `-l` (env logger)

Delete the following flags from the VC:

- `--latency-measurement-service`
- `--disable-run-on-all`
- `--produce-block-v3`

Delete an old `AttestationId` from the operation_pool that was deprecated long ago.


pawanjay176 (commented):
There are a few deprecated flags in beacon_node/cli too. self-limiter, http-spec-fork etc

michaelsproul:
Thanks @pawanjay176! I missed those! Have deleted them now in a31bb49098d5be4464c486a58cdbd9a59bdedfe1

michaelsproul:
If we want to remove `--http-allow-sync-stalled` we are blocked on this change to Kurtosis:

- https://github.com/ethpandaops/ethereum-package/pull/805

Let's see if that gets merged promptly, but if not we can leave the flag for now.

pawanjay176 (approved):

macladson (commented):
There's also the `env_log` flag which is marked as deprecated

michaelsproul:
@mergify queue

PR #6492, opened 2024-10-15
Trigger sampling after gossip block is execution valid

dapplion:
## Issue Addressed

Resolve a PeerDAS todo

## Proposed Changes

Wait for a gossip block to be execution valid before triggering sampling. Otherwise, the block might have invalid transactions and trigger us to ask peers for data that does not exist. However, a proposer could still include valid blob transactions and withhold data, causing the same situation. Since gossip blocks are proposer signature valid, we have some DOS protection.

I'm 50/50 on this change being necessary, thoughts?


pawanjay176:
I think without any time sensitive requirements on sampling, doing it post execution verification makes more sense

pawanjay176 (approved):

dapplion:
The damage the proposer can do is to force us to trigger sampling for a block that actually has no data. This is equivalent to the proposer publishing a valid block with data but withholding the data.

So I'll close for now

PR #6493, opened 2024-10-16
Migrate account manager cli client to clap derive

eserilev:
## Issue Addressed

Partially https://github.com/sigp/lighthouse/issues/5900

## Proposed Changes

Migrate account-manager to clap derive


macladson (changes requested):
I think this PR needs some of the same love from #6300 since `lighthouse am --help` also doesn't work.

common/clap_utils/src/lib.rs
```diff
@@ -57,6 +57,21 @@ pub fn parse_hardcoded_network(
     Eth2NetworkConfig::constant(network_name.as_str())
 }
 
+/// If `name` is in `matches`, parses the value as a path. Otherwise, attempts to find the user's
+/// home directory and appends `default` to it.
```
macladson: I think this comment needs updating. Maybe also explain how this function is different from `v1`

eserilev:
this one should be g2g for another review

common/directory/src/lib.rs
```diff
@@ -62,6 +62,24 @@ pub fn parse_path_or_default_with_flag(
     )
 }
 
+/// If `arg` is in `matches`, parses the value as a path.
+///
+/// Otherwise, attempts to find the default directory for the `testnet` from the `matches`
+/// and appends `flag` to it.
```
macladson: Same here

macladson (changes requested):
Will leave a more thorough review soon, but noticed some comments that need updating

common/clap_utils/src/lib.rs
eserilev: updated

common/directory/src/lib.rs
eserilev: updated

account_manager/src/validator/modify.rs
```diff
@@ -1,91 +1,42 @@
...
-        Some((unknown, _)) => {
-            return Err(format!(
-                "{} does not have a {} command. See --help",
-                CMD, unknown
-            ))
-        }
-        _ => return Err(format!("No command provided for {}. See --help", CMD)),
+    let (enabled, sub_matches) = match modify_config {
+        Modify::Enable(sub_matches) => (true, Box::new(sub_matches) as Box<dyn Modifiable>),
+        Modify::Disable(sub_matches) => (false, Box::new(sub_matches) as Box<dyn Modifiable>),
```
macladson: Unless I'm missing something I think we can remove both the `Box` and `Modifiable` if we just do: 

```suggestion
    let (enabled, pubkey, is_all) = match modify_config {
        Modify::Enable(enable) => (true, enable.pubkey.clone(), enable.all),
        Modify::Disable(disable) => (false, disable.pubkey.clone(), disable.all),
    };
```

lighthouse/src/main.rs
```diff
@@ -756,6 +745,14 @@ fn run<E: EthSpec>(
     }
 
     match LighthouseSubcommands::from_arg_matches(matches) {
+        Ok(LighthouseSubcommands::AccountManager(account_manager_config)) => {
+            eprintln!("Running account manager for {} network", network_name);
+            // Pass the entire `environment` to the account manager so it can run blocking operations.
+            account_manager::run(&account_manager_config, matches, environment)?;
```
macladson: We should reorder this for consistency with the other subcommands
```suggestion
            account_manager::run(matches, &account_manager_config, environment)?;
```

macladson (changes requested):

lighthouse/src/main.rs
eserilev: [c0ce1cc](https://github.com/sigp/lighthouse/pull/6493/commits/c0ce1cceeab464925ea9d7a5cc4e2316fe35f186)

account_manager/src/validator/modify.rs
eserilev: c0ce1cc
9cfcc59

eserilev:
hi @chong-he could i bug you to review this PR when you get the chance :)

chong-he:
> hi @chong-he could i bug you to review this PR when you get the chance :)

I wouldn't say I am capable to review this PR now, but I am happy to do some testing on this, will post if anything. 

Thanks!

chong-he:
After chatting with @michaelsproul , I am going to give reviewing this PR a try

account_manager/src/validator/cli.rs
```diff
@@ -0,0 +1,317 @@
...
+#[clap(
+    about = "Creates new validators from an existing EIP-2386 wallet using the EIP-2333 HD key \
+            derivation scheme."
+)]
+pub struct Create {
+    #[clap(
+        long,
+        value_name = "WALLET_NAME",
+        help = "Use the wallet identified by this name"
+    )]
```
chong-he, line 51: Currently the flags are not sorted alphabetically. We can add `display_order = 0` to make it sorted, as in the validator client and database cli files.

```suggestion
        help = "Use the wallet identified by this name",
        display_order = 0
    )]
```

This applies to all flags in the `cli.rs` file

common/clap_utils/src/lib.rs
```diff
@@ -57,6 +57,20 @@ pub fn parse_hardcoded_network(
     Eth2NetworkConfig::constant(network_name.as_str())
 }
 
+/// If `path` is `Some`, return it, else return the default path
+pub fn parse_path_with_default_in_home_dir_v2(
```
chong-he: Small nit pick, would be good to put this after the function `parse_path_with_default_in_home_dir` for consistency

chong-he (changes requested):
Edit: after a closer look, I think the logging of `No logfile path provided, logging to file is disabled` is actually a feature, not a bug. It is telling users that logging to file is disabled, and it can be enabled with `--logfile-dir` to save all logs to a file. Please ignore the comments below

Looks pretty good overall. ~~There is just this thing that I found:~~


~~For commands in `lighthouse am`, for example, when we run: `lighthouse account wallet create --name test`, this first line is printed when it shouldn't (unless `--logfile-dir` is specified):~~
```
No logfile path provided, logging to file is disabled
Running account manager for mainnet network
wallet-dir path: "/home/ck/.lighthouse/mainnet/wallets"
```

~~It's from here: 
https://github.com/sigp/lighthouse/blob/3e6b0bd0a36dfa434a779812b648d15c9e67025f/lighthouse/environment/src/lib.rs#L217-L221~~

~~This looks like is because from here:~~
https://github.com/sigp/lighthouse/blob/3e6b0bd0a36dfa434a779812b648d15c9e67025f/lighthouse/src/main.rs#L569-L588

~~where `account_manager` falls into the `None` arm of the match automatically. We can add an `if` condition in the logging in `lighthouse/lighthouse/environment/src/lib.rs` to only print it for some subcommands, but that sounds a bit of a hack. I am sure you have a better solution for this.~~

~~I think `logfile-dir` for  `account_manager` is not really relevant, so would be nice to not have that line printed (current `lighthouse db` also have this line printed)~~

beacon_node/http_api/src/inclusion_list_duties.rs
macladson, line 1: Looks like this file snuck in from somewhere else

macladson (commented):
I'll be dedicating some cycles soon to getting the existing `clap` PRs merged as I'm keen to see these happen!

beacon_node/http_api/src/inclusion_list_duties.rs
eserilev: oops, I've removed that file. thanks!

account_manager/src/lib.rs
```diff
@@ -13,27 +15,31 @@ pub const VALIDATOR_DIR_FLAG: &str = "validator-dir";
 pub const VALIDATOR_DIR_FLAG_ALIAS: &str = "validators-dir";
```
macladson: I think these constants are now unused since we define these in the actual struct definitions.

account_manager/src/validator/cli.rs
```diff
@@ -0,0 +1,317 @@
...
+use std::path::PathBuf;
+
+use super::exit::DEFAULT_BEACON_NODE;
+
+#[derive(Parser, Clone, Deserialize, Serialize, Debug)]
+#[clap(about = "Provides commands for managing Eth2 validators.")]
+pub struct Validator {
+    #[clap(
+        long,
+        value_name = "VALIDATOR_DIRECTORY",
```
macladson, line 16: I think we have lost the previous alias `--validators-dir`, which we can add back like this:


```suggestion
    #[clap(
        long,
        alias = "validators-dir",
        value_name = "VALIDATOR_DIRECTORY",
```

account_manager/src/validator/cli.rs
```diff
@@ -0,0 +1,317 @@
...
+        help = "Only presign the voluntary exit message without publishing it"
+    )]
+    pub presign: bool,
+}
+
+#[derive(Parser, Clone, Deserialize, Serialize, Debug)]
+#[clap(
+    about = "Imports one or more EIP-2335 passwords into a Lighthouse VC directory, \
+            requesting passwords interactively. The directory flag provides a convenient \
+            method for importing a directory of keys generated by the eth2-deposit-cli \
```
macladson: Did this get renamed recently? It was `ethstaker-deposit-cli` previously 

```suggestion
            method for importing a directory of keys generated by the ethstaker-deposit-cli \
```

account_manager/src/lib.rs
macladson: There might be more unused constants in other modules too

account_manager/src/wallet/cli.rs
```diff
@@ -0,0 +1,204 @@
...
+        value_name = "WALLET_TYPE",
+        value_enum,
+        default_value_t = WalletType::Hd,
+        help = "The type of wallet to create. Only HD (hierarchical-deterministic) \
+                wallets are supported presently..",
+    )]
+    pub r#type: WalletType,
+}
+
+pub trait NewWallet {
```
macladson: Should this trait exist in `cli.rs`? Seems like we should reserve the `cli.rs` file only for cli definitions

macladson (changes requested):

account_manager/src/lib.rs
eserilev: [555c100](https://github.com/sigp/lighthouse/pull/6493/commits/555c1001f26e35e21e9cdd3eb7552b6dc9ce26bf)

account_manager/src/validator/cli.rs
eserilev: [113c00e](https://github.com/sigp/lighthouse/pull/6493/commits/113c00e21b7f107077e4998aedf3213cd77020de)

account_manager/src/validator/cli.rs
eserilev: [77d08eb](https://github.com/sigp/lighthouse/pull/6493/commits/77d08eb0923fd99b434669fd8fd3e727ec0e987a)

account_manager/src/wallet/cli.rs
eserilev: [98159fb](https://github.com/sigp/lighthouse/pull/6493/commits/98159fb099a05126222178488070a4830051cecf)

account_manager/src/validator/cli.rs
eserilev: i found out that we can define the display order at the struct level instead of individual fields!

common/clap_utils/src/lib.rs
eserilev: its now listed after

eserilev:
I think I tackled all the feedback. Should be g2g for another review

macladson (commented):
Looks like the `/lighthouse/tests` relied on those constants after all. We could either leave them, or hard code them in the lighthouse tests directly. It was kinda nice that they were imported constants, but most of the other cli flag tests just hard code them anyway so I don't think we lose that much realistically.

eserilev:
sorry jsut getting around to this now. the constants are in the test file now

PR #6496, opened 2024-10-16
Electra alpha8 spec updates

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Implements changes for electra from https://github.com/ethereum/consensus-specs/releases/tag/v1.5.0-alpha.7 and https://github.com/ethereum/consensus-specs/releases/tag/v1.5.0-alpha.8

1. Fixed partial withdrawals count https://github.com/ethereum/consensus-specs/pull/3943
2. Remove get_active_balance https://github.com/ethereum/consensus-specs/pull/3949
3. Remove queue_entire_balance_and_reset_validator https://github.com/ethereum/consensus-specs/pull/3951
4. Switch to compounding when consolidating with source==target https://github.com/ethereum/consensus-specs/pull/3918
5. Queue deposit requests and apply them during epoch processing https://github.com/ethereum/consensus-specs/pull/3818

Other changes from the releases are mostly cosmetic/ do not require changes on our end.
Can be reviewed per commit for each change. However, b28d0106299345b3b72e1fedd2c780c9b393c75f contains bug fixes from all the changes


consensus/types/src/deposit_request.rs
```diff
@@ -36,7 +36,7 @@ impl DepositRequest {
             pubkey: PublicKeyBytes::empty(),
             withdrawal_credentials: Hash256::ZERO,
             amount: 0,
-            signature: Signature::empty(),
+            signature: SignatureBytes::empty(),
```
pawanjay176, line 39: Should we change the type to SignatureBytes for the other request types as well for consistency?

consensus/state_processing/src/genesis.rs
```diff
@@ -120,6 +120,9 @@ pub fn initialize_beacon_state_from_eth1<E: EthSpec>(
         let post = upgrade_state_to_electra(&mut state, Epoch::new(0), Epoch::new(0), spec)?;
         state = post;
 
+        // TODO(electra): do we need to iterate over an empty pending_deposits list here and increase balance?
+        // in accordance with `initialize_beacon_state_from_eth1` function from the spec
```
dapplion: No because we have already populated the validators array in line 42 and 45

consensus/state_processing/src/per_block_processing/process_operations.rs
```diff
@@ -450,59 +449,69 @@ pub fn apply_deposit<E: EthSpec>(
...
+                exit_epoch: spec.far_future_epoch,
+                withdrawable_epoch: spec.far_future_epoch,
+                effective_balance: 0,
+                slashed: false,
+            };
+            let max_effective_balance =
+                validator.get_max_effective_balance(spec, state.fork_name_unchecked());
+            validator.effective_balance = std::cmp::min(
+                amount.safe_sub(amount.safe_rem(spec.effective_balance_increment)?)?,
+                max_effective_balance,
```
dapplion: This code is unnecessary, amount is hardcoded to 0

consensus/state_processing/src/per_block_processing/process_operations.rs
```diff
@@ -658,11 +675,73 @@ pub fn process_consolidation_requests<E: EthSpec>(
...
+
+    Ok(true)
+}
+
 pub fn process_consolidation_request<E: EthSpec>(
     state: &mut BeaconState<E>,
     consolidation_request: &ConsolidationRequest,
     spec: &ChainSpec,
 ) -> Result<(), BlockProcessingError> {
+    dbg!("here");
```
dapplion: delete

consensus/state_processing/src/upgrade/electra.rs
```diff
@@ -57,7 +58,32 @@ pub fn upgrade_to_electra<E: EthSpec>(
...
+
+        let validator = post
+            .validators_mut()
+            .get_mut(index)
+            .ok_or(Error::UnknownValidator(index))?;
+        validator.effective_balance = 0;
+        validator.activation_eligibility_epoch = spec.far_future_epoch;
+        let pubkey = validator.pubkey;
+        let withdrawal_credentials = validator.withdrawal_credentials;
+        dbg!("is this getting hit");
```
dapplion: delete

consensus/state_processing/src/per_block_processing/process_operations.rs
```diff
@@ -450,59 +449,69 @@ pub fn apply_deposit<E: EthSpec>(
...
             return Ok(());
         }
 
-        let new_validator_index = state.validators().len();
-
         // [Modified in Electra:EIP7251]
-        let (effective_balance, state_balance) = if state.fork_name_unchecked() >= ForkName::Electra
-        {
-            (0, 0)
+        if state.fork_name_unchecked() >= ForkName::Electra {
```
ethDreamer:
```suggestion
        if state.fork_name_unchecked().electra_enabled() {
```

consensus/state_processing/src/per_block_processing/process_operations.rs
ethDreamer: It's not unnecessary (this is logic for setting effective balance which is in the spec), but it IS a bug because amount has been hardcoded to zero. I think this bit of code is getting more and more difficult to manage as we add forks so I've created a PR to simplify this whole piece of code:

https://github.com/sigp/lighthouse/pull/6515

it needs a review to ensure we're compliant with the spec across all forks. But if we merge that into this PR it'll simplify this a ton.

ethDreamer: nvm this isn't a bug because the spec set `amount` to `0` much earlier as well.. but the simplification should still be done.

consensus/state_processing/src/per_block_processing/process_operations.rs
```diff
@@ -658,11 +675,73 @@ pub fn process_consolidation_requests<E: EthSpec>(
...
+    // Verify pubkey exists
+    let Some(source_index) = state
+        .pubkey_cache()
+        .get(&consolidation_request.source_pubkey)
+    else {
+        // source validator doesn't exist
+        return Ok(false);
+    };
+
+    let source_validator = state.get_validator(source_index)?;
```
ethDreamer, line 654: This last line is the only reason this function returns a `Result<bool>` rather than a `bool`. But it shouldn't ever return an error if the line above returned a pubkey. But since you need to fetch this validator here anyway, just combine them so this function has no reliance on the pubkey cache:
```rust
let source_validator = match state.get_validator(source_index) {
    Ok(v) => v,
    Err(_) => return false, 
};
```
then you can refactor this whole function to just return a `bool`

consensus/state_processing/src/per_block_processing/process_operations.rs
```diff
@@ -658,11 +675,73 @@ pub fn process_consolidation_requests<E: EthSpec>(
...
+    // Verify the source withdrawal credentials
+    if let Some(withdrawal_address) = source_validator.get_eth1_withdrawal_credential(spec) {
+        if withdrawal_address != consolidation_request.source_address {
+            return Ok(false);
+        }
+    } else {
+        // Source doen't have execution withdrawal credentials
```
ethDreamer:
```suggestion
        // Source doesn't have eth1 withdrawal credentials
```
the distinction here is important

consensus/state_processing/src/per_block_processing/process_operations.rs
```diff
@@ -658,11 +675,73 @@ pub fn process_consolidation_requests<E: EthSpec>(
...
     spec: &ChainSpec,
 ) -> Result<(), BlockProcessingError> {
+    dbg!("here");
+    if is_valid_switch_to_compounding_request(state, consolidation_request, spec)? {
+        let Some(source_index) = state
+            .pubkey_cache()
+            .get(&consolidation_request.source_pubkey)
+        else {
+            // source validator doesn't exist. This is unreachable as `is_valid_switch_to_compounding_request`
+            // will return false in that case.
```
ethDreamer, line 701: If you implement the function above, the previous function no longer relies on the pubkey cache, so technically this could get hit if the cache is stale.. in which case I think we'd probably wanna build it and check again?

consensus/types/src/validator.rs
```diff
@@ -152,6 +152,17 @@ impl Validator {
...
+    pub fn get_eth1_withdrawal_credential(&self, spec: &ChainSpec) -> Option<Address> {
+        self.has_eth1_withdrawal_credential(spec)
+            .then(|| {
+                self.withdrawal_credentials
+                    .as_slice()
+                    .get(12..)
+                    .map(Address::from_slice)
+            })
+            .flatten()
+    }
```
ethDreamer: i hate that we added this function back just for a single place in the spec.. this is gonna cause a bug at some point due to confusing with `get_execution_withdrawal_credential`..

consensus/types/src/beacon_state.rs
```diff
@@ -2214,12 +2169,10 @@ impl<E: EthSpec> BeaconState<E> {
             .validators_mut()
             .get_mut(validator_index)
             .ok_or(Error::UnknownValidator(validator_index))?;
-        if validator.has_eth1_withdrawal_credential(spec) {
-            AsMut::<[u8; 32]>::as_mut(&mut validator.withdrawal_credentials)[0] =
-                spec.compounding_withdrawal_prefix_byte;
+        AsMut::<[u8; 32]>::as_mut(&mut validator.withdrawal_credentials)[0] =
+            spec.compounding_withdrawal_prefix_byte;
```
ethDreamer, line 2205: hmmm we should probably refactor the [`change_withdrawal_credentials()`](https://github.com/sigp/lighthouse/blob/v5.3.0/consensus/types/src/validator.rs#L158) method on the `Validator` so that it accepts a prefix.

consensus/types/src/deposit_request.rs
ethDreamer: curious why we did this?

consensus/state_processing/src/per_epoch_processing/single_pass.rs
```diff
@@ -342,20 +345,66 @@ pub fn process_epoch_single_pass<E: EthSpec>(
...
+                    pubkey: deposit.pubkey,
+                    withdrawal_credentials: deposit.withdrawal_credentials,
+                    amount: deposit.amount,
+                    signature: deposit.signature,
+                },
+                spec,
+            )
+            .is_ok()
+            {
+                let mut validator = Validator {
```
ethDreamer: this will also be significantly simplified now that we have `add_validator_to_registry()` :D

consensus/state_processing/src/per_block_processing/process_operations.rs
pawanjay176: Fixed with merging unstable and 37c7658 to get it more in line with the spec

consensus/state_processing/src/per_block_processing/process_operations.rs
pawanjay176: Resolved in [9f85074](https://github.com/sigp/lighthouse/pull/6496/commits/9f850748548612f004ff009dc3e8fce7d6829898)

consensus/state_processing/src/per_block_processing/process_operations.rs
pawanjay176: Fixed

consensus/types/src/validator.rs
pawanjay176: I agree, spent quite a bit of time fighting the tests on this one. I'm just gonna use the function directly here instead

consensus/types/src/beacon_state.rs
pawanjay176: This way seems cleaner to me tbh, we are changing it in place instead of creating the entire byte array from scratch

consensus/types/src/validator.rs
pawanjay176: Fixed in 707338d

consensus/state_processing/src/genesis.rs
pawanjay176: was talking about [this](https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#testing)
```python

    # Process deposit balance updates
    validator_pubkeys = [v.pubkey for v in state.validators]
    for deposit in state.pending_deposits:
        validator_index = ValidatorIndex(validator_pubkeys.index(deposit.pubkey))
        increase_balance(state, validator_index, deposit.amount)
    state.pending_deposits = []
```
Added in b2b728d

consensus/types/src/deposit_request.rs
pawanjay176: Because we need to allow for invalid bls signature bytes to be sent over the engine api. Otherwise, it becomes a deserialization error when we get the payload from the EL instead of an invalid deposit during processing

consensus/types/src/beacon_state.rs
ethDreamer: Can the `change_withdrawal_credentials()` function not be refactored to also do it in place?

pawanjay176: that function also changes the execution address, not just the prefix. Can do it, but i'm not sure what it adds 🤷‍♂️

ethDreamer: Gotcha. I guess I'm just arguing for encapsulating it as a function on the validator. Honestly just realizing the naming of this sucks. `0x01` is `eth1` credentials (despite wanting to move away from "eth1" and towards "execution") but "execution" now means and both `0x01` and `0x02` 😩

I suppose we could fix this in a future PR.. but like what do you think about being explicit:
`eth1` -> `0x01` for example `has_0x01_withdrawal_credentials`

and we could rename `change_withdrawal_credentials` to `change_to_0x01_credentials` and add a new function `change_to_0x02_credentials` which encapsulates what you have above. `has_execution_withdrawal_credentials` remains the same and is more consistent now.

pawanjay176: Yeah I agree the naming is super confusing, agree with doing the renaming in a future PR

consensus/state_processing/src/per_epoch_processing/single_pass.rs
```diff
@@ -369,10 +372,60 @@ pub fn process_epoch_single_pass<E: EthSpec>(
...
+                        deposit_data.amount,
+                        spec,
+                    )?;
+                    added_validators
+                        .push((deposit_data.pubkey, state.validators().len().safe_sub(1)?));
+                }
+            }
+        }
+        if conf.effective_balance_updates {
+            // Re-process effective balance updates for validators affected by top-up of new validators.
```
pawanjay176, line 392: I'm very sus about this. Seems to have fixed the deposits processing bug though
cc @michaelsproul

pawanjay176: For some added context, we want deposit processing to work when `process_pending_deposits` processes 2 deposits where deposit 1 is a new validator and deposit 2 is a topup for the new validator.
Since `process_effective_balance_updates` happens after `process_pending_deposits`, we need the effective balance update happening after adding the new validator

beacon_node/execution_layer/src/block_hash.rs
```diff
@@ -221,29 +228,30 @@ mod test {
...
-            excess_blob_gas: Some(0x0u64),
-            parent_beacon_block_root: Some(Hash256::from_str("f7d327d2c04e4f12e9cdd492e53d39a1d390f8b1571e3b2a22ac6e1e170e5b1a").unwrap()),
+            blob_gas_used: Some(786432),
+            excess_blob_gas: Some(44695552),
+            parent_beacon_block_root: Some(Hash256::from_str("f3a888fee010ebb1ae083547004e96c254b240437823326fdff8354b1fc25629").unwrap()),
+            requests_root: Some(Hash256::from_str("9440d3365f07573919e1e9ac5178c20ec6fe267357ee4baf8b6409901f331b62").unwrap()),
         };
         let expected_hash =
-            Hash256::from_str("a7448e600ead0a23d16f96aa46e8dea9eef8a7c5669a5f0a5ff32709afe9c408")
+            Hash256::from_str("61e67afc96bf21be6aab52c1ace1db48de7b83f03119b0644deb4b69e87e09e1")
```
michaelsproul, line 254: Where is this test case from?

consensus/state_processing/src/genesis.rs
```diff
@@ -123,6 +123,19 @@ pub fn initialize_beacon_state_from_eth1<E: EthSpec>(
...
+        for pending_deposit in pending_deposits.into_iter() {
+            if let Some(validator_index) = state.pubkey_cache().get(&pending_deposit.pubkey) {
+                state
+                    .balances_mut()
+                    .get_mut(validator_index)
+                    .ok_or(Error::UnknownValidator(validator_index))?
+                    .safe_add(pending_deposit.amount)?;
+            }
+        }
+        *state.pending_deposits_mut()? = List::empty();
```
michaelsproul: What is this doing? How can we be iterating over an empty list -- doesn't this just not do anything?

consensus/state_processing/src/per_block_processing/process_operations.rs
```diff
@@ -621,11 +632,87 @@ pub fn process_consolidation_requests<E: EthSpec>(
...
+        .pubkey_cache()
+        .get(&consolidation_request.source_pubkey)
+    else {
+        // source validator doesn't exist
+        return false;
+    };
+
+    let Ok(source_validator) = state.get_validator(source_index) else {
+        // Validator must exist in the state else it wouldn't exist in the pubkey cache either
+        return false;
```
michaelsproul: I think it's safer to return an `Err` here if we think this is unreachable

consensus/state_processing/src/per_block_processing/process_operations.rs
```diff
@@ -699,5 +782,10 @@ pub fn process_consolidation_request<E: EthSpec>(
             target_index: target_index as u64,
         })?;
 
+    let target_validator = state.get_validator(target_index)?;
+    // Churn any target excess active balance of target and raise its max
+    if target_validator.has_eth1_withdrawal_credential(spec) {
+        state.switch_to_compounding_validator(target_index, spec)?;
+    }
```
michaelsproul, line 786: Just noting we need to delete this for `alpha.9`.

consensus/types/src/pending_deposit.rs
```diff
@@ -8,7 +9,6 @@ use tree_hash_derive::TreeHash;
     arbitrary::Arbitrary,
     Debug,
     PartialEq,
-    Eq,
```
michaelsproul, line 11: Did we lose `Eq` because of `PublicKeyBytes`/`SignatureBytes` not implementing it? I guess we don't need it.

michaelsproul (commented):
I've reviewed all but 2 files, but one of those files is `single_pass` :sweat_smile: Will come back for that tomorrow

beacon_node/execution_layer/src/block_hash.rs
pawanjay176: I got it from a devnet 4 block iirc.

consensus/state_processing/src/genesis.rs
pawanjay176: My understanding was this applies pending_deposits to the state during initialization if the state contains pending_deposits. Here's the part of the [spec](https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.8/specs/electra/beacon-chain.md#testing)
```python
for deposit in state.pending_deposits:
        validator_index = ValidatorIndex(validator_pubkeys.index(deposit.pubkey))
        increase_balance(state, validator_index, deposit.amount)
    state.pending_deposits = []
```

consensus/types/src/pending_deposit.rs
pawanjay176: Yep that was it

consensus/state_processing/src/genesis.rs
michaelsproul: Ohh right, I was confused because we are calling `upgrade_state_to_electra` immediately prior to this, which sets the `pending_deposits` to the empty list:

https://github.com/sigp/lighthouse/blob/6d10456912b3c39b8a8c9089db76e8ead20608a0/consensus/state_processing/src/genesis.rs#L120

https://github.com/sigp/lighthouse/blob/6d10456912b3c39b8a8c9089db76e8ead20608a0/consensus/state_processing/src/upgrade/electra.rs#L165

I think we might need to move the `apply_deposit` logic _after_ upgrading to Electra, so that the `pending_deposits` list gets populated. Currently it happens here:

https://github.com/sigp/lighthouse/blob/6d10456912b3c39b8a8c9089db76e8ead20608a0/consensus/state_processing/src/genesis.rs#L36-L43

Or maybe we just delete the whole `initialize_beacon_state_from_eth1` function and say we don't support starting from eth1 genesis any more :sweat_smile:

It's interesting that the EF tests for `initialize_beacon_state_from_eth1` already pass. Maybe processing the deposits with the pre-Electra logic is actually equivalent, and we could just delete the code here that loops over `pending_deposits`?

michaelsproul:
@pawanjay176 I've optimised and simplified part of the logic in `PendingDepositsContext::new`. We can avoid looking up the validator index twice, and we also don't need to handle the case for withdrawn validators that are also new, because that's impossible:

- 0af6b0fe88a60ba7cb68500c089723cf2d355991

consensus/state_processing/src/per_epoch_processing/single_pass.rs
michaelsproul: I agree with this assessment of the situation. It's a bit gnarly that we need to do an effective balance update here outside the main single pass loop, but I think it's the best way.

I was thinking it might make sense to put this block of code after the logic for processing pending consolidations because that's the order used by the spec:

```python
process_pending_deposits(state)  # [New in Electra:EIP7251]
process_pending_consolidations(state)  # [New in Electra:EIP7251]
process_effective_balance_updates(state)  # [Modified in Electra:EIP7251]
```

However, the spec also requires that consolidations are for known validators (both source and target), so that should not interact at all with our newly added validators.

michaelsproul: The only other thing I would consider doing would be getting rid of the `added_validators` vec and just running effective balance updates for all the validators from `num_validators..` on wards.

Ideally we could even use something like `iter_from` that returns CoW pointers (`iter_cow_from`), however this doesn't exist in Milhouse yet.

michaelsproul: Opened an issue on Milhouse:

- https://github.com/sigp/milhouse/issues/57

michaelsproul:
Pushed another commit which I think fixes a consensus bug :hot_pepper:

- 21591a332392797f3a37125ce2bcea18ef59c635

Let me know if you agree with my reasoning @pawanjay176. The EF tests pass with both versions of the code, which implies test coverage could be improved here.

consensus/state_processing/src/genesis.rs
pawanjay176: Yeah i'm okay with that. It seemed redundant to me initially but i just implemented it for completeness.
I'm fine with removing it

pawanjay176:
Great find 🙌  Turns out there was a spec test to catch this specific case, but it was written incorrectly. This has been fixed in the linked consensus specs PR above

michaelsproul:
Nice! Thanks for following that up!

consensus/state_processing/src/genesis.rs
pawanjay176: Removed in d5ec058

michaelsproul:
Some of the tests seem to be failing due to builder stuff. Did you mean to include mock-builder changes in this PR?

pawanjay176:
Ahh I didn't realise I pushed the mock builder changes to this PR. I'll revert them and fix the tests.
Will make a separate PR for the mock builder changes

michaelsproul:
I think I've fixed another consensus bug in this commit:

- ebabd0f83616d2920f69c07fa193af3530ffa30e

The spec iterates over `post.validators`, but we were iterating over a clone of `validators` taken prior to modifying `post.validators` in the `pre_activation` loop. So if there are any validators that are pre-activation with compounding withdrawal credentials, we would **double credit them** by pushing a PendingDeposit in the pre_activation loop here:

https://github.com/sigp/lighthouse/blob/ebabd0f83616d2920f69c07fa193af3530ffa30e/consensus/state_processing/src/upgrade/electra.rs#L68-L76

And then _another_ deposit if they meet the conditions `has_compounding_withdrawal_credential` (true) and `balance > spec.min_activation_balance` (true):

https://github.com/sigp/lighthouse/blob/ebabd0f83616d2920f69c07fa193af3530ffa30e/consensus/state_processing/src/upgrade/electra.rs#L81-L85

https://github.com/sigp/lighthouse/blob/ebabd0f83616d2920f69c07fa193af3530ffa30e/consensus/types/src/beacon_state.rs#L2170-L2192

Additionally, in this case, the balance of the validator would be set to `spec.min_activation_balance` rather than 0 (the correct value). So their total balance including deposits would end up being `balance (first deposit) + excess_balance (second deposit) + min_activation_balance (state) = 2 * balance`.

michaelsproul:
I also simplified some sorting in this commit (non-semantic change):

- 75efd4ffbf9449182a44fdae2a30629a018e1695

michaelsproul (approved):
Looks good. Let's merge this so we can get other Electra PRs into unstable on top

pawanjay176:
Great find with the other accounting bug as well! 🙏 I looked through it and also agree with your solution.

michaelsproul:
@mergify queue

michaelsproul:
@mergify requeue

michaelsproul:
Yay!

PR #6497, opened 2024-10-16
Add individual by_range sync requests

dapplion:
## Issue Addressed

Part of

To address PeerDAS sync issues we need to make individual by_range requests within a batch retriable. We should adopt the same pattern for lookup sync where each request (block/blobs/columns) is tracked individually within a "meta" request that group them all and handles retry logic.

## Proposed Changes

- Building on https://github.com/sigp/lighthouse/pull/6398

second step is to add individual request accumulators for `blocks_by_range`, `blobs_by_range`, and `data_columns_by_range`. This will allow each request to progress independently and be retried separately.

Most of the logic is just piping, excuse the large diff. This PR does not change the logic of how requests are handled or retried. This will be done in a future PR changing the logic of `RangeBlockComponentsRequest`.

### Before

- Sync manager receives block with `SyncRequestId::RangeBlockAndBlobs`
- Insert block into `SyncNetworkContext::range_block_components_requests`
- (If received stream terminators of all requests)
  - Return `Vec<RpcBlock>`, and insert into `range_sync`

### Now

- Sync manager receives block with `SyncRequestId::RangeBlockAndBlobs`
- Insert block into `SyncNetworkContext:: blocks_by_range_requests`
- (If received stream terminator of this request)
  - Return `Vec<SignedBlock>`, and insert into `SyncNetworkContext::components_by_range_requests `
  - (If received a result for all requests)
    - Return `Vec<RpcBlock>`, and insert into `range_sync`


beacon_node/network/src/sync/range_sync/range.rs
```diff
@@ -837,7 +853,12 @@ mod tests {
...
 
-        rig.expect_chain_segment();
-        rig.expect_chain_segment();
+        // TODO: To actually force range sync to send batches to the processor this tests need to be
+        // refactored into using the event driven pattern of the lookup tests. Otherwise
+        // `complete_range_block_and_blobs_response` doesn't actually complete any requests in
+        // practice
+        //
+        // rig.expect_chain_segment();
+        // rig.expect_chain_segment();
```
dapplion: I want to migrate the range sync tests to the event-driven format but would prefer to do so in a separate PR

beacon_node/lighthouse_network/src/service/api_types.rs
```diff
@@ -43,12 +47,50 @@ pub struct DataColumnsByRootRequestId {
     pub requester: DataColumnsByRootRequester,
 }
 
+#[derive(Debug, Hash, PartialEq, Eq, Clone, Copy)]
+pub struct BlocksByRangeRequestId {
+    pub id: Id,
```
realbigsean, line 53: i'm understanding this is the sub-request id and `requester.id` is the meta-request id, can we rename one of them? it's sort of confusing

beacon_node/network/src/sync/network_context/requests/blobs_by_range.rs
```diff
@@ -0,0 +1,42 @@
...
+impl<E: EthSpec> ActiveRequestItems for BlobsByRangeRequestItems<E> {
+    type Item = Arc<BlobSidecar<E>>;
+
+    fn add(&mut self, item: Self::Item) -> Result<bool, LookupVerifyError> {
+        if item.slot() < self.request.start_slot
+            || item.slot() >= self.request.start_slot + self.request.count
+        {
+            return Err(LookupVerifyError::UnrequestedSlot(item.slot()));
+        }
+        // TODO: Should check if index is within bounds
```
realbigsean: As in the blob index? might as well throw that check in now right

beacon_node/network/src/sync/network_context.rs
```diff
@@ -211,21 +208,9 @@ pub struct SyncNetworkContext<T: BeaconChainTypes> {
 
 /// Small enumeration to make dealing with block and blob requests easier.
 pub enum BlockOrBlob<E: EthSpec> {
```
realbigsean: Might be time to rename this.. maybe `RangeBlockComponent` because it should be the wrapper for the return type of `RangeBlockComponentRequest`?

beacon_node/network/src/sync/network_context.rs
```diff
@@ -341,113 +351,59 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
-        let expected_blobs = if matches!(batch_type, ByRangeRequestType::BlocksAndBlobs) {
-            debug!(
-                self.log,
-                "Sending BlobsByRange requests";
-                "method" => "BlobsByRange",
-                "count" => request.count(),
-                "epoch" => epoch,
-                "peer" => %peer_id,
-            );
+        let _blocks_req_id = self.send_blocks_by_range_request(peer_id, request.clone(), id)?;
```
realbigsean, line 367: The `_blocks_req_id` here is unused and `blobs_req_id` is only used to check if a blobs request was made. I don't see these `send_` methods used anywhere else so I think we can return `Result<(), ..>` rather than `Result<BlocksByRangeRequestId, ..>` for blocks. Maybe a bool for blobs. Looks like for custody columns, we care about the count requested but not particular ids

beacon_node/network/src/sync/network_context.rs
```diff
@@ -484,54 +440,55 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
-            Some(sender_id)
-        } else {
-            debug!(self.log, "Sync range request failed"; "request_id" => request_id);
-            None
-        }
-    }
-
     /// Received a blocks by range or blobs by range response for a request that couples blocks '
     /// and blobs.
     pub fn range_block_and_blob_response(
```
realbigsean: another rename suggestion `range_block_component_response`

beacon_node/network/src/sync/network_context.rs
```diff
@@ -484,54 +440,55 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
 
-        let (_, info) = entry.get_mut();
-        match block_or_blob {
-            BlockOrBlob::Block(maybe_block) => info.add_block_response(maybe_block),
-            BlockOrBlob::Blob(maybe_sidecar) => info.add_sidecar_response(maybe_sidecar),
-            BlockOrBlob::CustodyColumns(column) => info.add_data_column(column),
+        if let Err(e) = {
+            let request = entry.get_mut();
+            match block_or_blob {
+                BlockOrBlob::Block(resp) => match resp {
```
realbigsean: Can do this here and for blobs and columns as well
```
resp.map(|(blocks, _)| {
    request.add_blocks(blocks);
})
```

realbigsean (changes requested):
Looks really good overall!

beacon_node/network/src/sync/network_context.rs
dapplion: In a future PR `RangeBlockComponentsRequest` will track individual requests by ID. Left this variable named but unused as a hint that it should be used.

beacon_node/lighthouse_network/src/service/api_types.rs
dapplion: What name(s) do you have in mind? I can add docs otherwise

beacon_node/network/src/sync/network_context.rs
dapplion: Addressed in 5fa291240

beacon_node/network/src/sync/network_context.rs
dapplion: Addressed in 5fa291240

beacon_node/network/src/sync/network_context.rs
dapplion: Addressed in 5fa291240

beacon_node/network/src/sync/network_context/requests/blobs_by_range.rs
dapplion: Addressed in 5fa291240

beacon_node/network/src/sync/network_context/requests/blocks_by_range.rs
```diff
@@ -0,0 +1,40 @@
...
+    fn add(&mut self, item: Self::Item) -> Result<bool, LookupVerifyError> {
+        if item.slot().as_u64() < *self.request.start_slot()
+            || item.slot().as_u64() >= self.request.start_slot() + self.request.count()
+        {
+            return Err(LookupVerifyError::UnrequestedSlot(item.slot()));
+        }
+
+        self.items.push(item);
+
+        Ok(self.items.len() >= *self.request.count() as usize)
```
jimmygchen, line 42: Should we check if a block for the slot already exists in `self.items`, to handle the case where the same block is sent twice? or is this handled elsewhere?

beacon_node/network/src/sync/network_context/requests/data_columns_by_range.rs
```diff
@@ -0,0 +1,43 @@
...
+        if !self.request.columns.contains(&item.index) {
+            return Err(LookupVerifyError::UnrequestedIndex(item.index));
+        }
+
+        self.items.push(item);
+
+        Ok(self.items.len() >= self.request.count as usize * self.request.columns.len())
+    }
+
+    fn consume(&mut self) -> Vec<Self::Item> {
```
jimmygchen, line 51: Looks like this impl can be moved to `ActiveRequestItems`?

beacon_node/network/src/sync/network_context/requests/blocks_by_range.rs
dapplion: Added the check for duplicate data

beacon_node/network/src/sync/network_context/requests/data_columns_by_range.rs
dapplion: It can't as it uses state in `DataColumnsByRangeRequestItems`

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -29,18 +28,13 @@ pub struct RangeBlockComponentsRequest<E: EthSpec> {
     /// Used to determine if the number of data columns stream termination this accumulator should
     /// wait for. This may be less than the number of `expects_custody_columns` due to request batching.
     num_custody_column_requests: Option<usize>,
-    /// The peers the request was made to.
-    pub(crate) peer_ids: Vec<PeerId>,
-    max_blobs_per_block: usize,
```
dapplion, line 34: Removed this from the state, since `into_responses_with_blobs` has access to the spec

beacon_node/network/src/sync/network_context/requests/blocks_by_range.rs
dapplion: Added the duplicate data check to all by_range requests

dapplion:
Rebased on latest unstable, @jimmygchen ready for another review

beacon_node/network/src/sync/network_context.rs
```diff
@@ -198,13 +170,22 @@ pub struct SyncNetworkContext<T: BeaconChainTypes> {
...
+    data_columns_by_range_requests:
+        ActiveRequests<DataColumnsByRangeRequestId, DataColumnsByRangeRequestItems<T::EthSpec>>,
 
     /// Mapping of active custody column requests for a block root
     custody_by_root_requests: FnvHashMap<CustodyRequester, ActiveCustodyRequest<T>>,
 
     /// BlocksByRange requests paired with BlobsByRange
-    range_block_components_requests:
-        FnvHashMap<Id, (RangeRequestId, RangeBlockComponentsRequest<T::EthSpec>)>,
+    components_by_range_requests:
```
jimmygchen, line 188: nit: comment above can be updated to include `DataColumnsByRange` or the overall components by range request.

beacon_node/lighthouse_network/src/service/api_types.rs
jimmygchen: What about changing `requester` to `parent_request_id` or `range_request_id`? and also would be helpful to have some docs too.

```rust
/// The Id of the overall By Range request for block components.
pub parent_request_id: ComponentsByRangeRequestId
```

I agree calling it `requester` here is a bit confusing and it did took me a while to wrap my head around the fields - as the requester is actually `request_id.requester.requester`

if we rename it, it would be
- `request_id.id`: id of the sub request
- `request_id.parent_request_id.id`: id of the overall by range request
- `request_id.parent_request_id.requester`:  requester containing the `batch_id`

I think it's a bit clearer?

beacon_node/network/src/sync/network_context/requests/data_columns_by_range.rs
jimmygchen: oh yeah sorry!

beacon_node/network/src/sync/network_context/requests/blobs_by_range.rs
```diff
@@ -0,0 +1,53 @@
...
+            .iter()
+            .any(|existing| existing.slot() == blob.slot() && existing.index == blob.index)
+        {
+            return Err(LookupVerifyError::DuplicatedData(blob.slot(), blob.index));
+        }
+
+        self.items.push(blob);
+
+        // Skip check if blobs are ready as it's rare that all blocks have max blobs
+        Ok(false)
```
jimmygchen, line 50: Do we verify blob inclusion proofs anywhere else?

beacon_node/network/src/sync/tests/range.rs
```diff
@@ -116,7 +121,7 @@ impl TestRig {
                     NetworkMessage::SendRequest {
                         peer_id,
                         request: RequestType::BlobsByRange(_),
-                        request_id: AppRequestId::Sync(SyncRequestId::RangeBlockAndBlobs { id }),
+                        request_id: AppRequestId::Sync(SyncRequestId::BlobsByRange(id)),
```
jimmygchen, line 212: Just a heads up that I've added data column handling for these tests - if this one gets approved first I'll address it in my PR :)

beacon_node/network/src/sync/manager.rs
```diff
@@ -1224,27 +1243,26 @@ impl<T: BeaconChainTypes> SyncManager<T> {
...
     /// Handles receiving a response for a range sync request that should have both blocks and
     /// blobs.
-    fn range_block_and_blobs_response(
+    fn on_range_components_response(
         &mut self,
-        id: Id,
+        id: ComponentsByRangeRequestId,
         peer_id: PeerId,
-        block_or_blob: BlockOrBlob<T::EthSpec>,
+        block_or_blob: RangeBlockComponent<T::EthSpec>,
```
jimmygchen: I think we can also rename the parameter name to `range_block_component`?

beacon_node/network/src/sync/manager.rs
```diff
@@ -1224,27 +1243,26 @@ impl<T: BeaconChainTypes> SyncManager<T> {
 
     /// Handles receiving a response for a range sync request that should have both blocks and
     /// blobs.
-    fn range_block_and_blobs_response(
+    fn on_range_components_response(
         &mut self,
-        id: Id,
+        id: ComponentsByRangeRequestId,
```
jimmygchen: perhaps rename this to `range_request_id`? 
The `id.id` below is a bit confusing to me.

beacon_node/network/src/sync/network_context.rs
```diff
@@ -267,37 +239,59 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
-                        Some(SyncRequestId::RangeBlockAndBlobs { id: *id })
-                    } else {
-                        None
-                    }
-                });
-
-        let failed_block_ids = self
-            .blocks_by_root_requests
+        // Note: using destructuring pattern without a default case to make sure we don't forget to
+        // add new request types to the this function. Otherwise, lookup sync can break and lookups
```
jimmygchen:
```suggestion
        // add new request types to this function. Otherwise, lookup sync can break and lookups
```

beacon_node/network/src/sync/network_context.rs
```diff
@@ -356,116 +350,63 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
-                (Some(column_indexes), Some(num_of_custody_column_req))
+                (
+                    Some(column_indexes.into_iter().collect::<Vec<_>>()),
+                    Some(data_column_requests),
+                )
             } else {
                 (None, None)
             };
 
         // TODO(pawan): this would break if a batch contains multiple epochs
```
jimmygchen: I think this comment can e removed?

beacon_node/network/src/sync/network_context.rs
```diff
@@ -502,54 +443,43 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
-    pub fn range_block_and_blob_response(
+    pub fn range_block_component_response(
         &mut self,
-        request_id: Id,
-        block_or_blob: BlockOrBlob<T::EthSpec>,
-    ) -> Option<BlocksAndBlobsByRangeResponse<T::EthSpec>> {
-        let Entry::Occupied(mut entry) = self.range_block_components_requests.entry(request_id)
-        else {
+        id: ComponentsByRangeRequestId,
+        block_or_blob: RangeBlockComponent<T::EthSpec>,
```
jimmygchen: Should we rename this to `block_component` or `range_block_component`?

beacon_node/network/src/sync/network_context.rs
```diff
@@ -999,6 +1038,41 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+        self.report_rpc_response_errors(resp, peer_id)
+    }
+
+    #[allow(clippy::type_complexity)]
+    pub(crate) fn on_data_columns_by_range_response(
+        &mut self,
+        id: DataColumnsByRangeRequestId,
+        peer_id: PeerId,
+        rpc_event: RpcEvent<Arc<DataColumnSidecar<T::EthSpec>>>,
+    ) -> Option<RpcResponseResult<Vec<Arc<DataColumnSidecar<T::EthSpec>>>>> {
```
jimmygchen: Could replace `Vec<..>` with `DataColumnSidecarList`?

beacon_node/network/src/sync/network_context.rs
```diff
@@ -1153,21 +1227,21 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
-        );
-        metrics::set_gauge_vec(
-            &metrics::SYNC_ACTIVE_NETWORK_REQUESTS,
-            &["range_blocks"],
-            self.range_block_components_requests.len() as i64,
-        );
+        for (id, count) in [
+            ("blocks_by_root", self.blocks_by_root_requests.len()),
+            ("blobs_by_root", self.blobs_by_root_requests.len()),
+            ("data_columns_by_root", self.blocks_by_root_requests.len()),
```
jimmygchen:
```suggestion
            ("data_columns_by_root", self.data_columns_by_root_requests.len()),
```

beacon_node/network/src/sync/network_context.rs
```diff
@@ -1153,21 +1227,21 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+            ("blocks_by_range", self.blocks_by_range_requests.len()),
+            ("blobs_by_range", self.blobs_by_range_requests.len()),
+            ("data_columns_by_range", self.blocks_by_range_requests.len()),
```
jimmygchen:
```suggestion
            ("data_columns_by_range", self.data_columns_by_range_requests.len()),
```

beacon_node/network/src/router.rs
```diff
@@ -611,7 +611,7 @@ impl<T: BeaconChainTypes> Router<T> {
     ) {
         let request_id = match request_id {
             AppRequestId::Sync(sync_id) => match sync_id {
-                id @ SyncRequestId::RangeBlockAndBlobs { .. } => id,
+                id @ SyncRequestId::BlocksByRange { .. } => id,
```
jimmygchen, line 627: Is it worth matching the `SyncRequestId` for other `_by_range_response` functions too?

jimmygchen (changes requested):
Looks great! 

The `id` naming was a bit confusing to me at first, but overall the logic is actually simpler than I thought 👍

beacon_node/lighthouse_network/src/service/api_types.rs
dapplion: Fixed with 11f903022

dapplion: Fixed with https://github.com/sigp/lighthouse/pull/6497/commits/11f903022

beacon_node/network/src/router.rs
dapplion: Not really, I would actually change this to a let Some() match instead of two matches (in a future PR) and leave as is to minimize diff

beacon_node/network/src/sync/manager.rs
dapplion: Fixed with https://github.com/sigp/lighthouse/pull/6497/commits/20d97941e

beacon_node/network/src/sync/manager.rs
dapplion: Fixed with https://github.com/sigp/lighthouse/pull/6497/commits/20d97941e

beacon_node/network/src/sync/network_context.rs
dapplion: yep https://github.com/sigp/lighthouse/pull/6497/commits/000bc5512

beacon_node/network/src/sync/network_context.rs
dapplion: Yeah! https://github.com/sigp/lighthouse/pull/6497/commits/000bc5512

beacon_node/network/src/sync/network_context.rs
dapplion: 👍  https://github.com/sigp/lighthouse/pull/6497/commits/000bc5512

beacon_node/network/src/sync/network_context.rs
dapplion: Uff nice find! https://github.com/sigp/lighthouse/pull/6497/commits/d99232a97

beacon_node/network/src/sync/network_context.rs
dapplion: https://github.com/sigp/lighthouse/pull/6497/commits/d99232a97

beacon_node/network/src/sync/network_context/requests/blobs_by_range.rs
dapplion: Missing check! Added in https://github.com/sigp/lighthouse/pull/6497/commits/d72b61d7a for both blob and columns

beacon_node/network/src/sync/network_context.rs
dapplion: Done

realbigsean (approved):
Nice, couldn't find any issues

beacon_node/lighthouse_network/src/service/api_types.rs
jimmygchen: Nice, much easier to read. Thanks for the docs

jimmygchen (approved):
Looks great! Thanks

PR #6498, opened 2024-10-16
Generalize sync block lookup tests

dapplion:
## Issue Addressed

As noted in 
- https://github.com/sigp/lighthouse/pull/6497#discussion_r1803877443

The existing range sync tests should adopt the event-driven style of block lookup tests.

## Proposed Changes

This PR just moves the existing tests to a new directory structure so git doesn't trip and retains all the sweet blame in the 3000 lines of code of lookup tests. A future PR will migrate the range sync tests.

```
beacon_node/network/src/sync/tests/
├── lookups.rs
├── mod.rs
└── range.rs
```

The idea is that lookup and range will each `impl TestRig` for what they need to keep code separate. The common functions can either be made `pub(crate)` (if already in `lookups.rs`) or moved to `mod.rs`.


realbigsean (approved):
Looks straightforward

jimmygchen (approved):
Make sense!

jimmygchen:
@mergify queue

PR #6513, opened 2024-10-18
Send `IDONTWANT` on publish to avoid downloading data we already have

jimmygchen:
## Issue Addressed

With the introduction of `getBlobsV1` optimisation, nodes may be publishing blobs if they retrieved it from the EL but haven't received it from gossip. In these scenarios, the node already have the blobs and can save download bandwidth by sending out `IDONTWANT` to peers on the blobs we publish, so that it avoids downloading them from gossip.

This would apply to all messages > 1000 bytes (`idontwant_message_size_threshold`) though, and in most cases we won't be publishing the same messages as other, so perhaps we could also consider adding a `send_idontwant` bool flag to the `behavour.publish` function, and only set it to true when publishing blobs?


AgeManning (approved):

beacon_node/lighthouse_network/gossipsub/src/behaviour.rs
```diff
@@ -626,6 +626,11 @@ where
             return Err(PublishError::Duplicate);
         }
 
+        // Broadcast IDONTWANT messages
```
jxs: at this point we may still fail publishing the message, even if do not end up sending any IDONTWANT we have still iterated redundantly the mesh and `GossipPromises`, was there a particular reason for putting here instead of after publishing Jimmy?

jxs (commented):
Thanks for starting this JImmy, left a comment.
Meanwhile the [spec](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.2.md) is clear in stating that IDONTWANT's should be sent for received messages:
>In more specific terms, a new control message is introduced: IDONTWANT. It's primarily intended to notify mesh peers that the node already received a message and there is no need to send its duplicate.

and:

>When the peer receives the first message instance it immediately broadcasts (not queue for later piggybacking) IDONTWANT with the messageId to all its mesh peers. This could be performed prior to the message validation to further increase the effectiveness of the approach.

can we forward this conversation to the spec and open an issue there proposing to also update the spec?

beacon_node/lighthouse_network/gossipsub/src/behaviour.rs
jimmygchen: Good question - I didn't really put too much thought about the ordering when I did this. I think it's fine to send out `idontwant` even if we fail to publish the message for whatever reasons - we do have the message and don't need to download them from the peers. Does that make sense?

jxs: Thanks for elaborating Jimmy, fair point. Still we should put this after asserting `recipient_peers` is not empty and we published to at least one peer. IDONTWANT's are sent to the mesh peers, if we get `InsuficientPeers` it means the mesh peers is also empty and it's pointless to iterate `mesh_peers` again trying to publish. 
If we also fail to publish to at least one peer it means all handlers have its queues full, which also makes it pointless  to try to publish to them.

jxs (commented):
Re-reading on your comment Jimmy, 
> though, and in most cases we won't be publishing the same messages as other, so perhaps we could also consider adding a send_idontwant bool flag to the behavour.publish function, and only set it to true when publishing blobs

I think this makes sense, we probably should offer this option to users

beacon_node/lighthouse_network/gossipsub/src/behaviour.rs
jimmygchen: Make sense, addressed in https://github.com/sigp/lighthouse/pull/6513/commits/7c624a601a6a7dba266814a29f65ec65be2ec1e7. Thanks!

jimmygchen:
> I think this makes sense, we probably should offer this option to users

I had a chat with Age regarding this and we decided not to add this flag. 

The `idontwant_message_size_threshold` should limit the amount of `IDONTWANT` being sent to peers - if a message exceeds `idontwant_message_size_threshold`, it's probably worth sending an `IDONTWANT`.

Let me know what you think!

jxs (approved):
LGTM thanks Jimmy!

>>I had a chat with Age regarding this and we decided not to add this flag.
>
>The idontwant_message_size_threshold should limit the amount of IDONTWANT being sent to peers - if a message exceeds idontwant_message_size_threshold, it's probably worth sending an IDONTWANT.
>
>Let me know what you think!

Yeah makes sense

pawanjay176 (approved):

jimmygchen:
@mergify queue

PR #6515, opened 2024-10-18
Simplify Validator Creation and Align with Spec

ethDreamer:
## Issue Addressed

The current code for adding new validators doesn't align well with the spec and it's getting increasingly convoluted as changes are made in different forks. This simplifies it and brings it into much closer alignment with the spec.


ethDreamer:
I noticed this while reviewing:

* #6496

which was originally difficult to follow because of how convoluted our approach is getting.

pawanjay176 (approved):
LGTM

pawanjay176:
@mergify queue

PR #6519, opened 2024-10-20
Detect invalid proposer signature on RPC block processing

dapplion:
## Issue Addressed

Complements

by detecting if the proposer signature is valid or not during RPC block processing. In lookup sync, if the invalid signature signature is the proposer signature, it's not deterministic on the block root. So we should only penalize the sending peer and retry. Otherwise, if it's on the body we should drop the lookup and penalize all peers that claim to have imported the block


beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -1102,7 +1111,19 @@ impl<T: BeaconChainTypes> SignatureVerifiedBlock<T> {
                 parent: Some(parent),
             })
         } else {
-            Err(BlockError::InvalidSignature)
+            // Re-verify the proposer signature in isolation to attribute fault
+            let mut signature_verifier = get_signature_verifier(&state, &pubkey_cache, &chain.spec);
```
pawanjay176: does verifying just a single signature with `verify_signature_sets` significantly worse than doing regular single signature verifcation? I don't think it will be anything significant but just checking

pawanjay176 (approved):
LGTM. Just a minor nit

beacon_node/beacon_chain/src/block_verification.rs
dapplion: Good point, looked into the code and `get_signature_verifier` would multiply the single signature by a rand coefficient even if there's a single one so I switched to single sig verification in 6b111b9ba394e0f140339ad4dbbe586c48db692f

pawanjay176 (approved):

pawanjay176:
Tests are failing

dapplion:
This is ready for review. 

Motivation: used by https://github.com/sigp/lighthouse/pull/6321 to not re-download a block that has invalid body signature(s)

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -329,6 +323,17 @@ pub enum BlockError {
     InternalError(String),
 }
 
+/// Which specific signature(s) are invalid in a SignedBeaconBlock
+#[derive(Debug)]
+pub enum InvalidSignature {
+    // The outter signature in a SignedBeaconBlock
```
pawanjay176:
```suggestion
    // The outer signature in a SignedBeaconBlock
```

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -329,6 +323,17 @@ pub enum BlockError {
...
+    ProposerSignature,
+    // One or more signatures in BeaconBlockBody
+    BlockBodySignatures,
+    // One or more signatures in SignedBeaconBlock
```
pawanjay176, line 333: what kind of signature error would be not ProposerSignature and BlockBodySignature. Won't these 2 variants encompass all cases?!

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -652,7 +659,7 @@ pub fn signature_verify_chain_segment<T: BeaconChainTypes>(
     }
 
     if signature_verifier.verify().is_err() {
-        return Err(BlockError::InvalidSignature);
+        return Err(BlockError::InvalidSignature(InvalidSignature::Unknown));
```
pawanjay176, line 662: Why isn't this BlockBodySignatures

pawanjay176 (approved):
Looks good mostly, not sure of the necessity of the Unknown error variant

beacon_node/beacon_chain/src/block_verification.rs
dapplion: This signature set includes all signatures (block body + proposer signature) of multiple blocks. This codepath is used only for sync by range, so it's not useful to figure out if it's the proposer signature. So I set here to unknown to avoid having to re-verify all proposer signatures for no benefit.

beacon_node/beacon_chain/src/block_verification.rs
dapplion: See https://github.com/sigp/lighthouse/pull/6519/files#r1931489262

beacon_node/beacon_chain/src/block_verification.rs
pawanjay176: Okay that makes sense

pawanjay176 (approved):

PR #6520, opened 2024-10-21
Strict match of errors in backfill sync

dapplion:
## Issue Addressed

Same spirit as

Handling errors in sync with a fallback match is not great. In the case of backfill sync it's very easy to fix! =D 

> (tangent) Noting a chat with @pawanjay176 in https://github.com/sigp/lighthouse/pull/6321 `BeaconChainError` actually includes errors that are not internal. In this case it was some of the variants of `HistoricalBlockError`.. I'll review `BeaconChainError` to double check there's no other hidden scorable variants.

## Proposed Changes

Turns out that its processing returns almost only variants of `HistoricalBlockError`, so I changed `import_historical_block_batch` to return `HistoricalBlockError` instead of `BeaconChainError`.

All logic is the same, just made the match on network beacon processor cleaner a bit cleaner.


pawanjay176 (approved):
Neat

jimmygchen:
@mergify queue

PR #6521, opened 2024-10-21
Strict match of errors in range sync

dapplion:
## Issue Addressed

Same spirit as

Handling errors in sync with a fallback match is not great. Range sync requires a big match like lookup sync.

## Proposed Changes

Explicitly match all variants of `BlockError` in range sync. 

Marked as draft to gather feedback on the correctness of my assumptions on what to match each variant to.


dapplion:
- Closing for stale for https://github.com/sigp/lighthouse/issues/7129

PR #6524, opened 2024-10-21
Clarify command-line reference docs

michaelsproul:
## Proposed Changes

Change the heading for the CLI docs to _Command Line Reference (CLI)_. I think this name is better than the current _Built-In Documentation_ title which does not obviously convey that these docs have anything to do with the CLI.

## Additional Info

I've also deleted the old `cli.md` file which was unused.


chong-he:
LGTM!

jimmygchen (approved):

book/src/SUMMARY.md
```diff
@@ -54,13 +54,13 @@
   * [Merge Migration](./merge-migration.md)
   * [Late Block Re-orgs](./late-block-re-orgs.md)
   * [Blobs](./advanced-blobs.md)
-* [Built-In Documentation](./help_general.md)
+* [Command Line Reference (CLI)](./help_general.md)
```
jimmygchen, line 57: Btw the heading for the doc is `Lighthouse General Commands`, do we want to make it match the link too?

michaelsproul: good idea!

I changed it, but it doesn't match exactly because I thought it would be nice to keep "Lighthouse" in the page title

michaelsproul: a6540f9d67332174e0df4e69e93c7dc331635a9d

michaelsproul:
@mergify queue

Issue #4546, opened 2023-07-28
Penalize peers for sending us invalid rpc blobs


pawanjay176:
## Description

In the availability check in the block processing flow, getting the `BlockError::AvailabilityCheck` should always imply either:

1. The block proposer is faulty (sent inconsistent blocks and blobs over gossip)
2. We received the blob over rpc and it is invalid (inconsistent w.r.t the block).
3. It is an internal error (kzg setup not initialized/ db error etc)

In the second case, we need to penalize the right peers (those that sent us the incorrect blobs). This would involve storing the peer_id of the peer that sends us the rpc blobs until it has been verified against the blob.

PR #6525, opened 2024-10-21
Penalize peer if missing blobs

dapplion:
## Issue Addressed

During range sync, if a peer replies with empty blobs we do not downscore that peer. We should.

## Proposed Changes

Downscore peer if we can't match block and data when constructing an RpcBlock

The PRs below also move these two variants into not being internal
- https://github.com/sigp/lighthouse/pull/6521

### DIFF
--- a/beacon_node/beacon_chain/src/data_availability_checker/error.rs
+++ b/beacon_node/beacon_chain/src/data_availability_checker/error.rs
@@ -37,8 +37,6 @@ impl Error {
     pub fn category(&self) -> ErrorCategory {
         match self {
             Error::SszTypes(_)
-            | Error::MissingBlobs
-            | Error::MissingCustodyColumns
             | Error::StoreError(_)
             | Error::DecodeError(_)
             | Error::Unexpected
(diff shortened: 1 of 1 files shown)


beacon_node/beacon_chain/src/data_availability_checker/error.rs
```diff
@@ -49,6 +47,8 @@ impl Error {
             | Error::SlotClockError => ErrorCategory::Internal,
             Error::InvalidBlobs { .. }
             | Error::InvalidColumn { .. }
+            | Error::MissingBlobs
+            | Error::MissingCustodyColumns
```
jimmygchen, line 51: The category change makes sense but it doesn't look like we use this to determine peer penalty.

Related issue:
https://github.com/sigp/lighthouse/issues/4546

I think we'll need to handle `BlockError::AvailabilityCheck` in `handle_failed_chain_segment`:
https://github.com/sigp/lighthouse/blob/ee7fca3ebd277a58739c7a19046608b7556afc59/beacon_node/network/src/network_beacon_processor/sync_methods.rs#L711

jxs: it seems we use Jimmy, `category()` is called on:
https://github.com/sigp/lighthouse/blob/2e0eb6d1b8705bbda2ba56eb195d9cc7c6575e95/beacon_node/network/src/sync/block_lookups/mod.rs#L580-L605

with Lion's change `Error::MissingBlobs` and `Error::MissingCustodyColumns` are now reported as `ErrorCategory::Malicious` instead of `ErrorCategory::Internal` which makes it so that those nodes get reported via `report_peer`

jimmygchen:
This is now covered in #7352, closing.

PR #6526, opened 2024-10-21
Fix BlobsByRange by reverting PR6462

michaelsproul:
## Issue Addressed

This reverts PR:

- #6462

That PR accidentally broke `BlobsByRange` requests by adding a non-standard field to the SSZ representation (sent over the network). The result is that peers would not respond to Lighthouse's `BlobsByRange` requests and Lighthouse would be unable to sync.


jimmygchen (approved):
LGTM! Nice find 🎉 
We can revisit the change once https://github.com/sigp/lighthouse/pull/6329 is merged.

dapplion (approved):

jimmygchen:
@mergify queue

michaelsproul:
@mergify requeue

PR #6531, opened 2024-10-21
Use `make cli-local` in CI test suite to remove redundant docker

jimmygchen:
## Issue Addressed

I noticed that our `cli` job sometimes take quite a while and take a few `docker pull` retries before it completes. This is because it attempts to pull the 2.2 GB `github-runner` image to run the the cli script in a docker container.

I think we can safely replace the `cli` command to run with `cargo` directly instead of running it with a `github-runner` image, as the job runs in a container running the `github-runner` image, so this is a bit redundant and slow.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,15 +183,9 @@ test-exec-engine:
 # test vectors.
 test: test-release
 
-# Updates the CLI help text pages in the Lighthouse book, building with Docker.
-cli:
-	docker run --rm --user=root \
-	-v ${PWD}:/home/runner/actions-runner/lighthouse sigmaprime/github-runner \
-	bash -c 'cd lighthouse && make && ./scripts/cli.sh'
-
 # Updates the CLI help text pages in the Lighthouse book, building using local
 # `cargo`.
-cli-local:
+cli:
 	make && ./scripts/cli.sh
 
 # Check for markdown files
--- a/Makefile
+++ b/Makefile
@@ -183,9 +183,15 @@ test-exec-engine:
 # test vectors.
 test: test-release
 
+# Updates the CLI help text pages in the Lighthouse book, building with Docker (primarily for window users).
+cli:
+	docker run --rm --user=root \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -183,15 +183,9 @@ test-exec-engine:
 # test vectors.
 test: test-release
 
-# Updates the CLI help text pages in the Lighthouse book, building with Docker.
-cli:
```
michaelsproul, line 187: I think this mainly exists to help Windows users who can't execute the bash script locally. Maybe we could just change CI to call `make cli-local`?

chong-he: Yes the docker command is to support Windows users. I just tested that it is still working in Windows. 

Agree about changing the CI from `make cli` to `make cli-local` here:

https://github.com/sigp/lighthouse/blob/2e0eb6d1b8705bbda2ba56eb195d9cc7c6575e95/.github/workflows/test-suite.yml#L398

jimmygchen: Ah I see. I've updated CI to use `cli-local`

Makefile
```diff
@@ -183,7 +183,7 @@ test-exec-engine:
 # test vectors.
 test: test-release
 
-# Updates the CLI help text pages in the Lighthouse book, building with Docker.
+# Updates the CLI help text pages in the Lighthouse book, building with Docker (primarily for window users).
```
macladson:
```suggestion
# Updates the CLI help text pages in the Lighthouse book, building with Docker (primarily for Windows users).
```

macladson (approved):
Approved pending one smol nit sorry.
Tbh, I'm not sure why we weren't using `make cli-local` in CI in the first place :sweat_smile:

macladson (commented):
Oh also, the title isn't really accurate anymore, I'll change it to something like "Use make cli-local in CI"

jimmygchen:
@mergify queue

PR #6534, opened 2024-10-22
RequestType::max_responses for LightClientUpdatesByRange

ackintosh:
## Issue Addressed

In case of `RequestType::LightClientUpdatesByRange`, `RequestType::max_responses()` returns  the number that  [`LightClientUpdatesByRangeRequest::max_requested()`](https://github.com/sigp/lighthouse/blob/9aefb5539baff637d68deb3dd386ff45312f3573/beacon_node/lighthouse_network/src/rpc/methods.rs#L491-L493) returns, which is a constant value representing [the maximum number of LightClientUpdate instances in a single request](https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/p2p-interface.md#configuration). I think it should return the actual number of instances the request requires, instead of the constant value defined in the spec.

### DIFF
--- a/beacon_node/lighthouse_network/src/rpc/protocol.rs
+++ b/beacon_node/lighthouse_network/src/rpc/protocol.rs
@@ -791,7 +791,7 @@ impl<E: EthSpec> RequestType<E> {
             RequestType::LightClientBootstrap(_) => 1,
             RequestType::LightClientOptimisticUpdate => 1,
             RequestType::LightClientFinalityUpdate => 1,
-            RequestType::LightClientUpdatesByRange(req) => req.max_requested(),
+            RequestType::LightClientUpdatesByRange(req) => req.count,
         }
     }
 


jimmygchen (approved):
Good catch!

This is also consistent with [the spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/p2p-interface.md#lightclientupdatesbyrange), with the max size enforced in the [handling method](https://github.com/sigp/lighthouse/blob/9aefb5539baff637d68deb3dd386ff45312f3573/beacon_node/network/src/network_beacon_processor/rpc_methods.rs#L474):

> The response MUST NOT contain more than min(MAX_REQUEST_LIGHT_CLIENT_UPDATES, count) results

jimmygchen:
@mergify queue

jimmygchen:
Failed due to `crates.io` download issue, will requeue.

jimmygchen:
@mergify requeue

PR #6537, opened 2024-10-23
Add 'beacon_' prefix to PeerDAS metrics names

KatyaRyazantseva:
## Issue Addressed

This PR addresses issue #6018, the list of the PeerDAS metrics. [PeerDAS metrics specs PR](https://github.com/ethereum/beacon-metrics/pull/14)

## Proposed Changes

Added `beacon_` prefix to metrics names to align with PeerDAS metrics specs.


beacon_node/beacon_chain/src/metrics.rs
```diff
@@ -1812,7 +1812,7 @@ pub static KZG_VERIFICATION_BATCH_TIMES: LazyLock<Result<Histogram>> = LazyLock:
 pub static KZG_VERIFICATION_DATA_COLUMN_SINGLE_TIMES: LazyLock<Result<Histogram>> =
     LazyLock::new(|| {
         try_create_histogram_with_buckets(
-            "kzg_verification_data_column_single_seconds",
+            "beacon_kzg_verification_data_column_single_seconds",
             "Runtime of single data column kzg verification",
```
jimmygchen, line 1851: Note that we also have similar metrics (e.g. `kzg_verification_single_seconds`) for Deneb blobs, but we can probably leave them as it is, as they will eventually be replaced by the PeerDAS ones.

beacon_node/beacon_chain/src/metrics.rs
```diff
@@ -1875,14 +1875,14 @@ pub static DATA_AVAILABILITY_OVERFLOW_STORE_CACHE_SIZE: LazyLock<Result<IntGauge
...
         )
     });
 pub static DATA_AVAILABILITY_RECONSTRUCTED_COLUMNS: LazyLock<Result<IntCounter>> =
     LazyLock::new(|| {
         try_create_int_counter(
-            "data_availability_reconstructed_columns_total",
+            "beacon_data_availability_reconstructed_columns_total",
             "Total count of reconstructed columns",
         )
     });
```
jimmygchen, line 1923: Could you also update the 4 data column metrics below to include the prefix too?

jimmygchen (changes requested):
Thanks @KatyaRyazantseva, the prefix make sense, just one minor comment here to update a few other related metrics.

beacon_node/beacon_chain/src/metrics.rs
KatyaRyazantseva: Found 3 `kzg_data_column` metrics below. What about `data_availability_cache_size` metrics above?

beacon_node/beacon_chain/src/metrics.rs
KatyaRyazantseva: Found 3 `kzg_data_column` metrics below. What about `data_availability_cache_size` metrics above?

jimmygchen: The cache size metrics are Lighthouse specific and they are existing metrics - so I think we can leave them as it is.

jimmygchen (approved):
Looks good to me. Thanks @KatyaRyazantseva !

jimmygchen:
@mergify queue

PR #6541, opened 2024-10-26
Track beacon processor import result metrics

dapplion:
## Issue Addressed

Our current metrics to track the result of block imports are incomplete. A block import can be triggered by any block component, both from gossip and RPC. We should metric the `AvailabilityProcessingStatus` of block, blob, and custody column import to count all possible import events.

Also, I would like to track all errors that can occur during import. For example `BlockError::DuplicateFullyImported`, which show the efficiency of sync.

## Proposed Changes

Track the return of all `process_*` functions with the same helper `register_process_result_metrics`. Track the result `FullyImported` / `MissingComponents` / `Error` labeled by block component and provenance. 

## Additional Info

This PR has no breaking changes w.r.t. current metrics. However, it fixes the following metric:
- `beacon_processor_rpc_block_imported_total`: This metric is incorrect. Blocks with blobs synced by lookup sync are never tracked, as the block is always imported first and blobs second.

### DIFF
--- a/beacon_node/network/src/metrics.rs
+++ b/beacon_node/network/src/metrics.rs
@@ -2,7 +2,8 @@ use beacon_chain::{
     attestation_verification::Error as AttnError,
     light_client_finality_update_verification::Error as LightClientFinalityUpdateError,
     light_client_optimistic_update_verification::Error as LightClientOptimisticUpdateError,
-    sync_committee_verification::Error as SyncCommitteeError,
+    sync_committee_verification::Error as SyncCommitteeError, AvailabilityProcessingStatus,
+    BlockError,
 };
 use fnv::FnvHashMap;
 use lighthouse_network::{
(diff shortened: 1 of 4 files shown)


beacon_node/network/src/metrics.rs
```diff
@@ -59,6 +67,27 @@ pub static SYNC_COMMITTEE_SUBSCRIPTION_REQUESTS: LazyLock<Result<IntCounter>> =
...
+            "Total number of imported individual block components that resulted in missing components",
+            &["source", "component"],
+        )
+    },
+);
+pub static BEACON_PROCESSOR_IMPORT_ERRORS_PER_TYPE: LazyLock<Result<IntCounterVec>> =
+    LazyLock::new(|| {
+        try_create_int_counter_vec(
+            "beacon_processor_import_errors_total",
+            "Total number of block components verified",
```
jimmygchen: This should be `Total number of block components that was not verified`

jimmygchen (changes requested):
Nice, just a nit on the metric description and some merge conflicts, otherwise looks good!

jimmygchen (approved):
Looks good thanks!

jimmygchen:
@mergify queue

PR #6553, opened 2024-10-31
Revise Siren documentation

chong-he:
## Proposed Changes

Update Siren documentation in Lighthouse book. There are quite some changes compared to the original doc so I will leave some comments to sections to highlight the change. I have also combined `Configuration` and `Installation` together as they are quite related to start Siren.


book/src/ui-configuration.md
```diff
@@ -1,37 +1,109 @@
...
+
+### About self-signed SSL certificates
+
+By default, Siren will generate and use a self-signed certificate on startup.
+This will generate a security warning when you try to access the interface.
+We recommend to only disable SSL if you would access Siren over a local LAN or otherwise highly trusted or encrypted network (i.e. VPN).
+
+#### Generating persistent SSL certificates and installing them to your system
 
-> To access from another machine on the same network (192.168.0.200) set the Beacon Node and Validator Client `--http-address` as `192.168.0.200`. When this is set, the validator client requires the flag `--beacon-nodes http://192.168.0.200:5052` to connect to the beacon node.
```
chong-he, line 17: I am not sure about this part. I think this is not necessary if we can connect Siren via SSH. Proposing to remove this to make things simpler.

book/src/ui-configuration.md
```diff
@@ -1,37 +1,109 @@
...
+    ```
+    BEACON_URL=http://localhost:5052
+    VALIDATOR_URL=http://localhost:5062
+    API_TOKEN=R6YhbDO6gKjNMydtZHcaCovFbQ0izq5Hk
+    SESSION_PASSWORD=your_password
+    ```
+
+ 1. You can now start Siren with:
+    ```bash
+    docker run --rm -ti --name siren -p 4443:443 --env-file $PWD/.env --net host sigp/siren
```
chong-he: I still don't get the port 4443:443. Removing the flag `-p 4443:443` still works, though with a warning:
` metadata.metadataBase is not set for resolving social open graph or twitter images, using "http://localhost:3000". See https://nextjs.org/docs/app/api-reference/functions/generate-metadata#metadatabase`

Should we keep this? I am not sure if we keep this, but we haven't used it anywhere else later. Please feel free to correct me.

antondlr: `--net host` is the culprit here: it automatically exposes all ports to the host (including port 80)
it is really bad to advise it as default as it is hugely unsafe. 
for some reason it it necessary for people who installed docker though a `snap` though.

book/src/ui-configuration.md
antondlr: The aim for Siren is to be as easily accessible as possible, asking them to expose a port over a reverse ssh tunnel is way more complicated. Just exposing the port and accessing the webapp directly that way fits better in the `end-user application`

book/src/ui-authentication.md
```diff
@@ -2,12 +2,12 @@
 
 ## Siren Session
 
-For enhanced security, Siren will require users to authenticate with their session password to access the dashboard. This is crucial because Siren now includes features that can permanently alter the status of user validators. The session password must be set during the [installation](./ui-installation.md) process before running the Docker or local build, either in an `.env` file or via Docker flags.
+For enhanced security, Siren will require users to authenticate with their session password to access the dashboard. This is crucial because Siren now includes features that can permanently alter the status of user validators. The session password must be set during the [configuration](./ui-configuration.md) process before running the Docker or local build, either in an `.env` file or via Docker flags.
```
antondlr:
```suggestion
For enhanced security, Siren will require users to authenticate with their session password to access the dashboard. This is crucial because Siren now includes features that can permanently alter the status of the user's validators. The session password must be set during the [configuration](./ui-configuration.md) process before running the Docker or local build, either in an `.env` file or via Docker flags.
```

book/src/ui-configuration.md
```diff
@@ -1,37 +1,117 @@
-# Configuration
+# 📦 Installation
 
-Siren requires a connection to both a Lighthouse Validator Client and a Lighthouse Beacon Node.  
-To enable connection, you must generate .env file based on the provided .env.example
+Siren supports any operating system that supports container runtimes and/or NodeJS 18, this includes Linux, MacOS, and Windows. The recommended way of running Siren is by launching the [docker container](https://hub.docker.com/r/sigp/siren).
```
antondlr:
```suggestion
Siren supports any operating system that supports container runtimes and/or NodeJS 18, this includes Linux, MacOS, and Windows. The recommended way of running Siren is by launching the [docker container](https://hub.docker.com/r/sigp/siren).  
Ubuntu users: Please note that Docker installed as a snap is not ideal; the preferred way to install Docker on Ubuntu is outlined [here.](https://docs.docker.com/engine/install/ubuntu/)
```

book/src/ui-configuration.md
```diff
@@ -1,37 +1,117 @@
...
+These ports should be accessible from Siren. This means adding the flag `--http` on both beacon node and validator client.
 
 To enable the HTTP API for the beacon node, utilize the `--gui` CLI flag. This action ensures that the HTTP API can be accessed by other software on the same machine.
 
 > The Beacon Node must be run with the `--gui` flag set.
 
-If you require accessibility from another machine within the network, configure the `--http-address` to match the local LAN IP of the system running the Beacon Node and Validator Client.
+## Running the Docker container (Recommended)
+
+The common usage is to run Siren at a client computer connecting to a server running the node. The following guide is to setup Siren at a client computer and connect it to the server via SSH, so that Siren can be accessed and viewed from the client computer's browser.
```
antondlr: this is 100% not the case anymore with the "new" dockerised setup. 
the preferred way to run it nowadays is to have the docker container as close to the BN and VC as possible, running all the time, with its https port exposed to local LAN. 
(this was the whole reason of the architectural overahaul)

book/src/ui-configuration.md
```diff
@@ -1,37 +1,117 @@
...
+
+    ```
+    validator api issue, server response: 403
+    ```
+
+    which means that the API token is incorrect. Check that you have provided the correct token in the field `API_TOKEN` in `.env`.
+
+    When Siren is successfully run, you should see the log `LOG [NestApplication] Nest application successfully started +118ms`, indicating that Siren has started.
+
+ 1. On the client computer, SSH to the server:
```
antondlr: same comment as before; no ssh should be necessary. 
expose the https port on the node, access over https.

antondlr (commented):
if you can wait a couple days, I can make suggestion commits for the items I stipulated (https instead of ssh). 
and/or if you're confused we can have a call to discuss it, so I can explain.

book/src/ui-configuration.md
chong-he: The extra space will not pass the markdown linter so I resolved the conversation with no changes.

book/src/ui-configuration.md
chong-he: Revised in [2c6cd76](https://github.com/sigp/lighthouse/pull/6553/commits/2c6cd76a7e7580ebb69ca56510b448417ff91fc1)

book/src/ui-configuration.md
chong-he: As above

book/src/ui-configuration.md
antondlr: as we discussed IRL, because so many people will be running plain (non docker) lighthouse by following the installation guides, we'll go with `--net=host`.  The alternative would be that people have to allow access to their BN and VC API's on any network interface instead of the loopback address, which is a bigger security risk.  
I don't love this solution but it's the lesser of two evils, I think.

antondlr (approved):
lgtm!

chong-he:
As discussed with @antondlr, waiting for some testing and a final check before this is complete

michaelsproul:
Almost ready to merge this one?

chong-he:
> Almost ready to merge this one?

Got some tests and final checking pending, will try to get it done soon

antondlr (approved):
let's go!!

there's some stuff that will need rephrasing after magick93's pr gets merged but let's leave that out of scope for now

michaelsproul:
@mergify queue

PR #6554, opened 2024-10-31
Cargo update

AgeManning:
## Issue Addressed

Cargo update currently breaks something. Hoping for some help to fix. See CI


jimmygchen:
`local-testnet` issues are kurtosis related, addressed in https://github.com/sigp/lighthouse/pull/6555
Looking into sim tests now

jimmygchen:
Sim tests and local testnet tests are all passing now! 🎉

michaelsproul:
Maybe we can do a targeted cargo update of the crates we need to update? Doing wholesale `cargo update` isn't necessary for v6.0.0, especially if it's slowing down node startup

michaelsproul:
I've isolated the issue to the update of rust_eth_kzg. Opened this PR to pin it:

- https://github.com/sigp/lighthouse/pull/6608

PR #6555, opened 2024-10-31
Pin `kurtosis-cli` version

jimmygchen:
## Issue Addressed

CI seems to be failing everywhere we use Kurtosis.

There was a release yesterday (`1.4.0`), will try using an old version to see if it works.


jimmygchen:
Looks like 1.3.1 works!

michaelsproul (approved):

michaelsproul:
@mergify queue

PR #6558, opened 2024-11-01
Add error message for duration subtraction overflow in sim tests

jimmygchen:
## Issue Addressed

When our node startup time increases, sim tests may fail, as the `GENESIS_DELAY` is no longer sufficient and the tests requires all nodes to be started before genesis. 

When this happens, we get an error message on CI, which isn't very obvious - this PR adds some error logging:

```
thread 'main' panicked at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/time.rs:1141:31:
overflow when subtracting durations
```

Failing job:
https://github.com/sigp/lighthouse/actions/runs/11606498668/job/32318612076

### DIFF
--- a/testing/simulator/src/basic_sim.rs
+++ b/testing/simulator/src/basic_sim.rs
@@ -206,7 +206,7 @@ pub fn run_basic_sim(matches: &ArgMatches) -> Result<(), String> {
             node.server.all_payloads_valid();
         });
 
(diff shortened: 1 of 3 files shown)


macladson (approved):
Love it, very nice change! :pray:

AgeManning:
@mergifyio queue

PR #6561, opened 2024-11-02
Fix doc-test in `consensus` crate

macladson:
## Issue Addressed

I ran `make test` recently and noticed I was getting a doc-test failure in the `consensus` crate.

## Proposed Changes

Updates the import inside the doc test to use the correct crate.

## Additional Info

Curious to know why CI didn't catch this. To me that seems to indicate that `make nextest-release` doesn't cover the same set of tests as `make test`

### DIFF
--- a/consensus/types/src/runtime_var_list.rs
+++ b/consensus/types/src/runtime_var_list.rs
@@ -13,7 +13,7 @@ use std::slice::SliceIndex;
 /// ## Example
 ///
 /// ```
-/// use ssz_types::{RuntimeVariableList};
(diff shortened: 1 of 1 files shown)


jxs (approved):
LGTM, thanks!

michaelsproul:
@mergify queue

michaelsproul:
Good catch.

Based on a quick search it looks like nextest doesn't support running doctests, so maybe we should add a separate step just for them. Although I can imagine this might be almost 2x slower :\

- https://github.com/nextest-rs/nextest/issues/16

PR #6566, opened 2024-11-05
Deprecate `eth1` and `dummy-eth1` flags

michaelsproul:
## Proposed Changes

Deprecate the following beacon node flags:

- `--eth1`: No longer required because it is enabled by default. The deposit contract sync can be disabled via a separate flag `--disable-deposit-contract-sync`, which has existed since the merge.
- `--dummy-eth1`: Very rarely used and of limited utility.

### DIFF
--- a/beacon_node/client/src/config.rs
+++ b/beacon_node/client/src/config.rs
@@ -59,10 +59,6 @@ pub struct Config {
     /// Path where the blobs database will be located if blobs should be in a separate database.
     pub blobs_db_path: Option<PathBuf>,
     pub log_file: PathBuf,
-    /// If true, the node will use co-ordinated junk for eth1 values.
-    ///
-    /// This is the method used for the 2019 client interop in Canada.
-    pub dummy_eth1_backend: bool,
     pub sync_eth1_chain: bool,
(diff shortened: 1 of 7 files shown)


eserilev (commented):
nit 

https://github.com/sigp/lighthouse/blob/d10e81afda2ddea80b19daf9b996d78451186a9b/beacon_node/src/config.rs#L285-L294

since `sync_eth1_chain` is now `true` by default, I think this line is redundant. Also the comments about the deprecated `--eth-1` flag might no longer be relevant. 

EDIT:

also this line here may be redundant

https://github.com/sigp/lighthouse/blob/d10e81afda2ddea80b19daf9b996d78451186a9b/beacon_node/src/config.rs#L124

michaelsproul:
Nice catch, thanks @eserilev. I've removed that redundant code and also added a `conflicts_with` for `--staking` and `--disable-deposit-contract-sync`. Other than the `conflicts_with` clause, `staking` is basically identical to `http` now, but this may not remain the case forever, and I think our users appreciate this flag.

eserilev (approved):
LGTM!

michaelsproul:
@mergify queue

PR #6569, opened 2024-11-07
Fix rpc decoding for blobs by range/root

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Handle rpc responses for blobs correctly for electra. Add tests


jimmygchen (approved):
Nice catch! Good idea keeping the fork name matching for blobs as we may need to update it for the F* fork.

jimmygchen:
@mergify queue

Issue #5391, opened 2024-03-11
Import expired blobs


realbigsean:
## Description

- initializing a new archive node that wishes to store all blobs back to the deneb fork will need a way to import blobs into the db
- syncing them over the network might place an undue burden on archive nodes because you'd ban all non-archive peers (unless we make peer scoring changes around this). This might not work well at all if there are too few archive nodes. It *might* work just by updating `min_epochs_for_blob_sidecars_requests` to `u64::MAX`
- a better option would be to add a `--blob-provider` flag that points to an archive node's beacon API
- we could also potentially support importing a blobs db from a file

realbigsean:
we should try to include *an* option for this in the next release because I think it'll be an important feature once blobs start expiring on mainnet. I'm not sure which path is easier/more useful so would appreciate feedback. We could end up implementing both, but I think starting with one for the next release makes sense

**RPC download**

- if `--blob-provider $URL` is added, always download via beacon API up until the configured prune boundary
- could this run as a process mostly independent of sync? It would just be required to lag sync
  - how would this handle a scenario where blob data is invalid or unavailable - cause a crash?

**DB Import export**

When lighthouse is shut off:
- `lighthouse db export-blobs --file $FILENAME`
- `lighthouse db import-blobs --file $FILENAME`

On startup: 
- should lighthouse identify the db changes and go back to verify old blobs if the node is synced?
- should this only be possible on a fresh sync?
- do we need to add a flag `--archive-blobs` ? Or should this automatically be identified
- LH will need to be aware of old blobs during sync so as to not duplicate download and writes, and to verify blocks older than the prune boundary against local blobs
- how should gaps in the db here be handled if they are older than the prune boundary?

jonathanudd:
Any updates regarding this?

michaelsproul:
@jonathanudd Not yet sorry. We've been blocked a bit on implementing PeerDAS, but now that the bulk of that code is in `unstable` we can set to work on a future-compatible design here.

One complication is that with PeerDAS most nodes will cease to store whole blobs, so only supernodes (nodes that opt to store all blobs) will be able to implement the HTTP API for fetching/exporting blobs. Every other node will just have fragments ("columns").

[...]

PR #6570, opened 2024-11-07
MVP impl of `import-blobs`

michaelsproul:
## Issue Addressed

Closes:

- https://github.com/sigp/lighthouse/issues/5391

This is just a quick interim solution to allow blobs to be imported from an existing Lighthouse DB. We will continue working on a more general solution.

## Proposed Changes

New commands:

- `lighthouse db import-blobs --source-db /path/to/blobs_db`: import blobs from another node.
- `lighthouse db inspect-blobs`: iterate all blobs in the node's DB and look for gaps.
- `lighthouse db set-oldest-blob-slot`: manually hack the `oldest_blob_slot` (shouldn't be necessary if using `import-blobs`).


michaelsproul:
Recommended usage:

```
./lighthouse db import-blobs --datadir ./datadir/ --source-db ./blobs_to_import --debug-level debug
```

Running as the user with the correct permissions for both `./datadir ` and `./blobs_to_import`

michaelsproul:
Closing in favour of:

- https://github.com/sigp/lighthouse/pull/6656

PR #6574, opened 2024-11-07
Remove `yq` installation on CI

jimmygchen:
## Issue Addressed

The `yq` PPA repo has been down for a while today and we've been unable to get a passing CI.

~~Recently we changed our doc to suggest using `snap` (instead of `apt`) to install `yq`. (https://github.com/sigp/lighthouse/pull/6468) Michael also ran into some issues with the `yq` version installed with `apt`, so I thought it might be a good time to also switch our CI over to use `snap`.~~

**UPDATE:** it turns out that GitHub Runner images come with `yq` preinstalled, so I've removed the installation altogether.

### DIFF
--- a/.github/workflows/local-testnet.yml
+++ b/.github/workflows/local-testnet.yml
@@ -36,11 +36,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Install dependencies
+      - name: Install Kurtosis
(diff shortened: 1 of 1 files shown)


macladson (approved):
LGTM!

michaelsproul:
@mergify queue

PR #6578, opened 2024-11-11
Add additional metrics for idontwant

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Add the following metrics for idontwant messages:
1. Number of idontwant messages sent per topic
2. Number of messages we received that we had previously sent idontwants for

### DIFF
--- a/beacon_node/lighthouse_network/gossipsub/src/behaviour.rs
+++ b/beacon_node/lighthouse_network/gossipsub/src/behaviour.rs
@@ -1385,7 +1385,7 @@ where
                         "IWANT: Peer has asked for message too many times; ignoring request"
                     );
                 } else if let Some(peer) = &mut self.connected_peers.get_mut(peer_id) {
-                    if peer.dont_send.get(&id).is_some() {
+                    if peer.dont_send_received.get(&id).is_some() {
                         tracing::debug!(%peer_id, message=%id, "Peer already sent IDONTWANT for this message");
                         continue;
                     }
@@ -1817,10 +1817,10 @@ where
         // Calculate the message id on the transformed data.
         let msg_id = self.config.message_id(&message);
 
-        if let Some(peer) = self.connected_peers.get_mut(propagation_source) {
-            // Record if we received a message that we already sent a IDONTWANT for
-            if peer.dont_send.contains_key(&msg_id) {
-                if let Some(metrics) = self.metrics.as_mut() {
+        if let Some(metrics) = self.metrics.as_mut() {
+            if let Some(peer) = self.connected_peers.get_mut(propagation_source) {
+                // Record if we received a message that we already sent a IDONTWANT for to the peer
+                if peer.dont_send_sent.contains_key(&msg_id) {
                     metrics.register_idontwant_messages_ignored_per_topic(&raw_message.topic);
                 }
             }
@@ -2521,11 +2521,18 @@ where
 
         // Flush stale IDONTWANTs.
         for peer in self.connected_peers.values_mut() {
-            while let Some((_front, instant)) = peer.dont_send.front() {
+            while let Some((_front, instant)) = peer.dont_send_received.front() {
                 if (*instant + IDONTWANT_TIMEOUT) >= Instant::now() {
                     break;
                 } else {
-                    peer.dont_send.pop_front();
+                    peer.dont_send_received.pop_front();
+                }
+            }
+            while let Some((_front, instant)) = peer.dont_send_sent.front() {
+                if (*instant + IDONTWANT_TIMEOUT) >= Instant::now() {
+                    break;
+                } else {
+                    peer.dont_send_sent.pop_front();
                 }
             }
         }
@@ -2761,7 +2768,8 @@ where
                     .or_default()
                     .non_priority += 1;
             } else {
-                // IDONTWANT sent successfully
+                // IDONTWANT sent successfully.
+                peer.dont_send_sent.insert(msg_id.clone(), Instant::now());
                 if let Some(metrics) = self.metrics.as_mut() {
                     metrics.register_idontwant_messages_sent_per_topic(&message.topic);
                 }
@@ -2822,7 +2830,7 @@ where
         if !recipient_peers.is_empty() {
             for peer_id in recipient_peers.iter() {
                 if let Some(peer) = self.connected_peers.get_mut(peer_id) {
-                    if peer.dont_send.get(msg_id).is_some() {
+                    if peer.dont_send_received.get(msg_id).is_some() {
                         tracing::debug!(%peer_id, message=%msg_id, "Peer doesn't want message");
                         continue;
                     }
@@ -3176,7 +3184,8 @@ where
                 connections: vec![],
                 sender: RpcSender::new(self.config.connection_handler_queue_len()),
                 topics: Default::default(),
-                dont_send: LinkedHashMap::new(),
+                dont_send_received: LinkedHashMap::new(),
+                dont_send_sent: LinkedHashMap::new(),
             });
         // Add the new connection
         connected_peer.connections.push(connection_id);
@@ -3208,7 +3217,8 @@ where
                 connections: vec![],
                 sender: RpcSender::new(self.config.connection_handler_queue_len()),
                 topics: Default::default(),
-                dont_send: LinkedHashMap::new(),
+                dont_send_received: LinkedHashMap::new(),
+                dont_send_sent: LinkedHashMap::new(),
             });
         // Add the new connection
         connected_peer.connections.push(connection_id);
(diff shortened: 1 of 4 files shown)


beacon_node/lighthouse_network/gossipsub/src/behaviour.rs
```diff
@@ -1817,6 +1817,15 @@ where
         // Calculate the message id on the transformed data.
         let msg_id = self.config.message_id(&message);
 
+        if let Some(peer) = self.connected_peers.get_mut(propagation_source) {
+            // Record if we received a message that we already sent a IDONTWANT for
+            if peer.dont_send.contains_key(&msg_id) {
```
jxs: `dont_send` contains the `IDONWANT` messages the peer sent us, so that we can avoid sending them messages they have sent us `IDONTWANT` for. 
Here you are checking if the peer who sent us the message, has also sent previously an `IDONTWANT` message Pawan.

we don't currently keep track of the `IDONTWANT` messages we sent to which peers, you could do it by adapting the ` idontwant_messages_sent_per_topic` you added in this PR to also contain the peers besides the topic. Can you elaborate on why are you adding this Pawan? Is it for scoring purposes?

Nonetheless we should check first if we have metrics enabled, so that we don't spend iterations that are later discarded because the user doesn't have metrics.

beacon_node/lighthouse_network/gossipsub/src/behaviour.rs
```diff
@@ -2751,6 +2760,11 @@ where
                     .entry(*peer_id)
                     .or_default()
                     .non_priority += 1;
+            } else {
+                // IDONTWANT sent successfully
```
jxs:
```suggestion
                // IDONTWANT sent successfully.
```

beacon_node/lighthouse_network/gossipsub/src/metrics.rs
```diff
@@ -188,6 +188,12 @@ pub(crate) struct Metrics {
     /// The number of bytes we have received in every IDONTWANT control message.
     idontwant_bytes: Counter,
 
+    /// Number of IDONTWANT messages sent per topic
```
jxs:
```suggestion
    /// Number of IDONTWANT messages sent per topic.
```

beacon_node/lighthouse_network/gossipsub/src/metrics.rs
```diff
@@ -188,6 +188,12 @@ pub(crate) struct Metrics {
...
+    idontwant_messages_sent_per_topic: Family<TopicHash, Counter>,
+
+    /// Number of full messages we received that we previously sent a IDONTWANT for
```
jxs:
```suggestion
    /// Number of full messages we received that we previously sent a IDONTWANT for.
```

jxs:
```suggestion
    /// Number of full messages we received that we previously sent a IDONTWANT for.
```

beacon_node/lighthouse_network/gossipsub/src/metrics.rs
```diff
@@ -341,6 +347,18 @@ impl Metrics {
             metric
         };
 
+        // idontwant messages sent per topic
```
jxs:
```suggestion
        // IDONTWANT messages sent per topic
```

beacon_node/lighthouse_network/gossipsub/src/metrics.rs
```diff
@@ -341,6 +347,18 @@ impl Metrics {
...
+        let idontwant_messages_sent_per_topic = register_family!(
+            "idonttwant_messages_sent_per_topic",
+            "Number of IDONTWANT messages sent per topic"
+        );
+
+        // idontwant messages sent but message received per topic
```
jxs:
```suggestion
        // IDONTWANT messages sent but message received per topic
```

jxs (commented):
Hi Pawan, thanks for this, left some comments!

beacon_node/lighthouse_network/gossipsub/src/behaviour.rs
pawanjay176: Ohh yeah did not notice that, thanks! I added an `dont_send_sent` in the `PeerConnections` struct in 
820c4a4.
The reason I want this metric is to see how latency on the mesh peer connections affects the usefulness of IDONTWANTS. 
If our mesh peers send us IDONTWANTS even if we explicitly requested them not to, that could imply implementation bugs or latency on the connections. If latency is the issue, we could employ try sending the idontwants earlier and check if it makes a difference. I think this metric is useful for that feedback.
lmk what you think.

pawanjay176: This field is just used for metrics btw. I kept the same IDONTWAT_CAP as for the received idontwants struct.
The max memory it will take would be a few mbs at capacity 
((20 bytes msg id + 8 bytes instant) * 10000 * 8 mesh peers). 
For ethereum, since we only use idontwant for blocks and blobs, we should expect to get only max 7 idontwants/ 12 secs, so we can reduce this quite a bit imo

pawanjay176:
Marking as v6.0 as this will be useful for bandwidth metrics cc @michaelsproul

beacon_node/lighthouse_network/gossipsub/src/types.rs
```diff
@@ -123,8 +123,10 @@ pub(crate) struct PeerConnections {
     pub(crate) sender: RpcSender,
     /// Subscribed topics.
     pub(crate) topics: BTreeSet<TopicHash>,
-    /// Don't send messages.
-    pub(crate) dont_send: LinkedHashMap<MessageId, Instant>,
+    /// Don't send messages received from the peer.
```
jxs: nit:
```suggestion
    /// IDONTWANT messages received from the peer.
```
since we are modifying this, probably makes sense to make this more explicit, as there aren't don't send messages. Wdyt?

beacon_node/lighthouse_network/gossipsub/src/types.rs
```diff
@@ -123,8 +123,10 @@ pub(crate) struct PeerConnections {
...
+    pub(crate) dont_send_received: LinkedHashMap<MessageId, Instant>,
+    /// Don't send messages we sent to the peer.
```
jxs:
```suggestion
    /// IDONTWANT messages we sent to the peer.
```

beacon_node/lighthouse_network/gossipsub/src/metrics.rs
```diff
@@ -608,6 +628,20 @@ impl Metrics {
         self.idontwant_bytes.inc_by(bytes as u64);
     }
 
+    /// Register receiving the total bytes of an IDONTWANT control message.
```
jxs:
```suggestion
    /// Register receiving an IDONTWANT control message for a given topic.
```
right?

beacon_node/lighthouse_network/gossipsub/src/metrics.rs
```diff
@@ -341,6 +347,18 @@ impl Metrics {
             metric
         };
 
+        // IDONTWANT messages sent per topic
+        let idontwant_messages_sent_per_topic = register_family!(
+            "idonttwant_messages_sent_per_topic",
+            "Number of IDONTWANT messages sent per topic"
+        );
+
+        // IDONTWANT messages sent but message received per topic
```
jxs:
```suggestion
        // IDONTWANTs which were ignored, and we still received the message per topic
```

beacon_node/lighthouse_network/gossipsub/src/behaviour.rs
```diff
@@ -2751,6 +2767,16 @@ where
                     .entry(*peer_id)
                     .or_default()
                     .non_priority += 1;
+            } else {
```
jxs: nit: can we return early instead of using `else`?  So that the below part of code is on the main scope of the function and it's clearer that it is a success case

beacon_node/lighthouse_network/gossipsub/src/behaviour.rs
```diff
@@ -2512,11 +2521,18 @@ where
...
         for peer in self.connected_peers.values_mut() {
-            while let Some((_front, instant)) = peer.dont_send.front() {
+            while let Some((_front, instant)) = peer.dont_send_received.front() {
+                if (*instant + IDONTWANT_TIMEOUT) >= Instant::now() {
+                    break;
+                } else {
+                    peer.dont_send_received.pop_front();
+                }
+            }
+            while let Some((_front, instant)) = peer.dont_send_sent.front() {
```
jxs, line 2532: since this one is used only for metrics, can we also put it inside a `if let Some(metrics) = self.metrics.as_mut()`?

jxs (commented):
Thanks Pawan! looks good left some nits and a remark

beacon_node/lighthouse_network/gossipsub/src/behaviour.rs
pawanjay176: We only insert into `dont_send_sent` list if metrics are active, so if metrics is None, then the map will be empty anyway so it will be a no-op

jxs: ah right, thanks for explaining Pawan. Can we then leave a comment stating that?

pawanjay176: Done 3683c7e

jxs (approved):
Thanks Pawan!

michaelsproul (approved):

michaelsproul:
@mergify queue

PR #6581, opened 2024-11-12
Add a security section to the book

michaelsproul:
## Proposed Changes

This PR adds a section on security to the book which is intended for security researchers. I've also uploaded the 2020 security assessment by Trail of Bits which we have not previously shared publicly (but have permission from both ToB and SigP to do so now).


paulhauner (approved):

book/src/security.md
```diff
@@ -0,0 +1,12 @@
+Security
+========
```
chong-he: I think we need a header `#` here and a blank line after the header to make Markdown linter happy

```suggestion
# Security

```

book/src/security.md
```diff
@@ -0,0 +1,12 @@
+# Security
+
+
```
chong-he: I accidentally deleted the separating line `======`. this should do it.

```suggestion
========
```

dapplion: I think it was me! 🙏

michaelsproul:
@mergify queue

PR #6583, opened 2024-11-12
builder gas limit & some refactoring

ethDreamer:
I had forgotten to finish this one when I got distracted by other stuff but I think it's good to go now.

Lighthouse now enforces validator sovereignty over the gas limit. The overwhelming majority of builders already respect this and I've notified the relays for the few that didn't and they told me they've fixed the bug. I also did some refactoring for clarity and convenience.

### DIFF
--- a/beacon_node/execution_layer/src/lib.rs
+++ b/beacon_node/execution_layer/src/lib.rs
@@ -27,7 +27,7 @@ use sensitive_url::SensitiveUrl;
 use serde::{Deserialize, Serialize};
 use slog::{crit, debug, error, info, warn, Logger};
 use slot_clock::SlotClock;
-use std::collections::HashMap;
+use std::collections::{hash_map::Entry, HashMap};
 use std::fmt;
 use std::future::Future;
 use std::io::Write;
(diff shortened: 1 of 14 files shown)


consensus/types/src/chain_spec.rs
```diff
@@ -1423,6 +1442,10 @@ const fn default_max_per_epoch_activation_churn_limit() -> u64 {
     8
 }
 
+const fn default_gas_limit_adjustment_factor() -> u64 {
+    1024
```
michaelsproul, line 1446: Checked against https://ethereum.github.io/execution-specs/src/ethereum/cancun/fork.py.html#ethereum.cancun.fork.GAS_LIMIT_ADJUSTMENT_FACTOR:0

consensus/types/src/chain_spec.rs
```diff
@@ -1296,6 +1311,10 @@ pub struct Config {
     #[serde(with = "serde_utils::address_hex")]
     deposit_contract_address: Address,
 
+    #[serde(default = "default_gas_limit_adjustment_factor")]
+    #[serde(with = "serde_utils::quoted_u64")]
+    gas_limit_adjustment_factor: u64,
```
michaelsproul, line 1316: I wonder if we should leave this out of `Config` so that users _can't_ set it? It's not defined as part of the upstream `config.yaml` that the EF uses, so might make sense not to add it here as a non-standard config.

Then again, for most configs it just won't be present so this doesn't pose an issue.

beacon_node/execution_layer/src/lib.rs
```diff
@@ -318,10 +318,52 @@ impl<E: EthSpec, Payload: AbstractExecPayload<E>> BlockProposalContents<E, Paylo
...
     preparation_data: ProposerPreparationData,
+    gas_limit: Option<u64>,
+}
+
+impl ProposerPreparationDataEntry {
+    pub fn update(&mut self, updated: Self) -> bool {
+        let mut changed = false;
+        // Update `gas_limit` if `updated.gas_limit` is `Some` and:
+        // - `self.gas_limit` is `None`, or
+        // - both are `Some` but the values differ.
```
michaelsproul, line 345: Just to confirm, the idea here is to make any previous gas limit "sticky"? So if we receive a validator registration that has a gas limit, followed by a duplicate registration without a gas limit, we remember the gas limit after processing the 2nd registration rather than overwriting it with `None`?

beacon_node/execution_layer/src/lib.rs
```diff
@@ -2004,19 +1999,51 @@ impl fmt::Display for InvalidBuilderPayload {
...
+
+    // Adjust the gas limit safely
+    if target_gas_limit > parent_gas_limit {
+        let gas_diff = target_gas_limit.saturating_sub(parent_gas_limit);
+        parent_gas_limit.checked_add(std::cmp::min(gas_diff, max_gas_limit_difference))
+    } else {
+        let gas_diff = parent_gas_limit.saturating_sub(target_gas_limit);
+        parent_gas_limit.checked_sub(std::cmp::min(gas_diff, max_gas_limit_difference))
+    }
+}
```
michaelsproul, line 2029: I've checked this against the spec for `check_gas_limit` and I am convinced this is correct.

https://ethereum.github.io/execution-specs/src/ethereum/cancun/fork.py.html#ethereum.cancun.fork.check_gas_limit:0

It's unfortunate that the spec uses `max_adjustment_delta` for an inclusive (`>=`) check, meaning that the actual max is 1 less than the "max adjustment delta". Oh well.

michaelsproul (approved):
Looks good. Will merge to unstable shortly.

jimmygchen:
@mergify queue

jimmygchen:
@mergify dequeue

jimmygchen:
@mergify requeue

PR #6585, opened 2024-11-15
Test hdiff binary format stability

michaelsproul:
## Proposed Changes

Add a test checking the stability of the hdiff format so that we can avoid accidentally breaking it during an update.


dapplion (approved):
LGTM

Issue #6442, opened 2024-09-27
Web3Signer integration tests failing on MacOS


jimmygchen:
## Description

Our web3signer integration tests consistently fails on MacOS:

```
---- tests::sepolia_altair_types stdout ----
thread 'tests::sepolia_altair_types' panicked at testing/web3signer_tests/src/lib.rs:258:91:
called `Result::unwrap()` on an `Err` value: InvalidWeb3SignerClientIdentityCertificate(reqwest::Error { kind: Builder, source: Error { code: -25264, message: "MAC verification failed during PKCS12 import (wrong password?)" } })
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This has been an issue for a while but we never had a chance to fix it. 

I digged into this a while ago, and the issue looks like it's related to `security-framework` (macos `Security.Framework` bindings used by `openssl`)  - and it seems like this crate is no longer being actively maintained. 

Not sure if it's worth fixing it to make it work, or we could just disable these tests on MacOS.

ackintosh:
I have also looked into this, and have raised an issue in `security-framework`. 

https://github.com/kornelski/rust-security-framework/issues/216

ackintosh:
The problem was that the default `openssl pkcs12` encoding algorithm is not compatible with the algorithm used by the Apple Security Framework for pkcs.

https://github.com/kornelski/rust-security-framework/issues/216#issuecomment-2466496614

Creating the p12 file with the `-legacy` option resolves the   web3signer test failure on macOS. Our options to achieve this are as follows:

- ~~Overwrite the `lighthouse/key.p12` file with one generated using the `-legacy` option.~~  See the `Update` below.
- Create `lighthouse/key_legacy.p12` and use it only for macOS testing.

I don't fully understand the web3signer_tests, but it seems that the second option might be preferable since the problem occurs only on macOS?

---

**Update**
I confirmed that, on Linux, choosing the first option (`Overwrite ....`) results in the following error:

```
thread 'tests::sepolia_altair_types' panicked at testing/web3signer_tests/src/lib.rs:257:91:
[...]

PR #6588, opened 2024-11-16
Fix web3signer test fails on macOS

ackintosh:
## Issue Addressed

https://github.com/sigp/lighthouse/issues/6442

## Proposed Changes

- Created `lighthouse/key_legacy.p12` and using it only for macOS testing. (See https://github.com/sigp/lighthouse/issues/6442#issuecomment-2469252651)
- Changed the validity period of certificates from 36500 to 825. (See https://github.com/sigp/lighthouse/issues/6442#issuecomment-2474979183)
- Removed `-aes256` as it's ignored on exporting. (https://github.com/sigp/lighthouse/commit/3c8258f966548217050712873002b1140c90c1cf)

## Additional Info

The validity period of certificates has changed to 825 days, so it's worth considering how to update the certificates automatically:

- Create a GitHub Actions job to update the certificates periodically, for example, every year.
- Generate the certificates for each test run.


dapplion (approved):
nice find!

jimmygchen:
@mergify queue

jimmygchen:
@mergify dequeue

jimmygchen:
@mergify requeue

PR #6591, opened 2024-11-18
Fix v22 schema upgrade

michaelsproul:
## Proposed Changes

Fix some AnchorInfo wrangling that got messed up between the last time we tested the schema migration and now. This code needs to be tested manually at present, which is a bit unfortunate.


michaelsproul:
Confirmed working

dapplion (approved):

michaelsproul:
@mergify queue

PR #6595, opened 2024-11-19
Avoid computing rpc_blob_limits multiple times

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

rpc_blob_limits was being computed repeatedly even though its a constant value. Move it to a LazyLock block to avoid repeated computation.
Found this taking ~10% of cpu when I was playing around with flamegraph

### DIFF
--- a/beacon_node/lighthouse_network/src/rpc/protocol.rs
+++ b/beacon_node/lighthouse_network/src/rpc/protocol.rs
@@ -18,11 +18,11 @@ use tokio_util::{
 };
 use types::{
     BeaconBlock, BeaconBlockAltair, BeaconBlockBase, BeaconBlockCapella, BeaconBlockElectra,
-    BlobSidecar, ChainSpec, DataColumnSidecar, EmptyBlock, EthSpec, ForkContext, ForkName,
-    LightClientBootstrap, LightClientBootstrapAltair, LightClientFinalityUpdate,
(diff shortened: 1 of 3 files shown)


michaelsproul (approved):
Nice find!

michaelsproul:
The test failure looks related. It might be because `E: EthSpec` has been fixed to `MainnetEthSpec`? We could one lazy lock per spec if necessary

dapplion (approved):
nice find! what else showed up in the flamegraph?

pawanjay176:
Moved rpc_tests to run over MainnetEthSpec. I think it made a difference when minimal blob sizes were small. Don't think it makes much of a difference in terms of test speeds now. Not sure if there's any reason to run them over minimal cc @AgeManning for when you are back

pawanjay176:
> Not sure if there's any reason to run them over minimal 
We need it for the simulator tests. Running sim tests over MainnetEthSpec will take 8x longer. Can revisit this if we manage to replace sim tests with assertoor tests. Reverted back to minimal in rpc_tests

pawanjay176:
> what else showed up in the flamegraph?

Mostly signature verification and decompression taking up cpu time. Milhouse uses a lot of stack depth. Both are expected I guess.

michaelsproul (approved):
Looks good.

michaelsproul:
@mergify queue

PR #6600, opened 2024-11-21
Fetch blobs from EL prior to block verification

michaelsproul:
## Proposed Changes

Optimise `fetch_blobs` significantly, by fetching blobs from the EL prior to consensus and execution verification of the block.

We had noticed that we weren't getting many hits with fetch blobs, and this was because blobs were almost always arriving on gossip prior to us requesting them. Only a few times an hour would the `fetch_blobs` logic actually fire.

With this change I'm seeing much more frequent hits, without a substantial increase in publication bandwidth. In the last 30 mins running on mainnet there have been 116 hits, and 156 individual blobs published (out of 395 fetched).

Data here: https://docs.google.com/spreadsheets/d/1ZJIYbOPwNGa_veqUC0ywsOdzFYvh4aqJLMYqFoisA_E/edit?usp=sharing

This does imply that we're publishing around 35% of all blobs! But this will likely come down as more nodes chip in to publishing.


dapplion:
Any security considerations on triggering this logic before validating the block? The most damage a proposer can do is waste bandwidth on a bad proposal. This does not seem like a big issue and can be done anyway regardless of fetch_blobs.

Else the experimental results look great

michaelsproul:
We're doing this after gossip validation of the block, so we know that **the proposer's signature is valid and they are a legitimate proposer for the slot**.

Unless the proposer slashes themselves, the blob versioned hashes in the block header are the "true" (valid) versioned hashes for this slot. Alternatively the block could be completely invalid (but not slashable), in which case we will reject it upon completion of block processing.

As part of `fetch_blobs` we run gossip blob validation:

https://github.com/sigp/lighthouse/blob/6e1945fc5d2b2025d8ec7eb9b3f2b1eed94967ad/beacon_node/beacon_chain/src/fetch_blobs.rs#L131-L148

So if they are malformed (e.g. bad KZG proof), they will be rejected at this point.

TL;DR on the whole I think it's security-equivalent to processing blobs on gossip:

- We have a valid signature from the proposer committing to the blobs.
- We verify the blobs themselves before importing them.

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -1444,6 +1444,12 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
             }
         }
 
+        // Block is gossip valid. Attempt to fetch blobs from the EL using versioned hashes derived
+        // from kzg commitments, without having to wait for all blobs to be sent from the peers.
+        let publish_blobs = true;
+        self.fetch_engine_blobs_and_publish(block.clone(), block_root, publish_blobs)
+            .await;
```
michaelsproul: @jimmygchen points out that we probably shouldn't await this here as it will delay block processing. We could spawn a task and do it in parallel.

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -1447,8 +1447,16 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
         // from kzg commitments, without having to wait for all blobs to be sent from the peers.
         let publish_blobs = true;
-        self.fetch_engine_blobs_and_publish(block.clone(), block_root, publish_blobs)
-            .await;
+        let self_clone = self.clone();
+        let block_clone = block.clone();
+        self.executor.spawn(
+            async move {
+                self_clone
+                    .fetch_engine_blobs_and_publish(block_clone, block_root, publish_blobs)
```
dapplion, line 1455: Since this is running as a task, it's no longer bound by the beacon processor queue. Could someone spam gossip blocks and cause a lot of fetch blobs work?

michaelsproul: They would need to be beacon blocks with valid signatures, and this is a linear factor, so it can't really blow up much beyond the number of threads allocated to the beacon processor. E.g. if we have 16 threads in the BP, we might end up with 32 running tasks max, which are mostly I/O bound and should be handled just fine by Tokio.

We do this in a few other places, like when we check the payload with the EL:

https://github.com/sigp/lighthouse/blob/6329042628ea0afcbcbce3874284c78ba9aa41a7/beacon_node/beacon_chain/src/block_verification.rs#L1407-L1416

dapplion (approved):

michaelsproul:
@mergify queue

michaelsproul:
@mergify requeue

michaelsproul:
@mergify queue

michaelsproul:
@mergify requeue

michaelsproul:
@mergify requeue

michaelsproul:
@mergify requeue

michaelsproul:
@mergify requeue

michaelsproul:
@mergify requeue

michaelsproul:
@mergify queue

michaelsproul:
THANK FUCK.

PR #6603, opened 2024-11-23
Ensure non-zero bits for each committee bitfield comprising an aggregate

eserilev:
## Issue Addressed

https://github.com/ethereum/consensus-specs/pull/4002

## Proposed Changes

Add an additional check to ensure that on-chain attestations have at least one non-zero bit set per attesting committee bitfield. This change is part of the upcoming Electra alpha release


michaelsproul (approved):
LGTM!

michaelsproul:
@mergify queue

PR #6604, opened 2024-11-24
Increase idle connection timeout

AgeManning:
## Issue Addressed

It was reported that when Prysm connects to us, there is occasionally a small time in negotiating protocols before we set-up gossipsub that we have an idle connection. In this case Lighthouse immediately drops the connection. 

This results in us preventing Prysm from connecting to us some fraction of the time. This PR increases the timeout from 0 seconds to 1 second to help correct this.

### DIFF
--- a/beacon_node/lighthouse_network/src/service/mod.rs
+++ b/beacon_node/lighthouse_network/src/service/mod.rs
@@ -38,6 +38,7 @@ use std::num::{NonZeroU8, NonZeroUsize};
 use std::path::PathBuf;
 use std::pin::Pin;
 use std::sync::Arc;
+use std::time::Duration;
 use types::{
     consts::altair::SYNC_COMMITTEE_SUBNET_COUNT, EnrForkId, EthSpec, ForkContext, Slot, SubnetId,
 };
@@ -466,6 +467,8 @@ impl<E: EthSpec> Network<E> {
             let config = libp2p::swarm::Config::with_executor(Executor(executor))
                 .with_notify_handler_buffer_size(NonZeroUsize::new(7).expect("Not zero"))
                 .with_per_connection_event_buffer_size(4)
+                .with_idle_connection_timeout(Duration::from_secs(1)) // Other clients can timeout
+                // during negotiation
                 .with_dial_concurrency_factor(NonZeroU8::new(1).unwrap());
 
             let builder = SwarmBuilder::with_existing_identity(local_keypair)
--- a/beacon_node/lighthouse_network/src/service/mod.rs
+++ b/beacon_node/lighthouse_network/src/service/mod.rs
@@ -467,7 +467,7 @@ impl<E: EthSpec> Network<E> {
             let config = libp2p::swarm::Config::with_executor(Executor(executor))
                 .with_notify_handler_buffer_size(NonZeroUsize::new(7).expect("Not zero"))
                 .with_per_connection_event_buffer_size(4)
-                .with_idle_connection_timeout(Duration::from_secs(1)) // Other clients can timeout
+                .with_idle_connection_timeout(Duration::from_secs(10)) // Other clients can timeout
                 // during negotiation
                 .with_dial_concurrency_factor(NonZeroU8::new(1).unwrap());
 


AgeManning:
@jxs - Is probably not going to be awake early enough to approve this one, but I think it's not going to destroy the world if for some reason this turns out to be a mistake. We can revert after the RC

michaelsproul:
I'll just roll it into the RC branch without merging it

jxs (commented):
I had a conversation with @dennis-tra some weeks ago, the data seemed inconclusive as AFAIR we were connected to Prysm clients, is there new data that backs this decision?

btw I am not sure if `with_idle_connection_timeout()` affects negotiation, confirm [here](https://github.com/libp2p/rust-libp2p/blob/00588a543a135f71ee7053b5315fdbb7c6e492f4/swarm/src/connection.rs#L370-L395)

AgeManning:
Yeah. The discussion I had was just that through experimentation they identified that this is the cause for connection drops. Changing this value is expected to resolve it. 

I've not seen specific data on it, but I also didn't see much downside to having a 1 second buffer, just in case it does help. 

Curious to know if there is some downside here (we shouldn't have idle connections so long as gossipsub is negotiated).

AgeManning:
I guess. perhaps we maintain connections with crawlers and other clients that don't support gossipsub more easily with this PR. 

Potentially, in a future PR we should only allow a percentage of our peer count to not support gossipsub. Last I checked there was a decent amount, like 10-20% of peers that didn't support it.

dennis-tra:
Hi everyone,

yeah I talked to @AgeManning at DevCon about this issue. I also mentioned that we had a chat @jxs and that the data was inconclusive.

@jxs you mentioned that Lighthouse is indeed connected to Prysm peers but the crucial thing is in which direction these connections point. I presume most connections are pointing outwards from Lighthouse to Prysm peers. However, let me emphasize that this is just a hypothesis. I'm basing this on our observations from interacting with other rust-libp2p peers and a study of the relevant code.

In go-libp2p there's a brief moment in time when connecting to a peer before control is handed back to the user where all four conditions are true from your [referenced code snippet](https://github.com/libp2p/rust-libp2p/blob/00588a543a135f71ee7053b5315fdbb7c6e492f4/swarm/src/connection.rs#L370-L395). In go-libp2p the relevant part is [here](https://github.com/libp2p/go-libp2p/blob/288868c0d90e2d72a124da26caa14c77d35dbbdd/p2p/host/basic/basic_host.go#L832) (which is called from the user-facing Connect call [here](https://github.com/libp2p/go-libp2p/blob/288868c0d90e2d72a124da26caa14c77d35dbbdd/p2p/host/basic/basic_host.go#L808)) where we wait until the `identify` exchange has completed before we do anything else (like opening a new stream).

AgeManning:
@dennis-tra - Cool. Wondering if we can confirm this before we merge this into our next release. 

Is it possible for you to use this PR and see if it resolves the issues you have been seeing?

If they persist, then perhaps the problem lies elsewhere

beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -466,6 +467,8 @@ impl<E: EthSpec> Network<E> {
             let config = libp2p::swarm::Config::with_executor(Executor(executor))
                 .with_notify_handler_buffer_size(NonZeroUsize::new(7).expect("Not zero"))
                 .with_per_connection_event_buffer_size(4)
+                .with_idle_connection_timeout(Duration::from_secs(1)) // Other clients can timeout
```
jxs:
```suggestion
                .with_idle_connection_timeout(Duration::from_secs(10)) // Other clients can timeout
```

jxs (approved):
Hi @dennis-tra! Thanks for your explanation, meanwhile I also found https://github.com/libp2p/rust-libp2p/pull/4967/ which I wasn't aware and makes sense to be merged, so I think we can leave it to 10 second here while a `rust-libp2p` release with 10 seconds as the default isn't released

dennis-tra:
Great news, thanks @jxs! 10s sounds good.

michaelsproul:
@mergify queue

michaelsproul:
@mergify dequeue

michaelsproul:
@mergify queue

michaelsproul:
@mergify requeue

michaelsproul:
@mergify dequeue

michaelsproul:
@mergify requeue

michaelsproul:
@mergify requeue

Issue #4804, opened 2023-10-04
Support voluntary exits via validator manager


karalabe:
The validator_manager supports creating and importing validator keys. The creation command makes one big file containing all the keys; and the importer imports each key individually with a randomly generated password. There is no voluntary exit command via the validator_manager. This makes things a bit quirky.

- I need to point lighthouse to one of the imported keys, since I can't use the dump containing many keys.
- The only place where I have a set of keys is the validator itself, so I need to dig up the key from the live system.
- I also need to look through the validator_definitions and extract the correct random password.

It's a bit of a PITA even for a single validator. Doing this for many gets annoying very fast.

Would be nice to be able to generate voluntary exists via the validator_manager too.

aronfoster:
Would be nice if the validator_manager also added the ability to [delete keys](https://ethereum.github.io/keymanager-APIs/#/Local%20Key%20Manager/deleteKeys), which you'll want to do after exiting.

rnicoll:
Taking a look at this currently, hoping to have a first PR later this month.

chong-he:
@rnicoll Are you working on this? 

Happy to take on this if you aren't on it, thanks!

Edit: Working on this and also another issue with voluntary exit #4303

rnicoll:
@chong-he Sorry, just saw this - as you probably gathered I'd run out of available time, thanks for taking this over!

chong-he:
Voluntary exit using `lighhtouse vm` is added in #6612

Issue #4303, opened 2023-05-17
Add `--exit-epoch` flag to validator-manager


xenowits:
## Description

This is a feature request to add an `--exit-epoch` flag to the `lighthouse account validator exit` command.

This feature allows the user to set a specific epoch for the voluntary exit. It is supported by other validator clients such as [teku](https://docs.teku.consensys.net/reference/cli/subcommands/voluntary-exit#epoch), [lodestar](https://chainsafe.github.io/lodestar/reference/cli/#validator-voluntary-exit) and [nimbus](https://nimbus.guide/voluntary-exit.html).

## Version

None.

## Present Behaviour

Currently, a user can't specify the desired epoch to exit and the lighthouse VC broadcasts exit message with the current epoch.

## Expected Behaviour

The validator client should allow a user to set a desired exit epoch.

## Steps to resolve

* Add a new flag `--exit-epoch` to https://github.com/sigp/lighthouse/blob/stable/account_manager/src/validator/exit.rs. 
* Modify `publish_voluntary_exit()` [function](https://github.com/sigp/lighthouse/blob/stable/account_manager/src/validator/exit.rs#L148) to use the `epoch` provided above (if any).

michaelsproul:
@xenowits As a workaround, try using the keymanager API for exits: https://ethereum.github.io/keymanager-APIs/?urls.primaryName=dev#/Voluntary%20Exit/signVoluntaryExit

Info on setting up the Lighthouse VC API can be found here: https://lighthouse-book.sigmaprime.io/api-vc.html

michaelsproul:
This may also be already implemented by https://github.com/sigp/lighthouse/pull/3502, I'm not sure.

xenowits:
> This may also be already implemented by https://github.com/sigp/lighthouse/pull/3502, I'm not sure.

Thanks for the workaround! 

But, do you think it is feasible to introduce this feature or are there some limitations?

michaelsproul:
@xenowits It's definitely feasible. The `lighthouse validator-manager` is a new CLI that uses the API, so it would be super easy to include it there. We could also add it to the existing CLI easily enough, but would probably move people towards the validator-manager long term.

xenowits:
okay, so the `validator-manager` is a new CLI feature that is not yet released i suppose? 

Yeah, this subcommand would be cool to use for exits then 👍

xenowits:
@michaelsproul is the `lighthouse validator-manager` command released?

[...]

PR #6612, opened 2024-11-26
Add voluntary exit via validator manager

chong-he:
## Issue Addressed

* #4303
* #4804

## Proposed Changes

-Add voluntary exit feature to the validator manager
-Add delete all validators by using the keyword "all"

## Additional Info

I reuse some code from the `lighthouse account validator exit` file: 
https://github.com/sigp/lighthouse/blob/stable/account_manager/src/validator/exit.rs

Thank you @michaelsproul for the help and guidance along the way


validator_manager/src/exit_validators.rs
```diff
@@ -0,0 +1,576 @@
...
+                    .await
+                    .map_err(|e| format!("Failed to publish voluntary exit: {}", e))?;
+                eprintln!(
+                    "Successfully validated and published voluntary exit for validator {}",
+                    validator_to_exit
+                );
+            }
+
+            // Check validator status after publishing voluntary exit
+            if exit_status {
```
chong-he: The  `--status` is used to query the exit status of validators.  The reason behind introducing it is:

In `lighthouse account validator exit`, the status was shown after a sleep of, e.g., 12s

https://github.com/sigp/lighthouse/blob/0d90135047519f4c2ee586d50e560f7bb2ff9b10/account_manager/src/validator/exit.rs#L204

If we introduce a sleep here, then for each `validator_to_exit`, it will sleep for 12s, which is not really UX friendly. For practical consideration, the user's priority may just want to first publish the exit messages for all validators.

While there is a way to tweak it so that it only sleeps once and not for the rest of `validator_to_exit` (or shortening the sleep time), doing so will likely cause the exit status to be shown as:

`Voluntary exit for Validator {} is waiting to be accepted into the beacon chain.`

for the rest of `validator_to_exit` as the head block is not yet updated, which is not really useful either.

So I ended up introducing this `--status` flag to query the exit status separately. But I am not sure if this is a good idea. 

Keen on hear the thoughts of reviewers on this.

validator_manager/src/exit_validators.rs
```diff
@@ -0,0 +1,625 @@
...
+                    eprintln!(
+                        "Validator {} has not initiated voluntary exit. Validator status is: {}",
+                        validator_to_exit, updated_validator_data.status
+                    )
+                }
+            }
+        }
+    }
+
+    if merge {
```
chong-he: Not sure if this `--merge` is useful? I tested pasting the generated JSON on beaconcha.in and it doesn't seem to work. Maybe it is an issue with the generated JSON, or is it because submitting voluntary exit can only be done one at a time?

michaelsproul: Yeah let's get rid of it. The beacon API only supports posting exits one at a time:

https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Beacon/submitPoolVoluntaryExit

validator_manager/src/exit_validators.rs
```diff
@@ -0,0 +1,680 @@
...
+use std::time::Duration;
+use types::{ChainSpec, EthSpec, PublicKeyBytes};
+
+pub const CMD: &str = "exit";
+pub const BEACON_URL_FLAG: &str = "beacon-node";
+pub const VC_URL_FLAG: &str = "vc-url";
+pub const VC_TOKEN_FLAG: &str = "vc-token";
+pub const VALIDATOR_FLAG: &str = "validators";
+pub const EXIT_EPOCH_FLAG: &str = "exit-epoch";
+pub const SIGNATURE_FLAG: &str = "signature";
```
michaelsproul: I think maybe we should call this `--no-broadcast` or `--presign`. The "signature" name is a little bit too general I think, because all voluntary exits involve signatures.

validator_manager/src/exit_validators.rs
```diff
@@ -0,0 +1,680 @@
...
+            "Exits one or more validators using the HTTP API. It can \
+        also be used to generate voluntary exit message for a particular future epoch.",
+        )
+        .arg(
+            Arg::new(BEACON_URL_FLAG)
+                .long(BEACON_URL_FLAG)
+                .value_name("NETWORK_ADDRESS")
+                .help("Address to a beacon node HTTP API")
+                .action(ArgAction::Set)
+                .display_order(0),
```
michaelsproul: We could add a `conflicts_with` to make `--beacon-node` and `--presign` conflict? Users shouldn't be doing both.

validator_manager/src/exit_validators.rs
```diff
@@ -0,0 +1,680 @@
...
+        if beacon_url.is_some() {
+            let beacon_node = if let Some(ref beacon_url) = beacon_url {
+                BeaconNodeHttpClient::new(
+                    SensitiveUrl::parse(beacon_url.as_ref())
+                        .map_err(|e| format!("Failed to parse beacon http server: {:?}", e))?,
+                    Timeouts::set_all(Duration::from_secs(12)),
+                )
+            } else {
+                return Err("Beacon URL is not provided".into());
+            };
```
michaelsproul: This can be simplified just to `if let Some(ref beacon_url) = beacon_url { ... }` at the top-level. That way there is no need to have an error case.

i.e. remove the `if beacon_url.is_some()` and just use `if let Some(ref beacon_url) = beacon_url` for the outermost `if`.

validator_manager/src/exit_validators.rs
```diff
@@ -0,0 +1,680 @@
...
+            .await
+            .map_err(|e| format!("Failed to get validator details: {:?}", e))?
+            .ok_or_else(|| {
+                format!(
+                    "Validator {} is not present in the beacon state. \
+                Please ensure that your beacon node is synced and the validator has been deposited.",
+                    validator_to_exit
+                )
+            })?
+            .data;
```
michaelsproul: This formatting is weird. I suspect `cargo fmt` isn't working properly here because some of the strings are too long. Put a few more line breaks in the strings and it should start working again.

validator_manager/src/exit_validators.rs
```diff
@@ -0,0 +1,680 @@
...
+                            "Validator {} is not present in the beacon state",
+                            validator_to_exit
+                        )
+                    })?
+                    .data;
+
+                if validator_data.status == ValidatorStatus::ActiveOngoing
+                    && updated_validator_data.status == ValidatorStatus::ActiveOngoing
+                {
+                    eprintln!("Voluntary exit for Validator {} is waiting to be accepted into the beacon chain.", validator_to_exit);
```
michaelsproul: I would have thought that in practice, we would often just hit this case. Unless there's a block that includes the exit immediately after the post request?

Maybe if `--status` is provided, there should be a loop over all validators _after_ the messages have been posted, in which case most of them should have been included in blocks

michaelsproul (changes requested):
Looking good overall, just a few minor suggestions!

validator_manager/src/exit_validators.rs
chong-he: Removed in [419eda7](https://github.com/sigp/lighthouse/pull/6612/commits/419eda7ffc58cf37dd3f2553c8d26904abd38572)

validator_manager/src/exit_validators.rs
chong-he: Revised in [b579d40](https://github.com/sigp/lighthouse/pull/6612/commits/b579d408c22cc76fb4ebcc21ba984bceca653c60)

validator_manager/src/exit_validators.rs
chong-he: Added in [b579d40](https://github.com/sigp/lighthouse/pull/6612/commits/b579d408c22cc76fb4ebcc21ba984bceca653c60)

validator_manager/src/exit_validators.rs
chong-he: Simplified in [b579d40](https://github.com/sigp/lighthouse/pull/6612/commits/b579d408c22cc76fb4ebcc21ba984bceca653c60)

validator_manager/src/exit_validators.rs
chong-he: You're exactly right! I split the string to more lines and the formatting is working again, thank you. 

[b579d40](https://github.com/sigp/lighthouse/pull/6612/commits/b579d408c22cc76fb4ebcc21ba984bceca653c60)

validator_manager/src/exit_validators.rs
chong-he: I moved the check of voluntary exit status to `list_validators.rs` as discussed. By default `list` will check all validators, but I figure in practice, users may just want to check some validators. So I added a flag `--validators` to specify which validator to query.

[2407965](https://github.com/sigp/lighthouse/pull/6612/commits/24079658dbe659cc72df196364a543f5207219e4)

validator_manager/src/exit_validators.rs
chong-he: Resolved by https://github.com/sigp/lighthouse/pull/6612#discussion_r1926625488

validator_manager/src/exit_validators.rs
chong-he: Also added `conflicts_with` for the flag `--exit-epoch` and `--beacon-node`, and that `--exit-epoch` will require `--presign` as I think having `--exit-epoch` alone isn't meaningful. 

[683259d](https://github.com/sigp/lighthouse/pull/6612/commits/683259dbaa2ce515ce229bb18e9764e715fe0cf9)

validator_manager/src/list_validators.rs
```diff
@@ -31,47 +37,177 @@ pub fn cli_app() -> Command {
...
+    pub beacon_url: Option<SensitiveUrl>,
+    pub validators_to_display: Vec<PublicKeyBytes>,
 }
 
 impl ListConfig {
     fn from_cli(matches: &ArgMatches) -> Result<Self, String> {
+        let validators_to_display_str =
+            clap_utils::parse_optional::<String>(matches, VALIDATOR_FLAG)?;
+
+        // Keyword "all" to exit all validators, vector to be created later
```
michaelsproul:
```suggestion
        // Keyword "all" to list all validators, vector to be created later
```

book/src/validator_manager.md
```diff
@@ -30,6 +30,9 @@ The `validator-manager` boasts the following features:
 
 ## Guides
 
+- [Creating and importing validators using the `create` and `import` commands.](./validator_manager_create.md)
+- [Moving validators between two VCs using the `move` command.](./validator_manager_move.md)
+- [Managing validators such as exit, delete, import and list validators.](./validator_manager_api.md)
```
michaelsproul: This section looks to be duplicated below?

michaelsproul: I updated this in 9f3a152ec5e75dcd896ad2400ad22283f84fc163

book/src/help_vm.md
```diff
@@ -28,6 +28,10 @@ Commands:
   delete
           Deletes one or more validators from a validator client using the HTTP
           API.
+  exit
+          Exits one or more validators using the HTTP API. It can also be used
+          to generate a presign voluntary exit message for a particular future
```
michaelsproul: I think we should say _presigned_?

```suggestion
          to generate a presigned voluntary exit message for a particular future
```

michaelsproul (approved):

PR #2827, opened 2021-11-24
[Merged by Bors] - Correctly update range status when outdated chains are removed

divagant-martian:
We were batch removing chains when purging, and then updating the status of the collection for each of those. This makes the range status be out of sync with the real status. This represented no harm to the global sync status, but I've changed it to comply with a correct debug assertion that I got triggered while doing some testing.
Also added tests and improved code quality as per @paulhauner 's suggestions.

PR #6618, opened 2024-11-27
Write range sync tests in external event-driven form

dapplion:
## Issue Addressed

Re-write existing range sync tests into an event-driven format to accommodate the refactor introduced in

This format was introduced in 
- https://github.com/sigp/lighthouse/pull/5534

and does not make assumptions about internal functions, it treats sync as a black box inputting actual events used in production.

Note: I've removed the generics used in range sync since now we don't unit test directly


beacon_node/network/src/sync/tests/range.rs
```diff
@@ -1 +1,275 @@
...
+
+    // Sync should have requested a batch, grab the request
+    let _ = rig.find_blocks_by_range_request(&finalized_peer);
+
+    // Now the chain knows both chains target roots.
+    rig.remember_block(head_peer_block);
+    rig.remember_block(finalized_peer_block);
+
+    // Add an additional peer to the second chain to make range update it's status
+    rig.add_finalized_peer();
```
jimmygchen, line 243: I don't really understand what this line does - do we need to assert for some condition after doing this?

beacon_node/network/src/sync/tests/range.rs
```diff
@@ -1 +1,275 @@
...
+        }
+    }
+
+    fn create_canonical_block(&mut self) -> SignedBeaconBlock<E> {
+        self.harness.advance_slot();
+
+        let block_root =
+            tokio::runtime::Runtime::new()
+                .unwrap()
+                .block_on(self.harness.extend_chain(
```
jimmygchen: Maybe consider using `async/await` here instead of creating a new `tokio::Runtime` - might keep things simpler and avoid the extra overhead?

beacon_node/network/src/sync/tests/range.rs
```diff
@@ -1 +1,275 @@
...
+            .store
+            .get_full_block(&block_root)
+            .unwrap()
+            .unwrap()
+    }
+
+    fn remember_block(&mut self, block: SignedBeaconBlock<E>) {
+        tokio::runtime::Runtime::new()
+            .unwrap()
+            .block_on(self.harness.process_block(
```
jimmygchen: Same here

jimmygchen:
Nice! Looks clean and easy to read 👍

beacon_node/network/src/sync/tests/range.rs
dapplion: I respected the original test sequence as written in https://github.com/sigp/lighthouse/pull/2827, it covered a bug which this final line triggered a debug assertion

beacon_node/network/src/sync/tests/range.rs
dapplion: Switched to async :+1

beacon_node/network/src/sync/tests/range.rs
dapplion: Switched to async 👍

dapplion:
```
    sync::tests::range::state_update_while_purging
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 86 filtered out; finished in 7.58s

──── STDERR:             network sync::tests::range::state_update_while_purging
thread 'sync::tests::range::state_update_while_purging' panicked at /home/runner/work/lighthouse/lighthouse/beacon_node/beacon_chain/src/test_utils.rs:904:14:
called `Result::unwrap()` on an `Err` value: ExecutionLayerMissing
```

This test is failing in CI after switching to async, but can't reproduce locally

beacon_node/network/src/sync/tests/range.rs
jimmygchen: Got it, thanks for adding in the links, very helpful!

jimmygchen (approved):
I managed to reproduce the failing test locally, not sure why it was passing before 🤔  We do need it in this case I think, so we might have uncovered a lurking issue - adding a mock execution layer fixed it. 

The other changes look good, thanks!

jimmygchen:
@mergify queue

jimmygchen:
@mergify dequeue

jimmygchen:
@mergify requeue

PR #6621, opened 2024-11-27
Make `ValidatorIndex` visible

Zacholme7:
## Issue Addressed

N/A

## Proposed Changes

Change `ValidatorIndex` to pub and export it

## Additional Info

Needed for defining types in Anchor

### DIFF
--- a/consensus/types/src/beacon_state/pubkey_cache.rs
+++ b/consensus/types/src/beacon_state/pubkey_cache.rs
@@ -1,8 +1,6 @@
 use crate::*;
 use rpds::HashTrieMapSync as HashTrieMap;
 
-pub type ValidatorIndex = usize;
-
 #[allow(clippy::len_without_is_empty)]
 #[derive(Debug, PartialEq, Clone, Default)]
 pub struct PubkeyCache {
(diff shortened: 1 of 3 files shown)


consensus/types/src/beacon_state/pubkey_cache.rs
```diff
@@ -1,7 +1,7 @@
 use crate::*;
 use rpds::HashTrieMapSync as HashTrieMap;
 
-type ValidatorIndex = usize;
+pub type ValidatorIndex = usize;
```
jxs: this is a type alias, if `consensus` is exposing a function/method that requires `ValidatorIndex` we should probably either switch it to `usize` or make `ValidatorIndex` a [`newtype`](https://doc.rust-lang.org/rust-by-example/generics/new_types.html) to avoid breaking change issues

Zacholme7: Hey thanks for this! I started changing the `ValidatorIndex` to a newtype but it ended up requiring too many unexpected downstream changes for me to justify just the re-export. I would like to keep it a newtype in anchor so I believe it is best to just define it there, hence why I closed this pr.

PR #6625, opened 2024-11-27
Remove pending requests from ready_requests

ackintosh:
## Issue Addressed & Proposed Changes

The pending requests in `delay_requests` move to `ready_requests` before being returned by `poll_ready`. One request is returned per `poll_ready` call, so some pending requests might remain in `ready_requests`. We need to remove the pending requests from `ready_requests`, just like `delay_requests`, when a peer disconnects.


beacon_node/lighthouse_network/src/rpc/self_limiter.rs
```diff
@@ -165,9 +165,30 @@ impl<Id: ReqId, E: EthSpec> SelfRateLimiter<Id, E> {
...
+                if req_peer_id == &peer_id {
+                    failed_requests.push((*request_id, request.protocol()));
+                    // Remove the entry
+                    false
+                } else {
+                    // Keep the entry
+                    true
+                }
+            } else {
+                unreachable!()
```
jxs: submitted #6634 to avoid having this `unreachable!()` here

jxs (commented):
Nice find Akihito! submitted #6634 to ease it

beacon_node/lighthouse_network/src/rpc/self_limiter.rs
```diff
@@ -234,9 +233,25 @@ impl<Id: ReqId, E: EthSpec> SelfRateLimiter<Id, E> {
...
+                if req_peer_id == &peer_id {
+                    failed_requests.push((*request_id, req.protocol()));
+                    // Remove the entry
+                    false
+                } else {
+                    // Keep the entry
+                    true
+                }
+            } else {
+                unreachable!("Coding error: unexpected RPCSend variant {rpc_send:?}.")
```
dapplion: Is it necessary to have this `unreachable` here? What about a debug assert?

beacon_node/lighthouse_network/src/rpc/self_limiter.rs
```diff
@@ -549,4 +565,71 @@ mod tests {
...
+            // Check that the limiter allows the first request while other requests are added to the queue.
+            if i == 1 {
+                assert!(result.is_ok());
+            } else {
+                assert!(result.is_err());
+            }
+        }
+
+        // Wait until the tokens have been regenerated, then run `next_peer_request_ready`.
+        tokio::time::sleep(Duration::from_secs(3)).await;
```
dapplion: Can you set a custom quota so we don't have to wait 3 seconds here?

ackintosh:
@dapplion @jxs Thanks for your feedback! I've updated the PR accordingly.

dapplion (approved):
Looks good to me! The test is solid too

michaelsproul (approved):

PR #6626, opened 2024-11-27
Increase rpc rate limits

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

In https://github.com/sigp/lighthouse/pull/6093 , we bumped the rate limits for blobs by range significantly presumably for peerdas devnets. This limit is too permissive for mainnet imo, we should reduce it down significantly so as to not get overwhelmed with requests and also keeping the upload bandwidth at check.

~~I'm proposing to increase the limits by a factor of around 3. Allowing a single epoch of blocks (32 blocks max) to be requested per second or 160 per 5 seconds to account for bursts.~~
See https://github.com/sigp/lighthouse/pull/6626#issuecomment-2505240978 for justification for final numbers on this PR after testing

### DIFF
--- a/beacon_node/lighthouse_network/src/rpc/config.rs
+++ b/beacon_node/lighthouse_network/src/rpc/config.rs
@@ -104,14 +104,14 @@ impl RateLimiterConfig {
     pub const DEFAULT_META_DATA_QUOTA: Quota = Quota::n_every(2, 5);
     pub const DEFAULT_STATUS_QUOTA: Quota = Quota::n_every(5, 15);
     pub const DEFAULT_GOODBYE_QUOTA: Quota = Quota::one_every(10);
-    // We request 32 blocks per batch. In a mainnet like network, where we have many peers to choose
-    // from for range sync, we should ideally be spreading out the sync requests over multiple nodes
-    // and request not more than 2 epochs from a single peer every 5 seconds.
-    pub const DEFAULT_BLOCKS_BY_RANGE_QUOTA: Quota = Quota::n_every(160, 5);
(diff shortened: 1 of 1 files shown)


pawanjay176:
After doing some analysis, I think we should adjust the defaults to be 128 tokens/10 secs for blocksbyrange and 512 tokens/10 secs for blobsbyrange (target+1 blobs per block on average to account for high blob usage periods)

Just writing down the thought process for future reference.

With these numbers, our bandwidth usage per peer  would look like
- Base rate blocks: 128 blocks/10s * 100KB = 1,280 KB/s
- Base rate blobs: 512 blobs/10s * 128KB = 6,553.6 KB/s
- Total base bandwidth per peer: ~7.8 MB/s 
- with the token bucketing algo we use, this could go max upto ~15.6 MB/s

For a node requesting a batch (32 blocks + associated blobs) from a given peer every 3-5 seconds, this limit allows 4 full batch requests (1 batch = 1 epoch) per 10-second window per peer. Given that we have a big pool of nodes to sync from in mainnet and similar networks, this value makes sense for mainnet/gnosis imo. For devnets, we can raise the limits with cli flags. 
The 10-second window provides good spam resistance vs shorter window sizes since an attacker would need to spread out these requests to max out the bandwidth usage. 

Not touching data column limits in this PR.

jxs (commented):
Thanks for this Pawan! Can we leave a link to https://github.com/sigp/lighthouse/pull/6626#issuecomment-2505240978 on the comment for the detailed info?

michaelsproul (approved):

michaelsproul:
@mergify queue

PR #6631, opened 2024-11-29
Delete OTB verification service

michaelsproul:
## Proposed Changes

One step closer to deleting all the old "merge" code, this PR removes the optimistic transition block verifier. It served us well, but is no longer required!


dapplion (approved):
Not familiar with the code, but looks like a simple deletion

michaelsproul:
@mergify queue

michaelsproul:
@mergify requeue

PR #6632, opened 2024-11-29
Test: Add a test for processing multiple deposits within the same epoch for a given validator

bshastry:
## Issue Addressed

https://github.com/ethereum/consensus-specs/issues/4021

## Proposed Changes

The idea is to create pre/post electra tests that do the following
- create a deposit in block N, epoch E for validator X
- create another deposit in block N + 3, epoch E for validator X
- assert that the balance of the validator after processing the deposits in the epoch equals expectation

## Additional Info

The test case pre electra works. I have not been able to correctly model state transition to electra (this test currently panics due to the mock execution layer not being able to provide a valid payload), so help there would be appreciated. I have hence maked this PR, draft.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -278,5 +278,8 @@ lto = "fat"
 codegen-units = 1
 incremental = false
 
+[profile.release]
+debug = true  # Enables debug symbols
+
 [patch.crates-io]
 quick-protobuf = { git = "https://github.com/sigp/quick-protobuf.git", rev = "681f413312404ab6e51f0b46f39b0075c6f4ebfd" }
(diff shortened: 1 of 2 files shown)


bshastry:
This PR is waiting on CI workflow approval I think

bshastry:
The test case for Electra fails because I don't know how to process deposit requests, can you please guide me?

```
per_block_processing::tests::test_multiple_deposits_same_validator_same_epoch_post_electra ---
thread 'per_block_processing::tests::test_multiple_deposits_same_validator_same_epoch_post_electra' panicked at consensus/state_processing/src/per_block_processing/tests.rs:1322:5:
Found the following validation errors:
Validator 1 balance is 32001788855 gwei, expected 34001788855 gwei (initial: 32001788855 + deposits: 2000000000)
```

michaelsproul:
@bshastry, I noticed the upstream issue on consenus-specs is closed. I think this scenario is handled by the tests added in:

- https://github.com/ethereum/consensus-specs/pull/4024

If you still think adding Lighthouse-specific tests would be useful, I can help you write them. But my general feeling is that these sorts of tests are best included in `consensus-specs` so that all clients can run them.

bshastry:
Thank you for taking a look @michaelsproul 

Feel free to close this PR.

michaelsproul:
Sorry it took so long for me to get to! Lots going on

PR #6634, opened 2024-11-29
Use RpcSend on RPC::self_limiter::ready_requests

jxs:
## Issue Addressed

We don't need to store `BehaviourAction` for `ready_requests` and therefore avoid having an `unreachable!` on #6625.
Therefore this PR should be merged before it


ackintosh (commented):
Thanks for this @jxs! This is helpful for simplifying #6625, but do we still need `unreachable!` on #6625 like this?:

```rust
                if let RPCSend::Request(request_id, req) = event {
                    //
                } else {
                    unreachable!()
                }
```

jxs (commented):
> Thanks for this @jxs! This is helpful for simplifying #6625, but do we still need `unreachable!` on #6625 like this?:
> 
> ```rust
>                 if let RPCSend::Request(request_id, req) = event {
>                     //
>                 } else {
>                     unreachable!()
>                 }
> ```

we can just return `false` on that situation right? (same as before true)

ackintosh:
It will look like this if we return `false` instead of `unreachable!` on #6625:

```rust
    pub fn peer_disconnected(&mut self, peer_id: PeerId) -> Vec<(Id, Protocol)> {
        // ...

        self.ready_requests.retain(|(req_peer_id, event)| {
            if let RPCSend::Request(request_id, req) = event {
                if req_peer_id == &peer_id {
                    failed_requests.push((*request_id, req.protocol()));
                    // Remove the entry
                    false
                } else {
                    // Keep the entry
                    true
                }
            } else {
                // unreachable!()

                // Remove the unexpected RPCSend variant 
                false
            }
        });
```

We can do that, but I think `unreachable!` is more appropriate in this situation to clearly indicate that the `else` clause is unreachable.

ackintosh (approved):

ackintosh (approved):

Issue #5880, opened 2024-06-03
Blank line at the end of graffiti-file causes error


davidcann:
## Description

A blank line at the end of the graffiti-file causes an error and the entire file to be ignored. There is no mention of blank lines being a problem in the documentation here: https://lighthouse-book.sigmaprime.io/graffiti.html

## Version

stable (production release)

## Present Behaviour

The entire graffiti-file is ignored when there's a blank line at the end of the text file.

    error: InvalidLine("Missing delimiter: "), service: block

## Expected Behaviour

Blank lines should be ignored. All valid lines should be read.

## Steps to resolve

Add a blank line to the end of the file set via --graffiti-file, then wait for a block proposal or restart the validator client.

davidcann:
To solve this, add this before /validator_client/src/graffiti_file.rs line 68:

```
if line.trim().is_empty() {
    continue;
}
```

https://github.com/sigp/lighthouse/blob/3058b96f2560f1da04ada4f9d8ba8e5651794ff6/validator_client/src/graffiti_file.rs#L68

michaelsproul:
thanks for the bug report

if you'd like to submit a PR with the fix we can merge it!

shayanb:
@michaelsproul I took the liberty of submitting a PR for this issue :)

PR #6635, opened 2024-11-30
Fix for blank line in graffiti file

shayanb:
## Issue Addressed

https://github.com/sigp/lighthouse/issues/5880

## Proposed Changes

Skip empty line as described in the issue 5880

## Additional Info

### DIFF
--- a/validator_client/graffiti_file/src/lib.rs
+++ b/validator_client/graffiti_file/src/lib.rs
@@ -66,6 +66,9 @@ impl GraffitiFile {
 
         for line in lines {
             let line = line.map_err(|e| Error::InvalidLine(e.to_string()))?;
+            if line.trim().is_empty() {
+                continue;
+            }
             let (pk_opt, graffiti) = read_line(&line)?;
             match pk_opt {
                 Some(pk) => {
--- a/validator_client/graffiti_file/src/lib.rs
+++ b/validator_client/graffiti_file/src/lib.rs
@@ -136,9 +136,15 @@ mod tests {
     const CUSTOM_GRAFFITI1: &str = "custom-graffiti1";
     const CUSTOM_GRAFFITI2: &str = "graffitiwall:720:641:#ffff00";
     const EMPTY_GRAFFITI: &str = "";
+    // Newline test cases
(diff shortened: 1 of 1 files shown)


michaelsproul:
Hi @shayanb, do you think you could try to add a test too? You can extend or copy the existing test in that file

shayanb:
> Hi @shayanb, do you think you could try to add a test too? You can extend or copy the existing test in that file

Hi @michaelsproul , yes for sure. There are many places I could implement the test, my guess is that it fits in `lighthouse/tests/validator_client.rs`, I'll push the change shortly but please do let me know if the test should be implemented somewhere else.

shayanb:
@michaelsproul here you go, added tests for different new line scenarios. Let me know if anything else is needed here.

shayanb:
I'd like to add printing the graffiti in the logs for validator initialization, would it fit in this PR or I can create a new PR for that.

Basically this line in `validator_client/initialized_validators/src/lib.rs` (although not sure how to test it yet):

```
                                self.validators
                                    .insert(init.voting_public_key().compress(), init);
                                info!(
                                    self.log,
                                    "Enabled validator";
                                    "signing_method" => "local_keystore",
                                    "voting_pubkey" => format!("{:?}", def.voting_public_key),
                                    **"graffiti" => format!("{:?}", def.graffiti)**
                                );
```

michaelsproul:
@shayanb Thanks for the test, looks good!

I think we could add the graffiti logging in a separate PR. There's some subtlety there because the graffiti can change at runtime, so whatever we log on startup won't necessarily stay. It could still be nice to log it though

michaelsproul:
Cargo fmt is failing, but you can fix it with `cargo fmt --all` (you can probably also configure your editor to run `cargo fmt` on save).

shayanb:
@michaelsproul thanks for the tip, pushed the fmt commit.

I'll submit graffiti logging PR soon.

michaelsproul (approved):
Looks great, thanks

The CI failure is unrelated, will be fixed by:

- https://github.com/sigp/lighthouse/pull/6646

michaelsproul:
@mergify queue

michaelsproul:
@mergify requeue

PR #6646, opened 2024-12-02
Correct flakey CI tests

AgeManning:
## Issue Addressed

Some subnet service tests were failing occasionally. There were two bugs that I found. 

I've run these tests repeatedly without fail, so at the very least, should make these more stable.


michaelsproul:
Clippy is failing

michaelsproul:
Does the `test_subscribe_same_subnet_several_slots_apart` test need the same fix as the same_slot test? It is still failing on this branch: https://github.com/sigp/lighthouse/actions/runs/12114467475/job/33771078350?pr=6646

jimmygchen (approved):
Looks good to me. This should also fix the test @michaelsproul mentioned above.

michaelsproul:
Rebased on `release-v6.0.1` so we can merge there and then back-merge to unstable

michaelsproul:
@mergify queue

PR #6650, opened 2024-12-03
Remove heading that isn't rendered correctly

michaelsproul:
## Proposed Changes

Fix the header formatting which is not being applied correctly:

https://lighthouse-book.sigmaprime.io/security.html

We only need the `#` at the start to make `Security` the heading (`h1`) for the page.

### DIFF
--- a/book/src/security.md
+++ b/book/src/security.md
@@ -1,6 +1,5 @@
 # Security
 
-========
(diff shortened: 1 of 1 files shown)


chong-he:
LGTM!

michaelsproul:
@mergify queue

macladson (approved):
LGTM

Issue #6644, opened 2024-12-02
Remove promises after sending IDONTWANT


AgeManning:
## Description

Currently if we send an IWANT for a message-id and then later receive that message via a mesh peer, we send an IDONTWANT to these peers to try and cancel the IWANT message. 

Our implementation (from what I can tell) doesn't remove the promise, and when the peer we sent an IDONTWANT to rightfully refuses to send a message, we will downscore them for breaking their promise. 

We should remove the promise if we send an IDONTWANT. 

We get the peers we sent an IWANT to from here: https://github.com/sigp/lighthouse/blob/stable/beacon_node/lighthouse_network/gossipsub/src/behaviour.rs#L2698

We should remove this peer from this mapping when we send an IDONTWANT to them somewhere here: https://github.com/sigp/lighthouse/blob/unstable/beacon_node/lighthouse_network/gossipsub/src/behaviour.rs#L2773

AgeManning:
This is my bad. I think we already check this when handling messages.

PR #6651, opened 2024-12-03
Add remove_promise function to GossipPromises and call function when message received from peer prior to sending IDONTWANT

jking-aus:
## Issue Addressed

#6644 

## Proposed Changes

To address the issue raised in #6644 where we don't remove IWANT promises from peers after sending IDONTWANT: 

- Add remove_promise function to GossipPromises in the lighthouse_network crate. This function removes a given peer's promise and cleans up empty message entries.  
- Call this function prior to sending IDONTWANT to avoid incorrectly penalising peers. 

## Additional Info

Also includes tests.


jking-aus:
code already handles this behaviour -- PR not required

PR #6656, opened 2024-12-04
Allow importing of historical blobs via HTTP API

macladson:
## Issue Addressed

https://github.com/sigp/lighthouse/issues/5391

## Proposed Changes

Allows importing historical blobs into the blobs-db via the `/lighthouse/database/import_blobs` endpoint.
Supports SSZ through `/lighthouse/database/import_blobs_ssz`.

In both cases, the required format must deserialize to `Vec<BlobSidecarList<E>>`

Performs the following checks:
 - Checks that `block_root` is consistent across each `BlobSidecarList`.
 - Checks that `block_root` exists in the database.
 - Checks if a `BlobSidecarList` is already stored for that `block_root`. If it is, ensure it matches the `BlobSidecarList` we are attempting to store.


beacon_node/store/src/hot_cold_store.rs
```diff
@@ -43,6 +43,8 @@ use types::data_column_sidecar::{ColumnIndex, DataColumnSidecar, DataColumnSidec
 use types::*;
 use zstd::{Decoder, Encoder};
 
+const HISTORICAL_BLOB_BATCH_SIZE: usize = 1000;
```
michaelsproul: I think lets get rid of the batch size so that the entire Vec provided by the caller either commits or doesn't

beacon_node/http_api/src/lib.rs
```diff
@@ -4422,6 +4424,141 @@ pub fn serve<T: BeaconChainTypes>(
             },
         );
 
+    // POST lighthouse/database/verify_blobs
```
michaelsproul: stray comment

beacon_node/http_api/src/lib.rs
```diff
@@ -4422,6 +4424,141 @@ pub fn serve<T: BeaconChainTypes>(
...
+                            }
+                        }
+                    }
+
+                    if results.is_empty() {
+                        Ok(api_types::GenericResponse::from(
+                            "All blobs verified successfully".to_string(),
+                        ))
+                    } else {
+                        Ok(api_types::GenericResponse::from(results.join("\n")))
```
michaelsproul: this looks like it would be a 200, but should probably be a 500?

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -844,6 +844,93 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
+        }
+
+        let mut total_imported = 0;
+
+        let mut ops = vec![];
+
+        for blob_sidecar_list in historical_blob_sidecars {
+            // Ensure all block_roots in the blob list are the same.
+            let block_root = {
+                let first_block_root = blob_sidecar_list[0].block_root();
```
michaelsproul, line 865: Need to handle empty `blob_sidecar_list` safely here

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -844,6 +844,93 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
+                warn!(
+                    %block_root,
+                    num_blob_sidecars = blob_sidecar_list.len(),
+                    "Aborting blob import; block does not exist"
+                );
+                return Err(Error::InvalidBlobImport("Missing block".to_string()));
+            }
+
+            // Check if a `blob_sidecar_list` is already stored for this block root.
+            match self.get_blobs(&block_root) {
```
michaelsproul, line 890: Lets try adding a log with the time that this step takes

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -844,6 +844,93 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
+                {
+                    return Err(Error::InvalidBlobImport(
+                        "Inconsistent block roots".to_string(),
+                    ));
+                }
+                first_block_root
+            };
+
+            // Check block is stored for this block_root.
+            if !self.block_exists(&block_root)? {
```
michaelsproul, line 878: Could check blob consistency with the block here (when `verify == true`)

beacon_node/http_api/src/lib.rs
```diff
@@ -4638,6 +4644,205 @@ pub fn serve<T: BeaconChainTypes>(
...
+        .and(chain_filter.clone())
+        .then(
+            |query: api_types::ImportBlobsQuery,
+             blob_lists: Vec<BlobSidecarList<T::EthSpec>>,
+             task_spawner: TaskSpawner<T::EthSpec>,
+             chain: Arc<BeaconChain<T>>| {
+                task_spawner.blocking_json_task(Priority::P1, move || {
+                    if query.verify == Some(true) {
+                        for blob_list in &blob_lists {
+                            match verify_kzg_for_blob_list(blob_list.iter(), &chain.kzg) {
```
michaelsproul: For completeness we should also check `verify_blob_sidecar_inclusion_proof` for each blob.

common/eth2/src/lighthouse.rs
```diff
@@ -368,6 +372,58 @@ impl BeaconNodeHttpClient {
...
+
+        path.query_pairs_mut()
+            .append_pair("end_slot", &end_slot.to_string());
+
+        if let Some(verify) = verify {
+            path.query_pairs_mut()
+                .append_pair("verify", &verify.to_string());
+        }
+
+        self.get(path).await
```
michaelsproul: Should probably turn off the timeout here

beacon_node/store/src/hot_cold_store.rs
michaelsproul: Not necessary if we do the KZG inclusion proof verification in `http_api/lib.rs`

database_manager/src/blobs_manager/export.rs
```diff
@@ -0,0 +1,151 @@
...
+    spec: &ChainSpec,
+) -> Result<(), String> {
+    let beacon_node = SensitiveUrl::parse(&config.beacon_node)
+        .map_err(|e| format!("Unable to parse beacon node url: {e:?}"))?;
+
+    let client = BeaconNodeHttpClient::new(beacon_node, Timeouts::set_all(Duration::from_secs(12)));
+
+    let (_, is_synced) = ensure_node_synced(&client).await?;
+    if !is_synced {
+        if config.allow_unsynced {
```
michaelsproul: Lets move `allow_unsynced` above `ensure_node_synced` so we don't hit the BN at all (and can't timeout)

database_manager/src/blobs_manager/import.rs
```diff
@@ -0,0 +1,45 @@
...
+
+pub async fn import_blobs<E: EthSpec>(config: &ImportBlobs) -> Result<(), String> {
+    let beacon_node = SensitiveUrl::parse(&config.beacon_node)
+        .map_err(|e| format!("Unable to parse beacon node url: {e:?}"))?;
+
+    let client = BeaconNodeHttpClient::new(beacon_node, Timeouts::set_all(Duration::from_secs(12)));
+
+    let (_, is_synced) = ensure_node_synced(&client).await?;
+    if !is_synced {
+        if config.allow_unsynced {
```
michaelsproul: As for export, lets move this above `ensure_node_synced`.

database_manager/src/blobs_manager/import.rs
```diff
@@ -0,0 +1,45 @@
...
+            warn!("Beacon node is not synced");
+        } else {
+            return Err("Beacon node is not synced".to_string());
+        }
+    }
+
+    let blobs_ssz = std::fs::read(&config.input_file)
+        .map_err(|e| format!("Failed to read input file: {e:?}"))?;
```
michaelsproul, line 31: We should add batching here so we can experiment with importing less blobs at a time, and see if that prevents overloading the BN.

database_manager/src/blobs_manager/import.rs
```diff
@@ -0,0 +1,45 @@
...
+    // generated during export.
+
+    info!("Beginning blob import");
+
+    if config.skip_verification {
+        warn!("Skipping blob verification");
+    }
+
+    client
+        .post_lighthouse_database_import_blobs_ssz(blobs_ssz.into(), config.skip_verification)
```
michaelsproul: This func could take a vec of `BlobSidecarList` and handle the conversion back to SSZ internally

dapplion:
- Closing for now, revive if there's still interest. https://github.com/sigp/lighthouse/issues/9214

PR #6658, opened 2024-12-05
Stuck lookup v6

dapplion:
## Issue Addressed

PR 
- https://github.com/sigp/lighthouse/pull/6443

Introduced a regression on the lookup pruning logic. If a lookup loses all peers before the block request downloads the block, it will get stuck. This is a relatively frequent occurrence. This bug will result in occasional warn logs on the user's terminals.

**Example case events**

- Lookup created for block A, from peer X
- Peer X disconencts, lookup A now has 0 peers
- 10 seconds pass
- `drop_lookups_without_peers` runs on lookup A
  - `has_no_peers` returns true
  - elapsed condition returns true
  - `is_awaiting_event` returns true <- so it doesn't get pruned

`is_awaiting_event` checks:
- `awaiting_parent.is_some()` returns false
- `self.block_request_state.state.is_awaiting_event()` returns false (block awaiting download)
- `component_requests` is `WaitingForBlock`, so it returns true <- the bug is here

https://github.com/sigp/lighthouse/blob/193c7f82411bb7c28cec1959e2d7682d72d1452b/beacon_node/network/src/sync/block_lookups/single_block_lookup.rs#L174-L177

## Proposed Changes

Fix `is_awaiting_event` logic.

### DIFF
--- a/beacon_node/network/src/sync/block_lookups/single_block_lookup.rs
+++ b/beacon_node/network/src/sync/block_lookups/single_block_lookup.rs
@@ -171,7 +171,10 @@ impl<T: BeaconChainTypes> SingleBlockLookup<T> {
         self.awaiting_parent.is_some()
             || self.block_request_state.state.is_awaiting_event()
             || match &self.component_requests {
-                ComponentRequests::WaitingForBlock => true,
+                // If components are waiting for the block request to complete, here we should
+                // check if the`block_request_state.state.is_awaiting_event(). However we already
+                // checked that above, so `WaitingForBlock => false` is equivalent.
+                ComponentRequests::WaitingForBlock => false,
                 ComponentRequests::ActiveBlobRequest(request, _) => {
                     request.state.is_awaiting_event()
                 }


michaelsproul (approved):
Makes sense to me, and testing has shown that this greatly reduces the frequency of stuck lookups.

michaelsproul:
@mergify queue

beacon_node/network/src/sync/block_lookups/single_block_lookup.rs
```diff
@@ -171,7 +171,10 @@ impl<T: BeaconChainTypes> SingleBlockLookup<T> {
         self.awaiting_parent.is_some()
             || self.block_request_state.state.is_awaiting_event()
             || match &self.component_requests {
-                ComponentRequests::WaitingForBlock => true,
+                // If components are waiting for the block request to complete, here we should
+                // check if the`block_request_state.state.is_awaiting_event(). However we already
+                // checked that above, so `WaitingForBlock => false` is equivalent.
+                ComponentRequests::WaitingForBlock => false,
```
jimmygchen, line 177: I think this makes sense.

I wonder if this would drop any lookups that are awaiting processing?
https://github.com/sigp/lighthouse/blob/886d409b858a63f6a811abd9cf60e2a4d6024bc2/beacon_node/network/src/sync/block_lookups/single_block_lookup.rs#L500

jimmygchen: This is unlikely an issue - if we do need the block, external events should somehow keep this lookup in place, a single disconnected peer should not affect this.

PR #6661, opened 2024-12-05
Remove `ZeroizeString` in favour of `Zeroizing<String>`

dknopik:
This is the first part of my effort to make the `eth2` crate more lightweight.

## Proposed Changes

Remove `ZeroizeString` from `account_utils`, which uses the `Zeroize` derive macro from the `zeroize` crate, and use the `Zeroizing<T>` newtype provided by `zeroize` instead.

Our newtype did not add any value, except providing `without_newlines`, which replaced a one-liner and was used in only one location (outside of tests).

## Additional Info

`eth2` was depending on `account_utils` exclusively for `ZeroizeString`, which was the inspiration for this PR. Removing the `account_utils` dependency immediately made the crate lighter: `cargo tree -p eth2 | wc -l` went from `1995` down to `1865`.

There are more newtypes scattered in the codebase that _could_ be removed like this, but some arguably improve code readability by providing context specific names, such as `DerivedKey` and `LamportSecretKey`, so I left them alone for now.

### DIFF
--- a/beacon_node/store/Cargo.toml
+++ b/beacon_node/store/Cargo.toml
@@ -33,6 +33,7 @@ zstd = { workspace = true }
 bls = { workspace = true }
 smallvec = { workspace = true }
 logging = { workspace = true }
+#eth2 = { workspace = true }
 
 [[bench]]
 name = "hdiff"
--- a/beacon_node/store/Cargo.toml
+++ b/beacon_node/store/Cargo.toml
@@ -33,7 +33,6 @@ zstd = { workspace = true }
 bls = { workspace = true }
 smallvec = { workspace = true }
 logging = { workspace = true }
-#eth2 = { workspace = true }
 
 [[bench]]
 name = "hdiff"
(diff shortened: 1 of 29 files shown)


beacon_node/store/Cargo.toml
```diff
@@ -33,6 +33,7 @@ zstd = { workspace = true }
 bls = { workspace = true }
 smallvec = { workspace = true }
 logging = { workspace = true }
+#eth2 = { workspace = true }
```
michaelsproul:
```suggestion
```

michaelsproul (approved):
Looks good, thanks!

beacon_node/store/Cargo.toml
michaelsproul: I can't seem to commit this suggestion, but if you commit it we can merge shortly. After this PR is back-merged to unstable to fix CI:

- https://github.com/sigp/lighthouse/pull/6671

dknopik: ah, missed that line, good catch

sorry, will allow edits in the future!

pawanjay176 (approved):
This is great! Let's lighten them dependencies

michaelsproul:
@mergify queue

PR #6662, opened 2024-12-05
Log graffiti on validator initialization

shayanb:
## Issue Addressed

No related issues found, but this feature was discussed shortly with @michaelsproul here: https://github.com/sigp/lighthouse/pull/6635#issuecomment-2510303659

## Proposed Changes

Print out the initial graffiti set to the logs when initializing the validators. 

## Additional Info

This can help to see if the initial graffiti is set correctly.

I wasn't able to find an easy way to test this, however the test suite are all passing locally.


validator_client/initialized_validators/src/lib.rs
```diff
@@ -1285,6 +1284,7 @@ impl InitializedValidators {
                                     "Enabled validator";
                                     "signing_method" => "local_keystore",
                                     "voting_pubkey" => format!("{:?}", def.voting_public_key),
+                                    "graffiti" => format!("{:?}", def.graffiti)
```
michaelsproul, line 1287: Just noting that this will only log the graffiti for a validator if it is specified in the `validator_definitions.yml`, not other sources (e.g. graffiti file).

We're also printing the `Debug` representation here which will be a bit ugly like `None` or `Some("graffiti")`. Instead it would be good to format the `None` case explicitly and just display the string if it is `Some`. Something like `def.graffiti.map_or("none".to_string(), |graffiti| graffiti.to_string())`. You may need to implement `Display` for `GraffitiString`.

PR #6664, opened 2024-12-06
Enable lints for tests only running optimized

dknopik:
## Issue Addressed

Previously, tests gated with `#[cfg(not(debug_assertions))]` were not linted, as Clippy does not run in `--release` mode to save time. 

## Proposed Changes

This PR enables linting these tests by setting the compiler option `-C debug-assertions=no` when invoking Clippy. This retains the time saved by not optimizing while linting. Also, this PR fixes allllll the lint errors uncovered by this.

## Additional Info

For easier review, this PR is split into several commits:

1. **[fix automatically fixable or obvious lints](https://github.com/sigp/lighthouse/commit/a22e66a7efbd0b9dcff4f7620585323df11ccd3e)**: Most of the new errors were "machine fixable" (stuff like superfluous `&` or `.into()`s). Some errors were not machine fixable but really obvious to fix anyway (at my discretion). Feel free to scroll through this anyway. I checked machine fixable occurrences manually to make sure they're sane.
2. **[fix `suspicious_open_options` by removing manual options](https://github.com/sigp/lighthouse/commit/b5a8bd77497619544dad3f75b57ec5a2b0a07654)**: Here I modified the code beyond the scope of the lint to make it a bit nicer, but that is a matter of taste.
3. **[fix `await_holding_locks`](https://github.com/sigp/lighthouse/commit/3c872c95ad29787f0b1535b65bc8a063113edc1a)**: Here we held locks across await points, so I had to shuffle things around a bit. In one case, it was annoying to avoid as `NaiveAggregationPool` is not `Clone`able, so I disabled the lint here (as there is some precedent for that in the codebase). (edit: I now realize one lock is used for inter-test ratelimiting, so I reverted and disabled the lint there in a later commit)
4. **[avoid failing lint due to now disabled `#[cfg(debug_assertions)]`](https://github.com/sigp/lighthouse/commit/5fd8d37928db1a07379a4920b7e0e1c6b6425f13)**: This one is interesting: The newly set compiler flag caused an issue in this non-test code as the `None` branch was now empty. I pulled the check out of the `match` and then applied the lint suggestion. (edit: By accident, I flipped the condition which I fixed in a later commit) 
5. **[reduce future sizes in tests](https://github.com/sigp/lighthouse/commit/cdef1ffa1c706da14c17bda6a994e0d826a6cff4)**: Some async test functions now had a too large stack size. I mostly sprinkled some `Box::pin` around some of the invoked futures, and split up the test in one case.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -204,7 +204,7 @@ test-full: cargo-fmt test-release test-debug test-ef test-exec-engine
 # Lints the code for bad style and potentially unsafe arithmetic using Clippy.
 # Clippy lints are opt-in per-crate for now. By default, everything is allowed except for performance and correctness lints.
 lint:
-	cargo clippy --workspace --benches --tests $(EXTRA_CLIPPY_OPTS) --features "$(TEST_FEATURES)" -- \
+	RUSTFLAGS="-C debug-assertions=no $(RUSTFLAGS)" cargo clippy --workspace --benches --tests $(EXTRA_CLIPPY_OPTS) --features "$(TEST_FEATURES)" -- \
 		-D clippy::fn_to_numeric_cast_any \
 		-D clippy::manual_let_else \
 		-D clippy::large_stack_frames \
--- a/account_manager/src/validator/exit.rs
+++ b/account_manager/src/validator/exit.rs
@@ -409,6 +409,6 @@ mod tests {
         )
         .unwrap();
 
-        assert_eq!(expected_pk, kp.pk.into());
+        assert_eq!(expected_pk, kp.pk);
     }
 }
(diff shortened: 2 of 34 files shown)


dapplion:
This is a great addition!

michaelsproul (approved):
Looks great, thanks!

Regarding `await_holding_lock` I think it would be good to fix all of these occurrences eventually, as they could lead to flaky tests/random deadlocks. I think the easiest way to do it is to swap the lock out for a `tokio` one, which makes the lock acquisition itself `async`.

The only reason we have this lint `allow`'d in the VC is that it never seems to have caused problems in practice, and it would be a decently complicated refactor to make that lock async.

michaelsproul:
Merging this now before it gets into conflict with too many other PRs

michaelsproul:
@mergify queue

PR #6667, opened 2024-12-07
Compact more when pruning states

michaelsproul:
## Proposed Changes

@chong-he noticed that nodes running v6.0.0 do not reclaim disk space after pruning states. This is because the generic `compact` method on the database only compacts a handful of columns: `BeaconState`, `BeaconStateSummary`, `BeaconBlock`.

For pre-v6.0.0 databases, most of the data for historic states will have been stored in the `BeaconState` column, so users upgrading from v5.x to v6 should have been able to reclaim most of the used space. There are only a handful of old columns like `BeaconStateRoots` which won't have been compacted.

For nodes running v6 that prune states after upgrading, the compaction currently misses the two most important columns: `BeaconStateSnapshot` and `BeaconStateDiff`. I've added a new `compact_freezer` method which compacts all of the current-schema columns, and all the old schema columns for good measure.

### DIFF
--- a/beacon_node/store/src/hot_cold_store.rs
+++ b/beacon_node/store/src/hot_cold_store.rs
@@ -2484,6 +2484,45 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold> 
         Ok(())
     }
 
+    /// Run a compaction pass on the freezer DB to free up space used by deleted states.
+    pub fn compact_freezer(&self) -> Result<(), Error> {
+        let current_schema_columns = vec![
+            DBColumn::BeaconColdStateSummary,
+            DBColumn::BeaconStateSnapshot,
+            DBColumn::BeaconStateDiff,
+            DBColumn::BeaconStateRoots,
+        ];
+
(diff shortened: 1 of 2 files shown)


beacon_node/store/src/hot_cold_store.rs
```diff
@@ -2484,6 +2484,45 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
+        // We can remove this once schema V21 has been gone for a while.
+        let previous_schema_columns = vec![
+            DBColumn::BeaconState,
+            DBColumn::BeaconStateSummary,
+            DBColumn::BeaconBlockRootsChunked,
+            DBColumn::BeaconStateRootsChunked,
+            DBColumn::BeaconRestorePoint,
+            DBColumn::BeaconHistoricalRoots,
+            DBColumn::BeaconRandaoMixes,
+            DBColumn::BeaconHistoricalSummaries,
```
dapplion, line 2505: Hope we don't make the same mistake when rolling diffs on hot DB

dapplion (approved):

michaelsproul:
@mergify queue

PR #6673, opened 2024-12-09
Set Java Version to 21 to fix Web3Signer tests

jimmygchen:
## Issue Addressed

Latest [Web3Signer release 24.12.0 ](https://github.com/Consensys/web3signer/releases/tag/24.12.0)requires Java 21 runtime. 

GitHub runners uses Java 11 by default, hence Web3Signer fails to start in our release tests. The runner already has Java 21 pre-installed, this PR sets the used Java version to 21.


jimmygchen:
Oh just realised we have Java 19 installed in our custom runner image, we should update the version there instead.

jimmygchen:
Made another attempt as a workaround for now: https://github.com/sigp/lighthouse/pull/6674

PR #6671, opened 2024-12-09
Fix Kurtosis, web3signer and cargo-audit for CI

michaelsproul:
## Issue Addressed

Fix the failing Kurtosis jobs on CI, e.g.

https://github.com/sigp/lighthouse/actions/runs/12226785831/job/34102825704?pr=6669

## Proposed Changes

- Kurtosis fix: Remove the `-0` that was being erroneously added to the artifact name reference in the doppelganger script. Something must have changed upstream to break this.
- Web3signer fix: update to Java 21.
- Cargo audit: temporarily ignore vuln for `idna` until upstream patch is available.
- May-as-well: install the latest version of `kurtosis-cli`, rather than pinning `1.3.1`.

PR #6674, opened 2024-12-10
Set Java Version to 21 (required since Web3Signer 24.12.0).

jimmygchen:
## Issue Addressed

Latest [Web3Signer release 24.12.0 ](https://github.com/Consensys/web3signer/releases/tag/24.12.0) requires Java 21 runtime. 

Our GitHub runners uses Java 19, hence Web3Signer fails to start in our release tests. This PR install Java 21 using the `setup-java` action. We could also include this in our runner image to avoid downloading during CI runs.

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -83,6 +83,11 @@ jobs:
     runs-on: ${{ github.repository == 'sigp/lighthouse' && fromJson('["self-hosted", "linux", "CI", "large"]') || 'ubuntu-latest'  }}
     steps:
     - uses: actions/checkout@v4
+    # Set Java version to 21. (required since Web3Signer 24.12.0).
+    - uses: actions/setup-java@v4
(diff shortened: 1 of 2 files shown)


michaelsproul:
I'm going to cherry-pick your commit into this PR:

- https://github.com/sigp/lighthouse/pull/6671

Then merge that to `release-6.0.1` and back-merge that to `unstable`.

antondlr:
how did I miss that :-( 
I'll add it to our runner images. GH hosted runners have it already. afterwards, we can either revert or leave it there (my preference would be to revert).

antondlr:
[bumped](https://github.com/sigp/sigp-infra/commit/fcaa09d6fdcfc2c9463965cbaa7496e5a0004543), pulled,  and cycled on all runners. we can rollback this additional step now.
(we use openjre in the runners as opposed to Temurin here.. should we change that? )

PR #6675, opened 2024-12-10
Fix subnet service test

jimmygchen:
## Issue Addressed

This PR fixes an intermittently failing test:
https://github.com/sigp/lighthouse/actions/runs/12246911324/job/34163756917

It fails when the validator subnets happens to be in the permanent subscriptions - the existing check doesn't actually handle this correctly as `subnet_service.is_subscribed` checks for short-lived subscription. In the case where the subnet is already subscribed as a permanent subscription, there will be no new subscription events.

### DIFF
--- a/beacon_node/network/src/subnet_service/tests/mod.rs
+++ b/beacon_node/network/src/subnet_service/tests/mod.rs
@@ -166,9 +166,9 @@ async fn get_events<S: Stream<Item = SubnetServiceMessage> + Unpin>(
 }
 
 mod test {
-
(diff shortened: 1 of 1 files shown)


michaelsproul:
Age has a PR for this here:

- https://github.com/sigp/lighthouse/pull/6646

Solution seems similar. Maybe we can combine the two?

jimmygchen:
ah right yes it's the same thing with additional fixes, I'll close this one in favour of his one. Thanks!

PR #6677, opened 2024-12-10
Correct `/nat` API for libp2p

AgeManning:
## Issue Addressed

There was a mismatch in the metric naming, resulting in the `/nat` api always returning false, when the underlying metric may actually be true.

### DIFF
--- a/beacon_node/lighthouse_network/src/peer_manager/network_behaviour.rs
+++ b/beacon_node/lighthouse_network/src/peer_manager/network_behaviour.rs
@@ -141,10 +141,6 @@ impl<E: EthSpec> NetworkBehaviour for PeerManager<E> {
                 debug!(self.log, "Failed to dial peer"; "peer_id"=> ?peer_id, "error" => %ClearDialError(error));
                 self.on_dial_failure(peer_id);
             }
-            FromSwarm::ExternalAddrConfirmed(_) => {
-                // We have an external address confirmed, means we are able to do NAT traversal.
-                metrics::set_gauge_vec(&metrics::NAT_OPEN, &["libp2p"], 1);
-            }
             _ => {
                 // NOTE: FromSwarm is a non exhaustive enum so updates should be based on release
                 // notes more than compiler feedback
--- a/common/system_health/src/lib.rs
+++ b/common/system_health/src/lib.rs
@@ -235,14 +235,14 @@ pub fn observe_nat() -> NatState {
 
     let libp2p_ipv4 = lighthouse_network::metrics::get_int_gauge(
         &lighthouse_network::metrics::NAT_OPEN,
-        &["libp2p"],
+        &["libp2p_ipv4"],
     )
     .map(|g| g.get() == 1)
     .unwrap_or_default();
 
     let libp2p_ipv6 = lighthouse_network::metrics::get_int_gauge(
         &lighthouse_network::metrics::NAT_OPEN,
-        &["libp2p"],
+        &["libp2p_ipv6"],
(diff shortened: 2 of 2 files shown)


michaelsproul:
Can you rebase this on `release-v6.0.1` branch so we can include it in v6.0.1?

beacon_node/lighthouse_network/src/peer_manager/network_behaviour.rs
```diff
@@ -141,10 +141,6 @@ impl<E: EthSpec> NetworkBehaviour for PeerManager<E> {
                 debug!(self.log, "Failed to dial peer"; "peer_id"=> ?peer_id, "error" => %ClearDialError(error));
                 self.on_dial_failure(peer_id);
             }
-            FromSwarm::ExternalAddrConfirmed(_) => {
-                // We have an external address confirmed, means we are able to do NAT traversal.
-                metrics::set_gauge_vec(&metrics::NAT_OPEN, &["libp2p"], 1);
-            }
```
jimmygchen, line 147: Nice catch. This is now set in [`handle_established_inbound_connection`](https://github.com/sigp/lighthouse/blob/8cebc87d956707a385aa21aa75ef748e6391a586/beacon_node/lighthouse_network/src/peer_manager/network_behaviour.rs#L212-L218) instead since #6486.

jimmygchen (approved):
Nice catch! Just need to rebase this as Michael mentioned so we can include it in 6.0.1.

beacon_node/lighthouse_network/src/peer_manager/network_behaviour.rs
jxs: this shouldn't be removed, it tells us when `libp2p-upnp` was able to successfully map the address and port via UPnP on the router and therefore NAT traversal is activated. 
I had missed the code Jimmy linked it, but we receiving an inbound connection doesn't tell us about our NAT traversal settings in libp2p, we may be reached by a peer with a public address that we established a connection before and so the port may still be mapped in the router right?

AgeManning: I wasn't sure of when this message actually gets sent. 
If this gets sent when upnp worked successfully, then we have some competing conditions about what we say about our libp2p nat being open. 

Which is a stronger condition? 
UPnP successful or incoming connections? 
Both have error cases I think.
-  UPnP could work for a local router, but if behind CGNAT or another router or something, you would not be contactable and not get incoming connections. Maybe we could open the wrong ports also?
- Incoming connections is fallible as you pointed out ( I think there is a 30s window or something for a peer to disconnect and then re-connect to us. We could increase the number of incoming connections to like 3 or 5 to lower the probability. 

To make it simple, I think its easier to reason about with incoming connections. Maybe we have some number of connections (in discovery its 2, i think). If UPnP worked, then I'd expect we would get incoming connections. 

The other thing we don't account for is, if after it worked once, we dont check again. Maybe we should also check if we get 0 incoming peers, we should make it as false?

AgeManning: Happy to have some mix if you prefer to keep the UPnP message. Is UPnP the only behaviour that sends this? Could this get sent from other behaviours that might be also fallible?

AgeManning:
This has been rebased

michaelsproul:
Rebased again to remove stuff from `unstable` that was still lurking

michaelsproul:
@mergify queue

PR #6678, opened 2024-12-10
Enforce alphabetically ordered cargo deps

macladson:
## Proposed Changes

This is merely cosmetic, but something that has annoyed me for a while is that dependencies are arbitrarily ordered in every crate which makes it difficult to see at a glace whether a certain dependency is in a crate.

Add `cargo-sort` to CI to enforce that dependencies are alphabetically ordered.

## Additional Changes

This will require devs to order dependencies from now on (or run `cargo sort --workspace`).


.github/workflows/test-suite.yml
macladson, line 1: My IDE mangled the test suite, and interestingly enough, it has prevented most of the checks from running :thinking:

dapplion:
I share the annoyance, enforcing sorting sounds good to me

michaelsproul (approved):

michaelsproul:
@mergify queue

Issue #6452, opened 2024-10-01
Modularize or Simplify the eth2 crate


AgeManning:
## Description

The `common/eth2` crate appears to import the whole world. Other crates that depend on eth2 are compiling and building libp2p and a whole host of other things. 

It would be great if we could significantly reduce the dependency chain here. Either by splitting up some of the structs or removing unnecessary imports.

dapplion:
I remember asking myself this when doing some PoC a long time ago that depended on `common/eth2`

dapplion:
The dependency on libp2p is only for these types

```rust
use lighthouse_network::{types::SyncState, PeerInfo};
use lighthouse_network::{ConnectionDirection, Enr, Multiaddr, PeerConnectionStatus};
use lighthouse_network::PeerId;
```

hopinheimer:
hi @AgeManning  do you think I can swoop in on this?

dapplion:
From an offline chat with @AgeManning we suggested to:
- Return the serialized ENR as a String to avoid the dependency on discv5
- Import PeerID from its own crate
- Move the misc types `SyncState`, `ConnectionDirection`, etc to their own crate of only types and make it a common dependency of `lighthouse_network` and `common/eth2`

SkandaBhat:
If nobody else is working on this, can I take a crack at this? @AgeManning @dapplion

AgeManning:
Hey @SkandaBhat -  Sorry we have been away at a conference. 

@hopinheimer are you still looking into this? Or should @SkandaBhat have a crack?

hopinheimer:
Hey @AgeManning sorry just reached back from the conference.  

hi @SkandaBhat please feel free to pick this up I haven't started work on it.

SkandaBhat:
Sure thanks! @AgeManning please feel free to assign this to me! I am available this week and can create smaller PRs towards this issue.

Gua00va:
> From an offline chat with @AgeManning we suggested to:
> 
> * Return the serialized ENR as a String to avoid the dependency on discv5
> * Import PeerID from its own crate
> * Move the misc types `SyncState`, `ConnectionDirection`, etc to their own crate of only types and make it a common dependency of `lighthouse_network` and `common/eth2`

@dapplion for the 3rd point, should a new crate be created or should I put it in an existing one?

dapplion:
> for the 3rd point, should a new crate be created or should I put it in an existing one?

You can add a new crate `network_types`

[...]

PR #6679, opened 2024-12-11
Remove lighthouse_network dependency from eth2 crate

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

The eth2 which contains our http api client crate depends on the lighthouse_network crate for some networking types.
This is completely unnecessary imo because we can just return strings/serde_json::Value types for the networking types. 
These types don't require us using any of the functions on the type structs. They just exist for some typing.

This is an enormous source of pain for any dependency that pulls the eth2 crate for using a `BeaconNodeHttpClient` because importing lighthouse_network pulls in a bunch of other dependencies that just bloats the dep tree.

I don't see any reason why the networking related types cannot be represented as `String` or json values. Let me know if that is not the case. 

## Todo

- [ ] Test all affected endpoints

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2571,7 +2571,6 @@ dependencies = [
  "futures",
  "futures-util",
  "libsecp256k1",
- "lighthouse_network",
  "mediatype",
  "pretty_reqwest_error",
  "procfs",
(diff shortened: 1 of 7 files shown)


AgeManning:
The same as this? https://github.com/sigp/lighthouse/issues/6452

pawanjay176:
Yeah there's already work being done on this. I'm closing this for now

PR #6680, opened 2024-12-11
decouple `eth2` from `store` and `lighthouse_network`

dknopik:
## Issue Addressed

- #6452 (partially)

## Proposed Changes

Remove dependencies on `store` and `lighthouse_network`  from `eth2`. This was achieved as follows:

- depend on `enr` and `multiaddr` directly instead of using `lighthouse_network`'s reexports.
- make `lighthouse_network` responsible for converting between API and internal types.
- in two cases, remove complex internal types and use the generic `serde_json::Value` instead - this is not ideal, but should be fine for now, as this affects two internal non-spec endpoints which are meant for debugging, unstable, and subject to change without notice anyway. Inspired by #6679. The alternative is to move all relevant types to `eth2` or `types` instead - what do you think?

## Additional Info

In my opinion, this only partially fixes the issue above, as we probably still should split out lighthouse-specific features.


common/eth2/src/lighthouse.rs
```diff
@@ -37,14 +34,12 @@ four_byte_option_impl!(four_byte_option_u64, u64);
 four_byte_option_impl!(four_byte_option_hash256, Hash256);
 
 /// Information returned by `peers` and `connected_peers`.
-// TODO: this should be deserializable..
-#[derive(Debug, Clone, Serialize)]
-#[serde(bound = "E: EthSpec")]
-pub struct Peer<E: EthSpec> {
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Peer {
```
jxs: this type seems to only be used in the `http_api` crate, we can move it there right?

common/eth2/src/lighthouse.rs
```diff
@@ -503,7 +490,7 @@ impl BeaconNodeHttpClient {
     }
 
     /// `GET lighthouse/database/info`
-    pub async fn get_lighthouse_database_info(&self) -> Result<DatabaseInfo, Error> {
+    pub async fn get_lighthouse_database_info(&self) -> Result<serde_json::Value, Error> {
```
jxs: is this function used? I commented it and the tests compiled, so maybe we don't need it

jxs (commented):
Thanks for starting this Daniel. Left a couple comments that may help not use `serde_json::Value`

common/eth2/src/lighthouse.rs
dknopik: I think the `eth2` is intended to be usable by other projects, hence it supporting all the API that exists, regardless of if we use it ourself.

common/eth2/src/lighthouse.rs
dknopik: seems so, so yeah.
but we could also add the missing fns for `peers` and `connected_peers` to the crate

dknopik: To reintroduce `PeerInfo`, we need to either
a) move it to `eth2`, as we no longer depend on `lighthouse_network` (which is the point of this PR)
b) have a variant of it in `eth2` and `From` impls to map to it.
c) have a common crate (e.g. new `network_types`, or even a module in `types`) where the types live 

Option "a" might be controversial, as having these types in `eth2` might be confusing. Option "b" causes much almost same code. Option "c" might be nicest, but it would either create a new crate or make `types` even larger.

dknopik:
@jxs, applied the changes we discussed, feel free to take a look again

jxs (approved):
LGTM Daniel, thanks!

michaelsproul:
Dropping this from v7.0.0 as I imagine it's most useful in unstable, and we can do that later

michaelsproul:
Doing some triage. Would be good to merge this soon. Maybe @macladson could take a quick look as second reviewer? :pray:

realbigsean:
this cuts the number of deps in half, thank you @dknopik  😭

common/eth2/src/lighthouse.rs
```diff
@@ -361,19 +352,6 @@ impl BeaconNodeHttpClient {
         self.get_opt::<(), _>(path).await.map(|opt| opt.is_some())
     }
 
-    /// `GET lighthouse/database/info`
```
macladson, line 383: My feeling is that we shouldn't remove this function even if it's unused in Lighthouse. It feels strange to have an incomplete interface for the `/lighthouse` API. If the test is the issue, then I would argue to remove it only and leave the actual function. Perhaps the main checks in the test can be moved directly into the `store` or `http_api` crates somehow.

I'll also note that if we do decide to remove the function, I believe this would _technically_ count as a breaking change

macladson (commented):
Looks good, just a comment regarding `GET lighthouse/database/info`

common/eth2/src/lighthouse.rs
dknopik: > I'll also note that if we do decide to remove the function, I believe this would `technically` count as a breaking change

Yeah. But with the `eth2` crate being unpublished and not properly versioned at all, we are in "perma-unstable" territory anyway. Also, by that logic, one can consider any change in the underlying internal database types a breaking change.

@jxs and I had a discussion about removing it vs. returning `serde_json::Value` vs. returning some view of the internal type. We ended up at the conclusion of removing it because it has close to no benefit only returning `serde_json::Value`, while moving the underlying types to `eth2` (or providing some types there to map to) is needlessly complex for an API function we guarantee no stability for. That being said, personally I don't feel too strongly either way, so I am open to reintroduce it.

macladson: Yeah these are fair points, perhaps if the day comes where the `eth2` crate is publishable we can revisit these missing functions.

dknopik:
@michaelsproul @macladson 

From my point of view, and if I understood @macladson's last comment in thread above correctly, this is ready to go. Let me know if there are any remaining questions or concerns.

jxs (approved):

Issue #6681, opened 2024-12-11
Subscribing to attestation subnets for aggregating does not work correctly


povi:
## Description

If `[/eth/v1/validator/beacon_committee_subscriptions](https://ethereum.github.io/beacon-APIs/#/Validator/prepareBeaconCommitteeSubnet)` request contains multiple aggregator subscriptions to the same attestation subnet at different slots, it respects only the last subscription per subnet:

E.g. if there request contains these subscriptions:

```
[
  {
    "validator_index": "1",
    "committee_index": "1",
    "committees_at_slot": "2",
    "slot": "2",
    "is_aggregator": true
  },
  {
    "validator_index": "2",
    "committee_index": "2",
    "committees_at_slot": "2",
    "slot": "7",
    "is_aggregator": true
  },
  {
    "validator_index": "3",
    "committee_index": "1",
    "committees_at_slot": "2",
    "slot": "12",
    "is_aggregator": true
  }
]
```

it will only subscribe to attestation subnets for aggregating at slots 7 and 12, ignoring slot 2.

PR #6682, opened 2024-12-11
Fix subscribing to attestation subnets for aggregating (#6681)

povi:
* Prevent scheduled subnet subscriptions from being overwritten by other subscriptions from same subnet with additional scoping by slot

## Issue Addressed

#6681

## Proposed Changes

Additionally scope scheduled subnet subscriptions by slot to prevent them from being overwritten by other same subnet subscriptions with different slots.

### DIFF
--- a/beacon_node/network/src/subnet_service/mod.rs
+++ b/beacon_node/network/src/subnet_service/mod.rs
@@ -86,7 +86,7 @@ pub struct SubnetService<T: BeaconChainTypes> {
     subscriptions: HashSetDelay<Subnet>,
 
     /// Subscriptions that need to be executed in the future.
-    scheduled_subscriptions: HashSetDelay<Subnet>,
+    scheduled_subscriptions: HashSetDelay<ExactSubnet>,
 
     /// A list of permanent subnets that this node is subscribed to.
     // TODO: Shift this to a dynamic bitfield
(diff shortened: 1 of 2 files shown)


beacon_node/network/src/subnet_service/mod.rs
```diff
@@ -485,7 +485,7 @@ impl<T: BeaconChainTypes> SubnetService<T> {
         } else {
             // This is a future slot, schedule subscribing.
             self.scheduled_subscriptions
-                .insert_at(subnet, time_to_subscription_start);
+                .insert_at(ExactSubnet { subnet, slot }, time_to_subscription_start);
```
AgeManning, line 490: Nice! We did notice a potential regression here, I looked back over this code and didn't see it. Thanks for catching this!

Could we add a comment above here, saying that we need to include the slot to make the key unique to prevent overwriting the entry for the same subnet (to prevent future people like me from removing the exact subnet here) :)

AgeManning (commented):
Thank you! We were trying to find this!

michaelsproul:
Thanks for the fix @povi!

I've rebased your commit on the `release-v6.0.1` branch so that we can include it in our upcoming bugfix release.

beacon_node/network/src/subnet_service/mod.rs
povi: Added the comment

michaelsproul (approved):

michaelsproul:
@mergify queue

PR #6688, opened 2024-12-12
docs: remove repetitive words

ChengenH:


michaelsproul:
we don't accept airdrop farming PRs

PR #6695, opened 2024-12-14
Add Fulu boilerplate

macladson:
## Proposed Changes

Adds Fulu fork boilerplate.

## Additional Info

This makes no attempt to unify Fulu with PeerDAS. This will need to be a separate PR.


beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -5322,7 +5322,8 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             BeaconState::Bellatrix(_)
             | BeaconState::Capella(_)
             | BeaconState::Deneb(_)
-            | BeaconState::Electra(_) => {
+            | BeaconState::Electra(_)
+            | BeaconState::Fulu(_) => {
```
dapplion: Can we do `fork.after_bellatrix` here? the argument of using a match to force us to think about the fork is broken here, because we don't know what EIPs will exist in the fork and if we need a new handler. So now we are getting the worst of both worlds. a 1980 lines boilerplate PR and no guarantees or correctness

beacon_node/beacon_chain/src/test_utils.rs
```diff
@@ -918,9 +925,9 @@ where
             | SignedBeaconBlock::Altair(_)
             | SignedBeaconBlock::Bellatrix(_)
             | SignedBeaconBlock::Capella(_) => (signed_block, None),
-            SignedBeaconBlock::Deneb(_) | SignedBeaconBlock::Electra(_) => {
-                (signed_block, block_response.blob_items)
-            }
+            SignedBeaconBlock::Deneb(_)
+            | SignedBeaconBlock::Electra(_)
+            | SignedBeaconBlock::Fulu(_) => (signed_block, block_response.blob_items),
```
dapplion: This can also be `.after_deneb()`

beacon_node/execution_layer/src/engine_api/http.rs
```diff
@@ -882,9 +910,14 @@ impl HttpJsonRpc {
...
                     .map_err(Error::BadResponse)
             }
-            ForkName::Base | ForkName::Altair | ForkName::Deneb | ForkName::Electra => Err(
-                Error::UnsupportedForkVariant(format!("called get_payload_v2 with {}", fork_name)),
-            ),
+            ForkName::Base
+            | ForkName::Altair
+            | ForkName::Deneb
+            | ForkName::Electra
+            | ForkName::Fulu => Err(Error::UnsupportedForkVariant(format!(
```
dapplion: This should be a default case, we will never use v2 again

beacon_node/execution_layer/src/engine_api/http.rs
```diff
@@ -943,13 +977,46 @@ impl HttpJsonRpc {
...
 
+    pub async fn get_payload_v5<E: EthSpec>(
+        &self,
+        fork_name: ForkName,
+        payload_id: PayloadId,
+    ) -> Result<GetPayloadResponse<E>, Error> {
+        let params = json!([JsonPayloadIdRequest::from(payload_id)]);
+
+        match fork_name {
+            ForkName::Fulu => {
```
dapplion, line 980: can we use match(fulu) or after_fulu?

common/eth2_network_config/built_in_network_configs/mainnet/config.yaml
```diff
@@ -1,14 +1,14 @@
 # Mainnet config
 
 # Extends the mainnet preset
-PRESET_BASE: 'mainnet'
+PRESET_BASE: "mainnet"
```
dapplion: We should maintain same format as upstream file https://github.com/ethereum/consensus-specs/blob/dev/configs/mainnet.yaml

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Yeah I tend to agree here, I think matching enum variants are useful if all new variants need to be handled differently by default.

In our case for forks, they're are usually the opposite - they are unchanged unless we explicitly change them.

jimmygchen: I'm quite curious to know how much boilerplate we can reduce with `fork.after_<fork_name>`

common/eth2_network_config/built_in_network_configs/mainnet/config.yaml
macladson: Oof, this looks like the results of a rogue IDE. I've reverted this specific change in the various config files, but left some of the the other changes since they seemed okay.

beacon_node/execution_layer/src/engine_api/http.rs
macladson: Yeah, actually I think since each version of `get_payload` from v3 onwards is restricted to a single fork, we can match on the fork in question, then by default reject with `Error::UnsupportedForkVariant(...)`

We can do this for`v3` and `v4` as well.

dapplion:
@macladson I've applied the `fork_enabled` pattern on the remaining places where it is possible in this commit https://github.com/sigp/lighthouse/commit/a5437652cc1be7288e62eb870869f2fc9560482f please review and consider cherry-picking into this PR

To motivate my case against the `match fork_name`:
- Relying on the type system to direct our dev attention to certain places would make sense ONLY if we add a new variant to ForkName once the fork is fully spec'ed. Otherwise (like here, now) we just fill all those match cases assuming the fork is empty, losing the all benefits of this approach. Let's accept that we are unlikely to wait for a full spec before adding the boilerplate.

macladson:
Thanks for the commit! I agree with your reasoning and I've cherry-picked it into the main branch.

I ended up reverting the `web3signer.rs` changes as it actually uses a local variant of `ForkName` not `types::ForkName` so to remove that boilerplate, we'd either need to write a mapping from `types::ForkName => web3signer::ForkName` or rewrite the `web3signer` to use `types::ForkName`. I can't comment on why `types::ForkName` was chosen in the first place, but either way, it feels out of scope for this PR.

dapplion (approved):
Looks great! Send it

beacon_node/beacon_chain/src/beacon_block_streamer.rs
```diff
@@ -749,6 +750,7 @@ mod tests {
         let capella_fork_epoch = 4usize;
         let deneb_fork_epoch = 6usize;
         let electra_fork_epoch = 8usize;
+        let fulu_fork_epoch = 10usize;
         let num_blocks_produced = num_epochs * slots_per_epoch;
```
jimmygchen, line 754: `num_epochs` needs to be bumped for the test to transition to Fulu. 
nit: update test name from `to_electra` to `_to_fulu`.

beacon_node/beacon_chain/src/fulu_readiness.rs
```diff
@@ -0,0 +1,114 @@
...
+pub const FULU_READINESS_PREPARATION_SECONDS: u64 = SECONDS_IN_A_WEEK * 2;
+pub const ENGINE_CAPABILITIES_REFRESH_INTERVAL: u64 = 300;
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+#[serde(tag = "type")]
+pub enum FuluReadiness {
+    /// The execution engine is fulu-enabled (as far as we can tell)
+    Ready,
+    /// We are connected to an execution engine which doesn't support the V4 engine api methods
```
jimmygchen: `V5` instead of `V4`

lcli/src/mock_el.rs
```diff
@@ -19,6 +19,7 @@ pub fn run<E: EthSpec>(mut env: Environment<E>, matches: &ArgMatches) -> Result<
     let shanghai_time = parse_required(matches, "shanghai-time")?;
     let cancun_time = parse_optional(matches, "cancun-time")?;
     let prague_time = parse_optional(matches, "prague-time")?;
+    let osaka_time = parse_optional(matches, "fulu-time")?;
```
jimmygchen: `osaka-time`

beacon_node/beacon_chain/src/test_utils.rs
```diff
@@ -2863,6 +2864,25 @@ pub fn generate_rand_block_and_blobs<E: EthSpec>(
             message.body.blob_kzg_commitments = bundle.commitments.clone();
             bundle
         }
+        SignedBeaconBlock::Fulu(SignedBeaconBlockFulu {
+            ref mut message, ..
+        }) => {
+            // Get either zero blobs or a random number of blobs between 1 and Max Blobs.
+            let payload: &mut FullPayloadFulu<E> = &mut message.body.execution_payload;
+            let num_blobs = match num_blobs {
+                NumBlobs::Random => rng.gen_range(1..=E::max_blobs_per_block()),
```
jimmygchen, line 2873: no action here for now, just a note to update the usage here to `MAX_BLOBS_PER_BLOCK_ELECTRA` and `MAX_BLOBS_PER_BLOCK_FULU`

beacon_node/execution_layer/src/engine_api/http.rs
```diff
@@ -1214,6 +1264,13 @@ impl HttpJsonRpc {
                     Err(Error::RequiredMethodUnsupported("engine_newPayloadV4"))
                 }
             }
+            NewPayloadRequest::Fulu(new_payload_request_fulu) => {
+                if engine_capabilities.new_payload_v5 {
+                    self.new_payload_v5_fulu(new_payload_request_fulu).await
+                } else {
+                    Err(Error::RequiredMethodUnsupported("engine_newPayloadV4"))
```
jimmygchen: `engine_newPayloadV5`

beacon_node/execution_layer/src/test_utils/mock_builder.rs
```diff
@@ -592,17 +620,19 @@ pub fn serve<E: EthSpec>(
...
-                                    .as_electra()
-                                    .map_err(|_| reject("incorrect payload variant"))?
-                                    .into(),
-                                blob_kzg_commitments: maybe_blobs_bundle
-                                    .map(|b| b.commitments)
-                                    .unwrap_or_default(),
-                                value: Uint256::from(DEFAULT_BUILDER_PAYLOAD_VALUE_WEI),
-                                pubkey: builder.builder_sk.public_key().compress(),
-                            }),
+                            ForkName::Electra | ForkName::Fulu => {
```
jimmygchen: Should this be mapped to `BuilderBidFulu`?

beacon_node/execution_layer/src/test_utils/mock_builder.rs
```diff
@@ -644,17 +674,19 @@ pub fn serve<E: EthSpec>(
...
```
jimmygchen: Same here, `BuilderBidFulu`

beacon_node/lighthouse_network/src/types/pubsub.rs
```diff
@@ -242,6 +242,10 @@ impl<E: EthSpec> PubsubMessage<E> {
                                     SignedBeaconBlockElectra::from_ssz_bytes(data)
                                         .map_err(|e| format!("{:?}", e))?,
                                 ),
+                                Some(ForkName::Fulu) => SignedBeaconBlock::<E>::Electra(
+                                    SignedBeaconBlockElectra::from_ssz_bytes(data)
+                                        .map_err(|e| format!("{:?}", e))?,
+                                ),
```
jimmygchen, line 249: This should be `SignedBeaconBlockFulu`

consensus/types/src/config_and_preset.rs
```diff
@@ -50,7 +53,26 @@ impl ConfigAndPreset {
         let capella_preset = CapellaPreset::from_chain_spec::<E>(spec);
         let extra_fields = get_extra_fields(spec);
 
-        if spec.electra_fork_epoch.is_some()
+        if spec.fulu_fork_epoch.is_some()
+            || fork_name.is_none()
+            || fork_name == Some(ForkName::Fulu)
+        {
+            let deneb_preset = DenebPreset::from_chain_spec::<E>(spec);
```
jimmygchen: I think we can move `deneb_preset` to the top and drop support for returning `ConfigAndPresetCapella`?

consensus/types/src/beacon_state.rs
```diff
@@ -2277,13 +2336,27 @@ impl<E: EthSpec> BeaconState<E> {
...
+            BeaconState::Fulu(_) => {
+                let state = self.as_fulu_mut()?;
+
+                // Consume the balance and update state variables.
+                state.consolidation_balance_to_consume =
+                    consolidation_balance_to_consume.safe_sub(consolidation_balance)?;
+                state.earliest_consolidation_epoch = earliest_consolidation_epoch;
+                Ok(state.earliest_consolidation_epoch)
+            }
+            _ => unreachable!(),
```
jimmygchen: This will panic in the future if we forget to add the G fork here, and compiler wouldn't be able to catch. May be worth returning an `Error`?

testing/ef_tests/src/cases/operations.rs
```diff
@@ -308,6 +300,7 @@ impl<E: EthSpec> Operation<E> for BeaconBlockBody<E, FullPayload<E>> {
                 ForkName::Capella => BeaconBlockBody::Capella(<_>::from_ssz_bytes(bytes)?),
                 ForkName::Deneb => BeaconBlockBody::Deneb(<_>::from_ssz_bytes(bytes)?),
                 ForkName::Electra => BeaconBlockBody::Electra(<_>::from_ssz_bytes(bytes)?),
+                ForkName::Fulu => BeaconBlockBody::Electra(<_>::from_ssz_bytes(bytes)?),
```
jimmygchen: `BeaconBlockBody::Fulu`

testing/ef_tests/src/cases/operations.rs
```diff
@@ -363,6 +356,10 @@ impl<E: EthSpec> Operation<E> for BeaconBlockBody<E, BlindedPayload<E>> {
                     let inner = <BeaconBlockBodyElectra<E, FullPayload<E>>>::from_ssz_bytes(bytes)?;
                     BeaconBlockBody::Electra(inner.clone_as_blinded())
                 }
+                ForkName::Fulu => {
+                    let inner = <BeaconBlockBodyElectra<E, FullPayload<E>>>::from_ssz_bytes(bytes)?;
+                    BeaconBlockBody::Electra(inner.clone_as_blinded())
```
jimmygchen, line 361: This should be `BeaconBlockBodyFulu` right?

consensus/types/src/beacon_block.rs
```diff
@@ -1122,5 +1255,28 @@ mod tests {
...
+            };
+
+            assert_eq!(
+                BeaconBlock::from_ssz_bytes(&good_block.as_ssz_bytes(), &spec)
+                    .expect("good fulu block can be decoded"),
+                good_block
+            );
+            // TODO(fulu)
+            //BeaconBlock::from_ssz_bytes(&bad_block.as_ssz_bytes(), &spec)
+            //    .expect_err("bad fulu block cannot be decoded");
```
jimmygchen, line 1280: Why is this commented out? is this failing

jimmygchen (changes requested):
Nice work @macladson and thanks for the review @dapplion! I've added a few more comments, let me know what you think. I think we can merge this soon 🎉

consensus/types/src/beacon_state.rs
macladson: Ah good catch, I had intended to remove this

consensus/types/src/beacon_block.rs
macladson: Ah yeah, I forgot to add a comment explaining this. Since the method we use to invalidate the blocks is by setting the slot to an epoch before the fork epoch, it actually becomes a valid Electra block (since Fulu blocks don't yet contain any new fields) and thus doesn't error on the decode.

I figured the easiest way was just wait to uncomment this out until Fulu has actual features. Otherwise we could rewrite the way we test invalid block decoding.

jimmygchen: Oh, I incorrectly assumed the slot would invalidate the decoding, got it now - thanks!
Yeah i think it's fine to leave it as it is with a comment 👍

macladson: I've added a comment now, let me know if you think this is unreasonable :+1:

consensus/types/src/beacon_state.rs
macladson: I have also now added an exhaustive match statement rather than using a default. This seems like the kind of use case we want the compiler to direct us to.

consensus/types/src/beacon_block.rs
jimmygchen: Nice, thanks

jimmygchen (approved):
Looks great, thanks @macladson!

michaelsproul:
@mergify queue

michaelsproul:
@mergify dequeue

michaelsproul:
@mergify requeue

michaelsproul:
@mergify dequeue

michaelsproul:
@macladson this needs some updates for the clippy lints https://github.com/sigp/lighthouse/actions/runs/12702656555/job/35409240281?pr=6785

michaelsproul:
@mergify queue

PR #6700, opened 2024-12-16
Back merge v6.0.1 to unstable

michaelsproul:
## Proposed Changes

Merge the changes from v6.0.1 back into `unstable`. Notably this will fix Docker builds on `unstable`.

This PR should not be merged using Mergify. We just need an approval on it, and then it can be merged without squashing.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -13,8 +13,8 @@ concurrency:
   cancel-in-progress: true
 
 env:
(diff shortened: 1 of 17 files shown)


michaelsproul:
Ok so this won't actually fix Docker builds, because they're broken on `stable`. But I would still like to merge this and then fix Docker builds on `unstable` and backport the fix to a new release v6.0.2

jimmygchen (approved):

PR #6339, opened 2024-09-02
Integrate tracing

ThreeHrSleep:
## Issue Addressed
Tracing Integration
- [reference](https://github.com/eth-protocol-fellows/cohort-five/blob/5bbf1859e921065bd69f8671038ed16643465b86/projects/project-ideas.md?plain=1#L297)
## Proposed Changes
- [x] replace slog & log with tracing throughout the codebase 
- [x] implement custom crit log 
- [x] make relevant changes in the formatter
- [x] replace sloggers 
- [x] re-write SSE logging components

cc: @macladson @eserilev

PR #6705, opened 2024-12-16
Modularize validator store

dknopik:
This is a PR that modularises the validator store so that we can reuse the validator services in Anchor.

## Proposed Changes

- Create trait `ValidatorStore` with all functions used by the `validator_services`
- Make `validator_services` generic on `S: ValidatorStore`
- Introduce `LighthouseValidatorStore`, which has identical functionality to the old `ValidatorStore`
- Remove dependencies (especially `environment`) from `validator_services` and `beacon_node_fallback` in order to be able to cleanly use them in Anchor

## Additional Info

This also includes some changes from @jxs's closed PR #6727 to remove `E` from some `struct`s and make functions generic instead (where possible).


dknopik:
Will rebase this to clean up the messy history after 
- #6339 and
- #6727

are merged. Still, feel free to start reviewing already.

validator_client/http_api/Cargo.toml
```diff
@@ -32,19 +33,20 @@ serde = { workspace = true }
...
-tempfile =  { workspace = true }
-tokio =  { workspace = true }
-tokio-stream =  { workspace = true }
-tracing =  { workspace = true }
-types =  { workspace = true }
-url =  { workspace = true }
-validator_dir =  { workspace = true }
-validator_services =  { workspace = true }
-validator_store =  { workspace = true }
+
```
michaelsproul:
```suggestion
```

validator_client/http_api/src/tests.rs
```diff
@@ -1159,7 +1158,7 @@ async fn validator_derived_builder_boost_factor_with_process_defaults() {
         })
         .await
         .assert_default_builder_boost_factor(Some(80))
-        .assert_validator_derived_builder_boost_factor(0, None)
+        .assert_validator_derived_builder_boost_factor(0, Some(80))
```
michaelsproul, line 1161: Why did this change?

validator_client/signing_method/src/lib.rs
```diff
@@ -227,11 +232,7 @@ impl SigningMethod {
 
                 // Determine the Web3Signer message type.
                 let message_type = object.message_type();
-
-                if matches!(
-                    object,
-                    Web3SignerObject::Deposit { .. } | Web3SignerObject::ValidatorRegistration(_)
-                ) && fork_info.is_some()
+                if matches!(message_type, MessageType::ValidatorRegistration) && fork_info.is_some()
```
michaelsproul, line 235: Is this a bugfix?

validator_client/validator_services/src/block_service.rs
```diff
@@ -563,15 +579,9 @@ impl<T: SlotClock + 'static, E: EthSpec> BlockService<T, E> {
...
-
-        // Fallback to process-wide configuration if needed.
-        let maybe_builder_boost_factor = validator_builder_boost_factor.or_else(|| {
-            self.validator_store
-                .determine_default_builder_boost_factor()
-        });
+            .determine_builder_boost_factor(validator_pubkey);
 
-        if let Some(builder_boost_factor) = maybe_builder_boost_factor {
+        if let Some(builder_boost_factor) = validator_builder_boost_factor {
```
michaelsproul: This looks like it removes the process-wide builder boost factor?

Is this why the test behaviour changed? I feel like I'm missing something here.

michaelsproul (commented):
This is fantastic, very clean!

I'm highly motivated to get this merged soon, not just because of Anchor, but also because I want to solve this FIXME building on your branch:

https://github.com/sigp/lighthouse/blob/e42406d7b79a85ad4622f3a7440ff6468ac4c9e1/validator_client/validator_store/src/lib.rs#L209

It will require changing a few of the `ValidatorStore` methods to return `impl Future`, but I think we should merge this first, and then adapt Anchor to that change once it's also ready and merged.

validator_client/signing_method/src/lib.rs
dknopik: Good catch. Checking the `git blame`, this seems to originate from @jxs's https://github.com/sigp/lighthouse/pull/6727. @jxs, do you remember this? There is a non-zero chance that I messed this up when rebasing, though I do not remember ever touching this.

validator_client/validator_services/src/block_service.rs
dknopik: I took the opportunity to simplify the interface a bit here. Instead of letting the caller here fall back to a default that is fetched via `determine_default_builder_boost_factor`, it is now the responsibility of `determine_builder_boost_factor` to either pick a validator specific factor or fall back to a default. This is in line with the behaviour of e.g. `get_fee_recipient`.

No strong feelings for this change, so we can change it back if you prefer.

dknopik:
There is one more change that would be nice: currently, `sign_validator_exit` is part of the new `ValidatorStore` trait, as I wanted to keep the interface (mostly) the same. However, in the meantime I realized that `sign_validator_exit` is not actually used in the `validator_services` crate, but in the VC HTTP API, where the actual type "`LighthouseValidatorStore`" is available. So we could move it back to be a method of that type. This would make the Anchor code a bit nicer as the function signature is not sufficient for our needs, so we only return an error in there anyway (as it will never be called by the `validator_services`) and have a custom function for Anchor internal use.

Does that sound fine by you?

michaelsproul:
> Does that sound fine by you?

Yeah that sounds totally fine, happy for it to move to `LighthouseValidatorStore`!

validator_client/validator_services/src/block_service.rs
michaelsproul: The default is completely gone from `determine_builder_boost_factor` though, so this seems like a regression. I'll just put the logic back here if it's not needed for the `ValidatorStore` abstraction.

michaelsproul: Ah, I see the issue with putting it here. I'll try to consolidate the default logic into `LighthouseValidatorStore` so that Anchor doesn't need to worry about any of this. I'm guessing Anchor doesn't care about LH's config for this at all?

michaelsproul: Sorry, my bad. There _is_ a default in `determine_builder_boost_factor`. This code is just insane because of the combinatoric explosion of settings.

michaelsproul: Tried to simplify things a little by removing `BlockService::get_builder_boost_factor`. I think that should be OK for anchor too?

https://github.com/sigp/lighthouse/pull/6705/commits/508bbed3a02b7d1d28abc6157479da911898afbc

validator_client/http_api/src/tests.rs
michaelsproul: I understand this now. It's because we added the process-level defaults into `determine_builder_boost_factor`. Previously this test asserted that the boost factor for the key was `None` despite the process-level default, but now it is 80 (the process-level default), which is fine.

validator_client/validator_services/src/block_service.rs
dknopik: Looks good, thanks! Should be fine for Anchor, but I'll test it along with the change mentioned below.

michaelsproul (approved):
Let's goooo

PR #6711, opened 2024-12-17
Unsaturate dial negotiation queue

### LINKED ISSUE #6703: Stuck lookups involving block requests that never time out
## Description

We are investigating a pattern of stuck block lookups characterised by the log:

```
WARN Notify the devs a sync lookup is stuck ...
```

The root cause seems to be that we never receive a response for some BlocksByRoot requests, not even a timeout. This has been observed with Caplin peers.

Often this type of stuck lookup does not pose any risk to the node's sync status, as the block is imported via HTTP or gossip in the meantime. This may be incidental (and lucky), or related to the root cause in some way that we have not yet determined. It might be prudent to _remove lookups which are obsoleted by blocks imported via HTTP, as we do for gossip_.

## Version

Lighthouse v6.0.1

AgeManning:
## Issue Addressed

It seems to me there is a bug, where if we fail dials via io errors, we attempt a re-connection for a number of times before failing the RPC request. However, in the re-attempts we keep adding to the `dial_negotiation` count. 

I suspect, after a number of failures, the `dial_negotiation` count can saturate and become larger than our max of 8 concurrent dials. In which case, we will no longer perform any further dials. Our dial queue will just grow and grow and nothing will happen because the RPC thinks we have already too many dials. 

We've witnessed logs of this form. I haven't tested this, but I think my logic here is sound.

I think resolves #6703

### DIFF
--- a/beacon_node/lighthouse_network/src/rpc/handler.rs
+++ b/beacon_node/lighthouse_network/src/rpc/handler.rs
@@ -964,6 +964,9 @@ where
         request_info: (Id, RequestType<E>),
         error: StreamUpgradeError<RPCError>,
     ) {
+        // This dialing is now considered failed
+        self.dial_negotiated -= 1;
+
         let (id, req) = request_info;
 
         // map the error
@@ -988,9 +991,6 @@ where
             }
             StreamUpgradeError::Apply(other) => other,
         };
-
-        // This dialing is now considered failed
-        self.dial_negotiated -= 1;
 
         self.outbound_io_error_retries = 0;
         self.events_out


AgeManning:
For reviewers. 

The condition for us to send a message is here:
https://github.com/sigp/lighthouse/blob/stable/beacon_node/lighthouse_network/src/rpc/handler.rs#L781
```rust
if !self.dial_queue.is_empty() && self.dial_negotiated < self.max_dial_negotiated {
self.dial_negotiated += 1;
```
We add to this number. But notice this number must be less than self.max_dial_negotiated, which is some constant like 8. 

Then when this dial fails:
https://github.com/sigp/lighthouse/blob/stable/beacon_node/lighthouse_network/src/rpc/handler.rs#L962
```rust
        let (id, req) = request_info;

        // map the error
        let error = match error {
            StreamUpgradeError::Timeout => RPCError::NegotiationTimeout,
            StreamUpgradeError::Apply(RPCError::IoError(e)) => {
                self.outbound_io_error_retries += 1;
                if self.outbound_io_error_retries < IO_ERROR_RETRIES {
                    self.send_request(id, req);
                    return;
                }
                RPCError::IoError(e)
            }
            StreamUpgradeError::NegotiationFailed => RPCError::UnsupportedProtocol,
            StreamUpgradeError::Io(io_err) => {
                self.outbound_io_error_retries += 1;
                if self.outbound_io_error_retries < IO_ERROR_RETRIES {
                    self.send_request(id, req);
                    return;
                }
                RPCError::IoError(io_err.to_string())
            }
            StreamUpgradeError::Apply(other) => other,
        };

        // This dialing is now considered failed
        self.dial_negotiated -= 1;

        self.outbound_io_error_retries = 0;
        self.events_out
            .push(HandlerEvent::Err(HandlerErr::Outbound {
                error,
                proto: req.versioned_protocol().protocol(),
                id,
            }));
    }
}
```
We can sometimes not decrement this constant, because we re-try. I think this means we can artificially increase this number, over our limit and get the dial_queue stuck.

jimmygchen (approved):
Great find!

dapplion (approved):
Epic find

jimmygchen:
@mergify queue

jxs (approved):
LGTM age!

jimmygchen:
@mergify requeue

PR #6726, opened 2024-12-18
Add tests for EIP 7251.

bshastry:
## Issue Addressed

This PR addresses the testing coverage for EIP-7251's partial withdrawal functionality in Lighthouse, specifically the `is_partially_withdrawable_validator_electra` implementation. No existing issue number, but this adds test coverage for the Electra fork implementation.

## Proposed Changes

1. Adds test `test_partial_withdrawals_electra` in `consensus/types/src/validator.rs` that covers four scenarios:
   - Compounding validator (0x02 prefix) with 2048 ETH effective balance and excess balance
   - Non-compounding validator (0x01 prefix) with 2048 ETH effective balance
   - Compounding validator with 32 ETH effective balance
   - Non-compounding validator with 32 ETH effective balance

2. Test verifies that automatic partial withdrawals are only allowed when:
   - Validator has the correct effective balance for both Eth1 and compounding modes
   - Validator has excess balance above effective balance

## Additional Info

1. **Test Coverage**: This test specifically covers the automatic sweep mechanism for excess balances, which is distinct from user-initiated partial withdrawals.

2. **Future Considerations**:
   - May want to add additional test cases for edge cases around the effective balance values
   - Could extend to test interaction with different fork transitions
   - Consider adding integration tests with actual withdrawal processing

3. **Related Specs**:
   - [EIP-7251](https://eips.ethereum.org/EIPS/eip-7251)
   - [Electra Consensus Spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#modified-is_partially_withdrawable_validator)

4. **Review Notes**:
   - Test uses both 0x01 and 0x02 withdrawal credential prefixes
   - Balance values are chosen to test excess balance scenarios
   - All assertions include descriptive messages for failure reporting


dapplion:
@bshastry thanks for the contribution! Sorry for leaving the PR un-merged for that long. Since Electra fork is live and has spec tests, could you migrate this cases to spec tests? That way all clients can benefit from the extra coverage

PR #6727, opened 2024-12-18
Remove E from validator_client types where it's not required

jxs:
and the `Ethspec` parameters can be passed on function all time


validator_client/signing_method/src/lib.rs
```diff
@@ -227,11 +232,7 @@ impl SigningMethod {
 
                 // Determine the Web3Signer message type.
                 let message_type = object.message_type();
-
-                if matches!(
-                    object,
-                    Web3SignerObject::Deposit { .. } | Web3SignerObject::ValidatorRegistration(_)
-                ) && fork_info.is_some()
+                if matches!(message_type, MessageType::ValidatorRegistration) && fork_info.is_some()
```
jxs, line 235: `object` can never be a `Web3SignerObject::Deposit` as a `SignableMessage` doesn't convert to it, see line `202` above

validator_client/validator_store/Cargo.toml
```diff
@@ -10,7 +10,10 @@ path = "src/lib.rs"
 
 [dependencies]
 account_utils = { workspace = true }
+beacon_node_fallback = { workspace = true }
```
pawanjay176: Not really sure why we had to add dependencies here

validator_client/signing_method/src/lib.rs
pawanjay176: Feels kinda hard to keep track if `SignableMessage` starts converting to a `Web3SignerObject::Deposit` in the future. 
I don't see the issue with keeping it as is, but what you said makes sense as well.
Although, i don't have enough context on the web3signer to know for sure

pawanjay176 (approved):
LGTM. 
I'd recommend running this version of the VC on testnets before merging to make sure there isn't something that breaks unexpectedly

validator_client/validator_store/Cargo.toml
jxs: yup you are right, it was because I had remove `doppelganger_service` as a separate crate but I have since removed again, thanks Pawan!

AgeManning:
Whats the state of this guy? We still doing this?

dknopik:
> Whats the state of this guy? We still doing this?

Came here to ask the same thing :D

I implemented #6705 with this PR in mind. Unfortunately, there I introduce a new generic parameter to the duties service: `S: ValidatorStore` (which will have one struct impling it for each LH and Anchor), and as we need `E` in Anchors `ValidatorStore` impl, we need to reintroduce it in a few places in Lighthouse as well. That being said, some benefits introduced in this PR remain, so I propose to go ahead with this. I encourage anyone to check out PR #6705 and the diff between the two PRs, maybe we can find an even better way :)

jxs:
> > Whats the state of this guy? We still doing this?
> 
> Came here to ask the same thing :D
> 
> I implemented #6705 with this PR in mind. Unfortunately, there I introduce a new generic parameter to the duties service: `S: ValidatorStore` (which will have one struct impling it for each LH and Anchor), and as we need `E` in Anchors `ValidatorStore` impl, we need to reintroduce it in a few places in Lighthouse as well. That being said, some benefits introduced in this PR remain, so I propose to go ahead with this. I encourage anyone to check out PR #6705 and the diff between the two PRs, maybe we can find an even better way :)

thanks for the ping. I think we should merge @dknopik's PR instead as it contains these changes. So going to close this one

PR #6735, opened 2024-12-19
Refactor mock builder

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Adds electra support to the builder flow.
Refactor the mock builder to extract all builder functions into separate functions.
Also adds a `prepare_execution_layer` method that prepares the execution layer for payload creation for each slot.

The motivation is to provide a backend for a simple builder service that follows the builder api and produces execution payloads from the local mempool of the connected EL. This allows us to have a faster turnaround for testing changes to the builder api in kurtosis testnets and new fork devnets. 
See https://github.com/pawanjay176/rustic-builder/

## Additional info

The default parameters to the `MockBuilder::new_for_testing` produces a MockBuilder with same functionality as before, so the tests in lighthouse should work as before


realbigsean (dismissed):
Excellent work

So you know I have axum implemented for this api here, we can add this later though

https://github.com/sigp/ethereum_apis/blob/main/builder-server/src/server.rs

edit: nevermind you're already using it 😂

pawanjay176:
The bug Sean was referring to is that  in scenarios where validators are signalling a gas limit increase/decrease with the validator regisration message, the mock builder is not equipped to increase/decrease the gas limit of the payloads it builds. 
The gas limit is set in the connected EL at startup and runtime changes to it are not possible.

After spending some time trying to allow for this feature, I have come to the conclusion that its very hard to do and probably not worth the effort. The mock builder is meant to work as a backend in controlled testing scenarios of the builder flow. Its pretty easy to configure kurtosis to set the gas limit for the builder EL to be the same as all the CLs in the test environment so its easier to fix it in the testing config.

realbigsean (approved):

realbigsean:
@mergifyio queue

Issue #6709, opened 2024-12-17
Decouple PeerDAS network subnets from das core parameters


jimmygchen:
## Description

See consensus spec PR and discussions here: https://github.com/ethereum/consensus-specs/pull/3832

This change allow us to configure and tweak network parameters in isolation with das core parameters, e.g. changing number of column subnets without impacting custody requirement.

This has been merged and will likely be included in the next devnet.

jimmygchen:
Completed in #6736

PR #6736, opened 2024-12-20
Implement PeerDAS subnet decoupling (aka custody groups)

jimmygchen:
## Issue Addressed

#6709 

This spec change changes the custody unit from _subnets_ to _custody groups_, in order to decouple networking and DAS core parameters - this allow us to tweak DAS core and network parameters in isolation, i.e. changing number of column subnets will not have a direct impact to sampling parameters & security.

Spec: https://github.com/ethereum/consensus-specs/pull/3832

## Proposed Changes:

On startup, we construct `NetworkGlobals` and compute sampling subnets and sampling columns from `NodeId` - this remains the same but the computation logic has been updated to compute these from custody groups. (`node_id` => `custody_groups` => `subnets` and `columns`)

- [x] Introduce "custody groups" and compute custody subnets and columns with custody group.
- [x] Change `custody_subnet_count` (`csc`) to `custody_group_count` (`cgc`) in `MetaData` and `ENR`. ([spec](https://github.com/ethereum/consensus-specs/pull/4062))
- [x] Add a custody group to subnet mapping in PeerManager to avoid re-computing it when setting peer subnets based on their custody groups.
- [x] The `sampling_size` computation logic is has been updated to be base on custody groups instead of custody subnets. [code](https://github.com/sigp/lighthouse/blob/e094a243626af9cfec5768211407d7abb0adebfb/consensus/types/src/chain_spec.rs#L609-L619)  ([spec update](https://github.com/ethereum/consensus-specs/pull/4060))
- [x] Refactored `sampling_size` and `custody_group_count` logic into `ChainSpec` to keep them in one place.


jimmygchen:
@dapplion @pawanjay176 I think this PR is ready for review now. The failing spec tests are new electra tests, which Pawan and Michael are currently working on, and the DAS spec tests are all passing.

Please review and let me know what you think 🙏

consensus/types/src/chain_spec.rs
```diff
@@ -605,10 +606,33 @@ impl ChainSpec {
...
+            .safe_div(self.number_of_custody_groups)
+            .expect("Custody group count must be greater than 0")
+    }
+
+    /// Returns the number of column sidecars to sample per slot.
+    pub fn sampling_size(&self, custody_group_count: u64) -> u64 {
+        let columns_per_custody_group = self
+            .number_of_columns
+            .safe_div(self.number_of_custody_groups)
+            .expect("Custody group count must be greater than 0");
```
dapplion:
```suggestion
            .expect("number_of_custody_groups must be greater than 0");
```

dapplion: A user may run LH with a config that has `number_of_custody_groups: 0`, then LH will panic here. Should there be a sanity check somewhere?

consensus/types/src/chain_spec.rs
```diff
@@ -605,10 +606,33 @@ impl ChainSpec {
...
+
+        let custody_column_count = columns_per_custody_group
+            .safe_mul(custody_group_count)
+            .expect("Computing sampling size should not overflow");
```
dapplion: Are we sure `custody_group_count` is always checked when read from external sources? Add a comment here saying:
- ENR getter checks it's less than X
- On ReqResp Metadata messages?

EDIT: Metadata value is checked in https://github.com/sigp/lighthouse/pull/6736/files#diff-bb7fd4c05488f03a20af7fd1d00ccde4438d074889ce1fc20d8e919845837001R1433-R1434

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -555,7 +555,7 @@ impl<T: BeaconChainTypes> DataAvailabilityCheckerInner<T> {
         if custody_column_count != total_column_count {
             return ReconstructColumnsDecision::No("not required for full node");
         }
-        if received_column_count == self.spec.number_of_columns {
+        if received_column_count == total_column_count {
```
dapplion: Why not `>=`?

dapplion:
A few comments, else looks solid and good to go

beacon_node/lighthouse_network/src/discovery/enr.rs
```diff
@@ -67,16 +67,16 @@ impl Eth2Enr for Enr {
...
-            .map_err(|_| "Could not decode the ENR custody subnet count")?;
+    fn custody_group_count<E: EthSpec>(&self, spec: &ChainSpec) -> Result<u64, &'static str> {
+        let cgc = self
+            .get_decodable::<u64>(PEERDAS_CUSTODY_GROUP_COUNT_ENR_KEY)
+            .ok_or("ENR custody group count non-existent")?
+            .map_err(|_| "Could not decode the ENR custody group count")?;
 
-        if csc >= spec.custody_requirement && csc <= spec.data_column_sidecar_subnet_count {
-            Ok(csc)
+        if cgc >= spec.custody_requirement && cgc <= spec.number_of_custody_groups {
```
dapplion: Can we make this line equal to the one in `compute_peer_custody_groups`? It would be easier to follow.

consensus/types/src/chain_spec.rs
jimmygchen: Yeah this is intentional - If they're running a custom testnet and overrides this value to an invalid value (`0`), I think panicing is fine?

consensus/types/src/chain_spec.rs
jimmygchen: Yeah, so this method is only used for determining the sampling count for the running node, so the assumption here is these spec values MUST be valid, otherwise we panic. We currently don't need to determine the number of samples for other node, and i dont foresee that we'd need this.

If we're paranoid about this, we could make this return `Result` and panic on the caller site, to prevent potential future usage of this method unintentionally causing panic?

jimmygchen: btw re error handling in this PR, I have intentionally:
- handle errors when computing using a peer `cgc`, computing peer custody _should_  never panic.
- panic if we're computing custody for our own node (it means invalid `spec` config)

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: I think it's dangerous to allow for unexpected columns, because it means we either have a duplicate or we have a column that we don't expect - which may unintentionally satisfy the custody column count. I think we need to make sure it's an _exact_ match, given that we're only checking the `count`.

The code here validates for duplicate:
https://github.com/sigp/lighthouse/blob/e094a243626af9cfec5768211407d7abb0adebfb/beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs#L176-L178

Gossip validation should ensure we only get the columns we subscribe to. 

Do you see any valid scenario where this would be`>`?

beacon_node/lighthouse_network/src/discovery/enr.rs
jimmygchen: Good eyes! I've change this to the same format: https://github.com/sigp/lighthouse/pull/6736/commits/73cf91a1eb95fcb4b2639ba4b806e74cdc539a63, without the `!`. 

I prefer to avoid negation here for readability, whereas in `compute_peer_custody_groups` it's a bit easier to follow if we return invalid input early - this is my personal preference though, happy to change

consensus/types/src/chain_spec.rs
dapplion: I mean if we should panic earlier than here, like in the beacon node init sequence call a check_spec method. Also the code suggestion I made is to tell the user which config field is wrong. "Custody group count" is not obvious which field related to

consensus/types/src/chain_spec.rs
dapplion: You should add a comment on each expect justifying its safety, which the reasoning from here

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
dapplion: > I think it's dangerous to allow for unexpected columns

How does your PR disallow that? If somehow we actually have `received_column_count > total_column_count` your PR will trigger reconstruction, which makes no sense

dapplion: Consider all the possible ranges of values `received_column_count` could take. (`total_column_count` = `total`). In your PR, each range results in:

- `[0..total/2]`: not enough columns
- `[total/2..total]`: trigger reconstruction
- `[total..=total]`: all columns received
- `[total + 1..]`: trigger reconstruction

What I propose is to:

- `[0..total/2]`: not enough columns
- `[total/2..total]`: trigger reconstruction
- `[total..=total]`: all columns received
- `[total + 1..]`: all columns received

I think the later makes more sense right?

Another valid implementation would be

- `[0..total/2]`: not enough columns
- `[total/2..total]`: trigger reconstruction
- `[total..=total]`: all columns received
- `[total + 1..]`: error

which is better than triggering reconstruction if you get too many columns

consensus/types/src/chain_spec.rs
jimmygchen: Yep I've already applied the suggestion in https://github.com/sigp/lighthouse/pull/6736/commits/b3ecd95326166721cba9ce4ed1b8342d804361f5

jimmygchen: > I mean if we should panic earlier than here, like in the beacon node init sequence call a check_spec method

We could, but I don't think it makes much difference because this is called when initialising `DataAvailabilityChecker` and `NetworkGlobals` (on node startup), and I think it's reasonable to panic from there if the constructor params are invalid?

I think having validation on where we use the configuration makes sense - if we have a dedicated validator it could get out of date as it's likely we validate in the component consuming the configuration. No strong opinion here though.

consensus/types/src/chain_spec.rs
jimmygchen: Thanks, I've added some more comments, and I've updated this function to return a `Result` instead of panicing and leave the error handing from the caller site - I think longer term this is safer, as it's much clearer if to panic from the caller site and easier to justify - e.g. if chainspec is invalid on startup, we panic
https://github.com/sigp/lighthouse/pull/6736/commits/acbc155806d4d748e1c15871c0b047445bd1d286

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: Oh sorry you're right - I misread where this comment is for!  (I thought it was for `available` check, my bad)

We should either do `>=` or error here, in both cases it will be `ReconstructColumnsDecision::No`. I think the former is simpler so I'll go with that. This is probably not the best place to handle unexpected columns anyways.

jimmygchen: Fixed in 4a02b313f85511797dffb3627603498cbe9c8b44

jimmygchen:
@dapplion thanks for the thorough review! I've addressed all comments and resolved conflicts.

beacon_node/lighthouse_network/src/peer_manager/mod.rs
```diff
@@ -709,22 +730,39 @@ impl<E: EthSpec> PeerManager<E> {
...
+                                    self.subnets_by_custody_group
+                                        .get(&custody_index)
+                                        .cloned()
+                                        .unwrap_or_else(|| {
+                                            warn!(
+                                                self.log,
+                                                "Custody group not found in subnet mapping";
+                                                "custody_index" => custody_index,
+                                                "peer_id" => %peer_id
+                                            );
```
dapplion, line 756: This should never happen as we compute `subnets_by_custody_group` ourselves. Is it necessary to have a warn here?

dapplion (approved):
Looks good! We can review the error conditions later

beacon_node/lighthouse_network/src/peer_manager/mod.rs
jimmygchen: Yes agree that this _should_ not happen - we also validate peer's `custody_group_count` when computing custody groups.

What error handling are you suggesting? I don't think we should panic, or just swallow the error _if_ it happens in case of a bug or invalid input that somehow slip through our validation.

dapplion (approved):
LGTM

jimmygchen:
@mergify queue

Issue #1785, opened 2020-10-19
Remove head tracker in favour of fork choice


michaelsproul:
## Description

The head tracker contains redundant information that is already present in the fork choice block DAG. This is undesirable as it means there isn't just "one source of truth", and the two data structures need to be kept in sync across concurrent executions, which is highly non-trivial (see #1771 for more)

## Steps to resolve

* Write code to derive the list of heads from fork choice
* Remove the `HeadTracker` struct from the beacon chain
* Port the migration code to use the head data from fork choice
* Work out how to reconcile the pruning of fork choice with pruning of the database. Possible approaches:
     * Mark blocks deleted from disk as `deleted` in fork choice, until they are cleaned up by fork choice's own pruning mechanism
     * Mirror the pruning of the on-disk database in fork choice (i.e. prune the two structures together)

dapplion:
Some notes for non-michael readers and myself

HeadTracker is used for two things currently:
- As optimisation to query heads quickly to serve API requests. Premature optimization IMO, or could be integrated into the fork-choice and does not require to be persisted on disk.
- To optimize pruning hot data, more below

## How to prune hot data

Blocks and states are persisted in the hot DB by root. To figure out what to prune you need to build a block DAG and compute what branches are non-viable / abandoned. Some ways to do it:

1. **"Brute-force"**: stream all items in the DB's hot block column, build graph, prune.
2. **With a HeadTracker**: helps with tip discovery, then iterate each head chain to figure out if it's viable or not. With the current `RootsIterator` requires to load a full BeaconState every 8192 slots.
3. **With fork-choice prune output**: fork-choice prune can return a Vec of pruned proto-nodes, from which you can build a partial graph of abandoned chains, and prune.

The pruning routine must be crash safe, options 2 and 3 risk "forgetting" data or causing inconsistencies like https://github.com/sigp/lighthouse/issues/4773 if not implemented properly.

## Brute-force

The current implementation uses the `RootsIterator` which loads a full BeaconState every 8192 slots. With a state of size of 100MB (likely more represented in memory) that's 12 KB per slot. Blinded beacon blocks are ~50KB? so that's not _too_ far off. I don't see any immediate issues with such approach besides the others being more optimal. Even in the worse case, pruning runs on a background thread so streaming some hundred MB from the DB in the case of long non-finality does not sound too bad. Thoughts?

## Using fork-choice

[...]

Issue #6580, opened 2024-11-12
Database growth during non-finality


michaelsproul:
## Description

Lighthouse presently stores 1 state per epoch in the hot database during periods of non-finality. During a multi-week period of non-finality this would lead to hundreds of gigabytes of states being stored and the likely exhaustion of many nodes' disk storage.

## Steps to resolve

@dapplion and I have discussed adapting the HDiff approach (from https://github.com/sigp/lighthouse/pull/5978) to the hot database, and think we have a fairly good scheme that goes as follows:

1. Store hdiffs in the hot database with `state_root: Hash256` references rather than `slot: Slot`.
2. Retain the diff path from the most recent snapshot to the split state, while pruning everything else with a slot prior to the split slot.

michaelsproul:
Fixed by:

- https://github.com/sigp/lighthouse/pull/6750

:tada:

PR #6744, opened 2024-12-28
Drop head tracker for summaries DAG

dapplion:
## Issue Addressed

The head tracker is a persisted piece of state that must be kept in sync with the fork-choice. It has been a source of pruning issues in the past, so we want to remove it
- see https://github.com/sigp/lighthouse/issues/1785

When implementing tree-states in the hot DB we have to change the pruning routine (more details below) so we want to do those changes first in isolation.
- see https://github.com/sigp/lighthouse/issues/6580
- If you want to see the full feature of tree-states hot https://github.com/dapplion/lighthouse/pull/39

Closes https://github.com/sigp/lighthouse/issues/1785

## Proposed Changes

**Current DB migration routine**

- Locate abandoned heads with head tracker
- Use a roots iterator to collect the ancestors of those heads can be pruned
- Delete those abandoned blocks / states
- Migrate the newly finalized chain to the freezer

In summary, it computes what it has to delete and keeps the rest. Then it migrates data to the freezer. If the abandoned forks routine has a bug it can break the freezer migration.

**Proposed migration routine (this PR)**

- Migrate the newly finalized chain to the freezer
- Load all state summaries from disk
- From those, just knowing the head and finalized block compute two sets: (1) descendants of finalized (2) newly finalized chain
- Iterate all summaries, if a summary does not belong to set (1) or (2), delete

This strategy is more sound as it just checks what's there in the hot DB, computes what it has to keep and deletes the rest. Because it does not rely and 3rd pieces of data we can drop the head tracker and pruning checkpoint. Since the DB migration happens **first** now, as long as the computation of the sets to keep is correct we won't have pruning issues.


dapplion:
@michaelsproul If we stop updating the head_tracker in the persisted head, how can we downgrade safely? If you run an old version of Lighthouse it will read the persisted head which contains an empty head tracker.

dapplion:
Test `store_tests::finalizes_non_epoch_start_slot` panics at 

```
---- store_tests::finalizes_non_epoch_start_slot stdout ----
thread 'store_tests::finalizes_non_epoch_start_slot' panicked at /Users/lion/code/sigp/lighthouse/beacon_node/beacon_chain/src/test_utils.rs:836:14:
called `Result::unwrap()` on an `Err` value: HotColdDBError(MissingEpochBoundaryState(0x8da1c38d20071452e39a518001b61aa94ad871c6b95896203ddb0b68cfb89e5f))
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: beacon_chain::test_utils::BeaconChainHarness<beacon_chain::builder::Witness<slot_clock::manual_slot_clock::ManualSlotClock,beacon_chain::eth1_chain::CachingEth1Backend<E>,E,Hot,Cold>>::hot_state_exists
   4: beacon_chain_tests::store_tests::finalizes_non_epoch_start_slot::{{closure}}
   5: tokio::runtime::runtime::Runtime::block_on
   6: core::ops::function::FnOnce::call_once
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

This is part of an assertion in the test calling `hot_state_exists`, which now instead of returning an `Ok(bool)` returns an `Err`.

This state is not pruned, but is the `new_finalized_state_hash`

```
Jan 02 10:31:34.615 DEBG Starting database pruning, new_finalized_state_hash: 0x8da1c38d20071452e39a518001b61aa94ad871c6b95896203ddb0b68cfb89e5f, new_finalized_checkpoint: Checkpoint { epoch: Epoch(1), root: 0x401d573d920244cb9b9988da85f8093f6bedb1e68156774b87c173f6ded6623b }, module: beacon_chain::migrate:485
Jan 02 10:31:34.619 DEBG Extra pruning information, states_to_prune: {}, blocks_to_prune: {}, states_to_prune_count: 0, blocks_to_prune_count: 0, finalized_and_descendant_block_roots: 40, state_summaries_count: 40, newly_finalized_states_min_slot: 8, newly_finalized_state_roots: 1, newly_finalized_blocks_min_slot: 7, newly_finalized_blocks: 1, new_finalized_checkpoint: Checkpoint { epoch: Epoch(1), root: 0x401d573d920244cb9b9988da85f8093f6bedb1e68156774b87c173f6ded6623b }, module: beacon_chain::migrate:645
```

Michael: figure out which of the hot_state_exists cases triggers, plus what state root and slot

beacon_node/store/src/hot_cold_store.rs
michaelsproul, line 1: Note to self: need to remove state deletion from `migrate_database` so that it doesn't interfere with DAG-based pruning

beacon_node/beacon_chain/src/migrate.rs
```diff
@@ -504,249 +476,278 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> BackgroundMigrator<E, Ho
...
+
+        // We don't know which blocks are shared among abandoned chains, so we buffer and delete
+        // everything in one fell swoop.
+        let mut blocks_to_prune: HashSet<Hash256> = HashSet::new();
+        let mut states_to_prune: HashSet<(Slot, Hash256)> = HashSet::new();
+
+        for (slot, summaries) in state_summaries_dag.summaries_by_slot_ascending() {
+            for (state_root, summary) in summaries {
+                let should_prune =
+                    if finalized_and_descendant_block_roots.contains(&summary.latest_block_root) {
```
michaelsproul: Note to self: I think we should prune based on ancestors of the finalized _state_. This will more neatly handle the case where we have e.g.

- block at slot 30 (canonical)
- block at slot 31 building on 30 (not canonical)
- skipped slot at slot 32 that is finalized (also finalizes slot 31 as canonical)

In this example, the block at slot 31 descends from the finalized block but not from the finalized checkpoint. So it should be pruned, along with all its descendants.

dapplion:
@michaelsproul I have
- Fixed conflicts
- Added migration
- Remove any hot DB op from migrate_db, looks like an oversight when creating this PR initially
- Addressed the case where the finalized block is skipped, now we keep only descendants of the finalized checkpoint

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -1457,12 +1410,21 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
     ///
     /// Returns `(block_root, block_slot)`.
     pub fn heads(&self) -> Vec<(Hash256, Slot)> {
-        self.head_tracker.heads()
+        let head_slot = self.canonical_head.cached_head().head_slot();
+        self.canonical_head
+            .fork_choice_read_lock()
+            .proto_array()
+            .viable_heads::<T::EthSpec>(head_slot)
```
michaelsproul: This `head_slot` is being passed as the `current_slot`, which is not necessarily the same thing. I think we should probably read the slot clock and use its current slot?

michaelsproul: Head slot is the slot of the head block, which could be potentially quite old. This could affect fork choice calculations related to pulled tips, etc.

beacon_node/beacon_chain/src/migrate.rs
```diff
@@ -504,249 +473,312 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> BackgroundMigrator<E, Ho
...
-        if old_finalized_slot > new_finalized_slot {
-            return Ok(PruningOutcome::OutOfOrderFinalization {
-                old_finalized_checkpoint,
-                new_finalized_checkpoint,
-            });
-        }
+        // TODO(hdiff): if we remove the check of `old_finalized_slot > new_finalized_slot` can we
+        // ensure that a single pruning operation is running at once? If a pruning run is triggered
+        // with an old finalized checkpoint it can derive a stale hdiff set of slots and delete
+        // future ones that are necessary breaking the DB.
```
michaelsproul: I don't _think_ we have to worry about this, because finalization should never go backwards, and we sequence all migration messages (triggered on finalization) through a FIFO channel.

chong-he:
Done some testing using binary built with the commit 12fa5a8. 

1. The node can fully sync an archived node, although this error appears from time to time:

```
ERRO Error fetching block for peer, error: DBError(HotColdDBError(MissingExecutionPayload(0x8147657ef7e6e0331b59ae888714ba8d60094d499a5522e7efd9f75ea2bad298))), block_root: 0x8147657ef7e6e0331b59ae888714ba8d60094d499a5522e7efd9f75ea2bad298, module: network::network_beacon_processor::rpc_methods:817
```	

The execution client (Nethermind) is running normally, perhaps it's something to do with the DB? 
	
	
2. Upgrading to V23 caused panic.
   Steps to reproduce: Run Lighthouse V6 and checkpoint sync, wait for a minute or two, stop Lighthouse V6, change binary to this PR, and got error:

```
Feb 10 03:09:30.723 CRIT Task panic. This is a bug!              advice: Please check above for a backtrace and notify the developers, backtrace:    0: lighthouse::run::{{closure}}
```	

   The last line of the backtrace shows:
```	
29: clone3
          at ./misc/../sysdeps/unix/sysv/linux/x86_64/clone3.S:78
```	
(I have uploaded the full backtrace error in the link shared with you previously.)

When try to start agin with the PR binary, it failed:
	
```
Feb 10 03:09:48.935 INFO Blob DB initialized                     oldest_data_column_slot: None, oldest_blob_slot: Some(Slot(3589184)), path: "/home/ck/.lighthouse/holesky/beacon/blobs_db", service: freezer_db
Feb 10 03:09:48.940 CRIT Failed to start beacon node             reason: Unable to open database: SszDecodeError(InvalidByteLength { len: 32, expected: 64 })
Feb 10 03:09:48.940 INFO Internal shutdown received              reason: Failed to start beacon node
```	
	
If we change back to V6 binary, it produced the error:
	
```
Feb 10 03:10:25.051 INFO Blob DB initialized                     oldest_data_column_slot: None, oldest_blob_slot: Some(Slot(3589184)), path: "/home/ck/.lighthouse/holesky/beacon/blobs_db", service: freezer_db
Feb 10 03:10:27.614 INFO Starting checkpoint sync                remote_url: https://holesky.beaconstate.ethstaker.cc/, service: beacon
Feb 10 03:10:51.567 INFO Downloading genesis state               info: this may take some time on testnets with large validator counts, timeout: 180s, server: https://holesky.beaconstate.ethstaker.cc/, service: beacon
Feb 10 03:11:08.982 INFO Loaded checkpoint block and state       block_root: 0x6bb4a025905caa2b674e0d871afeaae9f6e15b8b32d6481b51981dd05ceb42f6, state_slot: 3604480, block_slot: 3604480, service: beacon
Feb 10 03:11:08.985 INFO Loaded deposit tree snapshot            deposits loaded: 483386, service: beacon
Feb 10 03:11:17.146 CRIT Failed to start beacon node             reason: Failed to initialize anchor info: AnchorInfoConcurrentMutation
Feb 10 03:11:17.146 INFO Internal shutdown received              reason: Failed to start beacon node
```
	
We can see that it is starting checkpoint sync again (even though the database is there and I don't use the flag `--purge-db-force`). It should skip checkpoint sync, but it doesn't.

It sort of "ignores" the corrupted database and starts a fresh checkpoint sync on its own.
	
3. Downgrading of database schema caused panic. Steps to reproduce: run a node with the PR, then stop the node, apply the downgrade command:
`/usr/local/bin/lighthouse6744 db migrate --to 22 --network holesky --datadir /path`

Last line of backtrace:
```
  18: _start
, message: None, location: /home/ck/testing/lighthouse/beacon_node/beacon_chain/src/schema_change/migration_schema_v23.rs:108:5
```

consensus/proto_array/src/proto_array.rs
```diff
@@ -1041,6 +1041,40 @@ impl ProtoArray {
...
+    /// Returns all nodes that have zero children and are viable heads
+    pub fn viable_heads<E: EthSpec>(&self, current_slot: Slot) -> Vec<&ProtoNode> {
+        self.nodes_without_children()
+            .into_iter()
+            .filter_map(|index| {
+                // An unknown index is not a viable head
+                if let Some(node) = self.nodes.get(index) {
+                    if self.node_is_viable_for_head::<E>(node, current_slot) {
+                        return Some(node);
+                    }
```
michaelsproul: **This check** is the reason for the `pruning_does_not_touch_blocks_prior_to_finalization` test failing at the moment.

The reason is:

- Head is at slot 40.
- Epoch 1 is finalized.
- The alternate head is at slot 15. It descends from the finalized block _but_ it is not viable to be the head because its justified/finalized checkpoints are inferior (still at genesis). Furthermore, the allowance for `voting_source.epoch + 2 >= current_epoch` no longer applies because the current epoch is already 5.

So we have a block at slot 15 which although it descends from the finalized head, is never viable to become head, and could easily have been pruned.

I think we have three options at this point:

1. Allow for this situation. Keep our implementation as-is in this PR, and update the test (and docs) to be permissive in the case of unviable heads that aren't pruned, OR
2. Remove the viability check from fork choice's function producing heads. This would be more consistent, and given that the list of heads is only used for informational purposes, seems low risk.
3. Try to prune this head when pruning at epoch 1. This would be a more drastic change, as it implies we should probably use fork choice to prune. This feels like it could open a whole can of worms, as we'd want to avoid holding a write lock on fork choice while pruning runs, yet this might be the only way to ensure that new blocks aren't imported atop blocks that we are pruning.

I think my preference for now is for (2), or maybe (1).

consensus/proto_array/src/proto_array.rs
```diff
@@ -1041,6 +1041,40 @@ impl ProtoArray {
...
+            // Add this node to the parent's child list if any
+            if let Some(parent_index) = node.parent {
+                childs_of.entry(parent_index).or_default().push(index);
+            }
+        }
+        childs_of
+            .into_iter()
+            .filter(|(_, childs)| childs.is_empty())
+            .map(|(index, _)| index)
+            .collect()
```
michaelsproul: Is there a difference between this and just:

```suggestion
        self.nodes
            .iter()
            .enumerate()
            .filter_map(|(i, node)| node.best_child.is_none().then_some(i))
            .collect()
```

consensus/proto_array/src/proto_array.rs
dapplion: Let's do 2 and add a comment on the function docs

consensus/proto_array/src/proto_array.rs
dapplion: Hmm right, we keep best_child updated

consensus/proto_array/src/proto_array.rs
michaelsproul: Done here: 7abbaebf1209111f5399e3f1b994616fc10a596c

Needed to check descent from the finalized checkpoint, as fork choice is pruned quite lazily, so there end up being nodes in fork choice that are not descended from finalization, and these were making some of the tests fail.

michaelsproul:
Found and fixed the payload deletion bug that @chong-he identified, in this commit: https://github.com/sigp/lighthouse/pull/6744/commits/abb3c3f99b107d497ee87cbafb568b2f0ab6b019

I also fixed up the migration so that it doesn't panic. Maybe if this passes CI we can retry testing, and look towards merging soon :tada:

beacon_node/beacon_chain/src/summaries_dag.rs
```diff
@@ -0,0 +1,399 @@
...
+                            state_summaries_by_block_root.get(&parent_block_root)
+                        {
+                            *parent_block_summaries
+                                .get(&previous_slot)
+                                // Should never error: summaries are continuous, so if there's an
+                                // entry it must contain at least one summary at the previous slot.
+                                .ok_or(Error::MissingStateSummaryAtSlot(
+                                    parent_block_root,
+                                    previous_slot,
+                                ))?
```
michaelsproul: We hit this error on Holesky. So either the node's DB is corrupt, or there's a bug and this is not unreachable.

I haven't had a chance to investigate yet.

> Mar 02 23:48:43.557 WARN Hot DB pruning failed                   error: PruningError(SummariesDagError("creating StateSumariesDAG", MissingStateSummaryAtSlot(0x54a7368f6e511e3c2adad25a87ea71798d0a8a426cfcfdc9f7a81a4b1a5f2588, Slot(3752575)))), service: beacon

michaelsproul: Thinking out loud:

The block root `0x54a7368f6e511e3c2adad25a87ea71798d0a8a426cfcfdc9f7a81a4b1a5f2588` is from the canonical chain at slot 3752570.

The block root here is the `parent_block_root`, so it appears we are dealing with a state summary at slot 3752576 for some non-canonical block at slot 3752576 which builds upon 3752570. We are erroring when trying to find the state summary for the previous slot 3752575. We've already checked whether there was a summary for the slot 3752576 block root at 3752575 and found none.

Either:

- The non-canonical block _is_ at slot 3752576 and we are missing the state summary for block 3752570 pulled up to 3752575.
- The non-canonical block is at a slot prior to 3752576 and we are missing the state summary for this block pulled up to 3752575.

I think we can differentiate these two cases by looking at the `latest_block_slot` in the summary. I will try that now.

michaelsproul: Haven't been able to repro this on Holesky. Maybe we could revive the fuzzer version of Hydra to try to break this

michaelsproul: Theory: 

This could happen because of temporary states, and an assumption in block processing that the pre-state is only advanced by 1 slot. Sequence of events that could trigger this:

- There's a block at slot 70. We start importing a block at slot 75. We write states with temporary flags for slots 71, 72, 73, 74. The block is invalid (or fails processing due to out-of-disk space), so we do not delete the temporary flags for these slots.
- There's a block at slot 76 building upon 70. It fetches as its pre-state an advanced state of slot 70, which is present in the state cache, and advanced through more than 1 slot, e.g. it's the state from slot 73 stored above.
- The calculation of initial `confirmed_state_roots` while processing block 76, sees that state 73 is not yet confirmed, so stages a deletion of its temporary state. HOWEVER, we do not stage deletion of temporary flags for 71 and 72, which is what leads to the bug:

https://github.com/sigp/lighthouse/blob/abb3c3f99b107d497ee87cbafb568b2f0ab6b019/beacon_node/beacon_chain/src/block_verification.rs#L1443-L1452
- Upon a restart, 71 and 72 get deleted during pruning of temporary states (which we saw a lot on Holesky), thus violating the continuity of state summaries on disk, i.e. we have a gap in the chain of state summaries between 70 and 73.

dapplion: Do we want to preserve the behaviour of test:
- `garbage_collect_temp_states_from_failed_block_on_startup`: Intermediary states of invalid block get pruned at startup

Note to self: `garbage_collect_temp_states_from_failed_block_on_finalization` is failing because `iter_temporary_state_roots` returns empty vec now

michaelsproul: maybe we can try adjusting the tests to the paradigm where we prune on startup wrt the current split?

dapplion: I'm looking into this. Running this PR's pruning code won't clean the advanced states of invalid blocks as they are descendant of finalized. We would need some special pruning routine to prune advanced states. We can't differentiate the advanced states of a no-longer-existing invalid block vs the advanced states of the head.

Consider a case where a network has a head with a lot of weight but block production stops for 2 weeks. To continue the chain you need that post head state advanced two weeks. That advanced state is valuable and should not be pruned. How can we distinguish that state from a fork for an invalid block?

I'm gravitating towards doing nothing at start-up. If someone wants to attack LH explicitly they can just create actually valid blocks and we always must take the loss. The garbage collection at start-up is moderately useful when the network is spamming bad blocks that skip a lot for some reason.

michaelsproul: yeah sounds good, so we just delete those two tests?

michaelsproul:
I've resolved conflicts with `unstable` after the merge of `tracing`. We should double check that I didn't mess up any logs or delete any code in that merge resolution, as it touched _a lot_ of stuff.

michaelsproul:
TODO:

- [x] can probably delete `prune_old_hot_states` https://github.com/sigp/lighthouse/pull/6744/commits/11cfa1c058f22df2cdad6f0a382d0f903282e0b1
- [x] can delete `load_state_temporary_flag` and friends https://github.com/sigp/lighthouse/pull/6744/commits/11cfa1c058f22df2cdad6f0a382d0f903282e0b1
- [x] delete all temporary flags in the upgrade migration https://github.com/sigp/lighthouse/pull/6744/commits/f4dd6fe060b3c590d1af788f517ded58c868f748

michaelsproul (approved):
Merging this down so we can simplify the conga line.

This has been running well on our infra.

michaelsproul:
Just going to do one more rollout and ensure logs are OK.

michaelsproul:
Yep this is looking solid. Merging! :tada:

Issue #6510, opened 2024-10-18
Store execution payloads during backfill if `--prune-payloads false`


michaelsproul:
## Description

@KolbyML notes that the ability to keep execution payloads during block backfill would be useful for syncing Trin (portal network client). It's a bit of a gotcha that at the moment Lighthouse throws out execution payloads during backfill sync, even when running with `--prune-payloads false`. We download the payloads from peers, and then throw them straight in the bin!

https://github.com/sigp/lighthouse/blob/2e0eb6d1b8705bbda2ba56eb195d9cc7c6575e95/beacon_node/beacon_chain/src/historical_blocks.rs#L121-L134

## Steps to resolve

If `prune-payloads` is `false`, store execution payloads in the hot DB as part of `import_historical_block_batch`.

## Additional Info

Should probably be implemented after the hdiff PR is merged, which modifies the backfill logic:

- https://github.com/sigp/lighthouse/pull/5978

KolbyML:
Just to clarify some terminology as we (EF Portal) use a lot of confusing terminology.

- Portal: is a collection of p2p protocols built on the Portal-wire-specification which is a framework for building distributed storage networks or p2p networks in general 
- Trin: is an implementation of the Portal specification and all the sub-protocols (history,state,beacon,etc). All networks are optional to implement.
- Trin Execution: is an Execution client we are implementing comparable to Geth/Nethermind/Besu/etc but with the key difference of we will not be using devp2p. So instead we will need to rely on “The Portal Networks”, archival formats, and Consensus clients to achieve what devp2p provides.

Our initial usecase of Trin Execution will be gossiping genesis-latest execution state diffs to the “Portal State Network”, but it will likely become a full Execution client released to the public in time. It also serves as a backup solution if we don’t get adequate adoption by major EL clients.

So this change would be very nice for syncing Trin Execution!

michaelsproul:
Unblocked, will try to get this in for the next release

chong-he:
I would like to give this a go

michaelsproul:
Part of the changes required for this feature (pruning) are in this PR:

- https://github.com/sigp/lighthouse/pull/6745

That PR updates the pruning to go back to the oldest_block_slot (rather than the slot of the downloaded checkpoint), which ensures that we will prune payloads between `oldest_block_slot` and `anchor_slot` if they are backfilled.

chong-he:
Closed by #6766

PR #6745, opened 2024-12-28
Use oldest_block_slot to break off pruning payloads

dapplion:
## Issue Addressed

Lighthouse DB persists an "anchor" which has some useful data. One is the `anchor_slot`. Tree-states on the hot DB wants to use that value and change it's meaning slightly. The only blocker for that change is `try_prune_execution_payloads` which uses the anchor_slot as optimization to break of early the pruning routine.

BUT it's an easy fix. From @michaelsproul 

> I was thinking maybe we could recycle the `anchor_slot` for use here, as it is _fairly useless_ and not used for much. The only load-bearing use of the `anchor_slot` that I can find is in `try_prune_execution_payloads` where it is used to halt the pruning process. However, this is not necessary as we could either:
>
> 1. Keep iterating back to Bellatrix, or
> 2. Stop iterating back once we find a payload that is missing, or
> 3. Use the `oldest_block_slot` to determine when to stop iterating back. This is my preferred option, as it is also compatible with storing execution payloads older than the `anchor_slot`, i.e. https://github.com/sigp/lighthouse/issues/6510

ref https://github.com/dapplion/lighthouse/pull/39#discussion_r1891039644

## Proposed Changes

This PR implements option 3 of the above citation, quite painless

### DIFF
--- a/beacon_node/store/src/hot_cold_store.rs
+++ b/beacon_node/store/src/hot_cold_store.rs
@@ -2629,7 +2629,7 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold> 
             "Pruning finalized payloads";
             "info" => "you may notice degraded I/O performance while this runs"
         );
-        let anchor_slot = self.get_anchor_info().anchor_slot;
+        let anchor_info = self.get_anchor_info();
 
         let mut ops = vec![];
         let mut last_pruned_block_root = None;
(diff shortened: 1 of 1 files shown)


beacon_node/store/src/hot_cold_store.rs
```diff
@@ -2670,10 +2670,10 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
                 ops.push(StoreOp::DeleteExecutionPayload(block_root));
             }
 
-            if slot == anchor_slot {
+            if slot < anchor_info.oldest_block_slot {
```
michaelsproul: I think we want to stop as soon as we hit the oldest block slot, i.e. `==`

```suggestion
            if slot <= anchor_info.oldest_block_slot {
```

michaelsproul (approved):

michaelsproul:
@mergify queue

michaelsproul:
@mergify dequeue

michaelsproul:
@mergify requeue

michaelsproul:
@mergify dequeue

michaelsproul:
@mergify queue

Issue #6746, opened 2024-12-29
feat: presigned validator exit txn


hamdiallam:
## Description

Would like to store a presigned validator exit transaction as a last resort mechanism for publishing an exit.

Rational
- Revealed presigned txn will result in potential unwanted exit. This okay as the validator can generate new keys and redeposit since the withdrawal address should always be set
- Can be a bit more liberal with how this presigned txns is backed up and accessed as a last resort mechanism

## Present Behaviour

```$ lighthouse account validator exit``` only supports sign + broadcast the withdrawal txn

## Added Behaviour

Added flag to log and exit the withdrawal txn

```
$ lighthouse account validator exit --presign
0x...
``

michaelsproul:
We are soft-deprecating `lighthouse account` commands in favour of `lighthouse validator-manager` ones. The PR for exits in the validator manager will be merged soon:

- https://github.com/sigp/lighthouse/pull/6612

I think we should add this feature to that branch

hamdiallam:
@michaelsproul thanks for the quick reply!

sounds great, i'll take a look at that branch

jimmygchen:
Completed in #6747

PR #6747, opened 2024-12-30
feat: presign for validator account

hamdiallam:
## Issue Addressed

#6746 

## Proposed Changes

Add a --presign flag to emit the json output to stdout instead of publishing the exit

## Additional Info

The validator-manager is also a surface area to support via the http client interface


account_manager/src/validator/exit.rs
```diff
@@ -154,9 +168,26 @@ async fn publish_voluntary_exit<E: EthSpec>(
...
+            .map_err(|e| format!("Unable to convert to JSON: {}", e))?;
+
+        println!("{}", string_output);
+        return Ok(());
+    }
+
     eprintln!(
         "Publishing a voluntary exit for validator: {} \n",
         keypair.pk
+
```
chong-he:
```suggestion
```

account_manager/src/validator/exit.rs
```diff
@@ -154,9 +168,26 @@ async fn publish_voluntary_exit<E: EthSpec>(
...
+        eprintln!(
+            "Successfully pre-signed voluntary exit for validator {}. Not publishing.",
+            keypair.pk
+        );
+
+        // Convert to JSON and print
+        let string_output = serde_json::to_string_pretty(&signed_voluntary_exit)
+            .map_err(|e| format!("Unable to convert to JSON: {}", e))?;
+
+        println!("{}", string_output);
```
chong-he, line 184: While working on #6612 , initially I also printed out the signature. But then I realise it probably isn't much useful in practice - if users want to pre-sign an exit, they probably want the signature saved to a file. So I modified in this commit in that PR:
https://github.com/sigp/lighthouse/pull/6612/commits/5a4f928dc6c1dc8e35ea1e91742cef4c8c9d2e58

We can hear from others about this and edit accordingly.

account_manager/src/validator/exit.rs
```diff
@@ -74,6 +76,15 @@ pub fn cli_app() -> Command {
                 .action(ArgAction::SetTrue)
                 .help_heading(FLAG_HEADER)
         )
+        .arg(
+            Arg::new(PRESIGN)
```
chong-he, line 80: Perhaps we want to standardise the flag name to be the same with #6612 , depends on which is preferred. I am happy to change the flag name in #6612 if `presign` is adopted.

chong-he (commented):
Thank you for the PR! I have tested it and the feature is working. I have left some comments. 

The CI check on formatting failed, also it would be great if you can sign the CLA.

I think it will also be nice to have a note highlighting the feature in the documentation, maybe before this section: https://lighthouse-book.sigmaprime.io/voluntary-exit.html#initiating-a-voluntary-exit

Thanks!

account_manager/src/validator/exit.rs
chong-he: I have changed the flag name in #6612 to `--presign` as per the comment: https://github.com/sigp/lighthouse/pull/6612#discussion_r1924579320

So please ignore the above

michaelsproul:
Marking this `waiting-on-author` as we need @hamdiallam to please sign the contributor agreement :pray:

hamdiallam:
Ah sorry for dropping the ball on this! Will address when I get home tonight

hamdiallam:
done!

account_manager/src/validator/exit.rs
michaelsproul: I think this is fine for now. Users can still pipe stdout to a file

michaelsproul (approved):

PR #6577, opened 2024-11-09
Add CLI flag for HTTP API token path (VC)

syjn99:
## Issue Addressed

#6563

## Proposed Changes

- Added `http-token-path` flag for users who want to set the token path explicity.
- Added documentation for CLI reference, etc.

## Additional Info

- Throw error when `http-token-path` is actually a directory.
- If parent directory doesn't exist, make parent directory.

PR #6748, opened 2025-01-01
Fix incorrect VC default HTTP token path when the `--datadir` flag is present

jimmygchen:
## Issue Addressed

Fixes a validator client bug introduced in #6577.

When a custom `--datadir` flag is provided, the default http token path changes to `$HOME/.lighthouse/mainnet/validators/api-token.txt`, rather than the previous default `<datadir>/validators/api-token.txt`

This causes a `CRIT` in some setups where the validator client does not have write access to the home directory (e.g. `eth-docker`):

```
CRIT Failed to start validator client        reason: Unable to create parent directories for "/nonexistent/.lighthouse/mainnet/validators/api-token.txt": Os { code: 13, kind: PermissionDenied, message: "Permission denied" }
``` 

## Proposed Changes

Set the correct default VC API token path, if a `datadir` or `validators-dir` is provided.

### DIFF
--- a/lighthouse/tests/validator_client.rs
+++ b/lighthouse/tests/validator_client.rs
@@ -345,7 +345,7 @@ fn http_store_keystore_passwords_in_secrets_dir_present() {
 }
 
 #[test]
-fn http_token_path_flag() {
+fn http_token_path_flag_present() {
     let dir = TempDir::new().expect("Unable to create temporary directory");
     CommandLineTest::new()
         .flag("http", None)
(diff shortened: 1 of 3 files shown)


jimmygchen:
This wasn't included in the v6.0.1 release, so a patch isn't required 😌

pawanjay176 (approved):
Took me a while to see the difference. Nice catch!

jimmygchen:
@mergify requeue

jimmygchen:
@mergify queue

syjn99:
Thanks for catching this! I made a mistake in my previous contribution, and I appreciate your help in fixing it.

jimmygchen:
All good, I missed it in review too 🙏

PR #6750, opened 2025-01-02
Hierarchical state diffs in hot DB

dapplion:
## Proposed Changes

This PR implements https://github.com/sigp/lighthouse/pull/5978 (tree-states) but on the hot DB. It allows Lighthouse to massively reduce its disk footprint during non-finality and overall I/O in all cases.

Closes https://github.com/sigp/lighthouse/issues/6580

Conga into https://github.com/sigp/lighthouse/pull/6744

### TODOs

- [x] Fix OOM in CI https://github.com/sigp/lighthouse/pull/7176
- [x] optimise store_hot_state to avoid storing a duplicate state if the summary already exists (should be safe from races now that pruning is cleaner)
- [x] mispelled: get_ancenstor_state_root
- [x] get_ancestor_state_root should use state summaries
- [x] Prevent split from changing during ancestor calc
- [x] Use same hierarchy for hot and cold

### TODO Good optimization for future PRs

- [ ] On the migration, if the latest hot snapshot is aligned with the cold snapshot migrate the diffs instead of the full states.
```
align slot  time
10485760    Nov-26-2024
12582912    Sep-14-2025
14680064    Jul-02-2026
```

### TODO Maybe things good to have

- [ ] Rename anchor_slot https://github.com/sigp/lighthouse/compare/tree-states-hot-rebase-oom...dapplion:lighthouse:tree-states-hot-anchor-slot-rename?expand=1
- [ ] Make anchor fields not public such that they must be mutated through a method. To prevent un-wanted changes of the anchor_slot

### NOTTODO

- [ ] Use fork-choice and a new method [`descendants_of_checkpoint`](https://github.com/sigp/lighthouse/commit/ca2388e1962b2ea60d3f22313a843285928c1542#diff-046fbdb517ca16b80e4464c2c824cf001a74a0a94ac0065e635768ac391062a8) to filter only the state summaries that descend of finalized checkpoint]

## Additional Info

Original WIP PR with lots of context and discussion 
- https://github.com/dapplion/lighthouse/pull/39

Some resources to get up to speed with HDiff concepts:
- https://github.com/dapplion/tree-states-review-guide/blob/main/persisted_hdiff.md
- https://hackmd.io/G82DNSdvR5Osw2kg565lBA


chong-he:
I have done some testing on this on the Holesky testnet as mentioned in #6775. Regular checkpoint sync works, so does syncing from a checkpoint >24 hours ago. Some findings:

1. Seeing the following logs when syncing an archive node:
```Jan 26 08:04:01.957 DEBG MissingHotStateSummary, existing_summaries: [(0x6cad1dd1b28bee2131667bfcfca619b89f7220898a7e2ef920df057435c86cf8, Slot(3483616)), (0xf191896412d173efc8a740acd8224d34c60108ab0327f5e21c9b803592386115,
Jan 26 08:04:01.957 ERRO Database write failed, action: reverting blob DB changes, error: MissingHotStateSummary(0x554bd524427dfc0a0efcabbdc65f087dcd053385fd378cc83aa78a5d9b9db167), service: freezer_db, module: store::hot_cold_store:1390
Jan 26 08:04:01.957 ERRO Database write failed!, error: MissingHotStateSummary(0x554bd524427dfc0a0efcabbdc65f087dcd053385fd378cc83aa78a5d9b9db167), msg: Restoring fork choice from disk, service: beacon, module: beacon_chain::beacon_chain:3957
```
(The first line of this is until slot 3498015, I truncated it)
Shortly after this, the following appears:
```
Jan 26 08:04:03.946 CRIT Beacon block processing error, error: DBError(MissingHotStateSummary(0x554bd524427dfc0a0efcabbdc65f087dcd053385fd378cc83aa78a5d9b9db167)), service: beacon, module: beacon_chain::beacon_chain:3377
Jan 26 08:05:49.160 CRIT Beacon block processing error, error: MissingBeaconBlock(0x831d25d70b5bdfe7544c0cac196cb781fcea3e9c7c5d6e902ad82059d27fc8ed), service: beacon, module: beacon_chain::beacon_chain:3377
Jan 26 08:12:17.256 WARN BlockProcessingFailure, outcome: MissingBeaconBlock(0x831d25d70b5bdfe7544c0cac196cb781fcea3e9c7c5d6e902ad82059d27fc8ed), msg: unexpected condition in processing block., module: network::network_beacon_processor::sync_methods:747
```

The beacon node wouldn't be able to start upon a restart:
```
Jan 27 05:57:58.731 CRIT Failed to start beacon node, reason: Failed to build beacon chain: Head block not found in store, module: lighthouse:721
```

Another point worth noting is that the logs seem to be in a loop. Similar logs appear repeatedly in the log file, to the extent that it would only take a few seconds/minutes to fill up a 200MB log file. These are the repeated logs:

```
Jan 26 08:02:47.632 DEBG Created block lookup, id: 1753067, awaiting_parent: none, block_root: 0x523bff6d855c77b4f14ed5d7639a3f4975cef5115e0f0631f2f730665229fbc6, peer_ids: [PeerId("16Uiu2HAm4qmfW7E7E9rkvNVAWeZNWMhZ4VefKJHhjtFDUKaJAGAb"), PeerId("16Uiu2HAkvWFRpCLuiRyxvse9JzXP5HjgTMJfwPcHUCCa5NGwaike"),  PeerId("16Uiu2HAm9pjmagyJFvMkn3bVDPWTVwgyLJ3QdsejndLJnRhz2wfZ")], service: lookup_sync, service: sync, module: network::sync::block_lookups:401
Jan 26 08:02:47.632 DEBG Dropping completed lookup, id: 1753067, block: 0x523bff6d855c77b4f14ed5d7639a3f4975cef5115e0f0631f2f730665229fbc6, service: lookup_sync, service: sync, module: network::sync::block_lookups:781
Jan 26 08:02:47.632 DEBG Continuing child lookup, block_root: 0xf8eb5af88a4ea962e91a6fa77c69e1efa17005c346e7921d4218d0e84ff5a149, id: 113701, parent_root: 0x523bff6d855c77b4f14ed5d7639a3f4975cef5115e0f0631f2f730665229fbc6, service: lookup_sync, service: sync, module: network::sync::block_lookups:733
Jan 26 08:02:47.632 DEBG Sending blobs for processing, id: 113701, block: 0xf8eb5af88a4ea962e91a6fa77c69e1efa17005c346e7921d4218d0e84ff5a149, service: sync, module: network::sync::network_context:1086
Jan 26 08:02:47.632 DEBG RPC blobs received, commitments: [0xb8b1…3151, 0xb0a9…8f7d, 0xb29d…876a], slot: 3497985, block_root: 0xf8eb5af88a4ea962e91a6fa77c69e1efa17005c346e7921d4218d0e84ff5a149, indices: [0, 1, 2], module: network::network_beacon_processor::sync_methods:267
Jan 26 08:02:47.632 WARN Error when importing rpc blobs, slot: 3497985, block_hash: 0xf8eb5af88a4ea962e91a6fa77c69e1efa17005c346e7921d4218d0e84ff5a149, error: ParentUnknown { parent_root: 0x523bff6d855c77b4f14ed5d7639a3f4975cef5115e0f0631f2f730665229fbc6 }, module: network::network_beacon_processor::sync_methods:318
Jan 26 08:02:47.632 DEBG Received lookup processing result, result: Err(ParentUnknown { parent_root: 0x523bff6d855c77b4f14ed5d7639a3f4975cef5115e0f0631f2f730665229fbc6 }), id: 113701, block_root: 0xf8eb5af88a4ea962e91a6fa77c69e1efa17005c346e7921d4218d0e84ff5a149, component: Blob, service: lookup_sync, service: sync, module: network::sync::block_lookups:552
Jan 26 08:02:47.632 DEBG Marking lookup as awaiting parent, parent_root: 0x523bff6d855c77b4f14ed5d7639a3f4975cef5115e0f0631f2f730665229fbc6, block_root: 0xf8eb5af88a4ea962e91a6fa77c69e1efa17005c346e7921d4218d0e84ff5a149, id: 113701,
```
(The first line of the above log is truncated)

2. On another try of syncing an archive node:
```
Jan 22 01:27:34.393 ERRO Database write failed, action: reverting blob DB changes, error: MissingHotStateSummary(0x21262323cc023403605249a57849c267debad4612d7c340523fab29ded71bdb5), service: freezer_db, module: store::hot_cold_store:1390
Jan 22 01:27:34.393 ERRO Database write failed!, error: MissingHotStateSummary(0x21262323cc023403605249a57849c267debad4612d7c340523fab29ded71bdb5), msg: Restoring fork choice from disk, service: beacon, module: beacon_chain::beacon_chain:3957
Jan 22 01:28:51.854 WARN Hot DB pruning failed, error: PruningError(MissingSummaryForFinalizedCheckpoint(0xd395c3f4e6d3eafd733a5554b3191e20f670c444916d868d68157aa07f0887fb)), service: beacon, module: beacon_chain::migrate:381
```

The beacon node wouldn't be able to start upon a restart:
```
Jan 22 03:55:27.130 INFO Blob DB initialized                     oldest_data_column_slot: None, oldest_blob_slot: Some(Slot(3336224)), path: "/var/lib/lighthouse_test1/beacon/blobs_db", service: freezer_db
Jan 22 03:55:28.186 CRIT Failed to start beacon node             reason: Unable to open database: HotColdDBError(MissingSplitState(0x855420d7ff98dca1dbe028ef624b18a81a027ab0cb1f95e4c048ad9b3c4510c4, Slot(3467232)))
Jan 22 03:55:28.186 INFO Internal shutdown received              reason: Failed to start beacon node
```


3. I have not successfully reconstructed all states (i.e., finished syncing an archived node) on this branch (due to the above errors). It is either stuck during historical blocks download or during state reconstruction. A similar error to no. 2 occurs regularly (during state reconstruction):

```
Jan 28 07:44:58.442 ERRO Database write failed                   action: reverting blob DB changes, error: Hdiff(LessThanStart(Slot(3504672), Slot(3504704))), service: freezer_db
Jan 28 07:44:58.443 WARN Hot DB pruning failed                   error: DBError(HotColdDBError(Rollback)), service: beacon
```
(slot 3504704 is the anchor slot from the database info)

4. Upgrading from v6.0.1: 
Checkpoint sync using v6.0.1, stop Lighthouse once it has checkpoint synced, then change the binary file to this PR branch. Sometimes it can upgrade smoothly, but sometimes it has error:

```
Jan 28 08:49:27.224 INFO Blob DB initialized                     oldest_data_column_slot: None, oldest_blob_slot: Some(Slot(3497888)), path: "/home/ck/.lighthouse/holesky/beacon/blobs_db", service: freezer_db
Jan 28 08:49:27.556 CRIT Failed to start beacon node             reason: Unable to open database: MigrationError("Migrating from SchemaVersion(22) to SchemaVersion(23): MigrationError(\"error computing states summaries dag MissingStateSummary(0x9d6efd378136fd8a08574b9a7f11b3f04ff5c5a4c15c0eaa64eac0c82f2aadee)\")")
Jan 28 08:49:27.557 INFO Internal shutdown received              reason: Failed to start beacon node
```

If this error occurs and I change back to v6.0.1 and let it syncs for a few minutes, and then change back again to this binary PR, it could work (upgrade is successful) and the database schema shows V23:

```
Jan 28 08:54:08.619 INFO Blob DB initialized                     oldest_data_column_slot: None, oldest_blob_slot: Some(Slot(3497504)), path: "/home/ck/.lighthouse/holesky/beacon/blobs_db", service: freezer_db
Jan 28 08:54:14.692 INFO Hot states migration in progress        summaries_written: 33, diff_written: 2
Jan 28 08:54:17.910 INFO Hot states migration complete           summaries_written: 99, diff_written: 3
Jan 28 08:54:20.345 INFO Refusing to checkpoint sync             msg: database already exists, use --purge-db to force checkpoint sync, service: beacon
```

However, I notice the `anchor_slot` is changed after upgrading to V23 (e.g., advancing by 1 epoch). Is this expected?

dapplion:
Working on a rebased version here https://github.com/sigp/lighthouse/compare/drop-headtracker...dapplion:lighthouse:tree-states-hot-rebase-2?expand=1

Looks much cleaner!

beacon_node/store/src/config.rs
```diff
@@ -62,18 +62,19 @@ pub struct StoreConfig {
...
     #[superstruct(only(V1))]
     pub slots_per_restore_point: u64,
     /// Prefix byte to future-proof versions of the `OnDiskStoreConfig` post V1
-    #[superstruct(only(V22))]
+    #[superstruct(only(V22, V24))]
     version_byte: u8,
-    #[superstruct(only(V22))]
+    #[superstruct(only(V22, V24))]
     pub hierarchy_config: HierarchyConfig,
+    // TODO(hdiff): Should persist the hot hierarchy_config too?
```
michaelsproul: yes

michaelsproul:
mispelled: `get_ancenstor_state_root`

EDIT: fixed now

michaelsproul:
TODO:

- ~~optimise `store_hot_state` to avoid storing a duplicate state if the summary already exists (should be safe from races now that pruning is cleaner)~~

DONE

dapplion:
I ran locally the failed test and it passes
```
cargo test -p fork_choice --release justified_checkpoint_updates_with_descendent_first_justification
```
re-running CI

dapplion:
Note: tests are failing as they are allocating ridiculous amounts of memory and getting killed by the runner or erroring directly. Need to dig into it

EDIT this loop was infinite looping pushing items into the array until OOM

https://github.com/sigp/lighthouse/blob/b49272f742322b221e13bb157ae6a00ccd276613/beacon_node/beacon_chain/src/summaries_dag.rs#L338-L345

The only cause for this too happen is some circular chain of summaries that point to themselves

dapplion:
TODO: get_ancestor_state_root should use the fork-choice to walk on old state roots.
- EDIT: Using the state summaries in disk in https://github.com/sigp/lighthouse/pull/7176

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -512,14 +512,16 @@ async fn epoch_boundary_state_attestation_processing() {
...
             .expect("block exists");
         let mut epoch_boundary_state = store
-            .load_epoch_boundary_state(&block.state_root())
+            .load_hot_state(&block.state_root())
             .expect("no error")
-            .expect("epoch boundary state exists");
+            .expect("epoch boundary state exists")
+            .0;
         let ebs_state_root = epoch_boundary_state.update_tree_hash_cache().unwrap();
         let mut ebs_of_ebs = store
```
dapplion, line 531: @michaelsproul I'm confused by this test. It appears to do:
- load state with `load_epoch_boundary_state` for root A
- compute the state root of above state, which _should_ be equal to root A
- load state with `load_epoch_boundary_state` for root A (again)

If that's correct why does it load the state twice? The test is failing because some attestations are later than finalized and `load_hot_state` returns None as the state is in the cold DB

dapplion: Ah `load_epoch_boundary_state` in unstable looks in both DBs. First hot db, then if None defaults to cold. So switching to get_state is the correct fix. I applied it in https://github.com/sigp/lighthouse/pull/7176/commits/2b050f4e8992a38af0fef116d0dde30f97241b9e

michaelsproul:
~~Could rename `diff_base_state_root` to `diff_base_state` (given that it contains a `slot`).~~

EDIT: done

michaelsproul:
Could also add some tests for `get_ancestor_state_root`. It should be straightforward to test.

beacon_node/beacon_chain/src/schema_change/migration_schema_v24.rs
```diff
@@ -0,0 +1,292 @@
...
+                        // Optimization: instead of having to load the state of each summary we load x32
+                        // less states by manually computing the HotStateSummary roots using the
+                        // computed state dag.
+                        //
+                        // No need to store diffs for states that will be reconstructed by replaying
+                        // blocks.
+                        //
+                        // 2. Convert the summary to the new format.
+                        let previous_state_root = if state_root == split.state_root {
+                            Hash256::ZERO
```
michaelsproul: I think this is unreachable (could be an error), because the `split.state_root` will never be stored as a `ReplayFrom`. The anchor slot is set to the split slot, so the split state will be stored as a snapshot.

`Error::SplitStateIsNotASnapshot`

beacon_node/beacon_chain/src/schema_change/migration_schema_v24.rs
```diff
@@ -0,0 +1,292 @@
...
+                                .previous_state_root(state_root)
+                                .map_err(|e| {
+                                    Error::MigrationError(format!(
+                                        "error computing previous_state_root {e:?}"
+                                    ))
+                                })?
+                        };
+
+                        let diff_base_state_root =
+                            if let Some(diff_base_slot) = storage_strategy.diff_base_slot() {
```
michaelsproul: Don't need to call this function because we can read the slot from the `ReplayFrom` which is matched above.

That also removes the `else` case with the zero root below.

beacon_node/beacon_chain/src/schema_change/migration_schema_v24.rs
```diff
@@ -0,0 +1,292 @@
...
+                    summaries_written = summaries_written,
+                    summaries_count = state_summaries_dag.summaries_count(),
+                    "Hot states migration in progress"
+                );
+            }
+        }
+    }
+
+    // TODO(hdiff): Should run hot DB compaction after deleting potentially a lot of states. Or should wait
+    // for the next finality event?
```
michaelsproul: Yeah I think we should do column-specific compaction after the migration completes (in `schema_change.rs`). I hope it should be reasonably fast because the deleted columns should now be empty.

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -3100,23 +3438,39 @@ impl StoreItem for HotStateSummary {
...
+            }
+        };
+        let diff_base_slot = storage_strategy.diff_base_slot();
+        let diff_base_state_root = if let Some(diff_base_slot) = diff_base_slot {
+            DiffBaseStateRoot::new(diff_base_slot, get_state_root(diff_base_slot)?)
         } else {
-            *state
-                .get_state_root(epoch_boundary_slot)
-                .map_err(HotColdDBError::HotStateSummaryError)?
+            DiffBaseStateRoot::zero()
```
michaelsproul: Maybe we should store a non-zero value here. We don't want to accidentally try loading a 0x0 state with the backwards iterator (get_ancestor_state_root).

We could even use an `Option<DiffBaseState>` for snapshots? (Need to make it serialize)

Or we could use something like:

```rust
DiffBaseStateRoot {
    slot: u64::MAX,
    state_root: Hash256::repeat_byte(0xff),
}
```

michaelsproul: This is way better than both actually:

```rust
#[derive(Encode, Decode)]
enum DiffBaseStateRoot {
   Snapshot,
   BaseState {
      slot: Slot,
      state_root: Hash256,
   }
}
```

michaelsproul:
Should address outstanding comments from the OOM PR too:

- https://github.com/sigp/lighthouse/pull/7176

michaelsproul:
I've just resolved the merge conflicts after the merge of `drop-headtracker`. It was a bit gnarly, but I think I got it right. We still have several TODOs to solve.

beacon_node/beacon_chain/src/schema_change/migration_schema_v24.rs
michaelsproul: Fixed in https://github.com/sigp/lighthouse/pull/6750/commits/088cfa03deec97ca56eb6bea4adaae35a2fa37a6

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -3577,45 +3583,72 @@ pub struct HotStateSummary {
...
     }
 
-    pub fn zero() -> Self {
-        Self {
-            slot: Slot::new(0),
-            state_root: Hash256::ZERO,
+    pub fn get_root(&self, slot: Slot) -> Result<Hash256, Error> {
+        match *self {
+            Self::Snapshot(stored_slot) => {
+                if slot == stored_slot {
```
dapplion: What's the purpose of tracking the slot in the Snapshot variant? And having a distinct error? We should never query `diff_base_state` for snapshots

michaelsproul: The ssz crate forced me to put something in there so I figured I would put the slot. You're right that it should never be used unless there's a config mismatch. It's not too bad though I don't think as the other mismatch checking code is similarly unreachable

michaelsproul: (this is because of how the enum_behaviours work — we could probably make the ssz crate more flexible but I don't want to do that rn)

I'll add a comment

dapplion: Or just do `Snapshot(u8)` and set it to zero with a comment that it's not used. Else add mental overload for no benefit

michaelsproul: I thought about that, but figured it's better to store something vaguely useful than a completely meaningless u8.

michaelsproul: I'll just change to a u8. I think you're right and it's a bit simpler

michaelsproul: Done in 54ded283112aaeb01f89efe0b3907082ca30ccab

beacon_node/beacon_chain/src/schema_change/migration_schema_v24.rs
michaelsproul: Fixed

beacon_node/store/src/hot_cold_store.rs
michaelsproul: Done with `OptionalDiffBaseState`

beacon_node/beacon_chain/src/schema_change/migration_schema_v24.rs
```diff
@@ -0,0 +1,383 @@
...
+            states_written += 1;
+        }
+
+        // Persist old summary
+        let epoch_boundary_state_slot = summary.slot - summary.slot % T::EthSpec::slots_per_epoch();
+        let old_summary = HotStateSummaryV22 {
+            slot: summary.slot,
+            latest_block_root: summary.latest_block_root,
+            epoch_boundary_state_root: state_summaries_dag
+                .ancestor_state_root_at_slot(state_root, epoch_boundary_state_slot)
```
michaelsproul, line 518: TODO: check that this handles the genesis state/slot

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -3130,19 +3467,188 @@ fn no_state_root_iter() -> Option<std::iter::Empty<Result<(Hash256, Slot), Error
...
+            .map_err(|_| StateSummaryIteratorError::OutOfBoundsInitialSlot)?
+    };
+
+    let mut previous_slot = None;
+
+    loop {
+        let state_summary = store
+            .load_hot_state_summary(&state_root)
+            .map_err(|e| StateSummaryIteratorError::LoadSummaryError(Box::new(e)))?
+            .ok_or(StateSummaryIteratorError::MissingSummary(state_root))?;
```
michaelsproul, line 3590: TODO: This could fail if state_root is pruned between when we took the split lock. We should probably hold the split lock from the start of this function (after the lookup in the state misses). This should be OK for performance because loading summaries is cheap and we're never iterating that far back.

michaelsproul:
TODO: remove `StateSummariesNotContiguous` failure case. We should try to gracefully handle this case and delete the junk state as part of the migration.

Related to:

- https://github.com/sigp/lighthouse/issues/7323

beacon_node/beacon_chain/src/summaries_dag.rs
```diff
@@ -163,34 +163,17 @@ impl StateSummariesDAG {
...
-                            // The test store_tests::finalizes_non_epoch_start_slot manages to send two
-                            // disjoint trees on its second migration.
-                            Hash256::ZERO
-                        }
+                        state_summaries_by_block_root
+                            .get(&parent_block_root)
+                            .and_then(|parent_block_summaries| {
+                                parent_block_summaries.get(&previous_slot)
+                            })
+                            .map_or(Hash256::ZERO, |(parent_state_root, _)| **parent_state_root)
```
michaelsproul: Rather than 0x0, we could use an enum here. I think we would also need this inside the `HotStateSummary`.

Another thing we could do would be to differentiate nodes that we don't expect to have a parent state (the split state) and nodes that should have one (everything else).

dapplion:
Changed the `DAGStateSummary` to hold a Result in this branch / commit https://github.com/sigp/lighthouse/commit/0ab1c7fb8f6dee5c3efb5270fec1e19e84bd190b
. What I don't like about this approach is that after the migration the `DAGStateSummary` still holds a Result but it will always be Ok

```rust
pub struct DAGStateSummary {
    ...
    pub previous_state_root: Result<Hash256, String>,
}
```

dapplion:
Idea: if migration fails because the DAG of states is broken, could we just drop all unfinalized data? Basically, drop the entire fork-choice and prune the blocks from the DB. Then we have an infallible migration. Provided we have the best possible code, it's best to require sync from a few unlucky users than to get them stuck. We have seen that the chance of the DB having a broken DAG is low and not everyone will update at the same time.

dapplion:
More thoughts:

At the migration we need to copy the summaries to the new DB column with the new format. The new format requires adding:
- `previous_state_root`
- `diff_base_state`

We don't need to copy ALL hot summaries. Only those to preserve the invariant:
- For each block there's a state_summary for each slot between finalized_slot and block_slot in the chain of ancestors of block
- For each state_summary there's a way to compute the full state of that state_root

**IDEA**: We don't need to copy all summaries. We can ignore the summaries for advanced states, and the non-descendants of finalized block. We can also over-copy and future pruning rounds will get rid of them.

#### Where to get previous_state_root from?

- From the DAG, like the current impl of the PR
- But we could also load the state at `state_root` and check its `state_roots` vector. With caching and finality not too far away we only need to load one state per head

#### Filter to reduce chance of bugs

- We can filter out state summaries not descendants of finalized. We can leverage the fork-choice to compute that list, does not seem too complicated https://github.com/sigp/lighthouse/commit/ca2388e1962b2ea60d3f22313a843285928c1542#diff-046fbdb517ca16b80e4464c2c824cf001a74a0a94ac0065e635768ac391062a8
- Filter all advanced states, like the one in https://github.com/sigp/lighthouse/issues/7323

is this a correct way of detecting advanced states?

```rust
struct Block {
  slot: Slot,
  highest_child: Option<Slot>,
}

fn is_advanced_state(state_summary: StateSummary) -> bool {
  let block: Block = get_block(state_summary.latest_block_root);
  match block.highest_child {
    Some(highest_child) => state_summary.slot > highest_child,
    None => state_summary.slot > block.slot,
  }
}
```

beacon_node/store/src/hot_cold_store.rs
michaelsproul: Fixed in:

- e1475a120b0fbb170323e67b1d19d90d3d0e1f42

beacon_node/store/src/config.rs
michaelsproul: we decided to keep this the same for hot and cold 7d8343f9e6f2db61b2c4fbc625bf1d3dd8f8d91b

beacon_node/beacon_chain/src/schema_change/migration_schema_v24.rs
```diff
@@ -0,0 +1,380 @@
...
+    let state_summaries_dag = StateSummariesDAG::new(state_summaries)
+        .map_err(|e| Error::MigrationError(format!("Error on new StateSumariesDAG {e:?}")))?;
+
+    let mut migrate_ops = vec![];
+    let mut states_written = 0;
+    let mut summaries_written = 0;
+    let mut last_log_time = Instant::now();
+
+    // TODO(tree-states): What about the anchor_slot? Is it safe to run the prior version of
+    // Lighthouse with an a higher anchor_slot than expected?
```
michaelsproul: Yeah it's fine, the `anchor_slot` wasn't used for anything except informational things prior to this.

It used to be used, to reject gossip blocks that were earlier than it, but then we removed this because it was redundant with rejecting blocks prior to the split/finalization.

beacon_node/store/src/metrics.rs
```diff
@@ -142,48 +142,6 @@ pub static BEACON_STATE_HOT_GET_COUNT: LazyLock<Result<IntCounter>> = LazyLock::
...
-        "store_beacon_state_write_total",
-        "Total number of beacon state writes the DB",
-    )
-});
-pub static BEACON_STATE_WRITE_BYTES: LazyLock<Result<IntCounter>> = LazyLock::new(|| {
-    try_create_int_counter(
-        "store_beacon_state_write_bytes_total",
-        "Total number of beacon state bytes written to the DB",
-    )
-});
```
michaelsproul: These metrics were used by the legacy encoding of the `BeaconState`. Many of them don't need replacing because we already have generic metrics with stats per column:

- DISK_DB_WRITE_BYTES
- DISK_DB_READ_BYTES
- DISK_DB_READ_COUNT
- DISK_DB_WRITE_COUNT
- DISK_DB_READ_TIMES
- DISK_DB_WRITE_TIMES

We also have hdiff-specific metrics, but could break these down by hot/cold? Or maybe that's not necessary seeing as the grid is the same, and the process should be very similar. Although I can imagine the different caching for hot and cold DBs could affect these times.

e.g.

- BEACON_HDIFF_READ_TIMES
- BEACON_HDIFF_DECODE_TIMES

wdyt @dapplion?

dapplion: I think it's worth to separate hot / cold, doesn't hurt and we should check the hypothesis 

> that's not necessary seeing as the grid is the same, and the process should be very similar

michaelsproul: Will do, I'll add this!

chong-he:
Done some testing with [c74913f](https://github.com/sigp/lighthouse/pull/6750/commits/c74913fb706a22aaa861016bd9c62953f112bf18), able to sync an archive node very smoothly. 

I found that the log: 
`
May 16 11:30:40.247 ERROR Request not found in active_inbound_requests. Response not sent  peer_id: 16Uiu2HAmAxSKQKrcYx6Jvso7jxKtVD9EFwyi8yvgoBwrkc6A5YBy, request_id: InboundRequestId { connection_id: ConnectionId(286664), substream_id: SubstreamId(4) }, response: BlocksByRange: Block slot: 430868, service: "network
`

is appearing quite frequently using the PR binary. This log looks like network related, however I don't see it when running Lighthouse v7.0.1 (see below).

I also sync using v7.0.1 and upgraded to this binary PR (during state reconstruction stage). During starting, this log appears:
`
May 19 17:41:06.526 WARN  State summaries DAG found more than one root  location: "migration", state_summaries_dag_roots: [(0xfe240dad7e625070bf09f0ff5a8e57bad041400f5fd37ce0b9b81f104a320a08, DAGStateSummary { slot: Slot(454555), latest_block_root: 0x691fbd0153224a5883d12fb5eb6cf2bd012580ae481850136908cd34014f85a4...<truncated>
`

A few minutes later (looks like related to the above):
`
May 19 17:48:01.491 WARN  Prune state summaries dag found more than one root  location: "pruning", new_finalized_state_root: 0x968c9607ea19155594daf9869494b83e8b90d2e20461678e62ec08995637d51d, split_prior_to_migration_slot: 454560, state_summaries_dag_roots_post_split: <truncated>
`

When running v7.0.1 for a few hours, I did not see the log about `ERROR Request not found in active_inbound_requests`, this error log appears 20 minutes after upgrading to this PR. 

Logs: https://drive.google.com/drive/folders/1lv2f2aPnz58M2zlbkAB1WXV7oqFoqUTd?usp=sharing

michaelsproul:
@chong-he Thanks for testing! Can you post the full untruncated logs from `State summaries DAG found more than one root`?

chong-he:
> @chong-he Thanks for testing! Can you post the full untruncated logs from `State summaries DAG found more than one root`?

First full log:
`May 19 17:41:06.526 WARN  State summaries DAG found more than one root  location: "migration", state_summaries_dag_roots: [(0xfe240dad7e625070bf09f0ff5a8e57bad041400f5fd37ce0b9b81f104a320a08, DAGStateSummary { slot: Slot(454555), latest_block_root: 0x691fbd0153224a5883d12fb5eb6cf2bd012580ae481850136908cd34014f85a4, latest_block_slot: Slot(454554), previous_state_root: 0x0000000000000000000000000000000000000000000000000000000000000000 }), (0x9e547e9783839a9fa5aa2880483fc2e40ea9cce560d1a308de00ef6b673225c3, DAGStateSummary { slot: Slot(454548), latest_block_root: 0xab2ebe57565b8f7faa4a7f8b500dbe45bbc517ab1bb5f40e7b03bab7caed8224, latest_block_slot: Slot(454547), previous_state_root: 0x0000000000000000000000000000000000000000000000000000000000000000 }), (0x7fcb96a0cd0adb78c71ff3f536bf74717d86dc845c856a69a2ddcfc7dc8f57d7, DAGStateSummary { slot: Slot(454529), latest_block_root: 0x308ed79c60899bc1b3b4408733140ac0078dc621feeea2930216811fc1cd96ff, latest_block_slot: Slot(454528), previous_state_root: 0x0000000000000000000000000000000000000000000000000000000000000000 }), (0x1a311f9b0f6c77172f0a02f67342c1434b640748f73a48d640dfd4b27b900188, DAGStateSummary { slot: Slot(454560), latest_block_root: 0xa517b202e6a10502ee0af43faf05788c587b92399d5f5c7998e645ee1f120061, latest_block_slot: Slot(454560), previous_state_root: 0x0000000000000000000000000000000000000000000000000000000000000000 })] 
`

Second full log:
`May 19 17:48:01.491 WARN  Prune state summaries dag found more than one root  location: "pruning", new_finalized_state_root: 0x968c9607ea19155594daf9869494b83e8b90d2e20461678e62ec08995637d51d, split_prior_to_migration_slot: 454560, state_summaries_dag_roots_post_split: [(0x6089a766b20f8e54b068c0e86a3b7da0a4dafe07aaf15b9c44797e0701dd21b7, DAGStateSummary { slot: Slot(454560), latest_block_root: 0x29114ec76ba1c598842626b2a2e83185a251d5543ebb9bcc343914a9c86e73da, latest_block_slot: Slot(454559), previous_state_root: 0xd103898210e1e7b70e8a78901b4c52172859b5e96d13211da269cef273c71672 }), (0x1a311f9b0f6c77172f0a02f67342c1434b640748f73a48d640dfd4b27b900188, DAGStateSummary { slot: Slot(454560), latest_block_root: 0xa517b202e6a10502ee0af43faf05788c587b92399d5f5c7998e645ee1f120061, latest_block_slot: Slot(454560), previous_state_root: 0xd103898210e1e7b70e8a78901b4c52172859b5e96d13211da269cef273c71672 })] 
`

michaelsproul:
@chong-he Sorry didn't see your full link to logs, I will take a look!

michaelsproul:
Ok so the warning logs that CK got are _expected_ on nodes that are upgraded from v7.0.x. These warnings shouldn't occur again after the first two times (once in the migration and then once in pruning). They are caused by "temporary states" not being deleted on startup, which is a concept that we're removing in this PR. Lion and I had talked about this and decided to just let the pruning code clean them up, but this more or less guarantees the warning. The alternative is to eagerly delete them in the migration before the warning is logged.

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -279,7 +288,11 @@ impl<E: EthSpec> HotColdDB<E, BeaconNodeBackend<E>, BeaconNodeBackend<E>> {
         // Load the previous split slot from the database (if any). This ensures we can
         // stop and restart correctly. This needs to occur *before* running any migrations
         // because some migrations load states and depend on the split.
-        if let Some(split) = db.load_split()? {
+        //
+        // V23: `load_split` needs to load a hot state summary, which need to be migrated from V22
+        // to V23. Attempting to `load_split` here before the migration will trigger an SSZ decode
+        // error. Instead we load the partial split, and load the full split after the migration.
+        if let Some(split) = db.load_split_partial()? {
```
```suggestion
        // V24: `load_split` needs to load a hot state summary, which need to be migrated from V22
        // to V24. Attempting to `load_split` here before the migration will trigger an SSZ decode
        // error. Instead we load the partial split, and load the full split after the migration.
        if let Some(split) = db.load_split_partial()? {
```

chong-he:
> Ok so the warning logs that CK got are _expected_ on nodes that are upgraded from v7.0.x. These warnings shouldn't occur again after the first two times (once in the migration and then once in pruning). They are caused by "temporary states" not being deleted on startup, which is a concept that we're removing in this PR. Lion and I had talked about this and decided to just let the pruning code clean them up, but this more or less guarantees the warning. The alternative is to eagerly delete them in the migration before the warning is logged.

I can confirm that the warning does not occur again after the first two times

michaelsproul:
@dapplion I found an issue with the optimised migration for archive nodes, which I've fixed in:

- https://github.com/sigp/lighthouse/pull/6750/commits/9800ab882ec796f8bbf38f0b9010d2692a5eee2f
- https://github.com/sigp/lighthouse/pull/6750/commits/b0191db4539dce266352833529e1e40a71fbd5eb

In those commits I added a FIXME which I think might not be an issue, although I want to add a test for it. In the case where we need to look up the state root of a slot prior to the split, I _think_ what we have now works OK because we jump along `diff_base_state` connections, and the only states that we _should_ be looking up with this function are ones that are diff base states of existing diffs. E.g if the split is 1024 and we need the state root of slot 512 for the next state we're storing, we will look up the slot 1024 summary, and then jump straight to the summary at 512, which is the diff base state for slot 1024. And I think this works with multiple layers.

michaelsproul:
Other TODOs I noticed while upgrading and downgrading:

- We should probably delete new-style state data on downgrade, both to save space and in case it is buggy/wrong (just in case).
- We should think about how we want hot states prior to the split to interact with the state cache. I think it might be easier to just never cache them.

michaelsproul:
Interesting, the broken test (`schema_downgrade_to_min_version`) is due to my optimisation for archive nodes as well! Upgrading an archive node leads to the split state _not_ being stored as a snapshot. So we'll have to remove that error and handle that case correctly.

beacon_node/beacon_chain/src/schema_change/migration_schema_v24.rs
```diff
@@ -0,0 +1,579 @@
...
+                        // We choose to not compute states during the epoch with block replay for
+                        // simplicity. Therefore we can only support a hot heriarchy config where the
+                        // lowest layer is >= log2(SLOTS_PER_EPOCH).
+                        // FIXME(tree-states): we need to remove this restriction.
+                        if slot % T::EthSpec::slots_per_epoch() != 0 {
+                            return Err(Error::MigrationError(format!(
+                                "Hot hierarchy config lowest value must be >= {}",
+                                T::EthSpec::slots_per_epoch().trailing_zeros()
+                            )));
+                        }
```
michaelsproul: We need to remove this restriction or it's going to screw over anyone running with diffs denser than per-epoch, which is something we describe support for in our docs:

https://lighthouse-book.sigmaprime.io/advanced_database.html#hierarchical-state-diffs

michaelsproul: Fixed in https://github.com/sigp/lighthouse/pull/6750/commits/5fb55be3ab7daa60529346425dec8f93e4fc9a78

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -3510,6 +3512,101 @@ async fn prune_historic_states() {
...
+                    %ancestor_slot,
+                    head_slot = %head_state.slot(),
+                    "Fetching ancestor state root"
+                );
+                let ancestor_state_root = store::hot_cold_store::get_ancestor_state_root(
+                    &store,
+                    &head_state,
+                    ancestor_slot,
+                )
+                .unwrap();
```
michaelsproul: Will change this to an `expect` and remove the log

beacon_node/beacon_chain/src/schema_change/migration_schema_v24.rs
```diff
@@ -0,0 +1,608 @@
...
+        for db_column in [
+            DBColumn::BeaconStateHotSummary,
+            DBColumn::BeaconStateHotDiff,
+            DBColumn::BeaconStateHotSnapshot,
+        ] {
+            migrate_ops.push(KeyValueStoreOp::DeleteKey(
+                db_column,
+                state_root.as_slice().to_vec(),
+            ));
+        }
```
michaelsproul: Could change this to just iter column and delete all of these (rather than just the state roots reachable from the DAG)

beacon_node/beacon_chain/src/schema_change/migration_schema_v24.rs
```diff
@@ -0,0 +1,608 @@
...
+                    Error::MigrationError(format!(
+                        "error computing ancestor_state_root_at_slot({state_root:?}, {epoch_boundary_state_slot}) {e:?}"
+                    ))
+                })?,
+        };
+        migrate_ops.push(KeyValueStoreOp::PutKeyValue(
+            DBColumn::BeaconStateSummary,
+            state_root.as_slice().to_vec(),
+            old_summary.as_ssz_bytes(),
+        ));
```
michaelsproul, line 529: This could use `StoreItem`

beacon_node/beacon_chain/src/schema_change/migration_schema_v24.rs
```diff
@@ -0,0 +1,608 @@
...
+                .min()
+                .ok_or(Error::MigrationError(
+                    "closest_layer_points must not be empty".to_string(),
+                ))?;
+
+        if previous_snapshot_slot >= anchor_info.state_upper_limit
+            && db
+                .hierarchy
+                .storage_strategy(split.slot, dummy_start_slot)
+                .is_ok_and(|strategy| !strategy.is_replay_from())
```
michaelsproul, line 226: Should maybe test that this doesn't affect replay_from split states during normal operation.

We should add a note about this limitation, in case we delete blocks prior to the split from the hot DB in future.

michaelsproul: I think it's ok during normal operation because `previous_state_root` is already filled in. The reason we can't migrate a ReplayFrom split state is that we can't compute the previous state root from the DAG (because the split doesn't have any ancestors in the DAG).

beacon_node/beacon_chain/src/schema_change/migration_schema_v24.rs
```diff
@@ -0,0 +1,608 @@
...
+
+                        let new_summary = HotStateSummary {
+                            slot,
+                            latest_block_root: old_summary.latest_block_root,
+                            latest_block_slot: old_summary.latest_block_slot,
+                            previous_state_root,
+                            diff_base_state,
+                        };
+                        let op = new_summary.as_kv_store_op(state_root);
+                        // It's not ncessary to immediately commit the summaries of states that are
```
michaelsproul:
```suggestion
                        // It's not necessary to immediately commit the summaries of states that are
```

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -243,12 +248,16 @@ impl<E: EthSpec> HotColdDB<E, BeaconNodeBackend<E>, BeaconNodeBackend<E>> {
...
+        debug!("Opening HotColdDB");
         config.verify::<E>()?;
 
         let hierarchy = config.hierarchy_config.to_moduli()?;
 
+        debug!(?hot_path, "Opening LevelDB");
         let hot_db = BeaconNodeBackend::open(&config, hot_path)?;
+
         let anchor_info = RwLock::new(Self::load_anchor_info(&hot_db)?);
+        debug!(?anchor_info, "Loaded anchor info");
```
michaelsproul, line 263: Could log the hierarchy config at INFO level if it is not the default

beacon_node/beacon_chain/src/schema_change/migration_schema_v24.rs
michaelsproul: Done in:

- https://github.com/sigp/lighthouse/pull/6750/commits/03fece6badbf018e1d923e4b2dce73b7031666ac

beacon_node/beacon_chain/src/schema_change/migration_schema_v24.rs
michaelsproul: On second thoughts, it doesn't matter

beacon_node/beacon_chain/src/schema_change/migration_schema_v24.rs
michaelsproul: Added a test in:

- https://github.com/sigp/lighthouse/pull/6750/commits/bb23f2a6ed82c11af27ce2d21caabf65eeae6b76

beacon_node/store/src/metrics.rs
```diff
@@ -142,66 +142,38 @@ pub static BEACON_STATE_HOT_GET_COUNT: LazyLock<Result<IntCounter>> = LazyLock::
...
+pub static BEACON_COLD_HDIFF_BUFFER_CLONE_TIMES: LazyLock<Result<Histogram>> =
+    LazyLock::new(|| {
+        try_create_histogram(
+            "store_cold_hdiff_buffer_clone_seconds",
+            "Time required to clone hierarchical diff buffer bytes",
+        )
+    });
+// This metric is not split hot/cold because it is recorded in a place where that info is not known.
```
michaelsproul, line 203: Need to fix `STORE_BEACON_HDIFF_BUFFER_COMPUTE_TIME` and the apply metric to be split between hot and freezer, and should also remove "buffer" from the name, as these metrics are about computing the diffs themselves, not the buffers.

michaelsproul: We should also add a histogram for hdiff sizes (hot and cold), at the time they are computed

michaelsproul:
Let's merge :grin:

michaelsproul (approved):

PR #6753, opened 2025-01-03
fully lint only explicitly to avoid unnecessary rebuilds

dknopik:
## Issue Addressed

When using `make lint` (updated in #6664), the full project is rebuilt as the environment is changed to include release-only tests. 

## Proposed Changes

Move this behaviour to `make lint-full` in order to avoid wasting dev time. Update CI to use `make lint-full`.


pawanjay176 (approved):
Nice

pawanjay176:
@mergify queue

PR #6755, opened 2025-01-05
Implement `getBlobSidecars` support for PeerDAS

### LINKED ISSUE #6251: Implement `getBlobSidecars` support for PeerDAS
## Description

The [`getBlobSidecars`](https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Beacon/getBlobSidecars) beacon API needs a bit of work to make it functional for PeerDAS, as full blobs will no longer be stored in every node, and will be stored in the form of data columns.

For supernodes and nodes with > 50% columns, this should be relatively straight forward, as the node would be able to reconstruct the blobs locally and serve them.

For regular full nodes, they would need to fetch at least 50% of columns, perform reconstruction before the blobs can be served.

jimmygchen:
## Issue Addressed

Closes #6251 .

## Proposed Changes

The `getBlobSidecars` endpoint now serves blob sidecars on PeerDAS-enabled networks. For this endpoint to work on PeerDAS, the server needs to be custodying more than 50% of data columns. This means the node would either need to be a supernode (`--subscribe-all-data-column-subnets`) or have more than 64 validators attached to it. ([validator custody](https://github.com/ethereum/consensus-specs/pull/3871) not implemented yet)

### DIFF
--- a/beacon_node/beacon_chain/src/kzg_utils.rs
+++ b/beacon_node/beacon_chain/src/kzg_utils.rs
@@ -276,7 +276,7 @@ pub fn reconstruct_blobs<E: EthSpec>(
                 let cell = data_column
                     .column
                     .get(row_index)
-                    .ok_or(format!("Missing data column at index {row_index}"))
+                    .ok_or(format!("Missing data column at row index {row_index}"))
                     .and_then(|cell| {
                         ssz_cell_to_crypto_cell::<E>(cell).map_err(|e| format!("{e:?}"))
                     })?;
@@ -342,7 +342,7 @@ pub fn reconstruct_data_columns<E: EthSpec>(
                 for data_column in data_columns {
                     let cell = data_column.column.get(row_index).ok_or(
                         KzgError::InconsistentArrayLength(format!(
-                            "Missing data column at index {row_index}"
+                            "Missing data column at row index {row_index}"
                         )),
                     )?;
 
--- a/beacon_node/beacon_chain/src/kzg_utils.rs
+++ b/beacon_node/beacon_chain/src/kzg_utils.rs
@@ -7,8 +7,9 @@ use std::sync::Arc;
 use types::beacon_block_body::KzgCommitments;
 use types::data_column_sidecar::{Cell, DataColumn, DataColumnSidecarError};
 use types::{
-    Blob, ChainSpec, ColumnIndex, DataColumnSidecar, DataColumnSidecarList, EthSpec, Hash256,
-    KzgCommitment, KzgProof, KzgProofs, SignedBeaconBlock, SignedBeaconBlockHeader,
+    Blob, BlobSidecar, BlobSidecarList, ChainSpec, ColumnIndex, DataColumnSidecar,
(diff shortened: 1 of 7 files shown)


beacon_node/http_api/src/block_id.rs
```diff
@@ -310,12 +324,48 @@ impl BlockId {
...
+                )
+                .collect::<Result<Vec<_>, _>>()?;
+
+            reconstruct_blobs(&chain.kzg, &data_columns, blob_indices, block).map_err(|e| {
+                warp_utils::reject::custom_server_error(format!(
+                    "Error reconstructing data columns: {e:?}"
+                ))
+            })
+        } else {
+            Err(warp_utils::reject::custom_server_error(
```
jimmygchen, line 365: Note: the status code for this hasn't been decided and spec may need to be updated.

beacon_node/beacon_chain/src/kzg_utils.rs
```diff
@@ -243,6 +244,82 @@ fn build_data_column_sidecars<E: EthSpec>(
...
+                    .ok_or(format!("Missing data column at row index {row_index}"))
+                    .and_then(|cell| {
+                        ssz_cell_to_crypto_cell::<E>(cell).map_err(|e| format!("{e:?}"))
+                    })?;
+
+                cells.push(cell);
+                cell_ids.push(data_column.index);
+            }
+
+            kzg.recover_cells_and_compute_kzg_proofs(&cell_ids, &cells)
```
dapplion: Is there a need for and_then here instead of?
```suggestion
            let (cells, _kzg_proofs) = kzg.recover_cells_and_compute_kzg_proofs(&cell_ids, &cells)
                .map_err(..)?;
```

dapplion (approved):
Looks good to me

beacon_node/beacon_chain/src/kzg_utils.rs
jimmygchen: No, I think your suggested approach is easier to read, I've applied the changes. Thanks!

consensus/types/src/blob_sidecar.rs
```diff
@@ -150,10 +150,10 @@ impl<E: EthSpec> BlobSidecar<E> {
         })
     }
 
-    pub fn new_with_existing_proof(
+    pub fn new_with_existing_proof<Payload: AbstractExecPayload<E>>(
```
pawanjay176, line 153: I didn't get how we are using this

jimmygchen: Previously this method requires full payload, but the new code i added passes a block with blinded payload, so I've made this generic to support both.

https://github.com/sigp/lighthouse/pull/6755/files#diff-8b85bc6756dc32dfe7385f21235257abddf0c3ad5f36ca66a4a4078e1209970bR307-R310

pawanjay176 (approved):

jimmygchen:
@mergify requeue

jimmygchen:
@mergify queue

PR #6756, opened 2025-01-06
Fix data columns not persisting for PeerDAS due to a `getBlobs` race condition

jimmygchen:
## Issue Addressed

This PR fixes a race condition with `getBlobs` in PeerDAS, causing data columns not to be persisted on block import.

This bug was discovered on a local devnet. The sequence of events below:
1. A new block is received on gossip. Fetch engine blobs is triggerd before block verification completes (optimisation from #6600).
2. Blobs is received from the EL, the node starts computing data columns, and passes a receiver to [`process_availability`](https://github.com/sigp/lighthouse/blob/fa6c4c02a38b998e2bae35bd768a1af241faff4a/beacon_node/beacon_chain/src/beacon_chain.rs#L3624-L3641).
3. If this happens before block is verified, the receiver [gets dropped](https://github.com/sigp/lighthouse/blob/fa6c4c02a38b998e2bae35bd768a1af241faff4a/beacon_node/beacon_chain/src/beacon_chain.rs#L3638). 
4. After the block is verified, we import the block, however at this point we've lost the receiver, and data column is not persisted. We're hitting this code branch:
https://github.com/sigp/lighthouse/blob/fa6c4c02a38b998e2bae35bd768a1af241faff4a/beacon_node/beacon_chain/src/beacon_chain.rs#L3929-L3931

`getBlobsSidecars` endpoint shows that both supernode and fullnode are missing columns

```
# fullnode
$ curl http://127.0.0.1:64385/eth/v1/beacon/blob_sidecars/148
{"code":500,"message":"INTERNAL_SERVER_ERROR: Insufficient data columns to reconstruct blobs: required 64, but only 0 were found.","stacktraces":[]}%

# fullnode
$ curl http://127.0.0.1:64505/eth/v1/beacon/blob_sidecars/148
{"code":500,"message":"INTERNAL_SERVER_ERROR: Insufficient data columns to reconstruct blobs: required 64, but only 8 were found.","stacktraces":[]}%

# supernode
$ curl http://127.0.0.1:64341/eth/v1/beacon/blob_sidecars/148
{"code":500,"message":"INTERNAL_SERVER_ERROR: Insufficient data columns to reconstruct blobs: required 64, but only 0 were found.","stacktraces":[]}%
```

## Proposed Changes

After some discussion with @michaelsproul , we think the optimisation #6600 is useful to keep, as it enables data columns to be computed and propagated to the network sooner. The solution proposed is to store the data column `Receiver` in the `DataAvailaiblityChecker` (new field in `BlockImportData`), so it's always available on block import.

## TODO

- [ ] Add some tests to cover the race condition if possible.


beacon_node/beacon_chain/src/block_verification_types.rs
```diff
@@ -355,14 +356,17 @@ impl<E: EthSpec> AvailabilityPendingExecutedBlock<E> {
...
+#[derive(Debug, Derivative)]
+#[derivative(PartialEq)]
 pub struct BlockImportData<E: EthSpec> {
     pub block_root: Hash256,
     pub state: BeaconState<E>,
     pub parent_block: SignedBeaconBlock<E, BlindedPayload<E>>,
     pub parent_eth1_finalization_data: Eth1FinalizationData,
     pub confirmed_state_roots: Vec<Hash256>,
     pub consensus_context: ConsensusContext<E>,
+    #[derivative(PartialEq = "ignore")]
```
dapplion, line 350: Can you add docs for the `data_column_recv` field? Or link to `PendingComponents` docs

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -602,7 +631,7 @@ impl<T: BeaconChainTypes> DataAvailabilityCheckerInner<T> {
 
         // Check if we have all components and entire set is consistent.
         if pending_components.is_available(self.sampling_column_count, log) {
-            write_lock.put(block_root, pending_components.clone());
+            write_lock.put(block_root, pending_components.clone_without_column_recv());
```
michaelsproul, line 662: These clones where we delete the receiver are OK because we should be using the receiver the first time it is returned inside an available block, right?

beacon_node/beacon_chain/src/block_verification_types.rs
jimmygchen: Good call, added docs thanks!

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: Yes exactly -  I added some docs to the function but I still think don't think it's immediately obvious:
https://github.com/sigp/lighthouse/blob/aabb5c101cc38618c27b62e80a74b7201b450004/beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs#L42-L44

Added some more comments above this line in https://github.com/sigp/lighthouse/pull/6756/commits/4173135299242ca420f3ba2439580d099c59f4b3:

https://github.com/sigp/lighthouse/blob/4173135299242ca420f3ba2439580d099c59f4b3/beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs#L552-L554

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -479,9 +507,17 @@ impl<T: BeaconChainTypes> DataAvailabilityCheckerInner<T> {
         // Merge in the blobs.
         pending_components.merge_blobs(fixed_blobs);
 
+        if data_column_recv.is_some() {
+            // If `data_column_recv` is `Some`, it means we have all the blobs from engine, and have
+            // started computing data columns. We store the receiver in `PendingComponents` for
+            // later use when importing the block.
+            pending_components.data_column_recv = data_column_recv;
```
dapplion, line 518: If `data_column_recv` is already Some means we have double reconstruction happening. Should we mind that case or do something?

dapplion (commented):
Looks good to me, just a minor comment. Thanks for the docs throughout

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: Yes, I chatted with Michael about this - so currently it's possible any of these three happens at the same time
- gossip block triggering EL `getBlobs` and column computation
- rpc block triggering EL `getBlobs` and column computation
- gossip/rpc column triggering column reconstruction (only one computation at a time)

I initially wanted to address these altogether, but it got a bit messy and I decided to leave this issue to a separate PR and to keep this PR small , as I wanted to get this PR merged first and start interop testing.

jimmygchen: Created issue https://github.com/sigp/lighthouse/issues/6764

dapplion (approved):
LGTM

dapplion (approved):

jimmygchen:
@mergify queue

PR #6761, opened 2025-01-07
Fix incorrect data column metric name

jimmygchen:
## Issue Addressed

Fixes incorrect data column metric name.


dapplion (approved):

jimmygchen:
@mergify queue

jimmygchen:
@mergify requeue

jimmygchen:
@mergify queue

PR #6766, opened 2025-01-08
Keep execution payload during historical backfill when prune-payloads set to false

chong-he:
## Issue Addressed

- #6510 

## Proposed Changes

- Keep execution payload during historical backfill when `--prune-payloads false` is set
- Add a field in the historical backfill debug log to indicate if execution payload is kept
- Add a test to check historical blocks has execution payload when `--prune-payloads false is set
- Very minor typo correction that I notice when working on this 

## Additional Info

- I have done manual testing that when the flag is set to false, we see that the number of keys returned:
`lighthouse db inspect --column exp --output sizes --datadir path` is around the range of: `head_slot - oldest_block_slot`, so I think it works

Thank you @michaelsproul for the hints in testing


beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -2471,6 +2471,17 @@ async fn weak_subjectivity_sync_test(slots: Vec<Slot>, checkpoint_slot: Slot) {
         .map(|s| s.beacon_block.clone())
         .collect::<Vec<_>>();
 
+    // If prune_payloads is false, the execution payload should exist in historical blocks
+    if !beacon_chain.store.get_config().prune_payloads {
```
michaelsproul: I think we should remove this `if`, and make it so that these tests always this check

We can set `prune_payloads: false` in the default config here: 

https://github.com/sigp/lighthouse/blob/beefb92b9bab6b32b9d0cf1f8b0ee6ec66f3f644/beacon_node/beacon_chain/tests/store_tests.rs#L75-L85

michaelsproul (changes requested):
Looks good, just one suggested change to the test

beacon_node/beacon_chain/tests/store_tests.rs
chong-he: Fixed in [42fb4d5](https://github.com/sigp/lighthouse/pull/6766/commits/42fb4d59db50a787fd3815c5faf6f119581ee517), thank you for your help!

Setting `prune_payloads` default to `false` in the test makes the tests that calling `check_chain_dump` to fail: 
https://github.com/sigp/lighthouse/actions/runs/13171847366/job/36763510649

From my understanding, this is because when `prune_payloads` is set to false, the execution payloads are kept for all slots in the test, not only after the split slot, so the second argument in `assert_eq!` is false. 

https://github.com/sigp/lighthouse/blob/029b4f21047c37d1ffde51c554737b1aa0880f97/beacon_node/beacon_chain/tests/store_tests.rs#L3535-L3546

I make the `assert_eq!` to become `assert!` in [5db1b78](https://github.com/sigp/lighthouse/pull/6766/commits/5db1b7858b587f08a21791e0f0024df1610671dc) to have the tests pass, however it makes the tests "less robust", so not sure if there are better ways to do this?

michaelsproul (approved):

PR #6777, opened 2025-01-09
Add libssl install to udeps task

eserilev:
I'm not really sure what happened between now and yesterday, but I do know libssl1.1 has reached EOL recently. The udeps ci task wants to access libssl1.1 for some reason, so I've added a step to install it. It's not the cleanest fix, if anyone has better ideas please let me know

Here's the original error message
`error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory`

And some informational links
https://stackoverflow.com/questions/72133316/libssl-so-1-1-cannot-open-shared-object-file-no-such-file-or-directory
https://openssl-library.org/post/2023-03-28-1.1.1-eol/

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -392,6 +392,10 @@ jobs:
           cache: false
       env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    - name: Fetch libssl1.1
+      run: wget http://nz2.archive.ubuntu.com/ubuntu/pool/main/o/openssl/libssl1.1_1.1.1f-1ubuntu2_amd64.deb
+    - name: Install libssl1.1
+      run: sudo dpkg -i libssl1.1_1.1.1f-1ubuntu2_amd64.deb
     - name: Create Cargo config dir
       run: mkdir -p .cargo
     - name: Install custom Cargo config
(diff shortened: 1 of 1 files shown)


jimmygchen (approved):
Thanks for looking into this 🙏 

I don't mind merging this to unblock CI!

I think our options are:
1. install libssl 1.1 on the runner (this PR)
2. revert to `ubuntu-22.04` - this is probably not a long term solution either
3. investigate possibility of removing ssl 1.1 dependency complelely (likely from reqwest?)

.github/workflows/test-suite.yml
```diff
@@ -392,6 +392,10 @@ jobs:
           cache: false
       env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    - name: Fetch libssl1.1
+      run: wget http://nz2.archive.ubuntu.com/ubuntu/pool/main/o/openssl/libssl1.1_1.1.1f-1ubuntu2_amd64.deb
```
michaelsproul:
```suggestion
      run: wget https://nz2.archive.ubuntu.com/ubuntu/pool/main/o/openssl/libssl1.1_1.1.1f-1ubuntu2_amd64.deb
```

michaelsproul (approved):
Agree we should just merge this to unblock CI, and then hope that `cargo-udeps` is updated so as not to require this.

.github/workflows/test-suite.yml
michaelsproul: If we're gonna install random `.deb`s from the internet, I think we should use HTTPS 😅

michaelsproul:
@mergify queue

PR #6781, opened 2025-01-10
Rust 1.84 lints

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Fix all lints from rust 1.84.

- The majority of the fixes are moving from `.map_or(false..` to `is_ok_and../is_some_and..` [unnecessay_map_or](https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_map_or)
- Removed some unneeded lifetimes from [extra_unused_lifetimes](https://rust-lang.github.io/rust-clippy/master/index.html#extra_unused_lifetimes)
- .len() on a `&str` returns the number of bytes anyway so we were using as_bytes() unnecesssarily [needless_as_bytes](https://rust-lang.github.io/rust-clippy/master/index.html#needless_as_bytes
)

### DIFF
--- a/consensus/types/src/graffiti.rs
+++ b/consensus/types/src/graffiti.rs
@@ -57,6 +57,7 @@ impl FromStr for GraffitiString {
     type Err = String;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
+        #[allow(clippy::needless_as_bytes)]
         if s.as_bytes().len() > GRAFFITI_BYTES_LEN {
             return Err(format!(
                 "Graffiti exceeds max length {}",
--- a/consensus/types/src/graffiti.rs
+++ b/consensus/types/src/graffiti.rs
@@ -57,8 +57,7 @@ impl FromStr for GraffitiString {
     type Err = String;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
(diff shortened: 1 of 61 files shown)


consensus/types/src/graffiti.rs
```diff
@@ -57,7 +57,7 @@ impl FromStr for GraffitiString {
     type Err = String;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        if s.as_bytes().len() > GRAFFITI_BYTES_LEN {
+        if s.len() > GRAFFITI_BYTES_LEN {
```
michaelsproul, line 60: This one is interesting. String length is in bytes, so makes sense:

https://doc.rust-lang.org/std/primitive.str.html#method.len

michaelsproul (approved):
Thanks Pawan!

michaelsproul:
Going to try merging this in the same batch as:

- https://github.com/sigp/lighthouse/pull/6777

michaelsproul:
@mergify queue

consensus/types/src/graffiti.rs
pawanjay176: Yeah i was confused as well, but the lint explicitly stated that they are equivalent

PR #6793, opened 2025-01-11
 Add EIP-7636 support

SirSpudlington:
# Proposed Changes
This adds the option to self-report EIP-7636 information

# Additional Info
This defaults to `true` as the reported information *should* not cause any issues.

### DIFF
--- a/beacon_node/lighthouse_network/src/discovery/enr.rs
+++ b/beacon_node/lighthouse_network/src/discovery/enr.rs
@@ -8,7 +8,7 @@ use crate::types::{Enr, EnrAttestationBitfield, EnrSyncCommitteeBitfield};
 use crate::NetworkConfig;
 use alloy_rlp::bytes::Bytes;
 use libp2p::identity::Keypair;
-use lighthouse_version::VERSION;
+use lighthouse_version::{version, VERSION};
 use slog::{debug, warn};
 use ssz::{Decode, Encode};
 use ssz_types::BitVector;
@@ -191,15 +191,14 @@ pub fn build_enr<E: EthSpec>(
 
     // Add EIP 7636 client information
     if config.send_eip_7636_info {
-        let version = VERSION.split("/").collect::<Vec<&str>>();
-        builder.client_info(
-            version.first().unwrap_or(&"Lighthouse").to_string(),
-            version
-                .get(1)
-                .unwrap_or(&env!("CARGO_PKG_VERSION"))
-                .to_string(),
-            None,
-        );
+        let name = VERSION
+            .split("/")
+            .collect::<Vec<&str>>()
+            .first()
+            .unwrap_or(&"Lighthouse")
+            .to_string();
+
+        builder.client_info(name, version().to_string(), None);
     }
 
     // Add QUIC fields to the ENR.
--- a/beacon_node/lighthouse_network/src/discovery/enr.rs
+++ b/beacon_node/lighthouse_network/src/discovery/enr.rs
@@ -8,6 +8,7 @@ use crate::types::{Enr, EnrAttestationBitfield, EnrSyncCommitteeBitfield};
 use crate::NetworkConfig;
 use alloy_rlp::bytes::Bytes;
 use libp2p::identity::Keypair;
+use lighthouse_version::VERSION;
 use slog::{debug, warn};
 use ssz::{Decode, Encode};
 use ssz_types::BitVector;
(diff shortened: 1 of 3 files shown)


michaelsproul:
I think github remembers the branch and doesn't re-run CI. Try amending your commit or pushing an empty commit with `git commit --allow-empty`

beacon_node/lighthouse_network/src/discovery/enr.rs
```diff
@@ -188,6 +189,19 @@ pub fn build_enr<E: EthSpec>(
         builder.udp6(udp6_port.get());
     }
 
+    // Add EIP 7636 client information
+    if config.send_eip_7636_info {
+        let version = VERSION.split("/").collect::<Vec<&str>>();
+        builder.client_info(
+            version.first().unwrap_or(&"Lighthouse").to_string(),
+            version
+                .get(1)
```
ackintosh: This would be `v6.0.1-8124e5c` for example, but according to the [spec](https://eips.ethereum.org/EIPS/eip-7636#specification), the `Version` string should be `6.0.1`.

beacon_node/lighthouse_network/src/discovery/enr.rs
```diff
@@ -188,6 +189,19 @@ pub fn build_enr<E: EthSpec>(
...
+                .unwrap_or(&env!("CARGO_PKG_VERSION"))
```
ackintosh: `env!("CARGO_PKG_VERSION")` results in `0.2.0` due to the version of the lighthouse_network crate, which includes `enr.rs`. However, this does not represent the correct version of Lighthouse.

ackintosh (commented):
I think adding some functions (like `version()`, `build_version()`) to the `lighthouse_version` crate and using these functions at `enr.rs` would be a good option to ensure the correct version info is obtained.

beacon_node/lighthouse_network/src/discovery/enr.rs
SirSpudlington: Resolved in recent commit.

beacon_node/lighthouse_network/src/discovery/enr.rs
SirSpudlington: Resolved in recent commit.

common/lighthouse_version/src/lib.rs
```diff
@@ -48,6 +48,25 @@ pub fn version_with_platform() -> String {
     format!("{}/{}-{}", VERSION, Target::arch(), Target::os())
 }
 
+/// Returns semantic versioning information only.
+///
+/// ## Example
+///
+/// `1.5.1`
+pub fn version() -> &'static str {
```
ackintosh, line 56: We can simply hardcode the version number as shown below:

```rust
pub fn version() -> &'static str {
    &"6.0.1"
}
```

There are hardcoded version numbers in [lighthouse_version/src/lib.rs](https://github.com/sigp/lighthouse/blob/440cb78098adbe4ecbc620aa31b6e52ee97d8f50/common/lighthouse_version/src/lib.rs#L20-L21). These are automatically updated using [change_version.sh](https://github.com/sigp/lighthouse/blob/46a06069c605842111d68b85b3e9aeea1974e202/scripts/change_version.sh#L11-L13) during the release procedure. So there is no need to worry about the maintainability of the version number here.

beacon_node/lighthouse_network/src/discovery/enr.rs
```diff
@@ -188,6 +189,18 @@ pub fn build_enr<E: EthSpec>(
...
     }
 
+    // Add EIP 7636 client information
+    if config.send_eip_7636_info {
+        let name = VERSION
+            .split("/")
+            .collect::<Vec<&str>>()
+            .first()
+            .unwrap_or(&"Lighthouse")
+            .to_string();
```
ackintosh: How about adding `client_name()` function (which simply returns &"Lighthouse") to the `lighthouse_version` crate and using it here to simplify the code?

beacon_node/lighthouse_network/src/discovery/enr.rs
```diff
@@ -188,6 +189,18 @@ pub fn build_enr<E: EthSpec>(
         builder.udp6(udp6_port.get());
     }
 
+    // Add EIP 7636 client information
+    if config.send_eip_7636_info {
```
ackintosh: We have `private` flag in the `Config` that determines whether to send client information to peers. This flag can be used instead of adding a new flag.

https://github.com/sigp/lighthouse/blob/8124e5ca5a1f6b67a5137bfa25be18285d7b80ee/beacon_node/lighthouse_network/src/config.rs#L118-L120

common/lighthouse_version/src/lib.rs
SirSpudlington: Being able to ignore maintainability is a rare circumstance that is always heavily favoured.

ackintosh (approved):
Thanks for the update. It looks good to me. 🙏

michaelsproul (approved):
Looks good, thanks!

michaelsproul:
Looks like `cargo fmt` needs to be run to fix formatting @JKincorperated 

Then we can merge!

michaelsproul:
@mergify queue

Issue #6706, opened 2024-12-16
Activate PeerDAS at Fulu Fork and remove `EIP7594_FORK_EPOCH`


jimmygchen:
## Description

The initial Fulu spec was merged in https://github.com/ethereum/consensus-specs/pull/3994, and `EIP7594_FORK_EPOCH` and `EIP7594_FORK_VERSION` no longer exists in the configuration; The spec now specifies PeerDAS to be activated at `FULU_FORK_EPOCH`.

We'll need to change PeerDAS activation to happen at the Fulu fork in Lighthouse, and remove the `eip7594` related configurations. There is a WIP PR for the Fulu fork boiler plate, and is required for this task.

jimmygchen:
#6695 merged \o/ This is unblocked.

PR #6795, opened 2025-01-13
Implement PeerDAS Fulu fork activation

jimmygchen:
## Issue Addressed

Addresses #6706 

## Proposed Changes

This PR activates PeerDAS at the Fulu fork epoch instead of `EIP_7594_FORK_EPOCH`. This means we no longer support testing PeerDAS with Deneb / Electrs, as it's now part of a hard fork.

## Additional Info

Apologies for the large diff - it was initially a relatively small change, but ended up being bigger because:
- Fulu tests were failing since we now activate PeerDAS at the Fulu fork, so I had to refactor test utils to support PeerDAS - the good news is that PeerDAS/Fulu now gets the same test coverage as the other forks \o/
- The tests revealed a range sync bug where range requests were sent to custody peers selected from the global peer list, rather than peers from the same syncing chain.
- This PR also changes Fulu to use `*_payload_v4` engine methods, as the v5 are not yet implemented, we should revert this commit once we're ready to switch over (I've add `TODO(fulu)` comments to the changes): https://github.com/sigp/lighthouse/pull/6795/commits/8cdf82ea763b1594e4adbbc2f1a047d08b931c61
- I've also introduced a `ForkName::latest_stable()` which returns `Electra`, so we don't run tests on Fulu by default.


beacon_node/lighthouse_network/src/rpc/protocol.rs
```diff
@@ -710,13 +712,10 @@ pub fn rpc_blob_limits<E: EthSpec>() -> RpcLimits {
...
 
-// TODO(peerdas): fix hardcoded max here
-pub fn rpc_data_column_limits<E: EthSpec>(fork_name: ForkName) -> RpcLimits {
+pub fn rpc_data_column_limits<E: EthSpec>(fork_name: ForkName, spec: &ChainSpec) -> RpcLimits {
     RpcLimits::new(
-        DataColumnSidecar::<E>::empty().as_ssz_bytes().len(),
-        DataColumnSidecar::<E>::max_size(
-            E::default_spec().max_blobs_per_block_by_fork(fork_name) as usize
-        ),
+        DataColumnSidecar::<E>::min_size(),
```
jimmygchen, line 644: Previously we allow for `DataColumnSidecar` with empty cell, which is invalid.

beacon_node/network/src/service.rs
```diff
@@ -789,32 +785,29 @@ impl<T: BeaconChainTypes> NetworkService<T> {
         }
     }
 
+    /// Keeping these separate from core topics because it has custom logic:
+    /// 1. Data column subscription logic depends on subscription configuration.
+    /// 2. Data column topic subscriptions will be dynamic based on validator balances due to
+    ///    validator custody.
```
jimmygchen, line 807: The downside with not including it in core fork topic is - we subscribe to PeerDAS topics as soon as Fulu is scheduled, rather than waiting until the fork.

We could potentially just duplicate the logic here:
https://github.com/sigp/lighthouse/blob/770d677a4e25df54f87789b0f6269c4be1149f96/beacon_node/lighthouse_network/src/service/mod.rs#L708

testing/ef_tests/src/cases.rs
```diff
@@ -89,26 +89,26 @@ pub use transition::TransitionTest;
 ///     to return `true` for the feature in order for the feature test vector to be tested.
 #[derive(Debug, PartialEq, Clone, Copy)]
 pub enum FeatureName {
-    Eip7594,
+    Placeholder,
```
jimmygchen: Keeping this around as it's likely going to be useful for feature prototyping in the future.

beacon_node/network/src/service.rs
dapplion: You should add a TODO for this

testing/ef_tests/check_all_files_accessed.py
```diff
@@ -47,7 +47,7 @@
     "bls12-381-tests/hash_to_G2",
     "tests/.*/eip6110",
     "tests/.*/whisk",
-    "tests/.*/eip7594",
+    "tests/.*/fulu",
```
dapplion: Why do we exclude fulu?

jimmygchen: This PR still runs alpha-8 and the tests won't pass until we merge custody group (#6736) and electra PRs (#6488)

beacon_node/network/src/service.rs
jimmygchen: Done in https://github.com/sigp/lighthouse/pull/6795/commits/4e2530221ee9ee143cc7fb448998067a378ec7d8

dapplion (approved):

dapplion:
Release tests are failing

jimmygchen:
Thanks! I've been working on the tests today. It turns out that fixing all tests for Fulu is quite a bit of work as i need to make sure we have columns (instead of blobs) for RPC block everywhere in the tests.  I'll continue on this tomorrow.

jimmygchen:
I've updated a few places in test utils to make it Fulu compatible and work with data columns, however there are a few that needs to be re-written for PeerDAS and is a deep rabbit hole. 

I've added a few `TODO(fulu)` to address them in the future, e.g. broadcast validation tests, and make them run on `ForkName::latest_stable` (Electra) instead of `ForkName::latest` (Fulu) - since we should be testing them with Electra anyway.

jimmygchen:
The failing network tests should hopefully be fixed by https://github.com/sigp/lighthouse/pull/6814

jimmygchen:
This PR is getting a lot larger than I expected (sorry 🙏 ) due to having to update tests to support PeerDAS in Fulu:
- Update tests to support processing data columns
- Found a range sync bug where range requests are sent to custody peers selected from the global peer list, rather than peers from the same syncing chain. (the main part of the fix is [here](https://github.com/sigp/lighthouse/pull/6795/files#diff-2a2c481817feb9f76f2ff94088fd03e7217ff4d108065649b179fb1aa1c38b2aR310-R327))

I'm hoping the latest commit fix all tests, and we can address the remaining issues separately, if any.

beacon_node/network/src/service.rs
jimmygchen: Existing issue for this #5894

jimmygchen:
Ignore the mergify approval :) Just testing the `trivial` label and it works!

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -3075,6 +3075,10 @@ async fn deneb_prune_blobs_happy_case() {
     let db_path = tempdir().unwrap();
     let store = get_store(&db_path);
 
+    if store.get_chain_spec().is_peer_das_scheduled() {
+        // TODO(fulu): add prune tests for Fulu / PeerDAS data columns.
+        return;
+    }
```
jimmygchen, line 3098: Added tracking issue: https://github.com/sigp/lighthouse/issues/6853

beacon_node/http_api/tests/broadcast_validation_tests.rs
```diff
@@ -7,16 +6,19 @@ use eth2::reqwest::StatusCode;
...
-    BlobSidecar, Epoch, EthSpec, FixedBytesExtended, ForkName, Hash256, MainnetEthSpec, Slot,
+    ColumnIndex, Epoch, EthSpec, FixedBytesExtended, ForkName, Hash256, MainnetEthSpec, Slot,
 };
 use warp::Rejection;
 use warp_utils::reject::CustomBadRequest;
 
 type E = MainnetEthSpec;
 
 /*
+ * TODO(fulu): write PeerDAS equivalent tests for these.
```
jimmygchen, line 20: Note that I've updated some of the tests to support data columns, but we currently run them on Electra only. The tests that needs to be written for Fulu are the scenarios covering gossip blobs.

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -1249,6 +1250,55 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+    ) -> Result<Option<BlobSidecarList<T::EthSpec>>, Error> {
+        let Some(block) = self.store.get_blinded_block(block_root)? else {
+            return Ok(None);
+        };
+
+        if self.spec.is_peer_das_enabled_for_epoch(block.epoch()) {
+            if let Some(columns) = self.store.get_data_columns(block_root)? {
+                let num_required_columns = self.spec.number_of_columns / 2;
+                let blobs_available = columns.len() >= num_required_columns as usize;
+                if blobs_available {
```
dapplion: nit but a better name would be "reconstruction_possible", or just inline the `columns.len() >= num_required_columns` in the if, I think it's easy enough to understand

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -5850,6 +5900,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
 
                 let kzg = self.kzg.as_ref();
 
+                // TODO(fulu): we no longer need blob proofs from PeerDAS and could avoid computing.
                 kzg_utils::validate_blobs::<T::EthSpec>(
```
dapplion, line 5904: Why are we validating blobs in the first place if we have compute the proofs? And what do you mean we don't need the KZG proofs anymore?

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -2286,7 +2299,12 @@ async fn weak_subjectivity_sync_test(slots: Vec<Slot>, checkpoint_slot: Slot) {
 
     let temp1 = tempdir().unwrap();
     let full_store = get_store(&temp1);
-    let harness = get_harness(full_store.clone(), LOW_VALIDATOR_COUNT);
+
+    // Run a supernode so the node has full blobs stored.
```
dapplion:
```suggestion
    // TODO(das): Run a supernode so the node has full blobs stored.
```

beacon_node/network/src/service.rs
```diff
@@ -805,32 +801,34 @@ impl<T: BeaconChainTypes> NetworkService<T> {
...
+    ///
+    /// TODO(das): The downside with not including it in core fork topic is - we subscribe to
```
dapplion, line 809: To be fixed with 
- https://github.com/sigp/lighthouse/pull/6849

beacon_node/network/src/sync/network_context.rs
```diff
@@ -358,6 +371,7 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
         batch_type: ByRangeRequestType,
         request: BlocksByRangeRequest,
         sender_id: RangeRequestId,
+        syncing_peers: impl Iterator<Item = PeerId>,
```
dapplion: This is not syncing peers, but the set of peers that claim to be able to serve `request`. Why take an iterator as arg instead of just `Vec<PeerId>` or a slice?

beacon_node/network/src/sync/network_context.rs
```diff
@@ -473,14 +487,18 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
         let mut peer_id_to_request_map = HashMap::new();
+        let syncing_peers = syncing_peers.collect::<Vec<_>>();
 
         for column_index in custody_indexes {
             // TODO(das): The peer selection logic here needs to be improved - we should probably
             // avoid retrying from failed peers, however `BatchState` currently only tracks the peer
             // serving the blocks.
-            let Some(custody_peer) = self.get_random_custodial_peer(*column_index) else {
+            let Some(custody_peer) =
+                self.choose_random_custodial_peer(*column_index, syncing_peers.iter())
```
dapplion: I think this change should not be bundled into this PR. Range sync syncing chains are not filled with peers immediately, instead, there's some delay as peers are status'd. I worry that
- Syncing is created
- In our global peer set we have enough peers on each column, but in this syncing chain there's only few peers
- Syncing chain attempts to make progress and passes the small set of peers to this function
- `Err(RpcRequestSendError::NoCustodyPeers)` condition is hit rather frequently

To do this change properly I think the syncing chains must be able to be "peer-less" for some time. Similar to lookups, but that would take some refactoring.
- WIP PR with a more complete solution https://github.com/sigp/lighthouse/pull/6864

The fact that range sync draws from the global set of peers is a known design tradeoff, unrelated to making PeerDAS Fulu compatible

dapplion (changes requested):
Fulu changes look good! I would remove the fix for selecting relevant peers in range sync and handle that in a separate PR

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Good question - the blob KZG proofs come from the EL (from the tx sender and EL validates them, not sure if this has changed), so we _should_ verify them, but I could be wrong. We also verify them again before we publish them.

Blob KZG proofs are no longer used from PeerDAS - we don't send BlobSidecars to peers anymore, and we only use the cell KZG proofs, so validating these blob proofs may not be useful.

beacon_node/network/src/sync/network_context.rs
jimmygchen: The name came from here (peers of a `SyncingChain`) - but can rename if it's misleading
https://github.com/sigp/lighthouse/blob/614f98452fb3eda2d910f956d2ef7190530be318/beacon_node/network/src/sync/range_sync/chain.rs#L178-L181

dapplion:
I have reverted the change that draws custody peers from the chain's peer set. The test `pause_and_resume_on_ee_offline` broke as a result.

To fix it I followed a different path than you did initially @jimmygchen. I see why you needed to do the change to only draw peers from the sync chain but that's not the root cause of the failing test.

Let's break down `pause_and_resume_on_ee_offline`

- Add peer A and create head chain A
- Stop engine so nothing is sent for processing
- Request to download blocks + columns on epoch 0 chain A
- Complete the requests above, all sent to peer A
- Chain A sends requests to peer A for epoch 1 chain A
- Add peer B and create finalized chain B
- Request to download blocks + columns on epoch 0 chain B - Now because we draw from the global pool there's a high chance that the custody requests will be split between peer A and peer B
- Complete the requests above, but because we filter requests to peer B the batch download will not be complete
- Enable engine, and expect batch epoch 0 chain A to be sent for processing (Ok, happens) + expect batch epoch 0 chain B to be sent for processing (failure, it's still on downloading mode)

To fix the tests I have made the `find_blocks_by_range_request` smarter with a better filter such that I can identify all column requests for a given chain even if split among different peers. We will need this improvement anyway when adding more tests. It's unrealistic for chains to have a single peer and in real networks custody by range requests are split among a lot of peers.

I also had to add an epoch filter as the extra request from chain A for epoch 1 was colliding with the complete attempt for chain B. As we add more tests we can make this filters more robust but I think it's good for now.

dapplion (approved):

jxs (commented):
Hi Jimmy, the network parts look good to me!

beacon_node/lighthouse_network/src/rpc/protocol.rs
```diff
@@ -386,7 +386,7 @@ impl SupportedProtocol {
             ProtocolId::new(Self::BlocksByRootV1, Encoding::SSZSnappy),
             ProtocolId::new(Self::PingV1, Encoding::SSZSnappy),
         ];
-        if fork_context.spec.is_peer_das_scheduled() {
+        if fork_context.fork_exists(ForkName::Fulu) {
```
pawanjay176: Maybe use `ChainSpec::is_peer_das_scheduled` here like before?

ForkContext will have `ForkName::Fulu` if 
`spec.fulu_fork_epoch.is_some()`. 
It will be Some even when fulu_fork_epoch will be `INT_MAX` then i suppose? So we'll support the rpc method when we don't have to.

beacon_node/lighthouse_network/src/rpc/protocol.rs
```diff
@@ -405,7 +405,7 @@ impl SupportedProtocol {
                 ProtocolId::new(SupportedProtocol::BlobsByRangeV1, Encoding::SSZSnappy),
             ]);
         }
-        if fork_context.spec.is_peer_das_scheduled() {
+        if fork_context.fork_exists(ForkName::Fulu) {
```
pawanjay176: same as above

beacon_node/network/src/service.rs
```diff
@@ -805,32 +801,34 @@ impl<T: BeaconChainTypes> NetworkService<T> {
...
-                    let topic =
-                        GossipTopic::new(gossip_kind, GossipEncoding::default(), fork_digest);
-                    if self.libp2p.subscribe(topic.clone()) {
-                        subscribed_topics.push(topic);
-                    } else {
-                        warn!(self.log, "Could not subscribe to topic"; "topic" => %topic);
-                    }
-                }
-            }
+        let column_subnets_to_subscribe = if self.subscribe_all_data_column_subnets {
```
pawanjay176, line 814: I really don't understand why we need the Cow here. Why can't you just collect and then take the reference in the if case

pawanjay176 (approved):
Looks good, mostly nits

beacon_node/lighthouse_network/src/rpc/protocol.rs
jimmygchen: Oh yeah that's true, nice catch

beacon_node/network/src/service.rs
jimmygchen: You're right, this is unnecessary - will remove

jimmygchen (commented):
@dapplion your change looks good to me!

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Renamed to `reconstruction_possible`

AgeManning (approved):
Pretend review, lgtm

PR #6796, opened 2025-01-13
UX Network Fixes

AgeManning:
## Issue Addressed

There were two things I came across during some recent testing, that this PR addresses. 

1 - The default port for IPv6 was set to 9090, which is confusing. I've set this to match its ipv4 counterpart (i.e 9000 and 9001). This makes more sense and is easier to firewall, for those firewalls that support both versions for a single rule. 

2 - Watching the NAT status of lighthouse, I notice we only set the field to 1 once the NAT is passed. We don't give it a default 0 (false). So we only see results when its successful. On peer disconnects, i've piggy-backed a loop of the connected peers to also watch and check for NAT status updates.


chong-he:
Maybe the docs need some update as well? I have tried to update here:
https://github.com/chong-he/lighthouse/commit/ed16a7166ffcbce948c002a0265e198388dc29e8
Feel free to use it or modify if anything is wrong

AgeManning:
Nice, good catch!

michaelsproul:
Nominated for inclusion in `v6.0.2`.

beacon_node/src/config.rs
```diff
@@ -985,6 +984,14 @@ pub fn parse_listening_addresses(
                 warn!(log, "When listening only over IPv6, use the --port flag. The value of --port6 will be ignored.");
             }
 
+            // If we are only listening on ipv6 and the user has specified --port6, lets just use
+            // that.
+            let port = if let Some(port6) = maybe_port6 {
+                port6
+            } else {
+                port
+            };
```
jxs: nit:
```suggestion
            let port = maybe_port6.unwrap_or(port);
```

beacon_node/src/config.rs
```diff
@@ -1051,6 +1058,13 @@ pub fn parse_listening_addresses(
             })
         }
         (Some(ipv4), Some(ipv6)) => {
+            // If --port6 is not set, we use --port
+            let port6 = if let Some(port6) = maybe_port6 {
+                port6
+            } else {
+                port
+            };
```
jxs: nit:
```suggestion
            let port6 = maybe_port6.unwrap_or(port);
```

jxs (approved):
LGTM Age, left two nitpicks

michaelsproul:
Re-targeting this at `unstable` due to cancellation of v6.0.2.

AgeManning:
@mergifyio queue

PR #6797, opened 2025-01-13
Dedup post merge readiness notifiers

dapplion:
## Issue Addressed

Lots of repeated code that just checks if the engine API has the required methods for the next fork. All readiness checks after the merge just check if a specific methods are in the list of capabilities.

## Proposed Changes

Delete boilerplate so future forks don't need to copypasta again.

The logic now will:
- iterate forks in descending order (latest first)
- find the first fork that is in a preparation period
- do the engine capabilities check

If a future fork requires more complex readiness checks, it can add a dedicated _readiness file like for the merge.


beacon_node/client/src/notifier.rs
```diff
@@ -63,7 +72,7 @@ pub fn spawn_notifier<T: BeaconChainTypes>(
                     );
                     eth1_logging(&beacon_chain, &log);
                     bellatrix_readiness_logging(Slot::new(0), &beacon_chain, &log).await;
-                    capella_readiness_logging(Slot::new(0), &beacon_chain, &log).await;
+                    post_capella_readiness_logging(Slot::new(0), &beacon_chain, &log).await;
```
dapplion: All forks after capella forgot to add the notifier here, resulting is the lack of readiness checks pre-genesis. Since it only affects devnets It was not noticed probably

beacon_node/client/src/notifier.rs
```diff
@@ -351,19 +357,6 @@ async fn bellatrix_readiness_logging<T: BeaconChainTypes>(
...
-        if !beacon_chain.is_time_to_prepare_for_capella(current_slot) {
-            error!(
-                log,
-                "Execution endpoint required";
-                "info" => "you need an execution engine to validate blocks, see: \
-                           https://lighthouse-book.sigmaprime.io/merge-migration.html"
-            );
-        }
-        return;
-    }
```
dapplion, line 365: No need to do this check here, we have the EL status upcheck service

beacon_node/client/src/notifier.rs
```diff
@@ -417,231 +410,139 @@ async fn bellatrix_readiness_logging<T: BeaconChainTypes>(
 }
 
 /// Provides some helpful logging to users to indicate if their node is ready for Capella
-async fn capella_readiness_logging<T: BeaconChainTypes>(
+async fn post_capella_readiness_logging<T: BeaconChainTypes>(
```
macladson: For clarity I think this should be:
```suggestion
async fn post_bellatrix_readiness_logging<T: BeaconChainTypes>(
```

macladson (changes requested):
Looks great! I'm kind of jealous I didn't think to do this first :smiling_face_with_tear: 
Just a small nit

beacon_node/client/src/notifier.rs
dapplion: Fixed with https://github.com/sigp/lighthouse/pull/6797/commits/af49dc8c0fbd65abb10474c426f52b7df3945dea

macladson (approved):
LGTM!

michaelsproul:
@dapplion or @macladson if one of you fixes the merge conflicts we can merge

dapplion:
Fixed conflicts and ready for review

michaelsproul:
sorry @dapplion bitrot got it :(

if you can fix up the Fulu conflicts lets merge

michaelsproul:
I would like to keep this, if you have some time to polish it up Lion

Issue #6790, opened 2025-01-10
Delete ineffectual RPC limits `SIGNED_BEACON_BLOCK_BELLATRIX_MAX`


michaelsproul:
## Description

We have a problem where every time we add a new fork we copy-pasta some stuff in this section of the code:

https://github.com/sigp/lighthouse/blob/0d90135047519f4c2ee586d50e560f7bb2ff9b10/beacon_node/lighthouse_network/src/rpc/protocol.rs#L76-L106

I claim that this is completely pointless because these limits are so high anyway. We should do away with the fiction that the fork changes are relevant here and just use some large constant for all forks after Bellatrix.

pawanjay176:
I'm in favour. Should have gotten rid of this way back during bellatrix when it started becoming pointless.

jimmygchen:
Completed in #6798

PR #6798, opened 2025-01-13
Remove ineffectual block RPC limits post merge

dapplion:
## Issue Addressed

After the merge the max SSZ size of a block is absurdly big. The size is actually bound by other constants, so computing the fork's actual max size is pointless.

We can get rid of more fork boilerplate.

Close https://github.com/sigp/lighthouse/issues/6790

## Proposed Changes

Use bellatrix's max value in block's rpc_block_limits_by_fork for post-merge forks

### DIFF
--- a/beacon_node/lighthouse_network/src/rpc/protocol.rs
+++ b/beacon_node/lighthouse_network/src/rpc/protocol.rs
@@ -17,12 +17,11 @@ use tokio_util::{
     compat::{Compat, FuturesAsyncReadCompatExt},
 };
 use types::{
-    BeaconBlock, BeaconBlockAltair, BeaconBlockBase, BeaconBlockCapella, BeaconBlockElectra,
-    BeaconBlockFulu, BlobSidecar, ChainSpec, DataColumnSidecar, EmptyBlock, EthSpec, EthSpecId,
-    ForkContext, ForkName, LightClientBootstrap, LightClientBootstrapAltair,
-    LightClientFinalityUpdate, LightClientFinalityUpdateAltair, LightClientOptimisticUpdate,
-    LightClientOptimisticUpdateAltair, LightClientUpdate, MainnetEthSpec, MinimalEthSpec,
-    Signature, SignedBeaconBlock,
+    BeaconBlock, BeaconBlockAltair, BeaconBlockBase, BlobSidecar, ChainSpec, DataColumnSidecar,
+    EmptyBlock, EthSpec, EthSpecId, ForkContext, ForkName, LightClientBootstrap,
+    LightClientBootstrapAltair, LightClientFinalityUpdate, LightClientFinalityUpdateAltair,
+    LightClientOptimisticUpdate, LightClientOptimisticUpdateAltair, LightClientUpdate,
+    MainnetEthSpec, MinimalEthSpec, Signature, SignedBeaconBlock,
 };
 
 // Note: Hardcoding the `EthSpec` type for `SignedBeaconBlock` as min/max values is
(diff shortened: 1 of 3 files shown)


beacon_node/lighthouse_network/src/rpc/protocol.rs
```diff
@@ -96,33 +68,6 @@ pub static SIGNED_BEACON_BLOCK_BELLATRIX_MAX: LazyLock<usize> =
     + types::ExecutionPayload::<MainnetEthSpec>::max_execution_payload_bellatrix_size() // adding max size of execution payload (~16gb)
     + ssz::BYTES_PER_LENGTH_OFFSET); // Adding the additional ssz offset for the `ExecutionPayload` field
 
-pub static SIGNED_BEACON_BLOCK_CAPELLA_MAX: LazyLock<usize> = LazyLock::new(|| {
```
jimmygchen, line 99: I think we can also remove the `FIXME` comment above

https://github.com/sigp/lighthouse/blob/2e1155490583ea654fa4ee0247fcf6556084365d/beacon_node/lighthouse_network/src/rpc/protocol.rs#L90-L93

beacon_node/lighthouse_network/src/rpc/protocol.rs
```diff
@@ -96,33 +68,6 @@ pub static SIGNED_BEACON_BLOCK_BELLATRIX_MAX: LazyLock<usize> =
...
-    *SIGNED_BEACON_BLOCK_CAPELLA_MAX_WITHOUT_PAYLOAD
-    + types::ExecutionPayload::<MainnetEthSpec>::max_execution_payload_capella_size() // adding max size of execution payload (~16gb)
```
jimmygchen, line 101: Looks like we can also remove the `max_execution_payload_capella_size` functions:

https://github.com/sigp/lighthouse/blob/ecdf2d891fb78d84e500ca8cf3a9deb9706263cf/consensus/types/src/execution_payload.rs#L131-L182

beacon_node/lighthouse_network/src/rpc/protocol.rs
```diff
@@ -55,33 +54,6 @@ pub static SIGNED_BEACON_BLOCK_ALTAIR_MAX: LazyLock<usize> = LazyLock::new(|| {
...
-        BeaconBlock::Electra(BeaconBlockElectra::full(&MainnetEthSpec::default_spec())),
-        Signature::empty(),
-    )
-    .as_ssz_bytes()
-    .len()
-});
-
-pub static SIGNED_BEACON_BLOCK_FULU_MAX_WITHOUT_PAYLOAD: LazyLock<usize> = LazyLock::new(|| {
-    SignedBeaconBlock::<MainnetEthSpec>::from_block(
-        BeaconBlock::Fulu(BeaconBlockFulu::full(&MainnetEthSpec::default_spec())),
```
jimmygchen, line 78: looks like we can also kill some of the `BeaconBlock*::full()` functions

jimmygchen (changes requested):
Yay to removing more code \o/ 
I think there are a few more things we can remove. Please see comments.

beacon_node/lighthouse_network/src/rpc/protocol.rs
dapplion: https://github.com/sigp/lighthouse/pull/6798/commits/4261ce0e49866782d0297667c41faf6c24faa9cc

beacon_node/lighthouse_network/src/rpc/protocol.rs
dapplion: https://github.com/sigp/lighthouse/pull/6798/commits/4261ce0e49866782d0297667c41faf6c24faa9cc

beacon_node/lighthouse_network/src/rpc/protocol.rs
dapplion: https://github.com/sigp/lighthouse/pull/6798/commits/4261ce0e49866782d0297667c41faf6c24faa9cc

jimmygchen (approved):
Looks great, thanks! -300 lines 🔪

jimmygchen:
@mergify queue

PR #4588, opened 2023-08-09
increase the max topic subscriptions #4581

qu0b:
## Issue Addressed

Which issue # does this PR address?

See: #4581

## Proposed Changes

Please list or describe the changes introduced by this PR.

I propose to increase the `max_subscribed_topics` to 400 from 200 to allow for nodes to be connected to all topics during a fork transition. I also propose to increase `max_subscriptions_per_request` to 200 from 160. 

Kudos to [antithesis](https://antithesis.com/).

PR #6803, opened 2025-01-14
Implement changes for EIP 7691

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Implement changes for EIP 7691. Mostly involves adding new config params and adding and whitelisting the new blob topics. 
The changes to RPC with respect to new max params are a little more involved so I'll make a separate PR for that.
This is intended to get a working version for devnet 5.

### DIFF
--- a/beacon_node/lighthouse_network/src/service/mod.rs
+++ b/beacon_node/lighthouse_network/src/service/mod.rs
@@ -16,7 +16,7 @@ use crate::rpc::{
 use crate::types::{
     attestation_sync_committee_topics, fork_core_topics, subnet_from_topic_hash, GossipEncoding,
     GossipKind, GossipTopic, SnappyTransform, Subnet, SubnetDiscovery, ALTAIR_CORE_TOPICS,
-    BASE_CORE_TOPICS, CAPELLA_CORE_TOPICS, DENEB_CORE_TOPICS, LIGHT_CLIENT_GOSSIP_TOPICS,
+    BASE_CORE_TOPICS, CAPELLA_CORE_TOPICS, LIGHT_CLIENT_GOSSIP_TOPICS,
 };
 use crate::EnrExt;
 use crate::Eth2Enr;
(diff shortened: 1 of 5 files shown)


beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -285,26 +285,23 @@ impl<E: EthSpec> Network<E> {
 
             let max_topics = ctx.chain_spec.attestation_subnet_count as usize
                 + SYNC_COMMITTEE_SUBNET_COUNT as usize
-                + ctx.chain_spec.blob_sidecar_subnet_count as usize
+                + ctx.chain_spec.blob_sidecar_subnet_count_electra as usize
```
michaelsproul, line 288: I see we already have an allowance for data column subnets irrespective of fork config, is it normal that we overestimate `max_topics` a little?

consensus/types/src/chain_spec.rs
```diff
@@ -618,14 +624,22 @@ impl ChainSpec {
...
+
     /// Return the value of `MAX_BLOBS_PER_BLOCK` appropriate for the fork at `epoch`.
     pub fn max_blobs_per_block(&self, epoch: Epoch) -> u64 {
         self.max_blobs_per_block_by_fork(self.fork_name_at_epoch(epoch))
     }
 
     /// Return the value of `MAX_BLOBS_PER_BLOCK` appropriate for `fork`.
     pub fn max_blobs_per_block_by_fork(&self, fork_name: ForkName) -> u64 {
-        if fork_name.electra_enabled() {
+        if fork_name >= ForkName::Electra {
```
michaelsproul: This is equivalent, and I think we were trying to prefer the `*_enabled()` functions, although I don't have a strong feeling either way.

michaelsproul (approved):
Looks good

beacon_node/lighthouse_network/src/service/mod.rs
michaelsproul: @jimmygchen would you mind weighing in on this if you have a sec? :pray:

jimmygchen: Yeah, I think this is ok 👍

jimmygchen: A bit of background on these max values - we've already overestimate this by 2x previously to avoid running into problems (e.g. if we forget to update these during an upgrade), as these max values are mainly used to protect us from malicious actors forcing us to allocate more memory, so it doesn't have to be super precise.

https://github.com/sigp/lighthouse/pull/4588#pullrequestreview-1623824024

jimmygchen (approved):
Looks good 👍

beacon_node/lighthouse_network/src/service/mod.rs
michaelsproul: awesome, thank you

michaelsproul:
@mergify queue

PR #6808, opened 2025-01-15
IPv6 By Default

AgeManning:
## Description

This enables ipv4 and ipv6 by default in Lighthouse.

### Motivation

IPv6 has been around for quite a few years now and most parts of the internet are supporting it. It is strange that a cutting-edge blockchain such as Ethereum doesn't support it natively. I've tested Lighthouse quite a bit and our recent upgrades should have no problem supporting IPv6. The problem is that we need a few more nodes on our DHT to also support IPv6 for some of the discovery mechanisms to function. So we need to bootstrap the network with a few IPv6 nodes and our original plan was to make lighthouse do this by default if the host system supports a globally routable IPV6 address. 

### What this PR does

If a user has specified a listening address of `0.0.0.0` and their node also has a globally routable IPv6 address (i.e their ISP and router is configured for IPv6, then we also listen on `::` and go to dual stack mode. This also adds a CLI to disable this default behaviour `--disable-ipv6`. 

### Why I no longer like this approach

- Security - If a user has opened their IPv6 firewall, and run lighthouse to listen on ipv4 `0.0.0.0` they may not expect us to secretly also open `::` and expose those ports when they have not specifically set it. I think if users are exposing ports on their computers they need to do it explicitly rather than us trying to force ipv6 on them. 
- Practicality - I think even if we do dual-stack by default, the majority of users who are behind a normal firewall, will still not see IPv6 traffic. They will need to manually intervene to allow IPv6 traffic through their firewall. If they are manually doing this, then they can just set `--ip-address ::` also to enable dual stack.

After some thought I think the best approach is just via education rather than code. We should make announcements and encourage users to support dual stack, rather than set it by default. 

I'll leave this PR around in case others feel differently. 

Also, this should not be merged until: https://github.com/sigp/discv5/pull/275


beacon_node/lighthouse_network/src/config.rs
```diff
@@ -265,6 +266,18 @@ impl Config {
         }
     }
 
+    /// A helper function to check if the local host has a globally routeable IPv6 address. If so,
+    /// returns true.
+    pub fn is_ipv6_supported() -> bool {
+        // If IPv6 is supported
+        let Ok(std::net::IpAddr::V6(local_ip)) = local_ipv6() else {
```
jxs, line 274: nit: we already depend on [`rtnetlink` ](https://docs.rs/rtnetlink/latest/rtnetlink/) via `rust-libp2p` and `if-watch`.
There's an example on how to do it [here](https://github.com/rust-netlink/rtnetlink/issues/63#issuecomment-2105034192), we could implement it here, or upsteam it as they mention in the discussion

jxs (commented):
Hi age! Left a comment

beacon_node/lighthouse_network/src/config.rs
AgeManning: Yeah, this is fair. 

It wasn't straight forward to see how to implement this, to ensure just globally routable IP. 

Happy to attempt to use this code snippet, or not import a new library, if its easy, maybe push a commit here, or if its harder we can leave for a future PR?

AgeManning:
Had a bit more of a think about this, now I'm back onboard for including this. 

I changed the logic around the "listening by default". 

Factors that changed my mind:

- We now only listen on ipv6 by default if the user doesn't specify ANY listening address. This is more inline with what I've seen in the industry. Currently Lighthouse, will default to listen on Ipv4, so now, if the host has a globally routable ipv6 address, it will also listen on ipv6 `::`.
- The arguments I made about firewalls and users being unaware that it was listening on IPv6, equally apply to listening by default on IPv4. I think now, if anything we are being consistent across the IP versions. 

Users are free to specify their listening address, this negates the need to enable/disable ipv4/6 as the user has full control. It is only in the case the user doesn't put any `--listen-address` CLI param, that we MAY now listen on IPv6.

beacon_node/lighthouse_network/src/config.rs
jxs: thanks for looking it out age! Yeah I also looked and yeah it's def not trivial, so we can leave for a future PR

AgeManning:
@mergifyio queue

michaelsproul (approved):
Stamp of approval to unblock merging. I haven't tested this, but the code change & concept seem sane.

PR #6810, opened 2025-01-15
Misc. dependency cleanup

dknopik:
## Issue Addressed

Anchor depends on some parts of Lighthouse. Due to certain dependencies between crates, it pulled in far more crates than necessary, causing longer builds.

## Proposed Changes

**[remove ensure_dir_exists](https://github.com/sigp/lighthouse/commit/a174d6780e2209bfe8c71b0a26cf16fb8364192d)**: `account_utils` and `validator_dir` used the `directory` crate for a single function: `ensure_dir_exists`. This function had questionable value, as `create_dir_all` is a no-op if the directory exists anyway. Remove `ensure_dir_exists` and the two dependencies on `directory`, which is potentially expensive to depend on due to its dependency on `clap_utils`.

**[group UNHANDLED_ERRORs into a generic](https://github.com/sigp/lighthouse/commit/6221455f91df1b62b73ab76dc620f7eb8cd5d370)**: `warp_utils` depends on `beacon_chain` to provide `beacon_chain_error`. This causes `validator_client` to depend on `beacon_chain` transitively, which is not nice.
Remove `beacon_chain_error` and similar functions and provide a generic `unhandled_error` instead, which behaves exactly the same.

**[Introduce separate `health_metrics` crate](https://github.com/sigp/lighthouse/commit/cd05981018e2b7adb4a0ce58033f41622557d4b5) by Sayan**: The `warp_utils` crate contained facilities for capturing health metrics. Anchor wants those metrics, but also wants to avoid pulling in `warp`, as we are `axum`-only, so we introduce a new crate called `health_metrics`. We use the opportunity to also move the logic for capturing this data from `eth2` to that new crate.

## Additional Info

Lighthouse also benefits from this, as the build can run more parallelized.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2571,9 +2571,7 @@ dependencies = [
  "lighthouse_network",
  "mediatype",
  "pretty_reqwest_error",
- "procfs",
  "proto_array",
- "psutil",
  "reqwest",
  "reqwest-eventsource",
  "sensitive_url",
@@ -3710,6 +3708,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "health_metrics"
+version = "0.1.0"
+dependencies = [
+ "eth2",
+ "metrics",
+ "procfs",
+ "psutil",
+]
+
+[[package]]
 name = "heck"
 version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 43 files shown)


AgeManning:
Does anchor depend on `warp_utils`? In principle we don't want anything to do with warp, but maybe there's stuff in there we needed?

AgeManning (approved):
Looks like an improvement to me

AgeManning:
@mergifyio queue

dknopik:
> Does anchor depend on `warp_utils`? In principle we don't want anything to do with warp, but maybe there's stuff in there we needed?

Not any more after the `health_metrics` crate was introduced.

PR #6813, opened 2025-01-17
fix: add libpq-dev to the packages that need to be installed as pre-requiste

jcesar-betrnk:
## Issue Addressed
This will fix the compilation error when issuing `cargo build --all`

- NA

## Proposed Changes

- Just add `libpq-dev` to the list of packages that needs to be installed before hand.

## Additional Info

- postgresql client is also used in this project

### DIFF
--- a/book/src/installation-source.md
+++ b/book/src/installation-source.md
@@ -28,7 +28,7 @@ operating system.
 Install the following packages:
 
 ```bash
-sudo apt update && sudo apt install -y git gcc g++ make cmake pkg-config llvm-dev libclang-dev clang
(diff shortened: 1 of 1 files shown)


michaelsproul:
Those docs are intended for users installing `lighthouse`, and `libpq-dev` is only required for the (experimental) binary `watch`. 

Users wanting to install `watch` will consult the docs here, which already mention `libpq`:

https://github.com/sigp/lighthouse/tree/stable/watch

Therefore I'm inclined to close this PR, as it is likely to cause users building from source to install a package they don't need.

PR #6814, opened 2025-01-17
Do not send column requests if there is no blob for the block.

jimmygchen:
## Issue Addressed

We currently send custody column requests to peers even on blocks that do not contain any blobs. This is unnecessary and is removed in this PR.

### DIFF
--- a/beacon_node/network/src/sync/block_lookups/common.rs
+++ b/beacon_node/network/src/sync/block_lookups/common.rs
@@ -177,11 +177,7 @@ impl<T: BeaconChainTypes> RequestState<T> for CustodyRequestState<T::EthSpec> {
         expected_blobs: usize,
         cx: &mut SyncNetworkContext<T>,
     ) -> Result<LookupRequestResult, LookupRequestError> {
-        if expected_blobs == 0 {
-            return Ok(LookupRequestResult::NoRequestNeeded("no columns to fetch"));
-        }
-
-        cx.custody_lookup_request(id, self.block_root)
+        cx.custody_lookup_request(id, self.block_root, expected_blobs)
             .map_err(LookupRequestError::SendFailedNetwork)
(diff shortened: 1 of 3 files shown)


beacon_node/network/src/sync/block_lookups/common.rs
```diff
@@ -174,9 +174,13 @@ impl<T: BeaconChainTypes> RequestState<T> for CustodyRequestState<T::EthSpec> {
         id: Id,
         // TODO(das): consider selecting peers that have custody but are in this set
         _peer_id: PeerId,
-        _: usize,
+        expected_blobs: usize,
         cx: &mut SyncNetworkContext<T>,
     ) -> Result<LookupRequestResult, LookupRequestError> {
+        if expected_blobs == 0 {
+            return Ok(LookupRequestResult::NoRequestNeeded("no columns to fetch"));
+        }
```
dapplion: Maybe move this into `custody_lookup_request` to group all LookupRequestResult conditions

jimmygchen: Done in https://github.com/sigp/lighthouse/pull/6814/commits/e6efd356873faed5f6b1f8f4f75597154c2cb04b

dapplion (approved):

jimmygchen:
Just found that the previous solution didn't actually work because we don't pass the blob count to `CustodyRequestState`.

I realise the we actually have `ComponentRequests::NotNeeded` but it was never set correctly and I think this is a better fix: https://github.com/sigp/lighthouse/pull/6814/commits/aad2a8eecbeee113b2bd6f6a310f2e4364c1dc9d

dapplion (approved):
Nice find! Better solution

jimmygchen:
@mergify queue

PR #6815, opened 2025-01-17
Fix custodial peer assumption on lookup custody requests

dapplion:
## Issue Addressed

Current unstable assumes that all peers to which we request columns by root, MUST have the columns in custody

https://github.com/sigp/lighthouse/blob/06329ec2d105fc60c4c7218561cff11fb6b398a3/beacon_node/network/src/sync/network_context/custody.rs#L286-L289

This statement is incorrect, as we pick a random peer from the global set of peers. The result, is a lot of downscore events like

```
VerifyError(NotEnoughResponsesReturned { actual: 0 })
```

for innocent peers.

## Proposed Changes

Provide the custody request with an up-to-date list of peers that claim to have imported the lookup block. If peers give us that signal we can expect and enforce that they MUST have the columns in custody. If we don't have any such peers, we draw from the global set, but don't enforce then to have the columns in custody.

Since we need share state between the custody request and the overall lookup request I introduced a Mutex. I don't see another way around it if we want the custody request to have the most up-to-date view of the peer set. With @michaelsproul we discussed using an immutable HashSet and clone it, but that would result in the custody request having a quickly stale view. Only when the overall custody request fails (after 3 retries) the peer set would update. 

Consider the following sequence of events, if we don't use a Mutex:
- Receive attestation for block A
- Lookup for block A is created with a single peer
- Custody request is created with a single peer
- Receive attestations for block A from 50 other peers
- (..) wait for custody request to fail 3 times (could take many seconds)
- Re-create custody request with 51 peers

During a significant duration the custody request has a stale view, rendering this feature not that useful.

### DIFF
--- a/beacon_node/network/src/sync/network_context.rs
+++ b/beacon_node/network/src/sync/network_context.rs
@@ -27,7 +27,8 @@ use lighthouse_network::service::api_types::{
     DataColumnsByRootRequester, Id, SingleLookupReqId, SyncRequestId,
 };
 use lighthouse_network::{Client, NetworkGlobals, PeerAction, PeerId, ReportSource};
-use rand::seq::SliceRandom;
+use parking_lot::RwLock;
+use rand::prelude::IteratorRandom;
 use rand::thread_rng;
 pub use requests::LookupVerifyError;
 use requests::{
@@ -308,8 +309,8 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
 
     pub fn get_random_custodial_peer(&self, column_index: ColumnIndex) -> Option<PeerId> {
         self.get_custodial_peers(column_index)
+            .into_iter()
             .choose(&mut thread_rng())
-            .cloned()
     }
 
     pub fn network_globals(&self) -> &NetworkGlobals<T::EthSpec> {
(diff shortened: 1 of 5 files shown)


beacon_node/network/src/sync/network_context.rs
```diff
@@ -562,9 +563,24 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+            .choose(&mut rand::thread_rng())
+            .copied()
+        else {
+            // Allow lookup to not have any peers and do nothing. This is an optimization to not
+            // lose progress of lookups created from a block with unknown parent before we receive
+            // attestations for said block.
+            // Lookup sync event safety: If a lookup requires peers to make progress, and does
+            // not receive any new peers for some time it will be dropped. If it receives a new
+            // peer it must attempt to make progress.
+            return Ok(LookupRequestResult::Pending("no peers"));
```
dapplion, line 581: This is equivalent to the previous implementation, I just moved the condition down the stack

jimmygchen (approved):
Nice solution, logic is pretty clean 👍

jimmygchen (approved):

jimmygchen:
@mergify queue

jimmygchen:
@mergify dequeue

jimmygchen:
@mergify queue

jimmygchen:
@mergify requeue

jimmygchen:
@mergify requeue

PR #6816, opened 2025-01-17
Avoid computing columns from EL blobs if block has already been imported

jimmygchen:
## Issue Addressed

From PeerDAS local testing I noticed some errors with failing to send computed columns to receiver:

https://github.com/sigp/lighthouse/blob/dd7591f7123dfe072631c0deb0abc1b78cc82733/beacon_node/beacon_chain/src/fetch_blobs.rs#L252

I believe there is a race condition, the sequence:
1. Fetch EL blobs triggered, received 4 blobs from the EL, triggered column computation
1. All 8 custody columns received via gossip
1. Columns written to db, block imported. (Column receiver dropped)
1. Computed columns but fail to send as receiver was dropped.

We should not even try to compute the columns if we've already imported the block from gossip. This consumes extra resources and bandwidth and is unnecessary.

### DIFF
--- a/beacon_node/beacon_chain/src/fetch_blobs.rs
+++ b/beacon_node/beacon_chain/src/fetch_blobs.rs
@@ -262,18 +262,21 @@ fn spawn_compute_and_publish_data_columns_task<T: BeaconChainTypes>(
                 }
             };
 
-            if let Err(e) = data_columns_sender.send(all_data_columns.clone()) {
-                error!(log, "Failed to send computed data columns"; "error" => ?e);
+            if data_columns_sender.send(all_data_columns.clone()).is_err() {
+                // Data column receiver have been dropped - block may have already been imported.
+                // This race condition exists because gossip columns may arrive and trigger block
+                // import during the computation. Here we just drop the computed columns.
+                debug!(
+                    log,
+                    "Failed to send computed data columns";
+                );
+                return;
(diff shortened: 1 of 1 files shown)


dapplion (commented):
We should address the error log if the receiver is dropped after import

jimmygchen:
> We should address the error log if the receiver is dropped after import

I've downgraded the error to `warn` and added some comments here in the Fulu PR:  https://github.com/sigp/lighthouse/pull/6795/files#diff-52c135e6abd229a6d90ebb9a954e3f28145ec85c311e95da86dd7a1c590e102fR251-R258

What do you propose the error handling to be? I think this PR should prevent this, and I think it's fine to just log it if it happens? If block has already been imported, then we just don't re-import and drop them .

michaelsproul:
Agree `debug` log is better than a warning for the fail to send case

jimmygchen:
Downgraded to `debug` and also just `return` without publishing as the block is likely available via gossip.
https://github.com/sigp/lighthouse/pull/6816/commits/0df238424b2ec9ee34f0f28aaa7218f3c3b2c83c

pawanjay176 (approved):
LGTM

michaelsproul (approved):

michaelsproul:
@mergify queue

Issue #6818, opened 2025-01-20
Attestation Rewards API endpoint broken on Pectra enabled networks


peterbitfly:
When executing the following API call against a lighthouse node running on `pectra-devnet-5` (`docker.ethquokkaops.io/dh/ethpandaops/lighthouse:single_attestation-b6d80eb`) we get the following error:

```
curl -X POST --data '[]' http://localhost:4000/eth/v1/beacon/rewards/attestations/826
```

```
{"code":500,"message":"INTERNAL_SERVER_ERROR: unexpected error: AttestationRewardsError","stacktraces":[]}
```

at the same time we see the following error in the ligthouse log:
```
lighthouse-1  | Jan 20 06:48:37.419 WARN Error processing HTTP API request       method: POST, path: /eth/v1/beacon/rewards/attestations/826, status: 500 Internal Server Error, elapsed: 84.008069ms
```

The requests works up until devnet-5 epoch 302. For later epoch all requests fail:
```
curl -X POST --data '[]' http://localhost:4000/eth/v1/beacon/rewards/attestations/302 --> works
curl -X POST --data '[]' http://localhost:4000/eth/v1/beacon/rewards/attestations/303 --> fails
```

I assume the endpoint fails once a validator has an effective balance > 32 ETH.

michaelsproul:
Nice catch, thanks!

Looks like you're exactly right and we need to update the maximum effective balance for which we calculate `ideal_rewards`.

michaelsproul:
I'll have a quick go at this now.

michaelsproul:
Fixed by:

- https://github.com/sigp/lighthouse/pull/6819

Part of `unstable` now and will be in the `v7.0.0-beta.0` release.

PR #6819, opened 2025-01-20
Update attestation rewards API for Electra

michaelsproul:
## Issue Addressed

Closes:

- https://github.com/sigp/lighthouse/issues/6818

## Proposed Changes

Use `MAX_EFFECTIVE_BALANCE_ELECTRA` (2048) for attestation reward calculations involving Electra.

Add a new `InteropGenesisBuilder` that tries to provide a more flexible way to build genesis states. Unfortunately due to lifetime jank, it is quite unergonomic at present. We may want to refactor this builder in future to make it easier to use.

## Additional Info

- [x] Needs a regression test.


michaelsproul:
Testing this is proving harder than I thought, but I should have something up tomorrow.

beacon_node/beacon_chain/tests/rewards.rs
```diff
@@ -36,6 +36,36 @@ fn get_harness(spec: ChainSpec) -> BeaconChainHarness<EphemeralHarnessType<E>> {
...
+
+    let harness = BeaconChainHarness::builder(E::default())
+        .spec(spec.clone())
+        .keypairs(KEYPAIRS.to_vec())
+        .with_genesis_state_builder(|builder| {
+            builder.set_initial_balance_fn(Box::new(move |i| {
+                spec.max_effective_balance_electra
+                    / (i as u64 + 1)
+                    / spec.effective_balance_increment
+                    * spec.effective_balance_increment
```
dapplion: After `i = 2048` the effective balance of the validator will be zero, feels strange?

michaelsproul: in these tests `i` is less than 2048, but yeah I'll add a minimum of 1 or something

michaelsproul: added a minimum of `min_ejection_balance`

eserilev (approved):
LGTM

dapplion (approved):

PR #6829, opened 2025-01-21
Expose blst internals

dknopik:
Add several functions to `Generic*` in `bls`, in cases where access to the underlying `blst` types is needed.

This will be needed in Anchor.


crypto/bls/src/generic_secret_key.rs
```diff
@@ -89,3 +94,20 @@ where
...
 }
+
+impl<Sig, Pub, Sec> GenericSecretKey<Sig, Pub, Sec>
+where
+    Sig: TSignature<Pub>,
+    Pub: TPublicKey,
+    Sec: TSecretKey<Sig, Pub> + Clone,
+{
+    /// Instantiates `Self` from a `point`.
+    /// Takes a reference, as moves might accidentally leave behind key material
```
michaelsproul, line 105: Can you elaborate on this? Is it because zeroize only happens on drop, and a move is implemented as a memcpy without a drop?

crypto/bls/src/generic_signature.rs
```diff
@@ -22,6 +25,16 @@ pub const INFINITY_SIGNATURE: [u8; SIGNATURE_BYTES_LEN] = [
     0,
 ];
 
+pub const INFINITY_SIGNATURE_UNCOMPRESSED: [u8; SIGNATURE_UNCOMPRESSED_BYTES_LEN] = [
```
michaelsproul, line 28: Could we add a unit test for something like `INFINITY_SIGNATURE.serialize_uncompressed() == INFINITY_SIGNATURE_UNCOMPRESSED`?

And `deserialize_uncompressed(INFINITY_SIGNATURE_UNCOMPRESSED) == INFINITY_SIGNATURE`

crypto/bls/src/generic_secret_key.rs
dknopik: Exactly.

Cloning here ensures that we have two owned `blst::min_pk::SecretKey`, which will be zeroed each on drop. Moving may or may not do a memcpy (up to rustc to inline or do other optimisations), and may cause data to be left behind. 

See also: https://docs.rs/zeroize/latest/zeroize/index.html#stackheap-zeroing-notes

crypto/bls/src/generic_signature.rs
dknopik: good idea, done!

crypto/bls/src/generic_secret_key.rs
michaelsproul: I suspect there are other places we've run afoul of this. Good to know.

I wonder how invasive it would be to use `Pin` on secret keys. Might be an exploration for another day

michaelsproul (approved):
LGTM.

michaelsproul:
@mergify dequeue

michaelsproul:
CI will fail until we merge:

- https://github.com/sigp/lighthouse/pull/7019

michaelsproul:
@mergify requeue

Issue #5244, opened 2024-02-15
Support page size > 4K in jemalloc


koraykoska:
## Description

Some ARM processors have a page size other than 4k (16k for example). lighthouse bn is unable to launch on those processors, with the error: `<jemalloc>: Unsupported system page size`

This is a know issue as can be seen here:

https://github.com/jemalloc/jemalloc/issues/2572

As explained in the above issue, jemalloc page size is a compile time option, but higher ones are ok to use on lower page sizes, the other way round doesn't work.

My suggestion would be to increase the page size to 16k during the jemalloc compilation phase on the ARM docker images. As it's officially supported to have it higher, this shouldn't break anything, just make 16k systems work.

## Version

v4.6.0

## Present Behaviour

On launch of the BN: <jemalloc>: Unsupported system page size

## Expected Behaviour

Just work normally. Page size shouldn't be an issue for normal applications.

## Steps to resolve

Right now the only way is to switch your kernel to a 4k page size. That's quite intrusive.

michaelsproul:
I think a better option might be a feature flag that compiles jemalloc with 16K pages, which we can then decide to turn on or off by default.

I'm hesitant to turn it on by default without assessing the performance impact

koraykoska:
@michaelsproul What about adding it by default to ARM builds? It's more common than 4k on ARM

michaelsproul:
@koraykoska Yeah I think that would probably be fine, but we should do some benchmarking first. Let's start with an optional feature, then we can switch it on for the ARM builds by default if it proves useful.

Btw, if you build from source _on_ the board, I think jemalloc will already autodetect the page size, won't it? So this issue is more about cross-compiling, which includes the case of building the release binary, right?

koraykoska:
@michaelsproul Yes it's about the cross-compilation and especially also the docker image.

I would be happy with a special docker tag like `arm-16k` or something.

michaelsproul:
Sounds good. I'll see if anyone from the team is interested in working on it

jtraglia:
@koraykoska @michaelsproul we ran into the same issue with Reth. This was our fix:

* https://github.com/paradigmxyz/reth/pull/7123

bcaddy:
I think I'm also having this issue. I'm trying to use Ruff on a Grace Hopper and getting this error
```
<jemalloc>: Unsupported system page size
<jemalloc>: Unsupported system page size
memory allocation of 5 bytes failed
Aborted
```

[...]

PR #6831, opened 2025-01-21
Increase jemalloc aarch64 page size limit (#5244)

janickm:
## Issue Addressed

#5244

## Proposed Changes

Pass `JEMALLOC_SYS_WITH_LG_PAGE=16` env to aarch64 cross-compilation to support systems with up to 64-KiB page sizes. This is backwards-compatible for the current (most usual) 4-KiB systems.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,17 +63,17 @@ install-lcli:
 build-x86_64:
 	cross build --bin lighthouse --target x86_64-unknown-linux-gnu --features "portable,$(CROSS_FEATURES)" --profile "$(CROSS_PROFILE)" --locked
 build-aarch64:
-    # JEMALLOC_SYS_WITH_LG_PAGE=16 tells jemalloc to support up to 64-KiB
-    # pages, which are commonly used by aarch64 systems.
-    # See: https://github.com/sigp/lighthouse/issues/5244
+	# JEMALLOC_SYS_WITH_LG_PAGE=16 tells jemalloc to support up to 64-KiB
+	# pages, which are commonly used by aarch64 systems.
+	# See: https://github.com/sigp/lighthouse/issues/5244
 	JEMALLOC_SYS_WITH_LG_PAGE=16 cross build --bin lighthouse --target aarch64-unknown-linux-gnu --features "portable,$(CROSS_FEATURES)" --profile "$(CROSS_PROFILE)" --locked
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -63,7 +63,10 @@ install-lcli:
 build-x86_64:
 	cross build --bin lighthouse --target x86_64-unknown-linux-gnu --features "portable,$(CROSS_FEATURES)" --profile "$(CROSS_PROFILE)" --locked
 build-aarch64:
-	cross build --bin lighthouse --target aarch64-unknown-linux-gnu --features "portable,$(CROSS_FEATURES)" --profile "$(CROSS_PROFILE)" --locked
+    # JEMALLOC_SYS_WITH_LG_PAGE=16 tells jemalloc to support up to 64-KiB
+    # pages, which are commonly used by aarch64 systems.
+    # See: https://github.com/sigp/lighthouse/issues/5244
+	JEMALLOC_SYS_WITH_LG_PAGE=16 cross build --bin lighthouse --target aarch64-unknown-linux-gnu --features "portable,$(CROSS_FEATURES)" --profile "$(CROSS_PROFILE)" --locked
```
michaelsproul, line 69: Can you please add the same to `build-lcli-aarch64`?

ph03: Sure, added :+1: (I thought only the "main" lighthouse target would be using the jemalloc feature, but I might be wrong)

janickm: Sure, added 👍 (I thought only the "main" lighthouse target would be using the jemalloc feature, but I might be wrong)

michaelsproul (approved):
Looks good.

I added the page size info to `lighthouse --version` so we can confirm when the setting is taking effect.

I had issues with Cargo not detecting the setting of the env var, like: 

1. Build without env var
2. Build with env var (no-op, but should rebuild)

I think this is maybe just something to be aware of. A `cargo clean` between 1 and 2 will sort it out.

michaelsproul:
Added the `backwards-incompat` label too so that we remember to mention this in the release notes.

michaelsproul:
Some benchmarks on my Mac showed an insignificant (~1%) difference between 16K pages (default on macOS) and 64K pages, when it came to CPU performance of `lcli skip-slots`.

PR #6597, opened 2024-11-20
add new features to mergify config

jxs:
## Issue Addressed

While addressing an issue on the rust-libp2p `mergify.yml` config file, finally had time to read its doc and suggest some nice things it has there to the lighthouse repo:

- automatically notify PR authors when their PR has conflicts with the target PR branch
- add the ability to PR's to the mergify queue by adding the `add-to-queue` label instead of having to reply to the PR with "@mergifyio queue"
- add the ability to merge trivial PRs without other approvals by adding the `trivial` label

## Notes
- the `lighthouse` team doesn't exist yet, we can create it or use the `sigp` team which has all its members
- took the opportunity to also change the merged commit message by instead of having the commits having the PR description sections, tell me wdyt @jimmygchen, @michaelsproul

PR #6841, opened 2025-01-22
Merge `stable` into `unstable`

michaelsproul:
## Issue Addressed

Update `unstable` for the mergify config change merged to `stable`:

- https://github.com/sigp/lighthouse/pull/6597

## Additional Info

We should merge this with a merge commit rather than a squash.


michaelsproul:
Actually I don't think we need this yet. We can just back-merge after releasing v6.0.2

michaelsproul:
Indecisive, sorry :sweat_smile:

PR #6847, opened 2025-01-23
Fork aware max values in rpc

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

In https://github.com/sigp/lighthouse/pull/6329 we changed `max_blobs_per_block` from a preset to a config value. 
We weren't using the right value based on fork in that PR. This is a follow up PR to use the fork dependent values.

In the proces, I also updated other places where we weren't using fork dependent values from the ChainSpec.

Note to reviewer: easier to go through by commit

### DIFF
--- a/beacon_node/lighthouse_network/src/rpc/codec.rs
+++ b/beacon_node/lighthouse_network/src/rpc/codec.rs
@@ -572,37 +572,30 @@ fn handle_rpc_request<E: EthSpec>(
         SupportedProtocol::BlocksByRangeV1 => Ok(Some(RequestType::BlocksByRange(
             OldBlocksByRangeRequest::V1(OldBlocksByRangeRequestV1::from_ssz_bytes(decoded_buffer)?),
         ))),
-        SupportedProtocol::BlocksByRootV2 => {
-            let max_request_blocks = spec.max_request_blocks(current_fork);
-            Ok(Some(RequestType::BlocksByRoot(BlocksByRootRequest::V2(
-                BlocksByRootRequestV2 {
-                    block_roots: RuntimeVariableList::from_ssz_bytes(
-                        decoded_buffer,
-                        max_request_blocks,
-                    )?,
-                },
-            ))))
-        }
-        SupportedProtocol::BlocksByRootV1 => {
-            let max_request_blocks = spec.max_request_blocks(current_fork);
-            Ok(Some(RequestType::BlocksByRoot(BlocksByRootRequest::V1(
-                BlocksByRootRequestV1 {
-                    block_roots: RuntimeVariableList::from_ssz_bytes(
-                        decoded_buffer,
-                        max_request_blocks,
-                    )?,
-                },
-            ))))
-        }
+        SupportedProtocol::BlocksByRootV2 => Ok(Some(RequestType::BlocksByRoot(
+            BlocksByRootRequest::V2(BlocksByRootRequestV2 {
+                block_roots: RuntimeVariableList::from_ssz_bytes(
+                    decoded_buffer,
+                    spec.max_request_blocks(current_fork),
+                )?,
+            }),
+        ))),
+        SupportedProtocol::BlocksByRootV1 => Ok(Some(RequestType::BlocksByRoot(
+            BlocksByRootRequest::V1(BlocksByRootRequestV1 {
+                block_roots: RuntimeVariableList::from_ssz_bytes(
+                    decoded_buffer,
+                    spec.max_request_blocks(current_fork),
+                )?,
+            }),
+        ))),
         SupportedProtocol::BlobsByRangeV1 => Ok(Some(RequestType::BlobsByRange(
             BlobsByRangeRequest::from_ssz_bytes(decoded_buffer)?,
         ))),
         SupportedProtocol::BlobsByRootV1 => {
-            let max_request_blob_sidecars = spec.max_request_blob_sidecars(current_fork);
             Ok(Some(RequestType::BlobsByRoot(BlobsByRootRequest {
                 blob_ids: RuntimeVariableList::from_ssz_bytes(
                     decoded_buffer,
-                    max_request_blob_sidecars,
+                    spec.max_request_blob_sidecars(current_fork),
                 )?,
             })))
         }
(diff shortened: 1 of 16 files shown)


beacon_node/network/src/sync/network_context.rs
```diff
@@ -455,7 +459,6 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
                 (None, None)
             };
 
-        // TODO(pawan): this would break if a batch contains multiple epochs
```
pawanjay176, line 458: Even if a batch contains multiple epochs (which it doesn't), this is a very niche case that would only break during fork boundaries.

beacon_node/lighthouse_network/src/rpc/codec.rs
```diff
@@ -572,44 +572,37 @@ fn handle_rpc_request<E: EthSpec>(
...
+            let max_request_blocks = spec.max_request_blocks(current_fork);
+            Ok(Some(RequestType::BlocksByRoot(BlocksByRootRequest::V1(
+                BlocksByRootRequestV1 {
+                    block_roots: RuntimeVariableList::from_ssz_bytes(
+                        decoded_buffer,
+                        max_request_blocks,
+                    )?,
+                },
+            ))))
+        }
```
jxs: nit:
```suggestion
        SupportedProtocol::BlocksByRootV1 => Ok(Some(RequestType::BlocksByRoot(
            BlocksByRootRequest::V1(BlocksByRootRequestV1 {
                block_roots: RuntimeVariableList::from_ssz_bytes(
                    decoded_buffer,
                    spec.max_request_blocks(current_fork),
                )?,
            }),
        ))),
```
we can just inline the `spec.max_request_blocks(current_fork)` wdyt Pawan?

beacon_node/lighthouse_network/src/rpc/codec.rs
```diff
@@ -572,44 +572,37 @@ fn handle_rpc_request<E: EthSpec>(
...
+        SupportedProtocol::BlobsByRangeV1 => Ok(Some(RequestType::BlobsByRange(
+            BlobsByRangeRequest::from_ssz_bytes(decoded_buffer)?,
+        ))),
         SupportedProtocol::BlobsByRootV1 => {
+            let max_request_blob_sidecars = spec.max_request_blob_sidecars(current_fork);
```
jxs: nit:
same we can inline the call below

beacon_node/lighthouse_network/src/rpc/codec.rs
```diff
@@ -1097,21 +1090,24 @@ mod tests {
         }
     }
 
-    fn bbroot_request_v1(spec: &ChainSpec) -> BlocksByRootRequest {
-        BlocksByRootRequest::new_v1(vec![Hash256::zero()], spec)
+    fn bbroot_request_v1(fork_name: ForkName) -> BlocksByRootRequest {
+        let fork_context = fork_context(fork_name);
```
jxs: nit:
same I'd inline all the `fork_context()` calls for cleaner diff and less lines of code

beacon_node/network/src/sync/manager.rs
```diff
@@ -258,10 +260,11 @@ pub fn spawn<T: BeaconChainTypes>(
     network_send: mpsc::UnboundedSender<NetworkMessage<T::EthSpec>>,
     beacon_processor: Arc<NetworkBeaconProcessor<T>>,
     sync_recv: mpsc::UnboundedReceiver<SyncMessage<T::EthSpec>>,
+    fork_context: Arc<ForkContext>,
```
jxs, line 263: nit: we can just pass the `ForkName` which is `Copy` and avoid clonning the `Arc` when passing it

jxs (approved):
LGTM Pawan, left some suggestions :)

beacon_node/network/src/sync/manager.rs
pawanjay176: `current_fork` can change during runtime so we may use the outdated value if we do that . its just an Arc clone so it's okay imo

jxs: makes sense, thanks for explaining Pawan!

beacon_node/lighthouse_network/src/rpc/handler.rs
```diff
@@ -855,6 +855,45 @@ where
...
+                    }));
+                    return self.shutdown(None);
+                }
+            }
+            RequestType::BlobsByRange(request) => {
+                let max_requested_blobs = request
+                    .count
+                    .saturating_mul(spec.max_blobs_per_block_by_fork(current_fork));
+                let max_allowed = spec.max_request_blob_sidecars(current_fork) as u64;
+                if max_requested_blobs > max_allowed {
```
michaelsproul, line 882: Doing this check during decoding was nice because it protected the rate-limiter, which is currently vulnerable to overflow (not a big issue, because peer penalties kick in shortly after). I'll open a follow-up PR to fix that.

beacon_node/lighthouse_network/src/rpc/handler.rs
```diff
@@ -855,6 +855,45 @@ where
         }
 
         let (req, substream) = substream;
+        let current_fork = self.fork_context.current_fork();
+        let spec = &self.fork_context.spec;
+
+        match &req {
+            RequestType::BlocksByRange(request) => {
+                let max_allowed = spec.max_request_blocks(current_fork) as u64;
```
michaelsproul, line 863: Have we ever thought about making the limit dependent on the protocol version or the active fork at the time of the `request.slot` range? It seems like we could inadvertently reject requests from clients that use different limits for pre-Deneb blocks.

consensus/types/src/chain_spec.rs
```diff
@@ -633,6 +644,17 @@ impl ChainSpec {
...
     }
 
+    /// Returns the highest possible value for max_request_blobs based on enabled forks.
+    ///
+    /// This is useful for upper bounds in testing.
+    pub fn max_request_blobs_upper_bound(&self) -> usize {
+        if self.electra_fork_epoch.is_some() {
+            self.max_request_blob_sidecars_electra as usize
+        } else {
+            self.max_request_blocks as usize
```
michaelsproul: Typo:

```suggestion
            self.max_request_blob_sidecars as usize
```

michaelsproul (dismissed):

beacon_node/lighthouse_network/src/rpc/handler.rs
jxs: Hi Michael, can you elaborate on what you mean by `request.slot` range? 
If it's pre Deneb `max_request_blocks` will return the value for [pre Deneb](https://github.com/sigp/lighthouse/blob/c228e57caaab5d63aeacf89cda5cf64ebe6975ce/consensus/types/src/chain_spec.rs#L624) right?

beacon_node/lighthouse_network/src/rpc/handler.rs
jxs: yeah It was a suggestion from me to not do it during `protocol.rs` and the `Encoder` and `Decoder` as decode should only care about encoding and decoding not protocol logic, 
thanks for opening the issue Michael!

pawanjay176: This check in the handler is still before hitting the rate limiter though. The rate limiter check happens in the rpc behaviour on a `HandlerEvent::Ok`. We don't send a `HandlerEvent::Ok` if the check above doesn't pass

beacon_node/lighthouse_network/src/rpc/handler.rs
michaelsproul: We are always using the current fork according to wall clock time, so if the request is for pre-Deneb data but the current fork is post-Deneb, we will use the Deneb value.

pawanjay176: I think we don't do it that way to keep things consistent between byroot and byrange requests (since we cannot know the slot when we get root requests).
I asked other CL teams how they handle it and it seems like prysm and lodestar do by slot, but nimbus does it by current fork.
In practice, I don't think this matters that much because all clients request 1/2 epochs per by_range request which doesn't hit the limits.

jxs (approved):

PR #6834, opened 2025-01-22
Cargo update - Mainly for discv5 update

AgeManning:
There has been a minor update to discv5. 

This PR does a cargo update to update all of lighthouse dependencies, including discv5. 

Note that this PR updates lighthouse MSRV to 1.82 as `netlink-proto` and `alloy-*` crates use later features.

PR #6608, opened 2024-11-25
Pin rust_eth_kzg to 0.5.1

michaelsproul:
## Proposed Changes

Temporarily pin `rust_eth_kzg` to 0.5.1 as 0.5.2 seems to have regressed some loading performance enough for our CI to start timing out in the simulator tests (32s timeout). See:

- https://github.com/sigp/lighthouse/pull/6607

PR #6848, opened 2025-01-23
Cargo update without `rust_eth_kzg`

michaelsproul:
## Proposed Changes

Update cargo dependencies while keeping `rust_eth_kzg` pinned to `0.5.1` due to the regression described in:

- https://github.com/sigp/lighthouse/pull/6608

The changes from that PR were not sufficient to actually pin the dependencies of `rust_eth_kzg`, because the dependencies from the workspace Cargo.toml file were not being used anywhere. To fix this, I've added them as explicit dependencies in `crypto/kzg/Cargo.toml`. With this change, `cargo update` no longer tries to update them.

## Additional Info

Replaces:

- https://github.com/sigp/lighthouse/pull/6834

Hopefully `cargo udeps` doesn't fire due to these deps not being used directly :pray: . From memory I think this case should be fine.


michaelsproul:
Ok, this will need the MSRV bump. This is a bit naughty, because the alloy deps shouldn't be bumping MSRV on patch releases.

michaelsproul:
MSRV bump is fine, but I've asked about Alloy's policy here just so we know how to handle this going forward:

- https://github.com/alloy-rs/core/pull/790

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM rust:1.80.0-bullseye AS builder
+FROM rust:1.84.0-bullseye AS builder
```
jimmygchen, line 1: I think we can also bump the version in `lcli` too?

https://github.com/sigp/lighthouse/blob/96b00ef66c82eb7c67e76bf6142922e43f71f8b5/lcli/Dockerfile#L4

jimmygchen (commented):
Looks good to me! Just added a minor comment to update `lcli` Dockerfile as well.

Marking this as `backwards-incompat` due to MSRV update.

Dockerfile
michaelsproul: Good catch! I thought of this and then forgot to do it 🤦‍♂️

michaelsproul:
Updated!

jimmygchen (approved):
LGTM

Issue #6439, opened 2024-09-27
Same block getting imported multiple times due to race condition


jimmygchen:
## Description

**TL;DR:** This is actually an existing race condition and can be triggered today. This can happen when blobs arrived from RPC and gossip around the same time and both triggering a block import. There's no impact other than extra work performed:  in the 2nd import, the state cache is expected to miss, as the previous import `pop`'d the state, so this will result in state reconstruction, which takes ~80ms.

### PeerDAS occurence

On `peer-das-devnet-2`, I'm seeing the same block getting written to the database multiple times (4 times within 2 ms!). This occurred on a supernode.

I think this is because we keep the `PendingComponent`s in the cache until we complete block import, so for each gossip component that made it through to the DA checker after the block is available, it would trigger another import:

https://github.com/sigp/lighthouse/blob/07afa6c2a57e5eff515482a62e064ff975e36966/beacon_node/beacon_chain/src/beacon_chain.rs#L3635-L3646

With PeerDAS supernodes, once we receive 64 columns, the remaining columns can come from reconstruction and block can be made available after this. However, we can still get columns from gossip (as they haven't been seen via gossip), and each of these gossip columns could trigger a block import, as the `PendingComponent` in the cache is "complete".

The change to keep `PendingComponent` in the DA cache was intentional and was made in the following PR to address an issue with sync lookup:
https://github.com/sigp/lighthouse/pull/5845

I'm not sure if there's a better way, but if we do need to keep the block in the DA Checker during import, perhaps we can do a check before trying to process a gossip block/blob/data_column  - if block has already been made available, just `return` instead of processing and re-importing.

jimmygchen:
Logs from `Sep 25` on lighthouse-geth-3 on `peerdas-devnet-2`:
- `05:29:05.122`: received block `0x647586a23cd593120c0277581cbf829294821abdd6f8306564b78351be4961d4`
- `05:29:05.929`: received 67 data columns
- `05:29:06.476`: `writing data_columns to store` (block import), data column reconstruction must have occurred for this to happen (on the 64th data column recieved)
- `05:29:06.482`: another `writing data_columns to store` - this should be from the 65th gossip data columns
- `05:29:06.489`: `Gossipsub data column processed, imported fully available block` - this proved the above write is triggered from gossip column
- `05:29:06.489`: `Sending pubsub messages..` this indicates reconstruction is complete and publish a bunch of reconstructed data columns
[...]

PR #6850, opened 2025-01-23
Generalize block available time metric for PeerDAS

dapplion:
## Issue Addressed

Currently `BlockDelays` metrics track `all_blobs_observed` which is has the "intention" to track the time at which the last blob for a block was received. This metric won't register for PeerDAS.

## Proposed Changes

To generalize this for PeerDAS
- Register the last blob or data column seen on `all_blobs_observed`

Should I update the name of the metric?

### DIFF
--- a/beacon_node/beacon_chain/src/data_availability_checker.rs
+++ b/beacon_node/beacon_chain/src/data_availability_checker.rs
@@ -5,6 +5,7 @@ use crate::block_verification_types::{
 use crate::data_availability_checker::overflow_lru_cache::{
     DataAvailabilityCheckerInner, ReconstructColumnsDecision,
 };
+use crate::validator_monitor::timestamp_now;
 use crate::{metrics, BeaconChain, BeaconChainTypes, BeaconStore};
 use kzg::Kzg;
 use slog::{debug, error, Logger};
@@ -230,6 +231,11 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
         block_root: Hash256,
         custody_columns: DataColumnSidecarList<T::EthSpec>,
     ) -> Result<Availability<T::EthSpec>, AvailabilityCheckError> {
+        let seen_timestamp = self
+            .slot_clock
+            .now_duration()
+            .ok_or(AvailabilityCheckError::SlotClockError)?;
+
         // TODO(das): report which column is invalid for proper peer scoring
         // TODO(das): batch KZG verification here, but fallback into checking each column
         // individually to report which column(s) are invalid.
@@ -238,7 +244,7 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
             .map(|column| {
                 let index = column.index;
                 Ok(KzgVerifiedCustodyDataColumn::from_asserted_custody(
-                    KzgVerifiedDataColumn::new(column, &self.kzg)
+                    KzgVerifiedDataColumn::new(column, &self.kzg, seen_timestamp)
                         .map_err(|e| AvailabilityCheckError::InvalidColumn(index, e))?,
                 ))
             })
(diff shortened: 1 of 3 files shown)


beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -565,6 +571,8 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
             &self.kzg,
             &pending_components.verified_data_columns,
             &self.spec,
+            // TODO(das): what timestamp to use when reconstructing columns?
+            timestamp_now(),
```
jimmygchen, line 575: On one hand I think would be a bit more consistent if we put it _after_ reconstruction, and right before we add the columns to `put_kzg_verified_data_columns`, similar to elsewhere; On the other hand, we can argue that we have these columns once we see the last column that triggered reconstruction. I don't see an issue going either way.

One scenario i can think of is -
1. receive column 64, and trigger reconstruction, last_seen timestamp is `now`
2. receive column 65 from gossip, last seen timestamp is `now + t`
3. block is available when reconstruction completes, and last seen timestamp is actually `now + t`, not the reconstructed columns that made the block available. 

Another scenario
1. receive column 64, and trigger reconstruction, last_seen timestamp is `now`
2. receive column 65-127 from gossip, last seen timestamp is `now + t`
3. block is made available before reconstruction completed
4. reconstruction completed, but we ignore the columns because they're already in the cache.
5. Note that we trigger `make_available` again here (bug), but it doesnt change the last seem timestamp, still `now + t`

(bug issue: https://github.com/sigp/lighthouse/issues/6439)

jimmygchen: Not suggesting a change here - but if you think it makes sense to use the timestamp before reconstruction (i.e. the last data column that triggered reconstruction to make this block available), we can remove this TODO

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -233,11 +234,17 @@ impl<T: BeaconChainTypes, O: ObservationStrategy> GossipVerifiedDataColumn<T, O>
 #[ssz(struct_behaviour = "transparent")]
 pub struct KzgVerifiedDataColumn<E: EthSpec> {
     data: Arc<DataColumnSidecar<E>>,
+    #[ssz(skip_serializing, skip_deserializing)]
```
jimmygchen, line 237: Do we still need `Encode` and `Decode` here? might be able to remove it now we don't store overflow cache anymore.

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -293,29 +300,38 @@ impl<E: EthSpec> CustodyDataColumn<E> {
 #[ssz(struct_behaviour = "transparent")]
 pub struct KzgVerifiedCustodyDataColumn<E: EthSpec> {
     data: Arc<DataColumnSidecar<E>>,
+    #[ssz(skip_serializing, skip_deserializing)]
```
jimmygchen, line 303: sam as above

jimmygchen (changes requested):
I've added a few minor comments, overall looks good to me!
I think it's ok to keep the same metric name, it's still means the same thing - the last timestamp we get all the required blobs to make block available (regardless of whether the transmit unit is column sidecar or blob sidecar).

dapplion:
@jimmygchen another solution for this metric is to track the timestamp for when `is_available` returns true for the first time. What do you think?

https://github.com/sigp/lighthouse/pull/6866 makes the current mechanism a bit awkward

jimmygchen:
> another solution for this metric is to track the timestamp for when `is_available` returns true for the first time. What do you think?

I think the downside of this is it no longer accurately tracks blob available time, but rather the block available time, which we already have metric for.

If we receive all columns before the block, then we end up tracking the block receive time instead of   `blobs_available_timestamp`.

jimmygchen:
@dapplion do you still want to continue working on this PR? should we close this?

dapplion:
- Moving to an issue https://github.com/sigp/lighthouse/issues/7477

PR #6851, opened 2025-01-23
Use data column batch verification consistently

dapplion:
## Issue Addressed

Resolve a `TODO(das)` to use KZG batch verification in `put_rpc_custody_columns`

## Proposed Changes

Uses `verify_kzg_for_data_column_list_with_scoring` in all paths that send more than one column. To use batch verification and have attributability of which peer is sending a bad column.

Needs to move `verify_kzg_for_data_column_list_with_scoring` into the type's module to convert to the KZG verified type.

### DIFF
--- a/beacon_node/beacon_chain/src/data_column_verification.rs
+++ b/beacon_node/beacon_chain/src/data_column_verification.rs
@@ -243,7 +243,7 @@ impl<E: EthSpec> KzgVerifiedDataColumn<E> {
     pub fn from_batch(
         data_columns: Vec<Arc<DataColumnSidecar<E>>>,
         kzg: &Kzg,
-    ) -> Result<Vec<Self>, (u64, KzgError)> {
+    ) -> Result<Vec<Self>, Vec<(ColumnIndex, KzgError)>> {
         verify_kzg_for_data_column_list_with_scoring(data_columns.iter(), kzg)?;
         Ok(data_columns
             .into_iter()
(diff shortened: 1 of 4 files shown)


beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -239,6 +239,18 @@ impl<E: EthSpec> KzgVerifiedDataColumn<E> {
     pub fn new(data_column: Arc<DataColumnSidecar<E>>, kzg: &Kzg) -> Result<Self, KzgError> {
         verify_kzg_for_data_column(data_column, kzg)
     }
+
+    pub fn from_batch(
+        data_columns: Vec<Arc<DataColumnSidecar<E>>>,
+        kzg: &Kzg,
+    ) -> Result<Vec<Self>, (u64, KzgError)> {
```
jimmygchen: nit:

```suggestion
    ) -> Result<Vec<Self>, (ColumnIndex, KzgError)> {
```

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -378,6 +390,38 @@ where
...
+/// Complete kzg verification for a list of `DataColumnSidecar`s.
+///
+/// If there's at least one invalid column, it re-verifies all columns individually to identify the
+/// first column that is invalid. This is necessary to attribute fault to the specific peer that
+/// sent bad data. The re-verification cost should not be significant. If a peer sends invalid data it
+/// will be quickly banned.
+pub fn verify_kzg_for_data_column_list_with_scoring<'a, E: EthSpec, I>(
+    data_column_iter: I,
+    kzg: &'a Kzg,
+) -> Result<(), (u64, KzgError)>
```
jimmygchen: Why don't we return a Vec of this? in case if there's multiple failures, we'd want to penalise all failed peers.

Currently if we reach the unreachable `Err((0, batch_err))`, we penalise the peer that served column index `0`. 

Not sure if this is really `unreachable` either, this assumes the KZG lib is bug free and it would not be possible for us to catch _if_  it happens. If we make this return value a `Vec`, then we can:
- return errors for multiple failures
- allow us to indicate to the caller that the unexpected scenario occured by passing an empty `vec`

dapplion: Changed to tracking Vec https://github.com/sigp/lighthouse/pull/6851/commits/d77ce79483e5470a9150884a8bb8baf660d5a8c5

beacon_node/beacon_chain/src/data_column_verification.rs
dapplion: https://github.com/sigp/lighthouse/pull/6851/commits/d77ce79483e5470a9150884a8bb8baf660d5a8c5

jimmygchen (approved):
Looks great!

Issue #6845, opened 2025-01-23
More Mergify improvements


michaelsproul:
We've adopted some new Mergify patterns in:

- https://github.com/sigp/lighthouse/pull/6842
- https://github.com/sigp/lighthouse/pull/6597

There are still a few outstanding issues.

- [x] I got the label name wrong for `ready-for-merge`. The config is incorrectly using `ready-to-merge`. We could rename the label, or update the config (which we probably want to do anyway).
- [ ] We should ideally remove approval on `trivial` PRs if the `trivial` label is removed.
- [x] We should not let the bot approve `trivial` PRs that have conflicts.
- [x] We should not let the bot queue PRs that are labelled `do-not-merge` (this is orthogonal to the recent changes, but might be nice to have).

michaelsproul:
cc @jxs

michaelsproul:
Closing as complete

PR #6852, opened 2025-01-23
Improve mergify config

jxs:
## Issue Addressed

And partially address https://github.com/sigp/lighthouse/issues/6845

- [x] use `ready-for-merge` label instead of `ready-to-merge`
- [ ] We should ideally remove approval on `trivial` PRs if the `trivial` label is removed.
- [x] Do not let the bot approve `trivial` PRs that have conflicts.
- [x] Do not let the bot queue PRs that are labelled `do-not-merge.

the missing one seems impossible as I didn't find a [condition](https://docs.mergify.com/configuration/conditions/) for a removed label.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -18,7 +18,7 @@ pull_request_rules:
       - base=unstable
       - label=trivial
       - author=@sigp/lighthouse
-      - conflict
+      - -conflict
     actions:
       review:
         type: APPROVE
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -18,6 +18,7 @@ pull_request_rules:
       - base=unstable
       - label=trivial
       - author=@sigp/lighthouse
+      - conflict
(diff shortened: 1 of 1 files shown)


.github/mergify.yml
```diff
@@ -18,6 +18,7 @@ pull_request_rules:
       - base=unstable
       - label=trivial
       - author=@sigp/lighthouse
+      - conflict
```
jimmygchen: should this be `conflict=false`?

jxs: Hi Jimmy, thanks for noticing it forgot to update, `= false` doesn't work for boolean types:
`Operators are invalid for Boolean attribute: `conflict``
I used the `-` operator to negate it, see [here](https://docs.mergify.com/configuration/conditions/#modifiers)

jimmygchen (approved):
Looks good! Thanks @jxs

PR #6856, opened 2025-01-24
Fix Redb implementation and add CI checks

eserilev:
Ensure Redb compiles and ensure we check for Redb compilation errors during CI

### DIFF
--- a/beacon_node/store/src/forwards_iter.rs
+++ b/beacon_node/store/src/forwards_iter.rs
@@ -158,6 +158,8 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> Iterator
             return None;
         }
         self.inner
+            .as_mut()
+            .ok()?
             .next()?
             .and_then(|(slot_bytes, root_bytes)| {
                 let slot = slot_bytes
--- a/beacon_node/store/src/forwards_iter.rs
+++ b/beacon_node/store/src/forwards_iter.rs
@@ -159,7 +159,6 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> Iterator
         }
         self.inner
             .as_mut()
-            .ok()?
             .next()?
             .and_then(|(slot_bytes, root_bytes)| {
                 let slot = slot_bytes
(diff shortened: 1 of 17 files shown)


beacon_node/store/src/forwards_iter.rs
```diff
@@ -158,6 +158,8 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> Iterator
             return None;
         }
         self.inner
+            .as_mut()
+            .ok()?
```
michaelsproul: I think we should put this `?` in the `::new` function where we call `iter_column_from`?

beacon_node/store/src/lib.rs
```diff
@@ -47,8 +47,9 @@ pub use types::*;
 
 const DATA_COLUMN_DB_KEY_SIZE: usize = 32 + 8;
 
-pub type ColumnIter<'a, K> = Box<dyn Iterator<Item = Result<(K, Vec<u8>), Error>> + 'a>;
-pub type ColumnKeyIter<'a, K> = Box<dyn Iterator<Item = Result<K, Error>> + 'a>;
+pub type ColumnIter<'a, K> =
+    Result<Box<dyn Iterator<Item = Result<(K, Vec<u8>), Error>> + 'a>, Error>;
+pub type ColumnKeyIter<'a, K> = Result<Box<dyn Iterator<Item = Result<K, Error>> + 'a>, Error>;
```
michaelsproul: I think these aliases should omit the `Result`, as it makes it less clear the iter methods can fail when `Result` is hidden behind an alias

michaelsproul (changes requested):
Looks good on the whole, just a few minor suggestions

beacon_node/store/src/forwards_iter.rs
eserilev: with the `ColumnIter` type change I just made, we no longer need the `?`

beacon_node/store/src/lib.rs
eserilev: updated to omit the `Result` and tweaked the Redb impl slightly

michaelsproul: looks good

michaelsproul (approved):

PR #6858, opened 2025-01-24
Refactor PeerManager to introduce Connectivity Module

diegomrsantos:
## Issue Addressed

https://github.com/sigp/lighthouse/issues/6860

## Proposed Changes

1. Introduction of Connectivity Struct:

     - Added a new Connectivity struct that uses the NetworkGlobalsProvider trait.
     - Encapsulates peer management parameters and methods such as dial_peer, peers_discovered, and maintain_peer_count.

2. NetworkGlobalsProvider Trait:

    - A new trait NetworkGlobalsProvider is introduced. The idea is to create a generic and minimal invasive interface to the current code which other projects are free to decide how to implement.
     - Methods include connected_outbound_only_peers, connected_or_dialing_peers, update_min_ttl, and should_dial.

3. Refactoring PeerManager:

     - PeerManager now includes the Connectivity struct.
     - Methods within PeerManager delegate connectivity-related operations to the Connectivity struct.
Removed direct use of discovery_enabled and network global operations from PeerManager.

4. File Changes:

     - connectivity.rs: Added implementation of the Connectivity struct.
     - mod.rs: Updated to include the new Connectivity struct and network_globals_provider.
     - network_behaviour.rs: Updated to use connectivity methods for checking connection limits.
     - network_globals_provider.rs: New file introducing the NetworkGlobalsProvider trait and its implementation for Arc<NetworkGlobals<E>>.

## Additional Info

The plan is to continue refactoring PeerManager and introducing new modules.


beacon_node/lighthouse_network/src/peer_manager/mod.rs
```diff
@@ -342,51 +349,17 @@ impl<E: EthSpec> PeerManager<E> {
...
-            debug!(self.log, "Skipping recursive discovery query after finding no useful results"; "results" => results_count);
-            metrics::inc_counter(&metrics::DISCOVERY_NO_USEFUL_ENRS);
-        } else {
-            // Queue another discovery if we need to
-            self.maintain_peer_count(to_dial_peers);
+        let wanted_peers = self.connectivity.peers_discovered(&results);
+        if wanted_peers > 0 {
+            self.events
+                .push(PeerManagerEvent::DiscoverPeers(wanted_peers));
+            // debug!(self.log,
```
AgeManning: Do we still need this log? 

I notice this is down lower in the file

beacon_node/lighthouse_network/src/peer_manager/connectivity.rs
```diff
@@ -0,0 +1,188 @@
...
+            {
+                // This should be updated with the peer dialing. In fact created once the peer is
+                // dialed
+                let peer_id = enr.peer_id();
+                if let Some(min_ttl) = min_ttl {
+                    self.network_globals_provider
+                        .update_min_ttl(&peer_id, *min_ttl);
+                }
+                if self.dial_peer(enr.clone()) {
+                    //debug!(self.log, "Added discovered ENR peer to dial queue"; "peer_id" => %peer_id);
```
AgeManning: Do we need this log?

beacon_node/lighthouse_network/src/peer_manager/connectivity.rs
```diff
@@ -0,0 +1,188 @@
...
+            //debug!(self.log, "Skipping recursive discovery query after finding no useful results"; "results" => results_count);
+            crate::metrics::inc_counter(&crate::metrics::DISCOVERY_NO_USEFUL_ENRS);
+            0
+        } else {
+            // Queue another discovery if we need to
+            self.maintain_peer_count(to_dial_peers)
+        }
+    }
+
+    /// This function checks the status of our current peers and optionally requests a discovery
```
AgeManning, line 140: These comments are now longer correct. This function returns a usize and doesn't request any discoveries. Maybe we need to change the name of the function to better reflect this change of logic.

beacon_node/lighthouse_network/src/peer_manager/connectivity.rs
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    pub fn next_peer_to_dial(&mut self) -> Option<Enr> {
+        self.peers_to_dial.pop()
+    }
+
+    /// Peers that have been returned by discovery requests that are suitable for dialing are
+    /// returned here.
+    ///
+    /// This function decides whether or not to dial these peers.
```
AgeManning, line 82: These comments should describe what the return value is used for as this logic has changed here and it might be easier for future people to understand when looking at what these functions do

AgeManning (commented):
This looks good to me. Seems like a good way to split things up. 

Historically there have been some sneaky bugs in the peer manager when we try to modify its workings. It usually requires a bit more testing than most other parts of code. 

We are currently in the planning of making a testnet release for an upcomming fork. We'll have to decide whether to put this in with electra or leave it for later. 

Have you done any testing on lighthouse for this?

beacon_node/lighthouse_network/src/peer_manager/mod.rs
diegomrsantos: I'm very sorry about that. I decided to make the code more similar to the original one. Please retake a look.

jimmygchen:
@diegomrsantos @AgeManning is this PR still required?

AgeManning:
I'm not sure. I think we may have gone another way inside libp2p, right @diegomrsantos ?

Issue #6514, opened 2024-10-18
Add SSZ Support to builder API


eserilev:
Spec: https://github.com/ethereum/builder-specs/pull/104

PR #6859, opened 2025-01-24
Add builder SSZ flow

eserilev:
Closes: https://github.com/sigp/lighthouse/issues/6514

- Added `DecodeByFork` trait. This is required for an axum extractor I introduced in this [PR](https://github.com/sigp/ethereum_apis/pull/8)
- Added an `AtomicBool` field on the builder client struct. When set to true, it indicates that SSZ is enabled for the builder. This flag is updated each time a response is received from `GET builder/header`.
- When a validator requests a blinded payload, we accept both JSON and SSZ but indicate a preference for SSZ
- Added logic to fallback to JSON if any response headers are missing


beacon_node/builder_client/src/lib.rs
```diff
@@ -64,13 +74,87 @@ impl BuilderHttpClient {
             server,
             timeouts: Timeouts::new(builder_header_timeout),
             user_agent,
+            ssz_enabled: Arc::new(false.into()),
```
michaelsproul, line 76: could use an `Arc<AtomicBool>`

beacon_node/builder_client/src/lib.rs
```diff
@@ -64,13 +74,87 @@ impl BuilderHttpClient {
...
+                };
+                T::from_ssz_bytes_by_fork(&response_bytes, fork_name)
+                    .map(|data| ForkVersionedResponse {
+                        version: Some(fork_name),
+                        metadata: EmptyMetadata {},
+                        data,
+                    })
+                    .map_err(Error::InvalidSsz)
+            }
+            ContentType::Json => {
```
michaelsproul, line 144: I think we may as well be dynamic here and set `ssz_enabled` to `false` if we get a JSON response

consensus/types/src/fork_versioned_response.rs
```diff
@@ -4,6 +4,11 @@ use serde::{Deserialize, Deserializer, Serialize};
 use serde_json::value::Value;
 use std::sync::Arc;
 
+pub trait ForkVersionDecode: Sized {
```
michaelsproul, line 7: I like this. Nice improvement

michaelsproul (approved):
Very nice! Looks good

michaelsproul:
I made a few tiny changes in this commit:

3eee2d188f1629c4f86fc9ba3e6391670cdc681b

Mostly just deleting a redundant `from_ssz_bytes` function.

PR #6861, opened 2025-01-24
Some sync/backfill format nits

dapplion:
## Proposed Changes

When working on unrelated changes I noted:

- An unnecessary closure left by a commit of some guy named @dapplion that can be removed
- match statements that can be simplified with the new let else syntax
- instead of mapping a result to ignore the Ok value, return

### DIFF
--- a/beacon_node/network/src/sync/backfill_sync/mod.rs
+++ b/beacon_node/network/src/sync/backfill_sync/mod.rs
@@ -425,9 +425,8 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
             Err(result) => {
                 let (expected_boundary, received_boundary, outcome) = match result {
                     Err(e) => {
-                        return self
-                            .fail_sync(BackFillError::BatchInvalidState(batch_id, e.0))
-                            .map(|_| ProcessResult::Successful);
+                        self.fail_sync(BackFillError::BatchInvalidState(batch_id, e.0))?;
+                        return Ok(ProcessResult::Successful);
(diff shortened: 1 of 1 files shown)


beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -388,67 +388,60 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
...
+                self.request_batches(network)?;
+                self.process_completed_batches(network)
+            }
+            Err(result) => {
+                let (expected_boundary, received_boundary, outcome) = match result {
+                    Err(e) => {
                         return self
-                            .fail_sync(BackFillError::BatchDownloadFailed(batch_id))
+                            .fail_sync(BackFillError::BatchInvalidState(batch_id, e.0))
                             .map(|_| ProcessResult::Successful);
```
michaelsproul: Another unnecessary closure here if you want to change it too

michaelsproul (commented):
Looks good, just code blocks moving around while retaining the same semantics. Happy to merge as-is or with that extra closure removed

beacon_node/network/src/sync/backfill_sync/mod.rs
dapplion: done https://github.com/sigp/lighthouse/pull/6861/commits/30ec5ea66b6760cea53216fef17976e123772b94

michaelsproul (approved):

PR #6862, opened 2025-01-24
Fix misc PeerDAS todos

dapplion:
## Issue Addressed

Address misc PeerDAS TODOs that are not too big for a dedicated PR

## Proposed Changes

I'll justify each TODO on an inlined comment

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -2924,10 +2924,9 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
     /// Only completed sampling results are received. Blocks are unavailable by default and should
     /// be pruned on finalization, on a timeout or by a max count.
     pub async fn process_sampling_completed(self: &Arc<Self>, block_root: Hash256) {
-        // TODO(das): update fork-choice
+        // TODO(das): update fork-choice, act on sampling result, adjust log level
         // NOTE: It is possible that sampling complets before block is imported into fork choice,
         // in that case we may need to update availability cache.
-        // TODO(das): These log levels are too high, reduce once DAS matures
         info!(self.log, "Sampling completed"; "block_root" => %block_root);
     }
 
--- a/beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
+++ b/beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
@@ -110,15 +110,6 @@ impl<E: EthSpec> PendingComponents<E> {
         self.get_cached_blobs().iter().flatten().count()
     }
 
-    /// Checks if a data column of a given index exists in the cache.
-    ///
-    /// Returns:
-    /// - `true` if a data column for the given index exists.
-    /// - `false` otherwise.
-    fn data_column_exists(&self, data_column_index: u64) -> bool {
-        self.get_cached_data_column(data_column_index).is_some()
-    }
-
     /// Returns the number of data columns that have been received and are stored in the cache.
     pub fn num_received_data_columns(&self) -> usize {
         self.verified_data_columns.len()
@@ -182,8 +173,7 @@ impl<E: EthSpec> PendingComponents<E> {
         kzg_verified_data_columns: I,
     ) -> Result<(), AvailabilityCheckError> {
         for data_column in kzg_verified_data_columns {
-            // TODO(das): Add equivalent checks for data columns if necessary
-            if !self.data_column_exists(data_column.index()) {
+            if self.get_cached_data_column(data_column.index()).is_none() {
                 self.verified_data_columns.push(data_column);
             }
         }
--- a/beacon_node/network/src/network_beacon_processor/gossip_methods.rs
+++ b/beacon_node/network/src/network_beacon_processor/gossip_methods.rs
@@ -1328,13 +1328,19 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                 return None;
             }
             Err(e @ BlockError::BlobNotRequired(_)) => {
-                // TODO(das): penalty not implemented yet as other clients may still send us blobs
-                // during early stage of implementation.
                 debug!(self.log, "Received blobs for slot after PeerDAS epoch from peer";
                     "error" => %e,
                     "peer_id" => %peer_id,
                 );
                 self.propagate_validation_result(message_id, peer_id, MessageAcceptance::Ignore);
+                // Receiving blobs when not required just spends unnecessary bandwidth of us. If the
+                // peer is malicious the damange is mild. If the peer is faulty, it's likely due to
+                // a non-updated node which will be quickly disconnected due to other faults.
+                self.gossip_penalize_peer(
+                    peer_id,
+                    PeerAction::MidToleranceError,
+                    "blob_not_required",
+                );
                 return None;
             }
         };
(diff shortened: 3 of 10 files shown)


beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -2924,10 +2924,9 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
     /// Only completed sampling results are received. Blocks are unavailable by default and should
     /// be pruned on finalization, on a timeout or by a max count.
     pub async fn process_sampling_completed(self: &Arc<Self>, block_root: Hash256) {
-        // TODO(das): update fork-choice
+        // TODO(das): update fork-choice, act on sampling result, adjust log level
         // NOTE: It is possible that sampling complets before block is imported into fork choice,
         // in that case we may need to update availability cache.
-        // TODO(das): These log levels are too high, reduce once DAS matures
         info!(self.log, "Sampling completed"; "block_root" => %block_root);
```
dapplion, line 2930: Merged these TODOs into one

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -182,8 +173,7 @@ impl<E: EthSpec> PendingComponents<E> {
         kzg_verified_data_columns: I,
     ) -> Result<(), AvailabilityCheckError> {
         for data_column in kzg_verified_data_columns {
-            // TODO(das): Add equivalent checks for data columns if necessary
-            if !self.data_column_exists(data_column.index()) {
+            if self.get_cached_data_column(data_column.index()).is_none() {
```
dapplion, line 176: Checks not necessary, we check the inclusion proof both via RPC and gossip. Inlined the function as it makes it easier to follow what's actually doing the `data_column_exists` check

beacon_node/network/src/sync/tests/lookups.rs
```diff
@@ -753,7 +752,6 @@ impl TestRig {
         }
 
         // Expect work event
-        // TODO(das): worth it to append sender id to the work event for stricter assertion?
```
dapplion, line 756: Not needed, there's a single request in this test

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -1328,13 +1328,19 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
                 );
                 self.propagate_validation_result(message_id, peer_id, MessageAcceptance::Ignore);
+                // Receiving blobs when not required just spends unnecessary bandwidth of us. If the
+                // peer is malicious the damange is mild. If the peer is faulty, it's likely due to
+                // a non-updated node which will be quickly disconnected due to other faults.
+                self.gossip_penalize_peer(
+                    peer_id,
+                    PeerAction::MidToleranceError,
+                    "blob_not_required",
+                );
```
dapplion: Implemented the penalty, all implementations should not send blobs by now @jimmygchen

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -336,9 +336,31 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+        let Some(slot) = custody_columns.iter().map(|d| d.slot()).next() else {
+            return;
+        };
+
+        if let Ok(current_slot) = self.chain.slot() {
+            if current_slot == slot {
+                let delay = get_slot_delay_ms(seen_timestamp, slot, &self.chain.slot_clock);
+                metrics::observe_duration(&metrics::BEACON_BLOB_RPC_SLOT_START_DELAY_TIME, delay);
+            }
+        }
```
dapplion, line 352: Use `seen_timestamp` is the same metric for blobs. Should we generalize the name?

beacon_node/network/src/sync/network_context/custody.rs
```diff
@@ -102,8 +103,6 @@ impl<T: BeaconChainTypes> ActiveCustodyRequest<T> {
         resp: RpcResponseResult<DataColumnSidecarList<T::EthSpec>>,
         cx: &mut SyncNetworkContext<T>,
     ) -> CustodyRequestResult<T::EthSpec> {
-        // TODO(das): Should downscore peers for verify errors here
```
dapplion, line 105: No, verify errors are handled in network_context.rs and downscored there

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -336,9 +336,31 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
         self: Arc<NetworkBeaconProcessor<T>>,
         block_root: Hash256,
         custody_columns: DataColumnSidecarList<T::EthSpec>,
-        _seen_timestamp: Duration,
+        seen_timestamp: Duration,
         process_type: BlockProcessType,
     ) {
+        // custody_columns must always have at least one element
+        let Some(slot) = custody_columns.iter().map(|d| d.slot()).next() else {
```
jimmygchen: Unnecessary to create an iterator if we just need to get the first element here

```suggestion
        let Some(slot) = custody_columns.first().map(|d| d.slot()) else {
```

beacon_node/network/src/network_beacon_processor/sync_methods.rs
jimmygchen: hmm I don't have a strong opinion here, I think it's fine to leave it as it is, and no need to introduce a breaking change. If people really find it confusing and want to update their dashboard (i doubt it lol) we can change.

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
jimmygchen: We already penalise peers for `BlobNotRequired` here:
https://github.com/sigp/lighthouse/blob/22c68ddf7de43e1d449c047466328493957de3b8/beacon_node/network/src/network_beacon_processor/gossip_methods.rs#L973-L987

Here's Pryms getting downscored in devnet 

```
Jan 23 10:15:00.128 DEBG Invalid gossip blob                     blob_index: 2, block_slot: 69, block_root: 0xcb9e88ec1f0505bbe527a7e2664bcf34e85c62943a636d74aabfe159c4aa2b52, outcome: BlobNotRequired(Slot(69))
Jan 23 10:15:00.128 DEBG Peer score adjusted                     score: -19.84, peer_id: 16Uiu2HAm3Dq3c3ubEQihCSFQ2dGu6TWwzazCTRwsS33AjtbENz8a, msg: bad_gossip_blob_ssz, service: libp2p
```

Why do we handle this variant in process gossip block here? Is it even possible to hit this?

jimmygchen (changes requested):
Re-reviewed, added some comments, otherwise looks good to merge soon.

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
dapplion: Oh right, it's unreachable here. Fixed https://github.com/sigp/lighthouse/pull/6862/commits/10f9f103dc623b30098a2f13148867dae630304a

jimmygchen (approved):
LGTM! Thanks for cleaning up the TODOs 🙏

PR #6864, opened 2025-01-25
Make range sync peer loadbalancing PeerDAS-friendly

dapplion:
## Issue Addressed

Range sync and backfill sync still assume that each batch request is done by a single peer. This assumption breaks with PeerDAS, where we request custody columns to N peers.

Issues with current unstable:

- Peer prioritization counts batch requests per peer. This accounting is broken now, data columns by range request are not accounted
- Peer selection for data columns by range ignores the set of peers on a syncing chain, instead draws from the global pool of peers
- The implementation is very strict when we have no peers to request from. After PeerDAS this case is very common and we want to be flexible or easy and handle that case better than just hard failing everything.

## Proposed Changes

- Upstream peer prioritization to the network context, it knows exactly how many active requests a peer (including columns by range)
- Upstream peer selection to the network context, now `block_components_by_range_request` gets a set of peers to choose from instead of a single peer. If it can't find a peer, it returns the error `RpcRequestSendError::NoPeer`
- Range sync and backfill sync handle `RpcRequestSendError::NoPeer` explicitly
  - Range sync: leaves the batch in `AwaitingDownload` state and does nothing. **TODO**: we should have some mechanism to fail the chain if it's stale for too long
  - Backfill sync: pauses the sync until another peer joins

### TODOs

- [ ] Re-add a mechanism to de-prioritize bad peers. Before we tracked peers that failed a specific batch and de-prioritize those. This is a bit trickier to do now because data columns by range request deal with peer groups. Instead we could
  - Count failures per peer in the network context and use those. I really wonder on the value of tracking failures on a specific batch vs all batches
  - Use the existing app peer score, which is a proxy for the count of failures
- [ ] Add tests :)

Note: this touches the mainnet path!

### DIFF
--- a/beacon_node/network/src/sync/backfill_sync/mod.rs
+++ b/beacon_node/network/src/sync/backfill_sync/mod.rs
@@ -879,33 +879,35 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
 
                     return Ok(());
                 }
-                Err(RpcRequestSendError::NoCustodyPeers) => {
-                    // If we are here the chain has no more synced peers
-                    info!(self.log, "Backfill sync paused"; "reason" => "insufficient_synced_peers");
-                    self.set_state(BackFillState::Paused);
-                    return Err(BackFillError::Paused);
-                }
-                Err(e) => {
-                    // NOTE: under normal conditions this shouldn't happen but we handle it anyway
-                    warn!(self.log, "Could not send batch request";
+                Err(e) => match e {
+                    RpcRequestSendError::NoPeer(no_peer) => {
+                        // If we are here the chain has no more synced peers
+                        info!(self.log, "Backfill sync paused"; "reason" => ?no_peer);
+                        self.set_state(BackFillState::Paused);
+                        return Err(BackFillError::Paused);
+                    }
+                    RpcRequestSendError::InternalError(e) => {
+                        // NOTE: under normal conditions this shouldn't happen but we handle it anyway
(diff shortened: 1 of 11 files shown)


beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -897,94 +843,71 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
...
-            self.send_batch(network, batch_id, peer)
-        } else {
-            // If we are here the chain has no more synced peers
-            info!(self.log, "Backfill sync paused"; "reason" => "insufficient_synced_peers");
-            self.set_state(BackFillState::Paused);
-            Err(BackFillError::Paused)
-        }
+        // TODO(das): previously here we de-prioritize peers that had failed to download or
+        // process a batch
+        self.send_batch(network, batch_id)
```
AgeManning, line 848: It looks like this function does nothing now and can just be replaced with `send_batch`?  (Unless implementing the TODO)

AgeManning (approved):
I didn't review this, but just wanted to throw in my 2 cents that I think the idea of bringing up the logic into network context is a good idea. 

There's still some todo's in here, I guess these can be addressed later.

dapplion:
Blocked by

jimmygchen:
@dapplion this PR got closed due to base branch being deleted, would you mind re-opening this against `unstable`?

Issue #6764, opened 2025-01-08
Only allow one data column computation at a time for a given block


jimmygchen:
## Description

See [this comment](https://github.com/sigp/lighthouse/pull/6756#discussion_r1906183914) for context.

Currently we allow for reconstruction and multiple column computation for the same block to happen at the same time, this is wasteful and we can avoid this.

In `PendingComponents`, we track reconstruction with a bool `reconstruction_started` and  we track blob to column computation with a `data_column_recv`. We should be able to replace the `reconstruction_started` usage with `data_column_recv` as well, and check if `data_column_recv` is `Some` before triggering either 1/ blob to column computation or 2/ column reconstruction.

https://github.com/sigp/lighthouse/blob/4173135299242ca420f3ba2439580d099c59f4b3/beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs#L23-L39

SunnysidedJ:
Hi, let me try working on this issue!

jimmygchen:
Hi @SunnysidedJ 

Thanks for helping out! - FYI with the recent proposals to offload proof computation from the beacon node, it's likely that we will get rid of the `data_column_recv` in order to reduce complexity, as the computation is no longer a long running task - see this comment here: https://github.com/sigp/lighthouse/pull/7117#discussion_r1991895794

I _think_ there will still be some race condition, but this may not be too big of an issue - I'll have to think about it a bit more. The change in #7117 will likely affect this, however we can't merge it yet because the [spec changes](https://github.com/ethereum/EIPs/pull/9378) for cell proof computation hasn't been finalised. It might be best to work on another issue for now.

One upcoming change that's kinda related is an optimisation to trigger `getBlobs` when the node receive a `data_column_sidecar` from gossip. Currently we only do this when we receive a block:

https://github.com/sigp/lighthouse/blob/cbf1c04a14861781fe503ec412cd05cce5a95a6c/beacon_node/network/src/network_beacon_processor/gossip_methods.rs#L1555-L1562

However we could also trigger `getBlobs` as soon as we receive  a valid data column - and if we do this, we'll have to handle concurrency as well (in case multiple columns triggered it simultaneously)

See the spec change here:
https://github.com/ethereum/consensus-specs/pull/4183/files#diff-81ad0f2992fd2a384c30b338ff6760b5b538a2f056b829db2d036ce42ba3090eR210

I haven't created an issue for this yet, but can do a writeup tomorrow if you're interested.

[...]

Issue #6465, opened 2024-10-04
Filter out the downloaded/computed sample columns and only store the custody columns


jimmygchen:
## Description

The initial version of Subnet Sampling for PeerDAS was implemented in #6410 - there's a TODO item from this PR to filter out the sampled columns:

https://github.com/sigp/lighthouse/blob/70194dfc6a3f4d10c9059610f889ff5a4e863a6a/beacon_node/beacon_chain/src/beacon_chain.rs#L7231-L7233

Currently all columns received, including columns for both custody and sampling are stored - however we only need to store the **custody columns**, as nodes are not required to serve non-custody columns (data column indices that the node is NOT assigned to custody), therefore it's not necessary to store them, and we should filter them out when persisting data columns.

We should NOT store the following columns:
- The extra columns downloaded for sampling purpose only (see below)
- Columns computed  from EL blobs, we currently store all columns we compute. If the `getBlobsV1` hit rate is high, we end up storing all columns, making the node consume the same amount of storage as a supernode! (i've added `fulu` label to this issue, because we should not go live without fixing this)

## Additional Info

Under PeerDAS, each node has a minimal custody requirement, which is [`chain_spec.custody_requirement: 8`](https://github.com/sigp/lighthouse/blob/70194dfc6a3f4d10c9059610f889ff5a4e863a6a/consensus/types/src/chain_spec.rs#L205) - currently this means validating and custody 8 assigned columns (computed based on node ID). Additionally, each node must sample at least `chain_spec.samples_per_slot: 16` in order to satisfy PeerDAS's sampling requirements.

In this case, the node has
- 8 columns that are custody columns, meaning the node advertises them and serve them to peers
- 8 additional columns that are sampling columns, meaning the node downloads and validate them for the purpose of sampling only, and do not store them.

Logic to derive sampling size is [here](https://github.com/sigp/lighthouse/blob/70194dfc6a3f4d10c9059610f889ff5a4e863a6a/consensus/types/src/chain_spec.rs#L693-L705).

jimmygchen:
Marking this as `blocked` for now, as it isn't very clear whether we want this change.

jimmygchen:
> We should NOT store the following columns:

> 1. The extra columns downloaded for sampling purpose only (see below)
[...]

PR #6866, opened 2025-01-25
Block availability data enum

dapplion:
## Issue Addressed

PeerDAS has undergone multiple refactors + the blending with the get_blobs optimization has generated technical debt.

A function signature like this

https://github.com/sigp/lighthouse/blob/f008b84079bbb6eb86de22bb3421dfc8263a5650/beacon_node/beacon_chain/src/beacon_chain.rs#L7171-L7178

Allows at least the following combination of states:
- blobs: Some / None
- data_columns: Some / None
- data_column_recv: Some / None
- Block has data? Yes / No
- Block post-PeerDAS? Yes / No

In reality, we don't have that many possible states, only:

- `NoData`: pre-deneb, pre-PeerDAS with 0 blobs or post-PeerDAS with 0 blobs
- `Blobs(BlobSidecarList<E>)`: post-Deneb pre-PeerDAS with > 0 blobs
- `DataColumns(DataColumnSidecarList<E>)`: post-PeerDAS with > 0 blobs
- `DataColumnsRecv(oneshot::Receiver<DataColumnSidecarList<E>>)`: post-PeerDAS with > 0 blobs, but we obtained the columns via reconstruction

^ this are the variants of the new `AvailableBlockData` enum

So we go from 2^5 states to 4 well-defined. Downstream code benefits nicely from this clarity and I think it makes the whole feature much more maintainable.

Currently `is_available` returns a bool, and then we construct the available block in `make_available`. In a way the availability condition is duplicated in both functions. Instead, this PR constructs `AvailableBlockData` in `is_available` so the availability conditions are written once

```rust
if let Some(block_data) = is_available(..) {
    let available_block = make_available(block_data);
}
```


beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -204,46 +185,62 @@ impl<E: EthSpec> PendingComponents<E> {
     ///
     /// Returns `true` if both the block exists and the number of received blobs / custody columns
     /// matches the number of expected blobs / custody columns.
```
jimmygchen: docs above need update

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -204,46 +185,62 @@ impl<E: EthSpec> PendingComponents<E> {
...
-            .unwrap_or("unknown".to_string());
+        let num_expected_blobs = block.num_blobs_expected();
+        if num_expected_blobs == 0 {
+            return Some(AvailableBlockData::NoData);
+        }
 
-        let num_received_blobs = self.num_received_blobs();
-        let num_received_columns = self.num_received_data_columns();
+        if spec.is_peer_das_enabled_for_epoch(block.epoch()) {
+            if self.verified_data_columns.len() >= custody_column_count {
```
jimmygchen: We've discussed about this before - we don't allow for duplicates when adding columns to the available cache, so if `len()` is greater than custody column count, we likely have added non-custody column to the cache, and this would result in: 
1. inconsistent columns in DB
2. incorrectly identifying a block as available where it isn't (getting some non custody columns, but custody columns may be missing)

With the current approach, we must be able to detect if there's a bug here.

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -204,46 +185,62 @@ impl<E: EthSpec> PendingComponents<E> {
...
-        debug!(
-            log,
-            "Component(s) added to data availability checker";
-            "block_root" => ?self.block_root,
-            "received_blobs" => num_received_blobs,
-            "expected_blobs" => expected_blobs_msg,
-            "received_columns" => num_received_columns,
-            "expected_columns" => expected_columns_msg,
-        );
+                // TODO(das) get only the columns that we need, filter the rest
```
jimmygchen: Is this filter necessary with the current approach? 
Is this `TODO` to account for peer sampling?
Feels like it's inconsistent in the context of subnet sampling though - because we could have enough `verified_data_columns` but end up storing less than `custody_column_count`?

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -204,46 +185,62 @@ impl<E: EthSpec> PendingComponents<E> {
...
+                    .verified_data_columns
+                    .iter()
+                    .map(|d| d.clone().into_inner())
+                    .collect::<Vec<_>>();
+                return Some(AvailableBlockData::DataColumns(data_columns));
+            }
 
-        let all_columns_received = expected_columns_opt
-            .is_some_and(|num_expected_columns| num_expected_columns == num_received_columns);
+            if let Some(data_columns_recv) = self.data_column_recv.take() {
```
jimmygchen: Nice, the ordering is good - so if we get enough gossip columns while computing, then we still make the block available  and not wait for the computation.

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -204,46 +185,62 @@ impl<E: EthSpec> PendingComponents<E> {
...
+                // columns, so we consider the block available
+                return Some(AvailableBlockData::DataColumnsRecv(data_columns_recv));
+            }
 
-        all_blobs_received || all_columns_received
+            None
+        } else {
+            // Before PeerDAS, blobs
+            let num_received_blobs = self.verified_blobs.iter().flatten().count();
+            if num_received_blobs >= num_expected_blobs {
```
jimmygchen: it's definitely a bug if `>` happens, I'm not sure if we just make it available and store the the "uncommitted blobs"

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -204,46 +185,62 @@ impl<E: EthSpec> PendingComponents<E> {
...
+                // TODO(das): Should do something special if `num_received_blobs > num_expected_blobs`?
```
jimmygchen: We can just log an error indicating a potential bug and not make the block available?

And there's also a chance that this will trigger panic due to `expect` usage below, if we allow `num_receive_blobs` to be more than `num_expected_blob`

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: I think the previous handling is good - and if there's unexpected blob there should be an error `AvailabilityCheckError::Unexpected`

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -204,46 +185,62 @@ impl<E: EthSpec> PendingComponents<E> {
...
+                let blobs_vec = self
+                    .verified_blobs
+                    .iter()
+                    .flatten()
+                    .map(|blob| blob.clone().to_blob())
+                    .collect::<Vec<_>>();
+                let max_blobs = spec.max_blobs_per_block(block.epoch());
+                let blobs = RuntimeVariableList::new(blobs_vec, max_blobs as usize)
+                    .expect("num_expect_blobs is less than max_blobs");
```
jimmygchen: We should avoid panic here

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -742,13 +740,24 @@ where
...
+
 /// A fully available block that is ready to be imported into fork choice.
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Debug)]
 pub struct AvailableBlock<E: EthSpec> {
     block_root: Hash256,
     block: Arc<SignedBeaconBlock<E>>,
-    blobs: Option<BlobSidecarList<E>>,
-    data_columns: Option<DataColumnSidecarList<E>>,
+    data: AvailableBlockData<E>,
```
jimmygchen: Should we call this field `blob_data`? `data` sounds very vague in this context - no strong opinion here though

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -276,63 +274,27 @@ impl<E: EthSpec> PendingComponents<E> {
...
-        import_data.data_column_recv = data_column_recv;
-
         let available_block = AvailableBlock {
             block_root,
             block,
-            blobs,
-            data_columns,
-            blobs_available_timestamp,
+            data: available_data,
+            blobs_available_timestamp: None,
```
jimmygchen: Is dropping this intentional?

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -365,6 +327,41 @@ impl<E: EthSpec> PendingComponents<E> {
                 None
             })
     }
+
+    pub fn status_str(
+        &self,
+        spec: &ChainSpec,
+        block_epoch: Epoch,
+        sampling_column_count: usize,
```
jimmygchen, line 319: nit: we usually put `spec` as the last parameter

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -562,12 +554,72 @@ impl<T: BeaconChainTypes> DataAvailabilityCheckerInner<T> {
...
+            pending_components.make_available(&self.spec, available_data, |diet_block| {
+                self.state_cache.recover_pending_executed_block(diet_block)
+            })
+        } else {
+            write_lock.put(block_root, pending_components);
+            Ok(Availability::MissingComponents(block_root))
+        }
+    }
+
+    /// The `data_column_recv` parameter is an optional `Receiver` for data columns that are
```
jimmygchen: `Receiver` not optional here

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -562,12 +554,72 @@ impl<T: BeaconChainTypes> DataAvailabilityCheckerInner<T> {
...
+    /// computed asynchronously. This method remains **used** after PeerDAS activation, because
```
jimmygchen: We can probably delete `This method remains **used** after PeerDAS activation`

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -562,12 +554,72 @@ impl<T: BeaconChainTypes> DataAvailabilityCheckerInner<T> {
...
+                PendingComponents::empty(
+                    block_root,
+                    self.spec.max_blobs_per_block(block_epoch) as usize,
+                )
+            });
+
+        // If `data_column_recv` is `Some`, it means we have all the blobs from engine, and have
+        // started computing data columns. We store the receiver in `PendingComponents` for
+        // later use when importing the block.
+        // TODO(das): Error or log if we overwrite a prior receiver
```
jimmygchen: We can link this issue:
https://github.com/sigp/lighthouse/issues/6764

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -562,12 +554,72 @@ impl<T: BeaconChainTypes> DataAvailabilityCheckerInner<T> {
...
+            .pop_entry(&block_root)
+            .map(|(_, v)| v)
+            .unwrap_or_else(|| {
+                PendingComponents::empty(
+                    block_root,
+                    self.spec.max_blobs_per_block(block_epoch) as usize,
+                )
+            });
+
+        // If `data_column_recv` is `Some`, it means we have all the blobs from engine, and have
```
jimmygchen: `data_column_recv` is no longer an option

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -1028,11 +1087,10 @@ mod test {
...
             kzg_verified_blobs.push(gossip_blob.into_inner());
             let availability = cache
-                .put_kzg_verified_blobs(root, kzg_verified_blobs.clone(), None, harness.logger())
+                .put_kzg_verified_blobs(root, kzg_verified_blobs.clone(), harness.logger())
                 .expect("should put blob");
-            assert_eq!(
-                availability,
-                Availability::MissingComponents(root),
+            assert!(
+                matches!(availability, Availability::MissingComponents(_)),
```
jimmygchen, line 1076: what's the reason for this change?

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3120,7 +3120,14 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             return Err(BlockError::DuplicateFullyImported(block_root));
         }
 
-        self.emit_sse_blob_sidecar_events(&block_root, blobs.iter().flatten().map(Arc::as_ref));
+        // process_engine_blobs is called for both pre and post PeerDAS. However, post PeerDAS
+        // consumers don't expect the blobs event to fire erraticly.
```
jimmygchen: Good catch

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -7200,34 +7206,17 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
-            if !blobs.is_empty() {
                 debug!(
-                    self.log, "Writing blobs to store";
+                    self.log, "Writing data_columns to store";
                     "block_root" => %block_root,
-                    "count" => blobs.len(),
+                    "count" => computed_data_columns.len(),
                 );
-                return Ok(Some(StoreOp::PutBlobs(block_root, blobs)));
+                Ok(Some(StoreOp::PutDataColumns(
```
jimmygchen, line 7265: Oops, forgot to raise an issue for this - we're storing all 128 columns here. Probably worth adding a TODO for the time being. I'll raise an issue tomorrow.

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -343,15 +347,14 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
...
         if self.blobs_required_for_block(&block) {
-            return if let Some(blob_list) = blobs.as_ref() {
+            return if let Some(blob_list) = blobs {
                 verify_kzg_for_blob_list(blob_list.iter(), &self.kzg)
                     .map_err(AvailabilityCheckError::InvalidBlobs)?;
                 Ok(MaybeAvailableBlock::Available(AvailableBlock {
                     block_root,
                     block,
-                    blobs,
+                    data: AvailableBlockData::Blobs(blob_list),
```
jimmygchen: I don't think it's an issue but just want to point out that we'll be using the `AvailableBlockData::Blobs` here instead of `AvailableBlockData::NoData` when the list is empty.

jimmygchen: Oh I think we'd actually add a `StoreOp` in this scenario, so we might end up storing an empty value

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -7168,29 +7169,34 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
-            } else if let Some(data_column_recv) = data_column_recv {
+        match block_data {
+            AvailableBlockData::NoData => Ok(None),
+            AvailableBlockData::Blobs(blobs) => {
+                debug!(
+                    self.log, "Writing blobs to store";
+                    "block_root" => %block_root,
+                    "count" => blobs.len(),
+                );
+                Ok(Some(StoreOp::PutBlobs(block_root, blobs)))
```
jimmygchen, line 7239: Should we keep the `is_empty` check for both blobs and columns so we dont't store empty values?

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -742,13 +740,24 @@ where
...
 }
 
+#[derive(Debug)]
+pub enum AvailableBlockData<E: EthSpec> {
+    NoData,
+    Blobs(BlobSidecarList<E>),
+    DataColumns(DataColumnSidecarList<E>),
+    /// An optional receiver for `DataColumnSidecarList`.
+    ///
+    /// This field is `Some` when data columns are being computed asynchronously.
```
jimmygchen: Docs need update, this is no longer an `Option`

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -779,39 +786,56 @@ impl<E: EthSpec> AvailableBlock<E> {
...
-    pub fn data_columns(&self) -> Option<&DataColumnSidecarList<E>> {
-        self.data_columns.as_ref()
+    pub fn data(&self) -> &AvailableBlockData<E> {
+        &self.data
+    }
+
+    pub fn has_blobs(&self) -> bool {
+        match self.data {
+            AvailableBlockData::NoData => false,
+            AvailableBlockData::Blobs(_) => true,
```
jimmygchen: As mentioned earlier in the Rpc block path, we might end up using this variant for an empty list

https://github.com/sigp/lighthouse/blob/05727290fbef27df72e854de0ca9280ee6347101/beacon_node/network/src/sync/block_sidecar_coupling.rs#L117-L134

beacon_node/beacon_chain/src/early_attester_cache.rs
```diff
@@ -70,14 +70,23 @@ impl<E: EthSpec> EarlyAttesterCache<E> {
...
 
-        let (_, block, blobs, data_columns) = block.deconstruct();
+        let (blobs, data_columns) = match block.data() {
+            AvailableBlockData::NoData => (None, None),
+            AvailableBlockData::Blobs(blobs) => (Some(blobs.clone()), None),
+            AvailableBlockData::DataColumns(data_columns) => (None, Some(data_columns.clone())),
+            // TODO(das): Once the columns are received, they will not be available in
+            // the early attester cache. If someone does a query to us via RPC we
+            // will get downscored.
+            AvailableBlockData::DataColumnsRecv(_) => (None, None),
```
jimmygchen, line 80: Good catch - this is an existing issue. I'll raise an issue tomorrow

jimmygchen: I think there's only little window where the downscoring can happen - once we finish computing then we'll be able to serve from the db

jimmygchen (changes requested):

jimmygchen (changes requested):
Nice, very elegant solution, and much cleaner.
I've added some comments, the main questions I have are the DA checker `is_available` checks and edge case handling.

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
dapplion: I have added a TODO in the line below to fix this case later. We were not handling this case before and it's out of scope for this PR IMO.

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
dapplion: I'm not sure what to do in this case. Erroring here is bad as it will prevent block import. This should not happen and points to an internal error. I would rather think how to handle the `>` cases in a separate PR.

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
dapplion: Yeah, we should be mindful that the sender will fail in that case

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
dapplion: Before we did not handle the `>` case, we had a `.take()` on the blobs iter, we would just ignore the extra blobs

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
dapplion: I added a `.take(max_blobs)` then
```
.expect("blobs_vec is less <= max_blobs because of take(max_blobs)");
```

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
dapplion: No, good catch

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
dapplion: `Availability` doesn't implement PartialEq because it includes the one_shot

beacon_node/beacon_chain/src/data_availability_checker.rs
dapplion: If blobs are empty we should use `NoData`. In what scenario de block has data and the blob list is not empty?

beacon_node/beacon_chain/src/beacon_chain.rs
dapplion: `AvailableBlockData::Blobs` enum by definition must have more than zero blobs. If we do a check here we should error, instead of ignoring the Op

beacon_node/beacon_chain/src/data_availability_checker.rs
dapplion: We should never construct the `AvailableBlockData::Blobs` variant with an empty blob list

dapplion: Added a check for `if num_expected_blobs == 0 { AvailableBlockData::NoData }` in both RPC paths

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Updated this issue to cover EL computed columns as well: https://github.com/sigp/lighthouse/issues/6465

beacon_node/beacon_chain/src/data_availability_checker.rs
jimmygchen: The RPC path. Looks like you've addressed them:

https://github.com/sigp/lighthouse/pull/6866#discussion_r1938074176

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
dapplion: I added a `.take(max_blobs)` then
```
.expect("blobs_vec is less <= max_blobs because of take(max_blobs)");
```

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -199,51 +180,71 @@ impl<E: EthSpec> PendingComponents<E> {
...
-    /// matches the number of expected blobs / custody columns.
-    pub fn is_available(&self, custody_column_count: usize, log: &Logger) -> bool {
-        let block_kzg_commitments_count_opt = self.block_kzg_commitments_count();
-        let expected_blobs_msg = block_kzg_commitments_count_opt
-            .as_ref()
-            .map(|num| num.to_string())
-            .unwrap_or("unknown".to_string());
+    /// Returns Some if the block has received all its required data for import. The return value
+    /// must be persisted in the DB along with the block.
+    pub fn is_available(
```
jimmygchen: nit on convention: the name `is_available` implies an immutable method that does not mutate `self` and return a bool - although I can't think of a better name.

beacon_node/beacon_chain/src/early_attester_cache.rs
jimmygchen: This is a very small window (~200ms) I'm not sure if it's worth doing this, inclined to leave it as it is and revisit if necessary.

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: Would you mind adding an issue for this?

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: The `take()` is unnecessary I believe, because we check for duplicates AND exact match before this PR

https://github.com/sigp/lighthouse/blob/a1b7d616b47604ec0cd1afb5543e03e68b629f96/beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs#L240-L244

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: As mentioned in the other comment, our code on `unstable` right now requires an exact match for either blobs or columns, why change this?

beacon_node/beacon_chain/src/data_availability_checker.rs
dapplion: `block_data` maybe?

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
dapplion: I have merged `is_available` fn into `make_available`

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3169,7 +3169,14 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             return Err(BlockError::DuplicateFullyImported(block_root));
         }
 
-        self.emit_sse_blob_sidecar_events(&block_root, blobs.iter().flatten().map(Arc::as_ref));
+        // process_engine_blobs is called for both pre and post PeerDAS. However, post PeerDAS
+        // consumers don't expect the blobs event to fire erraticly.
```
michaelsproul:
```suggestion
        // consumers don't expect the blobs event to fire erratically.
```

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -7218,29 +7219,34 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+                debug!(
+                    self.log, "Writing blobs to store";
+                    "block_root" => %block_root,
+                    "count" => blobs.len(),
+                );
+                Ok(Some(StoreOp::PutBlobs(block_root, blobs)))
+            }
+            AvailableBlockData::DataColumns(data_columns) => {
+                debug!(
+                    self.log, "Writing data_columns to store";
```
michaelsproul:
```suggestion
                    self.log, "Writing data columns to store";
```

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -7250,34 +7256,18 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
-                        "count" => data_columns.len(),
-                    );
-                    return Ok(Some(StoreOp::PutDataColumns(block_root, data_columns)));
-                }
-            }
-        } else if let Some(blobs) = blobs {
-            if !blobs.is_empty() {
                 debug!(
-                    self.log, "Writing blobs to store";
+                    self.log, "Writing data_columns to store";
```
michaelsproul:
```suggestion
                    self.log, "Writing data columns to store";
```

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -383,9 +392,8 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
         Ok(MaybeAvailableBlock::Available(AvailableBlock {
             block_root,
             block,
-            blobs: None,
             blobs_available_timestamp: None,
-            data_columns: None,
+            data: AvailableBlockData::NoData,
             spec: self.spec.clone(),
         }))
```
michaelsproul: This should be unreachable right? I wonder if we could simplify this, get rid of the early returns and just structure it like:

```rust
if block.num_expected_blobs() == 0 {
    // stuff
} else if blobs_required_for_block {
    // stuff
} else if data_columns_required {
    // stuff
} else {
   Err(..)
}
```

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -337,23 +341,29 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
...
+        } else if self.blobs_required_for_block(&block) {
+            return if let Some(blob_list) = blobs {
                 verify_kzg_for_blob_list(blob_list.iter(), &self.kzg)
                     .map_err(AvailabilityCheckError::InvalidBlobs)?;
                 Ok(MaybeAvailableBlock::Available(AvailableBlock {
                     block_root,
                     block,
-                    blobs,
+                    data: AvailableBlockData::Blobs(blob_list),
                     blobs_available_timestamp: None,
```
michaelsproul, line 352: It looks like you're just preserving existing behaviour, but it doesn't seem like we set `blobs_available_timestamp` anywhere in this method. Perhaps something to address in a follow-up PR like: 

- https://github.com/sigp/lighthouse/pull/6850

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -468,8 +482,7 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
                 MaybeAvailableBlock::Available(AvailableBlock {
```
michaelsproul, line 467: I think this is also unreachable because of the case added for `block.num_expected_blobs() == 0` above.

beacon_node/beacon_chain/src/data_availability_checker.rs
michaelsproul: I quite like `blob_data`. Data columns are still fundamentally blob-related

beacon_node/beacon_chain/src/historical_blocks.rs
```diff
@@ -144,17 +143,28 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+                .get_blobs_or_columns_store_op(block_root, block_data)
+                .map_err(|e| {
+                    HistoricalBlockError::StoreError(StoreError::DBError {
+                        message: format!("get_blobs_or_columns_store_op error {e:?}"),
+                    })
+                })?
+            {
+                for op in self.store.convert_to_kv_batch(vec![op])? {
+                    blob_batch.push(op);
+                }
```
michaelsproul:
```suggestion
                blob_batch.extend(self.store.convert_to_kv_batch(vec![op])?);
```

Simpler and also more efficient than multiple pushes

michaelsproul (commented):
A few suggested changes. I need to come back for overflow_lru_cache

beacon_node/beacon_chain/src/data_availability_checker.rs
dapplion: !! I just realized that's there an edge for the last else. 
- `blobs_required_for_block` = num expected blobs > 0 && da_check_required_for_epoch && !peerdas
- `columns_required_for_block` = num expected blobs > 0 && da_check_required_for_epoch && peerdas

So, what should we do if `num expected blobs > 0 && !da_check_required_for_epoch`, i.e. trying to sync a block with data prior to the retention window?

The name `da_check_required_for_epoch` is a bit sus. The check is ALWAYS required. But it's more like `da_check_possible_for_epoch`. As per the spec and discussion in Austria we should crash in that case, or at least require a flag where the user attests that a block is available without the full node being able to verify it.

dapplion:
@michaelsproul I have reverted the if else changes on `verify_kzg_for_rpc_block` and `verify_kzg_for_rpc_blocks` to the same behaviour as in stable. The default case at the end of the function (what you asked to change into an Err) is what allows LH to sync blocks outside the DA window. We should make that change explicitly in another PR after agreeing on
- https://github.com/sigp/lighthouse/issues/7011

dapplion:
Did a kurtosis run on 4d52ab6529d6bb7c668ff94b5443c574751fd2cf with deneb and a blob spammer. It finalized epoch 20 with blobs on all blocks I checked. No relevant ERR logs in all hosts

dapplion:
Doing a Kurtosis peerdas run I was hit by a probably unrelated error of interop compatibility between LH and Geth:

```
[cl-2-lighthouse-geth] Feb 19 05:38:55.078 WARN Execution engine call failed            error: BadResponse("Failed to convert json to execution requests : EmptyRequest(0)"), service: exec
[cl-2-lighthouse-geth] Feb 19 05:38:55.079 WARN Error processing HTTP API request       method: GET, path: /eth/v3/validator/blocks/37, status: 400 Bad Request, elapsed: 75.202982ms
```

This engine API error prevented block production on all blocks post-electra.

Also, this should be not a WARN since it prevents block production

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -179,128 +158,139 @@ impl<E: EthSpec> PendingComponents<E> {
...
-            verified_data_columns: vec![],
-            executed_block: None,
-            reconstruction_started: false,
-            data_column_recv: None,
-        }
-    }
-
     /// Verifies an `SignedBeaconBlock` against a set of KZG verified blobs.
     /// This does not check whether a block *should* have blobs, these checks should have been
     /// completed when producing the `AvailabilityPendingBlock`.
```
jimmygchen: Do you mean `AvailabilityPendingExecutedBlock` here?
```suggestion
    /// completed when producing the `AvailabilityPendingExecutedBlock`.
```

jimmygchen: Just realised this is existing comments that is outdated.

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -189,147 +158,139 @@ impl<E: EthSpec> PendingComponents<E> {
...
-        let is_peer_das_enabled = spec.is_peer_das_enabled_for_epoch(diet_executed_block.epoch());
-        let (blobs, data_columns) = if is_peer_das_enabled {
-            let data_columns = verified_data_columns
-                .into_iter()
-                .map(|d| d.into_inner())
-                .collect::<Vec<_>>();
-            (None, Some(data_columns))
+        let num_expected_blobs = block.num_blobs_expected();
+
+        let available_data = if num_expected_blobs == 0 {
```
jimmygchen: We could just call this `blob_data`?

jimmygchen (approved):
LGTM, would love to see this working on peerdas before we merge.

jimmygchen:
Tested locally and runs fine past the Fulu fork.

PR #6868, opened 2025-01-27
Make range sync chain Id sequential

dapplion:
## Issue Addressed

Currently, we set the `chain_id` of range sync chains to `u64(hash(target_root, target_slot))`, which results in a long integer. 

```
Jan 27 00:43:27.246 DEBG Batch downloaded, chain: 4223372036854775807, awaiting_batches: 0, batch_state: [p,E,E,E,E], blocks: 0, epoch: 0, service: range_sync
```

## Proposed Changes

Instead, we can use `network_context.next_id()` as we do for all other sync items and get a unique sequential (not too big) integer as id.

```
Jan 27 00:43:27.246 DEBG Batch downloaded, chain: 4, awaiting_batches: 0, batch_state: [p,E,E,E,E], blocks: 0, epoch: 0,  service: range_sync
```

Also, if a specific chain for the same target is retried later, it won't get the same ID so we can more clearly differentiate the logs associated with each attempt.

### DIFF
--- a/beacon_node/network/src/sync/range_sync/chain_collection.rs
+++ b/beacon_node/network/src/sync/range_sync/chain_collection.rs
@@ -9,6 +9,7 @@ use crate::metrics;
 use crate::sync::network_context::SyncNetworkContext;
 use beacon_chain::{BeaconChain, BeaconChainTypes};
 use fnv::FnvHashMap;
+use lighthouse_network::service::api_types::Id;
 use lighthouse_network::PeerId;
 use lighthouse_network::SyncInfo;
 use slog::{crit, debug, error};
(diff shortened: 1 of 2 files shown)


AgeManning (approved):
If this helps debugging moving away from a hash-based identification, then i'm all for it. 

The changes here seem fine to me.

beacon_node/network/src/sync/range_sync/chain_collection.rs
```diff
@@ -465,15 +466,17 @@ impl<T: BeaconChainTypes> ChainCollection<T> {
...
         } else {
             &mut self.head_chains
         };
-        match collection.entry(id) {
-            Entry::Occupied(mut entry) => {
-                let chain = entry.get_mut();
+
+        match collection
+            .iter_mut()
+            .find(|(_, chain)| chain.has_same_target(target_head_slot, target_head_root))
```
jimmygchen, line 477: If we have multiple chains with the same target, this will return the first one.
But this isn't possible I think because we only create a new chain when we cannot find one matching the same target (the `None` case), so I think this is fine.

jimmygchen (approved):
LGTM

PR #6869, opened 2025-01-27
Sync active request byrange ids logs

dapplion:
## Issue Addressed

Writing and running tests I noted that the sync RPC requests are very verbose now.

`DataColumnsByRootRequestId { id: 123, requester: Custody(CustodyId { requester: CustodyRequester(SingleLookupReqId { req_id: 121, lookup_id: 101 }) }) }`

Since this Id is logged rather often I believe there's value in 
1. Making them more succinct for log verbosity
2. Make them a string that's easy to copy and work with elastic

## Proposed Changes

Write custom `Display` implementations to render Ids in a more DX format

_ DataColumnsByRootRequestId with a block lookup_

```
123/Custody/121/Lookup/101
```

_DataColumnsByRangeRequestId_

```
123/122/RangeSync/0/5492900659401505034
```

- This one will be shorter after https://github.com/sigp/lighthouse/pull/6868

Also made the logs format and text consistent across all methods

### DIFF
--- a/beacon_node/lighthouse_network/src/service/api_types.rs
+++ b/beacon_node/lighthouse_network/src/service/api_types.rs
@@ -1,14 +1,13 @@
-use std::sync::Arc;
-
-use libp2p::swarm::ConnectionId;
-use types::{
-    BlobSidecar, DataColumnSidecar, Epoch, EthSpec, Hash256, LightClientBootstrap,
-    LightClientFinalityUpdate, LightClientOptimisticUpdate, LightClientUpdate, SignedBeaconBlock,
-};
-
 use crate::rpc::{
     methods::{ResponseTermination, RpcResponse, RpcSuccessResponse, StatusMessage},
     SubstreamId,
+};
+use libp2p::swarm::ConnectionId;
+use std::fmt::{Display, Formatter};
+use std::sync::Arc;
+use types::{
+    BlobSidecar, DataColumnSidecar, Epoch, EthSpec, Hash256, LightClientBootstrap,
+    LightClientFinalityUpdate, LightClientOptimisticUpdate, LightClientUpdate, SignedBeaconBlock,
 };
 
 /// Identifier of requests sent by a peer.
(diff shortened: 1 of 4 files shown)


beacon_node/lighthouse_network/src/service/api_types.rs
```diff
@@ -235,9 +234,138 @@ impl slog::Value for RequestId {
     }
 }
 
+// Since each request Id is deeply nested with various types, if rendered with Debug on logs they
+// take too much visual space. This custom Display implementations make the overall Id short while
+// not losing information
+impl Display for BlocksByRangeRequestId {
```
jxs, line 240: we can make it dry'er by defining a simple macro that impl's `Display` for the structs:
```rust
macro_rules! impl_display {
    ($structname: ty, $format: literal, $($field:ident),*) => {
        impl Display for $structname {
            fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
                write!(f, $format, $(self.$field,)*)
            }
        }
    };
}
```
and then calling it for the simple cases:
```rust
impl_display! {BlocksByRangeRequestId, "{}/{}", id, parent_request_id}
impl_display! {DataColumnsByRangeRequestId, "{}/{}", id, parent_request_id}
impl_display! {SingleLookupReqId, "{}/lookup/{}", req_id, lookup_id}
```

beacon_node/lighthouse_network/src/service/api_types.rs
```diff
@@ -235,9 +234,138 @@ impl slog::Value for RequestId {
...
+
+impl Display for SamplingRequester {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::ImportedBlock(block) => write!(f, "ImportedBlock/{block}"),
+        }
+    }
+}
+
+#[cfg(test)]
```
jxs, line 327: nit: 
I don't think we need tests for this, `Display` is reflected on the logs, if it's not according to what expected it's clear where one change, was there a special reason you added tests for?

beacon_node/network/src/network_beacon_processor/mod.rs
```diff
@@ -612,8 +612,13 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
         blocks: Vec<RpcBlock<T::EthSpec>>,
     ) -> Result<(), Error<T::EthSpec>> {
-        let is_backfill = matches!(&process_id, ChainSegmentProcessId::BackSyncBatchId { .. });
+        debug!(self.log, "Batch sending for process";
+            "blocks" => blocks.len(),
+            "id" => ?process_id,
+        );
+
         let processor = self.clone();
+        let id = process_id.clone();
```
jxs, line 621: nit: we can avoid cloning here, we can make `ChainSegmentProcessId` derive `Copy`

beacon_node/network/src/sync/network_context.rs
```diff
@@ -1102,14 +1106,38 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
     }
 
-    fn report_rpc_response_errors<R>(
+    fn on_rpc_response_result<I: std::fmt::Display, R, F: FnOnce(&R) -> usize>(
         &mut self,
+        id: I,
+        method: &'static str,
         resp: Option<RpcResponseResult<R>>,
         peer_id: PeerId,
+        get_count: F,
```
jxs, line 1118: why not just accept the count as an `usize`?

jxs (commented):
HI Lion, overall looks good to me, left some suggestions!

beacon_node/lighthouse_network/src/service/api_types.rs
dapplion: Mainly are regression tests, to ensure that a future mantainer understands the intended format and doesn't change it unintentionally

beacon_node/network/src/sync/network_context.rs
dapplion: Because I want to render the count of items in the Ok response

jimmygchen:
Tests are failing.

dapplion:
Let's wait to merge

jimmygchen:
@dapplion this PR got closed due to base branch being deleted, would you mind re-opening this against `unstable`?

beacon_node/lighthouse_network/src/service/api_types.rs
jimmygchen: +1 The test is a good idea here

beacon_node/lighthouse_network/src/service/api_types.rs
jimmygchen: Nice

PR #6872, opened 2025-01-28
Add more range sync tests

dapplion:
## Issue Addressed

Currently we have very poor coverage of range sync with unit tests. With the event driven test framework we could cover much more ground and be confident when modifying the code.

## Proposed Changes

Add two basic cases:
- Happy path, complete a finalized sync for 2 epochs
- Post-PeerDAS case where we start without enough custody peers and later we find enough

⚠️  If you have ideas for more test cases, please let me know! I'll write them


beacon_node/network/src/sync/tests/range.rs
```diff
@@ -439,3 +551,65 @@ fn pause_and_resume_on_ee_offline() {
...
+    if !r.fork_name.fulu_enabled() {
+        return;
+    }
+
+    let advanced_epochs: u64 = 2;
+    let remote_info = r.finalized_remote_info_advanced_by(advanced_epochs.into());
+
+    // Unikely that the single peer we added has enough columns for us. Find a way to make this test
+    // deterministic.
+    r.add_random_peer_not_supernode(remote_info.clone());
```
jimmygchen, line 592: We could just thread a fixed seed rng to these functions when generating peers?

```
StdRng::seed_from_u64(0xDEADBEEF0BAD5EEDu64)
```

For all peerdas tests we should probably make them deterministric, otherwise we could end up with flaky tests.

jimmygchen (changes requested):
Nice, just a small suggestion to make the test deterministic.

beacon_node/network/src/sync/tests/range.rs
dapplion: Made tests deterministic https://github.com/sigp/lighthouse/pull/6872/commits/2021abcd515bc4ac71c63c6f40e5c026ce18592e

Switched to ChaCha20 rng as Xor does not implement CryptoRng, required by the enr crate

jimmygchen (approved):
Nice, LGTM!

jimmygchen:
Ready for merge once release process is done.

PR #6882, opened 2025-01-29
Run Assertoor tests in CI

ackintosh:
## Issue Addressed

Added Assertoor tests to the local-testnet CI. 

## Proposed Changes

- The assertoor logs are included in the `logs-local-testnet` that is uploaded to GitHub Artifacts.
- Use `start_local_testnet.sh` so that we can also easily run the test locally.

## Additional Info

We can also schedule tests to run long-running tests, e.g., `run_lifecycle_test`.


jimmygchen:
@eserilev has done some prior work on this but didn't end up merging it: https://github.com/sigp/lighthouse/pull/6192

.github/workflows/local-testnet.yml
```diff
@@ -68,6 +68,20 @@ jobs:
         run: ./stop_local_testnet.sh local-blinded
         working-directory: scripts/local_testnet
 
+      - name: Start local testnet with Assertoor
+        run: ./start_local_testnet.sh -e local-assertoor -c -a -b false && sleep 60
+        working-directory: scripts/local_testnet
```
jimmygchen, line 57: I think we can probably have this replace the `Start local testnet` and `Start local testnet with blinded block production` tasks, as they don't currently do any verification.

jimmygchen (commented):
This looks good, thanks @ackintosh !
Just a question regarding replacing the existing kurtosis tests.

.github/workflows/local-testnet.yml
ackintosh: That makes sense, thanks!

ackintosh:
Thanks, Jimmy! I've updated this PR.

.github/workflows/local-testnet.yml
jimmygchen: Nice, let's merge this - I think we can update the network_params to have some nodes do blinded block production too.

jimmygchen (approved):
This is great thanks @ackintosh , good to add more confidence on block production and chain stability on CI runs!

Issue #6732, opened 2024-12-19
Lighthouse under-subscribes attestation subnets for aggregation


sauliusgrigaitis:
## Description

Lighthouse under-subscribes attestation subnets for aggregation.  Initially, I compared Grandine's and Lighthouse's networking behavior and found that Lighthouse stops subscribing to attestation subnets for aggregation completely. This was fixed by https://github.com/sigp/lighthouse/pull/6682 . However, even after this fix Lighthouse subscribes to significantly fewer number of attestations subnets.

## Version

The problem appears in 6.0.1 version. 

## Present Behaviour

6.0.1 version still subscribes to significantly fewer number of attestations subnets. A validator does not get directly penalized for it. However, this leads to a worse overall quality of aggregates in the network.

## Expected Behaviour

It should subscribe to the correct list of attestation subnets. 

## Steps to resolve

The easiest way to debug and fix it would be to print the subscription events in both 5.3 and 6.0.1 Lighthouse BN nodes behind Vouch with a large number of validators (I used 30K). And compare the difference of the behavior.

The image shows 6.0.1 at the first half and reverted 5.3 in the second half:

![Screenshot from 2024-12-19 11-06-36](https://github.com/user-attachments/assets/a57fb4ad-24b9-49ff-8ca1-e8810abf57d6)

michaelsproul:
cc @AgeManning

AgeManning:
> found that Lighthouse stops subscribing to attestation subnets for aggregation completely

This never happened. Lighthouse has always subscribed to subnets for aggregation. There was a drop that we noticed. The fix you mentioned corrected the case that we were aggregators for the same subnet, multiple times in one epoch, where 6.0.0 would subscribe just once, rather than multiple times.

We have revised the code a few times. The newer version is substantially simpler. When we tested, we did some quick math with high validator counts and the theoretical number of aggregate subscriptions checked out. It was hard for us to tell then which was the correct logic, the 5.3 potentially over-subscribing or 6.0.1 potentially under-subscribing. 

There might still be something here and it would be good to know which is correct, 5.3 or 6.0.1. I'll loop in @ackintosh and @jxs who might also be able to look into this.

sauliusgrigaitis:
> This never happened. 

Interesting, the reason why I requested Povi to check it was that I found there was no attestation traffic except for the 2 persistent subnets. Anyway, it's fixed now, so it doesn't matter.

[...]

PR #6890, opened 2025-01-30
Fix subnet unsubscription time

pawanjay176:
## Issue Addressed

Hopefully fixes https://github.com/sigp/lighthouse/issues/6732

## Proposed Changes

In our `scheduled_subscriptions`, we were setting unsubscription slot to be `current_slot + 1`. Given that we were subscribing to the subnet at `duty.slot - 1`, the unsubscription slot ended up being `duty.slot`. So we were unsubscribing to the subnet at the beginning of the duty slot which is insane.

Fixes the `scheduled_subscriptions` to unsubscribe at `duty.slot + 1`.

## Additional Info

Running this on kurtosis made me realise that we do not send subscriptions for the first couple of slots because of https://github.com/sigp/lighthouse/blob/unstable/validator_client/validator_services/src/duties_service.rs#L79-L85

Not sure if we want to handle that.

### DIFF
--- a/beacon_node/network/src/subnet_service/mod.rs
+++ b/beacon_node/network/src/subnet_service/mod.rs
@@ -636,7 +636,7 @@ impl<T: BeaconChainTypes> Stream for SubnetService<T> {
         match self.scheduled_subscriptions.poll_next_unpin(cx) {
             Poll::Ready(Some(Ok(exact_subnet))) => {
                 let ExactSubnet { subnet, slot } = exact_subnet;
-                // Set the `end_slot` for the subscription to be `duty.slot + 1` so that we unsubscribe 
+                // Set the `end_slot` for the subscription to be `duty.slot + 1` so that we unsubscribe
                 // only at the end of the duty slot.
                 if let Err(e) = self.subscribe_to_subnet_immediately(subnet, slot + 1) {
                     debug!(self.log, "Failed to subscribe to short lived subnet"; "subnet" => ?subnet, "err" => e);
--- a/beacon_node/network/src/subnet_service/mod.rs
+++ b/beacon_node/network/src/subnet_service/mod.rs
@@ -636,6 +636,8 @@ impl<T: BeaconChainTypes> Stream for SubnetService<T> {
         match self.scheduled_subscriptions.poll_next_unpin(cx) {
             Poll::Ready(Some(Ok(exact_subnet))) => {
                 let ExactSubnet { subnet, slot } = exact_subnet;
+                // Set the `end_slot` for the subscription to be `duty.slot + 1` so that we unsubscribe 
+                // only at the end of the duty slot.
                 if let Err(e) = self.subscribe_to_subnet_immediately(subnet, slot + 1) {
                     debug!(self.log, "Failed to subscribe to short lived subnet"; "subnet" => ?subnet, "err" => e);
                 }
--- a/beacon_node/network/src/subnet_service/mod.rs
+++ b/beacon_node/network/src/subnet_service/mod.rs
@@ -216,6 +216,12 @@ impl<T: BeaconChainTypes> SubnetService<T> {
             || self.permanent_attestation_subscriptions.contains(subnet)
     }
 
+    /// Returns whether we are subscribed to a permanent subnet for testing purposes.
+    #[cfg(test)]
+    pub(crate) fn is_subscribed_permanent(&self, subnet: &Subnet) -> bool {
+        self.permanent_attestation_subscriptions.contains(subnet)
+    }
+
     /// Processes a list of validator subscriptions.
     ///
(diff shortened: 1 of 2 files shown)


michaelsproul (approved):
Nice find!

The fix makes sense to me.

The testnet case is interesting, but I think a fix for that can wait for another PR. I guess we might want to special-case the VC's/BN's handling of the first ~2 slots of a network (feels kind of dumb), _or_ remove the loud warnings for late subscriptions from the BN and just track them with a metric.

michaelsproul:
Looks like one of the tests is failing. It might actually be expecting the old behaviour :eyes:

pawanjay176:
I fixed the failing test according to my understanding.
I think prior to #6146 , the test was checking if we bundled multiple unsubscriptions together into a single one (e.g. if we need to unsubscribe from subnet X at slot 9 and 12, then skip the unsubscribe at 9).

After the refactor, we decide to extend the unsubscription only in `subscribe_to_subnet_immediately` 1 slot before the subscription. So multiple unsubscriptions will get bundled together only in the case where there's overlap between an existing subscription and the new one (e.g. if we are subscribing at slot 11 and there's a planned unsubscription event at slot 12, then we just extend the subscription). 

TL;DR: the test `test_subscribe_same_subnet_several_slots_apart` wasn't refactored with the logic change in #6146 , and started failing when we fixed a bug. cbf7dca refactors the test.

jimmygchen (commented):
Nice find! 👍

beacon_node/network/src/subnet_service/tests/mod.rs
```diff
@@ -119,7 +118,7 @@ fn get_logger(log_level: Option<slog::Level>) -> Logger {
 static CHAIN: LazyLock<TestBeaconChain> = LazyLock::new(TestBeaconChain::new_with_system_clock);
 
 fn get_subnet_service() -> SubnetService<TestBeaconChainType> {
-    let log = test_logger();
+    let log = get_logger(TEST_LOG_LEVEL);
```
dapplion: Why not use the test_logger? We want to toggle logs programmatically with `--features test_logger`

beacon_node/network/src/subnet_service/tests/mod.rs
```diff
@@ -607,9 +604,18 @@ mod test {
 
         assert_eq!(no_events, []);
 
-        let second_subscribe_event = get_events(&mut subnet_service, None, 2).await;
+        let subscription_end_slot = current_slot + subscription_slot2 + 2; // +1 to get to the end of the duty slot, +1 for the slot to complete
```
dapplion, line 607: Place the comments above

beacon_node/network/src/subnet_service/mod.rs
```diff
@@ -629,9 +635,8 @@ impl<T: BeaconChainTypes> Stream for SubnetService<T> {
         // expire subscription.
         match self.scheduled_subscriptions.poll_next_unpin(cx) {
             Poll::Ready(Some(Ok(exact_subnet))) => {
-                let ExactSubnet { subnet, .. } = exact_subnet;
-                let current_slot = self.beacon_chain.slot_clock.now().unwrap_or_default();
-                if let Err(e) = self.subscribe_to_subnet_immediately(subnet, current_slot + 1) {
+                let ExactSubnet { subnet, slot } = exact_subnet;
+                if let Err(e) = self.subscribe_to_subnet_immediately(subnet, slot + 1) {
```
dapplion, line 641: You should add some comments referencing the issue and the reason for `slot + 1` in the code. Useful if we ever refactor this and git blame breaks

beacon_node/network/src/subnet_service/tests/mod.rs
pawanjay176: I set it to be consistent with `TestBeaconChain` since it uses `get_logger(TEST_LOG_LEVEL)`. 
Didn't know the `test_logger` feature was a thing. I'll change both to `test_logger()`

dapplion (approved):
LGTM

michaelsproul:
Fix for CI failure:

- https://github.com/sigp/lighthouse/pull/6901

AgeManning (approved):
Yeah nice catch. 

I wouldn't have imagined this would change the density of subscribed subnets between 5.3 and 6.0?

jimmygchen:
@mergify requeue

jimmygchen:
> Yeah nice catch.
> 
> I wouldn't have imagined this would change the density of subscribed subnets between 5.3 and 6.0?

Looks like this was introduced in #6146 which was part of the 6.0 release (My bad, I missed this in the review!). 

The old logic was good and same as the fix in this PR:
https://github.com/sigp/lighthouse/blob/08e8b92e5032044f253a7357aaf0a7c74f9f8b80/beacon_node/network/src/subnet_service/attestation_subnets.rs#L644-L647

PR #6898, opened 2025-02-01
Add check to Lockbud CI job

### LINKED ISSUE #6637: Lockbud CI job doesn't ever fail
## Description

The lockbud CI job checks for any deadlocks in the codebase. However, if a deadlock is found, the CI job doesn't actually fail. Instead it just prints out some text related to the deadlock and the CI job is marked as completed succesfully. 

image

It'd be much more helpful if the lockbud CI job actually failed when catching a deadlock.

There are two ways to potentially resolve this 

1. Contribute to lockbud to resolve the issue described [here](https://github.com/BurtonQin/lockbud/issues/74). 
2. Parse lockbud output like they do in this repo https://github.com/os-checker/os-checker and manually fail the CI job if a deadlock is found

eserilev:
Closes #6637

Adds a script to check if Lockbud failed. Thanks @magick93 for working on this!


eserilev:
heres a failure example:
image

scripts/ci/check-lockbud.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# Run lockbud to check for deadlocks and capture the output
+output=$(cargo lockbud -k deadlock -b -l tokio_util 2>&1)
+
+# Check if lockbud returned any issues
+if echo "$output" | grep -q '"bug_kind"'; then
```
jimmygchen, line 7: Note: this may produce false negatives if `cargo lockbud` failed for other reasons e.g. 
1. compilation error 
2. lockbud internal issues 

Although i think this is already better than before! Given that (1) compilation errors should fail other jobs as well, and we haven't seen (2) happened before (it may be very unlikely), I think we can go ahead and merge this.

jimmygchen (approved):

scripts/ci/check-lockbud.sh
magick93: Agree. 
This script was only intended to be a quick fix.

I do suggest we 
- centralize this script as currently its copy pasta between anchor and LH.
- ultimately improve the lockbud cli so it can return and error code with the result.

PR #6901, opened 2025-02-03
Cargo update for openssl vuln

michaelsproul:
## Issue Addressed

Run `cargo update` to address [RUSTSEC-2025-0004](https://rustsec.org/advisories/RUSTSEC-2025-0004), a vulnerability in `openssl`. I don't think we are affected, but this PR is required for us to pass `cargo audit` and unblock CI.

## Additional Info

The `rand 0.9` update being brought in is a bit sneaky on postgres' part. I suspect we'll want to update our crates to `rand 0.9` shortly.


jimmygchen (approved):
LGTM, thanks!

PR #6905, opened 2025-02-03
Address libp2p 0.55 deprecation of ConnectionHandler::{In,Out}boundOpenInfo

jxs:
## Issue Addressed
See [here](https://github.com/libp2p/rust-libp2p/issues/3268) for more info


jxs:
addressed upstream on https://github.com/libp2p/rust-libp2p/pull/5860.
We should be able to remove the `allow` attributes on the next patch release.

PR #6906, opened 2025-02-03
chore: update peerDAS KZG library to 0.5.3

kevaundray:
## Issue Addressed

This optimizes the time it takes to load the context, so that tests do not time out

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1576,9 +1576,9 @@ dependencies = [
 
 [[package]]
 name = "crate_crypto_internal_eth_kzg_bls12_381"
-version = "0.5.1"
+version = "0.5.3"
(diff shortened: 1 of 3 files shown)


kevaundray:
Initially branched off of stable, which is why there are force pushes

kevaundray:
Screenshot 2025-02-03 at 13 56 18

I get anywhere between 20-32% reduction, depending on the run. But the startup time is roughly the same at about 2.4-2.5 seconds.

To run these benchmarks:

1)  cd crypto/kzg
2) cargo bench

michaelsproul (approved):
Nice work, thanks! This does seem to have fixed the regression from 0.5.2!

PR #6914, opened 2025-02-05
Sync active request byrange ids logs

dapplion:
## Issue Addressed

- Re-opened PR from https://github.com/sigp/lighthouse/pull/6869

Writing and running tests I noted that the sync RPC requests are very verbose now.

`DataColumnsByRootRequestId { id: 123, requester: Custody(CustodyId { requester: CustodyRequester(SingleLookupReqId { req_id: 121, lookup_id: 101 }) }) }`

Since this Id is logged rather often I believe there's value in 
1. Making them more succinct for log verbosity
2. Make them a string that's easy to copy and work with elastic

## Proposed Changes

Write custom `Display` implementations to render Ids in a more DX format

_ DataColumnsByRootRequestId with a block lookup_

```
123/Custody/121/Lookup/101
```

_DataColumnsByRangeRequestId_

```
123/122/RangeSync/0/5492900659401505034
```

- This one will be shorter after https://github.com/sigp/lighthouse/pull/6868

Also made the logs format and text consistent across all methods

### DIFF
--- a/beacon_node/lighthouse_network/src/service/api_types.rs
+++ b/beacon_node/lighthouse_network/src/service/api_types.rs
@@ -1,14 +1,13 @@
-use std::sync::Arc;
-
-use libp2p::swarm::ConnectionId;
-use types::{
-    BlobSidecar, DataColumnSidecar, Epoch, EthSpec, Hash256, LightClientBootstrap,
-    LightClientFinalityUpdate, LightClientOptimisticUpdate, LightClientUpdate, SignedBeaconBlock,
-};
-
 use crate::rpc::{
     methods::{ResponseTermination, RpcResponse, RpcSuccessResponse, StatusMessage},
     SubstreamId,
+};
+use libp2p::swarm::ConnectionId;
+use std::fmt::{Display, Formatter};
+use std::sync::Arc;
+use types::{
+    BlobSidecar, DataColumnSidecar, Epoch, EthSpec, Hash256, LightClientBootstrap,
+    LightClientFinalityUpdate, LightClientOptimisticUpdate, LightClientUpdate, SignedBeaconBlock,
 };
 
 /// Identifier of requests sent by a peer.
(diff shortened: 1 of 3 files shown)


beacon_node/lighthouse_network/src/service/api_types.rs
```diff
@@ -235,9 +234,108 @@ impl slog::Value for RequestId {
...
+        }
+    };
+}
+
+// Since each request Id is deeply nested with various types, if rendered with Debug on logs they
+// take too much visual space. This custom Display implementations make the overall Id short while
+// not losing information
+impl_display!(BlocksByRangeRequestId, "{}/{}", id, parent_request_id);
+impl_display!(BlobsByRangeRequestId, "{}/{}", id, parent_request_id);
+impl_display!(DataColumnsByRangeRequestId, "{}/{}", id, parent_request_id);
```
jimmygchen, line 252: This is definitely better than before! although it may still be not as straight forward to search in an indexed log viewer, compared to other logs where we have key/values. 

e.g. 
```
123/122/RangeSync/0/5492900659401505034
```
This is quite hard to read without memorizing the order or looking at the code.

~~Also if we do want to go ahead with this format, would it make more sense to have the order the other way around? i.e. parent request id first?~~ (actually i realise it doesn't matter coz we have requester at the end)

beacon_node/lighthouse_network/src/service/api_types.rs
```diff
@@ -235,9 +234,108 @@ impl slog::Value for RequestId {
...
+            id: 123,
+            parent_request_id: ComponentsByRangeRequestId {
+                id: 122,
+                requester: RangeRequestId::RangeSync {
+                    chain_id: 54,
+                    batch_id: Epoch::new(0),
+                },
+            },
+        };
+        assert_eq!(format!("{id}"), "123/122/RangeSync/0/54");
```
jimmygchen, line 339: I found having consecutive numbers make it harder to guess what they are.

Would something like this be easier to understand?
123/Parent/122/RangeSync/Batch/0/Chain/54

jimmygchen (changes requested):
Just added my view on the new log format, no strong opinion though if this makes it easier for you to debug.

beacon_node/lighthouse_network/src/service/api_types.rs
dapplion: These are debug logs, only use will use this. Have you seen @AgeManning 's batch state codes? It's even harder to follow 🤣 . Remember that before we just had a number with no information at all. This IDs get logged quite a lot when sync is running I believe succinctness is good

jimmygchen: I disagree with this, because for anyone else to read this (even anyone from our team) we'd have to look it up in the code to know how to use this log, and I don't think adding `Parent` and `Batch` to it makes it a lot more verbose than it is now.

jimmygchen (approved):

jimmygchen (approved):
Please see comment above.
https://github.com/sigp/lighthouse/pull/6914#discussion_r1948274985

If it helps with debugging happy to merge now and revisit later

PR #6922, opened 2025-02-06
Make range sync peer loadbalancing PeerDAS-friendly

dapplion:
## Issue Addressed

- Re-opens https://github.com/sigp/lighthouse/pull/6864 targeting unstable

Range sync and backfill sync still assume that each batch request is done by a single peer. This assumption breaks with PeerDAS, where we request custody columns to N peers.

Issues with current unstable:

- Peer prioritization counts batch requests per peer. This accounting is broken now, data columns by range request are not accounted
- Peer selection for data columns by range ignores the set of peers on a syncing chain, instead draws from the global pool of peers
- The implementation is very strict when we have no peers to request from. After PeerDAS this case is very common and we want to be flexible or easy and handle that case better than just hard failing everything.

## Proposed Changes

- [x] Upstream peer prioritization to the network context, it knows exactly how many active requests a peer (including columns by range)
- [x] Upstream peer selection to the network context, now `block_components_by_range_request` gets a set of peers to choose from instead of a single peer. If it can't find a peer, it returns the error `RpcRequestSendError::NoPeer`
- [ ] Range sync and backfill sync handle `RpcRequestSendError::NoPeer` explicitly
  - [ ] Range sync: leaves the batch in `AwaitingDownload` state and does nothing. **TODO**: we should have some mechanism to fail the chain if it's stale for too long - **EDIT**: Not done in this PR
  - [x] Backfill sync: pauses the sync until another peer joins - **EDIT**: Same logic as unstable

### TODOs

- [ ] Add tests :)
- [x] Manually test backfill sync 

Note: this touches the mainnet path!


dapplion:
I would like to wait to merge this PR first to have more test coverage
- https://github.com/sigp/lighthouse/pull/6872

beacon_node/network/src/metrics.rs
```diff
@@ -746,16 +746,26 @@ pub fn update_sync_metrics<E: EthSpec>(network_globals: &Arc<NetworkGlobals<E>>)
...
+        }
     }
 
     for (sync_type, peer_count) in peers_per_sync_type {
         set_gauge_entry(&PEERS_PER_SYNC_TYPE, &[sync_type], peer_count);
     }
+
+    for (subnet, peer_count) in peers_per_column_subnet {
+        set_gauge_entry(
+            &PEERS_PER_COLUMN_SUBNET,
```
dapplion, line 766: `good_peers_on_sampling_subnets` is removed which previously set this metric

beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -871,106 +816,74 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
...
         for id in redownload_queue {
-            self.retry_batch_download(network, id)?;
+            self.send_batch(network, id)?;
         }
         // finally, re-request the failed batch.
-        self.retry_batch_download(network, batch_id)
-    }
-
-    /// Sends and registers the request of a batch awaiting download.
-    fn retry_batch_download(
```
dapplion, line 967: Now there's no different priorization between first send and retry. For both we prefer not failed peers with least active requests

beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -1012,35 +925,15 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
...
-                    .map(|requests| requests.is_empty())
-                    .unwrap_or(true)
-            })
-            .cloned()
-            .collect::<Vec<_>>();
-
-        idle_peers.shuffle(&mut rng);
-
-        while let Some(peer) = idle_peers.pop() {
+        loop {
```
dapplion: If the implementation of `include_next_batch` is correct we should never infinite loop here. But it's sus, I worry we introduce a bug in the future un-intentionally.

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -194,30 +186,8 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
...
-                    }
-                    self.retry_batch_download(network, id)?;
-                } else {
-                    debug!(self.log, "Batch not found while removing peer";
-                        "peer" => %peer_id, "batch" => id)
-                }
-            }
-        }
+    pub fn remove_peer(&mut self, peer_id: &PeerId) -> ProcessingResult {
+        self.peers.remove(peer_id);
```
dapplion, line 194: No need to track internal requests and fail them explicitly. If a disconnect interrupts an active request, since https://github.com/sigp/lighthouse/pull/6497 we will get an inject_error for that batch.

beacon_node/network/src/sync/network_context.rs
```diff
@@ -79,11 +79,18 @@ pub enum RpcResponseError {
...
-    /// Network channel send failed
-    NetworkSendError,
-    NoCustodyPeers,
-    CustodyRequestError(custody::Error),
-    SlotClockError,
+    /// No peer available matching the required criteria
+    NoPeer(NoPeerError),
+    /// These errors should never happen, including unreachable custody errors or network send
+    /// errors.
+    InternalError(String),
```
dapplion, line 93: I grouped errors with 2 variants as both backfill and forward range need to handle the `NoPeer` case and everything else is treated as an internal error

beacon_node/network/src/sync/network_context.rs
```diff
@@ -350,25 +357,95 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+            // - Backfill sync: will pause waiting for more peers to join
+            // - Forward sync: can never happen as the chain is dropped when removing the last peer.
+            return Err(RpcRequestSendError::NoPeer(NoPeerError::BlockPeer));
+        };
+
+        // Attempt to find all required custody peers before sending any request or creating an ID
+        let columns_by_range_peers_to_request =
+            if matches!(batch_type, ByRangeRequestType::BlocksAndColumns) {
+                let column_indexes = self.network_globals().sampling_columns.clone();
+                Some(self.select_columns_by_range_peers_to_request(&column_indexes, peers)?)
```
dapplion: Select column peers before sending the block request. So if we don't have enough custody peers we will error here

beacon_node/network/src/sync/network_context.rs
```diff
@@ -756,7 +878,25 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+                // - UnexpectedRequestId: Never happens: this Err is only constructed handling a
+                //   download or processing response
+                // - SendFailed: Should never happen unless in a bad drop sequence when shutting
+                //   down the node
+                e @ (CustodyRequestError::TooManyFailures
+                | CustodyRequestError::BadState { .. }
+                | CustodyRequestError::UnexpectedRequestId { .. }
+                | CustodyRequestError::SendFailed { .. }) => {
+                    RpcRequestSendError::InternalError(format!("{e:?}"))
+                }
```
dapplion, line 984: Coerce `CustodyRequestError` into `InternalError` so we don't have to handle the CustodyRequestError enum downstream

dapplion:
@jimmygchen I have reduced the scope of this PR. I intended to deprecate the check `good_peers_on_sampling_subnets` in this PR but it's a very sensitive change. I left that logic untouched, what we do now is:

- Check if `good_peers_on_sampling_subnets`
  - If no, don't create batch
  - If yes, create batch and send it
    - If we don't have enough custody peers, error and drop chain

In the future we can deprecate the `good_peers_on_sampling_subnets` check by allowing batches to remain in `AwaitingDownload` state. It's essentially duplicate code as we check for peer twice. It should make sync less likely to drop chains too, like we did in lookup sync by allowing batches to be peer-less for some time.

I added a `TODO(das)` to tackle in another PR

https://github.com/sigp/lighthouse/blob/1debbbf9aa455b3d4a49fd26e528e03d36cd1afe/beacon_node/network/src/sync/range_sync/chain.rs#L913-L918

beacon_node/network/src/sync/network_context.rs
ethDreamer: Inside `select_columns_by_range_peers_to_request()` you have:
```rust
fn select_columns_by_range_peers_to_request(
    &self,
    custody_indexes: &HashSet<ColumnIndex>,
    peers: &HashSet<PeerId>,
) -> Result<HashMap<PeerId, Vec<ColumnIndex>>, RpcRequestSendError> {
    ...
    // Re-compute here to account for the block peer
    let active_request_count_by_peer = self.active_request_count_by_peer();
}
```
so there you are expecting the block request to be sent before selecting the custody peers

beacon_node/network/src/sync/network_context.rs
```diff
@@ -379,70 +456,95 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+                        // Also account for requests that are not yet issued tracked in peer_id_to_request_map
+                        active_request_count_by_peer.get(peer).copied().unwrap_or(0)
+                            + peer_id_to_request_map
+                                .get(peer)
+                                .map(|columns| columns.len())
+                                .unwrap_or(0),
+                        // Random factor to break ties, otherwise the PeerID breaks ties
+                        rand::random::<u32>(),
+                        peer,
+                    )
```
ethDreamer, line 582: You're not taking into account failed peers (or what you called `peers_to_deprioritize` in `block_components_by_range_request()`) in this peer selection. Seems strange to account for when selecting a peer for the block but not when selecting peers for columns.

beacon_node/network/src/sync/network_context.rs
```diff
@@ -379,70 +456,95 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
-                            columns_by_range_request,
-                            id,
-                        )
-                    })
-                    .collect::<Result<Vec<_>, _>>()?;
-
-                (
-                    Some(column_indexes.into_iter().collect::<Vec<_>>()),
-                    Some(data_column_requests),
+        let data_column_requests =
```
ethDreamer: more idiomatic:
```rust
let data_column_requests = columns_by_range_peers_to_request
    .map(|peers_to_request| {
        peers_to_request
            .into_iter()
            .map(|(peer_id, columns)| {
                self.send_data_columns_by_range_request(
                    peer_id,
                    DataColumnsByRangeRequest {
                        start_slot: *request.start_slot(),
                        count: *request.count(),
                        columns,
                    },
                    id,
                )
            })
            .collect::<Result<Vec<_>, _>>()
    })
    .transpose()?;
```

dapplion: Fixed with 7ec350bd0

beacon_node/network/src/sync/network_context.rs
dapplion: Good find, fixed with 45f55282e

beacon_node/network/src/sync/network_context.rs
dapplion: Stale code and comment, fixed with 45f55282e

beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -441,13 +401,7 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
...
         }
 
-        // A stream termination has been sent. This batch has ended. Process a completed batch.
-        // Remove the request from the peer's active batches
-        self.active_requests
-            .get_mut(peer_id)
-            .map(|active_requests| active_requests.remove(&batch_id));
-
-        match batch.download_completed(blocks) {
+        match batch.download_completed(blocks, *peer_id) {
```
ethDreamer, line 402: I feel this should be re-architected. The `peer_id` passed into this function is just _the last peer_ to send us a component (block, blob, column). This peer then:
1. Gets passed to `download_completed()` where it's used to set `batch.state = BatchState::AwaitingProcessing(peer, ...)`
2. Other than logging, this is only ever used when we set `batch.state = BatchState::Processing(Attempt::new::<B, E>(peer, ..))`
3. From there, it's only ever used to append this peer to `failed_processing_attempts` should the processing fail.

This `failed_processing_attempts` list then enters into [`peers_to_deprioritize`](https://github.com/dapplion/lighthouse/blob/peer-loadbalancing/beacon_node/network/src/sync/network_context.rs#L404) or is used to do some [peer downscoring logic](https://github.com/dapplion/lighthouse/blob/peer-loadbalancing/beacon_node/network/src/sync/backfill_sync/mod.rs#L786) that's probably broken now that there's multiple peers to a batch.

We need to figure out how to treat these things now that there's multiple peers to any given batch and presumably any one of them could've been responsible for why the batch failed to process, rather than just punishing the *last peer* to send us a message.

It looks like you're kind of starting to rectify some of this logic here:
```rust
for peer in self.participating_peers.drain() {
    network.report_peer(peer, *penalty, "backfill_batch_failed");
}
```

ethDreamer (dismissed):

beacon_node/network/src/sync/backfill_sync/mod.rs
dapplion: I absolutely agree this has to be re-achitected. The change you suggest is part of https://github.com/sigp/lighthouse/issues/6258. I intend to make this PRs with a narrow scope to not drop a 3k PR with the hopes to have higher velocity.

Do you believe this PR _degrades_ current unstable in terms of peer tracking? Otherwise I would prefer not to implement this changes now, and defer to a future PR.

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -1096,27 +999,23 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
...
+                let batch_type = network.batch_type(epoch);
+                let optimistic_batch = BatchInfo::new(&epoch, EPOCHS_PER_BATCH, batch_type);
+                entry.insert(optimistic_batch);
+                self.send_batch(network, epoch)?;
             }
             return Ok(KeepChain);
         }
 
-        while let Some(peer) = idle_peers.pop() {
+        loop {
```
ethDreamer: this loop mirrors the logic you have above but missing this comment:
```
// Note: for this function to not infinite loop we must:
// - If `include_next_batch` returns Some we MUST increase the count of batches that are
//   accounted in the `BACKFILL_BATCH_BUFFER_SIZE` limit in the `matches!` statement of
//   that function.
```
it does seem like `send_batch()` is limited by `BATCH_BUFFER_SIZE` in this case. I think the `while let Some(batch_id)` version is cleaner.

dapplion: Good find! The duplicate code between forward and backfill sync is a bit annoying

beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -121,9 +122,6 @@ pub struct BackFillSync<T: BeaconChainTypes> {
     /// Sorted map of batches undergoing some kind of processing.
     batches: BTreeMap<BatchId, BatchInfo<T::EthSpec, BackFillBatchConfig>>,
 
-    /// List of peers we are currently awaiting a response for.
-    active_requests: HashMap<PeerId, HashSet<BatchId>>,
```
dapplion, line 125: The main point of this PR is getting rid of these `HashMap`s (here + forward sync). They track how many batches each peer is **currently** serving to load balance. On PeerDAS each batch is served by a set of peers, so doing the tracking here becomes more complicated. Plus the set of peer serving a batch will change once we implement individual column retries. So it's just simpler to count active requests in the network context which already tracks active requests per peers.

beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -314,45 +311,11 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
...
-                            batch_id = %id,
-                            error = "no synced peers",
-                            "Batch could not be retried"
-                        );
-                    }
-                } else {
-                    debug!(peer = %peer_id, batch = %id, "Batch not found while removing peer");
-                }
-            }
-        }
```
dapplion, line 354: Disconnection events trigger RPC errros, no need to handle this explicitly now

beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -441,13 +401,7 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
             return Ok(ProcessResult::Successful);
```
jimmygchen, line 399: Not part of this PR, but i think we can remove  the TODO above

beacon_node/network/src/sync/backfill_sync/mod.rs
jimmygchen: Agree too

beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -698,6 +651,8 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
                         );
 
                         for peer in self.participating_peers.drain() {
+                            // TODO(das): `participating_peers` only includes block peers. Should we
+                            // penalize the custody column peers too?
                             network.report_peer(peer, *penalty, "backfill_batch_failed");
```
jimmygchen, line 654: If we can't attribute the failure correctly, do we still want to penalise a random peer? we might end up penalising and lose a good peer.

beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -1115,34 +1030,16 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
...
-                // send the batch
-                self.send_batch(network, batch_id, peer)?;
-            } else {
-                // No more batches, simply stop
-                return Ok(());
-            }
+        // Note: for this function to not infinite loop we must:
+        // - If `include_next_batch` returns Some we MUST increase the count of batches that are
+        //   accounted in the `BACKFILL_BATCH_BUFFER_SIZE` limit in the `matches!` statement of
+        //   that function.
```
jimmygchen, line 1034: I don't really understand what this comment means, could you elaborate further?

beacon_node/network/src/sync/backfill_sync/mod.rs
jimmygchen: It doesn't seem like `participating_peers` is used anymore (i dont see any inserts). Is it still required?

beacon_node/network/src/sync/network_context/custody.rs
```diff
@@ -261,8 +257,12 @@ impl<T: BeaconChainTypes> ActiveCustodyRequest<T> {
                             // requests recently
                             self.failed_peers.contains(peer),
                             // Prefer peers with less requests to load balance across peers
-                            active_requests_by_peer.get(peer).copied().unwrap_or(0),
-                            // Final random factor to give all peers a shot in each retry
+                            active_request_count_by_peer.get(peer).copied().unwrap_or(0)
+                                + columns_to_request_by_peer
+                                    .get(peer)
+                                    .map(|columns| columns.len())
```
jimmygchen: Why do we care about the column count here? It's still a single request to each peer right?

beacon_node/network/src/sync/range_sync/batch.rs
```diff
@@ -302,14 +302,13 @@ impl<E: EthSpec, B: BatchConfig> BatchInfo<E, B> {
     #[must_use = "Batch may have failed"]
     pub fn download_failed(
         &mut self,
-        mark_failed: bool,
+        peer: Option<PeerId>,
```
jimmygchen, line 305: please update the method doc -  `mark_failed` is removed

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -281,11 +254,9 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
 
         // A stream termination has been sent. This batch has ended. Process a completed batch.
         // Remove the request from the peer's active batches
-        self.peers
-            .get_mut(peer_id)
-            .map(|active_requests| active_requests.remove(&batch_id));
 
-        let received = batch.download_completed(blocks)?;
+        // TODO(das): should use peer group here
```
jimmygchen: Be great to add gh issue links to some of these important TODOs, so we know it's tracked - and also get us into a habit of raising issues for TODOs

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -860,13 +826,9 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
...
         peer_id: PeerId,
     ) -> ProcessingResult {
-        // add the peer without overwriting its active requests
-        if self.peers.entry(peer_id).or_default().is_empty() {
-            // Either new or not, this peer is idle, try to request more batches
-            self.request_batches(network)
-        } else {
-            Ok(KeepChain)
-        }
+        self.peers.insert(peer_id);
```
jimmygchen, line 818: function doc above need to be updated

beacon_node/network/src/sync/backfill_sync/mod.rs
jimmygchen: Ah ok, i think I understand now - we need to update the batch state, so that we can filter for the batches that are in the buffer correctly when determining whether to start another batch.

jimmygchen (changes requested):

jimmygchen (changes requested):
@dapplion I'm almost done reviewing but ran out of time today, thought i'd publish the comments anyway and continue again later today or tomorrow.

beacon_node/network/src/sync/network_context.rs
jimmygchen: Nice!

beacon_node/network/src/sync/backfill_sync/mod.rs
jimmygchen: I attempted to add a test for this and was able to trigger a stack overflow when the batch state transition doesn't happen properly. Our state transition validation _should_ prevent this, but I think it would be prudent to add this as an extra defense in case the validation logic changes in the future.
https://github.com/dapplion/lighthouse/pull/42/files

beacon_node/network/src/sync/network_context.rs
```diff
@@ -372,25 +358,100 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
         }
     }
 
+    fn active_request_count_by_peer(&self) -> HashMap<PeerId, usize> {
+        let Self {
```
jimmygchen, line 406: Might be worth mentioning why we use explicit field ignoring here, instead of { .. } - to prevent forgetting to add future request types.

beacon_node/network/src/sync/network_context.rs
```diff
@@ -372,25 +358,100 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+
+        for peer_id in blocks_by_root_requests
+            .iter_request_peers()
+            .chain(blobs_by_root_requests.iter_request_peers())
+            .chain(data_columns_by_root_requests.iter_request_peers())
+            .chain(blocks_by_range_requests.iter_request_peers())
+            .chain(blobs_by_range_requests.iter_request_peers())
+            .chain(data_columns_by_range_requests.iter_request_peers())
+        {
+            *active_request_count_by_peer.entry(peer_id).or_default() += 1;
```
jimmygchen, line 437:
```suggestion
            *active_request_count_by_peer.entry(peer_id).or_default().saturating_add(1);
```

beacon_node/network/src/sync/network_context.rs
```diff
@@ -401,69 +462,93 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+                        .is_custody_peer_of(*column_index, peer)
+                })
+                .map(|peer| {
+                    (
+                        // If contains -> 1 (order after), not contains -> 0 (order first)
+                        peers_to_deprioritize.contains(peer),
+                        // Prefer peers with less overall requests
+                        // Also account for requests that are not yet issued tracked in peer_id_to_request_map
+                        active_request_count_by_peer.get(peer).copied().unwrap_or(0)
+                            + peer_id_to_request_map
```
jimmygchen: We're mixing request count + column count here, i feel like we shouldn't count the columns in the current request, because this could skew the numbers, e.g. if we have 2 active "requests" to peer_1, and now we are requesting 3 columns from peer_2 that has 0 active request, we will prioritise sending to peer_1, which actually has more active requests.

jimmygchen (changes requested):
Hey @dapplion, sorry about the delay with the review. I've added some comments. We should definitely do some testing before merging, but overall looks very good!

beacon_node/network/src/sync/backfill_sync/mod.rs
dapplion: It's used in line 651 of this file. And yes, scoring for custody errors is broken, but we should fix that in another PR

beacon_node/network/src/sync/network_context.rs
dapplion: There's no way this overflows, and we are not using safe math in sync in many places.

beacon_node/network/src/sync/backfill_sync/mod.rs
jimmygchen: What i meant was we never insert into this set. 

> It's used in line 651 of this file. And yes, scoring for custody errors is broken, but we should fix that in another PR

Agree, would you mind adding this to the range sync issue list or a separate issue to track?

beacon_node/network/src/sync/network_context/custody.rs
```diff
@@ -256,9 +252,12 @@ impl<T: BeaconChainTypes> ActiveCustodyRequest<T> {
                             // De-prioritize peers that have failed to successfully respond to
                             // requests recently
                             self.failed_peers.contains(peer),
-                            // Prefer peers with less requests to load balance across peers
-                            active_requests_by_peer.get(peer).copied().unwrap_or(0),
-                            // Final random factor to give all peers a shot in each retry
+                            // Prefer peers with less requests to load balance across peers. We
+                            // We batch requests to the same peer, so count existance in the
```
jimmygchen:
```suggestion
                            // Prefer peers with fewer requests to load balance across peers.
                            // We batch requests to the same peer, so count existence in the
```

beacon_node/network/src/sync/network_context.rs
jimmygchen: Agree with the latter comment (never say never, there's always bugs 😉 ), but fine with this as this is not consensus critical and overflow gets wrapped around in release mode.

jimmygchen (changes requested):
Changes look good. There are 2 failing range sync tests:
https://github.com/sigp/lighthouse/actions/runs/14340199806/job/40197309292

jimmygchen:
@dapplion @pawanjay176 as discussed, we need more testing on this PR, but would like to avoid having too many sync branches. Once this is reviewed, we can merge this to `peerdas-devnet-6` branch for testing all the sync fixes together.

I've updated the base branch for this PR.

jimmygchen:
I still can't figure out why the tests are failing, seems like adding peers doesn't cause a new chain to be created.
I think we'll need this PR to prevent Lighthouse from requesting from peers that are either out of sync or not on the same chain.

dapplion:
Fixed the test by requesting peers from the global peer pool, only synced peers. To request from the SyncingChain pool of peers we need more tweaks that we should do in another PR

jimmygchen (approved):
Changes look good to me.
I'll start testing this while waiting for Pawan's review.

beacon_node/lighthouse_network/src/types/globals.rs
```diff
@@ -206,6 +206,20 @@ impl<E: EthSpec> NetworkGlobals<E> {
             .collect::<Vec<_>>()
     }
 
+    /// Returns true if the peer is known and is a custodial of `column_index`
```
pawanjay176:
```suggestion
    /// Returns true if the peer is known and is a custodian of `column_index`
```

pawanjay176 (approved):
LGTM.
I like the direction of moving the peer selection to the network context. I have tested this in the mainnet context and found no issues. 
I haven't tested this yet on a peerdas network yet. 

Given that we still have a bunch of todos wrt the attribution, I think it might be a good idea to have a `peerdas-syncing` branch that we merge all sync improvements to and test extensively before merging it back to unstable. 
The new branch is only meant as a testing branch for all sync improvements so we still review properly before merging into it. 
 cc @dapplion @jimmygchen what do you guys think?

jimmygchen:
> LGTM. I like the direction of moving the peer selection to the network context. I have tested this in the mainnet context and found no issues. I haven't tested this yet on a peerdas network yet.
> 
> Given that we still have a bunch of todos wrt the attribution, I think it might be a good idea to have a `peerdas-syncing` branch that we merge all sync improvements to and test extensively before merging it back to unstable. The new branch is only meant as a testing branch for all sync improvements so we still review properly before merging into it. cc @dapplion @jimmygchen what do you guys think?

Yeah sounds good to me. I can also confirm this works fine with mainnet. I synced a node from scratch and it completed backfill successfully.

I haven't been able to get it to work under PeerDAS though. I'll check the logs to confirm we haven't broke anything there.

beacon_node/network/src/sync/backfill_sync/mod.rs
dapplion: I removed this line in https://github.com/sigp/lighthouse/pull/7352. Maybe @AgeManning can correct me but I don't see why we have to re-penalize peers when backfill sync fails.

https://github.com/sigp/lighthouse/pull/7352/files#diff-ff78d4cebe3393350c12a12dccf6040a286cba38733b46c8108a38a22b7ada0dR627-L702

jimmygchen:
I couldn't get sync to work on a local devnet, seems to be getting stuck due to metadata being acquired after peer is added to chain:

```
May 06 06:03:14.963 DEBUG Syncing new finalized chain                   id: 1, component: "range_sync"
May 06 06:03:14.963 DEBUG Waiting for peers to be available on sampling column subnets chain: 1service: "range_sync"
May 06 06:03:14.968 DEBUG Finalization sync peer joined                 peer_id: 16Uiu2HAmFWoqu6Y2EFTWzuZtdoupagnacQunLH2HoAkzT8ZuDi9y, component: "range_sync"
May 06 06:03:14.968 DEBUG Adding peer to known chain                    peer_id: 16Uiu2HAmFWoqu6Y2EFTWzuZtdoupagnacQunLH2HoAkzT8ZuDi9y, sync_type: Finalized, id: 1, component: "range_sync"
May 06 06:03:14.968 DEBUG Waiting for peers to be available on sampling column subnets chain: 1service: "range_sync"
May 06 06:03:14.968 DEBUG Waiting for peers to be available on sampling column subnets chain: 1service: "range_sync"
May 06 06:03:14.973 DEBUG Waiting for peers to be available on sampling column subnets chain: 1service: "range_sync"
```

I'm going to try revive #6975 and retest again.

jimmygchen:
Confirm this works on peerdas devnet too with #6975. Merging this now.

PR #6924, opened 2025-02-06
Fix attestation queue length metric

michaelsproul:
## Issue Addressed

We were using the wrong queue length for attestation work event metrics.


jimmygchen (approved):
Nice find!

PR #6926, opened 2025-02-06
Fix aggregate attestation v2 response

### LINKED ISSUE #6925: Electra /eth/v2/validator/aggregate_attestation returning incorrect response
When testing on devnet 6 we have been seeing an incorrect API response from `/eth/v2/validator/aggregate_attestation`. As per [the spec](https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Validator/getAggregatedAttestationV2) we should be receiving the fork `version` in the response. 

I double checked and we do not get the `version` key in the body or the `Eth-Consensus-Version` header in the response.

Lighthouse version:

```
Lighthouse v6.0.1-2193f6a
BLS library: blst
Allocator: jemalloc (4K)
Profile: release
```

Example call:

```shell
curl --header "Content-Type: application/json" --header "Accept: application/json"  --request GET --data '' "http://<lighthouse-node>/eth/v2/validator/aggregate_attestation?slot=12122&attestation_data_root=0x9b00809cfb8251fe3e6f41c9a6d12f0d9e876703c5ecc806177623308562af5e&committee_index=3"
```

Example response:

```json
{
  "data": {
    "aggregation_bits": "0x0000000000000000000000000000000012",
    "data": {
      "slot": "12122",
      "index": "0",
      "beacon_block_root": "0xd209078131a33c70d43aa0c00244bc63b5400e30d35ebfbd93a7f01131589e65",
      "source": {
        "epoch": "377",
        "root": "0x8a3795954ff133866cce8b656b640d388f518e0dac3ef92b9b432f69cf4077ed"
      },
      "target": {
        "epoch": "378",
        "root": "0x8d2b060dcd033e3cc71f4229bd551f0467b3337b1b091a758c4c11bc62fc1d81"
      }
    },
    "signature": "0xa2d81b21125795f12db78413c769ed5668d51a5432829bbfd0c2e238e08ece5cf5f7e03b5b76a700cb583fb06bef56d70de2b4b40150a2d91beb392b4ef84f4b335194d4d864af6382048fbc37ff176483e726e9a56be55ca3391a4d2e56565d",
    "committee_bits": "0x0800000000000000"
  }
}
```

eserilev:
Closes #6925

Ensure that we return a fork versioned response in the v2 case, plus a small refactor

### DIFF
--- a/beacon_node/http_api/src/aggregate_attestation.rs
+++ b/beacon_node/http_api/src/aggregate_attestation.rs
@@ -1,7 +1,8 @@
+use crate::api_types::GenericResponse;
 use crate::unsupported_version_rejection;
 use crate::version::{add_consensus_version_header, V1, V2};
 use beacon_chain::{BeaconChain, BeaconChainTypes};
-use eth2::types::{EndpointVersion, Hash256, Slot};
+use eth2::types::{self, EndpointVersion, Hash256, Slot};
 use std::sync::Arc;
 use types::fork_versioned_response::EmptyMetadata;
 use types::{CommitteeIndex, ForkVersionedResponse};
@@ -53,6 +54,7 @@ pub fn get_aggregate_attestation<T: BeaconChainTypes>(
                     e
                 ))
             })?
+            .map(GenericResponse::from)
             .ok_or_else(|| {
                 warp_utils::reject::custom_not_found("no matching aggregate found".to_string())
             })?;
--- a/beacon_node/http_api/src/aggregate_attestation.rs
+++ b/beacon_node/http_api/src/aggregate_attestation.rs
@@ -1,5 +1,5 @@
 use crate::unsupported_version_rejection;
-use crate::version::{V1, V2};
+use crate::version::{add_consensus_version_header, V1, V2};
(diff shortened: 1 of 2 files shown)


beacon_node/http_api/src/aggregate_attestation.rs
```diff
@@ -0,0 +1,60 @@
...
+            .ok_or_else(|| {
+                warp_utils::reject::custom_not_found("no matching aggregate found".to_string())
+            })?;
+        let fork_name = chain.spec.fork_name_at_slot::<T::EthSpec>(slot);
+        let fork_versioned_response = ForkVersionedResponse {
+            version: Some(fork_name),
+            metadata: EmptyMetadata {},
+            data: aggregate_attestation,
+        };
+        Ok(warp::reply::json(&fork_versioned_response).into_response())
```
michaelsproul: Pardon the lurking.. does this cover the Eth-Consensus-Version header?

eserilev: ah yep, added the header now as well. thanks!

michaelsproul (approved):
Looks good!

KaloyanTanev (approved):

PR #6927, opened 2025-02-06
Simplify into_responses_with_blobs logic

dapplion:
## Issue Addressed

- Pull simplification from https://github.com/sigp/lighthouse/pull/6469 
- Part of https://github.com/sigp/lighthouse/issues/6258

## Proposed Changes

The job of `into_responses_with_blobs` is to match block and blobs with the same block root. We can just use a HashMap and remove 41 lines of code.


realbigsean:
looks like there's a failing test here

pawanjay176 (commented):
Looks good. 
The one issue I see with this is we aren't enforcing the order of the blobs based on the index of the sidecars. 
We will throw the error as `KzgCommitmentMismatch` when trying to build the rpc block. Not sure if this would be an issue.

Edit: Actually the spec requires the sender to send it in the right order, so we were being quite permissive before
> The following blob sidecars, where they exist, MUST be sent in consecutive (slot, index) order

Edit 2: I think it is fine

pawanjay176 (approved):
LGTM

dapplion:
> Actually the spec requires the sender to send it in the right order, so we were being quite permissive before

We can enforce that if we want, but why would we do so if we don't need to? If we had some optimization we can expect that behaviour but I don't see any damage to us from peers sending blobs or blocks out of order

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -84,49 +84,26 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
...
+        let mut blobs_by_block = HashMap::<Hash256, Vec<Arc<BlobSidecar<E>>>>::new();
+        for blob in blobs {
+            let block_root = blob.block_root();
+            blobs_by_block.entry(block_root).or_default().push(blob);
         }
 
-        Ok(responses)
+        // Now collect all blobs that match to the block by block root. BlobsByRange request checks
+        // the inclusion proof so we know that the commitment is the expected.
+        // RpcBlock::new ensures that the count of blobs is consistent with the block
```
jimmygchen, line 102: We don't actually perform the check if the blob list is empty or `None`. 
Seems like it's worth checking in both cases?

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -84,49 +84,26 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
...
+        blocks
+            .into_iter()
+            .map(|block| {
+                let block_root = get_block_root(&block);
+                let max_blobs_per_block = spec.max_blobs_per_block(block.epoch()) as usize;
+                let blobs = blobs_by_block.remove(&block_root).unwrap_or_default();
```
jimmygchen, line 108: The existing code ensures the blobs are in order, and i think our code does assume they are - I think we should make sure they're in order here?

jimmygchen (changes requested):

beacon_node/network/src/sync/block_sidecar_coupling.rs
jimmygchen: Oh sorry just read Pawan's comment above. 

> but I don't see any damage to us from peers sending blobs or blocks out of order

Can you confirm this? IIRC we have some assumptions on the blob order in the code somewhere.

jimmygchen: There's one here but it's not affected by this change, although i feel like we should remove the assumption here
https://github.com/sigp/lighthouse/blob/5f053b0b6dfb7dba8f4455b6fea1d3e6777cea99/beacon_node/beacon_chain/src/kzg_utils.rs#L166-L176

Also the comment here mentions the assumption too: https://github.com/sigp/lighthouse/blob/279270533128ef8069765d04c280db993db12905/beacon_node/beacon_chain/src/block_verification_types.rs#L89-L91

beacon_node/beacon_chain/src/block_verification_types.rs
```diff
@@ -122,23 +122,10 @@ impl<E: EthSpec> RpcBlock<E> {
...
-                let blob_commitment = blob.kzg_commitment;
-                if blob_commitment != block_commitment {
-                    return Err(AvailabilityCheckError::KzgCommitmentMismatch {
-                        block_commitment,
-                        blob_commitment,
-                    });
-                }
-            }
+        if block.num_expected_blobs() != blobs.as_ref().map(|blobs| blobs.len()).unwrap_or(0) {
+            return Err(AvailabilityCheckError::MissingBlobs);
```
dapplion: Now we always check that the blob count is consistent with the block regardless of empty list or None.

beacon_node/beacon_chain/src/block_verification_types.rs
```diff
@@ -122,23 +122,10 @@ impl<E: EthSpec> RpcBlock<E> {
...
-            blobs.as_ref(),
-            block.message().body().blob_kzg_commitments(),
-        ) {
-            if blobs.len() != block_commitments.len() {
-                return Err(AvailabilityCheckError::MissingBlobs);
-            }
-            for (blob, &block_commitment) in blobs.iter().zip(block_commitments.iter()) {
-                let blob_commitment = blob.kzg_commitment;
-                if blob_commitment != block_commitment {
-                    return Err(AvailabilityCheckError::KzgCommitmentMismatch {
```
dapplion, line 134: We don't need to check the commitment equivalence anymore because:
- We check the inclusion proof of each blob
- We match blobs and blocks by block root

Therefore it's impossible for the commitment to not be equivalent

beacon_node/network/src/sync/block_sidecar_coupling.rs
dapplion: Fixed with 16c95f0fe958cace1f0c0f2b5509127af7ac2216

beacon_node/network/src/sync/network_context/requests/blobs_by_range.rs
```diff
@@ -36,12 +36,30 @@ impl<E: EthSpec> ActiveRequestItems for BlobsByRangeRequestItems<E> {
...
+                Ordering::Greater => {
+                    if blob.index != 0 {
+                        return Err(LookupVerifyError::NotSorted("non-zero index at slot start"));
+                    }
+                }
+                Ordering::Equal => {
+                    if blob.index == prev.index {
+                        return Err(LookupVerifyError::DuplicatedData(blob.slot(), blob.index));
+                    } else if blob.index != prev.index + 1 {
+                        return Err(LookupVerifyError::NotSorted("non-sequential indices"));
```
dapplion, line 53: We could group these two conditions into the "non-sequential" one. But I feel it's good to know if a sender is faulty due to duplicate data or non-sequential

beacon_node/network/src/sync/block_sidecar_coupling.rs
dapplion: b983fa9978b53bc85ac3448ed43a5ce200a1c0de enforces ordered responses on all by_range requests

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -84,49 +84,27 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
...
+        // Now collect all blobs that match to the block by block root. BlobsByRange request checks
+        // the inclusion proof so we know that the commitment is the expected.
+        // RpcBlock::new ensures that the count of blobs is consistent with the block
+        blocks
+            .into_iter()
+            .map(|block| {
+                let block_root = get_block_root(&block);
+                let max_blobs_per_block = spec.max_blobs_per_block(block.epoch()) as usize;
+                let mut blobs = blobs_by_block.remove(&block_root).unwrap_or_default();
+                blobs.sort_by(|a, b| a.index.cmp(&b.index));
```
jimmygchen: Given the latest changes to enforce order (https://github.com/sigp/lighthouse/pull/6927/commits/b983fa9978b53bc85ac3448ed43a5ce200a1c0de), blobs that arrive here would always be sorted right?

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -84,49 +84,27 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
...
+                let blobs = RuntimeVariableList::new(blobs, max_blobs_per_block)
+                    .map_err(|_| "Blobs returned exceeds max length".to_string())?;
+                RpcBlock::new(Some(block_root), block, Some(blobs)).map_err(|e| format!("{e:?}"))
+            })
+            .collect::<Result<Vec<_>, _>>()
```
jimmygchen, line 114: Just a note that we no longer return an error for the `Received sidecars that don't pair well` case, and we just drop the extra blob sidecars. Not sure if this is an issue.

beacon_node/network/src/sync/network_context.rs
```diff
@@ -861,6 +861,10 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
             "id" => ?id,
         );
 
+        // Send ascending indices so the DataColumnsByRange request handler expects ascending
+        // responses
+        request.columns.sort();
```
jimmygchen, line 867: We could also just make them sorted when we construct the request, so we don't have to mutate the request here?

jimmygchen (changes requested):

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -84,49 +84,27 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
...
-            .map_err(|_| "Blobs returned exceeds max length".to_string())?;
-            responses.push(RpcBlock::new(None, block, Some(blobs)).map_err(|e| format!("{e:?}"))?)
-        }
-
-        // if accumulated sidecars is not empty, throw an error.
-        if blob_iter.next().is_some() {
-            return Err("Received sidecars that don't pair well".to_string());
+        let mut blobs_by_block = HashMap::<Hash256, Vec<Arc<BlobSidecar<E>>>>::new();
+        for blob in blobs {
+            let block_root = blob.block_root();
```
jimmygchen, line 89: I wonder if we could just use slot here to avoid having to compute root for the block and the header in every blob?

dapplion: I don't like the idea as the sender could send distinct headers with the same slot number. Plus it feels like a minor micro-optimization on the sync path. I don't see how it will have any noticeable impact.

beacon_node/network/src/sync/network_context.rs
dapplion: We still have to call sort in `make_columns_by_range_requests` which isn't super ergonomic either. Now the handler expects a sorted response we should have the logic to ensure a sorted request as close to the handler as possible

beacon_node/network/src/sync/block_sidecar_coupling.rs
dapplion: Added a comment explaining why here https://github.com/sigp/lighthouse/pull/6927/commits/72b90d1a89865cbb48a31573a403b928e058185d

beacon_node/network/src/sync/block_sidecar_coupling.rs
dapplion: Removed in https://github.com/sigp/lighthouse/pull/6927/commits/72b90d1a89865cbb48a31573a403b928e058185d

beacon_node/beacon_chain/src/block_verification_types.rs
```diff
@@ -107,45 +107,24 @@ impl<E: EthSpec> RpcBlock<E> {
...
         // Treat empty blob lists as if they are missing.
         let blobs = blobs.filter(|b| !b.is_empty());
 
-        if let (Some(blobs), Ok(block_commitments)) = (
-            blobs.as_ref(),
-            block.message().body().blob_kzg_commitments(),
-        ) {
-            if blobs.len() != block_commitments.len() {
-                return Err(AvailabilityCheckError::MissingBlobs);
-            }
```
dapplion, line 130: Moved the check to the da_checker, which verifies KZG and ensures all blobs are present. We should revisit this anyway later for
- https://github.com/sigp/lighthouse/issues/7011

beacon_node/beacon_chain/src/block_verification_types.rs
```diff
@@ -107,45 +107,24 @@ impl<E: EthSpec> RpcBlock<E> {
         }
     }
 
-    /// Constructs a new `BlockAndBlobs` variant after making consistency
-    /// checks between the provided blocks and blobs. This struct makes no
-    /// guarantees about whether blobs should be present, only that they are
-    /// consistent with the block. An empty list passed in for `blobs` is
-    /// viewed the same as `None` passed in.
+    /// Constructs a new `BlockAndBlobs` variant.
```
jimmygchen, line 110: This is not true as the it could construct a `RpcBlockInner::Block` variant too.

beacon_node/network/src/sync/block_sidecar_coupling.rs
jimmygchen: yeah true, this should be really  fast given we're computing the root on the block header.

beacon_node/beacon_chain/src/block_verification_types.rs
jimmygchen, line 130: Do we still have a check somewhere that verifies the blob commitments matches the commitments in the block?

jimmygchen (changes requested):
Just a few more questions, I think we're close!
Have you tested this locally on mainnet code paths?

beacon_node/beacon_chain/src/block_verification_types.rs
dapplion: Yes, the inclusion proof asserts that equivalence. That check happens in the sync request handlers or gossip verification

jimmygchen (commented):
Ready for merge pending some mainnet & kurtosis testing.

PR #6928, opened 2025-02-06
Add PeerDAS metrics to track subnets without peers

dapplion:
## Proposed Changes

Currently we track a key metric `PEERS_PER_COLUMN_SUBNET` in a getter `good_peers_on_sampling_subnets`. Another PR https://github.com/sigp/lighthouse/pull/6922 deletes that function, so we have to move the metric anyway. This PR moves that metric computation to the metrics spawned task which is refreshed every 5 seconds.

I also added a few more useful metrics. The total set and intended usage is:

- `sync_peers_per_column_subnet`: Track health of overall subnets in your node
- `sync_peers_per_custody_column_subnet`: Track health of the subnets your node needs. We should track this metric closely in our dashboards with a heatmap and bar plot
- ~~`sync_column_subnets_with_zero_peers`: Is equivalent to the Grafana query `count(sync_peers_per_column_subnet == 0) by (instance)`. We may prefer to skip it, but I believe it's the most important metric as if `sync_column_subnets_with_zero_peers > 0` your node stalls.~~
- ~~`sync_custody_column_subnets_with_zero_peers`: `count(sync_peers_per_custody_column_subnet == 0) by (instance)`~~

### DIFF
--- a/beacon_node/network/src/metrics.rs
+++ b/beacon_node/network/src/metrics.rs
@@ -387,19 +387,6 @@ pub static PEERS_PER_CUSTODY_COLUMN_SUBNET: LazyLock<Result<IntGaugeVec>> = Lazy
         &["subnet_id"],
     )
 });
-pub static COLUMN_SUBNETS_WITH_ZERO_PEERS: LazyLock<Result<IntGauge>> = LazyLock::new(|| {
-    try_create_int_gauge(
-        "sync_column_subnets_with_zero_peers",
-        "Current count of total column subnets with zero peers",
-    )
-});
-pub static CUSTODY_COLUMN_SUBNETS_WITH_ZERO_PEERS: LazyLock<Result<IntGauge>> =
-    LazyLock::new(|| {
-        try_create_int_gauge(
-            "sync_custody_column_subnets_with_zero_peers",
-            "Current count of custody column subnets of this node with zero peers",
-        )
-    });
 pub static SYNCING_CHAINS_COUNT: LazyLock<Result<IntGaugeVec>> = LazyLock::new(|| {
     try_create_int_gauge_vec(
         "sync_range_chains",
@@ -787,7 +774,7 @@ pub fn update_sync_metrics<E: EthSpec>(network_globals: &Arc<NetworkGlobals<E>>)
     let custody_column_subnets = network_globals.sampling_subnets.iter();
 
     // Iterate all subnet values to set to zero the empty entries in peers_per_column_subnet
-    for subnet in all_column_subnets.clone() {
+    for subnet in all_column_subnets {
         set_gauge_entry(
             &PEERS_PER_COLUMN_SUBNET,
             &[&format!("{subnet}")],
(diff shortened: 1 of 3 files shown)


beacon_node/network/src/metrics.rs
```diff
@@ -374,11 +375,31 @@ pub static PEERS_PER_SYNC_TYPE: LazyLock<Result<IntGaugeVec>> = LazyLock::new(||
 });
 pub static PEERS_PER_COLUMN_SUBNET: LazyLock<Result<IntGaugeVec>> = LazyLock::new(|| {
     try_create_int_gauge_vec(
-        "peers_per_column_subnet",
+        "sync_peers_per_column_subnet",
         "Number of connected peers per column subnet",
```
jimmygchen, line 379:
```suggestion
        "Number of synced peers per column subnet",
```

beacon_node/network/src/metrics.rs
```diff
@@ -374,11 +375,31 @@ pub static PEERS_PER_SYNC_TYPE: LazyLock<Result<IntGaugeVec>> = LazyLock::new(||
 });
 pub static PEERS_PER_COLUMN_SUBNET: LazyLock<Result<IntGaugeVec>> = LazyLock::new(|| {
```
jimmygchen, line 376: Should rename this const name too.

beacon_node/network/src/metrics.rs
```diff
@@ -374,11 +375,31 @@ pub static PEERS_PER_SYNC_TYPE: LazyLock<Result<IntGaugeVec>> = LazyLock::new(||
...
+        "Number of connected peers per custody column subnet",
+        &["subnet_id"],
+    )
+});
+pub static COLUMN_SUBNETS_WITH_ZERO_PEERS: LazyLock<Result<IntGauge>> = LazyLock::new(|| {
+    try_create_int_gauge(
+        "sync_column_subnets_with_zero_peers",
+        "Current count of total column subnets with zero peers",
+    )
+});
```
jimmygchen: Can't we just use `sync_peers_per_column_subnet` for this?

and the query:
```
count(sync_peers_per_column_subnet == 0)
```

beacon_node/network/src/metrics.rs
```diff
@@ -374,11 +375,31 @@ pub static PEERS_PER_SYNC_TYPE: LazyLock<Result<IntGaugeVec>> = LazyLock::new(||
...
+pub static CUSTODY_COLUMN_SUBNETS_WITH_ZERO_PEERS: LazyLock<Result<IntGauge>> =
+    LazyLock::new(|| {
+        try_create_int_gauge(
+            "sync_custody_column_subnets_with_zero_peers",
+            "Current count of custody column subnets of this node with zero peers",
+        )
+    });
```
jimmygchen: Same as above, could use `sync_column_subnets_with_zero_peers`?

jimmygchen (changes requested):
Looks good, just added a comment.

beacon_node/network/src/metrics.rs
dapplion: Yeah I mentioned that in the description, we can skip the metric if we are ok doing a Grafana query

jimmygchen (approved):

michaelsproul:
Removing from merge queue temporarily while we merge v7 PRs

michaelsproul:
@mergify dequeue

jimmygchen:
@mergify requeue

jimmygchen:
@mergify requeue

Issue #6854, opened 2025-01-24
Unsubscribe from `blob_sidecars` gossip topic from the Fulu fork


jimmygchen:
## Description

Blob subnets are deprecated from the Fulu fork and Lighthouse should not subscribe to these topics from the Fulu fork.

See spec: https://github.com/ethereum/consensus-specs/blob/dev/specs/fulu/p2p-interface.md#deprecated-blob_sidecar_subnet_id

jimmygchen:
Prior work done by @dapplion here:
https://github.com/sigp/lighthouse/pull/5899/files#diff-667170557450f288a0ee39b61c2b28c01ca910a00244fd396c7c5e17620e44baR123

jimmygchen:
Closed by #6932

PR #5344, opened 2024-03-04
Correct the metrics for topic subscriptions

AgeManning:
Some topic subscription metrics were not being registered as a safety mechanism to prevent us from recording metrics for unbounded number of topics. 

The safety mechanism was too strict. 

This PR increases the number of topics we store metrics for, and registers all the subnets and sync committees specifically to ensure these topics are always correctly accounted for.

PR #6932, opened 2025-02-06
Unsubscribe blob topics at Fulu fork

dapplion:
## Issue Addressed

Addresses #6854.

PeerDAS requires unsubscribing a Gossip topic at a fork boundary. This is not possible with our current topic machinery.

## Proposed Changes

Instead of defining which topics have to be **added** at a given fork, we define the complete set of topics at a given fork. The new start of the show and key function is:

```rust
pub fn core_topics_to_subscribe<E: EthSpec>(
    fork_name: ForkName,
    opts: &TopicConfig,
    spec: &ChainSpec,
) -> Vec<GossipKind> {
    // ...
    if fork_name.deneb_enabled() && !fork_name.fulu_enabled() {
        // All of deneb blob topics are core topics
        for i in 0..spec.blob_sidecar_subnet_count(fork_name) {
            topics.push(GossipKind::BlobSidecar(i));
        }
    }
    // ...
}
```

`core_topics_to_subscribe` only returns the blob topics if `fork < Fulu`. Then at the fork boundary, we subscribe with the new fork digest to `core_topics_to_subscribe(next_fork)`, which excludes the blob topics.

I added `is_fork_non_core_topic` to carry on to the next fork the aggregator topics for attestations and sync committee messages. This approach is future-proof if those topics ever become fork-dependent.

Closes https://github.com/sigp/lighthouse/issues/6854

### DIFF
--- a/beacon_node/lighthouse_network/src/service/mod.rs
+++ b/beacon_node/lighthouse_network/src/service/mod.rs
@@ -289,7 +289,7 @@ impl<E: EthSpec> Network<E> {
                 }
             });
 
-            let max_topics_by_fork = current_and_future_forks
+            let all_topics_for_forks = current_and_future_forks
                 .map(|(fork, fork_digest)| {
                     all_topics_at_fork::<E>(fork, &ctx.chain_spec)
                         .into_iter()
@@ -307,11 +307,11 @@ impl<E: EthSpec> Network<E> {
 
             // For simplicity find the fork with the most individual topics and assume all forks
             // have the same topic count
-            let max_topics_at_any_fork = max_topics_by_fork
+            let max_topics_at_any_fork = all_topics_for_forks
                 .iter()
                 .map(|topics| topics.len())
                 .max()
-                .unwrap_or(0);
+                .expect("each fork has at least 5 hardcoded core topics");
 
             let possible_fork_digests = ctx.fork_context.all_fork_digests();
             let filter = gossipsub::MaxCountSubscriptionFilter {
@@ -356,7 +356,7 @@ impl<E: EthSpec> Network<E> {
             // If we are using metrics, then register which topics we want to make sure to keep
             // track of
             if ctx.libp2p_registry.is_some() {
-                for topics in max_topics_by_fork {
+                for topics in all_topics_for_forks {
                     gossipsub.register_topics_for_metrics(topics);
                 }
             }
@@ -734,8 +734,6 @@ impl<E: EthSpec> Network<E> {
             let topic = GossipTopic::new(kind, GossipEncoding::default(), new_fork_digest);
             self.subscribe(topic);
         }
-
-        // TODO(das): unsubscribe from blob topics at the Fulu fork
 
         // Already registered all possible gossipsub topics for metrics
     }
--- a/beacon_node/lighthouse_network/src/service/mod.rs
+++ b/beacon_node/lighthouse_network/src/service/mod.rs
@@ -14,9 +14,8 @@ use crate::rpc::{
     RequestType, ResponseTermination, RpcErrorResponse, RpcResponse, RpcSuccessResponse, RPC,
 };
 use crate::types::{
-    attestation_sync_committee_topics, fork_core_topics, subnet_from_topic_hash, GossipEncoding,
-    GossipKind, GossipTopic, SnappyTransform, Subnet, SubnetDiscovery, ALTAIR_CORE_TOPICS,
-    BASE_CORE_TOPICS, CAPELLA_CORE_TOPICS, LIGHT_CLIENT_GOSSIP_TOPICS,
+    all_topics_at_fork, core_topics_to_subscribe, is_fork_non_core_topic, subnet_from_topic_hash,
+    GossipEncoding, GossipKind, GossipTopic, SnappyTransform, Subnet, SubnetDiscovery,
 };
 use crate::EnrExt;
 use crate::Eth2Enr;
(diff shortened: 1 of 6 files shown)


beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -332,17 +356,9 @@ impl<E: EthSpec> Network<E> {
...
-                            GossipEncoding::default(),
-                            enr_fork_id.fork_digest,
-                        ))
-                        .into()
-                    })
-                    .collect::<Vec<TopicHash>>();
-                gossipsub.register_topics_for_metrics(topics_to_keep_metrics_for);
+                for topics in max_topics_by_fork {
+                    gossipsub.register_topics_for_metrics(topics);
+                }
```
jimmygchen, line 361: why do we want to register metric for all current + future topics on startup? 

~~This takes up unnecessary extra storage for metric databases and I don't think this actually works because there's a limit on how many unsubscribed topics you can register metric for, and there's 128 topics in PeerDAS.~~
https://github.com/sigp/lighthouse/blob/e7ea69647a4cb686d0c7a80f24b3f34d329a7e01/beacon_node/lighthouse_network/gossipsub/src/metrics.rs#L39-L41

beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -280,14 +279,39 @@ impl<E: EthSpec> Network<E> {
...
+                if fork >= ctx.fork_context.current_fork() {
+                    ctx.fork_context
+                        .to_context_bytes(fork)
+                        .map(|fork_digest| (fork, fork_digest))
+                } else {
+                    None
+                }
+            });
+
+            let max_topics_by_fork = current_and_future_forks
```
jimmygchen: There's no `max` usage here, can we call this `all_topics_for_forks`?

beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -280,14 +279,39 @@ impl<E: EthSpec> Network<E> {
...
+                })
+                .collect::<Vec<_>>();
+
+            // For simplicity find the fork with the most individual topics and assume all forks
+            // have the same topic count
+            let max_topics_at_any_fork = max_topics_by_fork
+                .iter()
+                .map(|topics| topics.len())
+                .max()
+                .unwrap_or(0);
```
jimmygchen: This should be unreachable but I think it's safe to do an `expect` here and exit
If we have 0 topics here something is fundamentally wrong.

beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -700,38 +716,28 @@ impl<E: EthSpec> Network<E> {
...
+        for kind in core_topics_to_subscribe::<E>(
+            new_fork,
             &self.network_globals.as_topic_config(),
+            &self.fork_context.spec,
         ) {
             let topic = GossipTopic::new(kind, GossipEncoding::default(), new_fork_digest);
             self.subscribe(topic);
         }
 
         // TODO(das): unsubscribe from blob topics at the Fulu fork
```
jimmygchen: This TODO can be removed

beacon_node/lighthouse_network/src/service/mod.rs
jimmygchen: Ignore my comment about storage - I don't actually know how the metrics are stored here, will look into it, ~~but I think the comment on max still applies.~~

jimmygchen: Sorry looks like I misunderstood how this works, this function just register it to make it an allowed topic.
And the fork digest label only gets created _if_ the metric is updated for the first time, so it shouldn't take up extra storage if we register future forks early.

Do you know the reason for only allowing sync committee topics previously?

Relevant PR
https://github.com/sigp/lighthouse/pull/5344

beacon_node/lighthouse_network/src/types/topics.rs
```diff
@@ -25,104 +25,110 @@ pub const BLS_TO_EXECUTION_CHANGE_TOPIC: &str = "bls_to_execution_change";
...
+    pub enable_light_client_server: bool,
+    pub subscribe_all_subnets: bool,
     pub subscribe_all_data_column_subnets: bool,
     pub sampling_subnets: &'a HashSet<DataColumnSubnetId>,
 }
 
-/// Returns the core topics associated with each fork that are new to the previous fork
-pub fn fork_core_topics<E: EthSpec>(
-    fork_name: &ForkName,
+/// Returns all the topics the node should subscribe at `epoch`
```
jimmygchen: at `ForkName`

beacon_node/lighthouse_network/src/types/topics.rs
```diff
@@ -496,31 +501,61 @@ mod tests {
...
-        let core_topics = core_topics_to_subscribe::<E>(latest_fork, &spec, &topic_config);
-        // Need to check all the topics exist in an order independent manner
-        for topic in all_topics {
-            assert!(core_topics.contains(&topic));
+            sampling_subnets,
+        }
+    }
+
+    #[test]
+    fn base_topics_are_always_active() {
```
jimmygchen, line 529: Could you add a test to cover all topics for the latest fork? Similar to what the old test does.

jimmygchen (changes requested):
Looks good to me, just a few comments and questions

beacon_node/lighthouse_network/src/service/mod.rs
jimmygchen: Checked with Age and this is fine!

jimmygchen:
@dapplion looks like the new test is failing.

jimmygchen (approved):
The changes look good to me, we can merge once the test is fixed!

jimmygchen (approved):
LGTM!

jimmygchen:
@mergify requeue

PR #6923, opened 2025-02-06
Fuller implementation of `SingleAttestation`

michaelsproul:
## Issue Addressed

Closes:

- https://github.com/sigp/lighthouse/issues/6970

## Proposed Changes

This is severely broken + incomplete. Just opening so I can see the diff.

Issue #6805, opened 2025-01-15
Ensure `SingleAttestation` conversion happens on the `BeaconProcessor`


michaelsproul:
## Description

In our `SingleAttestation` MVP, the conversion from `SingleAttestation` to `Attestation` happens on the main Tokio executor:

- https://github.com/sigp/lighthouse/pull/6488

We should ensure this is optimised out prior to shipping Electra. Some ideas for that here:

- https://github.com/sigp/lighthouse/pull/6488/files#r1915836564

The other option is to more wholeheartedly embrace `SingleAttestation`:

- https://github.com/sigp/lighthouse/pull/6616

michaelsproul:
I'm going to work on a small standalone PR for this.

michaelsproul:
Fixed by:

- https://github.com/sigp/lighthouse/pull/6934

Part of `v7.0.0-beta.0`

PR #6934, opened 2025-02-07
Optimise and refine `SingleAttestation` conversion

michaelsproul:
## Issue Addressed

Closes

- https://github.com/sigp/lighthouse/issues/6805

## Proposed Changes

- Use a new `WorkEvent::GossipAttestationToConvert` to handle the conversion from `SingleAttestation` to `Attestation` _on_ the beacon processor (prevents a Tokio thread being blocked).
- Improve the error handling for single attestations. I think previously we had no ability to reprocess single attestations for unknown blocks -- we would just error. This seemed to be the case in both gossip processing and processing of `SingleAttestation`s from the HTTP API.
- Move the `SingleAttestation -> Attestation` conversion function into `beacon_chain` so that it can return the `attestation_verification::Error` type, which has well-defined error handling and peer penalties. The now-unused variants of `types::Attestation::Error` have been removed.

## Additional Info

The names of variables and variants are quasi-generic so that we can switch around the conversion direction, i.e. `Attestation<E>` -> `SingleAttestation`, and run the whole attestation pipeline on `SingleAttestation` only. This will be more efficient for `SingleAttestation`s as we can postpone the expensive committee lookup until after signature verification (which is the whole point of `SingleAttestation`). I made a start on this here, but it needs more work:

- https://github.com/sigp/lighthouse/pull/6923

See also:

- https://github.com/sigp/lighthouse/pull/6616

### DIFF
--- a/beacon_node/beacon_chain/src/attestation_verification.rs
+++ b/beacon_node/beacon_chain/src/attestation_verification.rs
@@ -60,9 +60,9 @@ use std::borrow::Cow;
 use strum::AsRefStr;
 use tree_hash::TreeHash;
 use types::{
-    Attestation, AttestationRef, BeaconCommittee, BeaconStateError::NoCommitteeFound, ChainSpec,
-    CommitteeIndex, Epoch, EthSpec, Hash256, IndexedAttestation, SelectionProof,
-    SignedAggregateAndProof, SingleAttestation, Slot, SubnetId,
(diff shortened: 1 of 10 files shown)


eserilev (approved):
LGTM!

Would be nice to get a second set of eyes on the review though

pawanjay176 (approved):
Nice and tidy! Tested this on devnet 6 and mainnet, logs and metrics looks good

PR #6936, opened 2025-02-07
Use old geth version due to breaking changes.

jimmygchen:
## Issue Addressed

A temporary workaround for the failing execution tests for geth.

https://github.com/sigp/lighthouse/actions/runs/13192297954

The test is broken due to the following breaking changes in geth that requires updating our tests:
1. removal of `personal` namespace in v1.14.12: See #30704
2. removal of `totalDifficulty` field from RPC in v1.14.11. See #30386.

Using an older version for now (` 1.14.10`) as we need to get things merged for the upcoming release.

Will create a separate issue to fix this.

### DIFF
--- a/testing/execution_engine_integration/src/geth.rs
+++ b/testing/execution_engine_integration/src/geth.rs
@@ -7,7 +7,10 @@ use std::{env, fs};
 use tempfile::TempDir;
 use unused_port::unused_tcp4_port;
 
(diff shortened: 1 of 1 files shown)


jimmygchen:
Issue raised to fix this: https://github.com/sigp/lighthouse/issues/6937

michaelsproul (approved):
LGTM, thanks for fixing

PR #6947, opened 2025-02-07
feat: adds CLI flags to delay publishing for edge case testing on PeerDAS devnets

### LINKED ISSUE #6919: Add CLI flags to delay publishing for edge case testing on PeerDAS devnets
It would be useful to add some CLI flags on the beacon node to test out some unhappy cases on PeerDAS devnets, for example:
- `--delay-block-publishing <seconds>` 
- `--delay-data-column-publishing <seconds>`

These can be used for testing late publishing and recovery from missed/orphaned blocks on devnets.

These should be hidden flags and must be explicitly state as testing only flags.

kamuik16:
## Issue Addressed

Closes #6919 

## Additional Info

- Added two optional config fields (`block_publishing_delay` and `data_column_publishing_delay`) to `ChainConfig`, with defaults set to `None`.

- CLI Flags:
Introduced hidden CLI arguments (`--delay-block-publishing` and `--delay-data-column-publishing`) to set delays (in seconds) for testing purposes.

- ~~For block publishing: Modified the `publish_block_p2p` closure to accept a delay parameter. If set, the closure calls `std::thread::sleep(delay)` before publishing.~~
- For data columns: Before calling `publish_column_sidecars`, the code awaits a `tokio::time::sleep(delay)`.

## Known Limitations

- `block_publishing_delay` only works in the default broadcast validation mode `Gossip`. This is the mode use in most testing. The delay was not added to the other mode to avoid additional complexity.
- `data_column_publishing_delay`: if `block_publishing_delay` is also used, then it will sleep for a minimum of `block_publishing_delay` seconds. This limitation is probably ok as it's not really important enough to worth potentially impacting mainnet code path.

### Reason for Different Sleep Methods:

- ~~`std::thread::sleep`: Used in the synchronous closure `publish_block_p2p` because async closures are unstable and `.await` cannot be used there.~~
- `tokio::time::sleep`: Used in the async part of the function to avoid blocking the executor.

I also might be completely wrong here, do correct me, or feel free to discard the PR if this is not the solution 😄.


beacon_node/http_api/src/publish_blocks.rs
```diff
@@ -103,8 +105,13 @@ pub async fn publish_block<T: BeaconChainTypes, B: IntoGossipVerifiedBlock<T>>(
...
                                   sender,
                                   log,
-                                  seen_timestamp|
+                                  seen_timestamp,
+                                  block_publishing_delay|
           -> Result<(), BlockError> {
+        // Add delay before publishing the block to the network.
+        if let Some(block_publishing_delay) = block_publishing_delay {
+            std::thread::sleep(block_publishing_delay);
+        }
```
ethDreamer: Okay so after messing with this for a while:

> * `std::thread::sleep`: Used in the synchronous closure publish_block_p2p because async closures are unstable and .await cannot be used there.

Yeah I couldn't see an easy way of converting both to an async context. It's a shame because we *are* in an async context already :/

beacon_node/http_api/src/publish_blocks.rs
```diff
@@ -207,6 +217,10 @@ pub async fn publish_block<T: BeaconChainTypes, B: IntoGossipVerifiedBlock<T>>(
     }
 
     if gossip_verified_columns.iter().map(Option::is_some).count() > 0 {
+        // Add delay before publishing the data columns to the network.
+        if let Some(data_column_publishing_delay) = data_column_publishing_delay {
+            tokio::time::sleep(data_column_publishing_delay).await;
+        }
```
ethDreamer, line 236: The timing here is particularly difficult to reason about due to the way this code is structured. It essentially boils down to:
```rust
if validation_level == Gossip {
    sleep(block_publishing_delay)
    publish_block()
    sleep(data_column_publishing_delay)
    publish_data_columns()
} else {
    sleep(data_column_publishing_delay)
    publish_data_columns()
    sleep(block_publishing_delay)
    publish_block()
}
```

this is probably not what you really wanted. A quick look through the code indicates we (at least in every example I saw) call this with `validation_level == Gossip`

beacon_node/src/cli.rs
```diff
@@ -1590,5 +1590,27 @@ pub fn cli_app() -> Command {
...
+            Arg::new("delay-data-column-publishing")
+                .long("delay-data-column-publishing")
+                .value_name("SECONDS") 
+                .action(ArgAction::Set)
+                .help_heading(FLAG_HEADER)
+                .help("TESTING ONLY: Artificially delay data column publishing by the specified number of seconds. \
+                       DO NOT USE IN PRODUCTION.")
+                .hide(true)
+                .display_order(0)
+        )
```
ethDreamer, line 1626: The call to `publish_block` is triggered by the validator client calling `/eth/v2/beacon/blocks` or similar endpoint. Currently this endpoint has a default timeout value of:

[`slot_duration / HTTP_PROPOSAL_TIMEOUT_QUOTIENT`](https://github.com/sigp/lighthouse/blob/v6.0.1/validator_client/src/lib.rs#L316) = `6 seconds`

So there's an implicit requirement that the sum of these two timeouts not exceed that if you don't want to see error messages on the validator client.

jimmygchen: Good point. Discussed with @ethDreamer offline and we're mostly interested in testing delays between the 0-4s mark, so this should be ok for the purpose of testing.

beacon_node/http_api/src/publish_blocks.rs
jimmygchen: Even though this is for testing only, it may significantly impact testing performance, and it may be even worse given we usually run testing under resource constraint environments (e.g. Kurtosis, 4+ nodes on a machine).

I'm thinking maybe we just do the block publishing delay for the `BroadcastValidation::Gossip` case, since we mainly use this in devnet testing?

```rust
    if BroadcastValidation::Gossip == validation_level && should_publish_block {
        // add delay here
        publish_block_p2p(
            block.clone(),
            sender_clone.clone(),
            log.clone(),
            seen_timestamp,
        )
        .map_err(|_| warp_utils::reject::custom_server_error("unable to publish".into()))?;
    }
```

and if we really want to cover the other two broadcast variants, async closure is stabilising in two weeks... 🤩

jimmygchen: ^ As @ethDreamer pointed out, issue with this is it also delays data columns publishing, and i don't see any easy way around it without making this function even more complex - although I think the specific scenario where we want to test block delay without column delay is pretty low value, we can potentially just leave this as a known issue.

beacon_node/http_api/src/publish_blocks.rs
jimmygchen: oops i commented at the wrong place,
https://github.com/sigp/lighthouse/pull/6947#discussion_r1953776534

beacon_node/src/config.rs
```diff
@@ -887,6 +887,14 @@ pub fn get_config<E: EthSpec>(
         .max_gossip_aggregate_batch_size =
         clap_utils::parse_required(cli_args, "beacon-processor-aggregate-batch-size")?;
 
+    if let Some(delay) = clap_utils::parse_optional(cli_args, "delay-block-publishing")? {
+        client_config.chain.block_publishing_delay = Some(Duration::from_secs(delay));
```
jimmygchen: I think we could use `from_secs_f64` here to allow us to play with seconds in floating points?

jimmygchen (changes requested):
Nice work @kamuik16!

Please see comments above, i'm ok with the trade-off of _not_ covering the following to keep things simple (it would be worth documenting in the CLI description though!)
1. **delay publishing blocks without delaying columns** - this scenario is less important, since we expect other nodes in the network to publish columns anyway due to distributed blob publishing.
2. **covering `BroadcastValidation::Consensus` and `ConsensusAndEquivocation::ConsensusAndEquivocation`**: if we just implement it for `Gossip` (which is the main one used in testing), then we shouldn't need the thread::sleep. 

What do you think?

jimmygchen:
@kamuik16 

Any thoughts on the comments?

It would be handy to have this available to test the current active devnet - 
would you have time to work on this in the next week? it's no prob at all if you're busy with other things - let us know and we can continue where you left off ☺️ 

Thanks!

kamuik16:
> @kamuik16
> 
> Any thoughts on the comments?
> 
> It would be handy to have this available to test the current active devnet - would you have time to work on this in the next week? it's no prob at all if you're busy with other things - let us know and we can continue where you left off ☺️
> 
> Thanks!

Hey @jimmygchen and @ethDreamer, thanks for the comments, though I understood, I do not have enough context to do the code changes suggested and it will take some time too. But also, I'm busy w/ other things too, it would be really great if you guys can continue where I left off.

jimmygchen:
No problem, thank you @kamuik16 for your work 🙏  We'll take care of the rest.

kamuik16:
> No problem, thank you @kamuik16 for your work 🙏 We'll take care of the rest.

Thanks and sorry for leaving in between :(

beacon_node/src/config.rs
jimmygchen: Done in https://github.com/sigp/lighthouse/pull/6947/commits/20368f0fc84c5bd07bc89e94cd7ee6cccd61f78f

beacon_node/http_api/src/publish_blocks.rs
jimmygchen: @ethDreamer Please see comment above - I ended up just delaying the `BroadcastValidation::Gossip` code path, using `tokio::time::sleep`
https://github.com/sigp/lighthouse/commit/e3957d30996773e873bd20d69c6705c1443b85ce

jimmygchen:
Hey @ethDreamer,
I've addressed our review comments, would you mind re-reviewing this please?
Thanks!

beacon_node/http_api/src/publish_blocks.rs
ethDreamer: okay so the the logic now reduces to:
```rust
if validation_level == Gossip {
    sleep(block_publishing_delay)
    publish_block()
    sleep(max(data_column_publishing_delay-block_publishing_delay, 0))
    publish_data_columns()
} else {
    sleep(data_column_publishing_delay)
    publish_data_columns()
    publish_block()
}
```
which seems okay for testing but maybe we want to leave a TODO or something to either refactor this or rip this out once we're done testing?

ethDreamer (approved):

beacon_node/http_api/src/publish_blocks.rs
jimmygchen: Added issue here: https://github.com/sigp/lighthouse/issues/6980

jimmygchen:
@mergify requeue

jimmygchen (approved):

PR #6950, opened 2025-02-07
Enable Light Client server by default

eserilev:
Deprecate the `light-client-server` flag, enable light client server by default and introduce a new flag `disable-light-client-server`

### DIFF
--- a/beacon_node/beacon_chain/src/chain_config.rs
+++ b/beacon_node/beacon_chain/src/chain_config.rs
@@ -124,7 +124,7 @@ impl Default for ChainConfig {
             genesis_backfill: false,
             always_prepare_payload: false,
             epochs_per_migration: crate::migrate::DEFAULT_EPOCHS_PER_MIGRATION,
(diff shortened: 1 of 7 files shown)


eserilev:
# Light client dashboard metrics

### Roughly 1 light client data processing request per slot
image

### Light client data computation time: 25ms 
image

###  State merkle proof computation times on the order of milliseconds
image

### No block cache misses in the last 30 days
image

### No light client data computation failures in the last 30 days
image

### Negligible relative increase in db store disk size over last 30 days
image

### No visible increase in CPU/memory usage over the last 30 days

image
image

### No visibile increase in read/write IO operations/Throughput 
image

image

Note that on local testing I also saw negligible increases in cpu usage, IO, memory usage and disk space when running the light client vs not running the light client server

michaelsproul (approved):
Looks great, thanks!

I've just added a WARN log so that users know to remove the old `--light-client-server` flag.

PR #6964, opened 2025-02-10
Allow compilation without LevelDB

michaelsproul:
## Proposed Changes

To enable _pure Rust_ builds of Lighthouse without LevelDB/LMDB/MDBX, this PR turns off LevelDB as a default feature at the `store` dependency level.

Without this change, the `--no-default-features` flag is not sufficient to compile Lighthouse without LevelDB, because it only disables the default features of the top-level crate (`./lighthouse/`). With this change, you can pass `--no-default-features` and get a binary that only has Redb available as an argument to `--beacon-node-backend`.


michaelsproul:
I need to re-enable at least one DB backend for the tests

PR #6969, opened 2025-02-11
Remove AppRequestId from api_types,

jxs:
It seems we are never matching on it to react on it but rather to log `crit!` as those situations should not exist first hand.
`Status` request handling doesnt seem to care if it's `Router` or `Sync`.
Therefore it seems redundant to have it, but please give it a check as I may have missed something

### DIFF
--- a/beacon_node/network/src/router.rs
+++ b/beacon_node/network/src/router.rs
@@ -17,9 +17,8 @@ use futures::prelude::*;
 use lighthouse_network::discovery::ConnectionId;
 use lighthouse_network::rpc::*;
 use lighthouse_network::{
-    rpc,
-    service::api_types::{AppRequestId, SyncRequestId},
-    MessageId, NetworkGlobals, PeerId, PeerRequestId, PubsubMessage, Response,
+    rpc, service::api_types::SyncRequestId, MessageId, NetworkGlobals, PeerId, PeerRequestId,
+    PubsubMessage, Response,
 };
 use logging::TimeLatch;
 use slog::{crit, debug, o, trace};
@@ -62,13 +61,13 @@ pub enum RouterMessage<E: EthSpec> {
     /// An RPC response has been received.
     RPCResponseReceived {
         peer_id: PeerId,
-        request_id: AppRequestId,
+        request_id: SyncRequestId,
         response: Response<E>,
     },
     /// An RPC request failed
     RPCFailed {
         peer_id: PeerId,
-        request_id: AppRequestId,
+        request_id: SyncRequestId,
         error: RPCError,
     },
     /// A gossip message has been received. The fields are: message id, the peer that sent us this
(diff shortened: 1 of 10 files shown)


beacon_node/network/src/router.rs
```diff
@@ -576,15 +578,12 @@ impl<T: BeaconChainTypes> Router<T> {
...
-                request_id,
-                error,
-            });
-        }
+    pub fn on_rpc_error(&mut self, peer_id: PeerId, request_id: SyncRequestId, error: RPCError) {
+        self.send_to_sync(SyncMessage::RpcError {
+            peer_id,
+            request_id,
+            error,
+        });
```
AgeManning, line 586: This looks like one case where we care if the message is from `Sync` or `Router`. 

The logic here I think has changed. Previously an error that occurred due to a Status message would not get forwarded to sync, but now it does. 

This could have follow on effects, but could be handled in sync I guess

beacon_node/network/src/router.rs
```diff
@@ -560,8 +559,11 @@ impl<T: BeaconChainTypes> Router<T> {
     fn send_status(&mut self, peer_id: PeerId) {
         let status_message = status_message(&self.chain);
         debug!(self.log, "Sending Status Request"; "peer" => %peer_id, &status_message);
-        self.network
-            .send_processor_request(peer_id, RequestType::Status(status_message));
+        self.network.inform_network(NetworkMessage::SendRequest {
+            peer_id,
+            request_id: SyncRequestId::Status,
+            request: RequestType::Status(status_message),
```
AgeManning, line 565: This has changed a status message type to be related to sync, I think.

AgeManning (commented):
Yeah nice. 

Looking at this, I think the main logic change is around the Status message. It looks like this PR now sends RPC errors related to status message to Sync. 

We probably should avoid doing that, but it might re-introduce this concept of Router vs Sync in case we want to leave it generic if we add more non-sync rpc things later. 

Not sure, curious about what you think

beacon_node/network/src/router.rs
jxs: but in this case we don't care if it's `Router` or `Sync` right?

beacon_node/network/src/router.rs
jxs: thanks Age, handled with c537b6e704c06dcee7052b7660b89c6b4a62d041

AgeManning:
Yeah. I think these changes look sane. 

Just to be clear tho. We are now assuming every RPC call is related to sync and no longer have a way to distinguish non-sync network messages. I think in the current state it's fine, because Lighthouse isn't really making non-sync messages. 

Although, it looks like there was a panic if we ran this, in the previous version, so it might be worth quickly running this PR to make sure it works as expected. 

An easy way to do it locally is here: https://github.com/sigp/lighthouse/tree/stable/testing/network_testing

AgeManning:
Closing in favour of #7238

Issue #6895, opened 2025-01-31
Range sync stuck due to insufficient peers on data column subnets when triggered


jimmygchen:
## Issue

When a head chain sync starts, if there isn't any peers on data column subnets, it will not progress and gets stuck unless some external events triggers it to resume (e.g. new peer added).

This scenario is easily reproducible with PeerDAS, because we don't immediately know the peer's `custody_group_count` until we get a metadata response back from them.

The sequence of events I observed:
1. Connected to multiple peers with advanced sync state, so we start a new head chain sync (log `New chain added to sync`, `New head chain started syncing`)
2. And immediately, sync thinks there isn't any peers on its custody data column subnets because we haven't obtained their metadata yet and don't know their custody count, therefore no batch gets sent (log`Waiting for peers to be available on custody column subnets`)
3. Later, we obtained their metadata (log `Obtained peer's metadata`), but this doesn't re-trigger range sync, so we're stuck until finalized sync kicks in.

A few possible solutions (not mutually exclusive):
1. Handle this in range request decoupling properly (#6258)
2. Compute `peer_info.custody_subnets` when peer is connected, and using the minimum custody requirement, as every peer must serve the minimum required column count - this way we're likely to have some peers in data column subnets even before obtaining metadata.
3. Update sync when we obtain peer metadata, and trigger resume

## Additional Info

Range sync currently rely on this check `syncing_chain.good_peers_on_sampling_subnets` before requesting batches from peer.
1.  https://github.com/sigp/lighthouse/blob/7d54a43243905b62e1ced8f56cd5ad0575b8638b/beacon_node/network/src/sync/range_sync/chain.rs#L1075
2. https://github.com/sigp/lighthouse/blob/7d54a43243905b62e1ced8f56cd5ad0575b8638b/beacon_node/network/src/sync/range_sync/chain.rs#L444-L448
3.  https://github.com/sigp/lighthouse/blob/7d54a43243905b62e1ced8f56cd5ad0575b8638b/beacon_node/network/src/sync/range_sync/chain.rs#L1175-L1181

[...]

PR #6975, opened 2025-02-11
Retry custody requests after peer metadata updates

dapplion:
## Issue Addressed

Closes https://github.com/sigp/lighthouse/issues/6895

We need sync to retry custody requests when a peer CGC updates. A higher CGC can result in a data column subnet peer count increasing from 0 to 1, allowing requests to happen.

## Proposed Changes

Add new sync event `SyncMessage::UpdatedPeerCgc`. It's sent by the router when a metadata response updates the known CGC


beacon_node/lighthouse_network/src/peer_manager/mod.rs
```diff
@@ -732,12 +733,16 @@ impl<E: EthSpec> PeerManager<E> {
...
+                .and_then(|meta_data| meta_data.custody_group_count().copied().ok());
+
             let custody_group_count_opt = meta_data.custody_group_count().copied().ok();
             peer_info.set_meta_data(meta_data);
 
             if self.network_globals.spec.is_peer_das_scheduled() {
-                // Gracefully ignore metadata/v2 peers. Potentially downscore after PeerDAS to
-                // prioritize PeerDAS peers.
+                // Gracefully ignore metadata/v2 peers.
+                // TODO(das) Potentially downscore after PeerDAS to prioritize PeerDAS peers.
```
jimmygchen: This should not be possible once peerdas is scheduled, we only send out v3 requests, so maybe this TODO can be removed:

https://github.com/sigp/lighthouse/blob/4894cc2ec7ee8fdc4042c3e53246a4b3534df2a6/beacon_node/lighthouse_network/src/service/mod.rs#L1166-L1174

beacon_node/network/src/sync/manager.rs
```diff
@@ -483,6 +486,13 @@ impl<T: BeaconChainTypes> SyncManager<T> {
         }
     }
 
+    fn updated_peer_cgc(&mut self, _peer_id: PeerId) {
```
jimmygchen, line 482: I think we want to resume by range request as well?

jimmygchen (changes requested):
Overall looks good to me! I think we need to attempt to make progress on range sync as well?

beacon_node/network/src/sync/manager.rs
dapplion: Fixed, I resume range sync aswell

jimmygchen: Nice, I'll run some tests today to confirm this fixes the issue.

jimmygchen (approved):

beacon_node/network/src/sync/manager.rs
jimmygchen: Doesn't look like this fixes the issue. I still don't see range requests until a few minutes later until we add a new finalized chain.

Right after startup, waiting for custody peers
```
Feb 12 04:33:07.181 DEBG Waiting for peers to be available on sampling column subnets, chain: 1, service: range_sync, service: sync, module: network::sync::range_sync::chain:1057
```

Got peer metadata response after 15s
```
Feb 12 04:33:22.271 DEBG Obtained peer's metadata, new_seq_no: 6, peer_id: 16Uiu2HAmJyaVGkRGR9ACqompkoge8T2x4KFH4KbzDkh7zz6uN2JX, service: libp2p, module: lighthouse_network::peer_manager:732
```

No range requests until ~5 mins later
```
Feb 12 04:38:07.303 DEBG Finalization sync peer joined, peer_id: 16Uiu2HAmJyaVGkRGR9ACqompkoge8T2x4KFH4KbzDkh7zz6uN2JX, service: range_sync, service: sync, module: network::sync::range_sync::range:143
Feb 12 04:38:07.305 DEBG New chain added to sync, id: 2, from: 38, to: 1071, end_root: 0x3be00d7ce6e52f7938fd588d909055f72469d0f09ce545d7b23077f2d6b40e8a, current_target: 38, batches: 0, peers: 1, state: Stopped, sync_type: Finalized, peer_id: 16Uiu2HAmJyaVGkRGR9ACqompk
oge8T2x4KFH4KbzDkh7zz6uN2JX, service: range_sync, service: sync, module: network::sync::range_sync::chain_collection:506
Feb 12 04:38:07.306 DEBG Sync RPC request sent, id: 4/3/RangeSync/39/1, peer: 16Uiu2HAmAAZ5wP6fvpe1b9tWNmgA2Wn8MsrNYVhkw5WohcAoaHKR, epoch: 39, slots: 32, method: BlocksByRange, service: sync, module: network::sync::network_context:788
Feb 12 04:38:07.307 DEBG Sync RPC request sent, id: 5/3/RangeSync/39/1, peer: 16Uiu2HAm9PijSZpm5QUphXRoBtkhUZPkGJ4Rgxk4Bny91oZPYZLG, columns: [74, 30, 39, 19, 63, 41, 52, 47, 58], epoch: 39, slots: 32, method: DataColumnsByRange, service: sync, module: network::sync::network_context:870
```

dapplion: Did you saw this log? `Received updated peer CGC message`

jimmygchen: I don't recall seeing this. I think I was using the right locally-built image, but can be worth re-testing to confirm if you have time.

jimmygchen: @dapplion could you retest this change?

jimmygchen: I've retested this, and it doesn't seem to trigger retry after obtaining the peers metadata, and the `Received updated peer CGC message` log was not observed.

```
May 06 07:16:46.854 DEBUG Waiting for peers to be available on custody column subnets chain: 1service: "range_sync"
...
# obtained peers metadata
May 06 07:17:01.389 DEBUG Obtained peer's metadata                      peer_id: 16Uiu2HAkz7SLRbDFNscs6RFDrD37oB5yCR9UaBcVpAiy1G54BuQW, new_seq_no: 6, service: "network"
May 06 07:17:01.393 DEBUG Obtained peer's metadata                      peer_id: 16Uiu2HAmKjMmG6VJG2oi5Gd7x4iszsv4Fm9JZWeEad62kTVF9dke, new_seq_no: 132, service: "network"
...
# no range request until ~5 mins later when a new chain is added
May 06 07:21:46.397 DEBUG New chain added to sync                       peer_id: "16Uiu2HAkz7SLRbDFNscs6RFDrD37oB5yCR9UaBcVpAiy1G54BuQW", sync_type: Finalized, id: 5, start_epoch: 0, target_head_slot: 193, target_head_root: 0xc1802494a0935c8bee449a412f982cc1903ec791565d
9cf9eaab233f2cd2bc90, component: "range_sync"
May 06 07:21:46.397 DEBUG Sync RPC request sent                         method: "DataColumnsByRange", slots: 32, epoch: 1, columns: [8, 69, 99, 90, 9, 84, 24, 71, 12, 121, 50, 68, 127, 66, 36, 26, 41, 107, 47, 52, 108, 98, 70, 100, 54, 35, 21, 60, 49, 120, 10, 72, 44, 93, 67, 0, 122, 19, 43, 62, 97, 115, 59, 95, 48, 11, 101, 116, 34, 20, 25, 18, 3, 124, 110, 57, 46, 83, 77, 105, 81, 106, 1, 45, 104, 22, 29, 53, 6, 32, 13, 119, 4, 78, 63, 86, 92, 7, 114, 73, 16, 17, 109, 28, 75, 102, 80, 40, 79, 51, 125, 38, 85, 89, 42, 39, 126, 113, 87, 96, 37, 88, 64, 112, 14, 118, 76, 117, 58, 82, 27, 94, 74, 23, 30, 111, 15, 2, 123, 33, 55, 5, 65, 91, 31, 56], peer: 16Uiu2HAmKjMmG6VJG2oi5Gd7x4iszsv4Fm9JZWeEad62kTVF9dke, id: 8/6/RangeSync/1/1, chain: 1, service: "range_sync"
May 06 07:21:46.397 DEBUG Sync RPC request sent                         method: "DataColumnsByRange", slots: 32, epoch: 1, columns: [61, 103], peer: 16Uiu2HAkz7SLRbDFNscs6RFDrD37oB5yCR9UaBcVpAiy1G54BuQW, id: 9/6/RangeSync/1/1, chain: 1, service: "range_sync"

File: beacon_node/lighthouse_network/src/service/mod.rs [hunk 3]
```diff
@@ -1858,9 +1858,11 @@ impl<E: EthSpec> Network<E> {
                         None
                     }
                     RpcSuccessResponse::MetaData(meta_data) => {
-                        self.peer_manager_mut()
+                        let updated_cgc = self
+                            .peer_manager_mut()
                             .meta_data_response(&peer_id, meta_data.as_ref().clone());
-                        None
+                        // Send event after calling into peer_manager so the PeerDB is updated
+                        self.build_response(id, peer_id, Response::MetaData(meta_data, updated_cgc))
```
jimmygchen: Found out why this doesn't work - `self.build_response` here always return `None` because the `AppRequestId` for `MetaData` is always `AppRequestId::Internal`:

https://github.com/sigp/lighthouse/blob/ba0b62ba175b45e6f050d0ce88c213262de87c50/beacon_node/lighthouse_network/src/service/mod.rs#L1387-L1398

jimmygchen: Addressed here:
https://github.com/sigp/lighthouse/pull/6975/commits/aab502f8a3d6eb17ea88b4ec09cd22472b916fdf

beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -1858,9 +1858,18 @@ impl<E: EthSpec> Network<E> {
...
-                        None
+                        self.
+                        // Send event after calling into peer_manager so the PeerDB is updated.
+                        // Manually build the response here, as we want to propagate the METADATA
+                        // message upwards, even though it's initiated from network.
+                        Some(NetworkEvent::ResponseReceived {
+                            peer_id,
+                            app_request_id: id,
+                            response: Response::MetaData(meta_data, updated_cgc),
+                        })
```
jimmygchen: I opted for manually building the response here. The alternative to change id of METADATA requests from `AppRequestId::Internal` to `AppRequestId::Router` is simpler and achieves the same thing, but:
1. it's technically incorrect , as the `MetaData` request is initiated from network / behaviour, not the router.
2. It has a side effect - the RPC error will also be propagated upwards:
https://github.com/sigp/lighthouse/blob/aab502f8a3d6eb17ea88b4ec09cd22472b916fdf/beacon_node/lighthouse_network/src/service/mod.rs#L1686-L1695
Which _might_ be fine given they're ignored:
https://github.com/sigp/lighthouse/blob/ba0b62ba175b45e6f050d0ce88c213262de87c50/beacon_node/network/src/router.rs#L526

@AgeManning @jxs any thoughts?

beacon_node/network/src/sync/manager.rs
jimmygchen: See https://github.com/sigp/lighthouse/pull/6975#discussion_r2075498456

beacon_node/lighthouse_network/src/service/mod.rs
jimmygchen: Confirmed this works - sync immediately resume range sync after updating peer CGC: 
```
May 06 23:29:04.641 DEBUG Obtained peer's metadata                      peer_id: 16Uiu2HAkvNBQcvufrNw5qnt3ALxYQaRm1127L9iuXYXkcHXTGu7U, new_seq_no: 132, service: "network"
May 06 23:29:04.641 DEBUG Received updated peer CGC message             peer_id: PeerId("16Uiu2HAkvNBQcvufrNw5qnt3ALxYQaRm1127L9iuXYXkcHXTGu7U"), service: "sync"
May 06 23:29:04.641 DEBUG Sync RPC request sent                         method: "BlocksByRange", slots: 32, epoch: 1, peer: 16Uiu2HAm4NasLfUMfU3cfsMrbSrLTkwL1RJaWjKC9sX8iEh92s54, id: 6/5/RangeSync/1/1, chain: 1, service: "range_sync"
May 06 23:29:04.641 DEBUG Sync RPC request sent                         method: "DataColumnsByRange", slots: 32, epoch: 1, columns: [126, 21, 102, 100, 40, 3, 95, 68], peer: 16Uiu2HAkvNBQcvufrNw5qnt3ALxYQaRm1127L9iuXYXkcHXTGu7U, id: 7/5/RangeSync/1/1, chain: 1, service: "range_sync"
```

pawanjay176: Is there any use for actually passing the full metadata to the router in the future? 
From what I understand, we just need to intimate sync that the peer_id has an updated CGC. So a potentially easier way might be to create a `RouterMessage::PeerMetadataUpdated(PeerId)` and pass that to the router.
It just feels a bit redundant to pass the metadata to the router if it hasn't actually updated.

jimmygchen: Yeah I agree - i think we still need to propagate an event up so `NetworkService` can send it to `Router`. Could potentially add a `NetworkEvent::PeerMetadataUpdated` so it doesn't introduce the RPC response side effects?

pawanjay176: sounds good 👍

AgeManning (commented):
From what I can see here, you only care if the custody columns get updated. 

I think we should try and maintain all our internal consistencies, i.e Internal Requests stay internal else it can get very confusing. 

The solution you have here has redundant cases, like if its not updated, we send the event up to the router which then throws it away. 

To keep everything consistent, I think we should do the following:
1. Add a new event to `NetworkEvent` i.e
`NetworkEvent::PeerUpdatedCustodyGroupCount(PeerId)`
2. In the handling of a metadata response, if the cgc is updated, then return the above event
3. Handle this event and send to sync 

This has the benefits of:
1. Not sending redundant messages with redundant information (i.e the original metadata)
2. We maintain the logic that internal messages stay internal
3. I think its a bit clearer the purpose of the message and why its there.

jimmygchen:
Thanks @pawanjay176 @AgeManning ! yeah make sense, I've updated the code as per your suggestions. I think it's much cleaner.

dapplion (commented):
✅  This approach looks much cleaner! Thanks @AgeManning

pawanjay176 (approved):
Nice!

jimmygchen:
@mergify requeue

PR #6984, opened 2025-02-11
Ensure `GET v2/validator/aggregate_attestation` is backwards compatible

### LINKED ISSUE #6983: Electra /eth/v2/validator/aggregate_attestation returning 404s for DENEB
When testing the current proposed lighthouse version for Electra on holesky (deneb) we always get a 404 when calling `/eth/v2/validator/aggregate_attestation`. 

Lighthouse version:

```
Lighthouse v6.0.1-62a0f25
BLS library: blst
Allocator: jemalloc (4K)
Profile: release
```

Example curl:

```shell
curl --header "Content-Type: application/json" --header "Accept: application/json"  --request GET --data '' "http://<lighthouse-node>/eth/v2/validator/aggregate_attestation?slot=3613623&attestation_data_root=0x25cf9b33dbb83a06ee303595348e984d2bcda2dca31658cb3d8632fed339ffbe&committee_index=32"
```

Example response:

```json
{
  "code": 404,
  "message": "NOT_FOUND: no matching aggregate found",
  "stacktraces": []
}
```

eserilev:
## Issue Addressed

Closes #6983

## Proposed Changes

`GET v2/validator/aggregate_attestation` is not backwards compatible. It only works for post electra attestations. This PR adds backwards compatibility and additional test coverage. We should include this in the upcoming 7.0 beta release if possible

### DIFF
--- a/beacon_node/http_api/src/aggregate_attestation.rs
+++ b/beacon_node/http_api/src/aggregate_attestation.rs
@@ -19,22 +19,41 @@ pub fn get_aggregate_attestation<T: BeaconChainTypes>(
     chain: Arc<BeaconChain<T>>,
 ) -> Result<Response<Body>, warp::reject::Rejection> {
     if endpoint_version == V2 {
-        let Some(committee_index) = committee_index else {
-            return Err(warp_utils::reject::custom_bad_request(
-                "missing committee index".to_string(),
-            ));
+        let fork_name = chain.spec.fork_name_at_slot::<T::EthSpec>(slot);
+        let aggregate_attestation = if fork_name.electra_enabled() {
+            let Some(committee_index) = committee_index else {
+                return Err(warp_utils::reject::custom_bad_request(
+                    "missing committee index".to_string(),
+                ));
+            };
+            chain
+                .get_aggregated_attestation_electra(slot, attestation_data_root, committee_index)
+                .map_err(|e| {
+                    warp_utils::reject::custom_bad_request(format!(
+                        "unable to fetch aggregate: {:?}",
+                        e
+                    ))
+                })?
+                .ok_or_else(|| {
+                    warp_utils::reject::custom_not_found("no matching aggregate found".to_string())
+                })?
+        } else {
+            chain
+                .get_pre_electra_aggregated_attestation_by_slot_and_root(
+                    slot,
+                    attestation_data_root,
+                )
+                .map_err(|e| {
+                    warp_utils::reject::custom_bad_request(format!(
+                        "unable to fetch aggregate: {:?}",
+                        e
+                    ))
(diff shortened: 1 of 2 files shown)


pawanjay176 (commented):
This looks good to me.
I tested it by using v2 for both deneb and electra by modifying the lighthouse VC here https://github.com/sigp/lighthouse/blob/431dd7c39828e97a1e1f59238cc4ecd8a011dade/validator_client/validator_services/src/attestation_service.rs#L564-L575

It works as expected. Wondering if we v2 all across.

beacon_node/http_api/src/aggregate_attestation.rs
```diff
@@ -18,13 +18,14 @@ pub fn get_aggregate_attestation<T: BeaconChainTypes>(
     endpoint_version: EndpointVersion,
     chain: Arc<BeaconChain<T>>,
 ) -> Result<Response<Body>, warp::reject::Rejection> {
-    if endpoint_version == V2 {
+    let fork_name = chain.spec.fork_name_at_slot::<T::EthSpec>(slot);
```
pawanjay176, line 21: The spec requires a `Eth-Consensus-Version` header for V2. 
Wondering if we are supposed to derive the fork name from there instead of the BN.
The spec isn't very clear on that imo

michaelsproul: I think deriving it from the slot and ignoring the header is fine. If they're inconsistent it will be a result of misconfiguration. We could add a `WARN` log if this is detected, but I think this misconfiguration would be detected by the VC separately anyway (in `is_compatible`).

beacon_node/http_api/tests/tests.rs
```diff
@@ -3555,54 +3555,51 @@ impl ApiTester {
...
-                .attestations()
-                .next()
-                .unwrap()
-                .clone_as_attestation();
+        assert_eq!(result, expected);
+
+        self
+    }
+
+    #[allow(clippy::await_holding_lock)] // This is a test, so it should be fine.
```
michaelsproul: I actually think we should avoid doing this, as the tests deadlocking could be a source of flaky CI

michaelsproul: Maybe we can collect the tuples of `(Slot, Hash256, u64)` in a vec and then drop the read lock

michaelsproul (approved):

michaelsproul:
> It works as expected. Wondering if we v2 all across.

I think at this late stage I'm fine with sticking to v1 prior to Electra. We could consider changing this for v7.0.0 proper though

michaelsproul:
@mergify queue

PR #6986, opened 2025-02-11
Penalize peers that send an invalid rpc request

ackintosh:
## Issue Addressed

Since https://github.com/sigp/lighthouse/pull/6847, invalid `BlocksByRange`/`BlobsByRange` requests, which do not comply with the spec, are [handled in the Handler](https://github.com/sigp/lighthouse/blob/3d16d1080f5b93193404967dcb5525fa68840ea0/beacon_node/lighthouse_network/src/rpc/handler.rs#L880-L911). Any peer that sends an invalid request is penalized and disconnected.

However, other kinds of invalid rpc request, which result in decoding errors, are just dropped. No penalty is applied and the connection with the peer remains.

## Proposed Changes

I have added handling for the `ListenUpgradeError` event to notify the application of an `RPCError:InvalidData` error and disconnect to the peer that sent the invalid rpc request.

I also added tests for handling invalid rpc requests.


beacon_node/lighthouse_network/src/rpc/handler.rs
```diff
@@ -837,6 +837,20 @@ where
...
+                    self.current_inbound_substream_id.0 += 1;
+
+                    self.events_out.push(HandlerEvent::Err(HandlerErr::Inbound {
+                        id: inbound_substream_id,
+                        proto: Protocol::DataColumnsByRange, // FIXME: replace this hardcoded protocol
+                        error: e.error,
+                    }));
+                    self.shutdown(None);
+                }
+            }
```
ackintosh: > // FIXME: replace this hardcoded protocol

@jxs 

The `RPCError::InvalidData` error is emitted by the inbound codec [here](https://github.com/sigp/lighthouse/blob/70194dfc6a3f4d10c9059610f889ff5a4e863a6a/beacon_node/lighthouse_network/src/rpc/codec.rs#L166-L171). The problem I'm facing is that the handler cannot determine which  protocol the invalid data belongs to. I think changing the type of [`InboundUpgrade::Error`](https://github.com/sigp/lighthouse/blob/70194dfc6a3f4d10c9059610f889ff5a4e863a6a/beacon_node/lighthouse_network/src/rpc/protocol.rs#L664) from `RPCError` to `(Protocol, RPCError)` might solves this. What do you think?

ackintosh: Updated in https://github.com/sigp/lighthouse/pull/6986/commits/2ff8cf2c123488fd1b3b300df410d25529350264

beacon_node/lighthouse_network/src/rpc/handler.rs
```diff
@@ -892,6 +892,20 @@ where
             ConnectionEvent::DialUpgradeError(DialUpgradeError { info, error }) => {
                 self.on_dial_upgrade_error(info, error)
             }
+            ConnectionEvent::ListenUpgradeError(e) => {
+                if matches!(e.error.1, RPCError::InvalidData(_)) {
```
jxs: you don't need the nested matches Akihito, you can do:
```suggestion
            ConnectionEvent::ListenUpgradeError(ListenUpgradeError {
                error: (_, RPCError::IoError(_)),
                ..
            }) => {
```

beacon_node/lighthouse_network/src/rpc/protocol.rs
```diff
@@ -652,7 +652,7 @@ where
     E: EthSpec,
 {
     type Output = InboundOutput<TSocket, E>;
-    type Error = RPCError;
+    type Error = (Protocol, RPCError);
```
jxs, line 678: why use a tuple instead of putting the `Protocol` inside `RPCError`?

jxs (commented):
Hi Akihito, overall this looks good to me, sorry for the delay in the review, left some comments

beacon_node/lighthouse_network/src/rpc/protocol.rs
ackintosh: I think [`RPCError`](https://github.com/sigp/lighthouse/blob/ca6fe1ba92f7c9956213e01ee3934e449772711b/beacon_node/lighthouse_network/src/rpc/protocol.rs#L901-L926) can be used independently in other contexts without being tied to a specific protocol. If we embedded `Protocol` into each variant, we'd need separate error types for different contexts.

beacon_node/lighthouse_network/src/rpc/handler.rs
ackintosh: Refactored in https://github.com/sigp/lighthouse/pull/6986/commits/7e09c470c720129ce18a18177c0f43deab8040db

beacon_node/lighthouse_network/src/rpc/protocol.rs
ackintosh: I saw a struct that has `Protocol` and `RPCError` as separate fields like this, and I thought it was a reasonable way to represent what the error is and where it happened separately:

https://github.com/sigp/lighthouse/blob/c56c63e6d24681ad4f91ddafb809f430bbf07c89/beacon_node/lighthouse_network/src/rpc/handler.rs#L938-L946

jimmygchen:
@ackintosh @jxs is this PR still relevant?

jxs (approved):
Thanks or the ping @jimmygchen LGTM

jimmygchen (approved):
I've reviewed the latest test fix, looks good and is passing 🎉

ackintosh:
@jxs @jimmygchen Thanks for your review. ✨

beacon_node/lighthouse_network/src/rpc/handler.rs
```diff
@@ -888,6 +889,21 @@ where
             ConnectionEvent::DialUpgradeError(DialUpgradeError { info, error }) => {
                 self.on_dial_upgrade_error(info, error)
             }
+            ConnectionEvent::ListenUpgradeError(ListenUpgradeError {
+                error: (proto, error),
+                ..
+            }) if matches!(error, RPCError::InvalidData(_)) => {
+                // Peer is not complying with the protocol. Notify the application and disconnect.
+                let inbound_substream_id = self.current_inbound_substream_id;
+                self.current_inbound_substream_id.0 += 1;
```
pawanjay176: Just trying to understand why we need to increment the id here

beacon_node/lighthouse_network/src/rpc/handler.rs
```diff
@@ -888,6 +889,21 @@ where
...
+
+                self.events_out.push(HandlerEvent::Err(HandlerErr::Inbound {
```
pawanjay176, line 896: IIUC, this error gets pushed up the stack such that we get a 
`DEBUG RPC Error` log? If we don't, can we add a log to indicate this?
would have helped catching issues like here https://github.com/sigp/lighthouse/pull/7928

jimmygchen:
Thanks for the extra review @pawanjay176 
@ackintosh I'm not sure why the CLA check wasn't triggered, so it didn't satisfiy mergify condition, might have to find a way to trigger it, or re-open this PR :/

michaelsproul:
This might hack the CLA thing: https://cla-assistant.io/check/sigp/lighthouse?pullRequest=6986

jimmygchen:
Nice, looks like it's there now!

beacon_node/lighthouse_network/src/rpc/handler.rs
ackintosh: I intended to avoid reusing the same SubstreamId for another substream in [`on_fully_negotiated_inbound()`](https://github.com/sigp/lighthouse/blob/8faabe8197429b062b45f102315b2fca7aa3f02f/beacon_node/lighthouse_network/src/rpc/handler.rs#L920). But after reviewing your comment, I realized this increment isn't necessary since [we call `shutdown()`](https://github.com/sigp/lighthouse/blob/8faabe8197429b062b45f102315b2fca7aa3f02f/beacon_node/lighthouse_network/src/rpc/handler.rs#L905) immediately afterward. I will remove the increment. 🙏

beacon_node/lighthouse_network/src/rpc/handler.rs
ackintosh: Yes, that's correct. The error is propagated to PeerManager, and we see a [`RPC Error`](https://github.com/sigp/lighthouse/blob/ee1b6bc81bb4b7dafb95a95eabcfcaf6b76cea04/beacon_node/lighthouse_network/src/peer_manager/mod.rs#L525) log.

pawanjay176: Awesome, that should help us catching issues with rpc requests much easier. Thanks!

beacon_node/lighthouse_network/src/rpc/handler.rs
ackintosh: Removed in https://github.com/sigp/lighthouse/pull/6986/commits/6848e144bb10f89017286c16286b54528d9c38fc

pawanjay176 (approved):
LGTM.

jimmygchen:
@mergify requeue

beacon_node/lighthouse_network/tests/rpc_tests.rs
```diff
@@ -1732,3 +1736,175 @@ fn test_active_requests() {
         }
     })
 }
+
+#[test]
+fn test_request_too_large_blocks_by_range() {
```
jimmygchen, line 1794: looks like this test is now failing on CI
https://github.com/sigp/lighthouse/actions/runs/17784671968/job/50549871817

jimmygchen: passed in ubuntu, might jsut be flaky

ackintosh: I was able to reproduce the failure on my local machine, and I'm investigating it.

ackintosh: This test checks that when a node receives an invalid `BlobsByRange` request exceeding the maximum count, it bans the sender.

In most cases, on the sender side:

1. The handler notices that the substream has been closed, and [emits a `ConnectionHandlerEvent` (`RPCReceived::EndOfStream`)](https://github.com/sigp/lighthouse/blob/6848e144bb10f89017286c16286b54528d9c38fc/beacon_node/lighthouse_network/src/rpc/handler.rs#L760-L765).
2. After that, the handler shuts down due to the disconnect.

The sequence diagram below shows these details.  
The test expects the following: we receive an `EndOfStream` (`BlockByRange(None)`) response and a  `NetworkEvent::PeerDisconnected` event. 

image

However, as shown in the other diagram below, it seems that due to a race condition, the handler occasionally shuts down before it notices the substream disconnect. When this happens, the test fails with `Future timed out` error because no `EndOfStream` (`BlockByRange(None)`) response is received.

image

ackintosh: I think the missing `EndOfStream` response shown in the second diagram is not a problem, since the connection (not the subreams) has already been closed. From the sync's point of view, [the ongoing syncs are removed because of PeerDisconnected event](https://github.com/sigp/lighthouse/blob/522bd9e9c6ac167f2231525e937c9ebbcb86cf6e/beacon_node/network/src/sync/manager.rs#L488-L502).

ackintosh: So I think we should fix the test to check  only whether the sender was disconnected (i.e, the receiver banned the sender). What do you think? @jimmygchen @pawanjay176 @jxs

jimmygchen: Nice diagrams and thanks for the explanation! From a test perspective I think it makes sense to verify the eventual outcome we care about and yay to one less flaky test!

Just for my understanding - in the 2nd scenario, the sender won't receive a stream termination (`Response::BlocksByRange(None)`), but instead an RPC error that will be handled right?

1. The `FromSwarm::ConnectionClosed`  event will push the RPC error to here:
https://github.com/sigp/lighthouse/blob/ee1b6bc81bb4b7dafb95a95eabcfcaf6b76cea04/beacon_node/lighthouse_network/src/service/mod.rs#L1433-L1451

2. and eventually handled for `BlocksByRange` here
https://github.com/sigp/lighthouse/blob/b7d78a91e03d4b3975806e2460bded01825f5a92/beacon_node/network/src/sync/manager.rs#L471-L473

ackintosh: That's slightly different. In the 2nd scenario, the RPC (or Service) doesn't emit an RPC error. The `FromSwarm::ConnectionClosed` event is indeed handled in [`RPC::on_swarm_event`](https://github.com/sigp/lighthouse/blob/7b5be8b1e74cf0ce71aeda71e017dd426920e955/beacon_node/lighthouse_network/src/rpc/mod.rs#L355), but no RPC error occurs because there are neither [rate-limited outbound requests](https://github.com/sigp/lighthouse/blob/7b5be8b1e74cf0ce71aeda71e017dd426920e955/beacon_node/lighthouse_network/src/rpc/mod.rs#L375-L386) nor [pending outbound requests](https://github.com/sigp/lighthouse/blob/7b5be8b1e74cf0ce71aeda71e017dd426920e955/beacon_node/lighthouse_network/src/rpc/mod.rs#L399-L416).

Instead, [`PeerManager::on_connection_closed` handles the `FromSwarm::ConnectionClosed` event](https://github.com/sigp/lighthouse/blob/ee1b6bc81bb4b7dafb95a95eabcfcaf6b76cea04/beacon_node/lighthouse_network/src/peer_manager/network_behaviour.rs#L139-L145) and emits a `PeerManagerEvent::PeerDisconnected` event. This event is then propagated through Service, Router and finally to [SyncManager](https://github.com/sigp/lighthouse/blob/522bd9e9c6ac167f2231525e937c9ebbcb86cf6e/beacon_node/network/src/sync/manager.rs#L819-L822). [SyncManager handles it as `RPCError::Disconnected`](https://github.com/sigp/lighthouse/blob/522bd9e9c6ac167f2231525e937c9ebbcb86cf6e/beacon_node/network/src/sync/manager.rs#L492). Finally, that `RPCError::Disconnected` is handled for `BlocksByRange` here:

https://github.com/sigp/lighthouse/blob/522bd9e9c6ac167f2231525e937c9ebbcb86cf6e/beacon_node/network/src/sync/manager.rs#L471-L473

ackintosh:
@jimmygchen Thanks for your feedback! I've updated the tests to remove response validation.

beacon_node/lighthouse_network/src/rpc/handler.rs
```diff
@@ -888,6 +889,18 @@ where
             ConnectionEvent::DialUpgradeError(DialUpgradeError { info, error }) => {
                 self.on_dial_upgrade_error(info, error)
             }
+            ConnectionEvent::ListenUpgradeError(ListenUpgradeError {
+                error: (proto, error),
+                ..
+            }) if matches!(error, RPCError::InvalidData(_)) => {
```
dapplion: Why only match and shutdown on `InvalidData`, what about all other variants in `RPCError`?

beacon_node/lighthouse_network/src/rpc/handler.rs
```diff
@@ -888,6 +889,24 @@ where
...
             }
+            ConnectionEvent::ListenUpgradeError(ListenUpgradeError {
+                error: (proto, error),
+                ..
+            }) => {
+                if matches!(
+                    error,
+                    RPCError::InvalidData(_) | RPCError::SSZDecodeError(_)
+                ) {
+                    // Peer is not complying with the protocol.
```
pawanjay176: Do we need to disconnect here immediately or should we just propagate the error and let the peer manager decide? 
cc @jxs

jxs: thanks for the ping Pawan. Yeah exactly, `Behaviour`s should not decide to disconnect peers on their own they should just report to the application layer for it to decide then, so that the disconnection is coordinated among all the `Behaviour`s. Lighthouse already has a mechanism for that:

https://github.com/ackintosh/lighthouse/blob/c4437d2927cdcdccbae725e4eef77286d53b2501/beacon_node/lighthouse_network/src/service/mod.rs#L1754-L1762

in this case Rpc should just report to the application layer for it to decide then how to penalize and act accordingly

ackintosh: @jxs @pawanjay176 Thanks for the feedback! I've updated the PR in https://github.com/sigp/lighthouse/pull/6986/commits/f666c7a8609e927dd69a6223b3f402bd1aa95f87 .
Note that I also removed the `shutdown()` call at lines 938–954.

pawanjay176 (approved):
Nice!

beacon_node/lighthouse_network/src/rpc/handler.rs
pawanjay176: Good call on removing the other shutdown calls and letting the peer manager handle it.

Issue #6959, opened 2025-02-10
De-duplicate `enable_light_client_server`


michaelsproul:
The `enable_light_client_server` config is parsed 3 separate times at startup, and shimmed into 3 different structs. It would be nice if we could just define it in one place and have a single source of truth.

https://github.com/sigp/lighthouse/blob/0344f68cfdea246d49ca9f4cfc641c3d20433270/beacon_node/src/config.rs#L178-L179

https://github.com/sigp/lighthouse/blob/0344f68cfdea246d49ca9f4cfc641c3d20433270/beacon_node/src/config.rs#L182-L184

https://github.com/sigp/lighthouse/blob/0344f68cfdea246d49ca9f4cfc641c3d20433270/beacon_node/src/config.rs#L1421-L1422

We are flipping the flag polarity in this PR, but the same issue remains:

- https://github.com/sigp/lighthouse/pull/6950

PR #6988, opened 2025-02-11
De-duplicate `enable_light_client_server`

EllipticPoint:
## Issue Addressed

https://github.com/sigp/lighthouse/issues/6959

## Proposed Changes

- Use `Arc<bool>` type to create a single source of truth
- A new field `enable_light_client_server: Arc<bool>` has been added to [`Config`](https://github.com/EllipticPoint/lighthouse/blob/e0b5603b946de1d9ef2e827cd874fa37a8818d7c/beacon_node/client/src/config.rs#L86).
- The 3 existing struct fields mentioned in the issue have been changed to `Arc<bool>`

### DIFF
--- a/beacon_node/http_api/src/lib.rs
+++ b/beacon_node/http_api/src/lib.rs
@@ -155,7 +155,7 @@ pub struct Config {
     pub enable_beacon_processor: bool,
     #[serde(with = "eth2::types::serde_status_code")]
     pub duplicate_block_status_code: StatusCode,
-    pub enable_light_client_server: bool,
+    pub enable_light_client_server: Arc<bool>,
     pub target_peers: usize,
 }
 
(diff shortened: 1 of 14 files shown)


beacon_node/http_api/src/lib.rs
```diff
@@ -155,7 +155,7 @@ pub struct Config {
     pub enable_beacon_processor: bool,
     #[serde(with = "eth2::types::serde_status_code")]
     pub duplicate_block_status_code: StatusCode,
-    pub enable_light_client_server: bool,
+    pub enable_light_client_server: Arc<bool>,
```
michaelsproul, line 158: I don't think this really addresses the problem of there not being a single source of truth. It's the same situation as before with three struct fields, and they're manually kept in sync.

If you don't mind, I'll just close this PR, as I think to achieve the single-source-of-truth would be a larger refactor that probably isn't worth doing atm.

michaelsproul: This is more along the lines of what I was thinking:

- https://github.com/sigp/lighthouse/pull/7017

Maybe you could extend my PR to remove the config field from the network crate?

EllipticPoint: Happy to close this PR as I don't think I'll have the time anytime soon to properly do a larger refactor.

michaelsproul: Nws, thanks for your efforts nonetheless!

Issue #6989, opened 2025-02-12
Supernode unable to progress range sync due to unknown RPC error


jimmygchen:
I've been trying to sync a fresh supernode to `peerdas-devnet-4`, and finalized sync seems to frequently get stuck.

I've observed RPC requests sent and completed but eventually the batch failed with RPC Error - but i don't see any RPC error for this batch during this period. 

See some partial logs below 
- 15 requests were sent for this batch epoch = 41, range request id = 35
- 15 requests were completed
- and a `Batch failed. RPC Error` followed immediately, with no RPC error logs found in this period.

I'm suspecting we have an RPC error log suspressed somewhere, it's not obvious from the log what the rpc error is, and the node just keep retrying to sync the same batch and unable to make progress.

```
Feb 12 04:38:17.144 DEBG Sync RPC request sent, id: 36/35/RangeSync/41/1, peer: 16Uiu2HAmDdyTRG4dESR8v9nnAbdwuqWXHDQ8hLwhgmSDV1mooiw9, epoch: 41, slots: 32, method: BlocksByRange, service: sync, module: network::sync::network_context:788
Feb 12 04:38:17.144 DEBG Sync RPC request sent, id: 37/35/RangeSync/41/1, peer: 16Uiu2HAmH5PwwsfXLCGqSi2xFgPu7Zy9xrzXKTqc5ZwxkkdhYbVy, columns: [105, 37, 35, 36, 76, 54, 96, 26], epoch: 41, slots: 32, method: DataColumnsByRange, service: sync, module: network::sync::network_context:870
Feb 12 04:38:17.144 DEBG Sync RPC request sent, id: 38/35/RangeSync/41/1, peer: 16Uiu2HAm9PijSZpm5QUphXRoBtkhUZPkGJ4Rgxk4Bny91oZPYZLG, columns: [100, 3, 13, 126, 63, 7, 103, 85, 41], epoch: 41, slots: 32, method: DataColumnsByRange, service: sync, module: network::sync::network_context:870
Feb 12 04:38:17.144 DEBG Sync RPC request sent, id: 39/35/RangeSync/41/1, peer: 16Uiu2HAm35d5GWdmRH2XtjM4PMi6MXyCSwXgZi5KwkGNijbSmARB, columns: [89, 8, 79, 110, 75, 15, 109, 58, 1], epoch: 41, slots: 32, method: DataColumnsByRange, service: sync, module: network::sync::network_context:870
[...]

PR #6990, opened 2025-02-12
Add missing error log and remove redundant id field from lookup logs

jimmygchen:
## Issue Addressed

Partially #6989. 

This PR adds the missing error log when a batch fails due to issues with converting the response into `RpcBlock`. See the above linked issue for more details.

Adding this log reveals that we're completing range requests with missing columns, hence causing the batch to fail. It looks like we've hit the case where we've received enough stream terminations, but not all columns are returned.

```
Feb 12 06:12:16.558 DEBG Failed to convert range block components into RpcBlock, error: No column for block 0xc5b6c7fa02f5ef603d45819c08c6519f1dba661fd5d44a2fc849d3e7028b6007 index 18, id: 3456/RangeSync/116/3432, service: sync, module: network::sync::network_context:488
```

I've also removed some redundant `id` logging, as the `id` debug representation is difficult to read, and is now being logged as part of `req_id` in a more succinct format (relevant PR: #6914)


beacon_node/network/src/sync/network_context.rs
```diff
@@ -486,6 +486,14 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
             let blocks = request
                 .into_responses(&self.chain.spec)
+                .inspect_err(|e| {
+                    debug!(
+                        self.log,
+                        "Failed to convert range block components into RpcBlock";
+                        "id" => %id,
+                        "error" => e
+                    )
+                })
```
dapplion: Superseded by https://github.com/sigp/lighthouse/pull/6991

beacon_node/network/src/sync/network_context/custody.rs
```diff
@@ -179,7 +176,6 @@ impl<T: BeaconChainTypes> ActiveCustodyRequest<T> {
             Err(err) => {
                 debug!(self.log,
                     "Custody column download error";
-                    "id" => ?self.custody_id,
```
dapplion: This is a good fix as req_id includes custody_id

beacon_node/network/src/sync/network_context.rs
jimmygchen: Deleted

jimmygchen: Updated, please re-review @dapplion

eserilev (approved):

PR #6991, opened 2025-02-12
Log range sync download errors

dapplion:
## Issue Addressed

Currently range sync download log errors just say `error: rpc_error` which isn't helpful. The actual error is suppressed unless logged somewhere else.

## Proposed Changes

Log the actual error that caused the batch download to fail as part of the log that states that the batch download failed.


jimmygchen (approved):
Looks good to me, running into this issue again, lets merge it!

PR #6993, opened 2025-02-12
Fix light client plumbing in beacon processor

michaelsproul:
## Issue Addressed

Our Holesky nodes running with the light client enabled were logging messages about full queues:

> Feb 12 22:09:28.949 ERRO Work queue is full                      queue: unknown_light_client_optimistic_update, queue_len: 128, msg: the system has insufficient resources for load, service: bproc

I thought this might be genuine overload, but it turns out this queue was never being read from!

## Proposed Changes

- [x] Rename light-client related queues in the beacon processor for clarity.
- [x] Ensure all light-client related queues are being popped from. 

## Additional Info

I noticed that the `sampling_result_queue` is never being read from either, but I've added a `TODO(das)` to address that in future.

### DIFF
--- a/beacon_node/beacon_processor/src/lib.rs
+++ b/beacon_node/beacon_processor/src/lib.rs
@@ -109,8 +109,6 @@ pub struct BeaconProcessorQueueLengths {
     gossip_voluntary_exit_queue: usize,
     gossip_proposer_slashing_queue: usize,
     gossip_attester_slashing_queue: usize,
-    finality_update_queue: usize,
-    optimistic_update_queue: usize,
     unknown_light_client_update_queue: usize,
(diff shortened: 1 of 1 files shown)


jimmygchen (approved):
Good catch, LGTM!

michaelsproul:
For the record: the impact of the bug isn't that we **never** processing light client updates from gossip. It's that we never process them once queued. It takes a while for the queue to build up, because if the processor is not at capacity, we will still schedule and run processing immediately here:

https://github.com/sigp/lighthouse/blob/029b4f21047c37d1ffde51c554737b1aa0880f97/beacon_node/beacon_processor/src/lib.rs#L1301-L1302

michaelsproul:
@mergify queue

Issue #6937, opened 2025-02-07
Fix failing execution tests with latest geth version


jimmygchen:
We made a workaround to use geth `v1.10.10` instead of `master` to  get CI to pass for the time being:

https://github.com/sigp/lighthouse/pull/6936

Our tests need to be updated to work with latest geth, due to these breaking changes:

- removal of personal namespace in v1.14.12: See [#30704](https://github.com/ethereum/go-ethereum/pull/30704)
- removal of totalDifficulty field from RPC in v1.14.11. See [#30386](https://github.com/ethereum/go-ethereum/pull/30386).

michaelsproul:
I think the removal of total difficulty was also causing some of our readiness loggers to error. From [`interop`](https://discord.com/channels/595666850260713488/892088344438255616/1336021354591289406) channel on ETH R&D:

> {"msg":"Unable to check genesis execution payload","level":"ERRO","ts":"2025-02-03T17:07:31.659060494Z","service":"slot_notifier","error":"ExecutionLayerGetBlockByNumberFailed(EngineError(Api { error: Json(Error(\"missing field `totalDifficulty`\", line: 0, column: 0)) }))"}

It's cosmetic, but would be nice to fix (only relevant for devnets)

michaelsproul:
Removing the v7.0.0 tag from this, but leaving the issue open. We should still merge @EllipticPoint's PR:

- https://github.com/sigp/lighthouse/pull/6996

PR #6996, opened 2025-02-13
Fix execution engine integration tests with latest geth version

EllipticPoint:
## Issue Addressed

https://github.com/sigp/lighthouse/issues/6937

## Proposed Changes

- Use `ethers-rs` [`Signer`](https://www.gakonst.com/ethers-rs/middleware/signer.html) middleware for local signing and sending raw txs to geth
- ~~Set `totalDifficulty = 0` through `serde` default if the block does not contain a `totalDifficulty` field~~


michaelsproul:
Hey sorry for the slow review @EllipticPoint. Can you fix up merge conflicts and then we'll try to review and merge?

michaelsproul:
Calling in @macladson to help with review.

macladson (changes requested):
Hey @EllipticPoint, just noticed that you merged `stable` into your branch. Normally this would be fine, but we actually have some extra commits on the `stable` branch which we want to keep there for now. Would you be able to do a `git rebase -i unstable` and remove the extra commits please? :pray:

macladson (commented):
Hmm, still seeing some of those commits. Let me know if you need any help cleaning it up!

macladson (changes requested):
Sorry @EllipticPoint, one more thing.
It looks like the `Cargo.lock` file either got deleted or a `cargo update` got ran.
In Lighthouse we generally want ensure that most dependency updates are done intentionally.

If you could please run `git restore --source=unstable Cargo.lock` to reset the lockfile, then build Lighthouse again (normally a `make lint` is sufficient), this will ensure that the lockfile is updated only with changes specific to this PR

macladson (approved):
This looks solid to me!
CI isn't passing because of an unrelated `cargo audit` failure.
Once that fix is merged this PR will just need a `git merge unstable` and then it'll be good to go

PR #6998, opened 2025-02-13
Track request IDs in RangeBlockComponentsRequest

dapplion:
## Issue Addressed

Part of 

## Proposed Changes

`RangeBlockComponentsRequest` handles a set of by_range requests. It's quite lose on these requests, not tracking them by ID. We want to implement individual request retries, so we must make `RangeBlockComponentsRequest` aware of its requests IDs. We don't want the result of a prior by_range request to affect the state of a future retry. Lookup sync uses this mechanism.

Now `RangeBlockComponentsRequest` tracks:

```rust
pub struct RangeBlockComponentsRequest<E: EthSpec> {
    blocks_request: ByRangeRequest<BlocksByRangeRequestId, Vec<Arc<SignedBeaconBlock<E>>>>,
    block_data_request: RangeBlockDataRequest<E>,
}

enum RangeBlockDataRequest<E: EthSpec> {
    NoData,
    Blobs(ByRangeRequest<BlobsByRangeRequestId, Vec<Arc<BlobSidecar<E>>>>),
    DataColumns {
        requests: HashMap<
            DataColumnsByRangeRequestId,
            ByRangeRequest<DataColumnsByRangeRequestId, DataColumnSidecarList<E>>,
        >,
    },
}

enum ByRangeRequest<I: PartialEq + std::fmt::Display, T> {
    Active(I),
    Complete(T),
}
```

I have merged `is_finished` and `Into_responses` into the same function. Otherwise, we need to duplicate the logic to figure out if the requests are done.


beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -1,89 +1,156 @@
 use beacon_chain::{
     block_verification_types::RpcBlock, data_column_verification::CustodyDataColumn, get_block_root,
```
T10876, line 2: الو

pawanjay176 (approved):
LGTM. Much cleaner

realbigsean (approved):
Looks really good, way simpler

jimmygchen:
Nice, thanks for the reviews @pawanjay176 @realbigsean. I'm going to run some tests on this branch and we can merge!

jimmygchen (approved):
Tested this branch and able to sync to both mainnet and peerdas-devnet-5.

PR #7006, opened 2025-02-15
feat: implement new beacon APIs(accessors for pending_deposits/pending_partial_withdrawals)

### LINKED ISSUE #7003: Implement new beacon APIs for pending deposits/partial withdrawals
Two new informational APIs have been added to the beacon API spec in this PR:

- https://github.com/ethereum/beacon-APIs/pull/500

We should implement them. It should be straight-forward. We can error with a 400 if a pre-Electra state is requested.

syjn99:
## Issue Addressed

Resolves #7003 

## Proposed Changes

Added two endpoints as https://github.com/ethereum/beacon-APIs/pull/500 proposed:
- `/eth/v1/beacon/states/{state_id}/pending_deposits`
- `/eth/v1/beacon/states/{state_id}/pending_partial_withdrawals`

### DIFF
--- a/beacon_node/http_api/src/lib.rs
+++ b/beacon_node/http_api/src/lib.rs
@@ -1133,15 +1133,6 @@ pub fn serve<T: BeaconChainTypes>(
                         .map_state_and_execution_optimistic_and_finalized(
                             &chain,
                             |state, execution_optimistic, finalized| {
-                                if !state.fork_name_unchecked().electra_enabled() {
-                                    return Err(warp_utils::reject::pre_electra_not_supported(
-                                        format!(
-                                            "state at epoch {} is not activated for Electra",
-                                            state.current_epoch()
-                                        ),
-                                    ));
-                                }
-
                                 let Ok(deposits) = state.pending_deposits() else {
                                     return Err(warp_utils::reject::custom_bad_request(
                                         "Pending deposits not found".to_string(),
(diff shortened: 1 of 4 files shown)


beacon_node/http_api/src/lib.rs
```diff
@@ -1125,6 +1125,81 @@ pub fn serve<T: BeaconChainTypes>(
...
+        .then(
+            |state_id: StateId,
+             task_spawner: TaskSpawner<T::EthSpec>,
+             chain: Arc<BeaconChain<T>>| {
+                task_spawner.blocking_json_task(Priority::P1, move || {
+                    let (data, execution_optimistic, finalized) = state_id
+                        .map_state_and_execution_optimistic_and_finalized(
+                            &chain,
+                            |state, execution_optimistic, finalized| {
+                                if !state.fork_name_unchecked().electra_enabled() {
```
michaelsproul: we probably don't need this check or the new error type because the case where the state doesn't have the field is handled below with `custom_bad_request`

michaelsproul (commented):
Looks good at a glance, you're definitely on the right track. I've triggered CI so you can see if the tests pass. You can also run locally, just not in debug mode

michaelsproul:
You can fix the cargo format failure by running `cargo fmt --all`

syjn99:
```
RUST_MIN_STACK=16777216 cargo test beacon_get_state_info_electra -p http_api --test bn_http_api_tests -- --nocapture
```

@michaelsproul Hi, thanks to #7007, I found that my test successfully passed with the commit. However, I'm afraid that default `BeaconChainHarness` will always return empty list of `pending_deposits`/`pending_partial_withdrawals` as it is initialized with default value. Do I have to add a case for comparing **not emptied** list? If then, would you mind letting me know where the good place to start?

michaelsproul:
I think we should include this in v7.0.0 as we have some users asking about it.

@syjn99 Do you mind rebasing your changes on the `release-v7.0.0` branch? :pray:

michaelsproul:
When you rebase, just take out all the commits from `unstable` (use `rebase -i`).

michaelsproul (approved):
LGTM

Just pushed one small commit to use `List` in the serialization rather than `Vec`. This saves some copying of data (`List` and `Vec` have the same serialized representation anyway, so we can mix and match them).

PR #7008, opened 2025-02-18
Drop verify single RpcBlock fns

dapplion:
## Issue Addressed

Noted that we never have to verify individual RpcBlocks for sync. Lookup sync sends the block alone always, and range sync sends chain segments.

## Proposed Changes

- Removed verify_kzg_for_rpc_block
- Removed `IntoExecutionPendingBlock` impl for RpcBlock. Now lookup sync blocks use the SignedBeaconBlock `IntoExecutionPendingBlock` impl which is the same
- Switched tests to import RpcBlocks with process_chain_segment


michaelsproul:
Should we merge this before:

- https://github.com/sigp/lighthouse/pull/6866

?

dapplion:
> Should we merge this before

Independent, I can fix conflicts later.

Issue #7000, opened 2025-02-14
Ensure SSZ builder works


michaelsproul:
@pawanjay176 is investigating issues with our SSZ builder implementation, added in:

- https://github.com/sigp/lighthouse/pull/6859

There may be no fault in our impl, it could be an issue on the mev-boost side.

Either way, we may want to harden our error handling so that we disable SSZ if we get an error response.

We could set `ssz_enabled` to `result.is_ok()` here, rather than setting it to true:

https://github.com/sigp/lighthouse/blob/1888be554cb365c419aac993528bc6697728f793/beacon_node/builder_client/src/lib.rs#L135-L142

Similarly, here:

https://github.com/sigp/lighthouse/blob/1888be554cb365c419aac993528bc6697728f793/beacon_node/execution_layer/src/lib.rs#L1905-L1908

PR #7009, opened 2025-02-19
Fix builder API headers

pawanjay176:
## Issue Addressed

Resolves https://github.com/sigp/lighthouse/issues/7000

## Proposed Changes

Set the accept header on builder to the correct value when requesting ssz.

This PR also adds a flag to disable ssz over the builder api altogether. In the case that builders/relays have an ssz bug, we can react quickly by asking clients to restart their nodes with the `--disable-ssz-builder` flag to force json. I'm not fully convinced if this is useful so open to removing it or opening another PR for it.

Testing this currently.

### DIFF
--- a/beacon_node/builder_client/src/lib.rs
+++ b/beacon_node/builder_client/src/lib.rs
@@ -159,10 +159,10 @@ impl BuilderHttpClient {
     }
 
     /// Return `true` if the most recently received response from the builder had SSZ Content-Type.
-    ///
+    /// Return `false` otherwise.
     /// Also returns `false` if we have explicitly disabled ssz.
     pub fn is_ssz_enabled(&self) -> bool {
-        self.disable_ssz && self.ssz_used.load(Ordering::SeqCst)
+        !self.disable_ssz && self.ssz_used.load(Ordering::SeqCst)
     }
 
     async fn get_with_timeout<T: DeserializeOwned, U: IntoUrl>(
--- a/beacon_node/builder_client/src/lib.rs
+++ b/beacon_node/builder_client/src/lib.rs
@@ -7,8 +7,7 @@ use eth2::types::{
 use eth2::types::{FullPayloadContents, SignedBlindedBeaconBlock};
 pub use eth2::Error;
 use eth2::{
-    ok_or_error, StatusCode, CONSENSUS_VERSION_HEADER, CONTENT_TYPE_HEADER,
-    JSON_CONTENT_TYPE_HEADER, SSZ_CONTENT_TYPE_HEADER,
+    ok_or_error, StatusCode, CONSENSUS_VERSION_HEADER, CONTENT_TYPE_HEADER, SSZ_CONTENT_TYPE_HEADER,
 };
 use reqwest::header::{HeaderMap, HeaderValue, ACCEPT};
 use reqwest::{IntoUrl, Response};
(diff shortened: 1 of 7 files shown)


michaelsproul:
let's target this at the release-v7.0.0-beta.0 branch, which we'll use for future releases too (I should rename the branch most likely)

pawanjay176:
Tested this with lighthouse and other CL clients with the mock builder and ssz support works as expected.
the `--builder-disable-ssz` flag also leads to all requests made with json.

michaelsproul:
Nice, ty. Rebased on a new branch called `release-v7.0.0` which descends from `v7.0.0-beta.0` (not unstable)

beacon_node/builder_client/src/lib.rs
```diff
@@ -57,14 +62,20 @@ pub struct BuilderHttpClient {
     server: SensitiveUrl,
     timeouts: Timeouts,
     user_agent: String,
-    ssz_enabled: Arc<AtomicBool>,
+    /// Only use json for all requests/responses types.
+    disable_ssz: bool,
+    /// Indicates that the `get_header`` response had content-type ssz
+    /// so we can set content-type header to ssz to make the `submit_blinded_blocks`
+    /// request.
+    ssz_used: Arc<AtomicBool>,
```
michaelsproul: This might be a better name?

```suggestion
    ssz_available: Arc<AtomicBool>,
```

beacon_node/src/cli.rs
```diff
@@ -1460,6 +1460,15 @@ pub fn cli_app() -> Command {
                 .action(ArgAction::Set)
                 .display_order(0)
         )
+        .arg(
+            Arg::new("builder-disable-ssz")
+                .long("builder-disable-ssz")
+                .value_name("BOOLEAN")
+                .help("Disables sending requests using ssz over the builder api")
```
```suggestion
                .help("Disables sending requests using SSZ over the builder API.")
```

book/src/help_bn.md
```diff
@@ -28,6 +28,8 @@ Options:
           network. Multiaddr is also supported.
       --builder <builder>
           The URL of a service compatible with the MEV-boost API.
+      --builder-disable-ssz
+          Disables sending requests using ssz over the builder api
```
michaelsproul:
```suggestion
          Disables sending requests using SSZ over the builder API.
```

michaelsproul (approved):
Nice, substance looks good, just a few formatting nits

beacon_node/builder_client/src/lib.rs
```diff
@@ -124,15 +136,15 @@ impl BuilderHttpClient {
...
             return serde_json::from_slice(&response_bytes).map_err(Error::InvalidJson);
         };
 
         let content_type = self.content_type_from_header(&headers);
 
         match content_type {
             ContentType::Ssz => {
-                self.ssz_enabled.store(true, Ordering::SeqCst);
+                self.ssz_available.store(true, Ordering::SeqCst);
                 T::from_ssz_bytes_by_fork(&response_bytes, fork_name)
```
eserilev, line 148: in #7000 Michael also mentioned we should harden our error handling here by disabling SSZ if `from_ssz_bytes_by_fork` returns an error. Additionaly we could also even fall back to deserializing to JSON if SSZ decoding fails, to further protect against buggy builder impls. Might be overkill, but just a thought

pawanjay176: Yeah I thought about that. I think its better for the user to explicitly disable the ssz workkflow with a flag rather than us automatically doing this. With ssz failing in `get_header`, we will fallback to local block production so its not the worst thing.
Also falling back to json automatically might end up hiding a potential issue in the relay/builder/mev-boost.
Wdyt?

Issue #6610, opened 2024-11-26
DVT Compatibility


AgeManning:
## Description

There are two endpoints in the beaconapi specs, specifically designed for DVT middleware:
- https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Validator/submitBeaconCommitteeSelections
- https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Validator/submitSyncCommitteeSelections

We have a `--distributed` flag in the VC now. We should implement these endpoints on the BN and when this flag is set on the VC use these endpoints. 

There is further information and testing links in this comment: https://github.com/sigp/lighthouse/pull/4867#issuecomment-2493041658

michaelsproul:
My understanding is that we do not need to implement these on the BN: just to use them from the VC.

michaelsproul:
I will investigate

OisinKyne:
> My understanding is that we do not need to implement these on the BN: just to use them from the VC.

Correct. You just need to call them with the VC, charon returns an aggregated signature in exchange for the partial one sent, and then the aggregation selection process continues as normal from there :)

OisinKyne:
Hey @michaelsproul, have you had a chance to take a look at this issue yet? :)

michaelsproul:
@OisinKyne No, sorry! It fell off my todo list, will try to make some time for it now Electra is stabilising

michaelsproul:
@chong-he is going to take a look at this.

This is the point to get started (for the aggregate attestation API):

https://github.com/sigp/lighthouse/blob/029b4f21047c37d1ffde51c554737b1aa0880f97/validator_client/validator_services/src/duties_service.rs#L128-L150

We need to:

- Define the type for `BeaconCommitteeSelectionRequest` (from `beacon-api`).
- If `--distributed` mode is enableld: add an HTTP call to the aggregate committee API (using `first_success` on the list of beacon nodes is fine).
- Use the response from the API as the `selection_proof`, and keep the same logic for checking whether the validator is an aggregator (`is_aggregator`).

OisinKyne:
Hey @chong-he, let me know if you have any questions on what's involved. :)

chong-he:
> Hey [@chong-he](https://github.com/chong-he), let me know if you have any questions on what's involved. :)

[...]

PR #7016, opened 2025-02-20
Implement `selections` Beacon API endpoints to support DVT middleware

chong-he:
## Issue Addressed

* #6610 

## Proposed Changes

- [x] Add `beacon_committee_selections` endpoint
- [x] Test beacon committee aggregator and confirmed working
- [x] Add `sync_committee_selections` endpoint
- [x] Test sync committee aggregator and confirmed working

## Additional Info

Thank you @michaelsproul and @macladson for the help and guidance


common/eth2/src/lib.rs
```diff
@@ -2664,6 +2664,22 @@ impl BeaconNodeHttpClient {
         )
         .await
     }
+
+    /// `POST validator/beacon_committee_selections`
+    pub async fn post_validator_beacon_committee_selections(
+        &self,
+        selections: &[BeaconCommitteeSelection],
+    ) -> Result<GenericResponse<Vec<SelectionProof>>, Error> {
```
michaelsproul: This line is saying that we expect a `Vec<SelectionProof>` as the response. A `SelectionProof` is just a standalone signature. But the middleware will actually return a `Vec` of objects like this:

```json
{
  "data": [
    {
      "validator_index": "1",
      "slot": "1",
      "selection_proof": "0x1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505cc411d61252fb6cb3fa0017b679f8bb2305b26a285fa2737f175668d0dff91cc1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505"
    }
  ]
}
```

These objects are `BeaconCommitteeSelection`s, so we need to change the type here to `Vec<BeaconCommitteeSelection>`.

chong-he: Resolved in [0d724b1](https://github.com/sigp/lighthouse/pull/7016/commits/0d724b18a0e93b3f42e1b4dd7bbc43eb7ff1f96c), thanks

chong-he:
With the commit [8b2f058](https://github.com/sigp/lighthouse/pull/7016/commits/8b2f058d7be12022dc62ce26cf00e8a4deb43c85) where VC sends partial signatures in parallel for all validators for a slot, I believe the aggregated attestations are successful (during testing) now, thank you @michaelsproul !

Lighthouse VC logs:
```
{"log":"Mar 20 06:36:24.046 INFO Successfully published attestation      type: aggregated, slot: 37, committee_index: 0, head_block: 0xdb186315afecb3d1134d03eb2d5b04b83a21ede56be6563e3bff304b103bf4b6, signatures: 24, aggregator: 58, service: attestation\n","stream":"stderr","time":"2025-03-20T06:36:24.04690072Z"}
```
for every slot now

Charon logs:
```
{"log":"06:36:24.074 INFO bcast      Successfully submitted attestation aggregations to beacon node {\"delay\": \"8.074945477s\", \"duty\": \"37/aggregator\", \"peer\": \"twinkling-pen\", \"protocol\": \"/charon/parsigex/2.0.0\"}\n","stream":"stderr","time":"2025-03-20T06:36:24.074999639Z"}
```
for every slot now too

@KaloyanTanev could you give it a try and confirm that the beacon committee selection endpoint is working as intended? 

Thanks

KaloyanTanev:
Tested on local Kurtosis chain and it works!

KaloyanTanev (approved):

chong-he:
> Tested on local Kurtosis chain and it works!

Thanks for confirming, I am now working on the sync committee selection endpoint

consensus/types/src/sync_selection_proof.rs
```diff
@@ -7,11 +7,12 @@ use crate::{
...
 use ethereum_hashing::hash;
 use safe_arith::{ArithError, SafeArith};
+use serde::{Deserialize, Serialize};
 use ssz::Encode;
 use ssz_types::typenum::Unsigned;
 use std::cmp;
 
-#[derive(arbitrary::Arbitrary, PartialEq, Debug, Clone)]
+#[derive(arbitrary::Arbitrary, PartialEq, Debug, Clone, Serialize, Deserialize)]
 pub struct SyncSelectionProof(Signature);
```
michaelsproul, line 18: I wonder if we need `serde(transparent)`?

validator_client/validator_services/src/sync.rs
```diff
@@ -521,129 +522,316 @@ pub async fn fill_in_aggregation_proofs<T: SlotClock + 'static, E: EthSpec>(
...
+                                Ok(true) => {
+                                    if let Some(Some(duty)) =
+                                        validators.get(&selection.validator_index)
+                                    {
+                                        debug!(
+                                            log,
+                                            "Validator is sync aggregator";
+                                            "validator_index" => selection.validator_index,
+                                            "slot" => selection.slot,
+                                            "subnet_id" => %subnet_id,
```
michaelsproul: For debugging, we could log `?proof` here

validator_client/validator_services/src/sync.rs
```diff
@@ -521,129 +522,316 @@ pub async fn fill_in_aggregation_proofs<T: SlotClock + 'static, E: EthSpec>(
...
+                            .produce_sync_selection_proof(
+                                &duty.pubkey,
+                                proof_slot,
+                                subnet_id.into(),
+                            )
+                            .await;
+
+                        match sync_selection_proof {
+                            Ok(proof) => {
+                                let sync_committee_selection = SyncCommitteeSelection {
```
michaelsproul: Could also log the partial proof here (to make sure it's different from the full proof)

michaelsproul:
Could log the `selection_proof` and `aggregator_index` when posting the SignedContributionAndProof here:

https://github.com/sigp/lighthouse/blob/b21451fc114810e4e36777ec6b3c3d5072e15de9/validator_client/validator_services/src/sync_committee_service.rs#L398

This selection proof should be the same as the full proof.

validator_client/validator_services/src/sync.rs
```diff
@@ -521,129 +522,316 @@ pub async fn fill_in_aggregation_proofs<T: SlotClock + 'static, E: EthSpec>(
...
+                                break;
+                            }
+
+                            let selection = &valid_results[i];
+                            let subnet_id = SyncSubnetId::new(selection.subcommittee_index);
+
+                            // Convert the response to a SyncSelectionProof
+                            let proof = SyncSelectionProof::from(
+                                selection_response.selection_proof.clone(),
+                            );
```
michaelsproul: For debugging, we could check here that the `selection_response` is consistent with `duty.validator_index`, `duty.slot`, and `selection_response.subcommittee_index` is the same as `subnet_id`

^If Charon is not keeping track of the slots properly, this could be causing the error.

michaelsproul:
One more comment from debugging, the signature that is failing for sync committee selection proofs is the aggregated selection proof, here:

https://github.com/ObolNetwork/charon/blob/22ce8f0ef250cef7d4942e14ba0571f8fe357584/core/validatorapi/validatorapi.go#L928

chong-he:
As per [ef90462](https://github.com/sigp/lighthouse/pull/7016/commits/ef90462edbe9f1eafb39f9feb029d8db00a74097), testing shows that:

Lighthouse VC:
```
{"log":"Mar 27 13:05:18.097 INFO Successfully published sync contributions, slot: 43, num_signers: 128, beacon_block_root: 0x965f3d26d40f79949eff6ec2a7e5e6a4f59984beb78173b0ea98100774981f7b, subnet: 2, service: sync_committee\n","stream":"stderr","time":"2025-03-27T13:05:18.09860353Z"}
```

Charon:
```
{"log":"13:05:18.129 INFO bcast      Successfully submitted sync committee contributions to beacon node {\"delay\": \"8.128996601s\", \"duty\": \"43/sync_contribution\", \"peer\": \"calm-street\", \"protocol\": \"/charon/parsigex/2.0.0\"}\n","stream":"stderr","time":"2025-03-27T13:05:18.129227056Z"}
```
for every slot. This should imply that the `sync_committee_selection` endpoint is working. Waiting confirmation from Obol. 

Edit: confirmed by @KaloyanTanev that Sync committee selection endpoint is working as intended, thanks for testing it. I will clean up the code a bit to get the PR ready for review

validator_client/validator_services/src/sync.rs
chong-he: Added logging in [7ac6867](https://github.com/sigp/lighthouse/pull/7016/commits/7ac6867c5bca9b9f2f8c6bc145b361c17c4891eb)

validator_client/validator_services/src/sync.rs
chong-he: Added logging in [7ac6867](https://github.com/sigp/lighthouse/pull/7016/commits/7ac6867c5bca9b9f2f8c6bc145b361c17c4891eb)

validator_client/validator_services/src/sync.rs
chong-he: Added logging in [7ac6867](https://github.com/sigp/lighthouse/pull/7016/commits/7ac6867c5bca9b9f2f8c6bc145b361c17c4891eb)

dknopik:
Hello @chong-he and @michaelsproul,

I found this PR and am interested in the changes in `fill_in_selection_proofs` as Anchor's selection proofs do not work with the current state from `unstable`, but this PR fixes it partially. Unfortunately, there are some differences between `--distributed` mode and the behaviour required by Anchor, so I propose we make this PR a bit more general in order to accommodate everyone's needs.

To illustrate the differences:
|  | VC | VC with `--distributed` | Anchor |
|--------|--------|--------|--------|
| Look ahead | 8 | 1 | 0 (current slot only)* |
| Run at x seconds into slot | 6 | 6 | 0* |
| Use middleware endpoints | No | Yes | No |
| Sign in parallel | No | Yes | Yes** | 

*: Alternatively, this can be expressed as lookahead of 1, ran 12 seconds into the slot.
**: Ideally, results should handled as they come in, since some signatures may fail late. In other words, await a stream instead of using `join_all`.

We could bundle these configuration options in a struct (e.g. `SelectionProofConfig`) and pass it to the duties service. This way, Lighthouse can pass in the required values based on whether `--distributed` is specified, and Anchor can pass in its values.

Looking forward to hear your thoughts. Let me know if you have any questions.

validator_client/validator_services/src/duties_service.rs
```diff
@@ -1153,7 +1167,7 @@ async fn fill_in_selection_proofs<T: SlotClock + 'static, E: EthSpec>(
                             &duty,
                             &duties_service.validator_store,
                             &duties_service.spec,
-                            duties_service.distributed,
+                            !config.selections_endpoint, // non-distributed case
```
dknopik: I think the negation is not necessary here

validator_client/validator_services/src/duties_service.rs
```diff
@@ -1157,9 +1162,10 @@ async fn fill_in_selection_proofs<T: SlotClock + 'static, E: EthSpec>(
                         )
                         .await?;
                         Ok((duty, opt_selection_proof))
-                    },
-                ))
-                .await
+                    })
+                    .collect::<FuturesUnordered<_>>();
+
+                futures.collect::<Vec<_>>().await
```
dknopik: This effectively waits for all results - so, used like this, `FuturesUnordered` has no benefit over `join_all`.

`FuturesUnordered` implements `Stream`, and thus `StreamExt`, which means that we can use this: https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.next

This creates a future that yields a single result from the `FuturesUnordered` (or `None` if there are no more futures), allowing us to use it in a loop like this:

```rust
while let Some(result) = futures.next().await {
    // handle result
}
```

As we only wait for the next result, not all results, we can process results as they come in.

dknopik (commented):
Thanks, looks good so far!

validator_client/validator_services/src/duties_service.rs
chong-he: Thanks so much, I got myself confused when trying to figure the logic here, fixed in: [af676d4](https://github.com/sigp/lighthouse/pull/7016/commits/af676d43bee4b51c46ebee088cc1226e57041a56)

validator_client/validator_services/src/sync.rs
```diff
@@ -393,12 +406,22 @@ pub async fn poll_sync_committee_duties<T: SlotClock + 'static, E: EthSpec>(
...
             duties_service.context.executor.spawn(
                 async move {
+                    let config = SelectionProofConfig {
+                        lookahead_slot: sub_duties_service
+                            .sync_duties
+                            .aggregation_pre_compute_slots(),
+                        computation_offset: Duration::from_secs(12),
+                        selections_endpoint: sub_duties_service.distributed,
+                        parallel_sign: sub_duties_service.distributed,
+                    };
```
dknopik: In order for Anchor to be able to pass in its own config values, we must be able to pass the `SelectionProofConfig` from outside the crate. So, it would be nice if the config is passed to the `SyncDutiesMap` instead of the `distributed` flag. Basically, the `SelectionProofConfig` would be created where the validator services are used.

chong-he: Revised in [2bcb984](https://github.com/sigp/lighthouse/pull/7016/commits/2bcb984f80f68a3af28404173f2d4ea5b58ef274). From my understanding this change is only in `sync.rs`, not `duties_service.rs`, because in `duties_service.rs`, it is already on it's own crate. Feel free to correct me if I am wrong.

On a side note, one thing I am unsure is that `computation_offset` is not used in `sync.rs`, so I let it be a 0 value, but that's a bit dodgy. How should I go around with this? Strangely, the compiler does not complain that the field isn't used?

consensus/types/src/sync_selection_proof.rs
chong-he: Added in [43b209f](https://github.com/sigp/lighthouse/pull/7016/commits/43b209f5292a711ba3bb9ce9c6163e64a2c2693f)

validator_client/validator_services/src/duties_service.rs
chong-he: `FuturesUnordered` for both `duties_service.rs` (for beacon commitee selection) and `sync.rs` (for sync committee selection applied in latest commit: [05df47b](https://github.com/sigp/lighthouse/pull/7016/commits/05df47bed8714ed3a445611bf483cdcd7630ea0f)

validator_client/validator_services/src/sync.rs
dknopik: Sorry, I mean something different. 
Let me explain by linking some code.

https://github.com/sigp/lighthouse/blob/9459c362dde98fd8a993594e6218500e678f5b5c/validator_client/src/lib.rs#L448-L461

Here, Lighthouse initialises the duties service, and passes in the `distributed` flag. Anchor's code looks similar:

https://github.com/sigp/anchor/blob/b0b41846fffd4f12948e882faebd4745b25de270/anchor/client/src/lib.rs#L474-L484

Currently, we pass in `true` for distributed. However, with your changes, we need to be able to pass in a `SelectionProofConfig`, as we do not want to use the middleware endpoints. So I suggest removing the `distributed` flag from the `DutiesService` and allow the using code (Lighthouse VC or Anchor) to pass in a `SelectionProofConfig` instead. Lighthouse can create the correct config and pass it into the service, checking the distributed flag. And Anchor can pass its configuration in. 

If Lighthouse needs differing configs for Sync Selection proofs and Attestation Selection proofs, it might make sense to allow passing in the config via the `AttestationService` and the `SyncCommitteeService` instead.

Let me know if you have any questions!

validator_client/src/lib.rs
```diff
@@ -444,19 +456,38 @@ impl<E: EthSpec> ProductionValidatorClient<E> {
...
+                parallel_sign: false,
+            }
+        };
+
         let duties_context = context.service_context("duties".into());
         let duties_service = Arc::new(DutiesService {
             attesters: <_>::default(),
             proposers: <_>::default(),
-            sync_duties: SyncDutiesMap::new(config.distributed),
+            sync_duties: SyncDutiesMap::new(selection_proof_config.clone()),
```
dknopik: You could pass in a different config value here.

validator_client/validator_services/src/duties_service.rs
```diff
@@ -140,10 +130,10 @@ async fn make_selection_proof<T: SlotClock + 'static, E: EthSpec>(
     duty: &AttesterData,
     validator_store: &ValidatorStore<T, E>,
     spec: &ChainSpec,
-    config: &SelectionProofConfig,
+    duties_service: &DutiesService<T, E>,
     beacon_nodes: &Arc<BeaconNodeFallback<T, E>>,
 ) -> Result<Option<SelectionProof>, Error> {
-    let selection_proof = if config.selections_endpoint {
+    let selection_proof = if duties_service.selection_proof_config.selections_endpoint {
```
dknopik: I think it would be nicer to pass only the config to this function.

validator_client/validator_services/src/duties_service.rs
```diff
@@ -1186,17 +1160,32 @@ async fn fill_in_selection_proofs<T: SlotClock + 'static, E: EthSpec>(
 
     while !duties_by_slot.is_empty() {
         if let Some(duration) = slot_clock.duration_to_next_slot() {
-            sleep(duration.saturating_sub(config.computation_offset)).await;
+            sleep(
+                duration.saturating_sub(
+                    duties_service
+                        .sync_duties
```
dknopik: why `sync_duties`? I think we are in an attestation context here. Same below.

validator_client/validator_services/src/duties_service.rs
```diff
@@ -1231,7 +1224,7 @@ async fn fill_in_selection_proofs<T: SlotClock + 'static, E: EthSpec>(
                             &duty,
                             &duties_service.validator_store,
                             &duties_service.spec,
-                            &config,
+                            &duties_service,
                             &duties_service.beacon_nodes,
```
dknopik: `&duties_service.selection_proof_config` should work here.

validator_client/validator_services/src/sync.rs
```diff
@@ -109,7 +108,7 @@ impl<E: EthSpec> SyncDutiesMap<E> {
 
     /// Number of slots in advance to compute selection proofs
     fn aggregation_pre_compute_slots(&self) -> u64 {
-        if self.distributed {
+        if self.selection_proof_config.parallel_sign {
             AGGREGATION_PRE_COMPUTE_SLOTS_DISTRIBUTED
         } else {
             E::slots_per_epoch() * AGGREGATION_PRE_COMPUTE_EPOCHS
```
dknopik: We should remove this function, as the pre-compute slots should be used from the config directly.

validator_client/src/lib.rs
```diff
@@ -444,19 +456,38 @@ impl<E: EthSpec> ProductionValidatorClient<E> {
...
+            SelectionProofConfig {
+                lookahead_slot: SELECTION_PROOF_SLOT_LOOKAHEAD_DVT,
+                computation_offset: slot_clock.slot_duration() / SELECTION_PROOF_SCHEDULE_DENOM,
+                selections_endpoint: true,
+                parallel_sign: true,
+            }
+        } else {
+            SelectionProofConfig {
+                lookahead_slot: SELECTION_PROOF_SLOT_LOOKAHEAD,
+                computation_offset: Duration::from_secs(6),
```
dknopik: Should this duration also be `slot_clock.slot_duration() / SELECTION_PROOF_SCHEDULE_DENOM`?

EDIT: Ah, fixed already :)

dknopik (commented):
Thanks! I've added a few notes.

validator_client/validator_services/src/duties_service.rs
chong-he: This should be resolved by now

validator_client/validator_services/src/sync.rs
chong-he: This should be resolved by the latest commit [69e7e60](https://github.com/sigp/lighthouse/pull/7016/commits/69e7e60f93e38e888e88f8e88e148c9f1b919500)

validator_client/src/lib.rs
chong-he: The above comments are all resolved by now

chong-he:
> Thanks! I've added a few notes.

Thank you so much for the review and help. All revised accordingly, let me know if further changes required.

validator_client/validator_services/src/sync.rs
```diff
@@ -498,134 +486,297 @@ pub async fn poll_sync_committee_duties_for_period<S: ValidatorStore, T: SlotClo
...
-                            slot = %proof_slot,
-                            error = ?e,
-                            "Error determining is_aggregator"
-                        );
-                        None
+                // Create futures to produce proofs.
+                let duties_service_ref = &duties_service;
+                let futures = subnet_ids.iter().map(|subnet_id| async move {
+                    // Construct proof for prior slot.
+                    let proof_slot = slot - 1;
```
dknopik, line 720: Do you know why this is needed (here and above)? It seems weird to me to compute the proof for the previous slot.

chong-he: I honestly don't know why. @michaelsproul might have an answer for this.

In the distributed mode, I follow the same logic to also compute `slot - 1`

dknopik:
I am still confused about the `slot - 1`. But when removing it, this PR looks good to me from the perspective of the functionality needed by Anchor. @michaelsproul do you know why that is there? By my understanding of the Altair spec, the proof for the previous slot is not at all relevant for the current slot.

validator_client/validator_services/src/sync.rs
michaelsproul: @dknopik Sync committee messages have a confusing number of offsets from the current slot built in, but I _think_ what we have now is correct.

From: https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/validator.md#prepare-sync-committee-message

> If a validator is in the current sync committee (i.e. is_assigned_to_sync_committee() above returns True), then for every slot in the current sync committee period, the validator should prepare a SyncCommitteeMessage for the previous slot (slot - 1) according to the logic in get_sync_committee_message as soon as they have determined the head block of slot - 1. This means that when assigned to slot a SyncCommitteeMessage is prepared and broadcast in slot-1 instead of slot.

And then in the aggregate verification, we have another offset to deal with the fact that sync contributions and aggregates are for the previous slot (https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/p2p-interface.md#sync_committee_contribution_and_proof):

```
# Committees assigned to `slot` sign for `slot - 1`
# This creates the exceptional logic below when transitioning between sync committee periods
```

This is super confusing to reason about though, so I think we should definitely add a test (maybe in the simulator) to assert that we don't mess up the contribution at the period boundary.

dknopik: Ah, ok, thanks. I understand that we are treating `slot` as the duty slot, and `proof_slot` (= `slot - 1`) as the slot in which we have to actually sign the head root (`SyncCommitteeMassage`) and, if we are an aggregator, the contribution. However, in the context of this code, we start iterating over the slots to (pre-)compute selection proofs for starting at `SlotClock::now()` (let's call that slot `X`). However, for slot `X`, everything to do was done already, as we sign the `SyncCommitteeMessage` at `(X-1)+4s` and the contribution at `(X-1)+8s`. So there is no need at all to compute the selection proof for duty at slot `X` anymore.

So I guess the fix is to iterate from `X + 1` to `pre_compute_slot + 1` instead of from `X` to `pre_compute_slot`. Which is _almost_ equivalent to removing `let proof_slot = slot - 1;`, except for one check and logging.

michaelsproul: Oh yeah I think you are right. I think we should raise a PR with a test for this separate to this PR

dknopik (approved):
Now that the off-by-one is fixed, this is good to go from an Anchor point of view :)

Thank you @chong-he! 

@michaelsproul can you please take a look when you have time?

validator_client/validator_services/src/sync.rs
```diff
@@ -498,138 +486,301 @@ pub async fn poll_sync_committee_duties_for_period<S: ValidatorStore, T: SlotClo
...
-                        "Arithmetic error computing subnet IDs"
-                    );
-                    continue;
-                }
-            };
+        // For distributed mode
+        if duties_service
+            .sync_duties
+            .selection_proof_config
+            .parallel_sign
```
michaelsproul, line 614: I think we could think about re-combining these two cases in future, as the only substantial difference is the usage of `FuturesUnordered` in the non-distributed case. The non-distributed case is _still parallel_ but it waits for all of the proofs to come in before writing them to the in-memory map. Whereas in the distributed case we add proofs as they become available. `FuturesUnordered` would likely be fine for both cases, and would save us a lot of duplication of logic.

I've opened an issue:

- https://github.com/sigp/lighthouse/issues/7983

michaelsproul (approved):
I've pushed two commits to address my review comments. The most important change was removing `.data[0]` which can panic.

- https://github.com/sigp/lighthouse/pull/7016/commits/239160e494a399356bd8eac0bf2bab502c2c2e22
- https://github.com/sigp/lighthouse/pull/7016/commits/b7d6d2bd147550c0e0405518fa37ca10a18a3723

PR #7017, opened 2025-02-20
Remove LC server config from HTTP API

michaelsproul:
## Issue Addressed

Partly addresses

- https://github.com/sigp/lighthouse/issues/6959

## Proposed Changes

Use the `enable_light_client_server` field from the beacon chain config in the HTTP API. I think we can make this the single source of truth, as I think the network crate also has access to the beacon chain config.

### DIFF
--- a/beacon_node/http_api/src/lib.rs
+++ b/beacon_node/http_api/src/lib.rs
@@ -155,7 +155,6 @@ pub struct Config {
     pub enable_beacon_processor: bool,
     #[serde(with = "eth2::types::serde_status_code")]
     pub duplicate_block_status_code: StatusCode,
-    pub enable_light_client_server: bool,
     pub target_peers: usize,
 }
 
@@ -171,7 +170,6 @@ impl Default for Config {
             sse_capacity_multiplier: 1,
             enable_beacon_processor: true,
             duplicate_block_status_code: StatusCode::ACCEPTED,
-            enable_light_client_server: true,
             target_peers: 100,
         }
     }
@@ -294,18 +292,6 @@ pub fn prometheus_metrics() -> warp::filters::log::Log<impl Fn(warp::filters::lo
         );
         metrics::observe_timer_vec(&metrics::HTTP_API_PATHS_TIMES, &[path], info.elapsed());
     })
-}
-
-fn enable(is_enabled: bool) -> impl Filter<Extract = (), Error = warp::Rejection> + Clone {
-    warp::any()
-        .and_then(move || async move {
-            if is_enabled {
-                Ok(())
-            } else {
-                Err(warp::reject::not_found())
-            }
-        })
-        .untuple_one()
 }
 
 /// Creates a server that will serve requests using information from `ctx`.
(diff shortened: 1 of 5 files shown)


beacon_node/http_api/src/lib.rs
```diff
@@ -2467,11 +2466,13 @@ pub fn serve<T: BeaconChainTypes>(
...
         .and(warp::header::optional::<api_types::Accept>("accept"))
         .then(
-            |chain: Arc<BeaconChain<T>>,
+            |light_client_server_enabled: Result<(), Rejection>,
+             chain: Arc<BeaconChain<T>>,
              task_spawner: TaskSpawner<T::EthSpec>,
              block_root: Hash256,
              accept_header: Option<api_types::Accept>| {
                 task_spawner.blocking_response_task(Priority::P1, move || {
+                    light_client_server_enabled?;
```
jimmygchen, line 2475: Any reason why not to place this above the task spawner so we avoid spawning a blocking task just to return the error? I noticed we're also doing the same thing on other endpoints.

beacon_node/http_api/src/lib.rs
```diff
@@ -2485,10 +2486,12 @@ pub fn serve<T: BeaconChainTypes>(
         .and(warp::path::end())
         .and(warp::header::optional::<api_types::Accept>("accept"))
         .then(
-            |chain: Arc<BeaconChain<T>>,
+            |light_client_server_enabled: Result<(), Rejection>,
+             chain: Arc<BeaconChain<T>>,
              task_spawner: TaskSpawner<T::EthSpec>,
              accept_header: Option<api_types::Accept>| {
                 task_spawner.blocking_response_task(Priority::P1, move || {
+                    light_client_server_enabled?;
```
jimmygchen, line 2494: Same as above.

beacon_node/http_api/src/lib.rs
```diff
@@ -2580,11 +2585,13 @@ pub fn serve<T: BeaconChainTypes>(
...
         .and(warp::header::optional::<api_types::Accept>("accept"))
         .then(
-            |chain: Arc<BeaconChain<T>>,
+            |light_client_server_enabled: Result<(), Rejection>,
+             chain: Arc<BeaconChain<T>>,
              task_spawner: TaskSpawner<T::EthSpec>,
              query: LightClientUpdatesQuery,
              accept_header: Option<api_types::Accept>| {
                 task_spawner.blocking_response_task(Priority::P1, move || {
+                    light_client_server_enabled?;
```
jimmygchen, line 2594: Same as above.

beacon_node/http_api/src/lib.rs
michaelsproul: We need to return a future here, so we'd have to double nest `async move`, which is probably fine.

But also, I don't think we should care about performance for endpoints that users shouldn't be calling anyway. The common case is that you might call this once, it fails, and then you restart the node with the light client server enabled

jimmygchen (approved):
Looks good to me!

PR #7021, opened 2025-02-21
update codeowners, to be more specific

jxs:
## Issue Addressed

I keep being notified for PR's like https://github.com/sigp/lighthouse/pull/7009 where it doesn't touch the specified directories in the `CODEOWNERS` file.
After reading the [docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) not having a forward slash in beginning of the path means:
>  In this example, @octocat owns any file in an apps directory
> anywhere in your repository.

whereas with the slashes:
> In this example, @doctocat owns any file in the `/docs`
> directory in the root of your repository and any of its
> subdirectories.

this update makes it more rigid for the files in the jxs ownership

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,2 @@
-beacon_node/network/                @jxs
-beacon_node/lighthouse_network/     @jxs
+/beacon_node/network/                @jxs
+/beacon_node/lighthouse_network/     @jxs


michaelsproul:
I think the reason you were notified for 7009 is that we changed the target branch, which brought in everything from `unstable` to the PR, and then we rebased so the commits from unstable were no longer part of the PR

It might be better if we rebase first and then change the target branch, but github actions has a quirk where it won't re-trigger the target branch check when the target changes, meaning another rebase/dummy commit is required to get it to update

michaelsproul (approved):
Change looks to good to me regardless. I just don't think it will silence all the false positives because there are lots of PRs that either temporarily touch these files (due to merge commits/rebases) or which make very minor lint-level changes.

michaelsproul:
@mergify requeue

Issue #6875, opened 2025-01-28
Enable arithmetic lint in rate-limiter


michaelsproul:
The rate-limiter is called prior to application-level checks on request validity, so we should be careful to avoid integer overflow in its code, as it is operating on untrusted data.

There is integer overflow here:

https://github.com/sigp/lighthouse/blob/029b4f21047c37d1ffde51c554737b1aa0880f97/beacon_node/lighthouse_network/src/rpc/rate_limiter.rs#L416-L426

The `tokens` value is derived from untrusted user input:

https://github.com/sigp/lighthouse/blob/029b4f21047c37d1ffde51c554737b1aa0880f97/beacon_node/lighthouse_network/src/rpc/rate_limiter.rs#L327-L333

The `count` from the request is untrusted and can be set arbitrarily:

https://github.com/sigp/lighthouse/blob/029b4f21047c37d1ffde51c554737b1aa0880f97/beacon_node/lighthouse_network/src/rpc/protocol.rs#L794-L811

This was briefly fixed for `BlobsByRange` requests on `unstable` when we implemented a size check during decoding, but will regress again when this PR is merged:

- https://github.com/sigp/lighthouse/pull/6847

The fix that existed in unstable was coincidental, so I think it is better if we address the root cause and avoid overflowing integer arithmetic in the rate limiter altogether. We can probably replace most uses of `+`, `-`, `*` by their `saturating_` equivalents to get the right semantics.

We can also enable the [`arithmetic_side_effects`](https://rust-lang.github.io/rust-clippy/master/index.html#arithmetic_side_effects) lint, which we use in `consensus` crates. See:

- https://github.com/sigp/lighthouse/issues/1013

This issue is not deemed security-sensitive because other mechanisms exist to defend the node from peers that would bypass the rate limiter. Any request with an invalid `count` will be rejected by application-level checks and result in the peer being quickly banned.

Thanks to `@gln` for reporting this bug as part of the Immunefi Ethereum Attackathon.

jimmygchen:
Addressed in #7025 🎉

PR #7019, opened 2025-02-20
Rust 1.85 lints

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

2 changes:
1. Replace Option::map_or(true, ...) with is_none_or(...)
2. Remove unnecessary `Into::into` blocks where the type conversion is apparent from the types

PR #7025, opened 2025-02-21
Enable arithmetic lint in rate-limiter

ackintosh:
## Issue Addressed

https://github.com/sigp/lighthouse/issues/6875

## Proposed Changes

- Enabled the linter in rate-limiter and fixed errors.
- Changed the type of `Quota::max_tokens` from `u64` to `NonZeroU64` because `max_tokens` cannot be zero.
- Added a test to ensure that a large value for `tokens`, which causes an overflow, is handled properly.


ackintosh:
- [`check-code` failure ](https://github.com/sigp/lighthouse/actions/runs/13466656894/job/37633770391?pr=7025)will be solved in #7019
- [`release-tests-ubuntu` failure](https://github.com/sigp/lighthouse/actions/runs/13466656894/job/37633774494?pr=7025): We are facing an issue with downloading web3signer, possibly due to an expired github token.

jimmygchen:
@ackintosh ready to review this one - would you mind resolving the conflicts please?

ackintosh:
@jimmygchen Thanks! I've resolved the conflicts and updated the PR.

beacon_node/lighthouse_network/src/rpc/config.rs
```diff
@@ -100,24 +100,30 @@ pub struct RateLimiterConfig {
 }
 
 impl RateLimiterConfig {
-    pub const DEFAULT_PING_QUOTA: Quota = Quota::n_every(2, 10);
-    pub const DEFAULT_META_DATA_QUOTA: Quota = Quota::n_every(2, 5);
-    pub const DEFAULT_STATUS_QUOTA: Quota = Quota::n_every(5, 15);
+    pub const DEFAULT_PING_QUOTA: Quota = Quota::n_every(NonZeroU64::new(2).unwrap(), 10);
```
jimmygchen, line 103: We could use a `const fn` like the below to ensure compile-time guarantee that the value provided is valid, ~~whereas `unwrap` is performed at runtime:~~
https://github.com/sigp/lighthouse/blob/b0c374c1cabd7fd960b4dde9ec9b0b26a960ca0c/consensus/types/src/non_zero_usize.rs#L3-L8

jimmygchen: Oops I got it wrong here, `unwrap` will indeed result in a  at compile time error, please ignore me!

jimmygchen (approved):
Nice, thanks for adding the tests and the explanation!

Issue #6767, opened 2025-01-08
Implement basic validator data column custody


jimmygchen:
## Description

Implement spec: https://github.com/ethereum/consensus-specs/pull/3871

This isn't mandatory for `peerdas-devnet-4`, but the change is compatible with the devnet - it means potentially storing more columns when running a node with validators. The spec hasn't been merged yet but client teams have started looking into this. We could either wait for the spec to be merged or implement this behind a CLI flag for testing purpose until the spec change is merged.

## Tasks

Following the design in https://hackmd.io/@dapplion/validator_custody_no_backfill

- [x] Remove CGC from data_availability checker #7033
- [ ] make all codepaths read the CGC from a Mutex,
- [ ] carefully persist new consistency data to disk
```rust
cgc_updates: Vec<(Slot, u64)>,
backfilled_cgc: (Slot, u64),
```
- [x] restart the backfill under certain conditions (separated issue created #7603)
- [x] update gossip subscriptions
- [ ] cache the to-advertise CGC somewhere and update ENR + status at that moment

dapplion:
Linking this WIP while I focus on Holesky fixes https://github.com/sigp/lighthouse/compare/unstable...dapplion:lighthouse:validator-custody-state

jimmygchen:
Basic validator custody covered in #7578 🎉

PR #7033, opened 2025-02-24
Remove CGC from data_availability checker

dapplion:
## Issue Addressed

- Part of https://github.com/sigp/lighthouse/issues/6767

Validator custody makes the CGC and set of sampling columns dynamic. Right now this information is stored twice:
- in the data availability checker
- in the network globals

If that state becomes dynamic we must make sure it is in sync updating it twice, or guarding it behind a mutex. However, I noted that we don't really have to keep the CGC inside the data availability checker. All consumers can actually read it from the network globals, and we can update `make_available` to read the expected count of data columns from the block.


beacon_node/beacon_chain/src/block_verification_types.rs
```diff
@@ -104,6 +109,8 @@ impl<E: EthSpec> RpcBlock<E> {
         Self {
             block_root,
             block: RpcBlockInner::Block(block),
+            // Block has zero columns
+            custody_columns_count: 0,
```
dapplion, line 113: An RPC blocks without columns will never read this value

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -631,19 +621,16 @@ impl<T: BeaconChainTypes> DataAvailabilityCheckerInner<T> {
...
         if pending_components.reconstruction_started {
             return ReconstructColumnsDecision::No("already started");
         }
-        if custody_column_count != total_column_count {
-            return ReconstructColumnsDecision::No("not required for full node");
-        }
         if received_column_count >= total_column_count {
             return ReconstructColumnsDecision::No("all columns received");
         }
+        // Only supernodes receive >= 50% of columns
```
dapplion: Removed the "not required for full node" condition as it's redundant and would force us to pipe the custody group here somehow

beacon_node/beacon_chain/src/historical_blocks.rs
```diff
@@ -94,27 +94,12 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
-            .count()
-            .saturating_mul(n_blob_ops_per_block);
-
         let mut expected_block_root = anchor_info.oldest_block_parent;
         let mut prev_block_slot = anchor_info.oldest_block_slot;
         let mut new_oldest_blob_slot = blob_info.oldest_blob_slot;
         let mut new_oldest_data_column_slot = data_column_info.oldest_data_column_slot;
 
-        let mut blob_batch = Vec::<KeyValueStoreOp>::with_capacity(blob_batch_size);
+        let mut blob_batch = Vec::<KeyValueStoreOp>::new();
```
dapplion, line 101: Using `with_capacity` here feels like a micro-optimization unless I'm missing something. Dropped it as we don't have access to the column count here anymore

beacon_node/beacon_chain/src/test_utils.rs
```diff
@@ -815,6 +815,11 @@ where
         (0..self.validator_keypairs.len()).collect()
     }
 
+    pub fn get_sampling_column_count(&self) -> usize {
+        // Default column custody count
+        8
```
dapplion: This should have no effect on tests I believe

beacon_node/network/src/sync/tests/range.rs
```diff
@@ -449,7 +449,15 @@ fn build_rpc_block(
             RpcBlock::new(None, block, Some(blobs.clone())).unwrap()
         }
         Some(DataSidecars::DataColumns(columns)) => {
-            RpcBlock::new_with_custody_columns(None, block, columns.clone(), spec).unwrap()
+            RpcBlock::new_with_custody_columns(
+                None,
+                block,
+                columns.clone(),
+                // TODO
+                columns.len(),
```
dapplion, line 457: Assumes the CGC value is equal to all the columns received, which makes sense. We can change it later if we want to do more complex tests

beacon_node/beacon_chain/src/block_verification_types.rs
```diff
@@ -145,13 +152,16 @@ impl<E: EthSpec> RpcBlock<E> {
         Ok(Self {
             block_root,
             block: inner,
+            // Block is before PeerDAS
+            custody_columns_count: 0,
         })
     }
```
ethDreamer, line 159: might be worth having consistent terminology for:
1. Blocks
2. Post-Deneb Blocks (blocks with blobs)
3. Post-PeerDAS Blocks (blocks with custody columns)

Not sure what terminology that would be, but if we don't already have one / can't think of one, we should at least rename the above `new` function to something like `new_with_blobs` or something.

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -195,8 +195,14 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
                     ));
                 }
 
-                RpcBlock::new_with_custody_columns(Some(block_root), block, custody_columns, spec)
-                    .map_err(|e| format!("{e:?}"))?
+                RpcBlock::new_with_custody_columns(
+                    Some(block_root),
+                    block,
+                    custody_columns,
+                    expects_custody_columns.len(),
```
ethDreamer, line 264: The `expects_custody_columns` field on `RangeBlockComponentsRequest` could use a documentation comment

beacon_node/beacon_chain/src/block_verification_types.rs
dapplion: I want to get rid of RpcBlock completely, so this assignment here will become irrelevant. Instead have a chain segment as an enum as you suggest

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
ethDreamer: I had a quick look and didn't see any existing check to ensure this wasn't being called on a non-full node. Why do you say it's redundant?

Could you not do this check with:
```rust
if let Some(custody_columns_count) = pending_components
    .executed_block
    .as_ref()
    .map(|diet_block| diet_block.custody_columns_count())
{
    if custody_columns_count != total_column_count {
        return ReconstructColumnsDecision::No("not required for full node");
    }
}
```

beacon_node/lighthouse_network/src/types/globals.rs
```diff
@@ -33,6 +33,8 @@ pub struct NetworkGlobals<E: EthSpec> {
     /// The computed sampling subnets and columns is stored to avoid re-computing.
     pub sampling_subnets: HashSet<DataColumnSubnetId>,
     pub sampling_columns: HashSet<ColumnIndex>,
+    /// Constant custody group count (CGC) set at startup
+    custody_group_count: u64,
```
ethDreamer, line 37: You think it might be worth using one naming convention (like `custody_group_count`) everywhere instead of also using `custody_columns_count`?

ethDreamer: Also this isn't behind any `RwLock` or anything but it was my understanding that this could change as the node synced more columns / added more validators? Or is that just not implemented yet?

ethDreamer (dismissed):
some tests still need to be fixed

beacon_node/lighthouse_network/src/types/globals.rs
dapplion: not implemented yet

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
dapplion: What you suggest will cause weird behavior:
- if block is there: reconstruction happens only if custody_columns_count = 100%
- if block is not there: reconstruction happens only if custody_columns_count = 50%

Block being there or not before attempting to reconstruction depends on network conditions and timings.

beacon_node/network/src/sync/block_sidecar_coupling.rs
dapplion: Added

beacon_node/beacon_chain/src/test_utils.rs
dapplion: It did 🤣 fixed with 54608d822

dapplion:
CI passing now

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: Are nodes that custody 50% of columns considered supernodes?
I'm not sure if there's any reason to  reconstruct for these nodes, as they are not required to re-publish to their non-custody columns.

> If the node is subscribed to the subnet corresponding to the column, it MUST send the reconstructed DataColumnSidecar to its topic mesh neighbors. If instead the node is not subscribed to the corresponding subnet, it SHOULD still expose the availability of the DataColumnSidecar as part of the gossip emission process.

I don't quite understand why it means to "expose the availability of the DataColumnSidecar as part of the gossip emission process"

https://github.com/ethereum/consensus-specs/blob/dev/specs/fulu/das-core.md#reconstruction-and-cross-seeding

jimmygchen: Also, wouldn't this change cause this error since we're sending 128 columns to the  DA checker upon successful reconstruction:
> "too many columns got {num_received_columns} expected {num_expected_columns}"

If we do end up reconstructing for these nodes that don't custody 100% of columns, we would have to filter before putting them into the DA checker.

beacon_node/lighthouse_network/src/types/globals.rs
```diff
@@ -121,6 +120,16 @@ impl<E: EthSpec> NetworkGlobals<E> {
         self.listen_multiaddrs.read().clone()
     }
 
+    /// Returns true if this node is configured as a PeerDAS supernode
+    pub fn is_supernode(&self) -> bool {
+        self.custody_group_count == self.spec.number_of_columns
```
jimmygchen:
```suggestion
        self.custody_group_count == self.spec.number_of_custody_groups
```

beacon_node/beacon_chain/src/test_utils.rs
```diff
@@ -615,6 +615,13 @@ where
 
         let chain = builder.build().expect("should build");
 
+        let sampling_column_count = if self.import_all_data_columns {
+            chain.spec.number_of_custody_groups as usize
+        } else {
+            // Default minimum value
+            8
```
jimmygchen: nit: could just do `chain.spec.custody_requirement`?

beacon_node/beacon_chain/tests/block_verification.rs
```diff
@@ -1810,3 +1829,14 @@ async fn import_execution_pending_block<T: BeaconChainTypes>(
         }
     }
 }
+
+fn get_cgc<E: EthSpec>(blobs_opt: &Option<DataSidecars<E>>) -> usize {
+    if let Some(data_sidecars) = blobs_opt.as_ref() {
+        match data_sidecars {
+            DataSidecars::Blobs(b) => b.len(),
```
jimmygchen: CGC is not relevant for blobs, might as well just return 0?

beacon_node/beacon_chain/tests/block_verification.rs
```diff
@@ -991,6 +992,7 @@ async fn block_gossip_verification() {
     let (chain_segment, chain_segment_blobs) = get_chain_segment().await;
 
     let block_index = CHAIN_SEGMENT_LENGTH - 2;
+    let cgc = 0;
```
jimmygchen: I think CGC isn't relevent for gossip verifcation, but we could still use  `spec.custody_requirement` here - the tests are run for every fork so it would be good if the functions are called in a similar way to production.

beacon_node/beacon_chain/tests/block_verification.rs
```diff
@@ -1339,7 +1354,7 @@ async fn verify_block_for_gossip_slashing_detection() {
         )
         .await
         .unwrap();
-    unwrap_err(harness.chain.verify_block_for_gossip(block2).await);
+    unwrap_err(harness.chain.verify_block_for_gossip(block2, 0).await);
```
jimmygchen: same as the earlier comment about CGC. It would never be 0 in Fulu.

beacon_node/beacon_chain/tests/payload_invalidation.rs
```diff
@@ -1050,7 +1050,7 @@ async fn invalid_parent() {
 
     // Ensure the block built atop an invalid payload is invalid for gossip.
     assert!(matches!(
-        rig.harness.chain.clone().verify_block_for_gossip(block.clone()).await,
+        rig.harness.chain.clone().verify_block_for_gossip(block.clone(), 0).await,
```
jimmygchen: Same as earlier comment about CGC in gossip verification tests - be good to make input as close to prod as possible.

beacon_node/network/src/sync/tests/range.rs
jimmygchen: Sounds good to me - in that case we could remove the TODO or add a note for future readers.

beacon_node/lighthouse_network/src/types/globals.rs
```diff
@@ -121,6 +120,16 @@ impl<E: EthSpec> NetworkGlobals<E> {
...
+    }
+
+    /// Returns the count of custody columns this node must sample for block import
+    pub fn custody_columns_count(&self) -> u64 {
+        self.custody_group_count
```
jimmygchen: This is `custody_group_count` not `custody_columns`. If `custody_columns` is what we want then we should do the conversion when setting this.

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: > Removed the "not required for full node" condition as it's redundant and would force us to pipe the custody group here somehow

Piping this through should be fairly straight forward as reconstruction are called from `NetworkBeaconProcessor` and has access to `globals`

https://github.com/sigp/lighthouse/blob/187467de28b37c5b933b8aba046c01a9aa32e839/beacon_node/network/src/network_beacon_processor/mod.rs#L1008-L1012

jimmygchen (changes requested):
@dapplion this is looking good! and thanks for the first round of review @ethDreamer 🙏 
I've added a few comments. Let me know what you think!

beacon_node/lighthouse_network/src/types/globals.rs
dapplion: Fixed with fea34b4b31c24f4f633b137b039615c8d7522c29

beacon_node/lighthouse_network/src/types/globals.rs
dapplion: Fixed with fea34b4b31c24f4f633b137b039615c8d7522c29

beacon_node/beacon_chain/tests/block_verification.rs
dapplion: 2ba26543e5f039b1ee9774bb1c4c2f21dfa2a5e1

beacon_node/beacon_chain/tests/block_verification.rs
dapplion: 2ba26543e5f039b1ee9774bb1c4c2f21dfa2a5e1

beacon_node/beacon_chain/tests/payload_invalidation.rs
dapplion: 2ba26543e5f039b1ee9774bb1c4c2f21dfa2a5e1

beacon_node/beacon_chain/tests/block_verification.rs
dapplion: 2ba26543e5f039b1ee9774bb1c4c2f21dfa2a5e1

beacon_node/beacon_chain/src/test_utils.rs
dapplion: 2ba26543e5f039b1ee9774bb1c4c2f21dfa2a5e1

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: @dapplion made a fix here to ensure this isn't called on a non supernode
https://github.com/sigp/lighthouse/pull/7033/commits/62660577f7dd8f326ba906ef5e003584c29453dc

jimmygchen (approved):
LGTM! be great to test this on a peerdas local devnet before merging.

jimmygchen:
I've tested this locally with Kurtosis and is working as expected.

PR #7034, opened 2025-02-25
Cache validator registration only after successful publish

realbigsean:
We are caching validator registrations after signing is successful rather than after publish is successful. And the purpose of this cache is to suppress publishing. Here, I move the caching to post-publish.

We re-publish registrations regardless of this cache every epoch so this isn't a huge issue. The place where this helps is kurtosis testing with the `mev` flow. On starting a new testnet, we fail the initial publish, and this bug means we need to wait till the next epoch to actually test the `mev` flow.


jimmygchen (approved):
Nice, good find!

Issue #7040, opened 2025-02-26
Holesky Rescue [INFORMATIONAL]


paulhauner:
# Holesky Rescue - Feb 2025

This issue acts as a "front page" for the rescue efforts on Holesky. Holesky failed the Electra upgrade due to a config issue in some ELs.

The config issue in Nethermind, Geth & Besu resulted in an invalid block being justified. This means that for some users it will be impossible to start attesting on the valid chain without being slashed. Lighthouse will natively protect against this slashing; if you are affected your VC will log errors about slashing protection but you will not be slashed.

The community is still developing a unified approach to solving the slashing issue, but it is a strong possibility that these validators are destined to be slashed or leaked out. Even if your validators cannot attest, they can still produce blocks which is very valuable. **We encourage all Holesky validators to try and get their setups online and functioning.**

## Lighthouse Advice

If you're a Holesky validator, we recommend you use the [v7.0.0-beta.1 release](https://github.com/sigp/lighthouse/releases/tag/v7.0.0-beta.1).

There are some VC flags that may be relevant to your scenario:

- `--beacon-nodes-sync-tolerances 1000,2000,3000`
    - This flag may no longer be necessary due to recent improvements in block frequency.
- `--disable-attesting`
    - Only run this flag if you are struggling to sync a CL and EL pair and/or running out of memory. Once you are synced, remove this flag to allow your validator to attest to the canonical chain.

No additional flags are required for the BN.

## Disabling Slashing Protection

DO NOT DO THIS UNTIL INSTRUCTED TO DO SO

If your validators won't attest due to CRIT errors from the slashing protection database, you should _disable_ slashing protection in coordination with the rest of the validator set (monitor Discord/Telegram). To disable slashing protection in Lighthouse:

1. Stop the VC. Delete the `slashing_protection.sqlite` file located in `$datadir/validators`. You should also delete `slashing_protection.sqlite-journal`.
2. Start the VC again, using the flag `--init-slashing-protection`. The flag is required to force Lighthouse to start in the absence of a slashing protection DB. A new slashing protection DB will be created, but will not contain any of the historic information that would block your validators from signing.

Once slashing protection is disabled, your validators will most likely get slashed. Our hope is that they contribute to finalizing Holesky before doing so (hence the importance of a coordinated slashing).

## EL Client Advice

- Make sure your EL is running a patched version (not relevant for Reth & Erigon, which do not have a config bug).
- If your EL was affected by the config issue, you may need to drop the EL database and resync.
[...]

PR #7041, opened 2025-02-26
v7.0.0-beta.2 (Holesky Rescue)

paulhauner:
## Issue Addressed

Related to https://github.com/sigp/lighthouse/issues/7040

## Proposed Changes

Various hacks and tweaks to ensure LH syncs a valid Holesky chain.

Most notably, we explicitly reject the invalid payload that was justified (`0x2db899881ed8546476d0b92c6aa9110bea9a4cd0dbeb5519eb0ea69575f1f359`).

## Additional Info

- **Use this branch if you are running a node on Holesky**


validator_client/src/lib.rs
```diff
@@ -325,7 +325,7 @@ impl<E: EthSpec> ProductionValidatorClient<E> {
                     get_validator_block: slot_duration / HTTP_GET_VALIDATOR_BLOCK_TIMEOUT_QUOTIENT,
                 }
             } else {
-                Timeouts::set_all(slot_duration)
+                Timeouts::set_all(25 * slot_duration)
```
michaelsproul, line 328: @macladson could you make a clean PR for a CLI flag accomplishing the same effect against the `release-v7.0.0` branch? :pray:

michaelsproul: PR is now here: https://github.com/sigp/lighthouse/pull/7047

michaelsproul:
Trying to get CI running/passing on this so we have a bit more confidence in the release.

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -1505,34 +1518,50 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
-                .put_state(*state_root, block_root, &state)?;
-            debug!(
-                self.log,
-                "Cached state";
-                "state_root" => ?state_root,
-                "slot" => state.slot(),
-            );
+            if update_cache {
+                state.update_tree_hash_cache()?;
+                state.build_all_caches(&self.spec)?;
```
michaelsproul: @jimmygchen This is why the test is failing, we are now only building the state caches when `update_cache` is `true`. We should do it all the time so that the caller gets nice states with caches built.

jimmygchen: Nice, I've made a fix in https://github.com/sigp/lighthouse/pull/7071. Thanks!

DanielJP-Crypto (approved):

michaelsproul:
Closing as no longer relevant.

PR #7042, opened 2025-02-26
Add block ban flag `--invalid-block-roots`

eserilev:
Add a new flag that allows users to ban block roots listed in a file. Also make sure the hardcoded invalid block root is only considered on holesky


jimmygchen (approved):
Nice, LGTM!

michaelsproul:
I've cherry-picked _all_ the block ban related changes to this PR and rebased on `release-v7.0.0` so we can have a clean history with each feature in its own PR.

michaelsproul:
I reverted the `ignore` for the tests, because they were caused by the fork choice invalidation changes, which we do not want to keep.

jimmygchen (approved):
Looks good 👍

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -281,7 +295,10 @@ pub enum BlockError {
     /// If it's actually our fault (e.g. our execution node database is corrupt) we have bigger
     /// problems to worry about than losing peers, and we're doing the network a favour by
     /// disconnecting.
-    ParentExecutionPayloadInvalid { parent_root: Hash256 },
+    ParentExecutionPayloadInvalid {
+        parent_root: Hash256,
+    },
+    KnownInvalidExecutionPayload(Hash256),
```
dapplion, line 287: Add comment on top of this variant to have same format

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -1326,6 +1343,19 @@ impl<T: BeaconChainTypes> ExecutionPendingBlock<T> {
         chain: &Arc<BeaconChain<T>>,
         notify_execution_layer: NotifyExecutionLayer,
     ) -> Result<Self, BlockError> {
+        let invalid_holesky_block = {
+            if let Ok(invalid_block_root) = Hash256::from_str(
+                "2db899881ed8546476d0b92c6aa9110bea9a4cd0dbeb5519eb0ea69575f1f359",
```
dapplion: Is it intentional to hardcode a block here? You can add this block root in the default value of the CLI flag invalid_block_roots

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -1326,6 +1343,19 @@ impl<T: BeaconChainTypes> ExecutionPendingBlock<T> {
...
+            ) {
+                block_root == invalid_block_root && chain.spec.deposit_chain_id == 17000
+            } else {
+                false
+            }
+        };
+        if chain.config.invalid_block_roots.contains(&block_root) || invalid_holesky_block {
```
dapplion: This flag can be exploited for censoring, you should restrict usage to a specific network

eserilev: ive constrained this flag to holesky only

beacon_node/beacon_chain/src/block_verification.rs
eserilev: this block root is now added when we initialize the config value

beacon_node/beacon_chain/src/block_verification.rs
eserilev: added a comment

beacon_node/beacon_chain/src/block_verification.rs
michaelsproul: I disagree with both of Lion's suggested changes, sorry :sweat_smile: 

I think we need the flag to be not limited to Holesky so that we can potentially use it to get out of trouble on other catastrophically failing networks (without having to patch). We don't have another solution for preventing optimistic sync of the wrong chain at the moment.

I think this also means we should keep the hardcoded Holesky block as a separate check. It doesn't make sense as the flag default if the flag works on multiple networks.

eserilev: got it, I had assumed that this was strictly a holesky thing. If the plan is to keep this flag until we have a better solution for optimistic sync of the wrong chain, I'm cool with that

eserilev: i've reverted the changes back to what they were originally

jimmygchen (approved):
Re-reviewed, looks good!

beacon_node/beacon_chain/src/block_verification.rs
realbigsean: We can get rid of this when we implement checkpoint sync from an unfinalized checkpoint

jimmygchen: @realbigsean would that prevent us from importing an invalid block though? wouldn't sync still attempt to sync with the invalid chain?

jimmygchen:
I'm going to move the invalid block check so it happens sooner.

beacon_node/beacon_chain/src/block_verification.rs
realbigsean: you’d have to use a checkpoint ahead of the invalid block, and since we’d backfill from there we’d avoid it. Part of the complexity of unfinalized checkpoint sync would be making sure you ignore chain tips older than the checkpoint

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -800,6 +800,18 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+            // Penalise peers for sending us banned blocks.
+            BlockError::KnownInvalidExecutionPayload(block_root) => {
+                warn!(
+                    self.log,
+                    "Received block known to be invalid";
+                    "block_root" => ?block_root,
+                );
+                Err(ChainSegmentFailed {
+                    message: format!("Banned block: {block_root:?}"),
+                    peer_action: Some(PeerAction::LowToleranceError),
```
michaelsproul: Let's do `Fatal` here

michaelsproul (changes requested):

beacon_node/network/src/network_beacon_processor/sync_methods.rs
eserilev: 4aa59b6ab2aeceafee52f97c0eb5d2eb8f7b7445

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3324,6 +3340,8 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         block_source: BlockImportSource,
         notify_execution_layer: NotifyExecutionLayer,
     ) -> Result<AvailabilityProcessingStatus, BlockError> {
+        self.check_invalid_block_roots(block_root)?;
```
michaelsproul: I don't think doing the check here is sufficient to prevent the parent state load. This function is called from `gossip_methods.rs` _after_ the `GossipVerifiedBlock` is constructed:

https://github.com/sigp/lighthouse/blob/0f5e6801499bf230d249e851b639b58dac5dcccc/beacon_node/network/src/network_beacon_processor/gossip_methods.rs#L1634-L1642

I think we need to get in there before the block is fully gossip verified, as in the case of a proposer shuffling cache miss (likely), we will load the parent here:

https://github.com/sigp/lighthouse/blob/0f5e6801499bf230d249e851b639b58dac5dcccc/beacon_node/beacon_chain/src/block_verification.rs#L922-L924 

I think the check should probably be in `block_verification.rs` somewhere. I'll have a look and see what the best option is.

jimmygchen: Ah good catch. 
What about beginning of the `GossipVerifiedBlock` constructor, and return block error as soon as it detects invalid root:

https://github.com/sigp/lighthouse/blob/505b4be45737b3e9a58c84475afa93623277a989/beacon_node/beacon_chain/src/block_verification.rs#L808-L811

michaelsproul: RPC blocks don't go through this path, but maybe we could add a separate check for them?

eserilev: [e2499a7](https://github.com/sigp/lighthouse/pull/7042/commits/e2499a790e5a859e1844b2353fc03307cd6e0df9)

I'm a bit iffy on the check I made for RPC blocks. I _think_ it does the job and avoids any parent state loads

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3340,6 +3356,17 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         self.remove_notified(&block_root, r)
     }
 
+    /// Check for known and configured invalid block roots before processing.
+    pub fn check_invalid_block_roots(&self, block_root: Hash256) -> Result<(), BlockError> {
+        let is_invalid_holesky_block =
+            block_root == *INVALID_HOLESKY_BLOCK_ROOT && self.spec.deposit_chain_id == 17000;
```
dapplion: You should move this hardcoded value to `beacon_node/src/config.rs` and make it part of the flag parsing logic. Do we want to offer a way to run the binary and accept this invalid block root? i.e. make the default flag value equal to `INVALID_HOLESKY_BLOCK_ROOT` and if someone runs `--invalid-block-roots ""` the check is disabled. This can be useful to do Holesky testing or other purposes

beacon_node/src/config.rs
```diff
@@ -903,6 +904,34 @@ pub fn get_config<E: EthSpec>(
...
+                            "block_root" => s,
+                            "error" => ?e,
+                        );
+                        None
+                    }
+                },
+            )
+            .collect();
+        client_config.chain.invalid_block_roots = invalid_block_roots;
+    }
```
dapplion, line 939: i.e. 
```
else if chain == holesky {
  invalid_block_roots = [2db899881ed8546476d0b92c6aa9110bea9a4cd0dbeb5519eb0ea69575f1f359]
}

beacon_node/beacon_chain/src/beacon_chain.rs
michaelsproul: yeah I found myself wanting this the other day, I'll amend

beacon_node/src/config.rs
michaelsproul: Addressed in https://github.com/sigp/lighthouse/pull/7042/commits/fe41e182199480ebba85e21aebfd31fa21dcc3d7

beacon_node/beacon_chain/src/beacon_chain.rs
michaelsproul: Addressed in https://github.com/sigp/lighthouse/pull/7042/commits/fe41e182199480ebba85e21aebfd31fa21dcc3d7

beacon_node/beacon_chain/src/beacon_chain.rs
michaelsproul: I think the check in `sync_methods` was a bit dodgy because it returned without propagating the error. Based on my rudimentary understanding, sync doesn't like it when things like this happen (it doesn't know the result of the lookup, and could get stuck).

Instead I moved it into `SignatureVerifiedBlock::new`, which is called from the RPC path.

michaelsproul: c405dfbf396c09f18fd388c470e413bd53223236

michaelsproul (approved):

michaelsproul:
@mergifyio requeue

PR #7044, opened 2025-02-26
Use `sync_tolerance_epochs` flag to control the proposer prep routines

jimmygchen:
## Issue Addressed

Replace the `2 + 2 == 5` hacks from `holesky-rescue` and use the existing `sync_tolerance_epochs` flag to control the proposer prep routines.


michaelsproul:
Should merge this after (manually) merging this PR:

- https://github.com/sigp/lighthouse/pull/7045

michaelsproul:
Merged the original sync tolerance PR to unstable, so this now just contains the new changes.

beacon_node/beacon_chain/src/chain_config.rs
```diff
@@ -16,6 +16,9 @@ pub const DEFAULT_PREPARE_PAYLOAD_LOOKAHEAD_FACTOR: u32 = 3;
 /// Fraction of a slot lookahead for fork choice in the state advance timer (500ms on mainnet).
 pub const FORK_CHOICE_LOOKAHEAD_FACTOR: u32 = 24;
 
+/// Default sync tolerance epochs.
+pub const DEFAULT_SYNC_TOLERANCE_EPOCHS: u64 = 16;
```
jimmygchen: I haven't really given enough thoughts to this, so this PR changes the following
- `PREPARE_PROPOSER_HISTORIC_EPOCHS` from 4 to `sync_tolerance_epochs`
- `get_pre_payload_attributes` epochs from 2 to `sync_tolerance_epochs`
- `DEFAULT_SYNC_TOLERANCE_EPOCHS` for http block production from 8 to `sync_tolerance_epochs`

`sync_tolerance_epochs` is now defaulted to 16.

We discussed about `DEFAULT_SYNC_TOLERANCE_EPOCHS` a while ago but I forgot why it was 8.

@AgeManning  had a comment here https://github.com/sigp/lighthouse/issues/6880#issuecomment-2623531026

michaelsproul: I think lets change this back to 8.

Now that we have a config flag for it, we have the option of getting people to set it in adverse chain conditions (sparse chain with legitimately very few blocks).

I'm worried that setting it to 16 might have contributed to the plethora of shitty sidechains we saw on Holesky. Whatever value we set here is essentially a bound on "we will build a shitty side chain during sync/stuckness of up to this length". I'd even be tempted to set the default lower, maybe 2 (as Age suggested originally) or 4.

michaelsproul: As agreed on the call today, I've lowered the default to 2 in this commit: c7105d1edcde3e33ff98dd56dcfed82616a06e1d.

In case of a sparse chain and liveness issues, we can advise validators to bump up the sync tolerance temporarily.

michaelsproul (approved):

PR #7046, opened 2025-02-26
Add `--disable-attesting` flag to validator client

michaelsproul:
## Proposed Changes

Cleaned up and isolated version of the `--disable-attesting` flag for the VC, from the `holesky-rescue` branch:

- https://github.com/sigp/lighthouse/pull/7041

I figured we don't need the `--disable-attesting` flag on the BN for now, and it was a much more invasive impl.


jimmygchen (approved):
Nice, moving it to the VC makes sense - also avoids clogging the beacon processor on P0 requests.

PR #7047, opened 2025-02-26
Add `--long-timeouts-multiplier` CLI flag

macladson:
## Proposed Changes

Adds the `--long-timeouts-multiplier` flag.
Allows granular control for VC timeouts which has proved useful in Holesky.

## Additional Info

Defaults to `1` and for clarity requires use of the `--use-long-timeouts` flag in order to function. 
A multiplier of `0` is technically possible which I believe will disable timeouts.

### DIFF
--- a/book/src/help_vc.md
+++ b/book/src/help_vc.md
@@ -250,7 +250,7 @@ Flags:
       --long-timeouts-multiplier <LONG_TIMEOUTS_MULTIPLIER>
           If present, the validator client will use a multiplier for the timeout
           when making requests to the beacon node. This only takes effect when
-          the `--use-long_timeouts` flag is present. The timeouts will be the
+          the `--use-long-timeouts` flag is present. The timeouts will be the
           slot duration multiplied by this value. This flag is generally not
           recommended, longer timeouts can cause missed duties when fallbacks
           are used. [default: 1]
--- a/book/src/help_vc.md
+++ b/book/src/help_vc.md
@@ -247,6 +247,13 @@ Flags:
           contain sensitive information about your validator and so this flag
           should be used with caution. For Windows users, the log file
           permissions will be inherited from the parent folder.
+      --long-timeouts-multiplier <LONG_TIMEOUTS_MULTIPLIER>
+          If present, the validator client will use a multiplier for the timeout
+          when making requests to the beacon node. This only takes effect when
+          the `--use-long_timeouts` flag is present. The timeouts will be the
+          slot duration multiplied by this value. This flag is generally not
+          recommended, longer timeouts can cause missed duties when fallbacks
+          are used. [default: 1]
       --metrics
           Enable the Prometheus metrics HTTP server. Disabled by default.
       --prefer-builder-proposals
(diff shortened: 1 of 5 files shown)


validator_client/src/lib.rs
```diff
@@ -325,7 +325,7 @@ impl<E: EthSpec> ProductionValidatorClient<E> {
                     get_validator_block: slot_duration / HTTP_GET_VALIDATOR_BLOCK_TIMEOUT_QUOTIENT,
                 }
             } else {
-                Timeouts::set_all(slot_duration)
+                Timeouts::set_all(slot_duration.saturating_mul(config.long_timeouts_multiplier))
```
jimmygchen, line 328: This would still use the default timeouts if fallback node exists, is this the desired behaviour for this flag?

macladson: This `else` branch is only used if either there is no fallback or if `--use-long-timeouts` is set.
If no fallback node exists and `--use-long-timeouts` is not set then it will still use the default multiplier of `1` which is the same as just being `slot_duration`. Since the value is not configurable unless `--use-long-timeouts` is set, I figured this was fine

jimmygchen: ah right, sorry I got confused. So it would never use the quick timeouts as long as `config.use_long_timeouts` is `true`, makes sense.

book/src/help_vc.md
```diff
@@ -247,6 +247,13 @@ Flags:
           contain sensitive information about your validator and so this flag
           should be used with caution. For Windows users, the log file
           permissions will be inherited from the parent folder.
+      --long-timeouts-multiplier <LONG_TIMEOUTS_MULTIPLIER>
+          If present, the validator client will use a multiplier for the timeout
+          when making requests to the beacon node. This only takes effect when
+          the `--use-long_timeouts` flag is present. The timeouts will be the
```
dapplion:
```suggestion
          the `--use-long-timeouts` flag is present. The timeouts will be the
```

validator_client/src/cli.rs
```diff
@@ -107,6 +107,20 @@ pub struct ValidatorClient {
     )]
     pub use_long_timeouts: bool,
 
+    #[clap(
+        long,
+        requires = "use_long_timeouts",
+        default_value_t = 1,
+        help = "If present, the validator client will use a multiplier for the timeout \
+                when making requests to the beacon node. This only takes effect when \
+                the `--use-long_timeouts` flag is present. The timeouts will be the slot \
```
michaelsproul:
```suggestion
                the `--use-long-timeouts` flag is present. The timeouts will be the slot \
```

validator_client/src/cli.rs
```diff
@@ -107,6 +107,20 @@ pub struct ValidatorClient {
...
+        default_value_t = 1,
+        help = "If present, the validator client will use a multiplier for the timeout \
+                when making requests to the beacon node. This only takes effect when \
+                the `--use-long-timeouts` flag is present. The timeouts will be the slot \
+                duration multiplied by this value. This flag is generally not recommended, \
+                longer timeouts can cause missed duties when fallbacks are used.",
+        display_order = 0,
+        help_heading = FLAG_HEADER,
+    )]
+    pub long_timeouts_multiplier: u32,
```
michaelsproul, line 122: Needs a CLI test in `lighthouse/test/validator_client.rs`

michaelsproul (changes requested):

michaelsproul (approved):
LGTM

PR #7048, opened 2025-02-26
Remove `watch`

macladson:
## Proposed Changes

Removes the WIP chain indexer known as beacon watch.

## Additional Info

Development on beacon watch has halted at this point, and so it has now become a maintenance, testing and compilation burden for all devs. This could be spun out into its own repo at a later date if there is renewed interest, but for now, I believe that other teams in the ecosystem are better placed to provide this kind of tooling. Either way, it does not belong in the Lighthouse repo and should be removed.


book/src/setup.md
```diff
@@ -17,9 +17,6 @@ The additional requirements for developers are:
   some dependencies. See [`Installation Guide`](./installation.md) for more info.
 - [`java 17 runtime`](https://openjdk.java.net/projects/jdk/). 17 is the minimum,
   used by web3signer_tests.
-- [`libpq-dev`](https://www.postgresql.org/docs/devel/libpq.html). Also known as
-  `libpq-devel` on some systems.
-- [`docker`](https://www.docker.com/). Some tests need docker installed and **running**.
```
macladson, line 22: I believe removing `watch` also removes these compilation dependencies, would someone be able to confirm this?

michaelsproul: Yeah, agree. We don't use Docker for any of the other tests.

michaelsproul (approved):
So long and thanks for all the fish `watch` :wave:

PR #7050, opened 2025-02-26
Make ExecutionBlock::total_difficulty Optional

ryanschneider:
## Issue Addressed

This change makes the `total_difficulty` field in `ExecutionBlock` an `Option<Uint256>` since newer clients are no longer including the `totalDifficulty` field.

I think this will fix https://github.com/sigp/lighthouse/issues/6937 but I was actually more focused on the builder registration case described below.

In our [builder-playground](https://github.com/flashbots/builder-playground) we setup a local devnet using lighthouse, reth, and mev-boost-relay.  After upgrading to reth 1.2.0 and lighthouse v7.0.0.beta.0 for Pectra, we noticed that the validator registration process was _sometimes_ failing with:

```
Feb 25 23:35:25.038 ERRO Unable to publish proposer preparation to all beacon nodes, error: Some endpoints failed, num_failed: 1 http://localhost:3500/ => RequestFailed(ServerMessage(ErrorMessage { code: 400, message: "BAD_REQUEST: error updating proposer preparations: ForkchoiceUpdate(EngineError(Api { error: Json(Error(\"missing field `totalDifficulty`\", line: 0, column: 0)) }))", stacktraces: [] })), service: preparation
Feb 25 23:35:25.099 WARN Unable to publish validator registrations to the builder network, error: Some endpoints failed, num_failed: 1 http://localhost:3500/ => RequestFailed(ServerMessage(ErrorMessage { code: 400, message: "BAD_REQUEST: error updating proposer preparations: ForkchoiceUpdate(EngineError(Api { error: Json(Error(\"missing field `totalDifficulty`\", line: 0, column: 0)) }))", stacktraces: [] })), service: preparation
```

What was even more confusing, was that it was sometimes working, which actually led to a wild goose chase thinking it was a networking issue.  However, when tracing through the LH code, I came across this comment:

https://github.com/sigp/lighthouse/blob/70194dfc6a3f4d10c9059610f889ff5a4e863a6a/beacon_node/beacon_chain/src/beacon_chain.rs#L6048-L6049

This explained why it sometimes worked, in our playground we run lighthouse with `--prepare-payload-lookahead 8000` thus there was always a 4-second window where the call wasn't made.

But, if the call was made, then this code would 100% fail with updated reth:

https://github.com/sigp/lighthouse/blob/unstable/beacon_node/execution_layer/src/lib.rs#L1688-L1692

Which would then mapped to a `Error::ForkchoiceUpdate` in `update_execution_engine_forkchoice`.

## Proposed Changes

Anyways, the fix was to make `total_difficulty` Optional, and then to update any code paths where it was used.  In doing so, I assume that if the EL doesn't include total difficulty then the chain is already post-merge.

## Additional Info

I left `execution_layer::test_utils::PoWBlock::total_difficulty` non-optional since its used for testing the PoW transition, and total difficulty was an expected field at that time.

### DIFF
--- a/beacon_node/beacon_chain/src/bellatrix_readiness.rs
+++ b/beacon_node/beacon_chain/src/bellatrix_readiness.rs
@@ -171,7 +171,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
                 return BellatrixReadiness::NotSynced;
             }
             let params = MergeConfig::from_chainspec(&self.spec);
-            let current_difficulty = el.get_current_difficulty().await.ok();
+            let current_difficulty = el.get_current_difficulty().await.ok().flatten();
             BellatrixReadiness::Ready {
                 config: params,
                 current_difficulty,
--- a/beacon_node/execution_layer/src/engine_api.rs
+++ b/beacon_node/execution_layer/src/engine_api.rs
@@ -150,7 +150,7 @@ pub struct ExecutionBlock {
 impl ExecutionBlock {
     pub fn terminal_total_difficulty_reached(&self, terminal_total_difficulty: Uint256) -> bool {
         self.total_difficulty
-            .map_or(true, |td| td >= terminal_total_difficulty)
+            .is_none_or(|td| td >= terminal_total_difficulty)
     }
 }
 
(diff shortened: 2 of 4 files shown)


michaelsproul (approved):
Looks good.

Eventually I'd like to delete all this merge-related/difficulty-related stuff too.

PR #7051, opened 2025-02-26
Implement pessimistic sync

michaelsproul:
## Proposed Changes

Add a new flag which allows optimistic sync to be turned off.


michaelsproul:
Closing (again), as ELs require optimistic sync to sync. I think we could add a mode that turns off optimistic sync after initial sync

PR #7052, opened 2025-02-26
Publish EL Info in Metrics

ethDreamer:
## Issue Addressed

Since we now know the EL version, we should publish this to our metrics periodically.


jimmygchen (approved):
Looks good 👍

Issue #7053, opened 2025-02-27
Lighthouse OOM mitigations


michaelsproul:
Short term plan:

1. Move banned block checks _higher_ in block verification to prevent repeat state lookups (before every instance of `load_parent` in `block_verification.rs`)
2. Encourage use of `--state-cache-size 4` to avoid bad state cache pruning logic that is keeping 128x 180MB epoch boundary states around (~24GB of states).
3. (DONE) Remove block root lookups from status processing. We are getting killed looking up old states to compute the block root. We need a more aggressive version of this PR: https://github.com/sigp/lighthouse/pull/5481.

Point (1) is intended to fix an OOM that happens to nodes that are in sync and forced to process junk.

Point (2) fixes OOMs during head sync due to lots of epoch boundary states being retain.

To investigate later:

1. Why are epoch boundary state diffs so large (180MB+), given that we should be basing them off each other while syncing sequential blocks? Answer: `balances` and `inactivity_scores`.
2. Is an earlier invalid block check sufficient to prevent OOM while synced? Are there are other states or valid side chains which are forcing us to load states and use too much memory?
3. Why is sync sending us so many copies of the invalid block? Is there parallelism that is causing the OOM near the head?

Future plans (long-term fixes):

1. Implement the PromiseCache concept used for attestation committees for beacon states. This is quite subtle to get right, a version was previously attempted but abandoned (https://github.com/sigp/lighthouse/pull/5313). Tracking issue: https://github.com/sigp/lighthouse/issues/5112
2. Implement size-based pruning for the state cache. This is possible with my WIP changes from: https://github.com/sigp/lighthouse/pull/6532. However, that code is quite immature and the pruning itself is expensive (1.5s-4s or more), so we cannot ship this quickly. There is also some subtlety around deciding which states to prune based on size (we could use a similar heuristic to the existing `cull` method on the 20% largest states).
3. Re-think pruning logic in `cull` so that it doesn't hang on to so many useless epoch boundary states.

michaelsproul:
Merged status processing fix to `holesky-rescue` branch:

- https://github.com/sigp/lighthouse/pull/7054

michaelsproul:
Just thought of another source of unbounded state lookups: `BlocksByRoot` and `BlocksByRange`.

It might be time to build a dedicated in-memory DAG of block roots which we can use instead of the state-based block iterators.

cc @dapplion

eserilev:
[...]

PR #5481, opened 2024-03-26
Optimise status processing

michaelsproul:
## Proposed Changes

Avoid expensive `block_root_at_slot` call in status processing. This proved to be very time consuming on networks without finality, like Goerli.

PR #7054, opened 2025-02-27
Optimise status processing on release-v.7.0.0-beta.2 (TESTING branch)

jimmygchen:
## Issue Addressed

Taken from the #5481 branch, but with latest changes from `holesky-rescue` merged for testing on holesky.

Could potentially address #7053, in combination with `--state-cache-size=4`


michaelsproul (approved):
I'd forgotten how good this change was, it seems it already completely avoids state lookups ;)

PR #7056, opened 2025-02-27
Delete expensive chain compatibility checks in check_peer_relevance

dapplion:
## Issue Addressed

Extends https://github.com/sigp/lighthouse/pull/5481 not to do **any** DB reads when processing status messages. These have proven to be very expensive during the Holesky incident

## Proposed Changes

I added extensive comments to justify the changes, please check those :)

A working solution that doesn't read anything from the chain is 92a39362eaedb42d4c3203f5bfeb4db17045e461. Then 33870c1621f59518eabf156e3fa88eac905b9cbc covers a bit more ground by using the existing cached head which should be free

### DIFF
--- a/beacon_node/network/src/network_beacon_processor/rpc_methods.rs
+++ b/beacon_node/network/src/network_beacon_processor/rpc_methods.rs
@@ -13,11 +13,12 @@ use lighthouse_network::{PeerId, PeerRequestId, ReportSource, Response, SyncInfo
 use methods::LightClientUpdatesByRangeRequest;
 use slog::{debug, error, warn};
 use slot_clock::SlotClock;
+use std::cmp::Ordering;
 use std::collections::{hash_map::Entry, HashMap};
 use std::sync::Arc;
 use tokio_stream::StreamExt;
 use types::blob_sidecar::BlobIdentifier;
-use types::{Epoch, EthSpec, Hash256, Slot};
+use types::{EthSpec, Hash256, Slot};
 
 impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
     /* Auxiliary functions */
@@ -73,7 +74,6 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
         remote: &StatusMessage,
     ) -> Result<Option<String>, BeaconChainError> {
         let local = self.chain.status_message();
-        let start_slot = |epoch: Epoch| epoch.start_slot(T::EthSpec::slots_per_epoch());
 
         let irrelevant_reason = if local.fork_digest != remote.fork_digest {
             // The node is on a different network/fork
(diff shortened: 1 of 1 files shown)


beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -93,41 +93,67 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
-                    .block_root_at_slot(remote_finalized_slot, WhenSlotSkipped::Prev)
-                    .map(|root_opt| root_opt != Some(remote.finalized_root))?
-                {
-                    Some("Different finalized chain".to_string())
-                } else {
-                    None
+            match remote.finalized_epoch.cmp(&local.finalized_epoch) {
+                // This peer has a finalized epoch less than ours. This peer is likely to not be
+                // useful to us, as they want to sync from us. As a nice gesture, we could do work
+                // from them and check if their finalized_root matches our imported chain.
```
pawanjay176, line 100:
```suggestion
                // for them and check if their finalized_root matches our imported chain.
```

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -93,41 +93,67 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                // However, this check is unncessary, as if we are incompatible the peer will be
+                // eventually disconnected anyway. Consider the two cases:
+                //
+                // - Peer is our chain: we should forward peer to sync to allow them to sync from us
```
pawanjay176, line 104:
```suggestion
                // - Peer is on our chain: we should forward peer to sync to allow them to sync from us
```

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -93,41 +93,67 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                        // Definitely on same finalized chain, forward peer to sync
+                        None
+                    } else {
+                        Some(format!(
+                            "Different finalized chain, expected {} got {}",
+                            local.finalized_root, remote.finalized_root
+                        ))
+                    }
+                }
+                // Peer claims to know about a finalized checkpoint ahead of hours. Send peer to
```
pawanjay176, line 153:
```suggestion
                // Peer claims to know about a finalized checkpoint ahead of ours. Send peer to
```

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -93,41 +93,67 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                                None
+                            } else {
+                                Some(format!(
+                                    "Different finalized chain, expected {} got {}",
+                                    block_root_at_remote_finalized_epoch, remote.finalized_root
+                                ))
+                            }
+                        }
+                        // Out of bounds root check, allow the peer to connected to us and expect
+                        // they to disconnect or be disconnected if we are in incompatible chains
```
pawanjay176, line 138:
```suggestion
                        // them to disconnect or be disconnected if we are in incompatible chains
```

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -93,41 +93,67 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                                // Definitely on same finalized chain, forward peer to sync
+                                None
+                            } else {
+                                Some(format!(
+                                    "Different finalized chain, expected {} got {}",
+                                    block_root_at_remote_finalized_epoch, remote.finalized_root
+                                ))
+                            }
+                        }
+                        // Out of bounds root check, allow the peer to connected to us and expect
```
pawanjay176, line 137:
```suggestion
                        // Out of bounds root check, allow the peer to connect to us and expect
```

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -93,41 +93,67 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
-                // Peer's finalized checkpoint is older than anything in our DB. We are unlikely
-                // to be able to help them sync.
-                Some("Old finality out of range".to_string())
-            } else if remote_finalized_slot < self.chain.store.get_split_slot() {
-                // Peer's finalized slot is in range for a quick block root check in our freezer DB.
-                // If that block root check fails, reject them as they're on a different finalized
-                // chain.
-                if self
-                    .chain
-                    .block_root_at_slot(remote_finalized_slot, WhenSlotSkipped::Prev)
```
pawanjay176, line 119: If this check is indeed cheap, I would prefer to keep it as it would cause make it harder for peers to sync in a chaotic network if we remove this.
I do not have a sense of how cheap this is though.

michaelsproul: It's very cheap, reading a single block root from the freezer is fast (<3ms)

michaelsproul:
Closing as not required, but feel free to reopon

PR #7057, opened 2025-02-27
switch to upstream gossipsub

jxs:
## Issue Addressed

We forked `gossipsub` into the lighthouse repo sometime ago so that we could iterate quicker on implementing back pressure and IDONTWANT. 
Meanwhile we have pushed all our changes upstream and we are now the main maintainers of `rust-libp2p` this allows us to use upstream `gossipsub` again. 
Nonetheless we still use our forked repo to give us freedom to experiment with features before submitting them upstream


ackintosh:
Thanks for this PR! I'm checking if there are any improvements from our fork that might have been missed during upstreaming.

beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -817,17 +817,8 @@ impl<E: EthSpec> Network<E> {
...
-            Err(_) => {
-                warn!(self.log, "Failed to unsubscribe from topic"; "topic" => %libp2p_topic);
-                false
-            }
-            Ok(v) => {
-                // Inform the network
-                debug!(self.log, "Unsubscribed to topic"; "topic" => %topic);
-                v
-            }
-        }
```
AgeManning, line 830: Why did we remove these logs?

beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -912,13 +903,11 @@ impl<E: EthSpec> Network<E> {
...
         }
 
-        if let Err(e) = self.gossipsub_mut().report_message_validation_result(
+        self.gossipsub_mut().report_message_validation_result(
             &message_id,
             propagation_source,
             validation_result,
-        ) {
-            warn!(self.log, "Failed to report message validation"; "message_id" => %message_id, "peer_id" => %propagation_source, "error" => ?e);
-        }
```
AgeManning, line 921: And these ones?

beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -1256,13 +1245,11 @@ impl<E: EthSpec> Network<E> {
...
                         debug!(self.log, "Could not decode gossipsub message"; "topic" => ?gs_msg.topic,"error" => e);
                         //reject the message
-                        if let Err(e) = self.gossipsub_mut().report_message_validation_result(
+                        self.gossipsub_mut().report_message_validation_result(
                             &id,
                             &propagation_source,
                             MessageAcceptance::Reject,
-                        ) {
-                            warn!(self.log, "Failed to report message validation"; "message_id" => %id, "peer_id" => %propagation_source, "error" => ?e);
-                        }
```
AgeManning, line 1265: Here also

beacon_node/lighthouse_network/src/service/mod.rs
jxs: cause we no longer err on `unsubscribe()` it always succeeds at the `Behavior` level

beacon_node/lighthouse_network/src/service/mod.rs
jxs: same reason, we no longer fail at this level, message validation report always succeeds

AgeManning (approved):
Ok, cool. LGTM

ackintosh:
I noticed that #6873 has not been upstreamed, but I'm not sure if we still need these metrics. cc: @pawanjay176

pawanjay176:
Thanks @ackintosh . I think one of the metrics in the PR is buggy, been meaning to fix it but haven't gotten around to it so its good that it doesn't makes it upstream.

pawanjay176 (approved):

ackintosh:
@jxs @pawanjay176 
These metrics have also not been upstreamed. They looks useful to me, what do you think? I can submit a PR to upstream them if they're good.

https://github.com/sigp/lighthouse/blob/e7ea69647a4cb686d0c7a80f24b3f34d329a7e01/beacon_node/lighthouse_network/gossipsub/src/metrics.rs#L188-L195

jxs:
@pawanjay176 @ackintosh wdyt of adding them temporarily on https://github.com/sigp/rust-libp2p/pull/570 until we confirm IDONTWANTs are saving us bandwidth?

ackintosh:
That makes sense to me.

ackintosh (approved):

jimmygchen (approved):

PR #7058, opened 2025-02-28
Load block roots from fork choice where possible when serving `BlocksByRange` requests

jimmygchen:
## Issue Addressed

Partially #7053.

## Proposed Changes

Load block roots from fork choice where possible to avoid loading state from disk when serving block by range requests.

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -7282,6 +7282,39 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
 
         Ok(None)
     }
+
+    /// Retrieves block roots within some slot range from fork choice.
+    /// Returns `None` if the provided slot is not found.
+    pub fn block_roots_from_fork_choice(
+        &self,
+        start_slot: u64,
+        count: u64,
+    ) -> Option<Vec<Hash256>> {
+        let head_block_root = self.canonical_head.cached_head().head_block_root();
+        let fork_choice_read_lock = self.canonical_head.fork_choice_read_lock();
+        let block_roots_iter = fork_choice_read_lock
+            .proto_array()
+            .iter_block_roots(&head_block_root);
+        let end_slot = start_slot.saturating_add(count);
+        let mut roots = vec![];
+
+        for (root, slot) in block_roots_iter {
+            if slot < end_slot && slot >= start_slot {
+                roots.push(root);
+            }
+            if slot < start_slot {
+                break;
+            }
+        }
+
+        if roots.is_empty() {
+            None
+        } else {
+            // return in ascending slot order
+            roots.reverse();
+            Some(roots)
+        }
+    }
 }
 
 impl<T: BeaconChainTypes> Drop for BeaconChain<T> {
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -7283,13 +7283,8 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         Ok(None)
     }
 
-    /// Retrieves block roots within some slot range from fork choice.
-    /// Returns `None` if the provided slot is not found.
-    pub fn block_roots_from_fork_choice(
-        &self,
-        start_slot: u64,
-        count: u64,
-    ) -> Option<Vec<Hash256>> {
+    /// Retrieves block roots (in ascending slot order) within some slot range from fork choice.
+    pub fn block_roots_from_fork_choice(&self, start_slot: u64, count: u64) -> Vec<Hash256> {
         let head_block_root = self.canonical_head.cached_head().head_block_root();
         let fork_choice_read_lock = self.canonical_head.fork_choice_read_lock();
         let block_roots_iter = fork_choice_read_lock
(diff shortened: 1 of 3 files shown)


beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -7282,6 +7282,39 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+            if slot < start_slot {
+                break;
+            }
+        }
+
+        if roots.is_empty() {
+            None
+        } else {
+            // return in ascending slot order
+            roots.reverse();
```
michaelsproul: could `drop(fork_choice_read_lock)` before doing the reverse operation, in order to free up the lock ever so slightly sooner

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -681,64 +681,24 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
             "start_slot" => req.start_slot(),
         );
 
-        let forwards_block_root_iter = match self
+        let block_roots_timer = std::time::Instant::now();
+        let (block_roots, block_roots_source) = if let Some(block_roots) = self
```
michaelsproul: Rather than attempting to access fork choice and seeing if it works, I think we should check if the start slot is newer than finalization (`start_slot >= finalized_slot`), and use fork choice in that case. Fork choice has to have every chain newer than finalization.

This will be more efficient, as it fixes the issue with skipped slots. If a peer requests a range of unfinalized slots that is empty, we can just return nothing. Currently, we will try fork choice, get `None`, and then try `get_block_roots_from_store` and get the block roots from the skipped slots. This is not actually helpful, as later in the function we'll drop the blocks that are out of range anyway:

https://github.com/sigp/lighthouse/blob/2cb71e2211e7a778407c17ab7a6704a93966bca2/beacon_node/network/src/network_beacon_processor/rpc_methods.rs#L746-L758

In other words, the new fork choice impl allows us to avoid looking up blocks from skipped slots and then discarding them.

Something that this code change is not currently handling is requests that span across the finalized slot. If `start_slot < finalized_slot` and `end_slot > finalized_slot`, then the current impl will return partial data (all the blocks newer than finalization). Instead of this, short-term I think it would be better to fallback on the old disk-based iterator, which is accomplished by the same code change I suggested above: only use fork choice if the start slot is newer than finalization.

Longer term, we should use some kind of hybrid iterator so we can easily span the finalized slot without any mucking around. For now I think we don't need to worry about this though, because these requests should be infrequent.

michaelsproul (changes requested):

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -7282,6 +7282,39 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+
+    /// Retrieves block roots within some slot range from fork choice.
+    /// Returns `None` if the provided slot is not found.
+    pub fn block_roots_from_fork_choice(
+        &self,
+        start_slot: u64,
+        count: u64,
+    ) -> Option<Vec<Hash256>> {
+        let head_block_root = self.canonical_head.cached_head().head_block_root();
+        let fork_choice_read_lock = self.canonical_head.fork_choice_read_lock();
```
michaelsproul, line 7289: Technically we should probably avoid locking fork choice from an async task, but atm I think we have bigger fish to fry. It's nowhere near as bad as doing 10s of state loads 😱

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
jimmygchen: Thanks for the explanation. I've applied the change to check for finalization in  https://github.com/sigp/lighthouse/pull/7058/commits/bd093d95337ea7e1b0b010bc01994775f4be03c8

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Done in https://github.com/sigp/lighthouse/pull/7058/commits/bd093d95337ea7e1b0b010bc01994775f4be03c8

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -843,17 +788,121 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+        let block_roots_timer = std::time::Instant::now();
+        let finalized_slot = self
+            .chain
+            .canonical_head
+            .cached_head()
+            .finalized_checkpoint()
+            .epoch
+            .start_slot(T::EthSpec::slots_per_epoch());
+
+        let (block_roots, block_roots_source) = if req_start_slot >= finalized_slot.as_u64() {
```
dapplion, line 817: Is the `finalized_slot` in the head state the variable that strictly defines fork-choice availability? Looks correct on a first pass, but just double checking. A more complete impl would consider that fork-choice is pruned lazily so we may have block roots available in the fork-choice with slots less than `finalized_slot`

michaelsproul: Close enough for now IMO, we are at least inaccurate in a safe way

Something to consider for the clean impl though

michaelsproul (approved):
LFG

PR #7059, opened 2025-02-28
Manual finalization endpoint

eserilev:
Introduce a new lighthouse endpoint that allows a user to force a manual finalization. This migrates data to the freezer db and prunes sidechains which may help reduce disk space issues on holesky at the moment.


beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -1707,6 +1707,16 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         }
     }
 
+    pub fn manually_finalize_state(&self, state_root: Hash256, checkpoint: Checkpoint) {
+        let notif = ManualFinalizationNotification {
```
dapplion, line 1734: Here we should sanity check that this checkpoint is an actual checkpoint to not mess up the DB too much

testing/.DS_Store
michaelsproul, line 1: can delete

eserilev:
one quick and dirty thing we might be able to try 
```rust
// add some for loop here so we only fetch at most one state per roots iterator
for {
    let state_roots = RootsIterator::new(&store, finalized_state)
        .take_while(|result| match result {
            Ok((_, _, slot)) => *slot >= current_split_slot,
            Err(_) => true,
        })
    // do the hot db -> cold db migration + pruning
 }
  ```

**Comment by michaelsproul** [comment]
I think the `RootsIterator` is _already_ inside a loop, so there should only be one at a time:

https://github.com/sigp/lighthouse/blob/404e48d740778181ba18ce2803562fc9f8c4f19b/beacon_node/beacon_chain/src/migrate.rs#L602-L630

Similarly the `RootsIterator` itself jumps back over states one at a time too.

The only thing I can think is that we're putting these states in the cache, and that's enough to cause an OOM with `--state-cache-size 32`

File: beacon_node/beacon_chain/src/beacon_chain.rs (hunk 1 above)
Comment (michaelsproul) [reply beacon_node/beacon_chain/src/beacon_chain.rs:1734]: 100%

I think we should load the `HotStateSummary` for the `state_root` and check the provided `slot` and `block_root` against it (should match `slot` and `latest_block_root` from the summary).

File: beacon_node/beacon_chain/src/migrate.rs [hunk 2]
```diff
@@ -190,6 +198,10 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> BackgroundMigrator<E, Ho
         Ok(())
     }
 
+    pub fn process_manual_finalization(&self, notif: ManualFinalizationNotification) {
+        let _ = self.send_background_notification(Notification::ManualFinalization(notif));
```
jimmygchen: I don't think we ever run store migrator in the foreground other than testing, but probably should support it given we do it  for all 3 other notifications, and if we want to test this with `BeaconChainHarness`, it won't be immediately obvious why this doesn't get run.

michaelsproul: Good point. We should do this for the clean version of this PR

michaelsproul (approved):

PR #7066, opened 2025-03-03
Split block root lookups between fork choice and store on BBR response

realbigsean:


beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -815,16 +815,36 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
 
         let (block_roots, block_roots_source) = if req_start_slot >= finalized_slot.as_u64() {
+            // If the entire requested range is after finalization, use fork_choice
             (
                 self.chain
                     .block_roots_from_fork_choice(req_start_slot, req_count),
                 "fork_choice",
             )
-        } else {
+        } else if req_start_slot + req_count <= finalized_slot.as_u64() + 1 {
```
michaelsproul:
```suggestion
        } else if req_start_slot + req_count <= finalized_slot.as_u64() {
```

Thinking aloud here: the end slot (`req_start_slot + req_count`) is _exclusive_, so the last slot we will access will be `req_start_slot + req_count - 1`. The iterator will avoid loading a state if the `end_slot` (inclusive) is less than the split slot, here:

https://github.com/sigp/lighthouse/blob/bf955c7543dac8911a6f6c334b5b3ca4ef728d9c/beacon_node/store/src/forwards_iter.rs#L264-L272


So we'll avoid loading a hot state (good) in the case where:

```rust
req_start_slot + req_count - 1 < split.slot
```

i.e.

```rust
req_start_slot + req_count <= split.slot
```

so we don't need that +1 and can use

```
req_start_slot + req_count <= finalized_slot.as_u64()
```

Said another way: we allow the request's end state to be at most the finalized slot, and guarantee that all data prior to that lies in the freezer. This is safer and more conservative than with the +1.

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -815,16 +815,36 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+            // If the entire requested range is before finalization, use store
             (
                 self.get_block_roots_from_store(req_start_slot, req_count)?,
                 "store",
             )
+        } else {
+            // Split the request at the finalization boundary
+            let count_from_store = finalized_slot.as_u64() - req_start_slot + 1;
```
michaelsproul:
```suggestion
            let count_from_store = finalized_slot.as_u64() - req_start_slot;
```

Just want to fetch the slots `< finalized_slot` from store, so in the case where `req_state_slot == finalized_slot - 1`, we want this count to be 1.

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -815,16 +815,36 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+            let count_from_fork_choice = req_count - count_from_store;
+            let start_slot_fork_choice = finalized_slot.as_u64() + 1;
```
michaelsproul:
```suggestion
            let start_slot_fork_choice = finalized_slot.as_u64();
```

Want to fetch all slots `>= finalized_slot` from fork choice.

michaelsproul (approved):

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
realbigsean: Could only reason by example here:

> `req_start_slot + req_count <= finalized_slot.as_u64()`

A request for the first epoch would be `req_start_slot = 1` with `req_count = 32`. If the first epoch is finalized, the finalized slot is 32. Right? This request can be served just from the cold db. This is why I added the extra + 1

michaelsproul: I think we don't actually write the block root for the finalized slot itself to the cold DB though. The writes happen here:

https://github.com/sigp/lighthouse/blob/e12719c40483f17bbe43ab3fa911b5c5ba1bd12e/beacon_node/store/src/hot_cold_store.rs#L3164-L3169

The iterator is a `RootsIterator` created from the new split slot, which notably does not include the slot of the state it starts from (i.e. the split slot). 

https://github.com/sigp/lighthouse/blob/e12719c40483f17bbe43ab3fa911b5c5ba1bd12e/beacon_node/store/src/hot_cold_store.rs#L3146-L3152

The DB invariants are (usually) of the form:

```rust
if slot < split.slot then { freezer } else { hot }
```

PR #7067, opened 2025-03-03
Prevent writing to state cache when migrating the database

eserilev:
Added a new flag `update_cache` to `get_state`. When set to false, prevent writing fetched state to the cache


beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -821,7 +821,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             .get_blinded_block(&block_root)?
             .ok_or(Error::MissingBeaconBlock(block_root))?;
         let state = self
-            .get_state(&block.state_root(), Some(block.slot()))?
+            .get_state(&block.state_root(), Some(block.slot()), true)?
```
michaelsproul, line 824: Only used in tests so doesn't matter

beacon_node/beacon_chain/src/builder.rs
```diff
@@ -299,7 +299,11 @@ where
             .map_err(|e| descriptive_db_error("genesis block", &e))?
             .ok_or("Genesis block not found in store")?;
         let genesis_state = store
-            .get_state(&genesis_block.state_root(), Some(genesis_block.slot()))
+            .get_state(
+                &genesis_block.state_root(),
+                Some(genesis_block.slot()),
+                true,
```
michaelsproul, line 305: This is ok as we almost always have the genesis state in the freezer

michaelsproul (approved):

michaelsproul:
I'll PR this to `release-v7.0.0` with a `Co-author` tag for you @eserilev

PR #7068, opened 2025-03-03
Add http endpoint to add trusted peer

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Add a http endpoint `lighthouse/add_peer` that adds a trusted peer to the peerdb and dials it every heartbeat in case it gracefully disconnects.


michaelsproul (approved):
Used this local with `--disable-discovery` and it started syncing  effortlessly! Looks good!

PR #7069, opened 2025-03-03
state cache with fork awareness

realbigsean:
Trying to make the state cache aware of forks, pruning based on least recently accessed chain of states, and only maintaining a max of 4 states per chain.

Right now, tracking block ancestry in the cache. I tried initially by using the block roots in each state, but that spun out of control


realbigsean:
went with a different solution

PR #7072, opened 2025-03-04
Manual compaction endpoint

eserilev:
For some reason manual finalization isn't triggering compaction. I'll try investigating, but in the meantime this PR introduces an endpoint that allows users to manually trigger background compaction.


michaelsproul (approved):
Looks good!

jimmygchen (approved):
LGTM!

Issue #5745, opened 2024-05-09
Drop some boilerplate in lighthouse_network


dapplion:
## Description

There are some comments that do not provide any new information and are just noise

https://github.com/sigp/lighthouse/blob/3b42c8c5930648c5bcd8ee04cdadb75ad73a045f/beacon_node/lighthouse_network/src/rpc/methods.rs#L434-L441

https://github.com/sigp/lighthouse/blob/3b42c8c5930648c5bcd8ee04cdadb75ad73a045f/beacon_node/lighthouse_network/src/rpc/protocol.rs#L235-L239

https://github.com/sigp/lighthouse/blob/3b42c8c5930648c5bcd8ee04cdadb75ad73a045f/beacon_node/lighthouse_network/src/rpc/rate_limiter.rs#L84-L89

https://github.com/sigp/lighthouse/blob/3b42c8c5930648c5bcd8ee04cdadb75ad73a045f/beacon_node/lighthouse_network/src/service/api_types.rs#L37-L41

Matches like this could group conceptually protocol and use `A | B | C => return`

https://github.com/sigp/lighthouse/blob/3b42c8c5930648c5bcd8ee04cdadb75ad73a045f/beacon_node/lighthouse_network/src/peer_manager/mod.rs#L601-L614

guha-rahul:
Can i take it?

PR #7076, opened 2025-03-04
chore: removed unnecessary noise comments and refactored some code

startup-dreamer:
Closes: #5745 

## Proposed Changes

Removed the suggested comments and deduplicated some of the matches.


startup-dreamer:
@jxs let me know if I missed something.

michaelsproul:
This PR is trivial and of little value + suspected airdrop farm.

PR #7084, opened 2025-03-06
Reject attestations to blocks prior to the split

### LINKED ISSUE #7083: Reject attestations to blocks prior to the split
With manual finalization, we added checks to reject blocks that conflict with the `split` point, which are effectively finalized.

We didn't add any checks to reject similar attestations. This is leading to a lot of noisy cache miss logs and "failed promises" in the committee cache on Holesky.

> Mar 06 03:47:03.153 DEBG Committee cache miss                    head_block_root: 0xe156a286e3a24c81a5afb2c4a47cbd1a75be4ea25f5670c08dc5837a053244d1, shuffling_id: Epoch(118047), service: beacon
 
> Mar 06 03:47:14.357 WARN State cache missed backtrace: 0: store::hot_cold_store::HotColdDB<E,Hot,Cold>::get_advanced_hot_state 1: network::network_beacon_processor::gossip_methods::<impl network::network_beacon_processor::NetworkBeaconProcessor<T>>::process_gossip_attestation_to_convert 2: core::ops::function::FnOnce::call_once{{vtable.shim}} 3: tokio::runtime::task::raw::poll 4: std::sys::backtrace::__rust_begin_short_backtrace 5: core::ops::function::FnOnce::call_once{{vtable.shim}} 6: std::sys::pal::unix::thread::Thread::new::thread_start 7: <unknown> 8: <unknown>

> Mar 06 03:47:14.366 DEBG Invalid attestation from network type: "unaggregated", peer_id: 16Uiu2HAmLAWYx8PBL8CQRTQAGLLvRtdMK2B7b4exaPju2fkLt57h, block: 0xe156a286e3a24c81a5afb2c4a47cbd1a75be4ea25f5670c08dc5837a053244d1, reason: BeaconChainError(MissingBeaconState(0x08f94dfd6adc67d6cfee4c4a99ecb2f53042f9808c7697f785292a2a141ac706))

This doesn't seem to have much performance impact, as no states end up being actually loaded from disk, however it does obscure legitimate state cache and committee cache misses, which is not ideal.

eserilev:
Closes #7083

Reject and penalize peers for gossiping attestations for blocks prior to the current split point


dapplion (approved):

beacon_node/beacon_chain/src/attestation_verification.rs
```diff
@@ -266,6 +266,16 @@ pub enum Error {
...
+    /// The attestation conflicts with finalization, no need to propagate.
+    ///
+    /// ## Peer scoring
+    ///
+    /// It's unclear if this attestation is valid, but it conflicts with finality and shouldn't be
+    /// propogated.
+    NotFinalizedDescendant {
+        attestation_block_root: Hash256,
+        attestation_slot: Slot,
+    },
```
michaelsproul: We already have a very similar error in the form of `HeadBlockFinalized`. I think it would be cleaner to use that

beacon_node/beacon_chain/src/attestation_verification.rs
```diff
@@ -1361,6 +1374,30 @@ pub fn verify_committee_index<E: EthSpec>(attestation: AttestationRef<E>) -> Res
...
+) -> Result<(), Error> {
+    // If we have a split block newer than finalization then we also ban attestations which are not
+    // descended from that split block.
+    let fork_choice = chain.canonical_head.fork_choice_read_lock();
+    let split = chain.store.get_split_info();
+    let attestation_block_root = attestation.data().beacon_block_root;
+    let is_descendant_from_split_block =
+        split.slot == 0 || fork_choice.is_descendant(split.block_root, attestation_block_root);
+
+    if fork_choice.is_finalized_checkpoint_or_descendant(attestation_block_root)
```
michaelsproul: I think it might be better to fold these checks into `verify_head_block_is_known`, which is already trying to do some checks for descent from finalization. That function currently makes the (wrong) assumption that a block present in fork choice descends from finalization (this isn't true in general because fork choice is sometimes pruned on a delay).

So I think we can add these checks in the case where `block_opt` is `Some`

michaelsproul (changes requested):

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -2893,6 +2893,27 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                 );
                 self.propagate_validation_result(message_id, peer_id, MessageAcceptance::Ignore);
             }
+            AttnError::NotFinalizedDescendant {
```
michaelsproul: Can remove this case as this error no longer exists

beacon_node/beacon_chain/src/attestation_verification.rs
```diff
@@ -1361,6 +1364,29 @@ pub fn verify_committee_index<E: EthSpec>(attestation: AttestationRef<E>) -> Res
...
+    attestation: AttestationRef<T::EthSpec>,
+    chain: &BeaconChain<T>,
+) -> Result<(), Error> {
+    // If we have a split block newer than finalization then we also ban attestations which are not
+    // descended from that split block.
+    let fork_choice = chain.canonical_head.fork_choice_read_lock();
+    let split = chain.store.get_split_info();
+    let attestation_block_root = attestation.data().beacon_block_root;
+    let is_descendant_from_split_block =
+        split.slot == 0 || fork_choice.is_descendant(split.block_root, attestation_block_root);
```
michaelsproul: Needs a fix similar to:

- https://github.com/sigp/lighthouse/pull/7105

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
eserilev: [9d2ebf7](https://github.com/sigp/lighthouse/pull/7084/commits/9d2ebf706cdcd1601deabe0b5f2c6cea10dd90ec)

beacon_node/beacon_chain/src/attestation_verification.rs
eserilev: Added the check in `verify_head_block_is_known` when `block_opt` is `Some`

beacon_node/beacon_chain/src/attestation_verification.rs
eserilev: 72ef42b

michaelsproul (approved):

PR #7085, opened 2025-03-07
Ensure finalized block is the correct fork variant when constructing light client updates

eserilev:
This fixes a bug where, after a fork, we construct invalid light client updates for at least two epochs (until we reach finality). This is because the beacon states finalized checkpoint lies on the previous fork for at least the first two epochs. We will try to use that finalized checkpoint to construct a finalized light client header for the current fork variant instead of the previous fork variant.

This PR adds a check that the `finalized_header.fork_name == attested_header.fork_name`. If the check fails, we construct a "default" finalized light client header, one whose values are all set to `0x0`.

Note that this also resolves an issue that we thought was related to light client data construction requiring fork epochs to be divisible by 256. I've tested on kurtosis with `electra_epoch = 1` and the node was able to construct light client updates without any problems.

I've also re-introduced verifying light client updates in `basic-sim`


dapplion (approved):
Nice find and solution!

consensus/types/src/light_client_update.rs
```diff
@@ -218,7 +218,15 @@ impl<E: EthSpec> LightClientUpdate<E> {
                     LightClientHeaderAltair::block_to_light_client_header(attested_block)?;
 
                 let finalized_header = if let Some(finalized_block) = finalized_block {
-                    LightClientHeaderAltair::block_to_light_client_header(finalized_block)?
+                    if finalized_block
+                        .fork_name(chain_spec)
+                        .map_err(|_| Error::InconsistentFork)?
+                        == ForkName::Altair
```
michaelsproul: this should be `Altair` _or_ `Bellatrix`, I think? You could bind the fork name in the `match` arm like:

```rust
fork_name @ ForkName::Altair | fork_name @ ForkName::Bellatrix => {
    if finalized_block
       .fork_name(chain_spec)
       .map_err(|_| Error::InconsistentFork)?
       == fork_name
}
```

Maybe we could also use `fork_name_unchecked` here too to make it a bit neater.

testing/simulator/src/checks.rs
```diff
@@ -329,6 +334,20 @@ pub(crate) async fn verify_light_client_updates<E: EthSpec>(
...
+        let light_client_updates = client
+            .get_beacon_light_client_updates::<E>(sync_committee_period, 1)
+            .await
+            .map_err(|e| format!("Error while getting light client update: {:?}", e))?
+            .ok_or(format!("Light client update not found {slot:?}"))?;
+
+        // Ensure we're only storing a single light client update for the given sync committee period
+        if light_client_updates.len() != 1 {
+            return Err(format!(
+                "{} light client update(s) was returned when only one was expected.",
```
michaelsproul:
```suggestion
                "{} light client updates were returned when only one was expected.",
```

Nitpick: we know the number isn't 1 so we can use the plural (0 is plural)

michaelsproul (changes requested):
Looks good

Just a few suggested edits

consensus/types/src/light_client_update.rs
eserilev: Good catch, and yes, this is much cleaner. thanks! 
89168a3

dapplion (approved):

michaelsproul (approved):

PR #7086, opened 2025-03-07
Address cargo audit failure `RUSTSEC-2025-0009`

eserilev:
Bump ring version to resolve cargo audit 
https://github.com/sigp/lighthouse/actions/runs/13713278847/job/38353627642?pr=7085
https://rustsec.org/advisories/RUSTSEC-2025-0009

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2656,7 +2656,7 @@ dependencies = [
  "bls",
  "hex",
  "num-bigint-dig",
- "ring 0.16.20",
+ "ring 0.17.8",
  "sha2 0.9.9",
  "zeroize",
 ]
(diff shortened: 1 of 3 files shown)


eserilev:
I guess we'd merge to release-v7.0.0 and then cherry-pick to unstable?

michaelsproul:
Wondering if we should just merge an `--ignore` so we can unblock CI. As I understand we're waiting for @jxs to merge some stuff upstream in libp2p so we can avoid `ring 0.16` completely.

jxs:
Hi @michaelsproul released `libp2p-tls` `0.6.1` this should be addressed with
 `cargo update -p libp2p-tls`

eserilev:
I've updated libp2p-tls, and removed the relevant cargo audit ignore

jimmygchen (approved):
Looks good!

Issue #7039, opened 2025-02-26
Release v7.0.0-beta.clean


michaelsproul:
Plan for release `v7.0.0-beta.clean`:

Merge cleaned up versions of the necessary changes from `holesky-rescue` into `release-v7.0.0`. This will make it easier to reconcile the release with `unstable` in future.

TODO:

- [x] Backport https://github.com/sigp/lighthouse/pull/7059. Impl is already clean IMO, so we can likely just cherry-pick this commit. @eserilev: https://github.com/sigp/lighthouse/pull/7103
- [x] Backport https://github.com/sigp/lighthouse/pull/7072. As above. @eserilev.
- [x] Backport https://github.com/sigp/lighthouse/pull/7067. @eserilev / @michaelsproul: https://github.com/sigp/lighthouse/pull/7095.
- [ ] Backport fork choice/blocks by range changes from https://github.com/sigp/lighthouse/pull/7058 and https://github.com/sigp/lighthouse/pull/7066. These are combined into one PR by @dapplion: https://github.com/sigp/lighthouse/pull/7098.
- [x] Backport https://github.com/sigp/lighthouse/pull/7054. @michaelsproul. New PR: https://github.com/sigp/lighthouse/pull/7082.
- [x] Backport https://github.com/sigp/lighthouse/pull/7055. @jimmygchen. New PR: https://github.com/sigp/lighthouse/pull/7101
- [x] Backport `lighthouse/add_peer` https://github.com/sigp/lighthouse/pull/7068. @pawanjay176. New PR: #7198 
- [x] Backport https://github.com/sigp/lighthouse/pull/7071. This one is a bit tricky as the CI fixes are relevant across a few different PRs. https://github.com/sigp/lighthouse/pull/7103

Done previously:

- [x] Add CLI flag for additional block bans, but keep default ban: https://github.com/sigp/lighthouse/pull/7042
- [x] Specialise default block ban for Holesky: https://github.com/sigp/lighthouse/pull/7042
- [x] Remove `2 + 2 == 5` hacks: https://github.com/sigp/lighthouse/pull/7044
- [x] Cherry pick `--disable-attesting` changes for the VC into a clean PR: https://github.com/sigp/lighthouse/pull/7046
- [x] Add a flag for lengthening the VC timeouts used by `--use-long-timeouts` (we hardcoded 25x in `holesky-rescue`): https://github.com/sigp/lighthouse/pull/7041/files#r1971104870. @macladson 
[...]

PR #7087, opened 2025-03-07
Holesky rescue + hot tree states

dapplion:
Merged these two PRs
- https://github.com/sigp/lighthouse/pull/6744
- https://github.com/sigp/lighthouse/pull/6750

into 

- https://github.com/sigp/lighthouse/pull/7041

it's an alternative to 

- https://github.com/sigp/lighthouse/pull/7080

Hoped to be stable but not yet proven. So far I've

- [x] Run branch #6750 on Mainnet from checkpoint sync. I could sync for ~1 hours without errors
- [ ] Run branch #7087 on Holesky from checkpoint sync, against finalized checkpoint on https://checkpoint-sync.holesky.ethpandaops.io/ - on average it results in ~1MB / written diff
  - [x] Sync some epochs without errors
  - [ ] Reach the head without errors 
- [ ] Migrate existing Holesky big DB into #7087
  - [ ] successful migration (measure DB size before / after, note that there's no compaction)
  - [ ] Reach the head without errors

PR #7095, opened 2025-03-10
State cache tweaks

michaelsproul:
## Issue Addressed

Backport of:

- https://github.com/sigp/lighthouse/pull/7067

For:

- https://github.com/sigp/lighthouse/issues/7039

## Proposed Changes

- Prevent writing to state cache when migrating the database
- Add `state-cache-headroom` flag to control pruning
- Prune old epoch boundary states ahead of mid-epoch states
- Never prune head block's state
- Avoid caching ancestor states unless they are on an epoch boundary
- Log when states enter/exit the cache

Co-authored-by: Eitan Seri-Levi <eserilev@ucsc.edu>


dapplion:
Could you also add this logs to known when states enter / exit the cache?

https://github.com/sigp/lighthouse/pull/7087/commits/5b840946e97072da2d6b6b802deb726cd39d0136

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -816,7 +816,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             .get_blinded_block(&block_root)?
             .ok_or(Error::MissingBeaconBlock(block_root))?;
         let state = self
-            .get_state(&block.state_root(), Some(block.slot()))?
+            .get_state(&block.state_root(), Some(block.slot()), true)?
```
dapplion, line 821: The function `rev_iter_block_roots_from` is only used for tests, why `true` here? Maybe we can move this body to store_tests to have only less codepath to think about.

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -1519,8 +1520,9 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
                     })?
                     .ok_or(Error::NoStateForSlot(slot))?;
 
+                // Since this slot is less than the current head, don't cache the state
```
dapplion: Why?

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -6900,7 +6902,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
 
             let mut beacon_state = self
                 .store
-                .get_state(&beacon_state_root, Some(beacon_block.slot()))?
+                .get_state(&beacon_state_root, Some(beacon_block.slot()), true)?
```
dapplion, line 6914: Since this is used for analysis only, why cache states? Please write a justification comment above every .get_state call

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -7053,7 +7055,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
                 if signed_beacon_block.slot() % T::EthSpec::slots_per_epoch() == 0 {
                     let block = self.get_blinded_block(&block_hash).unwrap().unwrap();
                     let state = self
-                        .get_state(&block.state_root(), Some(block.slot()))
+                        .get_state(&block.state_root(), Some(block.slot()), true)
```
dapplion, line 7069: Since this is used for analysis only, why cache states?

beacon_node/beacon_chain/src/light_client_server_cache.rs
```diff
@@ -320,7 +320,7 @@ impl<T: BeaconChainTypes> LightClientServerCache<T> {
 
         // Compute the value, handling potential errors.
         let mut state = store
-            .get_state(block_state_root, Some(block_slot))?
+            .get_state(block_state_root, Some(block_slot), false)?
```
dapplion, line 326: Why false?

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -1015,7 +1022,7 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
             state_root
         };
         let mut opt_state = self
-            .load_hot_state(&state_root)?
+            .load_hot_state(&state_root, true)?
```
dapplion, line 1026: Why always true?

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -3169,8 +3203,9 @@ pub fn migrate_database<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>>(
             // Store slot -> state_root and state_root -> slot mappings.
             store.store_cold_state_summary(&state_root, slot, &mut cold_db_ops)?;
         } else {
+            // We purposely do not update the state cache to store this state
```
dapplion: Why?

beacon_node/store/src/state_cache.rs
```diff
@@ -250,8 +263,8 @@ impl<E: EthSpec> StateCache<E> {
         // This could probably be more efficient in how it interacts with the block map.
         for state_root in advanced_state_roots
             .iter()
-            .chain(mid_epoch_state_roots.iter())
```
dapplion, line 253: Please add a comment explaining the why of the ordering

beacon_node/store/src/state_cache.rs
```diff
@@ -236,7 +249,7 @@ impl<E: EthSpec> StateCache<E> {
                 }
             } else if is_advanced {
                 advanced_state_roots.push(state_root);
-            } else {
+            } else if state.get_latest_block_root(state_root) != self.head_block_root {
```
dapplion, line 262:
```suggestion
            } else if state.get_latest_block_root(state_root) != self.head_block_root {
                // Never prune the head state
```

consensus/state_processing/src/per_block_processing/tests.rs
```diff
@@ -1116,7 +1116,7 @@ async fn block_replayer_peeking_state_roots() {
         .unwrap();
     let parent_state = harness
         .chain
-        .get_state(&parent_block.state_root(), Some(parent_block.slot()))
+        .get_state(&parent_block.state_root(), Some(parent_block.slot()), true)
```
dapplion, line 1123: What about defining a constant `CACHE_STATE_IN_TESTS: bool = true`, so we don't have to think about cases, and can be tweaked easily if we need to debug.

beacon_node/beacon_chain/src/beacon_chain.rs
eserilev: I've added a TODO to move this to store_tests, I'll take care of that in a separate PR

In general we cache states in test fn's to make CI go brr

beacon_node/beacon_chain/src/beacon_chain.rs
eserilev: Michael added better comments here

beacon_node/beacon_chain/src/beacon_chain.rs
eserilev: added justifications above every .get_state call

beacon_node/beacon_chain/src/light_client_server_cache.rs
eserilev: we'd like to have the light client server be as minimally invasive as possible. This state should already be cached amyways, but I think its nice to prevent the light client server from ever writing to the state cache

beacon_node/store/src/hot_cold_store.rs
eserilev: we are actually caching this state again a few lines below. its a bit redundant, but doesnt seem to matter much

beacon_node/store/src/hot_cold_store.rs
eserilev: This is some hot state we're looking to migrate to the freezer db, we probably dont want to cache it. I added some better comments here

consensus/state_processing/src/per_block_processing/tests.rs
eserilev: [55f6d4f](https://github.com/sigp/lighthouse/pull/7095/commits/55f6d4f2b27248549d6b61b2a3026c373ad408ee)

beacon_node/store/src/state_cache.rs
eserilev: [5d62892](https://github.com/sigp/lighthouse/pull/7095/commits/5d6289283f81f9705923dad9be09fe34f9742171)

beacon_node/beacon_chain/src/attestation_rewards.rs
```diff
@@ -47,8 +47,10 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             .state_root_at_slot(state_slot)?
             .ok_or(BeaconChainError::NoStateForSlot(state_slot))?;
 
+        // This branch is reached from the HTTP API. We assume the user wants
```
dapplion, line 50: Can you add some tag to all these comments? Like
```suggestion
        // `update_cache = true` This branch is reached from the HTTP API. We assume the user wants
```
or
```suggestion
        // `get_state(.., true)` This branch is reached from the HTTP API. We assume the user wants
```
Will help future generations

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -1463,20 +1480,43 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
+                );
+                // Write the state anyway. Testing of this branch shows that states may exist in the
+                // cache but not be yet ever stored in the DB
+            }
+            PutStateOutcome::Finalized => {} // ignore
+        }
+
+        // TODO(hdiff): is this optimization necessary? Computing diffs is expensive so we may want
+        // to avoid it.
+        if self.load_hot_state_summary(state_root)?.is_some() {
```
dapplion: An `TODO(hdiff)` sneaked in. This check relies on the assumption that states are only ever deleted on finalization pruning. Otherwise we need a lock. I'm not sure if this check is same in this branch.

beacon_node/store/src/state_cache.rs
```diff
@@ -33,26 +33,33 @@ pub struct SlotMap {
...
+    head_block_root: Hash256,
+    headroom: NonZeroUsize,
 }
 
 #[derive(Debug)]
 pub enum PutStateOutcome {
     Finalized,
     Duplicate,
-    New,
+    // Includes deleted states as a result of this insertion
```
dapplion:
```suggestion
    /// Includes deleted states as a result of this insertion
```

book/src/help_bn.md
```diff
@@ -385,6 +385,9 @@ Options:
           Number of validators per chunk stored on disk.
       --slots-per-restore-point <SLOT_COUNT>
           DEPRECATED. This flag has no effect.
+      --state-cache-headroom <N>
+          Minimum number of states to cull from the state cache when it gets
+          full [default: 1]
```
dapplion, line 390: Should this flag be hidden?

michaelsproul: I don't mind it not being hidden. Or maybe we hide it until we implement:

- https://github.com/sigp/lighthouse/issues/7135

beacon_node/beacon_chain/src/attestation_rewards.rs
michaelsproul: I don't think this is really helpful compared to just grepping for `get_state`. I feel like we're bikeshedding these comments a bit at this point

beacon_node/store/src/hot_cold_store.rs
michaelsproul: I've deleted this check, as I want to keep this PR minimal and `release-v7.0.0` is working OK wrt temporary states/etc AFAICT. The bugs we saw on tree-states hot were caused by partially removing the temporary states concept.

jimmygchen (approved):
Reviewed the new diffs after Lion's review, looks good 👍

PR #7096, opened 2025-03-10
Add BlocksBy* protocols with specific block_root

dapplion:
## Issue Addressed

_@AgeManning pointed to me that we can have experimental ReqResp protocols in Lighthouse :) so here we go_

Implements 
- https://github.com/ethereum/consensus-specs/pull/3845

a new ReqResp protocol to make have better attributability of malicious sync peers when range syncing. See the spec PR for more details

## Proposed Changes

Since the spec is not agreed upon yet, this PR implements a new namespaces "feature" protocol suffixed by the spec PR number: `BlocksByRangeV2_3845`, `BlobsByRangeV1_3845`, `DataColumnsByRangeV1_3845`. Note that we may have a new `BlocksByRangeV3` protocol in the future before the 3845 feature is merged, thus having `BlocksByRangeV2_3845` and `BlocksByRangeV3_3845`.

These new protocol can exist in Lighthouse nodes only who will use them between themselves. They can also be deprecated in the future without coordination if the feature is merged or abandoned.

**Requester changes**

On a syncing chain of type Head that with peers with a Status message (head_root, head_slot), include `head_root` in the request.

**Server changes**

- I cherry-picked https://github.com/sigp/lighthouse/pull/7058 + https://github.com/sigp/lighthouse/pull/7066 to have fork-choice access

Then instead of querying the fork-choice for blocks with the head root, query by the request block_root.

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -7269,6 +7269,31 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             }
         }
     }
+
+    /// Retrieves block roots (in ascending slot order) within some slot range from fork choice.
+    pub fn block_roots_from_fork_choice(&self, start_slot: u64, count: u64) -> Vec<Hash256> {
+        let head_block_root = self.canonical_head.cached_head().head_block_root();
+        let fork_choice_read_lock = self.canonical_head.fork_choice_read_lock();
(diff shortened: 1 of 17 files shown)


jimmygchen:
@dapplion I'm reopening this - i assume you still want this?

dapplion:
- I'll close as stale and track this idea in https://github.com/sigp/lighthouse/issues/7678 depending on testing this modification may not be necessary

PR #7102, opened 2025-03-10
Add new HTTP endpoint to get supply for a given state root

alecalve:
## Issue Addressed

Computing how much ETH is on the Consensus layer requires currently querying `/eth/v1/beacon/states/{state_id}/validator_balances` and summing up all balances of all validators.

That data is 68MB of JSON which has to be downloaded and processed.

## Proposed Changes

This PR instead adds a simple enpoint where the computation is done at the node level, the returned data is therefore minimal.


michaelsproul:
Looks like this needs a sneaky `cargo fmt --all`

beacon_node/http_api/src/lib.rs
```diff
@@ -4448,6 +4448,23 @@ pub fn serve<T: BeaconChainTypes>(
             })
         });
 
+    // GET lighthouse/supply/{state_root}
+    let get_lighthouse_supply = warp::path("lighthouse")
+        .and(warp::path("supply"))
```
macladson:
```suggestion
    // GET lighthouse/analysis/global_validator_supply/{state_root}
    let get_lighthouse_supply = warp::path("lighthouse")
        .and(warp::path("analysis"))
        .and(warp::path("global_validator_supply"))
```

I think this fits the criteria for an `analysis` endpoint.
Also I think it might be worth using a more explicit name but I can't find an option I like. `global_validator_supply` seems a bit verbose, but `global_balances` seems a bit vague.

book/src/api_lighthouse.md
```diff
@@ -775,6 +775,28 @@ Caveats:
...
+  "data": [
+    674144000000000,
+    false,
+    true
+  ]
+}
+```
+
+The two boolean flags in the response are respectively:
+* `execution_optimism`: whether the response was computed using optimistically synced data
```
macladson:
```suggestion
* `execution_optimistic`: whether the response was computed using optimistically synced data
```

macladson (changes requested):
Thanks for the PR! This will be a neat endpoint to have. Just a couple of minor things

alecalve:
> Looks like this needs a sneaky `cargo fmt --all`

I ran the command, no files were changed

beacon_node/http_api/src/lib.rs
alecalve: I've changed the endpoint's path.

book/src/api_lighthouse.md
```diff
@@ -775,6 +775,28 @@ Caveats:
...
+    674144000000000,
+    false,
+    true
+  ]
+}
+```
+
+The two boolean flags in the response are respectively:
+* `execution_optimistic`: whether the response was computed using optimistically synced data
+* `finalized`: whether the response was computed over finalized dasta
```
macladson:
```suggestion
* `finalized`: whether the response was computed over finalized data
```

macladson (changes requested):
Looks like it needs a `make mdlint` to fix the Markdown formatting.

Also noticed a typo (we spellcheck the docs so this would've got picked up by CI as well)

alecalve:
For Markdown linting, I took the liberty of moving the check earlier in the test process as it's fast to run. In this MR, the suite failed nearly 15mins into testing when it could have failed much earlier.

macladson:
> For Markdown linting, I took the liberty of moving the check earlier in the test process as it's fast to run. In this MR, the suite failed nearly 15mins into testing when it could have failed much earlier.

Nice, that's a good change! 
And thanks for being prompt in implementing the suggestions, I'll test this out more thoroughly soon

alecalve:
@macladson I have no idea why the windows release tests failed, are they known to be flaky or could my changes somehow be the cause?

macladson:
> @macladson I have no idea why the windows release tests failed, are they known to be flaky or could my changes somehow be the cause?

Not sure what happened but I retriggered the tests and it has passed now, must've just been flaky

eserilev:
Hi @alecalve thanks for all your work so far! I was wondering if you'd be up for writing a test case for this endpoint? If not, I can go ahead and push up a test case myself. Thanks!

alecalve:
> Hi @alecalve thanks for all your work so far! I was wondering if you'd be up for writing a test case for this endpoint? If not, I can go ahead and push up a test case myself. Thanks!

I'll give it a try!

eserilev:
@alecalve  this looks great!

There are two small lint issues that are blocking CI. If you run `make lint-full` you should see them.

alecalve:
I wasn't aware of this final step of linting, I only ran `cargo fmt`. To make `clippy` happy, I split the HTTP test function into GET/POST methods to avoid triggering this:

```
error: this function may allocate 513901 bytes on the stack
    --> beacon_node/http_api/tests/tests.rs:7342:1
     |
7342 |   #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
7343 |   async fn lighthouse_endpoints() {
7344 | /     ApiTester::new()
7345 | |         .await
     | |______________- this is the largest part, at 37296 bytes for type `{async fn body of tests::ApiTester::new()}`
```

dapplion:
After reviewing the PR we would rather not have to mantain another endpoint if downstream consumers can compute it locally from JSON with a line of `jq`. Thanks for the contribution @alecalve

PR #7103, opened 2025-03-10
Pseudo finalization endpoint

eserilev:
## Issue Addressed

This is a backport of:
-  https://github.com/sigp/lighthouse/pull/7059
- https://github.com/sigp/lighthouse/pull/7071

For:
- https://github.com/sigp/lighthouse/issues/7039

## Proposed Changes

Introduce a new lighthouse endpoint that allows a user to force a pseudo finalization. This migrates data to the freezer db and prunes sidechains which may help reduce disk space issues on non finalized networks like Holesky

We also ban peers that send us blocks that conflict with the manually finalized checkpoint.

There were some CI fixes in https://github.com/sigp/lighthouse/pull/7071 that I tried including here

Co-authored with: @jimmygchen  @pawanjay176 @michaelsproul


michaelsproul:
I've cherry-picked the split check fix from this PR:

- #7105

It would be _awesome_ if we could add a unit test for manual finalization that shows it working correctly, and the node continuing to work after achieving real finalization.

I think we could add such a test in `beacon_node/beacon_chain/tests`.

eserilev:
Added a test that
- creates an un-finalized chain
- pseudo finalizes + runs some checks
- extends & finalizes the chain + runs some checks

michaelsproul (approved):
Looks good

I will add one more test case using `epochs_per_migration` to delay the split update, as I think that case is also interesting.

michaelsproul:
Regression test added and passing. Adding this to the queue.

PR #7109, opened 2025-03-11
Remove duplicate sync_tolerance_epochs config

michaelsproul:
## Proposed Changes

Delete duplicate sync tolerance epoch config in the HTTP API which is unused.

We introduced the `sync-tolerance-epoch` flag in this PR:

- https://github.com/sigp/lighthouse/pull/7030

Then refined it in this PR:

- https://github.com/sigp/lighthouse/pull/7044

Somewhere in the merge of `release-v7.0.0` into `unstable`, the config from the original PR which had been deleted came back. I think I resolved these conflicts, so my bad.


jimmygchen (approved):
LGTM!

PR #7112, opened 2025-03-11
Change `genesis-state-url-timeout`

antondlr:
## Issue Addressed

Timeouts sometimes occur on downloading the Holeksy genesis state from AWS, we've had reputable outside reports on this.  
It's around 200MB and hosted in APAC, it makes sense to bump the default, at least for Holesky. 

## Proposed Changes

Bump default timeout from 180 to 300 secs

## Additional Info

After interop, we noticed an outsized bill for this AWS bucket. This change could prevent or alleviate this in the future, if there are ever lots of genesis syncs on Holeksy again.

### DIFF
--- a/book/src/help_bn.md
+++ b/book/src/help_bn.md
@@ -167,7 +167,7 @@ Options:
           then this value will be ignored.
       --genesis-state-url-timeout <SECONDS>
           The timeout in seconds for the request to --genesis-state-url.
-          [default: 180]
+          [default: 300]
       --graffiti <GRAFFITI>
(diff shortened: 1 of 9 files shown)


michaelsproul (changes requested):
Needs an update to the book: `make && make cli-local`

And an update to the test that checks the default value

michaelsproul (approved):

michaelsproul:
the `genesis_state_url_default` test is broken (`grep "fn genesis_state_url_default"`)

macladson (approved):
This LGTM!

Issue #7100, opened 2025-03-10
High read IO due to blob pruning


michaelsproul:
Some users report CPU iowait increases and increased read IO running v7.0.0-beta. I think it's probably because we're reading all the blobs every epoch 😅. We changed our blob pruning algorithm as a result of:

- https://github.com/sigp/lighthouse/issues/6559

Graphs from an affected machine below:

CPU iowait (yellow):

![CPU iowait](https://github.com/user-attachments/assets/f6d8209a-d0ce-4b21-bd22-60306e953eac)

Read IO spiking to 250MB/s every epoch:

![Increased read IO](https://github.com/user-attachments/assets/22efba49-4750-4f8d-9de6-6414f37e60e8)

Given that I think we want to continue with abolishing `oldest_blob_slot` and allowing partial blob storage, I don't think we should revert to the previous algorithm. A quick fix is to change the frequency at which blob pruning runs to around once per day (`--epochs-per-blob-prune 256`). This would be an easy change to include in v7.0.0.

There might be another pruning algorithm we could use that is a hybrid of the previous one and the new one. Maybe we could iterate block roots backwards from the start of the data availability period, deleting blobs until we reach a block that is already pruned? This way we would only need to read the blobs to be pruned. However it has two downsides:

- We will stop pruning if there is a gap in blob availability in the database, and this could happen on healthy nodes if we support partial blob storage. Counter: maybe switching from a node with partial archival blob storage to archived historical blobs could require a more thorough manual prune?
- We need a backwards block iterator on the freezer block roots. Currently we only have a forwards iterator. This is not insurmountable, but needs to be written carefully given our recent issues with state loads and historical iterators.

eserilev:
I'd like to work on this

michaelsproul:
Removed v7.0.0 labels as we have merged the basic fix already. Leaving this issue open to track the more thorough fix.

michaelsproul:
Fixed (hopefully for the last time), in:

- https://github.com/sigp/lighthouse/pull/8266

PR #7113, opened 2025-03-11
Set `epochs-per-blob-prune` default to 256

eserilev:
## Issue Addressed

Partially #7100

## Proposed Changes

Set blob pruning to default to once per day

### DIFF
--- a/lighthouse/tests/beacon_node.rs
+++ b/lighthouse/tests/beacon_node.rs
@@ -1972,7 +1972,7 @@ fn prune_blobs_on_startup_false() {
 fn epochs_per_blob_prune_default() {
     CommandLineTest::new()
         .run_with_zero_port()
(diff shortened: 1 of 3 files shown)


lighthouse/tests/beacon_node.rs
```diff
@@ -1972,7 +1972,7 @@ fn prune_blobs_on_startup_false() {
 fn epochs_per_blob_prune_default() {
     CommandLineTest::new()
         .run_with_zero_port()
-        .with_config(|config| assert!(config.store.epochs_per_blob_prune == 1));
+        .with_config(|config| assert!(config.store.epochs_per_blob_prune == 256));
```
michaelsproul: Surprised Clippy didn't spot this one!

```suggestion
        .with_config(|config| assert_eq!(config.store.epochs_per_blob_prune, 256));
```

michaelsproul (approved):

PR #7116, opened 2025-03-12
Slasher HTTP API to import attestations from blocks

### LINKED ISSUE #7110: Slasher HTTP API to import attestations from blocks
In light of the recent Holesky incident, it may be useful to be able to manually import blocks and attestations to the slasher.

On Holesky, many slashers were shutdown during the non-finality period, meaning they missed the detection of many surround votes made by validators. One way we could get some of these slashings would be to manually feed blocks from the invalid chain to the slasher, so it can ingest the invalid attestations contained in those blocks and detect slashings.

The complication is, attestations in blocks require their `BeaconState`/committee to be converted into _indexed attestations_. And the slasher _needs_ indexed attestations (it needs to know which validators are attesting to the bad chain).

So I think we probably need a few new endpoints:

### Import attestations

- Path: `/lighthouse/slasher/import_attestations`
- Method: POST
- Request body: `Vec<IndexedAttestation>`
- Response: status code only

This endpoint is more important. It would import all of the indexed attestations to the slasher:

https://github.com/sigp/lighthouse/blob/bf955c7543dac8911a6f6c334b5b3ca4ef728d9c/beacon_node/beacon_chain/src/beacon_chain.rs#L4305-L4330

### Export indexed attestations

- Path: `/lighthouse/indexed_attestations`
- Method: POST
- Request body: `SignedBeaconBlock` (or `SignedBlindedBeaconBlock`)
- Response: `Vec<IndexedAttestation>`

This would need to be run on a node with the dodgy chain import (possibly imported via `lcli http-sync`). It would convert all of the attestations in the block into indexed attestations using the committees from the parent `BeaconState`.

These indexed attestations could then be saved to disk and plumbed into the `/lighthouse/slasher/import_attestations` endpoint.

### Import block

- Path: `/lighthouse/slasher/import_block`
- Method: POST
- Request body: `SignedBeaconBlock` (or `SignedBlindedBeaconBlock`?)
- Response: status code only

This method is not as useful as I originally hoped, it would _just_ be able to import the block header to the slasher due to the shuffling issue described above.

The implementation should pass the block itself to the slasher, probably using `accept_block_header` like in block verification:

https://github.com/sigp/lighthouse/blob/bf955c7543dac8911a6f6c334b5b3ca4ef728d9c/beacon_node/beacon_chain/src/block_verification.rs#L578

eserilev:
## Issue Addressed

Closes #7110 

## Proposed Changes

Adds three new endpoints

- `/lighthouse/slasher/import_attestations`: imports indexed attestations to the slasher db
- `/lighthouse/slasher/export_attestations`: exports indexed attestations from a provided block
- `/lighthouse/slasher/import_block`: imports a block header to the slasher db

## Additional Info

The endpoints don't ever fail, we might want to handle specific failures and return relevant status codes in those cases.

TODOs:
- [x] Add test cases
- [ ] Test on holesky
- [ ] Support SSZ?

### DIFF
--- a/beacon_node/http_api/src/slasher.rs
+++ b/beacon_node/http_api/src/slasher.rs
@@ -1,0 +1,80 @@
+use std::sync::Arc;
+
+use beacon_chain::{BeaconChain, BeaconChainTypes};
+use either::Either;
+use serde_json::Value;
+use types::{IndexedAttestation, SignedBeaconBlock, SignedBlindedBeaconBlock};
+
+#[allow(clippy::type_complexity)]
+pub fn deserialize_block_payload<T: BeaconChainTypes>(
+    payload: Value,
+) -> Option<Either<SignedBeaconBlock<T::EthSpec>, SignedBlindedBeaconBlock<T::EthSpec>>> {
+    if let Ok(block) = serde_json::from_value::<SignedBeaconBlock<T::EthSpec>>(payload.clone()) {
+        Some(Either::Left(block))
+    } else {
+        serde_json::from_value::<SignedBlindedBeaconBlock<T::EthSpec>>(payload)
+            .ok()
+            .map(Either::Right)
+    }
+}
+
+pub fn import_indexed_attestations<T: BeaconChainTypes>(
+    indexed_attestations: Vec<IndexedAttestation<T::EthSpec>>,
+    chain: Arc<BeaconChain<T>>,
+) {
+    if let Some(slasher) = chain.slasher.as_ref() {
+        for indexed_attestation in indexed_attestations {
+            slasher.accept_attestation(indexed_attestation);
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


beacon_node/http_api/src/slasher.rs
```diff
@@ -0,0 +1,80 @@
...
+            .ok()
+            .map(Either::Right)
+    }
+}
+
+pub fn import_indexed_attestations<T: BeaconChainTypes>(
+    indexed_attestations: Vec<IndexedAttestation<T::EthSpec>>,
+    chain: Arc<BeaconChain<T>>,
+) {
+    if let Some(slasher) = chain.slasher.as_ref() {
```
michaelsproul, line 25: TODO: could also check signatures here

Issue #7108, opened 2025-03-11
Offloading KZG Proof Computation from the beacon node


jimmygchen:
EIP change: https://github.com/ethereum/EIPs/pull/9378
`execution-api` PR: https://github.com/ethereum/execution-apis/pull/630

The exact spec change is still pending and under active discussion, below is a rough list of what needs to be done:

- Update code to assemble `DataColumnSidecar`:
  - Extend blobs
  - Use cell proofs from EL (no computaion)
- Update integration with EL `getPayloadV5` and EL `getBlobsV2`
- `GetBlobSidecar` API
  - We no longer have the blob KZG proofs, may need a v2 to return cell proofs?

jimmygchen:
Marked as `blocked` for now as spec change hasn't been finalised.

jimmygchen:
@ethDreamer and I have started working on this as the spec seems to be getting close to finalised and geth already has an implementation.

jimmygchen:
Completed in #7117 🎉

PR #7117, opened 2025-03-12
Offloading KZG Proof Computation from the beacon node

jimmygchen:
## Issue Addressed

Addresses #7108

- Add EL integration for `getPayloadV5` and `getBlobsV2`
- Offload proof computation and use proofs from EL RPC APIs

## Tasks

- [x] EL interop ready for [peerdas-devnet-6](https://notes.ethereum.org/@ethpandaops/peerdas-devnet-6#)
- [x] Update `getBlobsSidecars` Beacon API 
- [x] Remove data column receiver and await for the computation, so that we can check data column slashing.
- [x] KZG verification [during block production](https://github.com/sigp/lighthouse/blob/898e3fa4f81bd2086a6a3a0ad23c0a946c0fe863/beacon_node/beacon_chain/src/beacon_chain.rs#L5910-L5920) 
- [x] Update failing Fulu tests
- [x] Update EF spec tests


beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -148,51 +171,116 @@ pub async fn fetch_and_process_engine_blobs<T: BeaconChainTypes>(
...
+        debug!(
+            log,
+            "Ignoring EL blobs response";
+            "info" => "block has already been imported",
+        );
+        return Ok(None);
+    }
+
+    // TODO: we could get rid of the receiver and just await on the blocking task, given the
+    // compute cell operation is very cheap.
```
dapplion: That sounds good 👍

consensus/types/src/lib.rs
```diff
@@ -272,7 +272,14 @@ pub type Address = fixed_bytes::Address;
...
 pub type BLSFieldElement = Uint256;
 pub type Blob<E> = FixedVector<u8, <E as EthSpec>::BytesPerBlob>;
-pub type KzgProofs<E> = VariableList<KzgProof, <E as EthSpec>::MaxBlobCommitmentsPerBlock>;
+// Note on List limit:
+// - Deneb to Electra: `MaxBlobCommitmentsPerBlock`
+// - Fulu: `MaxCellCommitmentsPerBlock`
+// We choose to use a single type (with the larger value from Fulu as `N`) instead of having to
+// introduce a new type for Fulu. This is to avoid messy conversions and having to add extra types
+// with no gains - as `N` does not impact serialisation at all, and only affects merkleization,
+// which we don't current do on `KzgProofs` anyway.
```
dapplion, line 281: Agree

consensus/types/src/eth_spec.rs
```diff
@@ -146,6 +146,11 @@ pub trait EthSpec:
     /// Must be set to `BytesPerFieldElement * FieldElementsPerCell`.
     type BytesPerCell: Unsigned + Clone + Sync + Send + Debug + PartialEq;
 
+    /// The maximum number of cell commitments per block
+    ///
+    /// MaxBlobCommitmentsPerBlock * FieldElementsPerExtendedBlob / FieldElementsPerCell
+    type MaxCellCommitmentsPerBlock: Unsigned + Clone + Sync + Send + Debug + PartialEq;
```
jimmygchen: Should rename this to `MaxCellsPerBlock`, there's no such thing as cell commitments. (commitments are for blobs)

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -132,136 +170,176 @@ pub async fn fetch_and_process_engine_blobs<T: BeaconChainTypes>(
...
+                )
+                .discard_timer_on_break(&mut timer);
+                drop(timer);
+
+                let custody_columns = data_columns_result
+                    .map(|mut data_columns| {
+                        data_columns.retain(|col| custody_columns_indices.contains(&col.index));
+                        data_columns
+                    })
+                    .map_err(FetchEngineBlobError::DataColumnSidecarError)?;
```
jimmygchen, line 315: This filtering is required now because DA checker checks for column count when adding columns. Previously we only send it a one shot receiver so filtering was not required.

This should also address the bug - block made available via `getBlobs` will no longer store all 128 columns on full nodes. This addresses part of https://github.com/sigp/lighthouse/issues/6465.

beacon_node/beacon_chain/src/fetch_blobs.rs
jimmygchen: Completed in https://github.com/sigp/lighthouse/pull/7117/commits/d6197be93718db58c85c879d5cc3c4971c184a35

consensus/types/src/eth_spec.rs
jimmygchen: Addressed in https://github.com/sigp/lighthouse/pull/7117/commits/a5f9649c74aa79abbac90057b0e039c0bc0053eb

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -5798,15 +5799,30 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
-                    expected_kzg_commitments,
-                    blobs.iter().collect(),
-                    &kzg_proofs,
-                )
-                .map_err(BlockProductionError::KzgError)?;
+                if self
+                    .spec
+                    .is_peer_das_enabled_for_epoch(slot.epoch(T::EthSpec::slots_per_epoch()))
+                {
+                    // TODO: extend blobs and verify cells against proofs
```
dapplion:
```suggestion
                    // TODO(das): extend blobs and verify cells against proofs
```

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -473,10 +485,17 @@ fn verify_data_column_sidecar<E: EthSpec>(
...
-    if data_column.column.len() != data_column.kzg_commitments.len()
-        || data_column.column.len() != data_column.kzg_proofs.len()
-    {
-        return Err(GossipDataColumnError::InconsistentCommitmentsOrProofLength);
+    let cells_len = data_column.column.len();
+    let commitments_len = data_column.kzg_commitments.len();
+    let proofs_len = data_column.kzg_proofs.len();
+    if cells_len != commitments_len || cells_len != proofs_len {
+        return Err(
+            GossipDataColumnError::InconsistentCommitmentsOrProofLength {
```
dapplion: Why not split this into two errors? Weird to have an "Or" in an error name

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -125,136 +164,160 @@ pub async fn fetch_and_process_engine_blobs<T: BeaconChainTypes>(
...
-/// allows blobs / data columns to propagate without waiting for processing.
-///
-/// An `mpsc::Sender` is then used to send the produced data columns to the `beacon_chain` for it
-/// to be persisted, **after** the block is made attestable.
-///
-/// The reason for doing this is to make the block available and attestable as soon as possible,
-/// while maintaining the invariant that block and data columns are persisted atomically.
-fn spawn_compute_and_publish_data_columns_task<T: BeaconChainTypes>(
+/// Offload the data column computation to a blocking task to avoid holding up the async runtime.
+async fn spawn_compute_and_publish_data_columns_task<T: BeaconChainTypes>(
```
dapplion:
```suggestion
async fn compute_and_publish_data_columns_task<T: BeaconChainTypes>(
```
Name is misleading now

beacon_node/beacon_chain/src/kzg_utils.rs
```diff
@@ -43,6 +44,33 @@ pub fn validate_blob<E: EthSpec>(
...
 
+// TODO
+pub fn verify_cell_proof_batch<E: EthSpec>(
+    _kzg: &Kzg,
+    _cells: Vec<Cell<E>>,
+    _cell_indices: Vec<u64>,
+    _kzg_proofs: Vec<KzgProof>,
+    _kzg_commitments: Vec<KzgCommitment>,
+) -> Result<(), KzgError> {
+    todo!()
```
dapplion: unwrap

beacon_node/network/src/network_beacon_processor/mod.rs
```diff
@@ -842,6 +843,7 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
         block: Arc<SignedBeaconBlock<T::EthSpec, FullPayload<T::EthSpec>>>,
         block_root: Hash256,
         publish_blobs: bool,
+        custody_columns: HashSet<ColumnIndex>,
```
dapplion: Why add this new argument? You can just fetch it at the start of this function
```rust
let is_supernode = self.network_globals.is_supernode();
let custody_columns = self.network_globals.sampling_columns.clone();
```

consensus/types/src/eth_spec.rs
```diff
@@ -566,6 +573,7 @@ impl EthSpec for GnosisEthSpec {
     type MaxPendingDepositsPerEpoch = U16;
     type FieldElementsPerCell = U64;
     type FieldElementsPerExtBlob = U8192;
+    type MaxCellsPerBlock = U33554432;
```
dapplion, line 576: Could we make an alias named `VeryLongLen` or similar? To clarify that 33554432 is not a magic number

jimmygchen: It's documented at the trait level here:
https://github.com/sigp/lighthouse/blob/a5f9649c74aa79abbac90057b0e039c0bc0053eb/consensus/types/src/eth_spec.rs#L149-L152

Why would we want a generic name like `VeryLongLen` given this is a derive number from spec?

beacon_node/network/src/network_beacon_processor/mod.rs
jimmygchen: See comments here:
https://github.com/sigp/lighthouse/blob/d6197be93718db58c85c879d5cc3c4971c184a35/beacon_node/network/src/network_beacon_processor/sync_methods.rs#L201-L202

and here
https://github.com/sigp/lighthouse/blob/a5f9649c74aa79abbac90057b0e039c0bc0053eb/beacon_node/network/src/network_beacon_processor/gossip_methods.rs#L1558-L1559

I actually want to take the value from the `GossipVerifiedBlock` and `RpcBlock` but there's a bug in the `RpcBlock` right now (custody column count set to 0 on lookups), so I've added a `TODO` there.

beacon_node/beacon_chain/src/kzg_utils.rs
jimmygchen: Oh yeah, this was a `TODO` and wasn't called. I've implemented it in https://github.com/sigp/lighthouse/pull/7117/commits/7f65c506fc963c89199ef3458dedd58f4cdcb0b9

beacon_node/beacon_chain/src/fetch_blobs.rs
jimmygchen: It is still spawning a blocking task, but I guess it's not relevant to the caller, I'll update the method name.

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Implemented in https://github.com/sigp/lighthouse/commit/7f65c506fc963c89199ef3458dedd58f4cdcb0b9

beacon_node/beacon_chain/src/data_column_verification.rs
jimmygchen: I didn't really want to touch this - but ran into some issues with geth so I've only added fields to the error in this PR to help with debugging. Can we leave the refactor to a separate PR?

jimmygchen: Alright might as well include it given the size of the PR https://github.com/sigp/lighthouse/pull/7117/commits/5ce829989b132d1acf322e2f88ec2cff7c4bded4

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -7,34 +7,46 @@
...
 
 pub enum BlobsOrDataColumns<T: BeaconChainTypes> {
     Blobs(Vec<GossipVerifiedBlob<T, DoNotObserve>>),
     DataColumns(DataColumnSidecarList<T::EthSpec>),
 }
 
+pub enum EngineGetBlobsOutput<E: EthSpec> {
+    Blobs(FixedBlobSidecarList<E>),
+    DataColumns(DataColumnSidecarList<E>),
+}
```
dapplion: Can you add a `///` comment here specifying if these are all columns for the block or a filtered set. Plus specify that we they are considered to be valid due to trusted EL

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -125,136 +164,160 @@ pub async fn fetch_and_process_engine_blobs<T: BeaconChainTypes>(
...
-        if !blobs_to_import_and_publish.is_empty() {
-            publish_fn(BlobsOrDataColumns::Blobs(blobs_to_import_and_publish));
-        }
+        inc_counter(&metrics::BLOBS_FROM_EL_HIT_TOTAL);
+    }
 
-        None
-    };
+    // Partial blobs response isn't useful for PeerDAS, so we don't bother building and publishing data columns.
+    if num_fetched_blobs != num_expected_blobs {
```
dapplion, line 229: Above you check `num_fetched_blobs == 0`. Can `num_expected_blobs` be 0? Else you can remove the above condition. Otherwise add here the `inc_counter(&metrics::BLOBS_FROM_EL_MISS_TOTAL);` as this is also a miss.

beacon_node/beacon_chain/src/fetch_blobs.rs
dapplion: add this context as a code comment

beacon_node/network/src/network_beacon_processor/mod.rs
dapplion: But you have access here to network_globals, what's the need to take it from the block variable? That's only necessary inside the beacon_chain where we can't query network_globals

consensus/types/src/eth_spec.rs
dapplion: Ah ok

beacon_node/network/src/network_beacon_processor/mod.rs
jimmygchen: Because DA checker operates on the block `custody_columns_count`, not the global one.
https://github.com/sigp/lighthouse/blob/6f31d44343086ddaa70b396632c0eee724256454/beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs#L187

Network should compute the sampling columns and pass it on to fetch blobs.
If the global sampling columns changes, then this would not match, and block will not get imported.

beacon_node/beacon_chain/src/fetch_blobs.rs
jimmygchen: if `num_expected_blobs` is 0, it's returned early here
https://github.com/sigp/lighthouse/blob/73fe3175bd30c9ead1f66593435be56e76de5c9e/beacon_node/beacon_chain/src/fetch_blobs.rs#L78-L79

`BLOBS_FROM_EL_MISS_TOTAL` was originally used for tracking empty responses, but we can repurpose that. I generally avoid making these unrelated changes to keep the PR concise and easier to review, but happy to do it since you're reviewing :P
https://github.com/sigp/lighthouse/blob/70850fe58d5676d9072d418f04f4db7257c1513b/beacon_node/beacon_chain/src/metrics.rs#L1663

I think this file can use some refactoring, but I have also avoided doing this for the interest of keeping this moving quickly.

dapplion (approved):
LGTM!

PR #7119, opened 2025-03-12
Remove gas-limit flag dependency on builder-proposals

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Even though the VC doesn't affect gas limit without using an external builder, I don't think there's a hard requirement at the cli flag level for the requires("builder-proposals"). 

This is also causing issues with some devops scenarios as raised by @barnabasbusa

### DIFF
--- a/lighthouse/tests/validator_client.rs
+++ b/lighthouse/tests/validator_client.rs
@@ -487,7 +487,6 @@ fn no_gas_limit_flag() {
 fn gas_limit_flag() {
     CommandLineTest::new()
         .flag("gas-limit", Some("600"))
-        .flag("builder-proposals", None)
         .run()
         .with_config(|config| assert_eq!(config.validator_store.gas_limit, Some(600)));
 }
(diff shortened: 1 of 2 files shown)


barnabasbusa (approved):

michaelsproul:
I think we should at least log a warning if the builder flag isn't provided, because IMO accepting flags that do nothing is also bad UX

"placebo flags"

michaelsproul:
Thinking about this more I'm inclined to leave it as-is.

Most of the time with flags we follow a correct-by-construction sort of approach, where we are as strict with flag dependencies as possible. This is better for UX in a way, because it means the user doesn't need to check logs to know that a flag application was effective.

E.g. a recent change made along these lines (more strict dependencies):

- https://github.com/sigp/lighthouse/pull/6079

Maybe we should document this policy so that it's a bit clearer?

pawanjay176:
> This is better for UX in a way, because it means the user doesn't need to check logs to know that a flag application was effective.

I agree with you, for production scenarios, its better to be strict. 
Sorry @barnabasbusa , looks like you'll have to do some additional scripting to make it work :P

> Maybe we should document this policy so that it's a bit clearer?

Maybe CONTRIBUTING.md would be a good place?

michaelsproul:
Closing because we decided not to impl this change

Issue #6973, opened 2025-02-11
Add a beacon node flag to advertise a false custody group count for testing


jimmygchen:
For PeerDAS devnet testing, we're starting have stable devnets and looking to start doing some unhappy case testing. 

We already have `--malicious-withhold-count`, which allows the proposer to withhold a specified number of data columns during block publishing. It would be useful to have a flag that withholds custody columns over RPC too, for scenarios like:
1. A proposer uses `--malicious-withhold-count` to withhold some data columns - this may trigger a data column by root request from the mesh peer, and we could make the proposer not serve this requqest and force them to get it via alternatives like distributed blob building and column reconstruction - this would help us test the effectiveness of the solutions
2. A node pretends to be a supernode in the network but don't actually serve the data
 
## Proposed solution

Instead of touching all the RPC code paths to withhold data columns, we could simply have a CLI flag `--advertise-false-custody-group-count`, that just advertise a fake custody group count, and make peers think the node custodies more columns - that way we don't have to touch any RPC logic. The node just won't serve data they don't have.

This flag should be a hidden flag and clearly states it's for testing only and should NOT be used in production.

startup-dreamer:
Hi @jimmygchen, I think I can work on this.

jimmygchen:
Completed in #7593

PR #7120, opened 2025-03-13
Feat: add new `--advertise-false-custody-group-count` cli flag for PeerDAS testing

startup-dreamer:
## Issue Addressed

Closes: #6973

## Proposed Changes

This PR adds a new CLI flag `--advertise-false-custody-group-count` that allows nodes to advertise a false custody group count in their metadata for PeerDAS testing purposes. This enables testing scenarios where nodes need to mislead peers about their custody group count without altering the actual RPC logic.

### When this flag is set:
- If PeerDAS is scheduled, the node will advertise the specified custody group count in its metadata.
- If the specified count is outside the valid range (`<= number_of_custody_groups`), the node will log a warning and use the normal count.
- If PeerDAS is not scheduled, the flag has no effect.

### Future Work
This flag is intended to be temporary and will be removed before the PeerDAS release, as indicated by the TODO comment.

### DIFF
--- a/beacon_node/lighthouse_network/src/config.rs
+++ b/beacon_node/lighthouse_network/src/config.rs
@@ -150,7 +150,6 @@ pub struct Config {
     pub idontwant_message_size_threshold: usize,
 
     /// Advertise a false custody group count in metadata.
-    /// Makes peers think this node custodies more columns than it actually does.
     pub advertise_false_custody_group_count: Option<u64>,
 }
 
--- a/beacon_node/lighthouse_network/src/config.rs
+++ b/beacon_node/lighthouse_network/src/config.rs
@@ -148,6 +148,10 @@ pub struct Config {
     /// Configuration for the minimum message size for which IDONTWANT messages are send in the mesh.
     /// Lower the value reduces the optimization effect of the IDONTWANT messages.
     pub idontwant_message_size_threshold: usize,
+
+    /// Advertise a false custody group count in metadata.
+    /// Makes peers think this node custodies more columns than it actually does.
+    pub advertise_false_custody_group_count: Option<u64>,
 }
 
 impl Config {
@@ -372,6 +376,7 @@ impl Default for Config {
             invalid_block_storage: None,
             inbound_rate_limiter_config: None,
             idontwant_message_size_threshold: DEFAULT_IDONTWANT_MESSAGE_SIZE_THRESHOLD,
+            advertise_false_custody_group_count: None,
         }
     }
 }
--- a/beacon_node/lighthouse_network/src/service/mod.rs
+++ b/beacon_node/lighthouse_network/src/service/mod.rs
@@ -203,7 +203,8 @@ impl<E: EthSpec> Network<E> {
             if let Some(false_count) = config.advertise_false_custody_group_count {
                 false_count
             } else {
-                ctx.chain_spec.custody_group_count(config.subscribe_all_data_column_subnets)
+                ctx.chain_spec
+                    .custody_group_count(config.subscribe_all_data_column_subnets)
             }
         });
         let meta_data = utils::load_or_build_metadata(&config.network_dir, custody_group_count);
--- a/beacon_node/lighthouse_network/src/service/mod.rs
+++ b/beacon_node/lighthouse_network/src/service/mod.rs
@@ -199,8 +199,23 @@ impl<E: EthSpec> Network<E> {
 
         // Construct the metadata
         let custody_group_count = ctx.chain_spec.is_peer_das_scheduled().then(|| {
-            ctx.chain_spec
-                .custody_group_count(config.subscribe_all_data_column_subnets)
+            // Use the false custody group count if specified, otherwise use the normal one
+            if let Some(false_count) = config.advertise_false_custody_group_count {
+                // Ensure the false count is within valid range
+                if (ctx.chain_spec.custody_requirement..=ctx.chain_spec.number_of_custody_groups).contains(&false_count) {
+                    warn!(log, "Using false custody group count for testing"; "count" => false_count);
+                    false_count
+                } else {
+                    warn!(log, "Specified false custody group count is out of valid range, using normal count"; 
+                        "false_count" => false_count,
+                        "min" => ctx.chain_spec.custody_requirement,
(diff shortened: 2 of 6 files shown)


beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -199,8 +199,23 @@ impl<E: EthSpec> Network<E> {
 
         // Construct the metadata
         let custody_group_count = ctx.chain_spec.is_peer_das_scheduled().then(|| {
-            ctx.chain_spec
-                .custody_group_count(config.subscribe_all_data_column_subnets)
+            // Use the false custody group count if specified, otherwise use the normal one
+            if let Some(false_count) = config.advertise_false_custody_group_count {
+                // Ensure the false count is within valid range
+                if (ctx.chain_spec.custody_requirement..=ctx.chain_spec.number_of_custody_groups).contains(&false_count) {
```
dapplion: We just need to check that `false_count <= number_of_custody_groups` and that can happen when parsing the CLI flag. Logging here would be very verbose

beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -199,8 +199,23 @@ impl<E: EthSpec> Network<E> {
 
         // Construct the metadata
         let custody_group_count = ctx.chain_spec.is_peer_das_scheduled().then(|| {
```
dapplion, line 201: Does this affect both the ENR value and the value returned in the Status ReqResp message?

startup-dreamer: ENR is built with the metadata (false `custody_group_count` if provided) and The metadata returned in response to METADATA requests (which is part of the ReqResp protocol) so I'd say yes it affects both.

beacon_node/lighthouse_network/src/service/mod.rs
startup-dreamer: I treid to create this check when parsing the cli flag but couldn't get `number_of_custody_groups` but I don't think you can get it while parsing.

dapplion: We can't issue a warn log every time this code path is hit

startup-dreamer: No worries I can clean that up.

startup-dreamer: @dapplion 
Would it be better to show a single warning message when `advertise_false_custody_group_count` is provided and `false_count > number_of_custody_groups`? This way, the tester will be informed that `false_count` is out of bounds and not being used, as `custody_group_count` defaults to the standard value (`number_of_custody_groups` if supernode else `custody_requirement`) or should I just remove all the logs.

dapplion: if the user configures the client with an improper config value the best UX is:
- As soon as possible inform the user the value is wrong
- Don't allow the client to start with the invalid value

startup-dreamer: I've added the validation check thank you for your feedback @dapplion.

beacon_node/src/config.rs
```diff
@@ -49,6 +49,17 @@ pub fn get_config<E: EthSpec>(
 
     let mut client_config = ClientConfig::default();
 
+    if let Some(false_custody_group_count) =
+        clap_utils::parse_optional::<u64>(cli_args, "advertise-false-custody-group-count")?
```
dapplion: Why parse it twice? You can pass access to the spec to set_network_config

boot_node/src/config.rs
```diff
@@ -53,7 +53,7 @@ impl<E: EthSpec> BootNodeConfig<E> {
 
         let mut network_config = NetworkConfig::default();
 
-        set_network_config(&mut network_config, matches, &data_dir)?;
+        set_network_config(&mut network_config, matches, &data_dir, None)?;
```
startup-dreamer, line 56: Here `spec` is passed as `None` because unlike the beacon node the bootnode only needs to facilitate discovery and doesn't validate chain-specific data. Hopefully this is correct way to do it.

beacon_node/src/config.rs
startup-dreamer: Fixed it.

beacon_node/lighthouse_network/src/config.rs
```diff
@@ -148,6 +148,10 @@ pub struct Config {
     /// Configuration for the minimum message size for which IDONTWANT messages are send in the mesh.
     /// Lower the value reduces the optimization effect of the IDONTWANT messages.
     pub idontwant_message_size_threshold: usize,
+
+    /// Advertise a false custody group count in metadata.
+    /// Makes peers think this node custodies more columns than it actually does.
```
dapplion:
```suggestion
    /// Advertise a false custody group count in metadata.
```

beacon_node/lighthouse_network/src/service/mod.rs
dapplion: If the user restarts the node with a different value of the false_count flag, will we bump the `seq_number` of the ENR?

startup-dreamer: I rechecked I don't think it bumps the `seq_number` of ENR, Since the ENR is built from the chain spec and the 
`subscribe_all_data_column_subnets`, not from the false count flag (sorry for prev comment I am still getting familiar with the codebase). Does this make sense @dapplion 
https://github.com/sigp/lighthouse/blob/bf955c7543dac8911a6f6c334b5b3ca4ef728d9c/beacon_node/lighthouse_network/src/discovery/enr.rs#L256-L264

startup-dreamer: But I think I have to give more time to unserstand this.

jimmygchen:
Hi @startup-dreamer , there are some failing CI checks, would you mind taking a look please?

jimmygchen:
Sorry @startup-dreamer - I'm closing this PR as we haven't heard from you - it's now a bit stale and  requires quite a bit of rework with the introduction of validator custody. Thanks for your contribution though!

PR #7125, opened 2025-03-13
fix-new-Update contributing.md

helicopter-1:
fix to a new one


pawanjay176:
The current one redirects to .org

PR #7130, opened 2025-03-13
update gossipsub to the latest upstream revision

jxs:
## Issue Addressed

I feel it's preferable to do this explicitly by updating the revision on `Cargo.toml` rather than implicitly by letting `Cargo.lock` control the revision of the branch.


jimmygchen (approved):
Thanks @jxs, LGTM!

PR #7131, opened 2025-03-13
Add error reason on UnknownValidator

dapplion:
## Issue Addressed

This PR would be helpful to debug a current `UnknownValidator` error. Includes a hint on the size of the validator pubkey cache at the time of query.


michaelsproul:
Closing now that we've added a workaround and a root cause mitigation is in progress:

- https://github.com/sigp/lighthouse/issues/7216

PR #7132, opened 2025-03-13
Add logs when adding to pubkey cache

dapplion:
## Issue Addressed

Help debug errors similar to `UnknownValidator` where appears to be some pubkey cache corruption. Would be good to confirm that all keys are loaded properly.


jimmygchen:
I think this will be useful. Please fix build errors. Thanks!

jimmygchen (approved):

jimmygchen:
@dapplion Do we want to target this PR for the release 7.0.0 branch?

beacon_node/beacon_chain/src/validator_pubkey_cache.rs
```diff
@@ -64,6 +65,8 @@ impl<T: BeaconChainTypes> ValidatorPubkeyCache<T> {
             }
         }
 
+        debug!(indices = indices.len(), "Loaded pubkey cache from store");
```
michaelsproul, line 68: This isn't going to work on `release-v7.0.0` because we don't have tracing

jimmygchen (changes requested):
See Michael's comment above

michaelsproul:
Closing now that we've added a workaround and a root cause mitigation is in progress:

- https://github.com/sigp/lighthouse/issues/7216

PR #7133, opened 2025-03-14
Remove all prod eth1 related code

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

After the electra fork which includes EIP 6110, the beacon node no longer needs the eth1 bridging mechanism to include new deposits as they are provided by the EL as a `deposit_request`. So after electra + a transition period where the finalized bridge deposits pre-fork are included through the old mechanism, we no longer need the elaborate machinery we had to get deposit contract data from the execution layer. 

Since holesky has already forked to electra and completed the transition period, this PR basically checks to see if removing all the eth1 related logic leads to any surprises.


michaelsproul:
Hey @pawanjay176 I am keen to get this merged, would it help if I resolved merge conflicts and rolled this out on a few more nodes?

pawanjay176:
Resolved the conflicts. Would be awesome if we can roll this on some nodes with validators and check that block production works without any hitch 🙏

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -2398,13 +2378,11 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
         // If there's no eth1 chain then it's impossible to produce blocks and therefore
         // useless to put things in the op pool.
-        if self.eth1_chain.is_some() {
-            let (attestation, attesting_indices) =
-                verified_attestation.into_attestation_and_indices();
-            self.op_pool
-                .insert_attestation(attestation, attesting_indices)
-                .map_err(Error::from)?;
-        }
+        // TODO(pawan): do we want to differentiate between full node/staking node function
```
michaelsproul: I think the extra load is unlikely to be impactful, so we may as well keep things simple and not add another flag here.

It's possible we could use the `--staking` flag, but that would be a breaking change, and the default should probably be staking so we don't break people's setups.

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -660,18 +652,6 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
     }
 
-    /// Persists `self.eth1_chain` and its caches to disk.
-    pub fn persist_eth1_cache(&self) -> Result<(), Error> {
-        let _timer = metrics::start_timer(&metrics::PERSIST_ETH1_CACHE);
-
-        if let Some(eth1_chain) = self.eth1_chain.as_ref() {
-            self.store
-                .put_item(&ETH1_CACHE_DB_KEY, &eth1_chain.as_ssz_container())?;
-        }
```
michaelsproul, line 670: We should probably add a new schema version with a database migration that deletes this data. And a downgrade as well (probably a no-op).

beacon_node/http_api/src/lib.rs
```diff
@@ -4593,106 +4528,6 @@ pub fn serve<T: BeaconChainTypes>(
...
-                task_spawner.blocking_json_task(Priority::P1, move || {
-                    if chain.eth1_chain.is_some() {
-                        Ok(())
-                    } else {
-                        Err(warp_utils::reject::custom_not_found(
-                            "staking is not enabled, \
-                        see the --staking CLI flag"
-                                .to_string(),
-                        ))
-                    }
```
michaelsproul, line 4634: I wonder if we should keep this endpoint but just return `Ok(())`? In case we want to bring back some distinction between staking and non-staking in future.

beacon_node/src/config.rs
```diff
@@ -265,34 +263,6 @@ pub fn get_config<E: EthSpec>(
...
-
-    if cli_args.get_flag("eth1-purge-cache") {
-        client_config.eth1.purge_cache = true;
-    }
-
-    if let Some(follow_distance) =
-        clap_utils::parse_optional(cli_args, "eth1-cache-follow-distance")?
-    {
-        client_config.eth1.cache_follow_distance = Some(follow_distance);
-    }
```
michaelsproul: We should probably keep these 3 flags and log warnings here, seeing as we had not previously deprecated them. Usually we try to avoid fully removing flags unless they've been deprecated first.

beacon_node/src/config.rs
```diff
@@ -809,11 +761,6 @@ pub fn get_config<E: EthSpec>(
         }
     }
 
-    // Note: This overrides any previous flags that enable this option.
-    if cli_args.get_flag("disable-deposit-contract-sync") {
-        client_config.sync_eth1_chain = false;
-    }
```
michaelsproul: We should also deprecate this rather than fully removing it (log a warning here)

beacon_node/http_api/src/lib.rs
michaelsproul: Seeing as we used this in the VC when waiting for genesis, I think it's safer to keep this endpoint for now. This might avoid breaking some devnet setups unnecessarily (e.g. running LH VC v7.0.0 with LH BN v7.1.0).

michaelsproul:
I think the failing tests are fine as well, just need to update some expected state roots due to the changes to the eth1 data voting (not significant for what those tests are testing anyway).

beacon_node/src/config.rs
michaelsproul: Done in this commit:

- https://github.com/sigp/lighthouse/pull/7133/commits/4f413d07cacd8a3dfe9f709a41d8b7408d0846e4

I completely removed `--eth1` and `--disable-eth1` as we had already deprecated them in v7.

beacon_node/beacon_chain/src/beacon_chain.rs
michaelsproul: Added in:

- https://github.com/sigp/lighthouse/pull/7133/commits/924de422474a9d11fb8786e8bfd8041e23925304

beacon_node/http_api/src/lib.rs
michaelsproul: Put this back for now https://github.com/sigp/lighthouse/pull/7133/commits/bcf3c28d1956b6eb624576cd4a4e78f188266733

michaelsproul (approved):
I'm so keen for this PR!

pawanjay176:
Thanks for reviewing and fixing up everything Michael ❤️

PR #7137, opened 2025-03-14
Log `file appender` initialization errors properly

ThreeHrSleep:
## Issue Addressed

It'll help us debug the issue with rolling file appender initialization (`Failed to initialize libp2p rolling file appender`)

cc: @macladson


jimmygchen (approved):
Thanks LGTM!

Issue #6240, opened 2024-08-07
Use `MIN_EPOCHS_FOR_DATA_COLUMN_SIDECARS_REQUESTS` and `MAX_REQUEST_DATA_COLUMN_SIDECARS` for data columns


jimmygchen:
## Description

The current `das` branch incorrectly re-used the Deenb `MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS` for data column sidecars.

- There's a new config value for this, `MIN_EPOCHS_FOR_DATA_COLUMN_SIDECARS_REQUESTS`, which should be used instead. This should be configurable at run time (config value).
- Also added `MAX_REQUEST_DATA_COLUMN_SIDECARS`

Spec: https://github.com/ethereum/consensus-specs/blob/8f8ab03acf1656c54f3a81ef18878f853a1cc4c6/specs/_features/eip7594/p2p-interface.md#configuration

startup-dreamer:
I'll try to raise a pr for this.

jimmygchen:
Update: `MAX_REQUEST_DATA_COLUMN_SIDECAR` removed from spec
https://github.com/ethereum/consensus-specs/pull/4322

jimmygchen:
Completed in #7646

PR #7139, opened 2025-03-16
Feat: Add configuration values for data column (`MIN_EPOCHS_FOR_DATA_COLUMN_SIDECARS_REQUESTS` and `MAX_REQUEST_DATA_COLUMN_SIDECARS`)

startup-dreamer:
## Issue Addressed

Closes: #6240

## Proposed Changes

This pr adds dedicated configuration values for data column sidecars that were previously incorrectly using blob sidecar configuration values:

1. Added `MIN_EPOCHS_FOR_DATA_COLUMN_SIDECARS_REQUESTS` to replace the incorrect use of `MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS` for data column sidecars.
2. Added `MAX_REQUEST_DATA_COLUMN_SIDECARS` to make it configurable using configuration files.
3. Updated all network configuration files to include these new values.

These changes ensure proper separation between blob sidecar and data column sidecar configurations, allowing them to be independently configured at runtime without code modifications.

cc: @jimmygchen


beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -679,8 +679,9 @@ async fn availability_cache_maintenance_service<T: BeaconChainTypes>(
                 let cutoff_epoch = std::cmp::max(
                     finalized_epoch + 1,
                     std::cmp::max(
-                        current_epoch
-                            .saturating_sub(chain.spec.min_epochs_for_blob_sidecars_requests),
+                        current_epoch.saturating_sub(
+                            chain.spec.min_epochs_for_data_column_sidecars_requests,
+                        ),
```
dapplion: `do_maintenance` prunes blocks that are outside the da_window. Now the da_window size is dependant on if holds data columns or blobs. So if:
- block is for epoch < FULU_FORK_EPOCH, `min_epochs_for_blob_sidecars_requests` applies
- block is for epoch >= FULU_FORK_EPOCH, `min_epochs_for_data_column_sidecars_requests` applies

In a way it feels `min_epochs_for_data_column_sidecars_requests` is just `min_epochs_for_blob_sidecars_requests_fulu`.

So in this case the pruning function should conditionally use one value or the other depending on the block epoch

beacon_node/client/src/builder.rs
```diff
@@ -312,7 +312,7 @@ where
                         // a sync right before blobs start to disappear from the P2P
                         // network.
                         let reduced_p2p_availability_epochs = spec
-                            .min_epochs_for_blob_sidecars_requests
+                            .min_epochs_for_data_column_sidecars_requests
```
dapplion: This value must apply only post FULU

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -2706,7 +2706,8 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
         let min_current_epoch = self.get_split_slot().epoch(E::slots_per_epoch()) + 2;
         let min_data_availability_boundary = std::cmp::max(
             deneb_fork_epoch,
-            min_current_epoch.saturating_sub(self.spec.min_epochs_for_blob_sidecars_requests),
+            min_current_epoch
+                .saturating_sub(self.spec.min_epochs_for_data_column_sidecars_requests),
```
dapplion: Similar to https://github.com/sigp/lighthouse/pull/7139#discussion_r1997928871 must be conditional on the epoch we are attempting to prune

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -675,14 +682,18 @@ async fn availability_cache_maintenance_service<T: BeaconChainTypes>(
                     .fork_choice_read_lock()
                     .finalized_checkpoint()
                     .epoch;
+
+                // pre-fulu: use blob sidecar window, post-fulu: use data column window
+                let min_epochs = if chain.spec.is_peer_das_enabled_for_epoch(current_epoch) {
+                    chain.spec.min_epochs_for_data_column_sidecars_requests
+                } else {
+                    chain.spec.min_epochs_for_blob_sidecars_requests
+                };
```
dapplion, line 691: This will break in the transition. Consider a case where `min_epochs_for_blob_sidecars_requests = Inf` (never prune) and `min_epochs_for_data_column_sidecars_requests = 0` (prune instantly), after the Fulu fork you will prune all blobs which is incorrect. Instead of doing this if on the current epoch you need to do it on each blob / column you attempt to prune.

PR #7140, opened 2025-03-16
Fix panic when running VC 

ThreeHrSleep:
## Issue Addressed

~~https://github.com/sigp/lighthouse/issues/7147~~
Edit: VC was panicking during start up( in `debug` specifically)

### DIFF
--- a/lighthouse/src/main.rs
+++ b/lighthouse/src/main.rs
@@ -568,8 +568,8 @@ fn run<E: EthSpec>(
                     .join("logs"),
             ),
             Some(("validator_client", vc_matches)) => {
-                let base_path = if vc_matches.contains_id("validators-dir") {
-                    parse_path_or_default(vc_matches, "validators-dir")?
+                let base_path = if vc_matches.contains_id("validators_dir") {
+                    parse_path_or_default(vc_matches, "validators_dir")?
                 } else {
                     parse_path_or_default(matches, "datadir")?.join(DEFAULT_VALIDATOR_DIR)
                 };
--- a/lighthouse/tests/validator_client.rs
+++ b/lighthouse/tests/validator_client.rs
@@ -61,7 +61,7 @@ fn datadir_flag() {
 fn validators_and_secrets_dir_flags() {
     let dir = TempDir::new().expect("Unable to create temporary directory");
     CommandLineTest::new()
-        .flag("validators-dir", dir.path().join("validators").to_str())
+        .flag("validators_dir", dir.path().join("validators").to_str())
         .flag("secrets-dir", dir.path().join("secrets").to_str())
         .run_with_no_datadir()
         .with_config(|config| {
@@ -74,7 +74,7 @@ fn validators_and_secrets_dir_flags() {
 fn validators_dir_alias_flags() {
     let dir = TempDir::new().expect("Unable to create temporary directory");
     CommandLineTest::new()
-        .flag("validator-dir", dir.path().join("validators").to_str())
+        .flag("validator_dir", dir.path().join("validators").to_str())
         .flag("secrets-dir", dir.path().join("secrets").to_str())
         .run_with_no_datadir()
         .with_config(|config| {
--- a/lighthouse/tests/validator_client.rs
+++ b/lighthouse/tests/validator_client.rs
@@ -61,7 +61,7 @@ fn datadir_flag() {
 fn validators_and_secrets_dir_flags() {
     let dir = TempDir::new().expect("Unable to create temporary directory");
     CommandLineTest::new()
-        .flag("validators_dir", dir.path().join("validators").to_str())
+        .flag("validators-dir", dir.path().join("validators").to_str())
         .flag("secrets-dir", dir.path().join("secrets").to_str())
         .run_with_no_datadir()
         .with_config(|config| {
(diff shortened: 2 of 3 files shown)


lighthouse/tests/validator_client.rs
```diff
@@ -74,7 +74,7 @@ fn validators_and_secrets_dir_flags() {
 fn validators_dir_alias_flags() {
     let dir = TempDir::new().expect("Unable to create temporary directory");
     CommandLineTest::new()
-        .flag("validator-dir", dir.path().join("validators").to_str())
+        .flag("validator_dir", dir.path().join("validators").to_str())
```
jimmygchen: Not sure if I'm missing something here, why are we changing this flag?

ThreeHrSleep: i mainly changed it because it was panicking while i was running the tests (even though `--validator-dir` was working normally while running vc manually at that moment)...i tested again now & it seems to not be panicking anymore while running the tests...

i'll revert that change, thanks

lighthouse/src/main.rs
```diff
@@ -568,8 +568,8 @@ fn run<E: EthSpec>(
                     .join("logs"),
             ),
             Some(("validator_client", vc_matches)) => {
-                let base_path = if vc_matches.contains_id("validators-dir") {
-                    parse_path_or_default(vc_matches, "validators-dir")?
+                let base_path = if vc_matches.contains_id("validators_dir") {
+                    parse_path_or_default(vc_matches, "validators_dir")?
```
jimmygchen, line 572: I'm not sure this is the actual issue, I just tried running the latest `unstable` using:

```
  lighthouse validator_client \
    --validators-dir=/validator-data/keys-2 \
    --secrets-dir=/validator-data/secrets-2 \
   ...
```

And I can see that it's loaded the `validator_dir` correctly 

```
Mar 17 06:57:19.199 INFO Starting validator client                     beacon_nodes: ["http://beacon:5052/"], validator_dir: "/validator-data/keys-2"
```

I can't seem to reproduce this on a different machine (I was looking to see what the actual error is after you added logging), I wonder if it's a file permission issue 🤔 

Are you able to reproduce this?

ThreeHrSleep: i was constantly facing this (& still having this issue if i run `unstable`)
(even when i'm not using `--validators-dir`)
```
Mismatch between definition and access of `validators-dir`. Unknown argument or group id.  Make sure you are using the argument id and not the short or long flags

stack backtrace:
   0: rust_begin_unwind
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/std/src/panicking.rs:692:5
   1: core::panicking::panic_fmt
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/panicking.rs:75:14
   2: clap_builder::parser::error::MatchesError::unwrap
             at /Users/chow/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/clap_builder-4.5.32/src/parser/error.rs:32:9
   3: clap_builder::parser::matches::arg_matches::ArgMatches::contains_id
             at /Users/chow/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/clap_builder-4.5.32/src/parser/matches/arg_matches.rs:526:9
   4: lighthouse::run
             at ./lighthouse/src/main.rs:571:36
   5: lighthouse::main::{{closure}}
             at ./lighthouse/src/main.rs:477:35
   6: core::result::Result<T,E>::and_then
             at /Users/chow/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/result.rs:1353:22
   7: lighthouse::main
             at ./lighthouse/src/main.rs:454:18
   8: core::ops::function::FnOnce::call_once
             at /Users/chow/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5
```

macladson: This issue only seems to happen in `debug`, compiling in release makes it go away. Although it's unrelated to the actual usage of `--validators-dir` flag. You can simply run the VC and it will panic in `debug` with the mismatch error. Not sure what the solution here is

ThreeHrSleep: Thanks for pointing out the issue !!
As it shouldn't cause much trouble right now(& also because i'm not sure what the solution is either )
i'm going to ignore this for now & come back after dealing with the actual https://github.com/sigp/lighthouse/issues/7147

jimmygchen: Hi @ThreeHrSleep 

I've just managed to reproduce it on the same machine with latest `unstable`, it was indeed a file permission issue:
```
Failed to initialize libp2p rolling file appender: Failed to create directory: Permission denied (os error 13)
```

I think the handling is probably ok (im not sure what the behaviour was for file perms before tracing), however Im wondering if we could avoid building the discv5 and libp2p layers when running the validator client since they aren't relevant?

ThreeHrSleep: Thanks for the suggestion!!
just created a [new PR](https://github.com/sigp/lighthouse/pull/7163) to disable file logging for discv5 & libp2p when running VC

jimmygchen:
Closing this as I believe this is covered in #7168.

PR #7145, opened 2025-03-17
Support Hoodi testnet

michaelsproul:
## Proposed Changes

Hardcode config for the upcoming `hoodi` testnet so we can run with `--network hoodi`.

## Additional Info

Thanks to @antondlr for setting up the hosting of the genesis state on S3.


jimmygchen (approved):
LGTM!

Issue #7146, opened 2025-03-17
Debug level logs not written to log file (beacon.log) on `unstable`


jimmygchen:
This bug affects latest `unstable` and was introduced when tracing was merged.

## Actual Behaviour

Lighthouse debug logs no longer gets written to the debug log file (`beacon.log`), both by default and with the `--logfile-debug-level=debug` flag. Workaround right now is to set `RUST_LOG=debug` or use `--debug-level debug` as suggested by Mac, but discv5 logs in the beacon.log file gets very noisy.

## Expected Behaviour

- Debug level logs should be written to the log file by default.
- `--logfile-debug-level` flag should overwrite the default when supplied.
- discv5 logs should not be written to the log file, it should only be written to the `discv5.log` file.

jimmygchen:
Completed in #7149 & #7150, thanks @macladson ! 🙏

jimmygchen:
@macladson re-opening this issue - I'm running latest `unstable` and I'm still seeing discv5 logs in the `beacon.log` file.

I haven't got a chance to dig deeper, but the only log related CLI flag used was `--debug-level=info`.

macladson:
Fixed in #7168

PR #7149, opened 2025-03-17
Remove unnecessary `filter_layer` in logger builder

macladson:
## Issue Addressed

https://github.com/sigp/lighthouse/issues/7146

## Proposed Changes

Removes `filter_layer` from the builder as this was acting as a "global minimum". We don't actually need this, since we are using more granular control in `stdout_logging_layer` and `file_logging_layer`. Removing this restores control of the logfiles level back to the `--logfile-debug-level` flag (and defaults to debug).

## Additional Info

Discv5 logs are still present in the output, so that is still an ongoing issue.


jimmygchen (approved):
Nice find, thanks @macladson ! 🙏

PR #7150, opened 2025-03-17
Remove `discv5` logs from logfile output

macladson:
## Issue Addressed

#7146 

## Proposed Changes

Set `dep_logs` to false when building the `file_logging_layer`. This will remove both `discv5` and `libp2p` logs from the output.

## Additional Info

There's more cleanup required here but this is the simplest fix for now.


jimmygchen (approved):
Nice, thanks for the quick fix 🙏 

What are the remaining clean ups? would you mind creating issues so we can keep track of them? Is there anything that needs to be done for the  v7.0.0 release?

It looks like `dep_logs: true` is no longer used anywhere else, is it still worth keeping it?

PR #7160, opened 2025-03-18
feat: update hoodi config to use new fields

### LINKED ISSUE #7158: Use `MAX_PAYLOAD_SIZE` in place of `GOSSIP_MAX_SIZE`
Per this PR:

- https://github.com/eth-clients/hoodi/pull/5

The spec has moved from using `GOSSIP_MAX_SIZE` to `MAX_PAYLOAD_SIZE`.

While adding Hoodi in this PR (https://github.com/sigp/lighthouse/pull/7145) I opted just to continue using `GOSSIP_MAX_SIZE`. All these parameters have the default values currently, so it doesn't make any difference.

Eventually we should switch over to using `MAX_PAYLOAD_SIZE` and the new values. I don't think we need to worry about backwards-compatibility, because these params are relatively unimportant, and the only thing we will be breaking will be local testnet setups using old testnet dirs.

varun-doshi:
## Issue Addressed

Closes #7158 

## Proposed Changes

-  `MAX_PAYLOAD_SIZE` in place of `GOSSIP_MAX_SIZE`
- removes `MAX_CHUNK_SIZE`


michaelsproul:
This was intentionally omitted, as we haven't implemented those features yet:

- https://github.com/sigp/lighthouse/issues/7158

varun-doshi:
Should I close this?

PR #7164, opened 2025-03-18
Update ring to 0.17.14 to fix build compat

michaelsproul:
## Proposed Changes

Cross builds were failing since:

- https://github.com/sigp/lighthouse/pull/7086

This seems to have been due to a regression upstream in `ring` which is noted in the v0.17.14 release notes. I'm hoping that updating remedies it.

> Compatibility with GNU binutils 2.29 (used on Amazon Linux 2), and probably even earlier versions, was restored. It is expected that ring 0.17.14 will build on all the systems that 0.17.12 would build on.

https://github.com/briansmith/ring/blob/main/RELEASES.md#version-01714-2025-03-11

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7457,9 +7457,9 @@ dependencies = [
 
 [[package]]
 name = "ring"
-version = "0.17.13"
+version = "0.17.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


michaelsproul:
This did fix the build. Just needs a green tick

realbigsean (approved):
tick

Issue #7165, opened 2025-03-18
Cannot set `--debug-level` to `crit`


macladson:
### Issue

Since merging #6339, we can no longer set `--debug-level` to `crit` as `tracing` does not support the `CRIT` log level and we are using a shim to make it appear that `CRIT` logs are actually appearing.

See this example below:
```
> lighthouse bn --execution-endpoint "http://localhost:8551" --debug-level crit
Unsupported log level

> lighthouse bn --execution-endpoint "http://localhost:8551" --debug-level help
error: invalid value 'help' for '--debug-level <LEVEL>'
  [possible values: info, debug, trace, warn, error, crit]

```

### Possible Solutions
1. Remove `crit` as an option from the CLI entirely and map `--debug-level crit` to `--debug-level error` with a deprecation notice. This is simplest solution, users can just move to using `--debug-level error`, which while isn't as strict as `crit`, might be fine. I suspect most users aren't using `--debug-level crit` anyway.
2. Intercept the CLI value, convert it to `LevelFilter::Error`. We then need to find a way to restrict log output to only `ERROR`s which are using the `crit!` shim, I think we can slightly modify the log extractor to get this to work.

### Additional Info
The same applies to `--logfile-debug-level`.

dapplion:
> Remove crit as an option from the CLI entirely

👍

jimmygchen:
Completed in #7169 🎉

Issue #7155, opened 2025-03-17
Debug level log appearing in stdout when log level is set to INFO


jimmygchen:
This bug affects latest unstable and was introduced when tracing was merged.

## Actual Behaviour

I'm seeing some DEBUG level logs even when log level is set to `INFO` via the `--debug-level=info` flag

```
[cl-1-lighthouse-geth] Mar 17 12:22:02.893 DEBUG Batch downloaded                              epoch: 5618, blocks: 28, batch_state: "[p,E,E,E,E]", awaiting_batches: 0, chain: 3, service: "range_sync"
```

I don't see any other DEBUG logs though, not sure why only this log made it to stdout. 

## Expected Behaviour

DEBUG log should not be written to stdout when `--debug-level` is set to info.

chong-he:
A few more debug logs appearing in `journalctl`:

```
May 16 16:28:45.037 DEBUG New head chain started syncing                id: 1, component: "range_sync"
May 16 16:28:45.037 INFO  Sync state updated                            old_state: Stalled, new_state: Syncing Head Chain, service: "sync"
May 16 16:28:46.861 DEBUG Batch downloaded                              epoch: 13519, blocks: 31, batch_state: "[p,D,D,E,E]", awaiting_batches: 0, chain: 1, service: "range_sync"
May 16 16:28:46.876 DEBUG Batch downloaded                              epoch: 13520, blocks: 31, batch_state: "[P,p,D,E,E]", awaiting_batches: 1, chain: 1, service: "range_sync"
May 16 16:28:48.545 DEBUG Batch downloaded                              epoch: 13521, blocks: 21, batch_state: "[P,p,p,E,E]", awaiting_batches: 2, chain: 1, service: "range_sync"
May 16 16:28:54.000 INFO  Syncing                                       peers: "3", distance: "86 slots (17 mins)", est_time: "--"
May 16 16:28:54.000 WARN  Syncing deposit contract block cache          est_blocks_remaining: "48"
                     service: "network"
May 16 16:29:06.001 INFO  Syncing                                       peers: "2", distance: "54 slots (10        est_blocks_remaining: "48"
May 16 16:29:06.477 DEBUG New head chain started syncing                id: 11, component: "range_sync"
May 16 16:29:06.642 DEBUG Batch downloaded                              epoch: 13520, blocks: 31, batch_state: "[E,E,E,E,E]", awaiting_batches: 0, chain: 11, service: "range_sync"
[...]

Issue #7153, opened 2025-03-17
Unable to disable file logging


macladson:
Since merging #6339, we can no longer disable the file logger.
Previously, setting _either_ `--logfile-max-size 0` or `--logfile-max-number 0` would disable the file logger. 

This is additionally problematic because having `--logfile-max-size 0` will instead rotate the file every single log line. Users using `--logfile-compress` will have the additional issue of a race condition between the file getting compressed and it being rotated, and so will see hundreds of:

```
Couldn't compress log file: File IO error: No such file or directory (os error 2)
```

macladson:
Fixed in #7168

Issue #7148, opened 2025-03-17
libp2p.log no longer capture libp2p logs


jimmygchen:
This bug affects latest `unstable` and was introduced when tracing was merged.

## Actual Behaviour

I've noticed that libp2p logs are not being written to the libp2p.log file. I've tried running it with `RUST_LOG=debug`, which should log quite a bit, but the file was still empty.

## Expected Behaviour

libp2p logs should be written to the libp2p.log file.

ackintosh:
https://github.com/macladson/lighthouse/pull/4 should fix this. 🙏

macladson:
Fixed in #7168, credit to @ackintosh 🙏

PR #7168, opened 2025-03-18
Tracing cleanup

macladson:
## Issue Addressed

#7153
#7146 
#7147
#7148 -> Thanks to @ackintosh 

## Proposed Changes

This PR does the following:
1. Disable logging to file when using either `--logfile-max-number 0` or `--logfile-max-size 0`. Note that disabling the log file in this way will also disable `discv5` and `libp2p` logging.
1.  `discv5` and `libp2p` logging will be disabled by default unless running `beacon_node` or `boot_node`. This also should fix the VC panic we were seeing.
1. Removes log rotation and compression from `libp2p` and `discv5` logs. It is now limited to 1 file and will rotate based on the value of the `--logfile-max-size` flag. We could potentially add flags specifically to control the size/number of these, however I felt a single log file was sufficient. Perhaps @AgeManning has opinions about this?
1. Removes all dependency logging and references to `dep_log`.
1.  Introduces workspace filtering to file and stdout. This explicitly allows logs from members of the Lighthouse workspace, disallowing all others. It uses a proc macro which pulls the member list from cargo metadata at compile time. This might be over-engineered but my hope is that this list will not require maintenance.
1. Unifies file and stdout JSON format. With slog, the formats were slightly different. @threehrsleep worked to maintain that format difference, to ensure there was no breaking changes. If these format differences are actually problematic we can restore it, however I felt the added complexity wasn't worth it.
1. General code improvements and cleanup.

## Outstanding Issues

- This does not fix [certain `debug` logs related to the `sync` span being included in `INFO` logs](https://github.com/sigp/lighthouse/issues/7155). I suspect our spans will need to be reworked to fix this.
- [Trying to set `--debug-level crit` or `logfile-debug-level crit` will still fail.](https://github.com/sigp/lighthouse/issues/7165)  (Fix available) -> #7169 
- Logfile permission issues: #7170 #7171


ackintosh:
@macladson I've filed a PR to fix missing libp2p(gossipsub) logs. https://github.com/macladson/lighthouse/pull/4 Please have a look when you have time.

macladson:
> @macladson I've filed a PR to fix missing libp2p(gossipsub) logs. [macladson#4](https://github.com/macladson/lighthouse/pull/4) Please have a look when you have time.

Amazing, thank you so much for noticing this! :heart:

ackintosh:
Additionally, I noticed that the `--logfile-debug-level` option does not affect to `discv5.log` or `libp2p.log`. Currently both logs contain TRACE logs so the file size increases very quickly.

I've created https://github.com/macladson/lighthouse/pull/5 to fix this. 🙏

michaelsproul:
Removed the v7.0.0 label as we aren't releasing v7.0.0 from unstable, so it won't include tracing at all

AgeManning:
Single log file is fine with me :)

macladson:
> Additionally, I noticed that the `--logfile-debug-level` option does not affect to `discv5.log` or `libp2p.log`. Currently both logs contain TRACE logs so the file size increases very quickly.

I wonder if defaulting to `DEBUG` is actually fine here, then we can use `RUST_LOG` to control the verbosity of the discv5 and libp2p logs? Seems to me that `--logfile-debug-level info` (or higher) would yield mostly empty discv5/libp2p logfiles anyway, whereas `--logfile-debug-level info` is perfectly sensible for `beacon.log`. I agree that we should not default to `TRACE` though

common/lighthouse_macros/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "lighthouse_macros"
```
jimmygchen: This crate name is a bit generic and may encourage adding random macros into this crate.
What about something more specific, like `workspace_meta_macros` or `workspace_crates_macros`

macladson: I actually named it that with the idea that if we needed other Lighthouse-specific/misc macros, they could reuse the crate instead of creating a new one :thinking: 
Happy to change it of course

lighthouse/environment/src/lib.rs
```diff
@@ -204,7 +204,7 @@ impl<E: EthSpec> EnvironmentBuilder<E> {
     ) -> (
         Self,
         LoggingLayer,
-        LoggingLayer,
+        Option<LoggingLayer>,
```
jimmygchen, line 214: Worth adding a method doc here since we're returning two `LoggingLayer`s.

common/lighthouse_macros/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+edition = { workspace = true }
+
+[lib]
+proc-macro = true
+
+[dependencies]
+cargo_metadata = { workspace = true }
+proc-macro2 = { workspace = true }
```
jimmygchen: Doesn't look like this crate uses proc-macro2?

Cargo.toml
```diff
@@ -159,6 +161,7 @@ mockito = "1.5.0"
 num_cpus = "1"
 parking_lot = "0.12"
 paste = "1"
+proc-macro2 = "1"
```
jimmygchen: This seems unused

lighthouse/environment/src/tracing_common.rs
```diff
@@ -12,36 +15,44 @@ pub fn construct_logger<E: EthSpec>(
...
-    let (builder, file_logging_layer, stdout_logging_layer, sse_logging_layer_opt) =
+    let (builder, stdout_logging_layer, file_logging_layer, sse_logging_layer_opt) =
         environment_builder.init_tracing(logger_config.clone(), logfile_prefix);
 
-    let dependency_log_filter =
-        FilterFn::new(filter_dependency_log as fn(&tracing::Metadata<'_>) -> bool);
+    let libp2p_discv5_layer = if let Some(subcommand_name) = subcommand_name {
+        if subcommand_name == "beacon_node" || subcommand_name == "boot_node" {
+            if logger_config.max_log_size == 0 || logger_config.max_log_number == 0 {
+                // User has explictly disabled logging to file.
```
jimmygchen:
```suggestion
                // User has explicitly disabled logging to file.
```

common/lighthouse_macros/src/lib.rs
```diff
@@ -0,0 +1,39 @@
...
+                .packages
+                .iter()
+                .find(|package| &package.id == member_id)
+                .map(|package| package.name.clone())
+        })
+        .collect())
+}
+
+#[proc_macro]
+pub fn workspace_crates(_input: TokenStream) -> TokenStream {
```
jimmygchen: Is this macro meant for static use only? If so, we could just remove the `Result` and emit a `compile_error!()` at compile time if `cargo_metadata` fails.

common/logging/src/utils.rs
```diff
@@ -0,0 +1,33 @@
+use lighthouse_macros::workspace_crates;
+use std::collections::HashSet;
+use tracing_subscriber::filter::FilterFn;
+
+const WORKSPACE_CRATES: Result<&[&str], &str> = workspace_crates!();
```
jimmygchen: This is a compile-time constant and currently expands to `Ok(&[...])`, so it doesn't fail at compile time even though the macro already knows whether it failed or not at compile time. We could just panic during the build instead of expanding to a `Result` here and fail at runtime?

common/lighthouse_macros/Cargo.toml
jimmygchen: Yeah thats reasonable, happy to leave it for now.

common/logging/src/tracing_libp2p_discv5_logging_layer.rs
```diff
@@ -0,0 +1,113 @@
...
+use std::path::PathBuf;
+use tracing::Subscriber;
+use tracing_appender::non_blocking::{NonBlocking, WorkerGuard};
+use tracing_subscriber::{layer::Context, Layer};
+
+pub struct Libp2pDiscv5TracingLayer {
+    pub libp2p_non_blocking_writer: NonBlocking,
+    pub _libp2p_guard: WorkerGuard,
+    pub discv5_non_blocking_writer: NonBlocking,
+    pub _discv5_guard: WorkerGuard,
```
jimmygchen: Do `_libp2p_guard` and `_discv5_guard` have to be `pub`?

common/logging/src/tracing_libp2p_discv5_logging_layer.rs
```diff
@@ -0,0 +1,113 @@
...
+impl tracing_core::field::Visit for LogMessageExtractor {
+    fn record_debug(&mut self, _: &tracing_core::Field, value: &dyn std::fmt::Debug) {
+        self.message = format!("{} {:?}", self.message, value);
+    }
+}
+
+pub fn create_libp2p_discv5_tracing_layer(
+    base_tracing_log_path: Option<PathBuf>,
+    max_log_size: u64,
+) -> Option<Libp2pDiscv5TracingLayer> {
```
jimmygchen, line 62: The API for this function can be simplified - we know `base_tracing_log_path` when calling this function, so this can be simplified by removing `Option` here and only call this function if the log path is `Some`.

common/logging/src/tracing_libp2p_discv5_logging_layer.rs
```diff
@@ -0,0 +1,113 @@
...
+    if let Some(mut tracing_log_path) = base_tracing_log_path {
+        // Ensure that `tracing_log_path` only contains directories.
+        for p in tracing_log_path.clone().iter() {
```
jimmygchen, line 65: I don't really understand what this does.

Let's say if my `tracing_log_path` is `/home/user/lighthouse`
then the first iteration of this `for` loop would join the above with `home`, and because `/home/user/lighthouse/home` is not a directory, it would `pop` the last home out and break out of the for loop?

jimmygchen: Or is this trying to ensure `tracing_log_path` is a directory, if not, then it pop the last component off? but it still doesn't guarantee the result is a directory. Feels like a more predictable behaviour is to just error out if the provided path is not a directory?

common/logging/src/lib.rs
```diff
@@ -1,24 +1,24 @@
...
+use tracing_subscriber::EnvFilter;
 
 pub const MAX_MESSAGE_WIDTH: usize = 40;
 
 pub mod macros;
 mod sse_logging_components;
+mod tracing_libp2p_discv5_logging_layer;
 pub mod tracing_logging_layer;
 mod tracing_metrics_layer;
+pub mod utils;
```
jimmygchen: This doesn't have to be `pub` I think

jimmygchen (commented):
Hey @macladson 

Thanks for the clean up and fixing up the issues! 🙏 

I've added some comments, let me know your thoughts.

With the default `debug-level` and `logfile-debug-level`, I noticed a large amount of trace level logs in the `discv5.log` and `libp2p.log` files. @ackintosh has noticed this and created a fix for this (see [his comment](https://github.com/sigp/lighthouse/pull/7168#issuecomment-2742722083) above). However, even debug level logs for these are quite huge, and libp2p and gossip log files were `INFO` level by default previously (different to `beacon.log`), because they're _very_ noisy, and we had to use  the `RUST_LOG` env var to enable deps debug logs. I think we might want to maintain this behaviour due to rate of file size growth and disk IO?

lighthouse/src/main.rs
```diff
@@ -619,21 +616,50 @@ fn run<E: EthSpec>(
...
+    }
+
+    if let Some(libp2p_discv5_layer) = libp2p_discv5_layer {
+        logging_layers.push(
+            libp2p_discv5_layer
+                .with_filter(
+                    EnvFilter::builder()
+                        .with_default_directive(Level::DEBUG.into())
+                        .from_env_lossy(),
+                )
```
macladson, line 650: I also noticed we lost the ability to control the log level with `RUST_LOG` so I've restored it here. It will now default to `debug` and then be overridden by `RUST_LOG`

macladson:
Thanks for the review, I'll fix these up soon!

> However, even debug level logs for these are quite huge, and libp2p and gossip log files were `INFO` level by default previously (different to `beacon.log`), because they're _very_ noisy, and we had to use the `RUST_LOG` env var to enable deps debug logs. I think we might want to maintain this behaviour due to rate of file size growth and disk IO?

I think the main thing to consider is what the most useful information in these logs is. It could be the case that `INFO` logs generally don't provide anything worthwhile for debugging and so we have to set `RUST_LOG=debug` manually after discovering some issue. Seems to me this friction could be avoided if we logged at `debug` in the first place. I'm not convinced the old behaviour was better but I'm also not really in a position to make this determination.

As an aside about file size growth and IO, my understanding is that database read/writes dwarf disk IO to the point where logging is pretty insignificant. And the files themselves are each limited to 10MB so there's no concern about unbounded file growth.

common/logging/src/tracing_libp2p_discv5_logging_layer.rs
macladson: I noticed this too, but decided to leave it for now since it wasn't related to any of the current issues and the dep logfiles seem to work at the moment. I might revisit it afterwards though

common/lighthouse_macros/Cargo.toml
macladson: Ended up changing the name anyway :sweat_smile:

common/logging/src/tracing_libp2p_discv5_logging_layer.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+impl<S> Layer<S> for Libp2pDiscv5TracingLayer
+where
+    S: Subscriber,
+{
+    fn on_event(&self, event: &tracing::Event<'_>, _ctx: Context<S>) {
+        let meta = event.metadata();
+        let log_level = meta.level();
+        let timestamp = Local::now().format("%Y-%m-%d %H:%M:%S").to_string();
```
jimmygchen, line 23: We currently use this approach `Local::now().format()` for formatting timestamp in 3 places in logging, and this `format` function seems relatively expensive - it adds up significantly given it gets called very frequently.

I did a brief profiling and it showed that **~11.44%** of total CPU time is spent in `chrono` formatting functions, if I calculated correctly:

```bash
$ grep chrono lighthouse_2025-03-31-182111.collapsed | wc -l
1426
$ wc -l lighthouse_2025-03-31-182111.collapsed
12468 lighthouse_2025-03-31-182111.collapsed
```

 Might be worth optimising a bit here, e.g. consider alternatives, e.g. [`StrftimeItems`](https://github.com/chronotope/chrono/blob/6a853015b7ccc09ae807e277a76d56de58d65139/src/format/strftime.rs#L174-L192) or the [`time`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/time/struct.UtcTime.html) crate.

jimmygchen: Happy to create a separate issue to address this, given this PR already fixed a bunch of things and would be good to merge soon.

lighthouse/src/main.rs
jimmygchen: Nice, thanks!

common/logging/src/tracing_libp2p_discv5_logging_layer.rs
jimmygchen: I made an attempt using `StrftimeItems` and it's showing much better results (~3.58% overhead) - although still quite high just for logging timestamps.
https://github.com/macladson/lighthouse/pull/6

macladson: I wonder how much of that impact is due to using `Utc::now` vs using `StrftimeItems`. It would be nice if we could keep local time timestamps. From memory I believe this has been a point of feedback from users in the past

jimmygchen: Oh yeah, I reckon most of the impact is from switching to `StrftimeItems`, `DateTime::format` seems to be the more expensive operation (The former is recommended in the`chrono` repo over `DateTime::format` for performance-critical applications). I tried using `Utc::now` to avoid timezone conversion to see how much improvements we can get.

IMO local timezone is useful for users (mostly consuming stdout rather than the debug files), and debug files is mostly for us - in that case I think UTC would be less confusing - but yeah don't have a strong opinion on timezone. I think it would be useful to have a summary of logging changes from tracing somewhere, e.g. default logging level change, timestamp logging, CLI flags etc

jimmygchen (approved):
Hey @macladson ,

Thanks for the updates. It's looking much cleaner now!

I'm keen to merge this in soon and happy to leave the remaining comments if you prefer to address them separately, since none of them were introduced from your PR?
- https://github.com/sigp/lighthouse/pull/7168#discussion_r2020445664
- https://github.com/sigp/lighthouse/pull/7168#discussion_r2020458289
- https://github.com/sigp/lighthouse/pull/7168#discussion_r2022092027

common/logging/src/tracing_libp2p_discv5_logging_layer.rs
jimmygchen: Created #7232 to address this.

jimmygchen:
@mergify dequeue

jimmygchen:
Waiting for https://github.com/sigp/lighthouse/pull/7235 to be merged to unstable to unblock CI.

jimmygchen:
@mergify queue

PR #7169, opened 2025-03-19
Remove `crit` as an option from the CLI entirely

ThreeHrSleep:
## Issue Addressed

https://github.com/sigp/lighthouse/issues/7165


macladson (approved):
LGTM!

macladson:
Just noting that this is _technically_ a breaking change since if a user has `--debug-level crit` or `--logfile-debug-level crit` it will now fail. I don't think many users will have this issue. If there are concerns we could instead silently alias `crit` to `error` to avoid it

PR #7174, opened 2025-03-20
Fix xdelta3 output buffer issue

michaelsproul:
## Issue Addressed

Attempt to fix an issue whereby `xdelta3` would try to guess the size of the output buffer during diff application. If the guess was incorrect, an error would be returned and the diff application would not be re-attempted.

This branch uses an exponential doubling of the output buffer in the case of an insufficient space error, meaning that diffs should never fail due to insufficient space.

We are waiting on confirmation from a node operator on Holesky that this actually fixes the issue. We will also sync a Holesky archive node on our infra to try to repro the issue.

## Additional Info

Requires this upstream PR:

- https://github.com/sigp/xdelta3-rs/pull/5

### DIFF
--- a/beacon_node/store/src/hdiff.rs
+++ b/beacon_node/store/src/hdiff.rs
@@ -21,8 +21,8 @@ static EMPTY_PUBKEY: LazyLock<PublicKeyBytes> = LazyLock::new(PublicKeyBytes::em
 pub enum Error {
     InvalidHierarchy,
     DiffDeletionsNotSupported,
-    UnableToComputeDiff,
-    UnableToApplyDiff,
+    UnableToComputeDiff(xdelta3::Error),
+    UnableToApplyDiff(xdelta3::Error),
     BalancesIncompleteChunk,
     Compression(std::io::Error),
     InvalidSszState(ssz::DecodeError),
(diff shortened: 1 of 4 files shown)


michaelsproul:
My suspicion is that this was triggered on Holesky at slot 3670016, as that is around when the operator started seeing the errors. That slot is a multiple of `2^18`, so it would have resulted in a diff being stored from the snapshot: `2097152 -> 3670016`.

I don't understand why this diff would be _so_ large though, as there were already 1.66M validators on Holesky at this time (https://light-holesky.beaconcha.in/epoch/65536). It might have been a bunch of the churn generated by the non-finality incident, but we would need to see that diff to understand more.

I'm working on syncing a Holesky archive node so we can have a closer look.

beacon_node/store/src/hdiff.rs
```diff
@@ -323,9 +323,15 @@ impl BytesDiff {
     }
 
     pub fn compute_xdelta(source_bytes: &[u8], target_bytes: &[u8]) -> Result<Self, Error> {
-        let bytes = xdelta3::encode(target_bytes, source_bytes)
-            .ok_or(Error::UnableToComputeDiff)
-            .unwrap();
+        // TODO: Use a smaller estimate for the output diff buffer size, currently the xdelta3 lib
```
dapplion: label `TODO(hdiff)`

beacon_node/store/src/hdiff.rs
```diff
@@ -334,8 +340,31 @@ impl BytesDiff {
     }
 
     pub fn apply_xdelta(&self, source: &[u8], target: &mut Vec<u8>) -> Result<(), Error> {
-        *target = xdelta3::decode(&self.bytes, source).ok_or(Error::UnableToApplyDiff)?;
-        Ok(())
+        // Dynamic buffer allocation. This is a stopgap until we implement a schema change to store
```
dapplion: Add `TODO(hdiff)`

dapplion (commented):
LGTM

PR #7175, opened 2025-03-20
Clarify network limits

### LINKED ISSUE #6811: Clarify gossip limits
Spec: https://github.com/ethereum/consensus-specs/pull/4045

Unifies `GOSSIP_MAX_SIZE` and `MAX_CHUNK_SIZE` config values while clarifying how these limits apply to various gossipsub fields

pawanjay176:
## Issue Addressed

Resolves #6811 

## Proposed Changes

Rename `GOSSIP_MAX_SIZE` to `MAX_PAYLOAD_SIZE` and remove `MAX_CHUNK_SIZE` in accordance with the spec.

The spec also "clarifies"  the message size limits at different levels. The rpc limits are equivalent to what we had before imo.
The gossip limits have additional checks.

I have gotten rid of the `is_bellatrix_enabled`  checks that used a lower limit (1mb) pre-merge. Since all networks we run start from the merge, I don't think this will break any setups.


consensus/types/src/chain_spec.rs
```diff
@@ -712,6 +711,33 @@ impl ChainSpec {
...
+    /// Max allowed size of a raw, compressed message received over the network.
+    ///
+    /// https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#max_compressed_len
+    #[allow(clippy::arithmetic_side_effects)]
+    pub fn max_message_size(&self) -> usize {
+        std::cmp::max(
+            // 1024 to account for framing + encoding overhead
+            Self::max_compressed_len_snappy(self.max_payload_size as usize) + 1024,
+            //1MB
+            1024 * 1024,
```
dapplion, line 739: Why max it at 1MB? Doesn't this void the 10MB limit for mainnet?

consensus/types/src/chain_spec.rs
```diff
@@ -712,6 +711,33 @@ impl ChainSpec {
...
     }
 
+    /// Worst-case compressed length for a given payload of size n when using snappy.
+    ///
+    /// https://github.com/google/snappy/blob/32ded457c0b1fe78ceb8397632c416568d6714a0/snappy.cc#L218C1-L218C47
+    /// https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#max_compressed_len
+    #[allow(clippy::arithmetic_side_effects)]
+    fn max_compressed_len_snappy(n: usize) -> usize {
+        32 + n + (n / 6)
+    }
```
jxs: shouldn't we add guard rails here to not overflow? 
I.e. 
```suggestion
    fn max_compressed_len_snappy(n: usize) -> Option<usize> {
        32_usize.checked_add(n)?.checked_add(n/6)
    }
```

and then `unwrap_or_()` or `expect` with a message when using it

consensus/types/src/chain_spec.rs
```diff
@@ -712,6 +711,33 @@ impl ChainSpec {
...
+
+    /// Max allowed size of a raw, compressed message received over the network.
+    ///
+    /// https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#max_compressed_len
+    #[allow(clippy::arithmetic_side_effects)]
+    pub fn max_message_size(&self) -> usize {
+        std::cmp::max(
+            // 1024 to account for framing + encoding overhead
+            Self::max_compressed_len_snappy(self.max_payload_size as usize) + 1024,
```
jxs: same as above regarding guard rails

jxs (commented):
Hi, Pawan, overall looks good! Left two comments

consensus/types/src/chain_spec.rs
pawanjay176: Just following the spec here. https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#max_message_size

consensus/types/src/chain_spec.rs
pawanjay176: Yeah sorry was being lazy. Made it safe arithematic and added a test to make sure default values don't overflow

pawanjay176:
@michaelsproul marking this as v7.0.0 as I think we should have this before the mainnet release

consensus/types/src/chain_spec.rs
```diff
@@ -712,6 +711,35 @@ impl ChainSpec {
         }
     }
 
+    /// Worst-case compressed length for a given payload of size n when using snappy.
+    ///
+    /// https://github.com/google/snappy/blob/32ded457c0b1fe78ceb8397632c416568d6714a0/snappy.cc#L218C1-L218C47
+    /// https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#max_compressed_len
+    fn max_compressed_len_snappy(n: usize) -> Option<usize> {
+        32_usize.checked_add(n)?.checked_add(n / 6)
```
AgeManning, line 719: Rather than panic'ing (we have expects following this), why not just return usize::MAX if it were to overflow?

AgeManning (approved):
Looks like a lot of renaming, had a single comment. Otherwise lgtm

consensus/types/src/chain_spec.rs
pawanjay176: I think silently failing returning a higher max value might not be safe here and could potentially lead to consensus issues.
I have added a test at the end of this file to check that there are no panics based on default mainnet values.

PR #7176, opened 2025-03-20
Fix test OOM issues on tree-states-hot

michaelsproul:
WIP branch to debug test OOMs

### DIFF
--- a/beacon_node/beacon_chain/src/schema_change/migration_schema_v24.rs
+++ b/beacon_node/beacon_chain/src/schema_change/migration_schema_v24.rs
@@ -286,7 +286,7 @@ pub fn downgrade_from_v24<T: BeaconChainTypes>(
                 .load_hot_state(&state_root)?
                 .ok_or(Error::MissingState(state_root))?;
 
-            // Immediatelly commit the state. Otherwise we will OOM and it's stored in a different
+            // Immediately commit the state. Otherwise we will OOM and it's stored in a different
             // column. So if the migration crashes we just get extra harmless junk in the DB.
             let mut state_write_ops = vec![];
             store_full_state_v22(&state_root, &state, &mut state_write_ops)?;
--- a/beacon_node/beacon_chain/src/schema_change/migration_schema_v24.rs
+++ b/beacon_node/beacon_chain/src/schema_change/migration_schema_v24.rs
@@ -1,16 +1,17 @@
 use crate::{
     beacon_chain::BeaconChainTypes,
-    summaries_dag::{DAGStateSummaryV22, StateSummariesDAG},
+    summaries_dag::{DAGStateSummary, DAGStateSummaryV22, StateSummariesDAG},
 };
-use ssz::Decode;
+use ssz::{Decode, Encode};
 use ssz_derive::{Decode, Encode};
 use std::{
     sync::Arc,
     time::{Duration, Instant},
 };
 use store::{
-    get_full_state_v22, hdiff::StorageStrategy, hot_cold_store::DiffBaseStateRoot, DBColumn, Error,
-    HotColdDB, HotStateSummary, KeyValueStore, KeyValueStoreOp, StoreItem,
+    get_full_state_v22, hdiff::StorageStrategy, hot_cold_store::DiffBaseStateRoot,
+    store_full_state_v22, DBColumn, Error, HotColdDB, HotStateSummary, KeyValueStore,
+    KeyValueStoreOp, StoreItem,
 };
 use tracing::{debug, info, warn};
 use types::{EthSpec, Hash256, Slot};
@@ -94,8 +95,8 @@ pub fn upgrade_to_v24<T: BeaconChainTypes>(
         slots_count = summaries_by_slot.len(),
         min_slot = ?summaries_by_slot.first_key_value().map(|(slot, _)| slot),
         max_slot = ?summaries_by_slot.last_key_value().map(|(slot, _)| slot),
-        state_summaries_dag_roots = ?state_summaries_dag_roots,
-        hot_hdiff_start_slot = %hot_hdiff_start_slot,
+        ?state_summaries_dag_roots,
+        %hot_hdiff_start_slot,
         split_state_root = ?split.state_root,
         "Starting hot states migration"
     );
@@ -178,21 +179,22 @@ pub fn upgrade_to_v24<T: BeaconChainTypes>(
                                 })?
                         };
 
-                        let diff_base_state_root =
-                            if let Some(diff_base_slot) = storage_strategy.diff_base_slot() {
-                                DiffBaseStateRoot::new(
+                        let diff_base_state_root = if let Some(diff_base_slot) =
+                            storage_strategy.diff_base_slot()
+                        {
+                            DiffBaseStateRoot::new(
                                     diff_base_slot,
                                     state_summaries_dag
                                         .ancestor_state_root_at_slot(state_root, diff_base_slot)
                                         .map_err(|e| {
                                             Error::MigrationError(format!(
-                                                "error computing ancestor_state_root_at_slot {e:?}"
+                                                "error computing ancestor_state_root_at_slot({state_root:?}, {diff_base_slot}) {e:?}"
                                             ))
                                         })?,
                                 )
-                            } else {
-                                DiffBaseStateRoot::zero()
-                            };
+                        } else {
+                            DiffBaseStateRoot::zero()
+                        };
 
                         let new_summary = HotStateSummary {
                             slot,
@@ -227,8 +229,8 @@ pub fn upgrade_to_v24<T: BeaconChainTypes>(
             if last_log_time.elapsed() > Duration::from_secs(5) {
                 last_log_time = Instant::now();
                 info!(
-                    diffs_written = diffs_written,
-                    summaries_written = summaries_written,
+                    diffs_written,
+                    summaries_written,
                     summaries_count = state_summaries_dag.summaries_count(),
                     "Hot states migration in progress"
                 );
@@ -239,8 +241,8 @@ pub fn upgrade_to_v24<T: BeaconChainTypes>(
     // TODO(hdiff): Should run hot DB compaction after deleting potentially a lot of states. Or should wait
     // for the next finality event?
     info!(
-        diffs_written = diffs_written,
-        summaries_written = summaries_written,
+        diffs_written,
+        summaries_written,
         summaries_count = state_summaries_dag.summaries_count(),
         "Hot states migration complete"
     );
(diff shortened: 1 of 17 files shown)


michaelsproul:
The `Heaptrack` commit should probably be ripped out and pushed to `unstable` separately.

michaelsproul:
The current test failure is:

```
---- beacon_block_streamer::tests::check_all_blocks_from_altair_to_fulu stdout ----

thread 'beacon_block_streamer::tests::check_all_blocks_from_altair_to_fulu' panicked at beacon_node/beacon_chain/src/test_utils.rs:2712:14:
called `Result::unwrap()` on an `Err` value: BeaconChainError(DBError(HotColdDBError(HdiffGetPriorStateRootError(Slot(96), Slot(0)))))
```

dapplion:
@michaelsproul there's a key design decision when using the summaries for iteration.

https://github.com/sigp/lighthouse/blob/2b050f4e8992a38af0fef116d0dde30f97241b9e/beacon_node/store/src/hot_cold_store.rs#L3408-L3414

What slot do we start from?

**state_slot - 1**

 If we are attempting to store `state A` to the store, are we guaranteed that the summary of its parent is already persisted in disk? Could it be it will be written eventually but it's currently sitting in an ops vector? What about the migration path?

**state_slot - SLOTS_PER_HISTORICAL_VECTOR**

This _feels_ safer due to the big distance, but it can experience the same issue as above. Also, we could have a situation where we are attempting to get the ancestor root of an HDiff with slot < finalized_slot. And `state_slot - SLOTS_PER_HISTORICAL_VECTOR` is also < finalized_slot. In that case we can get a missing state summary as it has already been pruned as part of the migration.

What I'm setting for now is:

```rust
let initial_state_slot = max(state.slot() - SLOTS_PER_HISTORICAL_VECTOR, store.split_slot);
```

What do you think?

dapplion:
Test `weak_subjectivity_sync_unaligned_unadvanced_checkpoint` fails because the forwards block root iterator of both nodes don't match
Screenshot 2025-03-20 at 15 01 54

- left = forwards: `(0x6b6eef0cc6be9c55ca95d3bff66f1a78aa2aee8d2c73e3a0f958992a57999cfb, Slot(69))`
- right = expected: `(0xbf7bc306723de95400be0d8f4eccb9965c4b5ea5c10752d1ad84c94693c6f4a7, Slot(69))` 

The test `weak_subjectivity_sync_unaligned_advanced_checkpoint` has the exact same assertion error

@michaelsproul this tests look strange to me, questions
-  `weak_subjectivity_sync_unaligned_unadvanced_checkpoint` uses as checkpoint a slot that's not aligned, do we support this case?
- If we checkpoint sync into slot 72, where in the test are historical blocks fetched? I see this log in the consumer node `beacon_chain::historical_blocks: Ignoring some historic blocks oldest_block_slot=0 total_blocks=68 ignored=68`
- Does the consumer node have backfill and range sync enabled?

EDIT: The right, expected is correct. This is the chain of blocks built

- 68 0x6b6eef0cc6be9c55ca95d3bff66f1a78aa2aee8d2c73e3a0f958992a57999cfb
- 69 0xbf7bc306723de95400be0d8f4eccb9965c4b5ea5c10752d1ad84c94693c6f4a7 <- checkpoint block
- 70 skipped
- 71 skipped
- 72 skipped
- 73 0xde74591ddcd9fe0020fccffd9e13df0e3edc408af23e76db1c52a323cc1936ca

Somehow left thinks that block `0xbf7bc306723de95400be0d8f4eccb9965c4b5ea5c10752d1ad84c94693c6f4a7` is for slot 72 in the iterator.
- The forced import historical blocks imports up to 68 included

**Logs of the source checkpoint sync server node**

```
2025-03-20T18:46:48.977158Z DEBUG store::hot_cold_store: Storing hot state summary and diffs state_root=0x305fb468655195467b0a9a763714df9cb910c5a0d3a5bbf4009928fe21c74f21 slot=68 storage_strategy=ReplayFrom(Slot(64)) diff_base_state_root=64/0
x3ce62a358919b37aff4691b697b433e882f4037d88b905bd23606cddcc38d553 previous_state_root=0xc4f4a9e0266f78efc1b46ef06a8dae69131ccd9d07c46af392d625fb92aa87c5
2025-03-20T18:46:48.977204Z DEBUG beacon_chain::beacon_chain: Beacon block imported block_root=0x6b6eef0cc6be9c55ca95d3bff66f1a78aa2aee8d2c73e3a0f958992a57999cfb block_slot=68 source=range_sync

2025-03-20T18:46:48.987348Z DEBUG store::hot_cold_store: Storing hot state summary and diffs state_root=0x4698679e674a2d6e125b0e7c3eabbd9a3434213e79acabae07de52a8946d68ea slot=69 storage_strategy=ReplayFrom(Slot(64)) diff_base_state_root=64/0
x3ce62a358919b37aff4691b697b433e882f4037d88b905bd23606cddcc38d553 previous_state_root=0x305fb468655195467b0a9a763714df9cb910c5a0d3a5bbf4009928fe21c74f21
2025-03-20T18:27:21.284430Z DEBUG beacon_chain::beacon_chain: Beacon block imported block_root=0xbf7bc306723de95400be0d8f4eccb9965c4b5ea5c10752d1ad84c94693c6f4a7 block_slot=69 source=range_sync

2025-03-20T18:46:48.997590Z DEBUG store::hot_cold_store: Storing hot state summary and diffs state_root=0x14500ce15bc46fe39594aec134e6323aee2a012dc961f67cfd95e83daa71014c slot=70 storage_strategy=ReplayFrom(Slot(64)) diff_base_state_root=64/0
x3ce62a358919b37aff4691b697b433e882f4037d88b905bd23606cddcc38d553 previous_state_root=0x4698679e674a2d6e125b0e7c3eabbd9a3434213e79acabae07de52a8946d68ea
2025-03-20T18:46:48.997656Z DEBUG store::hot_cold_store: Storing hot state summary and diffs state_root=0xfee996191ecd94ee2b0e0d0bbd22b4784982b294a8fa7ae1fd5b8c9766277ef9 slot=71 storage_strategy=ReplayFrom(Slot(64)) diff_base_state_root=64/0
x3ce62a358919b37aff4691b697b433e882f4037d88b905bd23606cddcc38d553 previous_state_root=0x14500ce15bc46fe39594aec134e6323aee2a012dc961f67cfd95e83daa71014c
2025-03-20T18:46:48.997790Z DEBUG store::hot_cold_store: Storing hot state summary and diffs state_root=0x92c13c4dce9c76f9362cc679a6d08e8967d56da6ee2f6780f76522cda2c84223 slot=72 storage_strategy=ReplayFrom(Slot(64)) diff_base_state_root=64/0
x3ce62a358919b37aff4691b697b433e882f4037d88b905bd23606cddcc38d553 previous_state_root=0xfee996191ecd94ee2b0e0d0bbd22b4784982b294a8fa7ae1fd5b8c9766277ef9
2025-03-20T18:46:48.997987Z DEBUG store::hot_cold_store: Storing hot state summary and diffs state_root=0x5669b2d3a9e8127703814aa438ccd5bb9d4a8d6e3b1e2e288cc48c9bd6d94a21 slot=73 storage_strategy=ReplayFrom(Slot(64)) diff_base_state_root=64/0
x3ce62a358919b37aff4691b697b433e882f4037d88b905bd23606cddcc38d553 previous_state_root=0x92c13c4dce9c76f9362cc679a6d08e8967d56da6ee2f6780f76522cda2c84223

2025-03-20T18:27:21.291515Z DEBUG beacon_chain::beacon_chain: Beacon block imported block_root=0xde74591ddcd9fe0020fccffd9e13df0e3edc408af23e76db1c52a323cc1936ca block_slot=73 source=range_sync

2025-03-20T18:46:49.008362Z DEBUG store::hot_cold_store: Storing hot state summary and diffs state_root=0x8ddf4321d5c3e8b887e99ac27e3da5306cadfd16becc97f6eafa1423ac457568 slot=74 storage_strategy=ReplayFrom(Slot(64)) diff_base_state_root=64/0
x3ce62a358919b37aff4691b697b433e882f4037d88b905bd23606cddcc38d553 previous_state_root=0x5669b2d3a9e8127703814aa438ccd5bb9d4a8d6e3b1e2e288cc48c9bd6d94a21
2025-03-20T18:46:49.008385Z DEBUG beacon_chain::beacon_chain: Beacon block imported block_root=0x3c6a5a9728a05e1a2fc324a41ba24e421a5ad2f57aca0794c318adf3bd4bd2ab block_slot=74 source=range_sync
```

**Logs of the consumer node**

```
2025-03-20T18:46:49.387830Z DEBUG beacon_chain::builder: Storing split from weak subjectivity state slot=72 state_root=0x92c13c4dce9c76f9362cc679a6d08e8967d56da6ee2f6780f76522cda2c84223 block_root=0xbf7bc306723de95400be0d8f4eccb9965c4b5ea5c10
752d1ad84c94693c6f4a7
2025-03-20T18:46:49.388062Z DEBUG store::hot_cold_store: Storing hot state summary and diffs state_root=0x92c13c4dce9c76f9362cc679a6d08e8967d56da6ee2f6780f76522cda2c84223 slot=72 storage_strategy=Snapshot diff_base_state_root=0/0x000000000000
0000000000000000000000000000000000000000000000000000 previous_state_root=0xfee996191ecd94ee2b0e0d0bbd22b4784982b294a8fa7ae1fd5b8c9766277ef9
2025-03-20T18:46:49.389877Z  INFO beacon_chain::builder: Beacon chain initialized head_state=0x4698679e674a2d6e125b0e7c3eabbd9a3434213e79acabae07de52a8946d68ea head_block=0xbf7bc306723de95400be0d8f4eccb9965c4b5ea5c10752d1ad84c94693c6f4a7 head
_slot=69
```

^ It has a head at slot 69 for block `0xbf7bc306723de95400be0d8f4eccb9965c4b5ea5c10752d1ad84c94693c6f4a7`

michaelsproul:
> weak_subjectivity_sync_unaligned_unadvanced_checkpoint uses as checkpoint a slot that's not aligned, do we support this case?

Yeah we do. The test should pass as-is. To me this failure looks like a bug in our changes rather than an issue with the test.

> If we checkpoint sync into slot 72, where in the test are historical blocks fetched?

Right here: https://github.com/sigp/lighthouse/blob/de51d148771304e6ed21e3cd79b6a3c3a12b1c4e/beacon_node/beacon_chain/tests/store_tests.rs#L2449-L2476

https://github.com/sigp/lighthouse/blob/de51d148771304e6ed21e3cd79b6a3c3a12b1c4e/beacon_node/beacon_chain/tests/store_tests.rs#L2499-L2501

> Does the consumer node have backfill and range sync enabled?

Not exactly, we are manually feeding it blocks backwards and forwards to simulate backfill & range sync.

Probably just an off by one in one of the new iterators created?

michaelsproul:
For the backwards state root iterator, I think we might need to support fetching state summaries prior to the split, right? Because we allow diffs to exist in that region (prior to the split) if they are required to construct the split state itself.

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -2454,7 +2455,7 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
         } else {
             AnchorInfo {
                 anchor_slot,
-                oldest_block_slot: anchor_slot,
+                oldest_block_slot,
```
dapplion, line 2458: Quite a silly bug 🙄 probably I wrote that line before I understood the store well

beacon_node/store/src/reconstruct.rs
```diff
@@ -145,10 +144,8 @@ where
                             });
                         }
 
-                        self.compare_and_set_anchor_info_with_write(
-                            old_anchor,
-                            ANCHOR_FOR_ARCHIVE_NODE,
```
dapplion: Setting this const here innadvertedly mutated the anchor_slot, which would have broken HDiff reads

dapplion: We should rename `anchor_slot` and carefully review where it is mutated or set

dapplion: WIP, I will merge this branch first once CI is green then we can do the rename
https://github.com/sigp/lighthouse/compare/tree-states-hot-rebase-oom...dapplion:lighthouse:tree-states-hot-anchor-slot-rename?expand=1

beacon_node/beacon_chain/src/schema_change/migration_schema_v24.rs
```diff
@@ -249,10 +251,100 @@ pub fn upgrade_to_v24<T: BeaconChainTypes>(
...
+            states_written += 1;
+        }
+
+        // Persist old summary
+        let epoch_boundary_state_slot = summary.slot - summary.slot % T::EthSpec::slots_per_epoch();
+        let old_summary = HotStateSummaryV22 {
+            slot: summary.slot,
+            latest_block_root: summary.latest_block_root,
+            epoch_boundary_state_root: state_summaries_dag
+                .ancestor_state_root_at_slot(state_root, epoch_boundary_state_slot)
```
dapplion, line 303: Does this handle the genesis well?

beacon_node/beacon_chain/src/schema_change/migration_schema_v24.rs
```diff
@@ -249,10 +251,100 @@ pub fn upgrade_to_v24<T: BeaconChainTypes>(
...
+        .into_iter()
+        .flat_map(|(_, summaries)| summaries)
+    {
+        // If boundary state persist
+        if summary.slot % T::EthSpec::slots_per_epoch() == 0 {
+            let (state, _) = db
+                .load_hot_state(&state_root)?
+                .ok_or(Error::MissingState(state_root))?;
+
+            // Immediatelly commit the state. Otherwise we will OOM and it's stored in a different
```
michaelsproul:
```suggestion
            // Immediately commit the state. Otherwise we will OOM and it's stored in a different
```

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -1485,14 +1485,16 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
+        let summary = self.store_hot_state_summary(state_root, state, ops)?;
         self.store_hot_state_diffs(state_root, state, ops)?;
 
         debug!(
             ?state_root,
             slot = %state.slot(),
             storage_strategy = ?self.hot_storage_strategy(state.slot())?,
-            "Stored hot state summary and diffs"
+            diff_base_state_root = %summary.diff_base_state_root,
+            previous_state_root = ?summary.previous_state_root,
```
```suggestion
            diff_base_state_root = ?summary.diff_base_state_root,
            previous_state_root = ?summary.previous_state_root,
```

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -3361,20 +3378,106 @@ fn no_state_root_iter() -> Option<std::iter::Empty<Result<(Hash256, Slot), Error
     None
 }
 
+#[derive(Debug)]
+#[allow(dead_code)]
```
michaelsproul, line 3382: Could move this to specific fields that are only used by debug

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -3361,20 +3378,106 @@ fn no_state_root_iter() -> Option<std::iter::Empty<Result<(Hash256, Slot), Error
...
+            .map_err(|_| StateSummaryIteratorError::OutOfBoundsInitialSlot)?
+    };
+
+    let mut previous_slot = None;
+
+    loop {
+        let state_summary = store
+            .load_hot_state_summary(&state_root)
+            .map_err(|e| StateSummaryIteratorError::LoadSummaryError(Box::new(e)))?
+            .ok_or(StateSummaryIteratorError::MissingSummary(state_root))?;
```
michaelsproul, line 3450: This could fail if `state_root` is pruned between when we took the `split` lock. We should probably hold the split lock from the start of this function (after the lookup in the state misses). This should be OK for performance because loading summaries is cheap and we're never iterating that far back.

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -3361,20 +3378,106 @@ fn no_state_root_iter() -> Option<std::iter::Empty<Result<(Hash256, Slot), Error
...
+            Ordering::Greater => {} // keep going
+        }
+
+        // Jump to an older state summary that is ancestor of `state_root`
+        if target_slot <= state_summary.diff_base_state_root.slot {
+            // As an optimization use the HDiff state root to jump states faster
+            state_root = state_summary.diff_base_state_root.state_root;
+        } else {
+            // Else jump slot by slot
+            state_root = state_summary.previous_state_root;
```
michaelsproul, line 3478: This could be 0x0 in the case of the split state itself due to the initialisation in the migration here:

https://github.com/sigp/lighthouse/blob/06b19e33815c55ca3bdf44893fd7105a3c2f55c9/beacon_node/beacon_chain/src/schema_change/migration_schema_v24.rs#L170-L180

I think this could be a problem when we need to get a diff base state root prior to the split?

michaelsproul: I think actually that `Hash256::ZERO` is unreachable: https://github.com/sigp/lighthouse/pull/6750/files#r2025919484

beacon_node/store/src/reconstruct.rs
michaelsproul: Not doing this for now, seeing as anchor_slot seems to be working :crossed_fingers:

beacon_node/store/src/hot_cold_store.rs
michaelsproul: Fixed in:

- https://github.com/sigp/lighthouse/pull/6750/commits/088cfa03deec97ca56eb6bea4adaae35a2fa37a6
- https://github.com/sigp/lighthouse/pull/6750/commits/1184d7237e7c3cef031d208a763a5b3b076df23b

PR #7136, opened 2025-03-14
availability when fetching sync committees

Ayushdubey86:
## Issue Addressed

Which issue # does this PR address?
This is dummy pr for this issue #7115 

## Proposed Changes

Added logic to consider `state_upper_limit` when verifying full sync aggregates.
Modified `verify_full_sync_aggregates_up_to` to use `state_upper_limit` from `AnchorInfo`.
Introduced `HotColdDB` dependency to fetch `AnchorInfo`.

## Additional Info

PR is not final yet, but feel free to review it and give some pointers!

Issue #7115, opened 2025-03-12
Account for state availability when fetching sync committees


michaelsproul:
Short placeholder issue (will add more later):

Lighthouse currently loads the state from the start of the sync committee period when sync duties are requested. This can cause confusion when that state is not available, but _a state_ from the same period _is_ available. We could use the `state_upper_limit` to bound the slot of the state fetched, something like: `max(sync_committee_start_slot, state_upper_limit)` (if `state_upper_limit != INT_MAX`).

Ayushdubey86:
can i look into this?

michaelsproul:
@Ayushdubey86 sure!

Ayushdubey86:
Hey @michaelsproul , can you check out my pr if possible!

PR #7178, opened 2025-03-20
Account for state availability when fetching sync committees

Ayushdubey86:
Hello @michaelsproul , do check this pr out once free!

## Issue Addressed

Issue -> #7115 

## Proposed Changes
 This pr is a continuation of this [pr](https://github.com/sigp/lighthouse/pull/7136)


beacon_node/http_api/src/sync_committees.rs
```diff
@@ -108,11 +108,13 @@ fn duties_from_state_load<T: BeaconChainTypes>(
         // have to transition the head to start of the current period).
         //
         // We also need to ensure that the load slot is after the Altair fork.
+        let anchor_info = chain.store.get_anchor_info();
+        let state_upper_limit = anchor_info.state_upper_limit;
```
michaelsproul, line 112: We should account for the case where `state_upper_limit` is equal to `STATE_UPPER_LIMIT_NO_RETAIN` which we use to represent the configuration where _no_ historic states are stored. The "no retain" value is very high (`2^64 - 1`), so if we naively `max` it in here then the `load_slot` will always be `2^64 - 1`, which won't work.

beacon_node/http_api/src/sync_committees.rs
```diff
@@ -108,11 +108,13 @@ fn duties_from_state_load<T: BeaconChainTypes>(
...
         let load_slot = max(
-            chain.spec.epochs_per_sync_committee_period * sync_committee_period.saturating_sub(1),
-            altair_fork_epoch,
-        )
-        .start_slot(T::EthSpec::slots_per_epoch());
+            max(chain.spec.epochs_per_sync_committee_period * sync_committee_period.saturating_sub(1), 
+                altair_fork_epoch),
+            state_upper_limit
+        ).start_slot(T::EthSpec::slots_per_epoch());
```
michaelsproul: We don't want to call `start_slot` on `state_upper_limit`, because it is _already_ a `Slot`. The other values are epochs.

testing/simulator/src/checks.rs
```diff
@@ -172,15 +173,20 @@ pub async fn verify_fork_version<E: EthSpec>(
...
     network: LocalNetwork<E>,
+    db: Arc<HotColdDB<E, LevelDB<E>, LevelDB<E>>>,
     sync_committee_start_slot: Slot,
     upto_slot: Slot,
     slot_duration: Duration,
 ) -> Result<(), String> {
     slot_delay(upto_slot, slot_duration).await;
+    let anchor_info = db.get_anchor_info();
+    let state_upper_limit = anchor_info.state_upper_limit;
+    let effective_start_slot = std::cmp::max(sync_committee_start_slot, state_upper_limit);
```
michaelsproul: I think we can probably just revert all the changes in this file. These tests run with historic states enabled, so it doesn't make any difference.

If you would like to add a test, have a look at `beacon_node/beacon_chain/tests/store_tests.rs`. We could add a test that checkpoint syncs a node from some state, and then advances the chain and queries a historic sync committee which starts prior to the checkpoint. We would likely want to use the `MinimalEthSpec` so that sync committees aren't too long.

michaelsproul (changes requested):

michaelsproul:
Please rebase on `unstable` as well and remove the unrelated changes from `stable`.

Ayushdubey86:
> Please rebase on `unstable` as well and remove the unrelated changes from `stable`.

done

testing/simulator/src/checks.rs
Ayushdubey86: have removed it , can you run the pipeline once, in my recent local setup , i am unable to see tc passing/failing

beacon_node/http_api/src/sync_committees.rs
Ayushdubey86: done

beacon_node/http_api/src/sync_committees.rs
Ayushdubey86: done , do check it out once free!

macladson (commented):
Hi @Ayushdubey86, the PR seems to have commits from other authors related to the mergify config. These changes are only on the stable branch and not present in unstable. Since we need you to target unstable, these changes will need to be removed. I would recommend doing an interactive rebase (git rebase -i unstable) and removing those specific commits. Thanks!

PR #7182, opened 2025-03-20
Ensure sqlite and rusqlite are optional in `consensus/types`

ruuda:
## Issue Addressed

Even though the `consensus/types` crate has a feature named `sqlite`, it unconditionally depends on the `rusqlite` crate, which then depends on the `sqlite` crate — even when the feature is disabled. When the feature is disabled, the code that imports from `rusqlite` is disabled, so this dependency is not needed when the feature is disabled.

This is not a problem for Lighthouse itself, but I’m interested in using the types defined here in a different Rust project, which depends on a conflicting version of the `sqlite` crate.

## Proposed Changes

Ensure that the dependency on `rusqlite` is only present when the `sqlite` feature is enabled.

### DIFF
--- a/consensus/types/Cargo.toml
+++ b/consensus/types/Cargo.toml
@@ -38,7 +38,7 @@ rand_xorshift = "0.3.0"
 rayon = { workspace = true }
 regex = { workspace = true }
 rpds = { workspace = true }
-rusqlite = { workspace = true }
+rusqlite = { workspace = true, optional = true }
 safe_arith = { workspace = true }
 serde = { workspace = true, features = ["rc"] }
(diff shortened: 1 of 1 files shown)


michaelsproul (approved):
Nice improvement, ty

michaelsproul:
Hey @ruuda, you'll need to rebase on `unstable` again or push a new commit in order to re-trigger the target-branch-check. Unfortunately re-running it doesn't work

ruuda:
Ah, thanks, I was wondering why it complained about the base branch. Rebased it now.

michaelsproul (approved):

PR #7188, opened 2025-03-21
Fix the `getBlobs` metric and ensure it is recorded promptly to prevent miscounts

jimmygchen:
## Issue Addressed

From testing conducted by Sunnyside Labs, they noticed that the "expected blobs" are quite low on bandwidth constrained nodes. This observation revealed that we don't record the `beacon_blobs_from_el_expected_total` metric at all if the EL doesn't return any response. The fetch blobs function returns without recording the metric.

To fix this, I've moved `BLOBS_FROM_EL_EXPECTED_TOTAL` and `BLOBS_FROM_EL_RECEIVED_TOTAL` to as early as possible, to make the metric more accurate.


beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -73,13 +73,23 @@ pub async fn fetch_and_process_engine_blobs<T: BeaconChainTypes>(
         .as_ref()
         .ok_or(FetchEngineBlobError::ExecutionLayerMissing)?;
 
+    inc_counter_by(
+        &metrics::BLOBS_FROM_EL_EXPECTED_TOTAL,
+        num_expected_blobs as u64,
+    );
```
dapplion: You could also change to a histogram here. Still has a total but it's more granular. Plus add a counter for `FetchEngineBlobError::RequestFailed`. Then we can use `ERROR / (BLOBS_FROM_EL_EXPECTED_TOTAL + ERROR)` for the error rate. And know how often we request blobs for different blob counts.

jimmygchen: Updated, please re-review, thanks!

beacon_node/beacon_chain/src/metrics.rs
```diff
@@ -1662,28 +1662,37 @@ pub static DATA_COLUMN_SIDECAR_GOSSIP_VERIFICATION_TIMES: LazyLock<Result<Histog
...
+        "beacon_blobs_from_el_error_total",
+        "Number of failed blob fetches from the execution layer",
+    )
+});
+
+pub static BLOBS_FROM_EL_EXPECTED: LazyLock<Result<Histogram>> = LazyLock::new(|| {
+    try_create_histogram_with_buckets(
+        "beacon_blobs_from_el_expected",
         "Number of blobs expected from the execution layer",
+        linear_buckets(0.0, 4.0, 20),
```
dapplion: This buckets should have better resolution, maybe something like `[0,3,6,9,12,18,24,30]`

jimmygchen: The linear buckets has better granularity from 12 blobs than the one you defined.
Do you mean better resolution at lower blob counts?

pawanjay176 (approved):
LGTM sans lion's comment

beacon_node/beacon_chain/src/metrics.rs
jimmygchen: No strong opinion on this, updated in https://github.com/sigp/lighthouse/pull/7188/commits/6a788db663a4e00834ac654dcb49bf40e075874e

Issue #6296, opened 2024-08-22
Deterministic RNG in peer DAS publish block tests


michaelsproul:
## Description

It would be good to make the tests deterministic by using a seeded RNG here:

https://github.com/sigp/lighthouse/pull/5683#discussion_r1726333773

https://github.com/sigp/lighthouse/blob/ea331a97081b473d2db2828059b1df2450b48b13/beacon_node/http_api/src/publish_blocks.rs#L128

SunnysidedJ:
Hi, I'm a new joinee from Sunnyside Labs and I'd like to work on it. Thanks!

jimmygchen:
Completed in #7192, thanks @SunnysidedJ !

PR #7192, opened 2025-03-21
Fix for #6296: Deterministic RNG in peer DAS publish block tests

SunnysidedJ:
## Issue Addressed

#6296: Deterministic RNG in peer DAS publish block tests

## Proposed Changes

Made test functions to call publish-block APIs with true for the deterministic RNG boolean parameter while production code with false. This will deterministically shuffle columns for unit tests under broadcast_validation_tests.rs.

## Additional Info

1. Tested with "cargo test --workspace --release --features "$(TEST_FEATURES)" --exclude ef_tests --exclude beacon_chain --exclude slasher --exclude network --exclude web3signer_tests" which is same as "make test" but excluding web3signer_tests. The suggestion from the Lighthouse Book, update JDK/JRE to the latest (17 or above and I used 24), did not work. Please let me know if this is okay.
2. Tests related to the issue (bn_http_api_tests) was disabled by default. I manually enabled to run tests calling related API changes. See beacon_node/http_api/tests/main.rs. Many functions here result in stack overflow for some reason on my local machine.
3. Please clarify intention and scope of changes. I merely took the word-by-word on the issue #6296 by changing existing test code to take a fixed seed. Do you want to make it more customisable or use different design? Also, it'd be appreciated to know the intention of changing it to something deterministic. How does it affect testability, and what kinds of testing can we do further?


beacon_node/http_api/src/lib.rs
```diff
@@ -1260,6 +1260,7 @@ pub fn serve<T: BeaconChainTypes>(
                         BroadcastValidation::default(),
                         duplicate_block_status_code,
                         network_globals,
+                        false,
```
michaelsproul: I don't really _love_ adding a random bool parameter here. I wonder if a cleaner approach would be to add an `rng: Arc<RwLock<Box<dyn RngCore>>>` field to `BeaconChain` which is initialised either as the `thread_rng` or a seeded RNG depending on config? The `BeaconChainHarness` could initialise it in seeded mode, while the `BeaconChainBuilder` used in production could initialise it as `thread_rng`.

What do you think @SunnysidedJ?

SunnysidedJ: Thank you for making this comment with detailed pointers. I was also wondering a good place to pass RNG instances instead, but I wasn't able to due to my lack of knowledge in the codebase and Rust (I'm a beginner). One thing that I couldn't use was ThreadRng instance for production code because it is a thread-local object which cannot be passed between threads. Would a slight degrading of performance due to the use of OsRng be a concern for you? The logic otherwise seems the same according to its documentation. Nonetheless, I hope the new change is something closer to what you expect!

beacon_node/beacon_chain/src/builder.rs
```diff
@@ -106,6 +108,7 @@ pub struct BeaconChainBuilder<T: BeaconChainTypes> {
     task_executor: Option<TaskExecutor>,
     validator_monitor_config: Option<ValidatorMonitorConfig>,
     import_all_data_columns: bool,
+    rng: Option<Arc<RwLock<Box<dyn RngCore + Sync + Send>>>>,
```
jimmygchen: Could probably just be `Option<dyn RngCore>` here, as the builder doesn't need to be `Arc<RwLock<..`, and then instantiate the `Arc<RwLock` when actually building the chain.

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -494,6 +495,8 @@ pub struct BeaconChain<T: BeaconChainTypes> {
     pub data_availability_checker: Arc<DataAvailabilityChecker<T>>,
     /// The KZG trusted setup used by this chain.
     pub kzg: Arc<Kzg>,
+    /// RNG instance used by the chain. Currently used for shuffling column sidecars in block publishing.
+    pub rng: Arc<RwLock<Box<dyn RngCore + Sync + Send>>>,
```
jimmygchen: I can't think of an use case where we'd need to acquire a read lock for `rng`, so it might be more efficient to use `Mutex` here (and we could also drop `Sync`).

beacon_node/http_api/src/lib.rs
jimmygchen: Nice, thanks for making the change. I think `OsRng` is fine given we don't currently call this too frequently and only use it to shuffle a list of up to 128 items.

jimmygchen (changes requested):
@SunnysidedJ thanks for the PR, the changes looks good. I've added some comments above. Let me know what you think :)

beacon_node/http_api/src/lib.rs
SunnysidedJ: Yes, also just to add that OsRng is only used to reseed the RNG. Thus, it is barely ever used in this context (that output needed each time is very small to hit reseed threshold often) except for the initial seeding.

beacon_node/beacon_chain/src/builder.rs
SunnysidedJ: Agreed

beacon_node/beacon_chain/src/beacon_chain.rs
SunnysidedJ: Also agreed

jimmygchen (approved):
Looks good, thanks @SunnysidedJ 🙏

beacon_node/beacon_chain/src/builder.rs
```diff
@@ -31,6 +31,7 @@ use logging::crit;
 use operation_pool::{OperationPool, PersistedOperationPool};
 use parking_lot::{Mutex, RwLock};
 use proto_array::{DisallowedReOrgOffsets, ReOrgThreshold};
+use rand::rngs::StdRng;
```
jimmygchen: This needs to be in the `test` mod below.

jimmygchen (changes requested):
Please see comment above

beacon_node/http_api/src/publish_blocks.rs
```diff
@@ -519,7 +519,7 @@ fn publish_column_sidecars<T: BeaconChainTypes>(
             .len()
             .saturating_sub(malicious_withhold_count);
         // Randomize columns before dropping the last malicious_withhold_count items
-        data_column_sidecars.shuffle(&mut rand::thread_rng());
+        data_column_sidecars.shuffle(&mut **chain.rng.lock());
```
dapplion, line 524: Why not make `chain.rng` an option and default to thread_rng so we don't need to deal with locking on the production path?

SunnysidedJ: I'm not sure how you would make it default to an RNG that would not require a mutex unless you put an if-else here. If so, I think it's cleaner to not take that approach as it needs additional manual checks at usage.

jimmygchen: Yeah I think what Lion meant was to put an if-else - I don't see this making a noticeable difference, as shuffling 128 columns is very quick and I agree this approach is cleaner. Although you could argue the alternative more future proof if other parts of the beacon chain start using the `chain.rng` elsewhere. No strong opinion from my side.

michaelsproul: I don't mind the lock. Under the hood `thread_rng` is using a lock too.

If we wanted to abstract over the `if-else` optionality, we could define a method on `BeaconChain` like:

```rust
chain.with_rng(|rng| ..);
```

But I don't think that's really necessary

jimmygchen: Thanks @dapplion please see my comment above - I'm going to merge this now and happy to reconsider this if you feel strongly about it.

jimmygchen (approved):
LGTM - CI job is failing due to a Kurtosis issue unrelated to this PR, will merge when its fixed. Thanks @SunnysidedJ !

PR #7198, opened 2025-03-24
Admin add/remove peer

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Adds endpoints to add and remove trusted peers from the http api. The added peers are trusted peers so they won't be disconnected for bad scores. We try to maintain a connection to the peer in case they disconnect from us by trying to dial it every heartbeat.

### DIFF
--- a/beacon_node/lighthouse_network/src/peer_manager/mod.rs
+++ b/beacon_node/lighthouse_network/src/peer_manager/mod.rs
@@ -114,6 +114,7 @@ pub struct PeerManager<E: EthSpec> {
     metrics_enabled: bool,
     /// Keeps track of whether the QUIC protocol is enabled or not.
     quic_enabled: bool,
+    trusted_peers: HashSet<Enr>,
     /// The logger associated with the `PeerManager`.
     log: slog::Logger,
 }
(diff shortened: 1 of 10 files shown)


eserilev:
Is it possible to write a test case for this new endpoint?

beacon_node/lighthouse_network/src/peer_manager/mod.rs
```diff
@@ -943,6 +945,13 @@ impl<E: EthSpec> PeerManager<E> {
         }
     }
 
+    fn maintain_trusted_peers(&mut self) {
+        let trusted_peers = self.trusted_peers.clone();
```
michaelsproul, line 949: Minor inefficiency, but I don't think we need both `Clone`s here

michaelsproul: oh nvm, it's because `dial_peer` is `&mut self`

beacon_node/lighthouse_network/src/peer_manager/mod.rs
```diff
@@ -943,6 +945,13 @@ impl<E: EthSpec> PeerManager<E> {
...
+        for trusted_peer in &trusted_peers {
+            self.dial_peer(trusted_peer.clone());
+        }
```
michaelsproul: Can get rid of this one, but it's likely not really important

```suggestion
        for trusted_peer in trusted_peers {
            self.dial_peer(trusted_peer);
        }
```

michaelsproul (approved):

PR #7199, opened 2025-03-25
Add Electra forks to basic sim tests

jimmygchen:
## Issue Addressed

This PR adds transitions to Electra ~~and Fulu~~ fork epochs in the simulator tests.

~~It also covers blob inclusion verification and data column syncing on a full node in Fulu.~~

UPDATE: Remove fulu fork from sim tests due to https://github.com/sigp/lighthouse/pull/7199#issuecomment-2852281176


jimmygchen:
On my laptop, a fulu block takes up to 5s to import, and currently we run with 2s slot time (`speed-up-factor=2`). This should get better once we offload computation from the beacon node, but we may need to increase the slot time or reduce the number of blobs / node for this test to work.

jimmygchen:
I have tried slowing down the tests, and getting better results but still not perfect:
```
2025-03-25T11:39:48.3641052Z Mar 25 11:39:48.360 DEBUG Successfully verified gossip data column sidecar  slot: 131, block_root: 0xc196783f41eb37f8b4ef4cc4220f6f3776c0a73ef5e98d6d5d86e01b8ccec957, index: 113 
2025-03-25T11:39:48.3642654Z thread 'main' panicked at testing/simulator/src/basic_sim.rs:387:41:
2025-03-25T11:39:48.3643604Z called `Result::unwrap()` on an `Err` value: "There wasn't a block produced at every slot, got: 112, expected: 129"
```

I think we'll have to wait until #7117 is merged (offload proof computation from BN).

jimmygchen:
@dapplion this PR should be unblocked now with #7117 merged, please review when you have time - this will give us some more e2e test coverage on PeerDAS syncing.

jimmygchen:
After merging #7117, this seems to be doing a lot better, however it's still missing a few slots.
I've merged in #7304 to see if it improves things (cc @pawanjay176 )

jimmygchen:
Looks like we're really close after merging the above mentioned 2 PRs, only missing 1 slot now.

```
thread 'main' panicked at testing/simulator/src/basic_sim.rs:388:41:
called `Result::unwrap()` on an `Err` value: "There wasn't a block produced at every slot, got: 128, expected: 129, missed: [119]"
```

jimmygchen:
Note: we don't have getBlobsV2 in the EL version being used, which may improve things but i think it would be good to make sure this works without it.

dapplion (approved):
LGTM

jimmygchen:
So I pushed one more change to reduce the likelihood of miss blocks:
- changed number of supernodes from 6 -> 2 (1 BN and 1 proposer only node)
- change `speed-up-factor` from 3 -> 2, this means changing slot time from 2s -> 3s. 

If this still doesn't work, then we'd need to limit the number of blobs we produce. The current set up with 12 blobs, 6 supernodes and 2s slots isn't feasible on a CI runner.

jimmygchen:
Found the root cause of the failure - there was an incorrect `EthSpec` import, which resulted in wrong epoch calculation when running the minimal preset, fixed here and reverted the #7304 commit from this PR:
https://github.com/sigp/lighthouse/pull/7199/commits/337408500c0758925bb1c676cf728e03b304458a

.github/workflows/test-suite.yml
```diff
@@ -295,8 +295,8 @@ jobs:
       with:
           channel: stable
           cache-target: release
-    - name: Run a basic beacon chain sim that starts from Bellatrix
-      run: cargo run --release --bin simulator basic-sim
+    - name: Run a basic beacon chain sim that starts from Deneb
+      run: cargo run --release --bin simulator basic-sim --speed-up-factor 2
```
jimmygchen: The existing 2 second slots was a bit too short for Fusaka with up to 12 blobs.

jimmygchen:
I've noticed that only after Fulu fork we get delayed head block:
```
May 05 16:49:25.086 DEBUG Delayed head block                            block_root: 0x250eb95df22daa9be51b6f77ba52ccba9e95e7d7b499be2c2d9e224fd5c98a10, proposer_index: 21, slot: 9, total_delay_ms: 1086, observed_delay_ms: "175", blob_delay_ms: "unknown", consensus_time_ms: "4", execution_time_ms: "0", available_delay_ms: "180", attestable_delay_ms: "1074", imported_time_ms: "903", set_as_head_time_ms: "1" 
```

and notice the differences for `verify_kzg` between different nodes (2ms - 23ms for the exact same column):

```
May 05 16:49:24.534 DEBUG Data column validation timing                 total: 2, slot: 9, index: 121, timing: TimingInfo { verify_sidecar: 42ns, verify_subnet: 41ns, verify_future: 125ns, verify_finalized: 334ns, verify_first: 41ns, verify_inclusion: 2.792µs, verify_parent: 791ns, verify_slot: 42ns, verify_proposer: 726.083µs, verify_kzg: 2.20475ms, observe_slashable: 1.25µs, observe_gossip: 209ns } 
May 05 16:49:24.991 DEBUG Data column validation timing                 total: 12, slot: 9, index: 121, timing: TimingInfo { verify_sidecar: 41ns, verify_subnet: 0ns, verify_future: 333ns, verify_finalized: 458ns, verify_first: 416ns, verify_inclusion: 7.375µs, verify_parent: 1.791µs, verify_slot: 42ns, verify_proposer: 3.344792ms, verify_kzg: 8.668625ms, observe_slashable: 1.542µs, observe_gossip: 208ns } 
May 05 16:49:25.049 DEBUG Data column validation timing                 total: 26, slot: 9, index: 121, timing: TimingInfo { verify_sidecar: 42ns, verify_subnet: 125ns, verify_future: 292ns, verify_finalized: 542ns, verify_first: 416ns, verify_inclusion: 8µs, verify_parent: 1.833µs, verify_slot: 125ns, verify_proposer: 2.367833ms, verify_kzg: 23.76425ms, observe_slashable: 1.416µs, observe_gossip: 250ns } 
```

I wanted to be sure we're not slow down because of locks - but this is likely an issue on local setup with limited resources 
 - seeing the differences in verify kzg time. The worst case for verifying KZG in the sim tests is about ~100ms for a single column with up to 9 blobs, this is really high and should ideally be under 5ms for 6s slots.

It does perform better under kurtosis though - I don't see the same issue using a similar setup under kurtosis - 2 supernodes 2 fullnodes, 6s slot, and no missed slots for many epochs. I'm guessing the kurtosis setup potentially:
1.  allocates resources more evenly; OR
2. has `getBlobsV2` with a real EL, which allows skipping kzg verification in the CL.

I think this test would work if we set slot time to > 6 seconds, however it would take a lot longer - 6s slot takes ~30min, and it doesn't feel like it's worth it. We should consider writing a Kurtosis test for this.

For now it probably make sense to get the Electra fork and the bug fixes in. I'll remove the Fulu changes from the sim tests and we can move on from this .

.github/workflows/test-suite.yml
jimmygchen: Reverted after removing Fulu.

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -589,19 +588,19 @@ fn verify_proposer_and_signature<T: BeaconChainTypes>(
     chain: &BeaconChain<T>,
 ) -> Result<(), GossipDataColumnError> {
     let column_slot = data_column.slot();
-    let column_epoch = column_slot.epoch(E::slots_per_epoch());
+    let slots_per_epoch = T::EthSpec::slots_per_epoch();
+    let column_epoch = column_slot.epoch(slots_per_epoch);
```
jimmygchen, line 592: This is a bug in minimal preset - causing proposer shuffling miss for every gossip column.

michaelsproul (approved):

PR #7209, opened 2025-03-27
Prevent duplicate effective balance processing

michaelsproul:
## Proposed Changes

Ensure that each validator has its effective balance updated exactly once during epoch processing.


realbigsean (approved):
Nice one

PR #7213, opened 2025-03-27
Pubkey cache fix for skipped slots

michaelsproul:
## Issue Addressed

I suspect there's an issue with how the validator pubkey cache is updated after Electra:

Skipped slots can import new validators, i.e. validators get added to the state when advancing to the epoch boundary (because of new finalization) but not added to the cache. Then when we try to verify a chain segment or a block in the next epoch from one of these validators, we fail due to the cache not being updated 

## Proposed Changes

* [ ] Test case to reproduce the issue (WIP)
* [ ] Fix (no attempt made)

### DIFF
--- a/beacon_node/beacon_chain/tests/electra.rs
+++ b/beacon_node/beacon_chain/tests/electra.rs
@@ -1,13 +1,44 @@
 #![cfg(not(debug_assertions))] // Tests run too slow in debug.
 
-use beacon_chain::test_utils::BeaconChainHarness;
+use beacon_chain::{
+    builder::BeaconChainBuilder,
+    test_utils::{get_kzg, mock_execution_layer_from_parts, BeaconChainHarness, DiskHarnessType},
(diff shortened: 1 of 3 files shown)


beacon_node/beacon_chain/tests/electra.rs
```diff
@@ -0,0 +1,106 @@
...
+                .execution_requests_mut()
+                .unwrap()
+                .deposits
+                .push(deposit_request)
+                .unwrap();
+        })
+        .await;
+
+    // Compute correct state root.
+    // FIXME: this is kinda nasty
```
michaelsproul, line 105: it probably would have been cleaner to use the execution block generator to include the deposit requests, but it's a PITA in my experience, so I opted for just hacking the deposit into the block and recomputing the state root and signature

michaelsproul:
Closing as stale.

PR #7215, opened 2025-03-27
Pubkey cache test changes

eserilev:
This was my attempt at getting the test to pass. It passes with some caveats

Initially we were creating a block inserts some deposit requests into the execution request portion of the block and then trying to run `per_block_processing` & `process_block` These code paths never convert the execution deposit request into an actual deposit that gets included in the blocks body. Therefore the deposit never actually occurs.

I've altered the code a bit to explicitly include the deposits in the block body. `per_block_processing` actually processes the block succesfuly. However `process_block` throws an exception because we theres an expected deposit in the block, but no expected deposits on the state being used in that function.

I've commented out some of the code that causes us to either raise an exception or fail the test. In the meantime I'm working to gain more familiarity of the deposit flow so that I can be more useful.

### DIFF
--- a/beacon_node/beacon_chain/src/test_utils.rs
+++ b/beacon_node/beacon_chain/src/test_utils.rs
@@ -2173,6 +2173,38 @@ where
         block.body_mut().voluntary_exits_mut().push(exit).unwrap();
     }
 
+    /// Create a new block,with some deposits.
+    ///
+    /// The state returned is a pre-block state at the same slot as the produced block.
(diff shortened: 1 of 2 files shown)


michaelsproul:
I think after Electra all we should need is the deposit request, which goes into pending deposits and gets processed from there? I think the problem with my test might have been that the deposit amount was too large to be processed immediately

eserilev:
yeah you're right, i was looking at process_deposits and not process_deposit_requests further down belo

Issue #7216, opened 2025-03-28
Global pubkey cache is not crash safe


michaelsproul:
There's an issue with the global pubkey cache

We update the cache in memory prior to writing the new keys to disk. This can result in changes being made in memory but never making it to disk (e.g. if a block fails to be imported after the changes are made). Later blocks that would import the same keys to the cache then skip over them (because they are already in memory) and would not write them to disk.

The result is that the node will continue running OK until it is restarted. The in-memory cache is correct, but is out of sync with the on-disk cache.

I think we can solve this redundantly in two ways:

1. Maintain the following invariant: `key in memory -> key on disk`. To do this we will need to write keys to disk before ever updating the in-memory store. This has some performance challenges associated with holding the lock while performing IO (previous issue: https://github.com/sigp/lighthouse/issues/2327).
2. We can also defensively import keys from the head state upon restart.

The second change has lower impact, although it feels like a bit of a hack.

michaelsproul:
The workaround is merged, so this is essentially dealt with, but I've been thinking about cleaner fixes

One option to maintain the invariant might be to keep track of which keys have or haven't been flushed to disk from the in-memory cache. We could add them initially with `on_disk: false`, and then update that to `on_disk: true` after the database transaction commits. In the worst case another thread may add the same entries to the in-memory cache while the disk write is happening, which will then get written to disk twice, and updated twice (we should ensure the updates to `on_disk` in memory are idempotent, i.e. not compare-and-swap)

michaelsproul:
Removing the v7.0 tag as the merged workaround is working.

PR #7217, opened 2025-03-28
Top-up pubkey cache on startup

eserilev:
## Issue Addressed

This is a workaround for #7216

## Proposed Changes

In the case of gaps between the in-memory pub key cache and its on-disk representation, use the head state on startup to "top-up" the cache/db w/ any missing validators

### DIFF
--- a/beacon_node/beacon_chain/src/builder.rs
+++ b/beacon_node/beacon_chain/src/builder.rs
@@ -850,7 +850,7 @@ where
         let validator_pubkey_cache = self
             .validator_pubkey_cache
             .map(|mut validator_pubkey_cache| {
-                // If any validators weren't persisted to disk on previous runs, this will use the genesis_state to
+                // If any validators weren't persisted to disk on previous runs, this will use the head state to
                 // "top-up" the in-memory validator cache and its on-disk representation with any missing validators.
                 let pubkey_store_ops = validator_pubkey_cache
                     .import_new_pubkeys(&head_snapshot.beacon_state)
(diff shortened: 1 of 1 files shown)


beacon_node/beacon_chain/src/builder.rs
```diff
@@ -325,9 +325,20 @@ where
...
         );
 
-        let pubkey_cache = ValidatorPubkeyCache::load_from_store(store)
+        let mut pubkey_cache = ValidatorPubkeyCache::load_from_store(store.clone())
             .map_err(|e| format!("Unable to open persisted pubkey cache: {:?}", e))?;
 
+        // If any validators weren't persisted to disk on previous runs, this will use the genesis_state to
+        // "top-up" the in-memory validator cache and its on-disk representation with any missing validators.
+        let pubkey_store_ops = pubkey_cache
+            .import_new_pubkeys(&genesis_state)
```
michaelsproul: I don't think we wanna do this with the genesis state, we need the head state (or maybe the finalized state, but the head would be better)

michaelsproul: Here: https://github.com/sigp/lighthouse/blob/491091f317371aea92110487b6e8a0c960744a49/beacon_node/beacon_chain/src/builder.rs#L817-L820

michaelsproul: We'd do the top up in the `Ok` case here: https://github.com/sigp/lighthouse/blob/491091f317371aea92110487b6e8a0c960744a49/beacon_node/beacon_chain/src/builder.rs#L861-L864

michaelsproul (approved):
LGTM!

PR #7223, opened 2025-03-28
Update metrics.rs

zeevick10:


jimmygchen:
Hi @zeevick10 I'm closing this PR since there's no description and information why this change is required. Feel free to comment or add more details and we can reopen it if there's reason to do this change.

zeevick10:
> Hi @zeevick10 I'm closing this PR since there's no description and information why this change is required. Feel free to comment or add more details and we can reopen it if there's reason to do this change.

I adjusted for a range up to 0.1 seconds (100ms), I think a range of 10-150ms works well. Thanks!

Issue #7124, opened 2025-03-13
Improve HTTP API metrics and logs


michaelsproul:
@macladson and I noticed that we are missing some important routes from the HTTP metrics. This bloc probably needs an update:

https://github.com/sigp/lighthouse/blob/d60c24ef1cc0b5dfa930e1dd4fc85abc29e5fc4c/beacon_node/http_api/src/lib.rs#L212-L243

With `tracing`, we also lost the rather useful log:

https://github.com/sigp/lighthouse/blob/bf955c7543dac8911a6f6c334b5b3ca4ef728d9c/beacon_node/http_api/src/lib.rs#L205-L212

Improving this will help us see the difference in perf when upgrading from warp to axum

While we're restoring that log, we should ensure it always logs milliseconds as a float like:

```
elapsed_ms: 1.0
```

```
elapsed_ms: 0.001
```

```
elapsed_ms: 50000.0
```

This makes the logs easier to parse/categorise (no units to remove)

jimmygchen:
Fixed in #7225

PR #7225, opened 2025-03-29
Restore HTTP API logging and add more metrics

macladson:
## Issue Addressed

#7124 

## Proposed Changes

- Restores previous HTTP logging with tracing compatible syntax
- Adds metrics for certain missing endpoints (and alphabetized the existing ones)


michaelsproul (approved):
Looks good from a code PoV. I will test it once manually and then we can merge

macladson:
For the metrics, there might be some `v2` and `v3` endpoints I've missed. Since some of our endpoints are version agnostic now it's hard to tell at a glace which versions exist, but I'll cross check them with the spec

beacon_node/http_api/src/lib.rs
```diff
@@ -215,35 +215,66 @@ pub fn prometheus_metrics() -> warp::filters::log::Log<impl Fn(warp::filters::lo
...
                 .or_else(|| starts_with("v1/beacon/pool/sync_committees"))
-                .or_else(|| starts_with("v1/beacon/blocks/head/root"))
-                .or_else(|| starts_with("v1/validator/prepare_beacon_proposer"))
-                .or_else(|| starts_with("v1/validator/register_validator"))
+                .or_else(|| starts_with("v1/beacon/pool/voluntary_exits"))
+                .or_else(|| starts_with("v1/beacon/rewards/blocks"))
+                .or_else(|| starts_with("v1/beacon/rewards/attestations"))
+                .or_else(|| starts_with("v1/beacon/rewards/sync_committee"))
+                .or_else(|| starts_with("v1/beacon/rewards"))
+                .or_else(|| starts_with("v1/beacon/states"))
```
macladson, line 235: We are still using a catch-all from all `beacon/states/{state_id}` endpoints. At a later time it might be worth trying to split these out into their discrete routes.

michaelsproul: they're probably all equally slow, but yeah that's a decent idea

michaelsproul (approved):
Nice, thanks!

Issue #6853, opened 2025-01-24
Update store tests to cover data column pruning


jimmygchen:
## Description

There are a few `TODO`s in `store_tests.rs` to add tests to cover data column pruning that needs to be addressed:

https://github.com/sigp/lighthouse/blob/98863da429c4e45eada0a8e89acb0205589e9ca1/beacon_node/beacon_chain/tests/store_tests.rs#L3078-L3080

SunnysidedJ:
Hi @jimmygchen, I'd like to work on this issue as well.

dapplion:
> Hi @jimmygchen, I'd like to work on this issue as well.

Thanks for jumping on this, please let us know if you need help

jimmygchen:
Completed in #7228

Issue #6572, opened 2024-11-07
Remove `DataColumnInfo`


eserilev:
## Description

In #6559 were going ahead and removing `BlobInfo`. Its a bit of a pain to manage as it requires us to make an atomic transaction across two separate databases. Instead we're going to just iterate across the full blob db and prune blobs past the data availability window (additional info can be found in the linked issue)

I'd like to do the same for `DataColumnInfo`. Unless theres some additional need for this table that I'm unaware of, I'd like to go ahead and remove it.

michaelsproul:
Let's do it 🔥

michaelsproul:
Just noting that we currently use this (incorrectly) while processing RPC requests:

https://github.com/sigp/lighthouse/blob/e5b4983d6baf85770fe4539a565d8a2dd462bc53/beacon_node/network/src/network_beacon_processor/rpc_methods.rs#L1165-L1191

We should update to use the data column custody info, or the data availability boundary as appropriate.

michaelsproul:
I'm actually torn on removing this now. I'm writing down some pros and cons to help me make my mind up

## Pros to removing

- Remove cross-database transactions for blob/column metadata. Counterpoint: we could do this by storing the metadata in the blobs DB itself.

## Cons to removing

- Users loses all visibility into which blobs/columns are available in the database. While testing the `--complete-blob-backfill` flag it was quite nice to see progress in the HTTP API. Counterpoint: this can kind of be inferred? If `prune-blobs=false` we can _sort of_ guess that blobs are available back to `oldest_block_slot` (although this is not necessarily true), and otherwise we know they are available back to the data availability period?
- Need a breaking DB migration to remove it now, this creates an incompatibility between `v8.0.0-rc.X` and `v8.0.0` proper which might catch some users off guard. If we were to keep the metadata and just move it to the blobs DB, then a two-way migration would be possible and we could implement this in v8.1.0 rather than v8.0.0.

eserilev:
Yeah maybe they do have their usecase. For `DataColumnInfo` the oldest column slot is wrong as soon as the cgc changes, so we'd probably need to fix that

michaelsproul:
Conclusion from our meeting today:

- Keep `BlobInfo`. This will soon only be relevant for archive nodes backfilling beyond the DA period (i.e. `--complete-blob-backfill`). We might want to migrate it to the blobs DB in some future schema upgrade in order to simplify the cross-database transactions.
[...]

Issue #6120, opened 2024-07-17
stack overflow when running `overflow_cache_test_state_cache` in debug


jimmygchen:
## Description

Getting this error when running `cargo test` from the root directory:

```
thread 'tokio-runtime-worker' has overflowed its stack
```

Reproduced with `cargo nextest` as well. It doesn't occur on CI, but on our laptops. i think it may be due to different platform max stack size defaults.

The failing test identified is `overflow_cache_test_state_cache`

Based on [this comment](https://github.com/sigp/lighthouse/blob/2f0af2be8902596a78ac67a8d1b402bd944ff73e/beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs#L711), we might be able to get rid of this test once the state cache is removed in da checker:

```
// THIS TEST CAN BE DELETED ONCE TREE STATES IS MERGED AND WE RIP OUT THE STATE CACHE
```

The root cause of the stack overflow seems to be in the blob conversion here:
https://github.com/sigp/lighthouse/blob/969d12dc6f4b1162a9b75c7f6801415a39ed8eea/beacon_node/beacon_chain/src/kzg_utils.rs#L32

Not sure if there's an easy way around it given that the KZG library accepts `Blob` containing an owned array. It would be nice if it accepts [refs like the peerdas KZG lib](https://github.com/sigp/lighthouse/blob/094ee6036b935b2f7fd9e9b2e4250aef8ebd1110/beacon_node/beacon_chain/src/kzg_utils.rs#L14), so it doesn't require making a copy of the bytes (which will likely be unfeasible when we increase the blob count) 

## Version

5.2.1+ (current `unstable` branch)

alecalve:
It's still occurring for me when running `cargo test` on my laptop having checked out the tag `v6.0.0`.

michaelsproul:
Noting that we have another report from a Mac user:

- https://github.com/sigp/lighthouse/issues/6647

This does seem to be Mac specific, as we _do_ run this test in debug mode on CI:

https://github.com/sigp/lighthouse/blob/c042dc14d74352512b7632e0ee6ec07f1aa26b3a/.github/workflows/test-suite.yml#L241-L242

(I forgot we still run things in debug :sweat_smile: )

Patchoulis:
I noticed this was an issue, and I can confirm it also occurs on Linux(tested on Ubuntu). Tested it today to confirm.

jchavarri:
[...]

Issue #4987, opened 2023-12-06
Non-release tests using KZG verification stack overflow on macOS M1


realbigsean:
## Steps to reproduce

```
cd beacon_node/beacon_chain/src/data_availability_checker
cargo test -- overflow_cache
```
## Info 
I haven't investigated whether the stacks usage is high in this particular test prior to this call or if there's a different issue:

https://github.com/ethereum/c-kzg-4844/blob/main/bindings/rust/src/bindings/mod.rs#L234

I think we should explore updating  `c-kzg` bindings to accept slices so we don't have to do this transition from lighthouse types to `c-kzg` types (which is forcing this stack allocation, and also forces us to copy/clone blobs during verification)

mratsim:
A Blob is 131kB, MacOS default stack size is 512 kB, so a single Blob does use more than 25% of what's available.

Miscellaneous copies would lead to an overflow.

Now in general Blob should always be on the heap due to their size.

realbigsean:
Thanks @mratsim 

> Now in general Blob should always be on the heap due to their size.

I agree. We don't store them internally on the stack, but when we pass them to `c-kzg`, we're forced to initialize a blob-sized array (at the link in the PR description).  I'm interested in trying out `constantine`, either directly or via `rust-kzg`.

PR #7228, opened 2025-03-31
#6853 Adding store tests for data column pruning

SunnysidedJ:
## Issue Addressed

#6853 Update store tests to cover data column pruning

## Proposed Changes

Created a helper function `check_data_column_existence` which is a copy of `check_blob_existence` but checking data columns instead.
The helper function is then used to check whether data columns are also pruned when blobs are pruned if PeerDAS is enabled.

## Additional Info

There were three bugs discovered when writing these tests

1. Data column info was not updated post fulu
2. Blob info was continuing to be updated post fulu
3. `delete_if` was broken for leveldb

For bug 3, we now `seek` to the correct column before trying to delete values. The `prune_blobs_across_fork_boundary` test provides test coverage for this change. When there are both blobs and data columns in the blob store, `delete_if` properly prunes blobs (or data columns). Without `seek` blobs would never be pruned as soon as the first data column was inserted into the blob store


beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -3205,6 +3210,9 @@ async fn deneb_prune_blobs_no_finalization() {
         Some(deneb_fork_slot)
     );
     check_blob_existence(&harness, Slot::new(0), harness.head_slot(), true);
```
jimmygchen, line 3704: You might have already noticed, this test [runs multiple times](https://github.com/sigp/lighthouse/blob/166f6df5a973e5edd6faf17812f483260f50d2a1/Makefile#L145-L146 ) on CI for all forks.

The code that reads `FORK_NAME` from env var is here:
https://github.com/sigp/lighthouse/blob/6f31d44343086ddaa70b396632c0eee724256454/beacon_node/beacon_chain/src/test_utils.rs#L193-L206

So this assertion would fail when `FORK_NAME=fulu`, because we don't store any blobs from since the Fulu fork. I think it might be cleaner to start a separate test that verifies data column pruning instead of having to add conditions on every assertion.

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -3337,6 +3348,9 @@ async fn deneb_prune_blobs_margin_test(margin: u64) {
         Some(deneb_fork_slot)
     );
     check_blob_existence(&harness, Slot::new(1), harness.head_slot(), true);
+    if check_data_columns {
+        check_data_column_existence(&harness, Slot::new(1), harness.head_slot(), true);
+    }
```
jimmygchen: Same as above, when `FORK_NAME=fulu`, `check_blob_existence` would fail already before hitting this line.

jimmygchen (dismissed):

jimmygchen:
Thanks for the PR @SunnysidedJ ! 

Please see my comments above and let me know if it's unclear or if you have any questions!

beacon_node/beacon_chain/tests/store_tests.rs
SunnysidedJ: Sorry, I thought we still keep the blobs in Fulu. Changes have been made respectively. Is there a reason that we no longer keep blobs? Is it to save the storage?

SunnysidedJ:
Major changes:
- Separated tests for data column pruning.
- Blob pruning tests only run on Deneb while column ones only run on Fulu with PeerDAS enabled.
- Found that fork_boundary test is not working (see inline comments).

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -3222,28 +3224,34 @@ async fn deneb_prune_blobs_no_finalization() {
...
 
     let harness = get_harness(store.clone(), LOW_VALIDATOR_COUNT);
 
-    let num_blocks = E::slots_per_epoch() * 7;
+    let blocks_to_deneb = E::slots_per_epoch() * 7;
+    let blocks_to_fulu = E::slots_per_epoch() * 13;
 
-    // Finalize to epoch 5.
+    // Finalize to epoch 5 (Deneb).
     harness
```
SunnysidedJ, line 3745: Returns a stack overflow here. Seems like using the spec with customization is not working properly. I think we need a separate issue for this. Any thoughts?

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -3282,6 +3290,56 @@ async fn deneb_prune_blobs_fork_boundary() {
     assert_eq!(store.get_blob_info().oldest_blob_slot, Some(pruned_slot));
     check_blob_existence(&harness, Slot::new(0), pruned_slot - 1, false);
     check_blob_existence(&harness, pruned_slot, harness.head_slot(), true);
+
+    // Finalize to epoch 11 (Fulu)
```
SunnysidedJ: Added a logic to test that blobs get removed once we reach Fulu fork, but couldn't test it due to the stack overflow issue above

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -3422,6 +3481,303 @@ fn check_blob_existence(
...
+        panic!("TEST");
+        //return;
+    }
+
+    let harness = get_harness(store.clone(), LOW_VALIDATOR_COUNT);
+
+    let num_blocks = E::slots_per_epoch() * 7;
+
+    // Finalize to epoch 5.
+    harness
```
SunnysidedJ, line 4187: The same issue as above

beacon_node/beacon_chain/tests/store_tests.rs
SunnysidedJ: (The test fails at the unstable branch as well)

beacon_node/beacon_chain/tests/store_tests.rs
jimmygchen: Yes, that's one of the reasons - but also it's no longer useful to the CL from Fulu.
- The network form of blobs on the beacon chain is `DataColumnSidecar`, we'll stop using `BlobSidecar` on p2p from Fulu slots.
- Another reason is fullnodes will no longer have blobs, but just samples (DataColumns)

beacon_node/beacon_chain/tests/store_tests.rs
jimmygchen: Yeah this has been a known issue for a while, so these tests are always run with `--release`. This is because the `c-kzg` library forces us to initialise a blob-sized array. I think either moving to `rust-eth-kzg` or perhaps updating to the latest `c-kzg` library would resolve this (I haven't checked)
- https://github.com/sigp/lighthouse/issues/6120
- https://github.com/sigp/lighthouse/issues/4987

beacon_node/beacon_chain/tests/store_tests.rs
jimmygchen: ah yeah, could you run it with `--release`, unfortunately we haven't been able to tackle that issue yet.

beacon_node/beacon_chain/tests/store_tests.rs
SunnysidedJ: I see. I will try running them either on Windows or Linux machine for the completion of this task for now

jimmygchen: This should work on macOS if you run with `--release` or you can run the debug test with this env var: `RUST_MIN_STACK=4194304` (setting a higher minimum stack size)

SunnysidedJ:
When I enable Fulu and try blob pruning, the blobs aren't pruned correctly. It'd be appreciated if you could have a look on the commit that reproduces the problem: https://github.com/SunnysidedJ/lighthouse/commit/6217cf9d67e3976a0d7d333f3691c316ababaff3

eserilev:
I'm going to take a look at the fulu blob pruning test issues

eserilev:
The test is now passing. The reason it was failing was two fold:
- we werent advancing the slot clock before extending the chain
- we tried to fork from deneb directly to fulu

I've added an intermediary step to fork to electra and run some checks there. Not sure if theres any additional checks we want to make in this test.

eserilev:
ah still need to fix one other test, working on that now

eserilev:
a634758 we werent updating data column info after pruning.

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -3264,17 +3267,17 @@ async fn deneb_prune_blobs_fork_boundary() {
...
     check_blob_existence(&harness, Slot::new(0), pruned_slot - 1, false);
     check_blob_existence(&harness, pruned_slot, harness.head_slot(), true);
-
-    // Finalize to epoch 11 (Fulu)
+    // Finalize to epoch 11 (Electra)
+    harness.advance_slot();
     harness
         .extend_chain(
-            blocks_to_fulu as usize,
+            blocks_to_electra as usize,
```
SunnysidedJ, line 135: Shouldn't this be 13 (7 + 6 epochs in total)?

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -3286,33 +3289,72 @@ async fn deneb_prune_blobs_fork_boundary() {
...
+            let oldest_slot = data_availability_boundary.start_slot(E::slots_per_epoch());
+            assert_eq!(store.get_blob_info().oldest_blob_slot, Some(oldest_slot));
+            // Blobs should exist
+            check_blob_existence(&harness, oldest_slot, harness.head_slot(), true);
         } else {
-            fulu_fork_epoch.start_slot(E::slots_per_epoch())
+            // Fulu fork epoch
+            let oldest_slot = fulu_fork_epoch.start_slot(E::slots_per_epoch());
+            // Blobs should not exist
+            check_blob_existence(&harness, oldest_slot, harness.head_slot(), false);
```
SunnysidedJ: Need to also check that blobs don't exist from Slot 0

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -3286,33 +3289,72 @@ async fn deneb_prune_blobs_fork_boundary() {
...
     assert_eq!(store.get_blob_info().oldest_blob_slot, Some(pruned_slot));
 
+    let electra_first_slot = electra_fork_epoch.start_slot(E::slots_per_epoch());
+    // Check that blobs have been pruned up to the pruned slot
+    check_blob_existence(&harness, Slot::new(0), pruned_slot - 1, false);
+    // Check that blobs exist from the pruned slot to electra
+    check_blob_existence(&harness, pruned_slot, electra_first_slot - 1, true);
+    // Check that blobs exist from electra to the current head
+    check_blob_existence(&harness, electra_first_slot, harness.head_slot(), true);
+    // Finalize to epoch 15 (Fulu)
```
SunnysidedJ, line 1612: to 17?

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -3286,33 +3289,72 @@ async fn deneb_prune_blobs_fork_boundary() {
...
+            // Blobs should exist
+            check_blob_existence(&harness, oldest_slot, harness.head_slot(), true);
         } else {
-            fulu_fork_epoch.start_slot(E::slots_per_epoch())
+            // Fulu fork epoch
+            let oldest_slot = fulu_fork_epoch.start_slot(E::slots_per_epoch());
+            // Blobs should not exist
+            check_blob_existence(&harness, oldest_slot, harness.head_slot(), false);
+            // Data columns should exist
+            check_data_column_existence(&harness, oldest_slot, harness.head_slot(), true);
```
SunnysidedJ: Need to also check that columns don't exist from Slot 0

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -3286,33 +3289,72 @@ async fn deneb_prune_blobs_fork_boundary() {
...
 
         // Check oldest blob slot is not updated.
-        let oldest_slot = if data_availability_boundary < fulu_fork_epoch {
-            data_availability_boundary.start_slot(E::slots_per_epoch())
+        if data_availability_boundary < fulu_fork_epoch {
+            // Pre Fulu fork epochs
+            let oldest_slot = data_availability_boundary.start_slot(E::slots_per_epoch());
+            assert_eq!(store.get_blob_info().oldest_blob_slot, Some(oldest_slot));
+            // Blobs should exist
+            check_blob_existence(&harness, oldest_slot, harness.head_slot(), true);
```
SunnysidedJ: Need to also check that blobs don't exist from Slot 0

SunnysidedJ: Pulled the changes and I'm looking into why Slot 40 exists when I call `check_blob_existence(&harness, Slot::new(0), oldest_slot - 1, false)`

SunnysidedJ (commented):
Sorry, I have been busy with other things and couldn't take much attention to this PR. Thank you for taking a deeper look and unblocking some of the issues I faced with the test cases. Please let me know if you have worked further since last commit and if this needs to be done with priority so that I can clearly understand the status. I will continue working on this whenever I'm available!

beacon_node/beacon_chain/tests/store_tests.rs
eserilev: yes the chain was extended to epoch 13, but the finalized epoch is at 11, I've added some comments here:
06431e42dfc9e20bf020ae2668c46bcbeda38a85 to make things a bit clearer

beacon_node/beacon_chain/tests/store_tests.rs
eserilev: weve extended the chain to 17 but finalized epoch 15 (added some comments for clarity)

eserilev:
Theres an issue with the way we try to prune blobs after fulu. If there are data columns in the blobs db `delete_if`, in the leveldb impl, doesn't work as expected

``` rust
pub fn delete_if(
        &self,
        column: DBColumn,
        mut f: impl FnMut(&[u8]) -> Result<bool, Error>,
    ) -> Result<(), Error> {
        let mut leveldb_batch = Writebatch::new();
        let iter = self.db.iter(self.read_options());

        iter.take_while(move |(key, _)| key.matches_column(column))
            .for_each(|(key, value)| {
                if f(&value).unwrap_or(false) {
                    let _timer = metrics::start_timer(&metrics::DISK_DB_DELETE_TIMES);
                    metrics::inc_counter_vec(&metrics::DISK_DB_DELETE_COUNT, &[column.into()]);
                    leveldb_batch.delete(key);
                }
            });

        self.db.write(self.write_options().into(), &leveldb_batch)?;
        Ok(())
    }
```

When we fork to fulu there will be two `DBColumns` in the blob store, `BeaconBlob` and `BeaconDataColumn`. Because of the way these columns are serialized, `BeaconDataColumn` values will always preceed `BeaconBlob` values in the leveldb blob store. In `delete_if` when we try to delete `BeaconBlob` values while `BeaconDataColumn` values exist in the blob store, `iter.take_while` will return false on the first key and immediately exit. This issue doesn't occur when using redb beacuse redb has `Tables`

In the leveldb case I'm thinking for `delete_if` we should introduce a `seek` to forward the iterator to the first instance of a given `DBColumn`. I need to double check other `delete_if` usage to make sure that doesn't potentially break other things or cause performance degradation.

EDIT:

IT seems we only use `delete_if` in the blob store, so introducing `seek` should be fine

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -2889,6 +2885,31 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
         Ok(())
     }
+
+    fn update_blob_or_data_column_info(
+        &self,
+        start_epoch: Epoch,
+        end_slot: Slot,
+        blob_info: BlobInfo,
+        data_column_info: DataColumnInfo,
+    ) -> Result<(), Error> {
```
eserilev, line 3385: added this to ensure that we update data column info and not blob info post fulu (vice-versa pre fulu)

eserilev:
this should be good to go for a review

michaelsproul (approved):
This looks good to me. I think the changes and tests make sense.

The only open question is how to handle the abolition of `BlobInfo` and `DataColumnInfo`. There are open issues for these, so I think they can be handled separately:

- https://github.com/sigp/lighthouse/issues/8203
- https://github.com/sigp/lighthouse/issues/6572

michaelsproul:
@mergify requeue

jimmygchen:
@mergify requeue

PR #7230, opened 2025-03-31
Ensure `light_client/updates` endpoint returns spec compliant SSZ data

### LINKED ISSUE #7167: Create Light Client Update SSZ Responses based on Attested Header Slot and not Signature Slot
The Fork Digest `context` in SSZ response for the `/light_client/update` endpoint is using `signature_slot` as it's source for calculation and not the `update.attested_header.beacon.slot` that the beacon API spec mentions.

[This line](https://github.com/sigp/lighthouse/blob/stable/beacon_node/http_api/src/light_client.rs#L162) in lighthouse.

[Beacon API doc](https://ethereum.github.io/beacon-APIs/#/Beacon/getLightClientUpdatesByRange)

eserilev:
## Issue Addressed

Closes #7167

## Proposed Changes

- Ensure the fork digest is generated from ther light client updates attested header and not the signature slot
- Ensure the format of the SSZ response is spec compliant


common/eth2/src/types.rs
```diff
@@ -800,13 +800,13 @@ pub struct LightClientUpdatesQuery {
 }
 
 #[derive(Encode, Decode)]
-pub struct LightClientUpdateSszResponse {
-    pub response_chunk_len: Vec<u8>,
-    pub response_chunk: Vec<u8>,
+pub struct LightClientUpdateResponseChunk {
+    pub response_chunk_len: u8,
```
michaelsproul: This is meant to be a u64 I think?

The spec says:

```
(
  response_chunk_len: Little-endian Uint64 byte length of `response_chunk`
  response_chunk: (
    context: 4 byte `ForkDigest`
    payload: SSZ serialized payload bytes
  )
)
```

michaelsproul (changes requested):

common/eth2/src/types.rs
eserilev: thanks for catching that

eserilev: [6a60551](https://github.com/sigp/lighthouse/pull/7230/commits/6a60551d606a30e1e07609ba6dff7c48f6ea9df4)

michaelsproul (approved):

macladson (approved):

PR #7238, opened 2025-04-01
RPC RequestId Cleanup

AgeManning:
## Issue Addressed

I've been working at updating another library to latest Lighthouse and got very confused with RPC request Ids. 

There were types that had fields called `request_id` and `id`. And interchangeably could have types `PeerRequestId`, `rpc::RequestId`, `AppRequestId`, `api_types::RequestId` or even `Request.id`. 

I couldn't keep track of which Id was linked to what and what each type meant. 

So this PR mainly does a few things:
- Changes the field naming to match the actual type. So any field that has an  `AppRequestId` will be named `app_request_id` rather than `id` or `request_id` for example. 
- I simplified the types. I removed the two different `RequestId` types (one in Lighthouse_network the other in the rpc) and grouped them into one. It has one downside tho. I had to add a few unreachable lines of code in the beacon processor, which the extra type would prevent, but I feel like it might be worth it. Happy to add an extra type to avoid those few lines. 
- I also removed the concept of `PeerRequestId` which sometimes went alongside a `request_id`. There were times were had a `PeerRequest` and a `Request` being returned, both of which contain a `RequestId` so we had redundant information. I've simplified the logic by removing `PeerRequestId` and made a `ResponseId`. I think if you look at the code changes, it simplifies things a bit and removes the redundant extra info. 

I think with this PR things are a little bit easier to reasonable about what is going on with all these RPC Ids. 

NOTE: I did this with the help of AI, so probably should be checked

### DIFF
--- a/beacon_node/lighthouse_network/src/lib.rs
+++ b/beacon_node/lighthouse_network/src/lib.rs
@@ -122,6 +122,6 @@ pub use peer_manager::{
     ConnectionDirection, PeerConnectionStatus, PeerInfo, PeerManager, SyncInfo, SyncStatus,
 };
 // pub use service::{load_private_key, Context, Libp2pEvent, Service, NETWORK_KEY_FILENAME};
-pub use service::api_types::{Response, ResponseId};
+pub use service::api_types::{OutboundRequestId, Response};
 pub use service::utils::*;
 pub use service::{Gossipsub, NetworkEvent};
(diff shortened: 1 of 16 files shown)


jimmygchen:
Potentially relevant PR from @jxs : https://github.com/sigp/lighthouse/pull/6969

jxs (approved):
Thanks age LGTM, submitted a couple commits to help improve

jxs (approved):

Issue #7234, opened 2025-04-01
Unnecessary `Mutex` in logging layer


macladson:
In order to record span data (which allows us to write out the service name in the log, and also allows us to determine if certain errors are crits), we store them in an `Arc<Mutex<HashMap<_, _>>`. This has the downside of causing us to need to manage the lock, check if it's poisoned, etc. We are also using an unwrap here:

https://github.com/sigp/lighthouse/blob/bde0f1ef0b29608b6cf4e447987843a2b249e433/common/logging/src/tracing_logging_layer.rs#L426-L434

I believe we can use `tracing`'s built in `Extension` system to manage this data for us, something like this:

```rust
let mut extensions = span.extensions_mut(); 

let span_data = SpanData {
    name: attrs.metadata().name().to_string(),
    fields: visitor.fields,
 };

extensions.insert(span_data);
```

This would allow us to remove the `Mutex` and the `unwrap`.

chong-he:
This issue seems to be resolved by #7239?

PR #7239, opened 2025-04-01
Control span data through tracing Extensions

macladson:
## Issue Addressed

#7234 

## Proposed Changes

Removes the `Arc<Mutex<_>` which was used to store and manage span data and replaces it with the inbuilt `Extension` for managing span-specific data. This also avoids an `unwrap` which was used when acquiring the lock over the mutex'd span data.


eserilev:
This looks like a nice simplification to me. I'm trying to think through the different panic scenarios. If theres multiple threads accessing the same spans `extensions_mut` is there a race condition where `get_mut` returns `None` for both threads, but one of the threads ends up panicking? I think since we have span usage across async tasks, the race condition might be technically possible?

common/logging/src/tracing_logging_layer.rs
```diff
@@ -49,23 +45,28 @@ impl<S> Layer<S> for LoggingLayer
...
+            if extensions.get_mut::<SpanData>().is_some() {
+                return;
+            }
+
+            let span_data = SpanData {
+                name: attrs.metadata().name().to_string(),
+                fields: visitor.fields,
+            };
+
+            extensions.insert(span_data);
```
ackintosh: I think using `replace()` instead of `get_mut()` and `insert()` can address concerns about the  race conditions. For reference, the implementation of these methods is here: https://github.com/tokio-rs/tracing/blob/tracing-subscriber-0.3.19/tracing-subscriber/src/registry/extensions.rs#L86-L96

macladson: Oh nice, I must've missed this earlier, thanks!

ackintosh (approved):
Looks good to me!

michaelsproul (approved):
LGTM; didn't review

Issue #7226, opened 2025-03-29
ANSI color codes in log output even in non-interactive terminals


macladson:
In our previous logging implementation using `slog`, coloured terminal output was only applied in situations where the terminal could support ANSI color codes. With `tracing`, we now manually apply color codes to closely match the previous `slog` format. This causes it to appear everywhere that layer is used, so it now affects both `stdout` and non-interactive terminal modes by default. Certain applications are not able to parse these codes as actual color and instead show a mess of ANSI codes surrounding the actual log lines. The workaround now is to run with `--log-color=false`.

We need to add a check to ensure the terminal can support color before we apply these color codes.
Alternatively, `tracing-subscriber` has a ` with_ansi()` method which would allow us to disable all ANSI codes after they have been applied. This means we could unify our log formatting, which currently is split between color-enabled (with codes) and color-disabled (no codes), then just apply `with_ansi(false)` to the appropriate outputs.

macladson:
Completed in #7240

PR #7240, opened 2025-04-01
Disable log color when running in non-interactive mode

macladson:
## Issue Addressed

#7226 

## Proposed Changes

Checks whether the application is running in a terminal, or in non-interactive mode (e.g. using systemd). It will then set the value of `--log-color` to `false` when running non-interactively.

## Additional Information

Note that if the user is running non-interactively _and_ has explicitly set `--log-color=true`, the flag will be silently ignored and no color codes will be present in the output.


macladson (commented):
@antondlr, can you test this and let me know if this works for you?

antondlr:
> @antondlr, can you test this and let me know if this works for you?

can and will, thanks for picking this up!

antondlr (approved):
this LGTM, no issues found in testing

michaelsproul:
@mergify requeue

PR #7241, opened 2025-04-01
Prevent gaps between the global validator cache and its on-disk representation

### LINKED ISSUE #7216: Global pubkey cache is not crash safe

eserilev:
## Issue Addressed

Closes #7216 

## Proposed Changes

Add a new field to the cache `staged_indices`. When writing to the pubkey cache, we will always insert new validators into the `staged_indices` mapping. When we write validators to disk, we will generate the database operations from `staged_indices`. After successful writes to disk, we will flush the validators we wrote to disk  from `staged_indices`.

This should prevent us from ever having gaps between our in-memory validator cache and its on-disk representation. 

To be clear, an example of a gap would be:

in-memory cache: `[1,2,3,4,5]`
on-disk: `[1,2,4,5]` (3 is missing, but 4 and 5 were written to disk)

If a block were to fail import for any reason, we would still have the new validators included in `staged_indices` ready to be written to disk on a subsequent block import. If the program crashes before we're able to import the block to disk, we still won't have gaps. On start-up the block will need to be imported to disk anyways which will capture whatever new validators were missed and write them to disk.

With this change #7217 should technically not be needed anymore. But its a simple change and there is no harm in being defensive, so I've chosen not to revert it.


eserilev:
I think we may be able to write some tests for this change, going to explore that

michaelsproul:
Going to drop this from v7.0 seeing as the workaround fix seems to work well. We can include this in v7.1.0

dapplion:
Why not just stay with the workaround? Feels simpler to reason about

dapplion:
- Closing for staleness

PR #7244, opened 2025-04-02
Attempt to fix flaky network tests

jimmygchen:
Attempt to fix flaky network tests - removed the  double conversion (duration -> slot, slot -> duration) which may have resulted in precision loss.

### DIFF
--- a/beacon_node/network/src/subnet_service/tests/mod.rs
+++ b/beacon_node/network/src/subnet_service/tests/mod.rs
@@ -116,18 +116,16 @@ fn get_subnet_service() -> SubnetService<TestBeaconChainType> {
     )
 }
 
-// gets a number of events from the subscription service, or returns none if it times out after a number
-// of slots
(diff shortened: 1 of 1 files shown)


pawanjay176 (approved):
Not sure if this is the root cause, but it does seem cleaner

jimmygchen:
Posting the investigation I posted privately earlier:

I think the flakiness is likely caused by this calculation (floor division), which may result in the test not waiting for long enough

```rust
        let wait_slots = subnet_service
            .beacon_chain
            .slot_clock
            .duration_to_slot(subscription_end_slot)
            .unwrap()
            .as_millis() as u64
            / SLOT_DURATION_MILLIS;
```

e.g. if `duration_to_slot` is 7.9s and slot duration is 4s, then we only wait 1 slot, instead of waiting for 2 slots.

The failing tests usually take 4s rather than 8s, which might be justified by this.

The PR removes the calculation and passes the duration to the get_events function directly.

PR #7253, opened 2025-04-03
feat: better error for light_client/bootstrap endpoint

### LINKED ISSUE #7229: Add better error messaging to the `light_client/bootstrap` endpoint
Per the spec, it is required to serve light client bootstrap objects from finalized checkpoint block roots. Serving bootstrap objects for all other blocks is optional. Lighthouse only supports serving bootstrap objects for finalized checkpoint block roots. Currently, if a block root is used that isn't a finalized checkpoint block root the following error is returned

```
Sync committee branch for block root 0x7a734cd6e35b6c21f0cf55a8e3c36331a54f9774dface476be3b60d47e15f09a not found
```

We should instead return a better error message that informs the user that the provided block root is not a finalized checkpoint block root

varun-doshi:
## Issue Addressed

Closes #7229 

## Proposed Changes

Update error message to be more meaningful for user and provider reason for unavailbility

### DIFF
--- a/beacon_node/beacon_chain/src/light_client_server_cache.rs
+++ b/beacon_node/beacon_chain/src/light_client_server_cache.rs
@@ -374,7 +374,7 @@ impl<T: BeaconChainTypes> LightClientServerCache<T> {
         let Some(current_sync_committee_branch) = store.get_sync_committee_branch(block_root)?
         else {
             return Err(BeaconChainError::LightClientBootstrapError(format!(
-                "Sync committee branch for block root {:?} not found",
+                "Block Root {:?} is not a finalized checkpoint root. Sync Committee branch only availble for finalized checkpoint block roots.",
                 block_root
             )));
         };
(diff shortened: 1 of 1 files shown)


beacon_node/beacon_chain/src/light_client_server_cache.rs
```diff
@@ -374,7 +374,7 @@ impl<T: BeaconChainTypes> LightClientServerCache<T> {
         let Some(current_sync_committee_branch) = store.get_sync_committee_branch(block_root)?
         else {
             return Err(BeaconChainError::LightClientBootstrapError(format!(
-                "Sync committee branch for block root {:?} not found",
+                "Block Root {:?} is not a finalized checkpoint root. Sync Committee branch only availble for finalized checkpoint block roots.",
```
michaelsproul: There are actually multiple reasons why a sync committee branch for a block root may not exist. E.g. it might be that the light client server wasn't running when that block was processed, and so no update was computed for it.

Maybe we could say something like: `Sync committee branch for block root {:?} not found. It might not be a checkpoint block`

michaelsproul (changes requested):

PR #7258, opened 2025-04-04
Fullnodes to publish data columns from EL `getBlobs`

jimmygchen:
## Issue Addressed

Previously only supernode contributes to data column publishing in Lighthouse. 

Recently we've [updated the spec](https://github.com/ethereum/consensus-specs/pull/4183) to have full nodes publishing data columns as well, to ensure all nodes contributes to propagation. 

This also prevents already imported data columns from being imported again (because we don't "observe" them), and ensures columns that are observed in the [gossip seen cache](https://github.com/sigp/lighthouse/blob/d60c24ef1cc0b5dfa930e1dd4fc85abc29e5fc4c/beacon_node/beacon_chain/src/data_column_verification.rs#L492) are forwarded to its peers, rather than being ignored.


pawanjay176 (approved):
Nice!

beacon_node/network/src/network_beacon_processor/mod.rs
```diff
@@ -927,20 +927,20 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
                     BlobsOrDataColumns::Blobs(blobs) => {
                         self_cloned.publish_blobs_gradually(blobs, block_root);
                     }
                     BlobsOrDataColumns::DataColumns(columns) => {
-                        self_cloned.publish_data_columns_gradually(columns, block_root);
+                        let columns_to_publish = columns
+                            .into_iter()
+                            .filter(|c| custody_columns.contains(&c.index))
+                            .collect();
+                        self_cloned.publish_data_columns_gradually(columns_to_publish, block_root);
```
macladson: One small optimization here might be to skip the filtering if `self.network_globals.is_supernode()`?

jimmygchen: Updated in https://github.com/sigp/lighthouse/pull/7258/commits/0c60a365a1c15fd9739153fe67a1f88d21be3697

I was a bit tempted to remove the `is_supernode` function - with the introduction of validator custody, some full nodes now performs the same tasks as supernodes - including distributed blob publishing and reconstruction, but I'll leave this change out of this PR.

jimmygchen: With validator custody, this becomes a bit complicated - the node should be publishing the columns that it imported into its database - so we need to be using the `custody_group_count` tied to the Rpc or gossip verified beacon block instead of using the `NetworkGlobals`, because CGC is now dynamic:
- https://github.com/sigp/lighthouse/blob/6f31d44343086ddaa70b396632c0eee724256454/beacon_node/beacon_chain/src/block_verification.rs#L686
- https://github.com/sigp/lighthouse/blob/6f31d44343086ddaa70b396632c0eee724256454/beacon_node/beacon_chain/src/block_verification_types.rs#L34

However, there are a few moving pieces and this wiring hasn't been fully set up, I think it's best to wait until the following PR is completed:
- https://github.com/sigp/lighthouse/pull/7117
- https://github.com/sigp/lighthouse/pull/7262

eserilev (approved):
LGTM

PR #7262, opened 2025-04-04
Implement basic validator custody

dapplion:
## Issue Addressed

- Implements https://github.com/sigp/lighthouse/issues/6767 except the backfill restart

Also does a consistency check on the DB to make sure the stored columns are compatible

Not tested yet

## Proposed Changes

TODO


beacon_node/network/src/service.rs
```diff
@@ -943,6 +949,49 @@ impl<T: BeaconChainTypes> NetworkService<T> {
...
+                < MAX_SLOT_DISTANCE_BACKFILL_RESTART;
+            // Note: we don't check if backfill finished. If it did because we are close to genesis,
+            // we want to restart it anyway to backfill with the CGC. The only condition to NOT
+            // restart is if backfill went too far and thus we would waste too much bandwidth
+            // fetching the blocks again.
+            if backfill_started_recently {
+                // We need backfill sync to fetch batches with `CGC_f = cgc_at_finalized_slot`. Then
+                // `custody_group_count(oldest_block_slot) should now return `CGC_f`. So we have to
+                // delete the CGC updates with `update.slot < finalized_slot`
+                todo!();
```
michaelsproul: To prevent concurrent updates of `oldest_block_slot`, we can get `import_historical_block_batch` to handle the mutation for us.

We can add an argument to that function, something like `restart: bool`, and set the `oldest_block_{slot,root}` to the _last_ block in the batch passed in (allowing the `oldest_block_slot` to _go up_).

TODO: what if state reconstruction is running as well?

- Need to consider how state reconstruction depends on and modifies `oldest_block_slot`. Should maybe make `as_archive_anchor()` preserve the oldest block slot

Alternatively, make sure we only restart backfill if `oldest_block_slot > 0` (might affect some testnets with `FULU_FORK_EPOCH=0`, but not mainnet). Actually we need to consider `max(oldest_block_slot, fulu_fork_slot, clock_slot - da_window)`

PR #7269, opened 2025-04-07
Bump version to v7.1.0-beta.0 (not a release)

michaelsproul:
## Proposed Changes

Having merged the drop-headtracker PR we now have a DB schema change in `unstable` compared to `release-v7.0.0`:

- https://github.com/sigp/lighthouse/pull/6744

There is a DB downgrade available, however this needs to be applied manually and it's usually a bit of a hassle.

This PR bumps the version on `unstable` to `v7.1.0-beta.0` _without_ actually cutting a `v7.1.0-beta.0` release, so that we can tell at a glance which schema version a node is using.

## Additional Info

Our next release with hot tree-states _can_ be v7.1.0 rather than v8.0.0 because we have a schema downgrade for this change as well:

- https://github.com/sigp/lighthouse/pull/6750

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -808,7 +808,7 @@ dependencies = [
 
 [[package]]
 name = "beacon_node"
-version = "7.0.0-beta.5"
+version = "7.1.0-beta.0"
 dependencies = [
  "account_utils",
  "beacon_chain",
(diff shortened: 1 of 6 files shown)


jimmygchen (approved):
LGTM!

michaelsproul:
I'm not sure why that network test failed. Looks a bit sus.

PR #7271, opened 2025-04-07
Upgrade to c-kzg 2.1.0 and alloy-primitives 1.0

cakevm:
Status: Sadly upgrading `c-kzg` brings a performance degradation that let the simulator CI job failing. Increasing `GENESIS_DELAY` helps, but I wonder why this is happening now.

## Issue Addressed

Update `c-kzg` from `v1` to `v2`. My motivation here is that `alloy-consensus` now uses `c-kzg` in `v2` and this results in a conflict when using lighthouse in combination with latest alloy. I tried also to disable the `czkg` feature in alloy, but the conflict persisted.

See here for the alloy update to `c-kzg v2`:  https://github.com/alloy-rs/alloy/pull/2240

Error:
```
error: failed to select a version for `c-kzg`.
...
versions that meet the requirements `^1` are: 1.0.3, 1.0.2, 1.0.0

the package `c-kzg` links to the native library `ckzg`, but it conflicts with a previous package which links to `ckzg` as well:
package `c-kzg v2.1.0`
    ... which satisfies dependency `c-kzg = "^2.1"` of package `alloy-consensus v0.13.0`
    ... which satisfies dependency `alloy-consensus = "^0.13.0"` of package ...
...
```

## Proposed Changes

- Upgrade `alloy-consensus` to `0.14.0` and disable all default features
- Upgrade `c-kzg` to `v2.1.0`
- Upgrade `alloy-primitives` to `1.0.0`
- Adapt the code to the new API `c-kzg`
- There is now `NO_PRECOMPUTE` as my understand from https://github.com/ethereum/c-kzg-4844/pull/545/files we should use `0` here as `new_from_trusted_setup_no_precomp` does not precomp. But maybe it is misleading. For all other places I used `RECOMMENDED_PRECOMP_WIDTH` because `8` is matching the recommendation.
- `BYTES_PER_G1_POINT` and `BYTES_PER_G2_POINT` are no longer public in `c-kzg`
- I adapted two tests that checking for the `Attestation`  bitfield size. But I could not pinpoint to what has changed and why now 8 bytes less. I would be happy about any hint, and if this is correct. I found related a PR here: https://github.com/sigp/lighthouse/pull/6915
- Use same fields names, in json, as well as `c-kzg` and `rust_eth_kzg` for `g1_monomial`, `g1_lagrange`, and `g2_monomial`

## Additional Info

- The copy and flattening into the data structure for the `KzgSettings` seems suboptimal. The whole data structure could be already prepared for this like in `c-kzg`.
- It seem that the `c-kzg` lib would provide already those constants (e.g. with `c_kzg::ethereum_kzg_settings`) and maybe the `TrustedSetup` could be simplified in the future (e.g. removing it and provide the user to load custom `KzgSettings` from json.

PS: I did not tested this with a real network, I hope that the test coverage from other crates would highlight if kzg is broken with this change. I just compared by hand that (e.g. `g1_monomial`) is correctly read. I am open to add test for the correct loading of the `kzg`, but maybe it make sense to seek to a deeper integration of `c-kzg` in another step, because it requires much more changes. Or it will be removed with this issue here: https://github.com/sigp/lighthouse/issues/6107#issuecomment-2632542447


consensus/types/src/attestation.rs
```diff
@@ -601,12 +601,12 @@ mod tests {
         let attestation_data = size_of::<AttestationData>();
         let signature = size_of::<AggregateSignature>();
 
-        assert_eq!(aggregation_bits, 152);
+        assert_eq!(aggregation_bits, 144);
```
pawanjay176, line 664: I have no idea why this would change because of a dependency update.
cc @paulhauner

crypto/kzg/benches/benchmark.rs
```diff
@@ -25,8 +25,13 @@ pub fn bench_init_context(c: &mut Criterion) {
                 serde_json::from_reader(get_trusted_setup().as_slice())
                     .map_err(|e| format!("Unable to read trusted setup file: {}", e))
                     .expect("should have trusted setup");
-            KzgSettings::load_trusted_setup(&trusted_setup.g1_points(), &trusted_setup.g2_points())
-                .unwrap()
+            KzgSettings::load_trusted_setup(
+                &trusted_setup.g1_monomial(),
+                &trusted_setup.g1_lagrange(),
+                &trusted_setup.g2_monomial(),
+                rust_eth_kzg::constants::RECOMMENDED_PRECOMP_WIDTH as u64,
```
pawanjay176: Not sure why we are using a constant from rust_eth_kzg as an input parameter for c-kzg settings. I would prefer to have a different constant defined if c-kzg does not expose this. Maybe use `NO_PRECOMPUTE`?

crypto/kzg/src/lib.rs
```diff
@@ -85,8 +94,10 @@ impl Kzg {
 
         Ok(Self {
             trusted_setup: KzgSettings::load_trusted_setup(
-                &trusted_setup.g1_points(),
-                &trusted_setup.g2_points(),
+                &trusted_setup.g1_monomial(),
+                &trusted_setup.g1_lagrange(),
+                &trusted_setup.g2_monomial(),
+                rust_eth_kzg::constants::RECOMMENDED_PRECOMP_WIDTH as u64,
```
pawanjay176: same as above. would prefer to use `NO_PRECOMPUTE` here.

pawanjay176 (dismissed):
Looking good
Not sure how urgently we want this to be in unstable. I would prefer not to have this in any of the stable releases that go into the pectra releases for now

crypto/kzg/benches/benchmark.rs
cakevm: You are absolute right.I checked again and could not find a constant in `c-kzg` for this value.  I introduced for that reason an own constant value for the recommend precompute setting.

crypto/kzg/src/lib.rs
cakevm: Same resolution as above.

cakevm:
Thank you for taking the time to have a look. It’s not a priority. I came across this as a transient dependency and initially thought updating the library would be a quick win. But given the issues that have come up, that doesn’t seem to be the case. I’m fine with either closing this PR or leaving it open until there’s a decision on whether c-kzg will be removed.

consensus/types/src/attestation.rs
michaelsproul: I think smallvec sometimes implements changes to its layout and size between versions, so this is fine

michaelsproul:
I'm going to try a narrower and more targeted `cargo update`, as the `Cargo.lock` file is currently very noisy and hard to review.

crypto/kzg/src/lib.rs
```diff
@@ -21,6 +21,20 @@ pub use rust_eth_kzg::{
...
+///
+/// See: <https://github.com/ethereum/c-kzg-4844/pull/545/files>
+pub const PRECOMPUTE: u64 = 8;
+
+/// Disables the fixed-base multi-scalar multiplication optimization for computing
+/// cell KZG proofs, as recommended when `compute_cells_and_kzg_proofs` and
+/// `recover_cells_and_kzg_proofs` are not used.
+///
+/// See: <https://github.com/ethereum/c-kzg-4844/pull/545/files>
+pub const NO_PRECOMPUTE: u64 = 0;
```
jimmygchen, line 29: Note that we don't currently use `c-kzg4844` for PeerDAS, so the methods above from c-kzg aren't being used. We are using the `rust-eth-kzg` for PeerDAS support, and we're already doing precomputation there:
https://github.com/sigp/lighthouse/blob/6ee8a7df9a2be8b072e673b26c3a04be78ab9493/crypto/kzg/src/lib.rs#L95-L100

We should be able to just use `NO_PRECOMPUTE` for c-kzg until we decide to support using c-kzg for PeerDAS (https://github.com/sigp/lighthouse/issues/6107). Otherwise we end up doing precomputation for both libraries, and it would slow down startup time and double the memory usage to cache the precomputation results.

michaelsproul: Good catch! The simulator tests are failing, and it's probably because of the startup time regression.

jimmygchen: Right, from the link provided, it looks like it still takes 1.69s to precompute, even when precompute is set to 0:
https://github.com/ethereum/c-kzg-4844/pull/545/files

I think it would be ok to bump up genesis delay slightly to make this work.

cakevm: @jimmygchen Awesome! I absolutely missed that. Thank you for the insights about the different kzg libs. I adapted it to use always `NO_PRECOMPUTE` and increased the `GENESIS_DELAY`. Using now `38` instead of `32` seems to be the sweet spot.

cakevm:
I did now some research and guess I have maybe found something, that could cause this slower runtime in the tests with version 2. It looks like that the initialization behavior has changed to always init FK20 and support cell operations, which would not be required here. But lead to a longer init time. For more details see: https://github.com/ethereum/c-kzg-4844/issues/580

michaelsproul:
I would still like to merge this

michaelsproul (approved):

michaelsproul:
@mergify requeue

michaelsproul:
I don't see why it failed, retrying

Issue #7278, opened 2025-04-08
PeerDAS internal availability check failure


jimmygchen:
We're seeing the following errors on on a `peerdas-devnet-5` node `lighthouse-geth-1 `:

```
Apr 04 09:43:49.680 WARN Internal availability check failure           block_root: 0xc98c2fca6e57d239fd4ad65cd09a8113f1e85608758c0303
89d1914b671b7e41, error: Unexpected("too many columns got 128 expected 0"), service: "lookup_sync"
```

This is a supernode, but its expecting 0 columns in lookup sync. I think it's because `RpcBlock` from lookup is not coupled with columns when we instantiate an RpcBlock with block variant here:
https://github.com/sigp/lighthouse/blob/7a0388ef2aa8ea6de5cfa0be26dd000f011a6484/beacon_node/network/src/sync/block_lookups/common.rs#L103

This could also happen via the HTTP duplicate block import path, or whenever we call `new_without_blobs` where blobs are expected:
https://github.com/sigp/lighthouse/blob/0e6da0fcafe25f7532c9ee373df405c87e568802/beacon_node/beacon_chain/src/block_verification.rs#L1277

In these scenarios, the node would not be able to import the block.

Related PR: 
- https://github.com/sigp/lighthouse/pull/7033

jimmygchen:
Fixed in #7281

PR #7281, opened 2025-04-08
Fix wrong custody column count for lookup blocks

dapplion:
## Issue Addressed

Fixes
- https://github.com/sigp/lighthouse/issues/7278

## Proposed Changes

Don't assume 0 columns for `RpcBlockInner::Block`

### DIFF
--- a/beacon_node/beacon_chain/src/block_verification.rs
+++ b/beacon_node/beacon_chain/src/block_verification.rs
@@ -1272,9 +1272,16 @@ impl<T: BeaconChainTypes> IntoExecutionPendingBlock<T> for Arc<SignedBeaconBlock
         // Perform an early check to prevent wasting time on irrelevant blocks.
         let block_root = check_block_relevancy(&self, block_root, chain)
             .map_err(|e| BlockSlashInfo::SignatureNotChecked(self.signed_block_header(), e))?;
+        // TODO(das): This is wrong, as the block may have columns. However this codepath is
+        // currently un-used, and should be removed with https://github.com/sigp/lighthouse/pull/7008
+        let custody_column_count = 0;
         let maybe_available = chain
             .data_availability_checker
-            .verify_kzg_for_rpc_block(RpcBlock::new_without_blobs(Some(block_root), self.clone()))
+            .verify_kzg_for_rpc_block(RpcBlock::new_without_blobs(
+                Some(block_root),
+                self.clone(),
+                custody_column_count,
+            ))
             .map_err(|e| {
                 BlockSlashInfo::SignatureNotChecked(
                     self.signed_block_header(),
--- a/beacon_node/beacon_chain/src/block_verification.rs
+++ b/beacon_node/beacon_chain/src/block_verification.rs
@@ -1260,47 +1260,6 @@ impl<T: BeaconChainTypes> IntoExecutionPendingBlock<T> for SignatureVerifiedBloc
     }
 }
 
-impl<T: BeaconChainTypes> IntoExecutionPendingBlock<T> for Arc<SignedBeaconBlock<T::EthSpec>> {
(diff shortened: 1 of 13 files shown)


beacon_node/beacon_chain/src/block_verification_types.rs
```diff
@@ -103,14 +103,15 @@ impl<E: EthSpec> RpcBlock<E> {
...
         block_root: Option<Hash256>,
         block: Arc<SignedBeaconBlock<E>>,
+        custody_columns_count: usize,
     ) -> Self {
         let block_root = block_root.unwrap_or_else(|| get_block_root(&block));
 
         Self {
             block_root,
             block: RpcBlockInner::Block(block),
             // Block has zero columns
```
jimmygchen: This comment is no longer relevant
```suggestion
```

jimmygchen (approved):
Looks good to me, this is a beautiful quick fix 🎉

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -1260,40 +1260,6 @@ impl<T: BeaconChainTypes> IntoExecutionPendingBlock<T> for SignatureVerifiedBloc
     }
 }
 
-impl<T: BeaconChainTypes> IntoExecutionPendingBlock<T> for Arc<SignedBeaconBlock<T::EthSpec>> {
```
jimmygchen, line 1263: I think this is a good cleanup given the logic mostly duplicates with the `RpcBlock` impl - internally it creates an `RpcBlock` anyway - and we don't really use this much in production code path.

Only usage removed here: https://github.com/sigp/lighthouse/pull/7281/files#diff-6a53505befab4a489b4eca65f7141bdee09e1044b7d63dadc1d18e708cd06291R305

beacon_node/beacon_chain/src/test_utils.rs
```diff
@@ -2371,7 +2371,11 @@ where
 
         // Blobs are stored as data columns from Fulu (PeerDAS)
         if self.spec.is_peer_das_enabled_for_epoch(block.epoch()) {
-            let columns = self.chain.get_data_columns(&block_root).unwrap().unwrap();
+            let columns = self
+                .chain
+                .get_data_columns(&block_root)
+                .unwrap()
+                .or_default();
```
jimmygchen: I think we cant default this - in this case the node should have columns in the db because we checked the block has blobs.

jimmygchen: Fixed in https://github.com/sigp/lighthouse/pull/7281/commits/bbbfa9476d5eff5b576a70d7068103ccddcf5d92

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -266,7 +266,8 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
                 )
                 .map_err(|e| format!("{e:?}"))?
             } else {
-                RpcBlock::new_without_blobs(Some(block_root), block)
+                // Block has no data, expects zero columns
+                RpcBlock::new_without_blobs(Some(block_root), block, 0)
```
jimmygchen, line 270: I realised this should be fine if we always send the `custody_columns_count`, regardless of whether the block has blobs, because DA checker only uses it if it has blobs.

jimmygchen:
@dapplion please review https://github.com/sigp/lighthouse/pull/7281/commits/b9a9e84f301f3059f9f8494ea3033d02239cdb24, CI should pass now

dapplion:
> @dapplion please review [b9a9e84](https://github.com/sigp/lighthouse/commit/b9a9e84f301f3059f9f8494ea3033d02239cdb24), CI should pass now

✅ looks good to me

PR #7292, opened 2025-04-09
Test force docker pull python image

jimmygchen:
## Issue Addressed

Test pulling python image to fix Kurtosis tests. Do not merge.


jimmygchen (commented):
Update python tag

jimmygchen:
Should be fixed in latest ethereum package
https://github.com/ethpandaops/ethereum-package/pull/957

PR #7293, opened 2025-04-09
Always run build.rs for eth2_network_config

diegomrsantos:
## Issue Addressed

When eth2_network_config is cached,  the build script doesn't run, and the genesis file can't be found. See https://github.com/sigp/anchor/pull/240

## Proposed Changes

Make the build script always run using `println!("cargo:rerun-if-changed=non_existent_file");`

### DIFF
--- a/common/eth2_network_config/build.rs
+++ b/common/eth2_network_config/build.rs
@@ -7,6 +7,9 @@ use std::io;
 use zip::ZipArchive;
 
 fn main() {
+    // This file doesn't exist, so Cargo treats it as changed.
+    println!("cargo:rerun-if-changed=non_existent_file");
+
     for network in ETH2_NET_DIRS {
         match uncompress_state(network) {
             Ok(()) => (),


michaelsproul:
I'm worried this will mess with compile times for devs building locally. Rebuilding _every_ run seems extreme.

Isn't the upstream anchor issue just a bug in anchor's cache? i.e. needing to cache the genesis.ssz files as well. Maybe we should move them to the `target` directory

diegomrsantos:
> I'm worried this will mess with compile times for devs building locally. Rebuilding _every_ run seems extreme.
> 
> Isn't the upstream anchor issue just a bug in anchor's cache? i.e. needing to cache the genesis.ssz files as well. Maybe we should move them to the `target` directory

Does building the build script force a package build? We use the cache option available in the github action, it's not a custom cache.

diegomrsantos:
Reading https://doc.rust-lang.org/cargo/reference/build-scripts.html#change-detection, I understand it should just compile and execute the build script when building the package.

diegomrsantos:
But moving them to the target directory could be a better solution

diegomrsantos:
@michaelsproul should I close it?

michaelsproul:
@diegomrsantos Maybe close in favour of the `target` dir approach? Could you do some testing to ensure that the crate doesn't get rebuilt every time as well? It's not completely clear to me from the Cargo docs

We have a similar issue with the `lighthouse_version` crate. It seems to rebuild unnecessarily sometimes, I think because it contains the commit hash, and the repo's dirty state

michaelsproul:
Closing in favour of target dir approach

Issue #7295, opened 2025-04-09
Schedule Gnosis Pectra hard-fork


dapplion:
Schedule Pectra for Gnosis, to activate on Apr-30-2025 14:03:40 UTC (~1 week before Ethereum). Having releases by April 23rd (Ethereum planned deadline) is okay, but we will communicate to users if you choose to release earlier.

- Upstream schedule PR https://github.com/gnosischain/specs/pull/74
- Full config https://github.com/gnosischain/specs/blob/master/consensus/config/gnosis.yaml
- Config diff against Ethereum https://github.com/gnosischain/specs/blob/master/beacon_chain.md#config-diff

chong-he:
I would like to give this a go, also similar to #7274

chong-he:
The Gnosis Pectra fork config has been added in #7296

Issue #7274, opened 2025-04-08
Gnosis Electra epoch


michaelsproul:
We need the Gnosis Electra epoch.

The Gnosis chain devs tell me this will be available Soon (TM).

michaelsproul:
Closing in favour of:

- https://github.com/sigp/lighthouse/issues/7274

Which has more detail.

PR #7296, opened 2025-04-09
Gnosis Pectra fork epoch

chong-he:
## Issue Addressed

~~* #7274~~ 
* #7295


common/eth2_network_config/built_in_network_configs/gnosis/config.yaml
```diff
@@ -120,6 +120,18 @@ BLOB_SIDECAR_SUBNET_COUNT: 6
 # `uint64(6)`
 MAX_BLOBS_PER_BLOCK: 6
```
chong-he: Not sure about this? 

From the spec, it's showing 2: https://github.com/gnosischain/specs/blob/de1d51759c8bb786c7172740c66ec50cd8cb8d1d/consensus/config/gnosis.yaml#L140

consensus/types/src/chain_spec.rs
```diff
@@ -1258,7 +1258,7 @@ impl ChainSpec {
             })
             .expect("calculation does not overflow"),
             max_per_epoch_activation_exit_churn_limit: option_wrapper(|| {
-                u64::checked_pow(2, 8)?.checked_mul(u64::checked_pow(10, 9)?)
+                u64::checked_pow(2, 6)?.checked_mul(u64::checked_pow(10, 9)?)
```
chong-he, line 1261: I think I have to change this according to: https://github.com/gnosischain/specs/blob/de1d51759c8bb786c7172740c66ec50cd8cb8d1d/consensus/config/gnosis.yaml#L147

common/eth2_network_config/built_in_network_configs/gnosis/config.yaml
michaelsproul: Yeah we should update this to 2. Lion explained that the EL sets a lower limit, so we are not buggy at the moment, but we should use 2 pre-Electra

dapplion: it's changing from 6 to 2. It's strange because it's a live constant, but at the same time it's enforced by the EL. A proposer would have to waste a block proposal to create a chance of peers downscoring each other if the run distinct versions of the value.

chong-he: Updated in [3e77541](https://github.com/sigp/lighthouse/pull/7296/commits/3e775411ca733b4787e0f2e1ec8b82f752b2227e)

michaelsproul:
CK do you mind copying across the changes in `electra.yaml` from here: https://github.com/gnosischain/specs/blob/de1d51759c8bb786c7172740c66ec50cd8cb8d1d/consensus/preset/gnosis/electra.yaml

into `consensus/types/presets/gnosis`?

chong-he:
> CK do you mind copying across the changes in `electra.yaml` from here: https://github.com/gnosischain/specs/blob/de1d51759c8bb786c7172740c66ec50cd8cb8d1d/consensus/preset/gnosis/electra.yaml
> 
> into `consensus/types/presets/gnosis`?

The only difference is [this first line below](https://github.com/gnosischain/specs/blob/de1d51759c8bb786c7172740c66ec50cd8cb8d1d/consensus/preset/gnosis/electra.yaml#L32):
```
# `uint64(2**0)` (= 1)
MAX_CONSOLIDATION_REQUESTS_PER_PAYLOAD: 2
```

which the current LH branch looks correct? https://github.com/sigp/lighthouse/blob/aed562abef148e64ef119ea5400e3e06055e1285/consensus/types/presets/gnosis/electra.yaml#L32-L33
```
# `uint64(2**1)` (= 2)
MAX_CONSOLIDATION_REQUESTS_PER_PAYLOAD: 2
```

If `MAX_CONSOLIDATION_REQUESTS_PER_PAYLOAD: 2` is 2, then it should be 2**1 = 2?

michaelsproul:
@chong-he, good catch, looks like the comment in the upstream Gnosis preset is wrong then

michaelsproul (approved):

michaelsproul:
@mergify requeue

Issue #7294, opened 2025-04-09
`/eth/v2/beacon/pool/attestations` does not honor `committee_index`


mcdee:
If the `/eth/v2/beacon/pool/attestations` endpoint is passed no `committee_index` then it returns attestations from all committees.

If it is passed `committee_index=0` it returns attestations from all committees.

If it is passed `committee_index=n` for any `n>0` it returns nothing.

These last two situations are non-spec and I believe are bugs.

```
lighthouse --version
Lighthouse v7.0.0-beta.5-6d5a2be
BLS library: blst-portable
BLS hardware acceleration: true
SHA256 hardware acceleration: true
Allocator: jemalloc (4K)
Profile: maxperf
Specs: mainnet (true), minimal (false), gnosis (true)
```

michaelsproul:
Yep good catch.

This filter needs updating:

https://github.com/sigp/lighthouse/blob/bf955c7543dac8911a6f6c334b5b3ca4ef728d9c/beacon_node/http_api/src/lib.rs#L1880-L1885

We'll need to change the closure type to take `&Attestation` rather than `&AttestationData` so we can access the committee index post-Electra.

Needs a regression test too in the HTTP API crate.

eserilev:
im on it

eserilev:
Closed in [#7298](https://github.com/sigp/lighthouse/pull/7298)

PR #7298, opened 2025-04-09
Ensure `/eth/v2/beacon/pool/attestations` honors `committee_index`

eserilev:
## Issue Addressed

#7294

## Proposed Changes

Fix the filtering logic so that we actually filter by committee index for both `Base` and `Electra` attestations.

Added a tiny optimization when calculating committee_index to prevent unneeded memory allocations

Added a regression test


beacon_node/http_api/src/lib.rs
```diff
@@ -1999,11 +2000,14 @@ pub fn serve<T: BeaconChainTypes>(
...
+                    let query_filter = |data: &AttestationData, committee_index: Option<u64>| {
                         query.slot.is_none_or(|slot| slot == data.slot)
-                            && query
-                                .committee_index
-                                .is_none_or(|index| index == data.index)
+                            && query.committee_index.is_none_or(|index| {
+                                if let Some(committee_index) = committee_index {
+                                    return index == committee_index;
+                                }
+                                false
```
michaelsproul: this is `committee_index.is_some_and(|committee_index| index == committee_index)`

beacon_node/operation_pool/src/attestation_storage.rs
```diff
@@ -119,6 +119,18 @@ impl<E: EthSpec> CompactAttestationRef<'_, E> {
...
     }
 
+    pub fn committee_index(&self) -> Option<u64> {
+        match self.indexed {
+            CompactIndexedAttestation::Base(_) => Some(self.data.index),
+            CompactIndexedAttestation::Electra(indexed_att) => indexed_att
+                .committee_bits
+                .iter()
+                .enumerate()
+                .find(|&(_, bit)| bit)
```
michaelsproul: I think we need to account for aggregation here. Attestations in the op pool can cover multiple committee indices and if _any_ of these match our query index, we should return that attestation.

The aggregation happens in `aggregate_across_committees`, which persists the aggregates in the attestation storage:

https://github.com/sigp/lighthouse/blob/bf955c7543dac8911a6f6c334b5b3ca4ef728d9c/beacon_node/operation_pool/src/attestation_storage.rs#L335

We could use the `committee_indices` vec and check `committee_indices.contains(&query.committee_index)`? We should also add a test covering this case (might need to call `aggregate_across_committees` to prompt aggregation).

(We should maybe remove the `committee_index` methods here so we don't accidentally misuse them)

michaelsproul (changes requested):

beacon_node/operation_pool/src/attestation_storage.rs
eserilev: One of the issues with out test harness is that we only have attestations across a single committee. so the aggregation isnt doing much. I've added it anyways, but I had to make a field public which isnt the best solution

We should clean this up and get the test harness to create attestations across more than a single committee. I just didnt want to block v7 testing because of this. Let me know what you think

michaelsproul: Oh yeah good point. We can come back to test this more thoroughly, I've opened an issue:

- https://github.com/sigp/lighthouse/issues/7312

macladson (approved):

PR #7300, opened 2025-04-09
Downgrade light client errors

eserilev:
## Issue Addressed

Downgrade light client errors to debug

Error messages are alarming and usually indicate somethings wrong with the beacon node. The Light Client service is supposed to minimally impact users, and most will not care if the light client server is erroring. Furthermore, the only errors we've seen in the wild are during hard forks, for the first few epochs before the fork finalizes.


michaelsproul (approved):
Agree this is a good temporary fix before v7.0.0 until we can repro the fork failures.

PR #7303, opened 2025-04-10
Update and cleanup Electra preset

michaelsproul:
## Proposed Changes

- Update the bundled `electra.yaml` preset files for `mainnet` and `minimal` to match `consensus-specs` as of: https://github.com/ethereum/consensus-specs/commit/bf09edef17e2900258f7e37631e9452941c26e86
- Add the field `max_pending_deposits_per_epoch` to our runtime representation of `ElectraPreset`. This results in it appearing in `/eth/v1/config/spec` where it was previously absent.

## Additional Info

I have not updated the Gnosis preset. That is happening as part of:

- https://github.com/sigp/lighthouse/pull/7296


macladson (approved):
LGTM!

Issue #7203, opened 2025-03-26
Data column gossip verification taking more than 4s when blob count is above 20


jimmygchen:
In the latest PeerDAS devnet testing report published by Sunnyside Lab:
https://testinprod.notion.site/Sunnyside-Devnet-Updates-03-25-Internal-1c08fc57f54680339b2fe2dbc8232b6d

Findings show that the network degraded when blobs per block is above 20, and Lighthouse's data column verification time increases quite significantly, taking more than 4s, causing the network to breakdown:

<img width="687" alt="Image" src="https://github.com/user-attachments/assets/3f08fa08-d24d-4537-a51c-f8ffbadab476" />

Might be worth checking these metrics as well:
- `beacon_kzg_verification_data_column_batch_seconds`
- `beacon_kzg_verification_data_column_single_seconds`

jimmygchen:
I looked at the following verification metric, and got quite different numbers as the report. I'll reach out to Sunnyside Lab to confirm the metric they're looking at.
 - `histogram_quantile(0.50, sum(rate(beacon_data_column_sidecar_gossip_verification_seconds_bucket[5m])) by (le))`


<img width="779" alt="Image" src="https://github.com/user-attachments/assets/ae4513e0-37db-43af-8adb-e6ca448d24d1" />

jimmygchen:
In the test the 95th percentile was used (instead of 50th), and showing very different results:
> I used `histogram_quantile(0.95, rate(beacon_data_column_sidecar_gossip_verification_seconds_bucket{network=~'$network', execution_client=~'$execution', consensus_client=~'$consensus'}[$__rate_interval]))` so that we could see the data from each series. If we look in this data individually, full node(have even numbers like `lighthouse-geth-2`) have higher runtime than supernodes (with odd numbers)

<img width="761" alt="Image" src="https://github.com/user-attachments/assets/68bb2cd5-8c57-4375-bb3f-a97d4dcb1a30" />

pawanjay176:
I can take a look

jimmygchen:
Thanks @pawanjay176!

jimmygchen:
@pawanjay176 
I was thinking if the following issue may be related - multiple concurrent writes holding up the fork choice write lock
https://github.com/sigp/lighthouse/issues/6439

jimmygchen:
[...]

Issue #4447, opened 2023-06-29
Getting shuffling efficiently in gossip blobs signature verification


realbigsean:
## Description

If we get a cache miss in the `beacon_proposer_cache`, we need a way to get the state which provides the shuffling for the blob. In block processing we fall back to the snapshot cache, and then remove the state from the snapshot cache (in the normal case), and attach it to the block for use during block processing. 

This means if we're getting the block first and getting a proposer cache miss, when verifying blobs we'd be likely to load the state and parent block from the db when we shouldn't have to because they're in memory.  We do prime the`beacon_proposer_cache` after finding shuffling for the block, but there's some raciness there with blob arrival/verification.

jimmygchen:
This is the relevant WARN log we're getting:
> Aug 21 01:27:12.431 WARN Snapshot cache miss for blob verification, index: 3, block_root: 0xfd7c…a70a, service: beacon

@realbigsean mentioned that `tree-states` actually fixes this issue, but it's likely to be merged after deneb, so it would be good to fix this now. I'll take a deeper look.

jimmygchen:
Completed in #4646 . Tested this on devnet-8, sent a few blobs and didn't see any cache miss 🎉

PR #7304, opened 2025-04-10
Compute proposer shuffling only once in gossip verification

jimmygchen:
## Issue Addressed

When we perform data column gossip verification, we sometimes see multiple proposer shuffling cache miss simultaneously and this results in multiple threads computing the shuffling cache and potentially slows down the gossip verification. 

Proposal here is to use a `OnceCell` for each shuffling key to make sure it's only computed once. I have only implemented this in data column verification as a PoC, but this can also be applied to blob and block verification

Related issues:
- https://github.com/sigp/lighthouse/issues/4447
- https://github.com/sigp/lighthouse/issues/7203


jimmygchen:
I've pushed a PoC, however I think this is still problematic because we hold the lock for the entire duration of the computation.

jimmygchen:
Pushed another reworked version to avoid lock contention mentioned above @pawanjay176

eserilev (approved):
Nice!

jimmygchen:
Thanks @eserilev !

@pawanjay176 I wrote a [test](https://github.com/jimmygchen/lighthouse/commit/a811daa26c9a42036afb64e69b7e95e333ccb548) to measure time for each step in column verification with some numbers for each step, and comparison with `unstable`. This seems to work and could potentially perform better when columns are received in a staggered way.

## Single column verification

This single test (7 blobs) shows that `verify_proposer_and_signature` is the only step taking significant amount of time ~500ms. The rest are pretty negligible.

```
verify_data_column_sidecar: 29.458µs
verify_index_matches_subnet: 2.125µs
verify_sidecar_not_from_future_slot: 8.041µs
verify_slot_greater_than_latest_finalized_slot: 10.5µs
verify_is_first_sidecar: 3.75µs
verify_column_inclusion_proof: 218.125µs
verify_parent_block_and_finalized_descendant: 33.083µs
verify_slot_higher_than_parent: 750ns
verify_proposer_and_signature: 519.821916ms
verify_kzg_for_data_column: 4.024208ms
observe_slashable: 54.709µs
All verifications passed successfully
verify_total_time_0: 524.22925ms
```

## Running multiple concurrent verifications + shuffling cache miss

8 verifications triggered 100ms apart in a 8 worker threads test. 
On PeerDAS network the number of concurrent verifications could be higher.

### Without Optimisation (`unstable`)

**Result**: 
- Majority of `verify_proposer_and_signature` took 500+ ms, as they all tried to load state and compute shuffling. 
- 6 out of 8 columns took 500+ms total time.

```
verify_proposer_and_signature: 514.836333ms
verify_proposer_and_signature: 1.295792ms
verify_proposer_and_signature: 523.640709ms
verify_proposer_and_signature: 1.183375ms
verify_proposer_and_signature: 536.179417ms
verify_proposer_and_signature: 537.438958ms
verify_proposer_and_signature: 510.963667ms
verify_proposer_and_signature: 501.606625ms

verify_kzg_for_data_column: 5.755458ms
verify_kzg_for_data_column: 7.827084ms
verify_kzg_for_data_column: 18.534791ms
verify_kzg_for_data_column: 5.266708ms
verify_kzg_for_data_column: 5.164334ms
verify_kzg_for_data_column: 4.364958ms
verify_kzg_for_data_column: 4.01325ms
verify_kzg_for_data_column: 3.930458ms

verify_total_time_6: 9.262542ms
verify_total_time_0: 520.857667ms
verify_total_time_1: 542.32725ms
verify_total_time_7: 6.59375ms
verify_total_time_2: 541.493416ms
verify_total_time_3: 541.944417ms
verify_total_time_4: 515.2415ms
verify_total_time_5: 505.678333ms
```

### After Optimisation (`OnceCell`)

**Result**: 
- Only one instance of `verify_proposer_and_signature` took 500+ ms, and the rest are lower depending on when the call was made - state was only load once and result was reused.
- Columns that arrive later took less time to verify, as they don't try to compute shuffling cache again.

```
verify_proposer_and_signature: 82.829084ms
verify_proposer_and_signature: 490.811709ms
verify_proposer_and_signature: 185.32525ms
verify_proposer_and_signature: 388.889333ms
verify_proposer_and_signature: 287.246667ms
verify_proposer_and_signature: 931.125µs
verify_proposer_and_signature: 2.505083ms
verify_proposer_and_signature: 1.602709ms

verify_kzg_for_data_column: 5.510167ms
verify_kzg_for_data_column: 5.613208ms
verify_kzg_for_data_column: 5.590584ms
verify_kzg_for_data_column: 5.566792ms
verify_kzg_for_data_column: 5.530834ms
verify_kzg_for_data_column: 5.974208ms
verify_kzg_for_data_column: 13.915542ms

verify_total_time_0: 496.674875ms
verify_total_time_4: 88.488625ms
verify_total_time_3: 191.037667ms
verify_total_time_2: 292.921833ms
verify_total_time_1: 394.67475ms
verify_total_time_5: 4.596292ms
verify_total_time_6: 7.768875ms
verify_total_time_7: 17.04425ms
```

jimmygchen:
May be worth adding this to block verification as well, as we're likely to receive the block + 128 columns within a small window under PeerDAS.

dapplion:
@jimmygchen we can as well cache the proposer signature verification with a similar strategy: LRUCache + OnceCell. Have a cache `LRUCache<(BlockRoot, BlockSignature), Result>`, and use it in block, blob and data column proposer signature verification. In a supernode case we will amortize the 500 ms across 128 columns.

jimmygchen:
Do you see much benefit of adding this signature cache in addition to `beacon_proposer_cache`?
From the numbers above, if we hit the `beacon_proposer_cache`, the verification time is quite minimal: 0.9 - 2.5ms. Am I missing something?

dapplion:
> From the numbers above, if we hit the beacon_proposer_cache, the verification time is quite minimal: 0.9 - 2.5ms. Am I missing something?

Right not necessary for now, we can revisit later if you want to hyper optimize

pawanjay176 (approved):
LGTM. 
I think worth considering adding the same to block and blob validation after we have tested this on peerdas devnets for some time

jimmygchen:
Thanks all! 

I've found a [bug](https://github.com/sigp/lighthouse/pull/7199/commits/337408500c0758925bb1c676cf728e03b304458a) in data column verification which causes proposer shuffling cache miss on every data column, due to incorrect epoch calculation only in minimal preset, as it was hard coded to use `MainnetEthSpec`.

This shouldn't happen often at all. But this PR should improve the performance in case if it happens and not do `128 + 1` state load + shuffling computation.

Agree to do it on block + blob verification as well once this is proven to be stable for some time on PeerDAS.

I'll go ahead and merge this now.

PR #7324, opened 2025-04-15
Revise logging when `queue is full`

chong-he:
The error log:

`ERRO Attestation delay queue is full msg: check system clock, queue_size: 16384, service: bproc`

is usually due to system resources overloading (CPU/RAM/disk IO). Logging as `check system clock` could cause confusion to users. This PR revises the log to `system resources may be saturated`


macladson (approved):
LGTM!

michaelsproul (approved):
Makes sense to me, thanks!

PR #7326, opened 2025-04-16
Optimize `validate_data_columns`

eserilev:
Based on conversation here:
https://discord.com/channels/595666850260713488/1360117816228843570/1360291212464099539

We've seen metrics that show beacon block header, inclusion proof etc. scaling with blob count. This PR aims to make a small optimization to `validate_data_columns`. Will need to test this to see if this makes any difference when scaling up blob count, though in general I think this is a safe optimization to include regardless.


jimmygchen (approved):
Nice!

Issue #7122, opened 2025-03-13
Avoid attempting to serve BlobsByRange RPC requests on Fulu slots


jimmygchen:
From PeerDAS slots, nodes will not serve `BlobsByRange` requests, and will be deprecated in future releases - at least ~18 days after the Fulu fork - as nodes are still required to serve Blobs before the Fulu fork for up to 4096 epochs.

Before this interface is deprecated, we need to add logic to avoid attempts to serving these requests containing Fulu slots, return an error and potentially penalise the faulty peer.

SunnysidedJ:
Hi @jimmygchen, I will sign up for this :)

jimmygchen:
Awesome, thank you @SunnysidedJ  ☺

SunnysidedJ:
If we penalize peers requesting fulu slots, don't we as a client should avoid requesting it? The sync logic doesn't seem to avoid requesting them. Then, we will get penalized by other lighthouse nodes. Please correct me if I'm wrong

eserilev:
@SunnysidedJ Yes we should also avoid attempting to request `BlobsByRange` ~18 days after the Fulu fork. I haven't dug too deeply into the syncing code but if its not doing that then yes it seems we would get penalized by other peers.

jimmygchen:
Hi @SunnysidedJ we'd like to have this issue fixed - I'm aware that you're busy with devnet testing currently, if you don't mind we'll take over this one and see if we can build on top of your existing PR. Thanks for your help!

SunnysidedJ:
Hi @jimmygchen, no worries. Thank you for reviewing and tracking everything. Please feel free to take this over!

chong-he:
I am giving this a try, thanks @SunnysidedJ

jimmygchen:
Completed in #7756 🎉

PR #7328, opened 2025-04-17
Fix for #7122 Avoid attempting to serve BlobsByRange RPC requests on Fulu slots

SunnysidedJ:
## Issue Addressed

#7122 Avoid attempting to serve BlobsByRange RPC requests on Fulu slots

## Proposed Changes

Added a check if Fulu slot is requested to BlobsByRange RPC handler.

## Additional Info

Please see comments below


SunnysidedJ:
A question left for penalizing peers: https://github.com/sigp/lighthouse/issues/7122#issuecomment-2811455509

beacon_node/network/src/network_beacon_processor/tests.rs
```diff
@@ -1216,9 +1216,55 @@ async fn test_blobs_by_range() {
     if test_spec::<E>().deneb_fork_epoch.is_none() {
         return;
     };
+    let rig_slot = 64;
+    let mut rig = TestRig::new(rig_slot).await;
+    let slot_count = 32;
+    // can I use slots_per_epoch from types::EthSpec?
+    rig.enqueue_blobs_by_range_request(0, slot_count);
```
SunnysidedJ: This test fails on the unstable branch as well, by calling this line. This is because test_spec::<E>() by default sets Deneb slot as the original slot 8626176. Thus, the request handler sees it is below data availability boundary. I tried setting TestRig's slot to 8626176, but it takes too long to reach the slot to initialize it. Is there a better way to manually set a new Deneb epoch for testing?

SunnysidedJ:
Test in progress as commented on the test file

beacon_node/network/src/network_beacon_processor/tests.rs
eserilev: In order to run that test on `unstable` you need to do something like

``` bash
FORK_NAME=deneb cargo nextest run --release \
        --features "fork_from_env" \
        -p network -- test_blobs_by_range
```

For the fulu situation I guess this test (or a separate test if thats easier) should check that the rpc request works if the epoch is less than ~18 days from fulu fork and fail if greater than ~18days

beacon_node/network/src/network_beacon_processor/tests.rs
```diff
@@ -1,4 +1,4 @@
-#![cfg(not(debug_assertions))] // Tests are too slow in debug.
+//#![cfg(not(debug_assertions))] // Tests are too slow in debug.
```
dapplion: Unwanted diff

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -917,6 +918,17 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+        // Check Fulu/PeerDAS is enabled. Blobs are not stored since then.
+        // Only need to check if the last slot is a Fulu slot or above.
+        let request_end_slot = Slot::from(req.start_slot + req.count);
+        let request_end_epoch = request_end_slot.epoch(T::EthSpec::slots_per_epoch());
+        if self.chain.spec.is_peer_das_enabled_for_epoch(request_end_epoch) {
+            return Err((
+                RpcErrorResponse::InvalidRequest,
+                "Req including Fulu slots",
+            ))
+        }
```
dapplion, line 968: Hmm, rethinking this maybe this is too punishing? Why not allow the request to creep into Fulu and just check that the start slot is in Deneb. It's fine to return empty for slots in Fulu.

Lighthouse only does by_range requests for a single epoch, but other clients may have different logic.

beacon_node/network/src/network_beacon_processor/tests.rs
SunnysidedJ: Thanks @eserilev, will check with this config and update you soon!

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
SunnysidedJ: Yes, I agree that it could be bit too punishing. The issue was vague in a way that we didn't set how much punishment we will give. Do you think it is okay to accept the request as long as it contains pre-Fulu slots?

Also, it seems like there's no check from Lighthouse to not send Fulu slots as left on the comment https://github.com/sigp/lighthouse/issues/7122#issuecomment-2811455509. So, we could potentially be penalized by ourselves

dapplion: Feels like something that should be clarified in the spec to align behaviour on all clients. Do you want to raise an issue to the specs?

SunnysidedJ: Sure! Just raised a PR to the specs https://github.com/ethereum/consensus-specs/pull/4286

beacon_node/network/src/network_beacon_processor/tests.rs
SunnysidedJ: Yes, this worked. Thanks!

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -7146,6 +7146,11 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         let end_slot = start_slot.saturating_add(count);
         let mut roots = vec![];
 
+        // let's explicitly check count = 0 since it's a public function for readabiilty purpose.
```
SunnysidedJ, line 7171: The original code handles edge cases.

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -880,8 +880,16 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
             "Received BlobsByRange Request"
         );
 
+        if req.count == 0 {
```
SunnysidedJ, line 912: Again, let's handle corner cases explicitly at some point

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
SunnysidedJ: Changed as per the spec PR to not punish and return empty for Fulu slots

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
SunnysidedJ: Will apply the same to other similar functions in this file if agreed

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -917,6 +925,15 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
             };
         }
 
+        // Check Fulu/PeerDAS is in the range. Blobs are not served since then.
+        if self.chain.spec.is_peer_das_enabled_for_epoch(request_end_epoch) {
+            // Fulu epoch is Some type by the condition above.
+            let fulu_epoch = self.chain.spec.fulu_fork_epoch.unwrap();
+            let fulu_start_slot = fulu_epoch.start_slot(T::EthSpec::slots_per_epoch());
+            // See the justification for the formula in PR https://github.com/sigp/lighthouse/pull/7328
+            req.count = fulu_start_slot.as_u64().saturating_sub(req.start_slot);
```
SunnysidedJ, line 967: Justification:
Let's consider every case one-by-one.

1. Range requested lies entirely on pre-Fulu
Case `req.start_slot + req.count <= fulu_start_slot` then
`req.count = req.count`
2. Range requested lies entirely on Fulu
Case `req.start_slot >= fulu_start_slot` then
`req.count = 0`
3. Range requested lies on both pre-Fulu and Fulu
Case `req.start_slot < fulu_start_slot` then
`req.count = req.count - ((req.start_slot + req.count) - fulu_start_slot)`
`req.count = fulu_start_slot - req.start_slot`

Union of all cases results in `req.count = fulu_start_slot.saturating_sub(req.start_slot)`

SunnysidedJ:
Current status: Issue-specific tests undergoing
Implemented as per the latest spec PR but can easily changeable to other options mentioned in the PR

SunnysidedJ:
> Current status: Issue-specific tests undergoing

Having some questions atm

**Sync-related questions**

1. Lighthouse nodes should not request to sync Fulu slots. Where do we handle this if so?
2. Do we automatically request data columns instead?

If sync is to be implemented I'd happy to work on, but seems better to do it after https://github.com/sigp/lighthouse/pull/7352

**Other**

1. Do we need to implement for BlobsByRoot as well?
2. The call self.get_block_roots_for_slot_range() within the function handle_blobs_by_root_request_inner() contributes to the metric BEACON_PROCESSOR_GET_BLOCK_ROOTS_TIME. Is this something wanted?
3. The function get_block_roots_for_slot_range() isn't error-ing out when the requested range is beyond the slots the client has (I might be wrong). Is this also expected behaviour?
4. In the line in handle_blobs_by_root_request_inner() `match self.chain.get_blobs(&root) {`, does it actually return any blobs for Fulu slots? If this is guaranteed the change won't be needed(?)

jimmygchen:
> 2. Do we automatically request data columns instead?

Yes. When making range request, we determine the batch type here:
https://github.com/sigp/lighthouse/blob/a6bdc474db0147c4194123bbd71dcfb562b36ae4/beacon_node/network/src/sync/range_sync/chain.rs#L1100

> 1. Do we need to implement for BlobsByRoot as well?

Oh yeah, we should not try to serve BlobsByRoot for fulu slots

> 2. The call self.get_block_roots_for_slot_range() within the function handle_blobs_by_root_request_inner() contributes to the metric BEACON_PROCESSOR_GET_BLOCK_ROOTS_TIME. Is this something wanted?

I don't see it being used in `handle_blobs_by_root_request_inner`, it's only used in handing `by range` requests.

> 3. The function get_block_roots_for_slot_range() isn't error-ing out when the requested range is beyond the slots the client has (I might be wrong). Is this also expected behaviour?

Yeah it doesn't error out. It just doesn't return blocks for the slots that the node doesn't have.

> 4. In the line in handle_blobs_by_root_request_inner() `match self.chain.get_blobs(&root) {`, does it actually return any blobs for Fulu slots? If this is guaranteed the change won't be needed(?)

It doesn't return blobs, because there's no blobs in Fulu slot.
However for the `blobs_by_range` method, we do waste resources doing lookup on the block roots before calling `get_blobs`. We shouldn't attempt to do this if the requested slot is in Fulu.
https://github.com/sigp/lighthouse/blob/39eb8145f89e11e3fd3a6c9c3d1031e8772ac491/beacon_node/network/src/network_beacon_processor/rpc_methods.rs#L920-L921

SunnysidedJ:
Hi @jimmygchen, thank you for detailed answers. I've made changes based on them. It'd be appreciated if LH team can have a look once available :)

SunnysidedJ:
Status: review ready
* Added avoiding mechanism and its tests for BlobsByRoot() as well
* All tests passes

jimmygchen:
Thanks @SunnysidedJ! Would you mind fixing the code conflicts - will take a look once it's resolved.

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -279,7 +280,38 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                    // If we can't find the block in either cache, we can't determine the epoch.
+                    continue;
+                };
+
+                is_fulu_by_root.insert(*root, self.chain.spec.is_peer_das_enabled_for_epoch(epoch));
+            }
+
+            if *is_fulu_by_root.get(root).unwrap() {
+                continue;
+            }
```
eserilev, line 313: remove unwrap

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -279,7 +280,38 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                } else if let Some(block) = self.chain.early_attester_cache.get_block(*root) {
+                    block.message().epoch()
+                } else if let Some(block) = self.chain.store.get_cached_block(root) {
+                    block.message().epoch()
+                } else if let Ok(Some(block)) = self.chain.store.get_blinded_block(root) {
+                    block.message().epoch()
+                } else {
+                    // If we can't find the block in either cache, we can't determine the epoch.
+                    continue;
+                };
```
eserilev, line 306: I'm wondering if we're opening up ourselves to a potential DOS vector here? Bogus blobs by root requests will force us to query multiple caches and the store itself. I guess peer rate limiting protects us so maybe thats a non-issue.

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -919,6 +954,19 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
         }
 
+        // Check Fulu/PeerDAS is in the range. Blobs are not served since then.
+        if self
+            .chain
+            .spec
+            .is_peer_das_enabled_for_epoch(request_end_epoch)
+        {
+            // Fulu epoch is Some type by the condition above.
+            let fulu_epoch = self.chain.spec.fulu_fork_epoch.unwrap();
```
eserilev, line 964: unwrap

eserilev (changes requested):

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
eserilev: would it be better to use fork choice as an initial check for the block root?

jimmygchen:
Superceded by #7756. Thanks a lot for your help and added clarification in the spec @SunnysidedJ  🙏

Issue #7212, opened 2025-03-27
Call `engine_getBlobs` as soon as receiving a valid `data_column_sidecar` from gossip


jimmygchen:
The "fetch blob from EL" was an optimization introduced in #6268 to reduce the delay to block import - nodes can retrieve blobs from EL without waiting for them from gossip, hence making blocks attestable earlier.

In Deneb, `engine_getBlobs` can only be called when a node receives a valid beacon block from gossip, as it requires KZG commitments of the blobs to compute version hashes, so that it can query for the associated blobs from the EL.

Under PeerDAS, blobs are transmitted over the network in the form of `DataColumnSidecar`s, each of them contains ALL blob KZG commitments for the block. This means we can apply this same optimisation, and query the Engine API for blobs as soon as we receive a `DataColumnSidecar` from gossip.

From the latest [Distributed Blob Publishing spec PR](https://github.com/ethereum/consensus-specs/pull/4183):
> Honest nodes SHOULD query `engine_getBlobsV2` as soon as they receive a valid `beacon_block` or `data_column_sidecar` from gossip. If ALL blobs matching `kzg_commitments` are retrieved, they should convert the response to data columns, and import the result.
> 
> Implementers are encouraged to leverage this method to increase the likelihood of incorporating and attesting to the last block when its proposer is not able to publish data columns on time.

However, with this additional optimisation, any of the 128 columns + the beacon block from gossip can trigger this, which means there's a high likelihood of race condition on these operations, and we should handle the concurrency here accordingly, and make sure we don't perform multiple fetch blob operations at the same time, which could result in wasting CPU cycles and unnecessary IO (repeated writes / block import).

Related issues:
- https://github.com/sigp/lighthouse/issues/6439
- https://github.com/sigp/lighthouse/issues/6764

jimmygchen:
@SunnysidedJ this is the issue I mentioned earlier in #6764. Let me know if you have any questions!

SunnysidedJ:
Sure, working on this at the moment. Hopefully, I should be able to raise a PR soon

PR #7329, opened 2025-04-17
Fix for #7212 Call engine_getBlobs as soon as receiving a valid data_column_sidecar from gossip

SunnysidedJ:
## Issue Addressed

#7212 Call engine_getBlobs as soon as receiving a valid data_column_sidecar from gossip

## Proposed Changes

Copied how blobs are fetched when a block is gossiped. Test in progress, and code will be refactored to merge common code between fetching logics from block and column gossips

## Additional Info

Please find the comments below


SunnysidedJ:
Lock while fetching from EL yet enabled

pawanjay176 (commented):
This looks good functionally. 
However, I think there is scope to reduce duplication between the fetch_blobs_from_block v/s fetch_blobs_from_columns functions by simply changing the arguments to the block version and reusing them in the data_column functions. 

Repeating the logic opens up to more bugs during further refactors

pawanjay176:
@SunnysidedJ I added a commit based on your branch to illustrate my suggestion https://github.com/sigp/lighthouse/commit/6fe3c5bc2a372ab1a90dd976f81c16915a904590

Let me know if you disagree with the approach. Feel free to cherry-pick the commit to this PR

jimmygchen:
Thinking about this again, I have some thoughts about whether this would materially help reduce block delays
- The fetch EL blobs operation no longer compute proofs, and it's now a very fast operation.
- We would still have to wait for the block to arrive before it can be imported.

The diff doesn't look too bad though. I think we'd have to also avoid calling EL multiple times, which if exists, may worsen the race condition we currently have - there are many concurrent paths that lead to block import - if we do this we'd have to track carefully to ensure we don't attempt multiple imports.
1. reconstruction
2. gossip data columns
3. EL getblobs

#7493 greatly reduces the likelihood of this from happening, but does not completely prevent it (issue here https://github.com/sigp/lighthouse/issues/6439 ). If we do go down this path to trigger getBlobs from gossip data column, it would increase the chance of race condition  and we should definitely try to avoid duplicate processing & imports - which will add complexity - so it might be worth weighing the trade-offs here.

Issue #6208, opened 2024-07-30
Perform validation on `config.yaml`


dknopik:
## Description

When constructing custom devnets, usually a custom `config.yaml` is created. The only validation performed right now is a check whether the `PRESET_NAME` matches the `EthSpec` used.

When a user makes a mistake and provides an invalid `config.yaml`, it is silently accepted at first, but may cause unexpected behaviour later on. Users might not realize their mistake from this behaviour. For example: When specifying the same fork version for Deneb (activated from genesis) and Electra (far future), Lighthouse will subscribe to the subnets of Electra, but send blocks for Deneb - causing deserialisation errors. This can be avoided by implementing sanity checks that deny illegal `config.yaml`s.

## Version

Current `unstable` (`c7ded10`)

## Present Behaviour

Invalid `config.yaml`s are silently accepted, causing various errors later on.

## Expected Behaviour

Lighthouse should perform a set of sanity checks to identify clearly faulty configurations and tell the user.

## Steps to resolve

- [ ] Define good sanity checks
- [ ] Implement

jimmygchen:
Yeah we've seen this issue before and I think it's a good idea to validate the config.
Related: https://github.com/sigp/lighthouse/issues/3033#issuecomment-1659491285

Patchoulis:
I'll probably pick this one up later today.

PR #7341, opened 2025-04-21
Config Validation

Patchoulis:
## Issue Addressed

[#6208](https://github.com/sigp/lighthouse/issues/6208)

## Proposed Changes

A few changes. First, as discussed in the discord, the defaults have been removed to ensure that all expected fields are provided.

Second, I do validate that no additional values are present. However, as discussed with sproul, there are some legitimate reasons to add some for future forks. I ended up solving this by adding a small modifiable constant which allows specified optional fields to exist.

I'm doing some tests to ensure presets like SLOTS_PER_EPOCH are also disallowed, but this should also solve that problem.

## Additional Info


macladson (changes requested):
Thanks for the PR! Looks like this needs a quick `cargo fmt --all`
Some of the tests don't seem happy either

Patchoulis:
I'll do that, as well as take a look at the tests. Appreciate the review!

consensus/types/src/chain_spec.rs
```diff
@@ -1833,6 +1720,18 @@ impl Config {
...
+    /// This is intended to check those extra fields to see if it's for a future fork or not
+    pub fn validate_extra_fields(&self) -> Result<(), String> {
+        for key in self.future_fields.keys() {
+            if !FUTURE_FIELDS.contains(&key.as_str()) {
+                return Err(format!("Unknown or disallowed field found: {}", key));
+            }
+        }
+
+        Ok(())
+    }
```
michaelsproul, line 1746: I like this approach. Simple and flexible.

Let me know if you get stuck fixing any of the tests

Patchoulis:
Alright, I fixed up the tests.

Most of the issues were simply because the validation and removal of defaults meant I had to add the missing fields in the existing configs.

For this, I ended up using the original default values to maintain the same functionality as before.

PR #7309, opened 2025-04-10
Add riscv64 build support

cmd0s:
## Issue Addressed

Lighthouse does not compile on the _riscv64_ architecture due to a missing target configuration in the `bls_hardware_acceleration` function.

## Proposed Changes

Added support for compiling Lighthouse on the riscv64 architecture by explicitly handling the _riscv64_ target in the bls_hardware_acceleration function.
Specifically, this line was added:
```
#[cfg(target_arch = "riscv64")]
return false;
```
This prevents a compilation error on RISC-V.

## Additional Info

Tested on a Milk-V Megrez board running Debian for _riscv64_.
With this small change, Lighthouse compiles successfully on this architecture.
It was also successfully run and synchronized with both the mainnet and the Hoodi testnet.

PR #7346, opened 2025-04-22
Enable cross-compiling for riscv64 architecture

cmd0s:
## Issue Addressed

Lighthouse currently lacks support for cross-compilation targeting the `riscv64` architecture.

## Proposed Changes

This PR introduces initial support for cross-compiling Lighthouse to `riscv64`. The following changes were made:

- **Makefile**: Updated to support `cross` with `riscv64` as a target.
- **Cross.toml**: Added configuration specific to `riscv64`.
- **Documentation**: List 'build-riscv64' in `book/src/installation_cross_compiling.md`.

## Additional Info

- To ensure modern docker image support for riscv64 in Cross, this PR uses:
  ```sh
  cargo install cross --git https://github.com/cross-rs/cross
  ```
The default Docker images for `riscv64` in the Cross release version are based on outdated Ubuntu versions.

- These changes build upon PR #7309, where the following fallback was added to `bls_hardware_acceleration()`:
```rust
#[cfg(target_arch = "riscv64")]
return false;
```
Once stable Rust adds support for runtime RISC-V vector feature detection (expected after Rust 1.88), this can be replaced with:
```rust
#[cfg(target_arch = "riscv64")]
std::arch::is_riscv_feature_detected!("v")
```
At the time of writing, using this feature requires nightly Rust (1.88.0-nightly) and the 
```rust
#![feature(stdarch_riscv_feature_detection)]`
```
attribute in `lighthouse/src/main.rs.
Note: Linux kernel 6.14 is required for this feature to work properly at runtime.

BTW: I have successfully compiled Lighthouse 7.1.0-beta for riscv64 with Vector Extension support 
_(This requires more changes to the code and the build/runtime environment)_. 
I ran it on a Banana Pi F3 (SpacemiT-X60), which supports RVV 1.0, and fully synchronized an Ethereum node (Hoodi) where both Geth and Lighthouse run on the same device.
image
image


michaelsproul:
Hey @cmd0s, thanks for this! Looks great!

I've just pushed a commit (18871997f8baf957351fb2c82e5d9014e9de1192) to use the newer Cross docker image with the released `cross` binary. I think this is simpler than installing a new `cross` version inside the makefile (no side effects on the system, and no need to re-install every time).

I've tested on my machine and it works. I'll merge this PR if you can confirm it works on your end too. Thanks again!

cmd0s:
Hi,
I ran some tests.

The native build using `make` worked on x86_64, aarch64, riscv64, and Mac M4.
The `make build-riscv64` command worked on x86_64, but cross-compilation for riscv64 failed on both my Mac M4 and Raspberry Pi 5.

If the assumption is that cross-compilation is only meant to be run on x86_64, then that's fine.

michaelsproul:
Yeah the cross compilation tools only really work on x86.

Thanks for testing, I'll merge!

michaelsproul (approved):

PR #7347, opened 2025-04-22
Implement weak subjectivity safety checks

### LINKED ISSUE #7273: Implement weak subjectivity safety checks
See
- https://github.com/ethereum/consensus-specs/pull/4179

eserilev:
## Issue Addressed

Closes #7273

## Proposed Changes

https://github.com/ethereum/consensus-specs/pull/4179


eserilev:
hey @dapplion when you have a chance, could you take a look at what I have so far?

The general flow is

- user sets the weak subjectivity period checkpoint via the existing `wss-checkpoint` flag 
- on startup, we make the weak subjectivity calculation, using the user provided ws checkpoint, the head snapshot and the most recent finalized checkpoint

If the user doesn't set a weak subjectivity checkpoint, what are we supposed to do here?

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -2944,6 +2944,35 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
+        }
+
+        let finalized_epoch = finalized_state.slot().epoch(E::slots_per_epoch());
+
+        if finalized_epoch != ws_checkpoint.epoch {
+            return false;
+        }
+
+        let Ok(ws_period) = finalized_state.compute_weak_subjectivity_period(&self.spec) else {
+            // TODO(ws) failed to calculate ws period, log and continue?
```
dapplion: No, this should never fail propagate the error and abort start

beacon_node/beacon_chain/src/builder.rs
```diff
@@ -816,6 +816,24 @@ where
             ));
         }
 
+        // Check if the store is within the weak subjectivity period
+        if let Some(ws_checkpoint) = self.chain_config.weak_subjectivity_checkpoint {
+            let Ok(finalized_block) = fork_choice.get_finalized_block() else {
+                panic!("TODO(ws)")
+            };
+            if !store.is_within_weak_subjectivity_period(
```
dapplion: Just use the `head_snapshot.beacon_state` to compute the weak subjectivity period. You don't need to fetch the head state's finalize state. For WS sync, the head state = the weak subjectivity checkpoint state. You don't need to fetch any more data:
- compute ws_period from head state
- get current epoch from clock
- check `current_epoch <= head_state.epoch + ws_period`

No need to involve the store here

consensus/types/src/beacon_state.rs
```diff
@@ -2505,6 +2505,30 @@ impl<E: EthSpec> BeaconState<E> {
...
+        const SAFETY_DECAY: u64 = 10;
+
+        let total_active_balance = self.get_total_active_balance()?;
+        let balance_churn_limit = self.get_balance_churn_limit(spec)?;
+        let epochs_for_validator_set_churn = SAFETY_DECAY
+            .safe_mul(total_active_balance)?
+            .safe_div(balance_churn_limit.safe_mul(200)?)?;
+        let weak_subjectivity_period = spec
+            .min_validator_withdrawability_delay
+            .safe_mul(epochs_for_validator_set_churn)?;
```
dapplion: Where is this formula from? You should implement

https://github.com/ethereum/consensus-specs/blob/1937aff86b41b5171a9bc3972515986f1bbbf303/specs/phase0/weak-subjectivity.md?plain=1#L82

dapplion:
> If the user doesn't set a weak subjectivity checkpoint, what are we supposed to do here?

We should assert that the initial state is within the weak subjectivity period on any type of start.

We should also offer a flag such that
- If flag set and starting from outside weak subjectivity period: big warn log, continue start
- If flag not set and starting from outside weak subjectivity period: error, abort start

consensus/types/src/beacon_state.rs
eserilev: I'm using the updated forumla for electra

https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/weak-subjectivity.md#modified-compute_weak_subjectivity_period

I think the calc is simpler now that top ups are part of activation churn

dapplion: Ah right, please link it

eserilev: Ah but we probably need to support the pre-electra calculation as well

beacon_node/beacon_chain/src/builder.rs
```diff
@@ -816,6 +816,31 @@ where
...
+                warn!(
+                    head_slot=%head_state.slot(),
+                    %current_slot,
+                    "The current head state is outside the weak subjectivity period. It is highly recommended to purge your db and \
+                    checkpoint sync."
+                )
+            }
+            return Err(
+                "The current head state is outside the weak subjectivity period. It is highly recommended to purge your db and \
+                checkpoint sync. Alternatively you can accept the risks and ignore this error with the --ignore-ws-check flag.".to_string()
```
dapplion: We should either summarize "the risks" or link to some resource of our own. Otherwise it's quite opaque to the users.

eserilev: i've summarized the risks and linked to a blog post by vitalik

eserilev:
Thanks for all the help Lion. I've made changes based on your feedback and added additional test coverage

dapplion (approved):
Looks great now!

beacon_node/beacon_chain/src/builder.rs
```diff
@@ -845,6 +845,34 @@ where
...
+                    "The current head state is outside the weak subjectivity period. It is highly recommended to purge your db and \
+                    checkpoint sync. You are currently running a node that is susceptible to long range attacks. For more information please \
+                    read this blog post: https://blog.ethereum.org/2014/11/25/proof-stake-learned-love-weak-subjectivity"
+                )
+            }
+            return Err(
+                "The current head state is outside the weak subjectivity period. It is highly recommended to purge your db and \
+                checkpoint sync. It is possible to ignore this error with the --ignore-ws-check flag, but this could make the node \
+                susceptible to long range attacks. For more information please read this blog post: \
+                https://blog.ethereum.org/2014/11/25/proof-stake-learned-love-weak-subjectivity".to_string()
```
dapplion: Nice message but I would order as:
- "The current head state is outside the weak subjectivity period"
- you are susceptible to attacks
- read more about attacks here
- if you understand the risks and want to continue: flag

beacon_node/src/cli.rs
```diff
@@ -1404,6 +1404,17 @@ pub fn cli_app() -> Command {
                 .help_heading(FLAG_HEADER)
                 .display_order(0)
         )
+        .arg(
+            Arg::new("ignore-ws-check")
+                .long("ignore-ws-check")
+                .help("The Weak Subjectivity Period is the the maximum time a node can be offline and still \
+                safely sync back to the canonical chain without the risk of falling victim to long-range attacks. \
+                This flag disables the Weak Subjectivity check at startup, allowing users to run a node whose current head snapshot \
+                is outside the Weak Subjectivity Period. It is unsafe to disable the Weak Subjectivity check at startup.")
```
dapplion: Use similar message to the error log and link the post. The user needs to know that this skips a defense for an attack, the details of what the WS are not super relevant and can be delegated to the post

dapplion (approved):
Solid straightforward change, minor some comments on the wording of the messages. Otherwise this is good to go. Only risk towards mainnet path is the WS check error triggering as a false positive. This is unlikely as there are many tests covering the `head_state.compute_weak_subjectivity_period` computation

beacon_node/beacon_chain/src/builder.rs
eserilev: Updated messaging order

beacon_node/src/cli.rs
eserilev: [83f6ac4](https://github.com/sigp/lighthouse/pull/7347/commits/83f6ac436ba61cd413cb468a7dd3958f3a003365)

consensus/types/src/state/beacon_state.rs
```diff
@@ -2715,6 +2722,37 @@ impl<E: EthSpec> BeaconState<E> {
 
         Ok(())
     }
+
+    /// Returns the weak subjectivity period for `self`
+    pub fn compute_weak_subjectivity_period(&self, spec: &ChainSpec) -> Result<Epoch, Error> {
+        let total_active_balance = self.get_total_active_balance()?;
```
michaelsproul, line 3027: Here we are assuming the total active balance cache is initialized, which is fine (see comment below)

consensus/types/src/beacon_state.rs
```diff
@@ -2715,6 +2722,37 @@ impl<E: EthSpec> BeaconState<E> {
...
+        if fork_name.electra_enabled() {
+            let balance_churn_limit = self.get_balance_churn_limit(spec)?;
+            compute_weak_subjectivity_period_electra(
+                total_active_balance,
+                balance_churn_limit,
+                spec,
+            )
+        } else {
+            let active_validator_count = self
+                .get_active_validator_indices(self.slot().epoch(E::slots_per_epoch()), spec)?
```
michaelsproul: We may as well use `get_cached_active_validator_indices` here, as it will be faster.

We are building the caches on the head snapshot here, prior to calling this func:

https://github.com/sigp/lighthouse/blob/83f6ac436ba61cd413cb468a7dd3958f3a003365/beacon_node/beacon_chain/src/builder.rs#L829-L832

consensus/types/src/beacon_state.rs
```diff
@@ -2986,3 +3024,211 @@ impl<'de, E: EthSpec> ContextDeserialize<'de, ForkName> for BeaconState<E> {
...
+
+        // (600 * delta * (2 * t + T))
+        let epochs_for_validator_set_churn_denominator =
+            validator_churn_limit.safe_mul(600)?.safe_mul(
+                total_active_balance_per_validator
+                    .safe_mul(2)?
+                    .safe_add(max_effective_balance)?,
+            )?;
+
+        // N * (t * (200 + 12 * D) - T * (200 + 3 * D)) // (600 * delta * (2 * t + T))
```
michaelsproul: Ok this is super gnarly, I think we just delete this and return `0` if we are pre-Electra (so that the check always fails).

michaelsproul (dismissed):

consensus/types/src/beacon_state.rs
michaelsproul: If we delete the pre-Electra logic, we can also delete this

consensus/types/src/beacon_state.rs
eserilev: I removed the pre-electra stuff and just return 0 now

consensus/types/src/beacon_state.rs
eserilev: removed pre-electra logic

michaelsproul:
There are some tests failures. Maybe we could use a small non-zero constant pre-Electra to smooth over this issue, or we could update these tests to only run post-Electra.

eserilev:
Yeah I'm going to fix theses tests to run post Electra

eserilev:
`InvalidPayloadRig` wasnt using the `fork_from_env` stuff when constructing `spec` so we weren't actually running related tests against `RECENT_FORKS` in some cases.

`revert_minority_fork_on_resume` was an annoying test to try to get right when setting the WS calculation to 0 pre-electra. Setting the calculation to a min of 5 fixes the test. 

basic sim and fallback sim were also failing and I havent had a chance to look at why. I've ramped up the pre electra ws calc to 256 to fix this (5 was too small of a number apparently).

dapplion (approved):
Fixed conflicts, logic and error messages are good

eserilev:
@mergify requeue

eserilev:
@mergify requeue

dapplion:
@mergify requeue

PR #7349, opened 2025-04-23
Use latest stable fork in default ChainSpec

michaelsproul:
## Issue Addressed

The `E::default_spec()` functions are used in tests, but are a little strange in that they use the "real" spec for their `EthSpec`, e.g. the `MainnetEthSpec::default_spec()` is currently the spec for Ethereum mainnet!

This PR changes these functions to return simpler specs with the latest stable fork initialised from genesis. In test functions at low slot counts, this ensures we are testing the latest specs and not random phase0 stuff.

I suspect some of the tests will fail, so this is WIP for now.


michaelsproul:
Closing as I don't have time to finish this right now.

PR #7351, opened 2025-04-23
Revise consolidation info in Lighthouse book

chong-he:
## Issue Addressed

One of the information in the consolidation section in Lighthouse book is wrong. I realise this after reading https://ethereum.org/en/roadmap/pectra/maxeb/ and a further look at [EIP 7251](https://eips.ethereum.org/EIPS/eip-7251) which states:

`
Note: the system contract uses the EVM CALLER operation (Solidity: msg.sender) to get the address used in the consolidation request, i.e. the address that calls the system contract must match the 0x01 withdrawal credential recorded in the beacon state.
`

So the withdrawal credentials of both source and target validators need not be the same.

### DIFF
--- a/book/src/validator_consolidation.md
+++ b/book/src/validator_consolidation.md
@@ -19,11 +19,11 @@ Let's take a look at an example: Initially, validators A and B are both with 0x0
     - validator A has 64 ETH
     - validator B has 0 ETH (i.e., validator B has exited the beacon chain)
 
-The consolidation process can be repeated to consolidate more validators into validator A.
+The consolidation process can be repeated to consolidate more validators into validator A. The request is made by signing a transaction using the **withdrawal address** of the source validator. The withdrawal credential of the target validator can be different from the source validator.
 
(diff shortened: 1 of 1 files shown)


macladson (approved):
This seems correct to me as well.
Reading through the `process_consolidation_requests` function, I believe we only check that the source validator's actual withdrawal address is equal to the source address in the consolidation request.

https://github.com/sigp/lighthouse/blob/c13e069c9c633535336bca6d6f5f7b30720df629/consensus/state_processing/src/per_block_processing/process_operations.rs#L733-L734

The only checks we do on the target validator is making sure it has a `0x02` withdrawal address.

https://github.com/sigp/lighthouse/blob/c13e069c9c633535336bca6d6f5f7b30720df629/consensus/state_processing/src/per_block_processing/process_operations.rs#L742-L745

michaelsproul (approved):
Nice correction, I didn't catch this in the first review!

michaelsproul:
@mergify requeue

PR #7352, opened 2025-04-23
Fix PeerDAS sync scoring

dapplion:
## Issue Addressed

Range sync will not penalize the proper peer during PeerDAS sync for invalid column errors

## Proposed Changes

- Propagate errors with enough details to figure out which block component(s) are invalid
- Track which peer served what in range sync batches
- Use the above two ^ to downscore the correct peers

## Progress

- [ ] Implement
- [ ] Add tests


beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -677,29 +601,44 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
...
-                            score_adjustment = %penalty,
                             batch_epoch = %batch_id,
-                            "Backfill batch failed to download. Penalizing peers"
+                            error,
+                            "Backfill sync failed after attempting to process batch too many times"
                         );
 
-                        for peer in self.participating_peers.drain() {
-                            network.report_peer(peer, *penalty, "backfill_batch_failed");
-                        }
```
dapplion, line 655: @AgeManning rethinking this, the following situation is problematic:
- We request epoch 1000, which should contain 2 canonical blocks (block A, block B) but the serving peer X only sends the first.
- We request epoch 999 to peer Y expecting block B but receive block C.

I think currently we downscore peer Y and retry epoch 999, getting stuck and eventually failing the backfill sync. Then we downscore everyone that participated in sync. So yes, we penalize peer X but with a lot of collateral damage. Is this the reason for downscoring participating_peers? We could find a better strategy if this is the only reason for it.

AgeManning: Yeah I think you're right. In backfill sync, I dont think we have the concept of validated batches for handling missing blocks. 

If that's true, then yes this is probably overkill. It would deter us from using these peers in the future, but as you've mentioned in your new comment, we probably already downscore the peers in other cases. Although a peer not sending us a block might skip all penalties and be hard to downscore. 

Either way, I'm happy to remove this

jimmygchen:
Hi @dapplion there are quite a lot of conflicts with `unstable` now, would you mind resolving them?

beacon_node/network/src/sync/range_sync/range.rs
```diff
@@ -227,15 +228,15 @@ where
     pub fn blocks_by_range_response(
         &mut self,
         network: &mut SyncNetworkContext<T>,
-        peer_id: PeerId,
+        peer_id: BatchPeerGroup,
```
jimmygchen: rename to `batch_peer_group`

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -216,7 +216,7 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
         &mut self,
         network: &mut SyncNetworkContext<T>,
         batch_id: BatchId,
-        peer_id: &PeerId,
+        peer: BatchPeerGroup,
```
jimmygchen: rename to `batch_peer_group`

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -245,7 +245,7 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
         // Remove the request from the peer's active batches
 
         // TODO(das): should use peer group here https://github.com/sigp/lighthouse/issues/6258
-        let received = batch.download_completed(blocks, *peer_id)?;
+        let received = batch.download_completed(blocks, peer)?;
```
jimmygchen: The above `TODO(das)` can be removed?

beacon_node/network/src/sync/range_sync/batch.rs
```diff
@@ -275,10 +298,10 @@ impl<E: EthSpec, B: BatchConfig> BatchInfo<E, B> {
     pub fn download_completed(
         &mut self,
         blocks: Vec<RpcBlock<E>>,
-        peer: PeerId,
+        peer: BatchPeerGroup,
```
jimmygchen: `batch_peer_group`

beacon_node/network/src/sync/network_context.rs
```diff
@@ -604,11 +600,13 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
 
     /// Received a blocks by range or blobs by range response for a request that couples blocks '
     /// and blobs.
+    #[allow(clippy::type_complexity)]
     pub fn range_block_component_response(
         &mut self,
         id: ComponentsByRangeRequestId,
+        peer_id: PeerId,
```
jimmygchen, line 608: I find this `peer_id` quite confusing - this is the peer the sent the last RPC response right? Why do we need to keep track of this?

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -100,48 +117,59 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
...
-                req.finish(req_id, columns)
+                req.finish(req_id, columns, peer_id)
             }
         }
     }
 
-    pub fn responses(&self, spec: &ChainSpec) -> Option<Result<Vec<RpcBlock<E>>, String>> {
-        let Some(blocks) = self.blocks_request.to_finished() else {
+    #[allow(clippy::type_complexity)]
+    pub fn responses(
```
jimmygchen, line 127: Probably worth adding some docs to this function to describe what it does and what it returns

beacon_node/network/src/sync/range_sync/batch.rs
```diff
@@ -194,7 +220,7 @@ impl<E: EthSpec, B: BatchConfig> BatchInfo<E, B> {
         );
 
         for attempt in &self.failed_processing_attempts {
-            peers.insert(attempt.peer_id);
+            peers.insert(attempt.peer_id.block());
```
jimmygchen: does this method just returns failed block peers? maybe worth renaming.

beacon_node/network/src/sync/range_sync/batch.rs
```diff
@@ -213,7 +239,7 @@ impl<E: EthSpec, B: BatchConfig> BatchInfo<E, B> {
     }
 
     /// Returns the peer that is currently responsible for progressing the state of the batch.
-    pub fn processing_peer(&self) -> Option<&PeerId> {
+    pub fn processing_peer(&self) -> Option<&BatchPeerGroup> {
```
jimmygchen: Please rename to reflect peer group change and update docs

beacon_node/network/src/sync/range_sync/batch.rs
```diff
@@ -441,16 +463,16 @@ impl<E: EthSpec, B: BatchConfig> BatchInfo<E, B> {
 /// Represents a peer's attempt and providing the result for this batch.
 ///
 /// Invalid attempts will downscore a peer.
-#[derive(PartialEq, Debug)]
+#[derive(Debug)]
 pub struct Attempt {
     /// The peer that made the attempt.
-    pub peer_id: PeerId,
+    pub peer_id: BatchPeerGroup,
```
jimmygchen: Please update field name and doc

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -521,10 +520,22 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
...
+                peer_action,
+                // TODO(sync): propagate error in logs
+                error: _,
             } => {
-                // Penalize the peer appropiately.
-                network.report_peer(peer, *penalty, "faulty_batch");
+                // TODO(sync): De-dup between back and forwards sync
+                if let Some(penalty) = peer_action.block_peer {
+                    // Penalize the peer appropiately.
+                    network.report_peer(batch_peers.block(), penalty, "faulty_batch");
```
jimmygchen, line 530: is it worth differentiating between the faulty batch types, i.e. `faulty_batch_block` / `faulty_batch_columns`

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -37,11 +39,65 @@ pub enum ChainSegmentProcessId {
...
+                Some(PeerGroupAction::block_peer(PeerAction::LowToleranceError))
+            }
+            AvailabilityCheckError::InvalidColumn(errors) => Some(PeerGroupAction::column_peers(
+                &errors.iter().map(|(index, _)| *index).collect::<Vec<_>>(),
+                PeerAction::LowToleranceError,
+            )),
+            AvailabilityCheckError::ReconstructColumnsError(_) => None, // internal error
+            AvailabilityCheckError::KzgCommitmentMismatch { .. } => None, // should never happen after checking inclusion proof
+            AvailabilityCheckError::Unexpected(_) => None,                // internal
+            AvailabilityCheckError::SszTypes(_) => None,                  // ??
```
jimmygchen: This variant is slightly confusing, I wonder if we can remove it.
It is used here:
https://github.com/sigp/lighthouse/blob/e0b36501c82af44bf20262092885d9d8c3769b98/beacon_node/beacon_chain/src/block_verification_types.rs#L208


This error throws when the number of column sidecars exceeds 128, which I'm not sure if it's possible..? if not we could consider removing this.

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -37,11 +39,65 @@ pub enum ChainSegmentProcessId {
...
+            AvailabilityCheckError::MissingBlobs => None, // TODO(das) internal for now
```
jimmygchen: Why is this not attributed to block peer?

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -37,11 +39,65 @@ pub enum ChainSegmentProcessId {
...
+            AvailabilityCheckError::MissingCustodyColumns(columns) => Some(
+                PeerGroupAction::column_peers(columns, PeerAction::LowToleranceError),
```
jimmygchen: This feels a bit strict? we could instantly ban them within a single range request.

beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -628,31 +601,44 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
...
+                error,
             } => {
+                // TODO(sync): De-dup between back and forwards sync
+                if let Some(penalty) = peer_action.block_peer {
+                    // Penalize the peer appropiately.
+                    network.report_peer(batch_peers.block(), penalty, "faulty_batch");
+                }
+                for (column_index, penalty) in &peer_action.column_peer {
+                    if let Some(peer) = batch_peers.column(column_index) {
+                        network.report_peer(*peer, *penalty, "faulty_batch");
```
jimmygchen: This could result in an instant ban of a supernode peer if they fail one request. Do we want to ban the same peer multiple times from a single batch?

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -589,148 +647,75 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
     fn process_backfill_blocks(
         &self,
         downloaded_blocks: Vec<RpcBlock<T::EthSpec>>,
-    ) -> (usize, Result<(), ChainSegmentFailed>) {
-        let total_blocks = downloaded_blocks.len();
-        let available_blocks = match self
-            .chain
-            .data_availability_checker
-            .verify_kzg_for_rpc_blocks(downloaded_blocks)
```
jimmygchen, line 597: Has the KZG verification been moved elsewhere?

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -589,148 +647,75 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                    debug!(?e, "Backfill sync processing error");
+                } else {
+                    // All others are some type of internal error worth surfacing?
+                    warn!(?e, "Unexpected backfill sync processing error");
+                }
+
+                Err(ChainSegmentFailed {
+                    // Render the full error in debug for full details
+                    message: format!("{:?}", e),
+                    // This is an internal error, don't penalize the peer.
```
jimmygchen: This comment is outdated?

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -739,121 +724,89 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                Some(PeerGroupAction::block_peer(PeerAction::Fatal))
             }
-            other => {
-                debug!(
-                    msg = "peer sent invalid block",
-                    outcome = %other,
-                    "Invalid block received"
-                );
+            // TODO(sync): Should we penalize slashable blocks?
+            BlockError::Slashable => None,
```
jimmygchen: Is this relevant? it suggests `MidToleranceError` here:
https://github.com/sigp/lighthouse/blob/e0b36501c82af44bf20262092885d9d8c3769b98/beacon_node/beacon_chain/src/block_verification.rs#L297-L298

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -739,121 +724,89 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+            // Do not penalize peers for internal errors.
+            // BlobNotRequired is never constructed on this path
+            // TODO(sync): Double check that all `BeaconChainError` variants are actually internal
+            // errors in thie code path
```
jimmygchen, line 797: Yeah I think `BeaconChainError` has been used for internal errors - and is also the existing behaviour to not penalise peer(s) - we haven't ran into an scenario where we'd need to penalise peer for this.

beacon_node/network/src/network_beacon_processor/tests.rs
```diff
@@ -348,22 +348,6 @@ impl TestRig {
         }
     }
 
-    pub fn enqueue_gossip_data_columns(&self, col_index: usize) {
```
jimmygchen: This is still being used in multiple places.

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -447,7 +449,21 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
...
+                    let missing_custody_columns = expected_custody_indices
+                        .into_iter()
+                        .filter(|index| !received_indices.contains(index))
+                        .collect::<Vec<_>>();
+
+                    if !missing_custody_columns.is_empty() {
+                        return Err(AvailabilityCheckError::MissingCustodyColumns(
+                            missing_custody_columns,
+                        ));
+                    }
```
jimmygchen, line 467: I don't see an immediate problem here - although It's a bit weird seeing this logic here, I feel like the entire call tree is quite confusing and not doing what it's advertised to do 

`chain.process_chain_segment`
- `signature_verify_chain_segment` (we're suppose to signature verify)
    - `data_availability_checker.verify_kzg_for_rpc_blocks` (now we're doing kzg verify)
       - BUT we end up verifying that received columns from RPC match requested columns

beacon_node/beacon_chain/tests/block_verification.rs
```diff
@@ -367,7 +367,7 @@ async fn chain_segment_non_linear_parent_roots() {
 
     let (mut block, signature) = blocks[3].as_block().clone().deconstruct();
     *block.parent_root_mut() = Hash256::zero();
-    blocks[3] = RpcBlock::new_without_blobs(
+    blocks[3] = harness.build_rpc_block_from_store_blobs(
```
jimmygchen: I think this test would fail? this method is doing different thing to what we were previously doing:

- `RpcBlock::new_without_blobs` constructs an RPC block without assuming we have it in the database
- `build_rpc_block_from_store_blobs` loads block and blobs from the database and build an RPC block (for testing)

beacon_node/beacon_chain/tests/block_verification.rs
```diff
@@ -405,7 +405,7 @@ async fn chain_segment_non_linear_slots() {
             .collect();
     let (mut block, signature) = blocks[3].as_block().clone().deconstruct();
     *block.slot_mut() = Slot::new(0);
-    blocks[3] = RpcBlock::new_without_blobs(
+    blocks[3] = harness.build_rpc_block_from_store_blobs(
```
jimmygchen: Same as above

beacon_node/beacon_chain/tests/block_verification.rs
```diff
@@ -433,7 +433,7 @@ async fn chain_segment_non_linear_slots() {
             .collect();
     let (mut block, signature) = blocks[3].as_block().clone().deconstruct();
     *block.slot_mut() = blocks[2].slot();
-    blocks[3] = RpcBlock::new_without_blobs(
+    blocks[3] = harness.build_rpc_block_from_store_blobs(
```
jimmygchen: Same as above

beacon_node/beacon_chain/tests/block_verification.rs
```diff
@@ -578,11 +578,7 @@ async fn invalid_signature_gossip_block() {
             .into_block_error()
             .expect("should import all blocks prior to the one being tested");
         let signed_block = SignedBeaconBlock::from_block(block, junk_signature());
-        let rpc_block = RpcBlock::new_without_blobs(
-            None,
-            Arc::new(signed_block),
-            harness.sampling_column_count,
-        );
+        let rpc_block = harness.build_rpc_block_from_store_blobs(None, Arc::new(signed_block));
```
jimmygchen: Same as above

beacon_node/beacon_chain/tests/block_verification.rs
```diff
@@ -1771,11 +1767,7 @@ async fn import_duplicate_block_unrealized_justification() {
     // Create two verified variants of the block, representing the same block being processed in
     // parallel.
     let notify_execution_layer = NotifyExecutionLayer::Yes;
-    let rpc_block = RpcBlock::new_without_blobs(
-        Some(block_root),
-        block.clone(),
-        harness.sampling_column_count,
-    );
+    let rpc_block = harness.build_rpc_block_from_store_blobs(Some(block_root), block.clone());
```
jimmygchen: Same as above

beacon_node/beacon_chain/src/historical_blocks.rs
```diff
@@ -44,7 +51,56 @@ impl From<StoreError> for HistoricalBlockError {
...
+                Self::InvalidSignature(format!("{e:?}"))
+            }
+            // All these variants are internal errors or unreachable for historical block paths,
+            // which only check the proposer signature.
+            // BadBlsBytes = Unreachable
+            e @ (SignatureSetError::BeaconStateError(_)
+            | SignatureSetError::ValidatorUnknown(_)
+            | SignatureSetError::ValidatorPubkeyUnknown(_)
+            | SignatureSetError::IncorrectBlockProposer { .. }
+            | SignatureSetError::MismatchedPublicKeyLen { .. }
```
jimmygchen: I dont see this used anywhere, can this be removed?

beacon_node/beacon_chain/src/block_verification_types.rs
```diff
@@ -88,14 +118,20 @@ impl<E: EthSpec> RpcBlock<E> {
...
     BlockAndBlobs(Arc<SignedBeaconBlock<E>>, BlobSidecarList<E>),
-    /// This variant is used with parent lookups and by-range responses. It should have all
-    /// requested data columns, all block roots matching for this block.
-    BlockAndCustodyColumns(Arc<SignedBeaconBlock<E>>, CustodyDataColumnList<E>),
+    /// **Range sync**: Variant for all post-Fulu blocks regardless if they have data or not
+    /// **Lookup sync**: Not used
+    BlockAndCustodyColumns(
+        Arc<SignedBeaconBlock<E>>,
+        CustodyDataColumnList<E>,
+        Vec<ColumnIndex>,
```
jimmygchen: Would be a bit more readable converting into a struct variant.

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -632,6 +637,36 @@ pub fn signature_verify_chain_segment<T: BeaconChainTypes>(
         &chain.spec,
     )?;
 
+    // verify signatures before matching blocks and data
```
jimmygchen: Feel like we could do the cheaper block & blobs pairing before attempting signature and kzg verification - could we even lift the matching (currently in `da_checker.verify_kzg_for_rpc_blocks`) out of this function?

jimmygchen (changes requested):
Hey @dapplion, I've done a quick first pass and added some comments. I'll go through the changes again to make sure I understand them.

beacon_node/network/src/network_beacon_processor/sync_methods.rs
dapplion: > and is also the existing behaviour to not penalise peer(s) - we haven't ran into an scenario where we'd need to penalise peer for this.

- I think we did, for Electra an Attestation bitfield error caused by a malformed attestation turned out to be `BeaconChainError` https://github.com/sigp/lighthouse/pull/7382

beacon_node/network/src/sync/network_context.rs
dapplion: Consider a batch that has blocks and columns and issues 8 columns by range requests. `range_block_component_response` will be called 1+8 times, where `peer_id` is the peer that served each individual _by_range response. Added a comment to clarify this.

beacon_node/network/src/sync/range_sync/batch.rs
dapplion: Renamed

beacon_node/network/src/sync/range_sync/chain.rs
dapplion: Yeah, will add `faulty_batch_columns`. I think we should use a more descriptive error string anyway in the future, like the actual error

beacon_node/network/src/network_beacon_processor/sync_methods.rs
dapplion: Will change to Unexpected

beacon_node/network/src/network_beacon_processor/sync_methods.rs
dapplion: Done

beacon_node/network/src/network_beacon_processor/sync_methods.rs
dapplion: Downgraded to High

beacon_node/network/src/sync/backfill_sync/mod.rs
dapplion: Changed to penalize each column peer once

beacon_node/network/src/network_beacon_processor/sync_methods.rs
dapplion: Removed

beacon_node/network/src/network_beacon_processor/sync_methods.rs
dapplion: Changed to MidToleranceError

beacon_node/network/src/network_beacon_processor/sync_methods.rs
dapplion: Yeah, inside `import_historical_block_batch`. That function in current unstable does:

- verify block signatures
- import

With this PR `import_historical_block_batch` does:

- assert chain of block hashes
- verify kzg proofs
- verify signatures
- match sidecar headers (but it's actually computed before verify signatures)
- import

If you want to keep verification outside `import_historical_block_batch` we should also move the verify signatures out. Should I do that refactor in this PR? I moved the logic inside due to how `import_historical_block_batch` is structured.

beacon_node/beacon_chain/src/data_availability_checker.rs
dapplion: Right, we could and should move `verify_kzg_for_rpc_blocks` outside the da_checker because it doesn't use the cache. It just verifies things and returns available. The types we use for range sync force us to do it this way but I agree it's weird. I would love to refactor this into a more intuitive design but feels out of scope of this PR.

beacon_node/beacon_chain/src/block_verification.rs
dapplion: We should not do that. Consider the case:
- block peer sends us bad signature
- column peer sends us valid signature
- because they don't match we penalize column peer

beacon_node/beacon_chain/src/block_verification_types.rs
```diff
@@ -53,31 +53,61 @@ impl<E: EthSpec> RpcBlock<E> {
...
     pub fn custody_columns(&self) -> Option<&CustodyDataColumnList<E>> {
         match &self.block {
             RpcBlockInner::Block(_) => None,
             RpcBlockInner::BlockAndBlobs(_, _) => None,
-            RpcBlockInner::BlockAndCustodyColumns(_, data_columns) => Some(data_columns),
+            RpcBlockInner::BlockAndCustodyColumns(_, data_columns, _) => Some(data_columns),
+        }
+    }
+
+    pub fn non_matching_blobs_signed_headers(&self) -> Option<Vec<ColumnIndex>> {
```
pawanjay176: This is a very unclear name for the function. 
Can't think of a better name though lol. Maybe add some docs? 
Also, might be clearer if this returns `Vec<usize>` instead of `ColumnIndex`

beacon_node/beacon_chain/src/block_verification_types.rs
```diff
@@ -53,31 +53,61 @@ impl<E: EthSpec> RpcBlock<E> {
...
+                    .iter()
+                    .filter(|blob| &blob.signed_block_header.signature != block.signature())
+                    .map(|blob| blob.index)
+                    .collect(),
+            ),
+            RpcBlockInner::BlockAndCustodyColumns(..) => None,
+        }
+    }
+
+    pub fn non_matching_custody_columns_signed_headers(&self) -> Option<Vec<ColumnIndex>> {
```
pawanjay176: same as above

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -81,20 +81,21 @@ impl PeerGroupAction {
             AvailabilityCheckError::ReconstructColumnsError(_) => None, // internal error
             AvailabilityCheckError::KzgCommitmentMismatch { .. } => None, // should never happen after checking inclusion proof
             AvailabilityCheckError::Unexpected(_) => None,                // internal
-            AvailabilityCheckError::SszTypes(_) => None,                  // ??
-            AvailabilityCheckError::MissingBlobs => None, // TODO(das) internal for now
+            AvailabilityCheckError::MissingBlobs => {
+                Some(PeerGroupAction::block_peer(PeerAction::LowToleranceError))
```
jimmygchen: Missing blob doesn't feel as bad as serving invalid data, could possibly downgrade.

beacon_node/network/src/sync/network_context.rs
jimmygchen: Got it, thanks for adding the comment.

beacon_node/network/src/network_beacon_processor/sync_methods.rs
jimmygchen: All good, just making sure we still have it. I'm fine to leave it as it is and refactor in a separate PR if necessary.

beacon_node/beacon_chain/src/data_availability_checker.rs
jimmygchen: Sure, happy to leave this in a separate PR as some of this is existing code - however I think we should fix this as it will get harder to find where things are and become less maintainable over time. Please add it to the sync roadmap or raise a separate issue so we can keep track.

beacon_node/beacon_chain/src/block_verification.rs
jimmygchen: ok make sense 👍

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -632,6 +637,37 @@ pub fn signature_verify_chain_segment<T: BeaconChainTypes>(
         &chain.spec,
     )?;
 
+    // Verify signatures before matching blocks and data. Otherwise we may penalize blob or column
+    // peers for valid signatures if the block peer sends us an invalid signature.
+    let pubkey_cache = get_validator_pubkey_cache(chain)?;
```
jimmygchen, line 644: Can we put this in a block or add an explicit `drop` for the `pubkey_cache` read guard?

We previously had `drop(pubkey_cache);`

beacon_node/beacon_chain/src/block_verification_types.rs
```diff
@@ -53,31 +54,59 @@ impl<E: EthSpec> RpcBlock<E> {
...
+            RpcBlockInner::BlockAndCustodyColumns {
+                block,
+                data_columns,
+                ..
+            } => Some(non_matching_custody_columns_block_signature(
+                block,
+                &data_columns
+                    .iter()
+                    .map(|data_column| data_column.clone_arc())
+                    .collect::<Vec<_>>(),
```
jimmygchen: In the function we immediately create an `.iter()` from this vec, nit but unnecessary alloc here. We could make the function take an `impl Iterator`

beacon_node/beacon_chain/src/data_availability_checker/error.rs
```diff
@@ -35,13 +32,10 @@ pub enum ErrorCategory {
 impl Error {
     pub fn category(&self) -> ErrorCategory {
         match self {
-            Error::SszTypes(_)
-            | Error::MissingBlobs
-            | Error::MissingCustodyColumns
+            Error::MissingBlobs
+            | Error::MissingCustodyColumns(_)
```
jimmygchen: The categorisation here seem inconsistent with
https://github.com/sigp/lighthouse/blob/9292f82c1c23be6d9a1b9d36015aabf60342518e/beacon_node/network/src/network_beacon_processor/sync_methods.rs#L72

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -37,11 +39,66 @@ pub enum ChainSegmentProcessId {
...
+    fn from_availability_check_error(e: &AvailabilityCheckError) -> Option<Self> {
+        match e {
+            AvailabilityCheckError::InvalidBlobs(_) => {
+                Some(PeerGroupAction::block_peer(PeerAction::LowToleranceError))
+            }
+            AvailabilityCheckError::InvalidColumn(errors) => Some(PeerGroupAction::column_peers(
+                &errors.iter().map(|(index, _)| *index).collect::<Vec<_>>(),
+                PeerAction::LowToleranceError,
+            )),
+            AvailabilityCheckError::ReconstructColumnsError(_) => None, // internal error
```
jimmygchen: We categorise this as `Malicious` in `AvailabilityCheckError::category`

jimmygchen: Wonder if we could reuse `AvailabilityCheckError::category` as we use that for block lookups and gossip message penalties.

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -589,148 +648,77 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
-                        message: format!("{:?}", err_str),
-                        // This is an internal error, don't penalize the peer.
-                        peer_action,
-                    }),
-                )
+
+                if peer_action.is_some() {
+                    // All errors that result in a peer penalty are "expected" external faults the
+                    // node runner can't do anything about
+                    debug!(?e, "Backfill sync processing error");
```
jimmygchen, line 693: Out of scope for this PR, just a note: I think we can improve our logging more now with tracing and consider entering a span at the `send_chain_segment`  and instrument the key child methods, so we have the `ChainSegmentProcessId` logged.

jimmygchen (changes requested):
@dapplion I've re-reviewed the new changes. I'm feeling pretty good about this PR and I think we can merge soon. Nice work 👏 

Just a few more small comments and failing CI that needs fix.

jimmygchen:
I've done a round of testing on `peerdas-devnet-7`
- ✅  supernode genesis sync
- ✅  fullnode genesis
- ❌ backfill sync stuck for both supernode and fullnode at epoch `1298` on `peerdas-devnet-7` (epoch of the checkpoint block slot `41536`), due to `MismatchedBlockRoot`:
  > May 15 12:59:33.603 DEBUG Backfill sync processing error                e: MismatchedBlockRoot { block_root: 0x65cc66f5d5bce0ae2703a6e60f2addbebe43b260b0224f856ba9efa6a99ba2ca, expected_block_root: 0x2db46c3c7e2c5fd5e64b5152b6939c3408e5b3d3ef5d36b293bcad355fd8cd47 } 

Other issues I've noticed:
- Seeing a large number of self rate limiting logs:
  > May 15 06:57:27.494 DEBUG Self rate limiting due to the number of concurrent req
uests  peer_id: 16Uiu2HAmKo7eVTbfjUnKfKNTFM7MsGqUCnmFgrC9N5UfXVfUovc1, protocol:
 data_column_sidecars_by_range, service: "network"

- Lighthouse peers responding with unrequested slots, for both blobs and data columns. 
  > May 15 03:24:38.391 DEBUG Sync RPC request error                        id: 197/195/RangeSync/63/1, method: "BlobsByRangeRequest", error: VerifyError(UnrequestedSlot(Slot(2015))), service: "sync"

beacon_node/beacon_chain/src/data_availability_checker.rs
dapplion: Added

beacon_node/beacon_chain/src/block_verification.rs
dapplion: done

beacon_node/beacon_chain/src/data_availability_checker/error.rs
dapplion: Done

beacon_node/beacon_chain/src/block_verification_types.rs
dapplion: done

beacon_node/network/src/network_beacon_processor/sync_methods.rs
dapplion: Yes we could, but we should move it outside of the beacon_chain crate as we don't have access to PeerAction there. Also should be part of another refactor.

dapplion: Removed `ReconstructColumnsError` for `Unexpected`

dapplion:
> Lighthouse peers responding with unrequested slots, for both blobs and data columns.

Added this commit to have more info on the error https://github.com/sigp/lighthouse/pull/7352/commits/464921194f8be450b16b43362fc0fe1c783dc343 otherwise can you paste the request of those errors? Just grep the ID `197/195/RangeSync/63/1`

beacon_node/beacon_chain/src/historical_blocks.rs
```diff
@@ -67,6 +155,28 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         &self,
         mut blocks: Vec<AvailableBlock<T::EthSpec>>,
     ) -> Result<usize, HistoricalBlockError> {
+        // Verify that blobs or data columns signatures match
+        //
+        // TODO(das): We don't raise the `matching_sidecar_signatures_error` yet. We have to wait to
+        // return an invalid block signature error first. We may want to refactor this order in a
+        // later code change.
+        let matching_sidecar_signatures_error = blocks
```
pawanjay176: Maybe perform this operation only on `blocks_to_import` instead of all the blocks?

beacon_node/network/src/network_beacon_processor/sync_methods.rs
dapplion: Downgraded to HighToleranceError

pawanjay176: I think better to keep it low tolerance as it will be easier to find bugs on quick disconnects

dapplion:
Is this lint error affecting unstable / peerdas-devnet-7? It showed up in CI in this PR but we haven't changed the `ParticipationEpochSummary`. New clippy rules?

```
error: large size difference between variants
  --> consensus/state_processing/src/per_epoch_processing/epoch_processing_summary.rs:12:1
   |
12 | /   pub enum EpochProcessingSummary<E: EthSpec> {
13 | | /     Base {
14 | | |         total_balances: TotalBalances,
15 | | |         statuses: Vec<ValidatorStatus>,
16 | | |     },
   | | |_____- the second-largest variant contains at least 88 bytes
17 | | /     Altair {
18 | | |         progressive_balances: ProgressiveBalancesCache,
19 | | |         current_epoch_total_active_balance: u64,
20 | | |         participation: ParticipationEpochSummary<E>,
21 | | |         sync_committee: Arc<SyncCommittee<E>>,
22 | | |     },
   | | |_____- the largest variant contains at least 360 bytes
23 | |   }
   | |___^ the entire enum is at least 360 bytes
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant
   = note: `-D clippy::large-enum-variant` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(clippy::large_enum_variant)]`
help: consider boxing the large fields to reduce the total size of the enum
   |
20 -         participation: ParticipationEpochSummary<E>,
20 +         participation: Box<ParticipationEpochSummary<E>>,
```

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -100,48 +117,60 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
...
+                        .map(|blocks| (blocks, BatchPeers::new_from_block_peer(block_peer))),
+                )
             }
             RangeBlockDataRequest::DataColumns {
                 requests,
-                expected_custody_columns,
+                expected_column_to_peer,
             } => {
                 let mut data_columns = vec![];
+                let mut column_peers = HashMap::new();
```
pawanjay176, line 154: I don't understand why we are rebuilding this hashmap here. Why can't we just pass `expected_column_to_peer` to `BatchPeers::new`

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -199,106 +228,98 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
...
 impl<I: PartialEq + std::fmt::Display, T> ByRangeRequest<I, T> {
-    fn finish(&mut self, id: I, data: T) -> Result<(), String> {
+    fn finish(&mut self, id: I, data: T, peer_id: PeerId) -> Result<(), String> {
         match self {
             Self::Active(expected_id) => {
                 if expected_id != &id {
                     return Err(format!("unexpected req_id expected {expected_id} got {id}"));
                 }
-                *self = Self::Complete(data);
+                *self = Self::Complete(data, peer_id);
```
pawanjay176, line 312: I don't think we should be adding `PeerId` here. The column request is made to multiple peers. However, this will only contain the peer id for the last column response. This will cause issues if you try to use that peer id for any attributable error upstream

jimmygchen: `ByRangeRequest` is for a single request to a single peer.
The collection of `DataColumnByRange` requests for a single batch is `RangeBlockDataRequest`.

beacon_node/network/src/sync/block_sidecar_coupling.rs
jimmygchen: `expected_column_to_peer` is the map of columns to peers that we sent request out to. 
`column_peers` is a map of columns we received to the peer ids. 

The difference is `columns_requested_to_peer` vs `columns_received_to_peer`.

beacon_node/network/src/network_beacon_processor/sync_methods.rs
jimmygchen: Yeah i think low is fine given we just penalise peer once in a single request now.

beacon_node/beacon_chain/src/block_verification_types.rs
jimmygchen: +1 to `Vec<usize>` for blobs.

What about refactor the signatures to
```rust
fn match_block_and_blobs(&self) -> Result<(), Vec<u64>>
fn match_block_and_columns(&self) -> Result<(), Vec<ColumnIndex>>
```

beacon_node/network/src/sync/block_sidecar_coupling.rs
pawanjay176: wouldn't that be the same? the request ids would ensure that right?
not sure why we want to keep 2 separate structures and keep them in sync

jimmygchen: Yeah true

beacon_node/network/src/network_beacon_processor/sync_methods.rs
jimmygchen: `ReconstructColumnsError` may not be unexpected in the future if we start doing reconstruction on sync right?

beacon_node/network/src/sync/block_sidecar_coupling.rs
pawanjay176: This is what I was thinking https://github.com/pawanjay176/lighthouse/commit/18b2cf3f5aa7ee5dcb1df1bc5a4bd9e126828741

beacon_node/network/src/network_beacon_processor/sync_methods.rs
dapplion: It will still be unexpected. If we verify the columns to be valid, and have >= 50% reconstruction should always succeed. A reconstruction error is something wrong with the underlying crypto lib AFAIK

beacon_node/beacon_chain/src/block_verification_types.rs
dapplion: Renamed

beacon_node/beacon_chain/src/historical_blocks.rs
dapplion: Done

beacon_node/network/src/sync/block_sidecar_coupling.rs
dapplion: This peer tracking mechanism gets replaced anyway in https://github.com/dapplion/lighthouse/pull/45 as we retry individual data_columns_by_range requests. I can use your commit for this intermediary but we'll have to revert it later

beacon_node/network/src/sync/block_sidecar_coupling.rs
pawanjay176: Just to document this for later
Here are some logs:
```
May 15 13:27:37.798 DEBUG Batch processing failed                       batch_epoch: 682, first_block_slot: 21824, chain: 1, last_block_slot: 21855, imported_blocks: 0, error: AvailabilityCheck(MissingCustodyColumns([68, 121])), service: "sync"
May 15 13:27:37.857 DEBUG Batch processing result                       result: FaultyFailure { imported_blocks: 0, peer_action: PeerGroupAction { block_peer: None, column_peer: {121: HighToleranceError, 68: HighToleranceError} }, error: "AvailabilityCheck(MissingCustodyColumns([68, 121]))" }, batch_epoch: 682, batch_state: "[P,p,p,p,D,]", batch: BatchInfo { start_slot: Slot(21824), end_slot: Slot(21856), failed_processing_attempts: [], non_faulty_processing_attempts: 0, failed_download_attempts: [], state: Processing(16Uiu2HAmTrdcGDTXnoHsgz78oNsU3AeKwDFfCAwCTP4aA6q3sHUi), marker: PhantomData<network::sync::range_sync::batch::RangeSyncBatchConfig> }, chain: 1, service: "range_sync"
```
So ideally, it should have downscored the responsible peers here
https://github.com/sigp/lighthouse/blob/31f46a12be5b6f3eaf20d7eb3d90d15d977d98c1/beacon_node/network/src/sync/range_sync/chain.rs#L533-L546

But that never happened, because `batch_peers` was constructed incorrectly in `responses`.
https://github.com/sigp/lighthouse/blob/31f46a12be5b6f3eaf20d7eb3d90d15d977d98c1/beacon_node/network/src/sync/block_sidecar_coupling.rs#L149-L172

The column_peer that is returned in `to_finished()` would be the last responder for that particular request id which isn't what we want. Tbh, i think its hard to maintain the peers from the responses, that's why i suggested reusing the request peers

beacon_node/network/src/sync/block_sidecar_coupling.rs
pawanjay176: Yeah i'm fine with us handling that in another PR, just wanted to point out that this PR isn't downscoring the correct peers by itself yet.

dapplion:
Last remaining CI failure is a clippy warning 

```
error: large size difference between variants
  --> consensus/state_processing/src/per_epoch_processing/epoch_processing_summary.rs:12:1
   |
12 | /   pub enum EpochProcessingSummary<E: EthSpec> {
```

dapplion:
- I have added store tests from https://github.com/sigp/lighthouse/pull/7353

pawanjay176 (approved):
Looks good. Can start testing and fix the peer attribution in the other big sync PR

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -2599,17 +2602,18 @@ async fn weak_subjectivity_sync_test(slots: Vec<Slot>, checkpoint_slot: Slot) {
...
-                let mut sidecar = data_columns[0].clone_arc();
-                let mut_sidecar = Arc::make_mut(&mut sidecar);
-                if mut_sidecar.kzg_proofs[0] == KzgProof::empty() {
+                let mut data_column = (*(data_columns[0]).clone_arc()).clone();
+                if data_column.kzg_proofs[0] == KzgProof::empty() {
                     panic!("kzg_proof is already G1_POINT_AT_INFINITY")
                 }
-                mut_sidecar.kzg_proofs[0] = KzgProof::empty();
+                data_column.kzg_proofs[0] = KzgProof::empty();
+                data_columns[0] = CustodyDataColumn::from_asserted_custody(data_column.into());
```
dapplion, line 2616: Changed `Arc::make_mut` into a direct assignment to make sure we are mutating the original Vec

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -412,14 +415,11 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
 
         let all_data_columns = blocks
             .iter()
-            .filter(|block| self.data_columns_required_for_block(block.as_block()))
```
pawanjay176, line 413: I don't like that we have different filtering for blobs and columns verification here. 
its not expected from the name of the function or the docs, its just a temporary hack to get a test passing.
If we want to deal with this later, i'd rather we merge this without the hack and deal with the failing test before merging peerdas-devnet-7 to unstable. 
cc @jimmygchen

dapplion: I think now tests should pass still with that line. I'll add it back to this PR.

I'll make an issue to remove this line in unstable as I think it's quite problematic

- EDIT: https://github.com/sigp/lighthouse/issues/7501

pawanjay176: Awesome, thanks!

pawanjay176:
@jimmygchen Merge at will

jimmygchen (approved):
LGTM! Thanks for pushing this through 👍

PR #7359, opened 2025-04-26
Make sure to log SyncingChain ID

dapplion:
## Issue Addressed

Debugging an sync issue from @pawanjay176 I'm missing some key info where instead of logging the ID of the SyncingChain we just log "Finalized" (the sync type). This looks like some typo or something was lost in translation when refactoring things.

```
Apr 17 12:12:00.707 DEBUG Syncing new finalized chain                   chain: Finalized, component: "range_sync"
```

This log should include more info about the new chain but just logs "Finalized"

```
Apr 17 12:12:00.810 DEBUG New chain added to sync                       peer_id: "16Uiu2HAmHP8QLYQJwZ4cjMUEyRgxzpkJF87qPgNecLTpUdruYbdA", sync_type: Finalized, new_chain: Finalized, component: "range_sync"
```

## Proposed Changes

- Remove the Display impl and log the ID explicitly for all logs.
- Log more details when creating a new SyncingChain

### DIFF
--- a/beacon_node/network/src/sync/range_sync/chain.rs
+++ b/beacon_node/network/src/sync/range_sync/chain.rs
@@ -13,7 +13,6 @@ use logging::crit;
 use rand::seq::SliceRandom;
 use rand::Rng;
 use std::collections::{btree_map::Entry, BTreeMap, HashSet};
-use std::fmt;
 use strum::IntoStaticStr;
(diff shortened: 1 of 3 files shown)


pawanjay176 (approved):
Nice! I had to add similar logs in my branch that i was testing with.

PR #7361, opened 2025-04-27
Fix missing commas in logs

ackintosh:
## Issue Addressed

In beacon.log, commas are missing in case where multiple spans are present, for example:

```
Apr 05 07:03:07.646 DEBUG Sync's view on execution engine state updated  past_state: Online, new_state: Offline, service: "sync"service: "network_context"
```

A comma should be present between `service: "sync"` and `service: "network_context"` as shown in the example.

## Proposed Changes

This PR:
- makes `build_log_text()` testable and adds unit tests for it
- fixes the missing commas issue

### DIFF
--- a/common/logging/src/tracing_logging_layer.rs
+++ b/common/logging/src/tracing_logging_layer.rs
@@ -464,7 +464,7 @@ mod tests {
             Buffer { data: Vec::new() }
         }
 
-        fn to_string(self) -> String {
+        fn into_string(self) -> String {
             String::from_utf8(self.data).unwrap()
         }
     }
@@ -598,6 +598,6 @@ mod tests {
             &mut writer,
         );
 
-        assert_eq!(expected, &writer.to_string());
+        assert_eq!(expected, &writer.into_string());
     }
 }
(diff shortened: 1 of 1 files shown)


macladson (commented):
This has some of the same ideas as https://github.com/sigp/lighthouse/pull/7239, I'm wondering if you've had a chance to look at that PR? It would be good if we could make them compatible with each other.

macladson:
I've written up [a version of this PR](https://gist.github.com/macladson/3525ac623ea0491ce06fa0740349ef5a) on top of #7239 
There's a few changes (commas are handled slightly differently and the span order for multiple spans is reversed). 
I wanted to make sure you still get the credit for this fix so let me know what you think.

macladson:
Now that #7239 has been merged we can push this fix through. @ackintosh let me know if you want me to make the changes, or if you have time to fix it

ackintosh:
@macladson Thanks for pointing that out! I believe the fix in this PR has already been addressed by #7239. On the current unstable branch, I checked the logs and confirmed that entries with multiple spans are now correctly formatted, with the missing commas resolved:

```
Jul 29 06:51:24.629 DEBUG Sync's view on execution engine state updated  past_state: Online, new_state: Online, service: "network_context", service: "sync"
```

macladson:
Oh that's an interesting side-effect! I'll close this now then but we should keep an eye on the logs in case it still appears in very specific situations :+1:

PR #7367, opened 2025-04-28
blanket signed root

ltitanb:
## Issue Addressed

We ([Helix](https://github.com/gattaca-com/helix) relay), have recently switched our consensus types to Lighthouse. During integration we made a few small quality of life changes that would be great to upstream so we can get rid of the fork. Overall all the changes are quite small and don't change any core logic.

## Proposed Changes

Make a blanket implementation of `SignedRoot` for all impls of `TreeHash`

### DIFF
--- a/consensus/types/src/aggregate_and_proof.rs
+++ b/consensus/types/src/aggregate_and_proof.rs
@@ -144,6 +144,3 @@ impl<E: EthSpec> AggregateAndProof<E> {
         self.selection_proof().verify(validator_pubkey, message)
     }
 }
-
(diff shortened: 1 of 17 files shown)


michaelsproul (changes requested):
Making `SignedRoot` opt-in was an intentional design choice to statically prevent signing of objects that should never be signed.

At a glance, I would prefer to keep it that way. Which type did you find yourself wanting `SignedRoot` on that didn't have it already implemented? Could we add that implementation instead? Or could you add it downstream?

ltitanb:
Noted, will follow the same approach downstream!

PR #7368, opened 2025-04-28
tree hash for blobs bundle

ltitanb:
## Issue Addressed

ref: #7367

## Proposed Changes

Implement `TreeHash` for `BlobsBundle`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2731,6 +2731,8 @@ dependencies = [
  "ssz_types",
  "test_random_derive",
  "tokio",
+ "tree_hash",
+ "tree_hash_derive",
  "types",
  "zeroize",
 ]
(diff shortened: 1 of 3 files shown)


jimmygchen:
I don't believe this is required - please provide rationale if you think otherwise - happy to re-open.

jimmygchen:
Hi @ltitanb 

Sorry for closing this abruptly without details - I'd like to add more info here - in the CL [we don't expect](https://github.com/ethereum/consensus-specs/pull/4187#discussion_r2010761634) to hash this object, that's why this wasn't implemented.

In PeerDAS/Fulu, we've updated the length of `VariableList`  without introducing a new version, hence the hash produced would be different to before. If we need tree hashing on this object, then we'd need to introduce a new `BlobsBundle` version for PeerDAS, which requires a bit more boilerplate.

jimmygchen:
See https://github.com/sigp/lighthouse/blob/759b0612b37f354b70bd00298082e6aacb20ba89/consensus/types/src/lib.rs#L275-L282

ltitanb:
Noted, thanks for the explanation!

Issue #7286, opened 2025-04-09
Refactor `ForkVersionedResponse` and `ForkVersionDeserialize`


michaelsproul:
To make `ForkVersionDeserialize` even more type safe, we could remove the requirement that types implementing the trait also implement `DeserializeOwned`. If we have a type that needs fork info to deserialize correctly, it is probably a footgun for this type to also implement `Deserialize`, because that will bypass the fork-aware deserialization.

To make this viable we would need to:

- Remove the bound from `ForkVersionDeserialize`
- Make the `version` field mandatory in `ForkVersionedResponse` and ensure there is no unintended fallout of making this change.

PoulavBhowmick03:
Working on this

michaelsproul:
Fixed by:

- #7372

PR #7372, opened 2025-04-28
ContextDeserialize and Beacon API Improvements

ethDreamer:
## Issue Addressed

* #7286 
* BeaconAPI is not returning a versioned response when it should for some V1 endpoints
* these [strange functions with vX in the name that still accept `endpoint_version` arguments](https://github.com/sigp/lighthouse/blob/stable/beacon_node/http_api/src/produce_block.rs#L192)

This refactor is a prerequisite to get the fulu EF tests running.

### DIFF
--- a/beacon_node/http_api/src/lib.rs
+++ b/beacon_node/http_api/src/lib.rs
@@ -34,7 +34,7 @@ mod validators;
 mod version;
 use crate::light_client::{get_light_client_bootstrap, get_light_client_updates};
 use crate::produce_block::{produce_blinded_block_v2, produce_block_v2, produce_block_v3};
-use crate::version::fork_versioned_response;
+use crate::version::beacon_response;
 use beacon_chain::{
     attestation_verification::VerifiedAttestation, observed_operations::ObservationOutcome,
     validator_monitor::timestamp_now, AttestationError as AttnError, BeaconChain, BeaconChainError,
@@ -89,18 +89,17 @@ use tokio_stream::{
 use tracing::{debug, error, info, warn};
 use types::AttestationData;
 use types::{
-    fork_versioned_response::EmptyMetadata, Attestation, AttestationShufflingId, AttesterSlashing,
-    BeaconStateError, ChainSpec, Checkpoint, CommitteeCache, ConfigAndPreset, Epoch, EthSpec,
-    ForkName, ForkVersionedResponse, Hash256, ProposerPreparationData, ProposerSlashing,
-    RelativeEpoch, SignedAggregateAndProof, SignedBlindedBeaconBlock, SignedBlsToExecutionChange,
-    SignedContributionAndProof, SignedValidatorRegistrationData, SignedVoluntaryExit, Slot,
-    SyncCommitteeMessage, SyncContributionData,
+    Attestation, AttestationShufflingId, AttesterSlashing, BeaconStateError, ChainSpec, Checkpoint,
+    CommitteeCache, ConfigAndPreset, Epoch, EthSpec, ForkName, Hash256, ProposerPreparationData,
+    ProposerSlashing, RelativeEpoch, SignedAggregateAndProof, SignedBlindedBeaconBlock,
+    SignedBlsToExecutionChange, SignedContributionAndProof, SignedValidatorRegistrationData,
+    SignedVoluntaryExit, Slot, SyncCommitteeMessage, SyncContributionData,
 };
 use validator::pubkey_to_validator_index;
 use version::{
     add_consensus_version_header, add_ssz_content_type_header,
-    execution_optimistic_finalized_fork_versioned_response, inconsistent_fork_rejection,
-    unsupported_version_rejection, V1, V2, V3,
+    execution_optimistic_finalized_beacon_response, inconsistent_fork_rejection,
+    unsupported_version_rejection, ResponseIncludesVersion, V1, V2, V3,
 };
 use warp::http::StatusCode;
 use warp::hyper::Body;
@@ -1723,6 +1722,12 @@ pub fn serve<T: BeaconChainTypes>(
                         .fork_name(&chain.spec)
                         .map_err(inconsistent_fork_rejection)?;
 
+                    let require_version = match endpoint_version {
+                        V1 => ResponseIncludesVersion::No,
+                        V2 | V3 => ResponseIncludesVersion::Yes(fork_name),
+                        _ => return Err(unsupported_version_rejection(endpoint_version)),
+                    };
+
                     match accept_header {
                         Some(api_types::Accept::Ssz) => Response::builder()
                             .status(200)
@@ -1734,9 +1739,8 @@ pub fn serve<T: BeaconChainTypes>(
                                     e
                                 ))
                             }),
-                        _ => execution_optimistic_finalized_fork_versioned_response(
-                            endpoint_version,
-                            fork_name,
+                        _ => execution_optimistic_finalized_beacon_response(
+                            require_version,
                             execution_optimistic,
                             finalized,
                             block,
@@ -1796,9 +1800,15 @@ pub fn serve<T: BeaconChainTypes>(
                         .attestations()
                         .map(|att| att.clone_as_attestation())
                         .collect::<Vec<_>>();
-                    let res = execution_optimistic_finalized_fork_versioned_response(
-                        endpoint_version,
-                        fork_name,
+
+                    let require_version = match endpoint_version {
+                        V1 => ResponseIncludesVersion::No,
+                        V2 | V3 => ResponseIncludesVersion::Yes(fork_name),
+                        _ => return Err(unsupported_version_rejection(endpoint_version)),
+                    };
+
+                    let res = execution_optimistic_finalized_beacon_response(
+                        require_version,
                         execution_optimistic,
                         finalized,
                         &atts,
@@ -1845,9 +1855,8 @@ pub fn serve<T: BeaconChainTypes>(
                             }),
                         _ => {
                             // Post as a V2 endpoint so we return the fork version.
-                            execution_optimistic_finalized_fork_versioned_response(
-                                V2,
-                                fork_name,
+                            execution_optimistic_finalized_beacon_response(
+                                ResponseIncludesVersion::Yes(fork_name),
                                 execution_optimistic,
                                 finalized,
                                 block,
@@ -1901,9 +1910,8 @@ pub fn serve<T: BeaconChainTypes>(
                             }),
                         _ => {
                             // Post as a V2 endpoint so we return the fork version.
-                            let res = execution_optimistic_finalized_fork_versioned_response(
-                                V2,
-                                fork_name,
+                            let res = execution_optimistic_finalized_beacon_response(
+                                ResponseIncludesVersion::Yes(fork_name),
                                 execution_optimistic,
                                 finalized,
                                 &blob_sidecar_list_filtered,
@@ -2063,7 +2071,13 @@ pub fn serve<T: BeaconChainTypes>(
                         })
                         .collect::<Vec<_>>();
 
-                    let res = fork_versioned_response(endpoint_version, fork_name, &attestations)?;
+                    let require_version = match endpoint_version {
+                        V1 => ResponseIncludesVersion::No,
+                        V2 | V3 => ResponseIncludesVersion::Yes(fork_name),
+                        _ => return Err(unsupported_version_rejection(endpoint_version)),
+                    };
+
+                    let res = beacon_response(require_version, &attestations);
                     Ok(add_consensus_version_header(
                         warp::reply::json(&res).into_response(),
                         fork_name,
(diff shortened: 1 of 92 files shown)


Alleysira:
I don't know Rust :) Just want to add that endpoint `getPendingDeposits` and `getPendingPartialWithdrawals` also missing `version` field, similar to #7440.

ethDreamer:
> I don't know Rust :) Just want to add that endpoint `getPendingDeposits` and `getPendingPartialWithdrawals` also missing `version` field, similar to #7440.

Thank you! I've added that fix into this PR :)

beacon_node/http_api/src/lib.rs
```diff
@@ -1723,6 +1738,12 @@ pub fn serve<T: BeaconChainTypes>(
                         .fork_name(&chain.spec)
                         .map_err(inconsistent_fork_rejection)?;
 
+                    let require_version = match endpoint_version {
+                        V1 => ResponseIncludesVersion::No,
+                        V2 | V3 => ResponseIncludesVersion::Yes(fork_name),
```
michaelsproul: There is no V3 of `GET /beacon/blocks` at the moment, maybe this should be an error? If one gets added in future we may not want to be returning a V2 response here (it could be wrong)

I can see the argument for just leaving it too.

beacon_node/http_api/src/lib.rs
```diff
@@ -1796,9 +1816,15 @@ pub fn serve<T: BeaconChainTypes>(
                         .attestations()
                         .map(|att| att.clone_as_attestation())
                         .collect::<Vec<_>>();
-                    let res = execution_optimistic_finalized_fork_versioned_response(
-                        endpoint_version,
-                        fork_name,
+
+                    let require_version = match endpoint_version {
+                        V1 => ResponseIncludesVersion::No,
+                        V2 | V3 => ResponseIncludesVersion::Yes(fork_name),
```
michaelsproul: Same here, no V3 of this endpoint (`beacon/block/attestations`)

beacon_node/http_api/src/lib.rs
```diff
@@ -2063,7 +2087,13 @@ pub fn serve<T: BeaconChainTypes>(
                         })
                         .collect::<Vec<_>>();
 
-                    let res = fork_versioned_response(endpoint_version, fork_name, &attestations)?;
+                    let require_version = match endpoint_version {
+                        V1 => ResponseIncludesVersion::No,
+                        V2 | V3 => ResponseIncludesVersion::Yes(fork_name),
```
michaelsproul: No V3

beacon_node/http_api/src/lib.rs
```diff
@@ -2152,7 +2182,13 @@ pub fn serve<T: BeaconChainTypes>(
                             })
                             .collect::<Vec<_>>();
 
-                        let res = fork_versioned_response(endpoint_version, fork_name, &slashings)?;
+                        let require_version = match endpoint_version {
+                            V1 => ResponseIncludesVersion::No,
+                            V2 | V3 => ResponseIncludesVersion::Yes(fork_name),
```
michaelsproul: No V3

consensus/types/src/beacon_response.rs
```diff
@@ -0,0 +1,239 @@
...
+    pub metadata: M,
+    pub data: T,
+}
+
+// Used for responses to V1 endpoints that don't have a version field.
+/// The metadata of type M should be set to `EmptyMetadata` if you don't care about adding fields other than
+/// version. If you *do* care about adding other fields you can mix in any type that implements
+/// `Deserialize`.
+#[derive(Debug, PartialEq, Clone, Serialize)]
+pub struct UnVersionedResponse<T, M = EmptyMetadata> {
```
michaelsproul: Personally I think this looks nicer? "unversioned" is a valid word on its own

```suggestion
pub struct UnversionedResponse<T, M = EmptyMetadata> {
```

common/eth2/src/lib.rs
```diff
@@ -2383,7 +2392,8 @@ impl BeaconNodeHttpClient {
             )
             .await?;
 
-        self.get(path).await
+        // TODO: verify this is fork versioned.. couldn't find API docs for this endpoint
```
michaelsproul: Yes, it is: https://ethereum.github.io/beacon-APIs/?urls.primaryName=v2.5.0#/Validator/produceBlindedBlock

This API has been deprecated and deleted though, so we could probably remove our implementation on server side and client side.

common/eth2/src/types.rs
```diff
@@ -1050,51 +1049,71 @@ pub struct SseExtendedPayloadAttributesGeneric<T> {
...
+    {
+        #[derive(Deserialize)]
+        struct Helper {
+            proposal_slot: Slot,
+            proposer_index: u64,
+            parent_block_root: Hash256,
+            parent_block_number: u64,
+            parent_block_hash: ExecutionBlockHash,
+            payload_attributes: serde_json::Value,
+        }
```
michaelsproul: This `Helper` is missing `serde_utils::quoted_u64`.

Can't we just use `SseExtendedPayloadAttributesGeneric<serde_json::Value>` like we were using previously? As far as I can see it's exactly the same

consensus/context_deserialize/src/impls.rs
```diff
@@ -0,0 +1,102 @@
+use crate::ContextDeserialize;
+use serde::de::{Deserialize, DeserializeSeed, Deserializer, SeqAccess, Visitor};
+use std::marker::PhantomData;
+use std::sync::Arc;
```
michaelsproul, line 4:
```suggestion
use std::sync::Arc;

```

consensus/context_deserialize/src/impls.rs
```diff
@@ -0,0 +1,102 @@
...
+        Ok(Arc::new(C::context_deserialize(deserializer, context)?))
+    }
+}
+
+impl<'de, C, T> ContextDeserialize<'de, T> for Vec<C>
+where
+    C: ContextDeserialize<'de, T>,
+    T: Clone,
+{
+    fn context_deserialize<D>(deserializer: D, context: T) -> Result<Self, D::Error>
```
michaelsproul: T for context and C for the element of the Vec seems a bit backwards, but it's also fine

consensus/types/src/runtime_var_list.rs
```diff
@@ -217,6 +218,21 @@ where
...
+    C: ContextDeserialize<'de, ForkName>,
+{
+    fn context_deserialize<D>(deserializer: D, context: ForkName) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        // first parse out a Vec<C> using the Vec<C> impl you already have
+        let vec: Vec<C> = Vec::context_deserialize(deserializer, context)?;
+        let len = vec.len();
+        Ok(RuntimeVariableList::from_vec(vec, len))
```
michaelsproul: I actually think this is a use case for `ContextDeserialize<'de, usize>` if we can make that work.

I don't think it makes sense to deserialize an arbitrary length `RuntimeVariableList` (i.e. using `vec.len()` as `max_len`), because the whole point of `RuntimeVariableList` is that we provide a runtime bound and check it.

I can see that removing this impl breaks the HTTP client code for `get_blobs`. Maybe we could add the `ContextDeserialize<'de, usize>` impl and try to use that there instead. It may require fussing around with `ExecutionOptimisticFinalizedBeaconResponse` though in order to also parse the metadata. We kind of just want the `ChainSpec` as part of the context as well so we can get the right blob bound.

consensus/context_deserialize_derive/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+                )
+                .to_compile_error()
+                .into();
+            }
+        }
+    }
+
+    if ctx_types.is_empty() {
+        return quote! {
+            compile_error!("Usage: #[context_deserialize(Type1, Type2, ..., bound = \"...\")]");
```
michaelsproul, line 64: It looks like we've got a lot of machinery in this proc macro that we aren't using/testing (using more than 1 context type, and using `bound`). Did you write this code by hand or is it based on another proc macro?

I think if we want to keep these features we should add some tests that exercise these code paths.

michaelsproul (commented):
Really nice and clean PR on the whole.

I just had a few comments, many of which I've pushed preliminary fixes for. I'll link the commits under each of my comments so you can thumbs up them (or we can tweak further).

consensus/types/src/beacon_response.rs
michaelsproul: I've changed this in:

- 2ce2b28b1b455aa3378ad4cea4c2fe754bedb968

Lemme know if you prefer to change it back.

common/eth2/src/lib.rs
michaelsproul: Removed this TODO in this commit:

- 0a65f0f2a58e12259b6354838c280aaf2b153832

And opened an issue for the deletion of that API:

- https://github.com/sigp/lighthouse/issues/7456

common/eth2/src/types.rs
michaelsproul: I've made this change here:

- 4b6e9f286596fb1c33e6ccb54d01d2c8875c77f3

I think it's OK.

consensus/context_deserialize/src/milhouse.rs
```diff
@@ -0,0 +1,45 @@
...
+    C: Clone,
+{
+    fn context_deserialize<D>(deserializer: D, context: C) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        // First deserialize as a Vec.
+        // This is not the most efficient implementation as it allocates a temporary Vec. In future
+        // we could write a more performant implementation using `List::builder()`.
+        let vec = Vec::<T>::context_deserialize(deserializer, context)?;
```
michaelsproul, line 19: I simplified this impl in this commit:

- https://github.com/sigp/lighthouse/pull/7372/commits/14c3ce6364d8ddb55044662d6da898501b338ca2

It's the same efficiency as what you had + more obviously correct IMO. If we want a faster impl in future, the option is there.

beacon_node/http_api/src/lib.rs
ethDreamer: Yeah that makes sense, I'll remove the `V3`'s from the endpoints that don't have them.

consensus/types/src/beacon_response.rs
ethDreamer: LGTM!

consensus/context_deserialize/src/milhouse.rs
ethDreamer: Awesome! Thank you :)

common/eth2/src/lib.rs
ethDreamer: sounds good to me! thanks!

common/eth2/src/types.rs
ethDreamer: Good call! I missed that. Thanks.

beacon_node/http_api/src/lib.rs
michaelsproul: Ok cool, sounds good!

consensus/context_deserialize/src/impls.rs
ethDreamer: Yeah good point. I swapped them :)

consensus/types/src/runtime_var_list.rs
ethDreamer: I've tried to implement this in [`6008895`](https://github.com/sigp/lighthouse/pull/7372/commits/6008895e565f26b3882271a98e832c910c5bce00). Let me know what you think!

consensus/context_deserialize_derive/src/lib.rs
ethDreamer: I've added tests in [`e1852a7`](https://github.com/sigp/lighthouse/pull/7372/commits/e1852a7fd84dfd23434923db627ea549b9c4cb3e). Let me know if this is good :)

michaelsproul:
looks like those test failures are not flakes

> could not match any variant of PublishBlockRequest

beacon_node/http_api/src/lib.rs
```diff
@@ -1405,21 +1420,30 @@ pub fn serve<T: BeaconChainTypes>(
         .and(warp::path("beacon"))
         .and(warp::path("blocks"))
         .and(warp::path::end())
-        .and(warp_utils::json::json())
+        .and(warp::body::json())
+        .and(consensus_version_header_filter)
```
michaelsproul, line 1424: I think this is technically a breaking change, but this endpoint is deprecated anyway and we should probably delete it (and update our tests not to use it).

michaelsproul (approved):
Looks good to me! Really nice changes, can't wait to get them merged

beacon_node/http_api/src/lib.rs
michaelsproul: I've opened an issue to just delete these endpoints:

- https://github.com/sigp/lighthouse/issues/7478

consensus/types/src/runtime_var_list.rs
michaelsproul: I like!

michaelsproul:
@mergify requeue

Issue #7196, opened 2025-03-23
Remove references to `slog`


michaelsproul:
There's a TODO added in this PR about tracing:

- https://github.com/sigp/lighthouse/pull/7194

https://github.com/sigp/lighthouse/blob/cd5632cdd45c54d45e1fe4b36bc17561590302ff/common/logging/src/sse_logging_components.rs#L1

There might be some more `slog` comments lurking.

_Someone from SigP_ should go through an remove/rewrite them. We won't accept external PRs for this issue.

chong-he:
Closed by #7378

PR #7378, opened 2025-04-30
Change slog to tracing in comments

chong-he:
## Issue Addressed

* #7196


common/logging/src/sse_logging_components.rs
```diff
@@ -1,5 +1,4 @@
-// TODO(tracing) fix the comments below and remove reference of slog::Drain
-//! This module provides an implementation of `slog::Drain` that optionally writes to a channel if
+//! This module provides an implementation of `tracing` that optionally writes to a channel if
```
macladson:
```suggestion
//! This module provides an implementation of `tracing_subscriber::layer::Layer` that optionally writes to a channel if
```

macladson (changes requested):
We can be more precise here and mention the actual trait we are implementing like the original comment

macladson (approved):

Issue #7360, opened 2025-04-27
Range sync process_completed_batches assumes processing_target exists


dapplion:
@pawanjay176 reported the following error on peerdas devnets `Batch not found for current processing target 214`:

```
Apr 17 12:12:01.763 CRIT  Chain removed                                 sync_type: Finalized, chain: Finalized, reason: WrongChainState("Batch not found for current processing target 214"), op: "batch processing result", component: "range_sync"
```

The SyncingChain wanted to send batch epoch 214 for processing but was not found. The log is missing the chain ID, but from a previous log we know it's ID 2. 

```
Apr 17 12:12:01.763 DEBUG Chain advanced                                previous_start: 212, new_start: 213, processing_target: 213, chain: 2, service: "range_sync"
```

On `advance_chain(validating_epoch)` the `processing_target` is forced set to advance to at least `validating_epoch = 213`.

Batch 214 for chain 2 is never downloaded. I don't know if that epoch is within the chain target or not. Looking at received status messages, all of them are way ahead of 214.

```
Apr 17 12:12:01.236 DEBUG Peer transitioned sync state                  peer_id: 16Uiu2HAmRgAfmQwtyNKdEpbvCRttMuGj6FpTigPb49v3rZqNuvoD, new_state: "Advanced", our_head_slot: 6847, our_finalized_epoch: 212, their_head_slot: 58779, their_finalized_epoch: 1793, is_connected: true, service: "sync"
```

There are two conditions where we may not request batches:
- No idle peers
- in optimistic start && no good_peers_on_sampling_subnets

Then `process_completed_batches` assumes that the batch at epoch 214 exists, which is not true. The bug appears to be `process_completed_batches`, assuming that the target batch exists + increasing `processing_target` without checks.

### Proposed solution

We can change `process_completed_batches` into finding the first batch `AwaitingProcessing` and send it. This will remove the error condition where `processing_target` is inconsistent.

@AgeManning any remarks?

pawanjay176:
> I don't know if that epoch is within the chain target or not. 

FYI, I checked and it was within the chain target. 

I think the `good_peers_on_sampling_subnets ` check in `include_next_batch` breaks some invariant we had before where we expect a batch to be present in the `SyncingChain` if `request_batches` is called.

AgeManning:
Ok, so this took me a bit to get my head around. 

I was very surprised that this would error when we just didn't have idle peers. I thought for sure, we'd have caught this by now. 

[...]

PR #7391, opened 2025-05-02
Allow for sync state where batch is unknown

AgeManning:
#7360 shows a bug where we attempt to process a batch that is not existent. 

It can be non-existent for two reasons:
1. We couldn't download it because there were no idle peers
2. We are in an optimistic start and had no good peers on sampling subnets

Under these conditions, the chain will be removed because there process_id pointer doesn't point to a batch. 

This PR permits this state, by simply removing the error catch.


beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -434,12 +434,15 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
             // target when there is no sampling peers available. This is a valid state and should not
             // return an error.
             return Ok(KeepChain);
-        } else {
-            return Err(RemoveChain::WrongChainState(format!(
-                "Batch not found for current processing target {}",
-                self.processing_target
```
dapplion, line 411: this will be a rare event, should we add a debug log? If we ever wonder why the SyncingChain is not processing anything

AgeManning: Yeah good call.

AgeManning: I don't think it will ever stop processing, something will progress it, but agree a log would be handy

dapplion (approved):
LGTM

michaelsproul (approved):
Gonna test this for v7.1.0 and include it

michaelsproul:
Actually nah, lets leave this for 7.2

jimmygchen (approved):
Looks good to me too

PR #7387, opened 2025-05-01
docs: fix broken link to voluntary exit guide

GarmashAlex:
Updated the broken voluntary exit guide link to a working link

PR #7392, opened 2025-05-02
docs (faq.md): update 404-error link 

gap-editor:


macladson (commented):
This is not the correct link. A fix for the broken link already [has a PR](https://github.com/sigp/lighthouse/pull/7387) so I will close this one

gap-editor:
> This is not the correct link. A fix for the broken link already [has a PR](https://github.com/sigp/lighthouse/pull/7387) so I will close this one

@macladson my bad, you are right. but in #7387 invalid link is on 326 line, while my is on 243. sol let me fix my mistake, please!

macladson:
Ah you are right, I just saw the same link and assumed it was the same line, sorry about that!
I'll reopen it for now, although we might want to just combine the two PRs since they are related

gap-editor:
@macladson fixed link! sorry for my mistake and thank you for reopening!

macladson:
Ah I see why I got confused. They actually are the same line, it's just that your branch is correctly targeting `unstable` whereas the other PR is targeting `stable` which is why the line numbers are different. I'll give the original author a chance to fix their PR since they got in first but if I don't hear from them tomorrow I'll merge yours

macladson:
Completed in #7387

PR #7393, opened 2025-05-02
simulator: Fix `Failed to initialize dependency logging`

ackintosh:
## Issue Addressed

The simulator works but always emits the following message:

```
$ cargo run --release --bin simulator basic-sim
...
...
Failed to initialize dependency logging: attempted to set a logger after the logging system was already initialized
...
...
```

## Proposed Changes

This PR removes the initialization with `env_logger`. 

(Update)
With https://github.com/sigp/lighthouse/pull/7433 merged, the libp2p/discv5 logs are saved in separate files and respect the `RUST_LOG` env var for log level configuration.


jimmygchen (approved):
Nice, I noticed the warning too but didn't look into it - this is what it was. Thanks 🙏

testing/simulator/Cargo.toml
```diff
@@ -7,13 +7,10 @@ edition = { workspace = true }
 
 [dependencies]
 clap = { workspace = true }
-env_logger = { workspace = true }
 environment = { workspace = true }
-eth2_network_config = { workspace = true }
 execution_layer = { workspace = true }
 futures = { workspace = true }
 kzg = { workspace = true }
-logging = { workspace = true }
```
jimmygchen: @ackintosh it turns out we're still using the `logging` crate for `build_workspace_filter`, I think we add this back?

jimmygchen (changes requested):
See comment above re failing CI.

jimmygchen (approved):
LGTM!

Issue #7377, opened 2025-04-30
Update `DataColumnSidecarsByRoot` request to use `DataColumnsByRootIdentifier`


jimmygchen:
See spec: https://github.com/ethereum/consensus-specs/pull/4284

The request structure for `DataColumnSidecarsByRoot` has been modified to reduce the size of the request message. This will likely be required in the next devnet (peerdas-devnet-7)

jimmygchen:
@SunnysidedJ has expressed interest working on this. For some reasons I can't seem to assign this to him.

PoulavBhowmick03:
Can I work on this if possible? @jimmygchen

jimmygchen:
@SunnysidedJ is already working on this. 
Thanks for offering help though @PoulavBhowmick03

PoulavBhowmick03:
> [@SunnysidedJ](https://github.com/SunnysidedJ) is already working on this. Thanks for offering help though [@PoulavBhowmick03](https://github.com/PoulavBhowmick03)

Got it, no issues, would love to work on other tasks related to das!

jimmygchen:
Completed in #7399

PR #7399, opened 2025-05-05
`peerdas-devnet-7`: update `DataColumnSidecarsByRoot` request to use `DataColumnsByRootIdentifier`

SunnysidedJ:
## Issue Addressed

Update DataColumnSidecarsByRoot request to use DataColumnsByRootIdentifier #7377

## Proposed Changes

As described in https://github.com/ethereum/consensus-specs/pull/4284


SunnysidedJ:
Status: testing

SunnysidedJ:
Status: review ready

consensus/types/src/data_column_sidecar.rs
```diff
@@ -32,6 +34,51 @@ pub struct DataColumnIdentifier {
     pub index: ColumnIndex,
 }
```
jimmygchen: `DataColumnIdentifier` can be removed

beacon_node/network/src/sync/network_context/requests/data_columns_by_root.rs
```diff
@@ -12,14 +15,16 @@ pub struct DataColumnsByRootSingleBlockRequest {
 
 impl DataColumnsByRootSingleBlockRequest {
     pub fn into_request(self, spec: &ChainSpec) -> DataColumnsByRootRequest {
+        let number_of_columns = spec.number_of_columns as usize;
+        // TODO we aren't handling the case where self.indices > NUMBER_OF_COLUMNS defined by the
+        // spec. Do we do this else where? I think we shall use RuntimeVariableList::new() and
+        // handle errors.
```
jimmygchen: Yes agree that we should use `RuntimeVariableList::new()` instead, perhaps modify this function to `try_into_request` and return `Result` here

beacon_node/lighthouse_network/src/rpc/codec.rs
```diff
@@ -596,10 +597,12 @@ fn handle_rpc_request<E: EthSpec>(
...
             DataColumnsByRootRequest {
-                data_column_ids: RuntimeVariableList::from_ssz_bytes(
-                    decoded_buffer,
-                    spec.max_request_data_column_sidecars as usize,
-                )?,
+                data_column_ids:
+                    <RuntimeVariableList<DataColumnsByRootIdentifier>>::from_ssz_bytes_with_nested(
+                        decoded_buffer,
+                        spec.max_request_blocks(current_fork),
+                        spec.number_of_columns as usize,
```
jimmygchen, line 604: I see why you're using `max_request_blocks` here, because we don't know how many columns are in this request before decoding, and `MAX_REQUEST_DATA_COLUMN_SIDECARS` is derived from `max_request_blocks`.

However I think this will fail in the case where peer requests less than `number_of_columns` but more than `max_request_blocks` even though it may still be within the `MAX_REQUEST_DATA_COLUMN_SIDECARS` limit.

I think we can pass `MAX_REQUEST_DATA_COLUMN_SIDECARS` to the decode function and do the validation in the decode there instead?

consensus/types/src/data_column_sidecar.rs
```diff
@@ -32,6 +34,51 @@ pub struct DataColumnIdentifier {
...
+            return Ok(RuntimeVariableList::empty(max_len));
+        }
+
+        let vec = ssz::decode_list_of_variable_length_items::<Vec<u8>, Vec<Vec<u8>>>(
+            bytes,
+            Some(max_len),
+        )?
+        .into_iter()
+        .map(|bytes| {
+            // Split manually: first 32 bytes = block_root and next 4 bytes = vector tag
```
jimmygchen: I think it would be a bit cleaner to use `SszDecoderBuilder` here to decode, and call into a separate `DataColumnByRootIdentifier` decode function

jimmygchen: Something like this:

```rust
let mut builder = ssz::SszDecoderBuilder::new(&bytes);
builder.register_type::<Hash256>()?;
builder.register_anonymous_variable_length_item()?;

let mut decoder = builder.build()?;
let block_root = decoder.decode_next()?;
let indices = decoder.decode_next_with(|bytes| DataColumnsByRootIdentifier::from_ssz_bytes(bytes, num_columns))?;
Ok(DataColumnsByRootIdentifier {
    block_root,
    indices,
})
```

jimmygchen (changes requested):
@SunnysidedJ thanks a lot for this PR!
I'm still reviewing but thought I'd post my comments I have so far. I'll continue review this afternoon.

beacon_node/lighthouse_network/src/rpc/codec.rs
jimmygchen: Oh sorry you're actually right to use `max_request_blocks`, I missed this in the latest spec:
```
(
  List[DataColumnsByRootIdentifier, MAX_REQUEST_BLOCKS_DENEB]
)
```

jimmygchen: Please ignore above, resolving this comment. This is actually simpler.

consensus/types/src/chain_spec.rs
```diff
@@ -1781,7 +1793,10 @@ fn default_max_blobs_by_root_request() -> usize {
 }
 
 fn default_data_columns_by_root_request() -> usize {
-    max_data_columns_by_root_request_common(default_max_request_data_column_sidecars())
+    max_data_columns_by_root_request_common(
+        default_max_request_data_column_sidecars(),
```
jimmygchen: this should be `default_max_request_blocks()`?

consensus/types/src/chain_spec.rs
```diff
@@ -2083,6 +2098,7 @@ impl Config {
             max_blobs_by_root_request: max_blobs_by_root_request_common(max_request_blob_sidecars),
             max_data_columns_by_root_request: max_data_columns_by_root_request_common(
                 max_request_data_column_sidecars,
```
jimmygchen: `default_max_request_blocks()`

beacon_node/lighthouse_network/src/rpc/methods.rs
```diff
@@ -479,32 +479,17 @@ impl BlobsByRootRequest {
...
-    pub data_column_ids: RuntimeVariableList<DataColumnIdentifier>,
+    pub data_column_ids: RuntimeVariableList<DataColumnsByRootIdentifier>,
 }
 
 impl DataColumnsByRootRequest {
-    pub fn new(data_column_ids: Vec<DataColumnIdentifier>, spec: &ChainSpec) -> Self {
+    pub fn new(data_column_ids: Vec<DataColumnsByRootIdentifier>, spec: &ChainSpec) -> Self {
         let data_column_ids = RuntimeVariableList::from_vec(
             data_column_ids,
             spec.max_request_data_column_sidecars as usize,
```
jimmygchen: This should be `MAX_REQUEST_BLOCKS_DENEB`

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -360,24 +360,26 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
         let mut send_data_column_count = 0;
 
-        for data_column_id in request.data_column_ids.as_slice() {
-            match self.chain.get_data_column_checking_all_caches(
-                data_column_id.block_root,
-                data_column_id.index,
+        for data_column_ids_by_root in request.data_column_ids.as_slice() {
+            match self.chain.get_data_columns_checking_all_caches(
+                data_column_ids_by_root.block_root,
+                &Vec::from(data_column_ids_by_root.indices.clone()),
```
jimmygchen: I think we can do `data_column_ids_by_root.indices.as_slice()` here to avoid cloning?

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -405,16 +405,18 @@ impl<T: BeaconChainTypes> DataAvailabilityCheckerInner<T> {
     }
 
     /// Fetch a data column from the cache without affecting the LRU ordering
```
jimmygchen:
```suggestion
    /// Fetch data columns of a given `block_root` from the cache without affecting the LRU ordering
```

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -1106,23 +1106,33 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
-            .get_data_column(&DataColumnIdentifier { block_root, index })?
+            .get_data_columns(block_root)?
         {
-            return Ok(Some(column));
-        }
+            Ok(Some(all_cols))
+        } else if let Some(all_cols) = self.early_attester_cache.get_data_columns(block_root) {
+            Ok(Some(all_cols))
+        } else {
+            self.get_data_columns(&block_root)
```
jimmygchen: We probably want to avoid loading all data columns from disk and just load the required one here.

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -1106,23 +1106,33 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+        };
```
jimmygchen: Doesn't look like we need to wrap it with `Result` here and unwrap it below.

We could simplify this to something like

```rust
        let columns = if let Some(all_cols) = self
            .data_availability_checker
            .get_data_columns(block_root)?
        {
            Some(all_cols)
        } else if let Some(all_cols) = self.early_attester_cache.get_data_columns(block_root) {
            Some(all_cols)
        } else {
            self.get_data_columns(&block_root)?
        };
```

and then return `Ok(columns)` after filtering.

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -1106,23 +1106,33 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
 
-        if let Some(columns) = self.early_attester_cache.get_data_columns(block_root) {
-            return Ok(columns.iter().find(|c| c.index == index).cloned());
+        if let Ok(Some(columns)) = columns {
+            return Ok(Some(
+                columns
+                    .iter()
+                    .filter(|col| indices.contains(&col.index))
+                    .cloned()
+                    .collect(),
```
jimmygchen: you could do `into_iter` to avoid having to clone the elements

jimmygchen (changes requested):
I've added some more comments, let me know what you think. Thanks!

beacon_node/beacon_chain/src/beacon_chain.rs
SunnysidedJ: I can't see a way data columns to be not read by filtering out based on the index since `DataColumnSidecar` needs to be read and decoded first to get indices. We may gain some performance by reducing a number of clones of `Arc<DataColumnSidecar` and put less columns into cache (not sure if it's useful).

testing/ef_tests/tests/tests.rs
```diff
@@ -666,14 +666,6 @@ mod ssz_static {
             .run_for_feature(FeatureName::Fulu);
     }
 
-    #[test]
```
SunnysidedJ, line 669: Not sure if it could be removed

pawanjay176: Need to replace this with the new `DataColumnsByRootIdentifier` type

testing/ef_tests/src/type_name.rs
```diff
@@ -58,7 +58,6 @@ type_name_generic!(BeaconBlockBodyFulu, "BeaconBlockBody");
 type_name!(BeaconBlockHeader);
 type_name_generic!(BeaconState);
 type_name!(BlobIdentifier);
-type_name!(DataColumnIdentifier);
```
pawanjay176, line 61: needs replacement with the new type as EF tests usually cover all types in the spec

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -201,11 +201,8 @@ impl<T: BeaconChainTypes, O: ObservationStrategy> GossipVerifiedDataColumn<T, O>
         )
     }
 
-    pub fn id(&self) -> DataColumnIdentifier {
-        DataColumnIdentifier {
-            block_root: self.block_root,
-            index: self.data_column.index(),
-        }
+    pub fn id(&self) -> ColumnIndex {
```
pawanjay176: maybe worth renaming the method here

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: The data columns are stored in the blobs db using the block root and column index as key, so we should be able to retrieve it without loading all of them.

jimmygchen: https://github.com/sigp/lighthouse/blob/70850fe58d5676d9072d418f04f4db7257c1513b/beacon_node/store/src/hot_cold_store.rs#L860-L870

SunnysidedJ: Thanks, made changes regarding this!

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -1111,27 +1111,22 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         block_root: Hash256,
         indices: &[ColumnIndex],
     ) -> Result<Option<DataColumnSidecarList<T::EthSpec>>, Error> {
+        let indices: HashSet<ColumnIndex> = indices.iter().copied().collect();
+        let filter = |cols: DataColumnSidecarList<_>| {
+            cols.into_iter()
+                .filter(|col| indices.contains(&col.index))
+                .collect()
+        };
```
jimmygchen: I see a potential issue with this filter approach - if any of the cache contains partial results, this will return the partial results instead of trying to load the missing items from disk.

e.g. let say we have columns 1 - 10 in disk, but only 1-5 in one of the caches, and columns 1-10 is requested, then this function only returns 1-5. I think we might want to load columns individually to prevent this?

SunnysidedJ: Agreed. My bad. Is there a chance that other caches also return partially?

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -2035,22 +2048,53 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
     pub fn get_data_columns(
         &self,
         block_root: &Hash256,
+        mut indices: Option<&mut HashSet<ColumnIndex>>,
     ) -> Result<Option<DataColumnSidecarList<E>>, Error> {
-        if let Some(columns) = self.block_cache.lock().get_data_columns(block_root) {
+        if let Some(columns) = self
+            .block_cache
+            .lock()
+            .get_data_columns(block_root, &mut indices)
```
jimmygchen: I think we could potentially return partial results here as well.

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -1206,8 +1235,11 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
     pub fn get_data_columns(
         &self,
         block_root: &Hash256,
+        indices: Option<&mut HashSet<ColumnIndex>>,
```
jimmygchen: I think we can avoid having to supply a mutable ref on these interfaces across layers.

consensus/types/src/data_column_sidecar.rs
```diff
@@ -23,13 +25,47 @@ pub type ColumnIndex = u64;
...
-#[derive(
-    Serialize, Deserialize, Encode, Decode, TreeHash, Copy, Clone, Debug, PartialEq, Eq, Hash,
-)]
-pub struct DataColumnIdentifier {
+/// Container of data columns of a block root.
+#[derive(Serialize, Deserialize, Encode, Clone, Debug, PartialEq, Eq, Hash)]
+pub struct DataColumnsByRootIdentifier {
     pub block_root: Hash256,
-    pub index: ColumnIndex,
+    pub indices: RuntimeVariableList<ColumnIndex>,
```
jimmygchen: nit: we don't really use the field name for encoding but we try to match the field names in the spec. I'll update this to `columns`

beacon_node/beacon_chain/src/data_column_verification.rs
jimmygchen: method removed

testing/ef_tests/src/type_name.rs
jimmygchen: added

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: https://github.com/sigp/lighthouse/pull/7399#discussion_r2079686498

consensus/types/src/data_column_sidecar.rs
```diff
@@ -23,13 +24,47 @@ pub type ColumnIndex = u64;
...
+/// Identifies a set of data columns associated with a specific beacon block.
+#[derive(Encode, Clone, Debug, PartialEq)]
+pub struct DataColumnsByRootIdentifier {
     pub block_root: Hash256,
-    pub index: ColumnIndex,
+    pub columns: RuntimeVariableList<ColumnIndex>,
+}
+
+impl RuntimeVariableList<DataColumnsByRootIdentifier> {
+    pub fn from_ssz_bytes_with_nested(
```
jimmygchen, line 35: I think we can simplify this by removing the logic to decode the variable list and just reuse `RuntimeVariableList::from_ssz_bytes`

beacon_node/lighthouse_network/src/rpc/codec.rs
```diff
@@ -596,10 +597,12 @@ fn handle_rpc_request<E: EthSpec>(
...
         SupportedProtocol::DataColumnsByRootV1 => Ok(Some(RequestType::DataColumnsByRoot(
             DataColumnsByRootRequest {
-                data_column_ids: RuntimeVariableList::from_ssz_bytes(
-                    decoded_buffer,
-                    spec.max_request_data_column_sidecars as usize,
-                )?,
+                data_column_ids:
+                    <RuntimeVariableList<DataColumnsByRootIdentifier>>::from_ssz_bytes_with_nested(
+                        decoded_buffer,
+                        spec.max_request_blocks(current_fork),
```
jimmygchen, line 603: This is inconsistent with
https://github.com/sigp/lighthouse/blob/d54bffcb787d957b257ea5abd257674118674c95/beacon_node/lighthouse_network/src/rpc/methods.rs#L489

The spec says Deneb but I think current_fork would be compliant too and forward compatible. We do the same  for `max_request_blob_sidecars` above too. I'll update the other spot to use `current_fork`.

jimmygchen: Updated in https://github.com/sigp/lighthouse/pull/7399/commits/7b0e862c93fb3c2ae307f71ab1f1a976436e50ff

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -1106,23 +1106,52 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+            return Ok(None);
+        }
+
+        let mut columns = vec![];
+        let mut indices: HashSet<ColumnIndex> = indices.iter().copied().collect();
+
+        if let Some(all_cols) = self
             .data_availability_checker
-            .get_data_column(&DataColumnIdentifier { block_root, index })?
+            .get_data_columns(block_root)?
```
jimmygchen: I think we can simplify this if we swap the order around - early attester cache and database should always have all the columns we store (they only get added there if they pass DA check).

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Sorry, I was confused without looking into the details, and i might have caused the same confusion for you.

Let's go through what each of these caches store:
1. **`data_availability_checker`**: contains either partial or complete columns. Columns are stored here:
    - While waiting to pass DA check
    - For some time after passing DA check (before pruning)
2. **`early_attester_cache`**: Contains all custody columns for the block being attested. (they have passed DA check, but may not yet reach the database)
3. **`store`**: The persistent storage for all custody columns.

So with the above, it looks like we could keep the logic simple and never had to merge columns retreived from these components.

beacon_node/beacon_chain/src/beacon_chain.rs
SunnysidedJ: I see. Then, would the optimal order be:
1) Early Attester Cache
2) DB cache
3) DAC
4) DB disk?
I think (3) and (4) can be switched based on the preference: minimizing chance to hit DB vs code simplicity

jimmygchen: Yeah I'm working on this, will push my changes soon before I go offline.

beacon_node/store/src/hot_cold_store.rs
jimmygchen: My bad, please ignore.

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Addressed in https://github.com/sigp/lighthouse/pull/7399/commits/ee7109c6c51e7b356aa9643b0281be37ba815b7f

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: I went with the original order (DAC -> early attester -> DB cache -> DB disk), because:
1. There's a high chance that recent columns including both partial columns and fully avaialble columns would be in the DA checker, if the blocks are recent enough - which is the most frequent case when calling ByRoot 
2. Early attester cache: there's only a small window that the block would be here
3. DB cache: may contain partial results, depending what was loaded prior, hence 1 and 2 are preferrable over this
4. DB disk: only load the missing columns from disk, and put the loaded columns into the DB cache.

SunnysidedJ: That makes sense. Thank you so much for cleaning up

pawanjay176 (approved):
LGTM. Should get this merged soon to be devnet 7 compliant.
Can fix the ef_tests in a later PR.

CI is currently failing because of a kurtosis failure, working on fixing that.

jimmygchen:
I've pushed a data column caching implementation - I tried to simplify it but still end up being a bit complicated 😓 
I think this should work though, will do some testing now.

jimmygchen:
I think we can potentially just get rid of the `store.get_data_columns` implementation to keep caching simple.
The only place we need all columns is here and in the tests. For all range request and root requests, we require sender to supply indices.
https://github.com/sigp/lighthouse/blob/669932aa671c69013f6133cdda7cb6c19b0832ae/beacon_node/http_api/src/block_id.rs#L342

To simplify things we could just extract this into `chain.get_data_columns`, and use just `store.get_data_column` like above, and get rid of `store.get_data_columns`.

pawanjay176:
Fixed the early stream termination issue in 5e49d8e . Local lighthouse only devnet works fine now.

The beacon chain test is failing because the code here assumes that the data_columns vec would always be non-empty 
https://github.com/sigp/lighthouse/blob/92391cdac665256f60fab151d371cfe6a462cce0/beacon_node/beacon_chain/src/beacon_chain.rs#L3067-L3071

I think the above function makes the right assumption but the changes in `BeaconChain::get_data_columns` in this PR make it such that its possible to have empty columns in tests now. I think we should keep the older function signature of Result<Option<DataColumnSidecarList<T::EthSpec>>, Error> to indicate the absence of columns instead of ignoring it and returning an empty list.

SunnysidedJ:
Thank you for debugging @pawanjay176  and @jimmygchen. I've made a small change to get_data_columns to return like before `Result<Option<...>>` and the failed test passed locally.

jimmygchen (approved):
Looks good to me 👍  
Thanks!

PR #7401, opened 2025-05-06
Update all deps incl Tokio

michaelsproul:
## Proposed Changes

Attempt to fix the mysterious VC hangs by updating all deps, but especially `tokio`.

The suspected fix in Tokio is:

- https://github.com/tokio-rs/tokio/pull/7216


michaelsproul:
Dropping this from v7.0.2 as it doesn't fix the freezes. Will reopen a similar PR against unstable, but it's not urgent.

PR #7404, opened 2025-05-06
Siren installation improvement

AgeManning:
I think its cleaner and easier to use the docker-compose setup we have in Siren. 

I also think it's not great to recommend using `docker run --rm` as the container will be nuked when it closes. 

I've added installation instructions for docker compose and modified the docker run command. 

This should work once https://github.com/sigp/siren/pull/371 is merged to stable.


book/src/ui_installation.md
```diff
@@ -29,9 +56,9 @@ We recommend running Siren's container next to your beacon node (on the same ser
     cd Siren
     ```
 
- 1. Create a configuration file in the `Siren` directory: `nano .env` and insert the following fields to the `.env` file. The field values are given here as an example, modify the fields as necessary. For example, the `API_TOKEN` can be obtained from [`Validator Client Authorization Header`](./api_vc_auth_header.md)
+ 1. Create a configuration file in the `Siren` directory: `nano .env` and insert the following fields to the `.env` file. The field values are given here as an example, modify the fields as necessary. For example, the `API_TOKEN` can be obtained from
```
chong-he: The end part of this sentence seems to be accidentally deleted. 

```suggestion
 1. Create a configuration file in the `Siren` directory: `nano .env` and insert the following fields to the `.env` file. The field values are given here as an example, modify the fields as necessary. For example, the `API_TOKEN` can be obtained from [`Validator Client Authorization Header`](./api_vc_auth_header.md).
```

book/src/ui_installation.md
```diff
@@ -43,32 +70,35 @@ We recommend running Siren's container next to your beacon node (on the same ser
...
 
-    When Siren has successfully started, you should see the log `LOG [NestApplication] Nest application successfully started +118ms`, indicating that Siren has started.
+means that the validator client is not running, or the `--http` flag is not provided, or otherwise inaccessible from within the container. Another common error is:
 
- 1. Siren is now accessible at `https://<the-servers-ip>` (when used with `--net=host`). You will get a warning about an invalid certificate, this can be safely ignored.
+```
+validator api issue, server response: 403
+```
+
+    which means that the API token is incorrect. Check that you have provided the correct token in the field `API_TOKEN` in `.env`.
```
chong-he: This part is rendered as a code block, perhaps due to the indentation 

```suggestion
which means that the API token is incorrect. Check that you have provided the correct token in the field `API_TOKEN` in `.env`.
```

book/src/ui_installation.md
```diff
@@ -8,19 +8,46 @@ To ensure proper functionality, the Siren app requires Lighthouse v4.3.0 or high
...
+    specified as well as the validator clients `API_TOKEN`, which can be obtained from the [`Validator Client Authorization Header`](./api_vc_auth_header.md).
+
+ 1. Run the containers with docker compose
+    ```
+    docker compose up -d
+    ```
+
+ 1. You should now be able to access siren at the url (provided SSL is enabled):
+    ```
+    https://localhost
```
chong-he, line 52: How does this work for Siren running in remote server? Say the BN, VC and Siren all started in a server, and we want to browse Siren on the client computer. There needs to be some kind of connection between them I assume? 

In the current docs, with `--net host`, Siren can be accessible from a browser by providing the server IP address.

book/src/ui_installation.md
```diff
@@ -43,32 +70,35 @@ We recommend running Siren's container next to your beacon node (on the same ser
  1. You can now start Siren with:
 
     ```bash
-    docker run --rm -ti --name siren --env-file $PWD/.env --net host sigp/siren
+    docker run -ti --name siren --env-file $PWD/.env -p 443:443 sigp/siren
```
chong-he, line 82: I am not sure if this will still work, without the `--net host`. @antondlr and I have discussed about this before. Because there are users running BN and VC not using docker (e.g., systemd), so `--net host` is somehow required

chong-he (commented):
Will need some inputs from @antondlr too

book/src/ui_installation.md
AgeManning: The `--net host` is a footgun imo. @antondlr can chime in. It exposes unnecessary ports. 

The setup I've listed here does the exact same thing. It will expose port 443. If you are running this on the server you can provide the server IP address also. Here it is assumed its running on localhost. 

We can add a comment to say IP of host also to make it explicit.

book/src/ui_installation.md
AgeManning: oh, I guess maybe its the container trying to access local instances. I use the docker compose for this and it works fine. 

I guess @antondlr will have to chime in for --net host. My understanding is that it's not needed, but I might be missing something.

antondlr: yes, age is 100% right. `--net=host` is dangerous and should not be used lightly. 
the problem is, a LOT of our users have setup their nodes with the [somer esat guides](https://github.com/SomerEsat/ethereum-staking-guides), which runs CL/EL though systemd and limits the api ports to 127.0.0.1 ... which is not accessible through docker. 

a solution for that is to use `--add-host=host.docker.internal:host-gateway` on linux hosts, buuuutttt that doesn't work when Docker was installed as a Snap (which used to be default on Ubuntu but I think it no longer is?) 

I'm not sure how `--net=host` ever became the default, almost recommended way to run Siren... it should be a last resort fix, and only for people who understand the implications.

chong-he: For those running BN and VC without docker (systemd), without `--net=host`, will it work? I get how `--net host` is no good, is there another way to get around?

book/src/ui_installation.md
chong-he: Got it. In that case, I think it will be good to add the comment or in the content of the book to say IP is also ok, because I assume most users will probably run it remotely

book/src/ui_installation.md
AgeManning: I run BN and VC without docker using systemd and what I've written here works perfectly fine. 

I think the only case is if they set their HTTP IP in the CLI of BN and VC to 127.0.0.1 then it won't work.

AgeManning: I've added a comment about handling this edge case. I dont think we need `--net host`. People just need to expose their HTTP API I guess, or add the flag that anton said. 

I've commented on this in the docs now.

chong-he: @antondlr could you chime in on this? We discussed before that without `--net host`, those running systemd will not be able to connect to Siren. Changing the `BEACON_URL` to `http://docker.host.internal:5052` or adding `--add-host=host.docker.internal:host-gateway sigp/siren` also doesn't work

antondlr: we discussed this in detail at devcon, and it is indeed as age says: 
either expose the api on `0.0.0.0` (or, their specific IP, or anything else more broad than `127.0.0.1` really), _or_ run Siren with `--net=host`. 

> Changing the BEACON_URL to http://docker.host.internal:5052 or adding --add-host=host.docker.internal:host-gateway sigp/siren also doesn't work  

you'd have to do both, simultaneously, and this only works for regular docker installs iirc, not for "snap" installs (like you have on your server?). (and it's `host.docker.internal` not `docker.host.internal` but that's just a typo I assume?)

I can help debug / find a way for a specific setup like yours but there are no magic solutions here that preserve max security. the tradeoff is between exposing the BN/VC API more broadly, or running Siren with `net=host`, that will always be the crux of it for setups where Lighthouse doesn't run in Docker. 
which of those is "best" is a personal choice; we landed on (ab)using `net=host` to unblock people because that way, no change to people's running Lighthouse instance would be necessary, but it's definitely not without downsides.  
lmk if you want to go over it again why and how this all happens; I can explain in a call but typing it all out would be a course in itself :-)

chong-he: Yes, I remember, the "lesser of the two evils" - either exposing the API or using `--net host`

So long as it works, then it's good.

AgeManning: For the record, this doesn't have anything do with Systemd. I run lighthouse with systemd and without. I don't use the `net=host` and everything works fine. It is only for users who are not exposing their API, i.e have not set --http-address to <IP>. 

I expose my HTTP API so I can access it and query things, its only accessible on my internal LAN.

AgeManning:
I'm going to merge this, because I think its an improvement and users who are starting to use Siren might see an immediate benefit. 

Happy to do more PRs to resolve the `net=host` thing. Maybe we can add it as a note for users who really want to not expose their HTTP APIs.

AgeManning:
@mergifyio queue

michaelsproul (approved):

AgeManning:
@mergifyio refresh

PR #7409, opened 2025-05-07
Don't publish data columns reconstructed from RPC columns to the gossip network

jimmygchen:
## Issue Addressed

Don't publish data columns reconstructed from RPC columns to the gossip network, as this may result in peer downscoring if we're sending columns from past slots.


pawanjay176 (approved):
NIce catch!

Issue #7410, opened 2025-05-07
VC logs `Uninitialized` BN briefly on startup


michaelsproul:
Some VCs log this warning on startup

> May 07 04:37:53.491 WARN A connected beacon node errored during routine health check, endpoint: http://localhost:5052/, error: Uninitialized, module: beacon_node_fallback:503

It then resolves itself almost immediately.

> May 07 04:37:53.491 INFO Initialized beacon node connections, available: 1, total: 1, module: validator_client:673

jimmygchen:
Fixed in #7411

PR #7411, opened 2025-05-07
Silence `Uninitialized` warn log on start-up

macladson:
## Issue Addressed

#7410 

## Proposed Changes

Silences the `Uninitialized` warn log during routine beacon node health check. 

## Additional Info

`CandidateError::Uninitialized` will only ever occur during start-up during a race between the first beacon node update and the initialization of the slot clock, so we can safely silence it.

### DIFF
--- a/validator_client/beacon_node_fallback/src/lib.rs
+++ b/validator_client/beacon_node_fallback/src/lib.rs
@@ -486,11 +486,19 @@ impl<T: SlotClock, E: EthSpec> BeaconNodeFallback<T, E> {
 
         for (result, node) in results {
             if let Err(e) = result {
-                // We don't want to spam warns before Genesis and a node is only marked as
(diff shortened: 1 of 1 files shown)


michaelsproul (commented):
Is there an issue where the node could become uninitialised but then we don't see the error because of this silencing?

michaelsproul:
That would require a pretty major bug I guess, but maybe we should keep a debug log for this case?

macladson:
> Is there an issue where the node could become uninitialised but then we don't see the error because of this silencing?

Looking at the code paths, I don't see how this could ever happen. The slot clock would basically have to uninitialize itself.
Still, a `debug` log can't hurt.

michaelsproul (approved):

michaelsproul:
@mergify dequeue

michaelsproul:
@mergify requeue

Issue #6970, opened 2025-02-11
Optimise `SingleAttestation` implementation further


michaelsproul:
For v7.0.0 we took the a straight-forward approach for `SingleAttestation`s, translating them into `Attestation`s and keeping the core of Lighthouse's processing the same.

It would be more efficient to reverse the direction of this translation so that we can benefit from the structure of `SingleAttestation`s during verification, i.e. so we can check signatures prior to hitting committee caches.

We've begun on this, and plan to release it shortly after Electra goes live. More info and ideas contained in the WIP implementations:

- https://github.com/sigp/lighthouse/pull/6923 (newer)
- https://github.com/sigp/lighthouse/pull/6616 (older)

PR #7415, opened 2025-05-08
Increase default work queue size

michaelsproul:
## Issue Addressed

Fix for error message seen on some nodes which are struggling slightly after Pectra:

> ERROR Unable to queue converted SingleAttestation   error: no available capacity, slot: 4230400

Because we convert `SingleAttestation`s to regular attestations in a separate work event, we are now saturating the work event queue more often. This will remain an issue until a more optimised implementation of `SingleAttestation` processing is merged (likely for v7.1.0). See:

- https://github.com/sigp/lighthouse/issues/6970

## Proposed Changes

Double the size of the generic work event queue. This isn't guaranteed to fix the issue, as nodes could be struggling to an arbitrary degree. However it should make up for the approx doubling of work events introduced by Pectra so that only nodes logging errors prior to the fork are logging errors now.

## Additional Info

I'm rolling this out to some of our Holesky nodes that exhibit this behaviour and will report back on its efficacy.

### DIFF
--- a/beacon_node/beacon_processor/src/lib.rs
+++ b/beacon_node/beacon_processor/src/lib.rs
@@ -77,7 +77,7 @@ pub mod work_reprocessing_queue;
 /// The maximum size of the channel for work events to the `BeaconProcessor`.
 ///
 /// Setting this too low will cause consensus messages to be dropped.
-const DEFAULT_MAX_WORK_EVENT_QUEUE_LEN: usize = 16_384;
+const DEFAULT_MAX_WORK_EVENT_QUEUE_LEN: usize = 32_768;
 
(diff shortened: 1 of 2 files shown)


michaelsproul:
I forgot but we also still have the hidden flag for this, so users could crank up the queue size without a new release using this flag:

```
--beacon-processor-work-queue-len 32768
```

eserilev (approved):
Thanks for opening this PR. Just one smol test failure, but other than that LGTM!

michaelsproul:
This change doesn't seem to have been effective in preventing the queue from filling. We might need a higher value.

michaelsproul:
I'm inclined to close this in favour of the "proper" fix in v7.1.0:

- https://github.com/sigp/lighthouse/pull/7444

Issue #7414, opened 2025-05-07
Make kurtosis runs faster when lighthouse is not first participant


barnabasbusa:
We are using `/eth/v1/node/health` endpoint to check if a node is available. 

When we run two lighthouse nodes:
```yaml
participants:
  - cl_type: lighthouse
  - cl_type: lighthouse
```
There are no issues. 

When lighthouse is not the first participant like the example below:
quick way to replicate:
```yaml
participants:
  - cl_type: teku
  - cl_type: lighthouse
```
I'm getting:
```sh
curl -v http://127.0.0.1:33008/eth/v1/node/health
* Uses proxy env variable no_proxy == '*'
*   Trying 127.0.0.1:33008...
* Connected to 127.0.0.1 (127.0.0.1) port 33008
> GET /eth/v1/node/health HTTP/1.1
> Host: 127.0.0.1:33008
> User-Agent: curl/8.7.1
> Accept: */*
>
* Request completely sent off
< HTTP/1.1 503 Service Unavailable
< content-type: application/json
< server: Lighthouse/v7.0.1-e42406d/aarch64-linux
< content-length: 113
< date: Wed, 07 May 2025 15:57:07 GMT
<
* Connection #0 to host 127.0.0.1 left intact
{"code":503,"message":"SERVICE_UNAVAILABLE: beacon node is syncing: execution layer is offline","stacktraces":[]}
```

This bug seems to happen regardless the EL used or the but dependent on which CL is used.
Issues when the first participant is:
**teku** , **grandine,**

**No** issues when first participant is:
**prysm** , **lodestar**, **nimbus**

LH also reports:
```sh
May 07 16:03:16.811 ERRO Error updating deposit contract cache   error: Failed to update deposit cache: GetDepositLogsFailed("eth_getLogs call failed ServerMessage { code: -32000, message: \"invalid block range params\" }"), retry_millis: 60000, service: deposit_contract_rpc
...
May 07 15:58:41.599 WARN Error processing HTTP API request       method: GET, path: /eth/v1/node/health, status: 503 Service Unavailable, elapsed: 1.255048ms
```

This bug slows down the startup process of each enclave by ~30s

michaelsproul:
That's weird. We'll try to repro and fix.

We are planning to remove the deposit loading from the EL, that would remove that error and may coincidentally fix it

pawanjay176:
[...]

PR #7417, opened 2025-05-08
Change default EngineState to online

pawanjay176:
## Issue Addressed

Resolves https://github.com/sigp/lighthouse/issues/7414

## Proposed Changes

The health endpoint returns a 503 if the engine state is offline. The default state for the engine is `Offline`. So until the first request to the EL is made and the state is updated, the health endpoint will keep returning 503s. 

This PR changes the default state to Online to avoid that. I don't think this causes any issues because in case the EL is actually offline, the first fcu will set the state to offline. 

Pending testing on kurtosis.


pawanjay176:
@barnabasbusa can you test if this PR fixes your issue?

michaelsproul (approved):
Agree this seems like a safe change to make!

michaelsproul:
@pawanjay176 Looks like some tests need updating

barnabasbusa:
This PR fixed the bug for me. LGTM

Screenshot 2025-05-08 at 09 55 15

cl-2 with patched image
cl-3 with latest unstable image

barnabasbusa (approved):

PR #7418, opened 2025-05-08
Enable debug symbols by default in release builds

michaelsproul:
## Proposed Changes

Turn on debug symbols for release builds so that we get more informative backtraces.

This won't affect releases, which use maxperf. If users building from source want to remove debug symbols they can opt in to maxperf, although I suspect nobody will notice any detriment from this change.


michaelsproul:
Binary size increased to 1.9GB, so uh, yeah we probably shouldn't change the default release profile. Maybe a new profile called... `release-debug`?

michaelsproul:
Not viable right now. Closing.

Issue #6687, opened 2024-12-12
Update Remote Monitoring to Report Disk Usage for Datadir Mount Point


hqy2000:
## Description

In remote monitoring, [Lighthouse Advanced Metrics](https://lighthouse-book.sigmaprime.io/advanced_metrics.html?highlight=monitor#remote-monitoring) reports disk stats, such as `disk_node_bytes_free`, for the disk mounted at `/`. However, this may not correspond to the disk holding the actual beacon chain data.

## Version

Using the 6.0.0 executable from the release.

## Present Behavior

Currently, remote monitoring always reports the disk space for the root (`/`) mount point.  

Relevant code reference: https://github.com/sigp/lighthouse/blob/c042dc14d74352512b7632e0ee6ec07f1aa26b3a/common/eth2/src/lighthouse.rs#L183

## Expected Behavior

Remote monitoring should report the disk space for the `datadir` (or the default directory) instead of the root mount point. This is important as users may mount a separate drive specifically for beacon chain data.

## Steps to Resolve

1. Set up Lighthouse with remote monitoring enabled.
2. Configure `datadir` to point to a mount point on a second drive intended for storing beacon chain data.
3. Observe the reported disk stats to see if it's for the `datadir` mount point or `/`.

takaebato:
I'll work on this!

rose2221:
i'd like to work on this

heronimus:
Hi, I’m new here and still learning to understand the Lighthouse project structure. I’m interested in this issue, but after examining it in detail, I want to confirm a few things:

- The relevant code reference suggests changing the `SystemHealth` --> `disk_usage` default path to use the `datadir` value instead of the root path `("/")` for remote monitoring.
- However, the`impl SystemHealth` is used not only by remote monitoring but also by the `http_api` (`GET lighthouse/health`) and `http_metrics` (Prometheus /metrics) in both the `beacon_node` and `validator_client` modules.
- There are other data-related metrics that already use `datadir` (e.g., `store_disk_db_size` and `store_freezer_db_size`), but they only report disk DB size and some other data metrics related to `LevelDB`.

Changing this behavior will alter the results of all `disk_node_***` metrics for remote monitoring, `http_api`, and `http_metrics`. Do you think this approach is acceptable and expected?

---
I noticed that @michaelsproul  reviewed a previous related PR, your insights would be greatly appreciated. Thanks!

heronimus:
Another consideration:
[...]

PR #7421, opened 2025-05-08
Improve Disk Monitoring Logic

hqy2000:
## Issue Addressed

[https://github.com/sigp/lighthouse/issues/6687](https://github.com/sigp/lighthouse/issues/6687)

## Proposed Changes

This PR updates which disk is monitored by the system:

* **Beacon Node**: Now uses the `datadir` from the client's config (which provides a fallback value if not specified in the command line).
* **Validator Client**: Now uses the `datadir` from the command line. If it's not specified, `validation_dir` is used as a fallback.

Additionally, this PR updates the `Observe` trait by introducing a new method, `observe_with_path`, which allows specifying the path to observe. The existing `observe` method, which monitors the root folder, remains unchanged but is now only used by `ProcessHealth`, as it does not require disk information.

## Additional Info

While this PR addresses disk usage reporting in many cases, it may be necessary to update the API endpoint to support reporting multiple disks. This would be useful for advanced users who split the database across several disks, as discussed in https://github.com/sigp/lighthouse/issues/6687#issuecomment-2682441104.


validator_client/http_api/src/test_utils.rs
```diff
@@ -167,6 +167,7 @@ impl ApiTester {
     pub fn default_http_config() -> HttpConfig {
         HttpConfig {
             enabled: true,
+            monitoring_dir: tempdir().unwrap().path().to_path_buf(),
```
michaelsproul, line 182: I don't think this is safe, `tempdir()` will delete its contents on drop.

Issue #7403, opened 2025-05-06
Validator Client intermittently freezes on Linux kernel `6.14.4` -> `6.14.7`


michaelsproul:
## Summary

The Validator Client on Linux with kernel versions `6.14.4` -> `6.14.7` will intermittently freeze, stopping it from performing its duties.

If you're running a distro which closely follows Linux mainline (such as Arch Linux or Fedora) you may be affected.

Run `uname -r` to check if your kernel version is in the affected range.

**Note:** Ubuntu 24.04 and older use older kernels and thus are not affected. However, Ubuntu 25.04 is currently running the  6.14.0 kernel which is unaffected but it is possible that a future package upgrade will include one of the affected kernels so upgrade with caution.

## Solutions and Workarounds
This bug, caused by changes to the `eventpoll` code, has already been patched in the Linux mainline kernel and will be fixed in `6.14.8+`.

Once your distro allows you to update the kernel to `6.14.8` you can safely do so.

In the meantime, if you are running an affected kernel version you have a few options:

#### Install an LTS kernel

The procedure for this will differ depending on your distro but the below example is the instructions for Arch Linux:

```bash
sudo pacman -S linux-lts
```

If you are using `systemd-boot`, it should automatically generate the corresponding bootloader entries.

If you are using `grub` you will need to regenerate them:

```bash
sudo grub-mkconfig -o /boot/grub/grub.cfg
```

Reboot and you should see `linux-lts` included in the `grub` menu.

#### Downgrade your kernel

This will vary depending on your distro and for some distros it is very involved. Here are the instructions for Fedora 41:

```bash
# Find available kernels
sudo dnf list kernel --showduplicates

# Install a specific kernel. For example:
sudo dnf install kernel-6.11.4-301.fc41
```

`grub` entries will automatically be added, so reboot and select the new kernel from the list.

#### Run an API polling script

If you do not want to touch your kernel in case you break something there is a simple `bash` script you can run instead.

Due to the internals of `eventpoll`, when the VC receives an API call, it will wake from its freeze.

Because of this we can use a script running in the background which continuously polls the VC. Here is an example of such a script:

```bash
while sleep 5; do curl -s --fail "http://localhost:5062/lighthouse/auth" > /dev/null && echo "polled at $(date)"; done
```

This will keep the VC awake. 
Note that running a full VC metrics server with Grafana polling the VC will also keep it awake for the same reason.

## Acknowledgments

[...]

PR #7423, opened 2025-05-08
Use async RwLock for the VC's `initialized_validators`

michaelsproul:
## Issue Addressed

Remove a `parking_lot::RwLockReadGuard` which was being held across an await in the VC.

Possibly addressing:

- https://github.com/sigp/lighthouse/issues/7403

## Proposed Changes

Use a `tokio::sync::RwLock` which requires `await` to be unlocked.

Unfortunately this has a huge ripple effect, with many functions needing to be async :roll_eyes: 

I've been grinding through this refactor for hours and it isn't quite finished yet. I'm offline until May 13. If anyone wants to pick this up before then just comment below.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9790,6 +9790,7 @@ dependencies = [
  "parking_lot 0.12.3",
  "serde",
  "slot_clock",
+ "tokio",
  "tracing",
  "types",
  "validator_metrics",
--- a/common/warp_utils/src/task.rs
+++ b/common/warp_utils/src/task.rs
@@ -1,5 +1,6 @@
 use crate::reject::convert_rejection;
 use serde::Serialize;
+use std::future::Future;
 use warp::reply::{Reply, Response};
 
 /// A convenience wrapper around `blocking_task`.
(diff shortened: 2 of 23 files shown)


eserilev:
Was poking around a bit here trying to get things to compile and found a subtle tracing issue 

[subte tracing thing](https://github.com/sigp/lighthouse/pull/7423/commits/47aca45013b787de4a2627f441ee8183d8cead92)

``` Rust

info!(
    voting_validators = validator_store.num_voting_validators().await,
    "Loaded validator keypair store"
);
        
```

Calling a future inside a tracing macro isn't `Send`

michaelsproul:
Oh nice find! Good to know!

jimmygchen (commented):
🤖 This looks like a solid code quality improvement. Removing all the `handle.block_on()` calls and `TaskExecutor` handle checks makes the code much more idiomatic for async Rust. CI is passing and the refactor looks correct.

Are you planning to merge this? The original bug reports seem to have stopped, but the async refactor itself seems worthwhile. If you do want to merge, we should deploy to a testnet node soon after to validate under real conditions given the scope of the changes.

PR #7426, opened 2025-05-08
Pin kurtosis version to fix local testnet errors

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Pin kurtosis version to 4.6.0 to avoid the bug caused by https://github.com/ethpandaops/ethereum-package/pull/1007


pawanjay176:
Fixed upstream, so not required anymore.

PR #7427, opened 2025-05-08
Move notifier and latency service to `validator_services`

dknopik:
## Proposed Changes

We would like to reuse the `notifier` and `latency_service` in Anchor. To make this possible, this PR moves these from `validator_client` to `validator_services` and makes them use the new `ValidatorStore` trait is used so that the code can be reused in Anchor.


AgeManning (approved):

michaelsproul (approved):
LGTM!

PR #7436, opened 2025-05-10
chore: account for state availability when fetching sync committees #7435

### LINKED ISSUE #7115: Account for state availability when fetching sync committees

Ayushdubey86:
closes #7115 

Completely forgot about this, do check this @michaelsproul 

ref #7178


michaelsproul:
Needs a test

PR #7437, opened 2025-05-11
Remove reprocess channel

eserilev:
## Issue Addressed
Partially https://github.com/sigp/lighthouse/issues/6291

## Proposed Changes

This PR removes the reprocess event channel from being externally exposed. All work events are now sent through the single `BeaconProcessorSend` channel. I've introduced a new `Work::Reprocess` enum variant which we then use to schedule jobs for reprocess. I've also created a new scheduler module which will eventually house the different scheduler impls.

This is all needed as an initial step to generalize the beacon processor

A "full" implementation for the generalized beacon processor can be found here
https://github.com/sigp/lighthouse/pull/6448

I'm going to try to break up the full implementation into smaller PR's so it can actually be reviewed


michaelsproul:
Nice! Love this!

eserilev:
Since reprocess and regular work events will all come through a single channel, we may need to raise the default work queue size like we're doing here: https://github.com/sigp/lighthouse/pull/7415

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -1190,11 +1199,15 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
-                    ))) = send_result
+                        )),
+                    });
+                    if let Err(TrySendError::Full(WorkEvent {
+                        work:
+                            Work::Reprocess(ReprocessQueueMessage::DelayColumnReconstruction(
+                                reconstruction,
+                            )),
+                        ..
+                    })) = send_result
```
michaelsproul, line 1061: I did this in the most recent merge commit, would appreciate an extra set of eyes on it

eserilev: this looks okay to me!

michaelsproul (approved):
LGTM, and seems to be running well on our infra!

Issue #7441, opened 2025-05-12
next sync committee branch calculated incorrectly


benluelo:
lighthouse: https://ethereum-beacon-api.publicnode.com/eth/v1/beacon/light_client/updates?start_period=1420&count=6
lodestar: https://lodestar-mainnet.chainsafe.io/eth/v1/beacon/light_client/updates?start_period=1420&count=6

```diff
556c556
<         "0x504373599ac3b799767d44b05f3809f01b85057a24a4ea11b7f3c6a6a175837d",
---
>         "0x3c52adeac342b3ae75f8cccb0df305bd816c4ebe5dcc295edffa9824d4f3b60d",
609,610c609
<     },
<     "version": "deneb"
---
>     }
1166c1165
<         "0x3c52adeac342b3ae75f8cccb0df305bd816c4ebe5dcc295edffa9824d4f3b60d",
---
>         "0x02e2764784ce09de80c2bb7c82c708447302c444cf36844abb85ee3cb6b0cb6e",
1219,1220c1218
<     },
<     "version": "deneb"
---
>     }
1776c1774
<         "0x02e2764784ce09de80c2bb7c82c708447302c444cf36844abb85ee3cb6b0cb6e",
---
>         "0x396a997227554ba1bed4b784c1da37ae76b9f2278dbe3d2b2714a878041dbb73",
1831,1832c1829
<     },
<     "version": "electra"
---
>     }
2388c2385
<         "0x396a997227554ba1bed4b784c1da37ae76b9f2278dbe3d2b2714a878041dbb73",
---
>         "0x5bc70cbdee70d51698c6df292058adf901b07b8e7a2e96375413bb552f25f51e",
2443,2444c2440
<     },
<     "version": "electra"
---
>     }
3000c2996
<         "0x5bc70cbdee70d51698c6df292058adf901b07b8e7a2e96375413bb552f25f51e",
---
>         "0x00a6a63e163d072f7bfebc2293da83d797be448f4177112ded1cc85d9153b8ce",
3055,3056c3051
<     },
<     "version": "electra"
---
>     }
3612c3607
<         "0x00a6a63e163d072f7bfebc2293da83d797be448f4177112ded1cc85d9153b8ce",
---
>         "0xf3c0e9972a9cdac7253db71729b45a2750c376b9a0d78eadefa160fc0a430221",
3667,3668c3662
<     },
<     "version": "electra"
---
>     }
```

[...]

PR #7443, opened 2025-05-12
Next sync committee branch bug

eserilev:
## Issue Addressed

#7441

## Proposed Changes

Make sure we're correctly caching light client data

## Additional Info

Previously we were setting `current_sync_committee_branch` to `next_sync_committee_branch` (and vice versa). I've removed the `let (_,_,_)` to make things clearer


michaelsproul:
How come we didn't trigger this while verifying light client updates from the network?

michaelsproul (approved):
(Fix looks good btw)

eserilev:
> How come we didn't trigger this while verifying light client updates from the network?

Only the `LightClientFinalityUpdate` and `LightClientOptimisticUpdate` objects get verified over the network

The `current_ sync_committee_branch` and `next_sync_committee_branch` are only included in `LightClientBootstrap` and `LightClientUpdate` objects

michaelsproul:
Oh right, makes sense. Thanks!

PR #7444, opened 2025-05-12
Single attestation "Full" implementation

eserilev:
## Issue Addressed

#6970

## Proposed Changes

This allows for us to receive `SingleAttestation` over gossip and process it without converting. There is still a conversion to `Attestation` as a final step in the attestation verification process, but by then the `SingleAttestation` is fully verified.

I've also fully removed the `submitPoolAttestationsV1` endpoint as its been deprecated

I've also pre-emptively deprecated supporting `Attestation` in `submitPoolAttestationsV2` endpoint. See here for more info: https://github.com/ethereum/beacon-APIs/pull/531

I tried to the minimize the diff here by only making the "required" changes. There are some unnecessary complexities with the way we manage the different attestation verification wrapper types. We could probably consolidate this to one wrapper type and refactor this even further. We could leave that to a separate PR if we feel like cleaning things up in the future.

Note that I've also updated the test harness to always submit `SingleAttestation` regardless of fork variant. I don't see a problem in that approach and it allows us to delete more code :)


beacon_node/beacon_chain/src/attestation_verification.rs
```diff
@@ -832,42 +846,47 @@ impl<'a, T: BeaconChainTypes> IndexedUnaggregatedAttestation<'a, T> {
...
-            attestation.data(),
+            &attestation.data,
             &chain.spec,
         )?;
 
-        // Check to ensure that the attestation is "unaggregated". I.e., it has exactly one
-        // aggregation bit set.
-        let num_aggregation_bits = attestation.num_set_aggregation_bits();
-        if num_aggregation_bits != 1 {
-            return Err(Error::NotExactlyOneAggregationBitSet(num_aggregation_bits));
```
eserilev, line 843: `SingleAttestation` objects are guaranteed to be unaggregated so this check is not needed

eserilev: https://github.com/ethereum/consensus-specs/pull/3900/files#diff-0c7f98ee00b98da09639d6e0550fd915999da5c090717076907cc6521ce03a2eR69-R71

eserilev: https://github.com/ethereum/consensus-specs/blob/23e30671489c06ed3cc1d3662c16a09ef2f71176/specs/electra/p2p-interface.md?plain=1#L115-L117

beacon_node/beacon_chain/src/attestation_verification.rs
```diff
@@ -832,42 +846,47 @@ impl<'a, T: BeaconChainTypes> IndexedUnaggregatedAttestation<'a, T> {
...
+            if attestation.data.index != 0 {
+                return Err(Error::CommitteeIndexNonZero(
+                    attestation.data.index as usize,
+                ));
+            }
         }
 
-        // [New in Electra:EIP7549]
-        verify_committee_index(attestation)?;
-
```
eserilev, line 848: `SingleAttestation` objects are guaranteed to be unaggregated so no need for this check

beacon_node/beacon_chain/src/attestation_verification.rs
```diff
@@ -928,23 +947,56 @@ impl<'a, T: BeaconChainTypes> IndexedUnaggregatedAttestation<'a, T> {
...
+        let attestation_epoch = attestation.data.slot.epoch(T::EthSpec::slots_per_epoch());
+
+        let fork_name = chain
+            .spec
+            .fork_name_at_slot::<T::EthSpec>(attestation.data.slot);
+
+        let indexed_attestation = attestation.to_indexed(fork_name);
+
+        // Check that the attester is a member of the committee and return `committees_per_slot` which 
+        // is used in `verify_middle_checks` to calculate the expected `subnet_id`
```
eserilev: https://github.com/ethereum/consensus-specs/blob/23e30671489c06ed3cc1d3662c16a09ef2f71176/specs/electra/p2p-interface.md?plain=1#L110-L111

beacon_node/beacon_chain/src/attestation_verification.rs
eserilev: This check was initially introduced in Electra, but is removed for `SingleAttestation`

https://github.com/ethereum/consensus-specs/pull/3900/files#diff-0c7f98ee00b98da09639d6e0550fd915999da5c090717076907cc6521ce03a2eL63

beacon_node/beacon_chain/src/attestation_verification.rs
```diff
@@ -977,10 +970,55 @@ impl<'a, T: BeaconChainTypes> IndexedUnaggregatedAttestation<'a, T> {
...
+        let (committee_opt, committees_per_slot) = chain.with_committee_cache(
+            attestation.data.target.root,
+            attestation.data.slot.epoch(T::EthSpec::slots_per_epoch()),
+            |committee_cache, _| {
+                let committee_opt = committee_cache
+                    .get_beacon_committee(attestation.data.slot, attestation.committee_index)
+                    .map(|beacon_committee| beacon_committee.committee.to_vec());
+
+                Ok((committee_opt, committee_cache.committees_per_slot()))
+            },
```
eserilev, line 978: In the spirit of the `SingleAttestation` changes, I've moved all checks that require access to the shuffling to after signature verification

eserilev: These checks are
- Committee exists for slot and index
- Attester is in the committee
- expected subnet id == provided subnet id

beacon_node/beacon_chain/src/attestation_verification.rs
```diff
@@ -977,10 +970,55 @@ impl<'a, T: BeaconChainTypes> IndexedUnaggregatedAttestation<'a, T> {
...
+                index: attestation.committee_index,
+            });
+        };
+
+        if committee.contains(&(attestation.attester_index as usize)) {
+            return Err(Error::AttesterNotInCommittee {
+                attester_index: attestation.attester_index,
+                committee_index: attestation.committee_index,
+                slot: attestation.data.slot,
+            });
```
eserilev: https://github.com/ethereum/consensus-specs/blob/23e30671489c06ed3cc1d3662c16a09ef2f71176/specs/electra/p2p-interface.md?plain=1#L110-L111

beacon_node/beacon_chain/tests/attestation_verification.rs
```diff
@@ -1010,121 +1020,23 @@ async fn unaggregated_gossip_verification() {
...
         )
-        /*
-         * The following two tests ensure:
-         *
-         * The attestation is unaggregated -- that is, it has exactly one participating validator
-         * (len([bit for bit in attestation.aggregation_bits if bit == 0b1]) == 1).
-         */
-        .inspect_unaggregate_err(
-            "attestation without any aggregation bits set",
-            |tester, mut a, _| {
```
eserilev, line 1029: not a valid test case for `SingleAttestation`

beacon_node/beacon_chain/tests/attestation_verification.rs
```diff
@@ -1010,121 +1020,23 @@ async fn unaggregated_gossip_verification() {
...
-                assert!(matches!(
-                    err,
-                    AttnError::NotExactlyOneAggregationBitSet(0)
-                ))
-            },
-        )
-        .inspect_unaggregate_err(
-            "attestation with two aggregation bits set",
-            |tester, mut a, _| {
-                match &mut a {
```
eserilev, line 1063: not a valid test case for `SingleAttestation`

beacon_node/beacon_chain/tests/attestation_verification.rs
```diff
@@ -1010,121 +1020,23 @@ async fn unaggregated_gossip_verification() {
...
-        /*
-         * The following test ensures:
-         *
-         * The number of aggregation bits matches the committee size -- i.e.
-         *   `len(attestation.aggregation_bits) == len(get_beacon_committee(state, data.slot,
-         *   data.index))`.
-         */
-        .inspect_unaggregate_err(
-            "attestation with invalid bitfield",
-            |_, mut a, _| {
```
eserilev, line 1092: not a valid test case for `SingleAttestation`

eserilev:
We now move checks that require access to the shuffling to AFTER signature verification. This seems in line with the spirit of the `SingleAttestation` changes

https://github.com/ethereum/consensus-specs/pull/3900

beacon_node/beacon_chain/src/attestation_verification.rs
michaelsproul: Nice! I've removed this error variant completely in:

- b413e3fd1fa846d583f4f0666cf5fb671fe1456a

beacon_node/beacon_chain/src/test_utils.rs
```diff
@@ -2433,17 +2444,49 @@ where
...
+
+                unaggregated.push((single_attestation, Some(*subnet)));
             }
 
             if let Some(a) = maybe_signed_aggregate {
                 aggregated.push(a)
             }
         }
 
+        //TODO(single-attestation) convert to single attn
```
michaelsproul: This seems to already be implemented in the loop above?

beacon_node/beacon_chain/tests/attestation_verification.rs
```diff
@@ -275,15 +280,16 @@ struct GossipTester {
     /*
      * Valid unaggregated attestation
      */
-    valid_attestation: Attestation<E>,
+    valid_attestation: SingleAttestation,
     attester_validator_index: usize,
+    #[allow(dead_code)]
     attester_committee_index: usize,
```
michaelsproul: Maybe lets just remove this field

michaelsproul (approved):
LGTM, keen to roll this out on some testnets and see how it does!

beacon_node/beacon_chain/tests/attestation_verification.rs
eserilev: did a bit of cleanup here [418a26f](https://github.com/sigp/lighthouse/pull/7444/commits/418a26f344e91a406a051de172f14fcc06f2419b)

Issue #7389, opened 2025-05-02
Add network metrics of received objects per client type


dapplion:
Add metric to track total blocks synced by client type, or agent to track versions. Will be helpful to track issues with specific clients. Extend to time on sync mode per client. Bytes per client, received messages per client, etc. To see if a specific client dominates our networking

iri031:
Hi @dapplion can i start with this?

dapplion:
> Hi [@dapplion](https://github.com/dapplion) can i start with this?

Yes, thank you! Let me know if you need help

iri031:
Hi @dapplion, I created an initial version in the linked PR based on my understanding. Please review and suggest any changes required!

PR #7445, opened 2025-05-12
Network metrics per client

iri031:
## Issue Addressed

#7389 
## Proposed Changes
Metrics for: 
* Blocks received and imported per client
* Sync time per client 
* Bytes and messages received per client


beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -214,6 +214,19 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
         reprocess_tx: Option<mpsc::Sender<ReprocessQueueMessage>>,
         seen_timestamp: Duration,
     ) {
+        let attestation_size = attestation.as_ssz_bytes().len() as u64;
```
dapplion: This calls force us to spend cycles re-serializing a lot of SSZ objects. To track this for free we should do the tracking in the lighthouse-network layer which I'm not sure has access to peer client information. If not possible to track this for free skip tracking bytes per client for now

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -214,6 +214,19 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+
+        metrics::inc_counter_vec_by(
+            &metrics::BYTES_RECEIVED_PER_CLIENT,
+            &[&client],
+            attestation_size,
+        );
+
+        metrics::inc_counter_vec(
+            &metrics::MESSAGES_RECEIVED_PER_CLIENT,
+            &[&client]
```
dapplion: You should label this metric by [client, object_type] where object type could be
- `gossip_attestation`
- `gossip_block`
- `range_sync_block`
- `backfill_sync_block`

etc

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -241,6 +254,21 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
         packages: GossipAttestationBatch<T::EthSpec>,
         reprocess_tx: Option<mpsc::Sender<ReprocessQueueMessage>>,
     ) {
+        for package in &packages {
```
dapplion: You can move attestation metrics to `process_gossip_attestation_result` and only track valid objects

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -901,6 +981,21 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+        metrics::inc_counter_vec_by(
+            &metrics::BYTES_RECEIVED_PER_CLIENT,
+            &[&client],
+            blob_size,
+        );
+
+        metrics::inc_counter_vec(
+            &metrics::MESSAGES_RECEIVED_PER_CLIENT,
+            &[&client]
+        );
```
dapplion: You could add a helper method

```rust
fn track_messages_received_per_client(&self, peer_id: PeerId, object_type: &'static str) {
    let client = self.network_globals.client(&peer_id).kind.to_string();
    metrics::inc_counter_vec(
            &metrics::MESSAGES_RECEIVED_PER_CLIENT,
            &[&client, object_type]
        );
}
```
to de-duplicate a lot of code

beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
```diff
@@ -169,6 +169,11 @@ impl<E: EthSpec> PeerDB<E> {
         }
     }
 
+    /// Returns the sync start time of the peer if it exists
+    pub fn sync_start_time(&self, peer_id: &PeerId) -> Option<&Instant> {
+        self.peers.get(peer_id).and_then(|info| info.sync_start_time())
```
dapplion, line 174: I don't follow the meaning/purpose of sync start time, could you detail it?

beacon_node/network/src/sync/network_context.rs
```diff
@@ -372,6 +372,16 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
     }
 
+    pub fn client_version(&self, peer_id: &PeerId) -> String {
+        self.network_globals()
+            .peers
+            .read()
+            .peer_info(peer_id)
+            .map(|info| info.client().version.clone())
+            .unwrap_or_default()
+    }
```
dapplion: You have a single metric, not one per client type and one for client version. In mainnet is `version` field set for most clients? Could you test this and report back? If so, to what?

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
iri031: Moved to lighthouse-network

dapplion:
FYI @iri031 there are still some unaddressed comment

beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
iri031: I added this to calculate sync time per client. When a sync request is received I update this parameter to record the time when the sync started [here](https://github.com/iri031/lighthouse/blob/metrics-by-client/beacon_node/network/src/sync/manager.rs#L417) and later when the state of the peer changes to is_synced [here](https://github.com/iri031/lighthouse/blob/metrics-by-client/beacon_node/network/src/sync/manager.rs#L593)  I calculate the time taken to include in the metric.

beacon_node/network/src/sync/network_context.rs
iri031: How can I test for the mainnet?

gitToki: - One potential way to do it would be to test it directly in the mainnet with temporary logging to you're lighthouse client> 

first add analysis to network_contexte:
```
// Add to SyncNetworkContext impl
pub fn analyze_peer_versions_sample(&self) -> (usize, usize, Vec<String>) {
    let peers: Vec<PeerId> = self.network_globals()
        .peers
        .read()
        .connected_peers()
        .take(50) // sample 50 but you can adjust that
        .cloned()
        .collect();
    
    let mut with_version = 0;
    let mut without_version = 0;
    let mut examples = Vec::new();
    
    for peer_id in peers {
        let version = self.client_version(&peer_id);
        let client_type = self.client_type(&peer_id);
        
        if version.is_empty() {
            without_version += 1;
        } else {
            with_version += 1;
            if examples.len() < 8 {
                examples.push(format!("{}: {}", client_type, version));
            }
        }
    }
    
    (with_version, without_version, examples)
    }
```
then call this function every X min in sync manager:

```
// Add this in main sync loop or network service
let mut last_version_check = Instant::now();

// Inside your main loop:
if last_version_check.elapsed() > Duration::from_secs(300) { // here I took 5 min
    let (with_ver, without_ver, examples) = 
        self.network_context.analyze_peer_versions_sample();
    
    let total = with_ver + without_ver;
    let percentage = if total > 0 { 
        (with_ver as f64 / total as f64) * 100.0 
    } else { 0.0 };
    
    info!("MAINNET VERSION TEST: {}/{} peers have versions ({:.1}%). Examples: {:?}", 
          with_ver, total, percentage, examples);
          
    last_version_check = Instant::now();
}
```
 and then build and enable the log to get the info

```
cargo build --release


./target/release/lighthouse bn \
    --network mainnet \
    --datadir /your/datadir \
    --http \
    --log-level info \
    2>&1 | tee mainnet_version_test.log
```
you should get something like: MAINNET VERSION TEST: 42/50 peers have versions (84.0%). Examples: ["xx"]

there is around 8200 CL nodes that you can communicate with in the mainnet, so adjusting the sample might be good

- Another way could be to use a crawler, use discv5 to discover nodes (starting from boot-nodes) and then libp2p to the discovered nodes to look at the version field (first option might be better, it depends).

gitToki: Hi again, I modified my crawler to get the data that we needed and made it runs for the past few hours. Here is what I got @dapplion 

image



image

got "Unknown" client names result from parsing failures. It makes the data inaccurate, my bad for that

to get better data you can look here -> https://monitoreth.io/nodes

iri031:
> This pull request has merge conflicts. Could you please resolve them @iri031? 🙏

On it.

PR #7321, opened 2025-04-12
feat: add tests for filtering GET /beacon/pool/attestations

gap-editor:
Added `test_get_beacon_pool_attestations_filtered` to verify that the `/eth/v2/beacon/pool/attestations` endpoint correctly filters attestations based on the `slot` and `committee_index` query parameters, as required by #7312.

PR #7451, opened 2025-05-13
Add additional mergify rules to automate triaging

jimmygchen:
## Issue Addressed

Added a few more mergify automation rules to help improve PR management:
- Add `waiting-on-author` and remove `ready-for-review` on PRs with conflicts or failing CI checks.
- Automatically remove `waiting-on-author` and add `ready-for-review` once conflicts / failing CI checks have been resolved. (Note: this will require a "request for change" review action when asking user to make changes, i.e. if reviewer only add a regular comment, the `waiting-on-author` label may be removed by mergify)
- Close stale pull request after 30 days of inactivity, if it's in `waiting-on-author` state.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -12,6 +12,58 @@ pull_request_rules:
       comment:
         message: This pull request has merge conflicts. Could you please resolve them
           @{{author}}? 🙏
+      label:
+        add:
+          - waiting-on-author
+        remove:
+          - ready-for-review
+
+  - name: Ask to resolve CI failures
+    conditions:
+      - or:
+          - check-failure=test-suite-success
+          - check-skipped=test-suite-success
+          - check-failure=local-testnet-success
+          - check-skipped=local-testnet-success
+    actions:
+      comment:
+        message: Some required checks have failed. Could you please take a look @{{author}}? 🙏
+      label:
+        add:
+          - waiting-on-author
+        remove:
+          - ready-for-review
+
+  - name: Update labels when PR is unblocked
+    conditions:
+      - label=waiting-on-author
+      - -conflict
+      - check-success=test-suite-success
+      - check-success=local-testnet-success
+      - "#changes-requested-reviews-by = 0"
+    actions:
+      label:
+        remove:
+          - waiting-on-author
+        add:
+          - ready-for-review
+      comment:
+        message: >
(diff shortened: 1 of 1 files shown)


jxs (approved):
LGTM Jimmy, thanks for this!

michaelsproul (approved):

.github/mergify.yml
```diff
@@ -12,6 +12,58 @@ pull_request_rules:
...
+        remove:
+          - ready-for-review
+
+  - name: Update labels when PR is unblocked
+    conditions:
+      - label=waiting-on-author
+      - -conflict
+      - check-success=test-suite-success
+      - check-success=local-testnet-success
+      - "#changes-requested-reviews-by = 0"
```
michaelsproul: I think this won't fire very often. If someone requests changes, their request for changes hangs around until they dismiss it or re-review.

We can give it a go, but I think what we really need is a way of checking whether all review comments have been addressed. Maybe `#review-threads-unresolved = 0`?

michaelsproul: (from https://docs.mergify.com/configuration/conditions/)

jimmygchen: Good catch.

I had a look at a few open PRs and looks like neither of these would be sufficient
- `#changes-requested-reviews-by = 0` has the issue you mentioned, e.g. #7321 has a change requested so it will never fire until a maintainer changes the review status
- `#review-threads-unresolved = 0` may cause this rule to fire unintentionally, e.g. #7436 just had a change requested, but CI is passing and there's no conflict, therefore mergify will immediately remove the `waiting-on-author` label after it's added.

jimmygchen: For PRs by external contributors, this rule will never fire automatically because the CI workflow require approval - but it would be useful to automatically update the labels once the a maintainer approve the job and the CI job passes.

For PRs by maintainers, I think it would be easy enough for the author just update the labels.

jimmygchen: I've updated the rules to match for 
- no failing CI jobs instead of passing CI (because it requires approval)
- `"#review-requested > 0"` instead of the above, so it only fires automatically if a review is requested. This is a workaround for the above two issues, it's not ideal but i think easier than 1. dimissing the review, 2. having to create a thread to avoid `waiting-on-author` from getting removed.

michaelsproul (approved):

PR #7452, opened 2025-05-13
Add parking_lot deadlock detection

macladson:
Adds `parking_lot` deadlock detection for debugging the freezes occuring on the VC: https://github.com/sigp/lighthouse/issues/7403

It is added outside of the tokio executor so hopefully will continue running even if the executor locks up

Based on https://github.com/sigp/lighthouse/pull/7423 to limit any unrelated deadlock scenarios.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3431,6 +3431,12 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


macladson:
This didn't reveal anything so I'll close this now

Issue #7442, opened 2025-05-12
Missing implementation of `postStateValidatorIdentities` in Beacon-APIs


Alleysira:
Please check the [issue](https://github.com/Consensys/teku/issues/9417) in teku for details.

Thanks for your attention.

chong-he:
I would like to give this a go

chong-he:
Closed by #7462 , the endpoint is available in the `unstable` branch now, and will be available since Lighthouse v7.1.0

PR #7462, opened 2025-05-15
Implement `validator_identities` Beacon API endpoint

chong-he:
## Issue Addressed

* #7442 

## Additioanal information

Tested and it's working. Command: 
`curl -X 'POST' 'http://localhost:5052/eth/v1/beacon/states/head/validator_identities' -d '["1137772","1137773"]' | jq`

Before:
```
{
"code": 400
message": "BAD_REQUEST: Unsupported endpoint version: v1
"stacktraces": []
}
```

After:
```
{
"execution_optimistic": false
"finalized": false
"data": [
    {
index": "1137772
pubkey": "0x86a0a9460e34142b06346f6cd7eb574a6c110c431bb98bc3df4ba8609cb2c53c7b663e646eaf5671941a03a02049cea3
activation_epoch": "11784
    }
    {
index": "1137773
pubkey": "0xb47b2be67cadbab7f5be998599abc7f359a26df6fd5da060ca26e12753865d57256ec92f9c84ecf19a1d2aa80c5f9187
activation_epoch": "11785
    }
  ]
}
```

If a validator index is not known, it returns empty data, which matches the description in the beacon API: 
[no identity will be returned but this will not cause an error](https://ethereum.github.io/beacon-APIs/#/Beacon/postStateValidatorIdentities) :
```
{
  "finalized": false,
  "data": []
}
```


beacon_node/http_api/src/validators.rs
```diff
@@ -114,3 +114,47 @@ pub fn get_beacon_state_validator_balances<T: BeaconChainTypes>(
...
+
+                Ok((
+                    // From the BeaconState, extract the Validator data and convert it into ValidatorIdentityData type
+                    state
+                        .validators()
+                        .iter()
+                        .enumerate()
+                        // filter by validator id(s) if provided
+                        .filter(|(index, validator)| {
+                            ids_filter_set.as_ref().is_none_or(|ids_set| {
```
michaelsproul, line 147: This will return every validator if the request is `[]`, but it's not clear if that is desired by the spec:

- https://github.com/ethereum/beacon-APIs/issues/532

If we don't get an outcome on that issue soon we can merge to unstable and fix it up later if required.

michaelsproul: We have support from Nico, so we can proceed with implementing the `[] = all` behaviour.

Upstream PR will likely be merged soon:

- https://github.com/ethereum/beacon-APIs/pull/533

michaelsproul: @chong-he Maybe you can test this behaviour for this endpoint while fixing it for `balances` as well

chong-he: Will do. Currently, this endpoint when provided with `[]`, also returns no data. I will update this after fixing `validator_balances` endpoint

beacon_node/http_api/src/lib.rs
```diff
@@ -725,6 +725,27 @@ pub fn serve<T: BeaconChainTypes>(
...
+        .clone()
+        .and(warp::path("validator_identities"))
+        .and(warp::path::end())
+        .and(warp_utils::json::json())
+        .then(
+            |state_id: StateId,
+             task_spawner: TaskSpawner<T::EthSpec>,
+             chain: Arc<BeaconChain<T>>,
+             query: ValidatorIdentitiesRequestBody| {
+                task_spawner.blocking_json_task(Priority::P1, move || {
```
michaelsproul: We could consider making this P0 if we expect it to be used by validator clients. Or we could make it `P0` if the `state_id` is `Head` (because those requests should be very fast to process).

beacon_node/http_api/src/lib.rs
```diff
@@ -725,6 +725,27 @@ pub fn serve<T: BeaconChainTypes>(
             },
         );
 
+    // POST beacon/states/{state_id}/validator_identities
+    let post_beacon_state_validator_identities = beacon_states_path
+        .clone()
+        .and(warp::path("validator_identities"))
+        .and(warp::path::end())
+        .and(warp_utils::json::json())
```
michaelsproul: We should use your new `json_no_body` extractor now that it's in `unstable`

common/eth2/src/types.rs
```diff
@@ -694,6 +702,12 @@ pub struct ValidatorBalancesRequestBody {
     pub ids: Vec<ValidatorId>,
 }
 
+#[derive(Clone, Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct ValidatorIdentitiesRequestBody {
+    pub ids: Vec<ValidatorId>,
```
michaelsproul, line 708: Probably need the same `serde(default)` that we use for balances

chong-he: Resolved in [bd37444](https://github.com/sigp/lighthouse/pull/7462/commits/bd37444df8da073ec5e351c0cff48b9c80aa531c)

beacon_node/http_api/src/lib.rs
chong-he: Resolved in [bd37444](https://github.com/sigp/lighthouse/pull/7462/commits/bd37444df8da073ec5e351c0cff48b9c80aa531c)

beacon_node/http_api/src/lib.rs
chong-he: Modifed accordingly in [bd37444](https://github.com/sigp/lighthouse/pull/7462/commits/bd37444df8da073ec5e351c0cff48b9c80aa531c)

michaelsproul (approved):

PR #7463, opened 2025-05-15
Fix mergify infinite loop.

jimmygchen:
## Issue Addressed

Fix mergify loop on this PR:
https://github.com/sigp/lighthouse/pull/7451

Added a condition to check that PR is open.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -39,10 +39,11 @@ pull_request_rules:
   - name: Update labels when PR is unblocked
     conditions:
       - -closed
+      - -draft
       - label=waiting-on-author
       - -conflict
-      - check-failure!=test-suite-success
-      - check-failure!=local-testnet-success
+      - check-skipped!=test-suite-success
+      - check-skipped!=local-testnet-success
       - "#review-requested > 0"
     actions:
       label:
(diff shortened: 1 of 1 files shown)


michaelsproul (approved):

.github/mergify.yml
```diff
@@ -20,6 +21,7 @@ pull_request_rules:
 
   - name: Ask to resolve CI failures
     conditions:
+      - -closed
       - or:
           - check-failure=test-suite-success
           - check-skipped=test-suite-success
```
michaelsproul, line 27: Maybe we need to mirror the opposite of this condition in the ready-for-review task?

.github/mergify.yml
```diff
@@ -36,10 +38,12 @@ pull_request_rules:
...
   - name: Update labels when PR is unblocked
     conditions:
+      - -closed
+      - -draft
       - label=waiting-on-author
       - -conflict
-      - check-failure!=test-suite-success
-      - check-failure!=local-testnet-success
+      - check-skipped!=test-suite-success
+      - check-skipped!=local-testnet-success
```
michaelsproul: I think you need all 4 of these, otherwise we have the same problem in reverse don't we. If the checks fail, we add waiting-on-author, but then this job adds ready-for-review immediately after and then we loop.

jimmygchen: Added

.github/mergify.yml
jimmygchen: Added

michaelsproul (commented):
looks good

Issue #7398, opened 2025-05-05
create subcommand of validator-manager fails in debug mode


ch4r10t33r:
The `create` subcommand of `validator-manager` fails with the following error

```
thread 'main' panicked at /Users/partha/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/clap_builder-4.5.27/src/builder/debug_asserts.rs:220:13:
Command create: Argument or group 'at-most' specified in 'conflicts_with*' for 'count' does not exist
stack backtrace:
   0: rust_begin_unwind
             at /rustc/4eb161250e340c8f48f66e2b929ef4a5bed7c181/library/std/src/panicking.rs:692:5
   1: core::panicking::panic_fmt
             at /rustc/4eb161250e340c8f48f66e2b929ef4a5bed7c181/library/core/src/panicking.rs:75:14
   2: clap_builder::builder::debug_asserts::assert_app
             at /Users/partha/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/clap_builder-4.5.27/src/builder/debug_asserts.rs:220:13
   3: clap_builder::builder::command::Command::_build_self
             at /Users/partha/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/clap_builder-4.5.27/src/builder/command.rs:4404:13
   4: clap_builder::builder::command::Command::_build_subcommand
             at /Users/partha/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/clap_builder-4.5.27/src/builder/command.rs:4491:9
   5: clap_builder::parser::parser::Parser::parse_subcommand
             at /Users/partha/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/clap_builder-4.5.27/src/parser/parser.rs:722:27
   6: clap_builder::parser::parser::Parser::parse
             at /Users/partha/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/clap_builder-4.5.27/src/parser/parser.rs:497:17
   7: clap_builder::parser::parser::Parser::get_matches_with
             at /Users/partha/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/clap_builder-4.5.27/src/parser/parser.rs:57:13
   8: clap_builder::parser::parser::Parser::parse_subcommand
             at /Users/partha/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/clap_builder-4.5.27/src/parser/parser.rs:739:37
[...]

PR #7469, opened 2025-05-16
Delete `at-most` in `lighthouse vm create`

chong-he:
## Issue Addressed

* #7398 
* 
## Additional Info

The `at-most` is not a flag in `lighthouse vm create`, thus causing the error when running in debug mode.


michaelsproul:
Blocked on the Clippy fix

michaelsproul (approved):

PR #7470, opened 2025-05-16
Prevent mergify from updating labels while CI is still running

jimmygchen:
## Issue Addressed

Mergify adds `ready-for-review` before the CI jobs have started / are running.

I've stripped out all the negation conditions to avoid the rules getting fired too easily.


michaelsproul (approved):
LGTM

PR #7471, opened 2025-05-16
Rust clippy 1.87 lint fixes

eserilev:
## Issue Addressed

Fix clippy lints for `rustc` 1.87 

## Proposed Changes

clippy complains about `BeaconChainError` being too large. I went on a bit of a boxing spree because of this. We may instead want to `Box` some of the `BeaconChainError` variants?


michaelsproul (approved):
I think it would be a bit cleaner to box inside `BeaconChainError`, but in the interests of unblocking CI I'm happy to merge this and revisit it later.

michaelsproul:
At a glance none of the `BeaconChainError` variants look bloated to me, so I'm curious which one it is.

PR #7474, opened 2025-05-16
Empty list `[]` to return all validators balances

chong-he:
## Issue Addressed

The endpoint `/eth/v1/beacon/states/head/validator_balances` returns an empty data when the data field is `[]`. According to the beacon API spec, it should return the balances of all validators:

Reference: https://ethereum.github.io/beacon-APIs/#/Beacon/postStateValidatorBalances
`If the supplied list is empty (i.e. the body is []) or no body is supplied then balances will be returned for all validators.`
## Proposed Changes

This PR changes so that: `curl -X 'POST' 'http://localhost:5052/eth/v1/beacon/states/head/validator_balances' -d '[]' | jq` returns balances of all validators.

## Additional Info

One thing I am not sure, the spec also mentions when `no body is supplied`. If query without body:
`curl -X 'POST' 'http://localhost:5052/eth/v1/beacon/states/head/validator_balances'  | jq` 

it errors:
`{
  "code": 400,
  "message": "BAD_REQUEST: body deserialize error: Error(\"EOF while parsing a value\", line: 1, column: 0)",
  "stacktraces": []
}
`

~~Should this be fixed too?~~
Update: added a `json_no_body` function to the HTTP request that will return balances of all validators if no body is supplied [811691b](https://github.com/sigp/lighthouse/pull/7474/commits/811691bc3f138258ede41dc27ce982a036fe9054)

Thanks @michaelsproul for bringing this up


michaelsproul:
Oh yeah, we should fix the empty body case as well. We might need a new kind of warp filter, or we could try to make the `ValidatorBalancesRequestBody` type deserialize from `""`. But I think we probably need something like an "optional JSON" filter.

chong-he:
>  But I think we probably need something like an "optional JSON" filter.

Is this about modifying the function here / adding a new function here? https://github.com/sigp/lighthouse/blob/stable/common/warp_utils/src/json.rs

michaelsproul:
@chong-he, yeah! we could make a copy of that function that converts `bytes == []` into `None`, or something like that

michaelsproul (approved):
This looks good! I think using `Default` is a good way to do it!

Issue #7461, opened 2025-05-15
Gossip blobs/data columns processed even after block import via EL blobs


jimmygchen:
On **PeerDAS devnet-7**, gossip messages for data columns (blobs) are being processed even *after* the corresponding block has been imported via `engine_getBlobs`.

This behaviour stems from the combination of EL blob fetching and Lighthouse’s batch re-publishing mechanism, as outlined in our [blog post](https://blog.sigmaprime.io/peerdas-distributed-blob-building.html#solutions-and-possible-optimisations):

1. The node receives a block via gossip and attempts to fetch the associated blobs from the EL.
2. If all blobs are successfully returned, the node imports the beacon block and begins **re-publishing blobs/data columns in batches** ([code reference](https://github.com/sigp/lighthouse/blob/0f13029c7d516ef931d8043436b811c41de3cba4/beacon_node/network/src/network_beacon_processor/mod.rs#L985)).

### Why this happens

The batch publishing mechanism was introduced to conserve bandwidth and prevent network-wide duplication. Each batch contains a randomly selected subset of blobs, ideally allowing full blob coverage early on without flooding.

However, the current batch size is [`1`](https://github.com/sigp/lighthouse/blob/0f13029c7d516ef931d8043436b811c41de3cba4/beacon_node/network/src/network_beacon_processor/mod.rs#L1013), and the node only marks a blob as “observed” in the [beacon chain cache](https://github.com/sigp/lighthouse/blob/593390162f470906c8364d3a4117164a0a438e72/beacon_node/beacon_chain/src/beacon_chain.rs#L419-L422) *right before* it publishes it ([code reference](https://github.com/sigp/lighthouse/blob/0f13029c7d516ef931d8043436b811c41de3cba4/beacon_node/network/src/network_beacon_processor/mod.rs#L1018)).

This means:

* The node may import all 4 blobs from the EL, but only "observe" one at a time as it publishes them.
* There's a 200ms delay between publishing each blob.
* Until a blob is published and marked as observed, the node does not consider it “seen”.

As a result, blobs/data columns arriving via gossip before being published locally may:

* Be unnecessarily processed again,
* Trigger additional work (e.g. redundant block imports),
* Attempt to forward duplicate message already published earlier.

### Impact

This issue is particularly noticeable in **PeerDAS** due to the larger volume of data column messages. It contributes to performance degradation and is likely related to the behaviour described in [#6439](https://github.com/sigp/lighthouse/issues/6439).

### Ideas

[...]

PR #7493, opened 2025-05-20
Verify `getBlobsV2` response and avoid reprocessing imported data columns

jimmygchen:
## Issue Addressed

#7461 and partly #6439.

Desired behaviour after receiving `engine_getBlobs` response:

1. Gossip verify the blobs and proofs, but don't mark them as observed yet. This is because not all blobs are published immediately (due to staggered publishing). If we mark them as observed and not publish them, we could end up blocking the gossip propagation.
2. Blobs are marked as observed _either_ when:
   * They are received from gossip and forwarded to the network .
   * They are published by the node.

Current behaviour:
- ❗ We only gossip verify `engine_getBlobsV1` responses, but not `engine_getBlobsV2` responses (PeerDAS).
- ❗ After importing EL blobs AND before they're published, if the same blobs arrive via gossip, they will get re-processed, which may result in a re-import.

## Proposed Changes

1. Perform gossip verification on data columns computed from EL `getBlobsV2` response. We currently only do this for `getBlobsV1` to prevent importing blobs with invalid proofs into the `DataAvailabilityChecker`, this should be done on V2 responses too.
2. Add additional gossip verification to make sure we don't re-process a ~~blob~~ or data column that was imported via the EL `getBlobs` but not yet "seen" on the gossip network. If an "unobserved" gossip blob is found in the availability cache, then we know it has passed verification so we can immediately propagate the `ACCEPT` result and forward it to the network, but without re-processing it.

**UPDATE:** I've left blobs out for the second change mentioned above, as the likelihood and impact is very slow and we haven't seen it enough, but under PeerDAS this issue is a regular occurrence and we do see the same block getting imported many times.

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -3673,10 +3673,12 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         custody_columns: impl IntoIterator<Item = &'a DataColumnSidecar<T::EthSpec>>,
     ) -> Result<(), BlockError> {
         let mut slashable_cache = self.observed_slashable.write();
-        // Assumes all items in custody_columns are for the same block_root
-        if let Some(column) = custody_columns.into_iter().next() {
-            let header = &column.signed_block_header;
-            if verify_header_signature::<T, BlockError>(self, header).is_ok() {
+        for header in custody_columns
+            .into_iter()
+            .map(|c| c.signed_block_header.clone())
+            .unique()
+        {
+            if verify_header_signature::<T, BlockError>(self, &header).is_ok() {
                 slashable_cache
                     .observe_slashable(
                         header.message.slot,
@@ -3685,7 +3687,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
                     )
                     .map_err(|e| BlockError::BeaconChainError(Box::new(e.into())))?;
                 if let Some(slasher) = self.slasher.as_ref() {
-                    slasher.accept_block_header(header.clone());
+                    slasher.accept_block_header(header);
                 }
             }
         }
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -3545,7 +3545,9 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         if let Some(slasher) = self.slasher.as_ref() {
             slasher.accept_block_header(blob.signed_block_header());
         }
-        let availability = self.data_availability_checker.put_gossip_blob(blob)?;
+        let availability = self
+            .data_availability_checker
+            .put_gossip_verified_blobs(blob.block_root(), std::iter::once(blob))?;
 
         self.process_availability(slot, availability, || Ok(()))
             .await
@@ -3568,7 +3570,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
 
         let availability = self
             .data_availability_checker
-            .put_gossip_data_columns(block_root, data_columns)?;
+            .put_gossip_verified_data_columns(block_root, data_columns)?;
 
         self.process_availability(slot, availability, publish_fn)
             .await
@@ -3628,7 +3630,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             EngineGetBlobsOutput::Blobs(blobs) => {
                 self.check_blobs_for_slashability(block_root, blobs.iter().map(|b| b.as_blob()))?;
                 self.data_availability_checker
-                    .put_engine_blobs(block_root, blobs)?
+                    .put_gossip_verified_blobs(block_root, blobs)?
             }
             EngineGetBlobsOutput::CustodyColumns(data_columns) => {
                 self.check_columns_for_slashability(
@@ -3636,7 +3638,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
                     data_columns.iter().map(|c| c.as_data_column()),
                 )?;
                 self.data_availability_checker
-                    .put_engine_data_columns(block_root, data_columns)?
+                    .put_gossip_verified_data_columns(block_root, data_columns)?
             }
         };
 
@@ -3673,6 +3675,9 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         custody_columns: impl IntoIterator<Item = &'a DataColumnSidecar<T::EthSpec>>,
     ) -> Result<(), BlockError> {
         let mut slashable_cache = self.observed_slashable.write();
+        // Process all unique block headers - previous logic assumed all headers were identical and
+        // only processed the first one. However, we should not make assumptions about data received
+        // from RPC.
         for header in custody_columns
             .into_iter()
             .map(|c| c.signed_block_header.clone())
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -3146,7 +3146,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         self: &Arc<Self>,
         slot: Slot,
         block_root: Hash256,
-        engine_get_blobs_output: EngineGetBlobsOutput<T::EthSpec>,
+        engine_get_blobs_output: EngineGetBlobsOutput<T>,
     ) -> Result<AvailabilityProcessingStatus, BlockError> {
         // If this block has already been imported to forkchoice it must have been available, so
         // we don't need to process its blobs again.
@@ -3161,7 +3161,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         // process_engine_blobs is called for both pre and post PeerDAS. However, post PeerDAS
         // consumers don't expect the blobs event to fire erratically.
         if let EngineGetBlobsOutput::Blobs(blobs) = &engine_get_blobs_output {
-            self.emit_sse_blob_sidecar_events(&block_root, blobs.iter().flatten().map(Arc::as_ref));
+            self.emit_sse_blob_sidecar_events(&block_root, blobs.iter().map(|b| b.as_blob()));
         }
 
         let r = self
@@ -3574,15 +3574,15 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             .await
     }
 
-    fn check_blobs_for_slashability(
+    fn check_blobs_for_slashability<'a>(
         self: &Arc<Self>,
         block_root: Hash256,
-        blobs: &FixedBlobSidecarList<T::EthSpec>,
+        blobs: impl IntoIterator<Item = &'a BlobSidecar<T::EthSpec>>,
     ) -> Result<(), BlockError> {
         let mut slashable_cache = self.observed_slashable.write();
         for header in blobs
-            .iter()
-            .filter_map(|b| b.as_ref().map(|b| b.signed_block_header.clone()))
+            .into_iter()
+            .map(|b| b.signed_block_header.clone())
             .unique()
         {
             if verify_header_signature::<T, BlockError>(self, &header).is_ok() {
@@ -3609,7 +3609,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         block_root: Hash256,
         blobs: FixedBlobSidecarList<T::EthSpec>,
     ) -> Result<AvailabilityProcessingStatus, BlockError> {
-        self.check_blobs_for_slashability(block_root, &blobs)?;
+        self.check_blobs_for_slashability(block_root, blobs.iter().flatten().map(Arc::as_ref))?;
         let availability = self
             .data_availability_checker
             .put_rpc_blobs(block_root, blobs)?;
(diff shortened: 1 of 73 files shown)


beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -444,6 +458,21 @@ pub fn validate_data_column_sidecar_for_gossip<T: BeaconChainTypes, O: Observati
...
+    if chain
+        .data_availability_checker
+        .is_data_column_cached(&data_column.block_root(), &data_column)
+    {
+        if O::observe() {
+            observe_gossip_data_column(&data_column, chain)?;
+        }
+        // TODO should this really be an error?
+        return Err(GossipDataColumnError::PriorKnownUnpublished);
+    }
```
jimmygchen, line 484: I don't really like my solution here and considering an alternative here.

Here I want a way to mark a gossip data column as valid BUT not re-processing them - so we can propagate but not re-process and forward them onto the gossip network. Options I have in my mind:
1. Return an error like what I did there - although i don't feel like this is really an error and will result in extra slashing info processing [here](https://github.com/sigp/lighthouse/blob/fb4cda00bd07c2e20ec5086ea9c46ea7322c1e79/beacon_node/beacon_chain/src/data_column_verification.rs#L209-L212)
2. Return an `Result<(GossipVerifiedDataColumn, bool), _>` or an enum that distinguishes between an observed / unobserved data column. This feels a bit more correct but may look a bit weird?

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -180,7 +177,7 @@ async fn fetch_and_process_blobs_v1<T: BeaconChainTypes>(
         .process_engine_blobs(
             block.slot(),
             block_root,
-            EngineGetBlobsOutput::Blobs(fixed_blob_sidecar_list.clone()),
+            EngineGetBlobsOutput::Blobs(blobs_to_import_and_publish),
```
jimmygchen: Bug fix: we previously pass all blobs instead of `blobs_to_import_and_publish` to processing.

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -26,24 +27,17 @@ use tracing::debug;
...
+    BeaconStateError, Blob, BlobSidecar, ChainSpec, ColumnIndex, EthSpec, FullPayload, Hash256,
+    KzgProofs, SignedBeaconBlock, SignedBeaconBlockHeader, VersionedHash,
 };
 
-/// Blobs or data column to be published to the gossip network.
-pub enum BlobsOrDataColumns<T: BeaconChainTypes> {
+/// Result from engine get blobs to be passed onto `DataAvailabilityChecker` and published to the
+/// gossip network. The blobs / data columns have not been marked as observed yet, as they may not
+/// be published immediately.
+pub enum EngineGetBlobsOutput<T: BeaconChainTypes> {
```
jimmygchen: Consolidating `BlobsOrDataColumns` with this type as we don't really need a separate type.

beacon_node/beacon_chain/src/data_column_verification.rs
jimmygchen: I had a chat with Michael and we agreed that using an error here isn't too bad - as long as it gets handled at the call sites.

jimmygchen: I've tried adding handling code at the caller sites and realised I need to have the `GossipVerifiedDataColumn` wrapped in the error, e.g. [here](https://github.com/sigp/lighthouse/blob/268809a53069a343318918d7c523523f33f1bf0e/beacon_node/http_api/src/publish_blocks.rs#L413-L426). This makes it a bit ugly because it means the `GossipDataColumnError` and the parent enum types all need to be generic `T: BeaconChainTypes`.

I'm going to try a different approach.

jimmygchen: Ok, I've tried the enum approach and I'm not sure if it's necessary.

Whenever we get `GossipDataColumnError::PriorKnownUnpublished`, it means the data column was fetched via `engine_getBlobs` earlier, and is scheduled for batch publishing (but not yet published).

I think the only case we need to explicitly handle this is in `NetworkBeaconProcessor::process_gossip_data_column_sidecar`, i.e. when we receive a gossip data column; and here we immediately propagate the already processed data column without processing / importing.

The other place that we call gossip verify data columns is via the HTTP api `publish_block` - if it's already processed, we can just ignore it, and not include it in publishing again. It's not important to publish here because if it's imported via EL blobs, it's highly likely that the publishing is already scheduled (stagger publishing). We could reconsider this though if we want to "force" an immediate publish.

jimmygchen: I've just pushed a new commit to ensure that we handle this at all call sites: https://github.com/sigp/lighthouse/pull/7493/commits/7c5e55af7131fec47cd8cc02eaff714866fdb315

- **gossip**: propagate ACCEPT without importing
- **HTTP**: skip already processed blobs

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -181,6 +185,16 @@ pub struct GossipVerifiedDataColumn<T: BeaconChainTypes, O: ObservationStrategy
     _phantom: PhantomData<O>,
 }
 
+impl<T: BeaconChainTypes, O: ObservationStrategy> Clone for GossipVerifiedDataColumn<T, O> {
+    fn clone(&self) -> Self {
+        Self {
+            block_root: self.block_root,
+            data_column: self.data_column.clone(),
```
dapplion, line 192: Why don't you add a new property to `GossipVerifiedDataColumn` named `should_import`?

jimmygchen: I don't feel like import logic belongs to gossip verification, so I'd prefer to avoid that and leave the handling logic at the caller site (unless necessary), rather than having import logic at multiple places.

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3661,16 +3667,18 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
     ) -> Result<(), BlockError> {
         let mut slashable_cache = self.observed_slashable.write();
-        // Assumes all items in custody_columns are for the same block_root
-        if let Some(column) = custody_columns.first() {
-            let header = &column.signed_block_header;
-            if verify_header_signature::<T, BlockError>(self, header).is_ok() {
+        for header in custody_columns
+            .into_iter()
+            .map(|c| c.signed_block_header.clone())
+            .unique()
```
pawanjay176, line 3684: This would process multiple headers if they exist as opposed to the older behaviour where only one header is processed.
This won't happen based on how this function is called currently, but might be worth documenting this behaviour.

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -219,21 +235,14 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
     }
 
     /// Put a list of blobs received from the EL pool into the availability cache.
-    ///
-    /// This DOES NOT perform KZG verification because the KZG proofs should have been constructed
-    /// immediately prior to calling this function so they are assumed to be valid.
-    pub fn put_engine_blobs(
+    pub fn put_engine_blobs<O: ObservationStrategy>(
```
pawanjay176: This could potentially be merged with `put_gossip_blob` given they have the same logic?

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -248,18 +257,14 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
     /// This method is used if the EL already has the blobs and returns them via the `getBlobsV2`
     /// engine method.
     /// More details in [fetch_blobs.rs](https://github.com/sigp/lighthouse/blob/44f8add41ea2252769bb967864af95b3c13af8ca/beacon_node/beacon_chain/src/fetch_blobs.rs).
-    pub fn put_engine_data_columns(
+    pub fn put_engine_data_columns<O: ObservationStrategy>(
```
pawanjay176: This could also potentially be merged with `put_gossip_data_columns`?

beacon_node/beacon_chain/src/fetch_blobs.rs
```diff
@@ -314,8 +311,36 @@ async fn compute_and_publish_data_columns<T: BeaconChainTypes>(
...
+                        let col_index = col.index;
+                        match GossipVerifiedDataColumn::<T, DoNotObserve>::new(
+                            col,
+                            col_index,
+                            &chain_cloned,
+                        ) {
+                            Ok(verified) => Some(Ok(verified)),
+                            // Ignore already seen data columns
+                            Err(GossipDataColumnError::PriorKnown { .. })
+                            | Err(GossipDataColumnError::PriorKnownUnpublished) => None,
```
pawanjay176: can we match all errors here? if there's another error variant we add that we need to potentially ignore later, then it will take forever to find the bug with fetch_blobs

pawanjay176 (approved):
Looks good mostly, just nits

beacon_node/beacon_chain/src/data_availability_checker.rs
jimmygchen: Yeah good catch - since we now do verification on EL blobs, this function is no longer necessary (docs above is also outdated). 

With this change we do delay our import block time for a bit though and should make sure our gossip verification is fast (rn it's quite slow on large blob count) -  this just reminded me that skipping verification was an intentional decision made when this was originally implemented - i think if we go down this route, there might be some parts that we should optimise in the verify method, e.g. verifying the block header, signature etc - potentially we could just verify the kzg proof? cc @michaelsproul

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Yep I can add a line to document this - this is the same behaviour as blobs now. I didn't know we were only checking the header of the first data column, seems a bit too permissive.

beacon_node/beacon_chain/src/data_availability_checker.rs
jimmygchen: good catch!

beacon_node/beacon_chain/src/fetch_blobs.rs
jimmygchen: I generally agree with this, although in this case `PriorKnown`s are kind of like the exception here and we don't do any handling for other error variants, and just let errors propagate up. I imagine for new variants we'd want to do the same thing?

jimmygchen: I think matching all errors is generally good practice, though I believe we should stay pragmatic and evaluate on a case by case basis. I don't have a strong preference here, happy to make the change to get things moving, just wanted to share my thoughts.

jimmygchen: Addressed in 966dadae6f00f0e363953d42d6861e25388c7a2b

beacon_node/beacon_chain/src/data_availability_checker.rs
jimmygchen: Addressed in 966dadae6f00f0e363953d42d6861e25388c7a2b

I think it may be worth taking out gossip verification though, but will have to revive this function (to accept kzg verified columns instead of gossip columns) - let me know what you think.

beacon_node/beacon_chain/src/fetch_blobs.rs
jimmygchen: Addressed in 966dadae6f00f0e363953d42d6861e25388c7a2b

beacon_node/beacon_chain/src/data_availability_checker.rs
jimmygchen: Addressed in 966dadae6f00f0e363953d42d6861e25388c7a2b

I think it may be worth taking out gossip verification though, but will have to revive this function (to accept kzg verified columns instead of gossip columns) - let me know what you think.

beacon_node/beacon_chain/src/data_availability_checker.rs
jimmygchen: Addressed in https://github.com/sigp/lighthouse/commit/966dadae6f00f0e363953d42d6861e25388c7a2b

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Addressed in https://github.com/sigp/lighthouse/commit/966dadae6f00f0e363953d42d6861e25388c7a2b

jimmygchen:
Resolved merge conflicts but CI is failing due to an issue with GitHub actions.

beacon_node/beacon_chain/src/data_availability_checker.rs
pawanjay176: I think the main reason we do gossip verification is because we propagate the columns on gossip so we want to make sure its valid.
We can avoid the proposer signature validation which I would guess is the other factor which takes some significant amount time in gossip verification (other than kzg validation)

PR #7494, opened 2025-05-20
Update mergify rule to avoid unnecessarily firing some rules

jimmygchen:
I've noticed that mergify spams PRs with comments even before they're marked as reviewed. It also comments on PRs that are stale and not in the `ready-for-review` state.

This PR updates those mergify rules to not evaluate PRs that are not ready for review, so that it reduces noise and avoid updating stale PRs.


antondlr (approved):
lgtm!

macladson (approved):

Issue #7489, opened 2025-05-19
Add fork version fallback for `GET /eth/v1/beacon/blobs`


michaelsproul:
In our effort to follow the beacon API spec, we followed it a bit too hard and broke compatibility with [checkpointz](https://github.com/ethpandaops/checkpointz).

The issue is, since we refactored our fork-version handling, Lighthouse is now strict about receiving a `version` field in responses on `/eth/v1/beacon/blobs/{block_id}` which we use for checkpoint sync. The result is that checkpoint sync fails with `checkpointz`, because the server doesn't include the `version` field despite it being mandatory.

The fork versioning change was made in:

- https://github.com/sigp/lighthouse/pull/7372

To maintain backwards-compatibility for v7.1.0 we should consider relaxing our `version` parsing by using the `Electra` fork by default if none is specified. This is pretty janky, so we should **also delete this default** once enough `checkpointz` instances are updated.

Thanks to @barnabasbusa for reporting!

jimmygchen:
Completed in #7495

PR #7495, opened 2025-05-20
Fix blob download from checkpointz servers

michaelsproul:
## Issue Addressed

Closes:

- https://github.com/sigp/lighthouse/issues/7489

## Proposed Changes

Use `ForkName::latest_stable`, i.e. Electra when decoding blobs from a server that does not provide `version`. This is only a temporary workaround that should be reverted once `checkpointz` is fixed. Having a default fork is potentially incorrect, and glossing over bugs in servers in general is not ideal.

However, even in the case where we update `ForkName::latest_stable` to `Fulu`, this should continue to work, as the blob limit is likely to increase and the `RuntimeVariableList` will just have a slightly higher limit than necessary (which is OK so long as the server isn't buggy enough to violate the correct lower bound: e.g. if the block is an Electra one and the server sends 10 blobs, which exceeds the Electra max (9) but not the Fulu max).

## Additional Info

I've opened an issue to revert this PR before we proceed with Fulu on mainnet, but _after_ v7.1.0:

- https://github.com/sigp/lighthouse/issues/7496

### DIFF
--- a/common/eth2/src/lib.rs
+++ b/common/eth2/src/lib.rs
@@ -307,6 +307,8 @@ impl BeaconNodeHttpClient {
 
         #[derive(serde::Deserialize)]
         struct Helper {
+            // TODO: remove this default once checkpointz follows the spec
+            #[serde(default = "ForkName::latest_stable")]
             version: ForkName,
             #[serde(flatten)]
             metadata: serde_json::Value,


jimmygchen (approved):
LGTM, thanks!
(also having issues checkpoint syncing :/)

PR #7497, opened 2025-05-20
Pass blobs into `ValidatorStore::sign_block`

dknopik:
## Issue Addressed

While the Lighthouse implementation of the `ValidatorStore` does not really care about blobs, Anchor needs to be able to return different blobs from `sign_blocks` than what was passed into it, in case it decides to sign another Anchor node's block. Only passing the unsigned block into `sign_block` and only returning a signed block from it (without any blobs and proofs) was an oversight in #6705.

## Proposed Changes

- Replace `validator_store::{Uns,S}ignedBlock` with `validator_store::block_service::{Uns,S}ignedBlock`, as we need all data in there.
- In `lighthouse_validator_store`, just add the received blobs back to the signed block after signing it.


dknopik:
@michaelsproul Could you please take a look? :pray:

michaelsproul (approved):
Looks good

Nice to get rid of some of those leftover types in `block_service` too!

Issue #7002, opened 2025-02-14
Resolve light client errors at fork boundaries


michaelsproul:
I think we should pause light client computations if the fork epoch is misaligned in a way that breaks our calculations.

This should be necessary only for ~1 sync committee period around the fork.

It would be nice to avoid error logs, even if they are harmless, like:

> Feb 13 20:25:36.228 ERRO error computing light_client updates LightClientUpdateError(BeaconStateError(IncorrectStateVariant)), service: lc_update

eliotstock:
Just started seeing these immediately after Pectra. Should I be bothered?

michaelsproul:
@eliotstock No, they're harmless. Should be gone now that finalization has happened.

We'll fix it for future forks. 

We thought we'd already fixed this, but the fix wasn't effective in all circumstances

eliotstock:
Oh yeah, it's gone now. Cool, cheers.

michaelsproul:
Adding this to v8.0.0 because it happened again at Fulu. We should at least just downgrade the error to a debug log.

koxu1996:
I started seeing such errors just after Fulu was applied (nice ASCII art btw!):

```
ERROR error computing light_client updates LightClientUpdateError(BeaconStateError(IncorrectStateVariant))
```

Restarting Lighthouse didn't help.

michaelsproul:
Still an issue in rc.1 but the recently merged PR fixes this error for rc.2 and v8.0.0

- https://github.com/sigp/lighthouse/pull/8196

PR #7502, opened 2025-05-20
Allow for  constructing light client data across fork boundaries

eserilev:
## Issue Addressed

#7002

## Proposed Changes

`LightClientUpdate` data can contain an attested header and finalized header that originate from different forks. This usually happens during hardforks while an epoch in the new fork hasn't been finalized yet. This PR allows for us to construct `LightClientUpdate`s in this scenario.

### DIFF
--- a/consensus/types/src/light_client_update.rs
+++ b/consensus/types/src/light_client_update.rs
@@ -1,23 +1,21 @@
 use super::{EthSpec, FixedVector, Hash256, Slot, SyncAggregate, SyncCommittee};
-use crate::context_deserialize;
 use crate::light_client_header::LightClientHeaderElectra;
 use crate::LightClientHeader;
 use crate::{
-    beacon_state, test_utils::TestRandom, ChainSpec, ContextDeserialize, Epoch, ForkName,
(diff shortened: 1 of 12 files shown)


consensus/types/src/light_client_update.rs
```diff
@@ -529,6 +488,148 @@ impl<E: EthSpec> LightClientUpdate<E> {
...
+        // FINALIZED HEADER
+        builder.register_anonymous_variable_length_item()?;
+        if fork_name.electra_enabled() {
+            builder.register_type::<FinalityBranchElectra>()?;
+        } else {
+            builder.register_type::<FinalityBranch>()?;
+        }
+        builder.register_type::<SyncAggregate<E>>()?;
+        builder.register_type::<Slot>()?;
+        let mut decoder = builder.build()?;
```
eserilev, line 514: There is an `OffsetOutOfBounds` failure on line 506 when I try to register the finalized header as an anonymous variable length item. I'm not sure why

eserilev: `cargo test --release -p ef_tests --features "ef_tests" -- light_client_update`

PR #7505, opened 2025-05-21
Update `engine_getBlobsV2` response type and add `getBlobsV2` tests

jimmygchen:
## Issue Addressed

Update `engine_getBlobsV2` response type to `Option<Vec<BlobsAndProofV2>>`. See recent spec change [here](https://github.com/ethereum/execution-apis/pull/630).

Added some tests to cover basic fetch blob scenarios.


dapplion:
@jimmygchen Why not use mockEL to return arbitrary data, like `nil` for `engine_getBlobsV2`?

jimmygchen:
> @jimmygchen Why not use mockEL to return arbitrary data, like `nil` for `engine_getBlobsV2`?

It _might_ be possible, but would be:
1. more difficult to get it working
2. less flexible for more edge case scenarios
3. difficult to maintain due to the amount of set up required
4. I also expect it to be slower if we need to set up a real chain with blocks.

Currently we don't write as many tests because our test machineries are time consuming to setup and are more designed for integration-style tests - these tests are important but  expensive and not as flexible to cover edge case scenarios. Going forward I'm hoping to write more unit tests to make sure we cover important code path as much as possible - I think once we get the infra & pattern set up, this would be much easier to maintain and scale.

jimmygchen:
To be a bit more concrete, using mocks also allow me to easily make assertion about what's passed the the function and what should be returned , and number of calls made. I can easily manipulate the return value of calls to the `BeaconChain` as well, without having to set up a real beacon chain.

pawanjay176 (approved):
LGTM functionality wise, we should get it in before the fulu devnet.

Testing wise, I'm still not fully convinced that the unit tests with the mocking approach is entirely necessary in this specific part of the code.
A lot of our code in the `beacon_chain` crate is tightly coupled and I feel that writing unit tests for components which are tightly coupled with other components is a little forced.
 
However, I'm willing to give this a go and see how this sort of testing approach would benefit us in other parts of the code further down the line. We can revisit it later and reevaluate if it is useful or just a maintenance burden later on.

jimmygchen:
Thanks @pawanjay176 appreciate the feedback 🙏  yeah happy to think and discuss about unit testing a bit more.

PR #7506, opened 2025-05-21
Enable stdout logging in rpc_tests

ackintosh:
## Issue Addressed

Currently `test_delayed_rpc_response` is flaky (possibly specific to Windows?), but I'm not sure why.

## Proposed Changes

Enabled stdout logging in rpc_tests. Note that in nextest, std output is only displayed when a test fails.


jimmygchen (approved):
LGTM!

PR #7507, opened 2025-05-22
Use slice.is_sorted now that it's stable

michaelsproul:
## Proposed Changes

Use slice.is_sorted which was stabilised in Rust 1.82.0

I thought there would be more places we could use this, but it seems we often want to check strict monotonicity (i.e. sorted + no duplicates)


jimmygchen (approved):
LGTM!

PR #7510, opened 2025-05-22
Downscore and retry custody failures

dapplion:
## Issue Addressed

Partially addresses 

## Proposed Changes

TBD

1000 lines are for new unit tests :)

## Questions / TODO

- Should custody_by_range requests try all possible peers before giving up? i.e. should they ignore the failures counter for custody failures? Add a test for this behaviour.

**Tests to add**

- Peer does not send columns at all on specific index
- We find no-one serving columns on a specific index, how to recover
- Random tests where most peer are faulty and we have to keep cycling them. Use randomness and run the tests with different levels of fault % :=> build a simulator
- Permanently fail a single column many times, but then resolve the rest such that we can reconstruct.
- Send non-matching data columns
- Test downscoring of custody failures
- Test the fallback mechanism of peer selection
- Test a syncing chain reaching 0 peers, both for forward sync and backwards sync
- Test forwards sync on Fulu with blocks without data

### Future TODOs

- Ensure uniform use of `subscribe_all_data_column_subnets` to determine `is_supernode`
- Ensure consistent naming in sync network context handlers `on_xxx_response`, suggestion -> https://github.com/sigp/lighthouse/pull/7510#discussion_r2107334518
- Cleanup `add_peer_*` functions in sync tests. Best to do in another PR as it touches a lot of lookup sync tests (unrelated) 
- Track `requests_per_peer` to ensure that eventually we fetch data from all peers in the chain


beacon_node/network/src/sync/manager.rs
```diff
@@ -545,6 +545,9 @@ impl<T: BeaconChainTypes> SyncManager<T> {
         for (id, result) in self.network.continue_custody_by_root_requests() {
             self.on_custody_by_root_result(id, result);
         }
+        for (id, result) in self.network.continue_custody_by_range_requests() {
+            self.on_custody_by_range_result(id, result);
+        }
```
dapplion, line 560: Every interval (15 sec), we call continue_custody_by_range / by_root requests, which will cause the request to error if it has been alive for too long. This allows the requests to not error immediately if they do not have enough custody peers.

beacon_node/network/src/sync/manager.rs
```diff
@@ -442,6 +439,9 @@ impl<T: BeaconChainTypes> SyncManager<T> {
...
```
dapplion, line 453: Every time a peer joins, attempt to progress custody_by_root and custody_by_range requests

beacon_node/network/src/sync/mod.rs
```diff
@@ -3,7 +3,6 @@
 //! Stores the various syncing methods for the beacon chain.
 mod backfill_sync;
 mod block_lookups;
-mod block_sidecar_coupling;
```
dapplion, line 6: Logic moved to `beacon_node/network/src/sync/network_context/block_components_by_range.rs`

beacon_node/network/src/sync/network_context.rs
```diff
@@ -47,11 +48,13 @@ use tokio::sync::mpsc;
 use tracing::{debug, error, span, warn, Level};
 use types::blob_sidecar::FixedBlobSidecarList;
 use types::{
-    BlobSidecar, ColumnIndex, DataColumnSidecar, DataColumnSidecarList, EthSpec, ForkContext,
-    Hash256, SignedBeaconBlock, Slot,
+    BlobSidecar, ChainSpec, ColumnIndex, DataColumnSidecar, DataColumnSidecarList, Epoch, EthSpec,
+    ForkContext, Hash256, SignedBeaconBlock, SignedBeaconBlockHeader, Slot,
 };
 
-pub mod custody;
```
dapplion, line 54: Renamed existing `custody` module to `custody_by_root` and added a new one `custody_by_range`

beacon_node/network/src/sync/network_context.rs
```diff
@@ -297,6 +315,14 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
     /// Returns the ids of all the requests made to the given peer_id.
     pub fn peer_disconnected(&mut self, peer_id: &PeerId) -> Vec<SyncRequestId> {
+        self.active_requests()
+            .filter(|(_, request_peer)| *request_peer == peer_id)
+            .map(|(id, _)| id)
+            .collect()
+    }
+
+    /// Returns the ids of all active requests
+    pub fn active_requests(&mut self) -> impl Iterator<Item = (SyncRequestId, &PeerId)> {
```
dapplion, line 321: Changed this signature for tests, to have access to all active RPC requests

beacon_node/network/src/sync/network_context.rs
```diff
@@ -447,205 +485,95 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
-                                columns,
-                            },
-                            id,
-                        )
-                    })
-                    .collect::<Result<Vec<_>, _>>()?;
-
-                Ok((requests, column_to_peer_map))
-            })
-            .transpose()?;
```
dapplion, line 537: network_context no longer spawns _by_range requests, this logic is now inside `BlockComponentsByRangeRequest`

beacon_node/network/src/sync/network_context/block_components_by_range.rs
```diff
@@ -0,0 +1,550 @@
...
+    Base {
+        blocks_by_range_request:
+            ByRangeRequest<BlocksByRangeRequestId, Vec<Arc<SignedBeaconBlock<E>>>>,
+    },
+    // Two single concurrent requests for block + blobs
+    DenebEnabled {
+        blocks_by_range_request:
+            ByRangeRequest<BlocksByRangeRequestId, Vec<Arc<SignedBeaconBlock<E>>>>,
+        blobs_by_range_request: ByRangeRequest<BlobsByRangeRequestId, Vec<Arc<BlobSidecar<E>>>>,
+    },
```
dapplion, line 44: Maintains the same behaviour for mainnet:
- deneb: issue blocks + blobs requests at the same time
- fulu: issue blocks request first, then columns

beacon_node/network/src/sync/network_context.rs
```diff
@@ -72,32 +75,29 @@ impl<T> RpcEvent<T> {
...
-    CustodyRequestError(#[allow(dead_code)] CustodyRequestError),
-    BlockComponentCouplingError(#[allow(dead_code)] String),
+    RequestExpired(String),
+    InternalError(#[allow(dead_code)] String),
 }
 
 #[derive(Debug, PartialEq, Eq)]
 pub enum RpcRequestSendError {
-    /// No peer available matching the required criteria
-    NoPeer(NoPeerError),
```
dapplion, line 90: Requests do not error now on send if they don't have peers. Instead, custody_by_root and custody_by_range requests are left idle for some time, expecting peers.

beacon_node/lighthouse_network/src/types/globals.rs
```diff
@@ -245,6 +245,25 @@ impl<E: EthSpec> NetworkGlobals<E> {
         Self::new_test_globals_with_metadata(trusted_peers, metadata, config, spec)
     }
 
+    pub fn new_test_globals_as_supernode(
+        trusted_peers: Vec<PeerId>,
+        config: Arc<NetworkConfig>,
+        spec: Arc<ChainSpec>,
+        is_supernode: bool,
```
jimmygchen, line 282: I don't think it matters _currently_ - but thought might be worth mentioning - we use `config.subscribe_all_data_column_subnets` for determine the gossip topics, and it doesn't seem like we use it for determining `is_supernode` elsewhere, but something to keep in mind - if we want consistency maybe we can derive `is_supernode` from network config, but that may not be relevant from once we move to validator custody.

beacon_node/network/src/sync/manager.rs
```diff
@@ -1186,10 +1189,9 @@ impl<T: BeaconChainTypes> SyncManager<T> {
         block: RpcEvent<Arc<SignedBeaconBlock<T::EthSpec>>>,
     ) {
         if let Some(resp) = self.network.on_blocks_by_range_response(id, peer_id, block) {
-            self.on_range_components_response(
+            self.on_block_components_by_range_response(
```
jimmygchen, line 1217: I often get confused here as we have layers of `on_xxx_by_range_response` functions, what about renaming:
- `resp` to `result`
- `self.on_block_components_by_range_response` to `self.on_range_components_rpc_result`, and add a comment describing it:

```suggestion
            // Process it further if the response results in a completed request or a failure that needs to be handled.
            self.on_range_components_rpc_result(
```

jimmygchen: also might help to add some comment to `self.network.on_blocks_by_range_response`. I wrote these notes while reviewing:

```rust
    /// Processes the `BlocksByRange` response (`RpcEvent`), depending on the event, it may:
    /// - update the active request state; or do nothing.
    /// - complete the active request and return an `Ok` response containing a list of response objects.
    /// - propagate the error if necessary.
    pub(crate) fn on_blocks_by_range_response(
        &mut self,
        id: BlocksByRangeRequestId,
        peer_id: PeerId,
        rpc_event: RpcEvent<Arc<SignedBeaconBlock<T::EthSpec>>>,
    ) -> Option<RpcResponseResult<Vec<Arc<SignedBeaconBlock<T::EthSpec>>>>> {
```

beacon_node/network/src/sync/manager.rs
```diff
@@ -1215,18 +1216,46 @@ impl<T: BeaconChainTypes> SyncManager<T> {
...
-                RangeBlockComponent::CustodyColumns(id, resp),
-            );
+            // custody_by_range accumulates the results of multiple data_columns_by_range requests
+            // returning a bigger list of data columns across all the column indices this node has
+            // to custody
+            if let Some(result) =
+                self.network
+                    .on_custody_by_range_response(id.parent_request_id, id, peer_id, resp)
+            {
+                self.on_custody_by_range_result(id.parent_request_id, result);
```
jimmygchen, line 1253: Nice, this one is a bit more readable with the naming and comments.

beacon_node/network/src/sync/manager.rs
```diff
@@ -1215,18 +1216,46 @@ impl<T: BeaconChainTypes> SyncManager<T> {
...
-                id.parent_request_id,
-                peer_id,
-                RangeBlockComponent::CustodyColumns(id, resp),
-            );
+            // custody_by_range accumulates the results of multiple data_columns_by_range requests
+            // returning a bigger list of data columns across all the column indices this node has
+            // to custody
+            if let Some(result) =
+                self.network
+                    .on_custody_by_range_response(id.parent_request_id, id, peer_id, resp)
```
jimmygchen: We can probably remove the first parameter here, and get the `parenet_request_id` from the `id`.

beacon_node/network/src/sync/network_context.rs
```diff
@@ -1468,6 +1420,61 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+    ///
+    /// - `Some`: Request completed, won't make more progress. Expect requester to act on the result.
+    /// - `None`: Request still active, requester should do no action
+    #[allow(clippy::type_complexity)]
+    pub fn on_custody_by_range_response(
+        &mut self,
+        id: CustodyByRangeRequestId,
+        req_id: DataColumnsByRangeRequestId,
+        peer_id: PeerId,
+        resp: RpcResponseResult<Vec<Arc<DataColumnSidecar<T::EthSpec>>>>,
```
jimmygchen: Using `DataColumnSidecarList` can reduce the type complexity

beacon_node/network/src/sync/network_context.rs
```diff
@@ -1468,6 +1420,61 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+    ) -> Option<CustodyByRootResult<T::EthSpec>> {
```
jimmygchen: This should be `CustodyByRangeResult`. Thought I don't see a lot of benefits of having two type aliases for the same type - it doesn't provide more safety, and using it incorrectly but may add confusion.

beacon_node/network/src/sync/network_context.rs
```diff
@@ -1468,6 +1420,61 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+        // Note: need to remove the request to borrow self again below. Otherwise we can't
+        // do nested requests
+        let Some(mut request) = self.custody_by_range_requests.remove(&id) else {
+            // TOOD(das): This log can happen if the request is error'ed early and dropped
```
jimmygchen: typo in TODO

jimmygchen (commented):
@dapplion I've started reviewing this but haven't got to the meat of the changes - will continue tomorrow but I've submitted my comments so far.

beacon_node/lighthouse_network/src/types/globals.rs
dapplion: Good point, we should revisit after impl validator custody. Since this function is for unit tests without gossip, it doesn't matter

EDIT: Added a TODO to the PR body

beacon_node/network/src/sync/tests/lookups.rs
```diff
@@ -355,29 +370,63 @@ impl TestRig {
         self.expect_empty_network();
     }
 
-    pub fn new_connected_peer(&mut self) -> PeerId {
+    // Don't make pub, use `add_connected_peer_testing_only`
+    fn new_connected_peer(&mut self) -> PeerId {
+        self.add_connected_peer_testing_only(false)
+    }
+
+    // Don't make pub, use `add_connected_peer_testing_only`
```
jimmygchen: What's the reason for this comment?

beacon_node/network/src/sync/tests/lookups.rs
```diff
@@ -355,29 +370,63 @@ impl TestRig {
...
+    fn new_connected_supernode_peer(&mut self) -> PeerId {
+        self.add_connected_peer_testing_only(true)
+    }
+
+    pub fn add_connected_peer_testing_only(&mut self, supernode: bool) -> PeerId {
```
jimmygchen, line 402: `testing_only` seems redundant for a `TestRig` function

beacon_node/network/src/sync/tests/lookups.rs
```diff
@@ -355,29 +370,63 @@ impl TestRig {
...
+        let mut peer_custody_subnets = self
+            .network_globals
+            .peers
+            .read()
+            .peer_info(&peer_id)
+            .expect("peer was just added")
+            .custody_subnets_iter()
+            .map(|subnet| **subnet)
+            .collect::<Vec<_>>();
+        peer_custody_subnets.sort_unstable();
```
jimmygchen, line 418: feel like sorting may be useful for debugging in prod? should we just sort them after computing here https://github.com/sigp/lighthouse/blob/a497ec601cae969d363bd1575a0539ba0491203f/beacon_node/lighthouse_network/src/peer_manager/mod.rs#L746-L762

beacon_node/network/src/sync/tests/lookups.rs
```diff
@@ -355,29 +370,63 @@ impl TestRig {
...
         peer_id
     }
 
-    pub fn new_connected_supernode_peer(&mut self) -> PeerId {
-        let key = self.determinstic_key();
-        self.network_globals
-            .peers
-            .write()
-            .__add_connected_peer_testing_only(true, &self.harness.spec, key)
+    pub fn add_sync_peer(&mut self, supernode: bool, remote_info: SyncInfo) -> PeerId {
```
jimmygchen, line 426: I'm a little confused with all the add peer functions here, is it possible to consolidate some of them? 

- `add_connected_peer`: adds a peer with connected status (but no status yet)
- `add_connected_peer_with_status`: adds a connected peer and also send the `SyncInfo` to sync 

If not, I think some comments will help with readability here.

beacon_node/network/src/sync/tests/lookups.rs
```diff
@@ -1088,6 +1150,21 @@ impl TestRig {
         }
     }
 
+    pub fn expect_no_penalty_for_anyone(&mut self) {
+        self.drain_network_rx();
+        let downscore_events = self
+            .network_rx_queue
```
jimmygchen: Could re-use `filter_received_network_events`

beacon_node/network/src/sync/tests/lookups.rs
```diff
@@ -840,6 +889,19 @@ impl TestRig {
         }
     }
 
+    // Find, not pop
```
jimmygchen:
```suggestion
/// Similar to `pop_received_network_events` but finds matching events without removing them.
```

beacon_node/network/src/sync/tests/range.rs
```diff
@@ -34,10 +40,43 @@ pub(crate) enum DataSidecars<E: EthSpec> {
...
+    ),
 }
 
+impl ByRangeDataRequestIds {
+    fn peer(&self) -> PeerId {
+        match self {
+            Self::PreDeneb => panic!("no requests PreDeneb"),
+            Self::PrePeerDAS(_, peer, _) => *peer,
+            Self::PostPeerDAS(reqs) => {
+                if reqs.len() != 1 {
```
jimmygchen, line 57: Why exactly 1 here?

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -87,9 +87,11 @@ pub struct SyncingChain<T: BeaconChainTypes> {
...
 
     /// The peers that agree on the `target_head_slot` and `target_head_root` as a canonical chain
-    /// and thus available to download this chain from, as well as the batches we are currently
-    /// requesting.
-    peers: HashSet<PeerId>,
+    /// and thus available to download this chain from.
+    ///
+    /// Also, For each peer tracks the total requests done per peer as part of this SyncingChain
+    /// `HashMap<peer, total_requests_per_peer>`
+    peers: HashMap<PeerId, usize>,
```
jimmygchen: maybe rename this to something like `peers_to_requests_count` now the type has changed?

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -121,7 +123,40 @@ pub enum ChainSyncingState {
...
 
+#[cfg(test)]
+#[derive(Debug, Eq, PartialEq)]
+pub enum BatchStateSummary {
+    Downloading,
+    Processing,
+    AwaitingProcessing,
+    AwaitingValidation,
+    Unexpected(&'static str),
+}
```
jimmygchen: Why can't we reuse `BatchState`?

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -221,6 +256,12 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
         request_id: Id,
         blocks: Vec<RpcBlock<T::EthSpec>>,
     ) -> ProcessingResult {
+        // Account for one more requests to this peer
+        // TODO(das): this code assumes that we do a single request per peer per RpcBlock
```
jimmygchen: This would be true in the below case:

```
peerA: block
peerB: col_1, col_2
peerC: col_2, col_3,
```

we get 1 req for each peer.

BUT for this below scenario:

```
peerA: block, col_5
peerB: col_1, col_2
peerC: col_2, col_3,
```

we still get 1 for each peer, which isn't fully correct.

Or we could just use `BatchPeers` and count block peer separately and correctly.

Either way I don't think it makes a big difference - I think we can get rid of this `TODO` and document the assumptions.

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -880,12 +914,13 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
...
                 error = ?err,
-                %peer_id,
                 %request_id,
                 "Batch download error"
             );
             if let BatchOperationOutcome::Failed { blacklist } =
-                batch.download_failed(Some(*peer_id))?
+                // TODO(das): Is it necessary for the batch to track failed peers? Can we make this
+                // mechanism compatible with PeerDAS and before PeerDAS?
+                batch.download_failed(None)?
```
jimmygchen: We still the peer to track failed block peers, and this will break the current peer prioritisation for block and blobs. I think this needs a bit more thought.

jimmygchen:
I've added a few more comments. I've spent quite a bit of time reading but I'm really struggling with reviewing this in the current state, with potential missing pieces and a bunch of outstanding `TODO`s. I find it quite difficult to understand the changes, assumptions, intentions and how the working solution would eventually look like. 

I think it might be useful to go through the plan and code together with @pawanjay176, or re-review this once this is complete. What do you think?

beacon_node/network/src/sync/range_sync/chain.rs
dapplion: Changed to using BatchState with `matches!`

beacon_node/network/src/sync/tests/lookups.rs
dapplion: I agree the add_peer functions are a bit of a mess, but changing them will trigger a big diff on the lookup tests, which use them extensively

beacon_node/network/src/sync/tests/lookups.rs
dapplion: In the code snipped you linked we collect as a HashSet for set_custody_subnets

beacon_node/network/src/sync/tests/lookups.rs
dapplion: Added comments.

I agree the add_peer functions are a bit of a mess, but changing them will trigger a big diff on the lookup tests, which use them extensively

beacon_node/network/src/sync/tests/range.rs
dapplion: Added a comment to the function

```rust
    /// If there's a single active request, returns its peer, else panics
    fn peer(&self) -> PeerId {
```

Where this function is used, I expect a single request. If there are more and we return the first peer it feels a bit random an too loose

jimmygchen:
I've just merged latest `peerdas-devnet-7` into this branch and CI should pass now. Will do another round of review tomorrow.

I've triggerd a CI workflow to run the latest checkpoint sync test on this PR: https://github.com/jimmygchen/lighthouse/actions/runs/15412363885

jimmygchen:
Looks like it's struggling with backfill - below link is a test run on sepolia - both nodes transitioned to `Synced` without completing backfill. The logs are also available on the CI workflow:
https://github.com/jimmygchen/lighthouse/actions/runs/15412363885/job/43368155087

On PeerDAS, supernode struggles to sync to head, but fullnode did complete backfilling 1000 slots in 506 seconds:
https://github.com/jimmygchen/lighthouse/actions/runs/15412363885/job/43368155070

Logs are available on the workflow summary page: https://github.com/jimmygchen/lighthouse/actions/runs/15412363885

jimmygchen:
@dapplion Would you mind taking a look at the failing CI tests in the PR?

beacon_node/network/src/sync/manager.rs
dapplion: Added a note in the PR body

beacon_node/network/src/sync/range_sync/chain.rs
dapplion: After discussing with @pawanjay176 we decided to not track download failures. We rely on randomness to select a different peer as we expect the syncing chains to have a decent number of peers. For the custody by range requests there's an internal failed_peers set that will de-prioritize peers with network errors. The failed_peers at the batch level is still used to track peers that sent blocks or columns that failed processing and ensure that we retry from a different peer

pawanjay176:
Not sure if this is ready for review? Its crashing on mainnet with 

```
Jun 16 10:40:58.395 CRIT  Task panic. This is a bug!                    location: "/Users/pawan/ethereum/lighthouse/beacon_node/network/src/sync/network_context/block_components_by_range.
rs:461:18", msg_id: "TODO: don't do matching here: MissingBlobs", backtrace:    0: std::backtrace::Backtrace::create
```

dapplion:
- Superseded by https://github.com/sigp/lighthouse/issues/7678

Issue #7467, opened 2025-05-15
Support blob parameter forks (BPO) and replace `MAX_BLOBS_PER_BLOCK_FULU`


jimmygchen:
See spec PR on updated blob schedule format:
https://github.com/ethereum/consensus-specs/pull/4277

beacon API spec:
https://github.com/ethereum/beacon-APIs/pull/529

This will be required for `fulu-devnet-0` targeted for 27 May. @ethDreamer is already working on this.

PR #7521, opened 2025-05-26
Implement EIP-7892 BPO hardforks

ethDreamer:
## Issue Addressed

[EIP-7892: Blob Parameter Only Hardforks](https://eips.ethereum.org/EIPS/eip-7892)

#7467


consensus/types/src/chain_spec.rs
```diff
@@ -653,19 +653,48 @@ impl ChainSpec {
...
-    /// Return the value of `MAX_BLOBS_PER_BLOCK` appropriate for the fork at `epoch`.
+    /// Return the value of `MAX_BLOBS_PER_BLOCK` for the given `epoch`.
+    /// NOTE: this function is *technically* not spec compliant, but
+    /// I'm told this is what the other clients are doing for `devnet-0`..
     pub fn max_blobs_per_block(&self, epoch: Epoch) -> u64 {
-        self.max_blobs_per_block_by_fork(self.fork_name_at_epoch(epoch))
+        match self.fulu_fork_epoch {
+            Some(fulu_epoch) if epoch >= fulu_epoch => {
+                let mut max_blobs_per_block = self.max_blobs_per_block_electra;
+                let mut blob_schedule = self.blob_schedule.clone();
```
pawanjay176: why not sort the chainspec blob_schedule at construction from the config instead of sorting it everytime on function call?

consensus/types/src/chain_spec.rs
```diff
@@ -653,19 +653,48 @@ impl ChainSpec {
...
+                blob_schedule.sort_by_key(|entry| entry.epoch);
+                for entry in blob_schedule {
```
pawanjay176: Maybe we can do something like
```rust
    let index = self.blob_schedule.partition_point(|entry| entry.epoch <= epoch);
    if index > 0 {
        max_blobs = self.blob_schedule[index - 1].max_blobs_per_block;
    }
```

pawanjay176 (approved):
LGTM, just minor nits. 
I think we should merge this asap for the devnet, given the spec is a bit in flux, I think its okay to merge this with the TODOs and iterate as the spec gets clearer.

consensus/types/src/chain_spec.rs
```diff
@@ -653,19 +653,48 @@ impl ChainSpec {
...
     }
 
-    /// Return the value of `MAX_BLOBS_PER_BLOCK` appropriate for `fork`.
-    pub fn max_blobs_per_block_by_fork(&self, fork_name: ForkName) -> u64 {
-        if fork_name.fulu_enabled() {
-            self.max_blobs_per_block_fulu
-        } else if fork_name.electra_enabled() {
-            self.max_blobs_per_block_electra
+    // TODO(EIP-7892): remove this once we have fork-version changes on BPO forks
+    pub fn max_blobs_per_block_within_fork(&self, fork_name: ForkName) -> u64 {
```
jimmygchen, line 673: I think it's worth adding some unit tests for these functions.

jimmygchen: ah nevermind - just read the comment above

pawanjay176:
Tested this with the interop config in https://notes.ethereum.org/@ethpandaops/fusaka-devnet-0#Kurtosis-Interop-Conifg-Pre-devnet-testing

Let it run for 3 epochs which covered an increase and decrease in the blob count and it runs perfectly with the other clients. 🎉

consensus/types/src/chain_spec.rs
ethDreamer: Added some tests. This function isn't complex right now. I plan to get rid of it. But if it stuck around for multiple forks it would be more complex.

consensus/types/src/chain_spec.rs
ethDreamer: My concern was that the spec was constructed in multiple places and I want to ensure the blob schedule was always sorted. I've done that now by adding a wrapper `BlobSchedule` type. Let me know what you think :)

consensus/types/src/chain_spec.rs
ethDreamer: I've updated the function to be implemented on the `BlobSchedule`. It's simpler now. Let me know what you think :)

pawanjay176 (approved):
LGTM, the CI failures seem to be unrelated to this PR
cc @jimmygchen

jimmygchen:
Nice one! thanks 🎉

jimmygchen:
> LGTM, the CI failures seem to be unrelated to this PR cc @jimmygchen

Indeed, the RPC one has been fixed in #7522.
I created a PR for the `AvailabilityCheckError::Unexpected` failure here #7533.

jimmygchen:
@mergify requeue

jimmygchen:
> Indeed, the RPC one has been fixed in #7522. I created a PR for the `AvailabilityCheckError::Unexpected` failure here #7533.

Ah it was actually not that (even though it revealed a real bug that triggers the same failure) - the test is actually failing 100% now in its current state, because the failing test expects a blob, but the random block generator didn't create a block with blobs.

jimmygchen:
> > Indeed, the RPC one has been fixed in #7522. I created a PR for the `AvailabilityCheckError::Unexpected` failure here #7533.
> 
> Ah it was actually not that (even though it revealed a real bug that triggers the same failure) - the test is actually failing 100% now in its current state, because the failing test expects a blob, but the random block generator didn't create a block with blobs.

@ethDreamer I've added a setter method to the execution block generator to allow modifying blob generation behaviour on the fly.

Basically the test is failing because it expects blobs, but there's no blobs generated for the next block (passing before due to `rng`), and passing empty `Vec<DataColumnSidecar>` to DA checker would return an error. The fix ensures `rig.next_blobs` or `rig.next_data_column` is always non empty.

michaelsproul (approved):

jimmygchen:
@mergifyio requeue

Issue #7466, opened 2025-05-15
Flaky test in `release-tests-windows` CI job


jimmygchen:
https://github.com/sigp/lighthouse/actions/runs/15044891932/job/42284911795

This test appears to be flaky:
> ──── STDERR:             lighthouse_network::rpc_tests test_delayed_rpc_response
> 
> thread 'test_delayed_rpc_response' panicked at beacon_node\lighthouse_network\tests\rpc_tests.rs:1229:33:
> assertion failed: request_sent_at.elapsed() >
    (Duration::from_secs(QUOTA_SEC) - Duration::from_millis(100))
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

Here is the failing assertion:
https://github.com/sigp/lighthouse/blob/1324d3d3c4c20914545f6dadcb018a5b442a95a5/beacon_node/lighthouse_network/tests/rpc_tests.rs#L1226-L1233

Although it looks like it already has a 100ms margin added, which should probably be enough? Might need @ackintosh's help to figure out the right fix.

jimmygchen:
Hi @ackintosh, would you mind taking a look at this one?
I'm not sure if adding more margin to the assertion is the right fix, just wanted to check with you if things are working correctly.

ackintosh:
Stdout logging for the test has been enabled via https://github.com/sigp/lighthouse/pull/7506. The sequence diagram below shows what happened during the [test run](https://github.com/sigp/lighthouse/actions/runs/15198128212/job/42746928204?pr=7025) in which `test_delayed_rpc_response` failed.

<img width="1096" alt="Image" src="https://github.com/user-attachments/assets/7c6c8539-17e8-46d0-9dd8-ba5ee2340e27" />

The intent of `test_delayed_rpc_response` is to check that the rate-limiter is working on the receiver side in an end-to-end fashion. I think expanding the margin, to 500ms for example, would be a good idea.

jimmygchen:
Fixed in #7522, thanks @ackintosh and @diegomrsantos for help with investigating! 🎉

PR #7522, opened 2025-05-26
Mitigate flakiness in test_delayed_rpc_response

ackintosh:
## Issue Addressed

https://github.com/sigp/lighthouse/issues/7466

## Proposed Changes

Expanded the margin from 100ms to 500ms.

### DIFF
--- a/beacon_node/lighthouse_network/tests/rpc_tests.rs
+++ b/beacon_node/lighthouse_network/tests/rpc_tests.rs
@@ -1246,10 +1246,12 @@ fn test_delayed_rpc_response() {
                                 // The second and subsequent responses are delayed due to the response rate-limiter on the receiver side.
                                 // Adding a slight margin to the elapsed time check to account for potential timing issues caused by system
                                 // scheduling or execution delays during testing.
+                                // https://github.com/sigp/lighthouse/issues/7466
+                                let margin = 500;
                                 assert!(
                                     request_sent_at.elapsed()
                                         > (Duration::from_secs(QUOTA_SEC)
-                                            - Duration::from_millis(100))
+                                            - Duration::from_millis(margin))
                                 );
                                 if request_id == 5 {
                                     // End the test


ackintosh:
A different test failed.

https://github.com/sigp/lighthouse/actions/runs/15262744452/job/42923339626?pr=7522
>    thread 'network_beacon_processor::tests::test_rpc_block_reprocessing' panicked at beacon_node/network/src/network_beacon_processor/tests.rs:1259:5:
>    assertion `left == right` failed
>      left: 0x2b95a01b37ad3685e8045426fc4abb4b7df438f27c98ee5b6c007972bcae4d33
>     right: 0x9864352b6d2d5b4f40d754fda23c9d35e0efa45c850d6f55a40cdb4a1f1785d1

jimmygchen:
> A different test failed.
> 
> https://github.com/sigp/lighthouse/actions/runs/15262744452/job/42923339626?pr=7522
> 
> > thread 'network_beacon_processor::tests::test_rpc_block_reprocessing' panicked at beacon_node/network/src/network_beacon_processor/tests.rs:1259:5:
> > assertion `left == right` failed
> > left: 0x2b95a01b37ad3685e8045426fc4abb4b7df438f27c98ee5b6c007972bcae4d33
> > right: 0x9864352b6d2d5b4f40d754fda23c9d35e0efa45c850d6f55a40cdb4a1f1785d1

That looks like an existing flaky test, i think these real-time based tests can easily be flaky
https://github.com/sigp/lighthouse/blob/70194dfc6a3f4d10c9059610f889ff5a4e863a6a/beacon_node/network/src/network_beacon_processor/tests.rs#L1074-L1078

I'm wondering if the block processing is taking longer than 10ms here if the github runner is overwhelmed? Ideally we avoid having realtime-based tests, but it doesn't look like we can easily do that here. I think a retry-based approach is generally better than a single sleep - less flaky and avoid adding a longer sleep that potentially slows down the test.

e.g. 
```rust
    let max_retries = 3;
    let success = false;
    for _ in 0..max_retries {
        // Add an extra delay for block processing
        tokio::time::sleep(Duration::from_millis(10)).await;
        // head should update to next block now since the duplicate
        // cache handle was dropped.
        if next_block_root == rig.head_root() {
            success = true;
            break;
        }
    }
    assert!(
        success,
        "expected head_root to be {:?} but was {:?}",
        next_block_root,
        rig.head_root()
    );
```

jimmygchen:
@ackintosh I'm inclined to merge this as we're seeing it quite regularly on CI, we can address the other flaky tests separately. Thanks!

jimmygchen (approved):

ackintosh:
Thanks, Jimmy! I will create a separate PR to address the other flaky test.

PR #7420, opened 2025-05-08
Add `console-subscriber` feature for debugging

jimmygchen:
## Issue Addressed

Add `console-subscriber` feature for debugging tokio async tasks.

Usage:
- Build Lighthouse with `RUSTFLAGS=--cfg tokio_unstable` and `--features console-subscriber`, e.g.:
```
RUSTFLAGS=-"-cfg=tokio_unstable --remap-path-prefix=$(pwd)=." FEATURES=console-subscriber make
```
- Run the Lighthouse binary.
- Install `tokio-console` and run it in a terminal.

PR #7529, opened 2025-05-27
Add `console-subscriber` feature for debugging

jimmygchen:
## Issue Addressed

Add `console-subscriber` feature for debugging tokio async tasks.

Supersedes #7420 to work with `unstable`.

Usage:
- Build Lighthouse with `RUSTFLAGS=--cfg tokio_unstable` and `--features console-subscriber`, e.g.:
```
RUSTFLAGS=-"-cfg=tokio_unstable --remap-path-prefix=$(pwd)=." FEATURES=console-subscriber make
```
- Run the Lighthouse binary.
- Install `tokio-console` and run it in a terminal.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -613,6 +613,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-stream"
+version = "0.3.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


macladson (approved):
Code looks good to me! Not really sure how useful this will be exactly, but we may as well have it, especially since it's feature-gated.

jimmygchen:
Thanks @macladson ! 
I think this would be quite useful in seeing what's going on in the tokio runtime if something gets stuck or if taking a long time - we'd be able to see what task are running and what resources are being held.

PR #7530, opened 2025-05-28
Add e2e sync tests to CI

jimmygchen:
## Issue Addressed

This PR adds the following sync tests to CI workflow - triggered when a PR is labeled `syncing` - to ensure we have some e2e coverage on basic sync scenarios:
- [x] checkpoint sync to a live network (covers range and backfill sync for _current_ fork)
- [x] checkpoint sync to a running devnet (covers range and backfill sync for _next_ fork)

It seems to work fine running on github hosted runners - but if performance become an issue we could switch to using self hosted runners for sepolia sync test. (standard CPU runners have 4 CPU, 16 GB ram - i think it _should_ be enough on sepolia / devnet networks)

The following tests have been **removed** from this PR and moved to a separate issue *(#7550)
- [x] genesis sync on a local devnet (covers current and next fork)
  - [x] brief shutdown and restart (covers lookup sync)
  - [x] longer shutdown and restart (covers range sync)

I'm hoping to keep these e2e test maintenance effort to a minimum - hopefully longer term we could have some generic e2e tests that works for all clients and the maintenance effort can be spread across teams.

### Latest test run: 
https://github.com/sigp/lighthouse/actions/runs/15411744248

### Results:
image
image

#### logs are available as artifacts:
image


.github/workflows/local-testnet.yml
```diff
@@ -161,4 +199,4 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Check that success job is dependent on all others
-        run: ./scripts/ci/check-success-job.sh ./.github/workflows/local-testnet.yml local-testnet-success
+        run: ./scripts/ci/check-success-job.sh ./.github/workflows/local-testnet.yml local-testnet-success checkpoint-sync-test
```
jimmygchen: Added to ensure this optional job is excluded from CI requirement.

scripts/tests/checkpoint-sync-config.yaml
```diff
@@ -0,0 +1,13 @@
+participants:
+  - cl_type: lighthouse
+    cl_image: lighthouse:local
+    supernode: true
+  - cl_type: lighthouse
+    cl_image: lighthouse:local
+    supernode: false
+
+checkpoint_sync_enabled: true
+checkpoint_sync_url: "https://checkpoint-sync.sepolia.ethpandaops.io"
```
jimmygchen: Can't use hoodi because the checkpoint syncing to it takes too long. There's a 2m timeout in Kurtosis (not configurable).

.github/workflows/local-testnet.yml
```diff
@@ -146,6 +146,44 @@ jobs:
...
+        run: |
+          ./checkpoint-sync.sh
+        working-directory: scripts/tests
+
+      - name: Upload logs artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: logs-checkpoint-sync
+          path: |
+            scripts/local_testnet/logs
```
jimmygchen, line 213: logs aren't there - need to use `[start|stop]_local_testnet.sh` instead of running with kurtosis directly.

jimmygchen: Addressed in https://github.com/sigp/lighthouse/pull/7530/commits/547eac0c2bb91d99d081dabdac75697cf66c4c82

jimmygchen:
I think this PR is ready for review. The failing jobs due to the following:
1. **devnet checkpoint sync**: requires BPO to be merged to checkpoint sync to `fusaka-devnet-0`
2. **genesis-sync-test-***: there are some known assertoor failures that ethpandaops are currently investigating.

It's good that the test is already catching the 1st scenario above. The two issues should be addressed soon outside of this PR.

.github/workflows/local-testnet.yml
```diff
@@ -146,6 +146,88 @@ jobs:
...
+      matrix:
+        fork: [electra, fulu]
+        offline_secs: [ 120, 300]
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Checkout kurtosis-sync-test
+        uses: actions/checkout@v4
+        with:
+          repository: jimmygchen/kurtosis-sync-test
```
jimmygchen: I made some fixes on my fork. Upstream PR: https://github.com/ethpandaops/kurtosis-sync-test/pull/5

jimmygchen:
@chong-he I'm finding the checkpoint sync script quite useful in testing other PRs. As discussed earlier, we'd like to replace the genesis sync test with our own script - let's do that in a separate PR so we can merge this soon.

I'll remove the genesis sync tests from this PR.

scripts/tests/checkpoint-sync-config-devnet.yaml
```diff
@@ -0,0 +1,16 @@
...
+participants:
+  - cl_type: lighthouse
+    cl_image: lighthouse:local
+    supernode: true
+  - cl_type: lighthouse
+    cl_image: lighthouse:local
+    supernode: false
+
+checkpoint_sync_enabled: true
+checkpoint_sync_url: "https://checkpoint-sync.fusaka-devnet-0.ethpandaops.io"
```
michaelsproul, line 11: If we don't want to run the devnet tests, we can just leave off the `syncing` label altogether right? I'm just thinking of the case where we want to merge devnet-1 changes prior to devnet-1 existing

michaelsproul (approved):
LGTM, just one nit

scripts/tests/checkpoint-sync-config-devnet.yaml
jimmygchen: Yes these tests don't run by default and not required for merge - although we could make sepolia one mandatory.

Yeah i think in that case its probably ok for it to fail, and _i think_ mergify won't block it (i've updated the `local-testnet-success` job).

I've made a feature request (https://github.com/ethpandaops/kurtosis-sync-test/issues/4) to test checkpoint sync with a local testnet here, so longer term we could potentially getting rid of the devnet test here.

michaelsproul: Ok yeah sounds good

PR #7533, opened 2025-05-29
Prevent `AvailabilityCheckError` when there's no new custody columns to import

jimmygchen:
## Issue Addressed

Addresses a regression recently introduced when we started gossip verifying data columns from EL blobs

```
    failures:
        network_beacon_processor::tests::accept_processed_gossip_data_columns_without_import

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 90 filtered out; finished in 16.60s
    
  stderr ───

    thread 'network_beacon_processor::tests::accept_processed_gossip_data_columns_without_import' panicked at beacon_node/network/src/network_beacon_processor/tests.rs:829:10:
    should put data columns into availability cache: Unexpected("empty columns")
 ```

https://github.com/sigp/lighthouse/actions/runs/15309278812/job/43082341868?pr=7521

If an empty `Vec` is passed to the DA checker, it causes an unexpected error.

This PR addresses it by not passing an empty `Vec` for processing, and not spawning a task to publish.

### DIFF
--- a/beacon_node/beacon_chain/src/fetch_blobs/mod.rs
+++ b/beacon_node/beacon_chain/src/fetch_blobs/mod.rs
@@ -264,6 +264,7 @@ async fn fetch_and_process_blobs_v2<T: BeaconChainTypes>(
         return Ok(None);
     }
 
+    debug!(num_fetched_blobs, "All expected blobs received from the EL");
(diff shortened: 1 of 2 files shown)


michaelsproul (approved):
Nice fix, and nice test too!

PR #7540, opened 2025-05-29
Fulu EF tests v1.6.0-alpha.0

ethDreamer:
## Issue Addressed

Update to EF tests v1.6.0-alpha.0


jimmygchen:
Thanks @ethDreamer, I'll pick it up from here!

michaelsproul:
There's something weird going on when I test locally where some Capella and Deneb test files are detected as unused. Not sure what is going on there, and I need to run now.

> tests/mainnet/deneb/epoch_processing/inactivity_updates/pyspec_tests/random_inactivity_scores_empty_participation/pre_epoch.ssz_snappy

Ohhh, this might be a change in the test format? And we're not reading those files?

michaelsproul (approved):
LGTM ;)

consensus/types/src/runtime_var_list.rs
```diff
@@ -241,6 +242,62 @@ where
...
+
+// We can delete this once the upstream `vec_tree_hash_root` is modified to use a runtime max len.
+pub fn runtime_vec_tree_hash_root<T>(vec: &[T], max_len: usize) -> Hash256
+where
+    T: TreeHash,
+{
+    match T::tree_hash_type() {
+        TreeHashType::Basic => {
+            let mut hasher =
+                MerkleHasher::with_leaves(max_len.div_ceil(T::tree_hash_packing_factor()));
```
jimmygchen, line 273: I guess this is covered by the spec tests in LH - but we should add a unit test if we move this into the ssz library.

michaelsproul: Yeah agreed. I'll work on a PR for that

testing/ef_tests/src/handler.rs
```diff
@@ -21,15 +22,18 @@ pub trait Handler {
...
         Self::Case::is_enabled_for_fork(fork_name)
     }
 
-    fn is_enabled_for_feature(&self, feature_name: FeatureName) -> bool {
-        Self::Case::is_enabled_for_feature(feature_name)
+    fn is_enabled_for_feature(&self, _feature_name: FeatureName) -> bool {
+        // Feature gating is used for future forks that haven't been incorporated into a main fork yet.
+        // When a feature is ready to be part of a main fork, it should be removed from here and
+        // the tests should run normally under that fork.
+        false
```
jimmygchen: Is it necessary for this to return false instead of calling `Self::Case::is_enabled_for_feature(feature_name)`? (which is false by default)

testing/ef_tests/src/handler.rs
```diff
@@ -971,8 +1032,11 @@ impl<E: EthSpec> Handler for KZGVerifyCellKZGProofBatchHandler<E> {
         "verify_cell_kzg_proof_batch".into()
     }
 
-    fn is_enabled_for_feature(&self, feature_name: FeatureName) -> bool {
-        feature_name == FeatureName::Fulu
+    fn is_enabled_for_feature(&self, _feature_name: FeatureName) -> bool {
```
jimmygchen: I think we can remove these impls for the KZG handlers as the default impl returns `false`

testing/ef_tests/src/handler.rs
michaelsproul: Yeah I was wondering this, don't think it's necessary. I'll revert back to using `Self::Case::is_enabled_for_feature(feature_name)` now

jimmygchen (approved):

jimmygchen:
Looks great, really nice work @ethDreamer @michaelsproul 🥳

testing/ef_tests/src/handler.rs
michaelsproul: Nice catch. Removed in https://github.com/sigp/lighthouse/pull/7540/commits/6a17665b142290dced06db89898bf8ce5ac98ff4

testing/ef_tests/src/handler.rs
michaelsproul: Fixed in https://github.com/sigp/lighthouse/pull/7540/commits/6a17665b142290dced06db89898bf8ce5ac98ff4

jimmygchen (approved):
LGTM 🎉

PR #7544, opened 2025-06-02
Add a name to the Tokio task

ackintosh:
## Issue Addressed

The `console-subscriber` feature was added in https://github.com/sigp/lighthouse/pull/7529. However, the names of the running tasks are blank:

image

## Proposed Changes

Set the task name using `tokio::task::Builder`, which is availble when the `tokio_unstable` is enabled.

image

## Additional Info

As you can see in the screenshot, there are many tasks named as `libp2p`. I think it would be more informative if those tasks were named after their Behaviour name instead of just `libp2p` (e.g., `gossipsub`). But that would require changes to rust-libp2p.


jimmygchen (approved):
This looks good to me! Definitely make the output a lot useful, nice work!

PR #7518, opened 2025-05-26
Update `staking-deposit-cli` to `ethstaker-deposit-cli`

chong-he:
Based on the feedback on [Discord](https://discord.com/channels/605577013327167508/605577013331361793/1375339231970328657), we should update `staking-deposit-cli` to `ethstaker-deposit-cli` as the former has not been updated for a while (e.g., not supporting Hoodi). 

Ethstaker is now maintaining the deposit cli: https://www.reddit.com/r/ethstaker/comments/1ecuj5q/the_depositcli_has_long_needed_some_critical/

~There are some files like `validator_manager/src/common.rs` and `validator_manager/test_vectors/generate.py` still~ ~refering to `staking-deposit-cli`.~

~The `common.rs` is referring to this commit:~
~https://github.com/ethereum/staking-deposit-cli/blob/76ed78224fdfe3daca788d12442b3d1a37978296/~staking_deposit/credentials.py#L131-L144~
~which is changed in `ethstaker-deposit-cli`:~
~https://github.com/eth-educators/ethstaker-deposit-cli/blob/main/ethstaker_deposit/credentials.py~

~So I leave it untouched for now~
Edit: updated all to `ethstaker-deposit-cli` 

Other updates:
- added a note on Homebrew page as Lighthouse on Homebrew hasn't been updated since v7.0.0. [Review here](https://github.com/sigp/lighthouse/pull/7518/files#diff-fd14e06257fea0886d9281f1d8e833d40460103bb4406baf77ce2cd29ed2a520R8)
- added `--vc-url` in validator manager doc as per: https://discord.com/channels/605577013327167508/605577013331361793/1377638933579436032. [Review here](https://github.com/sigp/lighthouse/pull/7518/files#diff-0467838cad89c67f549e4e4e6b8c8c180eafb6b648b1e40001bb3c692c8174dcR3)

PR #7548, opened 2025-06-03
Update key generation in validator manager

chong-he:
## Issue Addressed

#7518 breaks the key generation process in `validator_manager/test_vectors/generate.py` after updating to `ethstaker-deposit-cli`. This PR updates the key generation process, tested and successfully generated the deposit data JSON files.


michaelsproul (approved):
Nice!

Issue #6837, opened 2025-01-22
Download checkpoint data columns from p2p peers


jimmygchen:
## Description

Our current implementation in PeerDAS (https://github.com/sigp/lighthouse/pull/6760) requires the checkpoint server to serve checkpoint blobs via the `beacon/blob_sidecars` HTTP endpoint. 

Once PeerDAS is activated, a full node will no longer be able to serve this endpoint as it won't store full blobs but instead store data columns (samples), and it won't store enough to be able to serve blobs (require >= 50% of all columns to reconstruct blob sidecars) - this means the checkpoint server will need to run a supernode or custody at least 50% of all data columns.

This approach is currently the simplest solution for testing purpose, as the alternative of retrieving columns from p2p peers requires breaking an existing invariant of having all blobs/columns for blocks. However, longer term it would be beneficial to implement the latter, and potentially before PeerDAS goes live.

See this comment for more details: https://github.com/sigp/lighthouse/pull/6760#issuecomment-2585970322

jimmygchen:
I had a brief conversation with @michaelsproul about this, we think the easiest approach would be just checkpoint sync without the columns, then reimport the block with new columns (potentially using the changes from validator custody).

~~Validator custody is still WIP, so we could just wait for that before we implement backfilling the checkpoint block. If it's easy to implement once validator custody is done then we may as well implement it, but even without it, missing columns for a single block shouldn't break things too badly.~~

michaelsproul:
Just noting that we still attempt to download blobs from the checkpoint sync URL in v8.0.0, here:

https://github.com/sigp/lighthouse/blob/e3ee7febce64c1b5a85c3ab0be0619571ee92d58/beacon_node/client/src/builder.rs#L412-L418

The change in this PR did not apply to the checkpoint sync URL code:

- https://github.com/sigp/lighthouse/pull/7549

PR #7549, opened 2025-06-03
Checkpoint sync without the blobs from Fulu

jimmygchen:
## Issue Addressed

Lighthouse currently requires checkpoint sync to be performed against a supernode in a PeerDAS network, as only supernodes can serve blobs.

This PR lifts that requirement, enabling Lighthouse to checkpoint sync from either a fullnode or a supernode (See https://github.com/sigp/lighthouse/issues/6837#issuecomment-2933094923)

Missing data columns for the checkpoint block isn't a big issue, but we should be able to easily implement backfill once we have the logic to backfill data columns.

### DIFF
--- a/beacon_node/beacon_chain/src/builder.rs
+++ b/beacon_node/beacon_chain/src/builder.rs
@@ -486,24 +486,36 @@ where
 
         // Verify that blobs (if provided) match the block.
         if let Some(blobs) = &weak_subj_blobs {
-            let commitments = weak_subj_block
-                .message()
-                .body()
-                .blob_kzg_commitments()
-                .map_err(|e| format!("Blobs provided but block does not reference them: {e:?}"))?;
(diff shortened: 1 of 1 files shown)


jimmygchen:
Started a checkpoint sync test run here (with the script from #7530)
https://github.com/jimmygchen/lighthouse/actions/runs/15421369524

eserilev (approved):
LGTM

backfill sync looks pretty slow, but that seems out of scope for this PR

jimmygchen:
> LGTM
> 
> backfill sync looks pretty slow, but that seems out of scope for this PR

Yeah it's not very consistent either - especially for supernodes. Lion is making some big changes to improve the peerdas sync, so can ignore the devnet tests now.

Sepolia checkpoint sync seems to take around 2-3 mins (for backfilling 1000 slots) quite consistently on these GitHub runners.

PR #7551, opened 2025-06-03
Do not compute sync selection proofs for the sync duty at the current slot

dknopik:
As aggregation has already been performed in the last third of the previous slot, we no longer need the selection proof for the duty at the current slot.

See also the discussion at https://github.com/sigp/lighthouse/pull/7016#discussion_r2113193333

CC @michaelsproul

### DIFF
--- a/validator_client/validator_services/src/sync.rs
+++ b/validator_client/validator_services/src/sync.rs
@@ -512,8 +512,12 @@ pub async fn fill_in_aggregation_proofs<S: ValidatorStore, T: SlotClock + 'stati
         "Calculating sync selection proofs"
     );
 
+    // Start at the next slot, as aggregation proofs for the duty at the current slot are no longer
+    // required since we do the actual aggregation in the slot before the duty slot.
+    let start_slot = current_slot.as_u64() + 1;
+
     // Generate selection proofs for each validator at each slot, one slot at a time.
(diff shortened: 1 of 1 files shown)


dknopik:
@chong-he 
Opposed to my comment in linked PR, I decided to not also add `1` to `pre_compute_slot` as in `prepare_for_aggregator_pre_compute` we take the `last_slot_of_period` as upper bound. Adding `1` to that causes another off-by-one, as that slot would then belong to the next period.

dknopik:
I'll also try to add a test _somewhere_

dknopik:
@michaelsproul 
Regarding testing this:

We already have `verify_full_sync_aggregates_up_to` in the simulator. Simulation is done for 16 epochs, and a sync committee period is 8 epochs, so we are testing this. Introducing an off-by-one intentionally by changing 

```rust
    for slot in (start_slot..=pre_compute_slot.as_u64()).map(Slot::new) {
```

to 

```rust
    for slot in (start_slot..pre_compute_slot.as_u64()).map(Slot::new) {
```

causes an error as expected: `called `Result::unwrap()` on an `Err` value: "Sync aggregate at slot 17 was not full, got: 0, expected: 32"`

Therefore I do not think we need another test explicitly for this.

michaelsproul (approved):

michaelsproul:
@mergify requeue

michaelsproul:
@mergify requeue

Issue #7545, opened 2025-06-03
Optimise verification on GetBlobsV2 response


jimmygchen:
We've recently introduced gossip verification to `getBlobsV2` response (#7493) to make sure
1. The proofs from the EL are validated
2. We don't re-process already imported data columns (`GossipDataColumnError::PriorKnown` and `GossipDataColumnError::PriorKnownUnpublished`)

We currently do this by performing full gossip verification here:
https://github.com/jimmygchen/lighthouse/blob/e33dadbf1a520e92ddec614f3b7133fcc9e4c572/beacon_node/beacon_chain/src/fetch_blobs/mod.rs#L340-L352

The full verification includes relatively expensive checks like signature verification, which isn't necessary since the block headers are already verified and we're just cloning it over to build `DataColumnSidecar`s. This impacts the block attestable time and import time. We could optimise this to have a verification mode that skips all validation on the `SignedBeaconBlockHeader` and still do the rest.

eserilev:
I'd like to work on this if no one else is already on it

jimmygchen:
Nice, thanks @eserilev

jimmygchen:
Completed in #7582

PR #7553, opened 2025-06-03
Optimize GetBlobsV2 Verification

eserilev:
## Issue Addressed

#7545

## Proposed Changes

Add a verification mode that skips block proposer and block signature checks.


beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -2201,9 +2203,10 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
     ) -> Result<GossipVerifiedDataColumn<T>, GossipDataColumnError> {
         metrics::inc_counter(&metrics::DATA_COLUMN_SIDECAR_PROCESSING_REQUESTS);
         let _timer = metrics::start_timer(&metrics::DATA_COLUMN_SIDECAR_GOSSIP_VERIFICATION_TIMES);
-        GossipVerifiedDataColumn::new(data_column_sidecar, subnet_id, self).inspect(|_| {
-            metrics::inc_counter(&metrics::DATA_COLUMN_SIDECAR_PROCESSING_SUCCESSES);
-        })
+        GossipVerifiedDataColumn::new(data_column_sidecar, subnet_id, CheckBlockHeader::No, self)
```
jimmygchen:
```suggestion
        GossipVerifiedDataColumn::new(data_column_sidecar, subnet_id, CheckBlockHeader::Yes, self)
```
We need to check header for gossip columns. We can only skip the header check if the `DataColumnSidecar` is constructed internally by the node from an already verified `SignedBeaconBlockHeader`.

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -199,19 +206,24 @@ impl<T: BeaconChainTypes, O: ObservationStrategy> GossipVerifiedDataColumn<T, O>
...
-                )
-            },
+        validate_data_column_sidecar_for_gossip::<T, O>(
+            column_sidecar,
+            subnet_id,
+            check_block_header,
+            chain,
         )
+        .map_err(|e| {
+            process_block_slash_info::<_, GossipDataColumnError>(
```
jimmygchen: I think `process_block_slash_info` is not necessary if the mode is `CheckBlockHeader::No` because it's also processing an already processed header.

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -486,7 +499,12 @@ pub fn validate_data_column_sidecar_for_gossip<T: BeaconChainTypes, O: Observati
     verify_column_inclusion_proof(&data_column)?;
     let parent_block = verify_parent_block_and_finalized_descendant(data_column.clone(), chain)?;
     verify_slot_higher_than_parent(&parent_block, column_slot)?;
-    verify_proposer_and_signature(&data_column, &parent_block, chain)?;
+
+    match check_block_header {
+        CheckBlockHeader::Yes => verify_proposer_and_signature(&data_column, &parent_block, chain)?,
```
jimmygchen: I think it's worth considering skipping some more checks:
- `verify_parent_block_and_finalized_descendant`: this acquires a fork choice read lock - so it could delay verification when the node is overwhelmed
- `verify_column_inclusion_proof ` isn't necessary for sidecars constructed internally, because we're verifying proofs we just constructed (on the critical path as well!) - although this isn't part of the header, so skipping it when `CheckBlockHeader::No` may be confusing.
- `chain.observed_slashable.write()`: it's likely a quick operation but i think we'd want to avoid acquiring locks unnecessarily - in this case it's observing the data from the header 

I realise I may have not explained the rationale in the issues clearly - the main thing is we can avoid verifying some properties of the `DataColumnSidecar` if the sidecar is _constructed internally_ by the node, e.g. block proposal, fetch blobs etc - and in these case we can skip:
1. the block header that we've already verified, and have just been cloned over to the sidecar
2. the kzg inclusion proof that we've just constructed

But for any `DataColumnSidecar`s received over gossip, we'd want to fully verify them.

For reference, the function to construct `DataColumnSidecar` is below, and and the only untrusted input is `Blobs` and `Proofs`:
https://github.com/sigp/lighthouse/blob/9779b4ba2c045a401a5894bfeedefcc9a0c1bf68/beacon_node/beacon_chain/src/kzg_utils.rs#L166-L172

beacon_node/http_api/src/publish_blocks.rs
```diff
@@ -407,8 +409,12 @@ fn build_gossip_verified_data_columns<T: BeaconChainTypes>(
         .map(|data_column_sidecar| {
             let column_index = data_column_sidecar.index;
             let subnet = DataColumnSubnetId::from_column_index(column_index, &chain.spec);
-            let gossip_verified_column =
-                GossipVerifiedDataColumn::new(data_column_sidecar, subnet.into(), chain);
+            let gossip_verified_column = GossipVerifiedDataColumn::new(
+                data_column_sidecar,
+                subnet.into(),
+                CheckBlockHeader::Yes,
```
jimmygchen: The `DataColumnSidecar` is constructed in the same function internally by the node, so we just do light verification here i believe.

jimmygchen (changes requested):
Nice, this looks good, thanks @eserilev !

I've added some questions regarding if we could skip more checks, let me know what you think! 

@michaelsproul we briefly talked about this before, so would be great to get your input too 🙏

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -483,10 +497,17 @@ pub fn validate_data_column_sidecar_for_gossip<T: BeaconChainTypes, O: Observati
...
+            let parent_block =
+                verify_parent_block_and_finalized_descendant(data_column.clone(), chain)?;
+            verify_slot_higher_than_parent(&parent_block, column_slot)?;
+            verify_proposer_and_signature(&data_column, &parent_block, chain)?;
+        }
+        ConstructedInternally::Yes => (),
+    };
+
     let kzg = &chain.kzg;
     let kzg_verified_data_column = verify_kzg_for_data_column(data_column.clone(), kzg)
```
jimmygchen, line 512: I might be overthinking, but I'm also tempted to include `observe_slashable` (without changing the ordering), because it acquires a write lock, and usually data columns arrives in burst (128 simultaneously), and the write lock can only be acquired by one thread at a time, so this write _could_ potentially slow things down even though it's a very fast operation.
```rust
    chain
        .observed_slashable
        .write()
```

jimmygchen: It might be worth writing some bench tests for gossip verification (in a later PR) - given that 1/ gossip verification time is critical to both the network and the performance of the node; 2/ the volume of verification is significantly higher now with data columns.

jimmygchen: Created issue for benchmark #7576

jimmygchen:
Sorry @eserilev, 
I was soo close to merging this, but Daniel pointed out we can optimise further by batch verifying internally constructed columns (#7582), since batch KZG verification gives us a performance gain. In those cases, we'd prefer to batch verify the KZG proofs, whereas `validate_data_column_sidecar_for_gossip` only handles single data column sidecars. Doing the verification with a separate function also avoids mixing strategies in  `validate_data_column_sidecar_for_gossip` and skewing the gossip validation metrics.

eserilev:
yeah daniels batch kzg verification pr looks super promising! i'll close this PR in favor of that one

PR #7554, opened 2025-06-03
EIP-7917 Deterministic proposer lookahead First Pass

ethDreamer:
## Issue Addressed

* [EIP-7917](https://eips.ethereum.org/EIPS/eip-7917)

This is the first pass, there will need to be some work to re-architect the beacon proposer's cache.


ethDreamer:
The [epoch_processing EF tests](https://github.com/ethereum/consensus-specs/blob/dev/tests/core/pyspec/eth2spec/test/fulu/epoch_processing/test_process_proposer_lookahead.py) aimed at testing `process_proposer_lookahead()` are borked for this release but at least I manually verified that our version gets the expected result.

ethDreamer:
Everything passes except the `ef_tests` because:
1. The `epoch_processing()` test for `process_proposer_lookahead()` is broken
2. We need to write an entirely new harness for `test_effective_balance_increase_changes_lookahead()`

PR #7557, opened 2025-06-04
Support for using existing jwt secret for mock-el

jimmygchen:
## Issue Addressed

This PR adds support for `mock-el` to use an existing jwt secret. This makes it possible to potentially integrate `mock-el` into test tooling like kurtosis `ethereum-package`, which passes an existing jwt key to the EL.


jimmygchen:
It doesn't seem straight forward to add to `ethereum-package`, will close this for the time being.

PR #7559, opened 2025-06-04
Changes for `fusaka-devnet-1`

ethDreamer:
## Issue Addressed

Changes for [fusaka-devnet-1](https://notes.ethereum.org/@ethpandaops/fusaka-devnet-1)

## Proposed Changes

[Consensus Specs v1.6.0-alpha.1](https://github.com/ethereum/consensus-specs/pull/4346)
* [EIP-7917: Deterministic Proposer Lookahead](https://eips.ethereum.org/EIPS/eip-7917)
* [EIP-7892: Blob Parameter Only Hardforks](https://eips.ethereum.org/EIPS/eip-7892)

### DIFF
--- a/consensus/state_processing/src/per_epoch_processing/single_pass.rs
+++ b/consensus/state_processing/src/per_epoch_processing/single_pass.rs
@@ -19,7 +19,7 @@ use types::{
     milhouse::Cow,
     ActivationQueue, BeaconState, BeaconStateError, ChainSpec, Checkpoint, DepositData, Epoch,
     EthSpec, ExitCache, ForkName, List, ParticipationFlags, PendingDeposit,
-    ProgressiveBalancesCache, RelativeEpoch, Unsigned, Validator,
+    ProgressiveBalancesCache, RelativeEpoch, Unsigned, Validator, Vector,
 };
 
 pub struct SinglePassConfig {
@@ -30,6 +30,7 @@ pub struct SinglePassConfig {
     pub pending_deposits: bool,
     pub pending_consolidations: bool,
     pub effective_balance_updates: bool,
+    pub proposer_lookahead: bool,
 }
 
 impl Default for SinglePassConfig {
@@ -48,6 +49,7 @@ impl SinglePassConfig {
             pending_deposits: true,
             pending_consolidations: true,
             effective_balance_updates: true,
+            proposer_lookahead: true,
         }
     }
 
@@ -60,6 +62,7 @@ impl SinglePassConfig {
             pending_deposits: false,
             pending_consolidations: false,
             effective_balance_updates: false,
+            proposer_lookahead: false,
         }
     }
 }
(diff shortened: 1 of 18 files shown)


consensus/types/src/beacon_state.rs
```diff
@@ -1071,28 +1097,38 @@ impl<E: EthSpec> BeaconState<E> {
             return Err(Error::SlotOutOfBounds);
         }
 
-        let seed = self.get_beacon_proposer_seed(slot, spec)?;
-        let indices = self.get_active_validator_indices(epoch, spec)?;
+        if let Ok(proposer_lookahead) = self.proposer_lookahead() {
+            // Post-Fulu
+            let index = slot.as_usize().safe_rem(E::slots_per_epoch() as usize)?;
```
michaelsproul, line 1103: This calculation only works if the `slot` is in the current epoch, which we check above. Although I wonder if eventually we might want to loosen that restriction so we can take advantage of the cache to access proposers in future epochs.

Maybe just worth adding a comment here to that effect.

consensus/state_processing/src/per_epoch_processing/single_pass.rs
```diff
@@ -460,9 +463,43 @@ pub fn process_epoch_single_pass<E: EthSpec>(
...
     }
 
+    if conf.proposer_lookahead && fork_name.fulu_enabled() {
+        process_proposer_lookahead(state, spec)?;
+    }
+
     Ok(summary)
 }
 
+// TOOO(EIP-7917): use balances cache
```
michaelsproul, line 473: This impl looks pretty speedy already, happy to merge as-is.

consensus/types/src/beacon_state.rs
```diff
@@ -1071,28 +1097,38 @@ impl<E: EthSpec> BeaconState<E> {
...
+            // Pre-Fulu
+            let seed = self.get_beacon_proposer_seed(slot, spec)?;
+            let indices = self.get_active_validator_indices(epoch, spec)?;
 
-        self.compute_proposer_index(&indices, &seed, spec)
+            self.compute_proposer_index(&indices, &seed, spec)
+        }
     }
 
     /// Returns the beacon proposer index for each `slot` in `self.current_epoch()`.
```
michaelsproul:
```suggestion
    /// Returns the beacon proposer index for each `slot` in `epoch`.
```

michaelsproul (approved):
Looks great!

I just had a few suggestions for comments, and I pushed one small test

michaelsproul:
Notably EF tests are passing since the tarballs got rebuilt to relocate the new sanity tests

consensus/types/src/beacon_state.rs
ethDreamer: done!

consensus/types/src/beacon_state.rs
ethDreamer: done!

PR #7560, opened 2025-06-05
EF tests v1.6.0-alpha.1 passing

michaelsproul:
## Proposed Changes

Simple change to get the EF tests passing on v1.6.0-alpha.1

This PR may not be necessary at all if the tarball is update to move these tests inside `sanity_blocks`, which I think might happen soon.


michaelsproul:
Closing in favour of:

- https://github.com/sigp/lighthouse/pull/7559

The EF test vectors got patched to make this change obsolete.

Issue #7550, opened 2025-06-03
Add e2e sync tests for lookup and range sync to CI


jimmygchen:
I made an attempt to add the following tests on to in #7530 
- genesis sync on a local devnet (covers current and next fork)
  - brief shutdown and restart (covers lookup sync)
  - longer shutdown and restart (covers range sync)

I tried using [`kurtosis-sync-tests`](https://github.com/ethpandaops/kurtosis-sync-test/blob/main/synctest.sh) and ran into a some issues with assertoor and I ended up removing the tests from the PR so we can move forward.

Given the simplicity of these tests, I think we could just have a simple bash script (similar to the checkpoint sync one in #7530) to do this (without having to debug assertoor):
1. Start a network
2. Immediately shutdown a node for some time
3. Restart the node and poll its sync status until it's fully sync 

The matrix I think we should cover:
https://github.com/sigp/lighthouse/blob/cfcc0d24b34e800ab58b8190d06f964b9cf97a2e/.github/workflows/local-testnet.yml#L221-L224

The commit for the previous attempt is here
https://github.com/sigp/lighthouse/pull/7530/commits/6c7e83925fe1ba682e8d439960fca50b025414dd

jimmygchen:
Done in #7561, thanks @chong-he

PR #7561, opened 2025-06-05
Add genesis sync test to CI

chong-he:
## Issue Addressed

* #7550 

Use existing code from  @jimmygchen in #7530 and modify for genesis sync test. Thanks @jimmygchen !


chong-he:
This PR is ready for review. In this test, we use the `minimal` preset so that it is faster. The local testnet script is also updated to build for [minimal](github.com/sigp/lighthouse/pull/7561/commits/e44df3c09e7835af9a8aa4ee2e14c764f103d634): 

The rebuild image `-b true` [(link here)](https://github.com/sigp/lighthouse/pull/7561/commits/635fbe6dce33a2de4ec5258da070c06d1168fe0d) is required or Kurtosis [wouldn't start](https://github.com/sigp/lighthouse/actions/runs/15479511703/job/43582752395)

For 120s stop duration, we trigger short distance sync, the lookup sync. We see in the logs:

```
DEBUG Created block lookup                          peers: [PeerId("16Uiu2HAkzuo2s4zS8Ab1FVPU1GAue212NT3YhHVRgJwE9S1HQ12J")], block_root: 0x149f7747579266942795085564847f3e903de77524ddb59de3077fae5b238432, awaiting_parent: "none", id: 1, service: "lookup_sync"
```

For 300s stop duration, we trigger long distance sync, the range sync. We see in the logs:

```
DEBUG Syncing new finalized chain                   id: 1, component: "range_sync"
DEBUG Sync RPC request sent                         method: "BlocksByRange", slots: 8, epoch: 0, peer: 16Uiu2HAm2pJM6EUvzVtN5JwaKyXBx8DAyg9dSXrNPBc81UCP8WBp, id: 3/2/RangeSync/0/1, chain: 1, service: "range_sync"
```

All sync completed very fast, under 5s. So I reduce the polling time in the script to 0.5s so that we can see some outputs in the terminal, and set a shorter timeout of 5 minutes.

scripts/tests/genesis-sync.sh
```diff
@@ -0,0 +1,151 @@
...
+exit_and_dump_logs() {
+    local exit_code=$1
+    echo "Shutting down..."
+    $SCRIPT_DIR/../local_testnet/stop_local_testnet.sh $ENCLAVE_NAME
+    echo "Test completed with exit code $exit_code."
+    exit $exit_code
+}
+
+# Start the nodes
+$SCRIPT_DIR/../local_testnet/start_local_testnet.sh -e $ENCLAVE_NAME -b true -n $CONFIG
```
jimmygchen: we can actually do `-b false` here because we already have the image from the previous step on CI, and we dont need to rebuild the same image.

chong-he: changed to `-b false` and add `spec-minimal` during the build, thank you 

[d7b0926](https://github.com/sigp/lighthouse/pull/7561/commits/d7b0926fe3dc0d0bd114992bdcab24ee5ad537b3)

scripts/tests/genesis-sync.sh
```diff
@@ -0,0 +1,151 @@
...
+ENCLAVE_NAME=${1:-genesis-sync-testnet}
+CONFIG=${2:-$SCRIPT_DIR/genesis-sync-config-electra.yaml}
+FORK_TYPE=${3:-electra}  # electra or fulu
+OFFLINE_DURATION_SECS=${4:-120} # stopped duration  of non validating nodes
+
+# Test configuration
+# ------------------------------------------------------
+# Interval for polling the /lighthouse/syncing endpoint for sync status
+# Reduce the polling time so that some progress can be seen
+POLL_INTERVAL_SECS=0.5
```
jimmygchen, line 18: This is very frequent, I think it's not necessary and we could probably live with something like 5s, or 6s (1 poll per slot)? Don't forgot HTTP API requests goes through `BeaconProcessor` as well and it's `Priority::P0`, which means it will be prioritised over most tasks.

https://github.com/sigp/lighthouse/blob/7e2df6b602a1c3593605eb21cd4761aebbdb90e7/beacon_node/http_api/src/lib.rs#L4384

scripts/tests/genesis-sync.sh
```diff
@@ -0,0 +1,151 @@
...
+  if [ "$sync_state" = "object" ]; then
+    status=$(echo "$response" | jq -r '.data | keys[0] // "Unknown"')
+    fields=$(echo "$response" | jq -r ".data.${status} | to_entries | map(\"\(.key): \(.value)\") | join(\", \")")
+    echo "${node_type} status: ${status}, ${fields}"
+  else
+    status=$(echo "$response" | jq -r '.data' 2>/dev/null)
+    echo "${node_type} status: ${status:-Unknown}"
+    
+    # The test is complete when the node is synced
+    if [ "$status" = "Synced" ]; then
```
jimmygchen, line 52: I removed the "Synced" check in the checkpoint sync script, because of the issue we talked about on the call - there's a bug and LH may switch to `Synced` even though it's not.

But I guess for forward sync this is fine, as it should only switch to this status if we're within certain distance. If this doesn' work well, we could consider using the standard `node/syncing` API, as it also returns the sync distance, which provides a bit more certainty.

jimmygchen: I looked at the `sync-electra-300s` logs, and realised it didn't even start syncing, i suspect LH is reporting `Synced` status too early. Would it be worth looking at the sync distance?

scripts/tests/genesis-sync.sh
chong-he: As discussed, the sync completed very fast, like under 2s, e.g., for `sync-electra-300s`, we see from the logs for `cl-3-lighthouse-geth`:

```
Jun 06 06:33:34.585 DEBUG Range request block roots retrieved           req_type: "BlocksByRange", start_slot: 8, req_count: 8, roots_count: 8, source: "store", elapsed: 89.035µs, finalized_slot: 32
Jun 06 06:33:34.774 DEBUG Range request block roots retrieved           req_type: "BlocksByRange", start_slot: 16, req_count: 8, roots_count: 8, source: "store", elapsed: 40.275µs, finalized_slot: 32 
Jun 06 06:33:34.774 DEBUG Range request block roots retrieved           req_type: "BlocksByRange", start_slot: 24, req_count: 8, roots_count: 8, source: "store", elapsed: 61.705µs, finalized_slot: 32
```

and the sync is completed about 0.2-0.3 second later:
```
Jun 06 06:33:35.000 INFO  Synced                                        peers: "3", exec_hash: "0xa95848dca51fbedfce6f84bd2d7f2df0b459a5672c8c8fb118ad44b321170f2d (unverified)", finalized_root: 0x57274ba74a8a091caa527a1af01269489c333da085503a0ae76d3534ecbe03d0, finalized_epoch: 4, epoch: 6, block: "0x2fb201cc9202c794b31212e49d8c828274fa76669db0453993bd8650a19dad10", slot: 52 
```

So if we poll every 5s, we may not see any output, unless we increase the sync distance (stop the node for longer periods of time

scripts/tests/genesis-sync.sh
chong-he: For forward sync it looks ok? This is the output for `sync-electra-300s`: 
https://github.com/sigp/lighthouse/actions/runs/15484117320/job/43595585222?pr=7561

We can increase the poll time to 1-2s if that's better? 

I have also tested using the beacon API `/eth/v1/node/syncing`, the output could look like this (also with 0.5s poll time):
```
olling sync status until nodes are synced or timeout of 5 mins
supernode status: Synced (head_slot: 51, sync_distance: 0)
supernode completed sync in 1 seconds
fullnode status: No response or null response
fullnode status: No response or null response
fullnode status: No response or null response
fullnode status: Syncing (head_slot: 23, sync_distance: 28, is_optimistic: true, el_offline: false)
fullnode status: Synced (head_slot: 51, sync_distance: 0)
fullnode completed sync in 3 seconds
Genesis sync test complete! Both supernode and fullnode have synced successfully.
Supernode time: 1 seconds
Fullnode time: 3 seconds
```

It has the `head_slot` (which I believe is the same as `target_slot` in the `lighthouse/syncing` endpoint (the beacon API also has the some other info in the response like `el_offline` and `is_optimistic`). If the beacon API is better, I can switch to that

scripts/tests/genesis-sync.sh
jimmygchen: Got it, let's go with this but bear in mind that it's a P0 and may impact performance, especially on low capacity machines like the runner.

scripts/tests/genesis-sync.sh
jimmygchen: Sure, we can evaluate that - i think your change is useful for the time being, and we'd like to merge it to start running it. We can improve this as we go.

jimmygchen (approved):

PR #7578, opened 2025-06-07
Implement basic validator custody framework (no backfill)

### LINKED ISSUE #6767: Implement basic validator data column custody

pawanjay176:
## Issue Addressed

Resolves #6767 

## Proposed Changes

This PR implements a basic version of validator custody. 
- It introduces a new `CustodyContext` object which contains info regarding number of validators attached to a node and  the custody count they contribute to the cgc.
- The `CustodyContext` is added in the da_checker and has methods for returning the current cgc and the number of columns to sample at head. Note that the logic for returning the cgc existed previously in the network globals.
- To estimate the number of validators attached, we use the `beacon_committee_subscriptions` endpoint. This might overestimate the number of validators actually publishing attestations from the node in the case of multi BN setups. We could also potentially use the `publish_attestations` endpoint to get a more conservative estimate at a later point.
- Anytime there's a change in the `custody_group_count` due to addition/removal of validators, the custody context should send an event on a broadcast channnel. The only subscriber for the channel exists in the network service which simply subscribes to more subnets. There can be additional subscribers in sync that will start a backfill once the cgc changes.

 TODO

- [ ] **NOT REQUIRED:** Currently, the logic only handles an increase in validator count and does not handle a decrease. We should ideally unsubscribe from subnets when the cgc has decreased. 
- [ ] **NOT REQUIRED:** Add a service in the `CustodyContext` that emits an event once `MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS ` passes after updating the current cgc. This event should be picked up by a subscriber which updates the enr and metadata.
- [x] Add more tests


beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -0,0 +1,433 @@
...
+        // This is the "weight" of the validator based on the effective balance
+        let num_validators_for_effective_balance =
+            effective_balance / spec.balance_per_additional_custody_group;
+        self.validators
+            .insert(validator_index, num_validators_for_effective_balance);
+
+        // Each `BALANCE_PER_ADDITIONAL_CUSTODY_GROUP` effectively contributes one unit of "weight".
+        let validator_count_at_epoch = self.validators.values().sum();
+        let validator_custody_requirement =
+            get_validators_custody_requirement(validator_count_at_epoch, spec);
```
dknopik: This causes a total of 3 additional custody groups if one has 3 validators with 32 ETH effective balance each, but only 2 additional custody groups if one has 2 validators with 48 ETH each (as we round down per validator) despite both scenarios having a total effective balance of 96 ETH. 

If I read the spec correctly https://github.com/ethereum/consensus-specs/blob/23e30671489c06ed3cc1d3662c16a09ef2f71176/specs/fulu/validator.md?plain=1#L125-L127 we need to sum first and then divide, to ensure we round correctly.

beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -885,6 +888,24 @@ impl<E: EthSpec> Network<E> {
         }
     }
 
+    /// Subscribe to all data columns determined by the cgc.
+    /// TODO(pawan): unsubscribe if count reduces
```
dknopik: The spec says we should keep with the highest cgc, so I think there is no scenario where we need to unsubscribe? But technically, it says `SHOULD`, so we _could_ still do this.
https://github.com/ethereum/consensus-specs/blob/23e30671489c06ed3cc1d3662c16a09ef2f71176/specs/fulu/validator.md?plain=1#L152-L157

dknopik (commented):
Took a look at this as an opportunity to refamiliarise myself with the DAS code and spec. I hope I looked at the correct spec version x) 

Also, just noting as I found no explicit TODO or code for this: I think we also need to update the ValidatorRegistrations if the effective balance changes due to beacon chain rewards, penalties, deposits and (partial) withdrawals.

beacon_node/lighthouse_network/src/service/mod.rs
jimmygchen: Yep,  I think we can also just not do it at all. I see little benefit of handling the reduce case - for a node operator thats running with significant amount of validator stake, this adds marginal benefit as they'll be running with supernode grade hardware and network. Keeping it not only keeps our code simple but also benefit the overall network.

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -0,0 +1,433 @@
...
+const CHANNEL_CAPACITY: usize = 10;
+
+/// TODO(pawan): this currently just registers increases in validator count.
+/// Does not handle decreasing validator counts
+#[derive(Default, Debug)]
+struct ValidatorRegistrations {
+    /// Set of all validators that is registered to this node along with its effective balance
+    /// in increments of `BALANCE_PER_ADDITIONAL_CUSTODY_GROUP`
+    ///
+    /// Key is validator index and value is effective_balance // BALANCE_PER_ADDITIONAL_CUSTODY_GROUP.
```
jimmygchen: Any reason why we don't keep the raw effective balance here? Don't we potentially end up with lower custody requirements if we do floor division first? 

e.g. Say if i have 2 vals with 63 eth each, then 
- divide individually then sum: 2 custody units
- summing up first then divide: 3 custody unit

jimmygchen: ah sorry just read @dknopik 's comment [above](https://github.com/sigp/lighthouse/pull/7578#discussion_r2133988876)! nvm

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -0,0 +1,433 @@
...
+    /// need to custody when receiving blocks over gossip/rpc or during sync.
+    pub fn custody_group_count(&self, spec: &ChainSpec) -> u64 {
+        if self.current_is_supernode {
+            return spec.number_of_custody_groups;
+        }
+        let validator_custody_count_at_head =
+            self.validator_custody_count_at_head.load(Ordering::Relaxed);
+
+        // If there are no validators, return the minimum custody_requirement
+        if validator_custody_count_at_head > 0 {
```
jimmygchen, line 220: Could this be simplified to
```rust
max(validator_custody_count_at_head, spec.custody_requirement)
```

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -0,0 +1,433 @@
...
+
+    pub fn subscribe(&self) -> Receiver<CustodyContextMessage> {
+        self.sender.subscribe()
+    }
+
+    /// Register a new validator index and updates the list of validators if required.
+    ///
+    /// Also modifies the internal structures if the validator custody has changed to
+    /// update the `custody_column_count`.
+    pub fn register_validator<E: EthSpec>(
```
jimmygchen: perhaps rename this to `register_validators` given we're accepting a `Vec`

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -0,0 +1,433 @@
...
+        );
+
+        assert_eq!(
+            custody_context.advertised_custody_group_count(&spec),
+            spec.custody_requirement,
+            "advertised custody count should not change"
+        );
+
+        // adding 1 more validator should increase the custody count
+        custody_context.register_validator::<E>(vec![(8, 32_000_000_000)], Slot::new(0), &spec);
```
jimmygchen: shouldn't the custdoy group count be `spec.validator_custody_requirement` until the stake reaches `8 * eth per custody increment`?

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -0,0 +1,433 @@
...
+        // add a new entry for the current epoch
+        if Some(validator_custody_requirement) != self.latest_validator_custody_requirement() {
+            self.epoch_validator_custody_requirements
+                .entry(epoch)
+                .and_modify(|old_custody| *old_custody = validator_custody_requirement)
+                .or_insert(validator_custody_requirement);
+        }
+        tracing::debug!(
+            %epoch,
+            validator_count = validator_count_at_epoch,
```
jimmygchen: `validator_count_at_epoch` is a bit confusing, not sure if i have a better name though, maybe something like `raw_validator_custody_group_count`?

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -0,0 +1,433 @@
...
+        std::cmp::max(count, spec.validator_custody_requirement),
+        spec.number_of_custody_groups,
+    )
+}
+
+/// Contains all the information the node requires to calculate the
+/// number of columns to be custodied when checking for DA.
+#[derive(Debug)]
+pub struct CustodyContext {
+    /// Columns to be custodied based on number of validators
```
jimmygchen: nit:

```rust
    /// Columns to be custodied based on the number of validators
```

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -0,0 +1,433 @@
...
+    persisted_is_supernode: bool,
+}
+
+impl From<&CustodyContext> for CustodyContextSsz {
+    fn from(context: &CustodyContext) -> Self {
+        CustodyContextSsz {
+            advertised_validator_custody: context
+                .advertised_validator_custody_count
+                .load(Ordering::Relaxed)
+                .clone(),
```
jimmygchen: these clones after `.load` are not required as `AtomicU64::load() ` returns `u64` ?

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -0,0 +1,433 @@
...
+                    updated_cgc,
+                    "Custody group count updated"
+                );
+                if let Err(e) = self
+                    .sender
+                    .send(CustodyContextMessage::HeadCustodyCountChanged {
+                        new_custody_count: updated_cgc,
+                    })
+                {
+                    tracing::error!(error=?e, "Failed to send custody context message");
```
jimmygchen: How do we handle this correctly?
What if the `send` fails, and the message would never be sent even though the context is updated, and it won't be retried again until the `new_validator_custody` changes again.
Should we flip the order and only update after successfully send?

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -668,6 +670,34 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+    /// Persists the custody information to disk.
+    pub fn persist_custody_context(&self) -> Result<(), Error> {
+        let custody_context: CustodyContextSsz = self
+            .data_availability_checker
+            .custody_context()
+            .as_ref()
+            .into();
+        debug!(?custody_context, "Persisting custody context to store");
+
+        if let Err(e) =
```
jimmygchen: Is it necessary to clear the custody context here?
Thinking about scenarios where persist function below fails, then our we'd have nothing in the DB.

Also I think we should probably persist this more often, so we're crash safe?  Perhaps when every time there's a change?

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -668,6 +670,34 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+        {
+            error!(error = ?e, "Failed to clear old custody context");
+        }
+
+        match persist_custody_context::<T::EthSpec, T::HotStore, T::ColdStore>(
+            self.store.clone(),
+            custody_context,
+        ) {
+            Ok(_) => info!("Saved custody state"),
+            Err(e) => error!(
```
jimmygchen: I think we want to return the error here? or is this intentional?

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -7187,6 +7217,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
 impl<T: BeaconChainTypes> Drop for BeaconChain<T> {
     fn drop(&mut self) {
         let drop = || -> Result<(), Error> {
+            self.persist_custody_context()?;
```
jimmygchen: Does the order here matter?

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -0,0 +1,433 @@
...
+    /// Set of all validators that is registered to this node along with its effective balance
+    /// in increments of `BALANCE_PER_ADDITIONAL_CUSTODY_GROUP`
+    ///
+    /// Key is validator index and value is effective_balance // BALANCE_PER_ADDITIONAL_CUSTODY_GROUP.
+    validators: HashMap<usize, u64>,
+    /// Maintains the validator custody requirement at a given epoch.
+    ///
+    /// Note: Only stores the epoch value when there's a change in custody requirement.
+    /// So if epoch 10 and 11 has the same custody requirement, only 10 is stored.
+    epoch_validator_custody_requirements: BTreeMap<Epoch, u64>,
```
jimmygchen, line 30: I think we might want to prune this map periodically?

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -0,0 +1,433 @@
...
+    pub fn register_validator<E: EthSpec>(
+        &mut self,
+        validator_index: usize,
+        effective_balance: u64,
+        slot: Slot,
+        spec: &ChainSpec,
+    ) {
+        let epoch = slot.epoch(E::slots_per_epoch());
+        // This is the "weight" of the validator based on the effective balance
+        let num_validators_for_effective_balance =
```
jimmygchen: Calling this validator might be a bit confusing now with maxEB, what about  `validator_custody_units` or something similar?

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -0,0 +1,433 @@
...
+        tracing::debug!(
+            %epoch,
+            validator_count = validator_count_at_epoch,
+            validator_custody_requirement,
+            "Registered validators"
+        );
+    }
+}
+
+/// Given the `count` of validators, return the custody requirement based on
```
jimmygchen: Same as above, might be worth coming up with a term for this.

beacon_node/lighthouse_network/src/types/globals.rs
```diff
@@ -255,7 +268,15 @@ impl<E: EthSpec> NetworkGlobals<E> {
...
         let enr_key: discv5::enr::CombinedKey = discv5::enr::CombinedKey::from_secp256k1(&keypair);
         let enr = discv5::enr::Enr::builder().build(&enr_key).unwrap();
-        NetworkGlobals::new(enr, metadata, trusted_peers, false, config, spec)
+        NetworkGlobals::new(
+            enr,
+            metadata,
+            trusted_peers,
+            false,
+            config,
+            spec.number_of_columns,
```
jimmygchen: should this be custody groups?
```suggestion
            spec.number_of_custody_groups,
```

beacon_node/lighthouse_network/src/types/topics.rs
```diff
@@ -126,7 +126,7 @@ pub fn all_topics_at_fork<E: EthSpec>(fork: ForkName, spec: &ChainSpec) -> Vec<G
         enable_light_client_server: true,
         subscribe_all_subnets: true,
         subscribe_all_data_column_subnets: true,
-        sampling_subnets: &sampling_subnets,
+        sampling_subnets: sampling_subnets,
```
jimmygchen: Use shorthand initializer

beacon_node/network/src/service.rs
```diff
@@ -270,7 +274,15 @@ impl<T: BeaconChainTypes> NetworkService<T> {
...
 
         // launch libp2p service
-        let (mut libp2p, network_globals) = Network::new(executor.clone(), service_context).await?;
+        let (mut libp2p, network_globals) = Network::new(
+            executor.clone(),
+            service_context,
+            beacon_chain
+                .data_availability_checker
+                .custody_context()
+                .sampling_count(&beacon_chain.spec),
```
jimmygchen: should this be the advertised custody group count ?

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -0,0 +1,433 @@
...
+            advertised_validator_custody_count: AtomicU64::new(0),
+            validator_custody_count_at_head: AtomicU64::new(0),
+            current_is_supernode: is_supernode,
+            persisted_is_supernode: is_supernode,
+            validator_registrations: Default::default(),
+            sender,
+        }
+    }
+
+    /// Deserialize a `CustodyContext` from SSZ bytes.
```
jimmygchen: We don't deserialize here, is this outdated?

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -0,0 +1,433 @@
...
+    pub fn advertised_custody_group_count(&self, spec: &ChainSpec) -> u64 {
+        if self.persisted_is_supernode {
+            return spec.number_of_custody_groups;
+        }
+        let advertised_validator_custody = self
+            .advertised_validator_custody_count
+            .load(Ordering::Relaxed);
+
+        // If there are no validators, return the minimum custody_requirement
+        if advertised_validator_custody > 0 {
```
jimmygchen: Could this be simplified to
`max(advertised_validator_custody, spec.custody_requirement)`

beacon_node/http_api/src/lib.rs
```diff
@@ -3768,26 +3768,44 @@ pub fn serve<T: BeaconChainTypes>(
...
+                    for subscription in committee_subscriptions.iter() {
+                        chain
+                            .validator_monitor
+                            .write()
+                            .auto_register_local_validator(subscription.validator_index);
+                        // Register the validator with the `CustodyContext`
+                        if let Ok(effective_balance) = chain
+                            .canonical_head
+                            .cached_head()
+                            .snapshot
```
jimmygchen: We can probably extract this to a variable to avoid re-acquiring the read lock

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -0,0 +1,433 @@
...
+                {
+                    tracing::error!(error=?e, "Failed to send custody context message");
+                }
+            }
+        }
+    }
+
+    /// The custody count that we advertise to our peers in our metadata and
+    /// enr values.
+    pub fn advertised_custody_group_count(&self, spec: &ChainSpec) -> u64 {
```
jimmygchen: This method is unused now - we'll need to call this on startup, and on backfill complete right?

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -0,0 +1,433 @@
...
+            spec.custody_requirement
+        }
+    }
+
+    /// The custody count that we use to custody columns currently.
+    ///
+    /// This function should be called when figuring out how many columns we
+    /// need to custody when receiving blocks over gossip/rpc or during sync.
+    pub fn custody_group_count(&self, spec: &ChainSpec) -> u64 {
+        if self.current_is_supernode {
```
jimmygchen, line 214: do we cover the case where the user changes from non supernode to supernode ? do we backfill or just require user to resync?

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -0,0 +1,433 @@
...
+                (6, 32_000_000_000),
+                (7, 32_000_000_000),
+            ],
+            Slot::new(0),
+            &spec,
+        );
+
+        assert_eq!(
+            custody_context.custody_group_count(&spec),
+            spec.validator_custody_requirement + spec.custody_requirement,
```
jimmygchen: I think we're not spec compliant here - in this scenario, CGC should be `spec.validator_custody_requirement`.

```python
min(max(count, VALIDATOR_CUSTODY_REQUIREMENT), NUMBER_OF_CUSTODY_GROUPS)
```

https://github.com/ethereum/consensus-specs/blob/dev/specs/fulu/validator.md#validator-custody

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -0,0 +1,433 @@
...
+impl CustodyContext {
+    /// Create a new custody default custody context object when no persisted object
+    /// exists.
+    ///
+    /// The `is_supernode` value is based on current cli parameters.
+    pub fn new(is_supernode: bool) -> Self {
+        let (sender, _) = channel(CHANNEL_CAPACITY);
+
+        Self {
+            advertised_validator_custody_count: AtomicU64::new(0),
```
jimmygchen: should this be called `advertised_custody_group_count`?
and default to `spec.custody_requirement`?

jimmygchen: ah right, i see that you're recording the _actual_ validator custody in this field, and returning the default in the functions. Resolving this.

beacon_node/beacon_chain/src/validator_custody.rs
jimmygchen: Addressed in https://github.com/sigp/lighthouse/pull/7578/commits/1adcc4a5406911fa8677a51507a787a9c1565e2c

beacon_node/beacon_chain/src/validator_custody.rs
jimmygchen: Addressed in https://github.com/sigp/lighthouse/pull/7578/commits/1adcc4a5406911fa8677a51507a787a9c1565e2c

beacon_node/beacon_chain/src/validator_custody.rs
jimmygchen: Addressed in https://github.com/sigp/lighthouse/pull/7578/commits/1adcc4a5406911fa8677a51507a787a9c1565e2c

beacon_node/lighthouse_network/src/types/topics.rs
jimmygchen: Addressed in https://github.com/sigp/lighthouse/pull/7578/commits/1adcc4a5406911fa8677a51507a787a9c1565e2c

beacon_node/beacon_chain/src/validator_custody.rs
jimmygchen: looks like [the impl](https://github.com/sigp/lighthouse/pull/7578/commits/1adcc4a5406911fa8677a51507a787a9c1565e2c) was correct, just need to update the tests. Addressed in 
https://github.com/sigp/lighthouse/blob/b23a4dcc787728109893b0e09520a8fde15d1c8f/beacon_node/beacon_chain/src/validator_custody.rs#L93-L98

beacon_node/network/src/service.rs
jimmygchen: No, we actually want to subscribe to `sample_count` number of subnets, but use `advertised_custody_group` for metadata and ENR

beacon_node/lighthouse_network/src/service/mod.rs
pawanjay176: Yeah I don't think we should do it either. Just wanted to gauge interest about doing it in this PR, I think we could reconsider this at a later point.

beacon_node/beacon_chain/src/validator_custody.rs
pawanjay176: Yeah nice catch. Thanks for fixing! ❤️

beacon_node/beacon_chain/src/validator_custody.rs
pawanjay176: I just wanted the condition to be explicit for clarity, but can change it to what you said

jimmygchen: Yeah originally I was also concerned about the case where the value is > 0 but less than `spec.custody_requirement`, but this is ensured here, so no strong opinion to keep it as it is.
https://github.com/sigp/lighthouse/blob/b23a4dcc787728109893b0e09520a8fde15d1c8f/beacon_node/beacon_chain/src/validator_custody.rs#L93-L98

beacon_node/beacon_chain/src/validator_custody.rs
pawanjay176: The send would only fail if there are no receivers. I think since we atleast have one receiver in the network service, this should never fail because the network service loop should ideally start before the http service and if it doesn't, I think its good if we detect it early so we can change some assumptions here regarding ordering of operations.

beacon_node/beacon_chain/src/beacon_chain.rs
pawanjay176: Yeah probably not necessary to clear it, I just copied the other persist functions. I'll fix that up.

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -344,90 +336,92 @@ mod tests {
...
+        }
+    }
 
-        assert_eq!(
-            custody_context.advertised_custody_group_count(&spec),
-            spec.custody_requirement,
-            "advertised custody count should not change"
-        );
+    #[test]
+    fn register_multiple_validators_should_update_cgc() {
```
jimmygchen, line 331: Note that this test actually register different validator each epoch, so it confirms that we don't decrease the validator custody - i should probably add a separate test to cover what i originally intended to cover (multiple validators at the same time, not different epochs)

beacon_node/beacon_chain/src/beacon_chain.rs
pawanjay176: > Also I think we should probably persist this more often, so we're crash safe? Perhaps when every time there's a change?

Not sure its necessary. Its not catastrophic if the custody context doesn't get persisted. In case of an unclean shutdown, I think we might have bigger issues with other core beacon chain stuff not getting persisted.

beacon_node/beacon_chain/src/validator_custody.rs
pawanjay176: Yeah I thought about this too.
I ended up not doing it because we aren't storing changes at every epoch and once we reach the `custody_count` here it will not change, so the map size will be sorta contained.

beacon_node/beacon_chain/src/validator_custody.rs
pawanjay176: Yeah I like `validator_custody_units` way better, thanks!

beacon_node/beacon_chain/src/validator_custody.rs
pawanjay176: In the current state of the PR, I was treating changes from `is_supernode == false => true` similar to the validator custody count increasing. Just don't update the cgc until the min_epochs_for_blob_requests has passed.
However, there's a case to be made to just start backfilling immediately in both cases and update the advertised count once the backfill is complete. 
I think if we have the machinery in sync to start backfill for columns, its pretty easy to trigger it from here. We can just trigger an event and have a subscriber in the sync manager. 
imo, we should be doing the backfill in both cases since its more robust and sync is slightly easier too (since we already have the blocks). However, i didn't want to touch that until after the sync refactor.

beacon_node/http_api/src/lib.rs
```diff
@@ -3768,26 +3768,44 @@ pub fn serve<T: BeaconChainTypes>(
...
+                    let mut registrations = Vec::new();
+                    for subscription in committee_subscriptions.iter() {
+                        chain
+                            .validator_monitor
+                            .write()
+                            .auto_register_local_validator(subscription.validator_index);
+                        // Register the validator with the `CustodyContext`
+                        if let Ok(effective_balance) = chain
+                            .canonical_head
+                            .cached_head()
```
jimmygchen: should use finalized state here

beacon_node/beacon_chain/src/validator_custody.rs
jimmygchen: Yeah true, the max could be 128. I'll leave it as it is now

beacon_node/beacon_chain/src/validator_custody.rs
jimmygchen: make sense

beacon_node/beacon_chain/src/validator_custody.rs
jimmygchen: Make sense. Leaving it as it is

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Removed clearing and persist custody context last.

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Doesn't matter. Same output.

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Updated

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -0,0 +1,403 @@
...
+use tokio::sync::broadcast::{channel, Receiver, Sender};
+use types::{ChainSpec, Epoch, EthSpec, Slot};
+
+const CHANNEL_CAPACITY: usize = 10;
+
+/// TODO(pawan): this currently just registers increases in validator count.
+/// Does not handle decreasing validator counts
+#[derive(Default, Debug)]
+struct ValidatorRegistrations {
+    /// Set of all validators that is registered to this node along with its effective balance
```
dapplion, line 20: The effective balance at what state? The spec defines that we must use the effective balance of the latest finalized state. Do we have the guarantee that this balance is the expected one, and matches the current finalized state of this DB?

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -0,0 +1,403 @@
...
+            tracing::debug!(
+                old_count = validator_custody_count_at_head,
+                new_count = new_validator_custody,
+                "Validator count at head updated"
+            );
+            self.validator_custody_count
+                .store(new_validator_custody, Ordering::Relaxed);
+
+            let updated_cgc = self.custody_group_count(spec);
+            // Send the message to network only if there are more columns subnets to subscribe to
```
dapplion, line 193: Isn't it quite leaky for the `beacon_chain` crate to be aware of the network and send it messages?

beacon_node/lighthouse_network/src/types/globals.rs
```diff
@@ -96,14 +88,40 @@ impl<E: EthSpec> NetworkGlobals<E> {
...
+            let subnets = compute_subnets_from_custody_group(*custody_index, &self.spec)
+                .expect("should compute custody subnets for node");
+            sampling_subnets.extend(subnets);
+        }
+
+        let mut sampling_columns = self.sampling_columns.write();
+        for custody_index in &custody_groups {
+            let columns = compute_columns_for_custody_group(*custody_index, &self.spec)
+                .expect("should compute custody columns for node");
+            sampling_columns.extend(columns);
```
dapplion, line 134: Noting that now we have two sources of truth (3 items total) that we must carefully synchronize:
- `NetworkGlobals::sampling_subnets`
- `NetworkGlobals::sampling_columns`
- `DataAvailabilityChecker::custody_context`

It's fine but having a single source of truth was the motivation of passing the `cgc` around

beacon_node/beacon_chain/src/validator_custody.rs
dapplion: Instead you can move this code to the network where you recv the validator custody events and call the network after `register_validators` in the HTTP api

beacon_node/http_api/src/lib.rs
```diff
@@ -3768,26 +3768,42 @@ pub fn serve<T: BeaconChainTypes>(
...
+                        });
+                    }
+                    chain
+                        .data_availability_checker
+                        .custody_context()
+                        .register_validators::<T::EthSpec>(
+                            registrations,
+                            chain.slot().unwrap(),
+                            &chain.spec,
+                        );
```
dapplion: In my POC I also used the builder registration, which is a much faster signal for those that use it

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -0,0 +1,403 @@
...
+    /// This should trigger actions downstream like
+    /// subscribing/unsubscribing new subnets/
+    /// backfilling required columns.
+    HeadCustodyCountChanged { new_custody_count: u64 },
+    /// The advertised custody count has changed.
+    ///
+    /// This should trigger downstream actions like setting
+    /// a new cgc value in the enr and metadata fields and
+    /// performing any related cleanup actions.
+    AdvertisedCustodyCountChanged { new_custody_count: u64 },
```
dapplion: Never used?

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -0,0 +1,403 @@
...
+
+    /// Register a new validator index and updates the list of validators if required.
+    pub fn register_validator<E: EthSpec>(
+        &mut self,
+        validator_index: usize,
+        effective_balance: u64,
+        slot: Slot,
+        spec: &ChainSpec,
+    ) {
+        let epoch = slot.epoch(E::slots_per_epoch());
```
dapplion: The benefit of tracking the epoch in the updates is to uphold the following invariant

> A given block will have the same CGC during its lifetime 

To prevent situations where we receive the block from gossip with a CGC of 8 and the moment of importing turns out we are at 16, failing import. But the columns have already been gossiped (and ignored by us because we were CGC 8). In that case we have to trigger lookup sync to fetch the remaining columns and recover. It may work but it feels messy. Instead we can schedule the update for the next epoch to prevent this edge case.

beacon_node/beacon_chain/src/validator_custody.rs
dapplion: Some notes of mine: I came to define the `AdvertisedCustodyCount` as:

Thanks to the CGC registry updates you can construct a function of `CGC(time)`. Consider `da_window` and `now` units of time. Then consider the set `da_window_CGCs` to contain all distinct values of `CGC(time)` function between `now - da_window` and `now`. Then the `AdvertisedCustodyCount = min(da_window_CGCs)`. This value represents the minimum CGC value that we can guarantee to have all expect columns for in our DB for the da window.

That is why it's important to set the registry update to an epoch in the future to guarantee the above.

beacon_node/beacon_chain/src/validator_custody.rs
jimmygchen: We register validators based on their effective balance of the finalized beacon state https://github.com/sigp/lighthouse/pull/7578/commits/ec76b5675ab7ec5ef372f2feb0ef874c9c12ad1a

beacon_node/beacon_chain/src/validator_custody.rs
pawanjay176: yeah that's way cleaner, thanks! I started out with some additional assumptions about updating the advertised validator custody that needed this, but I think its no longer relevant.

beacon_node/lighthouse_network/src/types/globals.rs
pawanjay176: The `sampling_subnets ` and `sampling_columns` are initialized with the cgc value and updated when the cgc is updated.  The alternative would be computing it on the fly which is unnecessary imo.
Passing the cgc around leaks fulu specifc context all over the beacon chain verification and the gossip methods when its not even relevant pre fulu, so not a fan of that approach.

beacon_node/http_api/src/lib.rs
pawanjay176: Might also make sense to use the `prepare_beacon_proposer` endpoint as that works for nodes not running mev workflows too.

beacon_node/beacon_chain/src/validator_custody.rs
jimmygchen: Addressed in https://github.com/sigp/lighthouse/pull/7578/commits/92a990a8ebce0eb42e13145f1026590d90cfe290

beacon_node/http_api/src/lib.rs
jimmygchen: Agree - although it would be great if we can get a clean version of this PR merged, and we can work on the remaining issues together (this + status updates etc)

beacon_node/beacon_chain/src/validator_custody.rs
jimmygchen: Agreed! addressed in https://github.com/sigp/lighthouse/pull/7578/commits/011def61c465c3375d11d7be2eea9bc85ec1a9d0

beacon_node/beacon_chain/src/validator_custody.rs
jimmygchen: Addressed in https://github.com/sigp/lighthouse/pull/7578/commits/92a990a8ebce0eb42e13145f1026590d90cfe290

beacon_node/network/src/service.rs
jimmygchen: Addressed in https://github.com/sigp/lighthouse/pull/7578/commits/cd7059c34b33c5435820a2cf3869287a7be6ad0e

We should pass CGC here, as the sampling count is derived in the function later.

jimmygchen (approved):
LGTM 

Next changes (separate PRs)
- add lowest block slot to status message, update it on CGC change, & use lowest block slot for peer selection when making rpc requests
- use additional http api to register validators
- backfill

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -0,0 +1,440 @@
...
+    sync::atomic::{AtomicU64, Ordering},
+};
+
+use parking_lot::RwLock;
+
+use ssz_derive::{Decode, Encode};
+use types::{ChainSpec, Epoch, EthSpec, Slot};
+
+/// A delay before making the CGC change effective to the data availability checker.
+const CUSTODY_CHANGE_DA_EFFECTIVE_DELAY_SECONDS: u64 = 30;
```
pawanjay176, line 12: can't we have this delay in slots as that's what how this constant is used below?

pawanjay176 (commented):
LGTM as well. Thanks for fixing up most of the outstanding issues!

beacon_node/beacon_chain/src/validator_custody.rs
jimmygchen: Discussed in person - using seconds so the delay is consistent on devnets with low slots per sec as well.

PR #7582, opened 2025-06-09
Batch verify KZG proofs for getBlobsV2

dknopik:
Instead of fully gossip verifying the fetched data columns, we instead only verify the KZG proofs in a single batch. Before, we filter by what's already observed by gossip and available in DA.

## Additional Info

An untested alternative to #7553

### DIFF
--- a/beacon_node/beacon_chain/src/fetch_blobs/fetch_blobs_beacon_adapter.rs
+++ b/beacon_node/beacon_chain/src/fetch_blobs/fetch_blobs_beacon_adapter.rs
@@ -1,21 +1,25 @@
 use crate::blob_verification::{GossipBlobError, GossipVerifiedBlob};
-use crate::data_column_verification::{GossipDataColumnError, GossipVerifiedDataColumn};
+use crate::data_column_verification::KzgVerifiedDataColumn;
 use crate::fetch_blobs::{EngineGetBlobsOutput, FetchEngineBlobError};
+use crate::observed_block_producers::ProposalKey;
 use crate::observed_data_sidecars::DoNotObserve;
 use crate::{AvailabilityProcessingStatus, BeaconChain, BeaconChainTypes};
 use execution_layer::json_structures::{BlobAndProofV1, BlobAndProofV2};
-use kzg::Kzg;
+use kzg::{Error as KzgError, Kzg};
 #[cfg(test)]
 use mockall::automock;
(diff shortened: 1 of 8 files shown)


beacon_node/beacon_chain/src/fetch_blobs/mod.rs
```diff
@@ -353,57 +355,55 @@ async fn compute_custody_columns_to_import<T: BeaconChainTypes>(
...
+                ) {
+                    custody_columns.retain(|col| !observed_columns.contains(&col.index));
+                    if custody_columns.is_empty() {
+                        return Ok(vec![]);
+                    }
+                }
+
+                // Only consider columns that are not already known to data availability.
+                if let Some(known_columns) = chain_adapter_cloned.cached_data_column_indexes(
+                    &block.signed_block_header().message.canonical_root(),
```
jimmygchen: we could pass `block_root` to this function to avoid computing it again

beacon_node/beacon_chain/src/fetch_blobs/fetch_blobs_beacon_adapter.rs
```diff
@@ -75,12 +79,28 @@ impl<T: BeaconChainTypes> FetchBlobsBeaconAdapter<T> {
         GossipVerifiedBlob::<T, DoNotObserve>::new(blob.clone(), blob.index, &self.chain)
     }
 
-    pub(crate) fn verify_data_column_for_gossip(
+    pub(crate) fn verify_data_columns_kzg(
+        &self,
+        data_columns: Vec<Arc<DataColumnSidecar<T::EthSpec>>>,
+    ) -> Result<Vec<KzgVerifiedDataColumn<T::EthSpec>>, KzgErrors> {
+        KzgVerifiedDataColumn::from_batch(data_columns, &self.chain.kzg)
```
jimmygchen, line 84: `KzgVerifiedDataColumn::from_batch` actually does up to two round of verifications, if batch verification fails, it attempts individual verification to identify bad columns - the purpose is to identify which peer sent the invalid columns. 

In our case here, we don't really need to know exactly which columns fails, so we can just run a single pass with `verify_kzg_for_data_column_list`

jimmygchen (changes requested):
Nice! just two minor comments.

jimmygchen (approved):
Nice, LGTM!

Issue #7577, opened 2025-06-07
Lookup sync: don't penalize peers that extend children of long chains


dapplion:
However, we then penalize peers who do the same action. This is inconsistent and can lead to penalties for honest peers. We should not penalize peers who extend known long chains. We could have to hashsets
- failed_chains: reject descendant lookups and penalize
- long_chains: reject descendant lookups and don't penalize

We should fix this in a future PR

_Originally posted by @dapplion in https://github.com/sigp/lighthouse/pull/7118#discussion_r2128094198_

gitToki:
Hi @dapplion could I take this ?

dapplion:
Yes @gitToki let me know if you need pointers

jimmygchen:
It looks like this is happening on devnet-3, we fall behind and fallback to range sync, and penalise the peer the served us data.

jimmygchen:
Closed by #8042

PR #7583, opened 2025-06-10
fix: Distinguish long chains from failed chains in block lookup sync

gitToki:
## Issue Addressed

[ticket](https://github.com/sigp/lighthouse/issues/7577#issuecomment-2959331589)

## Proposed Changes

When a chain reaches `PARENT_DEPTH_TOLERANCE`, the code was putting it in `failed_chains` and penalizing peers who reference blocks from that chain, even if the chain itself was valid

Changes:

- Add a separate `long_chains` for chains that are too long but not necessarily invalid
- Only penalize peers for blocks from `failed_chains` (actual processing failures)
- Don't penalize peers for blocks from `long_chains` (just drop)
- Reorganized struct field


macladson (changes requested):
Hey @gitToki thanks for the PR,

It looks like your branch was originally based on `stable` and so it has inherited some commits that we actually don't want in `unstable`. Can you do a rebase to remove those extra commits?

gitToki:
Closing this PR to create a clean one based on unstable. Will resubmit, sorry for that.

PR #7584, opened 2025-06-10
fix: Distinguish long chains from failed chains in block lookup sync

gitToki:
Issue Addressed

[ticket](https://github.com/sigp/lighthouse/issues/7577#issuecomment-2959331589) 

Proposed Changes

When a chain reaches PARENT_DEPTH_TOLERANCE, the code was putting it in failed_chains and penalizing peers who reference blocks from that chain, even if the chain itself was valid

Changes:

- Add a separate long_chains for chains that are too long but not necessarily invalid
- Only penalize peers for blocks from failed_chains (actual processing failures)
- Don't penalize peers for blocks from long_chains (just drop)
- Reorganized struct field

### DIFF
--- a/beacon_node/network/src/sync/block_lookups/mod.rs
+++ b/beacon_node/network/src/sync/block_lookups/mod.rs
@@ -114,6 +114,9 @@ pub struct BlockLookups<T: BeaconChainTypes> {
     /// A cache of failed chain lookups to prevent duplicate searches.
     failed_chains: LRUTimeCache<Hash256>,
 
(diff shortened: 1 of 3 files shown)


gitToki:
it should be good now 🫡 @jimmygchen

dapplion:
Hey @gitToki thanks for the contribution! But we are in the process of changing the syncing algorithm. The new version won't need to distinguish long chains since those will always be synced, see
- https://github.com/sigp/lighthouse/issues/7678

gitToki:
Oh okay, I will close this PR then if it's no longer relevant

PR #7587, opened 2025-06-11
Data column sidecar event

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Implement events for data column sidecar https://github.com/ethereum/beacon-APIs/pull/535

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -73,7 +73,9 @@ use crate::{
     kzg_utils, metrics, AvailabilityPendingExecutedBlock, BeaconChainError, BeaconForkChoiceStore,
     BeaconSnapshot, CachedHead,
 };
-use eth2::types::{EventKind, SseBlobSidecar, SseBlock, SseExtendedPayloadAttributes};
+use eth2::types::{
+    EventKind, SseBlobSidecar, SseBlock, SseDataColumnSidecar, SseExtendedPayloadAttributes,
+};
 use execution_layer::{
     BlockProposalContents, BlockProposalContentsType, BuilderParams, ChainHealth, ExecutionLayer,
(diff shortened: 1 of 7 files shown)


beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3191,6 +3204,31 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+        I: Iterator<Item = &'a DataColumnSidecar<T::EthSpec>>,
+    {
+        if let Some(event_handler) = self.event_handler.as_ref() {
+            if event_handler.has_data_column_sidecar_subscribers() {
+                let imported_data_columns = self
+                    .data_availability_checker
+                    .cached_data_column_indexes(block_root)
+                    .unwrap_or_default();
+                let new_data_columns =
+                    data_columns_iter.filter(|b| !imported_data_columns.contains(&b.index));
```
jimmygchen, line 3221: I think we can skip this check because the gossip verification should ensure no duplication?

pawanjay176: We need it filter duplicates for rpc verification though

beacon_node/beacon_chain/tests/events.rs
```diff
@@ -43,6 +47,42 @@ async fn blob_sidecar_event_on_process_gossip_blob() {
...
+        .deterministic_keypairs(8)
+        .fresh_ephemeral_store()
+        .mock_execution_layer()
+        .build();
+
+    // subscribe to blob sidecar events
+    let event_handler = harness.chain.event_handler.as_ref().unwrap();
+    let mut data_column_event_receiver = event_handler.subscribe_data_column_sidecar();
+
+    // build and process a gossip verified blob
```
jimmygchen:
```suggestion
    // build and process a gossip verified data column
```

jimmygchen (approved):
Nice!

PR #7588, opened 2025-06-11
Wait before column reconstruction

dknopik:
Column reconstruction is computationally intensive, and we want to wait a bit before we start as the remaining columns might come in via gossip or getblobsv2.

This is only a draft as there might be a better way to tackle this - we might e.g. start with reconstruction ASAP, but only on a single thread, and ramp up or cancel depending on arrival of other columns, but this would take some refactoring

### DIFF
--- a/beacon_node/beacon_processor/src/lib.rs
+++ b/beacon_node/beacon_processor/src/lib.rs
@@ -500,10 +500,12 @@ impl<E: EthSpec> From<ReadyWork> for WorkEvent<E> {
                 drop_during_sync: false,
                 work: Work::ChainSegmentBackfill(process_fn),
             },
-            ReadyWork::ColumnReconstruction(QueuedColumnReconstruction(process_fn)) => Self {
-                drop_during_sync: true,
-                work: Work::ColumnReconstruction(process_fn),
-            },
+            ReadyWork::ColumnReconstruction(QueuedColumnReconstruction { process_fn, .. }) => {
+                Self {
+                    drop_during_sync: true,
+                    work: Work::ColumnReconstruction(process_fn),
+                }
+            }
         }
     }
 }
--- a/beacon_node/beacon_processor/src/lib.rs
+++ b/beacon_node/beacon_processor/src/lib.rs
@@ -39,7 +39,7 @@
 //! task.
 
 use crate::work_reprocessing_queue::{
-    QueuedBackfillBatch, QueuedGossipBlock, ReprocessQueueMessage,
+    QueuedBackfillBatch, QueuedColumnReconstruction, QueuedGossipBlock, ReprocessQueueMessage,
 };
 use futures::stream::{Stream, StreamExt};
 use futures::task::Poll;
@@ -117,6 +117,7 @@ pub struct BeaconProcessorQueueLengths {
     rpc_custody_column_queue: usize,
     rpc_verify_data_column_queue: usize,
     sampling_result_queue: usize,
+    column_reconstruction_queue: usize,
     chain_segment_queue: usize,
     backfill_chain_segment: usize,
     gossip_block_queue: usize,
@@ -184,6 +185,7 @@ impl BeaconProcessorQueueLengths {
             rpc_verify_data_column_queue: 1000,
             unknown_block_sampling_request_queue: 16384,
             sampling_result_queue: 1000,
+            column_reconstruction_queue: 64,
             chain_segment_queue: 64,
             backfill_chain_segment: 64,
             gossip_block_queue: 1024,
@@ -498,6 +500,10 @@ impl<E: EthSpec> From<ReadyWork> for WorkEvent<E> {
                 drop_during_sync: false,
                 work: Work::ChainSegmentBackfill(process_fn),
             },
+            ReadyWork::ColumnReconstruction(QueuedColumnReconstruction(process_fn)) => Self {
+                drop_during_sync: true,
+                work: Work::ColumnReconstruction(process_fn),
+            },
         }
     }
 }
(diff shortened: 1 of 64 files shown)


beacon_node/beacon_processor/src/lib.rs
```diff
@@ -184,6 +185,7 @@ impl BeaconProcessorQueueLengths {
             rpc_verify_data_column_queue: 1000,
             unknown_block_sampling_request_queue: 16384,
             sampling_result_queue: 1000,
+            column_reconstruction_queue: 64,
```
jimmygchen, line 188: 64 feels a bit high for this - i think a single digist size would be enough? most likley this would be 1-2?

beacon_node/beacon_processor/src/work_reprocessing_queue.rs
```diff
@@ -54,6 +55,9 @@ pub const QUEUED_RPC_BLOCK_DELAY: Duration = Duration::from_secs(4);
 /// For how long to queue sampling requests for reprocessing.
 pub const QUEUED_SAMPLING_REQUESTS_DELAY: Duration = Duration::from_secs(12);
 
+/// For how long to queue delayed column reconstruction.
+pub const QUEUED_RECONSTRUCTION_DELAY: Duration = Duration::from_millis(150);
```
jimmygchen, line 59: It might be worth making this configurable - we'll likely have some numbers from pandaops soon now we expose the data column event streams.

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -386,8 +386,8 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                     // We don't publish columns reconstructed from rpc columns to the gossip network,
                     // as these are likely historic columns.
                     let publish_columns = false;
-                    if let Some(availability) = self
-                        .attempt_data_column_reconstruction(block_root, publish_columns)
+                    if let Some(availability) = Arc::clone(&self)
+                        .attempt_data_column_reconstruction(block_root, publish_columns, false)
```
jimmygchen: I probably need to think about this a bit more - this is the scenario where we make RPC request from peers and we didn't get everything we need - but we may have enough to perform reconstruction - in this case I think we'd want to eagerly perform it, because we are less likely to get them from gossip & the EL.

beacon_node/beacon_processor/src/lib.rs
dknopik: Yeah, I also can't think of a plausible scenario over 2. Let's pick 4 to be safe?

beacon_node/network/src/network_beacon_processor/sync_methods.rs
dknopik: Makes sense.

Trying to fix this I again realized how clumsily this is implemented, having to plumb `is_retry` through the methods. I'll refactor this a bit...

dknopik: Fixed

beacon_node/beacon_processor/src/lib.rs
dknopik: With the refactor this will trigger more often to push back the reconstruction - so 64 is needed at least (128 cols / 2)

jimmygchen: hmm wouldn't we only queue it once when it reaches 64? (and not trigger for 65, 66 ..etc)

jimmygchen:
@dknopik looks like there's a failing test related to the change:
```
    thread 'network_beacon_processor::tests::import_gossip_block_acceptably_early' panicked at beacon_node/network/src/network_beacon_processor/tests.rs:637:9:
      left: ["column_reconstruction", "worker_freed", "nothing_to_do"]
     right: ["delayed_import_block", "worker_freed", "nothing_to_do"]
```

beacon_node/beacon_processor/src/lib.rs
dknopik: It would queue for 65, 66, ... as the processor pushes back the deadline for every arriving column. Basically, we wait for another column for 150ms every time one arrives, instead of waiting for 150ms for all to arrive after we got half

beacon_node/beacon_processor/src/lib.rs
```diff
@@ -891,6 +902,8 @@ impl<E: EthSpec> BeaconProcessor<E> {
             FifoQueue::new(queue_lengths.rpc_verify_data_column_queue);
         // TODO(das): the sampling_request_queue is never read
         let mut sampling_result_queue = FifoQueue::new(queue_lengths.sampling_result_queue);
+        let mut column_reconstruction_queue =
+            FifoQueue::new(queue_lengths.column_reconstruction_queue);
```
jimmygchen, line 906: we never pop from this queue. Need to add it somewhere in this block based on priority
https://github.com/sigp/lighthouse/blob/b2f7a5befd87ca024915b92fda4f607702af26e9/beacon_node/beacon_processor/src/lib.rs#L1058

beacon_node/beacon_processor/src/lib.rs
```diff
@@ -498,6 +500,12 @@ impl<E: EthSpec> From<ReadyWork> for WorkEvent<E> {
                 drop_during_sync: false,
                 work: Work::ChainSegmentBackfill(process_fn),
             },
+            ReadyWork::ColumnReconstruction(QueuedColumnReconstruction { process_fn, .. }) => {
+                Self {
+                    drop_during_sync: true,
```
jimmygchen, line 505: Note that we will never perform reconstruction during sync (which is also an upcoming feature), but i think we can think about it a bit more.

jimmygchen (approved):
LGTM!

beacon_node/beacon_processor/src/lib.rs
jimmygchen: Addressed in https://github.com/sigp/lighthouse/pull/7588/commits/719df16df6a74c54050b160db90ef92f6dbc6edd

PR #7593, opened 2025-06-12
Advertise `--advertise-false-custody-group-count` for testing PeerDAS

jimmygchen:
## Issue Addressed

#6973

### DIFF
--- a/beacon_node/lighthouse_network/src/service/utils.rs
+++ b/beacon_node/lighthouse_network/src/service/utils.rs
@@ -1,6 +1,5 @@
 use crate::multiaddr::Protocol;
-use crate::rpc::methods::MetaDataV3;
-use crate::rpc::{MetaData, MetaDataV1, MetaDataV2};
+use crate::rpc::{MetaData, MetaDataV2, MetaDataV3};
 use crate::types::{EnrAttestationBitfield, EnrSyncCommitteeBitfield, GossipEncoding, GossipKind};
 use crate::{GossipTopic, NetworkConfig};
 use futures::future::Either;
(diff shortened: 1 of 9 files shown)


beacon_node/network/src/service.rs
```diff
@@ -767,7 +767,14 @@ impl<T: BeaconChainTypes> NetworkService<T> {
...
                     .subscribe_new_data_column_subnets(sampling_count);
-                self.libp2p.update_enr_cgc(new_custody_group_count);
+                if self
+                    .network_globals
+                    .config
+                    .advertise_false_custody_group_count
+                    .is_none()
+                {
+                    self.libp2p.update_enr_cgc(new_custody_group_count);
+                }
```
dknopik, line 777: Why do we not set the false count instead?

jimmygchen: The false count is set on startup, and this condition prevents it from getting updated and dong an extra write to disk

dknopik: Of course, no need to update it then.

beacon_node/lighthouse_network/src/service/utils.rs
```diff
@@ -165,38 +164,41 @@ pub fn strip_peer_id(addr: &mut Multiaddr) {
...
                         || persisted_metadata.syncnets != meta_data.syncnets
+                        || persisted_metadata.custody_group_count != meta_data.custody_group_count
                     {
                         meta_data.seq_number += 1;
                     }
                     debug!("Loaded metadata from disk");
                 }
                 Err(_) => {
-                    match MetaDataV1::<E>::from_ssz_bytes(&metadata_ssz) {
+                    match MetaDataV2::<E>::from_ssz_bytes(&metadata_ssz) {
```
dknopik, line 199: Should we fall back to V1 if this also fails? I guess it is reasonable to assume there are no more V1's around

jimmygchen: Yep! that was the assumption - this only happens if user upgrades from a LH version thats multiple major version behind to the latest version, which there's probably no reason to support.

dknopik (approved):
lgtm

eserilev (approved):

PR #7595, opened 2025-06-12
Fix flaky test_rpc_block_reprocessing

ackintosh:
## Issue Addressed

The test occasionally fails, likely because the 10ms fixed delay after block processing isn't insufficient when the system is under load.

https://github.com/sigp/lighthouse/pull/7522#issuecomment-2914595667

## Proposed Changes

Replace single assertion with retry loop.


jimmygchen (approved):
LGTM, thanks!

PR #7597, opened 2025-06-13
feat: better error message for light_client/bootstrap endpoint

### LINKED ISSUE #7229: Add better error messaging to the `light_client/bootstrap` endpoint

PoulavBhowmick03:
## Issue Addressed

Fixes #7229


michaelsproul (approved):
Seems fine, thanks!

PR #7598, opened 2025-06-13
Make block available when we have 50% of columns on supernodes

dknopik:
## Proposed Changes

When we, as a supernode, have 50% of columns, we are able to reconstruct the remaining 50%. If we have verified the columns, this operation (theoretically) cannot fail.

So, we can treat the block as available as it is only a matter of time until we have all columns. This PR experimentally imports the block into fork choice when we have 50% of columns.

Additionally, we add some logic to delay responding to get column by root if we do not have all columns in DA checker yet. But that one might worsen network conditions compared to just returning partial or no columns.

This hasn't been thoroughly tested yet

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -121,14 +121,14 @@ use std::io::prelude::*;
 use std::marker::PhantomData;
 use std::sync::Arc;
 use std::time::Duration;
-use tokio::{pin, select};
-use tokio::time::sleep;
 use store::iter::{BlockRootsIterator, ParentRootBlockIterator, StateRootsIterator};
 use store::{
     BlobSidecarListFromRoot, DatabaseBlock, Error as DBError, HotColdDB, HotStateSummary,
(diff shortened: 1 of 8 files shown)


jimmygchen:
We should also consider this for v8.0.0 - however need to be careful with this optimisation - because if we attest to the block and not able to serve column data, we might get penalised by peers for this.

jimmygchen:
> Additionally, we add some logic to delay responding to get column by root if we do not have all columns in DA checker yet. But that one might worsen network conditions compared to just returning partial or no columns.

Yeah there is a trade off here - if we vote for the block and not able to serve the data, we might be severely penalised by other peers.

On the other hand, if we queue the by root requests, it's hurts other peers slightly if they try to retrieve columns during the attestation window.

Need to think a bit more about this. Right now this isn't a big issue as we've delayed reconstruction, so getting 100% of columns doesn't seem to take that much extra time in most cases.

jimmygchen:
Removed 8.0.0 as it's not strictly mandatory for the initial fusaka fork with the propose blob count. Will consider it if we see any bottleneck.

PR #7611, opened 2025-06-16
Send byrange responses in the correct requested range

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

For responding to by_range requests , we should ideally only respond with items in the range `req.start_slot()..req.start_slot() + req.count`. 
We were not filtering the generated response for blobs and data columns, only for blocks. This PR adds the filtering for the sidecars as well.

### DIFF
--- a/beacon_node/network/src/network_beacon_processor/rpc_methods.rs
+++ b/beacon_node/network/src/network_beacon_processor/rpc_methods.rs
@@ -945,12 +945,18 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
             match self.chain.get_blobs(&root) {
                 Ok(blob_sidecar_list) => {
                     for blob_sidecar in blob_sidecar_list.iter() {
-                        blobs_sent += 1;
-                        self.send_network_message(NetworkMessage::SendResponse {
-                            peer_id,
-                            inbound_request_id,
-                            response: Response::BlobsByRange(Some(blob_sidecar.clone())),
-                        });
+                        // Due to skip slots, blobs could be out of the range, we ensure they
+                        // are in the range before sending
+                        if blob_sidecar.slot() >= request_start_slot
+                            && blob_sidecar.slot() < request_start_slot + req.count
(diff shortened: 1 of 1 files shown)


michaelsproul:
I wonder if there's a test for this that we could add? Do we have tests for the RPC methods directly?

pawanjay176:
> I wonder if there's a test for this that we could add? Do we have tests for the RPC methods directly?

I looked for any available testing infra around this but couldn't find any. I think we have a bunch of tests in sync but they are pretty isolated to that crate.

michaelsproul:
Does the error log that you were getting happen as part of normal operation? Maybe we could try to detect it from the simulator?

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -945,12 +945,18 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
-                            peer_id,
-                            inbound_request_id,
-                            response: Response::BlobsByRange(Some(blob_sidecar.clone())),
-                        });
+                        // Due to skip slots, blobs could be out of the range, we ensure they
+                        // are in the range before sending
+                        if blob_sidecar.slot() >= request_start_slot
+                            && blob_sidecar.slot() < request_start_slot + req.count
+                        {
+                            blobs_sent += 1;
```
michaelsproul, line 953: Related, the debug message we log about `blobs_sent` has always seemed wrong to me. In the case of skipped slots it's impossible to achieve `blocks_sent == req_count`, but the message we log kind of smells like an error:

https://github.com/sigp/lighthouse/blob/0d311dfc3cfefe33371694c93e086acae716ed1a/beacon_node/network/src/network_beacon_processor/rpc_methods.rs#L631-L639

I wonder if we should get rid of that `msg` seeing as it's kind of misleading.

michaelsproul (approved):
Let's merge this as-is for v7.1.0

PR #7612, opened 2025-06-16
docs: update broken link

JacksonEi:
## Issue Addressed

Fixes a broken link in the FAQ documentation related to submitting a voluntary exit message without running a beacon node.

## Proposed Changes

- Updated the outdated link to the voluntary exit guide:
  - **Old link**: `https://github.com/eth-educators/ethstaker-guides/blob/main/voluntary-exit.md`
  - **New link**: `https://github.com/eth-educators/ethstaker-guides/blob/main/docs/voluntary-exit.md`

#### before
before
#### after
after

This change ensures that users are redirected to the correct and current guide on how to use `ethdo` to submit a voluntary exit message. No other content or logic has been modified.

### DIFF
--- a/book/src/faq.md
+++ b/book/src/faq.md
@@ -323,7 +323,7 @@ Another possible reason for missing the head vote is due to a chain "reorg". A r
 
 ### <a name="vc-exit"></a> Can I submit a voluntary exit message without running a beacon node?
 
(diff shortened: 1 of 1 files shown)


macladson (commented):
This has actually been fixed already on `unstable`: https://github.com/sigp/lighthouse/blob/unstable/book/src/faq.md?plain=1#L243C337-L243C419

When we do our next release this fix will be present in `stable`

And actually, our book itself is built from `unstable` so if you go to [the actual page in question](https://lighthouse-book.sigmaprime.io/faq.html) you'll see that the link actually does work.

PR #7614, opened 2025-06-16
feat: Add reproducible builds release workflows and push images to DockerHub 

### LINKED ISSUE #7485: Test reproducible build correctness
We are adding reproducible builds in:

However we do not test that the builds actually reproduce across different runs.

We could add a job to CI that runs periodically and compiles a reproducible binary twice, and then verifies that the two builds are identical.

### LINKED ISSUE #7486: Add GitHub action to publish docker reproducible builds
Now that we have a reproducible Docker build implemented in #6799, we’d like to extend the release workflow to:

* Build the Docker images as part of the release workflow
* Publish the images to DockerHub

~~This should replace the current Docker image builds, which aren't built via a reproducible process.~~

^ For the time being, we could consider publishing a **separate** `reproducible` image, because the new profile has some properties that we may not want in a standard build: 
- `panic = "abort"`: we lose crash info, ideally we remove this - but not sure if this is necessary for reproducible builds.
- `overflow-checks = true`: this is not usually on, so it should be tested.

https://github.com/sigp/lighthouse/blob/2b398f03c2f3f42b3f76b64f4ee71ee7d9f7ab4d/Cargo.toml#L301-L306

Eventually it would be nice to consolidate, but we'd need some further testing.

MoeMahhouk:
## Issue Addressed
This pull request introduces workflows and updates to ensure reproducible builds for the Lighthouse project. It adds two GitHub Actions workflows for building and testing reproducible Docker images and binaries, updates the `Makefile` to streamline reproducible build configurations, and modifies the `Dockerfile.reproducible` to align with the new build process. Additionally, it removes the `reproducible` profile from `Cargo.toml`.

### New GitHub Actions Workflows:

* [`.github/workflows/docker-reproducible.yml`](diffhunk://#diff-222af23bee616920b04f5b92a83eb5106fce08abd885cd3a3b15b8beb5e789c3R1-R145): Adds a workflow to build and push reproducible multi-architecture Docker images for releases, including support for dry runs without pushing an image.

### Build Configuration Updates:

* [`Makefile`](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L85-R143): Refactors reproducible build targets, centralizes environment variables for reproducibility, and updates Docker build arguments for `x86_64` and `aarch64` architectures.
* [`Dockerfile.reproducible`](diffhunk://#diff-587298ff141278ce3be7c54a559f9f31472cc5b384e285e2105b3dee319ba31dL1-R24): Updates the base Rust image to version 1.86, removes hardcoded reproducibility settings, and delegates build logic to the `Makefile`.
* Switch to using jemalloc-sys from Debian repos instead of building it from source. A Debian version is [reproducible](https://tests.reproducible-builds.org/debian/rb-pkg/trixie/amd64/jemalloc.html) which is [hard to achieve](https://github.com/NixOS/nixpkgs/issues/380852) if you build it from source.

### Profile Removal:

* [`Cargo.toml`](diffhunk://#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542L289-L295): Removes the `reproducible` profile, simplifying build configurations and relying on external tooling for reproducibility.

## Additional Info
This is mainly a follow up to this work #6799 where I refine the reproducible build configuration to simplify the CI workflow to generate the reproducible images and pushes them to DockerHub. I also added a cron job workflow (inspired from the Reth repo) that checks every two days or pull requests that touches files that might affect reproducibility to catch potential regressions.
In case, this is too much, let me know and I can create a separate PR for this to be merged later when necessary

close #7486
close #7485


chong-he:
Doing some testing on this, will post the comment when ready

.github/workflows/reproducible-build.yml
```diff
@@ -0,0 +1,188 @@
...
+        run: |
+          make build-reproducible \
+            RUST_TARGET=x86_64-unknown-linux-gnu
+          cp target/x86_64-unknown-linux-gnu/release/lighthouse \
+            lighthouse-build-1
+          sha256sum lighthouse-build-1 > lighthouse-build-1.sha256
+
+      - name: Clean build artifacts and cache
+        run: |
+          make clean
```
chong-he: Currently this causes error in the CI: https://github.com/sigp/lighthouse/actions/runs/15705132140/job/44642272289?pr=7614

We can edit this line: https://github.com/sigp/lighthouse/blob/999b04517e35f8fce78525132d0b8e0a4816f31e/testing/state_transition_vectors/Makefile#L8

to become `rm -rf vectors/` to force remove

.github/workflows/reproducible-build.yml
```diff
@@ -0,0 +1,188 @@
...
+      - uses: actions/checkout@v4
+
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          target: aarch64-unknown-linux-gnu
+
+      - name: Install build dependencies and cross-compilation tools
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libclang-dev cmake gcc-aarch64-linux-gnu
```
chong-he: This wouldn't compile for now: https://github.com/sigp/lighthouse/actions/runs/15705132140/job/44642272292?pr=7614

I have tested it in my local branch and added some dependencies, you can refer to: https://github.com/chong-he/lighthouse/pull/8

.github/workflows/reproducible-build.yml
```diff
@@ -0,0 +1,188 @@
...
+        run: |
+          echo "=== Build 1 SHA256 (ARM64) ==="
+          cat lighthouse-build-1-arm64.sha256
+          echo "=== Build 2 SHA256 (ARM64) ==="
+          cat lighthouse-build-2-arm64.sha256
+          echo "=== Binary Comparison ==="
+          if cmp lighthouse-build-1-arm64 lighthouse-build-2-arm64; then
+            echo "✅ ARM64 binaries are identical - reproducible build PASSED"
+          else
+            echo "❌ ARM64 binaries differ - reproducible build FAILED"
```
chong-he: In my testing, the reproducible for ARM64 is failing as the SHA256 for both builds are different (x86_64 passes): https://github.com/chong-he/lighthouse/actions/runs/15815928865/job/44574695858

.github/workflows/release-reproducible.yml
```diff
@@ -0,0 +1,142 @@
+name: release-reproducible
```
chong-he: Some comments: 

1. From my understanding this is about pushing the docker image, without checking if it is reproducible? Would it be better if we can add a condition such as 
"if reproducible-build is successful, then only push this docker image"? 

2. In my testing, the ARM64 docker image is taking ~5 hours to build: https://github.com/chong-he/lighthouse/actions/workflows/release-reproducible.yml
I am not sure if it is because of my local repo, or due to the runner being slow, or due to configs in the file

3. Since this is also pushing docker images, would it be better to put under: https://github.com/sigp/lighthouse/blob/unstable/.github/workflows/docker.yml
We can also modify some parameters to be consistent with other docker builds, for example, the OS that's running on (pinned `ubuntu-22.04` or `ubuntu-latest`), `docker/build-push-action@v5` or `v6`

.github/workflows/reproducible-build.yml
```diff
@@ -0,0 +1,188 @@
...
+            lighthouse-build-1
+          sha256sum lighthouse-build-1 > lighthouse-build-1.sha256
+
+      - name: Clean build artifacts and cache
+        run: |
+          make clean
+          cargo cache -a
+          rm -rf target/
+
+      - name: Build Lighthouse (second build)
```
chong-he: The two builds in CI will take about 40 mins to complete, which could be time consuming for every PR to run. One option is to run theese CI checks when there is a new release. We could add a condition like the `syncing` test: 
https://github.com/sigp/lighthouse/blob/fd643c310c4e831821a18aca5c44038e661de86e/.github/workflows/local-testnet.yml#L178
to only trigger this reproducible build checks by having a label that's tagged with a release, for example, a label that starts with `v`

chong-he (commented):
Thanks for the PR. I left some comments after doing some testing

MoeMahhouk:
> Thanks for the PR. I left some comments after doing some testing

Thanks for reviewing it.
I will do some rework and refinements to this PR to fix the issues and address your comments.

.github/workflows/release-reproducible.yml
MoeMahhouk: >1. From my understanding this is about pushing the docker image, without checking if it is reproducible? Would it be better if we can add a condition such as
"if reproducible-build is successful, then only push this docker image"? 

You are correct, I left that out in case you don't want to block docker container releases based on reproducibility but we can for sure add a script that tests the reproducibility similar to reproducible build workflow and only then push the resulting artifacts once successful to dockerhub

>2. In my testing, the ARM64 docker image is taking ~5 hours to build: https://github.com/chong-he/lighthouse/actions/workflows/release-reproducible.yml
I am not sure if it is because of my local repo, or due to the runner being slow, or due to configs in the file

Yes, the github runner seems to be x86 architecture, so the docker build for ARM64 is being emulated by qemu and therefore very slow compared to building it on a native ARM architecture machine which takes something like 20  mins similar to the x86 build on a x86 architecture hardware. Do you have suggestions or options to customize/select the runner machine architecture for ARM builds?
Edit: I found out today that github is offering a [preview version of arm runners](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/). I am trying them now and seems to be working. It build the binary within docker build in less than 20mins similar to the x86 based runner

>3. Since this is also pushing docker images, would it be better to put under: https://github.com/sigp/lighthouse/blob/unstable/.github/workflows/docker.yml
We can also modify some parameters to be consistent with other docker builds, for example, the OS that's running on (pinned ubuntu-22.04 or ubuntu-latest), docker/build-push-action@v5 or v6

I'd say that is your call to make, I separated them here to be targeting reproducible docker container images instead than a normal docker image. Hence, the version pinning of all dependencies and the specific docker image hash we are building with to avoid any factor that might impact the reproducibility of the build process.

.github/workflows/reproducible-build.yml
MoeMahhouk: I will add that, I was testing locally with `make clean || true` to by-pass it but I'll apply your suggestion for more clean approach

.github/workflows/reproducible-build.yml
MoeMahhouk: I can remove the on pull requests and only leave the cronjob which runs it individually every 2 days and giving you insights in-case something broke reproducibility, wdyt?
We could also add the condition like syncing in case you still would like to have this be triggered for tagged pull requests.

.github/workflows/reproducible-build.yml
MoeMahhouk: yeah it is probably missing the `g++-aarch64-linux-gnu` as it might be needed for cross compilation for ARM architecture. I will experiment more on seeing what makes it build sucessfully

.github/workflows/reproducible-build.yml
MoeMahhouk: Yes, something I couldn't grasp on why the github runner is having some non-deterministic factors affecting reproducibility while running this locally does not cause different hashes.
I will bypass this issue by building those binaries within the reproducible docker that we have now and extract the binaries from those builds to compare. So far it seems more successful and having clean build environment without potential caching residues affecting reproducibility.

.github/workflows/release-reproducible.yml
chong-he: Reply to:
1. Makes sense, not sure what is the best way, we will hear from the devs about this
2. Good to know!
3. will hear from the devs too

.github/workflows/reproducible-build.yml
chong-he: Yes, I can see that it passes the CI now, for the reproducible-build

.github/workflows/reproducible-build.yml
chong-he: I see that the PR as of now is passing the CI for reproducible builds

.github/workflows/reproducible-build.yml
chong-he: Had a discussion with @michaelsproul , he thinks it will be good to have both:
- a cronjob that runs every 2 days so that if anything breaks it can be caught earlier as you said
- a condition of labels to only run with certain labels, to catch anything before a release

That should be pretty safe to ensure the release is reproducible, yet if anything happens that breaks it, it can be caught earlier

.github/workflows/release-reproducible.yml
chong-he: For no. 1, it would be good to have a condition that checks the reproducibility before pushing. 
For no. 3, after some thinking, I think having a separate file is a better way, so no need to put under the `docker.yml` file

MoeMahhouk: 1. I will adjust it shortly to check the reproducibility prior pushing the images.
3. Sounds good to me, I will leave them separate as it is right now

.github/workflows/reproducible-build.yml
MoeMahhouk: alright, I'll add extra check for `test-reproducibility` label alongside the cronjob to enable checking reproducibility upon triggering it with those either options

chong-he:
> > Thanks for the PR. I left some comments after doing some testing
> 
> Thanks for reviewing it. I will do some rework and refinements to this PR to fix the issues and address your comments.

Is there a reason why is this PR still a draft?

MoeMahhouk:
> > > Thanks for the PR. I left some comments after doing some testing
> > 
> > 
> > Thanks for reviewing it. I will do some rework and refinements to this PR to fix the issues and address your comments.
> 
> Is there a reason why is this PR still a draft?

Not really, I am waiting for your final feedback to open it for review/merge. 
I will open it now for review and if you find anything needed for iteration, we can switch it back again.

.github/workflows/release-reproducible.yml
chong-he: 1. I can see that there is a check before pushing the docker images, thanks! 
2. resolved using the arm runner now: `runner: ubuntu-24.04-arm`
3. Not relevant anymore

Thank you!

.github/workflows/reproducible-build.yml
chong-he: cron job is added `cron: "0 1 */2 * *"` and the condition of label is added as well: `contains(github.event.pull_request.labels.*.name, 'test-reproducible')`

Thank you

Makefile
```diff
@@ -82,36 +82,65 @@ build-lcli-aarch64:
...
+
+# Disable incremental compilation to avoid non-deterministic artifacts
+CARGO_INCREMENTAL_VAL = 0
+# Set C locale for consistent string handling and sorting
+LOCALE_VAL = C
+# Set UTC timezone for consistent time handling across builds
+TZ_VAL = UTC
+
+# Default features for lighthouse
+FEATURES ?= gnosis,slasher-lmdb,slasher-mdbx,slasher-redb,jemalloc
```
chong-he: I think this breaks the `cli-check` due to this: https://github.com/chong-he/lighthouse/pull/8/commits/824a5dc735ec0babc9c7e4bdead18b3ef22aa461

I think we still support `mdbx` but it's going to be removed some time? We will wait for the devs to comment on this

.github/workflows/release-reproducible.yml
```diff
@@ -0,0 +1,253 @@
...
+env:
+  DOCKER_REPRODUCIBLE_IMAGE_NAME: >-
+    ${{ github.repository_owner }}/lighthouse-reproducible
+  DOCKER_PASSWORD: ${{ secrets.DH_KEY }}
+  DOCKER_USERNAME: ${{ secrets.DH_ORG }}
+
+jobs:
+  extract-version:
+    name: extract version
+    runs-on: ubuntu-latest
```
chong-he: Should we stick to `ubuntu-22.04` here? The `docker.yml` is also running on this now: 
https://github.com/sigp/lighthouse/blob/999b04517e35f8fce78525132d0b8e0a4816f31e/.github/workflows/docker.yml#L29

I somehow recall that we usually go with the previous version of OS for maximal compatibility

.github/workflows/reproducible-build.yml
```diff
@@ -0,0 +1,245 @@
...
+          else
+            echo "❌ FAILED: Docker-built binaries differ"
+            echo "First 10 differences:"
+            cmp -l lighthouse-build-1 lighthouse-build-2 | head -10
+            exit 1
+          fi
+
+  build-aarch64:
+    name: test reproducible builds (aarch64)
+    runs-on: ubuntu-24.04-arm
```
chong-he: Similar comment as above, wondering should we use `ubuntu-22.04-arm` instead?

chong-he (commented):
I have tested this and it is working. Would be great to get some comments from the devs

I have added a label `test-reproducible`, that will trigger the reproducible builds CI to run. When we have release PR, we can add this label 
to run the check and ensure that the binaries are reproducible.

.github/workflows/release-reproducible.yml
MoeMahhouk: I have no strong opinion about this. I will defer this decision to you

.github/workflows/reproducible-build.yml
MoeMahhouk: same as my answer above

Makefile
MoeMahhouk: what do you suggest to change it into? I am open for any suggestions that suits the purpose

chong-he: We will wait for a review from @michaelsproul when he has the time

MoeMahhouk:
@chong-he , I just wanted to double check if this is waiting on me for anything else so far?

chong-he:
> @chong-he , I just wanted to double check if this is waiting on me for anything else so far?

No 

The PR is now waiting for other team members to review

jimmygchen:
Re-opening this PR as it was awaiting review.

antondlr:
I like this idea but would like it to be less maintenance-intensive. 
One thing that immediately stands out: 
we should not have granular versions like `RUST_IMAGE="rust:1.86-bullseye@sha256:1110399f568f1dbe838e58f15b4162d899cb95f450f5f0ffa739614f3a4c32f1` in the workflow/dockerfile but rather provide them as parameters.

MoeMahhouk:
> I like this idea but would like it to be less maintenance-intensive. One thing that immediately stands out: we should not have granular versions like `RUST_IMAGE="rust:1.86-bullseye@sha256:1110399f568f1dbe838e58f15b4162d899cb95f450f5f0ffa739614f3a4c32f1` in the workflow/dockerfile but rather provide them as parameters.

good idea, could you provide an example on that suggestion? do you have a reference link or would you like to take that up and refine it to be a parameter instead?

MoeMahhouk:
Hey folks, @chong-he @jimmygchen ,this PR got auto-closed due to inactivity, but the changes are current and needs review. Could we reopen this for another pass?

chong-he:
> Hey folks, @chong-he @jimmygchen ,this PR got auto-closed due to inactivity, but the changes are current and needs review. Could we reopen this for another pass?

Reopening, could you solve some conflicts in the branch? 
I will try to give this another review

MoeMahhouk:
> > Hey folks, @chong-he @jimmygchen ,this PR got auto-closed due to inactivity, but the changes are current and needs review. Could we reopen this for another pass?
> 
> Reopening, could you solve some conflicts in the branch? I will try to give this another review

solved the conflicts and merged. Let me know if you want me to do any other change

Makefile
chong-he: `jemalloc` is now the default

```suggestion
FEATURES ?= gnosis,slasher-lmdb,slasher-mdbx,slasher-redb
```

.github/workflows/reproducible-build.yml
```diff
@@ -0,0 +1,245 @@
...
+        uses: docker/setup-buildx-action@v3
+        with:
+          driver: docker  # Use docker driver instead of docker-container for reproducibility
+
+      - name: Build first reproducible Docker image
+        run: |
+          echo "=== Building first Docker image (x86_64) ==="
+          docker build -f Dockerfile.reproducible \
+            --build-arg RUST_TARGET="x86_64-unknown-linux-gnu" \
+            --build-arg RUST_IMAGE="rust:1.88-bullseye@sha256:8e3c421122bf4cd3b2a866af41a4dd52d87ad9e315fd2cb5100e87a7187a9816" \
```
chong-he: I think we can remove this line and instead just let it read from `Dockerfile.reproducible` for the RUST_IMAGE. I have tested it on my local repo and it works, see: https://github.com/chong-he/lighthouse/pull/9/files#diff-587298ff141278ce3be7c54a559f9f31472cc5b384e285e2105b3dee319ba31d

This also applies to all other instances that uses this `RUST_IMAGE`. Taking the idea from `Dockerfile`:
https://github.com/sigp/lighthouse/blob/e3ee7febce64c1b5a85c3ab0be0619571ee92d58/Dockerfile#L1

We can also remove the sha256 in the RUST_IMAGE in `Dockerfile.reproducible`. That will make it much easier to maintain. 

We also do not need a separate image on x86_64 or arm64 from my understanding, as docker is able to select the image based on the host: https://docs.docker.com/build/building/multi-platform/#difference-between-single-platform-and-multi-platform-images, so that will make things simpler

```suggestion

```

.github/workflows/reproducible-build.yml
```diff
@@ -0,0 +1,245 @@
...
+            echo "✅ SUCCESS: Docker-built binaries are identical!"
+            echo "✅ Reproducible Docker build PASSED for aarch64"
+          else
+            echo "❌ FAILED: Docker-built binaries differ"
+            echo "First 10 differences:"
+            cmp -l lighthouse-build-1-arm64 lighthouse-build-2-arm64 | head -10
+            exit 1
+          fi
+
+  summary:
```
chong-he: I would like to suggest to remove this part to make things simpler. 

I think it is sufficient to have 2 additional CIs for reproducible, one for x86_64, another one for arm64. If both passes then good, if any is failing, then we can look at the CI itself, so this `summary` job is not necessary imho

.github/workflows/release-reproducible.yml
```diff
@@ -0,0 +1,253 @@
+name: release-reproducible
```
chong-he: I would suggest to name this as `docker-reproducible` and also the same for the file name, so make it distinct right from the name

.github/workflows/docker-reproducible.yml
```diff
@@ -0,0 +1,253 @@
...
+        uses: docker/setup-buildx-action@v3
+        with:
+          driver: docker
+
+      - name: Verify reproducible builds (${{ matrix.arch }})
+        run: |
+          echo "🔄 Verifying reproducible builds for ${{ matrix.arch }}..."
+
+          # Build first image
+          echo "=== Building first verification image ==="
```
chong-he: I know I raised this before to have a reproducible check before publishing the image. I am sorry, thinking about it now, we already have the reproducible-CI checks, and this seems repetitive now. In the interest of making things simpler, should we remove the check? I am trying to make the changes as minimal as possible, so that it is easier to read and maintain. After all, the checks for reproducibility is running every 2 days so if anything, we can catch it earlier.

I am just considering about the consequence - if without the checks and the docker images are pushed, then it could be that the docker images are not reproducible in the worse outcome?

Dockerfile.reproducible
```diff
@@ -5,40 +5,21 @@ FROM ${RUST_IMAGE} AS builder
...
 
-RUN mv /app/target/${RUST_TARGET}/${PROFILE}/lighthouse /lighthouse
+# Move the binary to a standard location
+RUN mv /app/target/${RUST_TARGET}/release/lighthouse /lighthouse
 
 # Create a minimal final image with just the binary
 FROM gcr.io/distroless/cc-debian12:nonroot-6755e21ccd99ddead6edc8106ba03888cbeed41a
 COPY --from=builder /lighthouse /lighthouse
+
+EXPOSE 30303 30303/udp 9001 8545 8546
```
chong-he: I am not sure if this is necessary? The ports 8545, 8546, 30303 are for the EL

Makefile
```diff
@@ -81,36 +81,65 @@ build-lcli-aarch64:
...
-# Reproducible build for x86_64
-build-reproducible-x86_64:
+.PHONY: build-reproducible
+build-reproducible: ## Build the lighthouse binary into `target` directory with reproducible builds
+	SOURCE_DATE_EPOCH=$(SOURCE_DATE) \
+	RUSTFLAGS="${RUST_BUILD_FLAGS} --remap-path-prefix $$(pwd)=." \
+	CARGO_INCREMENTAL=${CARGO_INCREMENTAL_VAL} \
+	LC_ALL=${LOCALE_VAL} \
+	TZ=${TZ_VAL} \
+	cargo build --bin lighthouse --features "$(FEATURES)" --profile "$(PROFILE)" --locked --target $(RUST_TARGET)
```
chong-he: Not sure about this. Do we need this, given that we have 1 for x86_64 and 1 for arm64 below?

I see that the `build-reproducible-all` below also don't call this `build-reproducible`

chong-he (dismissed):
Look at this a second time and try to suggest some simplifications so that it is easier to maintain and make the code cleaner. 

Also, I would prefer to not have the symbol (green tick, cross etc) to keep the code clean.

Makefile
chong-he: For simplicity, we could probably use the `CROSS_FEATURES` that's already defined in the `Makefile`? or is there something in `CROSS_FEATURES` that make the reproducible invalid?

.github/workflows/release-reproducible.yml
chong-he: For consistency, let's use Ubuntu 22.04 here to be the same with docker.yml file

.github/workflows/docker-reproducible.yml
MoeMahhouk: we can remove the check but then we would have the scenario that the reproducible-build CI might break but this workflow still builds and publishes a "reproducible" docker build despite being broken.
Do you still want to remove the reproducibility check?

Yes, I recall we discussed this before and decided to add this check here despite repetitiveness to avoid such misleading scenario where shipped reproducible docker builds aren't actually reproducible

.github/workflows/release-reproducible.yml
MoeMahhouk: Sure, I can switch all to ubuntu 22.04 and even the arm one below

.github/workflows/release-reproducible.yml
MoeMahhouk: sure thing, I will rename the workflow and file accordingly

.github/workflows/docker-reproducible.yml
```diff
@@ -0,0 +1,253 @@
...
+            echo "- ✅ No build artifacts tampering"
+            echo "- ✅ Deterministic build process"
+          elif [[ "${{ github.event.inputs.dry_run }}" == "true" ]]; then
+            echo "🧪 **Overall**: Dry run completed successfully!"
+            echo "- Reproducibility verified ✅"
+            echo "- Ready for real release ✅"
+          else
+            echo "🚨 **Overall**: Release failed or incomplete"
+            echo "- Check logs above for details"
+          fi
```
MoeMahhouk: @chong-he do you also want me to remove the release summary here too?

.github/workflows/reproducible-build.yml
MoeMahhouk: I am pinning the sha256 for better tracking and reproducibility stability because if any dependency gets updated for that specific image version of rust:1.88, it would also changes its hash history and eventually also impact the reproducibility of the generated lighthouse binary.
If you are ok to proceed with such trade-off, I can remove the sha256 pinning to the rust image.

Regarding removing the whole build-arg, I believe that would be necessary since we are not passing the --platform to the docker build command for the image to identify which one to select. 
Nonetheless, if you decide to keep the sha256 for stable reproducibility tracking, then I believe we can't remove this line as it wouldn't be able to detect which sha256 to fetch for which platform, right?

.github/workflows/reproducible-build.yml
MoeMahhouk: alright, I will remove the summary job for the sake of making stuff simpler

Dockerfile.reproducible
MoeMahhouk: Sure, which ones do you recommend to replace them with?

Makefile
MoeMahhouk: Sure we can try using the cross features instead

Makefile
MoeMahhouk: yes, this one is being used with the new revamped Dockerfile.reproducible.
The other two below is for developer purposes if one would like to trigger the build locally manually

Dockerfile.reproducible
chong-he: I think we can just delete this line, can't we? When running the image, users will usually specify the port themselves, so this feels redundant

.github/workflows/docker-reproducible.yml
chong-he: yes I think we can remove the release summary here as well, thanks

Makefile
```diff
@@ -81,36 +81,62 @@ build-lcli-aarch64:
...
 
-# extracts the current source date for reproducible builds
-SOURCE_DATE := $(shell git log -1 --pretty=%ct)
+# Environment variables for reproducible builds
+# Initialize RUSTFLAGS
+RUST_BUILD_FLAGS =
+# Remove build ID from the binary to ensure reproducibility across builds
+RUST_BUILD_FLAGS += -C link-arg=-Wl,--build-id=none
+# Remove metadata hash from symbol names to ensure reproducible builds
+RUST_BUILD_FLAGS += -C metadata=''
```
chong-he, line 90: ~~why do we need 3 `RUST_BUILD_FLAGS` here?~~

(I had this comment in the same time as 2 days ago but probably I forgot to click to publish the comment)

Edit: ignore the above, @michaelsproul mentioned this is ok

Makefile
chong-he: I see it now, looks good, thanks

.github/workflows/docker-reproducible.yml
```diff
@@ -0,0 +1,253 @@
+name: docker-reproducible
+
+on:
+  push:
+    tags:
+      - v*
```
michaelsproul, line 9: Could build this image under the same conditions as the regular docker image

```yaml
on:
    push:
        branches:
            - unstable
            - stable
        tags:
            - v*
```

And then we don't need `reproducible-build` as a separate job (which is highly duplicated anyway)

.github/workflows/docker-reproducible.yml
```diff
@@ -0,0 +1,253 @@
...
+    inputs:
+      dry_run:
+        description: >-
+          Enable dry run mode (builds images but skips push to registry)
+        type: boolean
+        default: false
+
+env:
+  DOCKER_REPRODUCIBLE_IMAGE_NAME: >-
+    ${{ github.repository_owner }}/lighthouse-reproducible
```
michaelsproul, line 14: Should do:

- `sigp/lighthouse-reproducible:vX.X.X` (for releases)
- `sigp/lighthouse-reproducible:latest` (for `stable` branch)
- `sigp/lighthouse-reproducible:latest-unstable` (for `unstable` branch)

.github/workflows/reproducible-build.yml
chong-he: I got what you mean. Referring to Michael' s comments below, we can just have the checks on docker-reproducible and don't need this file. This makes things much simpler

.github/workflows/docker-reproducible.yml
chong-he: Refer to comments from Michael's below

.github/workflows/docker-reproducible.yml
MoeMahhouk: we can do that in case you want to build and push to dockerhub a new image each time you push to the branches unstable/stable.
I agree that there is redundancy between this and reproducible-build. However, the latter runs periodically (every 2 days) and continuously checks if any new changes broke the reproducibility without building a docker image for each architecture and pushing it to dockerhub. 

Do you still want to remove reproducible-build and make it push a new image to dockerhub for each push to stable/unstable branches beside the release tags? @michaelsproul

.github/workflows/docker-reproducible.yml
MoeMahhouk: this is currently being adjusted in the final tag [below](https://github.com/sigp/lighthouse/pull/7614/files#diff-c338e5b32454c5245be015c80ac292a8bb5a3e6a9fe48bf2b1ba91f08e63fd8aR116). 

If you answer my question above with yes, we can add a way to identify those cases and put those labels in the final image name tag

Makefile
MoeMahhouk: yes, the short answer is that I extracted them out of the Dockerfile.reproducible file to make it cleaner

.github/workflows/docker-reproducible.yml
chong-he: yes, we think removing the `reproducible-build` is ok, we don't need the CI checks if the docker already checks it before publishing. 

So each time a merge is done to the stable or unstable, the docker images will be updated. If the reproducible is failing, then we can catch it earlier too.

Does the potentially frequent update of the docker images to unstable a concern for you or the use cases? If so, the stable reproducible can be the one to use

MoeMahhouk: alright, then if you are ok with building and publishing docker images for each push to stable/unstable too, then we can do it like that too.

The periodic CI workflow was only for the case to catch breaking changes early on. But now if you decide to add stable/unstable branch pushes, then it is not needed anymore. The only caveat I can think of now is that this might have slight time delay to shipping until this workflow passes but it shouldnt be a blocker either.

I will adjust accordingly and remove reproducible-builds.yml

MoeMahhouk:
I believe I addressed all your feedback/comments and also added reproducibility fix for jemalloc-sys which is used to be not reproducible on different host machines.
I kindly ask your review again on the changes as well as the responses to the comments if you are ok with the current state
@chong-he @michaelsproul

.github/workflows/docker-reproducible.yml
```diff
@@ -0,0 +1,189 @@
...
+        uses: docker/login-action@v3
+        with:
+          username: ${{ env.DOCKER_USERNAME }}
+          password: ${{ env.DOCKER_PASSWORD }}
+
+      - name: Push verified image (${{ matrix.arch }})
+        if: ${{ github.event_name != 'workflow_dispatch' }}
+        run: |
+          VERSION=${{ needs.extract-version.outputs.VERSION }}
+          IMAGE_TAG="${{ env.DOCKER_REPRODUCIBLE_IMAGE_NAME }}:${VERSION}-${{ matrix.arch }}"
```
chong-he, line 136: The published image tag doesn't have the `VERSION_SUFFIX` now (`stable`, `unstable`). Maybe we want to add that, to be in sync with the current naming convention

chong-he (commented):
Just a small comment here. I am unsure about other changes, will need another review from a dev

Can we also remove the echo that are not strictly necessary? For example, 

echo "Published: ${IMAGE_NAME}:${VERSION}"

.github/workflows/docker-reproducible.yml
MoeMahhouk: Can you please elaborate more? I didn't fully get what you mean.
All 3 different types of push follow here the same naming convention if I am not wrong. The version is extracted [here](https://github.com/sigp/lighthouse/pull/7614/files/12c057416da116a7809c29db2f1b7c57443e1a47#diff-c338e5b32454c5245be015c80ac292a8bb5a3e6a9fe48bf2b1ba91f08e63fd8aR19-R41), then they will be named as follow:
- push on stable branch: ${{ github.repository_owner }}/lighthouse-reproducible:latest-{amd64||arm64}
- push on unstable branch: ${{ github.repository_owner }}/lighthouse-reproducible:latest-unstable-{amd64||arm64}
- push with tag: ${{ github.repository_owner }}/lighthouse-reproducible:v{x.y.z}-{amd64||arm64}

Or did I misunderstand something here

MoeMahhouk:
> Just a small comment here. I am unsure about other changes, will need another review from a dev
> 
> Can we also remove the echo that are not strictly necessary? For example,
> 
> echo "Published: ${IMAGE_NAME}:${VERSION}"

I removed the majority of the "echo" statements. If you think any other one is not necessary, let me know and I'll adjust accordingly

.github/workflows/docker-reproducible.yml
chong-he: Sorry you are right. I was testing with only push with tag that's why I don't see the stable or unstable in the naming, my bad.

chong-he:
> > Just a small comment here. I am unsure about other changes, will need another review from a dev
> > Can we also remove the echo that are not strictly necessary? For example,
> > echo "Published: ${IMAGE_NAME}:${VERSION}"
> 
> I removed the majority of the "echo" statements. If you think any other one is not necessary, let me know and I'll adjust accordingly

Thanks. 

We need a review from @michaelsproul when he has time. I am not all that familiar with docker and stuff. I think this PR is now clean enough for a review.

dapplion (approved):

michaelsproul (approved):

Makefile
```diff
@@ -81,36 +81,70 @@ build-lcli-aarch64:
...
+
+# Disable incremental compilation to avoid non-deterministic artifacts
+CARGO_INCREMENTAL_VAL = 0
+# Set C locale for consistent string handling and sorting
+LOCALE_VAL = C
+# Set UTC timezone for consistent time handling across builds
+TZ_VAL = UTC
+
+# Default profile 
+PROFILE ?= release
```
michaelsproul: This is redundant

```suggestion
```

Makefile
```diff
@@ -81,36 +81,67 @@ build-lcli-aarch64:
...
+	TZ=${TZ_VAL} \
+	JEMALLOC_OVERRIDE=${JEMALLOC_OVERRIDE} \
+	cargo build --bin lighthouse --features "$(FEATURES_REPRODUCIBLE)" --profile "$(PROFILE)" --locked --target $(RUST_TARGET)
+
+.PHONY: build-reproducible-x86_64
+build-reproducible-x86_64: ## Build reproducible x86_64 Docker image
 	DOCKER_BUILDKIT=1 docker build \
 		--build-arg RUST_TARGET="x86_64-unknown-linux-gnu" \
 		--build-arg RUST_IMAGE=$(RUST_IMAGE_AMD64) \
-		--build-arg SOURCE_DATE=$(SOURCE_DATE) \
```
michaelsproul, line 95: How come we don't need to plumb the `SOURCE_DATE` in here any more? I don't see how it gets passed to the docker build

michaelsproul: Oh nvm, the docker file calls `make reproducible`. Got it.

michaelsproul:
@mergify requeue

MoeMahhouk:
Thank you for reviewing and merging the PR.
I just noticed that it successfully verified a reproducible build on unstable but it failed due to docker hub token permission issues.
Is the passed token has read-only permissions or did it expire and needs renewal? @michaelsproul

antondlr:
> Thank you for reviewing and merging the PR. I just noticed that it successfully verified a reproducible build on unstable but it failed due to docker hub token permission issues. Is the passed token has read-only permissions or did it expire and needs renewal? @michaelsproul

the ~~namespace~~ repository needs to be initialized first with a manual push, @realbigsean should have the correct permissions to do this

MoeMahhouk:
> > Thank you for reviewing and merging the PR. I just noticed that it successfully verified a reproducible build on unstable but it failed due to docker hub token permission issues. Is the passed token has read-only permissions or did it expire and needs renewal? @michaelsproul
> 
> the ~namespace~ repository needs to be initialized first with a manual push, @realbigsean should have the correct permissions to do this

oh yes, that makes sense. Thank you for confirming and the fast response!

michaelsproul:
I've just created the `sigp/lighthouse-reproducible` project and am re-running the build. Lets see if it works.

michaelsproul:
Didn't work. Are we sure the token isn't a fine-grained token that only works for `sigp/lighthouse`?

antondlr:
Bummer. I don't have the necessary permissions to check, unfortunately. 
I'll rustle some leaves.

realbigsean:
updated the deploy key in this repo with perms to deploy to the new docker repo

antondlr:
[it works! ](https://github.com/sigp/lighthouse/actions/runs/19881990720)
thanks Sean 🫶

PR #7617, opened 2025-06-17
feat: Add reproducible Debian package builds and distribution

### LINKED ISSUE #7618: Add reproducible Debian package builds and distribution
Now that we have reproducible builds and soon CI workflow to build those containers and push them to DockerHub (#7614), the next natural move is to also generate reproducible Debian packages and upload them as release assets.

## Why Debian packages?
Debian packages provide more comfort and would be nice for:

Building VM workloads (e.g., with mkosi)
Bare metal deployments without Docker overhead
Users who prefer system package managers
Integration with the Debian community and its reproducible builds culture (https://reproducible-builds.org/contribute/debian/

### Implementation
Build on existing reproducible build infrastructure using `cargo-deb`, test reproducibility in CI, and upload to GitHub Releases alongside existing artifacts.

Any thoughts?

MoeMahhouk:
## Issue Addressed

This pull request introduces workflows and configuration updates to automate the building, testing, and releasing of reproducible Debian packages for the Lighthouse project. Below are the key changes grouped by theme:

### New GitHub Actions Workflows

* **Debian Package Release Workflow**: Added `.github/workflows/release-deb.yml` to automate the building, testing, and releasing of reproducible Debian packages for multiple architectures (`x86_64` and `aarch64`). Includes steps for dry-run testing, artifact uploads, and checksum generation.

## Proposed Changes

- Added the required configuration for cargo-deb in lighthouse's Cargo.toml.
- Added Makefile targets to build the debian packages for different architectures and test their reproducibility
- Added CI workflow **release-deb.yml** that builds and pushes the debian packages into assets similar to release-reproducible.yml
- Added a simple **lighthouse.service** to be included in the generated lighthouse debian package

## Additional Info

This is an incremental work on https://github.com/sigp/lighthouse/pull/7614 and requires it to be merged first

close #7618


lighthouse/lighthouse.service
```diff
@@ -0,0 +1,34 @@
+[Unit]
+Description=Lighthouse Ethereum Beacon Chain Client
+Documentation=https://lighthouse-book.sigmaprime.io/
+After=network.target
+Wants=network.target
+
+[Service]
+Type=simple
+DynamicUser=true
+User=lighthouse
```
bakhtin: Both User and Group will most probably not exist on the target system. You should create it with [postinst maintainer scripts](https://github.com/kornelski/cargo-deb)

lighthouse/lighthouse.service
```diff
@@ -0,0 +1,34 @@
...
+Group=eth
```
bakhtin:
```suggestion
Group=lighthouse
```

lighthouse/lighthouse.service
```diff
@@ -0,0 +1,34 @@
...
+Restart=always
+RestartSec=5
+TimeoutStopSec=180
+Environment="RUST_LOG=info"
+ExecStart=/usr/bin/lighthouse \
```
bakhtin: You should not assume any specific use-case of a service. I'd simplify it to `/usr/bin/lighthouse bn`

lighthouse/lighthouse.service
```diff
@@ -0,0 +1,34 @@
+[Unit]
+Description=Lighthouse Ethereum Beacon Chain Client
+Documentation=https://lighthouse-book.sigmaprime.io/
+After=network.target
+Wants=network.target
+
+[Service]
+Type=simple
+DynamicUser=true
```
bakhtin:
```suggestion

```

bakhtin: I'd suggest to rely on static users/groups as DynamicUser has [complex interaction](https://0pointer.net/blog/dynamic-users-with-systemd.html) with the stateful objects on disk (e.g., LH database)

.github/workflows/reproducible-build.yml
```diff
@@ -0,0 +1,188 @@
+name: reproducible-build
```
bakhtin: I'd avoid imposing a responsibility for verifying reproducibility to third-parties (Lighthouse team in this case). Those who have reproducibility requirements should independently verify it relying on their infra.

.github/workflows/release-reproducible.yml
```diff
@@ -0,0 +1,142 @@
+name: release-reproducible
```
bakhtin, line 5: Can we make reproducible builds a default for Lighthouse? Not an additional target requiring extra effort to support

.github/workflows/reproducible-build.yml
bakhtin: The variance in the environment is not sufficient to say that the builds are reproducible. Check out how Debian independently verify reproducibility by building on different OSes with a variable set of tools.

.github/workflows/release-reproducible.yml
MoeMahhouk: I created a separate workflow for the reproducible container builds as separation of concerns similar to what the reth team did here https://github.com/paradigmxyz/reth/blob/main/.github/workflows/release-reproducible.yml

.github/workflows/reproducible-build.yml
MoeMahhouk: I also followed here the steps of reth and how they are doing it here https://github.com/paradigmxyz/reth/blob/main/.github/workflows/reproducible-build.yml
However, good point on if the lighthouse team would want this to be as part of their workflow or not. I would leave it for them to decide and then we can re-iterate on what the best way to do it

lighthouse/lighthouse.service
MoeMahhouk: I thought we wanted to pack, lighthouse, reth and rbuilder in the same group, right?

lighthouse/lighthouse.service
MoeMahhouk: yeah, this is just a dummy/simple systemd service file that will be anyway overwritten during deplyoments as we discussed offline, right?
Otherwise, how would you suggest it to be? just` /usr/bin/lighthouse --help`?

lighthouse/lighthouse.service
MoeMahhouk: good point, I will remove it

lighthouse/lighthouse.service
bakhtin: We should not assume that either reth or rbuilder will be installed on the system. LH should be self-sufficient. Also, since you're using `DynamicUser` you can omit User and Group all together.

bakhtin: I suggest to simplify the `[Service]` section to this:
```
[Service]
Type=exec # switched from 'simple' for better startup handling. See https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#Type=
DynamicUser=yes
StateDirectory=lighthouse
ExecStart=/usr/bin/lighthouse bn \
          --execution-endpoint http://localhost:8551 \
          --execution-jwt-secret-key 0000000000000000000000000000000000000000000000000000000000000000
          --datadir %S/lighthouse
```
This is unassuming, minimal, and still functional. An administrator can override the unit parameters if they want to.

lighthouse/lighthouse.service
bakhtin: see above

lighthouse/lighthouse.service
MoeMahhouk: good point. Adjusted as suggested

lighthouse/debian/postinst
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
```
bakhtin: You don't need this file. Your systemd unit will automatically create a directory and a user.

lighthouse/debian/postrm
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
```
bakhtin: This file also not needed

lighthouse/debian/postinst
MoeMahhouk: ah right, those are leftovers from the previous commits where I didn't have the DynamicUser.
Good catch. I will remove now

.github/workflows/release-reproducible.yml
```diff
@@ -0,0 +1,126 @@
...
+          DOCKER_SUCCESS=$(gh api /repos/${{ github.repository }}/actions/workflows/docker-reproducible.yml/runs \
+            --jq ".workflow_runs[] | select(.head_branch == \"$TAG\" and .conclusion == \"success\") | .id" \
+            | head -1)
+          
+          if [[ -n "$RELEASE_SUCCESS" ]] && [[ -n "$DOCKER_SUCCESS" ]]; then
+            echo "Both workflows completed successfully for $TAG"
+            echo "  - Release Suite: run $RELEASE_SUCCESS"
+            echo "  - docker-reproducible: run $DOCKER_SUCCESS"
+            echo "should_proceed=true" >> $GITHUB_OUTPUT
+          else
```
bakhtin, line 54: This block does not actually wait for the completion of another workflow that this workflow depends on, it checks the status and exits if not ready.
As per [GH docs](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_run) when multiple WFs specified in `workflows: [Release Suite, docker-reproducible]` only one needs to complete, the other could still be running when `release-reproducible` is triggered. So we need to actually wait for the completion periodically checking the status of a WF.

chong-he:
@MoeMahhouk is this still a work in progress?

.github/workflows/release-reproducible.yml
MoeMahhouk: yes exactly, that is why I introduced the `should_proceed` variable which will only be true if both of those workflows ran and completed successfully.
Therefore, if only one ran and the other didnt, it will immediately stop this workflow to spare resources and run again once the second workflow finalizes successfully. Check out the types: [completed] [above](https://github.com/sigp/lighthouse/pull/7617/files#diff-222af23bee616920b04f5b92a83eb5106fce08abd885cd3a3b15b8beb5e789c3R10)

MoeMahhouk:
> @MoeMahhouk is this still a work in progress?

@chong-he , yes it is. I will open it up for review this week after clearing up some feedback.
Then I would like to get your and your team's feedback on it to bring it to final shape

Makefile
```diff
@@ -332,3 +332,81 @@ clean:
...
+	else \
+		echo "No .deb package found. Run 'make deb-cargo' first."; \
+	fi
+
+
+.PHONY: remove-deb-local
+remove-deb-local: ## Remove installed lighthouse package
+	@echo "Removing lighthouse package..."
+	sudo dpkg -r lighthouse || true
+	sudo systemctl daemon-reload || true
```
MoeMahhouk: I can remove those targets if not needed to minimize PR code

Makefile
```diff
@@ -332,3 +332,81 @@ clean:
...
+	@echo "  remove-deb-local       - Remove installed package"
+	@echo "  clean-deb              - Clean up packaging artifacts"
+	@echo ""
+	@echo "Prerequisites:"
+	@echo "  - lighthouse/lighthouse.service file"  
+	@echo "  - README.md file in current directory"
+	@echo ""
+	@echo "Quick start:"
+	@echo "  make deb-cargo         - Build .deb for current RUST_TARGET"
+	@echo "  make install-deb-local - Test the package"
```
MoeMahhouk: same here, I can clean this up if not needed

chong-he:
cargo sort CI is failing, can you fix this? 

> > @MoeMahhouk is this still a work in progress?
> 
> @chong-he , yes it is. I will open it up for review this week after clearing up some feedback. Then I would like to get your and your team's feedback on it to bring it to final shape

cargo sort CI is failing, can you fix this?

MoeMahhouk:
> cargo sort CI is failing, can you fix this?
> 
> > > @MoeMahhouk is this still a work in progress?
> > 
> > 
> > @chong-he , yes it is. I will open it up for review this week after clearing up some feedback. Then I would like to get your and your team's feedback on it to bring it to final shape
> 
> cargo sort CI is failing, can you fix this?

Done

chong-he:
I have a look at this PR, from my understanding, this will include a `.deb` package in the release alongside with other binary files? 

I am wondering what is the use case here?

MoeMahhouk:
> I have a look at this PR, from my understanding, this will include a `.deb` package in the release alongside with other binary files?
> 
> I am wondering what is the use case here?

The use case here is that those reproducibly built debian packages can be utilized to accelerate workflows that require building images that includes lighthouse. For example, BuilderNet VM image. In mkosi, you could use debian packages during the image build process which would reduce the build time substantially. Beside that use case, there are also other use-cases for such users that prefer downloading and installing lighthouse using debian packages. In addition, the binary within is built reproducibly for security purposes.
There are probably other use-cases too but those that come to my mind now. Maybe @bakhtin could also share more here

bakhtin:
@chong-he I agree with what Moe described regarding the use-cases for a *.deb package. It also makes easier for people running nodes to install a *.deb package that comes with a systemd service definition baking in basic CLI options of Lighthouse simplifying managing the service.

dapplion:
Cleaned up unrelated changes that were pulled into the PR diff by merging `stable` into this branch. The `stable` branch contained commits from #7892 (beacon head monitor), #8748 (bootnode fix), and #8749 (v8.1.0 release) that reached `unstable` through a different merge path, causing them to show up as part of this PR's diff.

Merged latest upstream `unstable` and reset all non-deb-packaging files to match, so the PR diff now only contains the intended changes:

- `.github/workflows/docker-reproducible.yml` — extract `.deb` artifacts
- `.github/workflows/release-reproducible.yml` — sign and publish workflow
- `Dockerfile.reproducible` — build `.deb` in Docker
- `Makefile` — deb packaging targets
- `lighthouse/Cargo.toml` — cargo-deb metadata
- `lighthouse/debian/lighthouse.service` — systemd unit
- `lighthouse/debian/prerm` — pre-removal script

niccoloraspa:
Could we get this merged before next release? We would really love to use `deb` packages 🙏

antondlr:
@MoeMahhouk sorry for the silence on this, but it's been on my mind, and I'd like to propose a different direction: see #9147 
I'd very much appreciate your input on this matter!

MoeMahhouk:
> @MoeMahhouk sorry for the silence on this, but it's been on my mind, and I'd like to propose a different direction: see #9147 I'd very much appreciate your input on this matter!

@antondlr No problems, thanks for revisiting this PR back.
I have reviewed your proposal PR and left you comments directly on it. 
If I see correctly, the proposal PR does a solid refactoring but it missed the purpose of this PR which is debian packaging. 
Is that not wanted anymore?

PR #7620, opened 2025-06-19
Remove instrumenting log level

### LINKED ISSUE #7155: Debug level log appearing in stdout when log level is set to INFO

AgeManning:
## Issue Addressed

I think this should resolve #7155

This removes the level field from the instrumenting we were doing across a range of functions. The level will now default to the level of the log.

### DIFF
--- a/beacon_node/beacon_chain/src/validator_monitor.rs
+++ b/beacon_node/beacon_chain/src/validator_monitor.rs
@@ -406,7 +406,6 @@ pub struct ValidatorMonitor<E: EthSpec> {
 
 impl<E: EthSpec> ValidatorMonitor<E> {
     #[instrument(parent = None,
-        level = "info",
         name = "validator_monitor",
         skip_all
     )]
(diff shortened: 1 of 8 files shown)


michaelsproul (approved):
Nice!

Why were we doing this before??

AgeManning:
I don't know :man_shrugging:  - Artifact from the tracing upgrade?

michaelsproul:
@mergify requeue

Issue #5912, opened 2024-06-11
Database schema changes 2024-2025


michaelsproul:
## Description

This is a meta-issue for scheduling database schema changes in Lighthouse.

The Lighthouse v5.2.0 schema was v19.

v20 was:

- https://github.com/sigp/lighthouse/pull/5712

v21 was (released in Lighthouse v5.3.0):

- https://github.com/sigp/lighthouse/pull/5897

v22 was:

- https://github.com/sigp/lighthouse/pull/5978

v23 is:

- https://github.com/sigp/lighthouse/pull/6744

v24 is:

- https://github.com/sigp/lighthouse/pull/6750

v25 is:

- https://github.com/sigp/lighthouse/pull/7133

v26 is:

- Cleaning up on-disk Fusaka data accidentally stored pre-Fusaka (@jimmygchen)
- https://github.com/sigp/lighthouse/pull/7661

v27 is

- https://github.com/sigp/lighthouse/pull/7648

v28 is:

- https://github.com/sigp/lighthouse/issues/7760

Later:

- https://github.com/sigp/lighthouse/pull/7624

Future versions:

- Linear blocks
- `HotStateSummary` cleanup + `CompactBeaconState`. TODO: do it without `ValidatorTrait`.
- Abolish BlobInfo

michaelsproul:
Something else to consider is that we already used some of the versions in this range for preliminary tree-states releases. The latest being v24 which will _not_ be compatible with the final v24 envisioned here. Maybe we could skip a few numbers to avoid confusion

michaelsproul:
I've added V28 for the fork choice slimdown, which I am working on as part of:

- https://github.com/sigp/lighthouse/issues/7760

PR #7624, opened 2025-06-21
feat(beacon_node/beacon_chain): Removed DataColumnInfo

### LINKED ISSUE #6572: Remove `DataColumnInfo`

PoulavBhowmick03:
## Issue Addressed

Potentially Fixes #6572

## Proposed Changes

- Removed `DataColumnInfo` from `beacon_node/store/src/metadata.rs` and it's metadata
- Removed data column initialization in beacon chain builder
- Simplified RPC methods to use data_availability_boundary_slot directly
- Removed database operations for data column info storage/retrieval
- Cleaned up unused imports

### DIFF
--- a/beacon_node/beacon_chain/src/builder.rs
+++ b/beacon_node/beacon_chain/src/builder.rs
@@ -410,11 +410,6 @@ where
                 .init_blob_info(genesis.beacon_block.slot())
                 .map_err(|e| format!("Failed to initialize genesis blob info: {:?}", e))?,
         );
-        self.pending_io_batch.push(
-            store
-                .init_data_column_info(genesis.beacon_block.slot())
-                .map_err(|e| format!("Failed to initialize genesis data column info: {:?}", e))?,
-        );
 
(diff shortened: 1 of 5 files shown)


michaelsproul (changes requested):
Technically this change will also require a DB schema change in order to clean up the old data (to delete it). We'll also need a downgrade which recreates it. Seeing as the value is informational, the downgrade could just set the oldest data column slot to the `split` slot.

We are planning out schema upgrades and allocating versions in this issue:

- https://github.com/sigp/lighthouse/issues/5912

PR #7635, opened 2025-06-23
improve Rpc response logging

jxs:
## Issue Addressed

This PR mainly addresses two issues with the current ConnectionHandler logging:

-  Currently, only successful `BlocksByRange` and `BlobsByRange` responses sent are logged. This PR expands logging to include all response types, providing better visibility into the full range of interactions.

- The existing `ConnectionHandler` logs lack the `SubstreamId`, making it difficult to correlate responses sent with their corresponding `InboundRequestId`. This makes it impossible to find which `InboundRequestId` was responded to . 
This PR resolves that by logging when a request is dispatched to the `ConnectionHandler`, including the associated `InboundRequestId`.

While addressing the above issues, took the opportunity to unify `SubstreamId` and `ConnectionId` to `InboundRequestId` on `QueuedResponse` as it doesn't need to have them separated, we only need that when dispatching to the `ConnectionHandler`

### DIFF
--- a/beacon_node/lighthouse_network/src/rpc/handler.rs
+++ b/beacon_node/lighthouse_network/src/rpc/handler.rs
@@ -577,28 +577,15 @@ where
                                     self.inbound_substreams_delay.remove(delay_key);
                                 }
 
-                                // BlocksByRange is the one that typically consumes the most time.
-                                // Its useful to log when the request was completed.
-                                if matches!(info.protocol, Protocol::BlocksByRange) {
-                                    debug!(
-                                        peer_id = %self.peer_id,
-                                        connection_id = %self.connection_id,
-                                        duration = Instant::now()
-                                            .duration_since(info.request_start_time)
-                                            .as_secs(),
-                                        "BlocksByRange Response sent"
-                                    );
-                                }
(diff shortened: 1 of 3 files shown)


dknopik (commented):
Is it necessary to have this in unstable vs running this branch on a node for some time? I am a bit concerned about the increased log volume.

Otherwise LGTM

jxs:
> Is it necessary to have this in unstable vs running this branch on a node for some time? I am a bit concerned about the increased log volume.
> 
> Otherwise LGTM

necessary wouldn't say so, it would be helpful also for the future. But that's a good question, @antondlr wrt log volume do we have any data/context?

antondlr:
> I am a bit concerned about the increased log volume.

yeah ideally let's run it separately for a while on a node so we can compare (nominal) log volume. 
doesn't necessarily have to be blocking though imo. _if_ log volume increases too much, we can always address it in a followup PR. 

as far as I understand, in ideal conditions the impact will be minimal but with adverse network conditions, we could see seriously increased volumes? can we cap this somehow, just to make sure we survive turbulent waters?

jxs:
super seeded by https://github.com/sigp/lighthouse/pull/7663

PR #7644, opened 2025-06-25
Rust 1.89 compiler lint fix

chong-he:
## Issue Addressed

Fix lints for Rust 1.89 beta compiler


michaelsproul (approved):
Thanks CK, looks great!

It's funny because I think the compiler got us to remove all of these lifetimes in a previous version, and now it wants us to add them back :sweat_smile:

Issue #7647, opened 2025-06-25
Store node's earliest available data column on disk


eserilev:
According to the status v2 spec

```
When a value for `custody_group_count` is set, the `earliest_available_slot`
field in the status RPC message SHOULD reflect the slot at which the
`custody_group_count` was updated.
```

Currently we calculate `ealiest_available_slot` via `beacon_chain.store.get_anchor_info().oldest_block_slot`. This is problematic in situations where custody group count changes. When this happens the `earliest_available_slot` should be set to the slot where the cgc was changed. We need to add some additional metadata in the blobs db to keep track of the `earliest_available_slot` for the columns a node is required to custody.

jimmygchen:
Completed in #7648

PR #7648, opened 2025-06-25
Data column custody info

eserilev:
## Issue Addressed

#7647

## Proposed Changes

Introduces a new record in the blobs db `DataColumnCustodyInfo`

When `DataColumnCustodyInfo` exists in the db this indicates that a recent cgc change has occurred and/or that a custody backfill sync is currently in progress (custody backfill will be added as a separate PR). When a cgc change has occurred `earliest_available_slot` will be equal to the slot at which the cgc change occured. During custody backfill sync`earliest_available_slot` should be updated incrementally as it progresses.

~~Note that if `advertise_false_custody_group_count` is enabled we do not add a `DataColumnCustodyInfo` record in the db as that would affect the status v2 response.~~
(See comment https://github.com/sigp/lighthouse/pull/7648#discussion_r2212403389)

~~If `DataColumnCustodyInfo` doesn't exist in the db this indicates that we have fulfilled our custody requirements up to the DA window.~~
(It now always exist, and the slot will be set to `None` once backfill is complete)

StatusV2 now uses `DataColumnCustodyInfo` to calculate the `earliest_available_slot` if a `DataColumnCustodyInfo` record exists in the db, if it's `None`, then we return the `oldest_block_slot`.


beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -210,6 +210,7 @@ impl CustodyContext {
                 return Some(CustodyCountChanged {
                     new_custody_group_count: updated_cgc,
                     sampling_count: self.sampling_size(Some(effective_epoch), spec),
+                    slot: current_slot,
```
jimmygchen: I think we should be using the `effective_epoch` here - this is when the new CGC takes effect.

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -129,6 +137,9 @@ impl<E: EthSpec> BlockCache<E> {
             .get(block_root)
             .and_then(|map| map.get(column_index).cloned())
     }
+    pub fn get_data_column_custody_info(&mut self) -> Option<DataColumnCustodyInfo> {
```
jimmygchen: This doesn't require a mutable reference to `self` i think?

```suggestion
    pub fn get_data_column_custody_info(&self) -> Option<DataColumnCustodyInfo> {
```

beacon_node/store/src/metadata.rs
```diff
@@ -204,6 +205,16 @@ impl StoreItem for BlobInfo {
     }
 }
 
+/// Database parameter relevant to data column custody sync. There is only at most a single
+/// `DataColumnCustodyInfo` stored in the db. This record is added to the db when cgc
+/// count changes and is updated incrementally during data column custody backfill. Once custody backfill
+/// is complete the record is removed from the db.
+#[derive(Debug, PartialEq, Eq, Clone, Encode, Decode, Serialize, Deserialize, Default)]
+pub struct DataColumnCustodyInfo {
```
jimmygchen, line 213: Just confirming my understanding - this looks very similar to `DataColumnInfo` but I guess that's being removed soon in #7624, as it's stored in a different db - is this right?

beacon_node/network/src/status.rs
```diff
@@ -29,7 +29,18 @@ pub(crate) fn status_message<T: BeaconChainTypes>(beacon_chain: &BeaconChain<T>)
...
 
-    let earliest_available_slot = beacon_chain.store.get_anchor_info().oldest_block_slot;
+    // If there is no data column custody info in the db, that indicates that
+    // no recent cgc changes have occurred and no cgc backfill is in progress.
+    let earliest_available_slot = if let Ok(Some(data_column_custody_info)) =
+        beacon_chain.store.get_data_column_custody_info(false)
+    {
+        std::cmp::max(
+            beacon_chain.store.get_anchor_info().oldest_block_slot,
+            data_column_custody_info.earliest_data_column_slot,
```
jimmygchen: What's the reason for using `max` of these values instead of just returning `earliest_data_column_slot`? we should be able to serve anything within the data availability window?

beacon_node/beacon_chain/src/builder.rs
```diff
@@ -825,6 +825,10 @@ where
             ));
         }
 
+        store
+            .get_data_column_custody_info(true)
+            .map_err(|e| format!("Unable to fetch data column custody info {:?}", e))?;
```
jimmygchen: It's not immediately obvious what we're doing here as we're not using the return value - could add a comment to say we're priming the cache.


Or alternatively this could be done on demand in the `get_data_column_custody_info` function, check cache first and fallback to read from db, which saves having an extra function param and priming the cache on startup.

beacon_node/network/src/service.rs
```diff
@@ -753,6 +755,10 @@ impl<T: BeaconChainTypes> NetworkService<T> {
                     .advertise_false_custody_group_count
                     .is_none()
                 {
+                    // Update data column custody info with the slot at which cgc was changed.
+                    self.beacon_chain.store.put_data_column_custody_info(slot).unwrap_or_else(|e| {
```
jimmygchen: It feels a bit weird that the network is persisting non-network related metadata straight into the beacon store. It seems like we're crossing responsibilities here and relying on the fact that `store` isn't encapsulated.

So currently
- `http-api` registering validators via the beacon_chain (custody context)
- and then `CustodyContext` sends the update message to network
- `network` then persist the change directly to the `beacon_chain` store

An alternative might be
- `http-api` register the validators via `beacon_chain`
- `beacon_chain` computes the updates, send message to network, and then persist directly from within the beacon chain.

This isn't really a big deal I think, I'm just thinking out loud as I notice it. I'm also ok with leaving it as it is, seeing that network already holds a direct reference to store (via `self.store`).

jimmygchen (changes requested):
@eserilev Nice, this looks great overall! 
The main comment is that we should probably store the effective epoch/slot instead of the current slot - this is when the new CGC is applied in DA check.

beacon_node/store/src/metadata.rs
eserilev: yes, data column info is being deprecated which is why i didn't reuse it. I also thought `DataColumnCustodyInfo` would be a clearer name

beacon_node/beacon_chain/src/builder.rs
eserilev: the reason I wasn't falling back to the db is because data column custoy info only exist in the cache + db when a  recent cgc change occurs. So like 99% of the time we'd end falling back to the db which would make the cache sort of useless

beacon_node/network/src/service.rs
eserilev: yeah I think what you're saying makes sense, I can refactor things a bit here

beacon_node/network/src/status.rs
eserilev: yeah you're right if `data_column_custody_info` isn't `None` we should always just use that value

beacon_node/beacon_chain/src/builder.rs
jimmygchen: ah right, I missed that. Thanks!

jimmygchen: would be useful to add a comment here to say priming the cache though.

eserilev:
@michaelsproul have a question re: schema version increment

This PR introduces a new column, but the database will only contain a value for this column if a recent cgc change has occurred and custody backfill hasn't finished. In that case, is a version increment necessary? An upgrade should be a no-op. A downgrade might also be a no-op because we should not delete this column from the db if a record exists as that would break our ability to track the recent cgc change

beacon_node/beacon_chain/src/builder.rs
michaelsproul: This sort of conditional existence seems slightly trickier to reason about, maybe it would be more straightforward if we expected `DataColumnCustodyInfo` to always exist on disk, and we could have a value that means "no recent CGC changes"? Maybe an `Option<Slot>` just like `DataColumnInfo`. This would also mean we can detect cases like the `blobs_db` being deleted/replaced: we would error if the value is missing entirely.

We also wouldn't need to make one of these caches that is prone to erroring if it's not primed in the right order. i.e. we could make it so that `HotColdDB::get_data_column_custody_info` tries the in-memory value, and falls back to loading from disk if it's not found.

michaelsproul:
@eserilev I think a schema update is cleanest, especially if we go with my idea from here to make the `DataColumnCustodyInfo` stored always (when Fulu is enabled): https://github.com/sigp/lighthouse/pull/7648#discussion_r2181332857

As Jimmy pointed out via DM, a DB downgrade that deletes the value would also guard against starting up with a stale value in the scenario where we downgrade and then re-upgrade. I.e. run the new version, downgrade, run the old version for a bit, upgrade again --> end up using the stale value from the first time we ran. However this is probably not super relevant because by the time we are running Fulu on real networks, downgrading to a prior schema should not be allowed (we could even add a check in the downgrade that only allows downgrading if `fulu_fork_epoch.is_none()`).

michaelsproul:
The `schema_stability` test will also need an upgrade for the new column. And we could add an assert about the `DataColumnCustodyInfo` so that future changes to it (which are probably kinda likely) are caught.

beacon_node/beacon_chain/src/schema_change.rs
```diff
@@ -67,6 +68,14 @@ pub fn migrate_schema<T: BeaconChainTypes>(
             let ops = migration_schema_v26::downgrade_from_v26::<T>(db.clone())?;
             db.store_schema_version_atomically(to, ops)
         }
+        (SchemaVersion(26), SchemaVersion(27)) => {
+            let ops = migration_schema_v27::upgrade_to_v27::<T>(db.clone())?;
+            db.store_schema_version_atomically(to, ops)
```
jimmygchen: `db.store_schema_version_atomically` do the `ops` on the `hot_db` - not the `blobs_db` where `BeaconDataColumnCustodyInfo` is stored.

I think we need a additional call here to the `blobs_db` for the custody info.

beacon_node/beacon_chain/src/schema_change/migration_schema_v27.rs
```diff
@@ -0,0 +1,47 @@
...
+            earliest_data_column_slot: None,
+        };
+        vec![KeyValueStoreOp::PutKeyValue(
+            DBColumn::BeaconDataColumnCustodyInfo,
+            DATA_COLUMN_CUSTODY_INFO_KEY.as_slice().to_vec(),
+            data_column_custody_info.as_ssz_bytes(),
+        )]
+    } else {
+        // Delete it from the db if PeerDAS hasn't been scheduled
+        vec![KeyValueStoreOp::DeleteKey(
```
jimmygchen: `CustodyInfo` is introduced in v27, I think we don't need to delete it when upgrading from v26 to v27, as it doesn't exist in v26.

beacon_node/beacon_chain/src/schema_change/migration_schema_v27.rs
```diff
@@ -0,0 +1,47 @@
+use crate::BeaconChainTypes;
+use ssz::Encode;
+use std::sync::Arc;
+use store::metadata::DataColumnCustodyInfo;
+use store::metadata::DATA_COLUMN_CUSTODY_INFO_KEY;
+use store::{DBColumn, Error, HotColdDB, KeyValueStoreOp};
+use tracing::info;
+
+/// Add `DataColumnCustodyInfo` entry to v27.
+pub fn upgrade_to_v27<T: BeaconChainTypes>(
```
jimmygchen, line 6: You also need to update current schema version here to v27:

https://github.com/sigp/lighthouse/blob/ce1685541859aae8a0b2a85ada81742c3a16f27a/beacon_node/store/src/metadata.rs#L7

beacon_node/beacon_chain/src/schema_change/migration_schema_v27.rs
```diff
@@ -0,0 +1,47 @@
...
+    Ok(ops)
+}
+
+pub fn downgrade_from_v27<T: BeaconChainTypes>(
+    db: Arc<HotColdDB<T::EthSpec, T::HotStore, T::ColdStore>>,
+) -> Result<Vec<KeyValueStoreOp>, Error> {
+    if db.spec.is_peer_das_scheduled() {
+        return Err(Error::MigrationError(
+            "Cannot downgrade from v27 if peerDAS is scheduled".to_string(),
+        ));
```
jimmygchen, line 23: I think this entry only exists in v27 if peerdas is scheduled?

beacon_node/beacon_chain/src/validator_custody.rs
jimmygchen: this still need to be updated

beacon_node/beacon_chain/tests/schema_stability.rs
```diff
@@ -122,6 +124,7 @@ fn check_metadata_sizes(store: &Store<E>) {
         }
     );
     assert_eq!(DataColumnInfo::default().ssz_bytes_len(), 5);
+    assert_eq!(DataColumnCustodyInfo::default().ssz_bytes_len(), 5);
```
jimmygchen, line 137: We also need a function to load it from store and check the store bytes size, similar to the `check_custody_context` below

beacon_node/http_api/src/lib.rs
```diff
@@ -3878,6 +3842,15 @@ pub fn serve<T: BeaconChainTypes>(
                             current_slot,
                             &chain.spec,
                         ) {
+                            // Don't update custody info if we're advertising a
```
jimmygchen: Are we going to use this for backfilling? would it impact backfilling if we set it to none?

eserilev: yes, in general i wasnt planning on triggering backfill if we're advertising a false custody group count. does that sound right?

jimmygchen: i was thinking to have the adverise false cgc thing not impacting anything else, in case if we want to re-start the node without the flag after testing? but this isnt' a big issue, as we're likely going to remove / feature gate the flag before mainnet launch.

beacon_node/beacon_chain/src/schema_change/migration_schema_v27.rs
jimmygchen: Sorry i wasn't clear - i was going to say : we only want to delete this if peerdas is scheduled? 
^ thinking about this again, to be really correct, we should just always attempt to delete it in the case of downgrade - we can't guarantee if peerdas was scheduled when the node was running.

Again this is _very_ minor given this isn't on a live network, but thought it's worth pointing out anways.

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -217,7 +217,7 @@ impl CustodyContext {
                     new_custody_group_count: updated_cgc,
                     sampling_count: self
                         .num_of_custody_groups_to_sample(Some(effective_epoch), spec),
-                    slot: current_slot,
+                    effective_epoch: current_slot.epoch(E::slots_per_epoch()),
```
jimmygchen: Sorry i was probably unclear again, what i meant was to use the existing `effective_eppch` variable in scope:

```suggestion
                    effective_epoch,
```

jimmygchen (changes requested):
Hey @eserilev 
Looking good, i think we can merge this soon - I've left a few comments, the main one is the `effective_epoch` comment that needs fix I think.

beacon_node/store/src/metadata.rs
```diff
@@ -204,6 +205,16 @@ impl StoreItem for BlobInfo {
     }
 }
 
+/// Database parameter relevant to data column custody sync. There is only at most a single
+/// `DataColumnCustodyInfo` stored in the db. This record is added to the db when cgc
+/// count changes and is updated incrementally during data column custody backfill. Once custody backfill
+/// is complete the record is removed from the db.
```
michaelsproul: Description here is out of date (we no longer delete once custody backfill completes).

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -2389,6 +2421,33 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
+            let bytes_opt = self.blobs_db.get_bytes(
+                DBColumn::BeaconDataColumnCustodyInfo,
+                DATA_COLUMN_CUSTODY_INFO_KEY.as_slice(),
+            )?;
+
+            let Some(bytes) = bytes_opt else {
+                return Ok(None);
+            };
+
+            let data_column_custody_info = Some(DataColumnCustodyInfo::from_ssz_bytes(&bytes)?);
```
michaelsproul: I think you might be able to use `StoreItem` here to factor out some of this stuff, but it isn't essential

(would need to `impl StoreItem for DataColumnCustodyInfo`)

beacon_node/network/src/status.rs
```diff
@@ -29,8 +29,21 @@ pub(crate) fn status_message<T: BeaconChainTypes>(beacon_chain: &BeaconChain<T>)
         finalized_checkpoint.root = Hash256::zero();
     }
 
-    let earliest_available_slot = beacon_chain.store.get_anchor_info().oldest_block_slot;
+    let earliest_available_data_column_slot = beacon_chain
+        .store
+        .get_data_column_custody_info()
+        .ok()
```
michaelsproul, line 36: This is a biiit sketchy in that it swallows the DB read error, and then the logic below defaults to assuming we have full info.

Might be worth just adding a `NOTE` that we are making the assumption that `get_data_column_custody_info` won't fail

beacon_node/beacon_chain/src/schema_change/migration_schema_v27.rs
michaelsproul: There's two conflicting requirements here:

1. Keep backwards compatibility for current (testnet) PeerDAS nodes (allow downgrades v27 to v26 for peer DAS nodes)
2. Prevent users from breaking their node once PeerDAS is activated (disallow downgrades from v27 to v26 for peer DAS nodes)

Currently this migration is implementing (2). You're suggesting solving (1), which would be fine in the short-term, but should probably be changed to (2) eventually.

beacon_node/http_api/src/lib.rs
eserilev: I guess the only thing I was worried about is that updating custody info will affect the status v2 message. if we don't care about that, i'll ignore the false custody flag stuff

beacon_node/network/src/status.rs
eserilev: i've added a note

beacon_node/store/src/hot_cold_store.rs
eserilev: [0f2fc7d](https://github.com/sigp/lighthouse/pull/7648/commits/0f2fc7d842fce127ce487225e8f7da5be030149b)

beacon_node/store/src/metadata.rs
eserilev: updated description

beacon_node/beacon_chain/src/validator_custody.rs
eserilev: woops, no you werent unclear i just missed that

[dc04308](https://github.com/sigp/lighthouse/pull/7648/commits/dc043083375a855b4fe8ec5aeaa05a5539539860)

beacon_node/beacon_chain/src/schema_change/migration_schema_v27.rs
eserilev: is there a reason for solving (1) in the short term? does it give us better control over certain testing scenarios?

jimmygchen: Oh yeah that's true -  we disallow downgrade after a major version upgrade - I just didn't think about doing it in this PR.

Yeah i think solving (2) makes sense!

beacon_node/http_api/src/lib.rs
jimmygchen: yeah the advertise false CGC flag is a malicious test anyways, so statusv2 responses to peers doesn't really matter - but internally i think we should maintain the functionality, so we can turn flag on / off without having to re-sync.  I'll push a fix to drop this condition.

jimmygchen (approved):
LGTM!

michaelsproul (approved):
Looks good 😎

PR #7649, opened 2025-06-25
feat(val): Reconstructed Validator Attestation Service to construct SingleAttestation directly

### LINKED ISSUE #7634: Construct and sign `SingleAttestation` in the validators attestation service
The validator attestation service constructs an `Attestation` object that it signs, converts to a `SingleAttestation` and publishes to the network. We could remove the conversion step by just constructing the `SingleAttestation` directly and signing it. There is no need to support the pre-Electra unaggregated attestation format.

PoulavBhowmick03:
## Issue Addressed

Fixes #7634


eserilev (changes requested):
I think we can go a bit further in this refactor 

- In becon_chain.rs we have `produce_unaggregated_attestation`, this should produce a `SingleAttestation` instead of an `Attestation`
- We can safely remove the `Attestation` `empty_for_signing` method. It is used in various test cases, so those will also need to be updated to use `SingleAttestation::empty_for_signing`
- the `ValidatorStore` trait has `sign_attestation` method that should probably accept a `SingleAttestation` to sign instead of an `Attestation`. This will affect downstream tests

PoulavBhowmick03:
> I think we can go a bit further in this refactor
> 
> * In becon_chain.rs we have `produce_unaggregated_attestation`, this should produce a `SingleAttestation` instead of an `Attestation`
> * We can safely remove the `Attestation` `empty_for_signing` method. It is used in various test cases, so those will also need to be updated to use `SingleAttestation::empty_for_signing`
> * the `ValidatorStore` trait has `sign_attestation` method that should probably accept a `SingleAttestation` to sign instead of an `Attestation`. This will affect downstream tests

If I remove the `Attestation::empty_for_signing` function and use SingleAttestation, I would also have to introduce the types that Attestation has, for this part

```rust
        let attestation: SingleAttestation = SingleAttestation::empty_for_signing(
            index,
            committee_len,
            slot,
            beacon_block_root,
            state.current_justified_checkpoint(),
            Checkpoint {
                epoch,
                root: target_root,
            },
            &self.spec,
        )?;

        let attestation = match attestation {
            Attestation::Electra(mut attn) => {
                attn.aggregation_bits
                    .set(aggregation_bit_index, true)
                    .unwrap();
                Attestation::Electra(attn)
            }
            Attestation::Base(mut attn) => {
                attn.aggregation_bits
                    .set(aggregation_bit_index, true)
                    .unwrap();
                Attestation::Base(attn)
            }
        };
```
Because currently, eg, Electra or Base, these types dont exist in SingleAttestation
What should be the way to go here?

eserilev:
@PoulavBhowmick03 we can construct the `SingleAttestation` there directly. The match statement wont be necessary because its an unaggregated attestation (only one aggregation bit set, the `SingleAttestation` `attester_index` field)

PoulavBhowmick03:
> @PoulavBhowmick03 we can construct the `SingleAttestation` there directly. The match statement wont be necessary because its an unaggregated attestation (only one aggregation bit set, the `SingleAttestation` `attester_index` field)

ah okay. looking into it

PoulavBhowmick03:
@eserilev can you please open this PR. I will be committing my changes by tonight

eserilev:
For reference I'm pretty sure you can just hit the Reopen button below

PR #7653, opened 2025-06-26
Fix broken Lighthouse API anchor links in graffiti documentation

MozirDmitriy:
Replaced incorrect .html links with .md links for the Lighthouse API PATCH endpoint in graffiti.md. This ensures that all references to the validator graffiti update API point to the correct file and anchor, resolving broken link errors in the documentation build.


macladson (commented):
We actually build the Lighthouse docs from the `unstable` branch where this has already been fixed. [If you go to the actual page in question you can see the links work.](https://lighthouse-book.sigmaprime.io/validator_graffiti.html)

Issue #4788, opened 2023-09-28
State reconstruction API doesn't work if genesis backfill hasn't been enabled


michaelsproul:
## Description

The `/lighthouse/database/reconstruct` API is currently faulty in the case where the node was started without `--genesis-backfill` or `--reconstruct-historic-states`. The API will only attempt to trigger reconstruction, it won't trigger block backfill.

The error logged is:

> ERRO State reconstruction failed             error: MissingHistoricBlocks { oldest_block_slot: Slot(6225249) }, service: beacon

## Version

v4.5.0

## Present Behaviour

The reconstruct API only works if block backfill is enabled & complete.

## Expected Behaviour

We should either:

A) Make it so that the reconstruct API triggers block backfill by flipping that var (I think it's immutable currently), or
B) Clarify in the docs and the API response that this isn't supported.

## Workaround

Restart the BN with `--reconstruct-historic-states`.

michaelsproul:
I think we may as well fix the docs first (option B), and then think about option A (harder).

PR #7654, opened 2025-06-26
Improve /lighthouse/database/reconstruct API: return informative error if backfill not enabled

VolodymyrBg:
## Issue Addressed

This PR addresses issue #4788: State reconstruction API doesn't work if genesis backfill hasn't been enabled

## Proposed Changes

- The /lighthouse/database/reconstruct API endpoint now returns a clear, user-friendly error if the node was started without historic block backfill (i.e., without --genesis-backfill or --reconstruct-historic-states).
- The error message includes instructions to restart the Beacon Node with --reconstruct-historic-states.
- The API handler was refactored to propagate errors from the reconstruction logic to the HTTP response, instead of only logging them.
- Documentation (book/src/api-lighthouse.md) was updated to describe the new API behavior and provide example responses.
Internal logic in process_reconstruction was updated to return a Result, enabling error propagation to the API layer.

## Additional Info

- This change improves UX for users running archival nodes or researchers, making it clear why state reconstruction may fail and how to resolve it.
- No changes were made to the actual reconstruction logic; only error handling and documentation were improved.
- Future consideration: If dynamic enabling of backfill becomes possible, the API could be extended to trigger it directly.
- All changes are backward compatible for nodes already running with the correct flags.

### DIFF
--- a/beacon_node/beacon_chain/src/migrate.rs
+++ b/beacon_node/beacon_chain/src/migrate.rs
@@ -215,13 +215,15 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> BackgroundMigrator<E, Ho
         }
     }
 
-    pub fn process_reconstruction(&self) {
+    pub fn process_reconstruction(&self) -> Result<(), Error> {
         if let Some(Notification::Reconstruction) =
             self.send_background_notification(Notification::Reconstruction)
         {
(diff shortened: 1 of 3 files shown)


michaelsproul:
Hi @VolodymyrBg can you please re-target your PR at the `unstable` branch? We use `unstable` for all development

PR #7655, opened 2025-06-26
Final changes for `fusaka-devnet-2`

### LINKED ISSUE #7467: Support blob parameter forks (BPO) and replace `MAX_BLOBS_PER_BLOCK_FULU`

ethDreamer:
## Issue Addressed

Closes #7467. 

This PR primarily addresses [the P2P changes](https://github.com/ethereum/EIPs/pull/9840) in [fusaka-devnet-2](https://fusaka-devnet-2.ethpandaops.io/). Specifically:

* [the new `nfd` parameter added to the `ENR`](https://github.com/ethereum/EIPs/pull/9840)
* [the modified `compute_fork_digest()` changes for every BPO fork](https://github.com/ethereum/EIPs/pull/9840)

90% of this PR was absolutely hacked together as fast as possible during the Berlinterop as fast as I could while running between Glamsterdam debates. Luckily, it seems to work. But I was unable to be as careful in avoiding bugs as I usually am. I've cleaned up the things *I remember* wanting to come back and have a closer look at. But still working on this.

Progress:
* [x] get it working on `fusaka-devnet-2`
* [ ] [*optional* disconnect from peers with incorrect `nfd` at the fork boundary](https://github.com/ethereum/consensus-specs/pull/4407) - Can be addressed in a future PR if necessary
* [x] first pass clean-up
* [x] fix up all the broken tests
* [x] final self-review
* [x] more thorough review from people more familiar with affected code


Dustin4444 (approved):

Dustin4444 (approved):

lcli/src/generate_bootnode_enr.rs
```diff
@@ -15,6 +15,7 @@ pub fn run<E: EthSpec>(matches: &ArgMatches, spec: &ChainSpec) -> Result<(), Str
     let udp_port: NonZeroU16 = clap_utils::parse_required(matches, "udp-port")?;
     let tcp_port: NonZeroU16 = clap_utils::parse_required(matches, "tcp-port")?;
     let output_dir: PathBuf = clap_utils::parse_required(matches, "output-dir")?;
+    // FIXME: why is this being read from.. somewhere rather than just using the spec?
```
jimmygchen: We aren't actually using the spec from `testnet-dir` for `generate-bootnode-enr` at the moment.

The `spec` that we have access to here is the default mainnet/minimal/gnosis spec and shouldn't be used for all cases:
https://github.com/sigp/lighthouse/blob/ce8d0814ad71819b54c878f8c8a610c48e7eca40/lcli/src/main.rs#L636-L644

This closure loads the network config, and we should use this to derive the correct `chain_spec`
https://github.com/sigp/lighthouse/blob/ce8d0814ad71819b54c878f8c8a610c48e7eca40/lcli/src/main.rs#L695-L708

consensus/types/src/fork_context.rs
```diff
@@ -82,10 +105,10 @@ impl ForkContext {
         self.digest_to_fork.get(&context)
     }
 
-    /// Returns the context bytes/fork_digest corresponding to a fork name.
-    /// Returns `None` if the `ForkName` has not been initialized.
-    pub fn to_context_bytes(&self, fork_name: ForkName) -> Option<[u8; 4]> {
-        self.fork_to_digest.get(&fork_name).cloned()
+    // TODO: we *may* delete this entire object and just use the spec
```
jimmygchen: I think it's _possible_ but might be a bit cleaner to keep this for storing current context. Will keep this for now but also open to consider refactoring.

lcli/src/generate_bootnode_enr.rs
jimmygchen: For bootnodes the `enr_fork_id` and `nfd` are not really relevant I think - given the existing values and the fact that bootnodes on older versions continue to function after hard forks:

https://github.com/sigp/lighthouse/blob/e18b32b882ae56a941d508ba4edb632a8ce861fc/lcli/src/generate_bootnode_enr.rs#L35-L39

jimmygchen (changes requested):
Looks pretty good actually @ethDreamer !
I've made a PR #7722 for the cleanup and test fixes.

lcli/src/generate_bootnode_enr.rs
jimmygchen: Addressed in #7722

pawanjay176:
The sync failures are only on fulu. I'll be handling them in https://github.com/sigp/lighthouse/pull/7726

Gonna merge this one when the remaining checks pass

pawanjay176 (approved):
LGTM

jimmygchen (approved):

PR #7317, opened 2025-04-11
Replace HashSet with BitVector for permanent attestation subscriptions

VolodymyrBg:
## Issue Addressed

This PR addresses issue #XXX: Optimizing memory usage and performance of the subnet service.

## Proposed Changes

- Replaces HashSet<Subnet> with BitVector for storing permanent attestation subscriptions
- Updates related methods to work with bit operations instead of HashSet operations
- Modifies the subscription checks to handle the new BitVector implementation
- Updates test methods to work properly with the new data structure

## Additional Info

This change improves both memory efficiency and performance of the subnet service. BitVector is a more space-efficient representation for tracking subnet subscriptions, which is particularly important for nodes with many validators or high network activity.
The implementation follows a pattern already used elsewhere in the codebase for subnet bitfields, specifically in ENR attestation and sync committee bitfields, creating a more consistent approach to subnet tracking throughout the codebase.

PR #7656, opened 2025-06-27
Fix incorrect `waker` update condition

jimmygchen:
## Issue Addressed

This bug was first found and partially fixed by @VolodymyrBg in #7317 - this PR applies the same fix everywhere else.

The old logic updated the waker when it already matched the context, and did nothing when it was stale:

```rust
if waker.will_wake(cx.waker()) {
    self.waker = Some(cx.waker().clone());
}
```

This is the wrong way around. We only want to update the waker if it doesn't match the current context:

```rust
if !waker.will_wake(cx.waker()) {
    self.waker = Some(cx.waker().clone());
}
```

I don't think we've ever noticed any issues, but it’s a subtle bug that could lead to missed wakeups.


pawanjay176 (approved):
Nice catch!

PR #7658, opened 2025-06-27
fix: allow validator import flags with --validators-file (fixes #7651)

### LINKED ISSUE #7651: Validator manager import to allow `--suggested-fee-recipient` for all keystore files
In `lighthouse vm import`, the flag `--suggested-fee-recipient` currently is only tied to the `--keystore-file` (the keystore file format created using ethstaker-deposit-cli). So if a user creates the `--validators-file` using `lighthouse vm create` and using `--suggested-fee-recipient` flag during importing, it generates an unclear error message: 

```
error: the following required arguments were not provided:
  --password <STRING>
  --keystore-file <PATH_TO_KEYSTORE_FILE>

Usage: lighthouse validator_manager import --password <STRING> --keystore-file <PATH_TO_KEYSTORE_FILE> --validators-file <PATH_TO_JSON_FILE> --vc-token <PATH> --network <network> --suggested-fee-recipient <ETH1_ADDRESS>
```

We can remove the `.requires` line here: 
https://github.com/sigp/lighthouse/blob/999b04517e35f8fce78525132d0b8e0a4816f31e/validator_manager/src/import_validators.rs#L116

so that the fee recipient can be set during import for both types of keystore files

Thanks to @anisoptera for highlighting the issue https://discord.com/channels/605577013327167508/938294480279334912/1387583324922449991

Edit: some other flags are also having the same issue, such as `builder-boost-factor`, `builder-proposals`

Raveesh1007:
## Summary
This PR fixes issue #7651 by allowing validator import flags to work with both `--keystore-file` and `--validators-file` options, improving the UX for users importing validators.

## Changes Made
- **Removed `.requires(KEYSTORE_FILE_FLAG)` constraint** from the following flags:
  - `--suggested-fee-recipient`
  - `--gas-limit` 
  - `--builder-proposals`
  - `--builder-boost-factor`
  - `--prefer-builder-proposals`
- **Added missing `--enabled` flag** definition that was referenced in the `ImportConfig` struct but missing from CLI arguments

## Testing
-  Code compiles successfully
-  CLI arguments are parsed correctly
- The old error about `--password` and `--keystore-file` being required is resolved
- All flags now work with both `--keystore-file` and `--validators-file`

## Fixes
Closes #7651

### DIFF
--- a/validator_manager/src/import_validators.rs
+++ b/validator_manager/src/import_validators.rs
@@ -112,8 +112,7 @@ pub fn cli_app() -> Command {
                 .value_name("ETH1_ADDRESS")
                 .help("When provided, the imported validator will use the suggested fee recipient. Omit this flag to use the default value from the VC.")
                 .action(ArgAction::Set)
-                .display_order(0)
-                .requires(KEYSTORE_FILE_FLAG),
+                .display_order(0),
         )
         .arg(
             Arg::new(GAS_LIMIT)
@@ -122,8 +121,7 @@ pub fn cli_app() -> Command {
                 .help("When provided, the imported validator will use this gas limit. It is recommended \
                 to leave this as the default value by not specifying this flag.",)
                 .action(ArgAction::Set)
-                .display_order(0)
-                .requires(KEYSTORE_FILE_FLAG),
+                .display_order(0),
         )
         .arg(
             Arg::new(BUILDER_PROPOSALS)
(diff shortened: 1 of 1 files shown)


michaelsproul:
Hi @Raveesh1007, thanks for the contribution but I'm going to close this PR as it doesn't solve the issue, and the issue was also already assigned to @chong-he.

If you want to jump onto another issue please just write a comment saying you intend to work on it so we can avoid duplication of effort!

For future PRs it's also good if you can:

- Target the PR at `unstable`
- Sign the user contribution agreement

Thanks!

Raveesh1007:
Thanks I'll remember this and contribute with that in mind.

PR #7661, opened 2025-06-27
Custody persist fix

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Persist the epoch -> cgc values. This is to ensure that `ValidatorRegistrations::latest_validator_custody_requirement` always returns a `Some` value post restart assuming the `epoch_validator_custody_requirements` map has been updated in the previous runs.

### DIFF
--- a/beacon_node/beacon_chain/src/schema_change/migration_schema_v26.rs
+++ b/beacon_node/beacon_chain/src/schema_change/migration_schema_v26.rs
@@ -34,24 +34,31 @@ impl StoreItem for PersistedCustodyV1 {
 pub fn upgrade_to_v26<T: BeaconChainTypes>(
     db: Arc<HotColdDB<T::EthSpec, T::HotStore, T::ColdStore>>,
 ) -> Result<Vec<KeyValueStoreOp>, Error> {
-    let res = db.get_item::<PersistedCustodyV1>(&CUSTODY_DB_KEY);
-    let ops = match res {
-        Ok(Some(PersistedCustodyV1(v1))) => {
-            info!("Migrating `CustodyContext` to v26 schema");
-            let custody_context_v2 = CustodyContextSsz {
-                validator_custody_at_head: v1.validator_custody_at_head,
-                persisted_is_supernode: v1.persisted_is_supernode,
-                epoch_validator_custody_requirements: vec![],
-            };
-            vec![KeyValueStoreOp::PutKeyValue(
-                DBColumn::CustodyContext,
-                CUSTODY_DB_KEY.as_slice().to_vec(),
-                PersistedCustody(custody_context_v2).as_store_bytes(),
-            )]
+    let ops = if db.spec.is_peer_das_scheduled() {
+        match db.get_item::<PersistedCustodyV1>(&CUSTODY_DB_KEY) {
+            Ok(Some(PersistedCustodyV1(v1))) => {
+                info!("Migrating `CustodyContext` to v26 schema");
+                let custody_context_v2 = CustodyContextSsz {
+                    validator_custody_at_head: v1.validator_custody_at_head,
+                    persisted_is_supernode: v1.persisted_is_supernode,
+                    epoch_validator_custody_requirements: vec![],
+                };
+                vec![KeyValueStoreOp::PutKeyValue(
+                    DBColumn::CustodyContext,
+                    CUSTODY_DB_KEY.as_slice().to_vec(),
+                    PersistedCustody(custody_context_v2).as_store_bytes(),
+                )]
+            }
+            _ => {
+                vec![]
+            }
         }
-        _ => {
-            vec![]
-        }
+    } else {
+        // Delete it from db if PeerDAS hasn't been scheduled
+        vec![KeyValueStoreOp::DeleteKey(
+            DBColumn::CustodyContext,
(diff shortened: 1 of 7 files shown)


jimmygchen:
I'm creating a PR to only persist custdoy context _if_ peerdas is scheduled, and also make a schema change to un-store the existing format (as our testnet nodes already have this on disk).

Marking this as `blocked` for now until we merge that.

jimmygchen:
@pawanjay176  I'll just roll the change into this PR to reduce schema bumps (thanks @michaelsproul)

beacon_node/beacon_chain/src/schema_change/migration_schema_v26.rs
```diff
@@ -0,0 +1,83 @@
...
+        Ok(PersistedCustodyV1(custody_context))
+    }
+}
+
+/// Upgrade the `CustodyContext` entry to v26.
+pub fn upgrade_to_v26<T: BeaconChainTypes>(
+    db: Arc<HotColdDB<T::EthSpec, T::HotStore, T::ColdStore>>,
+) -> Result<Vec<KeyValueStoreOp>, Error> {
+    let res = db.get_item::<PersistedCustodyV1>(&CUSTODY_DB_KEY);
+    let ops = match res {
```
michaelsproul: Do you want to try deleting it if Fusaka is not scheduled? That might give us a bit more freedom to experiment with the format, without committing to writing data on non-Fusaka nodes

michaelsproul: OTOH, it's simpler if the DB schema isn't fork-dependent

jimmygchen: yeah good point, i'll add that

jimmygchen: Updated

jimmygchen (approved):
@pawanjay176's changes look good to me. 

This change is required because if `epoch_validator_custody_requirements` is not persisted, the node would revert back to only sampling the minimal custody requirement after a restart.

michaelsproul:
New schema stability test is working as expected! It will need updating

beacon_node/beacon_chain/src/schema_change/migration_schema_v26.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    fn from_store_bytes(bytes: &[u8]) -> Result<Self, Error> {
+        let custody_context = CustodyContextSszV1::from_ssz_bytes(bytes)?;
+        Ok(PersistedCustodyV1(custody_context))
+    }
+}
+
+/// Upgrade the `CustodyContext` entry to v26.
+pub fn upgrade_to_v26<T: BeaconChainTypes>(
```
dapplion, line 34: Why do we need a migration?

michaelsproul: Because the previous PR (https://github.com/sigp/lighthouse/pull/7578) started storing the `CustodyContextSsz` on disk by default, even when Fulu is disabled. This version of the code got rolled out on all our testnet infra while I was testing v7.1.0, so all those nodes have `CustodyContext`s in the old format. This PR changes the format of `CustodyContext`, so these nodes would fail to start if they were upgraded to run the new code without a migration.

In general, I want to try to maintain continuity of the database schema, even for commits on unstable. If we don't maintain it, we risk breaking testnet nodes used by us/pandaops/etc.

It's a little bit of extra overhead in the case where we are implementing a new fork, but I think it's worth it given the headaches caused by accidental schema changes. This recent issue is what motivated me to add this test:

- https://github.com/sigp/lighthouse/pull/7669

If we _do_ want to change the DB schema, we should at least be aware we are changing it, and consciously opt in to breaking some nodes. For Fulu, this PR also changes the `CustodyContext` so it's only stored on Fulu-enabled networks. This gives us a bit more wiggle room, as we can choose to _just_ break Fulu testnet nodes, rather than breaking _all_ nodes which ran the previous version of `unstable`.

beacon_node/beacon_chain/src/schema_change/migration_schema_v26.rs
```diff
@@ -0,0 +1,91 @@
...
+use crate::validator_custody::CustodyContextSsz;
+use crate::BeaconChainTypes;
+use ssz::{Decode, Encode};
+use ssz_derive::{Decode, Encode};
+use std::sync::Arc;
+use store::{DBColumn, Error, HotColdDB, KeyValueStoreOp, StoreItem};
+use tracing::info;
+
+#[derive(Debug, Encode, Decode, Clone)]
+pub(crate) struct CustodyContextSszV1 {
```
michaelsproul: Usually we use the first version where the data appeared to name it, but it doesn't matter too much: https://github.com/sigp/lighthouse/blob/stable/beacon_node/beacon_chain/src/schema_change/README.md#naming-conventions

```suggestion
pub(crate) struct CustodyContextSszV24 {
```

michaelsproul (approved):

beacon_node/beacon_chain/src/schema_change/migration_schema_v26.rs
jimmygchen: Thanks, updated!

PR #7663, opened 2025-06-28
Error from RPC `send_response` when request doesn't exist on the active inbound requests

jxs:
## Issue Addressed
Lighthouse is currently loggign a lot errors in the `RPC` behaviour whenever a response is received for a request_id that no longer exists in active_inbound_requests. This is likely due to a data race or timing issue (e.g., the peer disconnecting before the response is handled).
This PR addresses that by removing the error logging from the RPC layer. Instead, RPC::send_response now simply returns an Err, shifting the responsibility to the main service. The main service can then determine whether the peer is still connected and only log an error if the peer remains connected.
Thanks @ackintosh for helping debug!

### DIFF
--- a/beacon_node/lighthouse_network/src/rpc/mod.rs
+++ b/beacon_node/lighthouse_network/src/rpc/mod.rs
@@ -98,6 +98,13 @@ pub struct InboundRequestId {
     substream_id: SubstreamId,
 }
 
+// An Active inbound request received via Rpc.
+struct ActiveInboundRequest<E: EthSpec> {
+    pub peer_id: PeerId,
+    pub request_type: RequestType<E>,
+    pub peer_disconnected: bool,
+}
+
 impl InboundRequestId {
     /// Creates an _unchecked_ [`InboundRequestId`].
     ///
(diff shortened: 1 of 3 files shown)


beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -209,26 +215,41 @@ impl<Id: ReqId, E: EthSpec> RPC<Id, E> {
...
-            error!(%peer_id, ?request_id, %response, "Request not found in active_inbound_requests. Response not sent");
-            return;
+            return Err(response);
         };
 
+        if peer_disconnected {
+            trace!(%peer_id, ?request_id, %response,
+                "Discarding response, peer is no longer connected");
+            return Ok(());
+        }
```
ackintosh: Maybe this `if peer_disconnected {` block might be moved to before `self.send_response_inner(...)` on line 251, since we need to re-insert active_inbound_requests for certain types of requests (see line 238).

beacon_node/lighthouse_network/src/rpc/mod.rs
```diff
@@ -209,26 +215,41 @@ impl<Id: ReqId, E: EthSpec> RPC<Id, E> {
...
             && matches!(response, RpcResponse::Success(_))
         {
-            self.active_inbound_requests
-                .insert(request_id, (peer_id, request_type.clone()));
+            self.active_inbound_requests.insert(
+                request_id,
+                ActiveInboundRequest {
+                    peer_id,
+                    request_type: request_type.clone(),
+                    peer_disconnected: false,
```
ackintosh: If we address [this comment](https://github.com/sigp/lighthouse/pull/7663#discussion_r2191139400), the `peer_disconnected` value should come from active_inbound_requests:

```suggestion
                    peer_disconnected,
```

michaelsproul (approved):
Merging ahead of release ✅

Issue #7181, opened 2025-03-20
Feature request: allow me to initialize the p2p private key file with new flag


barnabasbusa:
Currently the lighthouse p2p private key is stored under beacon-data/network/key binary file. 

This is pretty inconvenient. It would be great if lighthouse would have a flag that would allow me to set the private key for the node using a new flag like `--p2p-priv-key=` that would point to a file. The content of this file is presumed to contain a hex  representing Secp256 key for example `e156614877adbfac274118a3be4b7dbfcd6d994126ab7e758dd7aa688f8ec0da`.

Lighthouse could convert this hex to binary, so that it wouldn't become a breaking change.
Then lighthouse should use this key to initialize its p2p address, which would allow us to test custom subnets for each peer.

Ayushdubey86:
can you assign this to me!

Ayushdubey86:
Hello @barnabasbusa@michaelsproul , Check out this [pr](https://github.com/sigp/lighthouse/pull/7191) once free!

AgeManning:
@barnabasbusa - What if we just used a hex encoding in our `key` file. You could just put whatever key you wanted into that file `beacon/network/key` without having to specify a new file via CLI? Or is that too cumbersome still?

barnabasbusa:
that’s also good, but wouldn’t that be a breaking change on your side?

AgeManning:
Yeah, kinda (we could support both formats). We do breaking changes a bit. If we just delete the old key create a new one in hex, I'm not sure anyone is really going to get upset about it, we just change PeerId's on restart, which is what Prysm does by default. 

But I also like the soln @Ayushdubey86 has already made, so can go with that.

Ayushdubey86:
Do checkout this [pr](https://github.com/sigp/lighthouse/pull/7191) once again, I have pushed further changes , @barnabasbusa @jimmygchen  @michaelsproul  @AgeManning

dapplion:
@Ayushdubey86 please do not tag people like this, relevant reviewers are notified of GH notifications already 🙏

masonreid1729:
Yes, there’s some kind of givеаway going on. I got some Ethereum there yesterday for different wallets. It’s unlikely they’ll take it back from me, but it’s a pity I didn’t have more wallets and couldn’t get more. But maybe it’s for the best.

<img width="373" alt="Image" src="https://github.com/user-attachments/assets/05cec7fc-91e3-4c46-92d5-a99ee58d5ef7" />

[...]

PR #7665, opened 2025-06-30
Transition network key to hex format

AgeManning:
## Issue Addressed

#7181 

## Proposed Changes

Instead of storing the network key as binary data we store it as hex, allowing users to modify it via the file. 

We can read old-binary forms, however we will migrate binary to hex as it will be the new standard.


michaelsproul:
Noting that this isn't backwards compat because old versions of LH won't work with the hex-formatted keys.

jimmygchen (approved):
Looks good!

jimmygchen:
@mergify requeue

PR #7666, opened 2025-06-30
Bump DB schema to v25

michaelsproul:
## Issue Addressed

When we removed the eth1 data, I wrote a v25 schema upgrade to delete the data on disk:

- https://github.com/sigp/lighthouse/pull/7133

However, I forgot to update the current schema version, so this change was never actioned.

## Proposed Changes

This PR updates the current schema version to v25 so that the migration runs.

## Additional Info

I'm getting this in before v26, which is required to fix up the persistence of validator custody data on disk before v7.1.0 is released. Presently we have introduced on-disk changes without a DB schema change in:

- https://github.com/sigp/lighthouse/pull/7578

### DIFF
--- a/beacon_node/store/src/metadata.rs
+++ b/beacon_node/store/src/metadata.rs
@@ -4,7 +4,7 @@ use ssz::{Decode, Encode};
 use ssz_derive::{Decode, Encode};
 use types::{Hash256, Slot};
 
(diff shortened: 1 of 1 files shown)


jimmygchen (approved):
LGTM

PR #7669, opened 2025-06-30
Test database schema stability

michaelsproul:
## Proposed Changes

This PR implements some heuristics to check for breaking database changes. The goal is to prevent accidental changes to the database schema occurring without a version bump.

## Additional Info

For example, this test would fail on the following PRs, forcing a schema change/test update:

- https://github.com/sigp/lighthouse/pull/7578 (already merged, broke the schema)
- https://github.com/sigp/lighthouse/pull/7661 (v26 added to prevent breakage)
- https://github.com/sigp/lighthouse/pull/7648 (needs a schema increment, probably v27)


jimmygchen (approved):
Nice!

PR #7675, opened 2025-06-30
feat: Optimise `HDiffBuffer` by using Arc instead of Vec

### LINKED ISSUE #6579: Try storing `Arc<Validator>` in `HDiffBuffer`
## Description

The `HDiffBuffer` now contains a `Vec<Validator>` which we could optimise by using `Arc`:

- https://github.com/sigp/lighthouse/pull/6556#discussion_r1833597899

The validator entries come from a milhouse List originally where they are arced, so we could clone these arcs and share memory between any live milhouse lists and hdiff buffers with the same data. For the freezer DB this isn't particularly likely to happen, but once we start implementing hdiffs in the hot DB, it will be quite common.

PoulavBhowmick03:
## Issue Addressed

Fixes #6579 

## Proposed Changes

Modified `HDiffBuffer` to use `Vex<Arc<Validator>>` instead of `Vec<Validator>`, and made necessary changes regarding that

### DIFF
--- a/beacon_node/store/src/hdiff.rs
+++ b/beacon_node/store/src/hdiff.rs
@@ -100,7 +100,7 @@ pub struct HDiffBuffer {
     state: Vec<u8>,
     balances: Vec<u64>,
     inactivity_scores: Vec<u64>,
-    validators: Arc<[Validator]>,
+    validators: Vec<Arc<Validator>>,
     historical_roots: Vec<Hash256>,
     historical_summaries: Vec<HistoricalSummary>,
 }
@@ -183,8 +183,11 @@ impl HDiffBuffer {
             // is post altair, all its items will show up in the diff as is.
             vec![]
         };
-        let validators = std::mem::take(beacon_state.validators_mut()).to_vec();
-        let validators = Arc::from(validators);
+        let validators = std::mem::take(beacon_state.validators_mut())
+            .iter()
+            .cloned()
+            .map(Arc::new)
+            .collect::<Vec<_>>();
         let historical_roots = std::mem::take(beacon_state.historical_roots_mut()).to_vec();
         let historical_summaries =
             if let Ok(historical_summaries) = beacon_state.historical_summaries_mut() {
@@ -222,8 +225,9 @@ impl HDiffBuffer {
                 .map_err(|_| Error::InvalidBalancesLength)?;
         }
 
-        *state.validators_mut() = List::try_from_iter(self.validators.iter().cloned())
-            .map_err(|_| Error::InvalidBalancesLength)?;
+        *state.validators_mut() =
+            List::try_from_iter(self.validators.iter().map(|arc| (**arc).clone()))
+                .map_err(|_| Error::InvalidBalancesLength)?;
 
         *state.historical_roots_mut() = List::try_from_iter(self.historical_roots.iter().copied())
             .map_err(|_| Error::InvalidBalancesLength)?;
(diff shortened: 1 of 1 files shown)


beacon_node/store/src/hdiff.rs
```diff
@@ -282,8 +283,11 @@ impl HDiff {
         self.balances_diff().apply(&mut source.balances, config)?;
         self.inactivity_scores_diff()
             .apply(&mut source.inactivity_scores, config)?;
-        self.validators_diff()
-            .apply(&mut source.validators, config)?;
+
+        let mut validators_vec = source.validators.to_vec();
+        self.validators_diff().apply(&mut validators_vec, config)?;
+        source.validators = Arc::from(validators_vec);
```
michaelsproul: I don't see how using `Arc<[Validator]>` helps when we do this conversion to and from `Vec<Validator>` whenever we apply a diff.

Are you trying to save memory or CPU cycles?

beacon_node/store/src/hdiff.rs
```diff
@@ -184,6 +184,7 @@ impl HDiffBuffer {
             vec![]
         };
         let validators = std::mem::take(beacon_state.validators_mut()).to_vec();
+        let validators = Arc::from(validators);
```
michaelsproul: If we did want to save memory, a promising approach might be to keep the `List`, which is a copy-on-write (persistent) data structure

However, `List`s have slower `O(log n)` indexing than vecs, and we do random-access indexing here:

https://github.com/sigp/lighthouse/blob/6be646ca1153301c24f030f3190d9941d3f6b285/beacon_node/store/src/hdiff.rs#L553

We also do it here, but this use could be replaced by a `zip` on two iterators:

https://github.com/sigp/lighthouse/blob/6be646ca1153301c24f030f3190d9941d3f6b285/beacon_node/store/src/hdiff.rs#L461

beacon_node/store/src/hdiff.rs
PoulavBhowmick03: You're right. i was trying to save memory as the issue mentioned, but converting to Vec<> fails that purpose.

michaelsproul: Oh yeah I forgot about that issue, I think the structure I was imagining was `Vec<Arc<Validator>>`

PoulavBhowmick03: Got it. It has to be a collection of individual Arcs. Made a mistake on wrapping all the validators in a single Arc. Will be changing that

beacon_node/store/src/hdiff.rs
```diff
@@ -183,7 +183,11 @@ impl HDiffBuffer {
             // is post altair, all its items will show up in the diff as is.
             vec![]
         };
-        let validators = std::mem::take(beacon_state.validators_mut()).to_vec();
+        let validators = std::mem::take(beacon_state.validators_mut())
+            .iter()
+            .cloned()
+            .map(Arc::new)
```
michaelsproul, line 189: This is maybe OK, but we could make it even better if we added an `iter_arc` method to `List`. In the case of validators, they are not packed in leaves, and are stored behind an `Arc`:

https://github.com/sigp/milhouse/blob/6da903ce8f783295714a68de4a2dcfc8b7c6ee01/src/leaf.rs#L17

Would you be interested in making a PR to `milhouse` to add this? I think the API would have to consider the case where `T` is packed, and return an error in this case. Something like:

```rust
pub fn iter_arc(&self) -> Result<impl Iterator<Item = &Arc<T>>, Error> {
   if T::tree_hash_type() == TreeHashType::Basic {
       // Can't return `Arc`s for packed leaves.
       return Err(Error::PackedLeavesNoArc);
   }
   // TODO
   Ok(iterator)
}
```

I'm happy to help advise on the implementation. Take a look at the other iterators in:
 
* https://github.com/sigp/milhouse/blob/main/src/iter.rs
* https://github.com/sigp/milhouse/blob/main/src/interface_iter.rs

PoulavBhowmick03: Sure! I will look into it and try to implement the `iter_arc` method to `List`. Leaving a comment on the issue in Milhouse

michaelsproul: Thanks!

PR #7679, opened 2025-07-01
Add a guide to partially reconstruct historic states to Lighthouse book

chong-he:
## Proposed Changes

The main change is adding a guide to partially reconstruct historic states to the FAQ. 

Other changes: 
- Update the database scheme info
- Delete the Homebrew issue as it has been solved in https://github.com/Homebrew/homebrew-core/pull/225877
- Update default gas limit in: [7cbf7f1](https://github.com/sigp/lighthouse/pull/7679/commits/7cbf7f15164788065328d6848c9034608badbc59)
- Updated the binary installation page [8076ca7](https://github.com/sigp/lighthouse/pull/7679/commits/8076ca79059cd312f08c885162ba1edd0c203772) as Lighthouse now supports aarch-apple binary built since v7.1.0


book/src/faq.md
```diff
@@ -190,6 +191,40 @@ If the node is syncing or downloading historical blocks, the error should disapp
...
+  curl "http://localhost:5052/lighthouse/database/info" | jq '.anchor'
+  ```
+
+  and look for the field `state_upper_limit`. It should show:
+
+  ```json
+  "state_upper_limit": "10485760",
+  ```
+
+Lighthouse will now start to reconstruct the historic states from `slot 10485760`. At this point, if you do not want a full state reconstruction, you may remove the flag `--reconstruct-historic-states`. When the process is completed, you will have the state data from `slot 10485760`.
```
macladson: So if I'm understanding correctly, the flag `--reconstruct-historic-states` is only necessary here to "trick" the database into thinking it is going to perform state reconstruction and so doesn't prune states? Otherwise this is just effectively checkpoint syncing to an older state and then syncing forwards as normal (minus state pruning). Does that mean that over time you'd have to repeat these steps to control state db growth?

michaelsproul: Yeah, once you sync like this your node keeps all states newer than the checkpoint. In practice it shouldn't be too bad, because state diffs are magic, but I guess users doing this are really interested in saving space. We could add a note along these lines?

chong-he: If state pruning is already disabled with the flag `--reconstruct-historic-states`, how to prune states again? Is it done via the manual pruning steps? 

So if saving space is a goal, then have to kind of do the manual state pruning from time to time? Because once `--reconstruct-historic-state` is used, automatic state pruning will not be active for the database from my understanding

michaelsproul: Yeah I added some text at the end here which just says to repeat the process

https://github.com/sigp/lighthouse/pull/7679/commits/9c6327c09cfafbe5c539225d4f99af2d76abcb98

chong-he:
Updated the binary installation page [8076ca7](https://github.com/sigp/lighthouse/pull/7679/commits/8076ca79059cd312f08c885162ba1edd0c203772) as Lighthouse now supports aarch-apple binary built since v7.1.0: `lighthouse-v7.1.0-aarch64-apple-darwin.tar.gz` in the release page: https://github.com/sigp/lighthouse/releases/tag/v7.1.0

michaelsproul (approved):

PR #7683, opened 2025-07-01
Update `SAMPLES_PER_SLOT` to be number of custody groups instead of data columns

jimmygchen:
## Issue Addressed

Update `SAMPLES_PER_SLOT` to be number of custody groups instead of data columns. This should have no impact on the current implementation as config currently maintains a `group:subnet:column` ratio of `1:1:1`.  **In short, this PR doesn't change anything for Fusaka, but ensures compliance with the spec and potential future changes.**

I've added separate methods to compute sampling columns and custody groups for clarity: `spec.sampling_size_columns` and `spec.sampling_size_custod_groups` 

See the clarifications in this PR for more details: https://github.com/ethereum/consensus-specs/pull/4251

### DIFF
--- a/beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
+++ b/beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
@@ -481,7 +481,8 @@ impl<T: BeaconChainTypes> DataAvailabilityCheckerInner<T> {
 
         if let Some(available_block) = pending_components.make_available(
             &self.spec,
-            self.custody_context.sampling_size(Some(epoch), &self.spec),
+            self.custody_context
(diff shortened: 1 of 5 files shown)


dapplion (approved):
Looks good! Tho when working on validator custody I found dangerous the interchangeability of cgc values, column count etc. We could define extra types like Slot for Cgc and DataColumnCount, or is that too much hastle? Of course in another PR

jimmygchen:
> Looks good! Tho when working on validator custody I found dangerous the interchangeability of cgc values, column count etc. We could define extra types like Slot for Cgc and DataColumnCount, or is that too much hastle? Of course in another PR

Sounds like a great idea to me, I'll make a PR!

PR #7684, opened 2025-07-01
Validator manager import to allow overriding fields with CLI flag

chong-he:
## Issue Addressed
* #7651


validator_manager/src/import_validators.rs
```diff
@@ -225,48 +229,92 @@ async fn run(config: ImportConfig) -> Result<(), String> {
...
-            }]
-        } else {
-            return Err(format!(
-                "One of the flag --{VALIDATORS_FILE_FLAG} or --{KEYSTORE_FILE_FLAG} is required."
-            ));
-        };
+            })?;
+
+        // Log the overridden note when one or more flags is supplied
+        if let Some(override_fee_recipient) = fee_recipient {
```
chong-he, line 258: This logging is very repetitive, I put them here mainly because the overriding of `suggested-fee-recipient`, but thought we should do the same for others too. Not sure if we want this logging?

michaelsproul:
@macladson would you mind reviewing this? :pray:

validator_manager/src/import_validators.rs
```diff
@@ -225,48 +229,92 @@ async fn run(config: ImportConfig) -> Result<(), String> {
...
+            eprintln!("Please note! --prefer-builder-proposals is provided. This will override existing prefer builder proposal setting defined in validators.json with: {}", override_prefer_builder_proposals);
+        }
+        if let Some(override_enabled) = enabled {
+            eprintln!("Please note! --enabled flag is provided. This will override existing setting defined in validators.json with: {}", override_enabled);
+        }
+
+        // Override the fields in validators.json file if the flag is supplied
+        for validator in &mut validators {
+            if let Some(override_fee_recipient) = fee_recipient {
+                validator.fee_recipient = Some(override_fee_recipient);
```
macladson, line 298: We could move the logging inside the for loop? I doubt we need to log anything in the case that `validators.len() == 0` and it would consolidate the `if let`s

chong-he: Thanks for the review. I got what you mean. 

Previously, before consolidating the `if let`, the message would be logged whenever the flags (e.g., `-suggested-fee-recipient`) are used together with `lighthouse vm import`, even though there are no validators

Now, after moving in the `if` loop, when there is no validators, the `if` loop will not be executed, so no message is logged. Makes complete sense

Revise in [6237ab3](https://github.com/sigp/lighthouse/pull/7684/commits/6237ab308c0529569c14fd3a7eccd8cd5486c361)

validator_manager/src/import_validators.rs
```diff
@@ -225,48 +229,99 @@ async fn run(config: ImportConfig) -> Result<(), String> {
...
-                .open(validators_format_path)
-                .map_err(|e| format!("Unable to open {:?}: {:?}", validators_format_path, e))?;
+        let validators_file = fs::OpenOptions::new()
+            .read(true)
+            .create(false)
+            .open(validators_format_path)
+            .map_err(|e| format!("Unable to open {:?}: {:?}", validators_format_path, e))?;
 
-            serde_json::from_reader(&validators_file).map_err(|e| {
+        // Define validators as mutable as that if CLI flag is supplied, the fields can be overridden
```
macladson:
```suggestion
        // Define validators as mutable so that if a relevant flag is supplied, the fields can be overridden.
```

validator_manager/src/import_validators.rs
```diff
@@ -225,48 +229,99 @@ async fn run(config: ImportConfig) -> Result<(), String> {
...
-        };
+            })?;
+
+        // Override the fields in validators.json file if the flag is supplied
+        for validator in &mut validators {
+            if let Some(override_fee_recipient) = fee_recipient {
+                validator.fee_recipient = Some(override_fee_recipient);
+                eprintln!(
+                    "Please note! --suggested-fee-recipient is provided. This will override existing fee recipient defined in validators.json with: {:?}",
+                    override_fee_recipient
```
macladson: Ah so as a consequence of moving the logs inside the for loop, this will log a separate instance of each `Please note!` log for every validator in your `validator.json` file. So we should either:
1. Add the validator public key to the log line. This makes it clear that each validator will have the field overridden.
2. Move the logging outside back outside the for loop but add a conditional that it doesn't run when `validators.len() == 0`.

I think I'm fine with either, it's just a matter of which we feel is more ergonomic (1 log per validator vs 1 log total for all validators). For 100s of validators, 1 log per validator would likely be too much. But if you had 2 validators, having a reminder that the override will apply to both might be nice

macladson (changes requested):
This looks pretty good! Just needs some fixes for the logging behaviour

validator_manager/src/import_validators.rs
chong-he: I am leaning towards no. 2, I will revise

chong-he: I added the check `if !validators.is_empty()` (for `validators.len() == 0`), but I am thinking that this may not be exactly accurate? 

For example, when we import new validators and the VC has no validators, then the logs `Please note! ...` wouldn't be shown, which is not really the intention. i.e., it's not only about modifying existing validators, it can also be used to modify the fields for newly added validators (including the scenario when the VC has no validators)

dapplion (commented):
Minor bug: `prefer_builder_proposals` is logged as an override (line 283) but is missing from the actual override loop (lines 298-314). The other five fields (`fee_recipient`, `gas_limit`, `builder_proposals`, `builder_boost_factor`, `enabled`) are all applied, but `prefer_builder_proposals` is skipped.

```rust
// Override the fields in validators.json file if the flag is supplied
for validator in &mut validators {
    if let Some(override_fee_recipient) = fee_recipient {
        validator.fee_recipient = Some(override_fee_recipient);
    }
    if let Some(override_gas_limit) = gas_limit {
        validator.gas_limit = Some(override_gas_limit);
    }
    if let Some(override_builder_proposals) = builder_proposals {
        validator.builder_proposals = Some(override_builder_proposals);
    }
    if let Some(override_builder_boost_factor) = builder_boost_factor {
        validator.builder_boost_factor = Some(override_builder_boost_factor);
    }
    if let Some(override_enabled) = enabled {
        validator.enabled = Some(override_enabled);
    }
    // missing: prefer_builder_proposals
}
```

chong-he:
> Minor bug: `prefer_builder_proposals` is logged as an override (line 283) but is missing from the actual override loop (lines 298-314). The other five fields (`fee_recipient`, `gas_limit`, `builder_proposals`, `builder_boost_factor`, `enabled`) are all applied, but `prefer_builder_proposals` is skipped.


Thank you for the comment, addressed in [here](https://github.com/sigp/lighthouse/pull/7684/commits/5fd3303ffc616fda101ef116c0b693106398dc79)

validator_manager/src/import_validators.rs
chong-he: I removed the `if !validators.is_empty()` condition with reason stated above: [2239b55](https://github.com/sigp/lighthouse/pull/7684/commits/2239b558de63ddcc9ed8e4a44e42109ebff7d338)

Let me know if this is no good and I am happy to revert the change

dapplion (approved):
LGTM!

macladson (approved):
Nice work! This is very clean

PR #7639, opened 2025-06-23
Draft: Add PostgresDB backend support for Lighthouse Beacon Node

owanikin:
## Proposed Changes

- Add initial scaffolding for postgres_impl.rs based on leveldb_impl.rs
- Setup feature flag in Cargo.toml for Postgres backend
- Add placeholder KeyValueStore trait impl using sqlx
- Integration points for Beacon Node backend to support Postgres

## Additional Info
- This is an early draft PR meant to gather feedback on approach, structure, and trait compatibility.

PR #7685, opened 2025-07-01
feat(store): Add PostgresDB backend with basic get/put and tests

owanikin:
## Issue Addressed

Supersedes: https://github.com/sigp/lighthouse/pull/7639  
Context: Initial draft PR for adding a PostgresDB backend to the Beacon Node store.

## Proposed Changes

- Adds `PostgresDB` backend implementation to `store` crate.
- Currently implemented methods:
  - `get_bytes`
  - `put_bytes`
- Includes a test in `tests/postgres_store.rs` that exercises these methods using a real Postgres database.

## Additional Info

- This is a **WIP** PR seeking early feedback before full integration into `interface.rs`.
- Other methods (`do_atomically`, `key_exists`, etc.) will be added after initial review, wiring, and interface agreement.
- Design goal: make the Beacon Node database pluggable with async-safe backends like Postgres for NFS compatibility.

@eserilev


beacon_node/store/src/migrations/20250626130431_init.sql
```diff
@@ -0,0 +1,9 @@
+create table if not exists store (
+    col text not null,
+    key bytea not null,
+    value bytea not null,
+    primary key (col, key)
+);
+
+INSERT INTO store (col, key, value)
+VALUES ('example', decode('aabbcc', 'hex'), decode('112233', 'hex'));
```
eserilev: I dont think we need this file?

beacon_node/store/tests/postgres_store.rs
```diff
@@ -0,0 +1,27 @@
...
+    db.put_bytes(DBColumn::BeaconBlock, key, value)
+        .await
+        .expect("put failed");
+
+    let result = db.get_bytes(DBColumn::BeaconBlock, key)
+        .await
+        .expect("get failed");
+
+    assert_eq!(result, Some(value.to_vec()));
+}
```
eserilev: We can delete this test eventually. We should instead enable our test suite to run with a postgres db backend enabled

eserilev:
Looks like a good start.

I don't think we'll want to provide a database url via env variable, we'll probably want to do that via a cli flag.

owanikin:
@eserilev I've pushed new commits addressing the recent work on the Postgres backend. It's ready for review when you have time. Let me know what you think.

beacon_node/store/Cargo.toml
```diff
@@ -25,14 +27,17 @@ redb = { version = "2.1.3", optional = true }
...
 state_processing = { workspace = true }
 strum = { workspace = true }
 superstruct = { workspace = true }
+tokio = { workspace = true }
 tracing = { workspace = true }
 tracing-subscriber = { workspace = true }
 types = { workspace = true }
 xdelta3 = { workspace = true }
 zstd = { workspace = true }
+dotenvy = "0.15.7"
```
eserilev: i don't think we need `dotenvy`

dapplion:
Closing for inactivity

Issue #7690, opened 2025-07-02
`BlockId::root` doesn't work at `oldest_block_slot`


michaelsproul:
While doing some testing I noticed that we're unable to load the block at the `oldest_block_slot` by slot:

```
curl -s "http://localhost:5052/eth/v1/beacon/blinded_blocks/3174625" | jq
{
  "code": 404,
  "message": "NOT_FOUND: beacon block at slot 3174625",
  "stacktraces": []
}
```

This is because `block_root_at_slot_skips_none` always tries to load the block root of the _previous slot_ to the one queried, which fails when requesting the `oldest_block_slot`.

https://github.com/sigp/lighthouse/blob/999b04517e35f8fce78525132d0b8e0a4816f31e/beacon_node/beacon_chain/src/beacon_chain.rs#L1036-L1047

jimmygchen:
Closed in #7693

PR #7693, opened 2025-07-02
Fix lookups of the block at `oldest_block_slot`

michaelsproul:
## Issue Addressed

Closes:

- https://github.com/sigp/lighthouse/issues/7690

## Proposed Changes

Another checkpoint sync related fix! See issue for a description of the bug.

We fix it by just loading the block root of the `oldest_block_slot`, rather than trying to load the slot prior, which will always fail.

## Additional Info

I've added a new part to the checkpoint sync tests to exercise this. I've confirmed that commenting out the fix causes the test to fail.


dapplion (approved):

pawanjay176 (approved):
Nice catch!

PR #7695, opened 2025-07-02
Bump default gas limit

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Bump the default gas limit to 45 million based on recommendation from EL teams https://x.com/vdWijden/status/1939234101631856969 and pandas https://ethpandaops.io/posts/gaslimit-scaling/


jimmygchen (changes requested):
We may also need to change the VC `--gas-limit` default here?
https://github.com/sigp/lighthouse/blob/058dae064184b63071828024a894b2590f7ea31a/validator_client/src/cli.rs#L388-L398

pawanjay176:
Great catch, I knew I was missing it somewhere! Fixed now

jimmygchen (approved):
Looks good, thanks!

PR #7702, opened 2025-07-04
Introduce a max delay bound of 1s for the column reconstruction

### LINKED ISSUE #7697: Add a max delay bound for column reconstruction to avoid indefinite waiting on staggered gossip
There was a recent optimisation made to delay reconstruction 
https://github.com/sigp/lighthouse/pull/7588

It has shown to be quite effective at improving supernode performance, however we should consider having a maximum delay on performing reconstruction. Reconstruction is currently triggered once we have at least 64 columns, and not received any more gossip columns for 150ms:
https://github.com/sigp/lighthouse/blob/f67084a571d14efb5473bf53fe32cf0128482973/beacon_node/beacon_processor/src/scheduler/work_reprocessing_queue.rs#L59

In the case where gossip columns arrives in a staggered manner (e.g. 100ms between columns from 64 to 70), it could delay reconstruction too much, and this should be a scenario to prevent against.

PoulavBhowmick03:
## Issue Addressed

Fixes #7697

## Proposed Changes

- Introduced a const `MAX_RECONSTRUCTION_DELAY` of 1s
- A new `ReconstructionBound` struct to group together the `DelayKey` and the `first enqueue`.
- A new `if` condition to check if the duration since the first enqueue surpasses the `MAX_RECONSTRUCTION_DELAY`

### DIFF
--- a/beacon_node/beacon_processor/src/scheduler/work_reprocessing_queue.rs
+++ b/beacon_node/beacon_processor/src/scheduler/work_reprocessing_queue.rs
@@ -29,6 +29,7 @@ use std::time::Duration;
 use strum::AsRefStr;
 use task_executor::TaskExecutor;
 use tokio::sync::mpsc::{self, Receiver, Sender};
+use tokio::time::Instant;
 use tokio_util::time::delay_queue::{DelayQueue, Key as DelayKey};
 use tracing::{debug, error, trace, warn};
 use types::{EthSpec, Hash256, Slot};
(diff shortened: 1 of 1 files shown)


jimmygchen:
@PoulavBhowmick03  we've decided to go with a different approach implemented in #7976, as we think reconstruction is better triggered at the 3s so we have enough time to reconstruct before the attestation deadline.

Thanks a lot for the PR 🙏

PoulavBhowmick03:
> @PoulavBhowmick03  we've decided to go with a different approach implemented in #7976, as we think reconstruction is better triggered at the 3s so we have enough time to reconstruct before the attestation deadline.
> 
> Thanks a lot for the PR 🙏 
> 
> 

Yes I went through the PR and spoke to Eitan about it, as the slot aware approach makes more sense for the reconstruction.
Anytime, looking forward to working on more stuffs and learning through it ❤️

PR #7705, opened 2025-07-07
Refine delayed head block logging

michaelsproul:
## Proposed Changes

Small tweak to `Delayed head block` logging to make it more representative of actual issues. Previously we used the total import delay to determine whether a block was late, but this includes the time taken for IO (and now hdiff computation) which happens _after_ the block is made attestable.

This PR changes the logic to use the attestable delay (where possible) falling back to the previous value if the block doesn't have one; e.g. if it didn't meet the conditions to make it into the attestable cache.

## Additional Info

I've also applied some refactors and simplifications related to the relocation of the `late_head` variable.

I got rid of some raw `SystemTime` usage as well, which improves repeatability of test cases by removing realtime variability. In production it should be exactly equivalent.

### DIFF
--- a/beacon_node/beacon_chain/src/canonical_head.rs
+++ b/beacon_node/beacon_chain/src/canonical_head.rs
@@ -38,7 +38,7 @@ use crate::{
     block_times_cache::BlockTimesCache,
     events::ServerSentEventHandler,
     metrics,
-    validator_monitor::{get_slot_delay_ms, timestamp_now},
+    validator_monitor::get_slot_delay_ms,
     BeaconChain, BeaconChainError as Error, BeaconChainTypes, BeaconSnapshot,
 };
 use eth2::types::{EventKind, SseChainReorg, SseFinalizedCheckpoint, SseHead, SseLateHead};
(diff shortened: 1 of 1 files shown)


beacon_node/beacon_chain/src/canonical_head.rs
```diff
@@ -1450,29 +1475,4 @@ fn observe_head_block_delays<E: EthSpec, S: SlotClock>(
...
-
-    if let Some(event_handler) = event_handler {
-        if !block_from_sync && late_head && event_handler.has_late_head_subscribers() {
-            let peer_info = block_times_cache.get_peer_info(head_block_root);
-            let block_delays = block_times_cache.get_block_delays(
-                head_block_root,
-                slot_clock
-                    .start_of(head_block_slot)
-                    .unwrap_or_else(|| Duration::from_secs(0)),
-            );
```
chong-he, line 1462: Good simplification, because `block_delays` was already defined above

beacon_node/beacon_chain/src/canonical_head.rs
```diff
@@ -1432,6 +1439,24 @@ fn observe_head_block_delays<E: EthSpec, S: SlotClock>(
                 set_as_head_time_ms = format_delay(&block_delays.set_as_head),
                 "Delayed head block"
             );
+            if let Some(event_handler) = event_handler {
+                if event_handler.has_late_head_subscribers() {
```
chong-he, line 1443: Nice! Because it is already in the `if !block_from_sync{}`  and now it's put in `if late_head{}` arm, so the condition becomes simpler

chong-he (approved):
Looks great! The code is a bit more simplified, and all changes are good. 

No changes required in my opinion

michaelsproul:
Thanks for the review @chong-he!

PR #7706, opened 2025-07-07
Update LH book for v7.1.0

michaelsproul:
## Proposed Changes

Update the book for upcoming v7.1.0 release.

This is targeted at `unstable` rather than `release-v7.1.0` because the book is built from `unstable`.

### DIFF
--- a/book/src/advanced_database_migrations.md
+++ b/book/src/advanced_database_migrations.md
@@ -17,7 +17,7 @@ validator client or the slasher**.
 
 | Lighthouse version | Release date | Schema version | Downgrade available? |
 |--------------------|--------------|----------------|----------------------|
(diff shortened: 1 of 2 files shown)


book/src/advanced_database_migrations.md
```diff
@@ -17,7 +17,7 @@ validator client or the slasher**.
 
 | Lighthouse version | Release date | Schema version | Downgrade available? |
 |--------------------|--------------|----------------|----------------------|
-| v7.1.0             | TBD 2025     | v23            | yes                  |
+| v7.1.0             | Jul 2025     | v26            | yes                  |
```
michaelsproul, line 20: Hopefully we can release before the end of the month :sweat_smile:

jimmygchen (approved):
LGTM! good catch on the gas limit, I missed that

PR #7707, opened 2025-07-07
Remove unneeded spans that caused debug logs to appear when level is set to `info`

### LINKED ISSUE #7155: Debug level log appearing in stdout when log level is set to INFO

jimmygchen:
## Issue Addressed

Fixes #7155.

It turns out the issue is caused by calling a function that creates an info span (`chain.id()` here), e.g.

```rust
debug!(id = chain.id(), ?sync_type, reason = ?remove_reason, op, "Chain removed");
```

I've remove all unneeded spans, especially getter functions - there's little reasons for span and they often get used in logging. We should also revisit all the spans after the release - i think we could make them more useful than they are today.

I've let it run for a while and no longer seeing any `DEBUG` logs.

### DIFF
--- a/beacon_node/network/src/sync/range_sync/chain.rs
+++ b/beacon_node/network/src/sync/range_sync/chain.rs
@@ -153,25 +153,21 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
     }
 
     /// Check if the chain has peers from which to process batches.
-    #[instrument(parent = None,fields(chain = self.id , service = "range_sync"), skip_all)]
     pub fn available_peers(&self) -> usize {
         self.peers.len()
     }
 
     /// Get the chain's id.
-    #[instrument(parent = None, fields(chain = self.id , service = "range_sync"), skip_all)]
     pub fn id(&self) -> ChainId {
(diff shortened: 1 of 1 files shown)


chong-he:
Did a quick test on this and can confirm that we are no longer seeing the DEBUG logs in terminal

macladson (approved):
Thanks for fixing this! I think certain spans will eventually have some utility for us but since we aren't utilizing them at all I support removing them :+1:

PR #7708, opened 2025-07-07
Make `notifier_service::notify` pub

dknopik:
## Issue Addressed

Anchor wants the `notify` function to run only in certain cases - so the `spawn_notifier` function is unsuitable for us.

## Proposed Changes

Anchor uses it's own `notify` function, which then calls `notifier_service::notify` (in most circumstances). To enable that, `notify` needs to be `pub`.


michaelsproul (approved):
LGTM!

Issue #7700, opened 2025-07-04
Skip KZG proof verification for blobs and proofs returned from the EL


jimmygchen:
Lighthouse currently verifies KZG proofs returned from the EL:
- `engine_getPayloadV5` (during block production, GET `v3/validator/blocks`)
- `engine_getBlobsV2` (during attestation window for data availability checks)

It was a good to have at the time it was implemented, because the verification is quite cheap at the current blob levels, but it grows as the blob count increases and could get quite significant for supernode after PeerDAS.

Both of these verifications are redundant and could be safely skipped, because the EL already verifies them as the blobs enter the mempool, and the EL is considered a trusted component (below from Francesco):
> If the EL is malicious it could say that invalid payloads are valid, which is not something the CL can verify

Additionally we could also skip the KZG verification during block publishing (POST `v2/beacon/blocks`), ONLY for locally built blocks, as we can't assume builder blobs are valid.

jimmygchen:
Completed in #7771 and #7713

PR #7713, opened 2025-07-08
Remove KZG verification from local block production and blobs fetched from the EL

jimmygchen:
## Issue Addressed

#7700 

## Proposed Changes

As described in title, the EL already performs KZG verification on all blobs when they entered the mempool, so it's redundant to perform extra validation on blobs returned from the EL.

This PR removes
- KZG verification for both blobs and data columns during block production
- KZG verification for data columns after fetch engine blobs call. I have not done this for blobs because it requires extra changes to check the observed cache, and doesn't feel like it's a worthy optimisation given the number of blobs per block.

This PR does not remove KZG verification on the block publishing path yet.

### DIFF
--- a/beacon_node/beacon_chain/src/fetch_blobs/mod.rs
+++ b/beacon_node/beacon_chain/src/fetch_blobs/mod.rs
@@ -311,6 +311,9 @@ async fn fetch_and_process_blobs_v2<T: BeaconChainTypes>(
         return Ok(None);
     }
 
+    // Up until this point we have not observed the data columns in the gossip cache, which allows
+    // them to arrive independently while this function is running. In publish_fn we will observe
+    // them and then publish any columns that had not already been observed.
     publish_fn(EngineGetBlobsOutput::CustodyColumns(
         custody_columns_to_import.clone(),
     ));
--- a/beacon_node/beacon_chain/src/fetch_blobs/mod.rs
+++ b/beacon_node/beacon_chain/src/fetch_blobs/mod.rs
@@ -14,7 +14,7 @@ mod tests;
 
 use crate::blob_verification::{GossipBlobError, GossipVerifiedBlob};
 use crate::block_verification_types::AsBlock;
-use crate::data_column_verification::KzgVerifiedCustodyDataColumn;
+use crate::data_column_verification::{KzgVerifiedCustodyDataColumn, KzgVerifiedDataColumn};
 #[cfg_attr(test, double)]
 use crate::fetch_blobs::fetch_blobs_beacon_adapter::FetchBlobsBeaconAdapter;
 use crate::kzg_utils::blobs_to_data_column_sidecars;
@@ -358,17 +358,24 @@ async fn compute_custody_columns_to_import<T: BeaconChainTypes>(
                 // `DataAvailabilityChecker` requires a strict match on custody columns count to
                 // consider a block available.
                 let mut custody_columns = data_columns_result
-                    .map(|mut data_columns| {
-                        data_columns.retain(|col| custody_columns_indices.contains(&col.index));
+                    .map(|data_columns| {
                         data_columns
+                            .into_iter()
+                            .filter(|col| custody_columns_indices.contains(&col.index))
+                            .map(|col| {
+                                KzgVerifiedCustodyDataColumn::from_asserted_custody(
+                                    KzgVerifiedDataColumn::from_execution_verified(col),
+                                )
+                            })
+                            .collect::<Vec<_>>()
                     })
                     .map_err(FetchEngineBlobError::DataColumnSidecarError)?;
 
                 // Only consider columns that are not already observed on gossip.
-                if let Some(observed_columns) = chain_adapter_cloned.known_for_proposal(
+                if let Some(observed_columns) = chain_adapter_cloned.data_column_known_for_proposal(
                     ProposalKey::new(block.message().proposer_index(), block.slot()),
                 ) {
-                    custody_columns.retain(|col| !observed_columns.contains(&col.index));
+                    custody_columns.retain(|col| !observed_columns.contains(&col.index()));
                     if custody_columns.is_empty() {
                         return Ok(vec![]);
                     }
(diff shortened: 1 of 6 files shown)


beacon_node/beacon_chain/src/fetch_blobs/mod.rs
```diff
@@ -358,17 +358,24 @@ async fn compute_custody_columns_to_import<T: BeaconChainTypes>(
...
                 let mut custody_columns = data_columns_result
-                    .map(|mut data_columns| {
-                        data_columns.retain(|col| custody_columns_indices.contains(&col.index));
+                    .map(|data_columns| {
                         data_columns
+                            .into_iter()
+                            .filter(|col| custody_columns_indices.contains(&col.index))
+                            .map(|col| {
+                                KzgVerifiedCustodyDataColumn::from_asserted_custody(
+                                    KzgVerifiedDataColumn::from_execution_verified(col),
```
pawanjay176, line 370: We haven't removed the gossip verification in fetch_blobs_v1 above which does kzg verification. Do we want to remove that as well?

pawanjay176 (commented):
Just noting here that we don't really gain anything from kzg verifying blobs returned by the builder when publishing because we have already signed the message by the time we receive the `BlobsBundle`.

beacon_node/beacon_chain/src/fetch_blobs/mod.rs
jimmygchen: Yeah I didn't do it because i thought the benefit was minimal with the blob count, and it requires introducing code changes to check gossip cache and da checker. But maybe it's worth doing it to keep things consistent so it's less confusing when we look at it in the future.

pawanjay176: i'm good with not doing it too. Keep this peerdas focused.

jimmygchen: Thanks! I don't have a strong preference either will keep it as it is and wait for @michaelsproul's review.

beacon_node/beacon_chain/src/fetch_blobs/mod.rs
```diff
@@ -378,26 +385,13 @@ async fn compute_custody_columns_to_import<T: BeaconChainTypes>(
...
                         return Ok(vec![]);
                     }
                 }
 
-                // KZG verify data columns before publishing. This prevents blobs with invalid
-                // KZG proofs from the EL making it into the data availability checker. We do not
-                // immediately add these blobs to the observed blobs/columns cache because we want
-                // to allow blobs/columns to arrive on gossip and be accepted (and propagated) while
-                // we are waiting to publish. Just before publishing we will observe the blobs/columns
-                // and only proceed with publishing if they are not yet seen.
```
michaelsproul, line 392: Should we keep the part of this comment about not marking data columns as seen here, while we wait to publish? AFAICT it is still relevant as publishing happens here:

https://github.com/sigp/lighthouse/blob/d7ab88cf858ec90696b1dbba199af34823d20e61/beacon_node/beacon_chain/src/fetch_blobs/mod.rs#L314-L316

The `publish_fn` in practice is:

https://github.com/sigp/lighthouse/blob/cfb1f7331064b758c6786e4e1dc15507af5ff5d1/beacon_node/network/src/network_beacon_processor/mod.rs#L1014-L1020

Which does still do the observation:

https://github.com/sigp/lighthouse/blob/cfb1f7331064b758c6786e4e1dc15507af5ff5d1/beacon_node/network/src/network_beacon_processor/mod.rs#L1057

michaelsproul (approved):
LGTM on the whole!

beacon_node/beacon_chain/src/fetch_blobs/mod.rs
michaelsproul: @jimmygchen What are the changes to check the gossip cache and DA checker required in the blobs/engine_v1 case? Isn't it similar in that we still do those checks prior to publishing? Here:

https://github.com/sigp/lighthouse/blob/cfb1f7331064b758c6786e4e1dc15507af5ff5d1/beacon_node/network/src/network_beacon_processor/mod.rs#L975

But yeah I'm happy to leave the blobs case as-is. No need to complicate it if the current logic works. Or we could clean it up and make it consistent, but in a separate PR?

beacon_node/beacon_chain/src/fetch_blobs/mod.rs
jimmygchen: Ah yeah good point! I remove them from here because it felt a bit out of place as the gossip verification code is no longer here, and we're no longer returning `GossipVerifiedDataColumns`.

I kept the comment here though but I think it doesn't provide enough context:
https://github.com/sigp/lighthouse/blob/5472cb85008b256512810b2163043e4276fc4486/beacon_node/beacon_chain/src/fetch_blobs/mod.rs#L44-L52

I think we could move this to where the `publish_fn` is invoked, wdyt?

beacon_node/beacon_chain/src/fetch_blobs/mod.rs
jimmygchen: so in data columns, we've removed gossip verification and replaced it with these - before publishing, we check if they're already observed:

https://github.com/sigp/lighthouse/blob/5472cb85008b256512810b2163043e4276fc4486/beacon_node/beacon_chain/src/fetch_blobs/mod.rs#L367-L385

I think we just need to do the same for blobs.

In proposer blob/data column publishing, we currently still do full gossip verification - so these functions don't exist for blobs yet, but are pretty trivial to add.

beacon_node/beacon_chain/src/fetch_blobs/mod.rs
michaelsproul: Yeah maybe just a short comment where `publish_fn` is invoked saying something like:

> Up until this point we have not observed the data columns in the gossip cache, which allows them to arrive independently while this function is running. In `publish_fn` we will observe them and then publish any columns that had not already been observed

beacon_node/beacon_chain/src/fetch_blobs/mod.rs
michaelsproul: I see, we are trying to avoid repeating the _import_ of anything already seen on gossip (or in DA checker), whereas the check here only prevents us from _publishing_ a column already seen in the cache:

https://github.com/sigp/lighthouse/blob/cfb1f7331064b758c6786e4e1dc15507af5ff5d1/beacon_node/network/src/network_beacon_processor/mod.rs#L1057

SGTM. Happy to leave this as is.

jimmygchen: I've implemented this - I'll create a separate PR for this as it's not a tiny change.

beacon_node/beacon_chain/src/fetch_blobs/mod.rs
jimmygchen: Nice, added in https://github.com/sigp/lighthouse/pull/7713/commits/c277d5092f176fc0271b9d808ac9850349446368. Thanks!

beacon_node/beacon_chain/src/fetch_blobs/mod.rs
jimmygchen: Addressed in https://github.com/sigp/lighthouse/pull/7771

PR #7717, opened 2025-07-08
Improve data column KZG verification metric buckets

jimmygchen:
## Issue Addressed

The current data column KZG verification buckets are not giving us useful info as the upper bound is too low. And we see most of numbers above 70ms for batch verification, and we don't know how much time it really takes.

This PR improves the buckets based on the numbers we got from testing. Exponential bucket seems like a good candidate here given we're expecting to increase blob count with a similar approach (possibly 2x each fork if it goes well).

### DIFF
--- a/beacon_node/beacon_chain/src/metrics.rs
+++ b/beacon_node/beacon_chain/src/metrics.rs
@@ -1809,12 +1809,12 @@ pub static KZG_VERIFICATION_BATCH_TIMES: LazyLock<Result<Histogram>> = LazyLock:
 /// * 64 columns batch: 257 ms
 /// * 128 columns batch: 508 ms
 pub static KZG_VERIFICATION_DATA_COLUMN_SINGLE_TIMES: LazyLock<Result<Histogram>> =
-    // 5 exponential buckets between 0.002 and 0.032 seconds, with more granularity on the lower end.
+    // 7 exponential buckets between 0.002 and 0.128 seconds, with more granularity on the lower end.
     LazyLock::new(|| {
             try_create_histogram_with_buckets(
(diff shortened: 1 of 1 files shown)


beacon_node/beacon_chain/src/metrics.rs
```diff
@@ -1803,26 +1803,30 @@ pub static KZG_VERIFICATION_BATCH_TIMES: LazyLock<Result<Histogram>> = LazyLock:
...
-    LazyLock::new(|| {
-        try_create_histogram_with_buckets(
-            "beacon_kzg_verification_data_column_single_seconds",
-            "Runtime of single data column kzg verification",
-            Ok(vec![
-                0.0005, 0.001, 0.0015, 0.002, 0.003, 0.004, 0.005, 0.007, 0.01, 0.02, 0.05,
-            ]),
-        )
-    });
+    // 5 exponential buckets between 0.002 and 0.032 seconds, with more granularity on the lower end.
```
michaelsproul: Isn't 32ms quite small for the largest bucket? Average should be around 5ms I know, but maybe we could push out a bit higher in case of slow hw (maybe 128ms?)

michaelsproul (approved):
Looks good! Just had one Q

beacon_node/beacon_chain/src/metrics.rs
jimmygchen: yeah agree - my thinking was if it's above 32ms, then something is seriously wrong and the node will likely suffer, but adding 2 extra buckets sounds reasonable to me.

jimmygchen: Updated!

pawanjay176 (approved):

PR #7722, opened 2025-07-09
`fusaksa-devnet-2` clean up and fixes

jimmygchen:
## Issue Addressed

Clean ups for #7655, mostly just adding doc comments and small refactors, no functionality change.

The original PR renamed some of the `fork_*` field names to `digest_*` , e.g. `digest_epoch`, `duration_to_next_digest` etc - it reads a bit strange at first, BPO forks are technically still hard forks - I think calling them forks instead of digests are a bit easier to understand. But I'd like to hear thoughts before proceeding with making this change.

### DIFF
--- a/beacon_node/lighthouse_network/tests/common.rs
+++ b/beacon_node/lighthouse_network/tests/common.rs
@@ -28,6 +28,9 @@ pub fn spec_with_all_forks_enabled() -> ChainSpec {
     chain_spec.deneb_fork_epoch = Some(Epoch::new(4));
     chain_spec.electra_fork_epoch = Some(Epoch::new(5));
     chain_spec.fulu_fork_epoch = Some(Epoch::new(6));
+
+    // check that we have all forks covered
+    assert!(chain_spec.fork_epoch(ForkName::latest()).is_some());
     chain_spec
 }
 
--- a/beacon_node/lighthouse_network/tests/common.rs
+++ b/beacon_node/lighthouse_network/tests/common.rs
@@ -11,7 +11,7 @@ use tracing::{debug, error, info_span, Instrument};
 use tracing_subscriber::EnvFilter;
 use types::{
     ChainSpec, EnrForkId, Epoch, EthSpec, FixedBytesExtended, ForkContext, ForkName, Hash256,
-    MinimalEthSpec, Slot,
+    MinimalEthSpec,
 };
 
 type E = MinimalEthSpec;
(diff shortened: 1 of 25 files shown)


consensus/types/src/fork_context.rs
```diff
@@ -23,96 +38,229 @@ impl ForkContext {
...
-                self.spec
-                    .compute_fork_digest(self.genesis_validators_root, epoch)
-            })
-            .unwrap_or_default()
+        let current_fork_epoch = self.current_fork_epoch();
+        self.epoch_to_forks
+            .range(current_fork_epoch..)
+            .nth(1)
+            .map(|(_, fork)| fork.fork_digest)
+            .unwrap_or_else(|| self.current_fork.read().fork_digest)
```
pawanjay176: maybe return an option here and handle the None case at the caller?

consensus/types/src/fork_context.rs
```diff
@@ -23,96 +38,229 @@ impl ForkContext {
...
     }
 
-    // TODO: we *may* delete this entire object and just use the spec
+    /// Returns the context bytes/fork_digest corresponding to an epoch.
+    /// See [`ChainSpec::compute_fork_digest`]
     pub fn context_bytes(&self, epoch: Epoch) -> [u8; 4] {
-        self.spec
-            .compute_fork_digest(self.genesis_validators_root, epoch)
+        self.epoch_to_forks
+            .range(..=epoch)
```
pawanjay176, line 136: pretty cool, didn't know this function exists

consensus/types/src/fork_context.rs
```diff
@@ -1,15 +1,30 @@
 use parking_lot::RwLock;
 
 use crate::{ChainSpec, Epoch, EthSpec, ForkName, Hash256, Slot};
-use std::collections::{ HashMap, HashSet};
+use std::collections::BTreeMap;
+
+#[derive(Debug, Clone)]
+pub struct HardFork {
```
pawanjay176, line 16: This also includes BPO forks though right? 
The naming seems to imply it excludes them. We could just make this private and call it `Fork`

jimmygchen: BPOs are hard forks too (it's also in the EIP) - in my mind the usual forks are `named forks` .
I can add some more docs to say it includes both named forks and BPO forks.

pawanjay176: > I can add some more docs to say it includes both named forks and BPO forks.

sounds good.

jimmygchen:
@pawanjay176 I've addressed the comments and fixed the tests 🤞  Please re-review, thanks!

beacon_node/lighthouse_network/tests/common.rs
```diff
@@ -11,41 +11,41 @@ use tracing::{debug, error, info_span, Instrument};
...
-    chain_spec.deneb_fork_epoch = Some(deneb_fork_epoch);
-    chain_spec.electra_fork_epoch = Some(electra_fork_epoch);
-    chain_spec.fulu_fork_epoch = Some(fulu_fork_epoch);
+    chain_spec.altair_fork_epoch = Some(Epoch::new(1));
+    chain_spec.bellatrix_fork_epoch = Some(Epoch::new(2));
+    chain_spec.capella_fork_epoch = Some(Epoch::new(3));
+    chain_spec.deneb_fork_epoch = Some(Epoch::new(4));
+    chain_spec.electra_fork_epoch = Some(Epoch::new(5));
+    chain_spec.fulu_fork_epoch = Some(Epoch::new(6));
+    chain_spec
```
pawanjay176, line 34: might be good to assert that the latest fork's fork_epoch is set or we will need to remember to update this for each fork

pawanjay176: Added in [555fb61](https://github.com/sigp/lighthouse/pull/7722/commits/555fb619fbd8f5402ff83d0395f18ac43809b257)

jimmygchen: Nice! Thanks

PR #7723, opened 2025-07-09
Reintroduce `--logfile` with deprecation warning

michaelsproul:
## Proposed Changes

Reintroduce the `--logfile` flag with a deprecation warning so that it doesn't prevent nodes from starting. This is considered preferable to breaking node startups so that users fix the flag, even though it means the `--logfile` flag is completely ineffective.

The flag was initially removed in:

- https://github.com/sigp/lighthouse/pull/6339


jimmygchen (approved):
LGTM!

Issue #7698, opened 2025-07-04
Switch VC <> BN block production call to SSZ (`eth/v3/validator/blocks`)


jimmygchen:
We currently use the JSON-based `eth/v3/validator/blocks` (POST) for block production. Lighthouse supports an SSZ equivalent which can reduce serde overhead.

Update the VC to use the SSZ variant for block requests when the BN supports it.

chong-he:
I am picking this one up, thanks @jimmygchen

jimmygchen:
Closed by #7727

PR #7727, opened 2025-07-10
Use SSZ by default when calling /eth/v3/validator/blocks

chong-he:
## Issue Addressed
* #7698

### DIFF
--- a/validator_client/validator_services/src/block_service.rs
+++ b/validator_client/validator_services/src/block_service.rs
@@ -524,22 +524,50 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> BlockService<S, T> {
         proposer_index: Option<u64>,
         builder_boost_factor: Option<u64>,
     ) -> Result<UnsignedBlock<S::E>, BlockError> {
-        let (block_response, _) = beacon_node
-            .get_validator_blocks_v3::<S::E>(
+        let block_response = match beacon_node
+            .get_validator_blocks_v3_ssz::<S::E>(
                 slot,
                 randao_reveal_ref,
                 graffiti.as_ref(),
                 builder_boost_factor,
             )
             .await
-            .map_err(|e| {
-                BlockError::Recoverable(format!(
-                    "Error from beacon node when producing block: {:?}",
-                    e
-                ))
-            })?;
+        {
+            Ok((ssz_block_response, _)) => {
+                info!(slot = slot.as_u64(), "Received unsigned block in SSZ");
 
-        let (block_proposer, unsigned_block) = match block_response.data {
+                ssz_block_response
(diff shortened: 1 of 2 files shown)


validator_client/validator_services/src/block_service.rs
```diff
@@ -524,22 +524,50 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> BlockService<S, T> {
...
             .await
-            .map_err(|e| {
-                BlockError::Recoverable(format!(
-                    "Error from beacon node when producing block: {:?}",
-                    e
-                ))
-            })?;
+        {
+            Ok((ssz_block_response, _)) => {
+                info!(slot = slot.as_u64(), "Received unsigned block in SSZ");
```
chong-he: Not sure if we need to log specifically as SSZ (and in JSON), or just leave it as the original log without `in SSZ`/`in JSON`, moving this line below and just log it after the `match`. 

Logging `in SSZ/JSON` provides a bit more info but don't know if it will be helpful? Moving the log after the `match` and keeping it the same as current is simpler

macladson: I don't mind either way, but I think it is redundant because of the `warn!` explaining the fallback to JSON.

validator_client/validator_services/src/block_service.rs
```diff
@@ -524,22 +524,50 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> BlockService<S, T> {
...
+
+                ssz_block_response
+            }
+            Err(e) => {
+                warn!(
+                    slot = slot.as_u64(),
+                    error = %e,
+                    "Beacon node does not support SSZ in block production, falling back to JSON"
```
macladson, line 541: I'm wondering is there a way to disable attempting to use SSZ in cases where we know SSZ isn't supported? I don't think it matters too much as the overhead of the API instantly erroring should be quite small, but it's worth considering in cases where the BN -> VC latency is very high

validator_client/validator_services/src/block_service.rs
chong-he: Makes sense, simplify in [4a3c268](https://github.com/sigp/lighthouse/pull/7727/commits/4a3c2686116cb975fdc9ff8b0c8e507a422129ef) (i.e., use the current info log)

validator_client/validator_services/src/block_service.rs
michaelsproul: What we do with the builder is we track the builder's SSZ support and turn it off after the first request fails. I don't think that additional complexity is worth it here however.

michaelsproul (approved):
LGTM!

PR #7728, opened 2025-07-10
Add Gloas boilerplate

macladson:
## Proposed Changes

Adds the required boilerplate code for the Gloas (Glamsterdam) hard fork. This allows PRs testing Gloas-candidate features to test fork transition.

This also includes de-duplication of post-Bellatrix readiness notifiers from #6797 (credit to @dapplion)

## Additional Info

Not all tests have been updated yet.


dapplion:
@macladson can you add this de-duplications? 
- https://github.com/sigp/lighthouse/pull/6797

beacon_node/execution_layer/src/engine_api/http.rs
```diff
@@ -1137,6 +1193,7 @@ impl HttpJsonRpc {
             new_payload_v3: capabilities.contains(ENGINE_NEW_PAYLOAD_V3),
             new_payload_v4: capabilities.contains(ENGINE_NEW_PAYLOAD_V4),
             new_payload_v5: capabilities.contains(ENGINE_NEW_PAYLOAD_V5),
+            new_payload_v6: capabilities.contains(ENGINE_NEW_PAYLOAD_V6),
```
dapplion: Why add a fictional v6 before it's actually used?

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -5700,6 +5700,48 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+                    kzg_commitments,
+                    maybe_blobs_and_proofs,
+                    maybe_requests,
+                    execution_payload_value,
+                ) = block_contents
+                    .ok_or(BlockProductionError::MissingExecutionPayload)?
+                    .deconstruct();
+
+                (
+                    BeaconBlock::Gloas(BeaconBlockGloas {
```
dapplion, line 5741: What's the point of adding a new block type before we actually need it? Could we wait for the first feature in Gloas?

macladson: I suppose we _could_ wait, but we will inevitably end up with this block type anyway so I would consider what I've added as boilerplate. Plus it reduces the overhead of devs adding new features. For example, if we have multiple people working on their own feature forks they can all reuse this code rather than all having to implement it themselves.

You could similarly argue that this _entire_ PR is unnecessary since we don't have any Gloas features yet, but that misses the point imo

beacon_node/execution_layer/src/engine_api/http.rs
macladson: See my comment regarding the new block type

beacon_node/beacon_chain/src/beacon_chain.rs
michaelsproul: Yeah I'm fine with defining `BeaconBlockGloas`

beacon_node/execution_layer/src/engine_api/http.rs
jimmygchen: btw new_payload_v5 doesn't exist in Fusaka, so we may be able to use v5 for G*

macladson: Ooh I was under the impression that both `new_payload` and `get_payload` would always increment each fork. In that case @dapplion is probably right and we can exclude these in the future.

beacon_node/beacon_chain/src/beacon_chain.rs
dapplion: Makes sense, I'm convinced

michaelsproul:
Can we resolve conflicts on this and merge?

Might be nice before your `types` refactor @macladson?

macladson:
I have rebased this code onto #7909 to clear up the confusion around payload versioning. I also removed the new versions of `get` and `new_payload` and Gloas will just reuse the Fulu ones for now. These aren't too hard to add later when required. As for any missing tests, I think we can just update those as we start adding Gloas content, it's not worth holding off on getting this merged.

Edit: With #7909 merged this is now ready for review

beacon_node/beacon_chain/src/beacon_block_streamer.rs
```diff
@@ -101,6 +101,7 @@ fn reconstruct_default_header_block<E: EthSpec>(
         ForkName::Deneb => ExecutionPayloadDeneb::default().into(),
         ForkName::Electra => ExecutionPayloadElectra::default().into(),
         ForkName::Fulu => ExecutionPayloadFulu::default().into(),
+        ForkName::Gloas => ExecutionPayloadGloas::default().into(),
```
shane-moore, line 104: Note: Comment below has no impact on this PR. just something we'll have to address in epbs branches, and noting here while reviewing since it's top of mind

When loading blinded blocks from the db, we run `reconstruct_default_header_block` to get a default `ExecutionPayloadHeader`.  If we expect the db to store blinded blocks for Gloas, then this `ForkName::Gloas => ExecutionPayloadGloas::default().into()` variant will be hit for creating a default payload, but when we try `let header_from_payload = ExecutionPayloadHeader::from(payload.to_ref());` right below it, `ExecutionPayloadHeader` is no longer supported in gloas, so it will fail.

I haven't thought enough about how the db will be refactored, but my intuition is that we will only store `Full` beacon blocks post-gloas, so we should be able to move Gloas into the variant:
```
 ForkName::Base | ForkName::Altair => {
            return Err(Error::PayloadReconstruction(format!(
                "Block with fork variant {} has execution payload",
                fork
            ))
            .into())
```

beacon_node/beacon_chain/src/test_utils.rs
```diff
@@ -3244,6 +3252,25 @@ pub fn generate_rand_block_and_blobs<E: EthSpec>(
             message.body.blob_kzg_commitments = bundle.commitments.clone();
             bundle
         }
+        SignedBeaconBlock::Gloas(SignedBeaconBlockGloas {
```
shane-moore, line 3255: Note: Comment below has no impact on this PR. just something we'll have to address in epbs branches, and noting here while reviewing since it's top of mind

we removed the gloas variant of `generate_rand_block_and_blobs` in eip-7732 branch since beacon blocks no longer contain execution payloads or blobs.

The goal of this test seems to be:
1. Generate a beacon block with blob commitments
2. Generate corresponding blob sidecars
3. Link them together with proofs

This won't work in gloas, since the kzg commitments will be revealed along with the `ExecutionPayloadEnvelope` instead of the `BeaconBlock`. We should probably make a comment in the epbs branches that we'll need a new test based off this separation

beacon_node/execution_layer/src/engine_api.rs
```diff
@@ -531,6 +541,34 @@ impl<E: EthSpec> ExecutionPayloadBodyV1<E> {
                     ))
                 }
             }
+            ExecutionPayloadHeader::Gloas(header) => {
```
shane-moore, line 544: we removed this Gloas variant from
```
impl<E: EthSpec> ExecutionPayloadBodyV1<E> {
    pub fn to_payload(
        self,
        header: ExecutionPayloadHeader<E>,
    ) -> Result<ExecutionPayload<E>, String> {
```
since there's no longer an `ExecutionPayloadHeader` superstruct variant for Gloas. There are `ExecutionBid` now in beacon blocks.  We shouldn't be using this `to_payload` method anymore for gloas for things like reconstructing blinded blocks since no more blinded blocks. instead, we'll still be using `get_payload` for retrieving locally built payloads and `new_payload` for EL verification of tsx received in `ExecutionPayloadEnvelope`

beacon_node/execution_layer/src/test_utils/execution_block_generator.rs
```diff
@@ -243,19 +246,23 @@ impl<E: EthSpec> ExecutionBlockGenerator<E> {
...
-                Some(fork_time) if timestamp >= fork_time => ForkName::Electra,
-                _ => match self.cancun_time {
-                    Some(fork_time) if timestamp >= fork_time => ForkName::Deneb,
-                    _ => match self.shanghai_time {
-                        Some(fork_time) if timestamp >= fork_time => ForkName::Capella,
-                        _ => ForkName::Bellatrix,
-                    },
-                },
-            },
+        let forks = [
```
shane-moore, line 249: much cleaner now 🤙

shane-moore:
LGTM, all changes seem isolated to gloas variants, so no side effects on current functionality. ofcourse could use a few more sets of 👀

ethDreamer (approved):
LGTM!

ethDreamer:
As Shane mentioned, there are many things in this PR that don't make total sense in the context of block payload separation. So we've been addressing them as they come up in the implementation of Gloas. It doesn't make sense to address them in this PR as it's beyond the scope and will start to cause conflicts with the changes we've already made.

beacon_node/client/src/notifier.rs
```diff
@@ -406,213 +400,146 @@ async fn bellatrix_readiness_logging<T: BeaconChainTypes>(
...
-                "Ready for Electra"
-            )
+fn methods_required_for_fork(
+    fork: ForkName,
+    capabilities: EngineCapabilities,
+) -> Vec<&'static str> {
+    let mut missing_methods = vec![];
+    match fork {
+        ForkName::Base | ForkName::Altair | ForkName::Bellatrix => {
+            unreachable!()
```
michaelsproul: I feel like this could just be a no-op rather than a panic?

The thinking being, if we introduce a bug here that makes this line reachable, it fundamentally doesn't matter and we should probably keep running rather than crashing.

beacon_node/store/src/partial_beacon_state.rs
```diff
@@ -16,7 +16,7 @@ use types::*;
 ///
 /// This can be deleted once schema versions prior to V22 are no longer supported.
```
michaelsproul, line 17: We'll be able to delete this once Fulu goes live on mainnet.

i.e. we can avoid making any more Gloas-related changes to this (hopefully).

michaelsproul (approved):
LGTM, let's merge

beacon_node/client/src/notifier.rs
michaelsproul: I removed this in:

- 7348f373503074e2a760cbd733cebb4cdbed82fb

PR #7729, opened 2025-07-10
Allow /validator apis to work pre-genesis

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Lighthouse BN http endpoint would return a server error pre-genesis on the `validator/duties/attester` and `validator/prepare_beacon_proposer` because `slot_clock.now()` would return a `None` pre-genesis. 

The prysm VC depends on the endpoints pre-genesis and was having issues interoping with the lighthouse bn because of this reason.

The proposer duties endpoint explicitly handles the pre-genesis case here
https://github.com/sigp/lighthouse/blob/538067f1ff9840d44e3c2ea60581e18aba8c4143/beacon_node/http_api/src/proposer_duties.rs#L23-L28

I see no reason why we can't make the other endpoints more flexible to work pre-genesis. This PR handles the pre-genesis case on the attester and prepare_beacon_proposer endpoints as well.

Thanks for raising @james-prysm.

### DIFF
--- a/beacon_node/http_api/src/attester_duties.rs
+++ b/beacon_node/http_api/src/attester_duties.rs
@@ -16,7 +16,12 @@ pub fn attester_duties<T: BeaconChainTypes>(
     request_indices: &[u64],
     chain: &BeaconChain<T>,
 ) -> Result<ApiDuties, warp::reject::Rejection> {
-    let current_epoch = chain.epoch().map_err(warp_utils::reject::unhandled_error)?;
+    let current_epoch = chain
+        .slot_clock
+        .now_or_genesis()
+        .map(|slot| slot.epoch(T::EthSpec::slots_per_epoch()))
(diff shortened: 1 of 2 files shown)


barnabasbusa:
Can confirm, prysm vc now works with `ethpandaops/lighthouse:pawanjay176-duties-pre-genesis-552421c`
```yaml
participants:
  - el_type: nethermind
    cl_type: lighthouse
    cl_image: ethpandaops/lighthouse:pawanjay176-duties-pre-genesis-552421c
    vc_type: prysm
    count: 2
additional_services:
  - dora
  ```

barnabasbusa (approved):

macladson (approved):
This looks good to me!

michaelsproul (approved):
LGTM, nice catch

michaelsproul:
@mergify requeue

PR #7733, opened 2025-07-10
Sync peer attribution

### LINKED ISSUE #7604: Use Status V2 `earliest_available_slot` for syncing
Spec: https://github.com/ethereum/consensus-specs/pull/4374

The above spec change has been partially implemented (#7590) and Lighthouse now serves and exchanges Status V2.

The remaining work:
1. Utilise the new `earliest_available_slot` info  for peer selection during sync
2. Update the `earliest_available_slot` when (covered in #7733 and #7603)
    - CGC increases (set to the slot when the effective CGC becomes effective and all advertises columns will be available)
    - Backfill (not implemented)

pawanjay176:
## Issue Addressed

Closes #7604

## Proposed Changes

Improvements to range sync including:

1. Contain column requests only to peers that are part of the SyncingChain
2. Attribute the fault to the correct peer and downscore them if they don't return the data columns for the request
3. Improve sync performance by retrying only the failed columns from other peers instead of failing the entire batch
4. Uses the earliest_available_slot to make requests to peers that claim to have the epoch. Note: if no earliest_available_slot info is available, fallback to using previous logic i.e. assume peer has everything backfilled upto WS checkpoint/da boundary

Tested this on fusaka-devnet-2 with a full node and supernode and the recovering logic seems to works well.
Also tested this a little on mainnet.

Need to do more testing and possibly add some unit tests.

### DIFF
--- a/beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
+++ b/beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
@@ -331,9 +331,7 @@ impl<E: EthSpec> PeerDB<E> {
             .filter(move |(peer_id, info)| {
                 // The custody_subnets hashset can be populated via enr or metadata
                 let is_custody_subnet_peer = info.is_assigned_to_custody_subnet(&subnet);
-                allowed_peers.contains(peer_id)
-                    && info.is_connected()
-                    && is_custody_subnet_peer
+                allowed_peers.contains(peer_id) && info.is_connected() && is_custody_subnet_peer
             })
             .map(|(peer_id, _)| peer_id)
     }
@@ -875,7 +873,7 @@ impl<E: EthSpec> PeerDB<E> {
             ) => {
                 // Update the ENR if one exists, and compute the custody subnets
                 if let Some(enr) = enr {
-                    info.set_enr(enr, );
+                    info.set_enr(enr);
                 }
 
                 match current_state {
--- a/beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
+++ b/beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
@@ -319,8 +319,8 @@ impl<E: EthSpec> PeerDB<E> {
             .map(|(peer_id, _)| peer_id)
     }
 
-    /// Returns an iterator of all good gossipsub peers that are supposed to be custodying
-    /// the given subnet id.
+    /// Returns an iterator of all peers that are supposed to be custodying
+    /// the given subnet id that also belong to `allowed_peers`.
     pub fn good_range_sync_custody_subnet_peer<'a>(
         &'a self,
         subnet: DataColumnSubnetId,
@@ -875,7 +875,7 @@ impl<E: EthSpec> PeerDB<E> {
             ) => {
                 // Update the ENR if one exists, and compute the custody subnets
                 if let Some(enr) = enr {
-                    info.set_enr(enr);
+                    info.set_enr(enr, );
                 }
 
                 match current_state {
--- a/beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
+++ b/beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
@@ -248,6 +248,34 @@ impl<E: EthSpec> PeerDB<E> {
             .map(|(peer_id, _)| peer_id)
     }
 
+    /// Returns all the synced peers from the list of allowed peers that claim to have the block
+    /// components for the given epoch based on `status.earliest_available_slot`.
+    ///
+    /// If `earliest_available_slot` info is not available, then return peer anyway assuming it has the
+    /// required data.
+    pub fn synced_peers_for_epoch<'a>(
+        &'a self,
+        epoch: Epoch,
+        allowed_peers: &'a HashSet<PeerId>,
+    ) -> impl Iterator<Item = &'a PeerId> {
+        self.peers
+            .iter()
+            .filter(move |(peer_id, info)| {
+                allowed_peers.contains(peer_id)
+                    && info.is_connected()
+                    && match info.sync_status() {
+                        SyncStatus::Synced { info } => {
+                            info.has_slot(epoch.end_slot(E::slots_per_epoch()))
+                        }
+                        SyncStatus::Advanced { info } => {
+                            info.has_slot(epoch.end_slot(E::slots_per_epoch()))
+                        }
+                        _ => false,
+                    }
+            })
+            .map(|(peer_id, _)| peer_id)
+    }
+
     /// Gives the `peer_id` of all known connected and advanced peers.
     pub fn advanced_peers(&self) -> impl Iterator<Item = &PeerId> {
         self.peers
--- a/beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
+++ b/beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
@@ -319,6 +319,25 @@ impl<E: EthSpec> PeerDB<E> {
             .map(|(peer_id, _)| peer_id)
     }
 
+    /// Returns an iterator of all good gossipsub peers that are supposed to be custodying
+    /// the given subnet id.
+    pub fn good_range_sync_custody_subnet_peer<'a>(
+        &'a self,
+        subnet: DataColumnSubnetId,
+        allowed_peers: &'a HashSet<PeerId>,
+    ) -> impl Iterator<Item = &'a PeerId> {
+        self.peers
+            .iter()
+            .filter(move |(peer_id, info)| {
+                // The custody_subnets hashset can be populated via enr or metadata
+                let is_custody_subnet_peer = info.is_assigned_to_custody_subnet(&subnet);
+                allowed_peers.contains(peer_id)
+                    && info.is_connected()
+                    && is_custody_subnet_peer
+            })
+            .map(|(peer_id, _)| peer_id)
+    }
+
     /// Gives the ids of all known disconnected peers.
     pub fn disconnected_peers(&self) -> impl Iterator<Item = &PeerId> {
         self.peers
--- a/beacon_node/lighthouse_network/src/peer_manager/peerdb/sync_status.rs
+++ b/beacon_node/lighthouse_network/src/peer_manager/peerdb/sync_status.rs
@@ -28,6 +28,19 @@ pub struct SyncInfo {
     pub earliest_available_slot: Option<Slot>,
 }
 
+impl SyncInfo {
+    /// Returns true if the provided slot is greater than `earliest_available_slot`.
+    ///
+    /// If `earliest_available_slot` does is None, then we just assume that the peer has the slot.
+    pub fn has_slot(&self, slot: Slot) -> bool {
+        if let Some(earliest_available_slot) = self.earliest_available_slot {
+            slot >= earliest_available_slot
+        } else {
+            true
+        }
+    }
+}
+
 impl std::cmp::PartialEq for SyncStatus {
     fn eq(&self, other: &Self) -> bool {
         matches!(
--- a/beacon_node/lighthouse_network/src/peer_manager/peerdb/sync_status.rs
+++ b/beacon_node/lighthouse_network/src/peer_manager/peerdb/sync_status.rs
@@ -29,9 +29,9 @@ pub struct SyncInfo {
 }
 
 impl SyncInfo {
-    /// Returns true if the provided slot is greater than `earliest_available_slot`.
+    /// Returns true if the provided slot is greater than or equal to the peer's `earliest_available_slot`.
     ///
-    /// If `earliest_available_slot` does is None, then we just assume that the peer has the slot.
+    /// If `earliest_available_slot` is None, then we just assume that the peer has the slot.
     pub fn has_slot(&self, slot: Slot) -> bool {
         if let Some(earliest_available_slot) = self.earliest_available_slot {
             slot >= earliest_available_slot
--- a/beacon_node/lighthouse_network/src/service/api_types.rs
+++ b/beacon_node/lighthouse_network/src/service/api_types.rs
@@ -1,4 +1,5 @@
 use crate::rpc::methods::{ResponseTermination, RpcResponse, RpcSuccessResponse, StatusMessage};
+use libp2p::PeerId;
 use std::fmt::{Display, Formatter};
 use std::sync::Arc;
 use types::{
@@ -61,6 +62,11 @@ pub struct DataColumnsByRangeRequestId {
     pub id: Id,
     /// The Id of the overall By Range request for block components.
     pub parent_request_id: ComponentsByRangeRequestId,
+    /// The peer id associated with the request.
+    ///
+    /// This is useful to penalize the peer at a later point if it returned data columns that
+    /// did not match with the verified block.
+    pub peer: PeerId,
 }
 
 /// Block components by range request for range sync. Includes an ID for downstream consumers to
@@ -306,6 +312,7 @@ mod tests {
                     batch_id: Epoch::new(0),
                 },
             },
+            peer: PeerId::random(),
         };
         assert_eq!(format!("{id}"), "123/122/RangeSync/0/54");
     }
--- a/beacon_node/network/src/sync/block_sidecar_coupling.rs
+++ b/beacon_node/network/src/sync/block_sidecar_coupling.rs
@@ -143,29 +143,6 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
         }
     }
 
-    // pub fn remove_custody_columns(
-    //     &mut self,
-    //     req_id: DataColumnsByRangeRequestId,
-    // ) -> Result<(), String> {
-    //     match &mut self.block_data_request {
-    //         RangeBlockDataRequest::NoData => {
-    //             Err("received data columns but expected no data".to_owned())
-    //         }
-    //         RangeBlockDataRequest::Blobs(_) => {
-    //             Err("received data columns but expected blobs".to_owned())
-    //         }
-    //         RangeBlockDataRequest::DataColumns {
-    //             ref mut requests, ..
-    //         } => {
-    //             let req = requests
-    //                 .get_mut(&req_id)
-    //                 .ok_or(format!("unknown data columns by range req_id {req_id}"))?;
-    //             req.remove_invalid_entries(req_id);
-    //             Ok(())
-    //         }
-    //     }
-    // }
-
     pub fn responses(
         &mut self,
         spec: &ChainSpec,
--- a/beacon_node/network/src/sync/block_sidecar_coupling.rs
+++ b/beacon_node/network/src/sync/block_sidecar_coupling.rs
@@ -33,7 +33,7 @@ enum RangeBlockDataRequest<E: EthSpec> {
             ByRangeRequest<DataColumnsByRangeRequestId, DataColumnSidecarList<E>>,
         >,
         /// The column indices corresponding to the request
-        column_peers: HashMap<DataColumnsByRangeRequestId, Vec<u64>>,
+        column_peers: HashMap<DataColumnsByRangeRequestId, Vec<ColumnIndex>>,
         expected_custody_columns: Vec<ColumnIndex>,
     },
 }
@@ -41,7 +41,7 @@ enum RangeBlockDataRequest<E: EthSpec> {
 #[derive(Debug)]
 pub struct CouplingError {
     pub(crate) msg: String,
-    pub(crate) column_and_peer: Option<(Vec<(u64, PeerId)>, PeerAction)>,
+    pub(crate) column_and_peer: Option<(Vec<(ColumnIndex, PeerId)>, PeerAction)>,
 }
 
 impl<E: EthSpec> RangeBlockComponentsRequest<E> {
@@ -50,7 +50,7 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
         blocks_req_id: BlocksByRangeRequestId,
         blobs_req_id: Option<BlobsByRangeRequestId>,
         data_columns: Option<(
-            Vec<(DataColumnsByRangeRequestId, Vec<u64>)>,
+            Vec<(DataColumnsByRangeRequestId, Vec<ColumnIndex>)>,
             Vec<ColumnIndex>,
         )>,
     ) -> Self {
@@ -609,7 +609,7 @@ mod tests {
         let columns_req_id = batched_column_requests
             .iter()
             .enumerate()
-            .map(|(i, column)| (columns_id(i as Id, components_id), column.clone()))
+            .map(|(i, columns)| (columns_id(i as Id, components_id), columns.clone()))
             .collect::<Vec<_>>();
 
         let mut info = RangeBlockComponentsRequest::<E>::new(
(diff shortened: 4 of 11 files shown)


beacon_node/lighthouse_network/src/peer_manager/peerdb/sync_status.rs
```diff
@@ -28,6 +28,19 @@ pub struct SyncInfo {
     pub earliest_available_slot: Option<Slot>,
 }
 
+impl SyncInfo {
+    /// Returns true if the provided slot is greater than `earliest_available_slot`.
```
jimmygchen:
```suggestion
    /// Returns true if the provided slot is greater than or equal to the peer's `earliest_available_slot`.
```

beacon_node/lighthouse_network/src/peer_manager/peerdb/sync_status.rs
```diff
@@ -28,6 +28,19 @@ pub struct SyncInfo {
...
+    ///
+    /// If `earliest_available_slot` does is None, then we just assume that the peer has the slot.
```
jimmygchen:
```suggestion
    /// If `earliest_available_slot` is None, then we just assume that the peer has the slot.
```

jimmygchen: This can only be `None` if the peer does not support StatusV2 right?
I guess we haven't formally deprecated V1 yet, so this behaviour is probably ok?

beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
```diff
@@ -828,7 +875,7 @@ impl<E: EthSpec> PeerDB<E> {
             ) => {
                 // Update the ENR if one exists, and compute the custody subnets
                 if let Some(enr) = enr {
-                    info.set_enr(enr);
+                    info.set_enr(enr, );
```
jimmygchen:
```suggestion
                    info.set_enr(enr);
```

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -891,7 +920,7 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
                 .network_globals()
                 .peers
                 .read()
-                .synced_peers()
+                .synced_peers_for_epoch(batch_id, &self.peers)
```
jimmygchen, line 921: The existing code already assumes `batch_id` == `epoch`, but i think we might want to start moving away from it, both for clarity and safety in case we want to change the batch size in the future. What about we get the epoch from the batch `start_slot`?

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -30,24 +32,38 @@ enum RangeBlockDataRequest<E: EthSpec> {
...
+        /// The column indices corresponding to the request
+        column_peers: HashMap<DataColumnsByRangeRequestId, Vec<u64>>,
         expected_custody_columns: Vec<ColumnIndex>,
     },
 }
 
+#[derive(Debug)]
+pub struct CouplingError {
+    pub(crate) msg: String,
+    pub(crate) column_and_peer: Option<(Vec<(u64, PeerId)>, PeerAction)>,
```
jimmygchen:
```suggestion
    pub(crate) column_and_peer: Option<(Vec<(ColumnIndex, PeerId)>, PeerAction)>,
```

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -30,24 +32,38 @@ enum RangeBlockDataRequest<E: EthSpec> {
...
+}
+
 impl<E: EthSpec> RangeBlockComponentsRequest<E> {
+    #[allow(clippy::type_complexity)]
     pub fn new(
         blocks_req_id: BlocksByRangeRequestId,
         blobs_req_id: Option<BlobsByRangeRequestId>,
-        data_columns: Option<(Vec<DataColumnsByRangeRequestId>, Vec<ColumnIndex>)>,
+        data_columns: Option<(
+            Vec<(DataColumnsByRangeRequestId, Vec<u64>)>,
```
jimmygchen: nit - for readability:

```suggestion
            Vec<(DataColumnsByRangeRequestId, Vec<ColumnIndex>)>,
```

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -30,24 +32,38 @@ enum RangeBlockDataRequest<E: EthSpec> {
             DataColumnsByRangeRequestId,
             ByRangeRequest<DataColumnsByRangeRequestId, DataColumnSidecarList<E>>,
         >,
+        /// The column indices corresponding to the request
+        column_peers: HashMap<DataColumnsByRangeRequestId, Vec<u64>>,
```
jimmygchen:
```suggestion
        column_peers: HashMap<DataColumnsByRangeRequestId, Vec<ColumnIndex>>,
```

beacon_node/network/src/sync/block_sidecar_coupling.rs
jimmygchen: Just a note here that we don't track the block root that had missing columns, so we just request the column for the entire range - might be something to optimise for when we get to higher blob counts.


for 48 blobs, each column is 96kb, so in the worst case if we're missing a single column for a block, we'd end up downloading the entire batch (96kb * 32 = 3072kb).


However, this is also an edge case - in normal cases, peer usually not return the columns or returns everything. The current logic handles this case fine, just to note a possible future optimisation. I think some clients are already using byRoots to retrieve missing columns, but in the current structure this feels pretty complicated to do (maybe something for tree sync)

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -485,7 +609,7 @@ mod tests {
         let columns_req_id = batched_column_requests
             .iter()
             .enumerate()
-            .map(|(i, _)| columns_id(i as Id, components_id))
+            .map(|(i, column)| (columns_id(i as Id, components_id), column.clone()))
```
jimmygchen: `columns` or `column_indices`

jimmygchen:
I wonder if we can now remove this check given we now have column retries?
https://github.com/sigp/lighthouse/blob/3182646d425dc497c024b636c8fdc870c9c65d2f/beacon_node/network/src/sync/range_sync/chain.rs#L1149-L1156

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -891,7 +920,7 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
                 .network_globals()
```
jimmygchen, line 918: The `TODO(das)` comment above can be removed i think?

beacon_node/network/src/sync/range_sync/chain.rs
jimmygchen: This may end up being a scope creep - there are actually assumptions on this in a few places, happy to leave it for a separate PR.

beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
jimmygchen: Fixed in https://github.com/sigp/lighthouse/pull/7733/commits/aa635897b53b7ffe0c7998ce8de2436d2ca825ef

beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
```diff
@@ -291,6 +319,23 @@ impl<E: EthSpec> PeerDB<E> {
...
+    /// Returns an iterator of all peers that are supposed to be custodying
+    /// the given subnet id that also belong to `allowed_peers`.
+    pub fn good_range_sync_custody_subnet_peer<'a>(
+        &'a self,
+        subnet: DataColumnSubnetId,
+        allowed_peers: &'a HashSet<PeerId>,
+    ) -> impl Iterator<Item = &'a PeerId> {
+        self.peers
+            .iter()
+            .filter(move |(peer_id, info)| {
```
jimmygchen, line 331: we no longer use the return value here, maybe we could optimise this function with `any` instead of `filter`, and return `bool` instead?

jimmygchen (changes requested):
Nice, the fixees look simple enough and seems to work really well on devnet.
My comments are mostly minor - let me know what you think.

beacon_node/lighthouse_network/src/peer_manager/peerdb/sync_status.rs
pawanjay176: This is essentially the behaviour we need to exhibit pre v2. 
Basically, we are saying "tell me what your earliest available slot is or i'll assume its backfilled all the way its supposed to be"

beacon_node/network/src/sync/block_sidecar_coupling.rs
pawanjay176: I think this would be unnecessarily complex to implement and wouldn't give us enough benefit to justify the complexity.

beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
pawanjay176: I didn't get this. You mean something like
```rust
fn is_good_range_sync_custody_subnet_peer(&self, peer_id: PeerId, subnet: DataColumnSubnetId) -> bool
```
and pass all peers from the peer list and just filter them at the caller?

beacon_node/network/src/sync/block_sidecar_coupling.rs
jimmygchen: Agreed

beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
jimmygchen: This function is only used here to determine if there's any peers on a particular custody subnet, we don't really use the return value other than checking the count > 0 - so perhaps we could change this function to `has_good_range_sync_peer_on_custody_subnet` and short-circuit it (using `any`) without iterating through all peers (128 times for superdnoes)

https://github.com/sigp/lighthouse/blob/3182646d425dc497c024b636c8fdc870c9c65d2f/beacon_node/network/src/sync/range_sync/chain.rs#L1098-L1110

jimmygchen: (Also this is likely a micro optimisation I wouldn’t be too worried about for this PR)

jimmygchen (approved):
Looking really good and able to sync on devnet 👍 
Happy to address backfill issues in a separate PR.

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -951,6 +978,50 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
...
+
+            let synced_peers = network
+                .network_globals()
+                .peers
+                .read()
+                .synced_peers()
+                .cloned()
+                .collect::<HashSet<_>>();
+
+            match network.retry_columns_by_range(
```
dapplion, line 1003: This retry logic only applies to forward sync and not backfill sync?

beacon_node/lighthouse_network/src/service/api_types.rs
```diff
@@ -61,6 +62,11 @@ pub struct DataColumnsByRangeRequestId {
     pub id: Id,
     /// The Id of the overall By Range request for block components.
     pub parent_request_id: ComponentsByRangeRequestId,
+    /// The peer id associated with the request.
+    ///
+    /// This is useful to penalize the peer at a later point if it returned data columns that
+    /// did not match with the verified block.
+    pub peer: PeerId,
```
dapplion, line 69: Adding the peer here is more for convenience of not having to pass another variable that maps DataColumnsByRangeRequestId -> PeerId?

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -235,30 +321,61 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
                     // TODO(das): on the initial version of PeerDAS the beacon chain does not check
                     // rpc custody requirements and dropping this check can allow the block to have
                     // an inconsistent DB.
-                    return Err(format!("No columns for block {block_root:?} with data"));
+
+                    // For now, we always assume that the block peer is right.
+                    // This is potentially dangerous as we can get isolated on a chain with a
+                    // malicious block peer.
+                    // TODO: fix this by checking the proposer signature before downloading columns.
```
dapplion, line 328: Even if we check the signature it doesn't defend against malicious proposers. The long term solution is tree-sync

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -235,30 +321,61 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
...
+                    let responsible_peers = column_to_peer.iter().map(|c| (*c.0, *c.1)).collect();
+                    return Err(CouplingError {
+                        msg: format!("No columns for block {block_root:?} with data"),
+                        column_and_peer: Some((responsible_peers, PeerAction::LowToleranceError)),
```
dapplion, line 332: LowToleranceError is very harsh we can disconnect everyone very quickly in case of bugs

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -235,30 +321,61 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
...
+                                column_and_peer: None,
+                            });
+                        };
+                        naughty_peers.push((*index, *responsible_peer));
+                    }
+                }
+                if !naughty_peers.is_empty() {
+                    return Err(CouplingError {
+                        msg: format!("Peers did not return column for block_root {block_root:?} {naughty_peers:?}"),
+                        column_and_peer: Some((naughty_peers, PeerAction::LowToleranceError)),
```
dapplion, line 359: Same as above

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -30,24 +32,38 @@ enum RangeBlockDataRequest<E: EthSpec> {
...
+        column_peers: HashMap<DataColumnsByRangeRequestId, Vec<ColumnIndex>>,
         expected_custody_columns: Vec<ColumnIndex>,
     },
 }
 
+#[derive(Debug)]
+pub struct CouplingError {
+    pub(crate) msg: String,
+    pub(crate) column_and_peer: Option<(Vec<(ColumnIndex, PeerId)>, PeerAction)>,
+}
```
dapplion, line 45: You could make this an enum

```rs
enum CouplingError {
    InternalError(String),
    CustodyFailure(String, Vec<(u64, PeerId)>, PeerAction),
}
```

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -235,30 +321,61 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
...
                     let remaining_indices = data_columns_by_index.keys().collect::<Vec<_>>();
-                    return Err(format!(
-                        "Not all columns consumed for block {block_root:?}: {remaining_indices:?}"
-                    ));
+                    // log the error but don't return an error, we can still progress with extra columns.
+                    tracing::error!(
+                        ?block_root,
+                        ?remaining_indices,
+                        "Not all columns consumed for block"
+                    );
```
dapplion, line 371: So this error can be triggered by bad peer behaviour, and there's nothing the user can do. It's not an internal error, I won't log anything or at most a `debug!` log

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -268,7 +385,9 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
         // Assert that there are no columns left for other blocks
         if !data_columns_by_block.is_empty() {
             let remaining_roots = data_columns_by_block.keys().collect::<Vec<_>>();
-            return Err(format!("Not all columns consumed: {remaining_roots:?}"));
+            // log the error but don't return an error, we can still progress with responses.
+            // this is most likely an internal error with overrequesting or a client bug.
+            tracing::error!(?remaining_roots, "Not all columns consumed for block");
```
dapplion, line 390: Same as above

beacon_node/network/src/sync/network_context.rs
```diff
@@ -441,6 +442,79 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+        }) else {
+            return Err("request id not present".to_string());
+        };
+
+        let active_request_count_by_peer = self.active_request_count_by_peer();
+
+        debug!(
+            ?failed_columns,
+            "Retrying only failed column requests from other peers"
+        );
```
dapplion, line 472: You should log the components_by_range ID here

beacon_node/network/src/sync/network_context.rs
```diff
@@ -441,6 +442,79 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
         active_request_count_by_peer
     }
 
+    /// Retries only the specified failed columns by requesting them again.
+    ///
+    /// Note: This function doesn't retry the whole batch, but retries specific requests within
+    /// the batch.
+    pub fn retry_columns_by_range(
+        &mut self,
+        request_id: Id,
```
dapplion, line 451: I tried never to pass just `Id` around to prevent mismatches and instead use the type `*RequestId` types. Since there are so many `Id` floating around a mismatch will cause this function to do nothing.

beacon_node/network/src/sync/network_context.rs
```diff
@@ -641,8 +726,12 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
             return Some(Err(e));
         }
 
-        if let Some(blocks_result) = entry.get().responses(&self.chain.spec) {
-            entry.remove();
+        if let Some(blocks_result) = entry.get_mut().responses(&self.chain.spec) {
+            if blocks_result.is_ok() {
+                // remove the entry only if it coupled successfully with
+                // no errors
+                entry.remove();
```
dapplion, line 733: Is there a retry counter somewhere? A peer should not be able cause a request to hang here forever in an infinite retry loop

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -826,11 +827,37 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
...
+            {
+                debug!(?batch_id, msg, "Block components coupling error");
+                // Note: we don't fail the batch here because a `CouplingError` is
+                // recoverable by requesting from other honest peers.
+                if let Some((column_and_peer, action)) = column_and_peer {
+                    let mut failed_columns = HashSet::new();
+                    let mut failed_peers = HashSet::new();
+                    for (column, peer) in column_and_peer {
+                        failed_columns.insert(*column);
+                        failed_peers.insert(*peer);
```
dapplion, line 843: Are these failed peers persisted between retries? Or they only the failed peers of retry N apply for the retry N+1?

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -891,7 +918,7 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
                 .network_globals()
                 .peers
                 .read()
-                .synced_peers()
+                .synced_peers_for_epoch(batch_id, &self.peers)
```
dapplion, line 921: Instead, you can iterate `self.peers` here and filter by peers that have epoch `batch_id`. All peers in the batch are synced connected peers

dapplion (commented):
Sorry I now realized I had a pending review in the other PR you closed. The main concern I have is the lack of a retry counter, where you can retry to infinity

beacon_node/network/src/sync/block_sidecar_coupling.rs
jimmygchen: Yeah i think this might be the case we're seeing on devnet-2.

jimmygchen:
> Sorry I now realized I had a pending review in the other PR you closed. The main concern I have is the lack of a retry counter, where you can retry to infinity

Thanks for the review @dapplion I've created an issue to keep track of this so we don't lose them #7744.

beacon_node/network/src/sync/range_sync/chain.rs
pawanjay176: Yeah i started working on it but it got merged before i could complete it lol. Will open another PR.

beacon_node/lighthouse_network/src/service/api_types.rs
pawanjay176: Pretty much yeah.

beacon_node/network/src/sync/block_sidecar_coupling.rs
pawanjay176: True, but its still good to have protection against malicious peers at least.

beacon_node/network/src/sync/block_sidecar_coupling.rs
pawanjay176: Why though? Not sending an error if you don't have something is malicious behaviour imo. Right now, its a bug on the majority of the clients that need to be fixed.

beacon_node/network/src/sync/block_sidecar_coupling.rs
pawanjay176: I like this, will do it in the backfill PR

beacon_node/network/src/sync/block_sidecar_coupling.rs
pawanjay176: Good call, I'll downgrade.

beacon_node/network/src/sync/range_sync/chain.rs
pawanjay176: Ohh yeah good point, i'll make the changes.

beacon_node/network/src/sync/network_context.rs
pawanjay176: Okay I'll look into this. Tbh, I am pretty confused by all the different id types, but this is helpful feedback

beacon_node/network/src/sync/network_context.rs
pawanjay176: In this version, I am assuming that there are some good peers that will eventually move the chain forward or all peers will get kicked and the chain will get removed.
Can add a retry counter in another PR.

beacon_node/network/src/sync/range_sync/chain.rs
pawanjay176: No they aren't. I tried to keep it simple here, but will try to add more retrying logic that holds some state in a subsequent PR.

beacon_node/network/src/sync/block_sidecar_coupling.rs
dapplion: Unless you want to check BLS signatures on the sync thread that would require you to send a beacon processor Work job. I suspect it will be a lot of code and complexity for little gain

beacon_node/network/src/sync/network_context.rs
dapplion: Not all errors result in downscore, so it possible to have infinite loops for chains with a single peer

PR #7734, opened 2025-07-11
Fix: RPC test failures

### LINKED ISSUE #7735: RPC tests fail due to concurrent tracing subscriber initialization
### Problem
When running the RPC tests using `cargo test --test rpc_tests` or `make test-full`, 11 out of 12 tests fail with a `SetGlobalDefaultError` indicating that "a global default trace dispatcher has already been set". This occurs because multiple tests running in parallel are attempting to initialize the global tracing subscriber.

### Expected Behavior
All RPC tests should pass successfully when run with `cargo test` or `make test-full`.

### Actual Behavior
Only 1 test passes (`test_tcp_blocks_by_range_chunked_rpc_terminates_correctly`), while 11 tests fail with the same error at `beacon_node/lighthouse_network/tests/common.rs:80:14`.

### Error Message
```
thread 'test_name' panicked at beacon_node/lighthouse_network/tests/common.rs:80:14:
called `Result::unwrap()` on an `Err` value: SetGlobalDefaultError("a global default trace dispatcher has already been set")
```

### Affected Tests

All 12 `rpc_tests`. On every run, 11/12 fail. The one that passes is not always the same, because this is a race condition.

### Steps to Reproduce
1. Clone the lighthouse repository
2. Set up the development environment as per the documentation
3. Run `cargo test --test rpc_tests` or `make test-full`
4. Observe the test failures

### Environment
- Branch: `stable` and `unstable`
- Rust version: `rustc 1.88.0 (6b00bc388 2025-06-23)`

### Root Cause
The issue stems from each test trying to initialize its own tracing subscriber using `tracing_subscriber::fmt::init()` or similar, but since the tracing subscriber is a global resource, only the first test to run succeeds in setting it up. Subsequent tests fail when they attempt to initialize it again.

I have implemented a fix for this issue and will be submitting a PR shortly.

danielrachi1:
## Issue Addressed

Fixes #7735 

## Proposed Changes

Use `tracing::subscriber::set_default` to ensure that each test/thread has its own subscirber.

### DIFF
--- a/beacon_node/lighthouse_network/tests/common.rs
+++ b/beacon_node/lighthouse_network/tests/common.rs
@@ -72,12 +72,18 @@ impl std::ops::DerefMut for Libp2pInstance {
 }
 
 #[allow(unused)]
-pub fn build_tracing_subscriber(level: &str, enabled: bool) {
+pub fn build_tracing_subscriber(
+    level: &str,
+    enabled: bool,
+) -> Option<tracing::subscriber::DefaultGuard> {
     if enabled {
-        tracing_subscriber::fmt()
-            .with_env_filter(EnvFilter::try_new(level).unwrap())
-            .try_init()
-            .unwrap();
+        Some(tracing::subscriber::set_default(
+            tracing_subscriber::fmt()
+                .with_env_filter(EnvFilter::try_new(level).unwrap())
+                .finish(),
+        ))
+    } else {
+        None
     }
 }
 
--- a/beacon_node/lighthouse_network/tests/rpc_tests.rs
+++ b/beacon_node/lighthouse_network/tests/rpc_tests.rs
@@ -57,7 +57,7 @@ fn test_tcp_status_rpc() {
     // Set up the logging.
     let log_level = "debug";
     let enable_logging = true;
-    build_tracing_subscriber(log_level, enable_logging);
+    let _subscriber = build_tracing_subscriber(log_level, enable_logging);
 
     let rt = Arc::new(Runtime::new().unwrap());
 
@@ -163,7 +163,7 @@ fn test_tcp_blocks_by_range_chunked_rpc() {
     // Set up the logging.
     let log_level = "debug";
     let enable_logging = true;
-    build_tracing_subscriber(log_level, enable_logging);
+    let _subscriber = build_tracing_subscriber(log_level, enable_logging);
 
     let messages_to_send = 6;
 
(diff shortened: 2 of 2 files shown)


beacon_node/lighthouse_network/tests/common.rs
```diff
@@ -74,10 +77,11 @@ impl std::ops::DerefMut for Libp2pInstance {
...
     if enabled {
-        tracing_subscriber::fmt()
-            .with_env_filter(EnvFilter::try_new(level).unwrap())
-            .try_init()
-            .unwrap();
+        INIT.call_once(|| {
+            tracing_subscriber::fmt()
+                .with_env_filter(EnvFilter::try_new(level).unwrap())
+                .init();
+        });
```
jimmygchen: Thanks for the PR!

I noticed a little limitation with this approach though - once tracing is initialised globally (via `init`)  with a specific log level (e.g., "info"), any subsequent tests that try to use a different level (e.g., "debug") won't see their preferred level applied, and it would not be obvious why debug logs aren't showing.

I think we could try to create test-local subscribers using the `tracing::subscriber::set_default` method or something similar, which returns a guard that unregisters the subscriber when dropped.

beacon_node/lighthouse_network/tests/common.rs
```diff
@@ -74,10 +74,11 @@ impl std::ops::DerefMut for Libp2pInstance {
 #[allow(unused)]
 pub fn build_tracing_subscriber(level: &str, enabled: bool) {
     if enabled {
-        tracing_subscriber::fmt()
-            .with_env_filter(EnvFilter::try_new(level).unwrap())
-            .try_init()
-            .unwrap();
+        tracing::subscriber::set_default(
```
jimmygchen: I _think_ the subscriber is immediately dropped because the return value is not used here. Have you tested this locally?

danielrachi1: Ah, my bad. I checked that the tests were passing, but didn't realize that no logs were showing. :man_facepalming:

jimmygchen (approved):
Looks good now! thanks @danielrachi1 🙏

jimmygchen:
@danielrachi1 looks like there's a clippy error, would you mind taking a look?
https://github.com/sigp/lighthouse/actions/runs/16768725687/job/47479016811?pr=7734

danielrachi1:
I forgot to run `cargo fmt` :man_facepalming: It should be ready now @jimmygchen

jimmygchen (approved):
LGTM thanks!

PR #7738, opened 2025-07-12
perf: optimize regex compilation in validator keystore detection

reallesee:
## Proposed Changes

- Replace repeated regex compilation in `is_voting_keystore` function with static LazyLock instances
- Add LazyLock import from std::sync
- Maintain identical functionality while improving performance

## Additional Info

This optimization is particularly important for validators managing many keystores, as the function is called repeatedly when scanning validator directories. Regex compilation is an expensive operation, and this change ensures it happens only once (on first use) rather than on every function call.

The change follows Rust best practices for handling expensive operations that can be cached and reused.

### DIFF
--- a/common/account_utils/src/validator_definitions.rs
+++ b/common/account_utils/src/validator_definitions.rs
@@ -16,6 +16,9 @@ use types::{graffiti::GraffitiString, Address, PublicKey};
 use validator_dir::VOTING_KEYSTORE_FILE;
 use zeroize::Zeroizing;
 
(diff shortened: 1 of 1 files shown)


michaelsproul:
I'm sorry but we can't accept your PR because it seems like a clear attempt at AI-driven airdrop farming. The evidence of this is:

- Private Github profile
- PR description written in an AI style
- PR targets the wrong branch (you/your AI did not read the contributor guide)

PR #7746, opened 2025-07-16
Add jemalloc profiling via HTTP API

jimmygchen:
## Issue Addressed

Add jemalloc profiling via HTTP API.

This is included at build time if `jemalloc` is enabled, however not activated by default. User will need to enable profiling via the curl command below.

## Commands

```
# Enables profiling
curl -X POST http://localhost:4000/lighthouse/malloc/prof_active -H "Content-Type: application/json" -d "true"

# Produce a dump
curl -X POST http://localhost:4000/lighthouse/malloc/prof_dump \
     -H "Content-Type: application/json" \
     -d '"/home/ubuntu/prof.dump"'

```

## Additional notes

- [x] DOCUMENTED: debug symbol needs to be enabled when building the binary
- [x] needs to confirm it works on different targets
  - [x] so far this works on ubuntu and macOS
- [x] needs to confirm profiling is inactive on startup - confirmed the memory growth was due to a sync bug (#7767)
- [ ] ~~might need a way to purge samples in memory~~ - I don't think this is strictly necessary at this point. We can add it if we need it later.

### DIFF
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -63,6 +63,7 @@
 * [FAQs](./faq.md)
 * [Protocol Developers](./developers.md)
   * [Lighthouse Architecture](./developers_architecture.md)
+  * [Memory Profiling](./memory_profiling.md)
 * [Security Researchers](./security.md)
 * [Archived](./archived.md)
   * [Merge Migration](./archived_merge_migration.md)
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -63,7 +63,7 @@
 * [FAQs](./faq.md)
 * [Protocol Developers](./developers.md)
   * [Lighthouse Architecture](./developers_architecture.md)
-  * [Memory Profiling](./memory_profiling.md)
+  * [Memory Profiling](./developers_memory_profiling.md)
 * [Security Researchers](./security.md)
 * [Archived](./archived.md)
   * [Merge Migration](./archived_merge_migration.md)
--- a/book/src/memory_profiling.md
+++ b/book/src/memory_profiling.md
@@ -1,0 +1,51 @@
+# Memory Profiling in Lighthouse
+
+Lighthouse ships with jemalloc enabled by default on Linux, with heap profiling (`prof:true,prof_active:false`) already configured. This guide explains how to capture and inspect heap profiles using `jeprof` to help diagnose memory issues. Use this profiling setup to catch leaks, regressions, or bloated allocation paths.
+
+## 1. Build Lighthouse with Debug Symbols
+
+To make the profiling data readable, build Lighthouse with debug symbols:
+
+```bash
+RUSTFLAGS="-C debuginfo=2" make
+```
+
+This ensures the installed `lighthouse` binary includes symbol information. `debug = true` in Cargo profiles is more expansive (enabling `opt-level = 0`, e.g.), but `debuginfo=2` is sufficient and better suited for profiling with optimized binaries.
+
+## 2. Run the Beacon Node
+
+Run the node as usual:
+
+```bash
+lighthouse bn ...
+```
+
+Let it run for a while to accumulate allocations if desired. Note that jemalloc only records allocations after profiling is activated - consider this when deciding when to start profiling.
+
+> **Be consistent:** When analyzing a profile dump, `jeprof` must be given the exact path to the binary used to launch the process. In this setup, it's simply `$(which lighthouse)`.
+
+## 3. Start Profiling and Dump Memory
+
+Enable jemalloc profiling:
+
+```bash
+curl -X POST http://localhost:5052/lighthouse/malloc/prof_active -H "Content-Type: application/json" -d "true"
+```
+
+Trigger a memory profile dump:
+
+```bash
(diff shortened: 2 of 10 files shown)


common/malloc_utils/src/jemalloc.rs
```diff
@@ -124,6 +125,22 @@ pub fn page_size() -> Result<usize, Error> {
...
+#[allow(dead_code)]
+pub fn prof_dump(filename: &str) -> Result<(), String> {
+    // Add a C-style `0x00` terminator to the filename and convert it to a
+    // `CStr`.
+    let mut terminated_filename = vec![0x00_u8; filename.len() + 1];
+    terminated_filename[..filename.len()].copy_from_slice(filename.as_ref());
+    let cstr = CStr::from_bytes_with_nul(&terminated_filename)
+        .map_err(|e| format!("Cannot convert filename to CStr: {e:?}"))?;
+
+    unsafe { raw::write("prof.dump".as_ref(), cstr) }
```
michaelsproul: LGTM, looking at the source this corresponds to `mallctl("prof.dump", NULL, NULL, filename, size)` which is exactly what we want

Ref:

- https://github.com/jemalloc/jemalloc/wiki/Use-Case%3A-Heap-Profiling
- https://docs.rs/tikv-jemalloc-ctl/0.6.0/src/tikv_jemalloc_ctl/raw.rs.html#137-147

lighthouse/Cargo.toml
```diff
@@ -38,6 +38,8 @@ console-subscriber = ["console-subscriber/default"]
 
 # Deprecated. This is now enabled by default on non windows targets.
 jemalloc = []
+# Optional profiling layer
+jemalloc-profilg = ["malloc_utils/jemalloc-profiling"]
```
michaelsproul: Typo:

```suggestion
jemalloc-profiling = ["malloc_utils/jemalloc-profiling"]
```

michaelsproul (approved):
LGTM! Let's merge before this bitrots (again)!!

michaelsproul:
Actually, don't we need some way to trigger this? Paul had an HTTP API I think

jimmygchen:
Sorry, I didn't include everything - I've pushed all changes now and I'm going to test it first as there were quite a bit of conflicts and changes that I had to make.

jimmygchen:
Got a bit stuck with getting debug symbols in the dump, will come back to this later.

jimmygchen:
Managed to get this to work, I've added a page to the lighthouse book with the instructions.

book/src/SUMMARY.md
```diff
@@ -63,6 +63,7 @@
 * [FAQs](./faq.md)
 * [Protocol Developers](./developers.md)
   * [Lighthouse Architecture](./developers_architecture.md)
+  * [Memory Profiling](./memory_profiling.md)
```
chong-he: Can we rename this to follow a prefix of the section, i.e., `developer_memory_profiling.md`?

book/src/memory_profiling.md
```diff
@@ -0,0 +1,51 @@
...
+```bash
+curl -X POST http://localhost:5052/lighthouse/malloc/prof_dump -H "Content-Type: application/json" -d '"/home/ubuntu/prof.dump"'
+```
+
+## 4. Analyze with `jeprof`
+
+Generate a visualization:
+
+```bash
+jeprof --svg $(which lighthouse) /home/ubuntu/prof.dump > profile.svg
```
chong-he: Giving it a try, after running this command, it is showing:
```
Using local file /home/ck/.cargo/bin/lighthouse.
Using local file /home/ck/prof.dump.
Dropping nodes with <= 1.8 MB; edges with <= 0.4 abs(MB)
sh: 1: dot: not found
```
Did I miss some steps?

jimmygchen: ah ok you need `graphviz` installed. 

```
sudo apt update
sudo apt install graphviz
```

Let me know if this works and I'll update the docs

chong-he: Works, thanks! So these need to be installed: 
`sudo apt install libjemalloc-dev`
`sudo apt install graphviz`

jimmygchen: Thanks, I'll update the docs 🙏

jimmygchen: Done in https://github.com/sigp/lighthouse/pull/7746/commits/3b01dd72f19418704c1778c48a5b81a1e63e7bfa

book/src/SUMMARY.md
jimmygchen: Done in https://github.com/sigp/lighthouse/pull/7746/commits/3b01dd72f19418704c1778c48a5b81a1e63e7bfa, thanks!

jimmygchen:
@michaelsproul I feel like we probably should try to merge this even if the output wasn't very useful this time - it's likely at some point in the future we're going to come back and look for this branch and try to revive it (happened twice already!)
I was a bit unsure because of the memory growth, but now we've identified the OOM cause, I feel more comfortable with this PR now.

We can probably try both jemalloc profiling and heaptrack again after the OOM fix.

chong-he (commented):
Looks good now for the documentation side of the change

jimmygchen:
Closing this one as we have heaptrack now and jemalloc didn’t provide us with meaningful result last time. Will repen if needed

Issue #7745, opened 2025-07-16
Send fcU before fully importing block


michaelsproul:
## The Problem

Presently Lighthouse doesn't send fcU until _after_ fully importing the block, here:

https://github.com/sigp/lighthouse/blob/cfb1f7331064b758c6786e4e1dc15507af5ff5d1/beacon_node/network/src/network_beacon_processor/gossip_methods.rs#L1499-L1505

The call to the EL happens from within that function, here-ish:

https://github.com/sigp/lighthouse/blob/cfb1f7331064b758c6786e4e1dc15507af5ff5d1/beacon_node/beacon_chain/src/canonical_head.rs#L1136-L1142

The problem is, even with recent optimisations, these calls are happening after state diff computations and disk writes, which add 1s+ of latency.

### Background thread for writes

We could add a dedicated background thread for finishing writes of `BeaconState`s to disk. This would allow `import_block` to complete sooner, and for the post-processing (like calling fork choice) to begin. This seems like it would be relatively simple to implement.

The added benefit is that we could also de-couple in-memory states from on-disk states, which would benefit the state advance timer as well. See:

- https://github.com/sigp/lighthouse/issues/4412

### Reorder fork choice update within import

Another option is to kick off a fork choice update at the point where we currently update the early attestation cache:

https://github.com/sigp/lighthouse/blob/cfb1f7331064b758c6786e4e1dc15507af5ff5d1/beacon_node/beacon_chain/src/beacon_chain.rs#L3910-L3913

That code is kind of messy and this might be a good opportunity to refactor it and clean it up.

michaelsproul:
I don't think `Reorder fork choice update within import` is particularly straight-forward. The problem is, if we try to call something like `recompute_head_at_slot_internal` where we previously updated the attester cache (i.e. in `import_block` before disk writes), then the logic that updates the `canonical_head` will error because the block & state won't have been written to disk (nor will they be available in the cache). I.e. we will almost always error here:

https://github.com/sigp/lighthouse/blob/cfb1f7331064b758c6786e4e1dc15507af5ff5d1/beacon_node/beacon_chain/src/canonical_head.rs#L644-L656

I don't see any easy way around this, other than introducing a disconnect between blocks & states cached in memory, and those stored on disk, which is more similar to idea 1. I guess an alternative would be to pass the block and state from `import_block` right through to `recompute_head` so they can be enshrined as the `canonical_head`?

PR #7752, opened 2025-07-17
Rough prototype sending fcU earlier

michaelsproul:
## Issue Addressed

This is option 2 from:

- https://github.com/sigp/lighthouse/issues/7745

## Proposed Changes

- Call the full `recompute_head*` routine in `import_block` so that fcU gets sent to the EL prior to the block being fully imported.
- To work around the assumptions of `recompute_head` (head block and state exist in store): pass the beacon block and state to the function.

## Additional Info

There are multiple problems with the current implementation but I'm curious to see how much this basic approach nets us before going further trying to fix the issues. Incomplete list of current flaws:

- I broke the attester cache's handling of blobs and columns (easy to fix)
- TODO


michaelsproul:
Gonna drop this for now. Can revisit in a Gloas context

PR #7753, opened 2025-07-17
Serialize bpo schedule in asending order

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Serializes the blob_schedule in ascending order to match other clients. This is needed to keep the output of `eth/v1/config/spec` http endpoint consistent across clients.
cc @barnabasbusa

### DIFF
--- a/consensus/types/src/chain_spec.rs
+++ b/consensus/types/src/chain_spec.rs
@@ -1469,7 +1469,7 @@ pub struct BlobParameters {
 
 // A wrapper around a vector of BlobParameters to ensure that the vector is reverse
 // sorted by epoch.
-#[derive(arbitrary::Arbitrary, Serialize, Debug, PartialEq, Clone)]
+#[derive(arbitrary::Arbitrary, Debug, PartialEq, Clone)]
 pub struct BlobSchedule(Vec<BlobParameters>);
 
 impl<'de> Deserialize<'de> for BlobSchedule {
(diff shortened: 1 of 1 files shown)


consensus/types/src/chain_spec.rs
```diff
@@ -2620,6 +2632,19 @@ mod yaml_tests {
             default_max_blobs_per_block_electra()
         );
         assert_eq!(spec.max_blobs_per_block_within_fork(ForkName::Fulu), 20);
+
+        // Check that serialization is in ascending order
+        let yaml = serde_yaml::to_string(&spec.blob_schedule).expect("should serialize");
+
+        // Deserialize back to Vec<BlobParameters> to check order
```
jimmygchen, line 2639: It's a little weird that a serde round trip result in different ordering. Maybe we should reverse again in deserialize? 

It currently don't cause any issue, but if it gets serialized -> deseralized -> serailzied again we'll end up having descending order - might be confusing if this functionality gets used by downstream projects

jimmygchen: Ah nevermind, this is done in the constructor

https://github.com/sigp/lighthouse/blob/03d297ddeb2f69452c0bac395705e3c38c514905/consensus/types/src/chain_spec.rs#L1486-L1490

jimmygchen (approved):
Looks good, thanks!

PR #7754, opened 2025-07-17
Fix enr loading from disk with cgc

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

During building an enr on startup, we weren't using the value in the custody context. 
This was resulting in the enr value getting updated when the cgc updates, the change getting persisted, but getting set back to the default on restart.
This PR takes the value explicitly from the custody context.

### DIFF
--- a/beacon_node/lighthouse_network/src/discovery/enr.rs
+++ b/beacon_node/lighthouse_network/src/discovery/enr.rs
@@ -159,6 +159,7 @@ pub fn build_or_load_enr<E: EthSpec>(
     local_key: Keypair,
     config: &NetworkConfig,
     enr_fork_id: &EnrForkId,
+    custody_group_count: Option<u64>,
     next_fork_digest: [u8; 4],
     spec: &ChainSpec,
 ) -> Result<Enr, String> {
@@ -166,7 +167,14 @@ pub fn build_or_load_enr<E: EthSpec>(
     // Note: Discovery should update the ENR record's IP to the external IP as seen by the
     // majority of our peers, if the CLI doesn't expressly forbid it.
     let enr_key = CombinedKey::from_libp2p(local_key)?;
-    let mut local_enr = build_enr::<E>(&enr_key, config, enr_fork_id, next_fork_digest, spec)?;
+    let mut local_enr = build_enr::<E>(
+        &enr_key,
+        config,
+        enr_fork_id,
+        custody_group_count,
+        next_fork_digest,
+        spec,
+    )?;
 
     use_or_load_enr(&enr_key, &mut local_enr, config)?;
     Ok(local_enr)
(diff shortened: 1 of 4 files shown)


beacon_node/lighthouse_network/src/discovery/enr.rs
```diff
@@ -159,14 +159,22 @@ pub fn build_or_load_enr<E: EthSpec>(
...
     let enr_key = CombinedKey::from_libp2p(local_key)?;
-    let mut local_enr = build_enr::<E>(&enr_key, config, enr_fork_id, next_fork_digest, spec)?;
+    let mut local_enr = build_enr::<E>(
+        &enr_key,
+        config,
+        enr_fork_id,
+        custody_group_count,
+        next_fork_digest,
+        spec,
+    )?;
```
jimmygchen, line 177: ah i didnt think this was needed because i thought we'd use the one from disk

https://github.com/sigp/lighthouse/blob/b43e0b446cd1d8a3f5b3c4a9f614f950ef672aa8/beacon_node/lighthouse_network/src/discovery/enr.rs#L123-L128

But then the above scenario evaluates to false because we set the CGC to the default value when building the ENR, so it doesn't survive across restarts.

Your fix makes sense 👍

beacon_node/lighthouse_network/src/discovery/enr.rs
```diff
@@ -271,14 +280,15 @@ pub fn build_enr<E: EthSpec>(
...
-            if let Some(false_cgc) = config.advertise_false_custody_group_count {
-                false_cgc
-            } else if config.subscribe_all_data_column_subnets {
-                spec.number_of_custody_groups
-            } else {
-                spec.custody_requirement
-            };
+        let custody_group_count = if let Some(cgc) = custody_group_count {
+            cgc
+        } else if let Some(false_cgc) = config.advertise_false_custody_group_count {
```
jimmygchen, line 285: oh i think we should prioritise `false_cgc` over the actual CGC here?

jimmygchen: ah right, it's getting passed in from here
https://github.com/sigp/lighthouse/blob/e7ff026b85fac8c96072ca2401ebfea762fe524c/beacon_node/lighthouse_network/src/service/mod.rs#L201-L204

So this is actually just a redundant /unreachable fallback?

jimmygchen (approved):
Looks good to me, thanks for the fix!

beacon_node/lighthouse_network/src/discovery/enr.rs
pawanjay176: I hadn't considered that we want false_cgc to be prioritized. But looks like we already do because of this.

PR #7755, opened 2025-07-18
Rough prototype for architectural changes needed to introduce execution proofs

kevaundray:
This PR is not meant to be merged. It's mainly to discover all of the architectural changes needed

Notes: https://hackmd.io/@kevaundray/BJeZCo5Tgx

## Issue Addressed

The addition of zkEVMs to Ethereum L1 allows the CL to no longer need an EL to verify execution payloads, the execution payload content is moreover replaced with a cryptographic proof that does not scale linearly with the size of the execution payload proof. We will still need the execution payload header.

The current idea for rolling out the changes for zkEVMs is to iteratively add the changes needed such that we can safely test them and uncover edge cases, blockers, etc. safely here means that it should not affect the existing CL logic. 

## Proposed Changes

**Proof generating nodes**

Someone must create the execution proofs for the execution payloads. Long term, when proofs are mandatory, we believe this will be the builders, since they are incentivised to not have their blocks be re-orged for being invalid. 

While we are in the interim stage where proofs are not mandatory, there is no incentive to create proofs. The idea here is that these proofs will be subsidised until then.

Proof generating nodes are the nodes whom will generate these proofs in the protocol.

_new_payload_

Nodes can opt to be proof generating nodes which means that whenever they receive a beacon block and need to verify the execution payload via engine_new_payload, they will also generate proofs and submit these proofs over the new proof subnets.

_get_payload_

These nodes can also opt to generate proofs for blocks they have proposed. I believe we could just have proof generation only in new_payload and omit it in get_payload -- the idea behind not doing it this way was that if you're the proposer, you can start generating proofs as soon as you receive the payload from the EL.

**Stateless attestor** 

_new_payload_

Stateless nodes do not have an EL attached (in practice) and will wait for proofs to be received on whichever proof subnets they have subscribed to, in order to validate the execution payload. 

The way it has been implemented in this PR is that if the node is a stateless validator, they will mark all payloads as optimistic, moreover, since fork choice is never modified, the node currently will always be in optimistic sync mode.

_get_payload_

Stateless nodes cannot locally build blocks because they have no EL state, they must use mev-boost. This PR does not add the mev-boost functionality, so stateless nodes cannot build blocks.

**Proof generation**

Currently proof generation has been stubbed out since it is not needed for running local testnets. The main feature we need here is determinstic output wrt the execution payload.

**Proof chain**

This code deliberately does not modify fork choice to incorporate proof verification. We instead have a structure that keeps track of what blocks have been proven alongside the beacon chain. This means that one could run a stateless validator alongside mainnet/testnet.

## Additional Info

- Ran some preliminary tests using kurtosis by running `./scripts/local_testnet/start_local_testnet.sh`


kevaundray:
A lot of the invasive changes were due to adding new gossipsub topics/subnets -- I've tried to put most of the new code into new files so that they don't touch the existing code, making the diff easier to see

book/src/advanced_stateless_integration.md
```diff
@@ -0,0 +1,455 @@
...
+    pub block_hash: ExecutionBlockHash,
+    /// The subnet ID where this proof was received/should be sent
+    pub subnet_id: ExecutionProofSubnetId,
+    /// Version of the proof format
+    pub version: u32,
+    /// Opaque proof data - structure depends on subnet_id and version
+    /// This contains cryptographic proofs from zkVMs or other proof systems
+    pub proof_data: Vec<u8>,
+    /// Timestamp when this proof was generated (Unix timestamp)
+    pub timestamp: u64,
```
kevaundray: This was removed

book/src/advanced_stateless_integration.md
kevaundray, line 1: This file is mainly here for notetaking, so can ignore

beacon_node/beacon_chain/src/beacon_chain_execution_proof.rs
```diff
@@ -0,0 +1,229 @@
...
+                );
+            }
+        }
+
+        // Remove pending blocks that now have sufficient proofs
+        let proven_blocks = self
+            .execution_payload_proof_store
+            .take_pending_blocks(&execution_block_hash);
+        // Note: That if we were to modify fork choice, it would likely be here, where we know what set of
+        // beacon blocks have valid execution payloads.
```
kevaundray: As the comment mentions, one could trigger head recomputation here based on the proven blocks

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3907,6 +3910,14 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
                 .map_err(|e| BlockError::BeaconChainError(Box::new(e.into())))?;
         }
 
+        // Register optimistic blocks for proof validation in stateless validation mode
+        if payload_verification_status.is_optimistic() {
+            if let Ok(execution_payload) = block.execution_payload() {
+                let execution_block_hash = execution_payload.block_hash().into();
+                self.register_optimistic_block_for_proof(block_root, execution_block_hash);
+            }
+        }
```
kevaundray: This is where the beacon block is registered in the proof store and we start listening for proofs of the particular execution_block_hash

beacon_node/beacon_chain/src/execution_payload.rs
```diff
@@ -540,3 +589,133 @@ where
...
+        }
+    }
+
+    Ok(())
+}
+
+fn extract_execution_payload<E: EthSpec>(
+    block: BeaconBlockRef<'_, E, FullPayload<E>>,
+) -> Result<ExecutionPayload<E>, BeaconStateError> {
+    // Extract ExecutionPayload using the FullPayload's execution_payload field
```
kevaundray:
```suggestion
```

beacon_node/beacon_chain/src/execution_proof_generation.rs
kevaundray, line 1: Proof generation and verification are essentially the two exported methods in here that have been stubbed out

beacon_node/beacon_chain/src/execution_proof_store.rs
kevaundray, line 1: This is likely the bulk of the PR, it is a "Proof Store" that keeps track of the blocks that we've proven, the proofs that we've received and general proof management

beacon_node/client/src/execution_proof_broadcaster.rs
kevaundray, line 1: This file is in charge of broadcasting proofs, it speaks with the proof store in order to know what exactly needs broadcasting

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -3167,4 +3167,124 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
             write_file(error_path, error.to_string().as_bytes());
         }
     }
+
+    /// Process a gossip execution proof message.
+    pub async fn process_gossip_execution_proof(
```
kevaundray, line 3259: Logic for handling gossip proof messages; the subnet ID lets us know the proof type/proving system

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -3167,4 +3167,124 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+            );
+            return;
+        }
+
+        // Validate subnet ID matches "proof ID"
+        //
+        // Note: `subnet_id_u64` was the subnet that the message was received on
+        // while `execution_proof.subnet_id` was the subnet ID embedded in the proof
+        // TODO: We could possibly remove the subnet ID being embedded in the execution proof
+        // TODO: Its really the ProofID, so we could change it to `execution_proof.proof_id`
```
kevaundray:
```suggestion
```

beacon_node/src/cli.rs
```diff
@@ -1509,6 +1509,36 @@ pub fn cli_app() -> Command {
...
+                .long("generate-execution-proofs")
+                .action(ArgAction::SetTrue)
+                .help_heading(FLAG_HEADER)
+                .help("Generate execution proofs for all blocks (both produced and received). \
+                       This makes the node act as a proof generator for the network. \
+                       Cannot be used with --stateless-validation.")
+                .display_order(0)
+        )
+        .arg(
+            Arg::new("stateless-min-proofs-required")
```
kevaundray, line 1500: `stateless-min-proofs-required` -- There will be more than one proof for a particular execution payload. This number essentially says how many proofs you want to receive before you deem the block to be valid. So its like a k-out-of-n

consensus/types/src/execution_proof_subnet_id.rs
```diff
@@ -0,0 +1,138 @@
...
+#[derive(arbitrary::Arbitrary, Clone, Copy, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct ExecutionProofSubnetId(#[serde(with = "serde_utils::quoted_u64")] u64);
+
+impl ExecutionProofSubnetId {
+    /// Create an ExecutionProofSubnetId from a u64, validating it's within bounds
+    ///
+    /// Note: bounds here relates to the fact that there is a maximum number of subnets
+    /// that we can have; it is the number of maximum number of proofs that we will accept.
+    pub fn new(id: u64) -> Result<Self, InvalidSubnetId> {
```
kevaundray, line 30: Main difference between this subnet wrapper type and others, like the ones for columns, is the fact that it has this validation check.

beacon_node/beacon_chain/src/chain_config.rs
```diff
@@ -142,6 +172,11 @@ impl Default for ChainConfig {
             prepare_payload_lookahead: Duration::from_secs(4),
             // This value isn't actually read except in tests.
             optimistic_finalized_sync: true,
+            stateless_validation: false,
+            generate_execution_proofs: false,
+            max_execution_payload_proofs: 10_000,
```
dapplion: What's the expected time between receiving a block and then receiving the proof? We had a similar debate for the cache the holds blocks without blobs where initially we made it such that it could hold a very large numbers of blocks by storing them on disk. However for simplicity we later change it to hold only items in memory since you since blobs are expected to come relatively fast. I'm not sure of the value of holding a block whose data arrives 1 day later. Similar reasoning may apply to this new cache for proofs.

beacon_node/beacon_chain/src/execution_payload.rs
```diff
@@ -540,3 +589,132 @@ where
...
+                "Failed to extract execution payload for proof generation: {:?}",
+                e
+            );
+            return;
+        }
+    };
+
+    // Spawn the proof generation task in the background
+    // WARNING: No resource limits or task counting is performed here.
+    // TODO: Implement a task queue with concurrency limits and resource monitoring.
```
dapplion: We could send a new task to the beacon processor, which would limit resource consumption. Do you expect this process to have heavy parallelization? Reconstruction is breaking a little our current model as it consumes too many threads @jimmygchen

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -96,20 +96,22 @@ pub enum DataColumnReconstructionResult<E: EthSpec> {
...
-/// Indicates if the block is fully `Available` or if we need blobs or blocks
-///  to "complete" the requirements for an `AvailableBlock`.
-pub enum Availability<E: EthSpec> {
+/// Indicates if the block is ready for import or if we need blobs, columns, or execution proofs
+/// to "complete" the requirements for an importable block.
+pub enum Importability<E: EthSpec> {
     MissingComponents(Hash256),
-    Available(Box<AvailableExecutedBlock<E>>),
+    ReadyForImport(Box<AvailableExecutedBlock<E>>),
 }
```
kevaundray: This change has not been propoerly propagated, because it would cause quite a lot of conflicts

kevaundray: Overarching change would be that da_checker becomes an fork_choice_import_checker

kevaundray: I may change this back as it causes a lot of sprinkled changes that may end up causing a lot of merge conflicts

kevaundray: reverted here https://github.com/sigp/lighthouse/pull/7755/commits/5013d8405730b2dff069ff2f187567ee3483b132 to reduce noise

kevaundray:
One restriction we can explicitly apply is: execution proofs must be signed by a validator (not just the current proposer) -- this allows us to automatically support the fallback technique, slash validators that sign over proofs that are incorrect -- this is also fine because we expect builders to be validators

(We could loosen up the condition that they get slashed if the execution proof is wrong -- not sure why to do this, but perhaps we can start off light and then make it stricter in the future)

0xalizk:
> slash validators that sign over proofs that are incorrect

Wouldn't this imply the enshrining of the verifiers?

kevaundray:
> > slash validators that sign over proofs that are incorrect
> 
> 
> 
> Wouldn't this imply the enshrining of the verifiers?

Yes thats a good point -- so it can't be done with the initial (optional) execution proofs feature

beacon_node/beacon_chain/src/execution_proof_verification.rs
```diff
@@ -179,7 +178,9 @@ fn validate_execution_proof_for_gossip<T: BeaconChainTypes, O: ObservationStrate
     }
 
     // TODO: Add timing validation based on slot
-    // TODO: Add duplicate proof detection
+    if O::observe() {
+        // TODO: Add duplicate proof detection
```
kevaundray, line 183: Related to this, one assumption also being made is that the block will arrive before the proofs. This assumption is reasonably sound given the time it takes to create proofs, however I can see it being broken in the following cases:

- The slot time is decreased (implying the execution payload gas limit is lowered) which would mean that proofs are created faster
- A node is not well-connected and does not receive the block until after the proof due to network lag. This may happen more frequently for empty blocks.

One solution here would be to store proofs for blocks currentSlot + N in the future, but we would need to store the slot number also in the execution proof message because an unknown beacon root cannot tell you how far in the future it is

beacon_node/beacon_chain/src/execution_proof_generation.rs
```diff
@@ -0,0 +1,300 @@
...
+    proof_id: ExecutionProofSubnetId,
+) -> ExecutionProof {
+    let execution_block_hash = payload.block_hash();
+    let block_number = payload.block_number();
+
+    // Simulate (some) proof computation delay
+    // In a real implementation, this would be the time needed for zkVM local proof generation
+    // or communication with external proof generation services
+    use rand::{Rng, rng};
+    let delay_ms = rng().random_range(1000..=3000);
```
dapplion, line 34: Should this be a flag to allow to test the system with slower proof generation?

beacon_node/beacon_chain/src/execution_proof_network.rs
```diff
@@ -0,0 +1,158 @@
...
+        task_name,
+    );
+}
+
+/// Get configured proof subnets for this node
+pub fn get_configured_proof_subnets<T: BeaconChainTypes>(chain: &Arc<BeaconChain<T>>) -> Vec<u64> {
+    // TODO: For now, the node will generate proofs for all available subnets.
+    // TODO: In the future, they should be able to configure this for proofs
+    // TODO: they can generate for. Mainly for altruistic nodes that want to
+    // TODO: seed the network.
```
dapplion: This should be a CLI argument where you choose the set of subnets

beacon_node/beacon_chain/src/execution_proof_network.rs
```diff
@@ -0,0 +1,158 @@
...
+        FullPayloadRef::Deneb(payload) => {
+            ExecutionPayload::Deneb(payload.execution_payload.clone())
+        }
+        FullPayloadRef::Electra(payload) => {
+            ExecutionPayload::Electra(payload.execution_payload.clone())
+        }
+        FullPayloadRef::Fulu(payload) => ExecutionPayload::Fulu(payload.execution_payload.clone()),
+        FullPayloadRef::Gloas(payload) => {
+            ExecutionPayload::Gloas(payload.execution_payload.clone())
+        }
```
dapplion: @michaelsproul is there a macro over the fields we can use here?

beacon_node/beacon_chain/src/execution_proof_verification.rs
```diff
@@ -0,0 +1,235 @@
...
+        Slot::new(0)
+    }
+
+    /// Convert into the inner verified execution proof.
+    pub fn into_inner(self) -> VerifiedExecutionProof {
+        self.proof
+    }
+}
+
+/// Validate an execution proof for gossip according to the rules defined in the consensus specs.
```
dapplion, line 160: Should we verify the ZK proof in gossip validation? How fast do you expect it to be?

beacon_node/lighthouse_network/src/types/topics.rs
```diff
@@ -91,6 +96,16 @@ pub fn core_topics_to_subscribe<E: EthSpec>(
         }
     }
 
+    // Subscribe to all execution proof subnets when stateless validation is enabled
+    if opts.stateless_validation {
```
dapplion, line 100: If all nodes subscribe to all execution proof subnets by default, why have separate subnets at all?

book/src/advanced_stateless_integration.md
dapplion: Can you move this file to a hackmd and link the hackmd at the top of the PR description?

consensus/types/src/execution_proof_subnet_id.rs
```diff
@@ -0,0 +1,139 @@
...
+/// but no node can exceed this protocol maximum.
+///
+/// The value of 8 subnets provides a good balance between:
+/// - Proof diversity (multiple independent proofs per block)
+/// - Network overhead (not too many gossip topics)
+/// - Resource requirements (reasonable for most nodes(?))
+///
+/// In reality, I do not think we will have 8, more closer to 3, though this is still being
+/// explored. This number could be larger if we consider combining different zkVMs with different guests.
+pub const MAX_EXECUTION_PROOF_SUBNETS: u64 = 8;
```
dapplion: If producers only subscribe and emit to a specific of subnets they support, and consumers only ready the set of incoming proofs they support, why do we need a max count of overall subnets?

beacon_node/beacon_chain/src/chain_config.rs
kevaundray: Current proof creation deadlines are around 8-10 seconds

beacon_node/beacon_chain/src/execution_payload.rs
kevaundray: I think in practice, the beacon chain will always make an API call here -- the proof generation process will require GPUs so it may be highly unlikely that the validator is on the same setup with a bunch of GPUs

beacon_node/beacon_chain/src/execution_proof_generation.rs
kevaundray: If you think so -- should note that we will only be testing (dummy proof generation here) because actual proof generation is too heavy (needs GPUs)

beacon_node/beacon_chain/src/execution_proof_network.rs
kevaundray: Should the user know what the subnets represent? ie subnet 0 might be the reth-risc0 subnet, oe just let them specify subnet0 and the mapping is noted somewhere else

beacon_node/beacon_chain/src/execution_proof_verification.rs
kevaundray: In the region of 100ms, but would need to benchmark this to give you concrete numbers for each zkEVM

beacon_node/lighthouse_network/src/types/topics.rs
kevaundray: This was just for the initial prototype -- nodes can choose how many subnets they subscribe to based on their bandwidth and thoughts around zkVM/client diversity

dapplion:
Some notes from my latest review:

- Who participates in the zk proofs subnets?
- We should add SSE for proofs as they are verified, this allows to build a 3rd party tool that tracks progress
- Need to modify the peer manager to mantain peers in the zk proof subnets
- Need to update ENR with zkproof subnets? Or a new metadata query to discover which proofs they listen to
Where do we trigger subscription of gossip topics?
- If all nodes subscribe to all execution proof subnets by default, why have separate subnets at all?

kevaundray:
> If all nodes subscribe to all execution proof subnets by default, why have separate subnets at all?

To rephrase this question; it is analogous to asking "execution proofs fungible" -- ie should nodes care if they receive a reth-sp1 proof vs a geth-risc0 proof. If they do then having separate subnets is useful and nodes can explicitly choose to receive one over the other, if they don't then having one subnet and receiving all of the different proofs on that subnet makes more sense

kevaundray:
Note from Teku; when Teku sends a proof before the block LH nodes will disconnect them since we currently assume that the block comes before the proof

kevaundray:
TODO: Decrease the number of subnets to 1

kevaundray:
Add issue to allow optional EL

kevaundray:
> Note from Teku; when Teku sends a proof before the block LH nodes will disconnect them since we currently assume that the block comes before the proof

This seems to be working as expected, ie proofs that arrive before the block are put into PendingComponents

beacon_node/beacon_chain/src/execution_payload.rs
```diff
@@ -452,6 +461,16 @@ where
...
         .ok_or(BlockProductionError::ExecutionLayerMissing)?;
 
+    // Check for stateless validation mode
+    if chain.config.stateless_validation {
+        // TODO(zkproofs): We could return an empty payload here until we hook up mev-boost
+        debug!(
+            "ERROR: Cannot produce blocks in stateless validation mode - no execution layer attached. \
+            TODO: Use MEV-boost for block production in stateless validation mode."
+        );
+        return Err(BlockProductionError::ExecutionLayerMissing);
```
dapplion: Add a new Err variant here specifically for this

scripts/local_testnet/network_params.yaml
```diff
@@ -1,12 +1,32 @@
...
+      - --stateless-min-proofs-required=3
+    count: 2
+  # Regular stateful nodes (normal operation, no proof generation)
+  - el_type: geth
+    el_image: ethereum/client-go:latest
+    cl_type: lighthouse
+    cl_image: lighthouse:local
+    cl_extra_params:
+      - --target-peers=3
+    count: 1
```
dapplion: You should add a new kurtosis params file. If you want to test this in CI add a new job

michaelsproul:
Hi Kev, I'm sorry this hasn't been reviewed and merged. We are going to try prioritising this once we've got the data availability checker (https://github.com/sigp/lighthouse/pull/8716) in. Gloas DA is likely to conflict with this PR because it also touches the execution payload verification pipeline. However once it's in it should streamline things.

kevaundray:
> Hi Kev, I'm sorry this hasn't been reviewed and merged. We are going to try prioritising this once we've got the data availability checker (#8716) in. Gloas DA is likely to conflict with this PR because it also touches the execution payload verification pipeline. However once it's in it should streamline things.

No worries! We've since moved development to https://github.com/eth-act/lighthouse/pull/24 (this should now also follow the version in the consensus specs)

kevaundray:
I'll close this to avoid confusion :)

PR #7756, opened 2025-07-18
Avoid attempting to serve blobs after Fulu fork

chong-he:
## Issue Addressed

* #7122


beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -279,22 +279,42 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+            let BlobIdentifier {
+                block_root: root,
+                index,
+            } = id;
+
             // First attempt to get the blobs from the RPC cache.
             if let Ok(Some(blob)) = self.chain.data_availability_checker.get_blob(id) {
+                // Check if the blob requested is from a Fulu slot, if so, skip the current blob id and proceed to the next
+                if let Some(fulu_slot) = fulu_start_slot {
+                    if blob.slot() >= fulu_slot {
```
michaelsproul: This can be simplified slightly using `fulu_start_slot.is_some_and(|fulu_slot| blob.slot() >= fulu_slot)`.

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -306,6 +326,20 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                     Ok(blobs_sidecar_list) => {
                         'inner: for blob_sidecar in blobs_sidecar_list.iter() {
                             if blob_sidecar.index == *index {
+                                // Same logic as above to check for Fulu slot
+                                if let Some(fulu_slot) = fulu_start_slot {
+                                    if blob_sidecar.slot() >= fulu_slot {
```
michaelsproul: As above

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: The `fulu_slot` is used in the logging below. So after changing, the compiler is saying `fulu_sot` not found. 

Should I remove `fulu_slot` from the logging?

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -279,22 +279,42 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
         let mut blob_list_results = HashMap::new();
         for id in request.blob_ids.as_slice() {
+            let BlobIdentifier {
+                block_root: root,
+                index,
+            } = id;
+
             // First attempt to get the blobs from the RPC cache.
             if let Ok(Some(blob)) = self.chain.data_availability_checker.get_blob(id) {
+                // Check if the blob requested is from a Fulu slot, if so, skip the current blob id and proceed to the next
```
jimmygchen: I think if the block is after fulu slot, this condition will never satisfy, as the result will always be `Ok(None)` as we never store blobs from Fulu. I don't think we need to handle it here as it should be unreachable.

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -279,22 +279,42 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
             .collect::<Vec<_>>();
         let mut send_blob_count = 0;
 
+        let fulu_start_slot = self
```
jimmygchen, line 298: not related to your PR, but I realised the logging for `BlobsByRoot outgoing response processed` below is wrong (i think we knew this but never managed to fix it), as it assumed there's only one block root in the request (it reads the `block_root` of the first `blob_ids`)

This is an actual production log - it kinda implies that there are multiple roots in the request, could fix it in a separate PR though, but thought it's relevant to mention:

> Jul 28 05:48:46.004 DEBUG BlobsByRoot outgoing response processed       peer_id: 16Uiu2HAmDUHfLY9HnuH2ne3EWYr5f4jbHctJFTPFQMAt8iwSFASa, requested_root: 0x70d411c60a0a6e83acc2790f8e7574fa3ca5810ccf8babbb1267b823d5a542e0, requested_indices: [0, 1, 2, 0, 1, 2, 3, 4, 5, 0, 1, 2, 3, 4, 5, 6, 0, 1, 2, 3, 0, 1, 2, 3, 4, 5, 6, 0, 1, 2, 3, 4, 5, 6], returned: 34

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -306,6 +326,20 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                     Ok(blobs_sidecar_list) => {
                         'inner: for blob_sidecar in blobs_sidecar_list.iter() {
                             if blob_sidecar.index == *index {
+                                // Same logic as above to check for Fulu slot
+                                if let Some(fulu_slot) = fulu_start_slot {
```
jimmygchen: This will also be unreachable. 

The purpose to avoid serving post-fulu slots in the byRange method is to save us work from doing block roots lookup - it doesn't seem as relevant here, as we've already hit the database and incurred the cost.

The spec also says the following, so we can't really penalise these peers during this deprecation period
> Clients SHOULD NOT penalize peers for requesting blob sidecars from FULU_FORK_EPOCH.

I think we could probably check all the caches to see if we have the block before hitting the database?

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -884,6 +918,36 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
         );
 
         let request_start_slot = Slot::from(req.start_slot);
+        // This variable may only change when the request_start_slot + req.count spans across the Fulu fork slot
+        let mut effective_count = req.count;
+
+        if let Some(fulu_epoch) = self.chain.spec.fulu_fork_epoch {
+            let fulu_start_slot = fulu_epoch.start_slot(T::EthSpec::slots_per_epoch());
+            let request_end_slot = request_start_slot + req.count - 1;
```
jimmygchen: we should probably use safe math here, as `req.count` is an external input and could lead to overflow, even though rate limiter by default should prevent large numbers reaching here - but we can't rely on it as it can be disabled.

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -884,6 +918,36 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                debug!(
+                    %peer_id,
+                    %request_start_slot,
+                    %fulu_start_slot,
+                    returned = 0,
+                    "BlobsByRange request is at or after a Fulu slot, returning empty response"
+                );
+                return Ok(());
+            // For the case that the request slots spans across the Fulu fork slot
+            } else if request_start_slot < fulu_start_slot && request_end_slot >= fulu_start_slot {
```
jimmygchen: `request_start_slot < fulu_start_slot` is redundant, and can be removed to simplify this

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -884,6 +918,36 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                effective_count = (fulu_start_slot - request_start_slot).as_u64();
+                debug!(
+                    %peer_id,
+                    %request_start_slot,
+                    %fulu_start_slot,
+                    requested = req.count,
+                    returned = effective_count,
```
jimmygchen: `returned` is probably a bit misleading, i think `effective_count` is fine?

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -884,6 +918,36 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
         );
 
         let request_start_slot = Slot::from(req.start_slot);
+        // This variable may only change when the request_start_slot + req.count spans across the Fulu fork slot
+        let mut effective_count = req.count;
```
jimmygchen: You could make the `if / else` block return the count, and that would allow you to remove the `mut`

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -884,6 +918,36 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+
+        if let Some(fulu_epoch) = self.chain.spec.fulu_fork_epoch {
+            let fulu_start_slot = fulu_epoch.start_slot(T::EthSpec::slots_per_epoch());
+            let request_end_slot = request_start_slot + req.count - 1;
+
+            // If the request_start_slot is at or after a Fulu slot, return empty response
```
jimmygchen:
```suggestion
            // If the request_start_slot is at or after a Fulu slot, return an empty response
```

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
jimmygchen: Additionally we could add a similar check to this in this function for safety, but for blobs

https://github.com/sigp/lighthouse/blob/befef72f92f742fab2846203deae4e326207274b/beacon_node/network/src/network_beacon_processor/rpc_methods.rs#L1077-L1083

jimmygchen (changes requested):
Hi @chong-he I've added some comments.
I think it might be worth adding something similar to data columns methods, to avoid block roots retrieval if the slots are before Fulu

https://github.com/sigp/lighthouse/blob/befef72f92f742fab2846203deae4e326207274b/beacon_node/network/src/network_beacon_processor/rpc_methods.rs#L1064-L1085

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: Revised in [bf20880](https://github.com/sigp/lighthouse/pull/7756/commits/bf208808a65bea6765b5071cf6039f18232827b0)

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: I will look into this logging issue, and we can fix in another PR?

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: Revised in [bf20880](https://github.com/sigp/lighthouse/pull/7756/commits/bf208808a65bea6765b5071cf6039f18232827b0), trying to avoid touching the database

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: Safe mathematics using `saturating_add` in [792c0d3](https://github.com/sigp/lighthouse/pull/7756/commits/792c0d320590e18bb92dffd7b79e37506099fd0b). 

By the way, in line 1081 as above, should it be:
`Request exceeded MAX_REQUEST_DATA_COLUMN_SIDECARS`?

If so, I can fix it in this PR too?

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: Fixed

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: Try to revise this in [32c3ab0](https://github.com/sigp/lighthouse/pull/7756/commits/32c3ab0d4acbd7f7ccbc0e331bdceffb76329b46), not sure if it is good?

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: Fixed

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: Fixed

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -279,8 +279,60 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                    .chain
+                    .data_availability_checker
+                    .get_execution_valid_block(root)
+                    .or_else(|| self.chain.early_attester_cache.get_block(*root))
+                {
+                    let slot = block.slot();
+                    retrieve_blob_slot.insert(*root, slot);
+                    slot
+                } else {
+                    match self.chain.get_blinded_block(root) {
```
chong-he: Not sure if this is expensive to look for?

jimmygchen: This also hits the database, but is way cheaper than loading a full beacon block or a blob.
Not sure if it's necessary though - or could be moved down, because we might be able to get the blobs directly from the DA checker (below) if it's already there, and could avoid a database read.

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
jimmygchen: Yeah separate PR is fine, it's not related - i brought it up because i noticed it while reviewing

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
jimmygchen: Yes, good catch

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
jimmygchen: yep that's better!

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
jimmygchen: Maybe we just attempt to serve the blobs if we don't see the block in the caches? 
This method will be deprecated in the release after the Fulu fork. It may not be worth adding the overhead for all requests before/after the fork. What do you think?

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: Fixed in [c762e91](https://github.com/sigp/lighthouse/pull/7756/commits/c762e918ce1b49e83f3ab9dc093434a459ac7ad3)

YourLUSTGod:
www.Makavelirecords666.bandcamp.com

YourLUSTGod:
Vivetuvidasincensura@gmail.com

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: > Maybe we just attempt to serve the blobs if we don't see the block in the caches?

So you mean if we don't see the block here: https://github.com/sigp/lighthouse/pull/7756/files#diff-45a20c4733163c4e73ce1a41a9bd86dc991684c4dba55054ffe547efe5f586dfR301-R310, then we just try serving the blobs (as in using the current code)? 

That means we can remove the part of the code in added in `BlobsByRoot` in this PR?

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: I look into it and it's a small change, so might as well change in this PR. I revised in this commit: https://github.com/sigp/lighthouse/pull/7756/commits/f79124465c4546733967c9cb978abdbbc0607818 
(also because it's easier as there will be no merge conflict later)

The `send_blob_count` could be redundant after the change in logging, but I keep it for now, waiting to hear your opinion on this

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: In [029e418](https://github.com/sigp/lighthouse/pull/7756/commits/029e418ffa41cd793dfccc2a692155459f2ab84b), I change to return None if the block root is not found in cache, and proceed as normal. Not sure if this is what you mean?

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -893,6 +951,49 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
         );
 
         let request_start_slot = Slot::from(req.start_slot);
+        let request_start_epoch = request_start_slot.epoch(T::EthSpec::slots_per_epoch());
+        // Should not send more than max request blob sidecars
+        if req.max_blobs_requested(request_start_epoch, &self.chain.spec)
+            > self.chain.spec.max_request_blob_sidecars_electra
```
jimmygchen: should we use the `spec.max_request_blob_sidecars(fork_name)` function here?
keeping those fields private means we don't leak the logic elsewhere -if we ever need to use a different config value for a different fork, we only need to update the above function.

beacon_node/network/src/network_beacon_processor/tests.rs
```diff
@@ -1365,3 +1373,137 @@ async fn test_blobs_by_range() {
...
+        } else {
+            panic!("unexpected message {:?}", next);
+        }
+    }
+    // Post-Fulu should return 0 blobs
+    assert_eq!(0, actual_count);
+}
+
+#[tokio::test]
+async fn test_blobs_by_range_spans_fulu_fork() {
```
jimmygchen, line 1686: I don't think this test would fail before your PR?
but i think it's useful that it covers blobs by range works before Fulu

beacon_node/network/src/network_beacon_processor/tests.rs
```diff
@@ -1365,3 +1373,137 @@ async fn test_blobs_by_range() {
...
+        } else {
+            panic!("unexpected message {:?}", next);
+        }
+    }
+    assert_eq!(blob_count, actual_count);
+}
+
+#[tokio::test]
+async fn test_blobs_by_root_post_fulu_should_return_empty() {
+    // Only test for Fulu fork
```
jimmygchen, line 1806: i think it would be worth adding a test to confirm it still work right before fulu fork?

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -268,21 +268,64 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
         inbound_request_id: InboundRequestId,
         request: BlobsByRootRequest,
     ) -> Result<(), (RpcErrorResponse, &'static str)> {
-        let Some(requested_root) = request.blob_ids.as_slice().first().map(|id| id.block_root)
+        let Some(_requested_root) = request.blob_ids.as_slice().first().map(|id| id.block_root)
```
jimmygchen: This block is now redundant and can be removed

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -268,21 +268,64 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
         let mut send_blob_count = 0;
 
+        let fulu_start_slot = self
+            .chain
+            .spec
+            .fulu_fork_epoch
+            .map(|epoch| epoch.start_slot(T::EthSpec::slots_per_epoch()));
+
         let mut blob_list_results = HashMap::new();
+        let mut retrieve_blob_slot = HashMap::new();
```
jimmygchen: I think it's fine being a bit more verbose here to help readability -  what about using the map naming convention `values_by_key`, e.g. `slots_by_block_root`?

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -268,21 +268,64 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                    .chain
+                    .data_availability_checker
+                    .get_execution_valid_block(root)
+                    .or_else(|| self.chain.early_attester_cache.get_block(*root))
+                {
+                    let slot = block.slot();
+                    retrieve_blob_slot.insert(*root, slot);
+                    Some(slot)
+                } else {
+                    // Blobs not found in cache, return None to avoid hitting the database
```
jimmygchen: This is the case where block is not found in cache, and we just need to hit the database right?

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -308,12 +347,27 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                                {
+                                    debug!(
+                                        %peer_id,
+                                        request_root = %root,
+                                        blob_slot = %blob_sidecar.slot(),
+                                        %fulu_slot,
+                                        "BlobsByRoot request is at or after Fulu slot, returning empty response"
+                                    );
+                                    break 'inner;
+                                }
```
jimmygchen: I think we can remove this block - if we have the blob then it means fulu is not activated and there's no need to check for fulu slot

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -893,6 +951,49 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+        let effective_count = if let Some(fulu_epoch) = self.chain.spec.fulu_fork_epoch {
+            let fulu_start_slot = fulu_epoch.start_slot(T::EthSpec::slots_per_epoch());
+            let request_end_slot = request_start_slot.saturating_add(req.count) - 1;
+
+            // If the request_start_slot is at or after a Fulu slot, return an empty response
+            if request_start_slot >= fulu_start_slot {
+                debug!(
+                    %peer_id,
+                    %request_start_slot,
+                    %fulu_start_slot,
```
jimmygchen: i think we can remove all the fulu start slot logging to reduce the logging noise - this doesn't give us much during debugging as there's many easy way to find out, e.g. /config/spec, network config etc

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -329,13 +383,17 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
-        );
+
+        // log once per block_root
+        for (block_root, blobs_indices) in &logging {
+            debug!(
+                %peer_id,
+                %block_root,
+                ?blobs_indices,
+                returned = send_blob_count,
+                "BlobsByRoot outgoing response processed"
```
jimmygchen: This log isn't accurate because `send_blob_count` is the total number of returned responses.
 
Maybe we could just log one entry, and the list of unique roots requested?

beacon_node/network/src/network_beacon_processor/tests.rs
chong-he: Yes this test wouldn't fail even without this PR change, which is why I think the test is not robust enough.

I have already started from mid-epoch slot, and the blob request spans across fork boundary. I wonder if I am missing something in the test?

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: Sure. The reason I am keeping this is, if we remove this, then it wouldn't have the check that if the request is empty before proceeding. wdyt?

Anyway I removed it in [0cbd9fd](https://github.com/sigp/lighthouse/pull/7756/commits/0cbd9fd57ead7accaf76f4b33f1a7608323b7e8a)

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: Thanks, revised in [0cbd9fd](https://github.com/sigp/lighthouse/pull/7756/commits/0cbd9fd57ead7accaf76f4b33f1a7608323b7e8a)
I didn't notice this function is available, when I use the Electra one I was feeling unsure about it

beacon_node/network/src/network_beacon_processor/tests.rs
chong-he: Added  a test in [0cbd9fd](https://github.com/sigp/lighthouse/pull/7756/commits/0cbd9fd57ead7accaf76f4b33f1a7608323b7e8a)

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: Rename the variable as suggested

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: Revise the comment to be clearer, thanks!

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: Removed in [0cbd9fd](https://github.com/sigp/lighthouse/pull/7756/commits/0cbd9fd57ead7accaf76f4b33f1a7608323b7e8a)

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: All fulu-related logging removed in [0cbd9fd](https://github.com/sigp/lighthouse/pull/7756/commits/0cbd9fd57ead7accaf76f4b33f1a7608323b7e8a)

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: Only log once after the revision [0cbd9fd](https://github.com/sigp/lighthouse/pull/7756/commits/0cbd9fd57ead7accaf76f4b33f1a7608323b7e8a)

chong-he:
Thanks for the detailed review @jimmygchen. I have address all comments as above.

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
jimmygchen: Resolving this comment. `send_blob_count` is still used in logging.

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -293,21 +293,51 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+        let fulu_start_slot = self
+            .chain
+            .spec
+            .fulu_fork_epoch
+            .map(|epoch| epoch.start_slot(T::EthSpec::slots_per_epoch()));
+
         let mut blob_list_results = HashMap::new();
+        let mut slots_by_block_root = HashMap::new();
+        // For logging purpose, to display one log per block root
+        let mut logging_index_by_block_root = HashMap::new();
```
jimmygchen: Doesn't look like the values are used, so we could just make a HashSet of  block roots - avoid `mut` if possible, having many `mut` objects mean you could mutate them multiple times in multiple places, making the code harder to maintain and more prone to bugs:

```rust
let logging_index_by_block_root = request
            .blob_ids
            .iter()
            .map(|blob_id| blob_id.block_root)
            .collect::<HashSet<_>>();
```

jimmygchen: Just realised you don't even need th above hashset if you have `slots_by_block_root`, because you could get the block roots with `.keys()`.

You can apply a similar idea to simplify construction of `slots_by_block_root`, eliminating `mut`:

```rust
let slots_by_block_root: HashMap<Hash256, Slot> = request
            .blob_ids
            .iter()
            .flat_map(|blob_id| {
                let block_root = blob_id.block_root;
                self.chain
                    .data_availability_checker
                    .get_execution_valid_block(&block_root)
                    .or_else(|| self.chain.early_attester_cache.get_block(block_root))
                    .map(|block| (block_root, block.slot()));
            })
            .collect();
```

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -339,6 +368,10 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                                     Response::BlobsByRoot(Some(blob_sidecar.clone())),
                                 );
                                 send_blob_count += 1;
+                                logging_index_by_block_root
+                                    .entry(*root)
+                                    .or_insert(Vec::new())
+                                    .push(*index);
                                 break 'inner;
```
jimmygchen: Unrelated to your change, but thought it's worth noting here as I already noticed - if we can express the same logic with iterator chains, I'd prefer them over named nested for loop - because it's much clearer and side-effect free, also much easier to read as well. 

e.g. 

```rust
                        if let Some(blob_sidecar) =
                            blobs_sidecar_list.iter().find(|b| b.index == *index)
                        {
                            self.send_response(
                                peer_id,
                                inbound_request_id,
                                Response::BlobsByRoot(Some(blob_sidecar.clone())),
                            );
                            send_blob_count += 1;
                        }
```

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -354,10 +387,10 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                 }
             }
         }
+
         debug!(
             %peer_id,
-            %requested_root,
-            ?requested_indices,
+            block_root = ?logging_index_by_block_root.keys(),
```
jimmygchen:
```suggestion
            block_roots = ?slots_by_block_root.keys(),
```

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: Resolving this as Fulu loggings are now removed

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: Resolving this as Fulu loggings are now removed

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: Good point, revised in [20719ae](https://github.com/sigp/lighthouse/pull/7756/commits/20719ae944863505140c80179aa8ca3fdb64cf9a)

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: Revised in [20719ae](https://github.com/sigp/lighthouse/pull/7756/commits/20719ae944863505140c80179aa8ca3fdb64cf9a)

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
chong-he: I simplify the loop here: [85cf611](https://github.com/sigp/lighthouse/pull/7756/commits/85cf611d4f3bcddcb074a3e15d7d1d8522e4f297)

beacon_node/network/src/network_beacon_processor/tests.rs
```diff
@@ -1670,6 +1678,204 @@ async fn test_blobs_by_range() {
...
+                actual_count += 1;
+            } else {
+                break;
+            }
+        } else {
+            panic!("unexpected message {:?}", next);
+        }
+    }
+    // Post-Fulu should return 0 blobs
+    assert_eq!(0, actual_count);
```
jimmygchen: I think the `test_blobs_by_range` actually covers the same code path? perhaps we could just add the 0 assertion to the test if the fork at genesis is fulu ?

beacon_node/network/src/network_beacon_processor/tests.rs
jimmygchen: i think the behaviour of this test is good - can't really test if we're doing early return vs reading from db, given both are returning empty lists

jimmygchen: only easy way is to do db mocking, but i think it's not worth doing it now

jimmygchen (approved):
Great work! Thanks for the patience and thoroughness 🙏

chong-he:
> Great work! Thanks for the patience and thoroughness 🙏

Thanks Jimmy. I was waiting for the CI to pass to only reply, thanks for the thorough review :heart:

PR #7758, opened 2025-07-18
fix: allow BLOB_SCHEDULE: [] field

barnabasbusa:
Now empty blob schedules from config.yaml (like BLOB_SCHEDULE: []) will be preserved and included in the /eth/v1/config/spec beacon API endpoint response instead of being omitted.

## Issue Addressed

## Proposed Changes

## Additional Info

### DIFF
--- a/consensus/types/src/chain_spec.rs
+++ b/consensus/types/src/chain_spec.rs
@@ -1744,7 +1744,6 @@ pub struct Config {
     #[serde(with = "serde_utils::quoted_u64")]
     custody_requirement: u64,
     #[serde(default = "BlobSchedule::default")]
-    #[serde(skip_serializing_if = "BlobSchedule::is_empty")]
     blob_schedule: BlobSchedule,
     #[serde(default = "default_validator_custody_requirement")]
     #[serde(with = "serde_utils::quoted_u64")]
--- a/consensus/types/src/config_and_preset.rs
+++ b/consensus/types/src/config_and_preset.rs
@@ -106,8 +106,7 @@ pub fn get_extra_fields(spec: &ChainSpec) -> HashMap<String, Value> {
     let hex_string = |value: &[u8]| format!("0x{}", hex::encode(value)).into();
     let u32_hex = |v: u32| hex_string(&v.to_le_bytes());
     let u8_hex = |v: u8| hex_string(&v.to_le_bytes());
-    
-    let mut fields = hashmap! {
+    hashmap! {
         "bls_withdrawal_prefix".to_uppercase() => u8_hex(spec.bls_withdrawal_prefix_byte),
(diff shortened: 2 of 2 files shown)


barnabasbusa:
CI fails for kurtosis because the new array datatype requires assertoor to run `ethpandaops/assertoor:fulu-support` image. 

We are looking into merging fulu supported branch into our master next week.

pawanjay176 (approved):

pawanjay176:
Waiting for assertoor fulu support to get merged.

barnabasbusa:
~~@pawanjay176 I have added a condition to only set the blob schedule in the api if fulu is not far away epoch, so it should now pass without needing to have to update assertoor (unless you test fulu case with assertoor)~~

Assertoor latest release is out. Should have fixed this issue.

jimmygchen:
> for any client that does not yet return `BLOB_SCHEDULE` from `/eth/v1/config/spec`, I would suggest to only return it if `FULU_FORK_EPOCH != FAR_FUTURE_EPOCH` as otherwise you might run into interop issues with older client releases, at least I know that Vouch is not able to parse the array right now

from Nico (https://discord.com/channels/595666850260713488/1392770505387409449/1397145131387453512)

michaelsproul:
haha called it

michaelsproul:
@barnabasbusa Trying to understand whether we should merge this.

What is your goal in making `blob_schedule: []` appear in the response? AFAICT the blob_schedule is irrelevant if Fulu is not scheduled, so it would be conceptually OK for us to avoid serializing it in when the Fulu fork epoch is not set (as Nico suggests). _Unless_ that is exactly the case where you want to test for consistency between different client impls?

I think we have two options:

- Merge your PR as-is
  - **Pros**: simple, consistent with other CLs
  - **Cons**: probably breaks compatibility with current/old Vouch versions on pre-Fulu networks
- Merge my change from here https://github.com/michaelsproul/lighthouse/commit/ea316310974614b4c349a78ff27a2ed177dfb35a
  - **Pros**: maintains Vouch compatibility by not including Fulu params before Fulu, still shows `blob_schedule: []` on Fulu networks
  - **Cons**: more complex implementation, doesn't allow testing `blob_schedule: []` in the case where the Fulu fork epoch is not set.
  
Let me know if the second solution would be acceptable for your use case, as I think it's better to avoid breaking Vouch, and more consistent with how we usually handle future forks.

barnabasbusa:
I'm happy with more complex solution to only have blob schedule when fulu is scheduled.

michaelsproul:
Niiice, thanks

Closing in favour of:

- https://github.com/sigp/lighthouse/pull/7779

Sorry to deny you a nice little green contribution dot @barnabasbusa 😭

PR #7761, opened 2025-07-18
Create `network_utils` crate

dknopik:
## Issue Addressed

Anchor currently depends on `lighthouse_network` for a few types and utilities that live within. As we use our own libp2p behaviours, we actually do not use the core logic in that crate. This makes us transitively depend on a bunch of unneeded crates (even a whole separate libp2p if the versions mismatch!)

## Proposed Changes

Move things we require into it's own lightweight crate.

## Additional Info

CC @diegomrsantos, wdyt?


macladson:
Love this idea! Just needs a `cargo sort --workspace` and a `cargo fmt` to pass CI

AgeManning (approved):
This refactor seems good to me

Issue #7744, opened 2025-07-15
Address #7733 review comments: Sync peer attribution


jimmygchen:
There were some good review comments made by @dapplion after #7733 was merged here:
https://github.com/sigp/lighthouse/pull/7733#pullrequestreview-3014128649

The main changes that still need to be done
1. Apply retry logic to backfill
2. Evaluate whether `LowToleranceError` for missing columns - this seem to have caused quick disconnections on devnet
3. Add a counter for retries
4. Persist failed peers across retries

jimmygchen:
#7762 addresses 1 & 3 above.

I had a conversation with @pawanjay176 and agreed that `LowTolerance` make sense if peer claims they have a column but is not able to serve it. We were disconnecting peers very frequently on devnet-3 because peers were not great serving data, but it has improved now. Will keep observing.

Regarding persisting failed peers across retries, we think it gives us little gain as the high penalties mean repeated offended will get banned pretty quickly.

Happy to close this issue now.

PR #7762, opened 2025-07-18
Backfill peer attribution

pawanjay176:
## Issue Addressed

Partly addresses https://github.com/sigp/lighthouse/issues/7744

## Proposed Changes

Implement similar peer sync attribution like in #7733 for backfill sync.


dapplion:
Can you implement max retries too before adding this to backfill?

jimmygchen:
I've tested this running a supernode on `fusaka-devnet-2` and it's no longer OOMing during backfill. Memory usage is consistently around 8 - 10G, and active sync requests metric remain low.

Looking great @pawanjay176

jimmygchen:
I'll do a round of review shortly. 
@dapplion it would be helpful to have your review too 🙏

jimmygchen:
@pawanjay176 I changed the base `branch` to `unstable` because I've already merged this into `fusaka-devnet-3` for testing.

beacon_node/network/src/sync/network_context.rs
```diff
@@ -722,12 +725,29 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
-        if let Some(blocks_result) = entry.get_mut().responses(&self.chain.spec) {
-            if matches!(blocks_result, Err(CouplingError::PeerFailure { .. })) {
-                debug!(
-                    entry=?entry.key(),
-                    "Peer failed to return columns"
-                );
+        let range_req = entry.get_mut();
+        if let Some(blocks_result) = range_req.responses(&self.chain.spec) {
+            if let Err(CouplingError::PeerFailure {
+                action,
```
dapplion: Seems like action will only be penalized if it's the last retry. That feels wrong as if the peer did something bad any retry attempt should be penalized regardless of ordering

pawanjay176: The `ExceededMaxRetries` error also gets downscored, but yeah this logic looks pretty contrived, i'm still trying to iterate on it.

.github/workflows/local-testnet.yml
```diff
@@ -173,7 +173,8 @@ jobs:
   # Tests checkpoint syncing to a live network (current fork) and a running devnet (usually next scheduled fork)
   checkpoint-sync-test:
     name: checkpoint-sync-test-${{ matrix.network }}
-    runs-on: ubuntu-latest
+    # Use self-hosted runner for Fusaka devnet testing as GitHub hosted ones aren't able to keep up with the chain.
+    runs-on: ${{ github.repository == 'sigp/lighthouse' && matrix.network == 'devnet' && fromJson('["self-hosted", "linux", "Kurtosis", "large"]') || 'ubuntu-latest'  }}
```
jimmygchen: Created a separate PR to fix and test this. https://github.com/sigp/lighthouse/pull/7804

beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -334,6 +337,48 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
...
+                        for peer in peers.iter() {
+                            network.report_peer(
+                                *peer,
+                                *action,
+                                "failed to return columns, exceeded retry attempts",
+                            );
+                        }
+                    }
+                    CouplingError::InternalError(msg) => {
+                        debug!(?batch_id, msg, "Block components coupling internal error");
```
jimmygchen: should this be `warn` or `error`?

pawanjay176: user can't really do much about this no?

jimmygchen: Yes that's true, but it would be quite difficult for us spot a bug if this is set to debug.

beacon_node/network/src/sync/network_context.rs
```diff
@@ -55,6 +55,9 @@ use types::{
 pub mod custody;
 mod requests;
 
+/// Max retries for block components after which we fail the batch.
+pub const MAX_COLUMN_RETRIES: usize = 3;
```
jimmygchen, line 59: Would `block_sidecar_coupling` be a better place for this, since this const is only use in that module?

beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -898,6 +942,54 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
         Ok(())
     }
 
+    /// Retries partial column requests within the batch by creating new requests for the failed columns.
+    #[instrument(parent = None,
+        fields(service = "backfill_sync"),
+        name = "backfill_sync",
+        skip_all
+    )]
```
jimmygchen: Ah I don't think this needs to be a root span. I've actually covered sending `DataColumnByRange requests` in the other tracing PR. I think the retry will just show up in the parent range request span, so we can delete this one.

https://github.com/sigp/lighthouse/pull/7831/files#diff-2a2c481817feb9f76f2ff94088fd03e7217ff4d108065649b179fb1aa1c38b2aR485-R493

jimmygchen: Oh I realised i deleted the import while resolving conflicts, will just delete this one to fix the build

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -35,13 +37,20 @@ enum RangeBlockDataRequest<E: EthSpec> {
         /// The column indices corresponding to the request
         column_peers: HashMap<DataColumnsByRangeRequestId, Vec<ColumnIndex>>,
         expected_custody_columns: Vec<ColumnIndex>,
+        attempt: usize,
```
jimmygchen, line 52: nit: plural
```suggestion
        attempts: usize,
```

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -192,22 +205,30 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
...
+                        requests.retain(|&k, _| k.peer != *peer);
                     }
                 }
 
                 Some(resp)
             }
-        }
+        };
+
+        // Increment the attempt once this function returns the response or errors
```
jimmygchen, line 269: This comment doesn't seem applicable here anymore

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -229,26 +250,23 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
...
 
             let mut blobs_buffer = vec![None; max_blobs_per_block];
             for blob in blob_list {
                 let blob_index = blob.index as usize;
                 let Some(blob_opt) = blobs_buffer.get_mut(blob_index) else {
-                    return Err(CouplingError {
-                        msg: "Invalid blob index".to_string(),
-                        column_and_peer: None,
-                    });
+                    return Err(CouplingError::InternalError(
```
jimmygchen: Why is this not a peer failure? if the blob_index is >= max_blobs_per_block

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -229,26 +250,23 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
...
+                        "Invalid blob index".to_string(),
+                    ));
                 };
                 if blob_opt.is_some() {
-                    return Err(CouplingError {
-                        msg: "Repeat blob index".to_string(),
-                        column_and_peer: None,
-                    });
+                    return Err(CouplingError::InternalError(
+                        "Repeat blob index".to_string(),
```
jimmygchen, line 307: Is this the case where the peer send us multiple blobs of the same indices? this also seems like a potential peer failure?

beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -1005,6 +1097,36 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
...
+        epoch: Epoch,
+        network: &SyncNetworkContext<T>,
+    ) -> bool {
+        if network.chain.spec.is_peer_das_enabled_for_epoch(epoch) {
+            // Require peers on all sampling column subnets before sending batches
+            let peers_on_all_custody_subnets = network
+                .network_globals()
+                .sampling_subnets()
+                .iter()
+                .all(|subnet_id| {
```
jimmygchen, line 1122: Potential optimisation here - i think we could create a function on the peer_db to accept the sampling subnets, and check that there's at least one peer in each of them with short-circuit - instead of iterating through all sampling subnets and all peers (up to 128 subnets * 100 peers = 12800 min iterations)

Something like this in the `PeerDB`:

```rust
    pub fn has_good_range_sync_peers_in_custody_subnets(
        &self,
        subnets: HashSet<DataColumnSubnetId>,
    ) -> bool {
        let mut subnets_clone = subnets.clone();
        for info in self.peers.values() {
            info.custody_subnets_iter().for_each(|subnet| {
                subnets_clone.remove(&subnet);
            });
            if subnets_clone.is_empty() {
                return true;
            }
        }
        false
    }
```

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -1084,11 +1090,13 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
...
-                        .network_globals()
+                    let peer_db = network.network_globals().peers.read();
+                    let peer_count = self
                         .peers
-                        .read()
-                        .good_range_sync_custody_subnet_peer(*subnet_id, &self.peers)
+                        .iter()
+                        .filter(|peer| {
+                            peer_db.is_good_range_sync_custody_subnet_peer(*subnet_id, peer)
+                        })
```
jimmygchen, line 1102: Same as above re efficiency, but not critical for this PR

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -35,13 +37,20 @@ enum RangeBlockDataRequest<E: EthSpec> {
...
 }
 
 #[derive(Debug)]
-pub struct CouplingError {
-    pub(crate) msg: String,
-    pub(crate) column_and_peer: Option<(Vec<(ColumnIndex, PeerId)>, PeerAction)>,
+pub(crate) enum CouplingError {
+    InternalError(String),
+    /// The peer we requested the data from was faulty/malicious
+    PeerFailure {
```
jimmygchen: This is data column specific, should we rename to something like `DataColumnPeerFailure`

beacon_node/network/src/sync/network_context.rs
jimmygchen: This has been addressed in https://github.com/sigp/lighthouse/pull/7762/commits/166468af9f3460d67bd4aeda0df3b2f4767dec9c i think

beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -898,6 +942,49 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
...
 
+    /// Retries partial column requests within the batch by creating new requests for the failed columns.
+    pub fn retry_partial_batch(
+        &mut self,
+        network: &mut SyncNetworkContext<T>,
+        batch_id: BatchId,
+        id: Id,
+        failed_columns: HashSet<ColumnIndex>,
+        mut failed_peers: HashSet<PeerId>,
+    ) -> Result<(), BackFillError> {
```
jimmygchen, line 956: Doesn't look like we need to return a `Result` here, it's always `Ok(())`

beacon_node/network/src/sync/backfill_sync/mod.rs
jimmygchen: oh btw this function does the same thing as the one here right?

https://github.com/sigp/lighthouse/blob/b473dc81ab04e5e97c89d3c5c5e46a96da2c15a3/beacon_node/network/src/sync/range_sync/chain.rs#L1083

Maybe if we add the above to `peer_db`, we could reduce some code duplication?

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -300,82 +315,66 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
...
-
-                    // For now, we always assume that the block peer is right.
-                    // This is potentially dangerous as we can get isolated on a chain with a
-                    // malicious block peer.
-                    // TODO: fix this by checking the proposer signature before downloading columns.
                     let responsible_peers = column_to_peer.iter().map(|c| (*c.0, *c.1)).collect();
-                    return Err(CouplingError {
-                        msg: format!("No columns for block {block_root:?} with data"),
-                        column_and_peer: Some((responsible_peers, PeerAction::LowToleranceError)),
+                    return Err(CouplingError::PeerFailure {
```
jimmygchen: I wonder if it makes sense to accumulate the errors and `continue` the for loop instead?

beacon_node/network/src/sync/block_sidecar_coupling.rs
pawanjay176: ohh nice catch! i just labeled all non data column errors as internal errors. will fix these

beacon_node/network/src/sync/block_sidecar_coupling.rs
pawanjay176: I tried to keep changes minimal here. Besides, if we just collect the error, we can't use `data_columns_by_index` below so we'll have to fail there. Its a bit inefficient, but it works

beacon_node/network/src/sync/block_sidecar_coupling.rs
jimmygchen: I think we're already downscoring the peer for the failed block & blob batch right, so this should have no impact on existing functionality, we just need a separate enum variant for clarity right?

pawanjay176: Yep, blob peer failure or something

beacon_node/network/src/sync/backfill_sync/mod.rs
pawanjay176: Ohh this is very neat. I'll add it in

beacon_node/network/src/sync/block_sidecar_coupling.rs
jimmygchen: I think we could still `continue` to pair the next block, but i guess if the peers' response is missing one block, we'd have to retry fetching columns for all blocks in the range anyways, so it's likely not very useful to accumulate the errors.

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -300,82 +315,66 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
...
                     });
                 };
 
                 let mut custody_columns = vec![];
                 let mut naughty_peers = vec![];
                 for index in expects_custody_columns {
                     if let Some(data_column) = data_columns_by_index.remove(index) {
-                        // Safe to convert to `CustodyDataColumn`: we have asserted that the index of
-                        // this column is in the set of `expects_custody_columns` and with the expected
-                        // block root, so for the expected epoch of this batch.
```
dapplion, line 342: Can you keep this comment?

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -179,6 +191,7 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
                     data_columns.extend(data.clone())
                 }
 
+                *attempt += 1;
```
jimmygchen, line 231: it seems a bit weird incrementing here at first, but it makes sense to me now -  this line is only reached when all requests are in the `Complete` state. might be worth adding a comment here to explain. (because moving this to a few lines earlier would break this, but it isnt immediately obvious to me.

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -300,82 +315,66 @@ impl<E: EthSpec> RangeBlockComponentsRequest<E> {
                 .insert(index, column)
                 .is_some()
             {
-                return Err(CouplingError {
-                    msg: format!("Repeated column block_root {block_root:?} index {index}"),
-                    column_and_peer: None,
-                });
+                return Err(CouplingError::InternalError(format!(
```
dapplion: Add a comment here saying that `DataColumnsByRangeRequestItems` ensures there are no duplicated indices and that peers return the requested indices, which we never request to more than one peer. Thus an internal error

beacon_node/network/src/sync/network_context.rs
```diff
@@ -715,10 +714,27 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+        let range_req = entry.get_mut();
+        if let Some(blocks_result) = range_req.responses(&self.chain.spec) {
+            if let Err(CouplingError::PeerFailure {
+                action: _,
+                error,
+                faulty_peers: _,
+                exceeded_retries,
+            }) = &blocks_result
+            {
+                // Remove the entry if its a peer failure **and** retry counter is exceeded
```
dapplion:
```suggestion
                // Remove the entry if it's a peer failure **and** retry counter is exceeded
```

beacon_node/network/src/sync/backfill_sync/mod.rs
pawanjay176: meant to return an error here if the batch doesn't exist. Should never happen and it shouldn't fail silently if it does

beacon_node/network/src/sync/backfill_sync/mod.rs
pawanjay176: I'll actually try to resolve this in a separate PR. needs a bit more work

beacon_node/network/src/sync/block_sidecar_coupling.rs
pawanjay176: done, dunno why i deleted it in the first place lol

beacon_node/network/src/sync/block_sidecar_coupling.rs
pawanjay176: I removed the error here because we can still continue with duplicates, added a comment like you said though

beacon_node/network/src/sync/block_sidecar_coupling.rs
pawanjay176: added a comment

pawanjay176:
@jimmygchen looks like the supernode backfill is failing because the CI machine is overwhelmed with reconstruction work. Maybe we can move the supernode task to a beefier machine after merging this PR?

beacon_node/network/src/sync/backfill_sync/mod.rs
jimmygchen: Created #7855

jimmygchen: @pawanjay176 I've assigned the issue to you since it seems like you've attempted it, but feel free to unassign if you are happy for others to pick it up instead.

jimmygchen (approved):
Thanks for addressing the comments @pawanjay176, looks great to me now! 🎉

PR #7763, opened 2025-07-19
fix: add exit on missing yq to prevent unclear script failures

Galoretka:
Added an explicit exit 1 after checking for the presence of the yq utility in start_local_testnet.sh.
Previously, if yq was not installed, the script would continue and fail later with a less clear error message when attempting to use yq.
Now, if yq is missing, the script will immediately exit with a clear message to the user.
This change makes the script’s behavior more predictable and helps users quickly identify missing dependencies.


michaelsproul:
Suspected bot PR, and you haven't signed the CLA.

PR #7764, opened 2025-07-21
Add heaptrack support

michaelsproul:
## Proposed Changes

Although we're working on jemalloc profiler support in https://github.com/sigp/lighthouse/pull/7746, heaptrack seems to be producing more sensible results.

This PR adds a heaptrack profile and a heaptrack feature so that we no longer need to patch the code in order to use heaptrack. This may prove complementary to jemalloc profiling, so I think there is no harm in having both.

## Additional Info

Usage:

```
FEATURES=heaptrack PROFILE=release-debug make
```

Then `heaptrack lighthouse ...` using the `lighthouse` binary that was just built.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -289,7 +289,7 @@ panic = "abort"
 codegen-units = 1
 overflow-checks = true
 
-[profile.heaptrack]
+[profile.release-debug]
 inherits = "release"
 debug = true
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -289,5 +289,9 @@ panic = "abort"
 codegen-units = 1
 overflow-checks = true
 
+[profile.heaptrack]
+inherits = "release"
+debug = true
+
 [patch.crates-io]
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -289,5 +289,9 @@ panic = "abort"
 codegen-units = 1
 overflow-checks = true
 
+[profile.heaptrack]
```
michaelsproul: We could call this profile something else if we want to reuse it with jemalloc?

Maybe `release-debug`?

jimmygchen: Is the overhead for having heaptrack high? 
I think having a mode that runs release with just debug symbols without anything else, and having the option to add heaptrack + jemalloc profiling would be useful - so maybe having them as feature would be good enough?

jimmygchen: `release-debug` sounds good to me!

michaelsproul: There's virtually no overhead for compiling a binary with `release-debug` and `features=heaptrack`, _except_ that we lose out on the efficiencies of jemalloc for tree-states. Running under heaptrack is a separate action, which does add quite a bit of overhead in terms of CPU

jimmygchen (approved):
LGTM

Issue #7749, opened 2025-07-17
Update LH to use Rust Edition 2024


michaelsproul:
Rust has _editions_ as well as versions. The 2024 edition has been out for a while, and brings some nice features like [`let`-chains](https://github.com/rust-lang/rust/pull/132833).

When I updated Milhouse to use the 2024 edition there were a few lifetime patterns that needed to be changed to bring us in line. For Lighthouse we can roll out the edition update one crate at a time, or in batches of crates. To avoid bitrot we probably shouldn't do the whole repo in one go.

This will probably be a backwards-incompat change as it will require us to bump the MSRV to at least 1.88.

michaelsproul:
@chong-he is going to take this :pray:

PR #7766, opened 2025-07-21
Update Rust Edition to 2024

chong-he:
## Issue Addressed

* #7749 

Thanks @dknopik and @michaelsproul for your help!


crypto/bls/src/impls/blst.rs
```diff
@@ -43,7 +42,7 @@ pub fn verify_signature_sets<'a>(
         return false;
     }
 
-    let rng = &mut rand::thread_rng();
+    let rng = &mut rand::rng();
```
chong-he, line 45: `thread_rng()` is deprecated in `rand 0.9.0` and above, replaced with `rand::rng()`:
https://docs.rs/rand/latest/rand/fn.thread_rng.html

crypto/bls/src/impls/blst.rs
```diff
@@ -55,7 +54,7 @@ pub fn verify_signature_sets<'a>(
         let mut vals = [0u64; 4];
         while vals[0] == 0 {
             // Do not use zero
-            vals[0] = rng.gen();
+            vals[0] = rng.random();
```
chong-he, line 57: All `gen()` methods renamed to `random()` in `rand 0.9.0` and above: https://docs.rs/rand/latest/rand/trait.Rng.html#method.gen
to avoid collision with the newly added reserved keyword `gen` in Rust Edition 2024: https://doc.rust-lang.org/edition-guide/rust-2024/gen-keyword.html

validator_client/http_api/src/api_secret.rs
```diff
@@ -1,6 +1,6 @@
 use filesystem::create_with_600_perms;
-use rand::distributions::Alphanumeric;
-use rand::{thread_rng, Rng};
+use rand::distr::Alphanumeric;
```
chong-he, line 2: Module `distributions` in [rand 0.8.5](https://docs.rs/rand/0.8.5/rand/distributions/index.html) renamed to `distr` in [rand 0.9.x](https://docs.rs/rand/0.9.1/rand/distr/index.html)

beacon_node/client/src/builder.rs
```diff
@@ -207,7 +207,8 @@ where
             .import_all_data_columns(config.network.subscribe_all_data_column_subnets)
             .validator_monitor_config(config.validator_monitor.clone())
             .rng(Box::new(
-                StdRng::from_rng(OsRng).map_err(|e| format!("Failed to create RNG: {:?}", e))?,
+                StdRng::try_from_rng(&mut OsRng)
```
chong-he, line 211: `OsRng` is giving error when compiling, consulted @dknopik and he suggested to use `try_from_rng`

beacon_node/network/Cargo.toml
```diff
@@ -55,5 +55,7 @@ gossipsub = { workspace = true }
 k256 = "0.13.4"
 kzg = { workspace = true }
 matches = "0.1.8"
-rand_chacha = "0.3.1"
+rand_08 = { package = "rand", version = "0.8.0" }
```
chong-he: To preserve an older version of `rand` crate where necessary, a double dependency approach is used here, thanks @dknopik

validator_client/http_api/src/tests/keystores.rs
```diff
@@ -1324,13 +1328,13 @@ async fn delete_concurrent_with_signing() {
         let all_pubkeys = all_pubkeys.clone();
 
         let handle = handle.spawn(async move {
-            let mut rng = SmallRng::from_entropy();
+            let mut rng: StdRng = SeedableRng::from_os_rng();
```
chong-he, line 1331: It appears that `from_entropy` in [rand 0.8.5](https://docs.rs/rand/0.8.5/rand/trait.SeedableRng.html#method.from_entropy) is changed to `from_os_rng` in [rand 0.9.x](https://docs.rs/rand/0.9.2/rand/trait.SeedableRng.html#method.from_os_rng), so updated here

chong-he (commented):
Some comments for the reviewers

beacon_node/beacon_chain/src/test_utils.rs
```diff
@@ -713,7 +713,7 @@ where
     pub fn set_mock_builder(
         &mut self,
         beacon_url: SensitiveUrl,
-    ) -> impl futures::Future<Output = ()> {
+    ) -> impl futures::Future<Output = ()> + use<E, Hot, Cold> {
```
macladson: Can you explain the need for this?

beacon_node/network/Cargo.toml
```diff
@@ -55,5 +55,7 @@ gossipsub = { workspace = true }
 k256 = "0.13.4"
 kzg = { workspace = true }
 matches = "0.1.8"
-rand_chacha = "0.3.1"
+rand_08 = { package = "rand", version = "0.8.5" }
+rand_chacha = "0.9.0"
+rand_chacha_08 = { package = "rand_chacha", version = "0.3.1" }
```
macladson: Should this be `rand_chacha_03`?

beacon_node/network/src/sync/manager.rs
```diff
@@ -38,8 +38,9 @@ use super::block_lookups::BlockLookups;
 use super::network_context::{
     CustodyByRootResult, RangeBlockComponent, RangeRequestId, RpcEvent, SyncNetworkContext,
 };
-use super::peer_sync_info::{remote_sync_type, PeerSyncType};
-use super::range_sync::{RangeSync, RangeSyncType, EPOCHS_PER_BATCH};
+
```
macladson: Accidental new line

```suggestion
```

beacon_node/network/src/sync/tests/lookups.rs
```diff
@@ -114,6 +115,7 @@ impl TestRig {
             network_rx,
             network_rx_queue: vec![],
             sync_rx,
+            rng_08,
```
macladson, line 118: Why do we need a 2nd rng here?

lighthouse/src/main.rs
```diff
@@ -99,7 +99,9 @@ fn build_profile_name() -> String {
 fn main() {
     // Enable backtraces unless a RUST_BACKTRACE value has already been explicitly provided.
     if std::env::var("RUST_BACKTRACE").is_err() {
-        std::env::set_var("RUST_BACKTRACE", "1");
+        unsafe {
+            std::env::set_var("RUST_BACKTRACE", "1");
+        }
```
macladson, line 109: We probably shouldn't add `unsafe` into the codebase without being extremely sure there are no other alternatives

macladson (changes requested):
I have looked through every file and it looks pretty good to me, just had one nit and a couple of questions

beacon_node/network/Cargo.toml
dknopik: When I suggested this I chose `_08` to indicate that this is the version compatible with rand `0.8`. But I was on the fence between `08` and `03`.

beacon_node/network/src/sync/tests/lookups.rs
dknopik: Because the test rig needs to pass the `rng_08` to a dependency we do not control which is still on rand 0.8. I think it is `secp256k1`

beacon_node/beacon_chain/src/test_utils.rs
michaelsproul: I think this is forcing the opaque type returned (the Future) not to capture the lifetime of `self`

https://doc.rust-lang.org/edition-guide/rust-2024/rpit-lifetime-capture.html

michaelsproul: We might be able to use `'static` instead, which is a bit clearer?

lighthouse/src/main.rs
michaelsproul: Hmm, I don't think we have another option at the moment:

- https://github.com/rust-lang/rust/issues/93346

beacon_node/beacon_chain/src/test_utils.rs
chong-he: To be honest, I was following the compiler's suggestion on this. 

Changed to `+ 'static` as suggested: [e9eba45](https://github.com/sigp/lighthouse/pull/7766/commits/e9eba45e8d098bfe0b1261db4086cabdeafb4e08) (and removed the fmt changed in [a7d5db1](https://github.com/sigp/lighthouse/pull/7766/commits/a7d5db1cfb75867f41aac8a222e6720251f4f96d))

Thanks!

beacon_node/network/Cargo.toml
chong-he: Change to `rand_chach_03` in [c144bd6](https://github.com/sigp/lighthouse/pull/7766/commits/c144bd6ad986d6785fecd193421f8d0422479c7a)

lighthouse/src/main.rs
chong-he: Yes, from the Rust Edition 2024, `std::env::set_var` is a new unsafe function: https://doc.rust-lang.org/edition-guide/rust-2024/newly-unsafe-functions.html

From the doc, we now have to the `unsafe {}` block explicitly

dknopik: Luckily, the invariant (no concurrent access) is pretty easy to assert here as we are at the very top of main and there are no other threads (unless someone is doing some `LD_PRELOAD` shenanigans or calling `main` from their crate or whatnot). Note also that the function is already unsafe to use, it was just made explicit with the edition. For Anchor I added a comment and yolo'd it.

beacon_node/network/Cargo.toml
macladson: Ah I see your point, I think that's a valid way of doing it. I'm fine with either

lighthouse/src/main.rs
michaelsproul: Yeah I think a comment would be ok.

macladson, line 109: Just ~~taking~~ _borrowing_ the comment from Anchor

```suggestion
        // `set_var` is marked unsafe because it is unsafe to use if there are multiple threads
        // reading or writing from the environment. We are at the very beginning of execution and
        // have not spun up any threads or the tokio runtime, so it is safe to use.
        unsafe {
            std::env::set_var("RUST_BACKTRACE", "1");
        }
```

chong-he:
It looks like Rust 1.89 is in effect and is already linting the chain `if` / `if let` loop: https://github.com/sigp/lighthouse/actions/runs/16826607902/job/47664447929

I revised all in this single commit: [8665e63](https://github.com/sigp/lighthouse/pull/7766/commits/8665e630c079e9e788a59465899f35d78240ca03), unfortunately this makes the diff and files changed increased by a lot

michaelsproul:
I think we should merge this ASAP and include it in `v8.0.0-rc.0`, so that it doesn't come into conflict with any other refactors.

@macladson you probably have the best ability to review quickly as you can review just the diff since your last review.

@chong-he there is a single conflict now which would be good to fix so we can merge immediately once Mac gives the :+1:

macladson (approved):
Ok I've gone through all the new changes and it all looks good to me :+1:
Nice work @chong-he!

Some of the new `&&` chaining reads a bit awkwardly in some places but I think overall the 2024 changes are an upgrade

chong-he:
> Ok I've gone through all the new changes and it all looks good to me 👍 Nice work @chong-he!
> 
> Some of the new `&&` chaining reads a bit awkwardly in some places but I think overall the 2024 changes are an upgrade

Thank you so much for your detailed review!

michaelsproul:
Woohoo! Thanks Mac for the review and CK for the impl! :heart:

PR #7767, opened 2025-07-21
Fix leak in range sync `components_by_range_requests`

jimmygchen:
## Issue Addressed

When a data column range sync fails with coupling error returned below, if the coupling error does not contain a column and peer (i.e. `column_and_peer == None`)

https://github.com/sigp/lighthouse/blob/9181ff43f623481c7c4d154a4b6dcb0f833ef75c/beacon_node/network/src/sync/block_sidecar_coupling.rs#L286-L292

The entire batch gets retried, without the previous `context.components_by_range_requests` entry being removed, this result in unbounded memory usage increase as shown in the metric:

image

Note: this PR is not tested and I'm not sure if the fix is good - but just wanted to illustrate the issue.

### DIFF
--- a/beacon_node/network/src/sync/range_sync/chain.rs
+++ b/beacon_node/network/src/sync/range_sync/chain.rs
@@ -885,6 +885,8 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
                     failing_batch: batch_id,
                 });
             }
+            // FIXME: hack to remove range request before a retry
+            network.remove_range_request_by_id(request_id);
             self.send_batch(network, batch_id)
         } else {
             debug!(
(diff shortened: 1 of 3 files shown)


beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -885,6 +885,8 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
                     failing_batch: batch_id,
                 });
             }
+            // FIXME: hack to remove range request before a retry
+            network.remove_range_request_by_id(request_id);
```
jimmygchen, line 889: Note that this hack didn't fix it

image

The metric i was looking at is `sync_active_network_requests`, type == `components_by_range`, recorded here:
https://github.com/sigp/lighthouse/blob/790c94c75a940dbdbf8dd5d354519c160719d780/beacon_node/network/src/sync/network_context.rs#L1705-L1708

This change what caused entries not to be removed i believe, its used by both range and backfill:

https://github.com/sigp/lighthouse/blob/790c94c75a940dbdbf8dd5d354519c160719d780/beacon_node/network/src/sync/network_context.rs#L745-L751

jimmygchen:
This will be fixed properly by @pawanjay176 in #7762.

PR #7770, opened 2025-07-22
Fix malloc_utils features (sysmalloc)

michaelsproul:
## Proposed Changes

Follow-up to:

- https://github.com/sigp/lighthouse/pull/7764

The `heaptrack` feature added in my previous PR was ineffective, because the jemalloc feature was turned on by the Linux target-specific dependency. 

This PR tweaks the features such that:

- The jemalloc feature is just used to control whether jemalloc is compiled in. It is enabled on Linux by the target-specific dependency (see `lighthouse/Cargo.toml`), and completely disabled on Windows.
- If the `sysmalloc` feature is set on Linux then it overrides jemalloc when selecting an allocator, _even if_ the jemalloc feature is enabled (and the jemalloc dep was compiled).


lcli/Cargo.toml
```diff
@@ -45,3 +43,9 @@ tracing-subscriber = { workspace = true }
 tree_hash = { workspace = true }
 types = { workspace = true }
 validator_dir = { workspace = true }
+
+[target.'cfg(not(target_os = "windows"))'.dependencies]
+malloc_utils = { workspace = true, features = ["jemalloc"] }
+
+[target.'cfg(target_os = "windows")'.dependencies]
+malloc_utils = { workspace = true, features = [] }
```
michaelsproul, line 51: These changes ensure that `lcli` actually uses jemalloc on Linux, which is potentially relevant for the benchmarking subcommands.

jimmygchen:
Added `backward-incompat` as `jemalloc` feature has been removed from the top level.

jimmygchen (approved):
The conditional compilation logic looks quite complex, but I think you've correctly covered the scenarios I can think of.

Looks good to me!

PR #7771, opened 2025-07-22
Remove KZG verification on blobs fetched from the EL

jimmygchen:
## Issue Addressed

Continuation of #7713, addresses comment about skipping KZG verification on EL fetched blobs:

https://github.com/sigp/lighthouse/pull/7713#discussion_r2198542501

### DIFF
--- a/beacon_node/beacon_chain/src/fetch_blobs/mod.rs
+++ b/beacon_node/beacon_chain/src/fetch_blobs/mod.rs
@@ -12,14 +12,14 @@ mod fetch_blobs_beacon_adapter;
 #[cfg(test)]
 mod tests;
 
-use crate::blob_verification::{GossipBlobError, GossipVerifiedBlob};
+use crate::blob_verification::{GossipBlobError, KzgVerifiedBlob};
 use crate::block_verification_types::AsBlock;
 use crate::data_column_verification::{KzgVerifiedCustodyDataColumn, KzgVerifiedDataColumn};
 #[cfg_attr(test, double)]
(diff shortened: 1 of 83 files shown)


beacon_node/beacon_chain/src/fetch_blobs/mod.rs
```diff
@@ -408,37 +409,28 @@ fn build_blob_sidecars<E: EthSpec>(
...
-            Err(e) => {
-                return Err(FetchEngineBlobError::BlobSidecarError(e));
-            }
-        }
+        )
+        .map_err(FetchEngineBlobError::BlobSidecarError)?;
+
+        sidecars.push(KzgVerifiedBlob::from_execution_verified(
+            Arc::new(blob_sidecar),
+            timestamp_now(),
```
michaelsproul, line 431: I would kind of like to stamp out usage of `timestamp_now` because it bypasses the slot clock and introduces non-determinism in tests.

However I notice that we don't actually have the slot clock here, so happy to leave this as-is for now. Adding a dep on the slot clock just for this is probably not worth it.

michaelsproul (approved):
LGTM

I just pushed one commit to remove an `Arc` clone haha.

beacon_node/beacon_chain/src/fetch_blobs/mod.rs
jimmygchen: Yeah agree, It also uses the `validator_monitor` mod which is probably shouldn't 😅

michaelsproul: Let's merge for now, can stamp out `timestamp_now` later :sweat_smile:

PR #7777, opened 2025-07-23
Simplify ConfigAndPreset

michaelsproul:
## Proposed Changes

I noticed that we are serving preset values for Fulu on mainnet nodes prior to the fork. This has already gone live in v7.1.0, but should hopefully be handled in a graceful way by API consumers.

This PR _reverts_ the serving of Fulu data prior to Fulu, by serving Fulu data only if Fulu is scheduled.


consensus/types/src/config_and_preset.rs
```diff
@@ -72,7 +74,7 @@ impl ConfigAndPreset {
                 extra_fields,
             })
         } else if spec.electra_fork_epoch.is_some()
-            || fork_name.is_none()
```
michaelsproul, line 79: This was unreachable before, as it had been copy-pasta'd to the Fulu case. Using `latest_stable_fork` prevents this kind of issue.

beacon_node/http_api/tests/tests.rs
```diff
@@ -2616,12 +2616,18 @@ impl ApiTester {
     }
 
     pub async fn test_get_config_spec(self) -> Self {
-        let result = self
-            .client
-            .get_config_spec::<ConfigAndPresetFulu>()
-            .await
-            .map(|res| ConfigAndPreset::Fulu(res.data))
-            .unwrap();
+        let result = if self.chain.spec.fulu_fork_epoch.is_some() {
```
jimmygchen: I think this would always return true today, as we have `FULU_FORK_EPOCH` set to `FAR_FUTURE_EPOCH` on the mainnet spec

```suggestion
        let result = if self.chain.spec.is_fulu_scheduled() {
```

consensus/types/src/config_and_preset.rs
```diff
@@ -43,7 +43,6 @@ pub struct ConfigAndPreset {
 }
 
 impl ConfigAndPreset {
-    // DEPRECATED: the `fork_name` argument is never used, we should remove it.
     pub fn from_chain_spec<E: EthSpec>(spec: &ChainSpec, fork_name: Option<ForkName>) -> Self {
```
jimmygchen: Should we remove `fork_name` here given the comment above?

michaelsproul: No, I think the comment is wrong

beacon_node/http_api/tests/tests.rs
michaelsproul: Huh, I thought far future got converted to None? I'm afk but otherwise what is the point of the option?

consensus/types/src/config_and_preset.rs
```diff
@@ -53,8 +52,11 @@ impl ConfigAndPreset {
         let deneb_preset = DenebPreset::from_chain_spec::<E>(spec);
         let extra_fields = get_extra_fields(spec);
 
+        // The latest stable fork is used if the `fork_name` is not specified.
+        let latest_stable_fork = ForkName::latest_stable();
```
jimmygchen: I think this function was originally intended for testing purposes and I'm not sure if we want the timing of changing this to impact the behaviour of the client, unless we re-purpose it.

Could we use `spec. is_fulu_scheduled` instead, and else fallback to Electra? and we could probably remove the deneb `else` branch below?

validator_client/http_api/src/tests.rs
```diff
@@ -206,13 +206,20 @@ impl ApiTester {
...
     pub async fn test_get_lighthouse_spec(self) -> Self {
-        let result = self
-            .client
-            .get_lighthouse_spec::<ConfigAndPresetFulu>()
-            .await
-            .map(|res| ConfigAndPreset::Fulu(res.data))
-            .unwrap();
-        let expected = ConfigAndPreset::from_chain_spec::<E>(&E::default_spec(), None);
+        let spec = E::default_spec();
+        let result = if spec.fulu_fork_epoch.is_some() {
```
jimmygchen: Same as above, might want to consider `spec.is_fulu_scheduled`?
One of these branches is going to be `unreachable` right? (because we're using `default_spec` in this test)
Also noting the http server has its own `spec`, maybe we want to keep a reference of the same `spec` in the `ApiTester`? 

https://github.com/sigp/lighthouse/blob/e34a9a0c65d5f8287c4e2e8805ee4ea7b36ac3a9/validator_client/http_api/src/tests.rs#L111-L133

consensus/types/src/config_and_preset.rs
jimmygchen: I don't see `fork_name` getting passed from anywhere though, is it used by something other than lighthouse or am i missing something?

beacon_node/http_api/tests/tests.rs
michaelsproul: Yeah it's the same thing, see `deserialize_fork_epoch`.

However I don't mind using the safer `is_fulu_scheduled` method.

jimmygchen (approved):
Looks good, thanks!

PR #7779, opened 2025-07-23
Skip serializing blob_schedule before Fulu

michaelsproul:
## Issue Addressed

Alternative to:

- https://github.com/sigp/lighthouse/pull/7758

## Proposed Changes

Serve the `blob_schedule` field on `/eth/v1/config/spec` _only_ when Fulu is enabled. If the blob schedule is empty, we will still serve it as `[]`, so long as Fulu is enabled.

## Additional Info

This touches the same code as:

- https://github.com/sigp/lighthouse/pull/7777

But this PR is higher prio for Fusaka testing.


pawanjay176 (approved):
LGTM, fixed the test and merged unstable

pawanjay176:
Actually this is quite annoying to deal with in the tests because a roundtrip ser/de isn't consistent. i.e. this test would fail
```rust
#[test]
    fn chain_spec_roundtrip() {
        let yamlconfig =
            ConfigAndPreset::from_chain_spec::<MainnetEthSpec>(&ChainSpec::mainnet(), None);

        let serialized = serde_yaml::to_string(&yamlconfig).unwrap();
        let deserialized: ConfigAndPreset = serde_yaml::from_str(&serialized).unwrap();
        assert_eq!(yamlconfig.config().blob_schedule, deserialized.config().blob_schedule);
    }
```
Not sure of a clean way to handle this other than setting fulu_fork_epoch all places where we do this kind of roundtrip.

michaelsproul:
Why doesn't it round trip? Is it because of the serialisation field? Maybe we should ignore that in the PartialEq impl of BlobSchedule

jimmygchen:
> Why doesn't it round trip? Is it because of the serialisation field? Maybe we should ignore that in the PartialEq impl of BlobSchedule

Yeah that's right
When we deserialise (`result` below), the `skip_serializing` value is set to `false`  (default), 
but when we create it from `ChainSpec` (`expected`), the `skip_serializing` value is set to true because Fulu is unset.

https://github.com/sigp/lighthouse/blob/257d2707182c2c1d073c53811b38e9944e07d932/validator_client/http_api/src/test_utils.rs#L257-L265

Yeah i think ignoring that field in `PartialEq` sounds good. Will push a commit now.

pawanjay176 (approved):
Ohh ignoring it in the partialeq impl is very clean, nice!

PR #7783, opened 2025-07-23
Expose functions to do preliminary slashing checks

dknopik:
Thiiiis one is a bit spicy:

Currently, the slashing database only exposes functions to atomically check slashability of an attestation/block and then insert the corresponding data into the database. As this is done atomically, we can fearlessly sign the thing as other threads will definitely see the inserted data.

In Anchor, we first decide with the SSV committee which attestation/block we should sign, and then create a partial signature and broadcast it. Until now, we did the slashing check-and-insert right before creating the partial signature, and will continue to do so. However, we want to be able to check whether a thing is slashable in the first step, in order to not decide on a invalid value and then sign nothing, as the the slashing check fails.

Basically, we want to move from:

```mermaid
flowchart LR
    A@{ shape: sm-circ, label: "Small start" } --> Z[Wait for block]
    Z --> B[Decide block]
    B --> C{Is safe?}
    C -->|Yes| D[Create partial sig]
    C -->|No| E[Sign nothing]
```
to 
```mermaid
flowchart LR
    A@{ shape: sm-circ, label: "Small start" } --> Z[Wait for block]
    Z --> C{"Is safe?<br>(check only)"}
    C -->|Yes| B[Decide block]
    C -->|No| Z
    B --> D{"Is safe?<br>(check and insert)"}
    D -->|Yes| H[Create partial sig]
    D -->|No| I[Sign nothing]
```

For this, we need to expose check only functions in the slashing database. This introduces a way for the user of the slashing database to shoot themselves in the foot.


AgeManning (commented):
I support these changes. 

@michaelsproul suggested a good modification. 

We probably should disallow these functions in lighthouse by doing something like this:
https://github.com/sigp/lighthouse/blob/stable/consensus/types/clippy.toml

and this:
https://github.com/sigp/lighthouse/blob/stable/consensus/types/src/lib.rs#L3-L10

.gitignore
```diff
@@ -9,7 +9,6 @@ perf.data*
 *.tar.gz
 /bin
 genesis.ssz
-/clippy.toml
```
dknopik, line 12: Not sure why this was here

validator_client/slashing_protection/src/slashing_database.rs
```diff
@@ -611,6 +611,7 @@ impl SlashingDatabase {
         block_header: &BeaconBlockHeader,
         domain: Hash256,
     ) -> Result<Safe, NotSafe> {
+        #[allow(clippy::disallowed_methods)]
```
dknopik, line 614: needed, as even disallowed methods may not call disallowed methods

dknopik (commented):
@michaelsproul good idea, done :)

michaelsproul:
CI should pass now, we've bumped the HTTP test timeout on unstable

michaelsproul (approved):
LGTM, thanks!

michaelsproul:
@mergify requeue

PR #7743, opened 2025-07-14
Feature gate arbitrary crate in the consensus types crate

ec2:
## Issue Addressed

Which issue # does this PR address?

## Proposed Changes

Puts the `arbitrary` crate behind a feature flag in the `types` crate. 

## Additional Info

Please provide any additional information. For example, future considerations
or information useful for reviewers.

PR #7787, opened 2025-07-24
Bump Rust version to 1.88

jimmygchen:
## Issue Addressed

In #7743, rust version was bumped:
- msrv to 1.87
- `Dockerfile` to 1.88

We also need to bump the other docker images as well, and might as well keep them all consistent at 1.88.

### DIFF
--- a/lighthouse/Cargo.toml
+++ b/lighthouse/Cargo.toml
@@ -4,7 +4,7 @@ version = "7.1.0"
 authors = ["Sigma Prime <contact@sigmaprime.io>"]
 edition = { workspace = true }
 autotests = false
-rust-version = "1.87.0"
+rust-version = "1.88.0"
 
 # Prevent cargo-udeps from flagging the dummy package `target_check`, which exists only
 # to assert properties of the compilation target.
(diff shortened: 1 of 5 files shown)


jimmygchen:
Updating self-hosted runners to 1.88, will re-run CI once that's done.

lighthouse/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "7.1.0"
 authors = ["Sigma Prime <contact@sigmaprime.io>"]
 edition = { workspace = true }
 autotests = false
-rust-version = "1.87.0"
+rust-version = "1.88.0"
```
michaelsproul, line 7: We don't technically need to bump the MSRV, right?

michaelsproul (approved):
LGTM

lighthouse/Cargo.toml
jimmygchen: We don't, but the previous PR bumped the dockerfile to 1.88, and we usually bump all the versions together - a bit easier to maintain IMO. 
OR we could keep everything at 1.87, what do you think?

PR #7792, opened 2025-07-25
Fix wrong columns getting processed on a CGC change

jimmygchen:
## Issue Addressed

This PR fixes a bug where wrong columns could get processed immediately after a CGC increase.

Scenario:
- The node's CGC increased due to additional validators attached to it (lets say from 10 to 11)
- The new CGC is advertised and new subnets are subscribed immediately, however the change won't be effective in the data availability check until the next epoch (See [this](https://github.com/sigp/lighthouse/blob/ab0e8870b4a35872574f5e33e9d13788c02ce250/beacon_node/beacon_chain/src/validator_custody.rs#L93-L99)). Data availability checker still only require 10 columns for the current epoch.
- During this time, data columns for the additional custody column (lets say column 11) may arrive via gossip as we're already subscribed to the topic, and it may be incorrectly used to satisfy the existing data availability requirement (10 columns), and result in this additional column (instead of a required one) getting persisted, resulting in database inconsistency.

## Proposed fix

- ~~In the `handle_gossip` function, we compute `should_process` for each data column that arrived, and pass it to the beacon processor - this column will get verified, propagated, but NOT processed / imported.~~ I've move this logic to `DataAvailabilityChecker` for consistency with RPC and better separation of concerns.
- This requires us to know the sampling columns for the current epoch, which is different to `globals.sampling_columns` (already updated to the new CGC). 
- To avoid having to maintain multiple column lists for current and next epoch, we compute the full ordered set of custody group count on startup, and this would allow us to quickly retrieve custody columns for both current and next epoch without having to recompute or maintain multiple lists.

The trade off here is we'd have to initialise the `CustodyContext` with the results from `get_custody_groups` _after_ network is initialised, which isn't ideal but will allow us to simplify and clean things up a lot by keeping custody related logic in `CustodyContext`. With this approach, we could potentially remove `network_globals.sampling_columns` too - I haven't done it in this PR though, wanted to see how this approach look like and get some feedback first.

The other options considered:
- Load `node_id` before `BeaconChain` is constructed, and have beacon chain also compute the custody columns. I think this leaks networking types into the `BeaconChain`, so IMO isn't great for separation of concerns and also introduces a circular dependency which will bite us later.
- Having network handle all the custody column related logic and only pass relevant data to the beacon processor. This approach splits custody logic between network and beacon chain - i believe the low cohesivsion sacrifices readability and maintainability.

### DIFF
--- a/beacon_node/beacon_chain/src/data_availability_checker.rs
+++ b/beacon_node/beacon_chain/src/data_availability_checker.rs
@@ -130,8 +130,8 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
         })
     }
 
-    pub fn custody_context(&self) -> Arc<CustodyContext<T::EthSpec>> {
-        self.custody_context.clone()
+    pub fn custody_context(&self) -> &Arc<CustodyContext<T::EthSpec>> {
+        &self.custody_context
     }
 
     /// Checks if the block root is currenlty in the availability cache awaiting import because
@@ -245,7 +245,7 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
         // a new epoch.
         let sampling_columns = self
             .custody_context
-            .sampling_columns_for_slot(slot, &self.spec);
+            .sampling_columns_for_slot(Some(slot), &self.spec);
         let verified_custody_columns = kzg_verified_columns
             .into_iter()
             .filter(|col| sampling_columns.contains(&col.index()))
@@ -289,7 +289,7 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
     ) -> Result<Availability<T::EthSpec>, AvailabilityCheckError> {
         let sampling_columns = self
             .custody_context
-            .sampling_columns_for_slot(slot, &self.spec);
+            .sampling_columns_for_slot(Some(slot), &self.spec);
         let custody_columns = data_columns
             .into_iter()
             .filter(|col| sampling_columns.contains(&col.index()))
@@ -892,7 +892,8 @@ mod test {
             .expect("should put rpc custody columns");
 
         // THEN the sampling size for the end slot of the same epoch remains unchanged
-        let sampling_columns = custody_context.sampling_columns_for_slot(cgc_change_slot, &spec);
+        let sampling_columns =
+            custody_context.sampling_columns_for_slot(Some(cgc_change_slot), &spec);
         assert_eq!(
             sampling_columns.len(),
             spec.validator_custody_requirement as usize // 8
@@ -970,7 +971,8 @@ mod test {
             .expect("should put gossip custody columns");
 
         // THEN the sampling size for the end slot of the same epoch remains unchanged
-        let sampling_columns = custody_context.sampling_columns_for_slot(cgc_change_slot, &spec);
+        let sampling_columns =
+            custody_context.sampling_columns_for_slot(Some(cgc_change_slot), &spec);
         assert_eq!(
             sampling_columns.len(),
             spec.validator_custody_requirement as usize // 8
@@ -995,12 +997,12 @@ mod test {
     fn init_custody_context_with_ordered_columns(
         custody_context: &Arc<CustodyContext<E>>,
         mut rng: &mut StdRng,
-        spec: &Arc<ChainSpec>,
+        spec: &ChainSpec,
     ) -> Vec<u64> {
         let mut all_data_columns = (0..spec.number_of_custody_groups).collect::<Vec<_>>();
         all_data_columns.shuffle(&mut rng);
         custody_context
-            .init_ordered_data_columns_from_custody_groups(all_data_columns.clone(), &spec)
+            .init_ordered_data_columns_from_custody_groups(all_data_columns.clone(), spec)
             .expect("should initialise ordered custody columns");
         all_data_columns
     }
--- a/beacon_node/beacon_chain/src/data_availability_checker.rs
+++ b/beacon_node/beacon_chain/src/data_availability_checker.rs
@@ -814,3 +814,207 @@ impl<E: EthSpec> MaybeAvailableBlock<E> {
         }
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::test_utils::{
+        generate_rand_block_and_data_columns, get_kzg, EphemeralHarnessType, NumBlobs,
+    };
+    use crate::CustodyContext;
+    use rand::prelude::StdRng;
+    use rand::seq::SliceRandom;
+    use rand::SeedableRng;
+    use slot_clock::{SlotClock, TestingSlotClock};
+    use std::collections::HashSet;
+    use std::sync::Arc;
+    use std::time::Duration;
+    use store::HotColdDB;
+    use types::{ChainSpec, ColumnIndex, EthSpec, ForkName, MainnetEthSpec, Slot};
+
+    type E = MainnetEthSpec;
+    type T = EphemeralHarnessType<E>;
+
+    /// Test to verify any extra RPC columns received that are not part of the "effective" CGC for
+    /// the slot are excluded from import.
+    #[test]
+    fn should_exclude_rpc_columns_not_required_for_sampling() {
+        // SETUP
+        let spec = Arc::new(ForkName::Fulu.make_genesis_spec(E::default_spec()));
+        let mut rng = StdRng::seed_from_u64(0xDEADBEEF0BAD5EEDu64);
+
+        let da_checker = new_da_checker(spec.clone());
+        let custody_context = &da_checker.custody_context;
+        let all_column_indices_ordered =
+            init_custody_context_with_ordered_columns(custody_context, &mut rng, &spec);
+
+        // GIVEN a single 32 ETH validator is attached slot 0
+        let epoch = Epoch::new(0);
+        let validator_0 = 0;
+        custody_context.register_validators(
+            vec![(validator_0, 32_000_000_000)],
+            epoch.start_slot(E::slots_per_epoch()),
+            &spec,
+        );
+        assert_eq!(
+            custody_context.num_of_data_columns_to_sample(Some(epoch), &spec),
+            spec.validator_custody_requirement as usize,
+            "sampling size should be the minimal custody requirement == 8"
+        );
+
+        // WHEN additional attached validators result in a CGC increase to 10 at the end slot of the same epoch
+        let validator_1 = 1;
+        let cgc_change_slot = epoch.end_slot(E::slots_per_epoch());
+        custody_context.register_validators(
+            vec![(validator_1, 32_000_000_000 * 9)],
+            cgc_change_slot,
+            &spec,
+        );
+        // AND custody columns (8) and any new extra columns (2) are received via RPC responses.
+        // NOTE: block lookup uses the **latest** CGC (10) instead of the effective CGC (8) as the slot is unknown.
+        let (_, data_columns) = generate_rand_block_and_data_columns::<E>(
+            ForkName::Fulu,
+            NumBlobs::Number(1),
+            &mut rng,
+            &spec,
+        );
+        let block_root = Hash256::random();
+        let requested_columns = &all_column_indices_ordered[..10];
+        da_checker
+            .put_rpc_custody_columns(
+                block_root,
+                cgc_change_slot,
+                data_columns
+                    .into_iter()
+                    .filter(|d| requested_columns.contains(&d.index))
+                    .collect(),
+            )
+            .expect("should put rpc custody columns");
+
+        // THEN the sampling size for the end slot of the same epoch remains unchanged
+        let sampling_columns = custody_context.sampling_columns_for_slot(cgc_change_slot, &spec);
+        assert_eq!(
+            sampling_columns.len(),
+            spec.validator_custody_requirement as usize // 8
+        );
+        // AND any extra columns received via RPC responses are excluded from import.
+        let actual_cached: HashSet<ColumnIndex> = da_checker
+            .cached_data_column_indexes(&block_root)
+            .expect("should have cached data columns")
+            .into_iter()
+            .collect();
+        let expected_sampling_columns = sampling_columns.iter().copied().collect::<HashSet<_>>();
+        assert_eq!(
+            actual_cached, expected_sampling_columns,
+            "should cache only the effective sampling columns"
+        );
+        assert!(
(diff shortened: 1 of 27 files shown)


jimmygchen:
I've labeled it as `do-not-merge` as this still requires test fixes and clean up on docs, but the idea is there and I'd like to get some early feedback.

beacon_node/client/src/builder.rs
```diff
@@ -787,6 +790,20 @@ where
...
+    chain: Arc<BeaconChain<T>>,
+    network_globals: &NetworkGlobals<T::EthSpec>,
+) -> Result<(), String> {
+    let node_id = network_globals.local_enr().node_id().raw();
+    let spec = &chain.spec;
+    let custody_groups_ordered = get_all_custody_groups_ordered(node_id, spec)
+        .map_err(|e| format!("Failed to compute custody groups: {:?}", e))?;
+    chain
+        .data_availability_checker
+        .custody_context()
```
jimmygchen, line 804: TODO: we can probably remove the `clone` in `custody_context()`

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -170,7 +184,30 @@ impl CustodyContext {
...
+            _phantom_data: PhantomData,
+        }
+    }
+
+    pub fn init_all_custody_groups_ordered(
+        &self,
+        all_custody_groups_ordered: Vec<CustodyIndex>,
+        spec: &ChainSpec,
+    ) -> Result<(), String> {
+        let mut ordered_custody_groups = vec![];
```
pawanjay176: I don't understand this.
We basically want to store an ordering of data columns for a given `NodeId`.
Why are we storing columns for every custody index here? Can't we just compute the ordering for the node id and take a slice [..cgc] for the required cgc value?

pawanjay176 (commented):
I like the direction of this, its less invasive than I thought originally

beacon_node/beacon_chain/src/validator_custody.rs
jimmygchen: yeah, this is due to custody groups decoupling - because number of columns could be different to custody groups. The alternative is storing all columns instead of groups, but it may not be as ergonomic because all the configs deal with groups (`sample_per_slot`,  `custody_requirement` etc)

`cgc` doesn't directly translate to columns either, we'd need to get the `sampling_size` (which also uses custody groups as unit), and then convert it to number of columns.

I think it's also doable though - so we'd
- store an ordered list of all columns
- when CGC changes, compute the `sampling_size` and then convert it to `sampling_size_columns`
- and take a slice of all columns using `sampling_size_columns`

The reason i did this was to potentially also store subnets in the `CustodyGroup` struct - it seems like a more logical grouping, and potentially remove `globals.sampling_subnets` and keep all custody related functions and values in `CustodyContext` - I think this is not necessary though, was just an idea that i had earlier, keeping networking logic in network is probably better.

pawanjay176: aghhh i hate the unnecessary abstraction with custody groups. I'm willing to bet that we will never change the columns per custody group and this will just be a painful tech debt forever 😭 

I also prefer having the columns to sample in one place to avoid syncing up everything if possible. 
I really like the approach of storing an ordered list of columns that you described. The computation is constant time so shouldn't be too expensive

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -800,3 +814,209 @@ impl<E: EthSpec> MaybeAvailableBlock<E> {
...
+    use store::HotColdDB;
+    use types::{ChainSpec, ColumnIndex, EthSpec, ForkName, MainnetEthSpec, Slot};
+
+    type E = MainnetEthSpec;
+    type T = EphemeralHarnessType<E>;
+
+    /// Test to verify any extra RPC columns received that are not part of the "effective" CGC for
+    /// the slot are excluded from import.
+    #[test]
+    fn should_exclude_rpc_columns_not_required_for_sampling() {
```
jimmygchen, line 854: Writing these tests make me want to use mocks again :sweat_smile: 

Without mocking, it requires understanding of the internals and how its dependencies work (in this case `custody_context`). The tests are just polluted with setting up the custody context. Whats worse is when the dependency behaviour changes, all these tests will break and require update.

beacon_node/beacon_chain/src/validator_custody.rs
jimmygchen: yeah the custody groups abstraction complicates a lot of things, including the tests.

I've updated it to store an ordered list of columns, i think it seems to work out quite well, and also managed to get rid of `network_globals.sampling_columns`, which allowed us to remove `globals` from a bunch of places 🎉

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -233,15 +233,24 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
...
             KzgVerifiedDataColumn::from_batch_with_scoring(custody_columns, &self.kzg)
                 .map_err(AvailabilityCheckError::InvalidColumn)?;
 
+        // Filter out columns that aren't required for custody for this slot
+        // This is required because `data_columns_by_root` requests the **latest** CGC that _may_
+        // not be yet effective for data availability check, as CGC changes are only effecive from
+        // a new epoch.
+        let sampling_columns = self
+            .custody_context
+            .sampling_columns_for_slot(Some(slot), &self.spec);
```
jimmygchen: If we move this to the `overflow_lru_cache`, we could avoid the below altogether and make the `num_of_data_columns_to_sample` function private.

https://github.com/sigp/lighthouse/blob/ba02fe1da09f3ac28f6d6c03553de106e9afdb64/beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs#L530-L532

beacon_node/beacon_chain/src/data_availability_checker.rs
jimmygchen: I'll make another attempt with mocks tomorrow to compare.

consensus/types/src/data_column_custody_group.rs
```diff
@@ -25,13 +24,32 @@ pub fn get_custody_groups(
     custody_group_count: u64,
     spec: &ChainSpec,
 ) -> Result<HashSet<CustodyIndex>, DataColumnCustodyGroupError> {
+    get_custody_groups_ordered(raw_node_id, custody_group_count, spec)
```
pawanjay176, line 27: I'm wondering if the hashset is required anymore? why don't we make the default return the ordered one according to the spec?

beacon_node/network/src/sync/network_context.rs
```diff
@@ -608,12 +613,7 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
             data_column_requests.map(|data_column_requests| {
                 (
                     data_column_requests,
-                    self.network_globals()
-                        .sampling_columns()
-                        .clone()
-                        .iter()
-                        .copied()
-                        .collect(),
+                    self.chain.sampling_columns_for_slot(None).to_vec(),
```
pawanjay176: don't we know the slot values here?

beacon_node/network/src/sync/network_context.rs
```diff
@@ -551,7 +551,12 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
         // Attempt to find all required custody peers before sending any request or creating an ID
         let columns_by_range_peers_to_request =
             if matches!(batch_type, ByRangeRequestType::BlocksAndColumns) {
-                let column_indexes = self.network_globals().sampling_columns();
+                let column_indexes = self
+                    .chain
+                    .sampling_columns_for_slot(None)
```
pawanjay176: don't we know the slot value here?

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -170,7 +177,35 @@ impl CustodyContext {
...
+    pub fn init_ordered_data_columns_from_custody_groups(
+        &self,
+        all_custody_groups_ordered: Vec<CustodyIndex>,
+        spec: &ChainSpec,
+    ) -> Result<(), String> {
+        let mut ordered_custody_groups = vec![];
+        for custody_index in all_custody_groups_ordered {
+            let columns = compute_columns_for_custody_group(custody_index, spec)
+                .map_err(|e| format!("Failed to compute columns for custody group {e:?}"))?;
+            ordered_custody_groups.extend(columns);
```
pawanjay176: this is `ordered_custody_columns` now no?

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -276,11 +311,37 @@ impl CustodyContext {
...
+    ///
+    /// # Parameters
+    /// * `slot` - Optional slot to determine sampling columns for. If not provided, returns sampling columns at head
+    /// * `spec` - Chain specification containing sampling parameters
+    ///
+    /// # Returns
+    /// A slice of ordered column indices that should be sampled for this slot based on the node's custody configuration
+    pub fn sampling_columns_for_slot(
+        &self,
+        slot: Option<Slot>,
```
pawanjay176: why not `sample_columns_for_epoch` considering we only switch the cgc at epoch boundaries? It'll be easier at sync too that way

pawanjay176: No need to have the EthSpec in the struct too that way

pawanjay176 (commented):
Looking good

beacon_node/network/src/sync/network_context.rs
jimmygchen: CGC can't change during sync so it's fine to use the slot I think, I was being a bit conservative and was preferring to download more and discard them in DA checker.

beacon_node/beacon_chain/src/validator_custody.rs
jimmygchen: yes good catch

beacon_node/beacon_chain/src/validator_custody.rs
jimmygchen: yeah I had considered it but thought slot would be more UX friendly for the caller - most of the time other than range sync we'd be using slots, so we'd have to do conversion at caller site.

`register_validator` also requires `EthSpec` though, feels like its safer and cleaner to have the generic at the struct level  for consistency (but maybe doenst matter because we only use one `EthSpec` anyway)

beacon_node/network/src/sync/network_context.rs
jimmygchen: same as above

consensus/types/src/data_column_custody_group.rs
jimmygchen: As discussed last week, I  had considered it but the diff is quite big so I'll do this in the next PR!

beacon_node/network/src/sync/network_context.rs
jimmygchen: We discussed this last week - CGC doesn't change while syncing, so there should be no differences here and a bit more conservative to request the CGC at head to avoid batches failing if CGC is updated during sync (not possible right now). This is also the current behaviour. Let me know if you prefer to be explicit here, I also don't think it's a bad idea.

beacon_node/beacon_chain/src/validator_custody.rs
jimmygchen: Fixed

beacon_node/beacon_chain/src/data_availability_checker.rs
jimmygchen: Changed my mind, doesn't seem to be worth the effort now since the tests have already been written and doesn't seem _too_ bad now. We could reconsider this if maintenance becomes difficult.

beacon_node/beacon_chain/src/data_availability_checker.rs
jimmygchen: Rethinking about this again, feel like this is a micro-refactor that's not worth it, i think performance-wise this is slightly better than using `contains` to check every columns. Will leave it as it is.

jimmygchen:
Hey @pawanjay176 

Thanks for the review, I've addressed most of your comments and left responses for a few. I don't have strong opinion on those though, happy to make the changes, but thought I'd mention the thoughts behind them.

Let me know what you think!

beacon_node/beacon_chain/src/validator_custody.rs
pawanjay176: I just felt that logically, it might be better to always have the assumption that all slots in an epoch have the same cgc. Having this function per slot seems like a potential footgun to me later on where we could refactor to different values at different slots within an epoch. But i'm fine with it either way

pawanjay176:
This is looking good, i'm testing this on the devnet with changing validator counts.

Update: testing this merged with my backfill branch and it works great 🎉  The only small thing is that it downloads different number of columns for different batches in backfill based on the custody count at head like i mentioned here https://github.com/sigp/lighthouse/pull/7792#discussion_r2252487293.

beacon_node/network/src/sync/network_context.rs
pawanjay176: this is relevant for backfill too though right? using the cgc at head could result in different backfill batches containing different number of columns based on the custody changes at head. 
Backfilling columns would have to account for different number of columns stored which isn't _that_ big a problem but its still a bit weird.

jimmygchen: Yeah great catch, I hadn’t thought about backfill. Will fix this 🙏

beacon_node/beacon_chain/src/validator_custody.rs
jimmygchen: fair point, I'll update it! thanks

beacon_node/network/src/sync/network_context.rs
jimmygchen: Added epoch

beacon_node/beacon_chain/src/validator_custody.rs
jimmygchen: Updated the function to accept a mandatory epoch.

jimmygchen:
Thanks @pawanjay176, I've addressed your review comments 🙏

beacon_node/network/src/sync/network_context.rs
jimmygchen: added

pawanjay176 (approved):
LGTM. Retested it with the changes and works great.

PR #7802, opened 2025-07-25
set process_cpu_seconds_total to type counter

### LINKED ISSUE #3108: Set process_cpu_seconds_total to type counter
## Description

Could be possible to replace type gauge for `process_cpu_seconds_total` metrics field to type counter?

## Version

v1.2.5

carekb684:
## Issue Addressed

Closes #3108

## Proposed Changes

Changed the Prometheus metric process_cpu_seconds_total from a Gauge to a Counter, which is more appropriate for representing cumulative CPU time.

## Additional Info
I found this issue and wanted to contribute.

I chose the approach of resetting the counter and then incrementing it with the new value for simplicity.
An alternative I considered was storing an extra state (previous_seconds) and incrementing only the calculated diff. However, I decided against this to avoid added complexity and state management, since the reset + increment pattern works well in this case.

### DIFF
--- a/common/health_metrics/src/metrics.rs
+++ b/common/health_metrics/src/metrics.rs
@@ -27,8 +27,8 @@ pub static PROCESS_SHR_MEM: LazyLock<Result<IntGauge>> = LazyLock::new(|| {
         "Shared memory used by the current process",
     )
 });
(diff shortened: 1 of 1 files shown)


dapplion:
Hey @carekb684 thanks for contributing! The approach looks solid. Can you run it locally, capture metrics with prometheus and grafana and show in a screen shot that the metric renders nicely?

PR #6532, opened 2024-10-22
State cache memory size WIP

michaelsproul:
## Proposed Changes

Add memory tracking based on:

- https://github.com/sigp/milhouse/pull/51

PR #7803, opened 2025-07-27
Memory Aware Caching in Lighthouse

### LINKED ISSUE #7449: Measure state cache size in memory
With this WIP Milhouse PR we can measure the exact size of the state cache in memory:

- https://github.com/sigp/milhouse/pull/51

That PR needs a bit more work to handle nested lists etc (for full generality), and possibly some API improvements to make it less error-prone.

I had a PoC PR to implement the state cache size in Lighthouse which we can revive once the Milhouse feature itself is ready:

- https://github.com/sigp/lighthouse/pull/6532

PoulavBhowmick03:
WIP

## Issue Addressed

Fixes #7449 

- PR for [Lighthouse: Better State Cache Heuristics](https://github.com/eth-protocol-fellows/cohort-six/blob/master/projects/project-ideas.md#lighthouse-better-state-cache-heuristics)
- [Project Proposal](https://github.com/eth-protocol-fellows/cohort-six/blob/master/projects/Lighthouse_Better_State_Cache_Heuristics.md)

### Memory tracking
* Introduced `BeaconStateMemorySize` (in `store/memsize.rs`) that wraps a `BeaconState` in `BeaconStateWrapper` and measures with Milhouse `MemoryTracker`.
* Added a histogram metric `beacon_state_memory_size_calculation_time` to time memory sizing calls.

### State cache API/behavior
* `StateCache` now has:

  * `max_cached_bytes`: configurable byte ceiling for the hot state cache.
  * `cached_bytes`: last measured total (not incrementally maintained).
  * `put_count`: counter to throttle re-measurement.
  * `needs_remeasure`: boolean, to check if there is need for remeasuring of the `memory_size`

* `put_state`:
  * Every `RECOMPUTE_INTERVAL` inserts calls the full re-measurement routine, if `needs_remeasure` is true.
* `delete_state` and related paths:

  * Remove entries and mappings only (no byte subtraction).. size is enforced by periodic measurement of the entire cache.
* Added `measure_cached_memory_size()`:

  * Runs a single `MemoryTracker` across all cached states and sums `differential_size` to avoid double-counting shared `Arc`s.
* Added `recompute_cached_bytes()`:

  * Measures if over `max_cached_bytes`, prune in batches (uses `headroom`, clamped to 5–64) → re-measure, loop until under the limit.
  * Updates gauge `store_beacon_state_cache_memory_size` and logs measurement time.

### Config / CLI / metrics

* New CLI flag: `--state-cache-max-bytes` (default 512 MiB).
* New constant: `DEFAULT_STATE_CACHE_MAX_BYTES`.
* `HotColdDB` updates the memory gauge with `state_cache.cached_bytes()`.
* Tests updated to assert the default for `max_state_cache_bytes`.


beacon_node/store/src/state_cache.rs
```diff
@@ -229,25 +242,31 @@ impl<E: EthSpec> StateCache<E> {
...
 
         // Insert the full state into the cache.
-        if let Some((deleted_state_root, _)) =
+        if let Some((deleted_state_root, removed_state)) =
             self.states.put(state_root, (state_root, state.clone()))
         {
             deleted_states.push(deleted_state_root);
+            self.cached_bytes = self
+                .cached_bytes
+                .saturating_sub(removed_state.memory_size());
```
dapplion: @michaelsproul can comment but can you just count the size of a single state? Does memory_size only count the size of the data owned by the first copy of the Arc?

dapplion: Can you add timer metrics on `memory_size` to measure how expensive it is in terms of time?

PoulavBhowmick03: Will do so. And the `memory_size` implementing Milhouse's `MemoryTracker` only counts the data owned by `Arc`, once, if that's what you meant? It does not overcount the shared `Arc` data

> Can you add timer metrics on `memory_size` to measure how expensive it is in terms of time?

michaelsproul: The `memory_size` of a state is probably larger than the amount of space we save by deleting it, because some of its memory might be shared with other states.

I don't think it's feasible to store enough information to update the `cached_bytes` after removing a state – we need to just recompute the cache size. This is one of the challenges of this approach. You fundamentally don't know how much memory you will recover unless you recount.

Maybe I'm wrong though? Maybe we could change the order that we count states so that the ones most likely to be pruned are tracked last. Then a state's differential size would be approximately how much memory we'd be able to recover by deleting it.

beacon_node/beacon_chain/src/chain_config.rs
```diff
@@ -10,7 +10,7 @@ pub const DEFAULT_RE_ORG_MAX_EPOCHS_SINCE_FINALIZATION: Epoch = Epoch::new(2);
 /// Default to 1/12th of the slot, which is 1 second on mainnet.
 pub const DEFAULT_RE_ORG_CUTOFF_DENOMINATOR: u32 = 12;
 pub const DEFAULT_FORK_CHOICE_BEFORE_PROPOSAL_TIMEOUT: u64 = 250;
-
+pub const DEFAULT_STATE_CACHE_MAX_BYTES: usize = 536870912;
```
michaelsproul: Just a stylistic thing but we could write this constant as `512 * (1 << 20)`, perhaps with an intermediate constant like `const MEGABYTE: usize = 1 << 20`.

beacon_node/src/cli.rs
```diff
@@ -816,6 +816,15 @@ pub fn cli_app() -> Command {
                 .action(ArgAction::Set)
                 .display_order(0)
         )
+        .arg(
+            Arg::new("state-cache-max-bytes")
+                .long("state-cache-max-bytes")
+                .value_name("BYTES")
+                .help("Specifies the maximum size of the state cache in bytes")
+                .default_value("536870912")
```
michaelsproul: This CLI param should probably be in MiB, as I doubt it's useful to set a finer granularity than 1 MiB.

beacon_node/src/config.rs
```diff
@@ -377,6 +377,12 @@ pub fn get_config<E: EthSpec>(
             .map_err(|_| "state-cache-size is not a valid integer".to_string())?;
     }
 
+    if let Some(max_bytes) = cli_args.get_one::<String>("state-cache-max-bytes") {
```
michaelsproul: `parse_optional` will be slightly less verbose here

beacon_node/beacon_chain/src/chain_config.rs
michaelsproul: Or `512 * 1024 * 1024` is even better!

michaelsproul: This config can probably be deleted actually, as you've got the default in the store config

beacon_node/store/src/memsize.rs
```diff
@@ -0,0 +1,44 @@
...
+/// BeaconState Wrapper for memory tracking.
+pub struct BeaconStateWrapper<'a, E: EthSpec>(pub &'a BeaconState<E>);
+
+impl<'a, E: EthSpec> MemorySize for BeaconStateWrapper<'a, E> {
+    fn self_pointer(&self) -> usize {
+        self.0 as *const _ as usize
+    }
+
+    fn subtrees(&self) -> Vec<&dyn MemorySize> {
+        vec![]
```
michaelsproul: You're missing all of the fields of the `BeaconState`, i.e. the interesting parts. See my previous draft PR:

https://github.com/michaelsproul/lighthouse/blob/e34fafab5977faae5362c3cc16f969256769502e/consensus/types/src/beacon_state/memsize.rs#L17-L90

beacon_node/store/src/memsize.rs
```diff
@@ -0,0 +1,44 @@
...
+/// Extension trait for approximate memory consumption of a `BeaconState`.
+pub trait BeaconStateMemorySize {
+    fn memory_size(&self) -> usize;
+}
+
+impl<E: EthSpec> BeaconStateMemorySize for BeaconState<E> {
+    fn memory_size(&self) -> usize {
+        let wrapper = BeaconStateWrapper(self);
+        // Timer for MemorySize
+        let timer = Instant::now();
```
michaelsproul: You can do this using `metrics::start_timer` and `drop` (no need for manual `Instant` use)

beacon_node/store/src/state_cache.rs
```diff
@@ -391,11 +413,75 @@ impl<E: EthSpec> StateCache<E> {
             .collect::<Vec<_>>();
 
         for state_root in &state_roots_to_delete {
+            self.cached_bytes = self.cached_bytes.saturating_sub(
+                self.states
+                    .peek(state_root)
+                    .map_or(0, |(_, state)| state.memory_size()),
```
michaelsproul: I don't think we can measure the size of a removed state like this, because `memory_size` as written will just calculate the size of this state in isolation, which ignores the structural sharing.

I think it's probably better just to not even try to update this on removal, because it would require a recompute. We can just wait for the next recompute (or trigger a recompute after deletion).

michaelsproul: Yeah, looks like you're already doing multiple recalcs inside `cull`, so this can just be deleted

beacon_node/store/src/state_cache.rs
```diff
@@ -391,11 +413,75 @@ impl<E: EthSpec> StateCache<E> {
...
+        while total_bytes > self.max_cached_bytes {
+            // Cull the cache until we are under the max_cached_bytes limit.
+            let deleted_states = self.cull(batch);
+            if deleted_states.is_empty() {
+                // No more states to delete, break out of the loop.
+                break;
+            }
+
+            // Recalculate the memory size after culling.
+            let (new_total_bytes, elapsed_time) = self.measure_cached_memory_size();
```
michaelsproul: Doing multiple recalcs in a loop might be a bit slow, but we can assess once `MemorySize` is implemented for real

michaelsproul: i.e. I expect the calulcation to take a lot longer once we actually examine the fields of the `BeaconState`, this issue:

 https://github.com/sigp/lighthouse/pull/7803#discussion_r2312957350

PoulavBhowmick03:
Migrated the `memsize.rs` to types, and modified the `impl MemorySize for BeaconState` to use the  `map_beacon_state_*_tree_list_fields_immutable!` macros in this [commit](https://github.com/sigp/lighthouse/pull/7803/commits/c5441b4d513055465dd4d89fc8a9e1f2eba4c08d). Depends on https://github.com/sigp/milhouse/pull/78

beacon_node/store/src/state_cache.rs
```diff
@@ -396,8 +416,69 @@ impl<E: EthSpec> StateCache<E> {
...
 
+    pub fn measure_cached_memory_size(&self) -> (usize, std::time::Duration) {
+        // start histogram
+        let timer = metrics::start_timer(&metrics::BEACON_STATE_MEMORY_SIZE_CALCULATION_TIME);
+
+        let mut tracker = MemoryTracker::default();
+        let mut total_bytes: usize = 0;
+
+        for (_, (_, state)) in &self.states {
+            total_bytes = total_bytes.saturating_add(tracker.track_item(state).differential_size);
```
michaelsproul: the `MemoryTracker` is already counting this, you can just read the `total_bytes` off the `MemoryTracker` at the end using `tracker.total_size()`

beacon_node/store/src/state_cache.rs
```diff
@@ -396,8 +416,69 @@ impl<E: EthSpec> StateCache<E> {
...
+        // If your helper returns (), add an explicit guard to see if the static is OK:
+        match &*metrics::STORE_BEACON_STATE_CACHE_MEMORY_SIZE {
+            Ok(_) => {
+                metrics::set_gauge(&metrics::STORE_BEACON_STATE_CACHE_MEMORY_SIZE, as_i64);
+            }
+            Err(e) => {
+                // TEMPORARY: log once to prove if registration failed
+                tracing::warn!("STORE_BEACON_STATE_CACHE_MEMORY_SIZE not registered: {e}");
+            }
+        }
```
michaelsproul: I'm guessing maybe AI wrote this for you, but you should simplify it to make it align more with the rest of Lighthouse. The way we would normally write this is just:

```rust
metrics::stop_timer(timer);
metrics::set_gauge(&metrics::STORE_BEACON_STATE_CACHE_MEMORY_SIZE, total_bytes as i64);
```

The established practise in LH is that we don't bother doing error logging for metrics ( we ignore registration failures), and we just use casts for integer ops that are very unlikely to fail (the cache size should not be able to exceed an `i64`).

beacon_node/store/src/state_cache.rs
```diff
@@ -396,8 +416,69 @@ impl<E: EthSpec> StateCache<E> {
...
+
+        (total_bytes, std::time::Duration::from_secs(0)) // or your elapsed, if you still return it
```
michaelsproul: If you use `metrics::stop_timer_with_duration` above, you can return the duration here. You should remove the AI's comment as well.

beacon_node/store/src/state_cache.rs
```diff
@@ -396,8 +416,69 @@ impl<E: EthSpec> StateCache<E> {
...
+
+    pub fn recompute_cached_bytes(&mut self) {
+        let mut total_bytes = self.measure_cached_memory_size().0; // sets gauge inside
+        self.cached_bytes = total_bytes;
+
+        // still okay to set again, but not required since measure_* already did it
+        metrics::set_gauge(
+            &metrics::STORE_BEACON_STATE_CACHE_MEMORY_SIZE,
+            i64::try_from(total_bytes).unwrap_or(i64::MAX),
+        );
```
michaelsproul: Seeing as this is unnecessary, it's best to delete it. The less code the better.

```suggestion
```

beacon_node/store/src/state_cache.rs
```diff
@@ -396,8 +416,69 @@ impl<E: EthSpec> StateCache<E> {
...
+
+            // measure again after culling; this will also update the gauge
+            total_bytes = self.measure_cached_memory_size().0;
+            self.cached_bytes = total_bytes;
+
+            metrics::set_gauge(
+                &metrics::STORE_BEACON_STATE_CACHE_MEMORY_SIZE,
+                i64::try_from(total_bytes).unwrap_or(i64::MAX),
+            );
+        }
```
michaelsproul: Unnecessary.

beacon_node/src/cli.rs
```diff
@@ -816,6 +816,15 @@ pub fn cli_app() -> Command {
                 .action(ArgAction::Set)
                 .display_order(0)
         )
+        .arg(
+            Arg::new("state-cache-max-MiB")
```
michaelsproul: Capitalisation in a flag name is very unusual. We should use either `-mb`, or just `state-cache-max-size` and the description can state that the quantity is in megabytes.

```suggestion
            Arg::new("state-cache-max-mb")
```

beacon_node/src/cli.rs
```diff
@@ -816,6 +816,15 @@ pub fn cli_app() -> Command {
...
+                .long("state-cache-max-MiB")
+                .value_name("MiB")
```
michaelsproul: Most users without CS degrees will recognise "megabytes" or MB

```suggestion
                .value_name("MEGABYTES")
```

beacon_node/src/cli.rs
```diff
@@ -816,6 +816,15 @@ pub fn cli_app() -> Command {
...
+                .help("Specifies the maximum size of the state cache in MiB")
```
```suggestion
                .help("Specifies the maximum size of the state cache in megabytes")
```

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -497,10 +499,20 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
             self.block_cache.lock().blob_cache.len() as i64,
         );
         let state_cache = self.state_cache.lock();
+        let (bytes, _age) = state_cache.measure_cached_memory_size();
+        metrics::set_gauge(
+            &metrics::STORE_BEACON_STATE_CACHE_MEMORY_SIZE,
+            i64::try_from(bytes).unwrap_or(i64::MAX),
+        );
```
michaelsproul: Redundant

beacon_node/store/src/metrics.rs
```diff
@@ -269,6 +269,19 @@ pub static STORE_BEACON_STATE_CACHE_SIZE: LazyLock<Result<IntGauge>> = LazyLock:
...
+        "Memory consumed by items in the beacon store state cache",
+    )
+});
+pub static BEACON_STATE_MEMORY_SIZE_CALCULATION_TIME: LazyLock<Result<Histogram>> =
+    LazyLock::new(|| {
+        try_create_histogram(
+            "beacon_state_memory_size_calculation_time",
+            "Time taken to calculate the memory size of a beacon state.",
+        )
+    });
```
michaelsproul: We should rename this metric (and update the description) to reflect the fact that we are timing the measurement of the _state cache_, not a single beacon state.

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -497,10 +499,20 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
+
         metrics::set_gauge(
             &metrics::STORE_BEACON_STATE_CACHE_SIZE,
             state_cache.len() as i64,
         );
+        metrics::set_gauge(
+            &metrics::STORE_BEACON_STATE_CACHE_MEMORY_SIZE,
+            state_cache.cached_bytes() as i64,
+        );
```
michaelsproul: Also redundant.

beacon_node/store/src/state_cache.rs
```diff
@@ -396,8 +416,69 @@ impl<E: EthSpec> StateCache<E> {
...
+                tracing::warn!("STORE_BEACON_STATE_CACHE_MEMORY_SIZE not registered: {e}");
+            }
+        }
+
+        (total_bytes, std::time::Duration::from_secs(0)) // or your elapsed, if you still return it
+    }
+
+    pub fn recompute_cached_bytes(&mut self) {
+        let mut total_bytes = self.measure_cached_memory_size().0; // sets gauge inside
+        self.cached_bytes = total_bytes;
```
michaelsproul: I would be inclined to put this inside `measure_cached_memory_size` or delete the `cached_bytes` field entirely. We don't really need it, I don't think.

michaelsproul (changes requested):

beacon_node/store/src/state_cache.rs
```diff
@@ -396,8 +410,41 @@ impl<E: EthSpec> StateCache<E> {
...
+        let timer = metrics::start_timer(&metrics::STATE_CACHE_MEMORY_SIZE_CALCULATION_TIME);
+
+        let mut tracker = MemoryTracker::default();
+        for (_, (_, state)) in &self.states {
+            // Track each state individually
+            tracker.track_item(state);
+        }
+        let total_bytes = tracker.total_size();
+
+        let duration = metrics::stop_timer_with_duration(timer);
```
dapplion: No need to stop the timer explicitly since it's at the end of the clousue. Just name the variable `_timer` and Drop will collect the metric

beacon_node/store/src/state_cache.rs
```diff
@@ -396,8 +410,41 @@ impl<E: EthSpec> StateCache<E> {
...
     }
-}
 
+    pub fn measure_cached_memory_size(&self) -> (usize, std::time::Duration) {
+        // start histogram
+        let timer = metrics::start_timer(&metrics::STATE_CACHE_MEMORY_SIZE_CALCULATION_TIME);
+
+        let mut tracker = MemoryTracker::default();
+        for (_, (_, state)) in &self.states {
+            // Track each state individually
```
dapplion, line 426: @michaelsproul should we add a second histogram here to track how much time it takes to compute the size of each state? I would assume the times should be roughtly equal but would be good to double check.

beacon_node/store/src/state_cache.rs
```diff
@@ -251,10 +260,15 @@ impl<E: EthSpec> StateCache<E> {
...
             deleted_states.push(deleted_state_root);
         }
-
         // Record the connection from block root and slot to this state.
         let slot = state.slot();
         self.block_map.insert(block_root, slot, state_root);
+        self.put_count += 1;
+
+        if self.put_count >= RECOMPUTE_INTERVAL {
+            self.recompute_cached_bytes();
```
dapplion, line 275: If this operation takes a long time and blocks the `put_state` operation, can that have unintended effects on the callers that now assume that interacting with the memory cache is fast? For the implementation of this I imagined that the recompute would happen on a separate thread once a while to not block other operations. Then pruning will just act on the latest value.

beacon_node/store/src/state_cache.rs
```diff
@@ -396,8 +410,41 @@ impl<E: EthSpec> StateCache<E> {
...
+        metrics::set_gauge(
+            &metrics::STORE_BEACON_STATE_CACHE_MEMORY_SIZE,
+            total_bytes as i64,
+        );
+        (total_bytes, duration)
+    }
+
+    pub fn recompute_cached_bytes(&mut self) {
+        let (mut total_bytes, _) = self.measure_cached_memory_size();
+        let batch = self.headroom.get().clamp(5, 64);
```
dapplion: Make these 5, 64 constants with an explanation

beacon_node/store/src/state_cache.rs
```diff
@@ -396,8 +410,41 @@ impl<E: EthSpec> StateCache<E> {
...
+
+        while total_bytes > self.max_cached_bytes {
```
dapplion: We don't need to perfectly match the cached_bytes every instant. Having this while here makes me worry of repeated `measure_cached_memory_size` calls that will make this function have variables cost and take too much CPU. Why not limit to a single cull operation per recompute_cached_bytes call? If you call `recompute_cached_bytes` once per interval in a separate thread the max cost this feature can have is bounded = a single `measure_cached_memory_size` call

beacon_node/beacon_chain/src/chain_config.rs
```diff
@@ -11,7 +11,7 @@ pub const DEFAULT_RE_ORG_MAX_EPOCHS_SINCE_FINALIZATION: Epoch = Epoch::new(2);
 /// Default to 1/12th of the slot, which is 1 second on mainnet.
 pub const DEFAULT_RE_ORG_CUTOFF_DENOMINATOR: u32 = 12;
 pub const DEFAULT_FORK_CHOICE_BEFORE_PROPOSAL_TIMEOUT: u64 = 250;
-
+pub const DEFAULT_STATE_CACHE_MAX_SIZE: usize = 1024 * 1024 * 1024;
```
michaelsproul: I think 4GB is probably a good default.

michaelsproul: I'll make a few changes and push them here if that's OK

beacon_node/store/src/state_cache.rs
michaelsproul: Yeah, agree we ideally shouldn't be doing this while holding the lock.

We've seen the state cache mutex getting held for extended periods of time (which is partly what we were trying to address in: https://github.com/sigp/lighthouse/pull/8176), and this has a real impact on the node's operation.

If we do add a background maintenance thread (or do it as part of the existing DB maintenance thread) I'm actually not sure how we can avoid holding the lock while working out what to prune. If we are modifying the cache to remove entries within a size bound, we sort of need to hold the lock by definition. Unless we use some persistent data structures, prune a snapshot, and then apply those prunes to the real cache (which might have changed a bit in the meantime).

lighthouse/tests/beacon_node.rs
```diff
@@ -1857,6 +1860,17 @@ fn state_cache_size_flag() {
...
+fn state_cache_max_bytes_default() {
+    CommandLineTest::new()
+        .run_with_zero_port()
+        .with_config(|config| {
+            assert_eq!(
+                config.store.state_cache_max_bytes,
+                DEFAULT_STATE_CACHE_MAX_BYTES
+            );
+        });
+}
```
michaelsproul, line 1882: We also need a test that uses the flag (now `--state-cache-max-mb`) and confirms that the supplied value makes it through to `config.store.state_cache_max_bytes`.

consensus/types/src/beacon_state/memsize.rs
```diff
@@ -0,0 +1,206 @@
...
+                map_beacon_state_gloas_tree_list_fields_immutable!(
+                    &'a _,
+                    self_inner,
+                    |_, self_field| {
+                        subtrees.push(self_field);
+                    }
+                );
+            }
+        }
+        subtrees
```
michaelsproul, line 114: We're missing the cache fields like:

https://github.com/sigp/lighthouse/blob/e5b4983d6baf85770fe4539a565d8a2dd462bc53/consensus/types/src/beacon_state.rs#L611

https://github.com/sigp/lighthouse/blob/e5b4983d6baf85770fe4539a565d8a2dd462bc53/consensus/types/src/beacon_state.rs#L617

https://github.com/sigp/lighthouse/blob/e5b4983d6baf85770fe4539a565d8a2dd462bc53/consensus/types/src/beacon_state.rs#L623

https://github.com/sigp/lighthouse/blob/e5b4983d6baf85770fe4539a565d8a2dd462bc53/consensus/types/src/beacon_state.rs#L629

https://github.com/sigp/lighthouse/blob/e5b4983d6baf85770fe4539a565d8a2dd462bc53/consensus/types/src/beacon_state.rs#L635

https://github.com/sigp/lighthouse/blob/e5b4983d6baf85770fe4539a565d8a2dd462bc53/consensus/types/src/beacon_state.rs#L642

Most of these caches use structural sharing. For the committee caches, each `Arc` entry is a subtree, and many of these will be shared amongst different states. My original PR included the logic for this, you can revive that: https://github.com/sigp/lighthouse/pull/6532/files#diff-dd153a2bc9b9b60b0e179424e2c7c0613f76e4fd86517c6e3767f1b98837b99c

For the caches that use `rpds` it might not be feasible to measure their structural sharing, because we can't implement `MemorySize` for them in a meaningful way. I think skipping these for now would be OK. In summary:

- `committee_caches` (steal code from my old PR)
- `progressive_balances_cache` (this is small with no sharing, already covered by `BeaconState::intrinsic_size`)
- `pubkey_cache`: uses `rpds`, skip for now
- `exit_cache`: small with no sharing, covered by intrinsic size
- `slashings_cache`: uses `rpds`, skip for now
- `epoch_cache`: uses an `Arc`! we should implement `MemorySize`

beacon_node/store/src/state_cache.rs
michaelsproul: I think we could still run this on one of our Holesky nodes to see how it does, in the meantime

consensus/types/src/epoch_cache.rs
```diff
@@ -147,3 +148,45 @@ impl EpochCache {
...
+
+    fn subtrees(&self) -> Vec<&dyn MemorySize> {
+        vec![]
+        // For pushing the vectors, we need to implement MemorySize for Vec<u64> and ActivationQueue.
+        // Q: Do we need to do that, or is it sufficient to just account for their sizes in intrinsic_size()?
+        // let mut subtrees: Vec<&dyn MemorySize> = vec![];
+        // subtrees.push(&self.effective_balances);
+        // subtrees.push(&self.base_rewards);
+        // subtrees.push(&self.activation_queue);
+        // subtrees
```
michaelsproul: Yeah should be good to just use intrinsic size, these fields are not subject to sharing and don't need `subtrees`

PoulavBhowmick03: got it!

beacon_node/beacon_chain/src/chain_config.rs
```diff
@@ -11,7 +11,6 @@ pub const DEFAULT_RE_ORG_MAX_EPOCHS_SINCE_FINALIZATION: Epoch = Epoch::new(2);
 /// Default to 1/12th of the slot, which is 1 second on mainnet.
 pub const DEFAULT_RE_ORG_CUTOFF_DENOMINATOR: u32 = 12;
 pub const DEFAULT_FORK_CHOICE_BEFORE_PROPOSAL_TIMEOUT: u64 = 250;
-
```
dapplion, line 14: Unwanted diff

beacon_node/store/src/state_cache.rs
dapplion: > If we do add a background maintenance thread (or do it as part of the existing DB maintenance thread) I'm actually not sure how we can avoid holding the lock while working out what to prune. If we are modifying the cache to remove entries within a size bound, we sort of need to hold the lock by definition. Unless we use some persistent data structures, prune a snapshot, and then apply those prunes to the real cache (which might have changed a bit in the meantime).

If we hold the lock to prune as part of the prune operation or as part of the put operation, aren't we still making the cache unavailable to writers for the same amount of time? However if we do it in a background thread we may or may not delay of put operation.

galadd:
Why is this remaining stale?

galadd:
Hi @PoulavBhowmick03, are you still planning on working on this? If not, I am happy to pick this up, rebase it against the latest, and address the outstandings

Also, @michaelsproul, is this still something you are open to merging if I drive it over the line

michaelsproul:
Hi @galadd, no we have decided not to pursue this approach. The overhead from calculating the true memory size of the cache entries is too high.

galadd:
Hi @michaelsproul, that actually makes sense. Is the cache performance stable for now?

I'm looking for other areas to contribute. Do you have any other open issues related to storage or networking that you think need attention?

galadd:
@michaelsproul, have you considered monitoring the process RSS, and just flush the cache when approaching system limit, that way we don't have to calculate the true memory size of cache entries

On one hand, other programs using the memory can cause early eviction, but temporary flushes under load > node crash. Also, we can tune the RSS threshold to give us a safety buffer

PR #7804, opened 2025-07-28
Use unique enclave names for CI Kurtosis tests on self hosted runners

jimmygchen:
## Issue Addressed

- Use self-hosted runners labelled `Kurstosis` to run Fusaka devnet tests on CI
- Use unique enclave names to avoid collisions on Kurtosis runner machines


jimmygchen:
Getting Kurtosis to work on our self hosted runner might not be straight forward given the current setup. 

It looks like the client isn't able to communicate with the engine via the API port due to access restrictions (the kurtosis client needs to be able to access the Kurtosis engine API)

Issue #7760, opened 2025-07-18
Slim down PersistedForkChoice


dapplion:
The fork-choice state is persistsed often and it's big, causing a lot of I/O. Currently, its size also grows during non-finality.

Some ideas:
- Don't persist fork-choice nodes, instead re-compute them on start-up from stored blocks. We would need to compute the cost of doing so in long non-finality. Probably we need to store the payload status of them, or maybe start them all as `syncing`
- Just persist the last votes once in a while. Persisting just votes should be smaller and constant size. We can experiment with how often we need to persist those votes too, or only persisting them on a clean shut-down. Worst case we will get new votes on the epoch after start-up. We can also run a compression algorithm on the votes as most of them have the same roots.

@michaelsproul

dapplion:
A mainnet node without subscribing all subnets had a `PersistedForkChoice` entry in the DB of 242 MB. Looking at the type, assuming 1.5M total indices 1M active let's try to account for those 242 MB.

```rust
pub struct PersistedForkChoice {
    pub fork_choice: fork_choice::PersistedForkChoice, // 242 - 60 = 166 MB
    pub fork_choice_store: PersistedForkChoiceStoreV17, // ~60 MB
}

pub struct PersistedForkChoice {
    // ProtoArrayForkChoice.as_bytes()
    pub proto_array_bytes: Vec<u8>,
    // slot worth of attestations 0.25 MB
    queued_attestations: Vec<QueuedAttestation>, // <== Double check, could be more depending on aggregation
}

// Written as ProtoArraySszContainer
pub struct ProtoArrayForkChoice {
    pub(crate) proto_array: ProtoArray,
    pub(crate) votes: ElasticList<VoteTracker>,
    pub(crate) balances: JustifiedBalances,
}

// Actual name in the codebase is `SszContainer`
pub struct ProtoArraySszContainer {
    // 108 MB ?? Duplicated list of votes
    pub votes: Vec<VoteTracker>,
    // 12 MB
    pub balances: Vec<u64>,
    pub prune_threshold: usize,
    pub justified_checkpoint: Checkpoint,
    pub finalized_checkpoint: Checkpoint,
    #[superstruct(only(V17))]
    // In 1 month of blocks 98 MB, 456 bytes / item
    pub nodes: Vec<ProtoNodeV17>,  // <== Double check, should be small but maybe not if pruning is broken?
    // In 1 month of blocks 8.6 MB, 40 bytes / item
[...]

PR #7805, opened 2025-07-28
Shrink persisted fork choice data

michaelsproul:
## Issue Addressed

Closes:

- https://github.com/sigp/lighthouse/issues/7760

## Proposed Changes

- [x] Remove `balances_cache` from `PersistedForkChoiceStore` (~65 MB saving on mainnet)
- [x] Remove `justified_balances` from `PersistedForkChoiceStore` (~16 MB saving on mainnet)
- [x] Remove `balances` from `ProtoArray`/`SszContainer`.
- [x] Implement zstd compression for votes
- [x] Fix bug in justified state usage
- [x] Bump schema version to V28 and implement migration.

## Testing Results

Total bytes written is way down:

write_bytes_total

Time spent writing to disk is also way down:

write_bytes_seconds


beacon_node/beacon_chain/src/canonical_head.rs
```diff
@@ -1014,10 +1014,42 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
         fork_choice: &BeaconForkChoice<T>,
     ) -> KeyValueStoreOp {
+        use ssz::Encode;
         let persisted_fork_choice = PersistedForkChoice {
             fork_choice: fork_choice.to_persisted(),
             fork_choice_store: fork_choice.fc_store().to_persisted(),
         };
+        let ssz_container = fork_choice.proto_array().as_ssz_container();
+        let fc_store = &persisted_fork_choice.fork_choice_store;
+        println!(
```
dapplion: Part of the WIP?

michaelsproul: yeah this is just me seeing how much space each bit uses

dapplion:
Closes https://github.com/sigp/lighthouse/issues/7760 or we want to go further?

michaelsproul:
This isn't finished yet, I'm working on the compression now and it gets the size down to 1.9MB total

beacon_node/beacon_chain/src/persisted_fork_choice.rs
```diff
@@ -39,4 +42,42 @@ macro_rules! impl_store_item {
...
+    ) -> Result<KeyValueStoreOp, Error> {
+        Ok(KeyValueStoreOp::PutKeyValue(
+            DBColumn::ForkChoice,
+            key.as_slice().to_vec(),
+            self.as_bytes(store_config)?,
+        ))
+    }
+}
+
+// FIXME(sproul): dedupe?
```
dapplion: Yes it makes sense as this crate already has store as dependency and this function uses StoreConfig

michaelsproul: fixed

beacon_node/store/src/config.rs
```diff
@@ -194,15 +194,23 @@ impl StoreConfig {
         }
     }
 
-    pub fn compress_bytes(&self, ssz_bytes: &[u8]) -> Result<Vec<u8>, Error> {
+    /// Compress bytes using zstd and the compression level from `self`.
+    pub fn compress_bytes(&self, ssz_bytes: &[u8]) -> Result<Vec<u8>, std::io::Error> {
```
jimmygchen, line 198: It feels a bit odd to have compression & decompression logic placed in `StoreConfig`. Can this be extracted to separate concerns?

michaelsproul: Yeah? I figured this is the place with the minimum dependency surface. Compress/decompress only _need_ the store config, not the full store. Moving them to the HotColdDB would make these functions harder to call/test (not that we're doing much of that).

I guess another option is standalone functions that take the compression level (or the store config) as arguments. Let me know what you were thinking

consensus/fork_choice/src/fork_choice.rs
```diff
@@ -795,8 +802,15 @@ where
         if unrealized_justified_checkpoint.epoch
             > self.fc_store.unrealized_justified_checkpoint().epoch
         {
-            self.fc_store
-                .set_unrealized_justified_checkpoint(unrealized_justified_checkpoint);
+            // Justification has recently updated therefore the justified state root should be in
+            // range of the head state's `state_roots` vector.
+            let unrealized_justified_state_root =
+                state.get_state_root_at_epoch_start(unrealized_justified_checkpoint.epoch)?;
```
jimmygchen, line 808: Is it practically possible that the justified state root is not in the head states's `state_roots` vector? should we handle this? 

If we have a long period without justification, would this cause epoch processing to keep failing with `SlotOutOfBounds`?

Might be easy to try this on minimal preset give we only keep historical state roots for 64 slots / 8 epochs.

michaelsproul: No I don't think it's possible to justify an epoch that's out of range. You can only justify an epoch up to 4 epochs in the past, and even the minimal spec provides more look back than this.

jimmygchen: ah yeah, we don't justify older epochs. With that constraint i think we're fine. Where is the 4 epochs coming from?

michaelsproul: It's either 3 or 4, it's in process_justification_and_finalization in the spec (I'm afk, going off memory). Think it might be 3 epochs for just and 4 for finalization

beacon_node/store/src/config.rs
jimmygchen: yeah i agree with minimum dependency but it doesn't really fit into responsibility of a `StoreConfig` object. I was just thinking standalone function or separate struct (which you originally had in `hdiff.rs`). I'm not seeing a big issue with this at the moment, but i think avoiding this pattern would keep the config object cleaner longer term.

beacon_node/beacon_chain/src/beacon_fork_choice_store.rs
```diff
@@ -162,9 +164,9 @@ where
     /// It is assumed that `anchor` is already persisted in `store`.
     pub fn get_forkchoice_store(
         store: Arc<HotColdDB<E, Hot, Cold>>,
-        anchor: &BeaconSnapshot<E>,
+        anchor: BeaconSnapshot<E>,
```
jimmygchen, line 168: Do we need ownership here? we're cloning the beacon state below, and i dont see this moved anywhere

beacon_node/beacon_chain/src/schema_change/migration_schema_v28.rs
```diff
@@ -0,0 +1,152 @@
...
+
+    let fc_store = BeaconForkChoiceStore::from_persisted_v17(
+        persisted_fork_choice_v17.fork_choice_store_v17,
+        justified_state_root,
+        unrealized_justified_state_root,
+        db.clone(),
+    )
+    .map_err(|e| {
+        Error::MigrationError(format!(
+            "Error loading fork choise store from persisted: {e:?}"
```
jimmygchen: small typo `fork choice`

consensus/fork_choice/src/fork_choice.rs
jimmygchen: thanks!

jimmygchen (approved):
Great optimisation 🎉  

I wasn't very familiar with this part of the codebase before so it took a while to review. I've reviewed this and it looks solid and clean to me. Really nice work!

jimmygchen:
Will leave it to you to merge if you'd like to do more testing or another pair of eyes (maybe @dapplion ?)

beacon_node/store/src/config.rs
jimmygchen: also just a nit, i don't really feel like this needs to change

Issue #7759, opened 2025-07-18
/eth/v1/beacon/light_client/update as ssz does not follow the spec


simon-jentzsch:
The spec defines the response when called with application/octet-stream as:

> 
> Sequence of zero or more response_chunk. Each successful response_chunk MUST contain a single LightClientUpdate payload:
> 
```
 (
   response_chunk_len: Little-endian Uint64 byte length of `response_chunk`
   response_chunk: (
     context: 4 byte `ForkDigest`
     payload: SSZ serialized payload bytes
   )
 )
```
> 
> For each response_chunk, a ForkDigest-context based on compute_fork_version(compute_epoch_at_slot(update.attested_header.beacon.slot)) is used to select the fork namespace of the Response type. Note that this fork_version may be different from the one used to verify the update.sync_aggregate, which is based on update.signature_slot.

Currnetly the light client updates returned by lighthouse fail while others like nimbus or lodestar follow this spec and work fine.

michaelsproul:
Is it a problem specifically with SSZ? Does LH work when you use JSON?

Can you share a script/program that you're using to test this?

simon-jentzsch:
It's not only the SSZ-Format, but also the JSON-Values are not working.
There are 2 issues:
1. even though I used count=1, the resulting array returned 2 LightClientUpdates, where the second one had an older slot not even matching the required period.
2. The merkle proof (branch) does produce the correct state_root.

I wrote a simply test-script which takes the json values, and verifies the merkle proof by calculating the state root. 
When I test this with the JSON-output from lodestar, it produces the correct state_root, but lighthouse does not.


Here is how to reproduce:

```sh
# fetch a light client update from lodestar
> curl https://lodestar-mainnet.chainsafe.io/eth/v1/beacon/light_client/updates?start_period=1487&count=1 > lodestar.json
> node light_client_update.js lodestar.json
valid state root 

# use the same script with lighthouse
>  curl https://gnosis-chiado.core.chainstack.com/<APIKEY>/eth/v1/beacon/light_client/updates?start_period=2140&count=1  > lighthouse.json
> node light_client_update.js lighthouse.json
State root mismatch
sync_root :  c90c77849e0476eb5af016f3d31e1ac92b9144fee7e322a68053658f2a9fc637
state_root_calculated :  ff64827679d5ca2ea540c824c39825d02b09df0db4ed16b6a56fb5f483644d32
[...]

PR #7806, opened 2025-07-28
Fix ssz formatting for `/light_client/updates` beacon API endpoint

eserilev:
## Issue Addressed

#7759

## Proposed Changes

We were incorrectly encoding the full response from `/light_client/updates` instead of only encoding the light client update

## Additional Notes

This repo has a script that verifies light client data. I've used the repo to verify lc updates generated from both lodestar and lighthouse
 
https://github.com/eserilev/verify_light_client_update/tree/main


beacon_node/http_api/src/light_client.rs
```diff
@@ -146,23 +148,24 @@ pub fn validate_light_client_updates_request<T: BeaconChainTypes>(
...
-    let response_chunk_len = fork_digest.len() + payload.len();
+    let response_chunk_len = fork_digest.len() + light_client_update.ssz_bytes_len();
 
     let response_chunk = LightClientUpdateResponseChunkInner {
         context: fork_digest,
-        payload,
+        payload: light_client_update,
     };
 
+    println!("response chunk len {:?}", response_chunk_len);
```
michaelsproul: Delete

common/eth2/src/lib.rs
```diff
@@ -981,6 +981,32 @@ impl BeaconNodeHttpClient {
...
+            .push("light_client")
+            .push("updates");
+
+        path.query_pairs_mut()
+            .append_pair("start_period", &start_period.to_string());
+
+        path.query_pairs_mut()
+            .append_pair("count", &count.to_string());
+
+        self.get_bytes_opt_accept_header(path, Accept::Ssz, self.timeouts.get_beacon_blocks_ssz)
```
michaelsproul: Should probably use the default timeout here?

```suggestion
        self.get_bytes_opt_accept_header(path, Accept::Ssz, self.timeouts.default)
```

common/eth2/src/types.rs
```diff
@@ -823,16 +824,44 @@ pub struct LightClientUpdatesQuery {
...
+        buf.extend_from_slice(&self.response_chunk_len.to_le_bytes());
+        buf.extend_from_slice(&self.response_chunk.context);
+        match &self.response_chunk.payload {
+            LightClientUpdate::Altair(lc) => lc.ssz_append(buf),
+            LightClientUpdate::Capella(lc) => lc.ssz_append(buf),
+            LightClientUpdate::Deneb(lc) => lc.ssz_append(buf),
+            LightClientUpdate::Electra(lc) => lc.ssz_append(buf),
+            LightClientUpdate::Fulu(lc) => lc.ssz_append(buf),
+        };
+    }
```
michaelsproul: This looks a bit sus to me, how does this differ from what `derive` generates?

eserilev: you're right, the match is exactly what `derive` generates. I cleaned things up a bit [67b45a9](https://github.com/sigp/lighthouse/pull/7806/commits/67b45a9ea6b15d710e028a184879b1fbf96dbbe6)

michaelsproul (approved):
LGTM!

PR #7807, opened 2025-07-29
Eip 7732 containers

shane-moore:
## Changes
- added `Constants` / `Misc` /  `Preset` / `Containers` from [consensus specs](https://ethereum.github.io/consensus-specs/specs/_features/eip7732/beacon-chain/#domain-types) while cross-referencing the old epbs [branch](https://github.com/sigp/lighthouse/tree/eip-7732), which already had some of the containers built out and just needed some minor tweaks per updates to the spec 

Note: I made a couple fixes to the gloas boilerplate to get `cargo build` to compile and `cargo test -p types` to be all passing in a separate [PR](https://github.com/sigp/lighthouse/pull/7810)

## Todo for Containers (didn't want this PR to get too big)
- `BeaconBlockBody`
- `BeaconState`


consensus/types/src/builder_pending_withdrawal.rs
```diff
@@ -0,0 +1,38 @@
...
+    Deserialize,
+    Encode,
+    Decode,
+    TreeHash,
+    TestRandom,
+)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+#[context_deserialize(ForkName)]
+pub struct BuilderPendingWithdrawal {
+    #[serde(with = "serde_utils::address_hex")]
```
shane-moore, line 25: we seem to serialize to `0xABC` instead of byte array when sending json over the wire

consensus/types/src/chain_spec.rs
```diff
@@ -468,6 +475,12 @@ impl ChainSpec {
             .is_some_and(|fulu_fork_epoch| fulu_fork_epoch != self.far_future_epoch)
     }
 
+    /// Returns true if `GLOAS_FORK_EPOCH` is set and is not set to `FAR_FUTURE_EPOCH`.
+    pub fn is_gloas_scheduled(&self) -> bool {
```
shane-moore: added this to resolve build issues since the function was referenced but didn't exist

consensus/types/src/config_and_preset.rs
```diff
@@ -160,7 +160,7 @@ mod test {
             .expect("error opening file");
         let mut mainnet_spec = ChainSpec::mainnet();
         // setting fulu_fork_epoch because we are roundtripping a fulu config
-        mainnet_spec.fulu_fork_epoch = Some(Epoch::new(42));
+        mainnet_spec.gloas_fork_epoch = Some(Epoch::new(42));
```
shane-moore: this test was failing because we were setting `mainnet_spec.fulu_fork_epoch` instead of `mainnet_spec.gloas_fork_epoch`, but the check at the end is for `gloas`: `assert_eq!(ConfigAndPreset::Gloas(from), yamlconfig);`

consensus/types/src/eth_spec.rs
```diff
@@ -448,6 +469,8 @@ impl EthSpec for MainnetEthSpec {
     type MaxAttestationsElectra = U8;
     type MaxWithdrawalRequestsPerPayload = U16;
     type MaxPendingDepositsPerEpoch = U16;
+    type PTCSize = U64; // todo: verify if needs to be U512 for some reason like in Mark's OG implementation
```
shane-moore: @ethDreamer, just wanted to check with you on this. pretty sure all good now but wanted to flag it anyways (https://ethereum.github.io/consensus-specs/specs/_features/eip7732/beacon-chain/#misc_1)

consensus/types/src/eth_spec.rs
```diff
@@ -585,6 +612,8 @@ impl EthSpec for GnosisEthSpec {
     type BytesPerCell = U2048;
     type KzgCommitmentsInclusionProofDepth = U4;
     type ProposerLookaheadSlots = U32; // Derived from (MIN_SEED_LOOKAHEAD + 1) * SLOTS_PER_EPOCH
+    type PTCSize = U64; // todo: verify if needs to be U512 for some reason like in Mark's OG implementation
```
shane-moore: @ethDreamer, just wanted to check with you on this. pretty sure all good now but wanted to flag it anyways (https://ethereum.github.io/consensus-specs/specs/_features/eip7732/beacon-chain/#misc_1)

consensus/types/src/execution_bid.rs
```diff
@@ -0,0 +1,47 @@
...
+    Decode,
+    Deserialize,
+    TreeHash,
+    Derivative,
+    TestRandom,
+)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+#[derivative(PartialEq, Hash)]
+#[context_deserialize(ForkName)]
+// This is what Potuz' spec calls an `ExecutionPayload` even though it's clearly a bid.
```
shane-moore: kept this as `ExecutionBid` per our convo, in lieu of `ExecutionPayloadHeader` per [spec](https://ethereum.github.io/consensus-specs/specs/_features/eip7732/beacon-chain/#executionpayloadheader)

consensus/types/src/execution_bid.rs
```diff
@@ -0,0 +1,47 @@
...
+pub struct ExecutionBid {
+    pub parent_block_hash: ExecutionBlockHash,
+    pub parent_block_root: Hash256,
+    pub block_hash: ExecutionBlockHash,
+    #[serde(with = "serde_utils::address_hex")]
+    pub fee_recipient: Address, // todo(eip-7732): verify if this needs address_hex serialization
```
shane-moore: we seem to serialize to 0xABC instead of byte array when sending json over the wire

consensus/types/src/payload_attestation_data.rs
```diff
@@ -0,0 +1,27 @@
...
+    TestRandom, TreeHash, Debug, Clone, PartialEq, Eq, Encode, Decode, Serialize, Deserialize, Hash,
+)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+#[context_deserialize(ForkName)]
+pub struct PayloadAttestationData {
+    pub beacon_block_root: Hash256,
+    pub slot: Slot,
+    pub payload_present: bool,
+}
+// todo(eip-7732): Mark's implementation  has PayloadStatus as an enum, but spec calls for a bool. Need to clarify this.
```
shane-moore: @ethDreamer, just flagging this spec change from your initial impl, [spec](https://ethereum.github.io/consensus-specs/specs/_features/eip7732/beacon-chain/#payloadattestationdata)

consensus/types/src/signed_beacon_block.rs
```diff
@@ -967,6 +967,7 @@ mod test {
         chain_spec.deneb_fork_epoch = Some(Epoch::new(4));
         chain_spec.electra_fork_epoch = Some(Epoch::new(5));
         chain_spec.fulu_fork_epoch = Some(Epoch::new(6));
+        chain_spec.gloas_fork_epoch = Some(Epoch::new(7));
```
shane-moore: this was missing in gloas boilerplate

consensus/types/src/builder_pending_withdrawal.rs
ethDreamer: This looks correct to me. Not sure what your question is exactly. These objects would only affect serialization on the Beacon HTTP API. We use a different set of objects for serialization on the engine API as those have different serialization requirements.

consensus/types/src/eth_spec.rs
```diff
@@ -75,6 +75,7 @@ pub trait EthSpec: 'static + Default + Sync + Send + Clone + Debug + PartialEq +
     type EpochsPerSlashingsVector: Unsigned + Clone + Sync + Send + Debug + PartialEq;
     type HistoricalRootsLimit: Unsigned + Clone + Sync + Send + Debug + PartialEq;
     type ValidatorRegistryLimit: Unsigned + Clone + Sync + Send + Debug + PartialEq;
+    type BuilderPendingWithdrawalsLimit: Unsigned + Clone + Sync + Send + Debug + PartialEq;
```
ethDreamer: small nit here - this is a state list length but this list appears as all phase0 state list lengths - others have been put along with their respective fork (see Electra's `PendingDepositsLimit` and `PendingPartialWithdrawalsLimit`)

consensus/types/src/eth_spec.rs
ethDreamer: you must have gotten confused because the spec you linked also has this as `512`?

consensus/types/src/execution_bid.rs
ethDreamer: So.. when gossiping payloads we always serialize with `SSZ` which is unaffected by the `serde` configuration. SSZ serialization/deserialization is achieved from `Encode` and `Decode` traits while JSON serialization/deserialization is done with the `Serialize` and `Deserialize` traits via `serde`.

There are two ways that consensus objects get serialized into JSON:
1. The beacon API (which while use these objects directly)
2. The engine API (where the objects are first converted to [`JSONVariant`](https://github.com/sigp/lighthouse/blob/stable/beacon_node/execution_layer/src/engine_api/json_structures.rs#L78) so that the serialization matches the engine API.. side note.. perhaps I should rename all instances from JSON to Engine as that would make more sense..

I don't believe this object ever enters the Engine API so it would only need to conform to the [beacon API](https://ethereum.github.io/beacon-APIs/#/Beacon) standards for JSON encoding.

consensus/types/src/indexed_payload_attestation.rs
```diff
@@ -0,0 +1,41 @@
...
+    Encode,
+    Decode,
+    Serialize,
+    Deserialize,
+)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+#[serde(bound = "E: EthSpec", deny_unknown_fields)]
+#[cfg_attr(feature = "arbitrary", arbitrary(bound = "E: EthSpec"))]
+#[context_deserialize(ForkName)]
+pub struct IndexedPayloadAttestation<E: EthSpec> {
```
ethDreamer, line 14: okay so i was under the impression that the `PayloadAttestation`s never go on chain but the existence of this object (and the latest spec) has them in the `BeaconBlock`.. i need to go figure out why..

consensus/types/src/payload_attestation.rs
```diff
@@ -0,0 +1,36 @@
...
+#[serde(bound = "E: EthSpec", deny_unknown_fields)]
+#[cfg_attr(feature = "arbitrary", arbitrary(bound = "E: EthSpec"))]
+#[derivative(PartialEq, Hash)]
+#[context_deserialize(ForkName)]
+pub struct PayloadAttestation<E: EthSpec> {
+    pub aggregation_bits: BitList<E::PTCSize>,
+    pub data: PayloadAttestationData,
+    pub signature: AggregateSignature,
+}
+
```
ethDreamer, line 22: Similarly here the `PayloadAttestation`s are [aggregated by the proposer](https://github.com/ethereum/consensus-specs/blob/master/specs/_features/eip7732/validator.md#constructing-the-new-payload_attestations-field-in-beaconblockbody).. this was the opposite of what I had understood.. I wonder if this was accidentally left in.. going to go check.

consensus/types/src/payload_attestation_data.rs
ethDreamer: yeah the spec used to have an enum of different values but has since been simplified.

ethDreamer (changes requested):
Great start! Only small changes and I need to go clarify some stuff about the spec I now realize.

consensus/types/src/eth_spec.rs
shane-moore: @ethDreamer, I'm reading 
image
as saying: PTCSize should be of type uint64 with a constant value of 512.  so I think we're ok to uses `U64` instead of `U512` in the `eth_spec.rs`?

consensus/types/src/builder_pending_withdrawal.rs
shane-moore: yep, just making sure we wanted hex encoding of `fee_recipient` for json. thanks for the context that this is only applicable to beacon http serialization

consensus/types/src/execution_bid.rs
shane-moore: for sure! I was familiar with `Encode` and `Decode` being for ssz, and `Serialize`/`Deserialize` being for json, but I did not know about the engine api having to convert to `JSONVariant` for serialization.  Will keep this in mind

consensus/types/src/eth_spec.rs
ethDreamer: Ohhhh no these types are special. The type is actually its value, not the number of bits in the type. Look at the `typenum` crate in rust.

shane-moore: ahh, I see! `typenum` lets you express that the type represents `512` in this case, nice

updated in latest commit https://github.com/sigp/lighthouse/pull/7807/commits/76f9f83d2455d12295966fb292d21590f263103e

consensus/types/src/eth_spec.rs
shane-moore: makes total sense, updated in https://github.com/sigp/lighthouse/pull/7807/commits/76f9f83d2455d12295966fb292d21590f263103e

Clay-1010:
￼

consensus/types/src/beacon_block_body.rs
```diff
@@ -127,17 +127,21 @@ pub struct BeaconBlockBody<E: EthSpec, Payload: AbstractExecPayload<E> = FullPay
     #[superstruct(only(Fulu), partial_getter(rename = "execution_payload_fulu"))]
     #[serde(flatten)]
     pub execution_payload: Payload::Fulu,
-    #[superstruct(only(Gloas), partial_getter(rename = "execution_payload_gloas"))]
-    #[serde(flatten)]
-    pub execution_payload: Payload::Gloas,
+    // execution_payload removed from Gloas, replaced with signed_execution_bid below
```
ethDreamer: I don't think this comment is necessary
```suggestion
```

consensus/types/src/beacon_block_body.rs
```diff
@@ -127,17 +127,21 @@ pub struct BeaconBlockBody<E: EthSpec, Payload: AbstractExecPayload<E> = FullPay
...
     #[superstruct(only(Capella, Deneb, Electra, Fulu, Gloas))]
     pub bls_to_execution_changes:
         VariableList<SignedBlsToExecutionChange, E::MaxBlsToExecutionChanges>,
-    #[superstruct(only(Deneb, Electra, Fulu, Gloas))]
+    // blob_kzg_commitments removed from Gloas, moved to `ExecutionPayloadEnvelope`
```
ethDreamer:
```suggestion
```

consensus/types/src/beacon_block_body.rs
```diff
@@ -127,17 +127,21 @@ pub struct BeaconBlockBody<E: EthSpec, Payload: AbstractExecPayload<E> = FullPay
...
+    #[superstruct(only(Deneb, Electra, Fulu))]
     pub blob_kzg_commitments: KzgCommitments<E>,
-    #[superstruct(only(Electra, Fulu, Gloas))]
+    // execution_requests removed from Gloas, moved to `ExecutionPayloadEnvelope`
```
ethDreamer:
```suggestion
```

consensus/types/src/beacon_state.rs
```diff
@@ -2103,6 +2121,20 @@ impl<E: EthSpec> BeaconState<E> {
         }
     }
 
+    pub fn is_parent_block_full(&self) -> bool {
+        match self {
+            BeaconState::Base(_) | BeaconState::Altair(_) => false,
```
ethDreamer, line 2126:
```suggestion
            BeaconState::Base(_) | BeaconState::Altair(_) => false,
            // TODO(EIP-7732): check the implications of this when we get to forkchoice modifications
```

consensus/types/src/dumb_macros.rs
```diff
@@ -0,0 +1,107 @@
+// These would usually be created by superstuct but now there's no longer a 1:1 mapping between
+// the variants for ExecutionPayload and the variants for
+// - ExecutionPayloadHeader
+// - FullPayload
+// - BlindedPayload
+
```
ethDreamer, line 7:
```suggestion
TODO(EIP-7732): get rid of this whole file and panics once the engine_api is refactored for ePBS
```

consensus/types/src/execution_bid.rs
```diff
@@ -0,0 +1,38 @@
...
+use test_random_derive::TestRandom;
+use tree_hash_derive::TreeHash;
+
+#[derive(
+    Default, Debug, Clone, Serialize, Encode, Decode, Deserialize, TreeHash, Derivative, TestRandom,
+)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+#[derivative(PartialEq, Hash)]
+#[context_deserialize(ForkName)]
+// This is what Potuz' spec calls an `ExecutionPayload` even though it's clearly a bid.
```
ethDreamer:
```suggestion
// This is what Potuz' spec calls an `ExecutionPayloadHeader` even though it's clearly a bid.
// https://github.com/ethereum/consensus-specs/blob/bba2c7be148d6d921d2ca5e1cc528f5daaf456d9/specs/gloas/beacon-chain.md#executionpayloadheader
```

consensus/types/src/signed_execution_bid.rs
```diff
@@ -0,0 +1,34 @@
...
+use test_random_derive::TestRandom;
+use tree_hash_derive::TreeHash;
+
+#[derive(
+    TestRandom, TreeHash, Debug, Clone, Encode, Decode, Serialize, Deserialize, Derivative,
+)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+#[derivative(PartialEq, Hash)]
+#[context_deserialize(ForkName)]
+pub struct SignedExecutionBid {
```
ethDreamer, line 16:
```suggestion
// https://github.com/ethereum/consensus-specs/blob/bba2c7be148d6d921d2ca5e1cc528f5daaf456d9/specs/gloas/beacon-chain.md#signedexecutionpayloadheader
pub struct SignedExecutionBid {
```

ethDreamer (commented):
looks good! just a couple of changes now

ethDreamer:
continuing this work in #7923

PR #7808, opened 2025-07-29
Add metrics to track beacon processor queue times

eserilev:
## Issue Addressed

This PR adds a created_timestamp to the beacon processor send channel. When work items are sent through that channel `try_send` will forward the work event along with the current timestamp to the beacon processor. When the work event is completed the `Drop` impl for `SendOnDrop` will track the time it took from work event creation to its completion. Previously we only had data on how long a work event took to process, but didn't have data on how long it sat in the queue + how long it took to process. 

## Additional notes

The aim was to minimize code diff in this PR so it could be easily reviewed. It might not be the cleanest solution ever, but we could always refactor things down the line

### DIFF
--- a/beacon_node/beacon_processor/src/lib.rs
+++ b/beacon_node/beacon_processor/src/lib.rs
@@ -1504,7 +1504,7 @@ impl<E: EthSpec> BeaconProcessor<E> {
     ) {
         let work_id = work.str_id();
         let work_type = work.to_type();
-        
+
         // This metric tracks how long a work event has been in the queue
         metrics::observe_timer_vec(
             &metrics::BEACON_PROCESSOR_QUEUE_TIME,
--- a/beacon_node/beacon_processor/src/lib.rs
+++ b/beacon_node/beacon_processor/src/lib.rs
@@ -1684,7 +1684,7 @@ impl Drop for SendOnDrop {
         metrics::observe_timer_vec(
             &metrics::BEACON_PROCESSOR_QUEUE_TIME,
             &[self.work_type.clone().into()],
-            self.created_timestamp - Instant::now(),
+            Instant::now() - self.created_timestamp,
         );
         if let Err(e) = self.tx.try_send(self.work_type.clone()) {
             warn!(
--- a/beacon_node/beacon_processor/src/lib.rs
+++ b/beacon_node/beacon_processor/src/lib.rs
@@ -258,7 +258,7 @@ impl Default for BeaconProcessorConfig {
 // The channels necessary to instantiate a `BeaconProcessor`.
 pub struct BeaconProcessorChannels<E: EthSpec> {
     pub beacon_processor_tx: BeaconProcessorSend<E>,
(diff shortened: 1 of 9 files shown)


beacon_node/beacon_processor/src/lib.rs
```diff
@@ -517,12 +517,15 @@ pub struct GossipAggregatePackage<E: EthSpec> {
...
 
 impl<E: EthSpec> BeaconProcessorSend<E> {
-    pub fn try_send(&self, message: WorkEvent<E>) -> Result<(), TrySendError<WorkEvent<E>>> {
+    pub fn try_send(
+        &self,
+        message: WorkEvent<E>,
+    ) -> Result<(), TrySendError<(WorkEvent<E>, Instant)>> {
         let work_type = message.work_type();
-        match self.0.try_send(message) {
+        match self.0.try_send((message, Instant::now())) {
```
jimmygchen: Have you considered creating this timestamp on the receiver side? Or are you creating the timestamp to also account for the `mpsc` polling delay?

https://github.com/sigp/lighthouse/blob/c5cc64105634c96a8288f8ed0532b46b489aeb94/beacon_node/beacon_processor/src/lib.rs#L939

I don't imagine we do any heavy blocking work in the loop, so it should be pretty quick to poll these. I think it would be a bit cleaner to create the timestamp when the event is polled, and avoid having to pass a timestamp over mpsc, but I could be missing something here.

eserilev: you're right, f650b65 should simplify things further

eserilev:
Running this on `toga` doesn't really reveal anything interesting. The worker threads are never maxed out so the time between queueing the work and popping it from the queue is 0. This metric will only be interesting when the BN is very busy or when the machine its running on is resource constrained

eserilev:
https://two018.spesi.io/dashboard/snapshot/72K85ocxZq4NHnB4U7nBNmQIDbJUCesu

beacon_node/beacon_processor/src/lib.rs
```diff
@@ -1486,8 +1496,22 @@ impl<E: EthSpec> BeaconProcessor<E> {
...
+        idle_tx: mpsc::Sender<WorkType>,
+    ) {
         let work_id = work.str_id();
+        let work_type = work.to_type();
+
+        // This metric tracks how long a work event has been in the queue
+        metrics::observe_timer_vec(
+            &metrics::BEACON_PROCESSOR_QUEUE_TIME,
+            &[work_type.into()],
+            Instant::now() - created_timestamp,
```
michaelsproul, line 1512: Usually we would use `created_timestamp.elapsed()` for this, although I note that the docs say this could change to a panic in future (if now < created). The safest option seems to be `saturating_duration_since`

michaelsproul (approved):
LGTM

michaelsproul:
I don't understand how the wait time can be 0 on these Holesky nodes when their queue lengths are consistently 3k+. If messages are only queued for a few microseconds, it doesn't seem likely we'd catch the queue with so many entries :thinking:

PR #7809, opened 2025-07-29
Fix `http-api` tests not getting run on CI

jimmygchen:
## Issue Addressed

The `http-api` tests job incorrectly runs the `beacon_chain` tests due to an copy paste error in the `Makefile`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -197,7 +197,7 @@ test-beacon-chain-%:
 test-http-api: $(patsubst %,test-http-api-%,$(RECENT_FORKS))
 
 test-http-api-%:
-	env FORK_NAME=$* cargo nextest run --release --features "beacon_chain/fork_from_env,$(TEST_FEATURES)" -p http_api
+	env FORK_NAME=$* cargo nextest run --release --features "beacon_chain/fork_from_env" -p http_api
 
 
 # Run the tests in the `operation_pool` crate for all known forks.
--- a/Makefile
+++ b/Makefile
@@ -193,7 +193,7 @@ test-beacon-chain: $(patsubst %,test-beacon-chain-%,$(FORKS))
 test-beacon-chain-%:
 	env FORK_NAME=$* cargo nextest run --release --features "fork_from_env,slasher/lmdb,$(TEST_FEATURES)" -p beacon_chain
 
-# Run the tests in the `http_api` crate for all known forks.
+# Run the tests in the `http_api` crate for recent forks.
 test-http-api: $(patsubst %,test-http-api-%,$(RECENT_FORKS))
 
 test-http-api-%:
--- a/Makefile
+++ b/Makefile
@@ -197,7 +197,7 @@ test-beacon-chain-%:
 test-http-api: $(patsubst %,test-http-api-%,$(RECENT_FORKS))
 
 test-http-api-%:
-	env FORK_NAME=$* cargo nextest run --release --features "fork_from_env,slasher/lmdb,$(TEST_FEATURES)" -p http_api
+	env FORK_NAME=$* cargo nextest run --release --features "beacon_chain/fork_from_env,$(TEST_FEATURES)" -p http_api
 
 
 # Run the tests in the `operation_pool` crate for all known forks.
--- a/Makefile
+++ b/Makefile
@@ -193,8 +193,8 @@ test-beacon-chain: $(patsubst %,test-beacon-chain-%,$(FORKS))
 test-beacon-chain-%:
 	env FORK_NAME=$* cargo nextest run --release --features "fork_from_env,slasher/lmdb,$(TEST_FEATURES)" -p beacon_chain
 
-# Run the tests in the `beacon_chain` crate for all known forks.
-test-http-api: $(patsubst %,test-beacon-chain-%,$(RECENT_FORKS))
+# Run the tests in the `http_api` crate for all known forks.
+test-http-api: $(patsubst %,test-http-api-%,$(RECENT_FORKS))
 
 test-http-api-%:
 	env FORK_NAME=$* cargo nextest run --release --features "fork_from_env,slasher/lmdb,$(TEST_FEATURES)" -p http_api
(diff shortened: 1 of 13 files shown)


jimmygchen:
Looks like we got throttled on github runners, will re-run the workflow later.

Makefile
```diff
@@ -193,8 +193,8 @@ test-beacon-chain: $(patsubst %,test-beacon-chain-%,$(FORKS))
 test-beacon-chain-%:
 	env FORK_NAME=$* cargo nextest run --release --features "fork_from_env,slasher/lmdb,$(TEST_FEATURES)" -p beacon_chain
 
-# Run the tests in the `beacon_chain` crate for all known forks.
-test-http-api: $(patsubst %,test-beacon-chain-%,$(RECENT_FORKS))
+# Run the tests in the `http_api` crate for all known forks.
```
chong-he: Just a small nit here, because I see $(RECENT_FORKS) in the command
```suggestion
# Run the tests in the `http_api` crate for recent forks.
```

chong-he:
Looks like we need to fix the command in the `http_api` test as it is failing: https://github.com/sigp/lighthouse/actions/runs/16590347328/job/46930002515

michaelsproul:
A lot of the builder-related tests are broken, I think since the change we made to not blind local blocks.

I think we need to update the tests to actually use the `mock_builder` so that we don't error at the publication step.

michaelsproul:
These tests are a nightmare. The mock-builder doesn't really work for producing real blinded blocks that we can unblind, and I'm still dealing with config issues. I think I'll have to come back to this next week unless someone else wants to jump on it

michaelsproul:
I got the tests passing for Electra, but I just noticed that the node doesn't seem to actually be using the payload from the builder, the logs show:

> 2025-07-31T07:26:09.704725Z  INFO execution_layer: Requested blinded execution payload relay_fee_recipient="0x2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a" relay_response_ms=71
 local_fee_recipient="0x2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a" local_response_ms=7 parent_hash=0x957126b7c0f77ed65d51b20a59c42a3cbddb8612f7564c205e637420e19a4384
2025-07-31T07:26:09.704732Z  INFO execution_layer: Received local and builder payloads relay_block_hash=0x9c638276377f4f0ab1100f7cd7a06ae48d56c22ad07ddd7860e6d486de8f3897 l
ocal_block_hash=0x8efa54a29595aabacc12d1c7fb5742e7545aa1588b91fe863da1c6e4d8cd2caf parent_hash=0x957126b7c0f77ed65d51b20a59c42a3cbddb8612f7564c205e637420e19a4384
2025-07-31T07:26:09.705792Z  INFO execution_layer: Local block is more profitable than relay block local_value=10000000000000000 relay_value=10000000000000000 boosted_relay_value=10000000000000000 builder_boost_factor=None

beacon_node/execution_layer/src/lib.rs
```diff
@@ -2033,11 +2035,10 @@ impl<E: EthSpec> ExecutionLayer<E> {
                         relay_response_ms = duration.as_millis(),
                         ?block_root,
                         "Failed to submit blinded block to the builder"
-                    )
+                    );
+                    Err(e)
```
michaelsproul, line 2039: @pawanjay176 For blinded_blocks v2 was it intentional that we weren't returning an error if the builder returned an error?

I needed to add an error here to give the broadcast validation tests some hope of passing -- otherwise we just get 200s regardless of what the builder returns, and we have no way of checking block validity in the BN because the builder never returns the full block.

beacon_node/http_api/src/lib.rs
```diff
@@ -1663,18 +1663,29 @@ pub fn serve<T: BeaconChainTypes>(
...
         .and(warp_utils::json::json())
+        .and(consensus_version_header_filter)
         .and(task_spawner_filter.clone())
         .and(chain_filter.clone())
         .and(network_tx_filter.clone())
         .and(network_globals.clone())
         .then(
             move |validation_level: api_types::BroadcastValidationQuery,
-                  blinded_block: Arc<SignedBlindedBeaconBlock<T::EthSpec>>,
+                  blinded_block_json: serde_json::Value,
```
michaelsproul, line 1654: This was a bug: we weren't using fork-aware decoding, so we would decode Fulu blinded blocks as Electra

beacon_node/http_api/tests/broadcast_validation_tests.rs
```diff
@@ -317,8 +321,22 @@ pub async fn consensus_gossip() {
     let error_response: eth2::Error = response.err().unwrap();
 
     /* mandated by Beacon API spec */
-    assert_eq!(error_response.status(), Some(StatusCode::BAD_REQUEST));
-    assert_server_message_error(error_response, "BAD_REQUEST: Invalid block: StateRootMismatch { block: 0x0000000000000000000000000000000000000000000000000000000000000000, local: 0x253405be9aa159bce7b276b8e1d3849c743e673118dfafe8c7d07c203ae0d80d }".to_string());
+    if tester.harness.spec.is_fulu_scheduled() {
+        assert_eq!(
+            error_response.status(),
+            Some(StatusCode::INTERNAL_SERVER_ERROR)
+        );
```
michaelsproul, line 336: This still doesn't really work (the tests are failing), but I think for now I'm giving up on getting accurate error codes for Fulu, because it requires too much fussing around with mock_builder's warp code.

beacon_node/execution_layer/src/lib.rs
pawanjay176: Ohh I don't think it was intentional. We do return the error from the builder_client, so this seems like an oversight.

michaelsproul: awesome, we should be able to get these passing soon-ish then

michaelsproul:
Current status:

- Electra tests are partly failing, but the failures _might_ be straightforward
- Fulu tests are mostly failing, might require more invasive fixes

The edit-debug cycle is super slow here, which makes progress painful.

jimmygchen:
`broadcast_validation_tests.rs` are now all passing on electra, but `interactive_tests` are failing 💀 
will continue next week!

michaelsproul:
Closing in favour of:

- https://github.com/sigp/lighthouse/pull/7943

Thanks Manas!

Issue #7815, opened 2025-07-31
Instrument tracing spans in block processing and importing


jimmygchen:
Instrument the **block processing and importing** code paths using tracing spans to enable end-to-end visibility and to improve debugging and performance analysis.

jimmygchen:
Completed in #7816

PR #7816, opened 2025-07-31
Instrument tracing spans for block processing and import

jimmygchen:
## Issue Addressed

#7815 

- removes all existing spans, so some span fields that appear in logs like `service_name` may be lost.
- instruments a few key code paths in the beacon node, starting from **root spans** named below:

* Gossip block and blobs
  * `process_gossip_data_column_sidecar`
  * `process_gossip_blob`
  * `process_gossip_block`
* Rpc block and blobs
  * `process_rpc_block`
  * `process_rpc_blobs`
  * `process_rpc_custody_columns` 
* Rpc blocks (range and backfill) 
  * `process_chain_segment` 
* `PendingComponents` lifecycle
  * `pending_components`

To test locally:
* Run Grafana and Tempo with https://github.com/sigp/lighthouse-metrics/pull/57
* Run Lighthouse BN with `--telemetry-collector-url http://localhost:4317`

Some captured traces can be found here: https://hackmd.io/@jimmygchen/r1sLOxPPeg

Removing the old spans seem to have reduced the memory usage quite a lot - i think we were using them on long running tasks and too excessively: 
image


jimmygchen:
Just reading the doc on spans, I think some of the async functions are instrumented incorrectly, will revisit this on Monday!
https://docs.rs/tracing/latest/tracing/span/struct.Span.html

I've checked all `enter()` and `entered` usage and I believe they're good now. 

Now seeing good traces in Grafana:
image

pawanjay176:
This is brilliant. I'm gonna play around with this later today

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -823,6 +825,8 @@ impl<T: BeaconChainTypes> GossipVerifiedBlock<T> {
     /// on the p2p network.
     ///
     /// Returns an error if the block is invalid, or if the block was unable to be verified.
+    #[instrument(name = "verify_gossip_block", skip_all)]
+    #[allow(clippy::too_many_arguments)]
```
eserilev: I dont think we need this `#[allow]`

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -87,14 +88,18 @@ impl<E: EthSpec> PendingComponents<E> {
...
+        let _guard = self.span.clone().entered();
+        debug!("Block added to pending components");
         *self.get_cached_block_mut() = Some(block)
     }
 
     /// Inserts a blob at a specific index in the cache.
     ///
     /// Existing blob at the index will be replaced.
     pub fn insert_blob_at_index(&mut self, blob_index: usize, blob: KzgVerifiedBlob<E>) {
+        let _guard = self.span.clone().entered();
```
eserilev, line 100: nit: I think `in_scope()` makes a bit more sense here. from what I understand the `guard` pattern is used when making multiple ops vs `in_scope()` for a single op.

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -273,12 +289,15 @@ impl<E: EthSpec> PendingComponents<E> {
 
     /// Returns an empty `PendingComponents` object with the given block root.
     pub fn empty(block_root: Hash256, max_len: usize) -> Self {
+        let span = debug_span!(parent: None, "pending_components", %block_root);
+        let _guard = span.clone().entered();
```
eserilev, line 292: i dont think this guard is doing anything

eserilev (approved):
This is great! Just a few minor nits from me, LGTM

beacon_node/beacon_chain/src/block_verification.rs
jimmygchen: ah yeah, thanks - I'll remove this, must have been a copy paste error 😅

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: I believe they do the same thing in this case, and guard is just slightly shorter in this case.

The main reason to use `in_scope` is when you want to instrument some synchronous operations (could be one or multiple ops) in an async function - it ensures that the `guard` is not held across await points and screw up the trace.

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: It does _i think_ - it starts the span and exits it when the function ends, but keeps it _open_ (as we keep the span reference), so that it can be entered later.

This is the `start_time` for the span.

image

beacon_node/beacon_chain/src/block_verification.rs
jimmygchen: Fixed

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
eserilev: oh i see, in order to track the start time you're entering the span initially, got it

jimmygchen:
@eserilev thanks for the review 🙏 
I've got some more changes & refactor on another branch, but will merge this one and do a separate PR since this one is getting too big and already reviewed!

PR #7819, opened 2025-07-31
Filter dependencies from SSE logging

AgeManning:
In the tracing upgrade we included dependency logging in the SSE logs. 

This PR removes dependencies from the SSE logs.


jimmygchen (approved):
Good catch!

PR #7821, opened 2025-07-31
fix: Replace const panic with compile-time error in non_zero_usize

crStiv:
Replaced the panic!() in new_non_zero_usize with a compile-time error using an empty array index. This prevents accidental runtime panics in const contexts.

Also added a runtime-safe alternative try_new_non_zero_usize() that returns a Result instead of panicking.

- Safer handling of zero values

- Keeps const-eval working as expected

- Makes intent clearer for consumers of the API

### DIFF
--- a/consensus/types/src/non_zero_usize.rs
+++ b/consensus/types/src/non_zero_usize.rs
@@ -1,8 +1,21 @@
 use std::num::NonZeroUsize;
 
+/// Creates a new NonZeroUsize from a usize value.
+///
+/// This function will cause a compile-time error if used with a zero value.
+/// For runtime values, use `try_new_non_zero_usize()` instead.
 pub const fn new_non_zero_usize(x: usize) -> NonZeroUsize {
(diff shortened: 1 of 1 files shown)


consensus/types/src/non_zero_usize.rs
```diff
@@ -1,8 +1,21 @@
...
+///
+/// This function will cause a compile-time error if used with a zero value.
+/// For runtime values, use `try_new_non_zero_usize()` instead.
 pub const fn new_non_zero_usize(x: usize) -> NonZeroUsize {
     match NonZeroUsize::new(x) {
         Some(n) => n,
-        None => panic!("Expected a non zero usize."),
+        None => {
+            // This creates a compile-time error for zero values by indexing an empty array
+            // This is safer than panic! and will catch errors at compile time
```
michaelsproul, line 12: This is already a compile time error because it's a const fn.

michaelsproul:
Closing as suspected bot PR + unnecessary.

PR #7822, opened 2025-07-31
fix: Replace panic with proper error handling in CLI flag conflict

crStiv:
Replaces a panic!() with a graceful error return when both --network and --testnet-dir flags are used together.

- Avoids crashing on invalid CLI usage

- Returns a clear error message instead

- Follows Rust best practices for error handling in CLI tools


michaelsproul:
This panic is not reachable, clap validates the arguments before this point. If you disagree, please reopen with a test case that demonstrates the panic.

PR #7825, opened 2025-08-04
Adjust DA checker cache size

jimmygchen:
## Issue Addressed

The current `OVERFLOW_LRU_CAPACITY` of `1024` seems a bit excessive now we rarely store more than 1 `PendingComponents` (under normal networking components). Additionally given the blob count increases, the max size of `PendingComponents` has also increased and is expected to increase further. 

This PR brings the max capacity of the cache down to `64`, which should be more than enough headroom but also give us  better protection from the network.

### DIFF
--- a/beacon_node/beacon_chain/src/data_availability_checker.rs
+++ b/beacon_node/beacon_chain/src/data_availability_checker.rs
@@ -38,12 +38,19 @@ use crate::observed_data_sidecars::ObservationStrategy;
 pub use error::{Error as AvailabilityCheckError, ErrorCategory as AvailabilityCheckErrorCategory};
 use types::non_zero_usize::new_non_zero_usize;
 
-/// The LRU Cache stores `PendingComponents` which can store up to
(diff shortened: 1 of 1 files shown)


michaelsproul:
We should probably also increase `STATE_LRU_CAPACITY_NON_ZERO` from 2, as I suspect that cache is sometimes causing us expensive cache misses?

jimmygchen:
> We should probably also increase `STATE_LRU_CAPACITY_NON_ZERO` from 2, as I suspect that cache is sometimes causing us expensive cache misses?

Do you have any number in mind?
I've mentioned the one occurrence on my node today (due to a race condition documented in #6439) and the reconstruction took 77ms. Probably not too bad, but I'm not sure what the worst case look like.

michaelsproul:
I was thinking the state cache size could be at least 16-32

jimmygchen:
Yeah sounds good, updated to 32.

michaelsproul (approved):
LGTM

PR #7826, opened 2025-08-05
Fix Clippy for Rust 1.90 beta

michaelsproul:
## Issue Addressed

Fix Clippy for recently released Rust 1.90 beta. There may be more changes required when Rust 1.89 stable is released in a few days, but possibly not :crossed_fingers:


macladson (approved):
LGTM!

PR #7827, opened 2025-08-05
Refactor `consensus/types`

macladson:
## Proposed Changes

Organize and categorize `consensus/types` into modules based on their relation to key consensus structures/concepts.
This is a precursor to a sensible public interface.

While this refactor is very opinionated, I am open to suggestions on module names, or type groupings if my current ones are inappropriate.

## Reviewer Guide

While this is a 2000+ line diff, I don't believe the review should be difficult as the vast majority of the changes are from the following:
1. Restructuring internal file imports (usually from something like `use crate::*` to a more explicit and verbose import. The imports will take the order `std, external, internal`  with a new line between each category of import. For example:
```rust
use std::collections::HashSet;

use serde::{Decode, Encode};

use crate::block::BeaconBlockBody;
```

This format is one of the available options in nightly `rustfmt`: https://rust-lang.github.io/rustfmt/?version=latest&search=#group_imports and is my personal preference.

2. Removing generic `Error` names in favour of more specific error names e.g. `BeaconStateError`.
3. Reorganizing exports into various `mod.rs` files. `lib.rs` can then just import each module directly.
4. Various temporary re-exports and facade modules to maintain backwards compatibility (see below).

### Import Structure

To (mostly) keep backwards compatibility with the rest of Lighthouse (and to stop this PR from exploding in size and scope), we export all public types at the root level. For example, you can do `use types::BeaconBlock`. There is also various facade modules to keep the original import structure. These can be removed 1-by-1 over time in small discreet PRs.

For us to transition the `types` crate to a public library, it is my belief that we should require imports be at the module level (except for very core types such as `Hash256`, `EthSpec`, `Slot`, etc). So we should instead require `use types::block::BeaconBlock`. In my view, this helps keep the namespaces and docs cleaner and enables easier IDE type discovery. Perhaps more subjectively, I think that it also aids DevEx as it becomes easier to see which types are being imported at a glance since related types are necessarily grouped together.

Therefore, we should update the rest of the `types` imports across Lighthouse to do the following:
- Remove all instances of `use types::*;`
- Swap all root level imports to use their module variants if applicable
- Remove the root-level re-exports from `types/src/lib.rs`

Similarly, we should remove any re-exports of the `types` crate in other crates (`eth2`, `beacon_chain`) which I don't believe is worth it. There are a plethora of situations where the same file will export types from multiple versions of the same crate through different re-exports. We should avoid situations like this and import only from `types` directly.

### External Crate Re-Exports

We are re-exporting several crates inside `types`.

```rust
// Temporary re-exports to maintain backwards compatibility for Lighthouse.
pub use bls::{
    get_withdrawal_credentials, AggregatePublicKey, AggregateSignature, Error as BlsError, Keypair,
    PublicKey, PublicKeyBytes, SecretKey, Signature, SignatureBytes, PUBLIC_KEY_BYTES_LEN,
    SIGNATURE_BYTES_LEN,
};
pub use context_deserialize::ContextDeserialize;
pub use context_deserialize_derive::context_deserialize;
pub use fixed_bytes::FixedBytesExtended;
pub use milhouse::{self, List, Vector};
pub use ssz_types::{typenum, typenum::Unsigned, BitList, BitVector, FixedVector, VariableList};
pub use superstruct::superstruct;
```

My opinion is that we should eventually remove all these re-exports (or at least most of them).

### Feature Gates

I think this refactor enables potential feature gating of certain groups of types. The most obvious one being `light_client`. The module is entirely distinct and not relied upon by any other module (outside of certain consts which could easily be moved).


macladson:
There seems to be some issue with the genesis sync tests but I don't believe they are specific to this PR

consensus/types/src/data_column_subnet_id.rs
```diff
@@ -65,15 +68,3 @@ impl From<&DataColumnSubnetId> for u64 {
...
-pub enum Error {
-    ArithError(ArithError),
-    InvalidCustodySubnetCount(u64),
-}
-
-impl From<ArithError> for Error {
-    fn from(e: ArithError) -> Self {
-        Error::ArithError(e)
-    }
-}
```
macladson, line 83: I removed this error type since from what I could tell it was unused

common/eth2/src/beacon_response.rs
macladson, line 1: I've moved these Response types to the `eth2` crate since it's a much more natural place for them to exist. I've added some re-export hackery to ensure the diff stays small, but we can tidy this up when we eventually refactor this crate

consensus/types/src/beacon_response.rs
```diff
@@ -1,12 +1,8 @@
-use crate::{ContextDeserialize, ForkName};
+use context_deserialize::ContextDeserialize;
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Deserializer, Serialize};
 use serde_json::value::Value;
-
-pub trait ForkVersionDecode: Sized {
```
pawanjay176, line 6: Any reason why we can't have ForkVersionDecode in `eth2` crate as well?

common/eth2/src/types.rs
```diff
@@ -19,10 +19,15 @@ use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
 use test_random_derive::TestRandom;
-use types::beacon_block_body::KzgCommitments;
 use types::test_utils::TestRandom;
 pub use types::*;
 
+// Temporary module and re-export hack to expose old `consensus/types` via `eth2/types`.
```
pawanjay176: Can we add a TODO for this to make searching for it easier

consensus/types/Cargo.toml
```diff
@@ -73,6 +73,10 @@ paste = { workspace = true }
 state_processing = { workspace = true }
 tokio = { workspace = true }
 
+[lints.clippy]
+module_inception = "allow"
+module_name_repetitions = "allow"
```
pawanjay176: Not sure if we need this anymore? lint passes without it as well

consensus/types/src/attestation/indexed_attestation.rs
```diff
@@ -208,9 +212,10 @@ impl<E: EthSpec> Hash for IndexedAttestation<E> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::MainnetEthSpec;
-    use crate::slot_epoch::Epoch;
-    use crate::test_utils::{SeedableRng, XorShiftRng};
+    use crate::{
+        Epoch, MainnetEthSpec,
```
pawanjay176: Should this be core::{Epoch, MainnetEthSpec}?

consensus/types/src/block/beacon_block.rs
```diff
@@ -865,7 +886,10 @@ impl fmt::Display for BlockImportSource {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::test_utils::{SeedableRng, XorShiftRng, test_ssz_tree_hash_pair_with};
+    use crate::{
+        MainnetEthSpec,
```
pawanjay176:
```suggestion
        core::MainnetEthSpec,
```

consensus/types/src/block/beacon_block_body.rs
```diff
@@ -1100,22 +1141,16 @@ impl<'de, E: EthSpec, Payload: AbstractExecPayload<E>> ContextDeserialize<'de, F
...
-    let commitments_joined = commitment_strings.join(", ");
-    let surrounded_commitments = format!("[{}]", commitments_joined);
-    surrounded_commitments
-}
-
 #[cfg(test)]
 mod tests {
     mod base {
         use super::super::*;
+        use crate::MainnetEthSpec;
```
pawanjay176:
```suggestion
        use crate::core::MainnetEthSpec;
```

consensus/types/src/block/beacon_block_body.rs
```diff
@@ -1100,22 +1141,16 @@ impl<'de, E: EthSpec, Payload: AbstractExecPayload<E>> ContextDeserialize<'de, F
...
         ssz_and_tree_hash_tests!(BeaconBlockBodyBase<MainnetEthSpec>);
     }
     mod altair {
         use super::super::*;
+        use crate::MainnetEthSpec;
```
pawanjay176:
```suggestion
        use crate::core::MainnetEthSpec;
```

consensus/types/src/block/signed_beacon_block.rs
```diff
@@ -919,6 +944,7 @@ pub mod ssz_tagged_signed_beacon_block_arc {
 #[cfg(test)]
 mod test {
     use super::*;
+    use crate::{MainnetEthSpec, block::EmptyBlock};
```
pawanjay176:
```suggestion
    use crate::{core::MainnetEthSpec, block::EmptyBlock};
```

consensus/types/src/core/chain_spec.rs
```diff
@@ -2581,6 +2589,7 @@ mod tests {
 #[cfg(test)]
 mod yaml_tests {
     use super::*;
+    use crate::MinimalEthSpec;
```
pawanjay176:
```suggestion
    use crate::core::MinimalEthSpec;
```

consensus/types/src/core/application_domain.rs
pawanjay176, line 1: maybe this belongs in the builder module instead of core?

consensus/types/src/data/mod.rs
pawanjay176, line 1: `data` seems a bit odd name for this module but I can't think of a better alternative 😅 Were there any other contenders?

consensus/types/src/deposit/deposit.rs
```diff
@@ -1,12 +1,12 @@
...
-use crate::*;
+use context_deserialize::context_deserialize;
 use serde::{Deserialize, Serialize};
 use ssz_derive::{Decode, Encode};
-use ssz_types::typenum::U33;
+use ssz_types::{FixedVector, typenum::U33};
 use test_random_derive::TestRandom;
 use tree_hash_derive::TreeHash;
 
+use crate::{Hash256, deposit::DepositData, fork::ForkName, test_utils::TestRandom};
```
pawanjay176:
```suggestion
use crate::{core::Hash256, deposit::DepositData, fork::ForkName, test_utils::TestRandom};
```

consensus/types/src/execution/execution_block_hash.rs
```diff
@@ -1,10 +1,11 @@
...
-use crate::test_utils::TestRandom;
+use std::fmt;
+
+use fixed_bytes::FixedBytesExtended;
 use rand::RngCore;
 use serde::{Deserialize, Serialize};
 use ssz::{Decode, DecodeError, Encode};
-use std::fmt;
+
+use crate::{Hash256, test_utils::TestRandom};
```
pawanjay176:
```suggestion
use crate::{core::Hash256, test_utils::TestRandom};
```

consensus/types/src/execution/execution_block_header.rs
```diff
@@ -17,10 +17,12 @@
 // LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
-use crate::{Address, EthSpec, ExecutionPayloadRef, Hash64, Hash256, Uint256};
 use alloy_rlp::RlpEncodable;
+use fixed_bytes::Uint256;
 use metastruct::metastruct;
 
+use crate::{Address, Hash64, Hash256, core::EthSpec, execution::ExecutionPayloadRef};
```
pawanjay176:
```suggestion
use crate::{core::{Address, Hash64, Hash256, EthSpec}, execution::ExecutionPayloadRef};
```

consensus/types/src/fork/fork_context.rs
```diff
@@ -152,8 +156,7 @@ impl ForkContext {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::MainnetEthSpec;
-    use crate::chain_spec::{BlobParameters, BlobSchedule};
+    use crate::{BlobParameters, BlobSchedule, MainnetEthSpec};
```
pawanjay176:
```suggestion
    use crate::core{BlobParameters, BlobSchedule, MainnetEthSpec};
```

consensus/types/src/light_client/error.rs
pawanjay176, line 1: Maybe the error type file can be named just error.rs since its already within the module?

consensus/types/src/state/historical_batch.rs
```diff
@@ -26,6 +31,7 @@ pub struct HistoricalBatch<E: EthSpec> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::MainnetEthSpec;
```
pawanjay176:
```suggestion
    use crate::core::MainnetEthSpec;
```

consensus/types/Cargo.toml
```diff
@@ -73,6 +73,10 @@ paste = { workspace = true }
 state_processing = { workspace = true }
 tokio = { workspace = true }
 
+[lints.clippy]
+module_inception = "allow"
```
pawanjay176, line 77: I'm not entirely convinced we need to do this. I sorta agree with the [lint](https://rust-lang.github.io/rust-clippy/master/index.html#module_inception) that having the option to import as use x::x; isn't great.
Is there any specific disadvantage in having the contents of x.rs in consensus/types/x/mod.rs ?

pawanjay176 (changes requested):
This is looking great 🎉  Thanks a lot for leading this effort and sorry for the delay in reviewing. 

Just a couple minor nits and a suggestion to adhere to `module_inception` lint. Lmk what you think

consensus/types/src/beacon_response.rs
pawanjay176: Ohh its because we need this in multiple places in types. Ignore me

consensus/types/src/core/application_domain.rs
macladson: The reason I didn't consider putting it in `builder` was because it gets used in `chain_spec.rs` so keeping it in `core` means we can potentially feature-gate `builder` types in the future

consensus/types/src/data/mod.rs
macladson: I called it `data` as in "data layer", I don't have any other alternatives though

consensus/types/Cargo.toml
macladson: There's not really a "disadvantage", but personally I prefer when `mod.rs` and `lib.rs` only contain meta code, like managing exports and module definitions. My perspective is that these files should generally never contain types or logic. But that perspective kind of leads to this module inception because the most natural name for the main file is the same as the module name.

As for `use my_crate::module::module::SomeType`, this can only occur within the crate itself, as a consumer of the crate you couldn't actually do this since the second (nested) module is private so I personally don't see it as a significant problem.

For example, you _could_ do:
`use crate::fork::fork::Fork;`
But not:
`use types::fork::fork::Fork;`
We would only ever allow:
`use types::fork::Fork;`

consensus/types/src/light_client/error.rs
macladson: Yeah makes sense to me :+1:

consensus/types/Cargo.toml
pawanjay176: Ohh yeah you're right about this only being a problem for within the crate. I'm fine either way in that case. 
Will leave this comment to see if anyone else has strong opinions

consensus/types/src/data/mod.rs
eserilev: in the future with epbs, i assume wed have payload envelopes under `data` as well? maybe `block_data` could be a potential name here? not super opinionated about this, but `data` feels a bit too generic

eserilev: or maybe wed just have like a `payload` submodule or something for the payload envelope. `data` seems fine to me

eserilev (approved):
Thanks for this, I can tell that there was a lot of thought put into these changes, its very much appreciated. I'm on board with the changes, we could always tweak things in subsequent PR's if needed

LGTM!

consensus/types/Cargo.toml
pawanjay176: I'm convinced this is better now after talking to @michaelsproul . Lets keep it.

pawanjay176 (approved):
LFG🚀 🚀

michaelsproul (approved):
Happy to merge too, thanks Mac for sticking with this, and Pawan for the thorough review as well!

kevaundray:
My favourite type of PRs!

Issue #7830, opened 2025-08-06
Instrument tracing spans for RPC data columns requests and responeses


jimmygchen:
Instrument the RPC request and response handling code paths using tracing spans to enable end-to-end visibility of data column rpc interactions with peers, e.g. what rpc requests each client are sending, and their success rate of serving requests.

### Questions to answer
- [x] what columns are clients requesting? are any extra non-advertise columns requested?
- [x] what columns are clients responding with? are there any missing columns
- [x] what is the duration and success rate for RPC responses

### fields to include
- `peer_id`
- `client`

### `active_custody_request`

Traces the entire duration of a custody request, which contains one to many `batch_columns_req` spans, each mapped to an actual RPC request with a `peer_id`.

jimmygchen:
Data column RPC paths instrumented in #7831

PR #7831, opened 2025-08-06
Implement tracing spans for data columm RPC requests and responses

jimmygchen:
## Issue Addressed

#7830 

### `syncing_chain`

Span to group events by syncing chain and add context to logs.

### `outgoing_range_request`

Traces a collection of outgoing range requests for a single slot range, tracking the requester (backfill or range sync).

Contains child spans that tracks individual requests to peers:
- `outgoing_blocks_by_range`
- `outgoing_columns_by_range`
- `outgoing_blobs_by_range`

image

### `outgoing_custody_request`

Traces the span of a collection of data column requests for a given `block_root`, which contains one to many `batch_columns_req` spans, each mapped to an actual RPC request with a `peer_id`, `client`, `missing_column_indexes` (recorded on response).

image

### `handle_data_columns_by_root_request`

Traces an incoming `DataColumnsByRoot` request, with fields `peer_id`, `non_custody_indices`, `client`.

### `handle_data_columns_by_range_request`

Traces an incoming `DataColumnsByRange` request, with fields `peer_id`, `non_custody_indices`, `client`.

**NOTE:** Outgoing `BlocksByRoot` and `BlobsByRoot` not implemented.


beacon_node/lighthouse_tracing/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+/// - Are root spans with names matching the predefined allowed list
+pub struct AllowedRootSpanFilter(&'static [&'static str]);
+
+impl AllowedRootSpanFilter {
+    pub fn new(allowed_names: &'static [&'static str]) -> Self {
+        AllowedRootSpanFilter(allowed_names)
+    }
+}
+
+impl<S> Filter<S> for AllowedRootSpanFilter
```
jimmygchen: Ignore this for now, this vibe coded impl is wrong and doesnt work (it's not actually checking for parent span)

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -826,7 +826,7 @@ impl<T: BeaconChainTypes> GossipVerifiedBlock<T> {
     /// on the p2p network.
     ///
     /// Returns an error if the block is invalid, or if the block was unable to be verified.
-    #[instrument(name = "verify_gossip_block", skip_all)]
+    #[instrument(name = "verify_gossip_block", skip_all, fields(block_root = tracing::field::Empty))]
```
jimmygchen, line 829: Fields need to be declared when span is created.

beacon_node/lighthouse_tracing/src/lib.rs
jimmygchen: Deleting it for now

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -990,14 +1090,51 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                .chain
+                .data_availability_checker
+                .custody_context()
+                .custody_columns_for_epoch(epoch_opt, &self.chain.spec);
+            requested_indices
+                .iter()
+                .filter(|subnet_id| !custody_columns.contains(subnet_id))
+                .collect::<Vec<_>>()
+        };
+        span.record("non_custody_indices", field::debug(non_custody_indices));
```
jimmygchen: This field is used to identify if peers are sending requests on columns we don't custody.

jimmygchen:
I've done some clean up and I think this is now ready for review.

Block lookup is not fully instrumented - so far I've only instrument custody columns by roots, I think eventually it would be nice to group all by roots calls for a single root under a root span. However this PR is getting huge, so i think it would be good to get some feedback and do the rest in a separate PR.

@eserilev would you mind reviewing this please?

eserilev (approved):
LGTM!

Issue #6647, opened 2024-12-02
Beacon block streamer test fails on macOS in debug mode


alecalve:
## Description

This is happening on a MacBook pro.

## Version

Lighthouse: `v6.0.0`
Cargo: `cargo 1.82.0 (8f40fc59f 2024-08-21)`
Rustup:

```
active toolchain
----------------

stable-aarch64-apple-darwin (default)
rustc 1.82.0 (f6e511eec 2024-10-15)
```

## Present Behaviour

Initial setup:
```
git clone git@github.com:sigp/lighthouse.git lighthouse-repro
cd lighthouse-repro
git checkout v6.0.0
```
### First failing test

Link to [code](https://github.com/sigp/lighthouse/blob/c042dc14d74352512b7632e0ee6ec07f1aa26b3a/beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs#L1000)

Command:
```
cargo test --package beacon_chain --lib -- data_availability_checker::overflow_lru_cache::test::overflow_cache_test_state_cache --show-output
```

Output:
```
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.91s
     Running unittests src/lib.rs (target/debug/deps/beacon_chain-d9a1d435f251ff26)

running 1 test

thread 'tokio-runtime-worker' has overflowed its stack
fatal runtime error: stack overflow
error: test failed, to rerun pass `-p beacon_chain --lib`
```

### Second failing test

Link to [code](https://github.com/sigp/lighthouse/blob/c042dc14d74352512b7632e0ee6ec07f1aa26b3a/beacon_node/beacon_chain/src/beacon_block_streamer.rs#L745)

Command:
```
cargo test --package beacon_chain --lib -- beacon_block_streamer --show-output
```

Output:
```
    Finished `test` profile [unoptimized + debuginfo] target(s) in 1.02s
     Running unittests src/lib.rs (target/debug/deps/beacon_chain-d9a1d435f251ff26)

running 1 test

thread 'tokio-runtime-worker' has overflowed its stack
fatal runtime error: stack overflow
error: test failed, to rerun pass `-p beacon_chain --lib`
```

## Expected Behaviour

The tests should complete.

michaelsproul:
Duplicate of:

- https://github.com/sigp/lighthouse/issues/6120

[...]

PR #7836, opened 2025-08-06
Integration tests ergonomics

### LINKED ISSUE #7785: Move integration tests submodules to their own directory
Integration tests submodules should be defined in their own directory. For example, if we want to define some common functions for a set of integration tests. The common module should be defined as: `crate/tests/common/mod.rs`, instead of `crate/tests/common.rs`.

This is explained in [The rust book, ch11-03](https://doc.rust-lang.org/book/ch11-03-test-organization.html#submodules-in-integration-tests). Or more succinctly in [rust by example](https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html).

Besides just adhering to standards, this has some practical implications. What motivated me to open this issue has to do with rust-analyzer. If you want to find references to a function defined in `tests/common.rs`, rust-analyzer will only list the ones in the current file. Eg: https://github.com/sigp/lighthouse/pull/7836#issuecomment-3162142439

### List of all files that follow this pattern

I *think* this command gives the full list of submodules of integration tests.

```txt
$ find . -path '*/tests/*' -name '*.rs' | xargs grep 'mod .*;'
./validator_client/slashing_protection/tests/main.rs:mod interop;
./validator_client/slashing_protection/tests/main.rs:mod migration;
./beacon_node/network/src/sync/tests/mod.rs:mod lookups;
./beacon_node/network/src/sync/tests/mod.rs:mod range;
./beacon_node/beacon_chain/tests/main.rs:mod attestation_production;
./beacon_node/beacon_chain/tests/main.rs:mod attestation_verification;
./beacon_node/beacon_chain/tests/main.rs:mod bellatrix;
./beacon_node/beacon_chain/tests/main.rs:mod block_verification;
./beacon_node/beacon_chain/tests/main.rs:mod capella;
./beacon_node/beacon_chain/tests/main.rs:mod events;
./beacon_node/beacon_chain/tests/main.rs:mod op_verification;
./beacon_node/beacon_chain/tests/main.rs:mod payload_invalidation;
./beacon_node/beacon_chain/tests/main.rs:mod rewards;
./beacon_node/beacon_chain/tests/main.rs:mod schema_stability;
./beacon_node/beacon_chain/tests/main.rs:mod store_tests;
./beacon_node/beacon_chain/tests/main.rs:mod sync_committee_verification;
./beacon_node/beacon_chain/tests/main.rs:mod tests;
./beacon_node/beacon_chain/tests/main.rs:mod validator_monitor;
./beacon_node/lighthouse_network/tests/rpc_tests.rs:mod common;
./beacon_node/http_api/tests/main.rs:pub mod broadcast_validation_tests;
./beacon_node/http_api/tests/main.rs:pub mod fork_tests;
./beacon_node/http_api/tests/main.rs:pub mod interactive_tests;
./beacon_node/http_api/tests/main.rs:pub mod status_tests;
./beacon_node/http_api/tests/main.rs:pub mod tests;
./lighthouse/tests/main.rs:mod account_manager;
./lighthouse/tests/main.rs:mod beacon_node;
./lighthouse/tests/main.rs:mod boot_node;
./lighthouse/tests/main.rs:mod exec;
./lighthouse/tests/main.rs:mod validator_client;
./lighthouse/tests/main.rs:mod validator_manager;
```

danielrachi1:
## Issue Addressed

Fixes #7785

### Main pattern

The "main pattern" refers to integration tests which:

1. Have multiple files.
2. Have a single point of entry, a `main.rs` file.
3. All tests are listed as submodules of that `main.rs` file.

In practice, this is achieved by:

1. Change `Cargo.toml` on each package to include:

```toml
[package]
autotests = false

[[test]]
name = "package_name_tests"
path = "tests/main.rs"
```

2. Create a `tests/main.rs` file on each package listing all other files
on `tests` as a module.

### `#![cfg(not(debug_assertions))]`

Some integration tests are configured to compile only on `--release`. This makes rust analyzer ignore them, and the IDEs to not offer LSP services on those files.

Setting `debug_assertions` to false, via `rust-analyzer.cargo.cfgs` fixes this.

## Proposed Changes

- [x] Update all integration tests with >1 files to follow the `main` pattern.
  - [x] `crypto/eth2_key_derivation/tests`
  - [x] `crypto/eth2_keystore/tests`
  - [x] `crypto/eth2_wallet/tests`
  - [x] `slasher/tests`
  - [x] `common/eth2_interop_keypairs/tests`
  - [x] `beacon_node/lighthouse_network/tests`
- [x] Set `debug_assertions` to false on `.vscode/settings.json`.
- [x] Document how to make rust analyzer work on integration tests files. In `book/src/contributing_setup.md`

---

Tracking a `rust-analyzer.toml` with settings like the one provided in `.vscode/settings.json` would be nicer. But this is not possible yet. For now, that config should be a good enough indicator for devs using editors different to VSCode.


michaelsproul:
I think this is going beyond moving _common_ code to submodules – it seems you're moving _all_ integration test code into submodules, which is not recommended in Rust docs AFACIT?

In [Rust by Example](https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html) the "File with test: tests/integration_test.rs" doesn't get moved to `tests/integration_tests/mod.rs`.

We also have another mechanism already in place to prevent recompilation. The `main.rs` file in each integration test directory prevents each file being compiled separately, and also allows for common modules, e.g.

https://github.com/sigp/lighthouse/blob/cfb1f7331064b758c6786e4e1dc15507af5ff5d1/lighthouse/tests/main.rs#L1-L8

Here the `exec` module doesn't contain any tests, it's just shared code.

Actually I think this might be the source of the confusion: it looks like you grepped for `mod X` and then moved all these modules. But most of these modules are themselves tests, because of the `main.rs` pattern.

michaelsproul:
Can you elaborate more on the rust-analyzer issue? What's an example file and a click-through that doesn't work with rust analyzer?

danielrachi1:
> Actually I think this might be the source of the confusion: it looks like you grepped for mod X and then moved all these modules. But most of these modules are themselves tests, because of the main.rs pattern.

Yes, that's exactly what happened. I got confused, indeed.

Still, Do you agree that *common* code should be moved to its own submodule/directory?

> Can you elaborate more on the rust-analyzer issue? What's an example file and a click-through that doesn't work with rust analyzer?

The file: `beacon_node/lighthouse_network/tests/common.rs` contains a function: `build_tracing_subscriber`. This function is used by `beacon_node/lighthouse_network/tests/rpc_tests.rs` in various tests.

If I go to `common.rs` right-click the function name and click "Find All References" I only get one result, the function definition in that file:

image

If I move common to `common/mod.rs`, all the references appear:

image

I'm using VSCode as an example here. The same thing happens in NeoVim.

michaelsproul:
> Still, Do you agree that common code should be moved to its own submodule/directory?

No, not necessarily. I think it's more consistent if we use the `main.rs` pattern everywhere, seeing as that has other benefits (compiling and linking one binary for all integration tests is faster than compiling and linking multiple binaries).

Can you check if adding `main.rs` to `beacon_node/lighthouse_network/tests/` fixes the rust analyzer issue?

michaelsproul:
I think `beacon_node/network/src/sync/tests` and `beacon_node/lighthouse_network` are the only crates without a `main.rs`? And I don't think the tests in `beacon_node/network/src/sync/tests` are even integration tests, because they're in the `src` directory. And they already have a `mod.rs` anyway.

danielrachi1:
> No, not necessarily. I think it's more consistent if we use the main.rs pattern everywhere, seeing as that has other benefits (compiling and linking one binary for all integration tests is faster than compiling and linking multiple binaries).

I didn't fully understand the main.rs pattern, but now I do, I hope.

Briefly:
1. Have a `tests/main.rs` file with all the tests listed as submodules.
2. In `Cargo.toml`: Set `autotests = false` and a single `[[test]]` section pointing to `tests/main.rs`. This tells cargo to not treat every file in `tests/` as a separate crate.
3. Now there's a single integration test, split into multiple modules. As suggested in the [cargo reference](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#integration-tests).

---

> Can you check if adding main.rs to beacon_node/lighthouse_network/tests/ fixes the rust analyzer issue?

I created: `beacon_node/lighthouse_network/tests/main.rs`:

```rs
// Not sure if necessary for this test
// #![cfg(not(debug_assertions))]

mod rpc_tests;
mod common;
```

Changed `beacon_node/lighthouse_network/Cargo.toml` to include:

```toml
[package]
autotests = false

[[test]]
name = "lighthouse_network_tests"
path = "tests/main.rs"
```

The test doesn't compile unless I move `common.rs` to its own directory:

```
$ cargo test --test lighthouse_network_tests
   Compiling lighthouse_network v0.2.0 (/home/daniel/code/lighthouse/beacon_node/lighthouse_network)
error[E0583]: file not found for module `common`
 --> beacon_node/lighthouse_network/tests/rpc_tests.rs:3:1
  |
3 | mod common;
  | ^^^^^^^^^^^
  |
  = help: to create the module `common`, create file "beacon_node/lighthouse_network/tests/rpc_tests/common.rs" or "beacon_node/lighthouse_network/tests/rpc_tests/common/mod.rs"
  = note: if there is a `mod common` elsewhere in the crate already, import it with `use crate::...` instead

For more information about this error, try `rustc --explain E0583`.
error: could not compile `lighthouse_network` (test "lighthouse_network_tests") due to 1 previous error
```

If I take the compiler's suggestion and create file `beacon_node/lighthouse_network/tests/rpc_tests/common/mod.rs` (+ fix the `use` statements):

- [x] The test compiles and runs normally.
- [x] Rust analyzer works on `common/mod.rs`

---

> I think beacon_node/network/src/sync/tests and beacon_node/lighthouse_network are the only crates without a main.rs? And I don't think the tests in beacon_node/network/src/sync/tests are even integration tests, because they're in the src directory. And they already have a mod.rs anyway.

This command lists all dirs that:
1. Are named `tests`.
2. Don't have `main.rs` inside them.
3. Its parent has `Cargo.toml`.

```
$ find . -type d -name "tests" -exec sh -c '[ ! -f "$1/main.rs" ] && [ -f "$(dirname "$1")/Cargo.toml" ] && echo "$1"' _ {} \;
./crypto/bls/tests
./crypto/eth2_key_derivation/tests
./crypto/eth2_keystore/tests
./crypto/eth2_wallet/tests
./slasher/tests
./consensus/fork_choice/tests
./consensus/context_deserialize_derive/tests
./testing/ef_tests/tests
./common/eth2_interop_keypairs/tests
./common/validator_dir/tests
./beacon_node/tests
./beacon_node/http_metrics/tests
./beacon_node/lighthouse_network/tests
./lighthouse/environment/tests
```

Some of them have a single file named `tests.rs`.

---

Okay, I think I should do this:

1. Scrap my previous idea. Literally delete the commit in this PR.
2. Implement the main.rs pattern, as described earlier, for... some of the directories listed above. Not sure about which ones to skip. Eg. Probably not necessary for the ones with a single `tests.rs` file.

I don't think we can avoid placing modules named `common` in `common/mod.rs`, as shown in the compiler error above. But I don't think this is an issue. The main.rs pattern and `common/mod.rs` are compatible. The tests are compiled as one binary, and rust analyzer works.

michaelsproul:
> I don't think we can avoid placing modules named common in common/mod.rs, as shown in the compiler error above. But I don't think this is an issue.

We can, you can remove the `mod common` from `rpc_tests.rs` and `use crate::common`. That's what the other tests do, e.g.

https://github.com/sigp/lighthouse/blob/cfb1f7331064b758c6786e4e1dc15507af5ff5d1/lighthouse/tests/beacon_node.rs#L1

For each module we only want one `mod X` across the whole project. So put it in `main.rs` and then delete it from `rpc_tests.rs`. That should work with rust analyzer too maybe?

macladson:
Perhaps my IDE is non-standard but I think there's a few things going on, one of which is `#![cfg(not(debug_assertions))]`. When this crate-level attribute is present, `rust-analyzer` won't process the file at all.  For example, if I try to find all references on the code in `lighthouse/tests/exec.rs` I find no results because the `mod exec;` in `main.rs` is gated behind the conditional.

When I do the same for the `lighthouse_network` tests, I get the same results that @danielrachi1 was seeing. I believe the reason for this is that since there is no `main.rs` and thus no `#![cfg(not(debug_assertions))]`, `rust-analyzer` is able to parse each file, but due to quirks in its analysis of integration test crates it is unable to link those files together which is why we only get references for each files own use of the function.

As soon as we move the code to `common/mod.rs`, `rust-analyzer` is able to figure it out, because (I think) the module is no longer marked as a separate test crate and thus `rust-analyzer` is able to connect the dots. I tried a lot of different ways to make this work but this was the only one that did.

So I think we have a few solutions:

1. Move `lighthouse_network` `common.rs` to `common/mod.rs` and nothing else. This enables `rust-analyzer` for the file and that's it.

2. Move `common.rs` to `/src/test_utils.rs`. This is a bit awkward though because none of the code is used in any unit tests.

3. Do 1 + add a `main.rs` so we get a consistent pattern across crates (we can consider not adding `#![cfg(not(debug_assertions))]`).

4. Do 3 + remove `#![cfg(not(debug_assertions))]` from all test code. We get the consistent `main.rs` pattern and this would fix `rust-analyzer` for all test code which is gated behind it. The comment explaining why this is attribute is present in the first place indicates that the tests run too slow in `debug` mode, but this might be an unnecessary restriction so long as we ensure we are running CI in `release`

michaelsproul:
Maybe another option is to set RUST_FLAGS in the IDE, like we do for `make lint-full`?

danielrachi1:
I agree with @macladson and I think implementing idea 4 would be the way to go. The concern about tests running slow on debug is mitigated in the `Makefile`: everything runs on `--release`.

With this, it won't be necessary to investigate how to solve the rust analyzer issue via IDE config. (I'm not a big fan of the idea anyway, but this is just personal preference. I prefer not having text editor configs per project.)

I'd like to have feedback on something before making changes: 
What integration tests should follow the `main.rs` pattern? 
All of them? (see the `find` command I ran in https://github.com/sigp/lighthouse/pull/7836#issuecomment-3162501519)
Only the ones that have multiple files? 
Only the tests for `lighthouse_network`?

cc @michaelsproul

michaelsproul:
I don't think removing `#![cfg(not(debug_assertions))]` is that straightforward, although it might still be possible. We _do_ run the tests in debug mode on CI:

https://github.com/sigp/lighthouse/blob/cfb1f7331064b758c6786e4e1dc15507af5ff5d1/.github/workflows/test-suite.yml#L224-L249

And we frequently get bug reports from devs who try to run tests in debug mode locally and wonder why they stack overflow (many IDEs default to doing this), see:

- https://github.com/sigp/lighthouse/issues/6647
- https://github.com/sigp/lighthouse/issues/6120

Re: `main.rs` I think it makes sense to do all integration tests with >1 file. E.g. `./crypto/bls/tests/tests.rs` is fine as is because it's just one file.

I think investigating the IDE config option would not take very long, and would be easy to document. Try setting `RUSTFLAGS="-C debug-assertions=no"` in the environment

danielrachi1:
The correct configuration is [rust-analyzer.cargo.cfg](https://rust-analyzer.github.io/book/configuration.html#cargo.cfgs). e.g. VSCode:

```json
{
    "rust-analyzer.cargo.cfgs": [
        "!debug_assertions",
    ]
}
```

This allows rust-analyzer to offer IDE services on files gated behind a `#![cfg(not(debug_assertions))]`, like: `lighthouse/tests/exec.rs` or `beacon_node/beacon_chain/tests/bellatrix.rs`.

My only concern is that in the future someone could gate a file with `#![cfg(debug_assertions)]`, and rust-analyzer would not work on that file. However, I couldn't find any code with that gate (that compiles only in debug). So... maybe it's not common, and a warning in the docs would be enough.

edit: moved the todo list to the PR description.

michaelsproul:
Related:

- https://github.com/rust-lang/rust-analyzer/issues/13529

It would be great if we could configure rust-analyzer in a config file and commit that to the repo, but it doesn't seem this is possible yet. Can we commit the VSCode config?

danielrachi1:
I decided to not gate the `main.rs` files with `#![cfg(not(debug_assertions))]` because most of the time individual tests files are gated. So, I suppose it's better to keep that choice module-by-module.

.gitignore
```diff
@@ -16,5 +16,3 @@ genesis.ssz
 /*.iml
 .idea
 
-# VSCode
-/.vscode
```
michaelsproul, line 19: I think we should keep this, and just force add the `.vscode/settings.json`. I'm wary of other junk from `.vscode` being included in people's PRs, e.g. `launch.json` etc

michaelsproul (approved):
LGTM, I'll just revert the gitignore change before merging

michaelsproul:
@mergify requeue

michaelsproul:
@danielrachi1 if you have a sec can you update the spellcheck list of allowed words? :pray:

then we can merge

danielrachi1:
@michaelsproul done :)

michaelsproul (approved):

danielrachi1:
@michaelsproul I forgot to add a word :man_facepalming: 

Should pass now.

For future PRs: How can I run this test locally?

michaelsproul:
I don't think we have documented a way to run it locally, I've never tried

Looks like it's this GH action:

https://github.com/sigp/lighthouse/blob/ced49dd265e01ecbf02b12073bbfde3873058abe/.github/workflows/test-suite.yml#L378-L379

michaelsproul:
@mergify requeue

Issue #7829, opened 2025-08-06
Extend fork choice debug endpoint with extra data


michaelsproul:
We currently implement only the most basic fields from:

- https://ethereum.github.io/beacon-APIs/#/Debug/getDebugForkChoice

Other clients implement additional fields, which would be useful for us to add. E.g. Prysm has:

```json
"extra_data": {
  "balance": "1108386000000000",
  "execution_optimistic": false,
  "target": "0xee98ad61066435260d2dfa2daf20047e7ff4742905cd52878a2943d064870b83",
  "timestamp": "1754461732",
  "unrealized_finalized_epoch": "384471",
  "unrealized_justified_epoch": "384472"
}
```

As seen on Xatu (https://forky.mainnet.ethpandaops.io/node/ethpandaops/mainnet/latitude-ash-mainnet-prysm-erigon-001).

Nimbus includes these extra fields:

```json
"extra_data": {
  "best_child": "0x0000000000000000000000000000000000000000000000000000000000000000",
  "best_descendant": "0x0000000000000000000000000000000000000000000000000000000000000000",
  "finalized_root": "0x7ca4d94cae84debaefd67693154f235c72b6d2ca351fbb4a4c016f53ec1516ae",
  "justified_root": "0xcef99bcc6a3fa0f5f22f9837a8706edef451972091a814a7e86726b88c42e526"
}
```

(see https://forky.mainnet.ethpandaops.io/node/ethpandaops/mainnet/sigma-mainnet-nimbus-nethermind-001).

In addition to these fields I think it would also be worth adding:

- `unrealized_justified_root`
- `unrealized_finalized_root`

These should all be already available inside fork choice's `ProtoNode`s, and just need to be plumbed through to the API.

chong-he:
I would like to work on this

michaelsproul:
Thanks CK!

chong-he:
Closed by #7845

PR #7845, opened 2025-08-08
Add extra data in `/eth/v1/debug/fork_choice`

chong-he:
## Issue Addressed

* #7829 

## Additional Info

Querying the endpoint `curl -X 'GET' 'http://localhost:5052/eth/v1/debug/fork_choice' | jq` yields:

Before:
```
  {
      "slot": "1033932",
      "block_root": "0xe1b860bbeec083c8edae37025366d3d62ac40bd27e3bd5baf0dfeb3bb74f9d13",
      "parent_root": "0x590d229e47b588449fc68c3f3047ac92e49637b8d2456c5abdaa7a80da2162ec",
      "justified_epoch": "32309",
      "finalized_epoch": "32308",
      "weight": "497146337500000",
      "validity": "valid",
      "execution_block_hash": "0xb7ccfe614cfd3ba54d11a10a5286e2e8365355d7f0bffbc456e5531d93b9b99b"
    }
```

After:
```
    {
      "slot": "1056126",
      "block_root": "0x2f4f8155e4c11a621b433570feabb60da6d303ed221e91c41160dde8ef126311",
      "parent_root": "0xde33eb2df97f3e87e1099872cfac19761bbf0379b62ef73afa2688b2fdfcefb5",
      "justified_epoch": "33002",
      "finalized_epoch": "33001",
      "weight": "2252276000000000",
      "validity": "valid",
      "execution_block_hash": "0x3870e095e8a2d9cd7f3e6c1bf26d2e06b071a982d294b774b922b9eea933a54a",
      "extra_data": {
        "target_root": "0x429b73834e076d946645b042cfe218c860c77fa6a0f1a2510a0f3ea5f13c0e83",
        "justified_root": "0xbf4ca2a5062b837822b02c28654535daa31ce465ffa7b7dc3134d68eacacbdff",
        "finalized_root": "0x009ed887bfb16ca18ed24e7463457b3be358b06fd8d8507eb9e46e0121322f09",
        "unrealized_justified_root": "0x429b73834e076d946645b042cfe218c860c77fa6a0f1a2510a0f3ea5f13c0e83",
        "unrealized_finalized_root": "0xbf4ca2a5062b837822b02c28654535daa31ce465ffa7b7dc3134d68eacacbdff",
        "unrealized_justified_epoch": "33003",
        "unrealized_finalized_epoch": "33002",
        "timestamp": "1754886936",
        "execution_status": "valid",
        "best_child": "0x71195e3c4a481d9074fd26ea4833c1a9267d77302690b9f3317b118d65c2c21b",
        "best_descendant": "0x71195e3c4a481d9074fd26ea4833c1a9267d77302690b9f3317b118d65c2c21b"
      }
```


beacon_node/http_api/src/lib.rs
```diff
@@ -2971,6 +2971,29 @@ pub fn serve<T: BeaconChainTypes>(
...
+                                    unrealized_finalized_root: node
+                                        .unrealized_finalized_checkpoint
+                                        .map(|checkpoint| checkpoint.root),
+                                    unrealized_justified_epoch: node
+                                        .unrealized_justified_checkpoint
+                                        .map(|checkpoint| checkpoint.epoch),
+                                    unrealized_finalized_epoch: node
+                                        .unrealized_finalized_checkpoint
+                                        .map(|checkpoint| checkpoint.epoch),
+                                    timestamp: timestamp_now().as_secs(),
```
chong-he: This field and `best_child` and `best_descendant` are returning numbers without quotes `""`. I see the [response](https://forky.mainnet.ethpandaops.io/node/ethpandaops/mainnet/sigma-mainnet-prysm-erigon-001) from Prysm is with the quotes, as they [convert it to string](https://github.com/OffchainLabs/prysm/blob/2ec5914b4adb48bad48caa3d7370d220f09cbc49/beacon-chain/rpc/eth/debug/handlers.go#L192). So wondering whether I can leave these fields as pure numbers?

chong-he (commented):
I have added most fields mentioned in #7829 , except for `balance` from the response of Prysm. Initially I thought it is the `weight` (which is already returned in the main response), but the data shows that `weight` is not equal to `balance` most of the time.

This is the field in the [source code](https://github.com/OffchainLabs/prysm/blob/2ec5914b4adb48bad48caa3d7370d220f09cbc49/beacon-chain/rpc/eth/debug/handlers.go#L190), which leads to [this struct](https://github.com/OffchainLabs/prysm/blob/develop/beacon-chain/core/epoch/precompute/type.go#L47-L49). I only see `weight` in `ProtoNode` in Lighthouse code. 

Other than that, this PR is ready for review. I am happy to add/remove any fields that are required/not necessary, or rearrange the fields to make it more readable

beacon_node/http_api/src/lib.rs
michaelsproul: Usually quoted numbers are better for compatibility.

For best child and best descendant does Prysm use a numeric index or a block root? To me it seems like a block root would be clearer

chong-he: The code is here (also found the definition for `balance`): https://github.com/OffchainLabs/prysm/blob/develop/beacon-chain/forkchoice/doubly-linked-tree/types.go#L60-L62

I think the *Node refers to the block root? 

In Lighthouse, it is defined as `Option<usize>` (index):
https://github.com/sigp/lighthouse/blob/cfb1f7331064b758c6786e4e1dc15507af5ff5d1/consensus/proto_array/src/proto_array.rs#L103

Do we have to change this?

michaelsproul: You can convert the index into a block root by looking it up in the array of nodes. E.g. if the index for best child is 5, use the block root of the 5th node as best child

For the balance vs weight, I think the difference is the attestation weight for the single block, vs the weight for that block and all its descendants. I think we can leave out the balance for now because it's a bit more involved to calculate it

chong-he: Changed `timestamp` to String, and `best_child` and `best_descendant` to display block_root instead

Thanks!

beacon_node/http_api/src/lib.rs
```diff
@@ -2978,6 +2978,35 @@ pub fn serve<T: BeaconChainTypes>(
...
+                                    unrealized_finalized_root: node
+                                        .unrealized_finalized_checkpoint
+                                        .map(|checkpoint| checkpoint.root),
+                                    unrealized_justified_epoch: node
+                                        .unrealized_justified_checkpoint
+                                        .map(|checkpoint| checkpoint.epoch),
+                                    unrealized_finalized_epoch: node
+                                        .unrealized_finalized_checkpoint
+                                        .map(|checkpoint| checkpoint.epoch),
+                                    timestamp: timestamp_now().as_secs().to_string(),
```
michaelsproul: I don't think we should include this field. I think it's meant to be the time at which the block arrived, which we don't have access to in `fork_choice`. We could try to read it from the `block_delays_cache`, but it's not guaranteed to be there. I would prefer to leave it out.

common/eth2/src/types.rs
```diff
@@ -1551,6 +1551,22 @@ pub struct ForkChoiceNode {
...
+pub struct ForkChoiceExtraData {
+    pub target_root: Hash256,
+    pub justified_root: Hash256,
+    pub finalized_root: Hash256,
+    pub unrealized_justified_root: Option<Hash256>,
+    pub unrealized_finalized_root: Option<Hash256>,
+    pub unrealized_justified_epoch: Option<Epoch>,
+    pub unrealized_finalized_epoch: Option<Epoch>,
+    pub timestamp: String,
+    pub execution_status: serde_json::Value,
```
michaelsproul: Why a `Value` rather than a `String`?

michaelsproul: I think a `String` would be simpler

beacon_node/http_api/src/lib.rs
chong-he: Revised in [714098c](https://github.com/sigp/lighthouse/pull/7845/commits/714098c6240184fcf35f09b0cb7438f5a1237f4e)

common/eth2/src/types.rs
chong-he: Fixed in [714098c](https://github.com/sigp/lighthouse/pull/7845/commits/714098c6240184fcf35f09b0cb7438f5a1237f4e)

michaelsproul (approved):
Nice work!

PR #7846, opened 2025-08-08
fix: paths so markdownlint-cli finds and fixes files properly

hexcow:
## Proposed Changes

the script had a bug where markdownlint-cli couldn’t find files or apply fixes because of wrong paths.
now it uses `-w /workdir` and absolute paths, so everything works smoothly.

### DIFF
--- a/scripts/mdlint.sh
+++ b/scripts/mdlint.sh
@@ -5,7 +5,7 @@
 # Run `make mdlint` from the root of the repository instead.
 
 # use markdownlint-cli to check for markdown files
-docker run --rm -v ./book:/workdir ghcr.io/igorshubovych/markdownlint-cli:latest '**/*.md' --ignore node_modules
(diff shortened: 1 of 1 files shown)


scripts/mdlint.sh
```diff
@@ -5,7 +5,7 @@
 # Run `make mdlint` from the root of the repository instead.
 
 # use markdownlint-cli to check for markdown files
-docker run --rm -v ./book:/workdir ghcr.io/igorshubovych/markdownlint-cli:latest '**/*.md' --ignore node_modules
+docker run --rm -v "$PWD/book":/workdir -w /workdir ghcr.io/igorshubovych/markdownlint-cli:latest "**/*.md" --ignore node_modules
```
chong-he, line 8: I am not sure if the change is necessary. The only place we use this script is to run as `make mdlint` from the root directory locally (and the CI runs automatically as checks):

https://github.com/sigp/lighthouse/blob/cfb1f7331064b758c6786e4e1dc15507af5ff5d1/Makefile#L247-L248

We will hear from the developers about this

dapplion:
Closing for inactivity

PR #7848, opened 2025-08-11
Initial Claude.md draft

jimmygchen:
## Issue Addressed

Add Initial `Claude.md` draft. Feel free to comment and make suggestions.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -151,7 +151,6 @@ Lighthouse is a modular Ethereum consensus client with two main components:
 ### Locks
 - Take great care to avoid deadlocks when working with fork choice locks - seek detailed review ([reference](https://github.com/sigp/lighthouse/blob/40c2fd5ff4215bcc5c2eed00dcb40dedd6bcc88b/beacon_node/beacon_chain/src/canonical_head.rs#L9-L32))
 - Keep lock scopes as narrow as possible to avoid blocking fast-responding functions like the networking stack
-- Consider using `try_lock()` patterns where appropriate to avoid blocking
 
 ### Async Patterns
 - Avoid blocking computations in async tasks
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -180,10 +180,43 @@ Lighthouse is a modular Ethereum consensus client with two main components:
 - Use safe math methods like `saturating_xxx` or `checked_xxx`
 - Critical that this crate behaves deterministically and MUST not have undefined behavior
 
+### Testing Patterns
+- **Use appropriate test types for the right scenarios**:
+  - **Unit tests** for single component edge cases and isolated logic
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,225 @@
...
+- `make nextest-release` - Run tests using nextest (faster parallel test runner)
+- `make test-beacon-chain` - Run beacon chain tests for all supported forks
+- `make test-slasher` - Run slasher tests with all database backend combinations
+- `make test-ef` - Download and run Ethereum Foundation test vectors
+- `make test-full` - Complete test suite including linting, EF tests, and execution engine tests
+- `cargo test -p <package_name>` - Run tests for a specific package
+- `FORK_NAME=electra cargo nextest run -p beacon_chain` - Run tests for specific fork
+
+### Linting and Code Quality  
+- `make lint` - Run Clippy linter with project-specific rules
```
michaelsproul, line 29: I think we should mention `lint-full`, it's probably more useful for Claude to run this as it also checks the tests.

CLAUDE.md
```diff
@@ -0,0 +1,225 @@
...
+
+### Building and Installation
+- `make install` - Build and install the main Lighthouse binary in release mode
+- `make install-lcli` - Build and install the `lcli` utility binary
+- `cargo build --release` - Standard Rust release build
+- `cargo build --bin lighthouse --features "gnosis,slasher-lmdb"` - Build with specific features
+
+### Testing
+- `make test` - Run the full test suite in release mode (excludes EF tests, beacon_chain, slasher, network, http_api)
+- `make nextest-release` - Run tests using nextest (faster parallel test runner)
```
michaelsproul, line 17: When I was playing around with claude code, I told it _not_ to run the full test suite when iterating, as it is too slow. Maybe we could add a note about the full suite's runtime (20 mins plus), and a suggestion that it only runs the full suite as a final check.

We could tell it to prefer to run individual tests

jimmygchen:
Also add a note about branching from unstable and target unstable when marking pull requests.

CLAUDE.md
jimmygchen: Thanks, added!

CLAUDE.md
jimmygchen: Yep, added!

CLAUDE.md
```diff
@@ -0,0 +1,296 @@
...
+
+### Safe Math in Consensus Crate
+```rust
+// ❌ Avoid - could panic
+let result = a + b;
+
+// ✅ Preferred
+let result = a.saturating_add(b);
+// or
+let result = a.checked_add(b).ok_or(ArithError::Overflow)?;
```
michaelsproul:
```suggestion
use safe_arith::SafeArith;

let result = a.safe_add(b)?;
```

michaelsproul (approved):
Looks good, happy to merge and tweak as required

PR #7851, opened 2025-08-11
Remove unused deps more aggressively 

macladson:
## Proposed Changes

I was playing around with `cargo machete` and noticed that it more aggressively detects unused deps than `cargo udeps`.
After running `cargo machete --with-metadata` and going through the list, despite a few false positives, there were quite a few (seemingly) unused dependencies.

## Additional Info
 Maybe we could consider switching to `cargo machete` over `cargo udeps` (or potentially run both) although I'd be concerned with the number of false positives. More conservative settings find significantly less and so is less compelling


boot_node/Cargo.toml
```diff
@@ -13,7 +13,6 @@ eth2_network_config = { workspace = true }
 ethereum_ssz = { workspace = true }
 hex = { workspace = true }
 lighthouse_network = { workspace = true }
-log = { workspace = true }
```
michaelsproul, line 16: We can probably delete this crate entirely from the top-level `Cargo.toml`. I don't think we want/need it for anything. Seems the only place it is used is `lcli`, and it isn't actually used there either.

beacon_node/store/Cargo.toml
```diff
@@ -17,19 +17,16 @@ ethereum_ssz = { workspace = true }
 ethereum_ssz_derive = { workspace = true }
 itertools = { workspace = true }
 leveldb = { version = "0.8.6", optional = true, default-features = false }
-logging = { workspace = true }
```
michaelsproul: Just flagging that there may be cases where we want to enable the `logging/test_logger` feature in tests, even if the `logging` crate is otherwise unused.

We may want to redesign how logging gets enabled for tests now that we are using tracing. Having logs go to stdout by default (rather than stderr) would be one option, but might still require a feature flag or env var to control this behaviour.

Right now, deleting this causes a regression in dev UX, because it's no longer possible to enable logs for anything using `BeaconChainHarness`, e.g. `store/src/iter.rs`, and `testing/ef_tests/src/cases/fork_choice.rs`. Maybe we could exempt the `logging` crate until we have a better design for test logging?

common/compare_fields/Cargo.toml
```diff
@@ -9,6 +9,3 @@ development = ["compare_fields_derive"] # used in doc-tests
 
 [dependencies]
 itertools = { workspace = true }
-
-[dev-dependencies]
-compare_fields_derive = { workspace = true }
```
michaelsproul: `cargo test` in this package fails now, as it tries to run the doc tests which depend on `compare_fields_derive`. I don't know why this failure isn't detected on CI?

Maybe it's something about how nextest/test runs over the whole workspace. Possibly something we should fix.

michaelsproul (commented):
Just a couple of Qs about things that look odd to me.

common/compare_fields/Cargo.toml
michaelsproul: Related:

- https://github.com/rust-lang/cargo/issues/6424

We might need an explicit CI job for doctests (probably heaps of them are bitrotted)

beacon_node/store/Cargo.toml
macladson: Yeah good call, I've restored all the logging deps

PR #7852, opened 2025-08-11
Re-export `context_deserialize_derive` inside `context_deserialize`

macladson:
## Proposed Changes

Re-export `context_deserialize_derive` inside of `context_deserialize` so they are both available from the same interface, which matches how popular crates (like `serde`) handle this. 

This also nests both crates inside a new `context_deserialize` directory which will make it easier to eventually spin out into a different repo (if/when) we decide to do that (plus I prefer it aesthetically). 

## Additional Info

I've also reorganized and added feature gates for external crate impls (`ssz_types`, and `milhouse`).


consensus/context_deserialize/context_deserialize/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+
+[features]
+default = ["derive"]
+derive = ["dep:context_deserialize_derive"]
+milhouse = ["dep:milhouse"]
+ssz = ["dep:ssz_types"]
+all = ["derive", "milhouse", "ssz"]
+
+[dependencies]
+context_deserialize_derive = { version = "0.1.0", path = "../context_deserialize_derive", optional = true }
```
michaelsproul, line 14: Should this not just be `workspace = true` rather than `path`?

michaelsproul (approved):
LGTM

Was there anything else you wanted to add before merging?

consensus/context_deserialize/context_deserialize/Cargo.toml
michaelsproul: Nvm, looks like that generates a warning about the version being unused. And the combo of version + path is more future proof (aligned with what we'll want when we pull this crate out).

PR #7853, opened 2025-08-11
Add a flag to disable getBlobs

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Add a flag to disable get blobs. I configured the flag to disable it regardless of version because its most likely something we use for testing anyway.


jimmygchen (approved):
LGTM, thanks!

PR #7859, opened 2025-08-12
feat: modularising eth2 [WIP]

hopinheimer:
We had a large monolithic `BeaconHttpClient` implementation. I have tried to split the implementation into functional based smaller chucks. While, I want to feature gate the client functions -- I'm still working on the making the tests work and passing CI


michaelsproul:
Maybe we could split this into multiple smaller PRs? These 3k+ line diffs are very time consuming to review, and subject to greater churn as the codebase changes.

Maybe we could modularise one module at a time?

macladson (commented):
I agree with @michaelsproul here.
What I'd recommend is:
- PR 1: `client` module
- PR 2: `config`, `debug`, `events` and `node` endpoints
- PR 3/4:  `beacon` and `validator` endpoints

hopinheimer:
facts! I was wonder this while posting the PR. I'll split it up in smaller PR accordingly. I just had one follow up query -- are you guys okay with the general direction of this PR?

michaelsproul:
I guess it seems like quite a lot of churn for minimal benefit. I think maybe a more targeted attack on the `lighthouse` feature and making its dependencies optional (like `lighthouse_network` and `proto_array`) would have more immediate benefit.

Or even splitting out the Lighthouse-specific client APIs into a separate wrapper, so that they aren't even part of the main `eth2` crate. This would help with pulling out the common logic into Bosun, while allowing us to leave the LH-specific logic with LH-deps here.

hopinheimer:
closing in accordance with #7878

PR #7862, opened 2025-08-12
Deprecate macOS-x86 binaries

antondlr:
## Issue Addressed

Rust is demoting x86 for macOS: https://blog.rust-lang.org/2025/08/07/Rust-1.89.0/
This makes it unfeasible to maintain such a build going forward.

## Proposed Changes

Stop publishing `x86_64-apple-darwin` binaries.

## Additional Information

Impacted users can always run the Linux Docker image.


michaelsproul (approved):
Thanks!

PR #7867, opened 2025-08-12
Disable sccache to fix Windows builds

michaelsproul:
## Proposed Changes

Quick fix to unblock Windows CI. I have a hammer and I'm using it.


jimmygchen (changes requested):
looks like we also need to disable the `sccache --show-stats` step:

https://github.com/sigp/lighthouse/actions/runs/16913585572/job/47921576950

jimmygchen (approved):
Passing!

Issue #7864, opened 2025-08-12
Upgrade rust-eth-kzg to latest version `0.8.0`


jimmygchen:
It's been a while since we upgraded `rust-eth-kzg` (we're currently on v0.5.4), and we should upgrade to the latest version before the fulu release scheduled in next few weeks.

jimmygchen:
Completed in #7870

PR #7870, opened 2025-08-13
Upgrade rust-eth-kzg to `0.8.0`

jimmygchen:
## Issue Addressed

#7864

The main breaking change in v0.8.0 is the `TrustedSetup` initialisation - it now requires a json string via `PeerDASTrustedSetup::from_json`.


eserilev (approved):
LGTM!

kevaundray (approved):
LGTM -- note that if you wanted to, you could call `TrustedSetup::default`: https://github.com/crate-crypto/rust-eth-kzg/blob/d49e9057ebce035e6a6024f32ae920fb20819ff8/crates/trusted_setup/src/lib.rs#L82

jimmygchen:
> LGTM -- note that if you wanted to, you could call `TrustedSetup::default`: https://github.com/crate-crypto/rust-eth-kzg/blob/d49e9057ebce035e6a6024f32ae920fb20819ff8/crates/trusted_setup/src/lib.rs#L82

Thanks @kevaundray ! 

I had a look into the code and currently this may be a bit of work until we remove / feature gate c-kzg and remove the `--trusted-setup-override` flag, as we currently pass the bytes array through from the [config](https://github.com/sigp/lighthouse/blob/ee9af3defd3edec07e83a2b4b372b70025d7e9f4/beacon_node/client/src/config.rs#L72) object, which is also used by `c-kzg`.

I think we can probably do this when we get to #6107, and it's likely going to simplify a lot of things (remove trusted setup bytes from client config etc)

PR #7872, opened 2025-08-13
Yeet env_logger into the sun

michaelsproul:
## Proposed Changes

- Remove explicit `env_logger` usage from `state_processing` tests and `lcli`.
- Set up tracing correctly for `lcli` (I've checked that we can see logs after this change).
- I didn't do anything to set up logging for the `state_processing` tests, as these are rarely run manually (they never fail). We could add `test_logger` in there on an as-needed basis.

## Additional Info

- This eliminates a dependency on `atty` which is abandonware with an [active RUSTSEC warning](https://rustsec.org/advisories/RUSTSEC-2024-0375.html).
- We still depend on a different version of `env_logger` via `quickcheck`. We could probably get rid of that by using `proptest` consistently.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -710,17 +710,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "atty"
-version = "0.2.14"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


jimmygchen (approved):
LGTM!

PR #7873, opened 2025-08-13
Try using sccache instead of disabling

antondlr:
## Issue Addressed

We temporarily can't build sccache on windows runners, but it's still available on linux. 
this smol change lets us use it when available, instead of disabling across the board. 

## Proposed Changes

The Windows runners now have a conditional check to disable (unset the `rustc-wrapper` env var) sccache in their entrypoint, just like the Linux ones have. 
Also the workflows no longer fail when `sccache --show-stats` fails.

## Additional Info

Additionally, bumped `checkout@v4` to `checkout@v5` to beat the bots proposing this oh-so-important life changing version bump.


michaelsproul (approved):
Thanks! 🤗

Issue #7855, opened 2025-08-12
Refactor data column RPC peer selection


jimmygchen:
See comment in this PR

https://github.com/sigp/lighthouse/pull/7762?notification_referrer_id=NT_kwDOAAtVarIxNzY4MDcyNzU3ODo3NDI3NjI#discussion_r2263107969

The duplicated function in range and backfill sync can be refactored into a function on `PeerDB`, and can be optimised using short-circuit as described in the above comment.

jimmygchen:
Completed in #7915

PR #4182, opened 2023-04-13
[Merged by Bors] - Separate BN for block proposals

AgeManning:
It is a well-known fact that IP addresses for beacon nodes used by specific validators can be de-anonymized. There is an assumed risk that a malicious user may attempt to DOS validators when producing blocks to prevent chain growth/liveness.

Although there are a number of ideas put forward to address this, there a few simple approaches we can take to mitigate this risk.

Currently, a Lighthouse user is able to set a number of beacon-nodes that their validator client can connect to. If one beacon node is taken offline, it can fallback to another. Different beacon nodes can use VPNs or rotate IPs in order to mask their IPs.

This PR provides an additional setup option which further mitigates attacks of this kind.

This PR introduces a CLI flag --proposer-only to the beacon node. Setting this flag will configure the beacon node to run with minimal peers and crucially will not subscribe to subnets or sync committees. Therefore nodes of this kind should not be identified as nodes connected to validators of any kind.

It also introduces a CLI flag --proposer-nodes to the validator client. Users can then provide a number of beacon nodes (which may or may not run the --proposer-only flag) that the Validator client will use for block production and propagation only. If these nodes fail, the validator client will fallback to the default list of beacon nodes.

Users are then able to set up a number of beacon nodes dedicated to block proposals (which are unlikely to be identified as validator nodes) and point their validator clients to produce blocks on these nodes and attest on other beacon nodes. An attack attempting to prevent liveness on the eth2 network would then need to preemptively find and attack the proposer nodes which is significantly more difficult than the default setup.

This is a follow on from: #3328

PR #7876, opened 2025-08-14
Sync fixes for `fusaka-devnet-4`

jimmygchen:
## Issue Addressed

- [ ] Remove penalties on failed block sidecar coupling as this is problematic in forky situations. Add them to fail peer only and deprioritise them instead.
  - May no longer be required with the new range sync approach, as we now query columns by root.
- [x] Add logic to maintain custody column peers and prevent pruning
  - #7915
- [x] Prioritise status v2
  - #7912
- [ ] Penalise peers for sending blocks containing invalid execution payload.
- [x] Only send lookup request to synced / advanced peers
  - #7913


beacon_node/src/config.rs
```diff
@@ -1480,10 +1480,12 @@ pub fn set_network_config(
     if parse_flag(cli_args, "proposer-only") {
         config.subscribe_all_subnets = false;
 
-        if cli_args.get_one::<String>("target-peers").is_none() {
-            // If a custom value is not set, change the default to 15
-            config.target_peers = 15;
-        }
```
dapplion: Why was this here in the first place?

beacon_node/lighthouse_network/src/peer_manager/mod.rs
```diff
@@ -759,6 +764,13 @@ impl<E: EthSpec> PeerManager<E> {
                                         })
                                 })
                                 .collect();
+                            let cgc = if custody_subnets.len() == 128 {
+                                "supernode".to_string()
+                            } else {
+                                format!("{:?}", custody_subnets)
+                            };
+
+                            debug!(cgc, ?peer_id, "Peer custodied subnets");
```
dapplion: Should unify this log with the one in line 735?

beacon_node/lighthouse_network/src/peer_manager/mod.rs
```diff
@@ -52,6 +52,11 @@ pub const PEER_RECONNECTION_TIMEOUT: Duration = Duration::from_secs(600);
 /// lower our peer count below this number. Instead we favour a non-uniform distribution of subnet
 /// peers.
 pub const MIN_SYNC_COMMITTEE_PEERS: u64 = 2;
+/// Avoid pruning sampling peers if subnet peer count is <= TARGET_SUBNET_PEERS.
+/// For non sampling columns, requirements is lower, but need to ensure there is peer for
+/// publishing.
+pub const MIN_SAMPLING_COLUMN_SUBNET_PEERS: u64 = TARGET_SUBNET_PEERS as u64;
+pub const MIN_NON_SAMPLING_COLUMN_SUBNET_PEERS: u64 = 1;
```
dapplion, line 60: Why do we need 1 peer and not 0 in this subnets?

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -1092,18 +1088,17 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
...
-                    let peer_count = self
+                    let peer_count = network
+                        .network_globals()
                         .peers
-                        .iter()
-                        .filter(|peer| {
-                            peer_db.is_good_range_sync_custody_subnet_peer(*subnet_id, peer)
-                        })
+                        .read()
+                        .good_custody_subnet_peer_range_sync(*subnet_id, epoch)
```
dapplion: How often do we call `good_peers_on_sampling_subnets`? Now it loops over all peers for all subnets (if supernode)

dapplion:
Did a first review pass and approach looks solid, just some minor comments

beacon_node/lighthouse_network/src/peer_manager/mod.rs
jimmygchen: I thought it would be useful to have at least one peer per column in the case where we're publishing data columns.

beacon_node/src/config.rs
jimmygchen: See #4182. Not sure if it's possible to do this from Fulu

beacon_node/network/src/sync/range_sync/chain.rs
jimmygchen: Yes this is a lot - I raised this in earlier reviews, I can address this
https://github.com/sigp/lighthouse/issues/7855

beacon_node/src/config.rs
jimmygchen: I'm going to revert the target peer related changes, as @AgeManning is covering them properly in #7877

beacon_node/lighthouse_network/src/peer_manager/mod.rs
jimmygchen: This one is logging the computed subnets actually, not the `cgc`. I'll update the log

beacon_node/network/src/sync/range_sync/chain.rs
jimmygchen: Addressed in https://github.com/sigp/lighthouse/pull/7876/commits/a250daa29beb64349fdc326a7d0a76f02fcc592c

beacon_node/lighthouse_network/src/peer_manager/mod.rs
jimmygchen: Updated in https://github.com/sigp/lighthouse/pull/7876/commits/a250daa29beb64349fdc326a7d0a76f02fcc592c

beacon_node/src/config.rs
jimmygchen: Reverted in https://github.com/sigp/lighthouse/pull/7876/commits/ec7956bf813ea3a62604d70c89f959cc083ba476

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -774,7 +775,19 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                        error = ?err,
+                        "Invalid execution payload rejected by EE"
+                    );
+                    Err(ChainSegmentFailed {
+                        message: format!(
+                            "Peer sent a block containing invalid execution payload. Reason: {:?}",
+                            err
+                        ),
+                        peer_action: Some(PeerAction::LowToleranceError),
+                    })
```
jimmygchen, line 789: @pawanjay176 do we need this change?

This may be due to optimistic EL right?
https://github.com/sigp/lighthouse/blob/522bd9e9c6ac167f2231525e937c9ebbcb86cf6e/beacon_node/beacon_chain/src/block_verification.rs#L430-L432

beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
```diff
@@ -300,6 +300,7 @@ impl<E: EthSpec> PeerDB<E> {
             .filter(move |(_, info)| {
                 // We check both the metadata and gossipsub data as we only want to count long-lived subscribed peers
                 info.is_connected()
+                    && info.is_synced_or_advanced()
```
jimmygchen, line 303: Need to think again about this change - this is used to determine if we have enough good peers when starting discovery query:

https://github.com/sigp/lighthouse/blob/c8bbe479199ef09c5e226aa089f462cdfc9b60c5/beacon_node/lighthouse_network/src/service/mod.rs#L1093-L1103

So the condition to push a `PeerManagerEvent::DiscoverSubnetPeers` below must also satisfy this 

https://github.com/sigp/lighthouse/blob/a250daa29beb64349fdc326a7d0a76f02fcc592c/beacon_node/lighthouse_network/src/peer_manager/mod.rs#L969-L992

Otherwise it might result in pushing a `DiscoverSubnetPeers` event to network but then it gets ignored by network.

jimmygchen:
@pawanjay176 I've done a bit of clean up, but there's still a bit of mess around maintaining peers & discovery:

https://github.com/sigp/lighthouse/pull/7876#discussion_r2278373889

I'll revisit next week.

jimmygchen:
Covered in #7912, #7915 and #7946

Issue #4962, opened 2023-11-30
Simple network analysis


AgeManning:
## Description

We are currently involved in some substantial network upgrades and in the process of making decisions about default values for lighthouse and the general limits we can push the software. 

One aspect of this is to collect a few metrics as a function of peer count. 

In particular we would like to measure the bandwidth, over both tcp and quic, and the total, average number of mesh peers and average number of nodes subscribed on subnets all as a function of peer-count. 

The metrics (at least the non-aggregated versions of them) already exist on the network dashboard, which can be found (and installed here: https://github.com/sigp/lighthouse-metrics/blob/master/dashboards/Network.json (this repo explains how to run the metrics also). 

What is left to do in this issue, is to run lighthouse with a specific --target-peers value for a few minutes (10-20, until we reach the target peer count (+- 10%)). Record the metrics or an average version of them and then repeat with a different --target--peers value. 
We probably want to do this for values [80,90,100,110,120,150,180,200,250,300] to get a relationship between these metrics and the peer count. 

We can run lighthouse with syncing and without an execution client to obtain these values. I write up of how to do this is here: https://github.com/sigp/lighthouse/pull/4961

We want to test these values on mainnet. There are a few upgrades coming this week (or next) which we probably want to get in first which should give improved numbers for these and improved metrics. 

Happy to help answer questions and provide guidance for anyone wishing to take this on.

armaganyildirak:
I would like to work on this.

AgeManning:
Resolved in #7877

Issue #6920, opened 2025-02-06
Investigate if the current peer count (100) is sufficient for PeerDAS


jimmygchen:
Lighthouse currently target 100 peers and may not be sufficient if the PeerDAS number of data column subnets = 128. @dknopik's did some simulation on this, and it shows that node can sometime struggle to find peers in their subnets. Having sufficient number of supernodes in the network help, however there's no guarantee that a node would have at least a few supernode peers.

For a supernode, it may struggle to find enough peers in all 128 subnet, this makes keeping the node in sync challenging, it would likely be worse during the initial sync.

However, increasing peer count also come with cost. Simulation shows that Lighthouse may consume large amount of memory with larger peer count, and could OOM.

- Investigate what number of peers is "sufficient" based on the current subnet count (128). Is increasing peer count necessary?
- Look into strategies that helps maintain a sufficient number of peers on the custody subnets
- Is there scenarios where we'd want higher peer count? It _may_ help to increase peer count temporarily during range sync, or under bad network conditions (as it's more difficult to find peers on the same chain AND custody the required columns)?

dapplion:
> Simulation shows that Lighthouse may consume large amount of memory with larger peer count, and could OOM

This shouldn't be the case, a non-mesh peer should consume barely no resources at all. Nimbus has been using ~300 peers for a long while with very low overall memory consumption

jimmygchen:
Yes I'm aware of Nimbus's higher peer count, I'm just pointing out the results from simulation - it may well be OOM due to other factors - but it's worth investigating and monitor the resource consumption as we experiment with different numbers.

jimmygchen:
Related issue: #6930

jxs:
Hi @jimmygchen do you have any data, repro on how to trigger the OOM so I can look at?

jimmygchen:
No I don't, i think the OOM could be due to the number of nodes @dknopik was running on his machine at the time, in combination with other issues (maybe memory growth due to non-finality etc) - so it _may_ not be a networking issue. We'll need to re-test.

AgeManning:
Resolved in #7877

PR #7877, opened 2025-08-14
Re-assess Lighthouse's peer count for Fusaka

AgeManning:
# Motivation 

It is becoming clear that Fuska will benefit from a more connected network. 

The requirement to download columns and the large number of columns require us to maintain quite a diverse peer set. We also need a diverse peer set for publishing attestations. Optimizing our peer-set becomes difficult when we are constrained by the maximum number of peers we can maintain connections with. 

Lighthouse (and the overall Ethereum network) gains stability when there are more connections between peers allowing us to discover and maintain a more diverse peer set to satisfy the custody and subnet requirements. 

I'm therefore proposing to doubling Lighthouse's default peer count to 200. 

# Justification

The downside of increasing a node's peer-count is simply resource consumption. More connected peers, means that a node needs to support more RPC requests and handle a great gossipsub load. This results in bandwidth, memory and CPU increases. However the increase is not linear as one might expect. Gossipsub bounds the number of peers that consume the most bandwidth and increasing out peer count doesn't increase the number of these peers. These are controlled by the mesh parameters. 

Therefore increasing out peer count won't linearly increase our resource consumption. A greater increase in peer count should (theoretically) result in a minimal increase in bandwidth and subsequent cpu/memory use. As the stability and performance in a Fusaka network is significantly more important to us than an increase in resource use, I think it's justified for us to make this trade-off. 

# Analysis

I'll add more analysis as we go. But preliminary results look like:

### 100 Peers (Current default)
For our default 100 peer target (running over an hour on mainnet and calculating averages):
```
  Average peers: 91
  LibP2P Bandwidth:
    Inbound:  374.17 KB/s
    Outbound: 269.33 KB/s
  Discovery Bandwidth:
    Inbound:  9.40 KB/s
    Outbound: 1.91 KB/s
  Total Bandwidth:
    Inbound:  383.58 KB/s
    Outbound: 271.24 KB/s
  Average memory usage: 4950.8 MB
```
### This PR 200 Peers (slight gossip modifications)
```
  Average peers: 183
  LibP2P Bandwidth:
    Inbound:  383.30 KB/s
    Outbound: 284.66 KB/s
  Discovery Bandwidth:
    Inbound:  10.35 KB/s
    Outbound: 2.26 KB/s
  Total Bandwidth:
    Inbound:  393.65 KB/s
    Outbound: 286.92 KB/s
 Average memory usage: 5339.9 MB
```
### Comparison
```
  Peer count change: +92 (+101.3%)
  LibP2P Inbound change: 9.13 KB/s (+2.4%)
  LibP2P Outbound change: 15.33 KB/s (+5.7%)
  Total Inbound change: 10.07 KB/s (+2.6%)
  Total Outbound change: 15.68 KB/s (+5.8%)
```

Screenshot_select-area_20250814175926

### Preliminary Results

Preliminary results show marginal cost to bandwidth and about a 10% increase in memory usage. Although this looks fairly promising I'll collect more data for a more informed decision. For the gains in stability, I think this looks promising and worth the extra resource consumption. 
  
Also, users can always lower their peer counts via `--target-peers` if the resource usage is getting too high for them.


AgeManning:
I've done some further tests. These are the results:

```
----------------------------------------------------------------------------------------------------------------------------------------------------------------
Peers  Total  TCP  QUIC  Dur(h) CPU%   Mem(Mb)  Total(MB/s) TCP(MB/s)  QUIC(MB/s)
----------------------------------------------------------------------------------------------------------------------------------------------------------------
100    103    70   33    1.1    79.44  4997     0.34        0.25       0.10
150    154    99   55    1.1    83.83  5190     0.31        0.26       0.05
200    205    128  77    1.1    83.69  5142     0.29        0.23       0.05
300    304    206  98    1.1    89.28  5308     0.37        0.31       0.06
400    405    256  149   1.1    95.25  5541     0.44        0.39       0.05
500    505    337  167   1.1    98.40  5579     0.39        0.36       0.03
750    752    478  274   1.1    100.58 6333     0.55        0.54       0.02
1000   998    656  340   1.1    135.62 6833     2.59        2.50       0.08
------------------------------------------------------------------------------------------------------------------------
```
The Total column represents the average peer count over the duration of the run. The TCP and QUIC are the average peers using TCP and QUIC over the duration of the run. We have split the total bandwidth up by transport in the TCP and QUIC fields at the end. 

The outlier is 1000 peers and the reported bandwidth sky-rocketed. This could be due to peers syncing from us. 

In graph form: 
Bandwidth:
Bandwidth per Peer Count
Memory:
Memory per Peer Count
CPU:
CPU Usage per Peer Count





It is someone probabilistic that with more peers we are more likely to get nodes syncing and requesting more data, so bandwidth can grow more than these results (which ran for about an hour each on mainnet). 

In either case. I think its fine to increase our peer count to 200 and potentially up to the 500 mark. 

This should resolve #4962 and #6920

AgeManning:
Overall I propose an initial increase to 200. Once we see how the network handles this, we consider an upgrade to 300 or 400 in a future release.

beacon_node/lighthouse_network/src/config.rs
```diff
@@ -421,8 +421,8 @@ impl From<u8> for NetworkLoad {
                 mesh_n_low: 4,
                 outbound_min: 3,
                 mesh_n: 8,
-                mesh_n_high: 12,
-                gossip_lazy: 3,
+                mesh_n_high: 10,
+                gossip_lazy: 2,
```
jimmygchen, line 425: Does the `network_load=3`  values need to be updated?

The config default is `4` here 
https://github.com/sigp/lighthouse/blob/b43e0b446cd1d8a3f5b3c4a9f614f950ef672aa8/beacon_node/lighthouse_network/src/config.rs#L355

but gets overriden by the CLI to `3`:

https://github.com/sigp/lighthouse/blob/522bd9e9c6ac167f2231525e937c9ebbcb86cf6e/beacon_node/src/cli.rs#L235-L239

So _i think_ we're using 3 by default?

jimmygchen: Are the charts & numbers above based on the network_load = 3?

AgeManning: oh damn! Nice catch!

Well, these tests were using the default. And I actually think 3 is a much nicer profile. I think dropping these values for 4 also are fine.

jimmygchen (approved):
Looks good to me!

PR #7878, opened 2025-08-14
Extracting the Error impl from the monolith `eth2` 

hopinheimer:
## Proposed Changes
 Currently the `eth2` crate lib file is a large monolith of almost 3000 lines of code. As part of the bosun migration we are trying to increase code readability and modularity in the lighthouse crates initially, which then can be transferred to bosun


beacon_node/builder_client/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 pub use eth2::Error;
+use eth2::ok_or_error;
```
macladson: This can get added back to the original `use eth2` statement below

beacon_node/client/src/builder.rs
```diff
@@ -21,7 +21,8 @@ use beacon_processor::{BeaconProcessor, BeaconProcessorChannels};
 use beacon_processor::{BeaconProcessorConfig, BeaconProcessorQueueLengths};
 use environment::RuntimeContext;
 use eth2::{
-    BeaconNodeHttpClient, Error as ApiError, Timeouts,
+    BeaconNodeHttpClient, Timeouts,
+    error::Error as ApiError,
```
macladson: Since we are exporting `Error` from the root, we can revert this line

common/eth2/src/error.rs
```diff
@@ -0,0 +1,267 @@
...
+                } else {
+                    false
+                }
+            }
+            _ => false,
+        }
+    }
+
+    /// Suggested backoff duration for retry
+    pub fn suggested_backoff(&self) -> Option<Duration> {
```
macladson: Is this code actually used anywhere? This (along with `is_retryable`) probably doesn't need to exist unless there's a compelling reason to add it.

lcli/src/http_sync.rs
```diff
@@ -2,7 +2,8 @@ use clap::ArgMatches;
 use clap_utils::{parse_optional, parse_required};
 use environment::Environment;
 use eth2::{
-    BeaconNodeHttpClient, Error, SensitiveUrl, Timeouts,
+    BeaconNodeHttpClient, SensitiveUrl, Timeouts,
+    error::Error,
```
macladson: Same as earlier (we can revert this line)

validator_client/http_api/src/test_utils.rs
```diff
@@ -7,7 +7,7 @@ use account_utils::{
 use deposit_contract::decode_eth1_tx_data;
 use doppelganger_service::DoppelgangerService;
 use eth2::{
-    Error as ApiError,
+    error::Error as ApiError,
```
macladson: And here

validator_client/http_api/src/tests.rs
```diff
@@ -13,7 +13,7 @@ use account_utils::{
 };
 use deposit_contract::decode_eth1_tx_data;
 use eth2::{
-    Error as ApiError,
+    error::Error as ApiError,
```
macladson: And here

common/eth2/src/error.rs
```diff
@@ -0,0 +1,267 @@
...
+impl std::error::Error for Error {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        match self {
+            Error::SseClient(error) => Some(error.as_ref()),
+            Error::InvalidJson(error) => Some(error),
+            Error::TokenReadError(_, error) => Some(error),
+            _ => None,
+        }
+    }
+}
```
macladson: Is this used at all?

common/eth2/src/error.rs
```diff
@@ -0,0 +1,267 @@
...
+
+// === Trait for Optional Responses ===
+
+/// Trait for converting a 404 error into an `Option<Response>`.
+pub trait ResponseOptional<T> {
+    fn optional(self) -> Result<Option<T>, Error>;
+}
+
```
macladson: Looks like we are duplicating the code from `mixin.rs`? We could consider moving the `ResponseOptional` trait somewhere more common if it gets widely used, but I'd argue it shouldn't be in `error.rs`

common/eth2/src/lighthouse_vc/http_client.rs
```diff
@@ -1,5 +1,6 @@
 use super::types::*;
-use crate::Error;
+use crate::error::Error;
```
macladson: Can revert this line

common/eth2/src/lighthouse.rs
```diff
@@ -6,7 +6,8 @@ mod block_rewards;
 pub mod sync_state;
 
 use crate::{
-    BeaconNodeHttpClient, DepositData, Error, Hash256, Slot,
+    BeaconNodeHttpClient, DepositData, Hash256, Slot,
+    error::Error,
```
macladson: Can revert this line
```suggestion
    BeaconNodeHttpClient, DepositData, Error,  Hash256, Slot,
```

common/eth2/src/error.rs
```diff
@@ -0,0 +1,267 @@
...
+            Error::InvalidHeaders(details) => write!(f, "Invalid response headers: {}", details),
+            Error::InvalidSsz(error) => write!(f, "SSZ decoding error: {:?}", error),
+            Error::TokenReadError(path, error) => {
+                write!(f, "Token read error from {}: {}", path.display(), error)
+            }
+            Error::NoServerPubkey => write!(f, "Server public key required but not configured"),
+            Error::NoToken => write!(f, "Authentication token required but not configured"),
+        }
+    }
+}
```
macladson: I think many of these manual impls are redundant and simple copy the `Error` variant name anyway. We could probably fallback to the original impl in that case?

common/eth2/src/mixin.rs
```diff
@@ -1,15 +1,15 @@
-use crate::{CONSENSUS_VERSION_HEADER, Error, types::Accept};
+use crate::{CONSENSUS_VERSION_HEADER, error::Error, types::Accept};
 use reqwest::{RequestBuilder, Response, StatusCode, header::ACCEPT};
 use std::str::FromStr;
 use types::ForkName;
 
 /// Trait for converting a 404 error into an `Option<Response>`.
 pub trait ResponseOptional {
-    fn optional(self) -> Result<Option<Response>, Error>;
+    fn optional(self) -> std::result::Result<Option<Response>, Error>;
```
macladson: Is there any reason to add the explicit `std::result` here? As far as I can tell we only have the standard `Result` anyway

macladson (changes requested):
There seems to be a lot of (in my opinion) unnecessary code which is inflating the diff a bit. We can also do some cleanup of the imports

hopinheimer:
clear out a lot of the unwanted stuff. it's a much leaner diff now

common/eth2/src/error.rs
```diff
@@ -0,0 +1,165 @@
...
+        Ok(response)
+    } else if let Ok(message) = response.json::<ResponseError>().await {
+        match message {
+            ResponseError::Message(message) => Err(Error::ServerMessage(message)),
+            ResponseError::Indexed(indexed) => Err(Error::ServerIndexedMessage(indexed)),
+        }
+    } else {
+        Err(Error::StatusCode(status))
+    }
+}
```
macladson, line 148: Unless there some particular reason why we need both we might be able to just remove `ok_or_error` entirely in favour of `success_or_error`

lcli/src/http_sync.rs
macladson, line 1: Just a nit: We can revert these changes to remove this file from the diff

macladson (commented):
Looks good to me! Just a couple of comments to consider

common/eth2/src/error.rs
```diff
@@ -0,0 +1,145 @@
...
+        Ok(response)
+    } else if let Ok(message) = response.json().await {
+        match message {
+            ResponseError::Message(message) => Err(Error::ServerMessage(message)),
+            ResponseError::Indexed(indexed) => Err(Error::ServerIndexedMessage(indexed)),
+        }
+    } else {
+        Err(Error::StatusCode(status))
+    }
+}
```
macladson, line 165: Hey @hopinheimer, sorry about this, but after some discussion with Michael, certain endpoints do distinguish between `200` and `202` and we want to handle those cases differently. Can you add back the original `ok_or_error` function? (The VC related one in `http_client.rs` can stay using `success_or_error` though

hopinheimer: Sorry it was my bad too. I was about to make the same argument. since  `fulu` the `/v2/builder/blinded_blocks` is handled differently. I'll revert the changes

common/eth2/src/lighthouse_vc/http_client.rs
```diff
@@ -1,5 +1,5 @@
 use super::types::*;
-use crate::Error;
+use crate::{Error, ok_or_error};
```
macladson: Sorry, I was a bit unclear, this should be `success_or_error` since it matches the behavior of the original `ok_or_error` function  that was in this file before. Let me know if that doesn't make sense

macladson (approved):
This looks good now!

PR #7882, opened 2025-08-15
Do not set TCP port in ENR on `discv5::Event::SocketUpdated`

dknopik:
## Issue Addressed

Updating the TCP port in our ENR when `discv5::Event::SocketUpdated` is emitted is incorrect. `discv5` reports the UDP socket - which may use a different port than the libp2p TCP socket, either because a different port is passed to `--port` and `--discovery-port` or because the port is NAT'ed differently. 

## Proposed Changes

Remove the call to `update_local_enr_socket`. Keep saving the ENR to disk/globals to persist the changes done by `discv5`.

## Additional Info

See also https://github.com/sigp/anchor/pull/502.

### DIFF
--- a/beacon_node/lighthouse_network/src/discovery/mod.rs
+++ b/beacon_node/lighthouse_network/src/discovery/mod.rs
@@ -1073,16 +1073,9 @@ impl<E: EthSpec> NetworkBehaviour for Discovery<E> {
                             metrics::inc_counter(&metrics::ADDRESS_UPDATE_COUNT);
                             // Discv5 will have updated our local ENR. We save the updated version
                             // to disk.
-
-                            if (self.update_ports.tcp4 && socket_addr.is_ipv4())
-                                || (self.update_ports.tcp6 && socket_addr.is_ipv6())
-                            {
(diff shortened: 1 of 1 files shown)


ackintosh:
https://lighthouse-book.sigmaprime.io/advanced_networking.html#nat-traversal-port-forwarding
> Lighthouse has an automated way of doing this for the UDP port. This means Lighthouse can detect its external UDP port. There is no such mechanism for the TCP port. As such, we assume that the external UDP and external TCP port is the same (i.e external 5050 UDP/TCP mapping to internal 9000 is fine). If you are setting up differing external UDP and TCP ports, you should explicitly specify them using the --enr-tcp-port and --enr-udp-port as explained in the following section.

By default, we assume that the external UDP and external TCP ports are the same. So I think we should update the TCP port if the user does not specify one via `--enr-tcp-port`.

dknopik:
Ah, okay, thanks. I was not aware of that :)

I am not sure if that is a good assumption to make if the user explicitly configures `--discovery-port` and `--port` to be different. Sure, the user could then NAT map them to the same external port, but that seems unlikely.

PR #7884, opened 2025-08-16
Do not filter validators by status if filter is an empty list

dknopik:
## Issue Addressed

https://github.com/ethereum/beacon-APIs/blob/69d2feb12a6047c4f21e628dea8b0135b7e9013a/apis/beacon/states/validators.yaml#L128-L130 says we need to not filter if the filter is an empty list.

## Proposed Changes

Add a check for `statuses.is_empty()`.

### DIFF
--- a/beacon_node/http_api/tests/tests.rs
+++ b/beacon_node/http_api/tests/tests.rs
@@ -1,30 +1,30 @@
 use beacon_chain::test_utils::RelativeSyncCommittee;
 use beacon_chain::{
-    test_utils::{AttestationStrategy, BeaconChainHarness, BlockStrategy, EphemeralHarnessType},
     BeaconChain, ChainConfig, StateSkipConfig, WhenSlotSkipped,
+    test_utils::{AttestationStrategy, BeaconChainHarness, BlockStrategy, EphemeralHarnessType},
 };
 use eth2::{
+    BeaconNodeHttpClient, Error,
+    Error::ServerMessage,
+    StatusCode, Timeouts,
     mixin::{RequestAccept, ResponseForkName, ResponseOptional},
     reqwest::RequestBuilder,
     types::{
         BlockId as CoreBlockId, ForkChoiceNode, ProduceBlockV3Response, StateId as CoreStateId, *,
     },
-    BeaconNodeHttpClient, Error,
-    Error::ServerMessage,
-    StatusCode, Timeouts,
 };
 use execution_layer::expected_gas_limit;
 use execution_layer::test_utils::{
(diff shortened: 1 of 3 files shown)


jimmygchen (approved):
Good catch!

michaelsproul:
I'm working on fixing the tests for this, or rather, adding a regression test.

It's a bit moot though because it won't run on CI until this is fixed:

- https://github.com/sigp/lighthouse/pull/7809

beacon_node/http_api/tests/tests.rs
```diff
@@ -1079,7 +1079,7 @@ impl ApiTester {
                         .get_beacon_states_validators(
                             state_id.0,
                             Some(validator_index_ids.as_slice()),
-                            None,
+                            Some(statuses.as_slice()),
```
michaelsproul, line 1082: We seemed not to be testing the statuses in this function at all, which I think was just a mistake.

beacon_node/http_api/tests/tests.rs
```diff
@@ -1113,7 +1121,13 @@ impl ApiTester {
 
                         let mut validators = Vec::with_capacity(validator_indices.len());
 
-                        for i in validator_indices {
+                        let expected_indices = if validator_indices.is_empty() {
+                            (0..state.validators().len() as u64).collect()
+                        } else {
+                            validator_indices.clone()
+                        };
```
michaelsproul, line 1128: This ensures we test the `validator_indices = []` case correctly, which was also previously incorrect.

common/eth2/src/types.rs
```diff
@@ -366,7 +366,7 @@ pub struct ValidatorIdentityData {
 // this proposal:
 //
 // https://hackmd.io/bQxMDRt1RbS1TLno8K4NPg?view
-#[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
```
michaelsproul, line 369: Required for the `HashSet` usage, which is consistent with how we do indices filtering, and might be ever-so-slightly faster (probably not).

michaelsproul:
I've pushed some changes and the relevant test passes now:

```
env FORK_NAME=electra cargo test -p http_api --release --features "beacon_chain/fork_from_env" -- beacon_get_state_info
```

dknopik:
@michaelsproul Thanks for taking care of the tests and improving the fix <3

michaelsproul:
No worries! Thanks for spotting this, thought we'd already fixed all these!

PR #7887, opened 2025-08-18
Revise data column rpc limits and queue sizes

jimmygchen:
## Issue Addressed

Revise data column rpc limits and queue sizes. Also removed some outdated TODOs for Fulu / das.


eserilev (approved):
LGTM!

I was actually hitting rpc limits testing custody sync when "converting" a full node to a super node so bumping these limit helps reduce edge cases I need to deal with :)

pawanjay176 (approved):
LGTM, we need to bump this if we start requesting everything by root in range sync as well

PR #7888, opened 2025-08-18
Add rustfmt config with edition 2024

michaelsproul:
## Issue Addressed

Since we updated to edition 2024 my Vim plugin for rustfmt is formatting code incorrectly, with 2018 settings:

https://github.com/rust-lang/rust.vim/blob/889b9a7515db477f4cb6808bef1769e53493c578/autoload/rustfmt.vim#L74-L75

Arguably this plugin is a bit junk, but I think it's fairly harmless to add this config.

## Proposed Changes

Add `rustfmt.toml`. This is a generic config file for `rustfmt` which is probably useful for `rustfmt` integration with other editors too.

We may want to add other config to `rustfmt.toml` over time as well, I think this was discussed recently.


jimmygchen (approved):
LGTM

PR #7890, opened 2025-08-18
Fulu update to spec v1.6.0-alpha.4

jimmygchen:
## Issue Addressed

Fulu update to spec [v1.6.0-alpha.4](https://github.com/ethereum/consensus-specs/releases/tag/v1.6.0-alpha.4).
- Make `number_of_columns` a preset
- Optimise `get_custody_groups` to avoid computing if cgc = 128
- Add support for additional typenum values in type_dispatch macro


consensus/types/src/chain_spec.rs
```diff
@@ -773,18 +772,20 @@ impl ChainSpec {
...
+    pub fn sampling_size_columns<E: EthSpec>(
+        &self,
+        custody_group_count: u64,
+    ) -> Result<usize, String> {
         let sampling_size_groups = self.sampling_size_custody_groups(custody_group_count)?;
 
-        let columns_per_custody_group = self
-            .number_of_columns
+        let columns_per_custody_group = (E::number_of_columns() as u64)
             .safe_div(self.number_of_custody_groups)
```
ethDreamer: nit.. might be worth using the `data_columns_per_group()` function defined above here

consensus/types/src/data_column_sidecar.rs
```diff
@@ -28,69 +26,11 @@ pub type Cell<E> = FixedVector<u8, <E as EthSpec>::BytesPerCell>;
...
-        let vec = ssz::decode_list_of_variable_length_items::<Vec<u8>, Vec<Vec<u8>>>(
-            bytes,
-            Some(max_len),
-        )?
-        .into_iter()
-        .map(|bytes| DataColumnsByRootIdentifier::from_ssz_bytes(&bytes, num_columns))
-        .collect::<Result<Vec<_>, _>>()?;
-
-        Ok(RuntimeVariableList::from_vec(vec, max_len))
-    }
```
ethDreamer, line 93: soooooo much better

consensus/types/src/chain_spec.rs
```diff
@@ -2020,10 +2008,7 @@ fn default_max_blobs_by_root_request() -> usize {
 }
 
 fn default_data_columns_by_root_request() -> usize {
```
ethDreamer, line 2007: Some of these functions feel like they'd benefit from using `LazyLock<usize>` but if they're only used a couple times then whatever.

ethDreamer (approved):
Looks good to me! Some small recommendations but not married to them.

consensus/types/src/chain_spec.rs
jimmygchen: good catch!

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -446,7 +446,7 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
             .map(CustodyDataColumn::into_inner)
             .collect::<Vec<_>>();
         let all_data_columns =
-            RuntimeVariableList::from_vec(all_data_columns, self.spec.number_of_columns as usize);
+            RuntimeVariableList::from_vec(all_data_columns, T::EthSpec::number_of_columns());
```
michaelsproul, line 449: Could this (eventually) become a `VariableList` now that the max length is determined at compile-time by the preset?

I guess it doesn't matter, as this vec seems only to be used for verification

consensus/types/src/chain_spec.rs
jimmygchen: yeah, I'd probably do that too if I was starting from scratch, but many of them are only used once at runtime. I'll leave them for now as it isn't a problem.

jimmygchen:
@mergifyio requeue

testing/ef_tests/src/lib.rs
```diff
@@ -1,11 +1,11 @@
 pub use case_result::CaseResult;
 pub use cases::WithdrawalsPayload;
 pub use cases::{
-    Case, DataColumnsByRootIdentifierWrapper, EffectiveBalanceUpdates, Eth1DataReset, FeatureName,
-    HistoricalRootsUpdate, HistoricalSummariesUpdate, InactivityUpdates,
-    JustificationAndFinalization, ParticipationFlagUpdates, ParticipationRecordUpdates,
-    PendingBalanceDeposits, PendingConsolidations, ProposerLookahead, RandaoMixesReset,
-    RegistryUpdates, RewardsAndPenalties, Slashings, SlashingsReset, SyncCommitteeUpdates,
+    Case, EffectiveBalanceUpdates, Eth1DataReset, FeatureName, HistoricalRootsUpdate,
```
michaelsproul, line 4: Does the reordering of these imports imply we aren't running `cargo fmt` on our tests?

michaelsproul (approved):
LGTM

beacon_node/beacon_chain/src/data_availability_checker.rs
jimmygchen: Yeah thats right. I tried doing that and it didn't make any difference, so i ended up reverting it

testing/ef_tests/src/lib.rs
jimmygchen: I removed `DataColumnsByRootIdentifierWrapper`, so the position of line break changed.

Issue #7820, opened 2025-07-31
Use the events api to attest on a head event


pawanjay176:
## Issue

Currently, we always attest to the slot by requesting the `attestation_data` 4 seconds into the slot. See https://github.com/sigp/lighthouse/blob/134039d0146bfbd5ac928a39fe1a9f7e57848166/validator_client/validator_services/src/attestation_service.rs#L163-L169

This isn't strictly following the spec as it mentions here https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/validator.md#attesting

> A validator should create and broadcast the attestation to the associated attestation subnet when either (a) the validator has received a valid block from the expected block proposer for the assigned slot or (b) 1 / INTERVALS_PER_SLOT of the slot has transpired (SECONDS_PER_SLOT / INTERVALS_PER_SLOT seconds after the start of slot) -- whichever comes first.

We skip (a) here and always attest 4 seconds into the slot. 

We should change this to become spec compliant. This also has benefits of spreading the attestation load throughout the slot instead of bursty traffic around 4seconds. 

## Proposed solution

We currently have the sse api on the http api which emits a [head event](https://github.com/sigp/lighthouse/blob/unstable/common/eth2/src/types.rs#L1165). 

Instead of waiting 4 seconds into the slot, we should essentially do something like this in the attestation service here https://github.com/sigp/lighthouse/blob/unstable/validator_client/validator_services/src/attestation_service.rs#L163
```rust
tokio::select!{
    _ => sleep(Duration::from_secs(4)),
    _ => poll_head_event_on_all_beacon_nodes(&self.beacon_nodes),
)
```

The `poll_head_event_on_all_beacon_nodes` would be the meat of the change imo. The function could use  [`select_all`](https://docs.rs/futures/latest/futures/prelude/stream/fn.select_all.html)
 to wait on the results of the event stream from all beacon nodes in https://github.com/sigp/lighthouse/blob/unstable/validator_client/validator_services/src/attestation_service.rs#L108 and return the first returned head event. 

## Testing

Best way to test this imo would be to run a kurtosis network and check that the attestations arrive at gossip earlier than 4 seconds into the slot. 
We can check the VC log `INFO Successfully published attestations type: unaggregated` to check that the blocks were published right after the head was updated on the BN. 

We could also add some custom logs/metrics to test attestation arrivals over gossip. Not sure if any such metrics already exist cc @michaelsproul 

## Additional stability

[...]

PR #7892, opened 2025-08-18
Use events API to eager send attestations 

hopinheimer:
addresses:

- #7820


validator_client/validator_services/src/attestation_service.rs
```diff
@@ -709,6 +714,38 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> AttestationService<S,
...
+                .await
+                .map_err(|e| format!("Failed to get event stream: {:?}", e))?;
+
+            // Poll once for a head event to trigger early attestation processing
+            if let Some(event_result) = event_stream.next().await {
+                let event =
+                    event_result.map_err(|e| format!("Head event stream error: {:?}", e))?;
+                match event {
+                    EventKind::Head(_) => {
+                        trace!("Received head event, triggering early attestation processing");
```
michaelsproul: This could probably be a debug log (we almost never show trace logs), and could include the slot and block root from the event, as well as something identifying the BN

validator_client/validator_services/src/attestation_service.rs
```diff
@@ -709,6 +714,38 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> AttestationService<S,
     }
 }
 
+async fn poll_head_event_on_all_beacon_nodes<T: SlotClock, E: EthSpec>(
+    beacon_nodes: &Arc<BeaconNodeFallback<T>>,
+) -> Result<(), String> {
+    beacon_nodes
+        .first_success(|beacon_node| async move {
```
michaelsproul: Related, this function is not parallel currently, so this function likely won't work very well if the first BN takes too long:

- https://github.com/sigp/lighthouse/issues/7908

An alternative would be to maintain event streams from _all_ BNs, and trigger the attestation process as soon as we receive a head event from _one_ of them. However this also has an issue in that we might use a BN to create the attestation that hasn't yet imported the block. E.g. if we have BNs `bn1,bn2,bn3` and we get a head event from `bn2`, then we'll try to attest using `bn1`, and build an attestation to the previous block because `bn1` hasn't imported the new head yet.

This also raises a related issue, which is that we need to keep track of the `slot` of the head event, because the BNs could be lagging and sending us `head` events for old slots.

A new design that handles both of these problems might be:

- Maintain `head` event streams from _all_ BNs.
- Begin the attestation process once a `head` event arrives that is superior to our current best known head (need to keep track of this).
- Option 1: Just try to attest using the BN(s) that sent you the head event (probably just 1 BN). Or, if we track the latest head on each BN in the background, some more BNs might be in this category.
- Option 2: Request attestations from all BNs, and only accept ones that match the latest known head.

hopinheimer: I went with the Option 1 but with an extra fail-safe branch to origin first_success if it encounters a race condition with the BN index that sent the latest head

validator_client/validator_services/src/attestation_service.rs
```diff
@@ -180,9 +205,25 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> AttestationService<S,
...
     /// attestation to the beacon node.
-    fn spawn_attestation_tasks(&self, slot_duration: Duration) -> Result<(), String> {
+    fn spawn_attestation_tasks(
+        &self,
+        slot_duration: Duration,
+        beacon_node_index: Option<usize>,
+    ) -> Result<(), String> {
+        info!(
+            "process attestation from beacon_node_index {:?}",
+            beacon_node_index
```
eserilev:
```suggestion
        info!(
            ?beacon_node_index,
            "process attestation from beacon node"
```

validator_client/validator_services/src/head_monitor_service.rs
```diff
@@ -0,0 +1,223 @@
+use beacon_node_fallback::BeaconNodeFallback;
+use eth2::types::EventTopic;
+use eth2::types::SseHead;
+use tokio::sync::mpsc;
+use tracing::{info, warn};
+
```
macladson: Should get rid of this extra line so `cargo-fmt` can do its thing and tidy these imports up

validator_client/src/lib.rs
```diff
@@ -79,6 +81,7 @@ pub struct ProductionValidatorClient<E: EthSpec> {
     context: RuntimeContext<E>,
     duties_service: Arc<DutiesService<ValidatorStore<E>, SystemTimeSlotClock>>,
     block_service: BlockService<ValidatorStore<E>, SystemTimeSlotClock>,
+    head_monitor_service: HeadMonitorService<ValidatorStore<E>, SystemTimeSlotClock>,
```
macladson: I am wondering if we should make this service optional? Doesn't seem like it helps much if there is only 1 connected BN. We don't necessarily have to make it CLI configurable but something as simple as only enabling it if `beacon_nodes.len() > 1` would be sufficient.

validator_client/validator_services/src/head_monitor_service.rs
```diff
@@ -0,0 +1,223 @@
...
+use eth2::types::EventKind;
+use futures::StreamExt;
+use parking_lot::RwLock;
```
macladson: Just a nit but I think for best practices we should replace this with a `tokio::sync::RwLock` otherwise tasks waiting for the lock would block the executor thread instead of yielding. The effect would be tiny but it would scale with more nodes so it's worth avoiding

validator_client/validator_services/src/head_monitor_service.rs
```diff
@@ -0,0 +1,223 @@
...
+
+impl<S: ValidatorStore + 'static, T: SlotClock + 'static> HeadMonitorService<S, T> {
+    // Starts the service to perpetually stream head events from connected beacon_nodes
+    pub fn start_update_service(self) -> Result<(), String> {
+        let executor = self.executor.clone();
+        let head_cache = self.beacon_head_cache.clone();
+
+        info!("Starting head monitoring service");
+        let interval_fut = async move {
+            let candidates = {
```
macladson: I think we need to add a loop here, otherwise if all connections are closed (temporary network drop for example), the `join_all` will complete and then the service will stop.

Also, since we are using `join_all` here, even once we add the loop, it will never restart until _all_ the SSE connections have terminated. My understanding is that these SSE connections are long-lived, so if one beacon node goes down briefly or is restarted, that task would complete and we will never use it again for head monitoring which is actually pretty bad.

The simplest solution is using a `select!` with a timeout, so it periodically restarts (as the comment on `HeadMonitorService` suggests). I'm not sure what a reasonable timeout would be but I think every slot would be too frequent but every epoch is probably too long.

validator_client/validator_services/src/head_monitor_service.rs
```diff
@@ -0,0 +1,223 @@
...
+                let candidates_guard = self.beacon_nodes.candidates.read().await;
+                candidates_guard.clone()
+            };
+            let mut tasks = vec![];
+
+            for candidate in candidates.iter() {
```
macladson: So I think there is a spooky edge case here with `update_candidates_list` from `beacon_node_fallback` (dynamically updating the beacon_node ordering from the API). If we called that endpoint and passed a new beacon node list, we would maintain those existing SSE connections (even ones no longer on our `beacon_nodes` list, assuming those nodes were still online). 

Also, and more dangerously, we would now have an outdated candidates list meaning when we receive a head event it could be from a beacon node whose index in our local candidates list doesn't even correspond to the actual node. 

For example, if we were connected to nodes [A, B, C], and we change the list via the API to [B, C, D], now when we receive a head event from B, we actually think it is index 1 (since it was when we initialized the service), even though in our actual candidate list, it is now index 0. So when we go to submit an attestation, we will send the preferred index of 1 which will actually trigger an attestation on node C even though the actual head event came from node B.

This will affect us until the next time the service is restarted. I don't think we should wait for the periodic refresh in this case, we should make the API trigger a refresh of this service.

Alternatively we could key the cache on the endpoint rather than the index. This would ensure we never send an attestation to the "wrong" BN. At worst, we receive a head event from a stale endpoint which is not found in our new list and thus it falls back to the regular `first_success` function. We could rework that function to be `first_success_from_endpoint` instead.

But I think I prefer this less since we would still end up with these stale/zombie connections until the next periodic restart. We could just trigger a restart of the service inside `update_candidates_list` instead.

validator_client/validator_services/src/head_monitor_service.rs
```diff
@@ -0,0 +1,223 @@
...
+    head_monitor_tx: Arc<mpsc::Sender<HeadEvent>>,
+}
+
+// Runs a non-terminating loop to update the `BeaconHeadCache` with the latest head received
+// from the candidate beacon_nodes. This is an attempt to stream events to beacon nodes and
+// potential start attestion duties earlier as soon as latest head is receive from any of the
+// beacon node in contrast to attest at the 1/3rd mark in the slot.
+//
+// The cache and the candidate BNs list are periodically refresh/purged to avoid race condition that may
+// be arise due to change in ranking of beacon_nodes thus affecting the indices of beacon_nodes
```
macladson: I believe ranking beacon_nodes doesn't touch their internal index (`candidate.index`). It just affects their actual ordering in the vec, which only really matters for `first_success`.

macladson (changes requested):

validator_client/validator_services/src/head_monitor_service.rs
hopinheimer: This is a nice catch thanks! I may have over-sighted the `update_candidates_list`. follow up question the triggering a service restart from `update_candidates_list` sound nice, but to stay consistently optimal we could target a restart after the first `4s` into the slot (this might be overkill but love to hear your thoughts)?

validator_client/src/lib.rs
hopinheimer: although, I have made the service optional for validator_client start

validator_client/validator_services/src/head_monitor_service.rs
```diff
@@ -0,0 +1,206 @@
...
+
+// Builder for `HeadMonitorService`
+#[derive(Default)]
+pub struct HeadMonitorServiceBuilder<S: ValidatorStore, T: SlotClock + 'static> {
+    validator_store: Option<Arc<S>>,
+    beacon_nodes: Option<Arc<BeaconNodeFallback<T>>>,
+    executor: Option<TaskExecutor>,
+    beacon_head_cache: Option<Arc<BeaconHeadCache>>,
+    head_monitor_tx: Option<Arc<mpsc::Sender<HeadEvent>>>,
+}
```
macladson: I'm a big fan of builder patterns in general, however it doesn't seem like it makes sense here since there is no construction of `HeadMonitorService` that you can build that can actually exclude one of the fields. So them being optional isn't actually adding value

validator_client/validator_services/src/head_monitor_service.rs
```diff
@@ -0,0 +1,206 @@
...
+    type Target = Inner<S, T>;
+
+    fn deref(&self) -> &Self::Target {
+        self.inner.deref()
+    }
+}
+
+impl<S: ValidatorStore + 'static, T: SlotClock + 'static> HeadMonitorService<S, T> {
+    // Starts the service to perpetually stream head events from connected beacon_nodes
+    pub fn start_update_service(self) -> Result<(), String> {
```
macladson: I think an alternate design we could consider here is moving the head monitor to `beacon_node_fallback` (rather than copying the caches). That way `beacon_node_fallback` could internally track its own head statuses independently.

We could create a channel like we do currently but pass the sender to `BeaconNodeFallback` via a `set_head_sender()` setter similar to what we have for the `slot_clock`. `BeaconNodeFallback` can then handle constructing its own cache.

Then. if the sender channel exists, we could then simply spawn a task like this inside the `start_fallback_updater_service` function (although it might need to move to be a method on `BeaconNodeFallback` for ergonomics) to send.

Let me know if you think that makes sense, I'm happy to discuss

macladson: Also worth pointing out that Anchor is using `BeaconNodeFallback` too so we want to avoid mangling the public interface too badly. The nice thing about this design is that `new()` remains untouched and we just add an extra (and optional) `set_head_sender` function.

validator_client/validator_services/src/head_monitor_service.rs
```diff
@@ -0,0 +1,206 @@
...
+                    .head_monitor_tx
+                    .ok_or("Cannot build HeadMonitorService without head_monitor_rx")?,
+            }),
+        })
+    }
+}
+
+// Helper to minimise `Arc` usage
+pub struct Inner<S, T> {
+    _validator_store: Arc<S>,
```
macladson: I think we can remove the validator_store from these structs entirely.

macladson (changes requested):

validator_client/validator_services/src/attestation_service.rs
```diff
@@ -180,9 +218,26 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> AttestationService<S,
...
+
     /// For each each required attestation, spawn a new task that downloads, signs and uploads the
     /// attestation to the beacon node.
-    fn spawn_attestation_tasks(&self, slot_duration: Duration) -> Result<(), String> {
+    fn spawn_attestation_tasks(
+        &self,
+        slot_duration: Duration,
+        beacon_node_index: Option<usize>,
+    ) -> Result<(), String> {
+        info!(
```
macladson: This should probably be a `debug!`

validator_client/beacon_node_fallback/src/lib.rs
```diff
@@ -415,6 +440,10 @@ impl<T: SlotClock> BeaconNodeFallback<T> {
         self.candidates.read().await.len()
     }
 
+    pub fn set_head_send(&mut self, head_monitor_send: Arc<mpsc::Sender<HeadEvent>>) {
+        self.head_monitor_send = Some(head_monitor_send);
+    }
+
```
macladson, line 464: Can we move this above `num_total` so that we can keep the `num` methods together. Also a small doc string might be nice for consistency with the other methods, especially since the method name is a bit ambiguous

validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
```diff
@@ -0,0 +1,148 @@
+use crate::BeaconNodeFallback;
+use eth2::types::EthSpec;
```
macladson: We should import from the types crate directly for this import:
```rust
use types::EthSpec;
``` 
The `eth2` re-export of `types` will get removed eventually so this would save a conflict later

validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
```diff
@@ -0,0 +1,148 @@
...
+use eth2::types::SseHead;
+use std::collections::HashMap;
+use tokio::sync::RwLock;
+
+use eth2::types::EventTopic;
+use tracing::{info, warn};
+
+use eth2::types::EventKind;
```
macladson: We should remove the extra lines here so imports can get properly ordered. Also combine the `eth2::types` imports:
```rust
eth2::types::{EventHead, EventTopic, SseHead}; 
```

validator_client/beacon_node_fallback/src/lib.rs
```diff
@@ -646,6 +679,41 @@ impl<T: SlotClock> BeaconNodeFallback<T> {
...
+        &self,
+        preferred_index: Option<usize>,
+        func: F,
+    ) -> Result<O, Errors<Err>>
+    where
+        F: Fn(BeaconNodeHttpClient) -> R + Clone,
+        R: Future<Output = Result<O, Err>>,
+        Err: Debug,
+    {
+        let candidates = self.candidates.read().await;
```
macladson: We should short-circuit this by checking if `preferred_index.is_none()` before we take the read lock since we don't need `candidates` at all if there is no preferred index.

validator_client/validator_services/src/attestation_service.rs
```diff
@@ -159,19 +178,38 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> AttestationService<S,
...
+
+                if current_slot <= *last_slot {
+                    debug!(?current_slot, "Attestation already initiated for the slot");
+                    continue;
+                }
+
+                match self.spawn_attestation_tasks(slot_duration, beacon_node_index) {
+                    Ok(_) => {
+                        *last_slot = current_slot;
+                        trace!(?current_slot, "Spawned attestation tasks");
```
macladson: We can probably remove this log since we already have one inside `spawn_attestation_tasks`

macladson (changes requested):

validator_client/beacon_node_fallback/src/lib.rs
```diff
@@ -396,6 +423,8 @@ impl<T: SlotClock> BeaconNodeFallback<T> {
             candidates: Arc::new(RwLock::new(candidates)),
             distance_tiers,
             slot_clock: None,
+            beacon_head_cache: Some(Arc::new(BeaconHeadCache::new())),
```
macladson: We should default this to `None` and only initialize it if `head_monitor_send == Some(_)`. Unless the head_cache is used elsewhere but I don't think it is

macladson (dismissed):
This is looking pretty good now, just a small nit.

We just need to decide if we want a CLI flag to disable the head monitoring. I'm leaning towards yes but we could make the flag hidden. What do you think?

Also, do you have any before/after benchmarks?

validator_client/beacon_node_fallback/src/lib.rs
hopinheimer: I had a small chat with @pawanjay176 and he said it would be to slowly roll out the feature thus, I have added the feature flag `--enable-beacon-head-monitor`

validator_client/src/cli.rs
```diff
@@ -466,6 +466,17 @@ pub struct ValidatorClient {
     )]
     pub beacon_nodes_sync_tolerances: Vec<u64>,
 
+    #[clap(
+        long,
+        help = "Enable the beacon head monitor so fallback head updates trigger duties when a lagging        primary is detected. \
```
michaelsproul: This line has too many spaces in it, we could probably flow it onto the next line,

```suggestion
        help = "Enable the beacon head monitor so fallback head updates trigger duties when a lagging primary is detected. \
```

validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
```diff
@@ -0,0 +1,147 @@
+use crate::BeaconNodeFallback;
+use std::collections::HashMap;
+use tokio::sync::RwLock;
+
+use types::EthSpec;
+
```
michaelsproul: These newlines mess up the ordering of deps, better to just keep them all in one block (no newlines), or organise by `std`/not `std`.

validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
```diff
@@ -0,0 +1,147 @@
...
+use slot_clock::SlotClock;
+use std::sync::Arc;
+
+type CacheHashMap = HashMap<usize, SseHead>;
+
+// This is used send the index derived from `CandidateBeaconNode` to the
+// `AttestationService` for further processing
+#[derive(Debug)]
+pub struct HeadEvent {
+    pub beacon_node_index: usize,
```
michaelsproul, line 17: I think we could use a unique `id` to identify each BN, rather than its index. The index seems too flaky and liable to break.

e.g. at the moment the function `update_all_candidates` will reorder the nodes, but not clear the cache, which will break the link between head events and BNs.

To assign IDs we could keep track of a field `next_beacon_node_id: AtomicUsize` in `BeaconNodeFallback`, and use this to assign IDs to candidate beacon nodes whenever they are _added_. It should be immutable after `CandidateBeaconNode` creation. New IDs can be assigned by just doing `next_beacon_node_id.fetch_add(1, Ordering::SeqCst)`.

validator_client/beacon_node_fallback/src/lib.rs
```diff
@@ -646,6 +688,39 @@ impl<T: SlotClock> BeaconNodeFallback<T> {
...
+
+            match Self::run_on_candidate(preferred_node, &func).await {
+                Ok(val) => return Ok(val),
+                Err(_) => {
+                    return self.first_success(func).await;
+                }
+            }
+        }
+
+        // Fall back to normal first_success behavior
```
michaelsproul: We should log a debug message here if the preferred candidate is not found, as this should be very unlikely once we use IDs rather than indices

michaelsproul: @macladson Is it intentional that the current `index` is immutable? I noticed that it isn't updated as part of `sort_nodes_by_health`?

https://github.com/hopinheimer/lighthouse/blob/409d937ed2e9eea224da9688299b834c4f674544/validator_client/beacon_node_fallback/src/lib.rs#L804-L827

So it doesn't correspond to the `index` of the beacon node within the vec. Maybe we can rename that `index` to `id` to make this clearer.

validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
michaelsproul: This comment is somewhat obsoleted by this other comment where I realised the index is immutable:

- https://github.com/sigp/lighthouse/pull/7892#discussion_r2566763742

I think we should rename it to `id` for clarity.

validator_client/beacon_node_fallback/src/lib.rs
macladson: Basically, the index refers to the original index of the candidate list when constructed during initialization. The ordering is defined by the order of the nodes in the `--beacon-nodes` list on startup.

This is immutable because we want to order nodes within health tiers by their original index (which preserves a sense of "the first node I put in my `--beacon-nodes` list will be the primary node").

I ended up calling this value the `user_index` in the `beacon_node_health` module.

The actual index inside the list is largely irrelevant and is never logged anywhere. It only ever gets used during `first_success` when iterating through the candidate list. Pretty much everything else is only concerned with the user_index. For example, the index exposed from a head event is this `user_index`.

In fact the term `index` is only ever used to refer to the list ordering in one place (in the `sort_nodes_by_health` function). Everywhere else, the term `index` always refers to the `user_index`.

But I agree, it can be bit confusing so I'd support a rename to either `user_index` or `id`

michaelsproul: user index sounds good to me!

validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
```diff
@@ -0,0 +1,144 @@
...
+// from the candidate beacon_nodes. This is an attempt to stream events to beacon nodes and
+// potential start attestion duties earlier as soon as latest head is receive from any of the
+// beacon node in contrast to attest at the 1/3rd mark in the slot.
+//
+//
+// The cache and the candidate BNs list are refresh/purged to avoid dangling reference conditions
+// that arise due to `update_candidates_list`.
+//
+// Starts the service to perpetually stream head events from connected beacon_nodes
+pub async fn poll_head_event_from_beacon_nodes<E: EthSpec, T: SlotClock + 'static>(
```
eserilev, line 81: It might be useful to write some unit tests here?

validator_client/beacon_node_fallback/src/lib.rs
```diff
@@ -646,6 +688,39 @@ impl<T: SlotClock> BeaconNodeFallback<T> {
         Err(Errors(errors))
     }
 
+    /// Try `func` on a specific beacon node by index first, then fall back to the normal order.
+    /// Returns immediately if the preferred node succeeds, otherwise falls back to first_success.
+    /// This is an insurance against potential race conditions that may arise.
+    pub async fn first_success_from_index<F, O, Err, R>(
```
eserilev: We should be able to write some unit tests for this new function. We already have some example tests below e.g.  `first_success_should_try_nodes_in_order`

validator_client/beacon_node_fallback/src/lib.rs
```diff
@@ -646,6 +688,39 @@ impl<T: SlotClock> BeaconNodeFallback<T> {
...
+            match Self::run_on_candidate(preferred_node, &func).await {
+                Ok(val) => return Ok(val),
+                Err(_) => {
+                    return self.first_success(func).await;
+                }
+            }
+        }
+
+        // Fall back to normal first_success behavior
+        self.first_success(func).await
```
dknopik: `first_success` will try again to contact the passed `preferred_index` BN. Should we avoid that?

validator_client/beacon_node_fallback/src/lib.rs
```diff
@@ -68,6 +75,24 @@ pub fn start_fallback_updater_service<T: SlotClock + 'static, E: EthSpec>(
...
+    // requirement of starting the head monitoring service or not.
+    if beacon_nodes_ref.head_monitor_send.is_some() {
+        let head_monitor_future = async move {
+            loop {
+                if let Err(err) =
+                    poll_head_event_from_beacon_nodes::<E, T>(beacon_nodes_ref.clone()).await
+                {
+                    warn!(error=?err, "Head service failed");
+                }
+            }
```
dknopik, line 96: This can spin loop if `poll_head_event_from_beacon_nodes` fails repeatedly. Maybe add some sleep between attempts?

validator_client/src/config.rs
```diff
@@ -129,6 +132,7 @@ impl Default for Config {
             builder_registration_timestamp_override: None,
             broadcast_topics: vec![ApiTopic::Subscriptions],
             enable_latency_measurement_service: true,
+            enable_beacon_head_monitor: false,
```
dknopik: IMO, the default should be `true`, as that is the spec behaviour

validator_client/validator_services/src/attestation_service.rs
```diff
@@ -108,10 +122,13 @@ pub struct Inner<S, T> {
     beacon_nodes: Arc<BeaconNodeFallback<T>>,
     executor: TaskExecutor,
     chain_spec: Arc<ChainSpec>,
+    head_monitor_rx: Option<Arc<Mutex<mpsc::Receiver<HeadEvent>>>>,
     disable: bool,
+    latest_attested_slot: Mutex<Slot>,
```
dknopik, line 126: this seems to be only used in the `interval_fut`, can this be a local variable in there, instead of a Mutex here?

validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
```diff
@@ -0,0 +1,369 @@
...
+pub struct HeadEvent {
+    pub beacon_node_index: usize,
+}
+
+/// Cache to maintain the latest head received from each of the beacon nodes
+/// in the `BeaconNodeFallback`.
+#[derive(Debug)]
+pub struct BeaconHeadCache {
+    cache: RwLock<CacheHashMap>,
+}
```
dknopik, line 27: Is this cache really needed? Can't we just send the slot of the new head to the attestation service via the `HeadEvent` above, and the attestation service checks if it is still relevant for it?

validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
```diff
@@ -0,0 +1,369 @@
...
+// that arise due to `update_candidates_list`.
+//
+// Starts the service to perpetually stream head events from connected beacon_nodes
+pub async fn poll_head_event_from_beacon_nodes<E: EthSpec, T: SlotClock + 'static>(
+    beacon_nodes: Arc<BeaconNodeFallback<T>>,
+) -> Result<(), String> {
+    let head_cache = beacon_nodes
+        .beacon_head_cache
+        .clone()
+        .expect("Unable to start head monitor without beacon_head_cache");
```
michaelsproul: This does not need to be an expect. Just return an error.

validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
```diff
@@ -0,0 +1,369 @@
...
+    let head_monitor_send = beacon_nodes
+        .head_monitor_send
+        .clone()
+        .expect("Unable to start head monitor without head_monitor_send");
```
michaelsproul: Should be an error.

book/src/help_vc.md
```diff
@@ -205,6 +205,13 @@ Flags:
       --distributed
           Enables functionality required for running the validator in a
           distributed validator cluster.
+      --enable-beacon-head-monitor
+          Enable the beacon head monitor so fallback head updates trigger duties
+          when a lagging        primary is detected. This keeps the attestation
```
michaelsproul: Spacing is messed up

validator_client/validator_services/src/attestation_service.rs
```diff
@@ -380,7 +432,7 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> AttestationService<S,
             .now()
             .ok_or("Unable to determine current slot from clock")?
             .epoch(S::E::slots_per_epoch());
-
+
```
michaelsproul: Unnecessary whitespace

validator_client/src/config.rs
michaelsproul: Agree. We definitely need a flag to disable it though, while in the early stages of adoption

validator_client/beacon_node_fallback/src/lib.rs
michaelsproul: Agree a sleep here would be useful. One slot is probably fine.

validator_client/beacon_node_fallback/src/lib.rs
michaelsproul: I think that's OK, we already have a lot of unnecessary redundancy in these methods, this is no worse than the existing behaviour :sweat_smile:

validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
michaelsproul: I think it's useful to have the head monitor handle this, as it might eventually be useful for things other than just notifying the attestation service.

E.g. when proposing a block we might want to query the head monitor, or the attestation consensus system might want to query the cached head of each BN to work out whether consensus exists (this is being worked on by Eitan in https://github.com/sigp/lighthouse/pull/8445)

validator_client/beacon_node_fallback/src/lib.rs
```diff
@@ -22,7 +25,11 @@ use std::time::{Duration, Instant};
 use std::vec::Vec;
 use strum::EnumVariantNames;
 use task_executor::TaskExecutor;
-use tokio::{sync::RwLock, time::sleep};
+
```
michaelsproul: Unnecessary whitespace

validator_client/validator_services/src/attestation_service.rs
```diff
@@ -1,14 +1,17 @@
 use crate::duties_service::{DutiesService, DutyAndProof};
-use beacon_node_fallback::{ApiTopic, BeaconNodeFallback};
+use tokio::sync::Mutex;
+
```
michaelsproul: Whitespace again

validator_client/validator_services/src/attestation_service.rs
```diff
@@ -108,10 +122,13 @@ pub struct Inner<S, T> {
     beacon_nodes: Arc<BeaconNodeFallback<T>>,
     executor: TaskExecutor,
     chain_spec: Arc<ChainSpec>,
+    head_monitor_rx: Option<Arc<Mutex<mpsc::Receiver<HeadEvent>>>>,
```
michaelsproul: I don't think this Mutex is necessary is it? Nor the Arc. This could probably just be `Option<mpsc::Receiver<HeadEvent>>`.

validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
```diff
@@ -0,0 +1,369 @@
...
+
+    pub async fn get(&self, beacon_node_index: usize) -> Option<SseHead> {
+        self.cache.read().await.get(&beacon_node_index).cloned()
+    }
+
+    pub async fn insert(&self, beacon_node_index: usize, head: SseHead) {
+        self.cache.write().await.insert(beacon_node_index, head);
+    }
+
+    pub async fn is_latest(&self, head: &SseHead) -> bool {
```
michaelsproul, line 51: Doc comments on these functions would be nice, particularly  `is_latest` should say that it checks whether the head's slot is `>=` _all_ cached heads

validator_client/validator_services/src/attestation_service.rs
hopinheimer: true.

validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
```diff
@@ -0,0 +1,369 @@
...
+        let cache = self.cache.read().await;
+        cache
+            .values()
+            .all(|cache_head| head.slot >= cache_head.slot)
```
michaelsproul, line 55: By using `>=` here we will fire approximately one head event for every node at every slot. I think this is probably OK, as the attestation service will de-dupe redundant events for the same slot. But we should be clear about documenting this behaviour (naively one might expect just one event per slot across all nodes, which is what you'd get using `>` instead of `>=`).

michaelsproul: I think this is definitely the correct behaviour, as the consensus service might want to wait for multiple nodes to agree on a head before proceeding.

validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
```diff
@@ -0,0 +1,379 @@
...
+    let candidates = {
+        let candidates_guard = beacon_nodes.candidates.read().await;
+        candidates_guard.clone()
+    };
+    let mut tasks = vec![];
+
+    for candidate in candidates.iter() {
+        let head_event_stream = candidate
+            .beacon_node
+            .get_events::<E>(&[EventTopic::Head])
```
dknopik, line 109: As `get_events` is used here, we should activate the `events` feature of `eth2` in the `Cargo.toml` of this crate

michaelsproul: Added in face04fc8a9444f1dd70a14721c7b1d3fb4efd0d

validator_client/src/cli.rs
```diff
@@ -476,6 +476,17 @@ pub struct ValidatorClient {
...
+    #[clap(
+        long,
+        help = "Disable the beacon head monitor which tries to attest as soon as any of the \
+                configured beacon nodes sends a head event. Leaving the service enabled is \
+                recommended, but disabling it can lead to reduced bandwidth and more predictable \
+                usage of the primary beacon node (rather than the fastest BN).",
+        display_order = 0,
+        help_heading = FLAG_HEADER
+    )]
+    pub disable_beacon_head_monitor: bool,
```
michaelsproul, line 488: FYI @hopinheimer had to switch this flag to be `--disable-beacon-head-monitor` because the head monitor is enabled by default. Previously we had a flag to enable (which did nothing) and no way to disable.

validator_client/validator_services/src/attestation_service.rs
```diff
@@ -159,19 +175,40 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> AttestationService<S,
...
+
+                let Some(current_slot) = self.slot_clock.now() else {
+                    error!("Failed to read slot clock after trigger");
                     continue;
+                };
+
+                let mut last_slot = self.latest_attested_slot.lock().await;
+
+                if current_slot <= *last_slot {
+                    debug!(?current_slot, "Attestation already initiated for the slot");
```
michaelsproul: This log is a bit noisy, it's happening even on my VC with only a single validator, and at slots my validator is not even attesting.

I think the flow is:

- Head event arrives
- Trigger attestation service
- Nothing to do (not an attester), update latest slot
- Go round the loop again
- Still the same slot, so wait for slot + 1/3
- 1/3 of slot arrives, we've already set `last_slot` -> log message

Maybe we should add a sleep at the end of this loop until the next slot so that we don't wrap round like this

validator_client/validator_services/src/attestation_service.rs
michaelsproul, line 264: A debug log here would be great, something like:

```rust
debug!(
    %slot,
    from_head_monitor = beacon_node_index.is_some(),
    "Starting attestation production"
);
```

Or we could infer the `from_head_monitor` from the timestamp of this log.

validator_client/validator_services/src/attestation_service.rs
michaelsproul: Maybe the sleep should be until the start of the original slot + 1 actually, just in case the attestation process takes a long time. We don't want "next slot" to actually be slot + 2 by the time this calculation is done

common/eth2/src/lib.rs
```diff
@@ -2800,7 +2800,14 @@ impl BeaconNodeHttpClient {
             .join(",");
         path.query_pairs_mut().append_pair("topics", &topic_string);
 
-        let mut es = EventSource::get(path);
+        // Long timeout for events, some events are quite infrequent.
+        let event_timeout = Duration::from_secs(5 * 60);
```
jimmygchen: not in scope for this PR, but I wonder if there's use case for longer timeout for other events, e.g. finalized checkpoint, chain_reorg etc 

do we consider this a breaking change given that we didn't have an explicit timeout?

validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
```diff
@@ -0,0 +1,418 @@
...
+use slot_clock::SlotClock;
+use std::collections::HashMap;
+use std::sync::Arc;
+use tokio::sync::RwLock;
+use tracing::{debug, info, warn};
+use types::EthSpec;
+
+type CacheHashMap = HashMap<usize, SseHead>;
+
+// This is used send the index derived from `CandidateBeaconNode` to the
```
jimmygchen: nit:

```suggestion
// This is used to send the index derived from `CandidateBeaconNode` to the
```

validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
```diff
@@ -0,0 +1,418 @@
...
+
+impl Default for BeaconHeadCache {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+// Runs a non-terminating loop to update the `BeaconHeadCache` with the latest head received
+// from the candidate beacon_nodes. This is an attempt to stream events to beacon nodes and
+// potential start attestion duties earlier as soon as latest head is receive from any of the
```
jimmygchen:
```suggestion
// potential start attestation duties earlier as soon as latest head is receive from any of the
```

validator_client/validator_services/src/attestation_service.rs
```diff
@@ -108,10 +121,13 @@ pub struct Inner<S, T> {
     beacon_nodes: Arc<BeaconNodeFallback<T>>,
     executor: TaskExecutor,
     chain_spec: Arc<ChainSpec>,
+    head_monitor_rx: Option<Mutex<mpsc::Receiver<HeadEvent>>>,
```
jimmygchen, line 124: ~~Do we need to use a mutex for this? looks like this is only used in  `interval_fut`?~~

Ah nvm I see why, this existing struct is wrapped in an `Arc` and we needed interior mutability. i think it's fine, refactoring would probably be more extra work

common/eth2/src/lib.rs
michaelsproul: I'm about to push a commit for this. Was just going to set it at 50x the slot duration.

michaelsproul: I don't think it qualifies as a breaking change though

michaelsproul: b2cc70339115c97af1f47a11627c032dc6127a53

michaelsproul: My commit allows downstream users (e.g. Anchor) to set the timeout however they like, so IMO the breakage should be minimal

michaelsproul (approved):

validator_client/src/config.rs
```diff
@@ -82,6 +82,8 @@ pub struct Config {
     pub broadcast_topics: Vec<ApiTopic>,
     /// Enables a service which attempts to measure latency between the VC and BNs.
     pub enable_latency_measurement_service: bool,
+    /// Enables the beacon head monitor that reacts to fallback head updates.
```
jimmygchen: I'm not sure why `fallback` is mentioned here,  is below a bit clearer?

```rust
...reacts to head updates from conntected beacon nodes
```

validator_client/beacon_node_fallback/src/lib.rs
```diff
@@ -380,6 +411,8 @@ pub struct BeaconNodeFallback<T> {
     pub candidates: Arc<RwLock<Vec<CandidateBeaconNode>>>,
     distance_tiers: BeaconNodeSyncDistanceTiers,
     slot_clock: Option<T>,
+    beacon_head_cache: Option<Arc<BeaconHeadCache>>,
+    head_monitor_send: Option<Arc<mpsc::Sender<HeadEvent>>>,
```
jimmygchen, line 418: Is this `Arc` necessary? ()

validator_client/validator_services/src/attestation_service.rs
```diff
@@ -210,7 +278,7 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> AttestationService<S,
                 async move {
                     let attestation_data = attestation_service
                         .beacon_nodes
-                        .first_success(|beacon_node| async move {
+                        .first_success_from_index(beacon_node_index, |beacon_node| async move {
```
jimmygchen: Quick Q on the behaviour here, may be too late to change, but wonder if you have considered this:

current behaviour from the PR:
- we poll for head events before 4s, we get a head event from node 1
- we call `get_attestation_data` with ` first_success_from_index`, so it tries node 1 first, if it fails, we fallback to getting attestation data from node 0 - but in this case node 0 may not have new head yet, and we could attest to old head early before 4s mark

alternative consideration:
- we poll for head events before 4s, and we get a head event from node 1
- we get attestation data from that node only, if it fails, we continue to wait for other head events; or wait until 4s, and we fallback to use `first_success`

validator_client/beacon_node_fallback/src/lib.rs
```diff
@@ -68,6 +74,31 @@ pub fn start_fallback_updater_service<T: SlotClock + 'static, E: EthSpec>(
...
+            loop {
+                if let Err(error) =
+                    poll_head_event_from_beacon_nodes::<E, T>(beacon_nodes_ref.clone()).await
+                {
+                    warn!(error, "Head service failed retrying starting next slot");
+                    let sleep_time = beacon_nodes_ref
+                        .slot_clock
+                        .as_ref()
+                        .and_then(|slot_clock| slot_clock.duration_to_next_slot())
+                        .unwrap_or_else(|| Duration::from_secs(12));
```
jimmygchen: please use `spec.get_slot_duration()` instead

validator_client/beacon_node_fallback/src/lib.rs
```diff
@@ -410,6 +445,15 @@ impl<T: SlotClock> BeaconNodeFallback<T> {
         self.slot_clock = Some(slot_clock);
     }
 
+    /// This the head monitor channel that streams events from all the beacon nodes that the
+    /// validator client is connected in the `BeaconNodeFallback`. This also initializes the
+    /// beacon_head_cache under the assumption the beacon_head_cache will always be needed when
+    /// head_monitor_send is set.
+    pub fn set_head_send(&mut self, head_monitor_send: Arc<mpsc::Sender<HeadEvent>>) {
+        self.head_monitor_send = Some(head_monitor_send);
+        self.beacon_head_cache = Some(Arc::new(BeaconHeadCache::new()));
```
jimmygchen, line 457: Minor design consideration - these two fields are coupled and form the monitor functionality, so we could potentially group these into its own struct, and have the monitor and caching functionality there, which means:
* we could avoid having to check for both fields both at the caller site and in the function
* we won't run into infinite loop of errors at the caller site (https://github.com/sigp/lighthouse/blob/7e0d15119fcbaafdc4009ba2aa8fdb969160b7bd/validator_client/beacon_node_fallback/src/lib.rs#L85) if the fields aren't correctly set.

Not critically important at this stage as we know it's correctly used, just a thought for future cleanup / designs.

validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
```diff
@@ -0,0 +1,418 @@
...
+    head_cache.purge_cache().await;
+
+    // Create Vec of streams, which we will select over.
+    let mut streams = vec![];
+
+    for candidate in &candidates {
+        let head_event_stream = candidate
+            .beacon_node
+            .get_events::<E>(&[EventTopic::Head])
+            .await;
```
jimmygchen, line 110: No action required but worth noting - there's potential room for optimisation here - if you have 5 BNs and each take 10-200ms to connect, it could add up, but i think it's fine since this function is only ran once on startup

validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
```diff
@@ -0,0 +1,418 @@
...
+
+        let head_event_stream = match head_event_stream {
+            Ok(stream) => stream,
+            Err(e) => {
+                warn!(error = ?e, node_index = candidate.index, "Failed to get head event stream");
+                continue;
```
jimmygchen, line 116: note: there's no retry on this, if a node failed at the time of startup, the VC doesnt register to the nodes' event stream when it recovers, right?

validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
```diff
@@ -0,0 +1,418 @@
...
+            }
+            Ok(event) => {
+                warn!(
+                    event_kind = event.topic_name(),
+                    candidate_index, "Received unexpected event from BN"
+                );
+                continue;
+            }
+            Err(e) => {
+                return Err(format!(
```
jimmygchen, line 184: What's the expected behaviour if a BN is restarted / shutdown?
I think it may cause this function to return and will have to sleep until next slot to re-create all streams?

jimmygchen: so if a BN is restarted, the VC may never re-establish the event streams with the BN until the VC is restarted I guess? (unless we handle re-establishing failed event streams)

currently the implementation satisfiy the happy path (BN always avaialble), however it:
* ignores BNs that fails to respond during VC startup
* _may_ ignore BNs if they were shutdown / restarted once, until VC restart

jimmygchen: May be worth consider handling errors per-stream and retrying independently, so a flaky BN doesn't disrupt monitoring of healthy BNs. This may be a bigger change though, and could
be a follow-up improvement.

michaelsproul: Yeah this is a bit imperfect, but it handles the case where the user re-arranges BNs at runtime (which is otherwise not handled). I think this is fine for an MVP as the head monitoring is non essential

validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
michaelsproul: That's right. I guess we could periodically retry, but happy to leave that for a V2

jimmygchen (changes requested):
@hopinheimer thanks for the nice work.

I've added some Qs and comments, let me know what you think. But overall it looks pretty solid!

michaelsproul:
Retargeted at the release-v8.1 branch so we can resume merging Gloas things to `unstable` while we do the final rounds of testing on this PR.

I'll do a back-merge from 8.1 to unstable later.

validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
jimmygchen: Agree this is a good MVP and we can raise a separate issue for the improvements

michaelsproul:
@hopinheimer We realised that the fallback to the other BNs wouldn't work due to the main loop waiting for the next slot on failure. So we've refactored quite substantially so that:

- On head event: request attestation data from that BN and check that it matches the expected head.
- If that works, we have our attestation data, proceed as before.
- If that doesn't work, wait until attestation deadline then run `first_success` on _all_ nodes to get attestation data.

These changes are in this commit:

- 0e9dab3878c8f594ac4e3b0094b71514424b3e54

In the process of this change I deleted `first_success_from_index` and adapted its tests (simpler now). I also made `spawn_attestation_tasks` async so we can just fetch the attestation data inline, and adjusted some of the math to account for this (avoiding calculating `duration_to_next_slot` early in case attestation data production takes substantial time (expected), and being explicit about _which_ slot we are waiting for).

jimmygchen (approved):

michaelsproul:
@mergify requeue

michaelsproul:
@mergify dequeue

michaelsproul:
@mergify requeue

khanbilal732 (approved):

khanbilal732 (approved):

PR #7896, opened 2025-08-19
Spec Mapping for configs.yml in Lighthouse

Harsh-Karia:
Lighthouse Spec Mapping

## Proposed Changes

Added all of the file paths and search terms for each of the spec implementations in configs.yml.

## Additional Info

Ethspecify checks pass for all implementations except 1:

AMBIGUOUS SEARCH: configs.CUSTODY_REQUIREMENT#fulu | 'CUSTODY_REQUIREMENT:' found 2 times in common/eth2_network_config/built_in_network_configs/mainnet/config.yaml

### DIFF
--- a/specrefs/.ethspecify.yml
+++ b/specrefs/.ethspecify.yml
@@ -1,0 +1,76 @@
+version: v1.6.0-alpha.3
+style: hash
+
+specrefs:
+  search_root: ..
+  files:
+    - configs.yml
+    - constants.yml
+    - containers.yml
+    - dataclasses.yml
+    - functions.yml
+    - presets.yml
+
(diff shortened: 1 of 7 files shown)


michaelsproul:
How do we maintain this? It looks like a massive amount of work from the diff

jtraglia:
> How do we maintain this? It looks like a massive amount of work from the diff

Hey @michaelsproul this isn't ready for review yet. We'll close the PR & re-open it later.

But to answer your question, I don't expect maintenance to be very difficult. See my answer here:

* https://github.com/OffchainLabs/prysm/pull/15592#issuecomment-3197646517

PR #7897, opened 2025-08-19
Avoid unnecessary database lookups in data column RPC requests

jimmygchen:
## Issue Addressed

This PR is an optimisation to avoid unnecessary database lookups when peer requests data columns that the node doesn't custody (advertised via `cgc`).

e.g. an extreme but realistic example - a full node only store 4 custody columns by default, but it may receive a range request of 32 slots with all 128 columns, and this would result in 4096 database lookups but the node is only able to get 128 (4 * 32) of them.

## Proposed Changes

- Filter data column RPC requests (`DataColumnsByRoot`, `DataColumnsByRange`) to only lookup columns the node custodies
- Prevents unnecessary database queries that would always fail for non-custody columns


beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -323,6 +323,37 @@ impl<E: EthSpec> CustodyContext<E> {
...
+        let custody_group_count = if let Some(epoch) = epoch_opt {
+            self.custody_group_count_at_epoch(epoch, spec) as usize
+        } else {
+            self.custody_group_count_at_head(spec) as usize
+        };
+
+        let all_columns_ordered = self
+            .all_custody_columns_ordered
+            .get()
+            .expect("all_custody_columns_ordered should be initialized");
```
michaelsproul, line 354: I'm not a huge fan of these `expect`s that are starting to creep in. Is it fundamentally caused by the fact that we need to late-initialize the custody context _after_ the network objects are initialized? Like:

1. Create `CustodyContext`
2. Initialize network
3. Initialize `CustodyContext` using network parameters

It might be worth looking into whether we can eliminate this pattern (maybe we can't, I imagine it's already been looked at).

michaelsproul (approved):
Nice optimisation, I just had one question which is probably better raised as a separate issue

beacon_node/beacon_chain/src/validator_custody.rs
jimmygchen: Yes that's right - the late initialize isn't pretty but it simplifies a lot of things and better than the alternatives we explored (see #7792).

I tried to avoid `expect` whenever we can but I think it's useful when 
1. it's caused by invalid config and the crash can be detected on startup 
2. it's caused by buggy initialisation code and the crash can be detected on startup 
3. it's irrecoverable - in this case if we can't verify DA we can't follow the chain correctly

In this case it's 2 & 3, and i think returning an error and handling it may just add more code without much benefits, as it's not recoverrable.

I also tried to make sure  it's encapsulated in one place when possible. Let me know what you think - happy to address this if there's a way to nicely eliminate this pattern.

michaelsproul: Yeah I think it's ok if there's no way to avoid the late-init. I would need to dig into it, and it doesn't seem like it would be easy

PR #7898, opened 2025-08-19
Add tempo to local testnet config and update fulu kurtosis config files

jimmygchen:
## Issue Addressed

This PR adds tempo to kurtosis config and will collect lighthouse traces on kurtosis local testnet. The traces can be viewed / queried from Grafana.

Also updated fulu kurtosis configs to use latest geth image.


michaelsproul (approved):
Nice improvement!

Issue #7135, opened 2025-03-14
Consolidating clap flags into `basic`, `advanced`, `hidden`


macladson:
### Issue

Currently we have a lot of CLI flags and arguments, many of which are totally irrelevant to 99% of users _however_ the flags are useful to have. We have generally built Lighthouse with the idea that a dynamic value adjusted by flag is always better than a hard-coded constant. I 100% agree with this, however it has led to a large amount of flag bloat, which in my opinion makes `--help` unwieldy to users. It becomes confusing and difficult to find flags which are relevant, especially as a new user.

### Potential Solutions

1. Be more aggressive with `.hidden(true)`. This has the downside of being unable to ever show the hidden flags in the help text, which even means the Lighthouse book won't have the other flags. A better solution in my opinion is:
 
2. Override the help headings, then write custom help functions which dynamically hide certain headings. For example it might look like:

```
➜ lighthouse bn --help-advanced
Usage: lighthouse [OPTIONS] --basic-field <FIELD>

Options:
  -h, --help     Print help
  -H --help-advanced      Print advanced help
  -V, --version      Print version

Basic Flags:
  -f, --basic-field <FIELD>  A basic field
  -s, --basic-setting <SETTING>    Some basic setting

Advanced Flags:
      --advanced-setting <SETTING>  Some advanced setting
```

Then, running `--help` would remove the Advanced flags from the output.

We could even add a `--help-hidden` flag which shows all the hidden flags (this might not be desired if we have certain experimental flags we want users to avoid, but they could technically just look at the source code. (We could even hide the `--help-hidden` command)

What is the appetite for solution #2. Do people feel that flag bloat is a real issue?

dapplion:
I would vote to be more aggressive with hidden flags. And then maybe display all of them in the book?

michaelsproul:
I think rather than fully hiding flags I like the idea of having advanced flags that only show under `--help-advanced`. And a hidden `--help-hidden` is a cute idea too.

AgeManning:
I support this too 👍 

I do think some flags are very specific to devs and can be hidden, because the devs can just lookup the code (I still do this for misc flags that I wrote myself, but forgot how to use), so support this, in combination with more aggressive hiding.

PR #7903, opened 2025-08-20
Add `--telemetry-service-name` CLI flag for OpenTelemetry service name override

jimmygchen:
## Issue Addressed

Allows users to customize the OpenTelemetry service name instead of using the hardcoded default `lighthouse`. Defaults to 'lighthouse-bn' for beacon node, 'lighthouse-vc' for validator client, or 'lighthouse' for other subcommands.

This is useful when analysing traces from multiple nodes, see Grafana screenshot below with service name overrides in Kurtosis (`ethereum-package` PR: https://github.com/ethpandaops/ethereum-package/pull/1160):

image

### DIFF
--- a/book/src/help_vm_move.md
+++ b/book/src/help_vm_move.md
@@ -85,6 +85,10 @@ Options:
       --telemetry-collector-url <URL>
           URL of the OpenTelemetry collector to export tracing spans (e.g.,
           http://localhost:4317). If not set, tracing export is disabled.
+      --telemetry-service-name <NAME>
+          Override the OpenTelemetry service name. Defaults to 'lighthouse-bn'
+          for beacon node, 'lighthouse-vc' for validator client, or 'lighthouse'
+          for other subcommands.
       --validators <STRING>
           The validators to be moved. Either a list of 0x-prefixed validator
           pubkeys or the keyword "all".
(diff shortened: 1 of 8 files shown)


book/src/help_vm_move.md
```diff
@@ -85,6 +85,10 @@ Options:
       --telemetry-collector-url <URL>
           URL of the OpenTelemetry collector to export tracing spans (e.g.,
           http://localhost:4317). If not set, tracing export is disabled.
+      --telemetry-service-name <NAME>
```
michaelsproul, line 88: This is an orthogonal issue, but probably we don't want these flags displayed prominently on short-lived subcommands like `lighthouse vm`. I think Mac is going to look into hiding a lot of the irrelevant flags by default:

- https://github.com/sigp/lighthouse/issues/7135

michaelsproul (approved):
LGTM! Nice feature!

book/src/help_vm_move.md
jimmygchen: Yeah i agree and I thought about that too, the downside is we'd have to duplicate the flags in BN and VC. Would be good to hide them!

On a separate note, we also have a few minor flag usage issues we need to fix, e.g. bootnode panic if run in debug mode.

Issue #7603, opened 2025-06-13
Implement validator custody backfill when CGC increases


jimmygchen:
See https://github.com/ethereum/consensus-specs/pull/4357

Backfill after CGC increase is optional in the spec, but implementing it provides several benefits:
1. A newly synced node would be able to query the node for recent historical data - potentially useful for L2s 
2. Be able to contribute to the overall network sync and make the above easier.
3. The column backfill/import machinery could be useful for downloading checkpoint data columns from peers (#6837)

TODO:

- [ ]  Data column import logic #7625
- [ ]  New channel to send messages to `SyncManager` #7625
- [ ]  Custody backfill service
- [ ]  Custody backfill metrics

eserilev:
I'd like to pick this one up

jimmygchen:
Completed in #7907 🎉

PR #7907, opened 2025-08-20
Custody backfill sync

eserilev:
## Issue Addressed

#7603

#### Custody backfill sync service
Similar in many ways to the current backfill service. There may be ways to unify the two services. The difficulty there is that the current backfill service tightly couples blocks and their associated blobs/data columns. Any attempts to unify the two services should be left to a separate PR in my opinion.

#### `SyncNeworkContext`
`SyncNetworkContext` manages custody sync data columns by range requests separetly from other sync RPC requests. I think this is a nice separation considering that custody backfill is its own service.

#### Data column import logic
The import logic verifies KZG committments and that the data columns block root matches the block root in the nodes store before importing columns

#### New channel to send messages to `SyncManager`
Now external services can communicate with the `SyncManager`. In this PR this channel is used to trigger a custody sync. Alternatively we may be able to use the existing `mpsc` channel that the `SyncNetworkContext` uses to communicate with the `SyncManager`. I will spend some time reviewing this.

## TODOs
- [x] Test with tracing changes from #7831
- [x] Make sure were verifying KZG before importing columns to the store
- [x] Set custody sync status to pending if cgc count changes during foward/backfill sync. Resume custody sync when its ready
- [x] Restart custody sync if a node shuts down during custody sync
- [x] Custody backfill metrics
- [x] Fix a bug where custody sync has issues running again after a succesful custody sync
- [x] Devnet-5 testing
- [x] Non-happy path testing
- [x] Update `ValidatorRegistrations.epoch_validator_custody_requirements` while custody syncing, or once its completed
- [x] Resolve TODO comments in the PR
- [x] Ensure custody sync can be started automatically on CGC change (currently this is disabled)
- [x] Review copy-pasta inline comments
- [ ] Remove manual custody sync backfill trigerring mechanism
- [ ] Fix notifier `speedo` logic

## Addtional notes

This needs to be throughouly tested before being included in `8.0.0-rc.0`.


eserilev:
I've tested one of the non-happy paths


```
ug 21 19:13:37.088 WARN  Some data columns are missing from the batch  epoch: Epoch(69), missing_slots: [Slot(2214), Slot(2213), Slot(2208), Slot(2211), Slot(2209), Slot(2216), Slot(2215), Slot(2212), Slot(2210), Slot(2217)] 
Aug 21 19:13:37.088 WARN  Custody backfill batch processing error       error: MissingDataColumns { missing_slots_and_data_columns: {Slot(2214): {82, 32, 61, 74, 46, 29, 124, 41, 86, 59, 88, 108, 9, 115, 11, 13, 31, 77, 89, 42, 7, 58, 76, 34, 107, 0, 113, 15, 114, 30, 17, 60, 111, 20, 109, 48, 6, 14, 62, 5, 112, 18, 93, 106, 95, 16, 37, 40, 84, 127, 126, 57, 51, 123, 70, 38, 80, 2, 96, 92, 55, 102, 23, 44, 67, 54, 68, 66, 110, 56, 81, 53, 69, 72, 100, 52, 91, 118, 122, 28, 25, 63, 39, 117, 26, 65, 87, 125, 45, 3, 97, 90, 64, 35, 104, 83, 49, 1, 98, 85, 120, 105, 75, 73, 19, 43, 24, 8, 27, 121, 21, 79, 99, 50, 47, 10, 4, 22, 119, 71, 116, 103, 101, 94, 33, 78, 36, 12}, Slot(2213): {82, 32, 61, 74, 46, 29, 124, 41, 86, 59, 88, 108, 9, 115, 11, 13, 31, 77, 89, 42, 7, 58, 76, 34, 107, 0, 113, 15, 114, 30, 17, 60, 111, 20, 109, 48, 6, 14, 62, 5, 112, 18, 93, 106, 95, 16, 37, 40, 84, 127, 126, 57, 51, 123, 70, 38, 80, 2, 96, 92, 55, 102, 23, 44, 67, 54, 68, 66, 110, 56, 81, 53, 69, 72, 100, 52, 91, 118, 122, 28, 25, 63, 39, 117, 26, 65, 87, 125, 45, 3, 97, 90, 64, 35, 104, 83, 49, 1, 98, 85, 120, 105, 75, 73, 19, 43, 24, 8, 27, 121, 21, 79, 99, 50, 47, 10, 4, 22, 119, 71, 116, 103, 101, 94, 33, 78, 36, 12}, Slot(2208): {82, 32, 61, 74, 46, 29, 124, 41, 86, 59, 88, 108, 9, 115, 11, 13, 31, 77, 89, 42, 7, 58, 76, 34, 107, 0, 113, 15, 114, 30, 17, 60, 111, 20, 109, 48, 6, 14, 62, 5, 112, 18, 93, 106, 95, 16, 37, 40, 84, 127, 126, 57, 51, 123, 70, 38, 80, 2, 96, 92, 55, 102, 23, 44, 67, 54, 68, 66, 110, 56, 81, 53, 69, 72, 100, 52, 91, 118, 122, 28, 25, 63, 39, 117, 26, 65, 87, 125, 45, 3, 97, 90, 64, 35, 104, 83, 49, 1, 98, 85, 120, 105, 75, 73, 19, 43, 24, 8, 27, 121, 21, 79, 99, 50, 47, 10, 4, 22, 119, 71, 116, 103, 101, 94, 33, 78, 36, 12}, Slot(2211): {82, 32, 61, 74, 46, 29, 124, 41, 86, 59, 88, 108, 9, 115, 11, 13, 31, 77, 89, 42, 7, 58, 76, 34, 107, 0, 113, 15, 114, 30, 17, 60, 111, 20, 109, 48, 6, 14, 62, 5, 112, 18, 93, 106, 95, 16, 37, 40, 84, 127, 126, 57, 51, 123, 70, 38, 80, 2, 96, 92, 55, 102, 23, 44, 67, 54, 68, 66, 110, 56, 81, 53, 69, 72, 100, 52, 91, 118, 122, 28, 25, 63, 39, 117, 26, 65, 87, 125, 45, 3, 97, 90, 64, 35, 104, 83, 49, 1, 98, 85, 120, 105, 75, 73, 19, 43, 24, 8, 27, 121, 21, 79, 99, 50, 47, 10, 4, 22, 119, 71, 116, 103, 101, 94, 33, 78, 36, 12}, Slot(2209): {82, 32, 61, 74, 46, 29, 124, 41, 86, 59, 88, 108, 9, 115, 11, 13, 31, 77, 89, 42, 7, 58, 76, 34, 107, 0, 113, 15, 114, 30, 17, 60, 111, 20, 109, 48, 6, 14, 62, 5, 112, 18, 93, 106, 95, 16, 37, 40, 84, 127, 126, 57, 51, 123, 70, 38, 80, 2, 96, 92, 55, 102, 23, 44, 67, 54, 68, 66, 110, 56, 81, 53, 69, 72, 100, 52, 91, 118, 122, 28, 25, 63, 39, 117, 26, 65, 87, 125, 45, 3, 97, 90, 64, 35, 104, 83, 49, 1, 98, 85, 120, 105, 75, 73, 19, 43, 24, 8, 27, 121, 21, 79, 99, 50, 47, 10, 4, 22, 119, 71, 116, 103, 101, 94, 33, 78, 36, 12}, Slot(2216): {82, 32, 61, 74, 46, 29, 124, 41, 86, 59, 88, 108, 9, 115, 11, 13, 31, 77, 89, 42, 7, 58, 76, 34, 107, 0, 113, 15, 114, 30, 17, 60, 111, 20, 109, 48, 6, 14, 62, 5, 112, 18, 93, 106, 95, 16, 37, 40, 84, 127, 126, 57, 51, 123, 70, 38, 80, 2, 96, 92, 55, 102, 23, 44, 67, 54, 68, 66, 110, 56, 81, 53, 69, 72, 100, 52, 91, 118, 122, 28, 25, 63, 39, 117, 26, 65, 87, 125, 45, 3, 97, 90, 64, 35, 104, 83, 49, 1, 98, 85, 120, 105, 75, 73, 19, 43, 24, 8, 27, 121, 21, 79, 99, 50, 47, 10, 4, 22, 119, 71, 116, 103, 101, 94, 33, 78, 36, 12}, Slot(2215): {82, 32, 61, 74, 46, 29, 124, 41, 86, 59, 88, 108, 9, 115, 11, 13, 31, 77, 89, 42, 7, 58, 76, 34, 107, 0, 113, 15, 114, 30, 17, 60, 111, 20, 109, 48, 6, 14, 62, 5, 112, 18, 93, 106, 95, 16, 37, 40, 84, 127, 126, 57, 51, 123, 70, 38, 80, 2, 96, 92, 55, 102, 23, 44, 67, 54, 68, 66, 110, 56, 81, 53, 69, 72, 100, 52, 91, 118, 122, 28, 25, 63, 39, 117, 26, 65, 87, 125, 45, 3, 97, 90, 64, 35, 104, 83, 49, 1, 98, 85, 120, 105, 75, 73, 19, 43, 24, 8, 27, 121, 21, 79, 99, 50, 47, 10, 4, 22, 119, 71, 116, 103, 101, 94, 33, 78, 36, 12}, Slot(2212): {82, 32, 61, 74, 46, 29, 124, 41, 86, 59, 88, 108, 9, 115, 11, 13, 31, 77, 89, 42, 7, 58, 76, 34, 107, 0, 113, 15, 114, 30, 17, 60, 111, 20, 109, 48, 6, 14, 62, 5, 112, 18, 93, 106, 95, 16, 37, 40, 84, 127, 126, 57, 51, 123, 70, 38, 80, 2, 96, 92, 55, 102, 23, 44, 67, 54, 68, 66, 110, 56, 81, 53, 69, 72, 100, 52, 91, 118, 122, 28, 25, 63, 39, 117, 26, 65, 87, 125, 45, 3, 97, 90, 64, 35, 104, 83, 49, 1, 98, 85, 120, 105, 75, 73, 19, 43, 24, 8, 27, 121, 21, 79, 99, 50, 47, 10, 4, 22, 119, 71, 116, 103, 101, 94, 33, 78, 36, 12}, Slot(2210): {82, 32, 61, 74, 46, 29, 124, 41, 86, 59, 88, 108, 9, 115, 11, 13, 31, 77, 89, 42, 7, 58, 76, 34, 107, 0, 113, 15, 114, 30, 17, 60, 111, 20, 109, 48, 6, 14, 62, 5, 112, 18, 93, 106, 95, 16, 37, 40, 84, 127, 126, 57, 51, 123, 70, 38, 80, 2, 96, 92, 55, 102, 23, 44, 67, 54, 68, 66, 110, 56, 81, 53, 69, 72, 100, 52, 91, 118, 122, 28, 25, 63, 39, 117, 26, 65, 87, 125, 45, 3, 97, 90, 64, 35, 104, 83, 49, 1, 98, 85, 120, 105, 75, 73, 19, 43, 24, 8, 27, 121, 21, 79, 99, 50, 47, 10, 4, 22, 119, 71, 116, 103, 101, 94, 33, 78, 36, 12}, Slot(2217): {82, 32, 61, 74, 46, 29, 124, 41, 86, 59, 88, 108, 9, 115, 11, 13, 31, 77, 89, 42, 7, 58, 76, 34, 107, 0, 113, 15, 114, 30, 17, 60, 111, 20, 109, 48, 6, 14, 62, 5, 112, 18, 93, 106, 95, 16, 37, 40, 84, 127, 126, 57, 51, 123, 70, 38, 80, 2, 96, 92, 55, 102, 23, 44, 67, 54, 68, 66, 110, 56, 81, 53, 69, 72, 100, 52, 91, 118, 122, 28, 25, 63, 39, 117, 26, 65, 87, 125, 45, 3, 97, 90, 64, 35, 104, 83, 49, 1, 98, 85, 120, 105, 75, 73, 19, 43, 24, 8, 27, 121, 21, 79, 99, 50, 47, 10, 4, 22, 119, 71, 116, 103, 101, 94, 33, 78, 36, 12}} } 
```

I'm guessing that since the slot is orphaned the import should actually succeed here (nothing would get imported). So i need to fix the error handling here a bit

EDIT: Ive fixed this using `forward_block_iter`

eserilev:
In order to reduce edge cases, we should wait until the cgc epoch reaches finality before begining custody backfill sync

If a node needs to custody backfilll during periods of nonfinality they can simply restart the node and checkpoint sync.

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -6872,6 +6876,24 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+                    if da_boundary_epoch < fulu_fork_epoch {
+                        Some(fulu_fork_epoch)
+                    } else {
+                        Some(da_boundary_epoch)
+                    }
+                } else {
+                    None
+                }
+            }
+            None => None, // If no DA boundary set, dont try to custody backfill
```
eserilev, line 7079: remove comment

jimmygchen:
@dapplion would you mind reviewing this please 🙏

dapplion:
While this approach is correct, we can have a simpler approach without blindly copying the machinery of the current backfill sync.

Backfill sync is conceptually simpler than forwards sync and we know exactly what data we are expecting.

The only annoyance is us requesting a range of slots that happen to have zero blocks. In that case, we need the "PendingValidation" step to handle that. However, we can make Custody backfill sync lag behind the regular backfill sync and use the roots of downloaded blocks to know exactly what data we are expecting. Then the sync process is deterministic:

- Request columns for a range of slots. We know exactly which block roots belong to those slots. If the headers of the columns don't match, instantly penalize pthe eer and retry.
- Then send columns to verify KZG proofs and BLS signatures, if correct, import; if incorrect, penalize peer and retry.

That's it, no need to track participating_peers or have pending validation batches.

beacon_node/network/src/sync/manager.rs
```diff
@@ -229,6 +268,9 @@ pub struct SyncManager<T: BeaconChainTypes> {
     /// Backfill syncing.
     backfill_sync: BackFillSync<T>,
 
+    /// Custody syncing.
+    custody_sync: CustodySync<T>,
```
dapplion: Should include backfill in the name?

beacon_node/network/src/sync/manager.rs
```diff
@@ -865,6 +951,89 @@ impl<T: BeaconChainTypes> SyncManager<T> {
...
+            }
+        }
+    }
+
+    pub(crate) fn handle_sync_service_message(&mut self, sync_service_message: SyncServiceMessage) {
+        match sync_service_message {
+            SyncServiceMessage::CustodyCountChanged { columns } => {
+                // Wait for the current epoch to finalize before starting custody sync
+                if let Err(e) = self.custody_sync.wait_for_finalization(columns) {
+                    tracing::warn!(error = ?e, "Failed to set custody backfill state to awaiting finalization");
```
dapplion:
```suggestion
                    warn!(error = ?e, "Failed to set custody backfill state to awaiting finalization");
```

beacon_node/network/src/sync/network_context.rs
```diff
@@ -293,8 +301,12 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
             blocks_by_range_requests: ActiveRequests::new("blocks_by_range"),
             blobs_by_range_requests: ActiveRequests::new("blobs_by_range"),
             data_columns_by_range_requests: ActiveRequests::new("data_columns_by_range"),
+            custody_sync_data_columns_by_range_requests: ActiveRequests::new(
```
dapplion: It's not necessary to introduce a new set of requests, just use `data_columns_by_range_requests`. And you change `DataColumnsByRangeRequestId::parent_request_id` into an enum that can either be `ComponentsByRangeRequestId` or `CustodySyncBatchRequestId`

beacon_node/network/src/sync/network_context.rs
```diff
@@ -1666,6 +1688,168 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+        // Attempt to find all required custody peers before sending any request or creating an ID
+        let columns_by_range_peers_to_request = {
+            let column_indexes = self
+                .chain
+                .sampling_columns_for_epoch(epoch)
+                .iter()
+                .cloned()
+                .collect();
+
+            Some(self.select_columns_by_range_peers_to_request(
```
dapplion: Why wrap with Some always?

beacon_node/network/src/sync/network_context.rs
```diff
@@ -1666,6 +1688,168 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+                            *peer_id,
+                            request.clone(),
+                            id,
+                        )
+                    })
+                    .collect::<Result<Vec<_>, _>>()
+            })
+            .transpose()?;
+
+        let range_data_column_batch_request = if let Some(result) = result {
```
dapplion: columns_by_range_peers_to_request is always Some

beacon_node/network/src/sync/network_context.rs
```diff
@@ -1666,6 +1688,168 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+                entry.remove();
+            }
+            // If the request is finished, dequeue everything
+            Some(data_column_result.map_err(RpcResponseError::BlockComponentCouplingError))
+        } else {
+            None
+        }
+    }
+
+    fn send_custody_sync_data_columns_by_range_request(
```
dapplion: Just use `send_data_columns_by_range_request`

common/eth2/src/lighthouse/sync_state.rs
```diff
@@ -15,6 +15,10 @@ pub enum SyncState {
     /// specified by its peers. Once completed, the node enters this sync state and attempts to
     /// download all required historical blocks.
     BackFillSyncing { completed: usize, remaining: usize },
+    /// The node is undertaking a custody backfill sync. This occurs for a node that has completed forward and
+    /// backfill sync and has undergone a custody count change. During custody backfill sync the node attempts
```
dapplion, line 19: > has completed forward and backfill sync

Is it true? Or it just

> The sync is waiting for the `earliest_available_data_column_slot`s epoch  to finalize before starting

The comments appear inconsistent

common/eth2/src/lighthouse/sync_state.rs
```diff
@@ -39,6 +43,24 @@ pub enum BackFillState {
...
+    /// The sync is waiting for the `earliest_available_data_column_slot`s epoch  to finalize before starting.
+    AwaitingFinalization,
+    /// The sync is partially completed and currently paused.
+    Paused,
+    /// We are currently backfilling custody columns.
+    Syncing,
+    /// A custody backfill sync has completed.
+    Completed,
+    /// Too many failed attempts at backfilling. Consider it failed.
+    Failed,
```
dapplion: With my suggestion above, backfill sync can only fail when our current peers send us invalid data too many times. We can resume syncing once a new peer joins. So is the `Failed` state necessary on its own?

common/eth2/src/lighthouse/sync_state.rs
```diff
@@ -39,6 +43,24 @@ pub enum BackFillState {
...
+    /// A custody sync should is set to Pending if the node is undergoing range/backfill syncing.
+    /// It should resume syncing after the node is fully synced.
+    Pending,
```
dapplion: I wonder if we can collapse `AwaitingFinalization` `Paused` `Failed` and `Pending` into a single state called `Pending(reason: String)` or `Paused(reason: String)`. And then external events like new peers or finalization call the resume function which checks that we have finalized, have peers, etc and then resume.

beacon_node/network/src/sync/custody_sync/batch.rs
```diff
@@ -0,0 +1,338 @@
...
+        let mut hasher = DefaultHasher::new();
+        data_columns.hash(&mut hasher);
+        let hash = hasher.finish();
+        Attempt { peer_id, hash }
+    }
+}
+
+#[derive(Debug)]
+/// A segment of a chain.
+pub struct CustodyBatchInfo<E: EthSpec> {
```
dapplion: You can generalize `range_sync::batch::BatchInfo` by making it generic over the data it holds on `BatchState::AwaitingProcessing` and re-use for custody backfill. A batch basically does:

- Track that a download for X is ongoing
- Handle download retries and track failed peers
- Track that a processing for X is ongoing
- Handle processing retries and track failed peers

Which is the same for range sync batch and custody batch. Do you really need to track `CustodyBatchInfo::columns` inside of the batch? Given an epoch you should be able to compute the columns in the network_context. Also if `CUSTODY_BACKFILL_EPOCHS_PER_BATCH` we don't need to track start_slot and end_slot. This is legacy code from before Deneb where each batch had more than one epoch in itself.

common/eth2/src/lighthouse/sync_state.rs
eserilev: they both need to be true

range sync and backfill sync must finish, and `earliest_available_data_column_slots` epoch must be finalized before custody backfill sync can start

common/eth2/src/lighthouse/sync_state.rs
eserilev: I've collapsed the extraneous states to a single `Pending(reason: String)` state

beacon_node/network/src/sync/custody_sync/batch.rs
eserilev: I've generalized `BatchInfo`

The diff at face value didn't decrease but thats because I moved some stuff around. I think we actually saved like 500-ish lines of code

common/eth2/src/lighthouse/sync_state.rs
eserilev: nope, no need for a `Failed` state. It's been removed

beacon_node/network/src/sync/network_context.rs
eserilev: I've removed the new set of requests and am using the existing data columns by range request type

beacon_node/beacon_chain/src/events.rs
```diff
@@ -312,3 +314,9 @@ impl<E: EthSpec> ServerSentEventHandler<E> {
         self.block_gossip_tx.receiver_count() > 0
     }
 }
+
+#[derive(Debug)]
+pub enum SyncServiceMessage {
```
pawanjay176: I'm not completely convinced we need this to be in beacon_chain.rs. 

iiuc, we are defining it here to trigger custody backfill as soon the `EarliestCustodyEpochFinalized` event happens in canonica_head. 

Instead of that, we could also just communicate to sync that the custody count has changed from the network service.

Then in manager.rs, here
https://github.com/eserilev/lighthouse/blob/7b1f7ac6e15024db8c711a6a842c3e1bd21a2b07/beacon_node/network/src/sync/manager.rs#L680

we can just check the earliest custody epoch has been finalized and then trigger custody backfill if it has. That way, we don't need to pass around the `SyncServiceMessage` everywhere around the beacon chain.

Let me know if i missed something here.

pawanjay176 (commented):
Still reviewing and testing this FYI.

beacon_node/beacon_chain/src/events.rs
eserilev: yeah I think you're right!

eserilev:
Theres a bug where the speedo notifier is showing the wrong estimated time for custody backfill sync

beacon_node/beacon_chain/src/events.rs
eserilev: I've implemented the changes here 
7d7eb24 

I think there's actually maybe a way to remove the new channel I introduced entirely. We could have the sync manager check `ValidatorRegistration::epoch_validator_custody_requirements` to figure out if we have any columns missing and trigger custody backfill sync. I'll take a stab at that and see how that goes

pawanjay176: Yeah that sounds like a good idea to me

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
```diff
@@ -0,0 +1,991 @@
...
+use crate::sync::{
+    backfill_sync::{ProcessResult, SyncStart},
+    batch::{BatchConfig, BatchInfo, BatchOperationOutcome, BatchProcessingResult, BatchState},
+    batch::{BatchId, ByRangeRequestType},
+    manager::CustodyBatchProcessResult,
+    network_context::{RpcResponseError, SyncNetworkContext},
+};
+
+/// The maximum number of batches to queue before requesting more.
+const BACKFILL_BATCH_BUFFER_SIZE: u8 = 20;
```
pawanjay176: I would prefer this to be much less. We recently changed this constant in full backfill to 5 to reduce the number of queued up backfill batches competing with reconstruction tasks in #7958

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
```diff
@@ -0,0 +1,991 @@
...
+            return self
+                .fail_sync(CustodyBackfillError::InvalidSyncState(format!(
+                    "Trying to process a batch that does not exist: {}",
+                    batch_id
+                )))
+                .map(|_| ProcessResult::Successful);
+        };
+
+        // NOTE: We send empty batches to the processor in order to trigger the processor
+        // result callback. This is done, because an empty batch could end a bad batch and catching those
```
pawanjay176: This comment doesn't apply here i think?

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
```diff
@@ -0,0 +1,991 @@
...
+        );
+
+        match result {
+            CustodyBatchProcessResult::Success {
+                imported_columns, ..
+            } => {
+                if let Err(e) = batch.processing_completed(BatchProcessingResult::Success) {
+                    self.fail_sync(CustodyBackfillError::BatchInvalidState(batch_id, e.0))?;
+                }
+                // If the processed batch was not empty, we can validate previous un-validated
```
pawanjay176: this doesn't really apply in case of custody backfill.
We can know for sure by just processing the batch if its valid or invalid, so we
should always advance if it wasn't an error even if it was 0 columns

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
```diff
@@ -0,0 +1,991 @@
...
+
+                let Some(fulu_fork_epoch) = self.beacon_chain.spec.fulu_fork_epoch else {
+                    return Err(CustodyBackfillError::InvalidSyncState(
+                        "Fulu epoch isn't schedlued".to_string(),
+                    ));
+                };
+
+                if batch_id == self.processing_target {
+                    // Advance processing target if we're above the Fulu fork epoch
+                    if self.processing_target > fulu_fork_epoch {
```
pawanjay176: I didn't quite get why we need this check. Why can't we progress processing target without this check?

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
```diff
@@ -0,0 +1,991 @@
...
+                            "Custody backfill batch failed to download. Penalizing peers"
+                        );
+                        self.fail_sync(CustodyBackfillError::BatchProcessingFailed(*batch_id))
+                            .map(|_| ProcessResult::Successful)
+                    }
+
+                    Ok(BatchOperationOutcome::Continue) => {
+                        // custody backfill can continue. Check if it can be progressed
+                        if *imported_columns > 0 {
+                            // TODO(custody-sync) is this actually true?
```
pawanjay176: this again does not apply in the custody backfill case imo.
The number of imported columns has nothing to do with the validity of the processed batch. Since we already have the blocks, we know for sure when processing a column batch if the received batch was valid or not independent of the previously downloaded batches. So at this point, we can just retry the batch with a different peer imo

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
```diff
@@ -0,0 +1,991 @@
...
+            let state = batch.state();
+            match state {
+                BatchState::AwaitingProcessing(..) => {
+                    return self.process_batch(network, self.processing_target);
+                }
+                BatchState::Downloading(..) => {
+                    // Batch is not ready, nothing to process
+                }
+                BatchState::AwaitingValidation(..) => {
+                    // This isn't a possible scenario
```
pawanjay176: Maybe log a crit here or fail sync? we should never come here ideally and we should ensure future refactors will handle this.

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
```diff
@@ -0,0 +1,991 @@
...
+                }
+                BatchState::Poisoned => unreachable!("Poisoned batch"),
+                BatchState::Failed | BatchState::AwaitingDownload | BatchState::Processing(_) => {
```
pawanjay176: AwaitingDownload is a valid state here i think

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
```diff
@@ -0,0 +1,991 @@
...
+            return Ok(ProcessResult::Successful);
+        }
+        Ok(ProcessResult::Successful)
+    }
+
+    /// Removes any batches previous to the given `validating_epoch`
+    ///
+    /// The `validating_epoch` must align with batch boundaries.
+    ///
+    /// If a previous batch has been validated and it had been re-processed, penalize the original
```
pawanjay176: this comment doesn't apply here i think

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
```diff
@@ -0,0 +1,991 @@
...
+        // We can now validate higher batches than the current batch. Here we remove all
+        // batches that are higher than the current batch. We add on an extra
+        // `BACKFILL_EPOCHS_PER_BATCH` as `split_off` is inclusive.
+        let removed_batches = self
+            .batches
+            .split_off(&(validating_epoch + CUSTODY_BACKFILL_EPOCHS_PER_BATCH));
+
+        for (id, batch) in removed_batches.into_iter() {
+            self.validated_batches = self.validated_batches.saturating_add(1);
+            // only for batches awaiting validation can we be sure the last attempt is
```
pawanjay176: this comment doesn't apply

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
```diff
@@ -0,0 +1,991 @@
...
+    ) -> Result<(), CustodyBackfillError> {
+        let span = info_span!(SPAN_CUSTODY_BACKFILL_SYNC_BATCH_REQUEST);
+        let _enter = span.enter();
+
+        if let Some(batch) = self.batches.get_mut(&batch_id) {
+            let synced_peers = self
+                .network_globals
+                .peers
+                .read()
+                .synced_peers()
```
pawanjay176: `synced_peers_for_epoch` here maybe?

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
eserilev: There was an edge case near the fulu fork epoch where we were setting the processing target to an epoch that was pre fulu. This check prevents that. I've added some comments and made the check a bit more clearer here: 51fc7ab

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
eserilev: 89c4549

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
eserilev: 89c4549

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
eserilev: 1d9750b

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
eserilev: [4aa92d2](https://github.com/sigp/lighthouse/pull/7907/commits/4aa92d23002607aea6982b39146e556a004346f9)

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
eserilev: added crit logging

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
eserilev: https://github.com/sigp/lighthouse/pull/7907/commits/494ac79d4a78706d2b80c250fde0b0b816062072

beacon_node/beacon_chain/src/events.rs
eserilev: I've completely removed that channel i introduced. the sync manager is now able to figure out when to trigger custody backfill sync

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -99,6 +101,29 @@ impl ValidatorRegistrations {
...
+                });
+
+            self.epoch_validator_custody_requirements
+                .entry(effective_epoch)
+                .and_modify(|old_custody| *old_custody = latest_validator_custody)
+                .or_insert(latest_validator_custody);
+        }
+    }
+
+    pub fn update_cgc(&mut self, new_custody: u64, effective_epoch: Epoch) {
```
eserilev: delete this after testing is done

beacon_node/http_api/src/lib.rs
```diff
@@ -4134,6 +4137,56 @@ pub fn serve<T: BeaconChainTypes>(
             },
         );
 
+    // POST lighthouse/custody_count
+    let post_lighthouse_increase_custody_count = warp::path("lighthouse")
```
eserilev: delete once were done testing

eserilev:
I think this should be ready for another round of reviews

eserilev:
Theres an edge case im not handling. When cgc changes during custody backfill sync we'll have a gap of columns that we didn't backfill. In my most recent commit i've ensured that we dont try to update our earliest data column status message in that case, but I still need to fix custody backfill sync

eserilev:
This is now in a much better spot. Some new things I've added

- Auto restart custody backfill sync if a cgc change occurs during an active sync
- Ensure we are safely updating `data_column_custody_info`
- Fetch only the columns that are missing for a given batch, this prevents duplicate column downloads
- Skip batches if we already have columns for those batches

beacon_node/network/src/sync/manager.rs
```diff
@@ -209,6 +213,36 @@ pub enum BatchProcessResult {
...
+#[derive(Debug)]
+pub enum CustodyBatchProcessResult {
+    /// The custody batch was completed successfully. It carries whether the sent batch contained blocks.
+    Success {
+        batch_id: Epoch,
+        #[allow(dead_code)]
+        sent_columns: usize,
+        imported_columns: usize,
+    },
+    /// The custody batch processing failed. It carries whether the processing imported any block.
```
jimmygchen: comment needs update here

beacon_node/http_api/src/lib.rs
```diff
@@ -4134,6 +4136,56 @@ pub fn serve<T: BeaconChainTypes>(
...
+                        .custody_context()
+                        .num_of_custody_groups_to_sample(effective_epoch, &chain.spec);
+
+                    let sampling_count = chain
+                        .data_availability_checker
+                        .custody_context()
+                        .num_of_custody_groups_to_sample(effective_epoch, &chain.spec);
+
+                    let _ = chain.store.put_data_column_custody_info(Some(
+                        (effective_epoch).end_slot(T::EthSpec::slots_per_epoch()),
```
jimmygchen: I understand this API will be deleted but just curious if its intentional to use `.end_slot` here given we use `.start_slot` for this everywhere else?

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -56,6 +59,160 @@ enum RangeBlockDataRequest<E: EthSpec> {
     },
 }
 
+pub struct RangeDataColumnBatchRequest<E: EthSpec> {
```
jimmygchen: Maybe we can use this struct for the enum variant too `RangeBlockDataRequest::DataColumns`?

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -56,6 +59,160 @@ enum RangeBlockDataRequest<E: EthSpec> {
...
+                self.requests.retain(|&k, _| k.peer != *peer);
+            }
+        }
+
+        Some(resp)
+    }
+
+    fn responses_with_custody_columns(
+        data_columns: DataColumnSidecarList<E>,
+        column_to_peer: HashMap<u64, PeerId>,
```
jimmygchen: nit: 

```suggestion
        column_to_peer: HashMap<ColumnIndex, PeerId>,
```

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -56,6 +59,160 @@ enum RangeBlockDataRequest<E: EthSpec> {
...
+                // delete it from the entries as we are going to make
+                // a separate attempt for those components.
+                self.requests.retain(|&k, _| k.peer != *peer);
+            }
+        }
+
+        Some(resp)
+    }
+
+    fn responses_with_custody_columns(
```
jimmygchen: I think it would be better to add `&self` to this function, as we have another function of the same name in this module, and this can be accessed from `RangeBlockComponentsRequest` functions and vice versa.

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -35,6 +40,36 @@ const BATCH_BUFFER_SIZE: u8 = 5;
...
+type RangeSyncBatches<E> = BTreeMap<BatchId, RangeSyncBatchInfo<E>>;
+
+/// The number of times to retry a batch before it is considered failed.
+const MAX_BATCH_DOWNLOAD_ATTEMPTS: u8 = 5;
+
+/// Invalid batches are attempted to be re-downloaded from other peers. If a batch cannot be processed
+/// after `MAX_BATCH_PROCESSING_ATTEMPTS` times, it is considered faulty.
+const MAX_BATCH_PROCESSING_ATTEMPTS: u8 = 3;
+
+#[derive(Debug)]
```
jimmygchen: Doesn't look like we need to implement `Debug` here

beacon_node/network/src/metrics.rs
```diff
@@ -212,6 +212,22 @@ pub static BEACON_PROCESSOR_RPC_BLOCK_IMPORTED_TOTAL: LazyLock<Result<IntCounter
             "Total number of gossip blocks imported to fork choice, etc.",
         )
     });
+// Custody Sync.
+pub static BEACON_PROCESSOR_CUSTODY_BACKFILL_COLUMN_IMPORT_SUCCESS_TOTAL: LazyLock<
+    Result<IntCounter>,
+> = LazyLock::new(|| {
+    try_create_int_counter(
+        "beacon_processor_custody_backfill_column_import_success_total",
+        "Total number of columns successfully processed.",
```
jimmygchen: nit: 

```suggestion
        "Total number of custody backfill columns successfully processed",
```

beacon_node/Cargo.toml
```diff
@@ -40,6 +40,7 @@ slasher = { workspace = true }
 store = { workspace = true }
 strum = { workspace = true }
 task_executor = { workspace = true }
+tokio = { workspace = true }
```
jimmygchen: doesn't look like we need `tokio` here?

common/eth2/src/types.rs
```diff
@@ -1513,6 +1513,12 @@ pub struct ManualFinalizationRequestData {
     pub block_root: Hash256,
 }
 
+#[derive(Debug, Serialize, Deserialize)]
```
jimmygchen: This seems unused?

beacon_node/beacon_chain/src/historical_data_columns.rs
```diff
@@ -0,0 +1,145 @@
...
+    fn from(e: StoreError) -> Self {
+        Self::StoreError(e)
+    }
+}
+
+impl<T: BeaconChainTypes> BeaconChain<T> {
+    /// Store a batch of historical data columns in the database.
+    ///
+    /// The data columns block roots and proposer signatures are verified with the existing
+    /// block stored in the DB. This function assumes that KZG proofs have already been verified.
```
jimmygchen: > This function assumes that KZG proofs have already been verified

It looks like this function also performs kzg verification.

beacon_node/beacon_chain/src/historical_data_columns.rs
```diff
@@ -0,0 +1,145 @@
...
+    ///
+    /// This function requires that the data column sidecar list contains columns for a full epoch.
+    ///
+    /// Return the number of `data_columns` successfully imported.
+    pub fn import_historical_data_column_batch(
```
jimmygchen, line 53: I think it's worth adding a span here, with `epoch` and `column_count` as fields

beacon_node/beacon_chain/src/historical_data_columns.rs
```diff
@@ -0,0 +1,145 @@
...
+                }
+            }
+        }
+
+        verify_kzg_for_data_column_list(historical_data_column_sidecar_list.iter(), &self.kzg)
+            .map_err(|_| HistoricalDataColumnError::InvalidKzg)?;
+
+        self.store.blobs_db.do_atomically(ops)?;
+
+        if !slot_and_column_index_to_data_columns.is_empty() {
```
jimmygchen, line 129: Should we persist the columns if there are missing columns?

jimmygchen: Doesn't `slot_and_column_index_to_data_columns` not empty mean we have some extra columns instead of missing?

beacon_node/beacon_chain/src/historical_data_columns.rs
```diff
@@ -0,0 +1,145 @@
...
+                missing_slots_and_data_columns: slot_and_column_index_to_data_columns
+                    .keys()
+                    .cloned()
+                    .collect::<Vec<_>>(),
+            });
+        }
+
+        self.data_availability_checker
+            .custody_context()
+            .backfill_custody_count_at_epoch(epoch);
```
jimmygchen: It's not clear to me this is a write operation, can we add `update_` to the function name?

eserilev:
Sorry I'm not comfortable getting this merged today. The changes I've made over the weekend are quite large. They should make custody backfill sync better, but I need to test this branch a bit further

beacon_node/http_api/src/lib.rs
eserilev: nah the slot here shouldnt matter much, this is just supposed to recreate a cgc change at some slot. it could be any random slot here

jimmygchen:
Hey @eserilev no worries, I also started reviewing yesterday but didn't get to finish. I'll submit my comment so far and finish the review after the release.

Thanks for pushing hard on this though, great effort 👏  I'll label this `v8.0.0 `.

jimmygchen (dismissed):

beacon_node/network/src/sync/block_sidecar_coupling.rs
eserilev: i'm going to avoid doing this for now if thats okay. I've added a beacon chain and epoch field to `RangeDataColumnBatchRequest` so the fields dont exactly match up with `RangeBlockDataRequest::DataColumns`

I think theres an opportunity to refactor here, and I'm hoping to take care of that in a future PR

beacon_node/beacon_chain/src/historical_data_columns.rs
eserilev: updated function name

beacon_node/beacon_chain/src/historical_data_columns.rs
eserilev: updated log message

beacon_node/beacon_chain/src/historical_data_columns.rs
eserilev: updated comment

common/eth2/src/types.rs
eserilev: removed unused derives

beacon_node/network/src/sync/range_sync/chain.rs
eserilev: removed

beacon_node/network/src/sync/block_sidecar_coupling.rs
eserilev: updated

eserilev:
There was another bug I came across while testing. If there was an epoch with no data columns in any slot, Custody backfill sync would not progress to the next epoch. I've fixed that issue

beacon_node/beacon_chain/src/historical_data_columns.rs
eserilev: [f7f9147](https://github.com/sigp/lighthouse/pull/7907/commits/f7f914736bb2d5079713fd5904cb7cb8122123d6)

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -56,6 +60,213 @@ enum RangeBlockDataRequest<E: EthSpec> {
     },
 }
 
+pub struct RangeDataColumnBatchRequest<T: BeaconChainTypes> {
```
dapplion: This has no coupling now since it's only columns, can you move to another file for clarity?

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -56,6 +60,213 @@ enum RangeBlockDataRequest<E: EthSpec> {
...
+        column_to_peer: HashMap<ColumnIndex, PeerId>,
+        expected_custody_columns: &HashSet<ColumnIndex>,
+        attempt: usize,
+    ) -> Result<DataColumnSidecarList<T::EthSpec>, CouplingError> {
+        let mut naughty_peers = vec![];
+        let mut result: DataColumnSidecarList<T::EthSpec> = vec![];
+
+        let forward_blocks_iter = self
+            .beacon_chain
+            .forwards_iter_block_roots_until(
```
dapplion: What happens if we try to iterate into a slot that's less than the min known block root?

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
```diff
@@ -0,0 +1,1118 @@
...
+        if let Some(batch) = self.batches.get_mut(&batch_id) {
+            let synced_peers = self
+                .network_globals
+                .peers
+                .read()
+                .synced_peers_for_epoch(batch_id, None)
+                .cloned()
+                .collect::<HashSet<_>>();
+
+            let request = batch.to_data_columns_by_range_request().map_err(|_| {
```
dapplion, line 1019: Here you can read from DB blinded blocks + block roots in the epoch `batch_id`
- Inform the by range request of those roots so coupling doesn't need to re-fetch them
- Make the by_range requests known exactly how many blocks and which ones they expect to fetch + which ones have data

Then we can get rid of all the `AwaitingValidation` + `advance_custody_backfill_sync` as we can know ahead of time if an epoch of data columns will be empty or not

eserilev: > Here you can read from DB blinded blocks + block roots in the epoch batch_id
>
>    - Inform the by range request of those roots so coupling doesn't need to re-fetch them
>    - Make the by_range requests known exactly how many blocks and which ones they expect to fetch + which ones have data

We need to fetch block data to ensure that:
- the slot isnt a skipped slot
- the block in the slot is expecting blobs

I'm not sure what the benefit is for reading the blinded blocks before making the by range request? 

> Then we can get rid of all the AwaitingValidation + advance_custody_backfill_sync as we can know ahead of time if an epoch of data columns will be empty or not

I think its going to be very unlikely that a whole epoch doesn't contain a single blob on mainnet. Optimizing for that scenario seems a bit overkill 

I might not be fully understanding your comment, maybe we can hop on a call monday to talk through this

beacon_node/network/src/sync/block_sidecar_coupling.rs
eserilev: i'm not sure if that's possible within the context of custody backfill sync. the sync manager allows custody backfill sync to start only when `anchor_info.block_backfill_complete()` returns true, i.e. if the db contains blocks up to the `genesis_backfill_slot`. On top of that custody backfill sync can only iterate back to the column da boundary.

I don't think it's possible for the column da boundary to be less than the `genesis_backfill_slot`. I could add a check to pause custody backfill sync and throw an `InvalidState` error if `genesis_backfill_slot` is greater than the column da boundary, but I dont think its a scenario that could happen in the first place

eserilev: if that assumption is incorrect lmk, i could always make custody backfill sync iterate back to `max(genesis_backfill_slot, column_da_boundary)`

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
dapplion: > I think its going to be very unlikely that a whole epoch doesn't contain a single blob on mainnet. Optimizing for that scenario seems a bit overkill

Some complexity of your code is there exactly to deal with the scenario where a full epoch has 0 blobs. Let's jump on a call to give you the full info :)

common/eth2/src/lighthouse/sync_state.rs
```diff
@@ -65,8 +84,8 @@ impl SyncState {
             SyncState::SyncingFinalized { .. } => true,
             SyncState::SyncingHead { .. } => true,
             SyncState::SyncTransition => true,
-            // Backfill doesn't effect any logic, we consider this state, not syncing.
-            SyncState::BackFillSyncing { .. } => false,
+            // Both backfill and custody backfill don't effect any logic, we consider this state, not syncing.
```
pawanjay176, line 87:
```suggestion
            // Both backfill and custody backfill don't effect any logic, we consider this state, not syncing.
```

beacon_node/beacon_chain/src/historical_data_columns.rs
```diff
@@ -0,0 +1,151 @@
...
+                        &block_root,
+                        data_column.clone(),
+                        &mut ops,
+                    );
+                    total_imported += 1;
+                }
+            }
+        }
+
+        // If we've made it to here with no columns to import, this is a valid situation
```
pawanjay176: This isn't a valid situation right?

beacon_node/beacon_chain/src/validator_custody.rs
pawanjay176: can we delete this now?

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -332,6 +361,16 @@ impl<E: EthSpec> CustodyContext<E> {
         &all_columns_ordered[..num_of_columns_to_sample]
     }
 
+    // TODO(custody-sync) delete this once it becoms unusued (after testing)
```
pawanjay176: can be deleted too now i guess?

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -3028,6 +3029,245 @@ async fn weak_subjectivity_sync_test(
...
+        .unwrap();
+
+    for block in block_root_iter {
+        let (block_root, _) = block.unwrap();
+        let data_columns = harness.chain.store.get_data_columns(&block_root).unwrap();
+        assert!(data_columns.is_some())
+    }
+}
+
+// This should verify that a data column sidecar containing mismatched block roots should fail to be imported.
```
pawanjay176, line 3252: Nice tests 👌

beacon_node/http_api/src/lib.rs
pawanjay176: can delete now?

beacon_node/network/src/network_beacon_processor/mod.rs
```diff
@@ -492,6 +492,22 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+        process_id: Epoch,
+        data_columns: DataColumnSidecarList<T::EthSpec>,
+    ) -> Result<(), Error<T::EthSpec>> {
+        let processor = self.clone();
+        let process_fn = move || processor.process_historic_data_columns(process_id, data_columns);
+
+        let work = Work::ChainSegmentBackfill(Box::new(process_fn));
+
+        self.try_send(BeaconWorkEvent {
+            drop_during_sync: true,
```
pawanjay176, line 507: I agree this should be dropped during sync, but this can't be called during sync afaiu.
Just curious because regular backfill seems to have it set to false

beacon_node/network/src/sync/manager.rs
```diff
@@ -1198,72 +1303,127 @@ impl<T: BeaconChainTypes> SyncManager<T> {
     /// blobs.
     fn on_range_components_response(
         &mut self,
-        range_request_id: ComponentsByRangeRequestId,
+        range_request_id: ColumnsByRangeParentRequestId,
```
pawanjay176: the naming is weird here. We get a range_request_id which is of type `ColumnsByRangeParentRequestId` but it could also potentially be blobs or no blobs?

beacon_node/lighthouse_network/src/service/api_types.rs
```diff
@@ -60,15 +60,38 @@ pub struct BlobsByRangeRequestId {
 pub struct DataColumnsByRangeRequestId {
     /// Id to identify this attempt at a data_columns_by_range request for `parent_request_id`
     pub id: Id,
-    /// The Id of the overall By Range request for block components.
-    pub parent_request_id: ComponentsByRangeRequestId,
+    /// The Id of the overall By Range request for either a components by range request or a custody backfill request.
+    pub parent_request_id: ColumnsByRangeParentRequestId,
```
pawanjay176: The naming change here seems misleading. What we want is maybe something like

```rust
pub enum ComponentsByRange {
    ComponentsWithBlocks(ComponentsByRangeRequestId),
    CustodyBackfillSync(CustodyBackFillBatchRequestId),
}
```

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -429,6 +432,112 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                        data_column_block_root,
+                        expected_block_root,
+                    } => {
+                        debug!(
+                            error = "no_block_found",
+                            ?data_column_block_root,
+                            ?expected_block_root,
+                            "Custody backfill batch processing error"
+                        );
+                        // The peer is faulty if they send blocks with bad roots.
```
pawanjay176, line 467: Since we are coupling before sending the columns, shouldn't this be an internal error?

pawanjay176 (changes requested):
Few more comments

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
```diff
@@ -0,0 +1,1162 @@
...
+            .custody_columns_for_epoch(None, &self.beacon_chain.spec)
+            .iter()
+            .cloned()
+            .collect::<HashSet<_>>();
+
+        let current_columns_at_epoch = custody_context
+            .custody_columns_for_epoch(Some(epoch), &self.beacon_chain.spec)
+            .iter()
+            .cloned()
+            .collect::<HashSet<_>>();
```
dapplion: I don't follow this logic, could you provide an example or test that it works?

beacon_node/network/src/sync/manager.rs
```diff
@@ -209,6 +213,36 @@ pub enum BatchProcessResult {
...
+        imported_columns: usize,
+    },
+    /// The custody batch processing failed. It carries whether the processing imported any data columns.
+    FaultyFailure {
+        batch_id: Epoch,
+        penalty: PeerAction,
+    },
+    NonFaultyFailure {
+        batch_id: Epoch,
+    },
```
dapplion: Can be simplified into
```suggestion
    Failure {
        batch_id: Epoch,
        penalty: Option<PeerAction>,
    },
```

dapplion: Or even

```rs
CustodyBatchProcessed { 
  batch_id: Epoch,
  result: Result<(), CustodyBatchProcessError>
}

CustodyBatchProcessError {
  penalty: Option<PeerAction>,
}
```

`imported_columns` is only used for a log that not super relevant and can be moved into the importer function

eserilev: Removed the extra enum, now we only have a Success and failure enum, w/ an optional peer action penalty in the failure case

beacon_node/lighthouse_network/src/service/api_types.rs
eserilev: renamed to `ByRangeParentRequestId`

beacon_node/network/src/network_beacon_processor/mod.rs
eserilev: yeah it cant be called during sync, but didnt think it would hurt to have this flag anyways

beacon_node/beacon_chain/src/validator_custody.rs
eserilev: i'll delete after we test on some of our infra

beacon_node/beacon_chain/src/historical_data_columns.rs
eserilev: It is, ive updated the comments a bit to clarify. This can happen if theres not blobs for an epoch

beacon_node/network/src/network_beacon_processor/sync_methods.rs
eserilev: we are not coupling before sending the columns. I had some code that i didnt mean to push up that may have been insinuating that we do that. But its been removed. We do the column verification against the blocks after receving the column responses (in range_data_columns_batch_request.rs)

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
eserilev: Heres a test case
db8c2e4

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -6913,6 +6913,113 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+                effective_epoch.start_slot(T::EthSpec::slots_per_epoch()),
+            ))?;
+            info!(?effective_epoch, ?cgc_at_effective_epoch, "Updated cgc?");
+        } else {
+            error!(
+                ?cgc_at_effective_epoch,
+                ?cgc_at_earliest_data_colum_epoch,
+                ?effective_epoch,
+                ?earliest_data_column_epoch,
+                "Couldn't update data column custody info"
```
dapplion, line 7034: Should you log the reason for failure?
```rs
let can_update_data_column_custody_info = if cgc_at_effective_epoch != cgc_at_earliest_data_colum_epoch {
  Some(format!("cgc_at_effective_epoch {cgc_at_effective_epoch} != cgc_at_earliest_data_colum_epoch {cgc_at_earliest_data_colum_epoch}"))
} else if effective_epoch != earliest_data_column_epoch - 1 {
  Some(format!("effective_epoch {effective_epoch} not below earliest_data_column_epoch {earliest_data_column_epoch}"))
} else {
  None
}
if let Some(can_update_data_column_custody_info) = can_update_data_column_custody_info {
  // log
} else {
  // update
}
```

eserilev: personally id rather not, i think it adds too much clutter. in my opinion the current message gives enough info to understand what the issue is

beacon_node/network/src/sync/manager.rs
```diff
@@ -624,25 +660,64 @@ impl<T: BeaconChainTypes> SyncManager<T> {
                                 error!(error = ?e, "Backfill sync failed to start");
                             }
                         }
+
+                        // If backfill is complete, check if we have a pending custody backfill to complete
+                        let anchor_info = self.chain.store.get_anchor_info();
+                        if anchor_info.block_backfill_complete(self.chain.genesis_backfill_slot) {
+                            match self.custody_backfill_sync.start(&mut self.network) {
```
pawanjay176, line 659: Maybe we can add a config option like `disable-custody-backfill` and not start it unless its false?

pawanjay176 (approved):
Great work on this.
I have looked through most changes that interact with existing code and I think this doesn't change logic in any existing code paths. I believe you have tested this enough in terms of the actual backfill functionality and we can fix small issues if they come up.

Just a small suggestion to have a kill switch cli flag in the manager to avoid starting custody-backfill altogether in case it is required. I don't think it will be required, but we can add it if we want to be extra safe. Also, could we run this on a few of our testnet infra before merging?

eserilev:
Thanks for the multiple rounds of reviews! And for all the help in general. 

I think the plan is to have the release candidate branch running on some testnet nodes over night and then cut the release tomorrow. 

If I have time later tonight I'll hijack a holesky node and trigger custody backfill sync on one of them

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -3169,6 +3171,245 @@ async fn weak_subjectivity_sync_test(
...
+    assert!(matches!(
+        error,
+        HistoricalDataColumnError::NoBlockFound { .. }
+    ));
+}
+
+// This should verify that a data column sidecar associated to a block root that doesn't exist in the store cannot
+// be imported.
+#[tokio::test]
+async fn test_import_historical_data_columns_batch_no_block_found() {
```
dapplion, line 3338: These two tests are identical, except that the first makes half the sidecars invalid, and the second one makes all of them invalid. 
https://www.diffchecker.com/mlBWqUKJ/

Are both tests necessary? If so, can you de-dup of make it obvious in the name?

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -4786,6 +5027,166 @@ async fn test_custody_column_filtering_supernode() {
     );
 }
 
+#[tokio::test]
+async fn test_missing_columns_after_cgc_change() {
+    let spec = test_spec::<E>();
+
+    let num_validators = 8;
```
dapplion, line 5093: Why 8 and not LOW_VALIDATOR_COUNT?

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -4786,6 +5027,166 @@ async fn test_custody_column_filtering_supernode() {
...
+    harness
+        .chain
+        .update_data_column_custody_info(Some(head_slot));
+
+    // We can only safely update custody column info 1 epoch at a time
+    // Skipping an epoch should return an error
+    harness
+        .chain
+        .safely_backfill_data_column_custody_info(head_slot.epoch(E::slots_per_epoch()) - 2)
+        .unwrap_err();
```
dapplion, line 5222: Should you assert the exact error to prevent false negatives?

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -4786,6 +5027,166 @@ async fn test_custody_column_filtering_supernode() {
...
+
+    // Iterate from the head epoch back to 0 and try to backfill data column custody info
+    for epoch in (0..head_slot.epoch(E::slots_per_epoch()).into()).rev() {
+        // This is an epoch before the cgc change took into effect, we shouldnt be able to update
+        // without performing custody backfill sync
+        if epoch <= epoch_after_increase.into() {
+            harness
+                .chain
+                .safely_backfill_data_column_custody_info(Epoch::new(epoch))
+                .unwrap_err();
```
dapplion, line 5232: Same, assert specific error

beacon_node/network/src/sync/batch.rs
```diff
@@ -132,36 +115,36 @@ impl<E: EthSpec, B: BatchConfig> fmt::Display for BatchInfo<E, B> {
 
 #[derive(Display)]
 /// Current state of a batch
-pub enum BatchState<E: EthSpec> {
+pub enum BatchState<D: Hash> {
```
dapplion, line 118: @eserilev re-ordered the file to match unstable and parse the diff better. I found that you derived Debug on BatchState, which would dump all the blocks or columns to the command line. Re-added the custom impl of Debug on BatchState that we have on unstable.

beacon_node/network/src/sync/batch.rs
```diff
@@ -60,28 +59,10 @@ pub trait BatchConfig {
     /// Note that simpler hashing functions considered in the past (hash of first block, hash of last
     /// block, number of received blocks) are not good enough to differentiate attempts. For this
     /// reason, we hash the complete set of blocks both in RangeSync and BackFillSync.
-    fn batch_attempt_hash<E: EthSpec>(blocks: &[RpcBlock<E>]) -> u64;
+    fn batch_attempt_hash<D: Hash>(data: &D) -> u64;
```
dapplion, line 62: Out of scope from this PR, but this mechanism does not make sense anymore. We are hashing RpcBlocks which contains data from **multiple** peers, so it cannot "track attempts" and downscore peers appropiately. If you look at the downscoring logic, it only affects block peers, while we track the responses of all peers (block + columns).

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
```diff
@@ -0,0 +1,1115 @@
...
+    /// The batch must exist and be ready for processing
+    fn process_batch(
+        &mut self,
+        network: &mut SyncNetworkContext<T>,
+        batch_id: BatchId,
+    ) -> Result<ProcessResult, CustodyBackfillError> {
+        // Check if we need to restart custody backfill sync due to a recent cgc change
+        if self.restart_if_required() {
+            return Ok(ProcessResult::Successful);
+        }
```
dapplion, line 474: I need to think more about this, but this restart strategy feels fragile. Let's label each custody backfill sync run by an integer. The first run is run 1. If a restart if needed the next run is run 2. At a given instant, there could be up to 5 batches of run 1 being downloaded and one batch processing. When those complete sync will receive events. It's possible that those events conflict with batches recently created for run 2. In that case we can have `WrongState` errors.

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
```diff
@@ -0,0 +1,1115 @@
...
+                                self.fail_sync(CustodyBackfillError::BatchInvalidState(
+                                    *batch_id, e.0,
+                                ))
+                                .map(|_| ProcessResult::Successful)
+                            }
+                            Ok(BatchOperationOutcome::Failed { blacklist: _ }) => {
+                                warn!(
+                                    score_adjustment = ?peer_action,
+                                    batch_epoch = %batch_id,
+                                    "Custody backfill batch failed to download. Penalizing peers"
```
dapplion, line 771: You should actually penalize peers here

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
```diff
@@ -0,0 +1,1115 @@
...
+    ) -> Result<ProcessResult, CustodyBackfillError> {
+        // Only process batches if custody backfill is syncing and only process one batch at a time
+        if self.state() != CustodyBackFillState::Syncing || self.current_processing_batch.is_some()
+        {
+            return Ok(ProcessResult::Successful);
+        }
+
+        // Don't try to process batches before the Fulu fork epoch since data columns don't exist
+        if let Some(fulu_fork_epoch) = self.beacon_chain.spec.fulu_fork_epoch
+            && self.processing_target < fulu_fork_epoch
```
dapplion, line 815: You should make this logic uniform with the is_complete check. Otherwise this could inifinte loop, backfill syncs thinks it's not complete but it never requests anything. Is it possible to process batches from before the fulu fork if there are never downloaded? Not sure why this check is necessary here.

beacon_node/network/src/sync/custody_backfill_sync/mod.rs
dapplion: Fixed with https://github.com/sigp/lighthouse/pull/7907/commits/020b7f2247801c22e2b182afd754a1a3f65f3e53

eserilev:
Lion and I pushed up some pretty big changes over the last few hours. I'm trying to get this tested, but unfortunately not comfortable merging until I can get it tested

eserilev:
I've tested on kurtosis and things seem to be working

pawanjay176 (approved):
Went through the remaining changes for peer attribution and the conditions look good to me. Lets get this in!

jimmygchen:
Nice work all, lets go 🚀

michaelsproul:
@mergify requeue

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -99,6 +101,25 @@ impl ValidatorRegistrations {
             None
         }
     }
+
+    /// Updates the `epoch_validator_custody_requirements` map by pruning all values on/after `effective_epoch`
+    /// and updating the map to store the latest validator custody requirements for the `effective_epoch`.
+    pub fn backfill_validator_custody_requirements(&mut self, effective_epoch: Epoch) {
```
jimmygchen, line 107: The changes in this file affects the `CustodyContext` behaviour and i think it's worth covering in unit tests. 

I suspect this may not work well with the semi-supernode changes. I'll make a PR with tests.

PR #7909, opened 2025-08-20
Use Fork variants instead of version for JsonPayload types

macladson:
## Issue Addressed

With Fulu, we increment the engine API version for `get_payload` but we do not also increment `new_payload`. 
In Lighthouse, we have a tight coupling between these version numbers and the Fork variants. 
For example, both `get_payload_v3` and `new_payload_v3` correspond to Deneb, `v4` to Electra, etc. 

However this coupling breaks with Fulu where only `get_payload_v5` is related to Fulu and `new_payload_v4` now also corresponds to Fulu (new_payload_v5 does not exist). While we can work around this, it creates a confusing situation where the versions and Fork variants are out of sync.

See the following code snippet where we are using the v4 endpoint, and yet passing a `V5` payload variant: https://github.com/sigp/lighthouse/blob/522bd9e9c6ac167f2231525e937c9ebbcb86cf6e/beacon_node/execution_layer/src/engine_api/http.rs#L849-L870

## Proposed Changes

1. Remove `new_payload_v5` as it is unused in Fulu.
2. Rename the `JsonExecutionPayload` and `JsonGetExecutionPayloadResponse` types to use Fork variants instead of version variants. This clarifies the relationship between them.

## Other Considerations

The engine API spec does refer to these types by version, however `ExecutionPayloadV4` and above do not actually exist since `ExecutionPayload` hasn't changed since Deneb. If we value close adherence to spec naming, we could consider reverting back to the old versioned JsonPayload types by implementing: https://github.com/sigp/lighthouse/pull/7841. By doing this, the `JsonExecutionPayload` type would stop at `V3` (Deneb). This is actually what is used in [the spec](https://github.com/ethereum/execution-apis/blob/main/src/engine/osaka.md#response)

### DIFF
--- a/beacon_node/execution_layer/src/engine_api.rs
+++ b/beacon_node/execution_layer/src/engine_api.rs
@@ -5,7 +5,7 @@ use crate::http::{
     ENGINE_GET_PAYLOAD_BODIES_BY_HASH_V1, ENGINE_GET_PAYLOAD_BODIES_BY_RANGE_V1,
     ENGINE_GET_PAYLOAD_V1, ENGINE_GET_PAYLOAD_V2, ENGINE_GET_PAYLOAD_V3, ENGINE_GET_PAYLOAD_V4,
     ENGINE_GET_PAYLOAD_V5, ENGINE_NEW_PAYLOAD_V1, ENGINE_NEW_PAYLOAD_V2, ENGINE_NEW_PAYLOAD_V3,
-    ENGINE_NEW_PAYLOAD_V4, ENGINE_NEW_PAYLOAD_V5,
+    ENGINE_NEW_PAYLOAD_V4,
 };
 use eth2::types::{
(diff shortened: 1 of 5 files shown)


michaelsproul (approved):
LGTM, this is a good clarification.

I think this helps us out in the short-term while we get Gloas merged in, and then we can come back for removing the payload type duplication.

ethDreamer:
Yeah let's just merge this and merge #7728 . The next thing i will probably do is rename everything in this file from `Json*` to `Engine*` since the entire reason these objects exist is because their json serialization on the `engine_api` is different from the `beacon_api`. But let's leave that for another PR.

ethDreamer (approved):

jimmygchen (approved):
LGTM too!

macladson:
@Mergifyio requeue

PR #7910, opened 2025-08-20
Update to spec v1.6.0-alpha.5

barnabasbusa:
## Issue Addressed

- https://github.com/ethereum/consensus-specs/pull/4508


michaelsproul:
Don't think we can merge this until there's a new spec release. The spec tests _should_ fail when these values are changed.

barnabasbusa:
alpha 5 should be cooking soon

jimmygchen (approved):
Looks good, thanks! 🙏

PR #7913, opened 2025-08-21
Only send lookup requests to peers that are synced or advacned

jimmygchen:


jimmygchen:
This change is coverd in #7915

PR #7915, opened 2025-08-21
Maintain peers across all data column subnets

### LINKED ISSUE #7865: Maintain peers across all custody column subnets
When a node is not able to connect to supernodes, it sometimes struggle to maintain peers across all 128 data column subnets. 

This means node won't be able to sync or keep up with head if it's not able to retrieve it's custody columns.

Similar to `maintain_sync_committee_peers`, we should try to maintain peers across all subnets, to ensure we can retrieve / publish data correctly.

There's also separate issue to bump peer count if necessary: https://github.com/sigp/lighthouse/issues/6920

jimmygchen:
## Issue Addressed

Closes:
- #7865
- #7855

Changes extracted from earlier PR #7876

This PR fixes two main things with a few other improvements mentioned below:
- Prevent Lighthouse from repeatedly sending `DataColumnByRoot` requests to an unsynced peer, causing lookup sync to get stuck
- Allows Lighthouse to send discovery requests if there isn't enough **synced** peers in the required sampling subnets - this fixes the stuck sync scenario where there isn't enough usable peers in sampling subnet but no discovery is attempted.

## Proposed Changes

- Make peer discovery queries if custody subnet peer count drops below the minimum threshold
- Update peer pruning logic to prioritise uniform distribution across all data column subnets and avoid pruning sampling peers if the count is below the target threshold (2)
- Check sync status when making discovery requests, to make sure we don't ignore requests if there isn't enough synced peers in the required sampling subnets
- Optimise some of the `PeerDB` functions checking custody peers
- Only send lookup requests to peers that are synced or advanced


AgeManning (commented):
I think this is an improvement, but we might want to make the data columns first class citizens. 

The overall goal of this logic originally was to get Lighthouse to reach a steady-state where it never had to do any discoveries. It would find and maintain a uniform set of subnet peers. 

There are two competing factors, discoveries which generate peers and pruning which remove the excess. If the pruning doesn't match what our discovery targets are, we might be in a perpetual state of discovering, then pruning the discovered peers. 

Before data columns, we would discover peers if we needed them for attestation subnets, then prune down to maintain a uniform set of attestation subnets, to prevent any future discoveries. 

With this change, we now have a new driving requirement, which is `has_good_peers_in_custody_subnet()`. We will try and discover peers constantly until we meet this requirement. 

But the pruning logic is still to maintain uniform attestation subnets. We only now don't prune peers that might help with our custody subnet requirement. 

I think now that things have changed, we should prioritize a uniform distribution on the data column custody and as a second priority, manage the attestation subnets. The reason being, is that the attestation subnets don't have a direct `maintain_custody_peers()` like function causing discoveries, and so I think its therefore less of a priority. 

Also, for attestation subnets we really need, we have a min_ttl which prevents them from being pruned when we need them. So we can rely on that to save the crucial ones from being dropped. 

So I think the pruning priorities should now be:
1. Maintain uniform distribution of data columns 
  1. a - Don't remove peers that we need for attestation subnets
   1. b - Dont remove peers that we need for sync committees
  2. If all of the above are satisfied, remove peers to make attestation subnets uniform.

jimmygchen:
Thanks @AgeManning , yeah I think your suggestion makes sense, I'll make this change.

beacon_node/lighthouse_network/src/peer_manager/mod.rs
```diff
@@ -949,6 +966,43 @@ impl<E: EthSpec> PeerManager<E> {
         }
     }
 
+    /// Run discovery query for additional custody peers if we fall below `TARGET_PEERS`.
```
AgeManning:
```suggestion
    /// Run discovery query for additional custody peers if we fall below `MIN_SAMPLING_COLUMN_SUBNET_PEERS`.
```

beacon_node/lighthouse_network/src/peer_manager/mod.rs
```diff
@@ -991,9 +1045,198 @@ impl<E: EthSpec> PeerManager<E> {
         }
     }
 
+    /// Build unified peer subnet information from connected peers.
+    ///
+    /// This creates a unified structure containing all subnet information for each peer,
+    /// excluding trusted peers and peers already marked for pruning.
+    fn build_peer_subnet_info(
```
AgeManning, line 1052: Maybe for a future PR. Rather than calculate this thing for every peer every heartbeat, we just change PeerInfo to store these naturally for each peer.

beacon_node/lighthouse_network/src/peer_manager/mod.rs
```diff
@@ -991,9 +1045,198 @@ impl<E: EthSpec> PeerManager<E> {
...
+                    .push(*peer_id);
+            }
+        }
+
+        custody_subnet_to_peers
+    }
+
+    /// Determine if a peer should be protected from pruning based on various criteria.
+    ///
+    /// Returns true if the peer should be protected (not pruned).
```
AgeManning: Maybe we should list the criteria here to assist future devs to understand this function

beacon_node/lighthouse_network/src/peer_manager/mod.rs
```diff
@@ -991,9 +1045,198 @@ impl<E: EthSpec> PeerManager<E> {
...
+    ) -> bool {
+        // Ensure we don't remove too many outbound peers
+        if candidate_info.info.is_outbound_only()
+            && self.target_outbound_peers()
+                >= connected_outbound_peer_count.saturating_sub(outbound_peers_pruned)
+        {
+            return true;
+        }
+
+        // Check data column sampling subnets
```
AgeManning, line 1136:
```suggestion
        // Check data column sampling subnets
        // If the peer exists in a smapling subnet that is less than or equal to MIN_SAMPLING_COLUMN_SUBNET_PEERS, we keep it
```

beacon_node/lighthouse_network/src/peer_manager/mod.rs
```diff
@@ -1023,7 +1266,10 @@ impl<E: EthSpec> PeerManager<E> {
     /// Prune peers in the following order:
     /// 1. Remove worst scoring peers
     /// 2. Remove peers that are not subscribed to a subnet (they have less value)
-    /// 3. Remove peers that we have many on any particular subnet
+    /// 3. Remove peers that we have many on any particular subnet, with some exceptions
+    ///    - Don't remove peers needed for data column sampling (≥ MIN_SAMPLING_COLUMN_SUBNET_PEERS)
+    ///    - Don't remove peers needed for sync committees (>=MIN_SYNC_COMMITTEE_PEERS)
+    ///    - Don't remove peers from the lowest density attestation subnets
     /// 4. Randomly remove peers if all the above are satisfied
```
AgeManning: I think we are now violating this last point. 

Assuming no peers commit any offense (no downscoring), it looks like we have a minimal set of peers, i.e:
MIN_SAMPLING_COLUMN_SUBNET_PEERS * no_of_subnets.  Because if we fill up all the subnets, to this constant, then all of those peers are protected from pruning. 

If our --target-peers is less than this number, I think we no longer prune down to our target. 

The 4th point here, was to just select random peers to prune down to our target.

beacon_node/lighthouse_network/src/peer_manager/mod.rs
```diff
@@ -1087,146 +1333,54 @@ impl<E: EthSpec> PeerManager<E> {
...
-                            removed_peer_index = Some(index);
-                            break;
+                    .map(|(subnet_id, _)| *subnet_id);
+
+                if let Some(densest_subnet) = custody_subnet_with_most_peers {
+                    // If we have successfully found a candidate peer to prune, prune it,
+                    // otherwise all peers on this subnet should not be removed due to our
+                    // outbound limit or min_subnet_count. In this case, we remove all
+                    // peers from the pruning logic and try another subnet.
+                    if let Some(candidate_peer) = self.find_prune_candidate(
```
AgeManning, line 1362: I think there is a bug here. 

We have built a set of `peers_to_prune`. But I dont think these peers are being accounted for when counting the peers on the columns. 

For example; if (because of scoring) we prune a column down to 1 peer, I think potentially we could still prune that last peer as a candidate, because we've not accounted for the peers that we are about to prune. I think we need to parse the `peers_to_prune` mapping into this and subsequent functions to ensure we're dynamically accounting for peers being pruned as we go.

beacon_node/lighthouse_network/src/discovery/mod.rs
```diff
@@ -1248,10 +1248,11 @@ mod tests {
         .unwrap();
         let globals = NetworkGlobals::new(
             enr,
-            MetaData::V2(MetaDataV2 {
+            MetaData::V3(MetaDataV3 {
```
AgeManning, line 1251: This is backwards compat right?

AgeManning (commented):
Left some comments

beacon_node/lighthouse_network/src/peer_manager/mod.rs
jimmygchen: We are removing the peer from the `custody_subnet_to_peers` map after we find a prune candidate, and passing that down to `find_prune_candidate()` and `should_protect_peer`, and the updated map is used when counting the peers, is this what you meant?

jimmygchen: https://github.com/sigp/lighthouse/blob/4a311980d0e9ba33a1223a1af03eb15dc0b2cea2/beacon_node/lighthouse_network/src/peer_manager/mod.rs#L1372-L1374

beacon_node/lighthouse_network/src/discovery/mod.rs
jimmygchen: Yes, we only use the cgc value if metadatav3 is returned

https://github.com/sigp/lighthouse/blob/4a311980d0e9ba33a1223a1af03eb15dc0b2cea2/beacon_node/lighthouse_network/src/peer_manager/mod.rs#L756-L762

a peer that returned v2 are just more likely to get pruned as it doesn't have any `custody_subnets`.

I'm just updating the test here to use the latest metadata.

beacon_node/lighthouse_network/src/peer_manager/mod.rs
jimmygchen: yeah that's true - this is true for sync committee too right? but the numbers are a lot lower.
and it also depends on the custody group count of each peers, by default each full node will subscribe to a min of 8 subnets but only advertise 4 of them.

I'll update the comment!

beacon_node/lighthouse_network/src/peer_manager/mod.rs
```diff
@@ -991,9 +1045,199 @@ impl<E: EthSpec> PeerManager<E> {
...
+        // Ensure we don't remove too many outbound peers
+        if candidate_info.info.is_outbound_only()
+            && self.target_outbound_peers()
+                >= connected_outbound_peer_count.saturating_sub(outbound_peers_pruned)
+        {
+            return true;
+        }
+
+        // Check data column sampling subnets
+        // If the peer exists in a smapling subnet that is less than or equal to MIN_SAMPLING_COLUMN_SUBNET_PEERS, we keep it
```
ackintosh:
```suggestion
        // If the peer exists in a sampling subnet that is less than or equal to MIN_SAMPLING_COLUMN_SUBNET_PEERS, we keep it
```

beacon_node/lighthouse_network/src/peer_manager/mod.rs
```diff
@@ -1087,146 +1334,54 @@ impl<E: EthSpec> PeerManager<E> {
...
-                                    {
-                                        *sync_committee_count =
-                                            sync_committee_count.saturating_sub(1);
-                                    }
-                                }
-                            }
-                            peers_to_prune.insert(candidate_peer);
-                        } else {
-                            peers_on_subnet.clear();
+                        // Simple cleanup: remove from all reverse lookups
```
ackintosh: I think `candidate_peer` should also be removed from `peer_subnet_info`, since `peer_subnet_info` is used to [check sync committee protection](https://github.com/sigp/lighthouse/blob/ac778c6bf91646ec9dfefc8784ddc31db48e8597/beacon_node/lighthouse_network/src/peer_manager/mod.rs#L1148-L1155).

ackintosh (commented):
Thanks for this PR, Jimmy! I've left some comments.

beacon_node/lighthouse_network/src/peer_manager/mod.rs
AgeManning: oh yeah checks out.

AgeManning (approved):
This looks good to me :)

beacon_node/lighthouse_network/src/peer_manager/mod.rs
jimmygchen: yeah good idea, with higher peer count it makes even more sense. I'll raise an issue for this.

beacon_node/lighthouse_network/src/peer_manager/mod.rs
jimmygchen: Oh yes nice catch. I'll make a fix.

beacon_node/lighthouse_network/src/peer_manager/mod.rs
jimmygchen: @ackintosh found a related bug with another map - I'm fixing it now.

beacon_node/lighthouse_network/src/peer_manager/mod.rs
jimmygchen: Added doc in [20483f9f](https://github.com/sigp/lighthouse/pull/7915/commits/20483f9feb7594c2d44b3972cdc49d295e66d014)

beacon_node/lighthouse_network/src/peer_manager/mod.rs
jimmygchen: Fixed in https://github.com/sigp/lighthouse/pull/7915/commits/20483f9feb7594c2d44b3972cdc49d295e66d014

beacon_node/lighthouse_network/src/peer_manager/mod.rs
jimmygchen: Fixed in https://github.com/sigp/lighthouse/pull/7915/commits/20483f9feb7594c2d44b3972cdc49d295e66d014

beacon_node/lighthouse_network/src/peer_manager/mod.rs
jimmygchen: Updated comment in https://github.com/sigp/lighthouse/pull/7915/commits/20483f9feb7594c2d44b3972cdc49d295e66d014

AgeManning (approved):

PR #7916, opened 2025-08-21
use DEFAULT_TARGET_PEERS for target peers everywhere

jxs:
## Issue Addressed
Was going to leave this as a comment on #7877 but when noticed it had already been merged.
we have `DEFAULT_TARGET_PEERS` which was set to 50 and only used on the `Default` impl for `peer_manager`'s `Config`, which then get's overridden by this `lighthouse_network::Config`s default 
This PR unifies everything on `DEFAULT_TARGET_PEERS`


jimmygchen (approved):
Good catch! Thanks @jxs

PR #7924, opened 2025-08-22
Use scoped rayon pool for backfill chain segment processing

jimmygchen:
## Issue Addressed

Part of #7866

- Continuation of #7921 

In the above PR, we enabled rayon for batch KZG verification in chain segment processing. However, using the global rayon thread pool for backfill is likely to create resource contention with higher-priority beacon processor work.

## Proposed Changes

This PR introduces a dedicated low-priority rayon thread pool `LOW_PRIORITY_RAYON_POOL` and uses it for processing backfill chain segments.

This prevents backfill KZG verification from using the global rayon thread pool and competing with high-priority beacon processor tasks for CPU resources.

However, this PR by itself doesn't prevent CPU oversubscription because other tasks could still fill up the global rayon thread pool, and having an extra thread pool could make things worse. To address this we need the beacon
  processor to coordinate total CPU allocation across all tasks, which is covered in:
- #7789


eserilev:
In the non-scoped rayon pool case, and zero delay reconstruction, i did see the node briefly become unsynced. Additionally I saw some instances of these log messages
```
WARN  Delayed proposer preparation                  prepare_slot: 263808, validator: 72
WARN  Error signalling fork choice waiter           error: ForkChoiceSignalOutOfOrder { current: Slot(263831), latest: Slot(263830) }, slot: 263830
```

I also saw spikes in gossip attestation times (up to 10s)
tracing

The numbers dropped to the 20ms range as soon as I switched to scoped rayon pool usage (at the 6:00-ish mark in the screenshot above). I also saw no sync issues or `WARN` log messages. This is also when column reconstruction is set to zero delay (i.e. reconstructing as soon as we get half the columns)

With scoped rayon pool usage reconstruction during backfill slows down a bit (as expected)
tracing

eserilev:
I've done a  bunch of testing on backfill with the global rayon pool vs scoped rayon pool usage. The biggest difference is that KZG verification takes more than double the time with a scoped rayon pool (using a  max of~25% of cpu threads) vs the global rayon pool. Additionally I did see lower cpu usage on average in the scoped pool case. My node **did not** have any issues following the chain in either case during backfill. So I can't argue that this change is absolutely necessary. But it is a safe and relatively simple optimization to make. It can potentially help protect nodes from having issues during backfill sync, so maybe thats a good enough reason to include this in our 8.0.0 release candidate

I haven't see any evidence that something like #7789 is required. It seems like the OS scheduler is good enough at figuring things out with the current scoped rayon pool usage. If we were to expand our scoped rayon pool usage to other work events, #7789 could potentially become more relevant.

In a future iteration (or in this PR) we could make the low priority rayon pool configurable via cli flag to give users additional control over the speed of backfill. This is probably unnecessarily at the moment, but could potentially become useful if we expand scoped rayon thread pool usage.

Another TODO could be to introduce a high priority rayon pool and always avoid scheduling rayon tasks on the global thread pool. It remains to be seen if that would be useful considering our current rayon usage.

eserilev:
I've added a high priority rayon pool thats used during backfill when the flag `--disable-backfill-rate-limiting` is used. This effectively speeds up kzg verification 2x vs using the low priority queue. The flag itself already warns that the node may have degraded attestation performance, so I dont think we need to warn users more than we already are.

eserilev:
Something about the changes to the backfill work event are causing a test to fail in a non-deterministic manner. I'm wondering if its because the backfill task is now blocking instead of async?

eserilev:
I made a small tweak to `assert_event_journal_with_timeout` to ignore `WORKER_FREED` and `NOTHING_TO_DO` events for the test case that was failing non-deterministically.

jimmygchen:
Hi @ethDreamer would you mind reviewing this please 🙏

beacon_node/beacon_processor/src/rayon_manager.rs
```diff
@@ -0,0 +1,37 @@
...
+pub const DEFAULT_LOW_PRIORITY_DIVISOR: usize = 4;
+const MINIMUM_LOW_PRIORITY_THREAD_COUNT: usize = 2;
+
+pub struct RayonManager {
+    /// Smaller rayon thread pool for lower-priority, compute-intensive tasks.
+    /// By default ~25% of CPUs or a minimum of 2 threads.
+    pub low_priority_threadpool: Arc<ThreadPool>,
+    /// Larger rayon thread pool for high-priority, compute-intensive tasks.
+    /// By default 100% of CPUs.
+    pub high_priority_threadpool: Arc<ThreadPool>,
```
jimmygchen: If we're allocating work to this thread pool, we might need to be a bit careful with using the global thread pool, because now we can allocate work to:
* global thread pool with `num_cpus` threads
* high_priority_threadpool with `num_cpus` threads
* low priority threadpool with 2 min threads

so we now let rayon use 2 * `num_cpus` + 2 threads, which may even cause more oversubscription.

jimmygchen: I see a few `into_par_iter` usages, but i think the only compute intensive ones are `reconstruction` and `verify_cell_proof_batch`

so maybe we should spawn those tasks with this thread pool too?

beacon_node/beacon_processor/src/rayon_manager.rs
```diff
@@ -0,0 +1,37 @@
+use rayon::{ThreadPool, ThreadPoolBuilder};
+use std::sync::Arc;
+
+pub const DEFAULT_LOW_PRIORITY_DIVISOR: usize = 4;
+const MINIMUM_LOW_PRIORITY_THREAD_COUNT: usize = 2;
```
jimmygchen: Maybe we can even lower this to 1 (suggested by @ethDreamer)
With [EIP-7870](https://eips.ethereum.org/EIPS/eip-7870#cpu), the min requirement for a full node is 4 phyiscal cores / 8 threads, so in production envs -  we should get 2 at the minimum. 

However, if the process is run on a containerised environment with CPU limits (eg. for devnet testing), then we might end up allocating more than we should, e.g. if `max_cpu=4`, we'd end up allocating half of the available threads. 

i think it probably safer to make this as low as possible (1), and this should be fine because as the name suggests, it's low priority - so running on a single thread should be fine.

beacon_node/beacon_processor/src/rayon_manager.rs
```diff
@@ -0,0 +1,37 @@
...
+    /// Smaller rayon thread pool for lower-priority, compute-intensive tasks.
+    /// By default ~25% of CPUs or a minimum of 2 threads.
+    pub low_priority_threadpool: Arc<ThreadPool>,
+    /// Larger rayon thread pool for high-priority, compute-intensive tasks.
+    /// By default 100% of CPUs.
+    pub high_priority_threadpool: Arc<ThreadPool>,
+}
+
+impl RayonManager {
+    pub fn new(low_prio_cpu_divisor: usize) -> Self {
```
jimmygchen: nit: we could implement a `default` function that uses the default constant, so it doesn't require us to import these constants everywhere we create this, and there's a possibility that we introduce more threadpools later with more parameters.

beacon_node/network/src/network_beacon_processor/tests.rs
```diff
@@ -623,11 +625,21 @@ impl TestRig {
         &mut self,
         expected: &[&str],
         timeout: Duration,
+        ignore_worker_freed: bool,
+        ignore_nothing_to_do: bool,
```
jimmygchen, line 674: > I made a small tweak to assert_event_journal_with_timeout to ignore WORKER_FREED and NOTHING_TO_DO events for the test case that was failing non-deterministically.

Do you think the non-deterministism was due to backfill rate limiting? would disabling it fixes it and speed up the tests?

```
let beacon_processor_config = BeaconProcessorConfig {
        enable_backfill_rate_limiting: false,
        ..Default::default()
    };
```

jimmygchen (commented):
@eserilev looks great, I've added some comments. Let me know what you think!

beacon_node/beacon_processor/src/rayon_manager.rs
eserilev: I dont think rayon is doing `2 * num_cpus` threads if we schedule one work event on the global pool and one work event on the high priority thread pool.

This is an article from one of the rayon maintainers 
https://smallcultfollowing.com/babysteps/blog/2015/12/18/rayon-data-parallelism-in-rust/

It's a bit old, but he explains the work stealing technique that rayon does to find idle threads. the tldr there being that rayon just constantly tries to steal idle threads to execute tasks in parallel.

I also checked with our AI overlords, and they mentioned that work scheduled in the global pool might have a bit more difficulty stealing threads from work scheduled in the scoped pool (and vice versa).  

As far as i can tell I think we could actually probably do away with the high priority thread pool (unless we want it to be `num_cpus < max_cpus`) as it doesn't really provide any benefit vs the global thread pool.

jimmygchen: i haven't dug enough, but my understanding was:
* work-stealing only exists between threads in the same threadpool
* "idle thread" means the rayon thread has no task in it's local work queue (not the phyiscal CPU core)

eserilev: yeah I'm just thinking that the global thread pool is providing the same functionality as the high prio thread pool with a bit less effort code wise. Because if we wanted to run something like reconstruction in the scoped high prio thread pool, I think we'd need to make the reconstruction work event a blocking task (like what you did with `ChainSegmentBackFill`)

jimmygchen: Yeah agree!

jimmygchen:
We've just observed that tree hash caching during block processing sometimes can take 5s+. It's likely that the task is sitting in the queues of rayon global threadpool. We might need to limit what uses the global threadpool, or bring back the high prio queue so these important tasks always get executed immediately.

beacon_node/beacon_processor/src/lib.rs
```diff
@@ -1603,7 +1607,17 @@ impl<E: EthSpec> BeaconProcessor<E> {
...
-            Work::ChainSegmentBackfill(process_fn) => task_spawner.spawn_async(process_fn),
+            Work::ChainSegmentBackfill(process_fn) => {
+                if self.config.enable_backfill_rate_limiting {
+                    task_spawner.spawn_blocking_with_rayon(
+                        self.rayon_manager.low_priority_threadpool.clone(),
+                        process_fn,
+                    )
+                } else {
+                    // use the global rayon thread pool if backfill rate limiting is disabled.
+                    rayon::spawn(process_fn)
```
jimmygchen: I think the recent metrics has shown that using the global rayon thread pool doesn't _always_ cause the CPU to be overwhelmed. It might happen more in the case where the node is able to download fast enough from peers, e.g. on devnet-5 we easily reach 200 peers, so downloads are usually easier and compute will always be the bottleneck - and probably explains why it wasn't an issue on denvet-3 with only 20 peers.

We might need to be a bit careful with this, because if we're downloading fast enough and using all the rayon threads, the node may not be able to perform its duties -  if we're backfilling fast enough, the node will frequently struggle with block processing and perform poorly on attestation, because `update_tree_hash_cache` will sometimes be queued until threads are freed up.

jimmygchen: With `rayon::spawn`, we're never sending idle signal on completion, might be easier to just use the existing `task_spawner.spawn_blocking`?

beacon_node/beacon_processor/src/rayon_manager.rs
```diff
@@ -0,0 +1,37 @@
...
+const DEFAULT_LOW_PRIORITY_DIVISOR: usize = 4;
+const MINIMUM_LOW_PRIORITY_THREAD_COUNT: usize = 2;
+
+pub struct RayonManager {
+    /// Smaller rayon thread pool for lower-priority, compute-intensive tasks.
+    /// By default ~25% of CPUs or a minimum of 2 threads.
+    pub low_priority_threadpool: Arc<ThreadPool>,
+    /// Larger rayon thread pool for high-priority, compute-intensive tasks.
+    /// By default 100% of CPUs.
+    pub high_priority_threadpool: Arc<ThreadPool>,
```
jimmygchen: I noticed that we're no longer using `high_priority_threadpool`, should we remove it?

jimmygchen (commented):
Nice, I've left some more comments, i think we're close!

beacon_node/beacon_processor/src/lib.rs
eserilev: woops that wasnt supposed to be here, thanks

beacon_node/beacon_processor/src/rayon_manager.rs
eserilev: ive changed the low prio threadpool to use just one thread

eserilev: I've changed the low prio threadpool to default to ~25% or at a minimum one thread

jimmygchen (commented):
Looks good!

beacon_node/network/src/network_beacon_processor/tests.rs
jimmygchen: Probably not related, we had backfill rate limiting disabled before.

michaelsproul (approved):
LGTM

jimmygchen:
@mergfiy requeue

PR #7925, opened 2025-08-24
Remove ttfb_timeout and resp_timeout

ackintosh:
## Issue Addressed

`TTFB_TIMEOUT` was deprecated in https://github.com/ethereum/consensus-specs/pull/3767.

## Proposed Changes

Remove `ttfb_timeout` from `InboundUpgrade` and other related structs.

(Update)
Also removed `resp_timeout` and also removed the `NetworkParams` struct since its fields are no longer used. https://github.com/sigp/lighthouse/pull/7925#issuecomment-3226886352

## Additional Info

I didn't remove `ttfb_timeout` field from [`ChainSpec`](https://github.com/sigp/lighthouse/blob/b4704eab4ac8edf0ea0282ed9a5758b784038dd2/consensus/types/src/chain_spec.rs#L220) because the field still exists in the [mainnet.yaml](https://github.com/ethereum/consensus-specs/blob/master/configs/mainnet.yaml#L166-L167) file.


pawanjay176 (commented):
LGTM. In terms of removing it from the config values, there is now a spec PR that removes both ttfb and resp timeouts https://github.com/ethereum/consensus-specs/pull/4532

I think we should still keep the resp timeout with a value of 10, not sure if we should change it. cc @AgeManning

AgeManning:
I agree, we should keep the timeout. It would stall sync and cause a bunch of issues if we don't timeout requests.

pawanjay176:
Awesome, @ackintosh can we remove `resp_timeout` config param as well and use a constant value set to 10 seconds for the stream timeout? 
We can merge that and then remove the config values once that gets upstreamed.

ackintosh:
Thanks @pawanjay176! I removed `resp_timeout` as well, and also removed the `NetworkParams` struct since its fields are no longer used. The `NetworkParams::max_payload_size` field had already been unused before.

pawanjay176 (approved):
LGTM, removed the `NetworkParams` struct like you mentioned and merged unstable.
Upstream PR also got merged, I think we should remove other references of `ttfb_timeout` and `resp_timeout` from the `ChainSpec` in a separate PR to give time for tooling to catch up. Will merge once CI passes

PR #7927, opened 2025-08-25
Fix RPC blocks not getting fully KZG verified

jimmygchen:
## Issue Addressed

Fix RPC blocks not getting fully KZG verified due to incorrect list truncation.

### DIFF
--- a/beacon_node/beacon_chain/src/data_availability_checker.rs
+++ b/beacon_node/beacon_chain/src/data_availability_checker.rs
@@ -19,7 +19,8 @@ use task_executor::TaskExecutor;
 use tracing::{debug, error, instrument};
 use types::blob_sidecar::{BlobIdentifier, BlobSidecar, FixedBlobSidecarList};
 use types::{
-    BlobSidecarList, ChainSpec, DataColumnSidecar, DataColumnSidecarList, Epoch, EthSpec, Hash256, SignedBeaconBlock, Slot,
+    BlobSidecarList, ChainSpec, DataColumnSidecar, DataColumnSidecarList, Epoch, EthSpec, Hash256,
+    SignedBeaconBlock, Slot,
 };
 
 mod error;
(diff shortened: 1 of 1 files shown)


beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -445,8 +445,6 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
             .flatten()
             .map(CustodyDataColumn::into_inner)
             .collect::<Vec<_>>();
-        let all_data_columns =
-            RuntimeVariableList::from_vec(all_data_columns, T::EthSpec::number_of_columns());
```
michaelsproul, line 449: What was this line _trying_ to do? I guess it was assuming these columns were for a single block and coercing them to a `RuntimeVariableList` in order to "check" the length?

michaelsproul (approved):
Fix looks correct to me.

michaelsproul:
Do you think it is worth adding a regression test for this? It seems very specific, and unlikely to happen again, but we could test verification of a batch of columns where the first 128 are valid, and everything after is invalid.

jimmygchen:
> Do you think it is worth adding a regression test for this? It seems very specific, and unlikely to happen again, but we could test verification of a batch of columns where the first 128 are valid, and everything after is invalid.

Yes, definitely worth it, I'll add a test. The bugs we found recently were all because of lack of test coverage from my side 🙏

beacon_node/beacon_chain/src/data_availability_checker.rs
pawanjay176: I think this pattern with the `RuntimeVarList::from_vec` of truncating to the length is a footgun for internal functions (non spec enforcing stuff) because we may be passing a vec with length greater than the `max_len` but the extra items get silently dropped and it breaks stuff elsewhere.

pawanjay176: Ohh just saw you made an issue 🙏

michaelsproul: PR here:

- https://github.com/sigp/lighthouse/pull/7930

jimmygchen: > What was this line _trying_ to do? I guess it was assuming these columns were for a single block and coercing them to a `RuntimeVariableList` in order to "check" the length?

Yeah exactly this i think

pawanjay176 (approved):

PR #7928, opened 2025-08-25
Fix `DataColumnsByRoot` request limit validation bug

### LINKED ISSUE #7926: Seeing immediate stream terminations on `DataColumnsByRoot` messages
@pawanjay176 noticed this issue while running a node, and this seems to be causing problems for sync

> There's something very weird about the DataColumnsByRoot message at the rpc messaging level. A simple rpc test of sending a request and receiving the required response fails by the receiver not showing any request recceived and the sender getting an immediate stream termination

This happens also on LH <> LH nodes, as a simple test in `rpc_tests.rs` confirms it's reproducible.

This failure is likely to cause issues on block lookups, and should also be reproducible with `./genesis-sync.sh` with 120s wait.

jimmygchen:
## Issue Addressed

Fixes #7926

This was a bug I introduced in #7890 and @pawanjay176 noticed it on some running nodes, and added a rpc test to confirm it.

The culprit is this line, where I failed to fill the vec to it's max size, so it doesn't calculate the max size properly, resulting in all `DataColumnByRoot` requests exceeding the max size during validation:
https://github.com/sigp/lighthouse/blob/d24a6d2a45f25dfdba8740d10c569226555a8143/consensus/types/src/chain_spec.rs#L1984

The PR fixes this and includes new regression tests for this fix.

### DIFF
--- a/beacon_node/lighthouse_network/tests/rpc_tests.rs
+++ b/beacon_node/lighthouse_network/tests/rpc_tests.rs
@@ -15,9 +15,10 @@ use tokio::runtime::Runtime;
 use tokio::time::sleep;
 use tracing::{Instrument, debug, error, info_span, warn};
 use types::{
-    BeaconBlock, BeaconBlockAltair, BeaconBlockBase, BeaconBlockBellatrix, BlobSidecar, ChainSpec,
-    EmptyBlock, Epoch, EthSpec, FixedBytesExtended, ForkName, Hash256, MinimalEthSpec,
-    RuntimeVariableList, Signature, SignedBeaconBlock, Slot,
+    BeaconBlock, BeaconBlockAltair, BeaconBlockBase, BeaconBlockBellatrix, BeaconBlockHeader,
+    BlobSidecar, ChainSpec, DataColumnSidecar, DataColumnsByRootIdentifier, EmptyBlock, Epoch,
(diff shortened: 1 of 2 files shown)


beacon_node/lighthouse_network/tests/rpc_tests.rs
```diff
@@ -949,6 +950,303 @@ fn test_tcp_blocks_by_root_chunked_rpc() {
...
+            Arc::downgrade(&rt),
+            current_fork_name,
+            spec.clone(),
+            Protocol::Tcp,
+            false,
+            None,
+        )
+        .await;
+
+        // BlocksByRoot Request
```
jimmygchen: Comment need update

beacon_node/lighthouse_network/tests/rpc_tests.rs
```diff
@@ -949,6 +950,303 @@ fn test_tcp_blocks_by_root_chunked_rpc() {
...
+
+        let rpc_request = RequestType::DataColumnsByRange(DataColumnsByRangeRequest {
+            start_slot: 320,
+            count: 32,
+            columns: (0..128).collect(),
+        });
+
+        // DataColumnsByRoot Response
```
jimmygchen: This one too

pawanjay176 (approved):
LGTM, thanks for fixing this up.

PR #7930, opened 2025-08-25
Delete `RuntimeVariableList::from_vec`

michaelsproul:
## Issue Addressed

This method is a footgun because it truncates the list. It is the source of a recent bug:

- https://github.com/sigp/lighthouse/pull/7927

## Proposed Changes

- Delete uses of `RuntimeVariableList::from_vec` and replace them with `::new` which does validation and can fail.
- Propagate errors where possible, unwrap in tests and use `expect` for obviously-safe uses (in `chain_spec.rs`).

## Additional Info

I've raised a similar PR for `ssz_types` here:

- https://github.com/sigp/ssz_types/pull/52


beacon_node/network/src/sync/network_context/requests/blobs_by_root.rs
```diff
@@ -22,6 +22,7 @@ impl BlobsByRootSingleBlockRequest {
                 .collect(),
             spec,
         )
+        .expect("single block lookup should always satisfy limit for BlobsByRootRequest")
```
pawanjay176: I think the users of these methods return Results so i would definitely prefer propagating the errors

michaelsproul: oh yeah good call, you're right

fixing now

michaelsproul: Fixed in:

da692a6daf43fa5a61a550f2a39d5d98fa14ce06

jimmygchen (approved):
LGTM, thanks for the fix!

PR #7931, opened 2025-08-25
Fix race condition causing data columns getting inserted into DA checker after block is imported

### LINKED ISSUE #6439: Same block getting imported multiple times due to race condition

jimmygchen:
## Issue Addressed

Closes #6439, but also covers additional race conditions described below 

Scenario:
1. Block becomes available via gossip, beacon chain starts [block import](https://github.com/sigp/lighthouse/blob/b18f5872588cbf88f3fd9c4040e81feb495a15a9/beacon_node/beacon_chain/src/beacon_chain.rs#L3853)
2. RPC columns batch arrive and hits `process_rpc_custody_columns` 
    * Fork choice check passes as the block is not yet in fork choice
    * RPC columns get verified 
 3. Block import completes - Block imported to fork choice AND removed from DA checker [here](https://github.com/sigp/lighthouse/blob/b18f5872588cbf88f3fd9c4040e81feb495a15a9/beacon_node/beacon_chain/src/beacon_chain.rs#L3840-L3841)
5. RPC data columns completes verification and gets imported into DA checker

This is not great because we may end up having _some_ data columns in DA checker (as RPC columns may be from multiple requests & peers) and may not be able to fully serve peers' requests.

## Proposed Changes

We had also considered Lion's proposal in this [comment](https://github.com/sigp/lighthouse/issues/6439#issuecomment-2881466485), it is clean but doesn't handle one specific case - when the `PendingComponent` gets removed after block import, if any RPC columns have already made it pass the fork choice check, this race condition may still be possible in a tiny window.

This PR proposes to use a LRU cache of a small size to track recently imported block roots, and performs check before processing block / blobs / columns.

NOTE: this hasn't been tested at all and we should add some tests for the `AlreadyAvailable` code path.

### DIFF
--- a/beacon_node/network/src/network_beacon_processor/sync_methods.rs
+++ b/beacon_node/network/src/network_beacon_processor/sync_methods.rs
@@ -317,6 +317,13 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                 );
                 self.chain.recompute_head_at_current_slot().await;
             }
+            Ok(AvailabilityProcessingStatus::AlreadyAvailable(hash)) => {
+                debug!(
+                    result = "rpc blobs not imported",
+                    block_hash = %hash,
+                    "Block has already been made available"
+                );
+            }
             Ok(AvailabilityProcessingStatus::MissingComponents(_, _)) => {
                 debug!(
                     block_hash = %block_root,
@@ -398,6 +405,13 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                         "Block components retrieved"
                     );
                     self.chain.recompute_head_at_current_slot().await;
+                }
+                AvailabilityProcessingStatus::AlreadyAvailable(hash) => {
+                    debug!(
+                        result = "custody columns not imported",
+                        block_hash = %hash,
+                        "Block components already available"
+                    );
                 }
                 AvailabilityProcessingStatus::MissingComponents(_, _) => {
                     debug!(
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -180,6 +180,7 @@ pub enum WhenSlotSkipped {
 pub enum AvailabilityProcessingStatus {
     MissingComponents(Slot, Hash256),
     Imported(Hash256),
+    AlreadyAvailable(Hash256),
 }
 
 impl TryInto<SignedBeaconBlockHash> for AvailabilityProcessingStatus {
@@ -2949,6 +2950,10 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
                                     ),
                                 };
                             }
+                            AvailabilityProcessingStatus::AlreadyAvailable(block_root) => {
+                                // Block was already available, treat as successful import.
+                                imported_blocks.push((block_root, block_slot));
+                            }
                         }
                     }
                     Err(BlockError::DuplicateFullyImported(block_root)) => {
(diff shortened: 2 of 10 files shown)


beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -399,6 +406,13 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                     );
                     self.chain.recompute_head_at_current_slot().await;
                 }
+                AvailabilityProcessingStatus::AlreadyAvailable(hash) => {
+                    debug!(
+                        result = "custody columns not imported",
+                        block_hash = %hash,
+                        "Block components already available"
+                    );
```
jimmygchen, line 414: I'm worried about the case where block is available but is awaiting import / recompute head. This could be a while and we may not want to send `BlockComponentProcessed` event until then

jimmygchen:
I'm going to close this as this isn't a great solution for the reason mentioned in the above comment. Will create a new PR.

jimmygchen:
@pawanjay176 

I've done another implementation keeping the blocks in the DA checker until it gets pruned, but i don't think it solves the problem mentioned above - we still need to handle the `AlreadyAvailable` case on caller site. It would also require rework of the `PendingComponents` span , as the pending components now stays in DA checker for ~64 blocks.

`AvailabilityProcessingStatus::AlreadyAvailable` is not the same as `AvailabilityProcessingStatus::Imported` or `BlockError::DuplicateFullyImported`, because there's no guarantee that the block has completed import and the caller may not be ready to perform the next operation e.g. 
- Complete the current lookup and process its child
- Recompute head

Without considering a redesign or significant refactor at this point, I see two potential options:
1. Allow block to be re-imported if the race occurs. Block re-import is idempotent and it's already happening anyway - but i think this PR can still fix two things:
    1. Prevent already imported components getting re-inserted into DA checker, potentially causing LH to serve partial data (the main bug we're addressing here)
    2. Prevent making a block available twice in the DA checker, this will _always_ result in state cache miss and state reconstruction [here](https://github.com/sigp/lighthouse/blob/522bd9e9c6ac167f2231525e937c9ebbcb86cf6e/beacon_node/beacon_chain/src/data_availability_checker/state_lru_cache.rs#L116-L120). This actually caused the availability cache write lock to beheld for longer, and make it even more likely to trigger this race.
2. Alternatively, we could store a promise-like object in the [`available_blocks`](https://github.com/sigp/lighthouse/blob/b18f5872588cbf88f3fd9c4040e81feb495a15a9/beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs#L374) LRU cache. This is potentially a cleaner fix but with more complexity and more effort required to make it robust.

Given we're late in the testing phase, I'd prefer avoiding a redesign / major refactor now if we can, and deferring it after the fork would allow us to address this race condition cleanly without adding more complexity now.

jimmygchen:
I made another attempt, i think it's now way simpler. (0 line of code added!) so I'm closing this one.

https://github.com/sigp/lighthouse/pull/7961

PR #7932, opened 2025-08-25
Gloas test fixes

shane-moore:
## Changes
- Until epbs related block production + validation modifications are all set, we will have this test not cover gloas and leave it as a `TODO` [check_all_blocks_from_altair_to_gloas](https://github.com/sigp/lighthouse/blob/92d9f502c42fb48c8fa557d7afb28b5b65072373/beacon_node/beacon_chain/src/beacon_block_streamer.rs#L719-L844)
   - similary, `network` CI/CD tests will be skipped, since most of the tests rely on an `body.execution_payload`, which isn't a thing under epbs anymore
- small formatting changes made since updating to rust `1.89`
- added `builder_pending_payments_limit` to `ethspec`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,11 @@ PROFILE ?= release
 # they run for different forks.
 FORKS=phase0 altair bellatrix capella deneb electra fulu gloas
 
+# List of all hard forks up to gloas. This list is used to set env variables for several tests so that
+# they run for different forks.
+# TODO(EIP-7732) Remove this once we extend network tests to support gloas
+FORKS_BEFORE_GLOAS=phase0 altair bellatrix capella deneb electra fulu
+
 # List of all recent hard forks. This list is used to set env variables for http_api tests
 RECENT_FORKS=electra fulu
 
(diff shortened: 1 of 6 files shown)


beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -5260,7 +5260,9 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
 
         // If required, start the process of loading an execution payload from the EL early. This
         // allows it to run concurrently with things like attestation packing.
-        let prepare_payload_handle = if state.fork_name_unchecked().bellatrix_enabled() {
+        let prepare_payload_handle = if state.fork_name_unchecked().bellatrix_enabled()
+            && !state.fork_name_unchecked().gloas_enabled()
```
ethDreamer:
```suggestion
            && !state.fork_name_unchecked().gloas_enabled()
            // TODO(EIP-7732): once we handle gloas block production, this is where we would typically get an execution bid.. we may reuse the `prepare_payload_handle`..
```

PR #7935, opened 2025-08-26
Add `BEACON_PROCESSOR_WORKERS_ACTIVE_GAUGE_BY_TYPE` metric

jimmygchen:
## Issue Addressed

Similar to `BEACON_PROCESSOR_WORKERS_ACTIVE_TOTAL` but this metric also records the work type.

This is useful in identifying the task when a worker is stuck due to a deadlock or something else, and usually difficult to debug in production / release mode.

### DIFF
--- a/beacon_node/beacon_processor/src/metrics.rs
+++ b/beacon_node/beacon_processor/src/metrics.rs
@@ -42,13 +42,6 @@ pub static BEACON_PROCESSOR_WORKERS_SPAWNED_TOTAL: LazyLock<Result<IntCounter>> 
             "The number of workers ever spawned by the gossip processing pool.",
         )
     });
-pub static BEACON_PROCESSOR_WORKERS_ACTIVE_TOTAL: LazyLock<Result<IntGauge>> =
-    LazyLock::new(|| {
-        try_create_int_gauge(
-            "beacon_processor_workers_active_total",
-            "Count of active workers in the gossip processing pool.",
-        )
-    });
 pub static BEACON_PROCESSOR_WORKERS_ACTIVE_GAUGE_BY_TYPE: LazyLock<Result<IntGaugeVec>> =
     LazyLock::new(|| {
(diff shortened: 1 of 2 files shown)


beacon_node/beacon_processor/src/metrics.rs
```diff
@@ -49,6 +49,14 @@ pub static BEACON_PROCESSOR_WORKERS_ACTIVE_TOTAL: LazyLock<Result<IntGauge>> =
...
         )
     });
+pub static BEACON_PROCESSOR_WORKERS_ACTIVE_GAUGE_BY_TYPE: LazyLock<Result<IntGaugeVec>> =
+    LazyLock::new(|| {
+        try_create_int_gauge_vec(
+            "beacon_processor_workers_active_gauge_by_type",
+            "Int gauge of the number of active workers per work type",
+            &["type"],
+        )
+    });
```
michaelsproul: Do you think we should remove the metric without the worker type? We can always re-derive it by summing across the `type`.

michaelsproul (approved):
LGTM, could also go for deleting the old metric and replacing it by this one, but happy to merge without that change.

beacon_node/beacon_processor/src/metrics.rs
jimmygchen: Yeah i think it would be good to delete it if backward incompatibility is not too much of an issue.

jimmygchen: Removed in https://github.com/sigp/lighthouse/pull/7935/commits/5b4c46c4d7df974210aa9f78cd16e4c82c4c5c4d

michaelsproul:
Yeah I don't think we need to worry about backwards compat for the old metric, probably we're the only ones looking at it. Good call marking this PR as backwards-compat just in case (so we can mention it in the release notes).

PR #7940, opened 2025-08-26
Notify lookup after gossip data column processing resulted in an import

jimmygchen:
## Issue Addressed

When gossip data column processing completes and results in a block import, sync is currently not notified of the successful import. This is inconsistent with how blob processing and block processing both notify sync.

This fix ensures lookup sync receives block import notifications when blocks become available through gossip data column.


pawanjay176 (approved):
Nice!

PR #7942, opened 2025-08-26
Fix data columns by range returning all columns

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

In https://github.com/sigp/lighthouse/pull/7897 , we seem to have modified data columns by range to return all the columns we have for the requested epoch disregarding what columns the peer requested.


jimmygchen (approved):
Thanks for the fix, I've added a regression test as well

jimmygchen:
@mergify requeue

PR #7943, opened 2025-08-26
Fix http api tests ci

hopinheimer:
still got stuff to finish, most of the commits are from #7809


'
michaelsproul, line 1: I got scared by the 8k line diff, but it's just this file that got copied 😅

beacon_node/beacon_chain/src/blob_verification.rs
```diff
@@ -96,7 +96,7 @@ pub enum GossipBlobError {
     /// ## Peer scoring
     ///
     /// We cannot process the blob without validating its parent, the peer isn't necessarily faulty.
-    BlobParentUnknown { parent_root: Hash256 },
+    ParentUnknown { parent_root: Hash256 },
```
michaelsproul, line 99: I feel like leaving this as-is and updating the error parsing might be a little cleaner.

beacon_node/http_api/tests/broadcast_validation_tests.rs
```diff
@@ -85,7 +85,13 @@ pub async fn gossip_invalid() {
     /* mandated by Beacon API spec */
     assert_eq!(error_response.status(), Some(StatusCode::BAD_REQUEST));
 
-    assert_server_message_error(error_response, "BAD_REQUEST: NotFinalizedDescendant { block_parent_root: 0x0000000000000000000000000000000000000000000000000000000000000000 }".to_string());
+    // Since Deneb, the invalidity of the blobs will be detected prior to the invalidity of the
+    // block.
+    let pre_finalized_block_root = Hash256::zero();
+    assert_server_message_error(
+        error_response,
+        format!("BAD_REQUEST: ParentUnknown {{ parent_root: {pre_finalized_block_root:?} }}"),
```
michaelsproul, line 93: This should probably remain as `BlobParentUnknown` per previous comment

beacon_node/http_api/tests/broadcast_validation_tests.rs
```diff
@@ -123,15 +129,11 @@ pub async fn gossip_partial_pass() {
...
-    let response: Result<(), eth2::Error> = tester
+    let response: Result<Response, eth2::Error> = tester
         .client
         .post_beacon_blocks_v2_ssz(&PublishBlockRequest::new(block, blobs), validation_level)
         .await;
-    assert!(response.is_err());
-
-    let error_response = response.unwrap_err();
-
-    assert_eq!(error_response.status(), Some(StatusCode::ACCEPTED));
```
michaelsproul, line 134: We can probably keep an assert on `response.unwrap().status()` being 202 here. It is a little more precise than just asserting `response.is_ok()`.

beacon_node/http_api/tests/broadcast_validation_tests.rs
```diff
@@ -770,25 +805,35 @@ pub async fn blinded_gossip_invalid() {
...
     assert!(response.is_err());
 
     let error_response: eth2::Error = response.err().unwrap();
-
+    let pre_finalized_block_root = Hash256::zero();
     /* mandated by Beacon API spec */
-    assert_eq!(error_response.status(), Some(StatusCode::BAD_REQUEST));
-    assert_server_message_error(error_response, "BAD_REQUEST: NotFinalizedDescendant { block_parent_root: 0x0000000000000000000000000000000000000000000000000000000000000000 }".to_string());
+    if tester.harness.spec.is_fulu_scheduled() {
+        assert_eq!(
```
michaelsproul, line 827: Lets add a comment here like:

```rust
// XXX: this should be a 400 but is a 500 due to the mock-builder being janky
```

beacon_node/http_api/tests/broadcast_validation_tests.rs
```diff
@@ -819,22 +864,18 @@ pub async fn blinded_gossip_partial_pass() {
...
+    let response: Result<Response, eth2::Error> = tester
         .client
-        .post_beacon_blinded_blocks_v2(&block_contents_tuple.0.clone_as_blinded(), validation_level)
+        .post_beacon_blinded_blocks_v2(&blinded_block, validation_level)
         .await;
-    assert!(response.is_err());
-
-    let error_response = response.unwrap_err();
-
-    assert_eq!(error_response.status(), Some(StatusCode::ACCEPTED));
```
michaelsproul, line 837: Again should keep this 202 assert I think

beacon_node/http_api/tests/broadcast_validation_tests.rs
```diff
@@ -952,25 +995,47 @@ pub async fn blinded_consensus_invalid() {
...
         .await;
     assert!(response.is_err());
 
     let error_response: eth2::Error = response.err().unwrap();
 
     /* mandated by Beacon API spec */
-    assert_eq!(error_response.status(), Some(StatusCode::BAD_REQUEST));
-    assert_server_message_error(error_response, "BAD_REQUEST: NotFinalizedDescendant { block_parent_root: 0x0000000000000000000000000000000000000000000000000000000000000000 }".to_string());
+    if tester.harness.spec.is_fulu_scheduled() {
+        assert_eq!(
```
michaelsproul, line 1043: Probably need the same comment about mock-builder as above

beacon_node/http_api/tests/broadcast_validation_tests.rs
```diff
@@ -1000,23 +1065,38 @@ pub async fn blinded_consensus_gossip() {
...
         .await;
-    assert!(response.is_err());
-
-    let error_response: eth2::Error = response.err().unwrap();
 
     /* mandated by Beacon API spec */
-    assert_eq!(error_response.status(), Some(StatusCode::BAD_REQUEST));
-    assert_server_message_error(error_response, "BAD_REQUEST: Invalid block: StateRootMismatch { block: 0x0000000000000000000000000000000000000000000000000000000000000000, local: 0x253405be9aa159bce7b276b8e1d3849c743e673118dfafe8c7d07c203ae0d80d }".to_string());
+    if tester.harness.spec.is_fulu_scheduled() {
+        assert!(response.is_ok());
```
michaelsproul: This doesn't look right. If the block is invalid and we are using `broadcast_validation=consensus`, then we should definitely get an error code here, even if we're running on Fulu.

beacon_node/http_api/tests/broadcast_validation_tests.rs
```diff
@@ -1215,24 +1322,36 @@ pub async fn blinded_equivocation_gossip() {
...
-    assert!(response.is_err());
-
-    let error_response: eth2::Error = response.err().unwrap();
 
     /* mandated by Beacon API spec */
-    assert_eq!(error_response.status(), Some(StatusCode::BAD_REQUEST));
-
-    assert_server_message_error(error_response, "BAD_REQUEST: Invalid block: StateRootMismatch { block: 0x0000000000000000000000000000000000000000000000000000000000000000, local: 0x253405be9aa159bce7b276b8e1d3849c743e673118dfafe8c7d07c203ae0d80d }".to_string());
+    if tester.harness.spec.is_fulu_scheduled() {
+        assert!(response.is_ok())
```
michaelsproul: As above I feel like this is very wrong. Should probably be a 500 like the other Fulu cases.

Oh! I think I know why this might be: in the mock-builder when we submit the full block to the BN we are not using `broadcast_validation=consensus`, we're just using `broadcast_validation=gossip`. I think we can probably get the tests to pass by hardcoding the mock-builder to use `broadcast_validation=consensus_and_equivocation` (when `broadcast_to_bn` is true), which is the validation that real relays use.

beacon_node/http_api/tests/tests.rs
```diff
@@ -1578,12 +1592,14 @@ impl ApiTester {
...
-                .is_err()
-        );
+        let response: Result<Response, Error> = self
+            .client
+            .post_beacon_blocks_v2(&PublishBlockRequest::from(block), None)
+            .await;
+
+        assert!(response.is_ok());
+
+        assert_eq!(response.expect("infailable").status(), StatusCode::ACCEPTED);
```
michaelsproul: typo, but can just unwrap:

```suggestion
        assert_eq!(response.unwrap().status(), StatusCode::ACCEPTED);
```

beacon_node/http_api/tests/tests.rs
```diff
@@ -1606,13 +1622,13 @@ impl ApiTester {
...
-                .await
-                .is_err()
-        );
+        let response: Result<Response, Error> = self
+            .client
+            .post_beacon_blocks_v2(&PublishBlockRequest::from(block), None)
+            .await;
 
+        assert!(response.is_ok());
+        assert_eq!(response.expect("infailable").status(), StatusCode::ACCEPTED);
```
michaelsproul:
```suggestion
        assert_eq!(response.unwrap().status(), StatusCode::ACCEPTED);
```

validator_client/validator_services/src/block_service.rs
```diff
@@ -494,22 +494,28 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> BlockService<S, T> {
...
                     .await
-                    .or_else(|e| {
-                        handle_block_post_error(e, signed_block.signed_block().message().slot())
-                    })?
+                {
+                    Ok(_) => {}
+                    Err(e) => {
+                        handle_block_post_error(e, signed_block.signed_block().message().slot())?
+                    }
+                }
```
michaelsproul: Nit: I'd slightly prefer to just add a `.map(|_| ())` to the existing expression, rather than adding a `match`.

validator_client/validator_services/src/block_service.rs
```diff
@@ -494,22 +494,28 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> BlockService<S, T> {
...
             }
             SignedBlock::Blinded(signed_block) => {
                 let _post_timer = validator_metrics::start_timer_vec(
                     &validator_metrics::BLOCK_SERVICE_TIMES,
                     &[validator_metrics::BLINDED_BEACON_BLOCK_HTTP_POST],
                 );
-                beacon_node
+                match beacon_node
```
michaelsproul: As above

michaelsproul (commented):
Thank you so much for grinding through these tests!

I have a few cosmetic suggestions, and one semantic one. Hopefully none of them are too painful to address (prioritise the semantic one, and we can leave some of the cosmetic ones if need be).

beacon_node/beacon_chain/src/blob_verification.rs
jimmygchen: IIRC the reason for this change is to match the `GossipDataColumnError` variant name to support Fulu: 
https://github.com/sigp/lighthouse/blob/a134d43446f776fe2a84f420854afbff76ca93d8/beacon_node/beacon_chain/src/data_column_verification.rs#L108

beacon_node/http_api/tests/broadcast_validation_tests.rs
hopinheimer: ok hardcoding the mock_builder to use broadcast_validation=consensus_and_equivocation makes sense. made the above the change and the tests are passing with 500

beacon_node/beacon_chain/src/blob_verification.rs
michaelsproul: oh ok yeah, that's fine then

michaelsproul (approved):
This is awesome, thank you so much for fixing this up!

PR #7944, opened 2025-08-26
Replace `INTERVALS_PER_SLOT` with explicit slot component times

eserilev:
## Issue Addressed
https://github.com/ethereum/consensus-specs/pull/4476

## Additional Info

This change affects many critical paths in lighthouse and should not be merged until after `8.0.0`

the Ethereum kurtosis package is working to support this change here:  https://github.com/ethpandaops/ethereum-package/pull/1168


dapplion:
Should this change consider that the slot time will change in the future? Or that's to be handled in a future PR

eserilev:
> Should this change consider that the slot time will change in the future? Or that's to be handled in a future PR

I think we should handle that in a future PR to keep this diff here as small as possible

beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -232,17 +232,17 @@ impl<E: EthSpec> Network<E> {
...
 
         let score_settings = PeerScoreSettings::new(&ctx.chain_spec, gs_config.mesh_n());
 
         let gossip_cache = {
-            let slot_duration = std::time::Duration::from_secs(ctx.chain_spec.seconds_per_slot);
-            let half_epoch = std::time::Duration::from_secs(
-                ctx.chain_spec.seconds_per_slot * E::slots_per_epoch() / 2,
+            let slot_duration = std::time::Duration::from_millis(ctx.chain_spec.slot_duration_ms);
+            let half_epoch = std::time::Duration::from_millis(
+                ctx.chain_spec.slot_duration_ms * E::slots_per_epoch() / 2,
```
eserilev: this should now be able to handle situations where a half epoch is not an integer

beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -232,21 +232,20 @@ impl<E: EthSpec> Network<E> {
...
         );
 
         let score_settings = PeerScoreSettings::new(&ctx.chain_spec, gs_config.mesh_n());
 
         let gossip_cache = {
-            let slot_duration = std::time::Duration::from_secs(ctx.chain_spec.seconds_per_slot);
-            let half_epoch = std::time::Duration::from_secs(
-                ctx.chain_spec.seconds_per_slot * E::slots_per_epoch() / 2,
+            let half_epoch = std::time::Duration::from_millis(
+                (ctx.chain_spec.get_slot_duration().as_millis() as u64) * E::slots_per_epoch() / 2,
```
eserilev, line 244: using millis to handle situations where a half epoch is not an integer

jimmygchen (commented):
🤖 Bug in `testing/simulator/src/basic_sim.rs:178`: The second line overwrites the division result and uses `seconds_per_slot` (seconds) instead of `slot_duration_ms` (milliseconds):

```rust
spec.slot_duration_ms /= speed_up_factor;
spec.slot_duration_ms = max(1_000, spec.seconds_per_slot);  // ❌ Wrong variable
```

Should be:
```rust
spec.slot_duration_ms /= speed_up_factor;
spec.slot_duration_ms = max(1_000, spec.slot_duration_ms);
```

Note that `fallback_sim.rs:184-185` has the correct implementation.

---

🤖 In `consensus/types/src/chain_spec.rs` (`get_slot_component_duration` method): Should we avoid `.expect()` here? This gets called in hot paths for every attestation/sync committee timing check. While overflow is unlikely with valid basis points, could we either return `Result<Duration, ArithError>` or add a comment explaining why the panic is safe? (e.g., the max intermediate value with mainnet parameters is far from u64::MAX)

---

🤖 In `beacon_node/network/src/network_beacon_processor/sync_methods.rs:306`: Stale comment reference mentions `slot_clock.single_lookup_delay()` which was removed in this PR. Should update or remove this reference.

---

🤖 General question: Are there plans to deprecate `seconds_per_slot`? Both fields now exist in `ChainSpec`. Should `seconds_per_slot` be marked deprecated or documented as maintained only for config backward compatibility?

testing/simulator/src/basic_sim.rs
```diff
@@ -175,8 +175,8 @@ pub fn run_basic_sim(matches: &ArgMatches) -> Result<(), String> {
     let latest_fork_version = spec.electra_fork_version;
     let latest_fork_start_epoch = ELECTRA_FORK_EPOCH;
 
-    spec.seconds_per_slot /= speed_up_factor;
-    spec.seconds_per_slot = max(1, spec.seconds_per_slot);
+    spec.slot_duration_ms /= speed_up_factor;
+    spec.slot_duration_ms = max(1_000, spec.seconds_per_slot);
```
jimmygchen: 🤖 Bug: The second line overwrites the division result and uses `seconds_per_slot` (seconds) instead of `slot_duration_ms` (milliseconds):

```rust
spec.slot_duration_ms /= speed_up_factor;
spec.slot_duration_ms = max(1_000, spec.seconds_per_slot);  // ❌ Wrong variable
```

Should be:
```rust
spec.slot_duration_ms /= speed_up_factor;
spec.slot_duration_ms = max(1_000, spec.slot_duration_ms);
```

Note that `fallback_sim.rs:184-185` has the correct implementation.

consensus/types/src/chain_spec.rs
```diff
@@ -865,6 +872,42 @@ impl ChainSpec {
...
+    pub fn get_sync_message_due(&self) -> Duration {
+        self.get_slot_component_duration(self.sync_message_due_bps)
+    }
+
+    /// Calculate the duration into a slot for a given slot component
+    fn get_slot_component_duration(&self, component_basis_points: u64) -> Duration {
+        Duration::from_millis(
+            component_basis_points
+                .safe_mul(self.slot_duration_ms)
+                .expect("should not overflow")
```
jimmygchen: 🤖 Should we avoid `.expect()` here? This gets called in hot paths for every attestation/sync committee timing check. While overflow is unlikely with valid basis points, could we either return `Result<Duration, ArithError>` or add a comment explaining why the panic is safe? (e.g., the max intermediate value with mainnet parameters is far from u64::MAX)

jimmygchen:
🤖 General question: Are there plans to deprecate `seconds_per_slot`? Both fields now exist in `ChainSpec`. Should `seconds_per_slot` be marked deprecated or documented as maintained only for config backward compatibility?

jimmygchen:
🤖 There's a stale comment reference in `beacon_node/network/src/network_beacon_processor/sync_methods.rs:302` that mentions `slot_clock.single_lookup_delay()` which was removed in this PR. Should update or remove this reference.

consensus/types/src/chain_spec.rs
eserilev: [c01abed](https://github.com/sigp/lighthouse/pull/7944/commits/c01abede7af5a1dfe6a80c7f3596feed5083ad4b)

eserilev: I've made it panic safe, though maybe we do want to panic if this calculation fails? It would be weird for a node to keep running if this calc fails.

testing/simulator/src/basic_sim.rs
eserilev: [71e9d22](https://github.com/sigp/lighthouse/pull/7944/commits/71e9d22ae4229dfb737db378cb1686e0c056fd66)

eserilev:
Fixed the above comments

jimmygchen:
Thanks for addressing the feedback! The fixes look good

validator_client/validator_services/src/attestation_service.rs
```diff
@@ -157,10 +157,15 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> AttestationService<S,
 
         let executor = self.executor.clone();
 
+        let unaggregated_attestation_due = self
+            .chain_spec
+            .get_unaggregated_attestation_due()
```
0xmrree: shouldn't function calls like these take in current epoch? - this way the gloas versions of timing vars can fetched when needed

eserilev: thats out of scope for this PR

consensus/types/src/core/chain_spec.rs
```diff
@@ -3099,4 +3147,66 @@ mod yaml_tests {
...
         }
     }
+
+    #[test]
+    fn test_slot_component_duration_calculations() {
+        let spec = ChainSpec::mainnet();
+
+        // Test unaggregated attestation (3333 bps = 33.33% of 12s = 4s)
+        let unagg_due = spec.get_unaggregated_attestation_due().unwrap();
+        assert_eq!(unagg_due, Duration::from_millis(3999)); // 12000 * 3333 / 10000
```
jimmygchen, line 3340: Is it intentional that the attestation deadline is 1ms earlier?

consensus/types/src/core/chain_spec.rs
```diff
@@ -896,6 +898,43 @@ impl ChainSpec {
         )
     }
 
+    /// Get the duration into a slot in which an unaggregated attestation is due
+    pub fn get_unaggregated_attestation_due(&self) -> Result<Duration, ArithError> {
+        self.get_slot_component_duration(self.attestation_due_bps)
```
jimmygchen: Is it worth pre-computing these so we don't have to compute and handle error at runtime? similar to the "Networking Derived" config above

eserilev: Yeah i was a bit on the fence about this. My reasoning for not pre-computing is that when we move to 6 second slot times,  we'll have to compute this at runtime. Also in epbs the attestation times change so this will also need to be computed at runtime. We'll need to add a `slot` as an argument to this function

consensus/types/src/core/chain_spec.rs
eserilev: yeah these deadlines are percentages now, e.g. the spec defines the unagg attestation deadline as 33.33% into the slot which is 3.99 seconds

consensus/types/src/core/chain_spec.rs
```diff
@@ -1783,6 +1823,8 @@ pub struct Config {
     #[serde(with = "serde_utils::quoted_u64")]
     seconds_per_slot: u64,
     #[serde(with = "serde_utils::quoted_u64")]
+    slot_duration_ms: u64,
```
michaelsproul: Needs a `default` for backwards compat with BNs that don't have this field

eserilev:
since the fix to chain spec the errors on `hi-h-vc-byte` have disappeared

image

consensus/types/src/core/chain_spec.rs
eserilev: [489021a](https://github.com/sigp/lighthouse/pull/7944/commits/489021a24bcd3c1c995833b97a3d4f6b58e4015d)

I default `slot_duration_ms` to `seconds_per_slot * 1000` if the value is missing (instead of just defaulting to 12000) so that we don't break certain gnosis and kurtosis set ups

michaelsproul: Yeah I like that

jimmygchen:
🤖 The aggregate and contribution timing in the validator client still uses the old `slot_duration / 3` calculation with TODOs to migrate:

- `attestation_service.rs:253`: `// TODO(slot-duration) this should pull from AGGREGATE_DUE_BPS`
- `sync_committee_service.rs:156`: `// TODO(slot-duration) this should pull from CONTRIBUTION_DUE_BPS`

Are these planned to be addressed in this PR, or is there a tracking issue for follow-up?

eserilev:
I created an issue to track the TODOs as well as the gloas related changes that will be required
https://github.com/sigp/lighthouse/issues/8686

jimmygchen:
One more references that could be replaced here:
* https://github.com/sigp/lighthouse/blob/af1d9b99911e2467a56ce9f308f06ec7a9ead3eb/beacon_node/beacon_chain/src/custody_context.rs#L117

beacon_node/network/src/service.rs
```diff
@@ -862,9 +862,9 @@ impl<T: BeaconChainTypes> NetworkService<T> {
             self.next_digest_update = Box::pin(next_digest_delay(&self.beacon_chain).into());
 
             // Set the next_unsubscribe delay.
-            let epoch_duration =
-                self.beacon_chain.spec.seconds_per_slot * T::EthSpec::slots_per_epoch();
-            let unsubscribe_delay = Duration::from_secs(UNSUBSCRIBE_DELAY_EPOCHS * epoch_duration);
+            let unsubscribe_delay = Duration::from_secs(
+                UNSUBSCRIBE_DELAY_EPOCHS * self.beacon_chain.spec.get_slot_duration().as_secs(),
```
jimmygchen: This should be multiplied by epoch duration instead I believe, otherwise we unsubscribe to the topic in 24 seconds.

beacon_node/beacon_chain/src/validator_custody.rs
```diff
@@ -0,0 +1,790 @@
+use parking_lot::RwLock;
```
jimmygchen: This file no longer exist on `unstable` (renamed to `custody_context` in #8254)

consensus/types/src/core/chain_spec.rs
jimmygchen: The problem with this is, this error can't really be correctly handled - we use `unwrap_or` on all caller sites with fallback values, which causes the node to function incorrectly:

attestation will never be produced:
https://github.com/sigp/lighthouse/blob/57011326c12c8fcdd83e252ed2cd33f536ee0aaf/validator_client/validator_services/src/attestation_service.rs#L160-L162

we'll see an error on every block as it will be detected as late block

https://github.com/sigp/lighthouse/blob/57011326c12c8fcdd83e252ed2cd33f536ee0aaf/beacon_node/http_api/src/publish_blocks.rs#L768-L776

If it's an invalid config value or a bug in this function, the node should really just crash on startup instead of continuing to run.

Would it work if we have those values derived on startup, and have a function that accepts slot and return the derived Duration?

beacon_node/beacon_chain/src/canonical_head.rs
```diff
@@ -1455,7 +1457,10 @@ fn observe_head_block_delays<E: EthSpec, S: SlotClock>(
 
         // Determine whether the block has been set as head too late for proper attestation
         // production.
-        let late_head = attestable_delay >= slot_clock.unagg_attestation_production_delay();
+        let late_head = attestable_delay
+            >= spec
+                .get_unaggregated_attestation_due()
+                .unwrap_or(block_delay_total);
```
jimmygchen: I don't think this is reachable currently, but why is it set to blcok_delay_total?

beacon_node/beacon_chain/src/light_client_finality_update_verification.rs
```diff
@@ -75,7 +75,10 @@ impl<T: BeaconChainTypes> VerifiedLightClientFinalityUpdate<T> {
             .slot_clock
             .start_of(rcv_finality_update.signature_slot())
             .ok_or(Error::SigSlotStartIsNone)?;
-        let one_third_slot_duration = Duration::new(chain.spec.seconds_per_slot / 3, 0);
+        let one_third_slot_duration = chain
```
jimmygchen: We should probably rename this variable

beacon_node/beacon_chain/src/light_client_optimistic_update_verification.rs
```diff
@@ -70,7 +70,12 @@ impl<T: BeaconChainTypes> VerifiedLightClientOptimisticUpdate<T> {
             .slot_clock
             .start_of(rcv_optimistic_update.signature_slot())
             .ok_or(Error::SigSlotStartIsNone)?;
-        let one_third_slot_duration = Duration::new(chain.spec.seconds_per_slot / 3, 0);
+
+        let one_third_slot_duration = chain
```
jimmygchen: please rename this too

beacon_node/beacon_chain/src/validator_monitor.rs
```diff
@@ -1210,13 +1214,15 @@ impl<E: EthSpec> ValidatorMonitor<E> {
...
+        spec: &ChainSpec,
     ) {
         let data = indexed_attestation.data();
         let epoch = data.slot.epoch(E::slots_per_epoch());
         let delay = get_message_delay_ms(
             seen_timestamp,
             data.slot,
-            Duration::from_secs(0),
+            spec.get_unaggregated_attestation_due()
+                .unwrap_or(Duration::from_secs(0)),
```
jimmygchen: I think this should remain unchanged, as it changes the metric defined here:

https://github.com/sigp/lighthouse/blob/c91345782a79d33408dc3c39698207a01116b99a/beacon_node/beacon_chain/src/metrics.rs#L1174

consensus/types/src/core/chain_spec.rs
```diff
@@ -1754,6 +1794,9 @@ pub struct Config {
 
     #[serde(with = "serde_utils::quoted_u64")]
     seconds_per_slot: u64,
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    slot_duration_ms: Option<MaybeQuoted<u64>>,
```
jimmygchen, line 1902: The BPS fields are missing in `Config` struct, which means the values from yaml files are ignored and we always use the values from the hardcoded `ChainSpec`. 

```rust
  pub proposer_reorg_cutoff_bps: u64,
  pub attestation_due_bps: u64,
  pub aggregate_due_bps: u64,
  pub sync_message_due_bps: u64,
  pub contribution_due_bps: u64,
```

jimmygchen (changes requested):
Hey @eserilev 

I've reviewed the PR, please see my comments - there are a few issues but most of the changes look safe and I think we're close to merging this! 

Let me know if you'd like to discuss on any of the comments or if they're unclear.

consensus/types/src/core/chain_spec.rs
eserilev: these have now been added

beacon_node/beacon_chain/src/validator_monitor.rs
eserilev: reverted this

beacon_node/network/src/service.rs
eserilev: fixed

michaelsproul (approved):
LGTM. Peer reviewed with @jimmygchen

jimmygchen (approved):

eserilev:
thanks guys!

PR #7945, opened 2025-08-27
Instrument `publish_block` code path 

jimmygchen:
## Issue Addressed

Instrument `publish_block` code path and log dropped data columns when publishing.

Example spans (running the devnet from my laptop, so the numbers aren't great)

image

image

### DIFF
--- a/beacon_node/http_api/src/publish_blocks.rs
+++ b/beacon_node/http_api/src/publish_blocks.rs
@@ -16,6 +16,7 @@ use eth2::types::{
 use execution_layer::{ProvenancedPayload, SubmitBlindedBlockResponse};
 use futures::TryFutureExt;
 use lighthouse_network::PubsubMessage;
+use lighthouse_tracing::SPAN_PUBLISH_BLOCK;
 use network::NetworkMessage;
 use rand::prelude::SliceRandom;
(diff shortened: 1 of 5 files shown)


beacon_node/http_api/src/publish_blocks.rs
```diff
@@ -532,7 +548,11 @@ fn publish_column_sidecars<T: BeaconChainTypes>(
             .saturating_sub(malicious_withhold_count);
         // Randomize columns before dropping the last malicious_withhold_count items
         data_column_sidecars.shuffle(&mut **chain.rng.lock());
-        data_column_sidecars.truncate(columns_to_keep);
+        let dropped_indices = data_column_sidecars
+            .drain(columns_to_keep..)
+            .map(|d| d.index)
+            .collect::<Vec<_>>();
+        debug!(indices = ?dropped_indices, "Dropping data columns from publishing");
```
jimmygchen, line 555: Added a log here when dropping columns with `--malicious-withhold-count`, this should help test the scenario where proposer doesn't have peers in any custody subnet to publish to

eserilev (approved):
LGTM!

jimmygchen:
@mergify requeue

PR #7957, opened 2025-08-28
Remove `expect` usage in `kzg_utils` 

jimmygchen:
## Issue Addressed

Remove `expect` usage in `kzg_utils` to handle the case where EL sends us invalid proof size instead of crashing.


michaelsproul (approved):
Looks good to me

michaelsproul:
Oh, one small conflict

PR #7958, opened 2025-08-28
Reduce backfill batch buffer size

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Currently, backfill is allowed to create upto 20 pending batches which is unnecessarily high imo. Forward sync also allows a max of 5 batches to be buffered at a time.  This PR reduces the batch size to match with forward sync. 

Having high number of batches is a little annoying with peerdas because we try to create and send 20 requests (even though we are processing them in a rate limited manner). Requests with peerdas is a lot more heavy as we distribute requests across multiple peers leading to lot of requests that may keep getting retried. This could take resources away from processing at head.


jimmygchen (approved):
Yeah I think this makes sense. 

The KZG verification takes a bit of time as well, so I doublt downloading more batches in advance is going to help us much.

jimmygchen:
@mergify requeue

PR #7962, opened 2025-08-29
Avoid serving validator endpoints while the node is far behind syncing head

jimmygchen:
## Issue Addressed

A performance issue was discovered when devnet-3 was under non-finality - some of the lighthouse nodes are "stuck" with syncing because of handling proposer duties HTTP requests.

These validator requests are higher priority than Status processing, and if they are taking a long time to process, the node won't be able to progress. What's worse is - under long period of non finality, the proposer duties calculation function tries to do state advance for a large number of [slots](https://github.com/sigp/lighthouse/blob/d545ddcbc7d97b24b5c15012d1a5f9a1dae90b2a/beacon_node/beacon_chain/src/beacon_proposer_cache.rs#L183) here, causing the node to spend all its CPU time on a task that doesn't really help, e.g. the computed duties aren't useful if the node is 20000 slots behind.

To solve this issue, we use the `not_while_syncing` filter to prevent serving these requests, until the node is synced. This should allow the node to focus on sync under non-finality situations.

### DIFF
--- a/beacon_node/http_api/src/lib.rs
+++ b/beacon_node/http_api/src/lib.rs
@@ -457,7 +457,7 @@ pub fn serve<T: BeaconChainTypes>(
                 move |network_globals: Arc<NetworkGlobals<T::EthSpec>>,
                       chain: Arc<BeaconChain<T>>| async move {
                     match *network_globals.sync_state.read() {
-                        SyncState::SyncingFinalized { .. } => {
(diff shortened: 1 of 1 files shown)


michaelsproul (approved):
LGTM. We need to be careful here, but I think this is safe enough for now as we have an escape hatch in case of liveness issues (`--sync-tolerance-epochs`).

PR #7964, opened 2025-09-01
Remove dependency on target_info.

SamWilsn:
## Proposed Changes

Remove dependency on target_info, use standard library instead.

## Additional Info

The target_info crate hasn't been updated in nine years. While this isn't a problem in and of itself, its functionality is available in Rust's standard library. Because target_info's manifest is basically empty, it makes packaging lighthouse more difficult.


michaelsproul (approved):
Love this! Thank you!

PR #7972, opened 2025-09-01
No default features for store

SamWilsn:
## Proposed Changes

Unless store's default features are disabled, it is impossible to unselect leveldb as a backend.


michaelsproul:
I'm keen for this, but I think we might need a few more changes to support LevelDB-free builds. I had a WIP here that I never quite finished:

- https://github.com/sigp/lighthouse/pull/6964

SamWilsn:
Ah, sorry, I didn't see that! I'll try and come back to this if/when I get my ebuild written.

michaelsproul:
Feel free to pinch changes from my PR if you think they're worthwhile!

Issue #7950, opened 2025-08-27
Remove data column gossip verification during data column publishing


jimmygchen:
We've noticed that data column gossip validation during block publishing takes up a lot of time, especially on lower capacity machines. After some testing (see numbers [here](https://hackmd.io/5qRLqpJaSFy7DOHSuyJWxw#Optimisations)) and internal discussions, we think the verification is redundant because:
- for locally built blocks, EL already verifies the cell proofs
- for externally built blocks, we don't get the blobs at all

The validation time adds up for 128 columns as the verification during block publishing currently runs sequentially on a single thread (as seen from the above linked post and the traces in this PR: https://github.com/sigp/lighthouse/pull/7945)

I think we can remove the gossip validation, but keep the logic to record columns in the observed sidecars cache.

jimmygchen:
Resovled in #7973, replaced kzg verification with some simple len checks.

PR #7973, opened 2025-09-02
Skip column gossip verification logic during block production

eserilev:
## Issue Addressed

#7950 

## Proposed Changes

Skip column gossip verification logic during block production as its redundant and potentially computationally expensive.

### DIFF
--- a/beacon_node/beacon_chain/src/data_column_verification.rs
+++ b/beacon_node/beacon_chain/src/data_column_verification.rs
@@ -222,6 +222,8 @@ impl<T: BeaconChainTypes, O: ObservationStrategy> GossipVerifiedDataColumn<T, O>
         column_sidecar: Arc<DataColumnSidecar<T::EthSpec>>,
         chain: &BeaconChain<T>,
     ) -> Result<Self, GossipDataColumnError> {
+        verify_data_column_sidecar(&column_sidecar)?;
+
         // Check if the data column is already in the DA checker cache. This happens when data columns
         // are made available through the `engine_getBlobs` method.  If it exists in the cache, we know
         // it has already passed the gossip checks, even though this particular instance hasn't been
@@ -238,24 +240,6 @@ impl<T: BeaconChainTypes, O: ObservationStrategy> GossipVerifiedDataColumn<T, O>
                 observe_gossip_data_column(&column_sidecar, chain)?;
             }
             return Err(GossipDataColumnError::PriorKnownUnpublished);
-        }
-
-        let cells_len = column_sidecar.column.len();
-        let commitments_len = column_sidecar.kzg_commitments.len();
-        let proofs_len = column_sidecar.kzg_proofs.len();
-
-        if cells_len != commitments_len {
-            return Err(GossipDataColumnError::InconsistentCommitmentsLength {
-                cells_len,
-                commitments_len,
-            });
-        }
-
-        if cells_len != proofs_len {
-            return Err(GossipDataColumnError::InconsistentProofsLength {
-                cells_len,
-                proofs_len,
-            });
         }
 
         Ok(Self {
--- a/beacon_node/beacon_chain/src/data_column_verification.rs
+++ b/beacon_node/beacon_chain/src/data_column_verification.rs
@@ -215,6 +215,17 @@ impl<T: BeaconChainTypes, O: ObservationStrategy> GossipVerifiedDataColumn<T, O>
         )
     }
 
+    /// Create a `GossipVerifiedDataColumn` from `DataColumnSidecar` for block production ONLY.
+    /// When publishing a block constructed locally, the EL will have already verified the cell proofs.
+    /// When publishing a block constructed externally, the proposer will only see the columns over gossip.
+    pub fn new_for_block_publishing(column_sidecar: Arc<DataColumnSidecar<T::EthSpec>>) -> Self {
+        Self {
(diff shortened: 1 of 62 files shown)


beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -276,6 +287,13 @@ impl<E: EthSpec> KzgVerifiedDataColumn<E> {
         Self { data: data_column }
     }
 
+    /// Create a `KzgVerifiedDataColumn` from `DataColumnSidecar` for block publishing ONLY.
+    /// When publishing a block constructed locally, the EL will have already verified the KZG commitment.
+    /// When publishing a block constructed externally, there will be no columns here.
+    pub(crate) fn new_for_block_publishing(data_column: Arc<DataColumnSidecar<E>>) -> Self {
```
jimmygchen: I think we could use `from_execution_verified()` above?

eserilev: updated to use `from_execution_verified`

jimmygchen:
@eserilev the sync job is failing, doesn't feel like your change would trigger this, but interestingly both fullnode and supernode are failing to sync, and stuck at `SyncTransition`, would you mind checking the logs to see what happened?
https://github.com/sigp/lighthouse/actions/runs/17475882997/job/49636337462?pr=7973

beacon_node/http_api/src/publish_blocks.rs
```diff
@@ -414,49 +419,22 @@ fn build_gossip_verified_data_columns<T: BeaconChainTypes>(
...
-                    error!(
-                        column_index,
-                        %slot,
-                        error = ?e,
-                        "Data column for publication is gossip-invalid"
-                    );
-                    Err(warp_utils::reject::custom_bad_request(format!("{e:?}")))
-                }
-            }
+            GossipVerifiedDataColumn::new_for_block_publishing(data_column_sidecar)
```
jimmygchen: I think we still need to do the following to avoid trying to publish & process them again:
1. check the availability cache and observed cache to make sure we exclude columns that we have seen
2. mark the column as seen in the observe cache

https://github.com/sigp/lighthouse/blob/9d2f55a39984958e53ccd7110c2cb76eade6b442/beacon_node/beacon_chain/src/data_column_verification.rs#L450-L466


We've seen a race condition where the proposer node gets the columns from peers before the proposer publishes - a scenario where other nodes was able to fetch blobs from the EL and compute them faster than the proposer node.

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -215,6 +215,36 @@ impl<T: BeaconChainTypes, O: ObservationStrategy> GossipVerifiedDataColumn<T, O>
...
+    /// When publishing a block constructed externally, there will be no columns here.
+    pub fn new_for_block_publishing(
+        column_sidecar: Arc<DataColumnSidecar<T::EthSpec>>,
+        chain: &BeaconChain<T>,
+    ) -> Result<Self, GossipDataColumnError> {
+        // Check if the data column is already in the DA checker cache. This happens when data columns
+        // are made available through the `engine_getBlobs` method.  If it exists in the cache, we know
+        // it has already passed the gossip checks, even though this particular instance hasn't been
+        // seen / published on the gossip network yet (passed the `verify_is_first_sidecar` check above).
+        // In this case, we should accept it for gossip propagation.
```
jimmygchen, line 231: I think we should also check for `is_first_sidecar` here, and we can avoid acquiring the cache write lock below (in the `observe` fn) if not needed - the cache read & write happens in burst around this time and it's time sensitive so it'd be great to avoid this.

jimmygchen (changes requested):
Just added a minor comment but i think we're close to merging this.

beacon_node/beacon_chain/src/data_column_verification.rs
eserilev: I've added the check you mentioned and also renamed the function to `verify_is_unknown_sidecar`. The original naming was a bit confusing to me, at first glance I was assuming the function was checking if its the first column sidecar in the list of sidecars (i.e. index == 1).

eserilev:
~~For the 300s sync tests, I found it a bit weird that the non vc nodes are supposed to sleep for 5min but the test also times out in 5min. I've increased the timeout threshold to 6min. Not sure if thats the underlying issue here, but it feels like a step in the right direction either way~~

nevermind, I see that the timeout is only relevant once the nodes are restarted

beacon_node/beacon_chain/src/data_column_verification.rs
jimmygchen: yeah the rename sounds good, IIRC it was originally named to match the spec, but I agree it's confusing without the full context:
> [IGNORE] The sidecar is the first sidecar for the tuple (block_header.slot, block_header.proposer_index, sidecar.index) with valid header signature, sidecar inclusion proof, and kzg proof.

jimmygchen (approved):
Nice, the added check looks good to me! Thanks

jimmygchen:
@eserilev http api tests are failing, looks like the tests need an update - we no longer expect a gossip data column failure, so the assertion needs to be updated for Fulu
https://github.com/sigp/lighthouse/actions/runs/17607093762/job/50020181539

jimmygchen:
@mergify requeue

Issue #7704, opened 2025-07-05
Self rate limiting due to number of concurrent requests for `data_column_sidecars_by_range`


jimmygchen:
When trying to sync on a PeerDAS-enabled devnet, we see a large number of self rate limiting logs. Range sync only sends up to 5 batches concurrently, but this changes with PeerDAS, as we may send more than 1 data column requests per batch - so might be worth investigating and see if the current limit makes sense for data column requests, as this will slow down sync:

```
Jul 05 18:08:02.446 DEBUG Self rate limiting due to the number of concurrent requests  peer_id: 16Uiu2HAmACmiJYvqiMWDiVN8tbwTpHqFWxj3E9bPvnRq1jeokGhH, protocol: data_column_sidecars_by_range, service: "network"
Jul 05 18:08:02.549 DEBUG Self rate limiting due to the number of concurrent requests  peer_id: 16Uiu2HAmACmiJYvqiMWDiVN8tbwTpHqFWxj3E9bPvnRq1jeokGhH, protocol: data_column_sidecars_by_range, service: "network"
Jul 05 18:08:02.651 DEBUG Self rate limiting due to the number of concurrent requests  peer_id: 16Uiu2HAmACmiJYvqiMWDiVN8tbwTpHqFWxj3E9bPvnRq1jeokGhH, protocol: data_column_sidecars_by_range, service: "network"
Jul 05 18:08:02.753 DEBUG Self rate limiting due to the number of concurrent requests  peer_id: 16Uiu2HAmACmiJYvqiMWDiVN8tbwTpHqFWxj3E9bPvnRq1jeokGhH, protocol: data_column_sidecars_by_range, service: "network"
```

jimmygchen:
@pawanjay176 wonder if you are aware of this - given you were able to sync with up to 20 slots /s on your branch?

dapplion:
We could address this by prioritizing peers with no active requests on the same protocol. With a similar function to this one but only for that specific topic
https://github.com/sigp/lighthouse/blob/b9c1a2b0c0997fd0530a995f7ea72f89b35ba22c/beacon_node/network/src/sync/network_context.rs#L406

jimmygchen:
This isn't really a bug. The request is queued due to number of concurrent requests for the same protocol - this is part of the spec. The logs may be a bit noisy on a smaller network, but i might be fine after we tweak the peer maintainance logic in #7916 and less noisy on a real network with 100+ peers (200 from the next release)

[...]

PR #7975, opened 2025-09-02
Improve `DataColumnByRange` request peer selection

jimmygchen:
## Issue Addressed

Address this comment [here](https://github.com/sigp/lighthouse/issues/7704#issuecomment-3049407760) to improve column range peer selection.

## Proposed Changes

Use only the data column by range active request count instead of overall request count when selecting request peers, and exclude the current request in the prioritisation. 

This would also prevent splitting a range request to too many requests to different peers.


jimmygchen:
Closing this as we're reworking range sync and this benefit from this is likely v. marginal.

PR #7976, opened 2025-09-02
Add max delay to reconstruction

eserilev:
## Issue Addressed

#7697

## Proposed Changes

If we're three seconds into the current slot just trigger reconstruction. I don't know what the correct reconstruction deadline number is, but it should probably be at least half a second before the attestation deadline

### DIFF
--- a/beacon_node/beacon_processor/src/scheduler/work_reprocessing_queue.rs
+++ b/beacon_node/beacon_processor/src/scheduler/work_reprocessing_queue.rs
@@ -175,6 +175,7 @@ pub struct QueuedBackfillBatch(pub AsyncFn);
 
 pub struct QueuedColumnReconstruction {
     pub block_root: Hash256,
+    pub slot: Slot,
     pub process_fn: AsyncFn,
 }
 
@@ -755,10 +756,12 @@ impl<S: SlotClock> ReprocessQueue<S> {
                 let mut reconstruction_delay = QUEUED_RECONSTRUCTION_DELAY;
                 if let Some(seconds_from_current_slot) =
                     self.slot_clock.seconds_from_current_slot_start()
+                    && let Some(current_slot) = self.slot_clock.now()
                     && seconds_from_current_slot >= RECONSTRUCTION_DEADLINE
+                    && current_slot == request.slot
                 {
                     // If we are at least `RECONSTRUCTION_DEADLINE` seconds into the current slot,
-                    // process reconstruction immediatly.
+                    // and the reconstruction request is for the current slot, process reconstruction immediately.
                     reconstruction_delay = Duration::from_secs(0);
                 }
                 match self.queued_column_reconstructions.entry(request.block_root) {
--- a/beacon_node/beacon_processor/src/scheduler/work_reprocessing_queue.rs
+++ b/beacon_node/beacon_processor/src/scheduler/work_reprocessing_queue.rs
@@ -752,25 +752,24 @@ impl<S: SlotClock> ReprocessQueue<S> {
                 }
             }
             InboundEvent::Msg(DelayColumnReconstruction(request)) => {
+                let mut reconstruction_delay = QUEUED_RECONSTRUCTION_DELAY;
+                if let Some(seconds_from_current_slot) =
+                    self.slot_clock.seconds_from_current_slot_start()
+                    && seconds_from_current_slot >= RECONSTRUCTION_DEADLINE
+                {
+                    // If we are at least `RECONSTRUCTION_DEADLINE` seconds into the current slot,
+                    // process reconstruction immediatly.
+                    reconstruction_delay = Duration::from_secs(0);
+                }
                 match self.queued_column_reconstructions.entry(request.block_root) {
                     Entry::Occupied(key) => {
-                        let mut reconstruction_delay = QUEUED_RECONSTRUCTION_DELAY;
-                        if let Some(seconds_from_current_slot) =
-                            self.slot_clock.seconds_from_current_slot_start()
-                            && seconds_from_current_slot >= RECONSTRUCTION_DEADLINE
-                        {
-                            // If we are at least `RECONSTRUCTION_DEADLINE` seconds into the current slot,
-                            // process reconstruction immediatly.
-                            reconstruction_delay = Duration::from_secs(0);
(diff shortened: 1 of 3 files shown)


beacon_node/beacon_processor/src/scheduler/work_reprocessing_queue.rs
```diff
@@ -82,6 +82,9 @@ pub const BACKFILL_SCHEDULE_IN_SLOT: [(u32, u32); 3] = [
     (4, 5),
 ];
 
+/// Trigger reconstruction if we are this many seconds into the current slot
+pub const RECONSTRUCTION_DEADLINE: Duration = Duration::from_millis(300);
```
kevaundray: Should this be three seconds?

beacon_node/beacon_processor/src/scheduler/work_reprocessing_queue.rs
```diff
@@ -751,9 +754,18 @@ impl<S: SlotClock> ReprocessQueue<S> {
...
+                            // If we are at least `RECONSTRUCTION_DEADLINE` seconds into the current slot,
+                            // process reconstruction immediatly.
+                            reconstruction_delay = Duration::from_secs(0);
+                        }
+
                         self.column_reconstructions_delay_queue
-                            .reset(key.get(), QUEUED_RECONSTRUCTION_DELAY)
+                            .reset(key.get(), reconstruction_delay);
                     }
                     Entry::Vacant(vacant) => {
```
jimmygchen, line 772: I think this still queues the first message? i.e. `Entry::Vacant` is returned

eserilev: heres the fix 3ba3b85

eserilev:
sync test failure. dont think its related to this PR. This is ready for review

beacon_node/beacon_processor/src/scheduler/work_reprocessing_queue.rs
```diff
@@ -749,16 +752,24 @@ impl<S: SlotClock> ReprocessQueue<S> {
                 }
             }
             InboundEvent::Msg(DelayColumnReconstruction(request)) => {
+                let mut reconstruction_delay = QUEUED_RECONSTRUCTION_DELAY;
+                if let Some(seconds_from_current_slot) =
+                    self.slot_clock.seconds_from_current_slot_start()
+                    && seconds_from_current_slot >= RECONSTRUCTION_DEADLINE
```
jimmygchen, line 760: This logic may not work if we're triggering reconstruction on blocks other than the current slot. 
Could potentially be triggered here:
https://github.com/sigp/lighthouse/blob/f19d4f6af124994678a11a765e50dc4be7ddac34/beacon_node/network/src/network_beacon_processor/sync_methods.rs#L403-L417

eserilev:
I should have written some tests for this in the first place. I've added those now, sorry about the back and forth.

eserilev:
sync test failures are flaky im digging into it, but hopefully isnt a blocker for this PR

beacon_node/network/src/network_beacon_processor/tests.rs
```diff
@@ -743,6 +808,162 @@ fn junk_message_id() -> MessageId {
     MessageId::new(&[])
 }
 
+// Test that column reconstruction is delayed for columns that arrive
+// at the beginning of the slot.
+#[tokio::test]
+async fn data_column_reconstruction_at_slot_start() {
+    let mut rig = TestRig::new(SMALL_CHAIN).await;
```
jimmygchen: We can add a check here to only create the test rig (slow to create) if it's >= Fulu, similar to how we did it for deneb tests:
```rust
    if test_spec::<E>().fulu_fork_epoch.is_none() {
        return;
    };
```

jimmygchen: same for the other tests too

beacon_node/network/src/network_beacon_processor/tests.rs
```diff
@@ -743,6 +808,162 @@ fn junk_message_id() -> MessageId {
...
+
+    let slot_start = rig
+        .chain
+        .slot_clock
+        .start_of(rig.next_block.slot())
+        .unwrap();
+
+    rig.chain
```
jimmygchen, line 836: I think we can remove these lines as it's not relevant to this test - keeping the tests focused makes it a bit easier to understand and maintain

beacon_node/network/src/network_beacon_processor/tests.rs
```diff
@@ -743,6 +808,162 @@ fn junk_message_id() -> MessageId {
...
+        // reconstruction up to this point
+        rig.assert_event_journal_does_not_complete_with_timeout(
+            &[WorkType::ColumnReconstruction],
+            Duration::from_millis(100),
+        )
+        .await;
+
+        // We've waited at least 150ms, reconstruction can now be triggered
+        rig.assert_event_journal_completes_with_timeout(
+            &[WorkType::ColumnReconstruction],
```
jimmygchen, line 864: I don' really understand how this test is passing, it looks like `num_data_columns` is only 8 here (default non supernode), and this shouldn't trigger reconstruction.

To set up a supernode, we need to add `.import_all_data_columns` when building the harness

jimmygchen (changes requested):
The changes look good.
I'm not sure if the tests are behaving as expected though, as the chain isn't a supernode by default.

beacon_node/network/src/network_beacon_processor/tests.rs
eserilev: ive removed the block gossiping checks. i've kept the slot checks as they are still relevant for these tests

beacon_node/network/src/network_beacon_processor/tests.rs
eserilev: I dont think it really matters if the node is a supernode or not. the reconstruction task still gets scheduled its just that `attempt_data_column_reconstruction` ends up immediately returning `None` in the non-supernode case. I added `import_all_data_columns` anyways though

jimmygchen (approved):
Looks great, thanks @eserilev 🎉

PR #7977, opened 2025-09-02
Fix tracing subscriber rust audit issue

eserilev:
https://rustsec.org/advisories/RUSTSEC-2025-0055

Upgraded `tracing-subscriber` to 0.3.2


jimmygchen:
Oops sorry @eserilev, I didn't notice this PR and already sent #7981 into them merge queue. 

btw I think we probably don't need to pin the patch version in Cargo.toml?

jimmygchen:
Thanks for fixing this though 🙏

eserilev:
Ah yep no need to pin the version, thanks for taking care of it Jimmy!

PR #7979, opened 2025-09-02
chore: add comment to PendingComponents

kevaundray:
## Issue Addressed

Adds doc comment

### DIFF
--- a/beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
+++ b/beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
@@ -26,6 +26,15 @@ use types::{
 ///
 /// The blobs are all gossip and kzg verified.
 /// The block has completed all verifications except the availability check.
+///
+/// There are currently three distinct hardfork eras that one should take note of:
+///     - Pre-Deneb: No availability requirements (Block is immediately available)
+///     - Post-Deneb, Pre-PeerDAS: Blobs are needed, but columns are not for the availability check
+///     - Post-PeerDAS: Columns are needed, but blobs are not for the availability check
+///
+/// Note: from this, one can immediately see that `verified_blobs` and `verified_data_columns`
+/// are mutually exclusive. i.e. If we are verifying columns to determine a block's availability
+/// we are ignoring the `verified_blobs` field.
 pub struct PendingComponents<E: EthSpec> {
     pub block_root: Hash256,
     pub verified_blobs: RuntimeFixedVector<Option<KzgVerifiedBlob<E>>>,


beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -26,6 +26,15 @@ use types::{
...
+///
+/// There are currently three distinct hardfork eras that one should take note of:
+///     - Pre-Deneb: No availability requirements (Block is immediately available)
+///     - Post-Deneb, Pre-PeerDAS: Blobs are needed, but columns are not for the availability check
+///     - Post-PeerDAS: Columns are needed, but blobs are not for the availability check
+///
+/// Note: from this, one can immediately see that `verified_blobs` and `verified_data_columns`
+/// are mutually exclusive. i.e. If we are verifying columns to determine a block's availability
+/// we are ignoring the `verified_blobs` field.
 pub struct PendingComponents<E: EthSpec> {
```
kevaundray, line 38: IIUC This structure will work across fork boundaries because a lot of the logic will explicitly handle the different forks at runtime. So merge_blobs if called during peerDAS epoch, will do nothing for example.

If that's correct then I think something like this:

```rust
pub enum PendingComponents<E: EthSpec> {
    PostDeneb {
        block_root: Hash256,
        executed_block: Option<DietAvailabilityPendingExecutedBlock<E>>,
        verified_blobs: RuntimeFixedVector<Option<KzgVerifiedBlob<E>>>,
        span: Span,
    },
    
    PostPeerDAS {
        block_root: Hash256,
        executed_block: Option<DietAvailabilityPendingExecutedBlock<E>>,
        verified_data_columns: Vec<KzgVerifiedCustodyDataColumn<E>>,
        reconstruction_started: bool,
        span: Span,
    },
```

Might express the intentions clearer and then it would be possible to panic if PendingComponents is PostPeerDAS but `merge_blobs` has been called for example.

jimmygchen: Thanks @kevaundray  - Agree the above is cleaner, although we may not refactor this further at this stage as we're in the process of redesigning DA checker but we'll keep this in mind, thanks!

jimmygchen (approved):
Thanks, the added comments look good.

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
kevaundray: Ah this is great to know -- is there currently a draft branch or doc for this refactor?

jimmygchen:
@mergify requeue

PR #7946, opened 2025-08-27
Forward sync columns by root

pawanjay176:
## Issue Addressed

N/A

## The Problem

Our current strategy of syncing blocks + columns by range works roughly as follows for each batch:

1. Find a peer from the current `SyncingChain` to fetch blocks from and send a `BlocksByRange` request
2. Find peer(s) from the global peer pool that should have columns for the same batch based on their status message and the columns they are supposed to custody , and send them a by `DataColumnsByRange` request at the same time
3. Once we get responses for all blocks + columns components, try to couple them by checking that the `block_root` and the `kzg_commitment` matches. If the coupling failed, try to re-request the failed columns. 
4. Send them for processing and  try to progress the chain

This strategy works decently well when the chain is finalizing as most of our peers are on the same chain.  However, in times of non-finality we need to potentially sync multiple head chains. 
This leads to issues with our current approach because the block peer and the data column peers might have a different view of the canonical chain due to multiple heads. So when we use the above approach, it is possible that the block peer returns us a batch of blocks for chain A while some or all data column peers send us the batch of data columns for a different chain B. Different data column peers might also be following different chains.

We initially tried to get around this problem by selecting column peers only from within the current `SyncingChain`. Each `SyncingChain` represents a `head_root` that we are trying to sync to and we group peers based on same `head_root`. That way, we know for sure that the block and column peers are on the same chain. This works in theory, but in practice, during long periods of non-finality, we tend to create multiple head chains based on the `head_root` and split the global peerset. Pre-fulu, this isn't a big deal since all peers are supposed to have all the blob data. 
But splitting peers with peerdas is a big challenge due to not all peers having the full data available. There are supernodes, but during bad network conditions, supernodes would be getting way too many requests and not even have any incoming peer slots. As we saw on fusaka devnets, this strategy leads to sync getting stalled and not progressing.

### 1. Use `DataColumnsByRoot` instead of `DataColumnsByRange` to fetch columns for forward sync

This is the main change. The new strategy would go as follows: 

1. Find a peer from the current `SyncingChain` to fetch blocks from and send a `BlocksByRange` request
2. Hold off on requesting for columns until we receive the response for the above blocks request
3. Once we get the blocks response, extract all the `block_roots` and trigger a `DataColumnsByRoot` request for every block in the response that has any blobs based on the `expected_kzg_commitments` field. 
[...]

PR #7984, opened 2025-09-03
Allow AwaitingDownload to be a valid in-between state

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Extracts (3) from https://github.com/sigp/lighthouse/pull/7946. 

Prior to peerdas, a batch should never have been in `AwaitingDownload` state because we immediataly try to move from `AwaitingDownload` to `Downloading` state by sending batches. This was always possible as long as we had peers in the `SyncingChain` in the pre-peerdas world. 

However, this is no longer the case as a batch can be stuck waiting in `AwaitingDownload` state if we have no peers to request the columns from. This PR makes `AwaitingDownload` to be an allowable in between state. If a batch is found to be in this state, then we attempt to send the batch instead of erroring like before. 
Note to reviewer: We need to make sure that this doesn't lead to a bunch of batches stuck in `AwaitingDownload` state if the chain can be progressed.

Backfill already retries all batches in AwaitingDownload state so we just need to make `AwaitingDownload` a valid state during processing and validation.

This PR explicitly adds the same logic for forward sync to download batches stuck in `AwaitingDownload`. 
Apart from that, we also force download of the `processing_target` when sync stops progressing. This is required in cases where `self.batches` has > `BATCH_BUFFER_SIZE` batches that are waiting to get processed but the `processing_batch` has repeatedly failed at download/processing stage. This leads to sync getting stuck and never recovering.

### DIFF
--- a/beacon_node/network/src/sync/range_sync/chain.rs
+++ b/beacon_node/network/src/sync/range_sync/chain.rs
@@ -1198,20 +1198,15 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
     ) -> bool {
         if network.chain.spec.is_peer_das_enabled_for_epoch(epoch) {
             // Require peers on all sampling column subnets before sending batches
+            let peer_db = network.network_globals().peers.read();
             network
                 .network_globals()
                 .sampling_subnets()
                 .iter()
                 .all(|subnet_id| {
-                    let peer_db = network.network_globals().peers.read();
-                    let peer_count = self
-                        .peers
-                        .iter()
-                        .filter(|peer| {
-                            peer_db.is_good_range_sync_custody_subnet_peer(*subnet_id, peer)
-                        })
-                        .count();
-                    peer_count > 0
+                    self.peers.iter().any(|peer| {
+                        peer_db.is_good_range_sync_custody_subnet_peer(*subnet_id, peer)
+                    })
                 })
         } else {
             true
--- a/beacon_node/network/src/sync/range_sync/chain.rs
+++ b/beacon_node/network/src/sync/range_sync/chain.rs
@@ -1163,28 +1163,6 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
             self.send_batch(network, batch_id)?;
         }
 
-        // Force requesting the `processing_batch` to progress sync if required
-        if !self.batches.contains_key(&self.processing_target) {
-            debug!(?self.processing_target,"Forcing requesting processing_target to progress sync");
-            if !self.good_peers_on_sampling_subnets(self.processing_target, network) {
-                debug!(
-                    src = "request_batches_processing",
-                    "Waiting for peers to be available on sampling column subnets"
-                );
-                return Ok(KeepChain);
-            }
-
-            if let Entry::Vacant(entry) = self.batches.entry(self.processing_target) {
-                let batch_type = network.batch_type(self.processing_target);
-                let processing_batch =
-                    BatchInfo::new(&self.processing_target, EPOCHS_PER_BATCH, batch_type);
-                entry.insert(processing_batch);
-                self.send_batch(network, self.processing_target)?;
-            } else {
-                self.attempt_send_awaiting_download_batches(network, "request_batches_processing")?;
-            }
-        }
-
         // No more batches, simply stop
         Ok(KeepChain)
     }
--- a/beacon_node/network/src/sync/range_sync/chain.rs
+++ b/beacon_node/network/src/sync/range_sync/chain.rs
@@ -350,7 +350,10 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
                     return Ok(KeepChain);
                 }
                 BatchState::Poisoned => unreachable!("Poisoned batch"),
-                BatchState::Processing(_) | BatchState::AwaitingDownload | BatchState::Failed => {
+                // Batches can be in `AwaitingDownload` state if there weren't good data column subnet
+                // peers to send the request to.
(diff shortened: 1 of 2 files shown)


beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -1119,6 +1163,28 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
...
+            }
+
+            if let Entry::Vacant(entry) = self.batches.entry(self.processing_target) {
+                let batch_type = network.batch_type(self.processing_target);
+                let processing_batch =
+                    BatchInfo::new(&self.processing_target, EPOCHS_PER_BATCH, batch_type);
+                entry.insert(processing_batch);
+                self.send_batch(network, self.processing_target)?;
+            } else {
+                self.attempt_send_awaiting_download_batches(network, "request_batches_processing")?;
```
AgeManning: This is not possible right? Because it is contained within a 
```
if !self.batches.contains_key(&self.processing_target) {
```

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -1119,6 +1163,28 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
...
+            }
+        }
```
AgeManning: I don't understand the point of this. The processing target is the epoch in which we need to process next. Why do we need to download it again?

This seems to handle an edge case and makes this more complex. It seems like it's covering for an underlying issue somewhere else?

AgeManning (commented):
I've had a look over this. 

It's been a while since I've looked over the sync code and things have changed a bit. 

I noticed that you've only permitted the AwaitingDownload state in backfill and nothing else. I didn't see any logic of where we enter the awaiting download state, but as your PR comment suggests it must currently be happening. The trigger to restart sync should be when we add new peers, in which case we call `request_batches()`. This looks to me like it will re-download batches. 

Just thinking about this quickly, it seems the logic happens, that we request to download a batch, but find we don't have enough column subnet peers. So we mark the batch as `AwaitingDownload`. Then we only resume sync once we have enough peers to cover all the column subnets. When we do, then we download the batch again. 

So my main question is; Is this `AwaitingDownload` state functionally equivalent to just not having downloaded the batch at all? 

If this is true, wouldn't it be easier to not add a new state of batch that is equivalent to us not downloading a batch. From memory we just use two pointers, processing_target and download_target. If we get a batch that we need more peers for, can't we just set hte download_target back to the start epoch of that batch?

I guess the problem we've got is that if we download 5 batches and 3 of them work somehow, we don't want to re-download all of them, just the ones that didn't work?

Curious about the thought here. 

Otherwise, changes look alright to me, i left some comments. 

With this stuff, we need to test pretty rigorously to shake out potential bugs.

beacon_node/network/src/sync/range_sync/chain.rs
pawanjay176: The issue i saw on devnet was this:
Batch 1,2,3,4,5 are sent for download. 2,3,4,5 were downloaded successfully and are now AwaitingProcessing. 
Batch 1 fails downloading a couple of times and the state changes to `AwaitingDownload`. Now batch 6 sneaks in at some point when batch 1 is still in the `AwaitingDownloading` state. 
So now `self.batches` contains 2,3,4,5,6 => AwaitingProcessing state while batch 1 can no longer get in.

I thought the simplest way to handle this would be force send the processing target. 

But now I think of it, this could have happened before I added a call to `attempt_send_awaiting_download_batches` here https://github.com/pawanjay176/lighthouse/blob/ce616022457c68686ea5ee9735f4755a210843a5/beacon_node/network/src/sync/range_sync/chain.rs#L817

So maybe we don't need this?

pawanjay176: I have done some more testing and have concluded that we do not need this. In my testing, i probably added the call to `attempt_send_awaiting_download_batches` in `start_syncing` along with this code for sending the `processing_target`. The sync getting stuck issue was probably fixed by the former and not the latter.

pawanjay176: Removed in 2241d9a

pawanjay176:
> So my main question is; Is this AwaitingDownload state functionally equivalent to just not having downloaded the batch at all?

I think yes.
 
> If this is true, wouldn't it be easier to not add a new state of batch that is equivalent to us not downloading a batch

I initially had the same thought, but when trying to come up with a name for this new BatchState variant, I felt that `AwaitingDownload` actually fit best. We basically tried to download, but couldn't so we are waiting for peers to download when they are available. The new state (maybe `AwaitingPeersForDownload`) seemed logically very similar to `AwaitingDownload` so I felt its okay to overload it.

beacon_node/network/src/sync/range_sync/chain.rs
pawanjay176: Just removed the entire block

AgeManning (approved):
Nice. LGTM!

jimmygchen (commented):
Thanks @pawanjay176!

I've resolved the conflicts from `unstable`. The only conflict is on the [`has_good_custody_range_sync_peer`](https://github.com/sigp/lighthouse/blob/c2a92f1a8ced6fa0ec357e72875b52a96b4fcbb1/beacon_node/lighthouse_network/src/peer_manager/peerdb.rs#L332) function I added in #7915, both are correct but i think the version on `unstable` is slightly more efficient with higher peer count - happy to change if needed, but would like to merge this and start testing on devnet.

Issue #7966, opened 2025-09-01
Backfill sync error logs due to insufficient custody peers


eserilev:
During backfill sync if a node has no synced peers for a column they're required to custody, Lighthouse shows a bunch of `ERROR` messages. We should probably downgrade these log messages as the node will eventually find peers that can serve those columns and continue on with backfill sync.

<img width="1708" height="709" alt="Image" src="https://github.com/user-attachments/assets/6b459c57-7701-4e02-86ae-e65aaf5b2e9f" />

anshalshukla:
Hey @eserilev, looks like an easy first issue to take up. Please assign this to me.

michaelsproul:
This also happens when backfilling beyond the DA window (https://github.com/sigp/lighthouse/pull/7751).

eserilev:
RangeSync will immediately start logging warns and errors on startup when it doesn't have peers for certain columns.

<img width="3840" height="2160" alt="Image" src="https://github.com/user-attachments/assets/6db853c5-6b14-40f9-b420-a497e2413594" />

PR #7985, opened 2025-09-03
Downgrade error log to warn

anshalshukla:
## Issue Addressed

Issue #7966 

## Proposed Changes

Changes logging level to Warn from Error.


eserilev:
hm I dont think downgrading to warn is what I had in mind. I think in the case where backfill is paused due to insufficient peers we can just log a single message. This is a bit more involved that just changing the log level. We want to be able to catch the specific error variant `InsufficientPeers` and log it accordingly. We also at the moment log a message for each column we have insufficient peers for, it would be better to just log one message.

PR #7986, opened 2025-09-03
Add `getBlobsV1` and `getBlobsV2` support to mock el server

jimmygchen:
## Issue Addressed

I just mirrored what we did with `get_payload`,  i.e. attempt to load blob bundles from `ExecutionBlockGenerator`. However, I'm not sure if the logic to get blobs from memory is useful in any case - perhaps when running a single node network 🤔  ?

For my testing purpose, this just prevents the CL from screaming errors, as  the mock EL currently always returns "method not supported".


michaelsproul (changes requested):
I would be fine with just returning `vec![]` every time, less code to maintain 😅

michaelsproul:
I would like to overhaul how we do the mock EL/mock builder/payload builder, as the current ones have lots of sharp edges

jimmygchen:
> I would be fine with just returning `vec![]` every time, less code to maintain 😅

Yeah agree - I wanted to do this initially, but I worry having inconsistency with other EL methods (that returns payload / blobs in memory) may set up a trap for others trying to use this in tests, i.e. expecting a result if it's set up in memory, but i might be overthinking.

michaelsproul:
Yeah that's fair, may as well merge as-is then seeing as you've already gone to the trouble of implementing it.

michaelsproul (approved):

jimmygchen:
> Yeah that's fair, may as well merge as-is then seeing as you've already gone to the trouble of implementing it.

I'm all in for less unnecessary code, so also don't mind if we don't want to include it now, or completely delete it once we overhaul the mock server

jimmygchen:
@mergify requeue

jimmygchen:
CI failing 
https://github.com/sigp/lighthouse/actions/runs/17778501914/job/50532087891

jimmygchen:
I'm closing this one as I don't think it's worth the effort to merge this in. We can revive later if we think it's useful.

jimmygchen:
new PR #8870

PR #7988, opened 2025-09-03
Simplify some fork-variant boilerplate

michaelsproul:
## Proposed Changes

Simplify some fork boilerplate by using `superstruct` mapping macros some more.

Due to a limitation of the current `superstruct` macros, we can't map between superstructs with different variants. The `BaseAndElectra` trait solves this by shimming the missing variants to their respective variants on the RHS.

This is probably not the best way to solve this, extending `superstruct`'s macros would likely be cleaner long term:

- https://github.com/sigp/superstruct/issues/54

## Additional Info

I've opened another PR with _just_ the non-spicy changes from this PR:

- https://github.com/sigp/lighthouse/pull/7989

### DIFF
--- a/consensus/types/src/beacon_block_body.rs
+++ b/consensus/types/src/beacon_block_body.rs
@@ -1,4 +1,5 @@
 use crate::test_utils::TestRandom;
+use crate::variant_mappings::BaseAndElectra;
 use crate::*;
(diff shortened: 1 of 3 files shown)


michaelsproul:
Closing in favour of doing this in `superstruct`.

Issue #7963, opened 2025-08-29
Add `User-Agent` HTTP header for validator client requests


KaloyanTanev:
Currently the Lighthouse validator client does not add any identification on itself, when sending request to the upstream.

Other validator clients opted in for a header `User-Agent` that identified them as such: `<CLIENT>/<VERSION>`, i.e.: [Prysm](https://github.com/OffchainLabs/prysm/pull/15574).

This would help a lot in a distributed validator middleware cases, where operators in a DV can better distinguish what their peers are running.

nadtech-hub:
I may take on it after finishing current one

anshalshukla:
I can take this up, please assign this to me.

PR #7990, opened 2025-09-03
add user agent in http requests

anshalshukla:
## Issue Addressed

Issue #7963 

## Proposed Changes

- Add user agent in all the Http requests
- Creates a new user agent function in `lighthouse_version` crate


eserilev (changes requested):
I don't think we need to introduce new crate dependencies for this. We already have ways to mock the beacon api, validator api etc and make requests

anshalshukla:
Right, I don’t actually need those. They were left over from something else I was experimenting with and went unnoticed.

common/eth2/src/lighthouse_vc/http_client.rs
```diff
@@ -698,3 +709,87 @@ async fn ok_or_error(response: Response) -> Result<Response, Error> {
         Err(Error::StatusCode(status))
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use mockito::{Matcher, Server};
```
eserilev: This probably won't compile now. 

We still want to write a test to check that a user agent is returned, we have some examples of api tests in our repo that you can reference. 

You can run our test suite via make test

eserilev (changes requested):

common/eth2/src/lighthouse_vc/http_client.rs
anshalshukla: I'll update the PR in a couple of days.

KaloyanTanev:
Hey guys, is there an update on this one?

anshalshukla:
hey @eserilev, please review this again, I've added testcases as well. I couldn't verify the request built using the reqwest crate as the headers are populated only when the request is broadcasted, I'm intercepting this request on a warp server to validate the presence of headers.

There are limited functions that reqwest exposes but if you can think of a better testing strategy I am happy to update my PR.

anshalshukla:
@jimmygchen let me know if you want me to resolve the conflicts and we are moving ahead with this. Also, if I can get a review on this.

KaloyanTanev:
@anshalshukla @eserilev 
Can we reopen and merge this one for v8.1.0?

anshalshukla:
I will rebase the changes again, it got closed because it wasn't reviewed

PR #7993, opened 2025-09-04
Add co-author to mergify commits

jimmygchen:
## Issue Addressed

We lost this feature after we moved to mergify. It would be good to keep the pull request co-authors in the the merge commits too.

I've tested this on my fork and it works:
https://github.com/jimmygchen/lighthouse/commit/774c38de5ff32acfc9393fa429f0bafc987292f4

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -105,6 +105,10 @@ queue_rules:
 
 
         {{ body | get_section("## Proposed Changes", "") }}
+      
+      {% for commit in commits | unique(attribute='email_author') %}
+      Co-Authored-By: {{ commit.author }} <{{ commit.email_author }}>
+      {% endfor %}
     queue_conditions:
       - "#approved-reviews-by >= 1"
       - "check-success=license/cla"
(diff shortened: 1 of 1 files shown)


jimmygchen:
I've also removed some automation and label updates that I feel a bit unnecessary and cause more trouble than good.

.github/mergify.yml
```diff
@@ -32,32 +27,6 @@ pull_request_rules:
     actions:
       comment:
         message: Some required checks have failed. Could you please take a look @{{author}}? 🙏
-      label:
-        add:
-          - waiting-on-author
-        remove:
-          - ready-for-review
```
michaelsproul: without this can't this action just loop? CI will fail and then it will keep spamming this comment because the conditions to trigger will remain true

michaelsproul (commented):
I feel like we should just leave the automation, or maybe just delete the one that goes from waiting on author to ready for review

jimmygchen:
> I feel like we should just leave the automation, or maybe just delete the one that goes from waiting on author to ready for review

Yeah i'll revert the automation change then, was a very weak preference because i had a feeling that people don't like it, but seems like you don't mind it :D

.github/mergify.yml
jimmygchen: I don't think it would trigger infinite loop if you don't change the conditions. I think it only action on them when the condition changes. Anyways, I'm reverting this

michaelsproul (approved):

Issue #7994, opened 2025-09-04
"Light client invalid finality update" from Lighthouse peers on fusaka-devnet-3


jimmygchen:
I'm seeing a lot of these messages from lighthouse peers on devnet-3 and potentially causing peering issues with other lighthouse nodes.

```
Sep 04 05:11:20.050 DEBUG Light client invalid finality update          peer_id: 16Uiu2HAmHM8Xe5a7fZSYdR68igMb7yTz1BySTkhJkrg3L1QReEsm, error: MismatchedSignatureSlot { local: Slot(306804), observed: Slot(306806) }
```

It looks like we try to compare the gossip lc finality update with our local update, however our node is not synced yet.

I think it might be due to this condition here:
https://github.com/sigp/lighthouse/blob/eef02afc9326c5e7f1855a995debdbcc481f5c14/beacon_node/beacon_chain/src/beacon_chain.rs#L4412-L4415

~~We  should be checking the delay against the current slot, not the `block_delay_total` (which compares against the block slot)~~ See michael's comment below

This might explain why we're always sending requests to the same clients - seems like we might be banning peers that gossip light client messages during sync.

jimmygchen:
@eserilev would you mind taking a look at this please 🙏

michaelsproul:
> We should be checking the delay against the current slot, not the block_delay_total (which compares against the block slot)

I might be confused, but what we have looks OK to me... If the block is old (we are syncing), then `block_delay_total` should be large. If it's older than 32 slots, then `block_delay_total` will be `>= self.slot_clock.slot_duration() * 32` and we won't try to send a LC update. Seems fine, no?

jimmygchen:
Yeah you're right!

This seems to happen when there isn't enough peer for the node to keep up with every slot, but it's constantly in `Synced` state, and mostly a few slots behind, and in this case it makes things worse by disconnecting the good light client peers.

jimmygchen:
Some logs here, I've seen 2649 occurrences over 3 hours.

```
beacon.log.6:Sep 02 10:42:08.140 DEBUG Light client invalid finality update          peer_id: 16Uiu2HAkz9r249RxF7Bt8TNvn1SBgAQ5NimhQs9Hqjho4WAydFWW, error: MismatchedSignatureSlot { local: Slot(294059), observed: Slot(294060) }
[...]

PR #7996, opened 2025-09-04
Don't penalize peers if locally constructed light client data is stale

eserilev:
## Issue Addressed

#7994

## Proposed Changes

We seem to be penalizing peers in situations where locally constructed light client data is stale. This PR ignores incoming light client data if our locally constructed light client data isn't up to date.


jimmygchen (approved):
LGTM, thanks for the quick fix!

PR #7997, opened 2025-09-04
Remove unused logging metrics

michaelsproul:
## Issue Addressed

@chong-he noticed that the INFO/WARN/ERRO log counts on our dashboards had stopped working. Since switching to `tracing` we are now tracking total events _per crate_, and the global counters are unused.

Per-crate metrics are here:

https://github.com/sigp/lighthouse/blob/cfb1f7331064b758c6786e4e1dc15507af5ff5d1/common/logging/src/tracing_metrics_layer.rs#L61-L63

## Proposed Changes

Delete the unused global counters from the source.

We can sum across the per-crate metric in our dashboards to restore the previous functionality.

### DIFF
--- a/common/logging/src/lib.rs
+++ b/common/logging/src/lib.rs
@@ -1,5 +1,3 @@
-use metrics::{IntCounter, Result as MetricsResult, try_create_int_counter};
-use std::sync::LazyLock;
 use std::time::{Duration, Instant};
 use tracing_subscriber::EnvFilter;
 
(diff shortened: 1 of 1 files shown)


chong-he (commented):
Looks good, just need to delete the dependencies input here:
https://github.com/sigp/lighthouse/blob/fd10b632740e441354c4547a12727c29dd288239/common/logging/src/lib.rs#L1-L2

jimmygchen (approved):

PR #8004, opened 2025-09-04
Grab newer version of leveldb-sys

SamWilsn:
## Proposed Changes

Patch leveldb-sys with the one from [this pull request](https://github.com/skade/leveldb-sys/pull/31).

## Additional Info

This commit grabs a version of leveldb-sys that:

 - Replaces snappy and leveldb with submodules (and updates to the latest releases).
 - Rewrites the CI workflow to be more modern.
 - Allows linking against the system version of leveldb and snappy by setting `LEVELDB_NO_VENDOR=1`.
 - Hopefully resolves a linking issue when using the vendored snappy.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5096,8 +5096,7 @@ dependencies = [
 [[package]]
 name = "leveldb-sys"
 version = "2.0.9"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9dd94a4d0242a437e5e41a27c782b69a624469ca1c4d1e5cb3c337f74a8031d4"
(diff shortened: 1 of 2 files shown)


michaelsproul:
I think the leveldb crates are basically unmaintained, so we might need our own fork if we want this

SamWilsn:
Yeah, I saw that your PR fixing paths with spaces has been sitting around for a while. What's your vibe on this? leveldb (minus snappy) isn't too broken as far as you're concerned, so I'd understand your reluctance to add this just for my packaging purposes.

michaelsproul:
Yeah I think my preference is to continue building LH against stock LevelDB, just with snappy disabled. Do you want snappy working just for the sake of completeness in packaging? Would it be much of a loss if the package just didn't offer snappy support?

SamWilsn:
Fixing snappy was just a bonus. I'm only interested in `LEVELDB_NO_VENDOR`, but I can work around not having it.

dapplion:
Closing for inactivity

Issue #7980, opened 2025-09-03
Lighthouse repeatedly sending DataColumnsByRoot requests for the same block on `fusaka-devnet-3`


jimmygchen:
## Issue 

Lighthouse repeatedly sending DataColumnsByRoot requests for the same block on `fusaka-devnet-3`.

This issue was reported by Prysm and Teku on `fusaka-devnet-3`
https://discord.com/channels/595666850260713488/892088344438255616/1412503910106796103

Pawan has checked the peer id (`16Uiu2HAmLyapMNMX9gRSZX3cMm58snMiyYfKSUu7ZzQDHkJ1cxah`) and it's `lighthouse-reth-2`, would be good to check the logs on the host.

jimmygchen:
@ethDreamer would you mind taking a look at this issue please? 🙏

ethDreamer:
So it looks like [this block was actually orphaned](http://web.archive.org/web/20250903211322/https://dora.fusaka-devnet-3.ethpandaops.io/slot/0xa7e61546b6ccea3b8ac4904200c44c74b1b3dbfd198ba5cf08b942f0efda6554), likely because it simply isn't available judging by all the responses from peers that they don't have the data.

Here's the order of events:
1. The block was [imported to the data availability checker](https://github.com/sigp/lighthouse/blob/10e72df3318c7140d7cb0605e2af1f19119ceab9/beacon_node/beacon_chain/src/beacon_chain.rs#L3482):
```
Sep 02 16:14:52.990 DEBUG Component added to data availability checker  component: "block", status: "block 1 data_columns 0/9", block_source: gossip, block_root: 0xa7e61546b6ccea3b8ac4904200c44c74b1b3dbfd198ba5cf08b942f0efda6554
Sep 02 16:14:52.990 DEBUG Beacon block awaiting blobs                   block_root: 0xa7e61546b6ccea3b8ac4904200c44c74b1b3dbfd198ba5cf08b942f0efda6554, slot: 295724, block_source: gossip, block_root: 0xa7e61546b6ccea3b8ac4904200c44c74b1b3dbfd198ba5cf08b942f0efda6554
```
2. Presumably, an attestation with this head block root was received.
[...]

PR #8005, opened 2025-09-04
Fix stuck data column lookups by improving peer selection and retry logic

jimmygchen:
## Issue Addressed

Fixes the issue described in #7980 where Lighthouse repeatedly sends `DataColumnsByRoot` requests to the same peers that return empty responses, causing sync to get stuck.

The root cause was we don't count empty responses as failures, leading to excessive retries to unresponsive peers.

## Proposed Changes

- Track per peer attempts to limit retry attempts per peer (`MAX_CUSTODY_PEER_ATTEMPTS = 3`)
- Replaced random peer selection with hashing within each lookup to prevent splitting lookup into too many small requests and improve request batching efficiency.
- Added `single_block_lookup` root span to track all lookups created and added more debug logs:

image

### DIFF
--- a/beacon_node/network/src/sync/network_context/custody.rs
+++ b/beacon_node/network/src/sync/network_context/custody.rs
@@ -7,19 +7,16 @@ use fnv::FnvHashMap;
 use lighthouse_network::PeerId;
 use lighthouse_network::service::api_types::{CustodyId, DataColumnsByRootRequester};
 use lighthouse_tracing::SPAN_OUTGOING_CUSTODY_REQUEST;
-use lru_cache::LRUTimeCache;
 use parking_lot::RwLock;
-use rand::Rng;
 use std::collections::HashSet;
 use std::time::{Duration, Instant};
 use std::{collections::HashMap, marker::PhantomData, sync::Arc};
-use tracing::{Span, debug, debug_span, field, warn};
+use tracing::{Span, debug, debug_span, warn};
 use types::{DataColumnSidecar, Hash256, data_column_sidecar::ColumnIndex};
 use types::{DataColumnSidecarList, EthSpec};
 
 use super::{LookupRequestResult, PeerGroup, RpcResponseResult, SyncNetworkContext};
 
-const FAILED_PEERS_CACHE_EXPIRY_SECONDS: u64 = 5;
 const MAX_STALE_NO_PEERS_DURATION: Duration = Duration::from_secs(30);
 
 pub struct ActiveCustodyRequest<T: BeaconChainTypes> {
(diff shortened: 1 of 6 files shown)


jimmygchen:
It seems like we're spliting into way too many requests than needed, and doesn't really help with sync speed

image

I'm going to try experiment replacing the random component when selecting peer and use something deterministic at each block, so we batch as much as possible, but without ending up selecting the same peer for every block lookup.

jimmygchen:
Interestingly removing the randomness seems to have caused more issues keeping up with the chain, it seems like the spreading requests out to different peers may be a better approach when peers aren't all serving by root requests well.

jimmygchen:
^ This is not true, it happened because devnet-3 went into non finality testing mode.

So I think the changes are good, I've compare it with some healthy nodes on devnet-3, and see a similar pattern

image

and some other unhealthy nodes, that likely ran into the "stuck" scenario and had to fallback to range sync:

image

beacon_node/network/src/sync/network_context/custody.rs
```diff
@@ -238,52 +224,29 @@ impl<T: BeaconChainTypes> ActiveCustodyRequest<T> {
...
-                            // Prioritize peers that claim to know have imported this block
-                            if lookup_peers.contains(peer) { 0 } else { 1 },
-                            // De-prioritize peers that have failed to successfully respond to
-                            // requests recently
-                            self.failed_peers.contains(peer),
-                            // Prefer peers with fewer requests to load balance across peers.
-                            // We batch requests to the same peer, so count existence in the
-                            // `columns_to_request_by_peer` as a single 1 request.
-                            active_request_count_by_peer.get(peer).copied().unwrap_or(0)
-                                + columns_to_request_by_peer.get(peer).map(|_| 1).unwrap_or(0),
```
eserilev, line 275: `columns_to_request_by_peer.get(peer).map(|_| 1).unwrap_or(0)` is removed in the new `select_column_peer` implementation. This does change things slightly. `active_request_count_by_peer` only gets updated after we loop through `column_requests` while `columns_to_request_by_peer` is updated at each iteration of the loop. 

I think this makes sense though because we haven't actually made a request yet while looping through `colum_requests`. If anything the old logic seems wrong

beacon_node/network/src/sync/network_context/custody.rs
```diff
@@ -345,11 +331,54 @@ impl<T: BeaconChainTypes> ActiveCustodyRequest<T> {
...
+        random_state: &RandomState,
+    ) -> Option<PeerId> {
+        // We draw from the total set of peers, but prioritize those peers who we have
+        // received an attestation or a block from (`lookup_peers`), as the `lookup_peers` may take
+        // time to build up and we are likely to not find any column peers initially.
+        let custodial_peers = cx.get_custodial_peers(column_index);
+        let mut prioritized_peers = custodial_peers
+            .iter()
+            .filter(|peer| {
+                // Exclude peers that we have already made too many attempts to without success.
```
eserilev: nit: If we're calling an empty response successful maybe the comment should just be `Exclude peers that we have already made too many attempts to.`

eserilev (approved):
This looks great! I just had one thing I wanted to mention, mostly to make sure I wasn't misunderstanding things, plus a tiny nit that doesn't really matter.

beacon_node/network/src/sync/network_context/custody.rs
jimmygchen: Yep, this was intentional and is removed to reduce request fragmentation
The old logic counts the pending requests so every request that we're "planning" to send to would be de-prioritized, so we end up having to split into many requests.

See comment above https://github.com/sigp/lighthouse/pull/8005#issuecomment-3265854738

beacon_node/network/src/sync/network_context/custody.rs
jimmygchen: yeah you're right, it's a bit confusing - it's considered "success" by specification, but not in practice if we can't complete the lookup, that's what I was trying to say. I can update the comment.

jimmygchen: Updated in e0ca662b

PR #8007, opened 2025-09-05
Reduce noise in `Debug` impl of `RuntimeVariableList`

jimmygchen:
## Issue Addressed

The default debug output of these types contains a lot of unnecessary noise making it hard to read. 

This PR removes the type and extra fields from debug output to make logs easier to read. 

`len` could be potentially useful in some cases, but this gives us flexibility to only log it separately if we need it.

Related PR in `ssz_types`:
- https://github.com/sigp/ssz_types/pull/57

### Before

Before:
```rust
RuntimeVariableList { vec: [42, 42, 42, 42], max_len: 4 }
```

After:

```rust
[42, 42, 42, 42]
```

### Before 

```rust
RuntimeFixedVector { vec: [42, 42, 42, 42], len: 4 }
```

### After

```rust
[42, 42, 42, 42]
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8905,14 +8905,14 @@ dependencies = [
 
 [[package]]
 name = "ssz_types"
-version = "0.12.0"
-source = "git+https://github.com/sigp/ssz_types?rev=c557336ae9b806ab00baa7e5c4ce446f3e205f2d#c557336ae9b806ab00baa7e5c4ce446f3e205f2d"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "75b55bedc9a18ed2860a46d6beb4f4082416ee1d60be0cc364cebdcdddc7afd4"
 dependencies = [
  "arbitrary",
- "derivative",
  "ethereum_serde_utils",
  "ethereum_ssz",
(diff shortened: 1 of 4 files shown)


michaelsproul (approved):
The `max_len` doesn't seem as nasty as the `typenum` we were logging in `ssz_types`, but happy to remove it anyway.

I guess having it could sometimes be useful for debugging config issues? Similar to what Pawan suggested on the ssz_types PR:

- https://github.com/sigp/ssz_types/pull/57#issuecomment-3257128030

jimmygchen:
> The `max_len` doesn't seem as nasty as the `typenum` we were logging in `ssz_types`, but happy to remove it anyway.
> 
> I guess having it could sometimes be useful for debugging config issues? Similar to what Pawan suggested on the ssz_types PR:
> 
> * [sigp/ssz_types#57 (comment)](https://github.com/sigp/ssz_types/pull/57#issuecomment-3257128030)

Yep I can implement something like 

```
[42, 42, 42, 42] (max_len=4)
[42, 42, 42, 42] (len=4)
```

michaelsproul (approved):
Looks nice

michaelsproul:
@mergify requeue

PR #8009, opened 2025-09-06
fix: extra fields in logs

### LINKED ISSUE #7995: Instrumentation leads to extra fields in logs
The instrumentation we've added has lead to the addition of the instrumented fields to logs. E.g. this state cache missed triggered by (presumably) blob verification, results in 3x block_root, 2x state_root, slot and index fields being logged:

> Aug 27 03:12:14.471 WARN  State cache missed                            state_root: 0xabbf84a7d9e8305916ddd9558f27241aa7a4bc22d5f4455d5ce4b6aea42a50b7, block_root: 0xc6dfece48906334a091f99cd5f07826ff00816b5765fd0caa34771efcafadafe, state_root: 0xabbf84a7d9e8305916ddd9558f27241aa7a4bc22d5f4455d5ce4b6aea42a50b7, block_root: 0xc6dfece48906334a091f99cd5f07826ff00816b5765fd0caa34771efcafadafe, index: 1, block_root: 0x6ddc3ab728146617dc9626938c53c285d8703b0aa234c0e32ec5f9d59fa967d3, slot: Slot(12453293)

This is despite the event in the code only logging the `state_root` and `block_root`:

https://github.com/sigp/lighthouse/blob/cfb1f7331064b758c6786e4e1dc15507af5ff5d1/beacon_node/store/src/hot_cold_store.rs#L1029

I think to fix this we probably want to remove redundant fields from each log event? Or maybe we should somehow exclude the fields from logs and just keep them in traces.

## Additional Info

In our example the duplication is especially confusing because the `block_root`s are different. One is the [block_root of the blob itself](https://beaconcha.in/slot/0x6ddc3ab728146617dc9626938c53c285d8703b0aa234c0e32ec5f9d59fa967d3), and the other two (lower-level) fields are the [parent block](https://beaconcha.in/slot/0xc6dfece48906334a091f99cd5f07826ff00816b5765fd0caa34771efcafadafe) being loaded from the DB.

Instrumentation added in:

- https://github.com/sigp/lighthouse/pull/7816

PoulavBhowmick03:
## Issue Addressed

Potentially fixes #7995

## Proposed Changes

changed `span_data` to a `HashMap` and added a new check to remove span fields whose base names are already present on the event.

### DIFF
--- a/common/logging/src/tracing_logging_layer.rs
+++ b/common/logging/src/tracing_logging_layer.rs
@@ -2,6 +2,7 @@ use crate::utils::is_ascii_control;
 
 use chrono::prelude::*;
 use serde_json::{Map, Value};
+use std::collections::{HashMap, HashSet};
 use std::io::Write;
 use tracing::Subscriber;
 use tracing::field::Field;
@@ -79,13 +80,26 @@ where
 
         event.record(&mut visitor);
 
-        let mut span_data = Vec::new();
+        let mut span_data = HashMap::new();
         if let Some(scope) = ctx.event_scope(event) {
             for span in scope.from_root() {
                 if let Some(data) = span.extensions().get::<SpanData>() {
-                    span_data.extend(data.fields.clone());
+                    for (k, v) in data.fields.clone() {
+                        span_data.insert(k, v);
+                    }
                 }
             }
+        }
+
+        // Remove span fields whose base names are already present on the event.
+        if !visitor.fields.is_empty() && !span_data.is_empty() {
+            let event_field_names: HashSet<&str> =
+                visitor.fields.iter().map(|(k, _)| k.as_str()).collect();
+
+            span_data.retain(|key, _| {
+                let base = key.rsplit('.').next().unwrap_or(key.as_str());
+                !event_field_names.contains(base)
+            });
         }
 
         // Remove ascii control codes from message.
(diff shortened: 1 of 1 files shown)


common/logging/src/tracing_logging_layer.rs
```diff
@@ -79,15 +80,28 @@ where
 
         event.record(&mut visitor);
 
-        let mut span_data = Vec::new();
+        let mut span_data = HashMap::new();
         if let Some(scope) = ctx.event_scope(event) {
             for span in scope.from_root() {
```
michaelsproul: Rather than looping over spans and then removing stuff, could we just take the _last_ (current) span and use its data?

michaelsproul: I feel like the problem is not only the duplication, but the fact that _all_ instrumented fields are showing up in log events.

michaelsproul: Chatted to @jimmygchen and we agreed this is the best approach.

We can also get rid of `from_root`, and just take the first span from `scope` (just `scope.next()` should work).

PoulavBhowmick03: one test is failing after my last commit. will fix it in soon

jimmygchen (approved):
LGTM

PR #8011, opened 2025-09-08
Migrating eip-3076

### LINKED ISSUE #7894: Extract EIP-3076 into its own crate
The `slashing_protection` crate contains our implementation of [EIP-3076](https://eips.ethereum.org/EIPS/eip-3076). This implementation is used by the `eth2` crate, which means we pull in all of `slashing_protection`, including its SQL-related stuff, as dependencies of `eth2`.

We should pull out the EIP-3076 related code into its own crate separate from `slashing_protection`, something like `common/eip-3076`.

Even when we split the `eth2` crate into two separate crates (one generic in Bosun and one Lighthouse-specific in Lighthouse), we may want the EIP-3076 stuff in the generic crate, as it is a cross-client format. In this case we'd pull EIP-3076 into Bosun as well.

nadtech-hub:
This should close [#7894](https://github.com/sigp/lighthouse/issues/7894)


common/eip_3076/interchange-tests/README.md
michaelsproul, line 1: The whole `interchange-tests` directory should not be committed. It was part of the `.gitignore` in `slashing_protection`, and should continue to be ignored.

The source lives here:

- https://github.com/eth-clients/slashing-protection-interchange-tests

michaelsproul (changes requested):
This PR doesn't accomplish the desired goal. You've moved all of the `slashing_protection` crate into `eip_3076`, but what we would ideally do, is just extract the "pure" EIP-3076 code (without database dependencies) into `eip_3076`, while leaving all of the database code and Lighthouse-specific logic in the `slashing_protection` crate.

This would allow us to reduce the dependency surface of crates like `eth2` which only need the "pure" EIP-3076 code, and _should not_ depend on Lighthouse's implementation details (our choice of sqlite for the slashing protection database).

nadtech-hub:
> This PR doesn't accomplish the desired goal. You've moved all of the `slashing_protection` crate into `eip_3076`, but what we would ideally do, is just extract the "pure" EIP-3076 code (without database dependencies) into `eip_3076`, while leaving all of the database code and Lighthouse-specific logic in the `slashing_protection` crate.
> 
> 
> 
> This would allow us to reduce the dependency surface of crates like `eth2` which only need the "pure" EIP-3076 code, and _should not_ depend on Lighthouse's implementation details (our choice of sqlite for the slashing protection database).

Makes sense, I'll do changes accordingly

michaelsproul:
Thanks!

Let us know if you have any questions, some of the stuff might be tricky to extract

nadtech-hub:
> Thanks!
> 
> 
> 
> Let us know if you have any questions, some of the stuff might be tricky to extract

It was matter of migrating interchange, signed_block and signed_attestation and corresponding tests, wasn't it? For some reason, I moved the whole thing)

michaelsproul:
I think just moving the interchange and its immediate dependents is the way to go, yeah.

common/eip_3076/src/signed_attestation.rs
michaelsproul, line 1: This is the wrong `SignedAttestation`. You should move this one: https://github.com/sigp/lighthouse/blob/fd10b632740e441354c4547a12727c29dd288239/validator_client/slashing_protection/src/interchange.rs#L36-L46.

common/eip_3076/src/signed_block.rs
michaelsproul, line 1: Also the wrong `SignedBlock`.

common/eip_3076/src/lib.rs
michaelsproul, line 1: I don't think `Safe`, `NotSafe` or `SigningRoot` need to be here either. Just move the `Interchange` type. Start by moving `interchange.rs`

nadtech-hub:
> I think just moving the interchange and its immediate dependents is the way to go, yeah.

I also thought that interchange file format needs be moved out along with inner data structures. Separating interchange from database is cumbersome

michaelsproul:
The interchange format is the main thing we need to pull out, even if it's just the type definition. Can you elaborate on the interaction with the database that makes it hard to extract?

nadtech-hub:
> The interchange format is the main thing we need to pull out, even if it's just the type definition. Can you elaborate on the interaction with the database that makes it hard to extract?

Interchange test modules are harder to migrate because of cohesion with db.
Interchange structure is extractable as a whole but tests will remain in slashing_protection.
Some data structures are replicated in interchange.

michaelsproul:
Yeah that sounds acceptable. The main thing we're trying to accomplish is pulling out the type definitions so we can remove the dependency of the `eth2` crate on `slashing_protection`. Instead `eth2` should depend on `eip_3076`.

common/eth2/Cargo.toml
```diff
@@ -29,8 +30,8 @@ reqwest-eventsource = "0.5.0"
 sensitive_url = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
-slashing_protection = { workspace = true }
 ssz_types = { workspace = true }
+slashing_protection = { workspace = true }
```
michaelsproul: This needs to be removed

common/eip_3076/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+[features]
+arbitrary-fuzz = ["types/arbitrary-fuzz"]
+portable = ["types/portable"]
+
+[dependencies]
+arbitrary = { workspace = true, features = ["derive"] }
+ethereum_serde_utils = { workspace = true }
+filesystem = { workspace = true }
+r2d2 = { workspace = true }
+r2d2_sqlite = "0.21.0"
```
michaelsproul: We need to remove the unnecessary dependencies here. You can run `cargo udeps` locally to check this (see: https://crates.io/crates/cargo-udeps).

common/eip_3076/src/lib.rs
michaelsproul: ^ @nadtech-hub this comment has not been addressed

common/eip_3076/interchange-tests/README.md
nadtech-hub: It's still ignored

common/eip_3076/Cargo.toml
nadtech-hub: useful thing

common/eip_3076/src/lib.rs
nadtech-hub: latest changes should be sufficient, shouldn't they?

PR #8013, opened 2025-09-09
bls: fix is_infinity when aggregating onto empty AggregateSignature

Galoretka:
- Ensure is_infinity reflects the actual aggregate when aggregating onto an empty AggregateSignature.
- Previously, starting from empty() and aggregating an infinity signature produced an aggregate whose underlying point was infinity but is_infinity stayed false. This broke eth_fast_aggregate_verify for empty pubkeys and desynced the flag from the serialized value.
- Update add_assign and add_assign_aggregate: when self was empty, set is_infinity = other.is_infinity; otherwise keep the existing conjunction behavior. This aligns with the doc that empty acts as “set to infinity before aggregation” and keeps is_infinity consistent with the bytes.


michaelsproul:
Can you please rebase on unstable and sign the CLA?

Your PR looks a bit like it might be AI-generated for farming contributions, but we will dig in properly to check.

PR #8016, opened 2025-09-09
Fix TOCTOU vulnerability in unused_port module

sashaodessa:
Replaces unsafe port-finding functions with secure bound socket APIs to eliminate race conditions.

### Problem
The existing `unused_tcp*_port()` and `unused_udp*_port()` functions had a classic **Time-of-Check-Time-of-Use (TOCTOU)** vulnerability:
- Functions would bind to port 0, get the assigned port number, then immediately close the socket
- Between getting the port number and later binding to it, another process could claim the same port
- The LRU cache provided false security and didn't prevent the race condition

### Solution
Added new safe APIs that return already-bound sockets:
- `bind_tcp4_any()` / `bind_tcp6_any()` → returns `TcpListener`
- `bind_udp4_any()` / `bind_udp6_any()` → returns `UdpSocket`


michaelsproul:
Thanks!

We are aware of this issue and had been phasing these functions out in favour of using 0 ports explicitly. The hard part is replacing the uses of the old functions in code that assumes it can determine an unused port without binding a socket, e.g. here:

https://github.com/sigp/lighthouse/blob/fd10b632740e441354c4547a12727c29dd288239/lighthouse/tests/beacon_node.rs#L955-L970

If you would be interested in coming up with clever fixes for these cases, that would be extremely helpful. That would then allow us to delete rather than deprecate the unsafe functions.

michaelsproul:
Most of the uses are in tests, but it's still important for us to get rid of them (if possible), as they sometimes contribute to CI flakiness.

beacon_node/src/config.rs
```diff
@@ -1058,24 +1055,21 @@ pub fn parse_listening_addresses(cli_args: &ArgMatches) -> Result<ListenAddress,
...
+            };
             // use zero ports if required. If not, use the specific quic port. If none given, use
             // the tcp port + 1.
-            let quic_port = use_zero_ports
-                .then(unused_port::unused_udp4_port)
-                .transpose()?
-                .or(maybe_quic_port)
-                .unwrap_or(if tcp_port == 0 { 0 } else { tcp_port + 1 });
+            let quic_port = if use_zero_ports {
+                0
```
daveth3t3chg33k: Extrapolate the function to classes

daveth3t3chg33k (commented):
Extrapolate the tcp function into the udp main class

sashaodessa:
@daveth3t3chg33k

michaelsproul:
@sashaodessa Can you also please rebase on `unstable`? All PRs need to be merged to `unstable`

sashaodessa:
@michaelsproul

michaelsproul:
Looks good at a glance, I've requested a review from some our networking experts on the changes that impact prod

.github/mergify.yml
```diff
@@ -105,6 +105,10 @@ queue_rules:
 
 
         {{ body | get_section("## Proposed Changes", "") }}
+      
+      {% for commit in commits | unique(attribute='email_author') %}
+      Co-Authored-By: {{ commit.author }} <{{ commit.email_author }}>
+      {% endfor %}
```
AgeManning, line 111: Are these changes relevant?

AgeManning (commented):
These changes look good to me. 

Agree a future PR to remove deprecation would be nice.

.github/mergify.yml
michaelsproul: No this is some leftover stuff from `stable` that hasn't been backmerged to `unstable` yet (see https://github.com/sigp/lighthouse/commits/stable/), it should be deleted from this PR.

michaelsproul:
I think we can probably just delete the deprecated functions in this PR, AFAICT all (?) usages have been removed

michaelsproul:
@sashaodessa Can you please update this PR with the latest changes from `unstable`?

sashaodessa:
> @sashaodessa Can you please update this PR with the latest changes from `unstable`?

@michaelsproul Hi, I flooded my MacBook and took it to a repair shop. Unfortunately, I only have my phone now. I`ll try from another device

michaelsproul:
This diff is way too large now

Issue #8012, opened 2025-09-09
Bump `ssz_types` version once v0.12.1 is released


jimmygchen:
In #8007 I've updated `ssz_types` to the current latest commit to get this change:
* https://github.com/sigp/ssz_types/pull/57

We should update the version to a proper released version once it's readybefore the v8.0.0 release.

jimmygchen:
I've reverted the change in #8007 as there are breaking changes in the upgrade that requires too many fixes. Will leave the upgrade to a separate PR.

michaelsproul:
I think we can do a quick sneaky v0.12.1 release (now) and do v0.13.0 later with the unsafe stuff.

- https://github.com/sigp/ssz_types/issues/59

PR #8032, opened 2025-09-11
Bump `ssz_types` to `v0.12.2`

macladson:
## Issue Addressed

https://github.com/sigp/lighthouse/issues/8012

## Proposed Changes

Replace all instances of `VariableList::from` and `FixedVector::from` to their `try_from` variants.

While I tried to use proper error handling in most cases, there were certain situations where adding an `expect` for situations where `try_from` can trivially never fail avoided adding a lot of extra complexity.


beacon_node/lighthouse_network/tests/rpc_tests.rs
```diff
@@ -1018,7 +1018,7 @@ fn test_tcp_columns_by_root_chunked_rpc() {
                 },
                 signature: Signature::empty(),
             },
-            column: vec![vec![0; E::bytes_per_blob()].try_into().unwrap()]
+            column: vec![vec![0; E::bytes_per_cell()].try_into().unwrap()]
```
michaelsproul, line 1021: Nice find!!

beacon_node/execution_layer/src/engine_api/json_structures.rs
```diff
@@ -673,6 +624,25 @@ impl From<JsonWithdrawal> for Withdrawal {
...
 }
+
+// Helper functions to convert between `VariableList<Withdrawal>` and `VariableList<JsonWithdrawal>`.
+// The `expect` is required to enable `from` impls between `ExecutionPayload` and `JsonExecutionPayload`.
+fn withdrawals_to_json<N>(list: VariableList<Withdrawal, N>) -> VariableList<JsonWithdrawal, N>
+where
+    N: Unsigned,
+{
+    VariableList::try_from_iter(list.into_iter().map(Into::into))
+        .expect("conversion to JSON should preserve withdrawals length")
```
michaelsproul: This is a little spicy, but I think it's OK for now.

beacon_node/lighthouse_network/src/rpc/methods.rs
```diff
@@ -30,14 +30,22 @@ pub const MAX_ERROR_LEN: u64 = 256;
 pub struct ErrorType(pub VariableList<u8, MaxErrorLen>);
 
 impl From<String> for ErrorType {
+    // This will panic if `string.as_bytes()` exceeds `MaxErrorLen`.
     fn from(s: String) -> Self {
-        Self(VariableList::from(s.as_bytes().to_vec()))
+        Self(
+            VariableList::try_from(s.as_bytes().to_vec())
+                .expect("length should not exceed MaxErrorLen"),
```
michaelsproul: Maybe we should keep the truncating behaviour here. It would be bad to crash LH due to an error being too long.

We don't necessarily need a method on `VariableList`, we could just do `.truncate()` on the bytes before passing them to `VariableList::try_from`.

beacon_node/lighthouse_network/src/rpc/methods.rs
```diff
@@ -30,14 +30,22 @@ pub const MAX_ERROR_LEN: u64 = 256;
...
     }
 }
 
 impl From<&str> for ErrorType {
+    // This will panic if `string.as_bytes()` exceeds `MaxErrorLen`.
     fn from(s: &str) -> Self {
-        Self(VariableList::from(s.as_bytes().to_vec()))
+        Self(
+            VariableList::try_from(s.as_bytes().to_vec())
+                .expect("length should not exceed MaxErrorLen"),
```
michaelsproul: As above. This impl could probably be the foundational one, and the `String` one could be implemented in terms of this.

consensus/types/src/attestation.rs
```diff
@@ -625,13 +625,17 @@ impl SingleAttestation {
     pub fn to_indexed<E: EthSpec>(&self, fork_name: ForkName) -> IndexedAttestation<E> {
         if fork_name.electra_enabled() {
             IndexedAttestation::Electra(IndexedAttestationElectra {
-                attesting_indices: vec![self.attester_index].into(),
+                attesting_indices: vec![self.attester_index]
+                    .try_into()
+                    .expect("Single index should not exceed MaxValidatorsPerSlot"),
```
michaelsproul: This probably fails clippy, maybe we should just make this method return a `Result`. It probably isn't used in that many places

testing/ef_tests/src/cases/ssz_generic.rs
```diff
@@ -326,6 +326,6 @@ where
             N::to_usize()
         )))
     } else {
-        Ok(decoded.into())
+        Ok(decoded.try_into().unwrap())
```
michaelsproul: Could probably remove the `unwrap` here and simplify, by combining the error case (`Too many values`) with the `try_into`. Something like `try_into().map_err(...)` and get rid of the manual check

michaelsproul (changes requested):
Looking good!

Just a few expects here and there that I wonder if we should remove

consensus/types/src/attestation.rs
macladson: I tried that initially, but `to_indexed` gets used in attestation verification which then requires a new error variant. That error variant then needs to be covered in match statements in some networking code and it ended up being really ugly just for this one error which can never occur anyway. The `expect` felt like the lesser of two evils. I'll keep trying and see if I can come up with something cleaner

testing/ef_tests/src/cases/ssz_generic.rs
macladson: Nice idea, I like it :+1:

michaelsproul:
Sorry @macladson, after discussing with Jimmy we decided it would be better to remove these `expect`s before merging. Can you have another go at propagating the errors, especially the ones for the JSON types, and anything that is reachable at runtime (anything outside a `static`/`const`).

beacon_node/execution_layer/src/engine_api/json_structures.rs
macladson: Here is the `expect`-less version: https://github.com/sigp/lighthouse/pull/8032/commits/6aa826bbb7c058f1ee2ecee1e220142efaa9f239

consensus/types/src/attestation.rs
macladson: `expect`-less version: https://github.com/sigp/lighthouse/pull/8032/commits/12c260a630e31d5dc67a140e9981118b48ce2450

beacon_node/lighthouse_network/src/rpc/methods.rs
```diff
@@ -29,15 +29,21 @@ pub const MAX_ERROR_LEN: u64 = 256;
...
-    fn from(s: String) -> Self {
-        Self(VariableList::from(s.as_bytes().to_vec()))
+impl From<&str> for ErrorType {
+    // This will truncate the error if `string.as_bytes()` exceeds `MaxErrorLen`.
+    fn from(s: &str) -> Self {
+        let mut bytes = s.as_bytes().to_vec();
+        bytes.truncate(MAX_ERROR_LEN as usize);
+        Self(
+            VariableList::try_from(bytes)
+                .expect("length should not exceed MaxErrorLen after truncation"),
```
macladson, line 39: Should I convert this whole impl to a `TryFrom`?

consensus/types/src/chain_spec.rs
```diff
@@ -2031,7 +2031,8 @@ fn max_data_columns_by_root_request_common<E: EthSpec>(max_request_blocks: u64)
 
     let empty_data_columns_by_root_id = DataColumnsByRootIdentifier {
         block_root: Hash256::zero(),
-        columns: VariableList::from(vec![0; E::number_of_columns()]),
+        columns: VariableList::try_from(vec![0; E::number_of_columns()])
+            .expect("VariableList::try_from should succeed"),
```
macladson: This function could also be a `Result<usize, Error>`

beacon_node/lighthouse_network/src/rpc/methods.rs
michaelsproul: I think for the error messages it is fine to truncate, otherwise we have errors upon errors

consensus/types/src/chain_spec.rs
michaelsproul: This one seems like it could be a bit of an annoying one to ripple, the error would bubble up to `ChainSpec::apply_to_chain_spec` which returns an `Option` (we'd probably have to silently drop the detail of the error and return `None`).

There is also a precedent for using `expect` in this function, there's an existing one on the line below this.

I think the cleanest fix would actually be to add `VariableList::repeat_full(0)` which creates a `VariableList<T, N>` of length `N` full of the provided value. This function would not need to return a `Result`. For `FixedVector` it could be even cleaner, just `::repeat(0)`.

macladson: Done in 8de1f30f5d6340b382eac312fef1652744c7e7d4

michaelsproul:
Tagging this for v8.1.0, we can merge shortly after we split off the branch for v8.0.0 (any day now)

michaelsproul (approved):

PR #8033, opened 2025-09-12
Add missing event in `PendingComponent` span and clean up sync logs

jimmygchen:
## Issue Addressed

I was looking into some long `PendingComponents` span and noticed the block event wasn't added to the span, so it wasn't possible to see when the block was added from the trace view, this PR fixes this.

image

Additionally I've noticed a lot of noises and confusion in sync logs due to the initial`peer_id` being included as part of the syncing chain span, causing all logs under the span to have that `peer_id`, which may not be accurate for some sync logs, I've removed `peer_id` from the `SyncingChain` span, and also cleaned up a bunch of spans to use `%` (display) for slots and epochs to make logs easier to read.

### DIFF
--- a/beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
+++ b/beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
@@ -657,11 +657,13 @@ impl<T: BeaconChainTypes> DataAvailabilityCheckerInner<T> {
             None
         };
 
-        debug!(
-            component = "block",
-            status = pending_components.status_str(num_expected_columns_opt),
-            "Component added to data availability checker"
-        );
+        pending_components.span.in_scope(|| {
+            debug!(
+                component = "block",
+                status = pending_components.status_str(num_expected_columns_opt),
+                "Component added to data availability checker"
+            );
+        });
 
         self.check_availability_and_cache_components(
             block_root,
--- a/beacon_node/network/src/sync/range_sync/batch.rs
+++ b/beacon_node/network/src/sync/range_sync/batch.rs
@@ -1,4 +1,5 @@
 use beacon_chain::block_verification_types::RpcBlock;
+use derivative::Derivative;
 use lighthouse_network::PeerId;
 use lighthouse_network::rpc::methods::BlocksByRangeRequest;
 use lighthouse_network::service::api_types::Id;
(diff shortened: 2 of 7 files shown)


beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -131,8 +131,14 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
...
-        skip(id),
-        fields(chain_id = %id)
+        skip_all,
+        fields(
+            chain_id = %id,
+            start_epoch = %start_epoch,
+            target_head_slot = %target_head_slot,
+            target_head_root = %target_head_root,
+            chain_type = %chain_type,
+        )
```
jimmygchen, line 141: skip all and explicitly add fields to use `Display` for some fields and exclude `peer_id`

beacon_node/network/src/sync/range_sync/batch.rs
```diff
@@ -114,6 +116,7 @@ pub struct BatchInfo<E: EthSpec, B: BatchConfig = RangeSyncBatchConfig> {
     /// Whether this batch contains all blocks or all blocks and blobs.
     batch_type: ByRangeRequestType,
     /// Pin the generic
+    #[derivative(Debug = "ignore")]
```
jimmygchen, line 119: the debug impl of `PhantomData` is very noisy and not really helpful in this case

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -666,11 +666,13 @@ impl<T: BeaconChainTypes> DataAvailabilityCheckerInner<T> {
...
-            status = pending_components.status_str(num_expected_columns_opt),
-            "Component added to data availability checker"
-        );
+        pending_components.span.in_scope(|| {
+            debug!(
+                component = "block",
+                status = pending_components.status_str(num_expected_columns_opt),
+                "Component added to data availability checker"
+            );
+        });
```
jimmygchen, line 675: log this under the span to make sure the debug event is  included in the trace/span

beacon_node/network/src/sync/range_sync/chain.rs
chong-he: So from my understanding this `skip_all` will skip all fields except explicitly states in the `fields()`, so it will skip the tracing for `peers: HashSet<PeerId>` as described in the PR

chong-he (approved):
Nice. I like the replacement of skip(xx) with `skip_all` and only specify the fields we want to include, making it easier to maintain and more future proof.

beacon_node/network/src/sync/range_sync/chain.rs
jimmygchen: yep that's right! and there's always a single peer when creating a lookup, and its often very misleading because new peers can get added to the lookup later, we'd then log incorrect peers when sending out requests.

PR #8039, opened 2025-09-12
Regression test for range sync CGC race condition

jimmygchen:
## Description

Regression test for a sync race condition where range sync gets stuck due to a CGC (custody group count) metadata timing issue. The original bug was discovered in CI kurtosis sync tests: https://github.com/sigp/lighthouse/actions/runs/17632496975/job/50102884940

The core fix (calling `attempt_send_awaiting_download_batches` on `SyncingChain::resume`) was merged in #8230. This PR adds a regression test that reproduces the exact scenario, plus related fixes.

**Changes:**

1. **Regression test** (`finalized_sync_not_enough_custody_peers_resume_after_peer_cgc_update`): reproduces the failing scenario and verifies that sync resumes correctly when CGC updates arrive.

2. **Moved to separate PR:** ~~**Count `AwaitingDownload` batches in buffer size check** (range sync + backfill sync): the `in_buffer` check previously only counted `Downloading` and `AwaitingProcessing` batches. This meant `AwaitingDownload` batches didn't count toward `BATCH_BUFFER_SIZE`, so new batches could be created while `AwaitingDownload` ones accumulated. Now `AwaitingDownload` is included so the buffer limit applies properly.~~

3. **Moved to separate PR**
: ~~**Handle `NoPeer` error without fake-failing the batch**: previously, when `send_batch` got a `NoPeer` error, it would fake-start the download (`start_downloading(1)`), then call `download_failed` to increment the failure count. This incorrectly penalized the batch for something that wasn't a real download failure. Now `NoPeer` just logs a debug message and leaves the batch in `AwaitingDownload` (a valid in-between state since #7984), letting it be retried when peers become available.~~


beacon_node/network/src/sync/tests/range.rs
```diff
@@ -601,3 +602,72 @@ fn finalized_sync_not_enough_custody_peers_on_start() {
...
+
+    // Peer 3 is connected and advanced
+    let peer_3 = r.new_connected_supernode_peer_no_metadata_custody_subnet();
+    r.send_sync_message(SyncMessage::AddPeer(peer_3, remote_info));
+    // We are still in finalized sync state (now with peer 3 added)
+    r.assert_state(RangeSyncType::Finalized);
+
+    for (i, p) in [peer_1, peer_2, peer_3].iter().enumerate() {
+        let peer_idx = i + 1;
+        println!("Peer {peer_idx}: {p:?}");
```
dapplion: Better this way so logs are ordered
```suggestion
        r.log(format!("Peer {peer_idx}: {p:?}"));
```

beacon_node/network/src/sync/tests/range.rs
```diff
@@ -601,3 +602,72 @@ fn finalized_sync_not_enough_custody_peers_on_start() {
...
+        return;
+    }
+    let mut r = TestRig::test_setup();
+
+    // GIVEN: the node is connected to 3 supernode peers:
+
+    // Peer 1 is synced (same finalized epoch)
+    let peer_1 = r.new_connected_supernode_peer_no_metadata_custody_subnet();
+    let remote_info = r.local_info().clone();
+    r.send_sync_message(SyncMessage::AddPeer(peer_1, remote_info));
```
dapplion, line 663: Add a helper function

```rs
fn add_peer(..) {
    r.send_sync_message(SyncMessage::AddPeer(peer, info))
}
```

beacon_node/network/src/sync/tests/range.rs
```diff
@@ -601,3 +602,72 @@ fn finalized_sync_not_enough_custody_peers_on_start() {
...
+
+    // Peer 2 is advanced (local finalized epoch + 2)
+    let advanced_epochs: u64 = 2;
+    let peer_2 = r.new_connected_supernode_peer_no_metadata_custody_subnet();
+    let remote_info = r.finalized_remote_info_advanced_by(advanced_epochs.into());
+    r.send_sync_message(SyncMessage::AddPeer(peer_2, remote_info.clone()));
+    // We expect a finalized chain to be created with peer 2, but no requests sent out yet due to missing custody info.
```
dapplion, line 670: You can assert this with
```rs
    r.expect_empty_network(); // expect no requests
```

beacon_node/network/src/sync/tests/range.rs
```diff
@@ -601,3 +602,72 @@ fn finalized_sync_not_enough_custody_peers_on_start() {
...
+        let peer_idx = i + 1;
+        println!("Peer {peer_idx}: {p:?}");
+    }
+
+    // WHEN: peer 1 sends its CGC via metadata response
+    let all_custody_subnets = (0..r.spec.data_column_sidecar_subnet_count)
+        .map(|i| i.into())
+        .collect::<HashSet<_>>();
+    r.set_peer_custody_subnets(&peer_1, all_custody_subnets.clone());
+    r.send_sync_message(SyncMessage::UpdatedPeerCgc(peer_1));
```
dapplion: Move this into a helper method for readability `set_peer_to_all_custody_subnets`

dapplion (approved):
Nice catch and good test! Some minor comments

beacon_node/network/src/sync/tests/range.rs
jimmygchen: This is really trivial and I could just do it, but I thought I'd mention what I think about these tests - I feel like a help function for this one liner doesn't really help with readability because:
1. It hides what component we're interacting with and the pattern and intent we're employing in these tests - i.e. interacting with sync manager, and making assertions on `network`. I worry that if we don't **be explicit** about focusing on message-driven tests and interacting mainly with public APIs, we'd easily end up reverting back to internal state manipulation and assertions, and making things more unmaintainable. 
3. Flexibility and dev UX: We have a bunch of `add*peer` and `new_*peer` functions, and they're already confusing enough that I have to look into the function impl every time i use them to know if it fits my purpose - some of them are adding peer to db, and some are adding to db + sending sync message. 

Not a strong opinion on this, it might just be me, but I struggled a bit when I'm working / debugging these tests, and had some thoughts about this. My preference is to avoid over-extracting these trivial functions unless the logic is more involved .e.g `find_blocks_by_range_request`.

beacon_node/network/src/sync/tests/range.rs
jimmygchen: Done

beacon_node/network/src/sync/tests/range.rs
jimmygchen: Done

beacon_node/network/src/sync/tests/range.rs
jimmygchen: Done

jimmygchen:
I had a look at the failing checkpoint sync tests (optionally triggered jobs), and I don't believe they are related to changes from this PR, i think what's happening is:
* The sepolia checkpoint server used is currently 4 days behind, so it takes longer to sync
* The devnet was under non finality tests, and it was just taking longer to sync.
* Github-hosted runner struggles to sync supernode on peerdas devnet due to capacity, to be resovled in #7804 once we have dedicated kurtosis runners.

jimmygchen:
@dapplion thanks for the review! I've addressed most of your comments.

pawanjay176 (commented):
FYI, running this on devnet 3 i'm getting a stuck chain with 40 peers. 
Need to investigate further

jimmygchen:
Thanks for raising this Pawan. Sorry i haven't managed to get to this today, will look into this tomorrow.

beacon_node/network/src/sync/tests/range.rs
dapplion: Fine either way! I see your point, let's leave it like this

dapplion (approved):

jimmygchen:
> FYI, running this on devnet 3 i'm getting a stuck chain with 40 peers. Need to investigate further

@pawanjay176 im going to remove the clean ups from this PR so we can merge this, because the flaky sync test is really quite annoying 😅

jimmygchen:
I've been running this branch on devnet-3 and it seems to be syncing fine and backfilling for the last ~9 hours:
image

beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -1063,7 +1063,9 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
         let in_buffer = |batch: &BackFillBatchInfo<T::EthSpec>| {
             matches!(
                 batch.state(),
-                BatchState::Downloading(..) | BatchState::AwaitingProcessing(..)
+                BatchState::AwaitingDownload
+                    | BatchState::Downloading(..)
+                    | BatchState::AwaitingProcessing(..)
```
dapplion: Looks good 👍 this was an oversight from https://github.com/sigp/lighthouse/pull/7984

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -1082,16 +1082,12 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
...
-                    // sync to work properly it must be okay to have "stalled" batches in
-                    // AwaitingDownload state. Currently it will error with invalid state if
-                    // that happens. Sync manager must periodicatlly prune stalled batches like
-                    // we do for lookup sync. Then we can deprecate the redundant
-                    // `good_peers_on_sampling_subnets` checks.
-                    e
-                    @ (RpcRequestSendError::NoPeer(_) | RpcRequestSendError::InternalError(_)) => {
+                    RpcRequestSendError::NoPeer(err) => {
+                        debug!(error = ?err, "Did not send batch request due to insufficient peers");
+                    }
```
dapplion: Fine to ignore the error here, batch will remain in AwaitingDownload state
- If a peer joins later: `attempt_send_awaiting_download_batches` is called ✅ 
- If a peer never joins: No timeout or pruning mechanism exists for stalled chains, can stall sync ⚠️ 

For safety we should have a pruning interval like for lookups

dapplion (changes requested):

dapplion:
Hey @jimmygchen, I pushed a follow-up on top of your branch that adds periodic pruning for stalled range sync chains: [`prune-stalled-range-sync-chains`](https://github.com/dapplion/lighthouse/tree/prune-stalled-range-sync-chains)

**Changes (4 files, +48 lines):**

1. **Periodic retry interval** (`manager.rs`): adds a 15s `tokio::time::interval` that calls `range_sync.resume()` as a safety net — retries `AwaitingDownload` batches in case a `UpdatedPeerCgc` event was missed.

2. **Per-batch stall detection** (`batch.rs`, `chain.rs`): adds `created_at: Instant` to `BatchInfo` to track batch age. In the `NoPeer` arm of `send_batch`, if a batch has been alive for >10 minutes (`CHAIN_MAX_STALL_DURATION`) without finding peers, the chain is failed with `ChainFailed { blacklist: false }`. This triggers `on_chain_removed` → STATUS to peers → chain gets recreated fresh when peers respond.

3. **Metrics** (`metrics.rs`):
   - `sync_range_chain_batch_lifetime_seconds` — histogram of batch lifetime from creation to completion
   - `sync_range_chain_failed_no_peers_total` — counter for chains failed due to stalled batches with no peers

This follows the pattern from lookup sync's periodic pruning, but adapted for range sync where the pruning happens on `NoPeer` rather than as a separate sweep.

jimmygchen:
@dapplion Thanks for the follow-up! The periodic retry and stall timeout are both good ideas. I spotted a gap in the timeout placement though — the check is in `send_batch()`'s `NoPeer` arm, but `attempt_send_awaiting_download_batches` only calls `send_batch` when `good_peers_on_sampling_subnets` returns true:

```rust
for batch_id in awaiting_downloads {
    if self.good_peers_on_sampling_subnets(batch_id, network) {
        self.send_batch(network, batch_id)?;  // timeout checked here
    } else {
        debug!("Waiting for peers...");  // timeout NOT checked here
    }
}
```

If there are zero peers on the right subnets, `send_batch` is never called and the timeout never fires — which is exactly the scenario where we'd most need it.

I've pulled the stall check out of `send_batch` and instead run it after the send loop in `attempt_send_awaiting_download_batches` as a separate `check_stalled_batches()` method. This catches both paths (NoPeer from network AND no peers on sampling subnets) in one place, and keeps `send_batch` focused on sending. Also gated the periodic `resume()` on `!self.range_sync.is_idle()` to avoid unnecessary work when the node is synced. Will push a commit shortly.

jimmygchen:
Tested this on two mainnet nodes to compare, and haven't noticed any differences.

I think to reproduce this we need to hit some peering issues, which may be hard to reproduce, i'll let it run for a bit and report back.

jimmygchen:
@dapplion please hold off from merging this. I'm seeing a trend of increasing `sync_active_network_requests` for `components_by_range` running this branch. Will investigate further tomorrow.

jimmygchen:
^ unrelated to this PR i think. I've made a PR to clean it up:

* https://github.com/sigp/lighthouse/pull/8890

jimmygchen:
@pawanjay176 @dapplion 
I believe I've addressed all the prior comments and this change is still required. Could you please re-review?
This PR doesn't fix the potential leaks, but prevents sending an unbounded number of batches upon resuming sync.
The leak is addressed in the link PR above.

jimmygchen:
@pawanjay176 I've looked at this PR again with @michaelsproul - this PR contains two changes that can be separated:
1. The regression test for a CGC race condition that has already been fixed a while ago - this is a low risk and low hanging fruit, but not urgent for v8.1.1.
2. The `NoPeer` clean up, which makes it possible to create a batch in `AwaitingDownload` state, and therefore needs to be bundled with the `in_buffer` check AND the stall detection. This is mainly a clean up. On current `unstable`, the batch never stays in `AwaitingDownload` after send batch returns, so these batches can't accumulate. Therefore this is not a required fix for v8.1.1 either.

jimmygchen:
Rebased on `unstable` and split this PR into two:

- This PR now contains only the **regression test** for the CGC race condition (change 1).
- The **NoPeer cleanup** (handle `NoPeer` without fake-failing the batch, count `AwaitingDownload` in buffer checks, stall detection, periodic retry) has been moved to #8919.

pawanjay176 (approved):
Good test, lets merge!

Issue #7678, opened 2025-06-30
Tree sync


dapplion:
Currently forwards sync fetches ranges of blocks by range from either head or latest finalized until reaching the target, from a set of peers with the same status messages. This approach has fundamental issues that make the sync complex and susceptible to griefing attacks. (I'll describe those later)

### Proposal

Divide forward sync into two phases:

1. Header backfill: construct an in-memory tree from all unknown received by peers (status, unknown attestations, etc) and backfill until reaching a known block or conflicting with finality.
2. Forward sync with knowledge of what block roots to expect at each slot and perfect peer grouping.

This model allows us to address all those issues and delete _a lot_ of code. All lookup sync, by_range routes, batching, optimistic batches, among others, are deleted.

#### Internal impl

PoC https://github.com/dapplion/lighthouse/pull/46

0. Add more test coverage of sync tests (agnostic of range or lookup sync) https://github.com/sigp/lighthouse/pull/8592
1. When creating a lookup fetch its ancestor headers until one is known or conflicts with finality
2. Introduce a new ReqResp route `headers_by_root` to make 1. fast and efficient
3. Make lookup sync downloads more efficient by buffering downloads (like range sync does)
4. Make lookup sync processing efficient by processing its components in batch 
5. Remove the max-depth limit of lookup and sync, and trigger lookup sync when a peer's status message contains an unknown root
6. Delete range sync

#### Spec

- [ ] Add `headers_by_root` p2p spec

### Related resources

- Early exploration on modifying the spec to address one of the issues above https://github.com/sigp/lighthouse/pull/7096

dapplion:
I propose to implement tree-sync progressively instead of in one big rewrite like https://github.com/dapplion/lighthouse/pull/46. The steps I propose next are well scoped and contained, so we will have a much simpler transition introducing the concepts of tree-sync one by one.

Tree sync is basically extending lookup sync to be able to handle very large chains of blocks, replacing the need for range sync. The key changes are:

0. Add more test coverage of sync tests (agnostic of range or lookup sync)
1. When creating a lookup fetch its ancestor headers until one is known or conflicts with finality
2. Introduce a new ReqResp route `headers_by_root` to make 1. fast and efficient
3. Make lookup sync downloads more efficient by buffering downloads (like range sync does)
4. Make lookup sync processing efficient by processing its components in batch 
5. Remove the max-depth limit of lookup and sync, and trigger lookup sync when a peer's status message contains an unknown root
6. Delete range sync

PR #8042, opened 2025-09-13
Don't penalize peers for extending ignored chains

dapplion:
## Issue Addressed

Lookup sync has a cache of block roots "failed_chains". If a peer triggers a lookup for a block or descendant of a root in failed_chains the lookup is dropped and the peer penalized. However blocks are inserted into failed_chains for a single reason:
- If a chain is longer than 32 blocks the lookup is dropped to prevent OOM risks. However the peer is not at fault, since discovering an unknown chain longer than 32 blocks is not malicious. We just drop the lookup to sync the blocks from range forward sync.

This discrepancy is probably an oversight when changing old code. Before we used to add blocks that failed too many times to process to that cache. However, we don't do that anymore. Adding a block that fails too many times to process is an optimization to save resources in rare cases where peers keep sending us invalid blocks. In case that happens, today we keep trying to process the block, downscoring the peers and eventually disconnecting them. _IF_ we found that optimization to be necessary we should merge this PR (_Stricter match of BlockError in lookup sync_) first. IMO we are fine without the failed_chains cache and the ignored_chains cache will be obsolete with [tree sync](https://github.com/sigp/lighthouse/issues/7678) as the OOM risk of long lookup chains does not exist anymore.

Closes https://github.com/sigp/lighthouse/issues/7577

## Proposed Changes

Rename `failed_chains` for `ignored_chains` and don't penalize peers that trigger lookups for those blocks

### DIFF
--- a/beacon_node/network/src/sync/block_lookups/mod.rs
+++ b/beacon_node/network/src/sync/block_lookups/mod.rs
@@ -59,7 +59,7 @@ mod single_block_lookup;
 /// reaches the maximum depth it will force trigger range sync.
 pub(crate) const PARENT_DEPTH_TOLERANCE: usize = SLOT_IMPORT_TOLERANCE;
 
-const FAILED_CHAINS_CACHE_EXPIRY_SECONDS: u64 = 60;
+const IGNORED_CHAINS_CACHE_EXPIRY_SECONDS: u64 = 60;
 pub const SINGLE_BLOCK_LOOKUP_MAX_ATTEMPTS: u8 = 4;
 
 /// Maximum time we allow a lookup to exist before assuming it is stuck and will never make
@@ -131,7 +131,7 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
     pub fn new() -> Self {
         Self {
             ignored_chains: LRUTimeCache::new(Duration::from_secs(
-                FAILED_CHAINS_CACHE_EXPIRY_SECONDS,
+                IGNORED_CHAINS_CACHE_EXPIRY_SECONDS,
             )),
             single_block_lookups: Default::default(),
         }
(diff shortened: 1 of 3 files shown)


beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -128,21 +130,21 @@ pub(crate) type BlockLookupSummary = (Id, Hash256, Option<Hash256>, Vec<PeerId>)
 impl<T: BeaconChainTypes> BlockLookups<T> {
     pub fn new() -> Self {
         Self {
-            failed_chains: LRUTimeCache::new(Duration::from_secs(
+            ignored_chains: LRUTimeCache::new(Duration::from_secs(
                 FAILED_CHAINS_CACHE_EXPIRY_SECONDS,
```
jimmygchen: Please update the const name `FAILED_CHAINS_CACHE_EXPIRY_SECONDS`

beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -331,11 +333,8 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
         cx: &mut SyncNetworkContext<T>,
     ) -> bool {
         // If this block or it's parent is part of a known failed chain, ignore it.
```
jimmygchen: please update comment

beacon_node/network/src/sync/tests/lookups.rs
```diff
@@ -1587,7 +1584,7 @@ fn test_child_lookup_not_created_for_failed_chain_parent_after_processing() {
 
     // At this point, the chain should have been deemed too deep and pruned.
     // The tip root should have been inserted into failed chains.
```
jimmygchen: please update comment

beacon_node/network/src/sync/tests/lookups.rs
```diff
@@ -1606,8 +1603,7 @@ fn test_child_lookup_not_created_for_failed_chain_parent_after_processing() {
 
     // THEN: The extending block should not create a lookup because the tip was inserted into failed chains.
```
jimmygchen: same as above

jimmygchen (changes requested):
Looks good to me, just a few minor doc update / naming comments

pawanjay176 (approved):
LGTM, sans jimmy's comments

jimmygchen (approved):
LGTM!

PR #8045, opened 2025-09-15
Consolidate `reqresp_pre_import_cache`  into `data_availability_checker`

jimmygchen:
## Issue Addressed

This PR consolidates the `reqresp_pre_import_cache` into the `data_availability_checker` for the following reasons:
- the `reqresp_pre_import_cache` suffers from the same TOCTOU bug we had with `data_availability_checker` earlier, and leads to unbounded memory leak, which we have observed over the last 6 months on some nodes.
- the `reqresp_pre_import_cache` is no longer necessary, because we now hold blocks in the `data_availability_checker` for longer since (#7961), and recent blocks can be served from the DA checker.  

This PR also maintains the following functionalities
- Serving pre-executed blocks over RPC, and they're now served from the `data_availability_checker` instead.
- Using the cache for de-duplicating lookup requests.

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -340,10 +340,6 @@ pub enum BlockProcessStatus<E: EthSpec> {
     ExecutionValidated(Arc<SignedBeaconBlock<E>>),
 }
 
-pub struct BeaconChainMetrics {
-    pub reqresp_pre_import_cache_len: usize,
-}
-
 pub type LightClientProducerEvent<T> = (Hash256, Slot, SyncAggregate<T>);
 
 pub type BeaconForkChoice<T> = ForkChoice<
@@ -362,9 +358,6 @@ pub type BeaconStore<T> = Arc<
         <T as BeaconChainTypes>::ColdStore,
     >,
 >;
-
-/// Cache gossip verified blocks to serve over ReqResp before they are imported
-type ReqRespPreImportCache<E> = HashMap<Hash256, Arc<SignedBeaconBlock<E>>>;
 
 /// Represents the "Beacon Chain" component of Ethereum 2.0. Allows import of blocks and block
 /// operations and chooses a canonical head.
@@ -462,8 +455,6 @@ pub struct BeaconChain<T: BeaconChainTypes> {
     pub(crate) attester_cache: Arc<AttesterCache>,
     /// A cache used when producing attestations whilst the head block is still being imported.
     pub early_attester_cache: EarlyAttesterCache<T::EthSpec>,
-    /// Cache gossip verified blocks to serve over ReqResp before they are imported
-    pub reqresp_pre_import_cache: Arc<RwLock<ReqRespPreImportCache<T::EthSpec>>>,
     /// A cache used to keep track of various block timings.
     pub block_times_cache: Arc<RwLock<BlockTimesCache>>,
     /// A cache used to track pre-finalization block roots for quick rejection.
@@ -1289,18 +1280,8 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
     /// chain. Used by sync to learn the status of a block and prevent repeated downloads /
     /// processing attempts.
     pub fn get_block_process_status(&self, block_root: &Hash256) -> BlockProcessStatus<T::EthSpec> {
-        if let Some(block) = self
-            .data_availability_checker
-            .get_execution_valid_block(block_root)
-        {
-            return BlockProcessStatus::ExecutionValidated(block);
-        }
-
-        if let Some(block) = self.reqresp_pre_import_cache.read().get(block_root) {
-            // A block is on the `reqresp_pre_import_cache` but NOT in the
-            // `data_availability_checker` only if it is actively processing. We can expect a future
-            // event with the result of processing
-            return BlockProcessStatus::NotValidated(block.clone());
+        if let Some(cached_block) = self.data_availability_checker.get_cached_block(block_root) {
+            return cached_block;
         }
 
         BlockProcessStatus::Unknown
@@ -3054,8 +3035,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
 
         self.emit_sse_blob_sidecar_events(&block_root, std::iter::once(blob.as_blob()));
 
-        let r = self.check_gossip_blob_availability_and_import(blob).await;
-        self.remove_notified(&block_root, r)
+        self.check_gossip_blob_availability_and_import(blob).await
     }
 
     /// Cache the data columns in the processing cache, process it, then evict it from the cache if it was
@@ -3092,15 +3072,13 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             data_columns.iter().map(|column| column.as_data_column()),
         );
 
-        let r = self
-            .check_gossip_data_columns_availability_and_import(
-                slot,
-                block_root,
-                data_columns,
-                publish_fn,
-            )
-            .await;
-        self.remove_notified(&block_root, r)
+        self.check_gossip_data_columns_availability_and_import(
+            slot,
+            block_root,
+            data_columns,
+            publish_fn,
+        )
+        .await
     }
 
     /// Cache the blobs in the processing cache, process it, then evict it from the cache if it was
@@ -3139,10 +3117,8 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
 
         self.emit_sse_blob_sidecar_events(&block_root, blobs.iter().flatten().map(Arc::as_ref));
 
-        let r = self
-            .check_rpc_blob_availability_and_import(slot, block_root, blobs)
-            .await;
-        self.remove_notified(&block_root, r)
+        self.check_rpc_blob_availability_and_import(slot, block_root, blobs)
+            .await
     }
 
     /// Process blobs retrieved from the EL and returns the `AvailabilityProcessingStatus`.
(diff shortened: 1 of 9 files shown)


beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3336,46 +3306,6 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         }
     }
 
-    /// Remove any block components from the *processing cache* if we no longer require them. If the
-    /// block was imported full or erred, we no longer require them.
-    fn remove_notified(
```
jimmygchen, line 3345: I still need to handle the case where block import was errored and remove them from cache.

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3427,6 +3351,15 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             );
         }
 
+        self.data_availability_checker
+            .put_pre_execution_block(block_root, unverified_block.block_cloned())?;
```
dapplion, line 3359: Noting that if the block is unverified, so it can be 10 MB. The current max size of the cache is 32 blocks so there's no risk of OOM.

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -96,9 +128,17 @@ impl<E: EthSpec> PendingComponents<E> {
...
-    /// Inserts a block into the cache.
-    pub fn insert_block(&mut self, block: DietAvailabilityPendingExecutedBlock<E>) {
-        *self.get_cached_block_mut() = Some(block)
+    /// Inserts an executed block into the cache.
+    pub fn insert_executed_block(&mut self, block: DietAvailabilityPendingExecutedBlock<E>) {
+        self.block = Some(CachedBlock::Executed(Box::new(block)))
+    }
+
+    /// Inserts a pre-execution block into the cache.
+    /// This does NOT override an existing executed block.
```
dapplion, line 137: Noting that this will override a `PreExecution` block, but it is fine since the block root is check so the data is exactly the same. Do we care about controlling more the state transitions of this enum?

dapplion (commented):
Nice simplification! Can't find any issues with it after mutliple passes

beacon_node/beacon_chain/src/beacon_chain.rs
dapplion: The damage to leave the invalid block in the cache is that we might serve the block over RPC after knowing it's invalid. Consider the case where we remove the block from the da_checker after noticing it's invalid and then someone triggers lookup sync and we fetch it again. We will serve the block over RPC while verification is happening for the second time. 

I don't see that big of an issue with leaving the invalid block in the da_checker until LRU evicts it. Otherwise, we can leave it there but mark it as "INVALID"

beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
jimmygchen: I don't understand this - it only insert it if the block is `None`, so if it's already set it will not override. Am I missing something?

I considered more controlled state transition but i think it's enough to ensure one direction in following order:

None -> Pre execution -> Executed

We can never override an executed block

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Thinking out loud -

If the block is invalid we will downscore the peer. 

If we leave the invalid block in the cache
- we might serve the invalid block over RPC (we already are doing it before execution) - however, we are unlikely to get downscored because we will never attest or build on top of thisblock.
- we avoid downloading the invalid block again, even if a peer sends us a child block / attestation

If we remove the invalid block from the cache
- we still serve the invalid block over RPC before execution, but NOT after it failed validation
- a peer might get us to fetch again and and serve us the invalid block, we will downscore them after execution
- it doesn't' really give us much because we could still serve the invalid block over RPC before execution, and even if we remove it, it could come back again

> I don't see that big of an issue with leaving the invalid block in the da_checker until LRU evicts it.

Yes agree. I actually think leaving it in the cache to prevent the lookup is better.

> Otherwise, we can leave it there but mark it as "INVALID"

A block that does not transition into executed block will never become available so we don't really need to mark them I believe.

jimmygchen: Thanks for raising these good points, I'm now inclined to leave it as it is, as it doesn't seem to help us that much by removing it, as we don't really benefit from doing it for extra code and complexity.

jimmygchen: Adding more thoughts after seeing the failed `block_in_processing_cache_becomes_invalid` test.

If the block fails execution initially for whatever reason (e.g. engine offline), and we keep it in the cache, then there's a chance that the chain might get stuck, because we will never perform lookup until the block is evicted from the cache. I think this is not great, I'll add code to remove it so we re-trigger a new lookup in this case.

if it is indeed an invalid block, peer will get penalised for getting us to do the lookup.

jimmygchen: Removing failed execution block in https://github.com/sigp/lighthouse/pull/8045/commits/5f2e85c5625bbfa4a99aef809f461854476e2bd5

pawanjay176: > we will never perform lookup until the block is evicted from the cache

Just to be clear, is this because we cannot re-trigger the execution of a block sitting in the da checker? not fully clear on this.

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3448,7 +3381,18 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+                .into_executed_block(execution_pending)
+                .await
+                .inspect_err(|_| {
+                    // If the block fails execution for whatever reason (e.g. engine offline),
+                    // and we keep it in the cache, then the node will NOT perform lookup and
+                    // reprocess this block until the block is evicted from DA checker, causing the
+                    // chain to get stuck temporarily if the block is canonical. Therefore we remove
+                    // it from the cache if execution fails.
+                    self.data_availability_checker
+                        .remove_block_on_execution_error(&block_root);
```
pawanjay176, line 3394: I agree its safer to just chuck it out of the cache and let lookup fetch the block regardless of the type of the execution error.

pawanjay176 (approved):
This is looking good to me. Just a question.
Also, would be good to test this on the devnet with stopping the EL node and checking if it recovers as expected.

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Yes exactly - because lookup will not re-download the block if it's in the DA checker:

https://github.com/sigp/lighthouse/blob/ee734d145665e435f8ea0abfbead37ff61b38549/beacon_node/network/src/sync/network_context.rs#L835-L853

so if it fails the first time because of unexpected error, e.g. EL timeout,  then it doesn't gets re-downloaded, until we fallback to range sync

jimmygchen:
> This is looking good to me. Just a question. Also, would be good to test this on the devnet with stopping the EL node and checking if it recovers as expected.

Yep I was able to see a block lookup immediately sent after engine is restarted

```
[cl-1-lighthouse-geth] Sep 19 04:08:48.641 INFO  New block received                            slot: 61844, root: 0x1405fffdae52a9caff2ffd5f75fa13a2ec98804ce9a14b3445e3cf0c1f341c8a
[cl-1-lighthouse-geth] Sep 19 04:08:54.001 INFO  Synced                                        peers: "206", exec_hash: "0x785f3ae2c9317692a6cc3184d7a9e8a60c6c0a6a92a0502ec1dc480ddc07b9ea (verified)", finalized_root: 0xfaec29bf7d6f46d6800dba641bf72eea1cbd810a209c2974fe8c9ab6538987cf, finalized_epoch: 1930, epoch: 1932, block: "   …  empty", slot: 61844
[cl-1-lighthouse-geth] Sep 19 04:09:06.000 INFO  Synced                                        peers: "200", exec_hash: "0x785f3ae2c9317692a6cc3184d7a9e8a60c6c0a6a92a0502ec1dc480ddc07b9ea (verified)", finalized_root: 0xfaec29bf7d6f46d6800dba641bf72eea1cbd810a209c2974fe8c9ab6538987cf, finalized_epoch: 1930, epoch: 1932, block: "   …  empty", slot: 61845
[cl-1-lighthouse-geth] Sep 19 04:09:10.869 INFO  Execution engine online
[cl-1-lighthouse-geth] Sep 19 04:09:10.869 INFO  Issuing forkchoiceUpdated                     forkchoice_state: ForkchoiceState { head_block_hash: 0x785f3ae2c9317692a6cc3184d7a9e8a60c6c0a6a92a0502ec1dc480ddc07b9ea, safe_block_hash: 0x2ef47dbb5c2a4ea7910e8a99567cb8c18e874ffa72c3d3aa161a8bfa0a118f4c, finalized_block_hash: 0x6b024dc3cba5a5aa69c80f5bcd468512a27e88caafa6098e2c6e8fb6a3c6edd6 }
[cl-1-lighthouse-geth] Sep 19 04:09:13.350 INFO  New RPC block received                        slot: 61844, hash: 0x1405fffdae52a9caff2ffd5f75fa13a2ec98804ce9a14b3445e3cf0c1f341c8a, block_root: 0x1405fffdae52a9caff2ffd5f75fa13a2ec98804ce9a14b3445e3cf0c1f341c8a
```

PR #8050, opened 2025-09-16
Another check to prevent duplicate block imports

michaelsproul:
## Issue Addressed

Attempt to address performance issues caused by importing the same block multiple times.

## Proposed Changes

- Check fork choice "after" obtaining the fork choice write lock in `BeaconChain::import_block`. We actually use an upgradable read lock, but this is semantically equivalent (the upgradable read has the advantage of not excluding regular reads).

The hope is that this change has several benefits:

1. By preventing duplicate block imports we save time repeating work inside `import_block` that is unnecessary, e.g. writing the state to disk. Although the store itself now takes some measures to avoid re-writing diffs, it is even better if we avoid a disk write entirely.
2. By returning `DuplicateFullyImported`, we reduce some duplicated work downstream. E.g. if multiple threads importing columns trigger `import_block`, now only _one_ of them will get a notification of the block import completing successfully, and only this one will run `recompute_head`. This should help avoid a situation where multiple beacon processor workers are consumed by threads blocking on the `recompute_head_lock`. However, a similar block-fest is still possible with the upgradable fork choice lock (a large number of threads can be blocked waiting for the first thread to complete block import).

## Additional Info

Longer term this will likely be obsoleted by @ethDreamer's block status tracker.

### DIFF
--- a/beacon_node/beacon_chain/src/canonical_head.rs
+++ b/beacon_node/beacon_chain/src/canonical_head.rs
@@ -397,7 +397,7 @@ impl<T: BeaconChainTypes> CanonicalHead<T> {
     pub fn fork_choice_upgradable_read_lock(
         &self,
     ) -> RwLockUpgradableReadGuard<'_, BeaconForkChoice<T>> {
-        let _timer = metrics::start_timer(&metrics::FORK_CHOICE_READ_LOCK_AQUIRE_TIMES);
+        let _timer = metrics::start_timer(&metrics::FORK_CHOICE_UPGRADABLE_READ_LOCK_AQUIRE_TIMES);
         self.fork_choice.upgradable_read()
     }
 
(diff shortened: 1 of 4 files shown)


beacon_node/beacon_chain/src/canonical_head.rs
```diff
@@ -389,6 +393,14 @@ impl<T: BeaconChainTypes> CanonicalHead<T> {
         self.fork_choice.read()
     }
 
+    /// Access an upgradable read-lock for fork choice.
+    pub fn fork_choice_upgradable_read_lock(
+        &self,
+    ) -> RwLockUpgradableReadGuard<'_, BeaconForkChoice<T>> {
+        let _timer = metrics::start_timer(&metrics::FORK_CHOICE_READ_LOCK_AQUIRE_TIMES);
```
jimmygchen: i think you wanted to add new metric, just a reminder

michaelsproul: fixed in 6ae992c5ecd30a61cae1c1ceb6429a35a93a9552

jimmygchen (approved):

PR #8051, opened 2025-09-16
Remove column reconstruction when processing rpc requests

eserilev:
Prevent the node from attempting to reconstruct data columns during RPC request processing. Instead allow sync to send out requests to fetch missing columns. Reconstruction is a computationally expensive task, especially at higher blob counts and should be avoided when possible.

### DIFF
--- a/beacon_node/network/src/network_beacon_processor/sync_methods.rs
+++ b/beacon_node/network/src/network_beacon_processor/sync_methods.rs
@@ -383,7 +383,7 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
             "RPC custody data columns received"
         );
 
-        let mut result = self
+        let result = self
             .chain
             .process_rpc_custody_columns(custody_columns)
             .await;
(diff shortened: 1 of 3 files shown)


beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -404,17 +404,6 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                         block_hash = %block_root,
                         "Missing components over rpc"
                     );
-                    // Attempt reconstruction here before notifying sync, to avoid sending out more requests
-                    // that we may no longer need.
-                    // We don't publish columns reconstructed from rpc columns to the gossip network,
-                    // as these are likely historic columns.
-                    let publish_columns = false;
-                    if let Some(availability) = self
-                        .attempt_data_column_reconstruction(block_root, publish_columns)
```
jimmygchen, line 413: i think we can remove the `publish_columns` flag on this function now

eserilev: removed

jimmygchen (approved):
Nice, thanks!

Issue #8018, opened 2025-09-09
Enable reconstruction for nodes custody-ing more than 50% of columns


eserilev:
In `attempt_data_column_reconstruction` if a node isn't defined as a super node based on cli parameters, reconstruction wont be attempted. Instead I think we should check if a nodes custody count > 50% of total columns to decide wether we can attempt reconstruction.

https://github.com/sigp/lighthouse/blob/fd10b632740e441354c4547a12727c29dd288239/beacon_node/network/src/network_beacon_processor/mod.rs#L876-L889

jimmygchen:
Completed in #8052

PR #8052, opened 2025-09-16
Enable reconstruction for nodes custodying more than 50% of columns and instrument tracing

jimmygchen:
Addresses #8018 and more below:

* Fixes a bug where a validator custody supernode never trigger reconstruction (currently only cli supernodes do)
* Remove code to trigger immediate reconstruction if the message fails to be sent to beacon processor. 
* Trigger reconstruction if sampling column is > 50% of columns.
* Add tracing to data column reconstruction.

### DIFF
--- a/beacon_node/network/src/network_beacon_processor/gossip_methods.rs
+++ b/beacon_node/network/src/network_beacon_processor/gossip_methods.rs
@@ -1067,7 +1067,7 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                                 slot: *slot,
                                 process_fn: Box::pin(async move {
                                     cloned_self
-                                        .attempt_data_column_reconstruction(block_root, true)
+                                        .attempt_data_column_reconstruction(block_root)
                                         .await;
                                 }),
                             },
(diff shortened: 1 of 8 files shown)


beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -1054,36 +1053,43 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
-                        ..
-                    })) = send_result
+                    if !self
+                        .chain
+                        .data_availability_checker
+                        .custody_context()
+                        .should_attempt_reconstruction(
+                            slot.epoch(T::EthSpec::slots_per_epoch()),
+                            &self.chain.spec,
+                        )
```
eserilev:
```suggestion
                    if self
                        .chain
                        .data_availability_checker
                        .custody_context()
                        .should_attempt_reconstruction(
                            slot.epoch(T::EthSpec::slots_per_epoch()),
                            &self.chain.spec,
                        )
```

eserilev:
other than my comment, this looks good to me

eserilev (approved):

jimmygchen:
@mergify dequeue

PR #8053, opened 2025-09-16
Reduce reconstruction queue capacity

eserilev:
Reduce reconstruction queue capacity to 1. This makes sure our nodes don't queue up a bunch of reconstruction tasks which can degrade node performance pretty significantly.

### DIFF
--- a/beacon_node/beacon_processor/src/lib.rs
+++ b/beacon_node/beacon_processor/src/lib.rs
@@ -1354,9 +1354,7 @@ impl<E: EthSpec> BeaconProcessor<E> {
                             Work::RpcCustodyColumn { .. } => {
                                 rpc_custody_column_queue.push(work, work_id)
                             }
-                            Work::ColumnReconstruction(_) => {
-                                column_reconstruction_queue.push(work, work_id)
-                            }
+                            Work::ColumnReconstruction(_) => column_reconstruction_queue.push(work),
                             Work::ChainSegment { .. } => chain_segment_queue.push(work, work_id),
                             Work::ChainSegmentBackfill { .. } => {
                                 backfill_chain_segment.push(work, work_id)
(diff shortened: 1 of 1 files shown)


beacon_node/beacon_processor/src/lib.rs
```diff
@@ -181,7 +181,7 @@ impl BeaconProcessorQueueLengths {
             // We don't request more than `PARENT_DEPTH_TOLERANCE` (32) lookups, so we can limit
             // this queue size. With 48 max blobs per block, each column sidecar list could be up to 12MB.
             rpc_custody_column_queue: 64,
-            column_reconstruction_queue: 64,
+            column_reconstruction_queue: 1,
```
jimmygchen, line 184: should we also make it a `LifoQueue` instead?

jimmygchen (approved):
LGTM

beacon_node/beacon_processor/src/lib.rs
pawanjay176: Not that it matters with a queue length of 1, but wouldn't it make more sense to potentially complete the reconstruction task for an older slot first so that we can import it first. So in that case, wouldn't Fifo make more sense?

eserilev: FIFO will error if we try to push an item to the queue when the queue is full

Also it's not really beneficial, from a validator performance perspective, to reconstruct an older slots columns.

pawanjay176: > Also it's not really beneficial, from a validator performance perspective, to reconstruct an older slots columns.

We only try to reconstruct if its required for the availability check right? So if we have 64 columns for slot 100, then we won't import it until we reconstruct everything. If we get another reconstruction task for slot 101 at this point, then its dependant on slot 100 getting imported successfully.

pawanjay176: Actually, we won't create a reconstruction task here for 101 because the parent isn't imported yet so this isn't really an issue.

eserilev: sorry yes, when I said "older slot" i sort of meant a block thats already been imported. I think what you said is much clearer

PR #8058, opened 2025-09-16
fix(rate_limiter): add missing prune calls for light client protocols

gitToki:
The prune method is missing pruning calls for light client protocol rate limiters. Memory was allocated but never freed.

It was affecting 4 protocols:
  - lc_bootstrap_rl
  - lc_optimistic_update_rl
  - lc_finality_update_rl
  - lc_updates_by_range_rl

I got a confirmation from @jimmygchen before sending this PR


dapplion:
Thanks for your contribution! But to prevent forgetting things like this in the future, can you adopt this pattern?

https://github.com/sigp/lighthouse/blob/fd10b632740e441354c4547a12727c29dd288239/beacon_node/network/src/sync/network_context.rs#L303-L320

If we add new fields late,r we will get a compilation error

gitToki:
Updated, thanks for the suggestion @dapplion

jimmygchen (approved):
Nice, thanks for the fix!

Issue #7974, opened 2025-09-02
Add a metric for optimistic sync status


michaelsproul:
We should add a metric tracking the optimistic sync status of the head block. Something like:

```
head_block_execution_verified 0|1
```

where:

- `0` represents optimistic sync (head block is `Syncing` in fork choice)
- `1` represents proper sync (head block is `Valid` in fork choice)

I think phrasing it in the positive (1 = good) makes sense, because this maps onto a nice big "SYNCED" metric on a Grafana dashboard.

chong-he:
I would like to pick this up

michaelsproul:
Perfect, thanks CK!

chong-he:
Closed by #8059

PR #8059, opened 2025-09-16
Add `optimistic_sync` metric

chong-he:
## Issue Addressed

* #7974 

## Proposed Changes

Display `0` if optimistic sync and `1` for normal sync

## Additional Info

Tested and it is working, the optimistic sync is triggered by running  `lcli mock-el`  without the flag `--all-payloads-valid true`, so the beacon node sync is optimistic under this situation.

Screenshot from 2025-09-16 18-37-51


beacon_node/client/src/metrics.rs
```diff
@@ -15,6 +15,13 @@ pub static IS_SYNCED: LazyLock<Result<IntGauge>> = LazyLock::new(|| {
     )
 });
 
+pub static IS_OPTIMISTIC_SYNC: LazyLock<Result<IntGauge>> = LazyLock::new(|| {
+    try_create_int_gauge(
+        "optimistic_sync",
+        "Metric to check if the beacon chain is in optimistic sync mode. 0 if optimistic sync and 1 if synced",
```
michaelsproul: I think this is counterintuitive. In my original comment I was thinking we would name the metric something like `execution_verified 0|1`, with 0 representing the optimistic case and 1 representing full sync. _Or_ if we keep the `optimistic_sync` name I think we should swap it so that 1 = optimistic, 0 = not optimistic. That might actually be the clearest. We would just need to display the 0 as "good" in our grafana chart

chong-he: Revised in [revise](https://github.com/sigp/lighthouse/pull/8059/commits/bcb66c9bf378b40596672facfa3a2c295937936f)

If you think the metric should be placed in another place (e.g., execution layer metrics) and rename to `execution_verified`, then I can revise to that

michaelsproul (approved):

beacon_node/client/src/metrics.rs
michaelsproul: Yeah looks good! Thanks!

PR #8065, opened 2025-09-17
Fix reprocess queue memory leak

michaelsproul:
## Issue Addressed

Fix a memory leak in the reprocess queue.

## Proposed Changes

If the vec of attestation IDs for a block is never evicted from the reprocess queue by a `BlockImported` event, then it stays in the map forever consuming memory. The fix is to remove the entry when its last attestation times out. We do similarly for light client updates.

In practice this will only occur if there is a race between adding an attestation to the queue and processing the `BlockImported` event, or if there are attestations for block roots that we never import (e.g. random block roots, block roots of invalid blocks).

## Additional Info

- [x] New metrics added to track the sizes of these maps.
- [x] Regression tests added.


jimmygchen (approved):
Nice fix and tests!

PR #8066, opened 2025-09-17
Make the block cache optional

michaelsproul:
## Proposed Changes

Address contention on the store's `block_cache` by allowing it to be disabled when `--block-cache-size 0` is provided, and also making this the default.

## Additional Info

It would also be good to add a metric on `block_cache.lock()`, but I think this would require abstracting/copying some of the `CanonicalHeadRwLock` stuff and I haven't done that yet.


beacon_node/store/src/config.rs
```diff
@@ -19,7 +19,7 @@ pub const DEFAULT_BACKEND: DatabaseBackend = DatabaseBackend::LevelDb;
 pub const PREV_DEFAULT_SLOTS_PER_RESTORE_POINT: u64 = 2048;
 pub const DEFAULT_SLOTS_PER_RESTORE_POINT: u64 = 8192;
 pub const DEFAULT_EPOCHS_PER_STATE_DIFF: u64 = 8;
-pub const DEFAULT_BLOCK_CACHE_SIZE: NonZeroUsize = new_non_zero_usize(64);
+pub const DEFAULT_BLOCK_CACHE_SIZE: usize = 64;
```
michaelsproul: I'll change this to 0 by default

michaelsproul:
Fixed the tests, there was just a small issue with us not deleting from the state cache when the block cache was disabled. The fix is to always delete from the state cache, I don't think this should impose too much overhead :pray: But re-testing with a backfilling supernode should confirm this.

jimmygchen (approved):
Looks good! 
Tested this earlier and this fixes the slow `persist_block_and_blobs` and rpc serving behaviour 🎉

PR #8067, opened 2025-09-17
Max reconstruction delay as a function of slot time

### LINKED ISSUE #8054: Make maximum reconstruction delay a function of slot duration
Currently the max delay for reconstruction is set to three seconds. For a network with 12 second slot times, that means we trigger reconstruction 1/4 of the way into the slot. For networks with different slot times (i.e. 6s) this is not the case. We should make the max delay for reconstruction 1/4 of the way into the slot regardless of slot time.

PoulavBhowmick03:
## Issue Addressed

Fixes #8054

### DIFF
--- a/beacon_node/beacon_processor/src/scheduler/work_reprocessing_queue.rs
+++ b/beacon_node/beacon_processor/src/scheduler/work_reprocessing_queue.rs
@@ -84,8 +84,9 @@ pub const BACKFILL_SCHEDULE_IN_SLOT: [(u32, u32); 3] = [
     (4, 5),
 ];
 
-/// Trigger reconstruction if we are this many seconds into the current slot
-pub const RECONSTRUCTION_DEADLINE: Duration = Duration::from_millis(3000);
+/// Fraction of slot duration after which column reconstruction is triggered, makes it easier for
+/// different slot timings to have a generalised deadline
+pub const RECONSTRUCTION_DEADLINE: f64 = 0.25;
 
 /// Messages that the scheduler can receive.
 #[derive(AsRefStr)]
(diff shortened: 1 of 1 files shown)


beacon_node/beacon_processor/src/scheduler/work_reprocessing_queue.rs
```diff
@@ -754,13 +755,17 @@ impl<S: SlotClock> ReprocessQueue<S> {
             }
             InboundEvent::Msg(DelayColumnReconstruction(request)) => {
                 let mut reconstruction_delay = QUEUED_RECONSTRUCTION_DELAY;
+                let slot_duration = self.slot_clock.slot_duration().as_millis() as f64;
```
michaelsproul: We avoid doing any float arithmetic due to rounding issues and general flakiness. Integer arithmetic in milliseconds would be my preference

jimmygchen: yep, might be worth adding this preference to `CLAUDE.md` too

PoulavBhowmick03: > We avoid doing any float arithmetic due to rounding issues and general flakiness. Integer arithmetic in milliseconds would be my preference

got it. will make that change

PoulavBhowmick03: > We avoid doing any float arithmetic due to rounding issues and general flakiness. Integer arithmetic in milliseconds would be my preference

in that case, i was initially having `RECONSTRUCTION_DEADLINE` as a fraction (1/4) and then multiplying the slot_duration in millis with the numerator and denominator of the same. should i keep that approach?

```rust
let reconstruction_deadline_millis =
                    (slot_duration_millis * RECONSTRUCTION_DEADLINE.0)
                        / RECONSTRUCTION_DEADLINE.1;
                let reconstruction_deadline = Duration::from_millis(reconstruction_deadline_millis);
```

eserilev: I think the simplest is probably something like slot_duration_ms * reconstruction_deadline / 100. Or just divide by 4. Seems a bit unnecessary to multiply by one

eserilev (approved):
Not the biggest fan of the tuple "fraction" but I think this fine and we need it for the upcoming release. Once we merge this https://github.com/sigp/lighthouse/pull/7944 we may be able use similar logic to calculate the max reconstruction deadline and clean this up a bit

michaelsproul (approved):
Looks good to me, but we should test on some live networks to make sure this behaves as expected (could merge to unstable first I guess)

beacon_node/beacon_processor/src/scheduler/work_reprocessing_queue.rs
```diff
@@ -756,13 +757,17 @@ impl<S: SlotClock> ReprocessQueue<S> {
             }
             InboundEvent::Msg(DelayColumnReconstruction(request)) => {
                 let mut reconstruction_delay = QUEUED_RECONSTRUCTION_DELAY;
+                let slot_duration = self.slot_clock.slot_duration().as_millis() as u64;
+                let reconstruction_deadline_millis =
+                    (slot_duration * RECONSTRUCTION_DEADLINE.0) / RECONSTRUCTION_DEADLINE.1;
+                let reconstruction_deadline = Duration::from_millis(reconstruction_deadline_millis);
                 if let Some(seconds_from_current_slot) =
                     self.slot_clock.seconds_from_current_slot_start()
```
jimmygchen, line 765: This function only gives us granuarity to the seconds I believe

https://github.com/sigp/lighthouse/blob/522bd9e9c6ac167f2231525e937c9ebbcb86cf6e/common/slot_clock/src/lib.rs#L108-L110

So on a network like gnosis (5s slot times), reconstruction is only triggered immediately if `seconds_from_current_slot_start` is 2s

PR #8075, opened 2025-09-18
Run reconstruction inside a scoped rayon pool

eserilev:
- Introduced a high priority rayon thread pool which uses ~80% of cpu, or at least 3 threads. 
- Created a `spawn_blocking_handle_with_rayon` fn inside `TaskExecutor`.
- Data column reconstruction now runs within the new scoped high priority rayon pool.
- Moved `RayonManager` reference to `TaskExecutor` 

It isn't straightforward to convert the full Reconstruction work event from an async to a blocking task. This approach is a temporary stopgap  until we are able to fully separate async tasks from compute heavy tasks.


jimmygchen:
This PR addresses #7719

common/task_executor/src/rayon_manager.rs
```diff
@@ -0,0 +1,57 @@
+use rayon::{ThreadPool, ThreadPoolBuilder};
+use std::sync::Arc;
+
+const DEFAULT_LOW_PRIORITY_CPU_PERCENTAGE: usize = 25;
+const MINIMUM_LOW_PRIORITY_THREAD_COUNT: usize = 1;
+
+const DEFAULT_HIGH_PRIORITY_CPU_PERCENTAGE: usize = 80;
+const MINIMUM_HIGH_PRIORITY_THREAD_COUNT: usize = 3;
```
jimmygchen: I think we could keep this as 1 as well. 
Supernodes on live network should always have more than this, however this may not be the case for devnets, smaller network or integration tests

beacon_node/beacon_processor/src/lib.rs
```diff
@@ -1609,10 +1605,7 @@ impl<E: EthSpec> BeaconProcessor<E> {
...
             Work::ChainSegmentBackfill(process_fn) => {
                 if self.config.enable_backfill_rate_limiting {
-                    task_spawner.spawn_blocking_with_rayon(
-                        self.rayon_manager.low_priority_threadpool.clone(),
-                        process_fn,
-                    )
+                    task_spawner.spawn_blocking_with_rayon(RayonPoolType::LowPriority, process_fn)
                 } else {
                     // use the global rayon thread pool if backfill rate limiting is disabled.
                     task_spawner.spawn_blocking(process_fn)
```
jimmygchen, line 1611: We may want to use the high priority pool for this too, because moving reconstruction to a different pool means we could allocate up to 180% of available CPUs when backfill + reconstruction happens simultaneously, instead of the current 100% (both tasks sharing the same global threadpool)

beacon_node/beacon_processor/src/lib.rs
```diff
@@ -1681,10 +1674,11 @@ impl TaskSpawner {
     }
 
     /// Spawns a blocking task on a rayon thread pool, dropping the `SendOnDrop` after task completion.
-    fn spawn_blocking_with_rayon<F>(self, thread_pool: Arc<ThreadPool>, task: F)
+    fn spawn_blocking_with_rayon<F>(self, rayon_pool_type: RayonPoolType, task: F)
     where
         F: FnOnce() + Send + 'static,
     {
+        let thread_pool = self.executor.rayon_manager.get_thread_pool(rayon_pool_type);
         self.executor.spawn_blocking(
```
jimmygchen: We could have a `spawn_blocking_with_rayon` function, and `executor.rayon_manger` can be made private (its generally more maintainable and more testable with high encapsulation)

common/task_executor/src/rayon_pool_provider.rs
```diff
@@ -0,0 +1,55 @@
...
+    LowPriority,
+}
+
+pub struct RayonManager {
+    /// Smaller rayon thread pool for lower-priority, compute-intensive tasks.
+    /// By default ~25% of CPUs or a minimum of 1 thread.
+    low_priority_thread_pool: Arc<ThreadPool>,
+    /// Larger rayon thread pool for high-priority, compute-intensive tasks.
+    /// By default ~80% of CPUs or a minimum of 1 thread.
+    high_priority_thread_pool: Arc<ThreadPool>,
```
jimmygchen, line 20: I think now we're kinda treating the global rayon thread pool as the highest priority pool now, given that's what time sensitive block processing tasks use - we're allocating reconstruction to this threadpool because we don't want it to impact the high pri tasks - might be worth mentioning this in comments. So this is more like mid-high priority right?

common/task_executor/src/rayon_manager.rs
```diff
@@ -0,0 +1,55 @@
...
+                .expect("failed to build high-priority rayon pool"),
+        );
+        Self {
+            low_priority_thread_pool,
+            high_priority_thread_pool,
+        }
+    }
+}
+
+impl RayonManager {
```
jimmygchen: minor: this struct isn't really doing any managing, and is more like a `RayonThreadPoolProvider`,  unless you have something else planned for this?

jimmygchen (changes requested):
Looks great! I've added some comments.
@michaelsproul I think another review on this would be useful if you have sometime 🙏

common/task_executor/src/lib.rs
```diff
@@ -226,6 +233,49 @@ impl TaskExecutor {
...
+            name,
+        ) {
+            self.spawn_monitor(task_handle, name);
+        }
+
+        match result.lock() {
+            Ok(mut guard) => guard.take(),
+            Err(e) => {
+                error!(error=?e, task_name=name, "Rayon pool lock is poisoned");
+                None
```
jimmygchen: I think we could return `Result` here to be more explicit about a failure here?

common/task_executor/src/lib.rs
```diff
@@ -226,6 +233,49 @@ impl TaskExecutor {
...
+    where
+        F: FnOnce() -> R + Send + 'static,
+        R: Send + 'static,
+    {
+        let thread_pool = self.rayon_manager.get_thread_pool(rayon_pool_type);
+
+        let result = Arc::new(Mutex::new(None));
+        let result_clone = result.clone();
+
+        if let Some(task_handle) = self.spawn_blocking_handle(
```
jimmygchen: I think we now have a tokio blocking thread sitting idle waiting for the rayon threads to complete? not sure if there's a better way though, i can think about it a bit more tomorrow

beacon_node/beacon_processor/src/lib.rs
eserilev: I guess it depends what the goal of the `disable-backfill-rate-limiting` flag is. If we want the fastest possible backfilling, 100% cpu usage (i.e. the global rayon threadpool) will do that. This comes at the cost of potentially degraded node performance which the flag clearly warns about.

common/task_executor/src/lib.rs
eserilev: b30e4ad 

I had to introduce a new error enum, because `std::sync::PoisonError` takes a reference of the mutex which is dropped at the end of the function.

common/task_executor/src/lib.rs
jimmygchen: This is a non issue I think, and is what Tokio blocking tasks are designed for

eserilev: I think we could do something like this

```rust
    pub async fn spawn_rayon_async<F, R>(
        &self,
        rayon_pool_type: RayonPoolType,
        task: F,
    ) -> Result<R, SomeError>
    where
        F: FnOnce() -> R + Send + 'static,
        R: Send + 'static,
    {
        let thread_pool = self.rayon_manager.get_thread_pool(rayon_pool_type);
        let (tx, rx) = tokio::sync::oneshot::channel();

        thread_pool.scope(|s| {
            s.spawn(|_| {
                let result = task();
                let _ = tx.send(result);
            });
        });

        rx.await.map_err(|_| TaskExecutorError::SomeError)
    }
```

eserilev: I think we could do something like

```rust
    pub async fn spawn_rayon_async<F, R>(
        &self,
        rayon_pool_type: RayonPoolType,
        task: F,
    ) -> Result<R, tokio::sync::oneshot::error::RecvError>
    where
        F: FnOnce() -> R + Send + 'static,
        R: Send + 'static,
    {
        let thread_pool = self.rayon_manager.get_thread_pool(rayon_pool_type);
        let (tx, rx) = tokio::sync::oneshot::channel();

        thread_pool.spawn(move || {
            let result = task();
            let _ = tx.send(result);
        });

        rx.await
    }
```

eserilev: Were only using rayon threads w/ this approach and it just seems simpler in general. I'm thinking this is better, what do you think?

jimmygchen: Yeah this looks better  and simpler without mutexes

beacon_node/beacon_processor/src/lib.rs
jimmygchen: yeah good point!

common/task_executor/src/lib.rs
jimmygchen: Thanks, however i think this is no longer needed if we go with. your suggested async + no mutex approach

common/task_executor/src/rayon_manager.rs
jimmygchen: not asking for a change - this is just personal preference and i don't mind keeping it as it is.

jimmygchen (changes requested):
Hey @eserilev I think your suggested approach looks good and i don't see any issue with it.

jimmygchen:
Updating this to `v8.0.0-rc.0` as I think it's important to have this in the testnet release, so reconstruction doesn't impact supernode block processing and attesting.

common/task_executor/src/rayon_manager.rs
eserilev: I agree with this sentiment

common/task_executor/src/rayon_pool_provider.rs
eserilev: I added some comments mentioning that critical/highest priority tasks should be run using the global thread pool

jimmygchen (approved):
Looking good, thanks! Will wait for another reviewer and some testing and then merge this.

I think it might be worth clarifying on the below functions, i.e. the difference isnt just blocking /async, we'd use the 2nd one of if we want to block the async thread to await for the return values (maybe fn naming + describe when to use each?)
- `spawn_blocking_with_rayon`
- `spawn_rayon_async`

eserilev:
Reconstruction inside the scoped high prio rayon pool (while setting reconstruction delay to 0)
tracing

tracing

eserilev:
When using the global rayon pool and setting reconstruction delay to 0

tracing
tracing

eserilev:
In the scoped high prio rayon pool with reconstruction delay set to 0 

over a one hour period

tree-hash-cache 90% of these were under 3ms
tracing

blobs_to_data_column_sidecars more than 90% were under 30ms
tracing

jimmygchen (approved):
Nice numbers 👏  pretty good improvements there, shows that the scoped rayon impl is quite effective.

PR #8082, opened 2025-09-18
Always upload sim test logs

jimmygchen:
## Issue Addressed

This CI job failed

https://github.com/sigp/lighthouse/actions/runs/17815533375/job/50647915897

But we lost the logs because they aren't uploaded when the job fails. This PR changes the step to always upload job, even in the case of failure.


antondlr (approved):
yup that should do it

PR #8083, opened 2025-09-18
Rename `--subscribe-all-data-column-subnets` to `--supernode` and make it visible in help

jimmygchen:
## Issue Addressed

Rename `--subscribe-all-data-column-subnets` to `--supernode` as it's now been officially accepted in the spec. Also make it visible in help in preparation for the fusaka release.

https://github.com/ethereum/consensus-specs/blob/dev/specs/fulu/p2p-interface.md#supernodes

### DIFF
--- a/lighthouse/tests/beacon_node.rs
+++ b/lighthouse/tests/beacon_node.rs
@@ -797,6 +797,13 @@ fn network_target_peers_flag() {
 #[test]
 fn network_subscribe_all_data_column_subnets_flag() {
     CommandLineTest::new()
+        .flag("subscribe-all-data-column-subnets", None)
+        .run_with_zero_port()
+        .with_config(|config| assert!(config.network.subscribe_all_data_column_subnets));
+}
+#[test]
+fn network_supernode_flag() {
+    CommandLineTest::new()
         .flag("supernode", None)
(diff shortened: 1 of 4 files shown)


eserilev (approved):

pawanjay176 (approved):
LGTM. Would be good to get a ✅ from @barnabasbusa as well to make sure we aren't breaking any setups unexpectedly.

lighthouse/tests/beacon_node.rs
```diff
@@ -797,11 +797,17 @@ fn network_target_peers_flag() {
 #[test]
 fn network_subscribe_all_data_column_subnets_flag() {
     CommandLineTest::new()
-        .flag("subscribe-all-data-column-subnets", None)
+        .flag("supernode", None)
```
michaelsproul, line 807: Should maybe keep the test for the old flag name too

jimmygchen: oh yeah i'll add a test for that.
Also i think its worth adding help text to mention why people would want to run supernode - i.e. get direct access to blobs via beacon API

jimmygchen:
@mergify requeue

PR #8094, opened 2025-09-19
Use HTTPS for xdelta3 in Cargo.toml

aviggiano:
## Issue Addressed

No issue

## Proposed Changes

Use HTTPS for dependency

## Additional Info

N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11165,7 +11165,7 @@ dependencies = [
 [[package]]
 name = "xdelta3"
 version = "0.1.5"
-source = "git+http://github.com/sigp/xdelta3-rs?rev=4db64086bb02e9febb584ba93b9d16bb2ae3825a#4db64086bb02e9febb584ba93b9d16bb2ae3825a"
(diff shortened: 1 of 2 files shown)


michaelsproul:
Thanks @aviggiano!

Not sure why we were using HTTP, that's a great catch.

Would you mind rebasing your changes on `unstable` and signing the CLA (see bot comment above mine)? I'll update the target branch, but GH won't pass CI until the branch is actually rebased.

Thanks

hopinheimer:
do you think we should add a CI job to check this just in case?

michaelsproul:
Yeah that would be good, there might be an existing cargo addon for this?

aviggiano:
yeah I caught this while trying to build lighthouse locally and it failed after I changed something else

I think a CI check would be nice

michaelsproul (approved):
LGTM, thanks

jimmygchen (dismissed):
Looks like the cargo lock check is failing and the lock file needs an update too.

https://github.com/sigp/lighthouse/actions/runs/17913557528/job/50941650269

michaelsproul (approved):

Issue #6293, opened 2024-08-22
Refactor validator monitor missed block tests


michaelsproul:
## Description

The validator monitor missed block tests are quite fragile and depend on hardcoded validator indices which break with each hard fork.

We should rewrite these tests to dynamically use correct validator indices.

The tests are also complicated by the "simulation" of missed blocks through the mutatation of the `block_roots` array. It would be simpler if they just avoided creating blocks at those slots using `extend_chain`.

## Version

Lighthouse v5.3.0

## Steps to resolve

Refactor the validator monitor tests, e.g. `produces_missed_blocks` so they don't use hardcoded validator indices or manual block_roots mutations.

PR #8101, opened 2025-09-22
Fix bugs in proposer calculation post-Fulu

michaelsproul:
## Issue Addressed

As identified by a researcher during the Fusaka security competition, we were computing the proposer index incorrectly in some places by computing without lookahead.

## Proposed Changes

- [x] Add "low level" checks to computation functions in `consensus/types` to ensure they error cleanly
- [x] Re-work the determination of proposer shuffling decision roots, which are now fork aware.
- [x] Re-work and simplify the beacon proposer cache to be fork-aware.
- [x] Optimise `with_proposer_cache` to use `OnceCell`.
- [x] All tests passing.
- [x] Resolve all remaining `FIXME(sproul)`s.
- [x] Unit tests for `ProtoBlock::proposer_shuffling_root_for_child_block`.
- [x] End-to-end regression test.
- [x] Test on pre-Fulu network.
- [x] Test on post-Fulu network.

## Testing

- The correctness of `ProtoBlock::proposer_shuffling_root_for_child_block` is tested by new tests added in e4713156e1c1013cf720540e8125794362588d99. They catch the bug that was fixed by 83128c40ef5d10b04906c0ad20fad23e69e57117.

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -6,8 +6,7 @@ use crate::attestation_verification::{
 use crate::attester_cache::{AttesterCache, AttesterCacheKey};
 use crate::beacon_block_streamer::{BeaconBlockStreamer, CheckCaches};
 use crate::beacon_proposer_cache::{
-    BeaconProposerCache, compute_proposer_duties_from_head,
-    ensure_state_can_determine_proposers_for_epoch,
+    BeaconProposerCache, ensure_state_can_determine_proposers_for_epoch,
 };
 use crate::blob_verification::{GossipBlobError, GossipVerifiedBlob};
 use crate::block_times_cache::BlockTimesCache;
@@ -4705,57 +4704,50 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             .snapshot
             .beacon_state
             .proposer_shuffling_decision_root_at_epoch(proposal_epoch, proposer_head, &self.spec)?;
-        let cached_proposer = self
-            .beacon_proposer_cache
-            .lock()
-            .get_slot::<T::EthSpec>(shuffling_decision_root, proposal_slot);
-        let proposer_index = if let Some(proposer) = cached_proposer {
-            proposer.index as u64
-        } else {
-            if head_epoch + self.config.sync_tolerance_epochs < proposal_epoch {
-                warn!(
-                    msg = "this is a non-critical issue that can happen on unhealthy nodes or \
-                              networks.",
-                    %proposal_epoch,
-                    %head_epoch,
-                    "Skipping proposer preparation"
-                );
 
-                // Don't skip the head forward more than two epochs. This avoids burdening an
-                // unhealthy node.
-                //
-                // Although this node might miss out on preparing for a proposal, they should still
-                // be able to propose. This will prioritise beacon chain health over efficient
-                // packing of execution blocks.
-                return Ok(None);
-            }
+        let Some(proposer_index) = self.with_proposer_cache(
+            shuffling_decision_root,
+            proposal_epoch,
+            |cache| cache.get_slot::<T::EthSpec>(shuffling_decision_root, proposal_slot)
+                         .map(|p| p.index as u64),
+            || {
+                if head_epoch + self.config.sync_tolerance_epochs < proposal_epoch {
+                    warn!(
+                        msg = "this is a non-critical issue that can happen on unhealthy nodes or \
+                               networks",
+                        %proposal_epoch,
+                        %head_epoch,
+                        "Skipping proposer preparation"
+                    );
 
-            let (proposers, decision_root, _, fork) =
-                compute_proposer_duties_from_head(proposal_epoch, self)?;
-
-            let proposer_offset = (proposal_slot % T::EthSpec::slots_per_epoch()).as_usize();
-            let proposer = *proposers
-                .get(proposer_offset)
-                .ok_or(BeaconChainError::NoProposerForSlot(proposal_slot))?;
-
-            self.beacon_proposer_cache.lock().insert(
-                proposal_epoch,
-                decision_root,
-                proposers,
-                fork,
-            )?;
-
-            // It's possible that the head changes whilst computing these duties. If so, abandon
-            // this routine since the change of head would have also spawned another instance of
-            // this routine.
-            //
-            // Exit now, after updating the cache.
-            if decision_root != shuffling_decision_root {
+                    // Don't skip the head forward too many epochs. This avoids burdening an
+                    // unhealthy node.
+                    //
+                    // Although this node might miss out on preparing for a proposal, they should
+                    // still be able to propose. This will prioritise beacon chain health over
+                    // efficient packing of execution blocks.
+                    Ok(None)
+                } else {
+                    let head = self.canonical_head.cached_head();
+                    Ok(Some((
+                        head.head_state_root(),
+                        head.snapshot.beacon_state.clone(),
+                    )))
+                }
+            },
+        ).or_else(|e| {
+            if let Error::ProposerCacheIncorrectState { .. } = e {
+                // It's possible that the head changes whilst computing these duties. If so, abandon
+                // this routine since the change of head would have also spawned another instance of
+                // this routine.
                 warn!("Head changed during proposer preparation");
-                return Ok(None);
+                Ok(None)
+            } else {
+                Err(e)
             }
-
-            proposer as u64
+        })? else {
+            warn!("Unable to determine proposer for proposer preparation");
+            return Ok(None);
         };
 
         // Get the `prev_randao` and parent block number.
@@ -6561,11 +6553,11 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         &self,
         shuffling_decision_block: Hash256,
         proposal_epoch: Epoch,
-        accessor: impl Fn(&mut BeaconProposerCache, Hash256, Epoch) -> Option<V>,
+        accessor: impl Fn(&mut BeaconProposerCache) -> Option<V>,
         state_provider: impl FnOnce() -> Result<Option<(Hash256, BeaconState<T::EthSpec>)>, Error>,
     ) -> Result<Option<V>, Error> {
         let mut cache = self.beacon_proposer_cache.lock();
-        if let Some(value) = accessor(&mut cache, shuffling_decision_block, proposal_epoch) {
+        if let Some(value) = accessor(&mut cache) {
             return Ok(Some(value));
         }
         drop(cache);
@@ -6610,7 +6602,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             state.fork(),
         )?;
 
-        if let Some(value) = accessor(&mut cache, shuffling_decision_block, proposal_epoch) {
+        if let Some(value) = accessor(&mut cache) {
             Ok(Some(value))
         } else {
             Ok(None)
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -6,7 +6,7 @@ use crate::attestation_verification::{
 use crate::attester_cache::{AttesterCache, AttesterCacheKey};
 use crate::beacon_block_streamer::{BeaconBlockStreamer, CheckCaches};
 use crate::beacon_proposer_cache::{
-    BeaconProposerCache, ensure_state_can_determine_proposers_for_epoch,
+    BeaconProposerCache, EpochBlockProposers, ensure_state_can_determine_proposers_for_epoch,
 };
 use crate::blob_verification::{GossipBlobError, GossipVerifiedBlob};
 use crate::block_times_cache::BlockTimesCache;
@@ -4708,8 +4708,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         let Some(proposer_index) = self.with_proposer_cache(
             shuffling_decision_root,
             proposal_epoch,
-            |cache| cache.get_slot::<T::EthSpec>(shuffling_decision_root, proposal_slot)
-                         .map(|p| p.index as u64),
+            |proposers| proposers.get_slot::<T::EthSpec>(proposal_slot).map(|p| p.index as u64),
             || {
                 if head_epoch + self.config.sync_tolerance_epochs < proposal_epoch {
                     warn!(
@@ -4726,27 +4725,28 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
                     // Although this node might miss out on preparing for a proposal, they should
                     // still be able to propose. This will prioritise beacon chain health over
                     // efficient packing of execution blocks.
-                    Ok(None)
+                    Err(Error::SkipProposerPreparation)
                 } else {
                     let head = self.canonical_head.cached_head();
-                    Ok(Some((
+                    Ok((
                         head.head_state_root(),
                         head.snapshot.beacon_state.clone(),
-                    )))
+                    ))
                 }
             },
-        ).or_else(|e| {
-            if let Error::ProposerCacheIncorrectState { .. } = e {
-                // It's possible that the head changes whilst computing these duties. If so, abandon
-                // this routine since the change of head would have also spawned another instance of
-                // this routine.
-                warn!("Head changed during proposer preparation");
-                Ok(None)
-            } else {
-                Err(e)
+        ).map_or_else(|e| {
+            match e {
+                Error::ProposerCacheIncorrectState { .. } => {
+                    warn!("Head changed during proposer preparation");
+                    Ok(None)
+                }
+                Error::SkipProposerPreparation => {
+                    // Warning logged for this above.
+                    Ok(None)
+                }
+                e => Err(e)
             }
-        })? else {
-            warn!("Unable to determine proposer for proposer preparation");
+        }, |value| Ok(Some(value)))? else {
             return Ok(None);
         };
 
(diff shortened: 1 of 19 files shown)


dapplion:
I looked at this with @michaelsproul today and the approach makes sense. I need a few more passes on my own to fully get the changes.

consensus/types/src/beacon_state.rs
```diff
@@ -998,6 +1012,45 @@ impl<E: EthSpec> BeaconState<E> {
...
+            // would be very dangerous as it would lead to conflicts between the *true* proposer as
+            // defined by `self.proposer_lookahead` and the output of this function.
+            // With MIN_SEED_LOOKAHEAD=1 (common config), this is equivalent to checking that the
+            // requested epoch is not the current epoch.
+            //
+            // We do not run this check if this function is called from `upgrade_to_fulu`,
+            // which runs *after* the slot is incremented, and needs to compute the proposer
+            // shuffling for the epoch that was just transitioned into.
+            if self.fork_name_unchecked().fulu_enabled()
+                && epoch < current_epoch.safe_add(spec.min_seed_lookahead)?
```
michaelsproul, line 1035: @jtraglia This is the exception to the assert I mentioned, which allows us to compute the proposer indices with 0 lookahead _only_ in the case of `upgrade_to_fulu` (when the state itself has not yet become the `Fulu` variant).

consensus/types/src/beacon_state.rs
```diff
@@ -1168,6 +1218,25 @@ impl<E: EthSpec> BeaconState<E> {
...
+                0
+            } else {
+                slots_per_epoch
+            };
+            return Ok(proposer_lookahead
+                .iter_from(start_offset)?
+                .take(slots_per_epoch)
+                .map(|x| *x as usize)
+                .collect());
+        }
```
michaelsproul, line 1238: @jtraglia Similarly, this block only comes into effect outside of the `upgrade_to_fulu` logic.

consensus/state_processing/src/epoch_cache.rs
```diff
@@ -123,7 +123,7 @@ pub fn is_epoch_cache_initialized<E: EthSpec>(
     let current_epoch = state.current_epoch();
     let epoch_cache: &EpochCache = state.epoch_cache();
     let decision_block_root = state
-        .proposer_shuffling_decision_root(Hash256::zero())
+        .epoch_cache_decision_root(Hash256::zero())
```
michaelsproul, line 126: The epoch cache is keyed on the last block of the previous epoch. Prior to Fulu this was the same as the proposer shuffling decision root, however that is no longer the case. So we use a separate function.

consensus/proto_array/src/proto_array_fork_choice.rs
```diff
@@ -160,6 +160,50 @@ pub struct Block {
     pub unrealized_finalized_checkpoint: Option<Checkpoint>,
 }
 
+impl Block {
+    /// Compute the proposer shuffling decision root of a child block in `child_block_epoch`.
+    ///
+    /// This function assumes that `child_block_epoch >= self.epoch`. It is the responsibility of
+    /// the caller to check this condition, or else incorrect results will be produced.
+    pub fn proposer_shuffling_root_for_child_block(
```
michaelsproul, line 168: This logic is a bit subtle. It was extracted from the 2-3 places where we used it previously (block/blob/data column verification).

It relates the proposer shuffling to the attester shufflings. The relationship is slightly different depending on whether Fulu is enabled. Prior to Fulu everything is shifted by an epoch, and then post Fulu things line up (proposer shuffling and attester shuffling for epoch N have the same decision block root).

beacon_node/beacon_chain/tests/validator_monitor.rs
```diff
@@ -234,21 +234,24 @@ async fn missed_blocks_basic() {
...
+    );
+
+    let decision_block_root = _state2
+        .proposer_shuffling_decision_root_at_epoch(epoch, Hash256::ZERO, &harness2.chain.spec)
+        .unwrap();
     assert_eq!(
         beacon_proposer_cache.lock().insert(
             epoch,
-            duplicate_block_root,
+            decision_block_root,
```
michaelsproul, line 248: Previously this test assumed that the `duplicate_block_root` was the decision root, but that is no longer true post Fulu.

beacon_node/beacon_chain/src/canonical_head.rs
```diff
@@ -829,7 +829,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         let head_slot = new_snapshot.beacon_state.slot();
         let dependent_root = new_snapshot
             .beacon_state
-            .proposer_shuffling_decision_root(self.genesis_block_root);
+            .attester_shuffling_decision_root(self.genesis_block_root, RelativeEpoch::Next);
```
michaelsproul, line 832: These used to be equivalent, but are no longer. This dependent root is meant to be the attestation shuffling for the next epoch, so it makes a lot more sense to use that explicitly!

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -948,61 +948,35 @@ impl<T: BeaconChainTypes> GossipVerifiedBlock<T> {
...
-
-            (proposer_index, state.fork(), Some(parent), block)
-        };
+        let block_slot = block.slot();
+        let mut opt_parent = None;
+        let proposer = chain.with_proposer_cache::<_, BlockError>(
+            proposer_shuffling_decision_block,
+            block_epoch,
+            |proposers| proposers.get_slot::<T::EthSpec>(block_slot),
+            || {
```
michaelsproul, line 961: I think we should putt back the same log message as before, and remove the cache miss log from `with_proposer_cache`, which has less context.

consensus/types/src/beacon_state.rs
jtraglia: I see, this makes sense. I think these checks (this one & the one mentioned in the comment below) might be a little too complex to add to the specs. I still prefer just removing these two helper functions sometime in the future. I'll try to implement these in the specs locally just to get a feel for what it would look like though.

beacon_node/beacon_chain/src/beacon_proposer_cache.rs
```diff
@@ -178,44 +187,60 @@ pub fn compute_proposer_duties_from_head<T: BeaconChainTypes>(
...
     Ok((indices, dependent_root, execution_status, state.fork()))
 }
 
-/// If required, advance `state` to `target_epoch`.
+/// If required, advance `state` to the epoch required to determine proposer indices in `target_epoch`.
 ///
 /// ## Details
 ///
 /// - Returns an error if `state.current_epoch() > target_epoch`.
 /// - No-op if `state.current_epoch() == target_epoch`.
```
jimmygchen, line 214: I think the comment here may need to be updated?

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -6559,6 +6554,70 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         }
     }
 
+    pub fn with_proposer_cache<V, E: From<BeaconChainError> + From<BeaconStateError>>(
```
jimmygchen, line 6557: could use some docs here

beacon_node/http_api/src/proposer_duties.rs
```diff
@@ -105,36 +106,29 @@ fn try_proposer_duties_from_cache<T: BeaconChainTypes>(
...
-                head_epoch, request_epoch
-            )));
-        }
-    };
+    // This code path can't handle requests for past epochs.
+    if head_epoch > request_epoch {
+        return Err(warp_utils::reject::custom_server_error(format!(
+            "head epoch {head_epoch} is later than request epoch {request_epoch}",
+        )));
+    }
```
jimmygchen, line 120: nit: can we lift this check earlier so we can skip some of the extra reads?

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -948,61 +948,35 @@ impl<T: BeaconChainTypes> GossipVerifiedBlock<T> {
...
-                parent_block
-                    .next_epoch_shuffling_id
-                    .shuffling_decision_block
-            } else {
-                parent_block.root
-            };
+            parent_block.proposer_shuffling_root_for_child_block(block_epoch, &chain.spec);
 
         // We assign to a variable instead of using `if let Some` directly to ensure we drop the
         // write lock before trying to acquire it again in the `else` clause.
```
jimmygchen, line 954: I think the comment above is no longer relevant here, now we've moved the lock management into `with_proposer_cache`

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -1191,6 +1191,271 @@ fn check_shuffling_compatible(
     }
 }
 
+/// These tests check the consistency of:
+///
+/// - ProtoBlock::proposer_shuffling_root_for_child_block, and
+/// - BeaconState::proposer_shuffling_decision_root{_at_epoch}
+async fn proposer_shuffling_root_consistency_test(parent_slot: u64, child_slot: u64) {
```
jimmygchen, line 1198: a little strange that these tests are under `store_tests.rs`, i guess it's more for convenience?

beacon_node/beacon_chain/tests/validator_monitor.rs
```diff
@@ -234,21 +234,24 @@ async fn missed_blocks_basic() {
 
     // Let's fill the cache with the proposers for the current epoch
     // and push the duplicate_block_root to the block_roots vector
+    assert_eq!(
+        _state2.set_block_root(prev_slot, duplicate_block_root),
+        Ok(())
+    );
+
+    let decision_block_root = _state2
```
jimmygchen, line 242: this can be renamed to `_state` now it's being used.

beacon_node/beacon_chain/tests/validator_monitor.rs
jimmygchen: Is this the case for this one too?
https://github.com/sigp/lighthouse/blob/9b4e4ebd721e9fb4bfaa4a8f7285d2173df29b66/beacon_node/beacon_chain/tests/validator_monitor.rs#L155

beacon_node/beacon_chain/tests/store_tests.rs
michaelsproul: I just write everything under store tests haha

jimmygchen:
We have this deployed to both live testnets and `fusaka-devnet-3` for a day now and I have not noticed any issues. I'm going to tick the testing boxes. But I'm going to keep trying to reproduce the attack on local devnets.

michaelsproul:
I think the attack on a live network might require a proposer cache miss. So potentially multiple skipped slots near an epoch boundary, plus the effective balance changes

consensus/proto_array/src/proto_array_fork_choice.rs
jimmygchen: nice refactoring, much better keeping this kind of logic in one place

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen, line 6557: This forces the caller to implement `From` for both error types, it feels a bit strange but given this works with existing code, i think it's fine to leave it as it is.

jimmygchen: i think we could be more explicit about why we're using this `assesor` pattern (it's probably a common pattern we use for lock management but may be worth mentioning)

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -641,65 +639,34 @@ fn verify_proposer_and_signature<T: BeaconChainTypes>(
...
-    let proposer_index = *epoch_proposers
-        .proposers
-        .get(column_slot.as_usize() % slots_per_epoch as usize)
-        .ok_or_else(|| BeaconChainError::NoProposerForSlot(column_slot))?;
-
-    let fork = epoch_proposers.fork;
+    let proposer_shuffling_root =
+        parent_block.proposer_shuffling_root_for_child_block(column_epoch, &chain.spec);
+
+    let proposer = chain.with_proposer_cache(
```
jimmygchen, line 645: nice, we also avoid building committee caches unnecessarily with `cheap_state_advance_to_obtain_committees` here and use `ensure_state_can_determine_proposers_for_epoch` instead

beacon_node/beacon_chain/src/beacon_proposer_cache.rs
```diff
@@ -178,44 +187,60 @@ pub fn compute_proposer_duties_from_head<T: BeaconChainTypes>(
...
 /// - It must be the case that `state.canonical_root() == state_root`, but this function will not
 ///   check that.
-pub fn ensure_state_is_in_epoch<E: EthSpec>(
+pub fn ensure_state_can_determine_proposers_for_epoch<E: EthSpec>(
```
jimmygchen, line 217: may be worth adding instrumenting tracing for this function (we used to call `cheap_state_advance_to_obtain_committees` which shows up in the gossip processing traces)

consensus/types/src/chain_spec.rs
```diff
@@ -865,6 +865,28 @@ impl ChainSpec {
         )
     }
 
+    /// Returns the slot at which the proposer shuffling was decided.
+    ///
+    /// The block root at this slot can be used to key the proposer shuffling for the given epoch.
+    pub fn proposer_shuffling_decision_slot<E: EthSpec>(&self, epoch: Epoch) -> Slot {
```
jimmygchen, line 871: At `epoch == fulu_fork_epoch`, I think this function returns incorrect decision slot, as the epochs proposer shuffling decision slot should be the slot at the end of epoch N - 1 right?

jimmygchen: ~~I _think_ this may not be a real issue since we'd be using `state.proposer_lookahead()` (initialised in `upgrade_to_fulu`) for proposer indices from the fulu fork epoch~~

jimmygchen (approved):
This is quite a difficult PR to review, as it covers quite a lot of code paths and edge cases, but after a few passes, i think it looks very solid! Impressed with how thorough you are with this one. I think you've handled all the code paths I can find and can't really spot any issues with correctness. 

I think this is good to merge and we can address the comments next week.

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -1191,6 +1191,271 @@ fn check_shuffling_compatible(
...
+        decision_root,
+        child_block_state
+            .proposer_shuffling_decision_root_at_epoch(child_block_epoch, Hash256::ZERO, spec)
+            .unwrap()
+    );
+}
+
+#[tokio::test]
+async fn proposer_shuffling_root_consistency_same_epoch() {
+    proposer_shuffling_root_consistency_test(32, 39).await;
```
dapplion, line 1271: Can do in another PR, but writing the numbers like will help understand them
```suggestion
    proposer_shuffling_root_consistency_test(
      E::slots_per_epoch().as_u64() * 4,
      E::slots_per_epoch().as_u64() * 5 - 1,
    ).await;
```

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -1191,6 +1191,271 @@ fn check_shuffling_compatible(
...
+    proposer_shuffling_root_consistency_test(32, 47).await;
+}
+
+#[tokio::test]
+async fn proposer_shuffling_root_consistency_two_epochs() {
+    proposer_shuffling_root_consistency_test(32, 55).await;
+}
+
+#[tokio::test]
+async fn proposer_shuffling_changing_with_lookahead() {
```
dapplion, line 1285: Nice test 💯

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -1191,6 +1191,271 @@ fn check_shuffling_compatible(
...
+        .clone();
+
+    // Craft a block with a deposit request and consolidation.
+    // XXX: This is a really nasty way to do this, but we need better test facilities in
+    // MockExecutionLayer to address this.
+    let deposit_request: DepositRequest = DepositRequest {
+        index: pre_deposit_state.eth1_deposit_index(),
+        pubkey: validator_to_topup.pubkey,
+        withdrawal_credentials: validator_to_topup.withdrawal_credentials,
+        amount: 63_000_000_000,
```
dapplion, line 1327: I assume you chose this value to find a balance that will trigger a change in the proposer selection

consensus/types/src/chain_spec.rs
```diff
@@ -865,6 +865,28 @@ impl ChainSpec {
...
+            epoch
+                .saturating_sub(self.min_seed_lookahead)
+                .start_slot(E::slots_per_epoch())
+                .saturating_sub(1_u64)
+        } else {
+            // Pre-Fulu the proposer shuffling decision slot for epoch N is the slot at the end of
+            // epoch N - 1 (note: +1 -1 for min_seed_lookahead=1 in all current configs).
+            epoch
+                .saturating_add(Epoch::new(1))
+                .saturating_sub(self.min_seed_lookahead)
```
dapplion, line 884: Why +1 - min_seed_lookahead instead of removing the `min_seed_lookahead` substraction?

consensus/types/src/chain_spec.rs
```diff
@@ -865,6 +865,28 @@ impl ChainSpec {
         )
     }
 
+    /// Returns the slot at which the proposer shuffling was decided.
+    ///
+    /// The block root at this slot can be used to key the proposer shuffling for the given epoch.
+    pub fn proposer_shuffling_decision_slot<E: EthSpec>(&self, epoch: Epoch) -> Slot {
+        if self.fork_name_at_epoch(epoch).fulu_enabled() {
```
dapplion, line 872: Should this change be 
```suggestion
        if self.fork_name_at_epoch(epoch + 1).fulu_enabled() {
```
? The first epoch of the fulu fork the proposers are computed at the transition which is decided by the last slot of the previous epoch

consensus/types/src/chain_spec.rs
jimmygchen: You've probably thought about this, but at the first slot of Fulu, I was wondering if the sanity check here would fail under some conditions

https://github.com/sigp/lighthouse/blob/4ceccd246386de82fafe59f783ef4b58da8d5c4f/beacon_node/beacon_chain/src/beacon_chain.rs#L6601-L6607


`proposer_shuffling_root_for_child_block` for the first fulu slot would return last slot at `N -1`
https://github.com/sigp/lighthouse/blob/6da576cdd917a46b98445f3cb7d0feb33334e0d8/consensus/proto_array/src/proto_array_fork_choice.rs#L194-L198

In the sanity check, `state.proposer_shuffling_decision_root_at_epoch` would return the root of last slot at `N-2`

https://github.com/sigp/lighthouse/blob/0ed80ff8398472a86d5704185ee8a15d34f94413/consensus/types/src/beacon_state.rs#L905-L907

consensus/types/src/chain_spec.rs
michaelsproul: So that we can be generic over min_seed_lookahead. If we remove it we are hardcoding min_seed_lookahead=1

beacon_node/beacon_chain/tests/store_tests.rs
michaelsproul: The churn limit is 64 ETH, so this is just the largest value that fits under that threshold and gets applied quickly. It also happens to change the shuffling with the default seed, and _I think_ should be likely to change it with such a small # of validators

consensus/types/src/chain_spec.rs
michaelsproul: I think `epoch - 1`, but yes. Working on a PR for this and other correctness fixes now.

consensus/types/src/chain_spec.rs
jimmygchen: Addressed in:

- https://github.com/sigp/lighthouse/pull/8128

consensus/types/src/chain_spec.rs
michaelsproul: - https://github.com/sigp/lighthouse/pull/8128

beacon_node/beacon_chain/tests/validator_monitor.rs
michaelsproul: This one is wrong, yeah, but doesn't actually do anything, because the `block_root` argument to `proposer_shuffling_decision_root` isn't used unless we're right on an epoch boundary (pre-Fulu), or there's a whole epoch of skipped slots (post Fulu).

I think I would prefer to completely rewrite these tests than tweaking them any more. There is an open issue for this:

- https://github.com/sigp/lighthouse/issues/6293

beacon_node/beacon_chain/tests/validator_monitor.rs
michaelsproul: These tests do this a lot already, I hate them :rofl:

Again, I reckon we just:

- https://github.com/sigp/lighthouse/issues/6293

PR #8102, opened 2025-09-22
Reduce `TARGET_BACKFILL_SLOTS` in checkpoint sync test

eserilev:
Our devnet checkpoint sync test is having a hard time finishing in under 10 minutes, I believe because of the higher blob counts on devnet3 (20-ish blobs). I tried disabling backfill rate limiting, but the test was still timing out. This PR reduces `TARGET_BACKFILL_SLOTS` from 1024 to 256 and disables backfill rate limiting so that the test doesn't time out.


jimmygchen (approved):
Looks good, thanks!

Issue #8085, opened 2025-09-19
Implement getBlobs endpoint


jimmygchen:
## Description

This spec was recently merged:
https://github.com/ethereum/beacon-APIs/pull/546

We should get this into our testnet releases so potential L2 users could consume this on testnets.

Thegaram:
Hi @jimmygchen @chong-he, I understand this did not make it into the testnet release. With PeerDAS going live, will `/eth/v1/beacon/blob_sidecars` stop working once Fusaka goes live on Sepolia?

pawanjay176:
`/eth/v1/beacon/blob_sidecars` should still work post fusaka. We however do not compute the correct blob kzg proof and instead stub it out with all zeroes. Would that work for your requirements since the getBlobs endpoint also doesn't actually return the kzg proof?

Thegaram:
Great, thank you! For us (Scroll), only the blob and the kzg commitment are needed, we do not use the kzg proof.

jimmygchen:
Hi @Thegaram could you describe how you use the kzg commitments please? Trying to figure out whether we should add it to the new endpoint

Thegaram:
We just use it to compute the `blobVersionedHash` [here](https://github.com/scroll-tech/go-ethereum/blob/06972688b320673254b3161b06df300b265b755b/rollup/da_syncer/blob_client/beacon_node_client.go#L150).

jimmygchen:
Oh right, so with the new `/eth/v1/beacon/ blobs` endpoint, you can now query the blob by version hashes - would you still require the commitments?

There's a thread on Eth R&D discord discussing about the requirement, if you're interested in participating:
https://discord.com/channels/595666850260713488/1423741917769891891/1424977894077366313

Thegaram:
Thank you! Actually I found this issue via that discussion, but [here I was told](https://github.com/ethereum/beacon-APIs/pull/546#discussion_r2297756644) that the new endpoint would still require the `block_id`.

In any case, if `/eth/v1/beacon/blob_sidecars` will continue to work for our use case, we'll continue to use it, otherwise it should be easy to switch to `/eth/v1/beacon/blobs`, assuming Lighthouse can still serve all recent blobs via this API.

jimmygchen:
> Thank you! Actually I found this issue via that discussion, but [here I was told](https://github.com/ethereum/beacon-APIs/pull/546#discussion_r2297756644) that the new endpoint would still require the `block_id`.

[...]

PR #8103, opened 2025-09-23
Implement `/eth/v1/beacon/blobs` endpoint

chong-he:
## Issue Addressed

* #8085 

## Proposed change

Rename current `blob_sidecars` endpoint from `get_blobs` to `get_blob_sidecars` to "give way" to the new get blobs endpoint. I thought this is the better way to have this new get_blobs endpoint name it this way in the code, but I am happy to change if that's not suitable/confusing

## Additional Info

Testing results:

Query with:
`curl -X 'GET' 'http://localhost:5052/eth/v1/beacon/blobs/12647385' | jq`

Response:

```
{
  "execution_optimistic": true,
  "finalized": true,
  "data": [
    "blob1"
    "blob2"
    "blob3" 
    "and so on"
  ]
}
```

Query with `versioned_hashes`:
`
curl -X 'GET' \
  'http://localhost:5052/eth/v1/beacon/blobs/12647385?versioned_hashes=0x01171d78b2bd7cf3c3e458f7337f37272bfe6972cdd07b6e12f399a94ff6627c&versioned_hashes=0x011384275182442b69831cc581de3b80fe60b13a8efd870f8b680422087b960e' \
  -H 'accept: application/json' | jq
`

Response:

```
{
  "execution_optimistic": true,
  "finalized": true,
  "data": [
      "blob1"
      "blob2"
  ]
}
```


beacon_node/http_api/src/block_id.rs
```diff
@@ -352,6 +353,62 @@ impl BlockId {
...
+        let block = BlockId::blinded_block_by_root(&root, chain)?.ok_or_else(|| {
+            warp_utils::reject::custom_not_found(format!("beacon block with root {}", root))
+        })?;
+
+        // Error if the block is pre-Deneb and lacks blobs.
+        let blob_kzg_commitments = block.message().body().blob_kzg_commitments().map_err(|_| {
+            warp_utils::reject::custom_bad_request(
+                "block is pre-Deneb and has no blobs".to_string(),
+            )
+        })?;
```
chong-he, line 375: This is reusing most of the code in the function `get_blinded_block_and_blob_list_filtered`: 
https://github.com/sigp/lighthouse/blob/3cb7e59be2ebcf66836dabae2c771b455822f654/beacon_node/http_api/src/block_id.rs#L313-L331

with `None` in indices and then convert the blob_sidecars to blobs. Another way to do this is to call `get_blinded_block_and_blob_list_filtered` to get the blob_sidecars (thus having less repetitive code), but I thought it would be better to call the "original" `get_blobs` function. Happy to edit this as fit

beacon_node/http_api/src/lib.rs
```diff
@@ -1950,6 +1951,58 @@ pub fn serve<T: BeaconChainTypes>(
...
+                                warp_utils::reject::custom_server_error(format!(
+                                    "failed to create response: {}",
+                                    e
+                                ))
+                            }),
+                        _ => {
+                            let hex_blobs: Vec<String> = blobs
+                                .iter()
+                                // Serialize the blob string as a hex string
+                                .map(|blob| format!("0x{}", hex::encode(blob.as_ref())))
```
chong-he: I do the conversion from String to Hex here in the JSON output

beacon_node/http_api/src/lib.rs
```diff
@@ -1950,6 +1951,58 @@ pub fn serve<T: BeaconChainTypes>(
...
+                                .collect();
+
+                            let res = execution_optimistic_finalized_beacon_response(
+                                ResponseIncludesVersion::No,
```
chong-he, line 1985: I see the example response should not have the "version" field, so I put a No here: https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Beacon/getBlobs

consensus/types/src/beacon_response.rs
```diff
@@ -25,6 +25,7 @@ pub struct ForkVersionedResponse<T, M = EmptyMetadata> {
 /// `Deserialize`.
 #[derive(Debug, PartialEq, Clone, Serialize)]
 pub struct UnversionedResponse<T, M = EmptyMetadata> {
+    #[serde(flatten)]
```
chong-he, line 28: Without this the response will have the `metadata` in the response and the "data" field is not in the same level as other fields:

```
{
  "metadata": {
    "finalized": false
  },
  "data": [
```

Adding this solves the issue and have the response to be the same format as in: https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Beacon/getBlobs

jimmygchen: I noticed this is used in the v1 version of two other v1 beacon API endpoints, which means these endpoints were responding with the incorrect format:
* [v1/beacon/blocks/{block_id}](https://github.com/sigp/lighthouse/blob/ee1b6bc81bb4b7dafb95a95eabcfcaf6b76cea04/beacon_node/http_api/src/lib.rs#L1733)
* [v1/beacon/blocks/{block_id}/attestations](https://github.com/sigp/lighthouse/blob/ee1b6bc81bb4b7dafb95a95eabcfcaf6b76cea04/beacon_node/http_api/src/lib.rs#L1807)

However, it looks like these endpoints have been deprecated and is no longer part of the [beacon API spec](https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#), i wonder if we should delete them too. (separate issue / PR of course)

testing/simulator/src/checks.rs
```diff
@@ -424,7 +424,7 @@ pub async fn verify_full_blob_production_up_to<E: EthSpec>(
         // the `verify_full_block_production_up_to` function.
         if block.is_some() {
             remote_node
-                .get_blobs::<E>(BlockId::Slot(Slot::new(slot)), None, &E::default_spec())
+                .get_blob_sidecars::<E>(BlockId::Slot(Slot::new(slot)), None, &E::default_spec())
```
jimmygchen: should we update this to test the new endpoint given we're deprecating the old one?

lcli/src/http_sync.rs
```diff
@@ -124,7 +124,7 @@ async fn get_block_from_source<T: EthSpec>(
             .unwrap()
             .unwrap();
         let blobs_from_source = source
-            .get_blobs::<T>(block_id, None, spec)
+            .get_blob_sidecars::<T>(block_id, None, spec)
```
jimmygchen, line 127: ah, the http sync feature will break after fulu, and will likely require a bit of work to get it working again (compute node's sampling columns, and download all columns via the [`getDebugDataColumnSidecars`](https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Debug/getDebugDataColumnSidecars) endpoint). This isn't high priority but thought it's worth pointing out. We could raise an issue for this so we don't forget.

common/eth2/src/lib.rs
```diff
@@ -1346,6 +1346,17 @@ impl BeaconNodeHttpClient {
         Ok(path)
     }
 
+    /// Path for `/v/beacon/blobs/{blob_id}`
+    pub fn get_blobs_path(&self, block_id: BlockId) -> Result<Url, Error> {
```
jimmygchen, line 1350: looks like this is unused in lighthouse, but makes sense to implement this in the crate, should we also create a `get_blobs` function?

beacon_node/http_api/tests/tests.rs
```diff
@@ -1956,7 +1956,7 @@ impl ApiTester {
 
         match self
             .client
-            .get_blobs::<E>(CoreBlockId::Slot(test_slot), None, &self.chain.spec)
+            .get_blob_sidecars::<E>(CoreBlockId::Slot(test_slot), None, &self.chain.spec)
```
jimmygchen, line 2030: Can we add some tests for `get_blobs`, some scenarios:
- pre deneb, a fullnode should return blobs
- from fulu, only supernode returns blobs
- from fulu, a fullnode should return an error

consensus/types/src/beacon_response.rs
jimmygchen: Could you add a round trip test for this as well? we have similar tests for `ForkVersionedResponse` in this file

beacon_node/http_api/src/lib.rs
```diff
@@ -1950,6 +1951,58 @@ pub fn serve<T: BeaconChainTypes>(
...
+                            .map_err(|e| {
+                                warp_utils::reject::custom_server_error(format!(
+                                    "failed to create response: {}",
+                                    e
+                                ))
+                            }),
+                        _ => {
+                            let hex_blobs: Vec<String> = blobs
+                                .iter()
+                                // Serialize the blob string as a hex string
```
jimmygchen: nit: blob bytes

beacon_node/http_api/src/block_id.rs
```diff
@@ -352,6 +353,62 @@ impl BlockId {
...
+
+        let max_blobs_per_block = chain.spec.max_blobs_per_block(block.epoch()) as usize;
+        let blob_sidecar_list = if !blob_kzg_commitments.is_empty() {
+            if chain.spec.is_peer_das_enabled_for_epoch(block.epoch()) {
+                Self::get_blobs_from_data_columns(chain, root, None, &block)?
```
jimmygchen: I think the performance of this is going to be quite bad with high blob count given we can't filter by indices, it means we have to load all data columns and reconstruct the entire data matrix, which could take a lot of time. (200ms per blob, with 20 blobs its' 4 seconds on a single thread)

I have a few suggestions:
- figure out the blobs indices first, by computing all versioned hashes from the `block.kzg_commitments`, and then use the indices to get only the required blob(s) (re-use existing `get_blobs_from_data_columns`)
- if we have all 128 columns, build the blobs without calling `recover_cells_and_compute_kzg_proofs`
- if we don't have all 128 columns, then recover the blobs using the above function, but use the `low_priority_thread_pool` instead of the global thread pool (as this may degrade beacon node performance briefly)

beacon_node/http_api/src/block_id.rs
```diff
@@ -352,6 +353,62 @@ impl BlockId {
...
+            } else {
+                //  // Use "None" for indices to get all blobs
+                Self::get_blobs(chain, root, None, max_blobs_per_block)?
```
jimmygchen: Same as above, would be useful to work out the indices first, so we don't load all blobs

beacon_node/http_api/src/block_id.rs
```diff
@@ -352,6 +353,62 @@ impl BlockId {
         Ok((block, blob_sidecar_list, execution_optimistic, finalized))
     }
 
+    #[allow(clippy::type_complexity)]
+    pub fn get_blobs_by_versioned_hashes<T: BeaconChainTypes>(
+        &self,
+        query: BlobsVersionedHashesQuery,
+        chain: &BeaconChain<T>,
+    ) -> Result<(Vec<Blob<T::EthSpec>>, ExecutionOptimistic, Finalized), warp::Rejection> {
```
jimmygchen: Since this is part of the http api crate, i think we could just return `UnversionedResponse<Vec<Blob>>` and don't need the clippy allow?

jimmygchen (changes requested):
Hey CK, thanks for the PR. I've added my thoughts, let me know what you think and if you have any questions!

beacon_node/http_api/src/block_id.rs
jimmygchen: yeah i think make sense to reuse what we have, we might have to tweak things a bit for performance though, see https://github.com/sigp/lighthouse/pull/8103#discussion_r2371266790

beacon_node/http_api/src/lib.rs
jimmygchen: yep, it was by design and discussed on the PR

jimmygchen:
btw very thorough with the spec compliance, i think the impl great in terms of correctness, and thanks for the test outputs. Nice work 👍

jimmygchen:
Bumping this to v8 mainnet release

beacon_node/beacon_chain/src/kzg_utils.rs
```diff
@@ -320,7 +320,7 @@ pub fn reconstruct_blobs<E: EthSpec>(
     };
 
     let blob_sidecars = blob_indices
-        .into_par_iter()
+        .into_iter()
```
jimmygchen, line 325: may be worth noting in the function doc that this function does NOT use rayon as this is primarily used by a non critical path in http api and will be slow if the node needs to reconstruct columns into blobs

beacon_node/http_api/src/block_id.rs
```diff
@@ -352,6 +354,87 @@ impl BlockId {
         Ok((block, blob_sidecar_list, execution_optimistic, finalized))
     }
 
+    #[allow(clippy::type_complexity)]
+    pub fn get_blobs_by_versioned_hashes<T: BeaconChainTypes>(
+        &self,
+        query: BlobsVersionedHashesQuery,
+        chain: &BeaconChain<T>,
+    ) -> Result<
+        UnversionedResponse<Vec<Blobs<T::EthSpec>>, ExecutionOptimisticFinalizedMetadata>,
```
jimmygchen: Should this be `Vec<Blob<..` instead of `Vec<Blobs<..`

common/eth2/src/lib.rs
chong-he: `get_blobs` function created, also used that in the test

lcli/src/http_sync.rs
chong-he: Is there any actionable item for this PR for now?

beacon_node/http_api/tests/tests.rs
chong-he: Added tests for: 
- pre Fulu, a fullnode should return blobs in `get_blob_sidecars()` - calling the `test_get_blobs` function
- post Fulu, a supernode is able to return blobs in `get_blobs_post_fulu_supernode()` - calling the `test_get_blobs` function
- post Fulu, a fullnode will error when asked for blobs in `get_blobs_post_fulu_full_node()`

beacon_node/beacon_chain/src/kzg_utils.rs
chong-he: Added a doc above the function

beacon_node/http_api/src/block_id.rs
chong-he: As discussed, change to `BlobWrapper` for a clearer struct name

beacon_node/http_api/src/block_id.rs
chong-he: I changed the output to UnversionedResponse, but I find that it still requires this: `#[allow(clippy::type_complexity)]` or the compiler will complain

lcli/src/http_sync.rs
jimmygchen: no, i'll create an issue tomorrow

chong-he:
> btw very thorough with the spec compliance, i think the impl great in terms of correctness, and thanks for the test outputs. Nice work 👍

Thanks so much for the thorough review. This PR is ready for another round of review.

Thank you both @michaelsproul and @jimmygchen

beacon_node/beacon_chain/src/kzg_utils.rs
```diff
@@ -337,12 +339,19 @@ pub fn reconstruct_blobs<E: EthSpec>(
...
-            let (cells, _kzg_proofs) = kzg
-                .recover_cells_and_compute_kzg_proofs(&cell_ids, &cells)
-                .map_err(|e| format!("Failed to recover cells and compute KZG proofs: {e:?}"))?;
+            let all_cells: Vec<_> = if data_columns.len() < E::number_of_columns() {
+                let (cells, _kzg_proofs) = kzg
+                    .recover_cells_and_compute_kzg_proofs(&cell_ids, &cells)
+                    .map_err(|e| {
+                        format!("Failed to recover cells and compute KZG proofs: {e:?}")
+                    })?;
+                cells.into_iter().collect()
```
jimmygchen: is it possible to avoid the `collect` into `Vec`  here? it doesn't seem necessary and we're just converting the vec back into `into_inter()` below.

beacon_node/client/src/builder.rs
```diff
@@ -413,7 +413,7 @@ where
                 let blobs = if block.message().body().has_blobs() {
                     debug!("Downloading finalized blobs");
                     if let Some(response) = remote
-                        .get_blobs::<E>(BlockId::Root(block_root), None, &spec)
+                        .get_blob_sidecars::<E>(BlockId::Root(block_root), None, &spec)
```
jimmygchen, line 415: This endpoint is being deprecated and will likely be removed soon - we can probably address this in a separate PR

beacon_node/http_api/src/block_id.rs
jimmygchen: ah right, yeah i's still quite nested. I think it's fine, it doesn't seem _too_ unreadable, given that we don't use this in multiple places. We _could_ add more aliases if we want, but it's not a big deal right now.

beacon_node/http_api/src/block_id.rs
```diff
@@ -352,6 +354,73 @@ impl BlockId {
...
+                .iter()
+                .flat_map(|versioned_hash| {
+                    blob_kzg_commitments.iter().position(move |commitment| {
+                        let computed_hash = kzg_commitment_to_versioned_hash(commitment);
+                        computed_hash == *versioned_hash
+                    })
+                })
+                .map(|index| index as u64)
+                .collect()
+        } else {
```
jimmygchen: I think we don't need this else statement here, because the `get_blobs_from_data_columns` and `get_blobs` functions accept `Option` and already handles this.

So this could just be simplified to:

```rust
        let blob_indices_opt = query.versioned_hashes.map(|versioned_hashes| {
            versioned_hashes
                .iter()
                .flat_map(|versioned_hash| {
                    blob_kzg_commitments.iter().position(move |commitment| {
                        let computed_hash = kzg_commitment_to_versioned_hash(commitment);
                        computed_hash == *versioned_hash
                    })
                })
                .map(|index| index as u64)
                .collect::<Vec<_>>()
        });
```

beacon_node/http_api/src/block_id.rs
```diff
@@ -352,6 +354,73 @@ impl BlockId {
...
+            }
+        } else {
+            BlobSidecarList::new(vec![], max_blobs_per_block)
+                .map_err(|e| warp_utils::reject::custom_server_error(format!("{:?}", e)))?
+        };
+
+        let blobs = blob_sidecar_list
+            .into_iter()
+            .map(|sidecar| BlobWrapper::<T::EthSpec> {
+                blob: sidecar.blob.clone(),
```
jimmygchen: Note that this is an expensive clone as `blob` isn't Arc'd.
However I don't see easy way around this unless we change the functions above to return `Blobs` instead.

beacon_node/http_api/src/block_id.rs
```diff
@@ -369,9 +438,9 @@ impl BlockId {
 
         let blob_sidecar_list_filtered = match indices {
             Some(vec) => {
-                let list: Vec<_> = blob_sidecar_list
+                let list: Vec<_> = vec
                     .into_iter()
-                    .filter(|blob_sidecar| vec.contains(&blob_sidecar.index))
+                    .flat_map(|index| blob_sidecar_list.get(index as usize).cloned())
```
jimmygchen: why do we need this change?

beacon_node/http_api/src/block_id.rs
jimmygchen: Just a note and no change required here

lcli/src/http_sync.rs
jimmygchen: Thinking about this again, it's a bit more complicated  because the `http-sync` feature does require full blobs to be published to the `publish_block` endpoint, which means post-fulu:
- `get_blob_sidecar` endpoint is deprecated, and cell kzg proofs are never returned from this endpoint, so publishing this won't work.
- `get_blobs` endpoint doesn't return KZG proofs, so it won't be sufficient to post to the publish block endpoint.

jimmygchen: Raised https://github.com/sigp/lighthouse/issues/8159.

beacon_node/client/src/builder.rs
jimmygchen: We're handling this gracefully at the moment so it's not a problem, i think we might eventually just remove this call and download columns from peers.

jimmygchen (changes requested):
Looks great, just a few minor comments and i think we can merge this soon.

beacon_node/http_api/src/block_id.rs
chong-he: Noted. 

If the function return `Blobs` (without the wrapper) then we don't have the nice feature of auto conversion from bytes to hex string :(

beacon_node/http_api/src/block_id.rs
chong-he: Before the change:

We iterate through the `blob_sidecar_list` and filters out when the indices are not contained in `vec` (the input indices). We have issues here in the test `test_get_blobs(true)` because the blobs created in the tests are identical: https://github.com/sigp/lighthouse/actions/runs/18151765807/job/51663901674?pr=8103
so say we have 2 blobs and 2 indices, the indices will be `[0, 0]`, and it gets filtered out in the code, giving an output of only 1 blob instead of 2, which fails the test

I think the relevant code for blob generator for test is here, where we can see it generates the same blobs: 
https://github.com/sigp/lighthouse/blob/e5b4983d6baf85770fe4539a565d8a2dd462bc53/beacon_node/execution_layer/src/test_utils/execution_block_generator.rs#L805-L831

After the change (by @michaelsproul):

We iterate through the index instead, and lookup the `blob_sidecar_list`, so as long as the `blob_sidecar_list` has the blob index, it will be returned. So even though the indices are the same, it will return the same repeated blobs

beacon_node/http_api/src/block_id.rs
chong-he: Great point, I have revised here: https://github.com/sigp/lighthouse/pull/8103/commits/434b3c58de65850770f1b0efc81eacfc44c4342e

beacon_node/beacon_chain/src/kzg_utils.rs
chong-he: I wrap them in `Box` to remove the `.collect()`, but I am not sure if it makes sense?

beacon_node/beacon_chain/src/kzg_utils.rs
```diff
@@ -337,13 +339,21 @@ pub fn reconstruct_blobs<E: EthSpec>(
...
-            let blob_bytes = cells
-                .into_iter()
+            let (all_cells, num_cells_original_blob): (Box<dyn Iterator<Item = _>>, usize) =
+                if data_columns.len() < E::number_of_columns() {
+                    let (cells, _kzg_proofs) = kzg
+                        .recover_cells_and_compute_kzg_proofs(&cell_ids, &cells)
+                        .map_err(|e| {
+                            format!("Failed to recover cells and compute KZG proofs: {e:?}")
+                        })?;
+                    let len = cells.len() / 2;
```
jimmygchen: `num_cells_original_blob` can just be be `E::number_of_columns() / 2` in both cases

jimmygchen: We can avoid the complex types and boxing here by collecting in each branch:

```rust
  let blob_bytes = if data_columns.len() < E::number_of_columns() {
      let (recovered_cells, _kzg_proofs) = kzg
          .recover_cells_and_compute_kzg_proofs(&cell_ids, &cells)
          .map_err(|e| {
              format!("Failed to recover cells and compute KZG proofs: {e:?}")
          })?;

      recovered_cells
          .into_iter()
          .take(num_cells_original_blob)
          .flat_map(|cell| cell.into_iter())
          .collect()
  } else {
      cells
          .into_iter()
          .take(num_cells_original_blob)
          .flat_map(|cell| cell.into_iter())
          .collect()
  };
```

This way we avoid collecting twice and avoid using `Box<dyn Interator>`

beacon_node/http_api/src/block_id.rs
jimmygchen: Thinking about this again, i think it will be cleaner if we just return `Vec<BlobSidecar>` from `get_blobs_from_data_columns` and `get_blobs` and avoid wrapping `BlobSidecar`s in `Arc`, and having to clone the blob later.

beacon_node/http_api/src/block_id.rs
jimmygchen: Right, I understand the fix now, but I don't understand how we're persisting duplicate indices, do you know why?

chong-he: Do you mean how do we get duplicate indices in the first place? From `get_blobs_by_versioned_hashes`, if the `versioned_hash` that we input to the function contains duplicated same `versioned_hash`, then we end up getting duplicated index, hence duplicated blobs from the function. 

The `versioned_hash` that we used in the test are exactly the same, which is why we get duplicated index in the first place.

beacon_node/beacon_chain/src/kzg_utils.rs
chong-he: Revised in https://github.com/sigp/lighthouse/pull/8103/commits/884e1171fe60ce2553049a875566260387eac7f0

beacon_node/http_api/src/block_id.rs
jimmygchen: ah got it now! thanks

beacon_node/http_api/src/block_id.rs
chong-he: Change the output as suggested, but I am not sure if that's the right way, because I am seeing more `.clone()` in the conversion, but managed to get rid of the clone in `blob: sidecar.blob.clone()`

beacon_node/http_api/src/block_id.rs
jimmygchen, line 351: This can be simplified to:

```rust
        let blob_sidecars = if !blob_kzg_commitments.is_empty() {
            if chain.spec.is_peer_das_enabled_for_epoch(block.epoch()) {
                Self::get_blobs_from_data_columns(chain, root, query.indices, &block)?
            } else {
                Self::get_blobs(chain, root, query.indices)?
            }
        } else {
            vec![]
        };

        let blob_sidecar_list = BlobSidecarList::new(
            blob_sidecars.into_iter().map(Arc::new).collect(),
            max_blobs_per_block,
        )
        .map_err(|e| warp_utils::reject::custom_server_error(format!("{:?}", e)))?;
```

beacon_node/http_api/src/block_id.rs
jimmygchen: ah i didn't realise `store.get_blobs` gives us Arc'd `BlobSidecar`s so we aren't actually saving any blob clones 🤦 
Sorry, your approach before was better (feel free to push back though!), and probably best to revert this commit.

beacon_node/http_api/src/block_id.rs
chong-he: If we revert the commit (which I have reverted), the output from `get_blobs` and `get_blobs_from_data_columns` will be `RuntimeVariableList<Arc<BlobSidecar<E>>>`, which the else branch having a `vec![]` is not of the same type?

beacon_node/http_api/src/block_id.rs
chong-he: reverted the commit

jimmygchen (approved):
Looks great now, thanks @chong-he !

michaelsproul:
@mergify requeue

Issue #8105, opened 2025-09-23
Attestation publishing failures due to insufficient peers on attestation subnets


jimmygchen:
## Description

Seeing this warning on nodes running latest `unstable`

```
Sep 22 06:55:43.408 WARN  Could not publish message                     error: NoPeersSubscribedToTopic, kind: beacon_attestation_17
```

Most of the nodes have around 200 peers due to the recent peer count increase, however it seems like the new peer balancing strategy implemented in #7915 isn't working well for attestation subnets.

## Version

`unstable`

## Expected Behaviour

Should have no warning when publishing attestations.

## Steps to resolve

Run the node with validators for some time.

jimmygchen:
Fixed in #8109

PR #8109, opened 2025-09-24
Only send data coumn subnet discovery requests after peerdas is scheduled

jimmygchen:
## Issue Addressed

#8105 (to be confirmed)

I noticed a large number of failed discovery requests after deploying latest `unstable` to some of our testnet and mainnet nodes. This is because of a recent PeerDAS change to attempt to maintain sufficient peers across data column subnets - this shouldn't be enabled on network without peerdas scheduled, otherwise it will keep retrying discovery on these subnets and never succeed.

Also removed some unused files.


beacon_node/lighthouse_network/src/peer_manager/mod.rs
```diff
@@ -1435,7 +1435,9 @@ impl<E: EthSpec> PeerManager<E> {
         self.update_peer_score_metrics();
 
         // Maintain minimum count for custody peers.
-        self.maintain_custody_peers();
+        if self.network_globals.spec.is_peer_das_scheduled() {
```
dapplion: Ideally we want to start doing this a little bit before the fork activates (say X epochs). This check `is_peer_das_scheduled` can be very early if a node updates its software right after the mainnet release and no one else has updated yet

pawanjay176: Agree. 2 slots before is when we subscribe to the new topics, so seems like a good time

jimmygchen: Thanks guys, good catch! I'll make the change

jimmygchen: I'm going to implement this a bit differently because `peer_manager` has no idea of the slot clock (and it probably shouldn't), i'm thinking to check for subscribed topics instead of 2 slots prior to Fulu - if we are subscribed to any column topics, then try to maintain custody peers.

i think the effect would be similar (with a slight delay), without introducing slot clock dependency.

jimmygchen: Done https://github.com/sigp/lighthouse/pull/8109/commits/8b7fa4a1d40f50733e38489f71a58cbb631baf88

pawanjay176 (approved):
LGTM. 
Just noting that we are doing the gossipsub topics check here since we don't have a slot clock in the peer manager to check 2 slots before the fulu fork.
even though using the topics seem a bit hacky, its okay to do imo since the alternative would involve a lot of code that would get thrown away once we are in fulu

Issue #8104, opened 2025-09-23
Stuck block lookup on nodes running latest `unstable`


jimmygchen:
## Description

Please provide a brief description of the issue.

## Version

`unstable`

## Present Behaviour

Seeing quite a lot of stuck lookup logs after latest `unstable` deployment to our sepolia and hoodi nodes:

```
Sep 22 03:06:54.358 WARN  Notify the devs a sync lookup is stuck        block_root: 0x7eb2f7a1660008dacb0ea735b9f5cf7210822ae09233a2f10bd6b30b365deff3, lookup: SingleBlockLookup { id: 420, block_request_state: BlockRequestState { state: SingleLookupRequestState { state: AwaitingDownload(block in processing cache), failed_processing: 0, failed_downloading: 0 } }, component_requests: ActiveBlobRequest(BlobRequestState { state: SingleLookupRequestState { state: Processed(no indices to fetch), failed_processing: 0, failed_downloading: 0 } }, 8), peers: 1, block_root: 0x7eb2f7a1660008dacb0ea735b9f5cf7210822ae09233a2f10bd6b30b365deff3, awaiting_parent: None, created: Instant { tv_sec: 2487708, tv_nsec: 655628474 }, span: Span { name: "single_block_lookup", level: Level(Debug), target: "network::sync::block_lookups::single_block_lookup", id: Id(8455515447213686799), module_path: "network::sync::block_lookups::single_block_lookup", line: 95, file: "beacon_node/network/src/sync/block_lookups/single_block_lookup.rs" } }
```

The most recent stuck lookup fix was #8005, it seemed to improve things looking at metrics at the time but may have introduced some other failure cases.

jimmygchen:
Hey @dapplion would you mind taking a look at this please?

jimmygchen:
Fixed in #8112

PR #8110, opened 2025-09-24
Send BlockComponentProcessed on HTTP block import

dapplion:
## Issue Addressed

- PR https://github.com/sigp/lighthouse/pull/8045 introduced a regression of how lookup sync interacts with the da_checker. 

Now in unstable block import from the HTTP API also insert the block in the da_checker while the block is being execution verified. If lookup sync finds the block in the da_checker in `NotValidated` state it expects a `GossipBlockProcessResult` message sometime later. That message is only sent after block import in gossip. 

I confirmed in our node's logs for 4/4 cases of stuck lookups are caused by this sequence of events:
- Receive block through API, insert into da_checker in fn process_block in put_pre_execution_block
- Create lookup and leave in AwaitingDownload(block in processing cache) state
- Block from HTTP API finishes importing
- Lookup is left stuck

Closes https://github.com/sigp/lighthouse/issues/8104

## Proposed Changes

This PR fixes the issue as when the HTTP block completes import the lookup continues thanks to the `BlockProcessResult` event.

The interaction of lookup sync with the da_checker feels very fragile. I reviewed https://github.com/sigp/lighthouse/pull/8045 and didn't realize that we had to send this event. The sync tests do not cover this code either. We only assert that certain events are sent to the beacon processor but we don't execute them. I would like to have a "better" fix at some point, such as storing a handle in the da_checker that lookup sync can await. However, this approach requires way more changes as sync is not async.

### DIFF
--- a/beacon_node/http_api/src/publish_blocks.rs
+++ b/beacon_node/http_api/src/publish_blocks.rs
@@ -538,6 +538,17 @@ async fn post_block_import_logging_and_response<T: BeaconChainTypes>(
     seen_timestamp: Duration,
     chain: &Arc<BeaconChain<T>>,
 ) -> Result<Response, Rejection> {
+    // Notify lookup sync of completing a block import. Necessary if a lookup was created after
+    // an HTTP block start import but before it completes.
+    self.send_sync_message(SyncMessage::BlockProcessResult {
+        block_root,
+        imported: matches!(
+            result,
+            Ok(AvailabilityProcessingStatus::Imported(_))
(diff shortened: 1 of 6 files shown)


dapplion:
- https://github.com/sigp/lighthouse/pull/8112 is a better solution for https://github.com/sigp/lighthouse/issues/8104

jimmygchen:
Thanks @dapplion, I agree that #8112 is a better solution 👍

PR #8111, opened 2025-09-24
Include block root in publish block logs

dapplion:
## Issue Addressed

Debugging https://github.com/sigp/lighthouse/issues/8104 it would have been helpful to quickly see in the logs that a specific block was submitted into the HTTP API. 

Because we want to optimize the block root computation we don't include it in the logs, and just log the block slot. I believe we can take a minute performance hit to have the block root in all the logs during block publishing.


beacon_node/http_api/src/publish_blocks.rs
```diff
@@ -107,8 +107,9 @@ pub async fn publish_block<T: BeaconChainTypes, B: IntoGossipVerifiedBlock<T>>(
     current_span.record("provenance", provenance);
 
     let block = unverified_block.inner_block();
+    let block_root = block.canonical_root();
```
jimmygchen: 1. We can do `block_root.unwrap_or_else` here, as block root is already calculated when calling from the publish blinded block path.
2. If you record this to the span, the `block_root` field will be available to both the span and all logs in this function:

```rust
current_span.record("block_root", ?block_root);
```

jimmygchen:
Hey @dapplion I think this is useful so I'm going to revive this one!

dapplion (commented):
✅ looks good! Thanks for continuing it @jimmygchen

jimmygchen (approved):

PR #8112, opened 2025-09-24
Only mark block lookups as pending if block is importing from gossip

dapplion:
## Issue Addressed

- PR https://github.com/sigp/lighthouse/pull/8045 introduced a regression of how lookup sync interacts with the da_checker. 

Now in unstable block import from the HTTP API also insert the block in the da_checker while the block is being execution verified. If lookup sync finds the block in the da_checker in `NotValidated` state it expects a `GossipBlockProcessResult` message sometime later. That message is only sent after block import in gossip. 

I confirmed in our node's logs for 4/4 cases of stuck lookups are caused by this sequence of events:
- Receive block through API, insert into da_checker in fn process_block in put_pre_execution_block
- Create lookup and leave in AwaitingDownload(block in processing cache) state
- Block from HTTP API finishes importing
- Lookup is left stuck

Closes https://github.com/sigp/lighthouse/issues/8104

## Proposed Changes

- https://github.com/sigp/lighthouse/pull/8110 was my initial solution attempt but we can't send the `GossipBlockProcessResult` event from the `http_api` crate without adding new channels, which seems messy.

For a given node it's rare that a lookup is created at the same time that a block is being published. This PR solves https://github.com/sigp/lighthouse/issues/8104 by allowing lookup sync to import the block twice in that case.

### DIFF
--- a/beacon_node/network/src/sync/network_context.rs
+++ b/beacon_node/network/src/sync/network_context.rs
@@ -49,8 +49,8 @@ use tokio::sync::mpsc;
 use tracing::{Span, debug, debug_span, error, warn};
 use types::blob_sidecar::FixedBlobSidecarList;
 use types::{
-    BlobSidecar, ColumnIndex, DataColumnSidecar, DataColumnSidecarList, EthSpec, ForkContext,
-    Hash256, SignedBeaconBlock, Slot,
+    BlobSidecar, BlockImportSource, ColumnIndex, DataColumnSidecar, DataColumnSidecarList, EthSpec,
+    ForkContext, Hash256, SignedBeaconBlock, Slot,
 };
 
 pub mod custody;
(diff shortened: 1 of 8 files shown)


eserilev (approved):
Allowing double imports seems fine to me, especially if these are rare. Wondering if there's a metric we could add somewhere to keep track of instances where a block has been imported twice? Can't find any issues with this PR, but another set of eyes on this would be nice.

beacon_node/network/src/sync/network_context.rs
```diff
@@ -835,14 +835,26 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+                    // are guaranteed to receive a `SyncMessage::GossipBlockProcessResult` that will
+                    // make progress on this lookup
+                    return Ok(LookupRequestResult::Pending("block in processing cache"));
+                }
+                BlockImportSource::Lookup
+                | BlockImportSource::RangeSync
+                | BlockImportSource::HttpApi => {
+                    // Lookup, RangeSync or HttpApi block import don't emit the GossipBlockProcessResult
+                    // event. If a lookup happens to be created during block import from one of
+                    // those sources just import the block twice. Otherwise the lookup will get
```
jimmygchen, line 854: Just a comment on the doc here:
- I think re-imports wouldn't happen to lookup responses because of `duplicate_cache` here
https://github.com/sigp/lighthouse/blob/78d330e4b7e2b76ab503cd88f4a365a6d7a0bcf0/beacon_node/network/src/network_beacon_processor/sync_methods.rs#L121
- for range sync its unlikely to race with lookups
- http api block imports are much less frequent,  so the impact of re-imports is much smaller

jimmygchen (approved):
Niice, thanks for fixing this! 

The fix looks good to me, be great to add some regression test for this as well.

beacon_node/network/src/sync/network_context.rs
jimmygchen: Also the impact of duplicate imports is not as bad since michael made the fix in #8050 
As described in the PR - longer term we'll have @ethDreamer's block status tracker to prevent block re-imports properly.

jimmygchen:
Merging this now since the fix is simple and makes sense. Great work finding the issue! 🕵️

beacon_node/network/src/sync/network_context.rs
dapplion: > longer term we'll have @ethDreamer's block status tracker to prevent block re-imports properly.

I noticed in this PR that we'll need to re-engineer sync a bit to for that to work. But working with Mark about it

PR #8113, opened 2025-09-24
Remove Server header from HTTP API responses to reduce information disclosure

viktorking7:
**Description**
### Summary
Removes the `Server` header from `beacon_node` HTTP API responses to avoid exposing version/platform details.

### Rationale
- Minimizes fingerprinting and targeted attacks
- Follows least-information best practice for public APIs

### Changes
- `beacon_node/http_api/src/lib.rs`: drop `Server` header; add a brief comment

### DIFF
--- a/beacon_node/http_api/src/lib.rs
+++ b/beacon_node/http_api/src/lib.rs
@@ -4910,8 +4910,7 @@ pub fn serve<T: BeaconChainTypes>(
         .recover(warp_utils::reject::handle_rejection)
         .with(tracing_logging())
         .with(prometheus_metrics())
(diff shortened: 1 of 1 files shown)


eserilev:
> Follows least-information best practice for public APIs

The beacon API isn't exposed publicly 

I'm not sure what this PR is aiming to solve, and I'm not a fan of accepting trivial first time contributions from private github profiles

michaelsproul:
Indeed the BN API is not intended to be public. If you're exposing parts of the BN API to the internet you should be using a load balancer, blocking some requests and stripping headers.

PR #8121, opened 2025-09-26
Fix bug in fork calculation at fork boundaries

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

In #8101 , when we modified the logic to get the proposer index post fulu, we seem to have missed advancing the state at the fork boundaries to get the right `Fork` for signature verification. 
This led to lighthouse failing all gossip verification right after transitioning to fulu that was observed on the holesky shadow fork
```
Sep 26 14:24:00.088 DEBUG Rejected gossip block                         error: "InvalidSignature(ProposerSignature)", graffiti: "grandine-geth-super-1", slot: 640 
Sep 26 14:24:00.099 WARN  Could not verify block for gossip. Rejecting the block  error: InvalidSignature(ProposerSignature)
```

I'm not completely sure this is the correct fix, but this fixes the issue with `InvalidProposerSignature` on the holesky shadow fork.

Thanks to @eserilev for helping debug this

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_proposer_cache.rs
+++ b/beacon_node/beacon_chain/src/beacon_proposer_cache.rs
@@ -234,12 +234,19 @@ pub fn ensure_state_can_determine_proposers_for_epoch<E: EthSpec>(
     if state.current_epoch() > maximum_epoch {
         Err(BeaconStateError::SlotOutOfBounds.into())
     } else if state.current_epoch() >= minimum_epoch {
-        // Fulu allows us to access shufflings in multiple epochs (thanks to lookahead).
-        // Pre-Fulu we expect `minimum_epoch == maximum_epoch`, and this branch covers that case.
+        if target_epoch > state.current_epoch() {
+            let target_slot = target_epoch.start_slot(E::slots_per_epoch());
+
+            // Advance the state into the same epoch as the block. Use the "partial" method since state
+            // roots are not important for proposer/attester shuffling.
+            partial_state_advance(state, Some(state_root), target_slot, spec)
+                .map_err(BeaconChainError::from)?;
+        }
         Ok(())
     } else {
         // State's current epoch is less than the minimum epoch.
         // Advance the state up to the minimum epoch.
+        tracing::debug!(state_current_epoch=?state.current_epoch(), ?minimum_epoch,fork=?state.fork(),"No We are here");
         partial_state_advance(state, Some(state_root), minimum_slot, spec)
             .map_err(BeaconChainError::from)
     }
--- a/beacon_node/beacon_chain/src/beacon_proposer_cache.rs
+++ b/beacon_node/beacon_chain/src/beacon_proposer_cache.rs
@@ -246,7 +246,6 @@ pub fn ensure_state_can_determine_proposers_for_epoch<E: EthSpec>(
     } else {
         // State's current epoch is less than the minimum epoch.
         // Advance the state up to the minimum epoch.
-        tracing::debug!(state_current_epoch=?state.current_epoch(), ?minimum_epoch,fork=?state.fork(),"No We are here");
         partial_state_advance(state, Some(state_root), minimum_slot, spec)
             .map_err(BeaconChainError::from)
     }


pawanjay176:
Kurtosis config I was using to test the change:
```yaml
participants_matrix:
  el:
    - el_type: besu
      el_image: ethpandaops/besu:main
  cl:
    - cl_type: grandine
      cl_image: ethpandaops/grandine:develop
      supernode: true
      cl_log_level: debug
      count: 3
    - cl_type: lighthouse
      cl_image: lighthouse-local:latest
      supernode: true
      cl_log_level: debug
    - cl_type: lighthouse
      cl_image: ethpandaops/lighthouse:unstable
      supernode: true
      cl_log_level: debug


network_params:
  fulu_fork_epoch: 1

additional_services:
  - dora
  - spamoor

global_log_level: debug
spamoor_params:
  image: ethpandaops/spamoor:master
  max_mem: 4000
  spammers:
    - scenario: eoatx
      config:
        throughput: 200
    - scenario: blobs
      config:
        throughput: 20
```

The unstable lighthouse node stalls at slot 32 (fulu slot) while this branch progresses fine

eserilev:
I was also able to repro the bug on current unstable, with pawans fix we are able to progress the chain just fine

eserilev (approved):

jimmygchen:
Nice find! I'll have to think about the fix a bit more.

Just to understand this scenario, this didn't happen on other paths (blob / column), because we `state_provider` function returns the advanced state to `slot` instead of parent state here:

https://github.com/sigp/lighthouse/blob/a08fd93e9a6485d3d1b4d1b6273294decbf40d66/beacon_node/beacon_chain/src/block_verification.rs#L957-L977

In gossip block verification >  `with_proposer_cache` > [`ensure_state_can_determine_proposers_for_epoch`](https://github.com/sigp/lighthouse/blob/4ceccd246386de82fafe59f783ef4b58da8d5c4f/beacon_node/beacon_chain/src/beacon_chain.rs#L6587) didn't try to advance the electra parent state to the same slot as `target_epoch` (fulu epoch) because the state should have sufficient lookahead.

So does the failure here mean the proposer indices for  fork (epoch `N`) calculated from the parent state (epoch `N-1`)  is different to the `proposer_lookahead` computed during fulu fork transition? And we can't rely on the lookahead computed from the electra parent state?

beacon_node/beacon_chain/src/beacon_proposer_cache.rs
```diff
@@ -234,8 +234,14 @@ pub fn ensure_state_can_determine_proposers_for_epoch<E: EthSpec>(
...
         Err(BeaconStateError::SlotOutOfBounds.into())
     } else if state.current_epoch() >= minimum_epoch {
-        // Fulu allows us to access shufflings in multiple epochs (thanks to lookahead).
-        // Pre-Fulu we expect `minimum_epoch == maximum_epoch`, and this branch covers that case.
+        if target_epoch > state.current_epoch() {
+            let target_slot = target_epoch.start_slot(E::slots_per_epoch());
+
+            // Advance the state into the same epoch as the block. Use the "partial" method since state
+            // roots are not important for proposer/attester shuffling.
+            partial_state_advance(state, Some(state_root), target_slot, spec)
```
jimmygchen, line 242: I'm not sure if this is required in all cases as we have 2 epochs of proposer lookahead from Fulu.

eserilev: this might only be required at fork boundaries, like when the `target_epoch` fork > `state.current_fork`

or maybe only at the fulu fork transition, i'm not fully across the fulu changes to proposer shuffling yet tbh

michaelsproul:
Good catch! I think another fix might be just to compute what the fork "will be" at some future epoch? So use something like `state.fork_at_epoch(epoch)` rather than `state.fork()`?

But yeah maybe just advancing the state fully into the new epoch is safer and clearer. We probably don't need the optimisation where we use the lookahead to advance less

beacon_node/beacon_chain/src/beacon_proposer_cache.rs
michaelsproul: IMO it should only be required at the boundary, but I am fine with the "safer" fix or just always doing the state advance. We can come back to optimise it later

michaelsproul: The optimised version would look something like setting minimum epoch to `max(minimum_epoch, fulu_fork_epoch)` when `target_epoch >= fulu_fork_eppch`.

jimmygchen:
> So does the failure here mean the proposer indices for fork (epoch `N`) calculated from the parent state (epoch `N-1`) is different to the `proposer_lookahead` computed during fulu fork transition? And we can't rely on the lookahead computed from the electra parent state?

@pawanjay176 pointed out that the proposer lookahead was correct, and it was to do with the proposer.fork 

> the proposer index that was calculated is correct. 
the Fork for the signature verification that we pushed in the beacon proposers cache is incorrect
we retrieved it here
https://github.com/sigp/lighthouse/blob/unstable/beacon_node/beacon_chain/src/block_verification.rs#L979

beacon_node/beacon_chain/src/beacon_proposer_cache.rs
jimmygchen: Pawan mentioned that the proposer index was correct but the fork used for signing was incorrect (electra).

Can we simply return `self.spec.fork_at_epoch(proposal_epoch)` here instead of `state.fork()`?

https://github.com/sigp/lighthouse/blob/4ceccd246386de82fafe59f783ef4b58da8d5c4f/beacon_node/beacon_chain/src/beacon_chain.rs#L6609-L6614

pawanjay176: yeah that makes sense to me, i don't know if not advancing the state would cause an issue in some other edge case though.

michaelsproul: Yeah I think that would work. Do we actually have a fork_at_epoch method that takes into account multiple forks activating at the same epoch?

michaelsproul: Yeah I think advancing the state is the more conservative approach. We can optimise further later

eserilev: I believe there's a few other places where we are inserting into the beacon proposer cache which might also potentially need to use the `fork_at_epoch` method if we don't advance the state

pawanjay176: Also to confirm, we don't need to build the committee cache after advancing the state like here right?
https://github.com/sigp/lighthouse/blob/c754234b2c94d90ed658788b5fb69ee405ed6cb7/beacon_node/beacon_chain/src/block_verification.rs#L2065-L2066

michaelsproul: > Also to confirm, we don't need to build the committee cache after advancing the state like here right?

Hmm good point. These caches are not required in general for the proposer cache, but we may need them in block verification. It wouldn't hurt to add them back here (I removed them by not using `cheap_state_advance`):

https://github.com/sigp/lighthouse/blob/c754234b2c94d90ed658788b5fb69ee405ed6cb7/beacon_node/beacon_chain/src/block_verification.rs#L962-L975

Although I think we end up doing it here anyway:

https://github.com/sigp/lighthouse/blob/c754234b2c94d90ed658788b5fb69ee405ed6cb7/beacon_node/beacon_chain/src/block_verification.rs#L1187-L1192

jimmygchen (approved):
Thanks @pawanjay176 @eserilev 
Great turnaround 👏  as discussed let’s merge this safer fix and optimise later!

PR #8125, opened 2025-09-27
Migrate derivative to educe

### LINKED ISSUE #7001: Remove `derivative` dependency in favour of `educe`
The `derivative` dependency is now unmaintained. We use it to customise `derive`d implementations of traits.

In our dependent crates we've already started switching to `educe` which offers similar functionality. We can switch Lighthouse over to `educe` too once those deps are updated, which will happen as part of:

- https://github.com/sigp/lighthouse/pull/6915

jchavarri:
## Issue Addressed

Fixes #7001.

## Proposed Changes

Mostly mechanical replacement of `derivative` attributes with `educe` ones.

### **Attribute Syntax Changes**

```rust
// Bounds: = "..." → (...)
#[derivative(Hash(bound = "E: EthSpec"))]
#[educe(Hash(bound(E: EthSpec)))]

// Ignore: = "ignore" → (ignore)  
#[derivative(PartialEq = "ignore")]
#[educe(PartialEq(ignore))]

// Default values: value = "..." → expression = ...
#[derivative(Default(value = "ForkName::Base"))]
#[educe(Default(expression = ForkName::Base))]

// Methods: format_with/compare_with = "..." → method(...)
#[derivative(Debug(format_with = "fmt_peer_set_as_len"))]
#[educe(Debug(method(fmt_peer_set_as_len)))]

// Empty bounds: removed entirely, educe can infer appropriate bounds
#[derivative(Default(bound = ""))]
#[educe(Default)]

// Transparent debug: manual implementation (educe doesn't support it)
#[derivative(Debug = "transparent")]
// Replaced with manual Debug impl that delegates to inner field
```

**Note**: Some bounds use strings (`bound("E: EthSpec")`) for superstruct compatibility (`expected ','` errors).

## Additional Info

`derivate` remains as a transitive dep through `alloy-primitives` -> `ruint` -> `ark-ff` -> `derivative`.


jchavarri:
> **Note**: Some bounds use strings (`bound("E: EthSpec")`) for superstruct compatibility (`expected ','` errors).

It seems this was fixed in recent versions of `darling`: https://github.com/TedDriggs/darling/releases/tag/v0.20.3. Updating `superstruct` to `0.10.0` solves the limitation.

Aside from that, trying to figure out / reproduce locally the ci build errors on the `check_consensus_block_proposals` task.

macladson (commented):
This looks good to me, but for consistency I think we may as well bump `superstruct` so we can get rid of the strings in some of the bounds. I've opened a PR here ->  https://github.com/sigp/lighthouse/pull/8133

macladson (changes requested):
@jchavarri We have bumped the `unstable` branch to `superstruct` version `0.10.0` 
Can you change the trait bounds to remove those strings?

jchavarri:
@macladson done! seems to work fine 🎉

consensus/types/src/attestation.rs
```diff
@@ -599,18 +600,8 @@ impl<'de, E: EthSpec> ContextDeserialize<'de, ForkName> for Vec<Attestation<E>>
...
-    Deserialize,
-    Decode,
-    Encode,
-    TestRandom,
-    Derivative,
-    TreeHash,
-    PartialEq,
-)]
+#[derive(Debug, Clone, Serialize, Deserialize, Decode, Encode, TestRandom, Educe, TreeHash)]
+#[educe(PartialEq)]
```
macladson: I think `Derivative` was unused in this struct, so we can probably just skip deriving `Educe` on it

testing/ef_tests/src/handler.rs
```diff
@@ -154,8 +154,8 @@ pub trait Handler {
 
 macro_rules! bls_eth_handler {
     ($runner_name: ident, $case_name:ident, $handler_name:expr) => {
-        #[derive(Derivative)]
-        #[derivative(Default(bound = ""))]
+        #[derive(Educe)]
+        #[educe(Default)]
```
macladson, line 158: This all compiles and works anyway so it's probably not a big deal, but I _think_ the `educe` syntax would be:
```rust
#[educe(Default())]
```

Otherwise I believe all the trait bounds would get added to the `Default` impl. This is some ancient code though so I'm not 100% sure on why `Derivative` was used here in the first place.

macladson (changes requested):
Looks pretty good to me! Just some minor comments

consensus/types/src/attestation.rs
jchavarri: 👍 Fixed in 3df8778.

testing/ef_tests/src/handler.rs
jchavarri: I got this syntax from the [educe docs](https://docs.rs/educe/latest/educe/#default). 🤔 

Changed in 10108bc.

macladson: Yeah the syntax you got from the docs was valid, but I think the `()` syntax is equivalent to the previous `bound = ""`. It's not very well documented though so I'm not 100% sure

macladson (approved):
After some internal discussion I've reverted the Default impl change since it's cleaner and is irrelevant for the tests. Sorry for the mix-up.

This PR looks good now so we can merge soon :rocket:

macladson (commented):
Hey @jchavarri 
Sorry for the delay. Looks like there are some conflicts, are you able to fix them?
Once it's fixed I'll push to get this merged ASAP

jchavarri:
@macladson no worries. Done!

PR #8128, opened 2025-09-28
Fix proposer shuffling decision slot at boundary

michaelsproul:
## Issue Addressed

Follow-up to the bug fixed in:

- https://github.com/sigp/lighthouse/pull/8121

This fixes the root cause of that bug, which was introduced by me in:

- https://github.com/sigp/lighthouse/pull/8101

Jimmy and Lion both identified the issue here:

- https://github.com/sigp/lighthouse/pull/8101#discussion_r2382258842
- https://github.com/sigp/lighthouse/pull/8101#discussion_r2382710356

## Proposed Changes

In the methods that compute the proposer shuffling decision root, ensure we don't use lookahead for the Fulu fork epoch itself. This is accomplished by checking if Fulu is enabled at `epoch - 1`, i.e. if `epoch > fulu_fork_epoch`.

I haven't updated the methods that _compute_ shufflings to use these new corrected bounds (e.g. `BeaconState::compute_proposer_indices`), although we could make this change in future. The `get_beacon_proposer_indices` method already gracefully handles the Fulu boundary case by using the `proposer_lookahead` field (if initialised).

## Additional Info

- [x] Unit test added for `proposer_shuffling_decision_slot` at the boundary
- [x] Updated the consistency tests to check cases around the fork boundary

### DIFF
--- a/beacon_node/beacon_chain/tests/store_tests.rs
+++ b/beacon_node/beacon_chain/tests/store_tests.rs
@@ -1195,14 +1195,18 @@ fn check_shuffling_compatible(
 ///
 /// - ProtoBlock::proposer_shuffling_root_for_child_block, and
 /// - BeaconState::proposer_shuffling_decision_root{_at_epoch}
-async fn proposer_shuffling_root_consistency_test(parent_slot: u64, child_slot: u64) {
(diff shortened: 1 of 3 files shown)


jimmygchen (approved):
Nice tests!

PR #8130, opened 2025-09-29
More proposer shuffling cleanup

michaelsproul:
## Issue Addressed

Addressing more review comments from:

- https://github.com/sigp/lighthouse/pull/8101

I've also tweaked a few more things that I think are minor bugs.

## Proposed Changes

- Instrument `ensure_state_can_determine_proposers_for_epoch`
- Fix `block_root` usage in `compute_proposer_duties_from_head`. This was a regression introduced in 8101 :grimacing: .
- Update the `state_advance_timer` to prime the next-epoch proposer cache post-Fulu.

## Additional Info

- [x] Add test for next epoch shuffling case in HTTP duties API. This test is `proposer_duties_from_head_fulu` and I have confirmed it fails when reverting the usage of `proposer_shuffling_decision_root_at_epoch` with the correct block root.
- [x] Add regression test for bug in `try_proposer_duties_from_cache`. See `proposer_duties_with_gossip_tolerance`.


beacon_node/http_api/src/proposer_duties.rs
```diff
@@ -119,6 +111,15 @@ fn try_proposer_duties_from_cache<T: BeaconChainTypes>(
         )));
     }
 
+    let head_decision_root = head
+        .snapshot
+        .beacon_state
+        .proposer_shuffling_decision_root_at_epoch(request_epoch, head_block_root, &chain.spec)
```
michaelsproul, line 117: We had regressed here (in 8101) when I removed this code:

https://github.com/sigp/lighthouse/blob/cfb1f7331064b758c6786e4e1dc15507af5ff5d1/beacon_node/http_api/src/proposer_duties.rs#L114-L126

My intent was to replace the manual calculation of the dependent root with the unified `proposer_shuffling_decision_root_at_epoch` function, but I mistakenly kept using the `proposer_shuffling_decision_root` function which computes the decision root for the state's _current epoch_. At epoch boundaries, this codepath could be called with `request_epoch == head_epoch + 1`, in which case `proposer_shuffling_decision_root` would be wrong (it would be the decision block for the current epoch instead of the next).

Using `proposer_shuffling_decision_root_at_epoch` fixes this, and is generic over pre-Fulu vs post-Fulu. It uses the request epoch to determine the decision slot, and then uses either the passed in `head_block_root` or state's block roots to compute the correct root.

beacon_node/beacon_chain/src/beacon_proposer_cache.rs
```diff
@@ -199,8 +200,7 @@ pub fn compute_proposer_duties_from_head<T: BeaconChainTypes>(
         .map_err(BeaconChainError::from)?;
 
     let dependent_root = state
-        // The only block which decides its own shuffling is the genesis block.
-        .proposer_shuffling_decision_root(chain.genesis_block_root, &chain.spec)
+        .proposer_shuffling_decision_root_at_epoch(request_epoch, head_block_root, &chain.spec)
```
michaelsproul, line 203: The previous logic here was relying on an assumption that is no longer true in the "optimised" version of #8101 (without https://github.com/sigp/lighthouse/pull/8121). The assumption is that we've advanced the state into the `request_epoch`, so the decision slot can never be newer than the state's slot, and therefore the block_root argument is not required (except at genesis). AND that we can use `state.current_epoch()` when determining the decision slot (which is fine because `state.current_epoch() == request_epoch` thanks to the advance).

With 8101 prior to Pawan's fix, it is still true that the block root argument isn't required, because we have advanced the state to the point where it can compute the shuffling for `request_epoch`. Prior to Fulu this means we've already done the epoch transition into `request_epoch`, and post-Fulu this means the earliest epoch we could be in is `request_epoch - 1`, and the decision block root in this case is the end of `request_epoch - 2` (i.e. less than our `state.slot()` and in range of `state.block_roots()`). However, what is no longer true is that we should use the decision root for `state.current_epoch()`, because this is not necessarily the same as `request_epoch`. E.g. post-Fulu we could have it that the state is sitting at `request_epoch - 1`, which would lead us to use the decision root from 1 epoch further in the past than we should.

This bug was introduced in #8101, and then re-patched by Pawan's changes in #8121 which force the state advance into `request_epoch` even when this isn't technically required post-Fulu. As an optimisation I would like to eventually revert #8121 so that we can do slightly fewer state advances. At which point the fix in this will become load-bearing.

beacon_node/http_api/src/proposer_duties.rs
michaelsproul: The impact of this is that in v8.0.0-rc.0 we might return the wrong duties for an epoch, and put a nonsensical entry into the cache. This could lead to a block proposal failure in the worst case (the VC would make a request for a block when it isn't actually the proposer), which would be caught by the proposer index checks in `produce_block` (which are not buggy). If the same VC holds the key for the true proposer then it might miss the proposal for that true proposer.

I think this is low impact enough that it is OK not to delay v8.0.0-rc.0 over, although we should fix it soon!

michaelsproul: For this to return the wrong duties in practice requires that there be an entry for `(current_epoch_decision_block_root, request_epoch)` in the cache. This should not happen, unless there's another bug somewhere that would add a value for this pair to the cache. As far as I can tell, there isn't (need to check the state advance).

michaelsproul: The other way we could end up with an entry for that key is if there's a long skip on an alternate chain using that shuffling. This is also quite unlikely, but I guess could be triggered intentionally by an attacker to force some LH proposers to fail proposing temporarily.

dapplion (approved):
LGTM

beacon_node/beacon_chain/src/beacon_proposer_cache.rs
michaelsproul: This is tested by `proposer_duties_from_head_fulu`. Reverting this line causes the test to fail.

beacon_node/beacon_chain/src/beacon_proposer_cache.rs
```diff
@@ -235,14 +239,6 @@ pub fn ensure_state_can_determine_proposers_for_epoch<E: EthSpec>(
...
         Err(BeaconStateError::SlotOutOfBounds.into())
     } else if state.current_epoch() >= minimum_epoch {
-        if target_epoch > state.current_epoch() {
-            let target_slot = target_epoch.start_slot(E::slots_per_epoch());
-
-            // Advance the state into the same epoch as the block. Use the "partial" method since state
-            // roots are not important for proposer/attester shuffling.
-            partial_state_advance(state, Some(state_root), target_slot, spec)
-                .map_err(BeaconChainError::from)?;
-        }
```
dapplion, line 244: why not necessary anymore?

michaelsproul: Pawan added this to work around the fact that we weren't advancing into the Fulu fork epoch to compute its shuffling (we were wrongly assuming we had lookahead). Now that we've fixed the [function to calculate the decision slot](https://github.com/sigp/lighthouse/blob/6ce8fd839d54a0af0a88bd766155fa10360acf65/consensus/types/src/chain_spec.rs#L880-L901), this is no longer a concern.

In this PR we've also started using `fork_at_epoch` so that the shuffling calculation works at future fork epochs (i.e. Gloas). This could be a good test case as well actually.

Removing the advance means we can take advantage of the lookahead post-Fulu. We avoid doing up to 1 epoch of unnecessary state advance.

michaelsproul: Gloas fork epoch test added in 3171c124406775118e4f5bf323f5b82e56237b32

beacon_node/http_api/src/proposer_duties.rs
michaelsproul: Test added in 97da066f9199978f9fe878aad044db4469af6923. I confirmed that this test fails when reverting this change.

michaelsproul:
Tests updated and ready for re-review.

I would like to get this in for v8.0.0-rc.2, as it fixes some bugs and should improve performance.

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -1501,6 +1520,120 @@ async fn proposer_shuffling_changing_with_lookahead() {
...
+async fn proposer_lookahead_gloas_fork_epoch() {
+    let gloas_fork_epoch = Epoch::new(4);
+    let mut spec = ForkName::Fulu.make_genesis_spec(E::default_spec());
+    spec.gloas_fork_epoch = Some(gloas_fork_epoch);
+
+    let db_path = tempdir().unwrap();
+    let store = get_store_generic(&db_path, Default::default(), spec.clone());
+    let validators_keypairs =
+        types::test_utils::generate_deterministic_keypairs(LOW_VALIDATOR_COUNT);
+    let harness = TestHarness::builder(MinimalEthSpec)
```
jimmygchen: nit: use `E` instead of `MinimalEthSpec` for consistency

beacon_node/http_api/tests/interactive_tests.rs
```diff
@@ -940,3 +940,106 @@ async fn queue_attestations_from_http() {
...
 }
+
+// Test that a request for next epoch proposer duties suceeds when the current slot clock is within
+// gossip clock disparity (500ms) of the new epoch.
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn proposer_duties_with_gossip_tolerance() {
+    let validator_count = 24;
+    let spec = ForkName::Fulu.make_genesis_spec(E::default_spec());
+
+    let tester = InteractiveTester::<E>::new(Some(spec.clone()), validator_count).await;
```
jimmygchen: we run all http-api tests on electra and fulu, so this test will get run twice with the same Fulu genesis conifg.Should we just pass `None` so we test two different forks?

jimmygchen (changes requested):
Looks good to me, I couldn't spot any issues, just some minor comments in test.

beacon_node/beacon_chain/tests/store_tests.rs
michaelsproul: Can't use `E`, it is not a value.

michaelsproul: Ah, could use `E::default()`, I'll do that

beacon_node/http_api/tests/interactive_tests.rs
michaelsproul: good call, generalised it to run for any fork

michaelsproul:
I've opened a follow-up PR here, but I don't think it's necessary for correctness, so leaving it out of v8.0.0(-rc.2):

- https://github.com/sigp/lighthouse/pull/8235

jimmygchen (approved):
LGTM!

PR #8133, opened 2025-09-29
Bump superstruct to `0.10.0`

macladson:
## Proposed Changes

Bump `superstruct` to the latest release `0.10.0`.
This version uses a later version of `darling` which is helpful for https://github.com/sigp/lighthouse/pull/8125


michaelsproul (approved):
LGTM, I see no reason not to include this in the next RC

PR #8134, opened 2025-09-29
Added Block Status Table

ethDreamer:
## Issue Addressed

This is useful for managing lock contention and will be really useful for gloas.

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -3035,7 +3035,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
                     return Err(BlockError::DuplicateFullyImported(block_root));
                 }
                 if status.is_past_pending() {
-                    return Err(BlockError::DuplicateImportStatusUnknown(block_root));
+                    return Err(BlockError::DuplicateImportStatusKnown(block_root, status));
                 }
             }
         }
@@ -3086,7 +3086,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
                     return Err(BlockError::DuplicateFullyImported(block_root));
                 }
                 if status.is_past_pending() {
-                    return Err(BlockError::DuplicateImportStatusUnknown(block_root));
+                    return Err(BlockError::DuplicateImportStatusKnown(block_root, status));
                 }
             }
         }
@@ -3131,7 +3131,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
                     return Err(BlockError::DuplicateFullyImported(block_root));
                 }
                 if status.is_past_pending() {
-                    return Err(BlockError::DuplicateImportStatusUnknown(block_root));
+                    return Err(BlockError::DuplicateImportStatusKnown(block_root, status));
                 }
             }
         }
@@ -3181,7 +3181,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
                     return Err(BlockError::DuplicateFullyImported(block_root));
                 }
                 if status.is_past_pending() {
-                    return Err(BlockError::DuplicateImportStatusUnknown(block_root));
+                    return Err(BlockError::DuplicateImportStatusKnown(block_root, status));
                 }
             }
         }
@@ -3282,7 +3282,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
                     return Err(BlockError::DuplicateFullyImported(block_root));
                 }
                 if status.is_past_pending() {
-                    return Err(BlockError::DuplicateImportStatusUnknown(block_root));
+                    return Err(BlockError::DuplicateImportStatusKnown(block_root, status));
                 }
             }
         }
@@ -3730,7 +3730,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
                     return Err(BlockError::DuplicateFullyImported(block_root));
                 }
                 if status.is_past_pending() {
-                    return Err(BlockError::DuplicateImportStatusUnknown(block_root));
+                    return Err(BlockError::DuplicateImportStatusKnown(block_root, status));
                 }
             }
         }
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -9,6 +9,7 @@ use crate::beacon_proposer_cache::{
     BeaconProposerCache, EpochBlockProposers, ensure_state_can_determine_proposers_for_epoch,
 };
 use crate::blob_verification::{GossipBlobError, GossipVerifiedBlob};
+use crate::block_status_table::{BlockStatus, BlockStatusTable, Error as BlockStatusTableError};
 use crate::block_times_cache::BlockTimesCache;
 use crate::block_verification::POS_PANDA_BANNER;
 use crate::block_verification::{
@@ -475,6 +476,8 @@ pub struct BeaconChain<T: BeaconChainTypes> {
     pub validator_monitor: RwLock<ValidatorMonitor<T::EthSpec>>,
     /// The slot at which blocks are downloaded back to.
     pub genesis_backfill_slot: Slot,
+    //// A table to track the status of blocks during data availability import
+    pub block_status_table: Arc<BlockStatusTable>,
     /// Provides a KZG verification and temporary storage for blocks and blobs as
     /// they are collected and combined.
     pub data_availability_checker: Arc<DataAvailabilityChecker<T>>,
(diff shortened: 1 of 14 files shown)


ethDreamer:
status_table (1)

ethDreamer:
## Basic Functionality

Whenever a component of a block is seen from the network:
```rust
if block_status_table.insert_pending(component.block_hash) {
    // block inserted with status PENDING
}

match da_cache.insert(component) {
    Availability::Available(block) => {
        if block_status_table.compare_and_swap_status(
            block.hash,
            BlockStatus::Pending,
            BlockStatus::Importing, // post gloas it will be BlockStatus::Executing
        ) {
            // proceed with importing the block into fork-choice
        }
        else {
            // someone beat you to it; you're done
        }
    }
    Availability::MissingComponents(..) => // you're done
}
```
any thread that needs to know the status of a block can subscribe for events:
```rust
let Some((status, mut rx)) = block_status_table.subscribe(block_hash) else {
    // block doesn't exist
}
if status.is_terminal() {
    // evaluate what to do based on status
}
while let Ok(status) = rx.recv().await {
    if status.is_terminal() {
        // decide what to do based on status
    }
}
```

beacon_node/beacon_chain/src/block_status_table.rs
```diff
@@ -0,0 +1,784 @@
...
+        block_hash: &Hash256,
+    ) -> Option<(BlockStatus, broadcast::Receiver<BlockStatus>)> {
+        // by subscribing before reading the current status, a race condition is possible where
+        // the subscribing thread receives a status and then when they await the channel, they receive
+        // the same status again. This is preferable to the reverse race condition where an old status is
+        // read and the subscriber doesn't send the update to the latest status.
+        let rx = self.notifiers.get(block_hash)?.subscribe();
+        let current_status = self.get_status(block_hash)?;
+        Some((current_status, rx))
+    }
```
dapplion, line 245: This function appears simple but the return value of `None` can have three meanings:
1. `block_hash` not in `self.notifiers`
2. `block_hash` not in `self.table`
3. Case 1 and 2

beacon_node/beacon_chain/src/block_status_table.rs
```diff
@@ -0,0 +1,784 @@
...
+/// efficient atomic operations on status transitions to prevent race conditions during
+/// block processing.
+#[derive(Debug)]
+pub struct BlockStatusTable {
+    /// The underlying concurrent hash map storing block status as AtomicU8
+    table: DashMap<Hash256, AtomicU8>,
+    /// Broadcast channels for notifying subscribers of status changes
+    notifiers: DashMap<Hash256, broadcast::Sender<BlockStatus>>,
+    /// Counter for tracking the number of entries (for metrics/debugging)
+    entry_count: AtomicUsize,
```
dapplion: Should be the initial design be the simplest approach possible?

```rust
pub enum BlockStatus {
    Pending,
    Executing(broadcast::Sender<BlockStatus>),
    Invalid,
    Importing(broadcast::Sender<BlockStatus>),
    Imported.
}

pub struct BlockStatusTable {
    table: DashMap<Hash256, BlockStatus>,
}
```

We can optimize later with more dashmaps if we happen to observe and measure performance degradation due to lock contention

beacon_node/beacon_chain/src/block_status_table.rs
```diff
@@ -0,0 +1,784 @@
...
+    ///
+    /// This operation is atomic and will only succeed if the current status
+    /// matches the expected status. Returns `true` if the swap succeeded,
+    /// `false` otherwise.
+    ///
+    /// # Arguments
+    /// * `block_hash` - The hash of the block to update
+    /// * `expected` - The expected current status
+    /// * `new` - The new status to set
+    pub fn compare_and_swap_status(
```
dapplion: This function returns `false` for two very distinct cases:
- if the block is not known
- if the expect status value doesn't match

That sounds that will make debuggability complicated if things go wrong. Why not be explicit about all possible error cases?

dapplion: I imagine in practice we just get a mutable reference to the dashmap and match over it. For example in sync

```rust
match block_status_map.get_mut(block_hash) {
    Pending => todo!?(),
    Executing(handle) => handle.await,
    Invalid => reject_block(),
    Importing(handle) => handle.await,
    Imported => complete_lookup_and_continue(),
}
```

beacon_node/beacon_chain/src/block_status_table.rs
ethDreamer: thanks! fixed in latest version :D

beacon_node/beacon_chain/src/block_status_table.rs
ethDreamer: i have changed the design to use a single dashmap and I think the code is both simpler and more efficient - let me know what you think! :D

beacon_node/beacon_chain/src/block_status_table.rs
ethDreamer: I've changed this function to return an error that will tell exactly what happened. Thanks! :D

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3018,22 +3021,31 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+        if status.is_past_pending() {
+            return Err(BlockError::DuplicateFullyImported(block_root));
         }
 
         // No need to process and import blobs beyond the PeerDAS epoch.
         if self.spec.is_peer_das_enabled_for_epoch(blob.epoch()) {
             return Err(BlockError::BlobNotRequired(blob.slot()));
         }
 
+        // TODO: there appears to be no check on the parent block being in fork choice here..
```
dapplion, line 3048: It's checked here https://github.com/sigp/lighthouse/blob/e5b4983d6baf85770fe4539a565d8a2dd462bc53/beacon_node/beacon_chain/src/blob_verification.rs#L475

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3018,22 +3021,31 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
-        if self
-            .canonical_head
-            .fork_choice_read_lock()
-            .contains_block(&block_root)
-        {
-            return Err(BlockError::DuplicateFullyImported(blob.block_root()));
+        // TODO: if somehow this blob is older than finalization, we need to not insert it - where is this checked?
+        let status = self
+            .block_status_table
+            .get_status_or_insert_pending(block_root, slot)
```
dapplion: Why insert here?

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3702,8 +3760,58 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
-                self.import_available_block(block).await
+                let block_root = block.block_root();
+                match self.block_status_table.transition(
+                    &block_root,
+                    BlockStatus::Pending,
+                    BlockStatus::Importing,
+                ) {
+                    Ok(_) => {
+                        let result = self.import_available_block(block).await;
+                        if let Ok(AvailabilityProcessingStatus::Imported(_)) = result.as_ref() {
```
dapplion: This is equivalent to a `match` with a default statement, which have been very problematic in the past as we modify the variants. Can you do a full match on result?

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3702,8 +3760,58 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+                                })?;
+                        }
+
+                        result
+                    }
+                    Err(e) => match e {
+                        // someone beat us to it - the block is already being imported
+                        BlockStatusTableError::InvalidStatusTransition { .. } => {
+                            Err(BlockError::DuplicateFullyImported(block_root))
+                        }
```
dapplion: Not true, this error will trigger if the current status is != Pending. If that status happens to be `Invalid` you will return `DuplicateFullyImported` which is incorrect

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3702,8 +3760,58 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+                        // other errors should not occur
+                        e => Err(BlockError::InternalError(format!(
```
dapplion: Avoid default match statement, just match explicitly the variants `BlockNotFound` / `InvalidStatus`. Are you sure `BlockNotFound` can never happen?

beacon_node/beacon_chain/src/block_status_table.rs
```diff
@@ -0,0 +1,938 @@
...
+}
+
+/// Represents the status of a block during the data availability import process.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[repr(u8)]
+pub enum BlockStatus {
+    /// Block (or blob) has been seen
+    Seen = 0,
+    /// Someone is running process_block() on the block
+    Processing = 1,
```
dapplion, line 27: As part of `process_block`, the block can be executing and importing, are these statuses overlapping?

dapplion:
- Closing for now, the idea is good but move to an issue until ready again

PR #8139, opened 2025-09-30
Migrate the `deposit_contract` crate to `alloy`

macladson:
## Issue Addressed

https://github.com/sigp/lighthouse/issues/6022

## Proposed Changes

Switches the `deposit_contract` crate to use the `alloy` ecosystem and removes the dependency on `ethabi`


macladson:
I think this is looking pretty good now, but I'm blocking on #8149 which we should merge first.

michaelsproul (approved):
Nice! :fire:

PR #8141, opened 2025-09-30
Move beacon processor work queue implementation to its own file

eserilev:
If we want to allocate worker threads differently for purely async tasks vs blocking tasks we will need to update how/when we queue/pop work events from the relevant queues. This PR, though purely aesthetic, might make those future changes a bit easier to implement while also "cleaning up" the beacon processor itself


macladson (approved):
LGTM!

eserilev:
@jimmygchen I'm going to build on top of this PR to make some of the beacon processor changes we talked about. I think this one is in a good spot to merge, lmk what you think

eserilev:
Heres a preview of the cpu bound and io bound worker management stuff we were talking about

https://github.com/eserilev/lighthouse/pull/23

It was actually a bit easier than I thought it'd be. I still need to double check that i categorized the work variants into io bound and cpu bound correctly, and do some local testing. Also need to think about what metrics we want to track for this change

macladson:
This is ready to be merged, shall I merge it @eserilev or do you want to wait for another set of eyes?

eserilev:
@macladson I dont think we need another set of eyes, happy to get this merged

jimmygchen:
@eserilev Ah, i think this PR reverted the fixes made in #8315 
This bug re-introduced i think: https://github.com/sigp/lighthouse/pull/8315#discussion_r2470477426

jimmygchen:
Made an issue to track the bug: https://github.com/sigp/lighthouse/issues/8648

PR #8140, opened 2025-09-30
Migrate `execution_engine_integration` to `alloy`

macladson:
## Issue Addressed

#6022 

## Proposed Changes
Migrate the `execution_engine_integration`  tests to the `alloy` ecosystem. This removes the last remaining `ethers` dependencies

PR #8149, opened 2025-10-01
Remove `ethers-core` from `execution_layer`

macladson:
## Issue Addressed

#6022
 
## Proposed Changes

Use `alloy_rpc_types::Transaction` to replace the `ethers_core::Transaction` inside the execution block generator.

## Additional Info

This along with #8139 and #8140 should remove `ethers` from the codebase entirely.
I also bump all alloy crate versions to ensure compatibility.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -140,19 +140,41 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "alloy-consensus"
-version = "0.14.0"
+version = "1.0.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c2179ba839ac532f50279f5da2a6c5047f791f03f6f808b4dfab11327b97902f"
+checksum = "3abecb92ba478a285fbf5689100dbafe4003ded4a09bf4b5ef62cca87cd4f79e"
 dependencies = [
  "alloy-eips",
  "alloy-primitives",
  "alloy-rlp",
+ "alloy-serde",
  "alloy-trie",
+ "alloy-tx-macros",
  "auto_impl 1.2.1",
+ "c-kzg",
  "derive_more 2.0.1",
  "either",
+ "k256 0.13.4",
  "once_cell",
+ "rand 0.8.5",
+ "secp256k1",
+ "serde",
+ "serde_json",
  "thiserror 2.0.12",
+]
+
+[[package]]
+name = "alloy-consensus-any"
+version = "1.0.42"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2e864d4f11d1fb8d3ac2fd8f3a15f1ee46d55ec6d116b342ed1b2cb737f25894"
+dependencies = [
+ "alloy-consensus",
+ "alloy-eips",
+ "alloy-primitives",
+ "alloy-rlp",
+ "alloy-serde",
+ "serde",
 ]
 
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -99,9 +99,10 @@ edition = "2024"
 
 [workspace.dependencies]
 account_utils = { path = "common/account_utils" }
-alloy-consensus = { version = "0.14.0", default-features = false }
-alloy-primitives = { version = "1.0", features = ["rlp", "getrandom"] }
-alloy-rlp = "0.3.4"
+alloy-consensus = { version = "=1.0.37", default-features = false }
+alloy-primitives = { version = "=1.4.0", features = ["rlp", "getrandom"] }
+alloy-rlp = "=0.3.12"
+alloy-rpc-types = "=1.0.37"
```
macladson: I was running into issues with compatibility so I've bumped (and pinned) the alloy crates to the same/compatible versions. When we go to update them, we can update them all at the same time to ensure we don't break compatibility. An alternative is to just import `alloy` and enable the features we need, but I suspect we'll end up with unnecessary bloat if we go down that path.

Cargo.lock
```diff
@@ -1043,6 +1313,22 @@ version = "0.8.0"
...
+
+[[package]]
+name = "bitcoin_hashes"
+version = "0.14.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bb18c03d0db0247e147a21a6faafd5a7eb851c743db062de72018b6b7e8e4d16"
+dependencies = [
+ "bitcoin-io",
+ "hex-conservative",
+]
```
michaelsproul, line 1272: I'm suss on these bitcoin deps. I don't see why they're needed, as it seems they only get turned on by the `secp256k1` feature of Alloy, which I wouldn't think we'd need?

macladson: I've added some stricter feature bounds, but `secp256k1` doesn't seem to want to leave the lockfile. It's an optional dep on `alloy-consensus` (and we aren't using the feature), and it doesn't look like we ever build it or its `bitcoin` dependencies. Not really sure what's going on there, I tried everything I could think of to remove it. Even if I manually edit the lockfile to remove it, Cargo automatically adds it again

michaelsproul: I'll have a crack at it

michaelsproul: Found the culprit:

https://github.com/alloy-rs/alloy/blob/808b88df4d627e98b3a3ea55ab54d96d264a23bb/crates/consensus/Cargo.toml#L94

Combined with this Cargo bug:

- https://github.com/rust-lang/cargo/issues/10801

So we aren't actually depending on `secp256k1` or these `bitcoin_hashes` crates, but Cargo includes them in the lockfile because of a bug. `cargo tree` _does not_ show these deps (hence the confusing output), but `cargo metadata | jq | less` _does_.

michaelsproul: Actually I think it's this line, as we're not using the `std` feature of `alloy-consensus`, but we _are_ using the `serde` feature:

https://github.com/alloy-rs/alloy/blob/808b88df4d627e98b3a3ea55ab54d96d264a23bb/crates/consensus/Cargo.toml#L123

macladson: Nice find! That's definitely it. Doesn't seem like we can do much about it though

michaelsproul: Yeah I think we just need to live with this.

michaelsproul (approved):

Issue #7638, opened 2025-06-23
Add PostgresDB as an optional DB implementation for the Beacon Node


owanikin:
The Lighthouse beacon node backend has been previously abstracted to handle different database implementations. Currently the two enabled database implementations are LevelDB and Redb. LevelDB is currently the most performant DB implementation, but we've seen issues where a LevelDB database hosted on a NFS (network file system) can get corrupted. This is because LevelDB (and probably Redb as well) is not designed to handle concurrent access and other inconsistencies/latencies introduced by NFS. Postgres, with the correct settings should be able to run on an NFS without these risks.

PR #8157, opened 2025-10-02
Postgres DB support for Lighthouse

owanikin:
## Issue Addressed
#7638 

Lays the groundwork for adding Postgres as an optional database backend.

## Proposed Changes

Added Postgres dependencies (`sqlx, tokio`) under a new postgres feature.

Defined Cargo feature alias `beacon-node-postgres` for Lighthouse builds.

Updated Makefile:

Added `beacon-node-postgres` to `CROSS_FEATURES` and `lint-full`.

Introduced a helper target (`make beacon-node-postgres`) to spin up Postgres in Docker and run Lighthouse with the Postgres feature enabled.


dapplion:
Thanks for the contribution! Closing for stale

PR #8158, opened 2025-10-02
Automatically pass spans into blocking handles

eserilev:
Currently we have to manually pass spans across threads (i.e. during `spawn_blocking_handle` invocations). This PR ensures that the current span is always passed across threads for:
- `spawn_blocking_handle`
- `spawn_blocking_with_rayon_async`
- `spawn_blocking_with_rayon`

If there is no current span, nothing will happen


beacon_node/lighthouse_tracing/src/lib.rs
```diff
@@ -29,7 +29,7 @@ pub const SPAN_PROCESS_CHAIN_SEGMENT: &str = "process_chain_segment";
 pub const SPAN_PROCESS_CHAIN_SEGMENT_BACKFILL: &str = "process_chain_segment_backfill";
 
 /// Fork choice root spans
-pub const SPAN_RECOMPUTE_HEAD: &str = "recompute_head_at_slot";
+pub const SPAN_RECOMPUTE_HEAD: &str = "recompute_head_at_slot_internal";
```
eserilev: I think I can delete this

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -2838,12 +2838,8 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
 
         // Filter uninteresting blocks from the chain segment in a blocking task.
         let chain = self.clone();
-        let filter_chain_segment = debug_span!("filter_chain_segment");
```
eserilev, line 2891: not sure if there's a parent span here, might need to keep the `debug_span!`

beacon_node/lighthouse_tracing/src/lib.rs
jimmygchen: Do you mean deleting this from the root spans? Why is this `_internal`?

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Ah yeah, I believe the span was used for instrumenting the `filter_chain_segment` function, but we can now just do this within the closure or at the `filter_chain_segment` function level.

The parent span should be created at the [`process_block`](https://github.com/sigp/lighthouse/blob/ff8b514b3f012537de8b99cd526d15bdb4610698/beacon_node/network/src/network_beacon_processor/sync_methods.rs#L583) function.

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -4483,15 +4467,10 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         //
         // Load the parent state from disk.
         let chain = self.clone();
-        let span = Span::current();
         let (state, state_root_opt) = self
             .task_executor
             .spawn_blocking_handle(
-                move || {
-                    let _guard =
-                        debug_span!(parent: span, "load_state_for_block_production").entered();
```
jimmygchen, line 4539: I think we want to keep this span, but we no longer need to pass the parent span explicitly

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -5047,13 +5026,10 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             .graffiti_calculator
             .get_graffiti(validator_graffiti)
             .await;
-        let span = Span::current();
         let mut partial_beacon_block = self
             .task_executor
             .spawn_blocking_handle(
                 move || {
-                    let _guard =
-                        debug_span!(parent: span, "produce_partial_beacon_block").entered();
```
jimmygchen, line 4971: same here,  be good to keep this or move the instrumentation to `produce_partial_beacon_block`

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -5089,14 +5065,10 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             match block_contents_type {
                 BlockProposalContentsType::Full(block_contents) => {
                     let chain = self.clone();
-                    let span = Span::current();
                     let beacon_block_response = self
                         .task_executor
                         .spawn_blocking_handle(
                             move || {
-                                let _guard =
-                                    debug_span!(parent: span, "complete_partial_beacon_block")
```
jimmygchen, line 5013: Same as above

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -5113,14 +5085,10 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
                 }
                 BlockProposalContentsType::Blinded(block_contents) => {
                     let chain = self.clone();
-                    let span = Span::current();
                     let beacon_block_response = self
                         .task_executor
                         .spawn_blocking_handle(
                             move || {
-                                let _guard =
-                                    debug_span!(parent: span, "complete_partial_beacon_block")
```
jimmygchen, line 5037: Same as above

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -5138,13 +5106,10 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             }
         } else {
             let chain = self.clone();
-            let span = Span::current();
             let beacon_block_response = self
                 .task_executor
                 .spawn_blocking_handle(
                     move || {
-                        let _guard =
-                            debug_span!(parent: span, "complete_partial_beacon_block").entered();
```
jimmygchen, line 5062: Same as above

beacon_node/beacon_chain/src/canonical_head.rs
```diff
@@ -513,21 +511,13 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
     /// situation can be rectified. We avoid returning an error here so that calling functions
     /// can't abort block import because an error is returned here.
     pub async fn recompute_head_at_slot(self: &Arc<Self>, current_slot: Slot) {
-        let span = info_span!(
```
jimmygchen, line 532: Not sure why this need to be deleted, were you thinking to create the spans in  `spawn_blocking_handle` automatically using the `task_name`?

It doesn't look like it's creating this span though

beacon_node/http_api/src/publish_blocks.rs
```diff
@@ -370,7 +365,6 @@ fn spawn_build_data_sidecar_task<T: BeaconChainTypes>(
                 let Some((kzg_proofs, blobs)) = proofs_and_blobs else {
                     return Ok((vec![], vec![]));
                 };
-                let _guard = debug_span!(parent: current_span, "build_data_sidecars").entered();
```
jimmygchen, line 371: think we should keep this one too?

jimmygchen (changes requested):
Nice, i think this definitely makes it easier to instrument code that spawns a blocking thread!

I wasn't sure what was the intention of removing some of the spans though, were you thinking to create them in the `spawn` functions, because we'll lose some traces here.

jimmygchen:
Hi @eserilev would you mind looking at the review comments and let me know if they make sense?
I think it'd still be useful to automatically pass spans into blocking handles.

beacon_node/lighthouse_tracing/src/lib.rs
eserilev: reverted this change

beacon_node/beacon_chain/src/beacon_chain.rs
eserilev: span is instrumented via attribute macro on `fn load_state_for_block_production`

beacon_node/beacon_chain/src/beacon_chain.rs
eserilev: span is instrumented via attribute macro on `fn complete_partial_beacon_block `

beacon_node/beacon_chain/src/beacon_chain.rs
eserilev: span is instrumented via attribute macro on fn complete_partial_beacon_block

eserilev:
sorry for leaving this hanging for so long, I've updated this PR based on your feedback

jimmygchen:
Thanks @eserilev I'll review this today!

jimmygchen:
@eserilev sorry for leaving this PR stale again 😂 

I'll work on getting this up to date again - there are a few calls in `gloas.rs` that needs update as well.

jimmygchen (approved):
I've merged `unstable` into this branch and applied the changes on a few new code paths. The remaining changes look good! 

@eserilev would you mind reviewing my latest changes before we merge this?

pawanjay176 (approved):

jimmygchen:
@mergify requeue

PR #8163, opened 2025-10-06
Fallback to getPayload v1 if v2 fails

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Post fulu, we should be calling the v2 api on the relays that doesn't return the blobs/data columns.

However, we decided to start hitting the v2 api as soon as fulu is scheduled to avoid unexpected surprises at the fork.
In the ACDT call, it seems like most clients are calling v2 only after the fulu fork. 
This PR aims to be the best of both worlds where we fallback to hitting v1 api if v2 fails. This way, we know beforehand if relays don't support it and can potentially alert them.

### DIFF
--- a/beacon_node/execution_layer/src/lib.rs
+++ b/beacon_node/execution_layer/src/lib.rs
@@ -1914,9 +1914,19 @@ impl<E: EthSpec> ExecutionLayer<E> {
     ) -> Result<SubmitBlindedBlockResponse<E>, Error> {
         debug!(?block_root, "Sending block to builder");
         if spec.is_fulu_scheduled() {
-            self.post_builder_blinded_blocks_v2(block_root, block)
+            let resp = self
+                .post_builder_blinded_blocks_v2(block_root, block)
                 .await
-                .map(|()| SubmitBlindedBlockResponse::V2)
+                .map(|()| SubmitBlindedBlockResponse::V2);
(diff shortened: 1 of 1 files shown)


michaelsproul (approved):
LGTM

beacon_node/execution_layer/src/lib.rs
```diff
@@ -1914,9 +1914,19 @@ impl<E: EthSpec> ExecutionLayer<E> {
...
         if spec.is_fulu_scheduled() {
-            self.post_builder_blinded_blocks_v2(block_root, block)
+            let resp = self
+                .post_builder_blinded_blocks_v2(block_root, block)
                 .await
-                .map(|()| SubmitBlindedBlockResponse::V2)
+                .map(|()| SubmitBlindedBlockResponse::V2);
+            // Fallback to v1 if v2 fails because the relay doesn't support it.
+            // Note: we should remove the fallback post fulu when all relays have support for v2.
+            if resp.is_err() {
```
jimmygchen, line 1923: minor note here: the check here strictly requires `202`. This is spec compliant but i feel like it's unnecessarily strict, and if any builder interface is not fully compliant e.g. returning 200, we'll end up sending v1 - this isn't an issue though.

But I feel like requiring a specific 2xx status code is too strict in general - it's meant to provide more granular info about a *successful* response - a 2xx response ideally should not be treated as a failure.

PR #8167, opened 2025-10-07
Quote `BeaconState::proposer_lookahead` in JSON repr

michaelsproul:
## Proposed Changes

Use quoted integers for `state.proposer_lookahead` when serializing JSON. This is standard for all integer fields, but was missed for the newly added proposer lookahead. I noticed this issue while inspecting the head state on a local devnet.

I'm glad we found this before someone reported it :P


eserilev (approved):
nice catch!

Issue #8046, opened 2025-09-15
Validator client does not ask for committee index 0 on attestation data request


KaloyanTanev:
Post electra validator clients are expected to ask only for attestation data with committee index 0.

It seems like [here](https://github.com/sigp/lighthouse/blob/fd10b632740e441354c4547a12727c29dd288239/validator_client/validator_services/src/attestation_service.rs#L355) Lighthouse should ask for 0, instead of the actual `committee_index`. Ideally, in cases where there are multiple keys serviced by the validator client that need to attest in the given slot, it should not repeat multiple times the same request with the same data.

N.B.: Currently, the validator client is getting successful responses, because the beacon nodes are usually disregarding this field and are simply returning the data. However, when the validator client is managing a lot of keys, i.e.: 960, it is sending on average 30x more requests to the beacon node than necessary.

michaelsproul:
Thanks, we were aware of this but didn't have an issue open. It could be a good one to prioritise after Fusaka

chong-he:
I would like to work on this

chong-he:
Closed by #8171

PR #8171, opened 2025-10-08
Always use committee index 0 when getting attestation data

chong-he:
## Issue Addressed
* #8046 

## Proposed Changes

Split the function `publish_attestations_and_aggregates` into `publish_attestations` and `handle_aggregates`, so that for attestations, only 1 task is spawned.

## Additional Info

Tested with Kurtosis on `minimal` preset, which has a `max_committees_per_slot` of 4.  The number of times the VC calls the BN for the endpoint `/eth/v1/validator/attestation_data` is given by the metric: `http_api_paths_times_count{path="v1/validator/attestation_data"}`

For 100 slots of data:
Before: `http_api_paths_times_count{path="v1/validator/attestation_data"} 398`
After: `http_api_paths_times_count{path="v1/validator/attestation_data"} 100`

From the test output, we see that the number of calls to the BN is reduced. On mainnet, the `max_committee_per_slot` is 64, the VC should only call the BN only once per slot for this endpoint


validator_client/validator_services/src/attestation_service.rs
```diff
@@ -180,23 +180,64 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> AttestationService<S,
...
 
+        // Create and publish an `Attestation` for all validators only once
+        // as the committee_index is the same (index 0) post-Electra
+        let attestation_duties: Vec<_> = self.duties_service.attesters(slot).into_iter().collect();
+
+        let attestation_data_cache = Arc::new(tokio::sync::RwLock::new(None));
+        let attestation_data_cache_clone = attestation_data_cache.clone();
+
+        let attestation_service = self.clone();
+        self.inner.executor.spawn_ignoring_error(
```
michaelsproul: Could use `spawn_handle` and then `await` the handle to get the attestation data from inside the future (rather than using the `attestation_data_cache` `RwLock`).

michaelsproul: In case the handle doesn't return the result, it's an error and we probably need to just stop the attestation process for this slot (return `Err(..)`)

validator_client/validator_services/src/attestation_service.rs
```diff
@@ -372,7 +388,6 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> AttestationService<S,
                     validator = ?duty.pubkey,
                     duty_slot = %duty.slot,
                     attestation_slot = %attestation_data.slot,
-                    duty_index = duty.committee_index,
```
michaelsproul: Could keep these logs (these won't be 0, because the duty still contains the real committee index).

validator_client/validator_services/src/attestation_service.rs
```diff
@@ -507,7 +520,8 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> AttestationService<S,
             ),
         }
 
-        Ok(Some(attestation_data))
+        drop(attestations_timer);
```
michaelsproul: This drop probably not needed (can just rename the timer to `_attestations_timer` and it will drop when it goes out of scope, when the function finishes)

validator_client/validator_services/src/attestation_service.rs
chong-he: Thank you for the comment, I revised in [0967696](https://github.com/sigp/lighthouse/pull/8171/commits/0967696831ce1c666baa6e85137c69483396f1d9)

validator_client/validator_services/src/attestation_service.rs
chong-he: Good point, revised in [0967696](https://github.com/sigp/lighthouse/pull/8171/commits/0967696831ce1c666baa6e85137c69483396f1d9)

validator_client/validator_services/src/attestation_service.rs
chong-he: Put a underscore before the timer and removed this line: [0967696](https://github.com/sigp/lighthouse/pull/8171/commits/0967696831ce1c666baa6e85137c69483396f1d9)

dapplion:
This feature conflicted in the past with DVT setups which rely on the committee index. Is that not the case anymore after Electra?

chong-he:
> This feature conflicted in the past with DVT setups which rely on the committee index. Is that not the case anymore after Electra?

Not sure which issue are you referring to about the DVT conflict? 

For post-Electra, when VC requests to the BN, it must have a `committee_index` of 0: 
https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Validator/produceAttestationData

KaloyanTanev:
@dapplion I can speak of Charon here. We did have conflicts indeed, but it was fixed in v1.4.3 (~5 months ago), a week after Electra fork.

Edit: To be precise, we had issues when a DV cluster had mix of VCs, some asking for attestation data with committee index 0, while others for the actual committee index.

validator_client/validator_services/src/attestation_service.rs
```diff
@@ -207,24 +247,43 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> AttestationService<S,
...
+                // Log an error if the handle fails and return, skipping aggregates
+                let Some(handle) = attestation_data_handle else {
+                    error!(slot = slot.as_u64(), "Failed to spawn attestation task");
+                    return;
+                };
+
+                let Ok(Some(Ok(attestation_data))) = handle.await else {
+                    error!(slot = slot.as_u64(), "Failed to spawn attestation task");
+                    return;
+                };
```
macladson: It's a little spicy but we could combine these two statements into one like this:


```suggestion
                let Some(attestation_data) =
                    async { attestation_data_handle?.await.ok()?.and_then(Result::ok) }.await
                else {
                    error!(slot = slot.as_u64(), "Failed to spawn attestation task");
                    return;
                };
```

macladson (commented):
Do we have to handle attestations pre-Electra as well? Seems like we would only request committee 0 even pre-Electra, but I could be misunderstanding the fork flow.

Note that this will spawn twice as many tasks (one for attestations, one for aggregates) but I actually think that in this case it will improve performance despite the context-switching overhead since these tasks are likely IO bound and so can yield quicker than before. (Plus of course the huge reduction in BN calls makes this significantly better of course)

eserilev:
> Do we have to handle attestations pre-Electra as well? Seems like we would only request committee 0 even pre-Electra, but I could be misunderstanding the fork flow.

no we don't need to handle constructing pre-electra attestations. we actually removed a lot of pre-electra attestation logic when we moved to `SingleAttestation`

validator_client/validator_services/src/attestation_service.rs
chong-he: This is elegant! I didn't know how to handle that before so I separate into 2 step but I know it's ugly, thanks for the comment

michaelsproul:
Seeing very high values for the `vc_attestation_service_task_times{attestations_http_get}` metric on Hoodi (like 4s+ for the 99th percentile, and 1.5-2s for the median). I think this PR could help speed that up a lot (by reducing the number of requests), so will prioritise review and merge.

michaelsproul:
This change did help a lot with reducing the time the VC needs to fetch attestations from the BN. Data taken from a Hoodi VC with 10k keys, connected to an overworked BN.



http-get-times

michaelsproul:
Marking this as backwards-incompat, as it will technically break pre-Electra devnets.

eserilev (approved):
Nice work, LGTM!!

michaelsproul (approved):

PR #8172, opened 2025-10-08
Remove `context_deserialize` and import from crates.io

macladson:
## Proposed Changes

Use the recently published `context_deserialize` and remove it from Lighthouse

## Additional Info

While we can technically merge this, I am going to keep this blocked until I have pushed a new version of `context_deserialize` to remove impls on foreign  types (`ssz_types`, `milhouse`, etc) and implemented it in those crates instead. This will reduce ongoing maintenance since we will not have to continually bump the deps in `context_deserialize`.

- https://github.com/sigp/ssz_types/pull/73
- https://github.com/sigp/ethereum_ssz/pull/61
- https://github.com/sigp/milhouse/pull/79
- https://github.com/sigp/context_deserialize/pull/1


macladson:
All deps have been bumped and this is ready to go now!

Cargo.lock
```diff
@@ -1233,7 +1233,7 @@ dependencies = [
  "educe",
  "eth2",
  "eth2_network_config",
- "ethereum_hashing",
+ "ethereum_hashing 0.7.0",
```
michaelsproul: Uh, we have duplication of `ethereum_hashing` :(

Probably ok for now, but we should fix this

I think this is from ssz_types bumping to 0.8 but other crates not bumping

michaelsproul: Actually, I think this is fine, just need to bump in LH.

michaelsproul (changes requested):
Needs a bump on ethereum_hashing (see above)

Cargo.lock
macladson: Good catch!

michaelsproul (approved):

PR #8174, opened 2025-10-08
Bump kzg library versions

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Update c-kzg and rust-eth-kzg to their latest versions. Also removes the patch version hardcoding in Cargo.toml.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -117,7 +117,7 @@ byteorder = "1"
 bytes = "1"
 # Turn off c-kzg's default features which include `blst/portable`. We can turn on blst's portable
 # feature ourselves when desired.
-c-kzg = { version = "2.1.0", default-features = false }
+c-kzg = { version = "2.1", default-features = false }
 cargo_metadata = "0.19"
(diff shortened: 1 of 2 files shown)


jimmygchen (approved):
LGTM, thanks!

Cargo.toml
```diff
@@ -117,7 +117,7 @@ byteorder = "1"
 bytes = "1"
 # Turn off c-kzg's default features which include `blst/portable`. We can turn on blst's portable
 # feature ourselves when desired.
-c-kzg = { version = "2.1.0", default-features = false }
+c-kzg = { version = "2.1", default-features = false }
```
michaelsproul, line 120: This doesn't actually do anything btw, a `2.1.0` requirement is identical to a `2.1`.

Both allow versions `>= 2.1.0 && < 3.0.0`.

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#default-requirements

pawanjay176: Makes sense. I just thought it was confusing to have the explicit patch version in the Cargo.toml (2.1.0) but having a higher patch version in the lockfile (2.1.5).

Issue #8147, opened 2025-10-01
Fork choice write lock held during database write


michaelsproul:
## Description

Currently we hold the fork choice write lock while writing to disk:

https://github.com/sigp/lighthouse/blob/e5b4983d6baf85770fe4539a565d8a2dd462bc53/beacon_node/beacon_chain/src/beacon_chain.rs#L4000-L4002

This poses a potential problem for fork choice lock readers, which will be blocked while waiting for this write to complete. We expect that writing to disk takes hundreds of milliseconds several seconds due to diff computation and overhead due to database I/O.

## Metrics

Metrics seem to indicate that this is not a significant performance issue:

```
# HELP beacon_fork_choice_read_lock_aquire_seconds Time taken to aquire the fork-choice read lock
# TYPE beacon_fork_choice_read_lock_aquire_seconds histogram
beacon_fork_choice_read_lock_aquire_seconds_bucket{le="0.0001"} 1025501427
beacon_fork_choice_read_lock_aquire_seconds_bucket{le="0.0004"} 1086019739
beacon_fork_choice_read_lock_aquire_seconds_bucket{le="0.0016"} 1254510271
beacon_fork_choice_read_lock_aquire_seconds_bucket{le="0.0064"} 1313781698
beacon_fork_choice_read_lock_aquire_seconds_bucket{le="0.0256"} 1316301315
beacon_fork_choice_read_lock_aquire_seconds_bucket{le="0.1024"} 1316615602
beacon_fork_choice_read_lock_aquire_seconds_bucket{le="0.4096"} 1316673956
beacon_fork_choice_read_lock_aquire_seconds_bucket{le="+Inf"} 1316682131
beacon_fork_choice_read_lock_aquire_seconds_sum 378137.6061273879
beacon_fork_choice_read_lock_aquire_seconds_count 1316682131
```

TODO: more analysis here

## Steps to resolve

Please describe the steps required to resolve this issue, if known.

dapplion:
Will provide more evidence but our early conclusion is:
- The possibility of holding the write lock for long real and can slow down reads
- However, metrics show that this happens very rarely
- So we will leave the code as is. The invariants we get from making disk and fork-choice writes atomic are very nice and don't want to lose them.

jimmygchen:
I noticed a bit of mismatch between traces and metrics on holesky nodes. 

Holesky node metrics over the last 24 hours, looks pretty good

[...]

PR #8176, opened 2025-10-09
Use OnceCell in HotHDiffBufferCache

michaelsproul:
## Issue Addressed

Partly addresses:

- https://github.com/sigp/lighthouse/issues/8147

## Proposed Changes

Use an `Arc` to avoid holding the `state_cache` lock while `HDiffBuffer::from_state` runs. This conversion function can take up to 1s, so it's good if we can release the lock while it runs. This can unblock other threads waiting on the lock.

### DIFF
--- a/beacon_node/store/src/state_cache.rs
+++ b/beacon_node/store/src/state_cache.rs
@@ -114,14 +114,10 @@ impl<E: EthSpec> StateCache<E> {
             drop(inner);
 
             metrics::inc_counter_vec(&metrics::STORE_BEACON_HDIFF_BUFFER_CACHE_HIT, HOT_METRIC);
-            let timer =
+            let _timer =
                 metrics::start_timer_vec(&metrics::BEACON_HDIFF_BUFFER_CLONE_TIME, HOT_METRIC);
-            let result = Some(buffer.clone());
-            drop(timer);
-            return result;
-        }
-
-        if let Some(state) = inner.get_by_state_root(state_root) {
+            Some(buffer.clone())
+        } else if let Some(state) = inner.get_by_state_root(state_root) {
             // Drop the lock before performing the conversion from state to buffer, as this
             // conversion is somewhat expensive.
             drop(inner);
@@ -132,10 +128,11 @@ impl<E: EthSpec> StateCache<E> {
 
             // Put the buffer back into the cache without re-locking, by using the OnceCell.
             buffer_cell.get_or_init(|| (slot, buffer.clone()));
-            return Some(buffer);
+            Some(buffer)
+        } else {
+            metrics::inc_counter_vec(&metrics::STORE_BEACON_HDIFF_BUFFER_CACHE_MISS, HOT_METRIC);
+            None
         }
-        metrics::inc_counter_vec(&metrics::STORE_BEACON_HDIFF_BUFFER_CACHE_MISS, HOT_METRIC);
-        None
     }
 }
 
--- a/beacon_node/store/src/state_cache.rs
+++ b/beacon_node/store/src/state_cache.rs
@@ -104,11 +104,7 @@ impl<E: EthSpec> StateCache<E> {
     pub fn get_hdiff_buffer_by_state_root(&self, state_root: Hash256) -> Option<HDiffBuffer> {
         let mut inner = self.inner.lock();
 
-        let buffer_cell = inner
-            .hdiff_buffers
-            .hdiff_buffers
-            .get_or_insert(state_root, || Arc::new(OnceCell::new()))
-            .clone();
+        let buffer_cell = inner.hdiff_buffers.hdiff_buffers.get(&state_root).clone();
 
         if let Some((_, buffer)) = buffer_cell.get() {
             drop(inner);
@@ -118,16 +114,25 @@ impl<E: EthSpec> StateCache<E> {
                 metrics::start_timer_vec(&metrics::BEACON_HDIFF_BUFFER_CLONE_TIME, HOT_METRIC);
             Some(buffer.clone())
         } else if let Some(state) = inner.get_by_state_root(state_root) {
+            // Insert an empty cell to the cache. Use get_or_insert to get a reference of the
+            // inserted value.
+            let buffer_cell = inner
+                .hdiff_buffers
+                .hdiff_buffers
+                .get_or_insert(&state_root, || Arc::new(OnceCell::new()))
+                .clone();
             // Drop the lock before performing the conversion from state to buffer, as this
             // conversion is somewhat expensive.
             drop(inner);
 
             metrics::inc_counter_vec(&metrics::STORE_BEACON_HDIFF_BUFFER_CACHE_HIT, HOT_METRIC);
-            let slot = state.slot();
-            let buffer = HDiffBuffer::from_state(state);
 
             // Put the buffer back into the cache without re-locking, by using the OnceCell.
-            buffer_cell.get_or_init(|| (slot, buffer.clone()));
+            let (_, buffer) = buffer_cell.get_or_init(|| {
+                let slot = state.slot();
+                let buffer = HDiffBuffer::from_state(state);
+                (slot, buffer.clone())
+            });
             Some(buffer)
         } else {
             metrics::inc_counter_vec(&metrics::STORE_BEACON_HDIFF_BUFFER_CACHE_MISS, HOT_METRIC);
--- a/beacon_node/store/src/state_cache.rs
+++ b/beacon_node/store/src/state_cache.rs
@@ -58,14 +58,11 @@ pub struct StateCacheInner<E: EthSpec> {
 
 /// Cache of hdiff buffers for hot states.
 ///
-/// This cache only keeps buffers prior to the finalized state, which are required by the
-/// hierarchical state diff scheme to construct newer unfinalized states.
-///
 /// The cache always retains the hdiff buffer for the most recent snapshot so that even if the
 /// cache capacity is 1, this snapshot never needs to be loaded from disk.
 #[derive(Debug)]
 pub struct HotHDiffBufferCache {
-    /// Cache of HDiffBuffers for states *prior* to the `finalized_state`.
+    /// Cache of HDiffBuffers.
     ///
     /// Maps state_root -> (slot, buffer).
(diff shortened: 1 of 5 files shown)


beacon_node/store/src/state_cache.rs
```diff
@@ -77,14 +86,67 @@ pub enum PutStateOutcome {
...
+            metrics::inc_counter_vec(&metrics::STORE_BEACON_HDIFF_BUFFER_CACHE_HIT, HOT_METRIC);
+            let slot = state.slot();
+            let buffer = HDiffBuffer::from_state(state);
+
+            // Put the buffer back into the cache without re-locking, by using the OnceCell.
+            buffer_cell.get_or_init(|| (slot, buffer.clone()));
+            return Some(buffer);
+        }
+        metrics::inc_counter_vec(&metrics::STORE_BEACON_HDIFF_BUFFER_CACHE_MISS, HOT_METRIC);
+        None
```
michaelsproul: Could consider deleting the once cell in this case, as we've created an entry and then left it empty. Alternatively we could return it for the caller to fill in once they load the state from disk.

beacon_node/store/src/state_cache.rs
```diff
@@ -77,14 +86,67 @@ pub enum PutStateOutcome {
...
+            .clone();
+
+        if let Some((_, buffer)) = buffer_cell.get() {
+            drop(inner);
+
+            metrics::inc_counter_vec(&metrics::STORE_BEACON_HDIFF_BUFFER_CACHE_HIT, HOT_METRIC);
+            let timer =
+                metrics::start_timer_vec(&metrics::BEACON_HDIFF_BUFFER_CLONE_TIME, HOT_METRIC);
+            let result = Some(buffer.clone());
+            drop(timer);
```
dapplion: Is it necessary to drop timer explicitly here?

dapplion: Opinionated, but it's easier to read to me and less lines using `if else`. I have the diff in this commit 9209c57b7

beacon_node/store/src/state_cache.rs
```diff
@@ -455,7 +501,8 @@ impl HotHDiffBufferCache {
     pub fn put(&mut self, state_root: Hash256, slot: Slot, buffer: HDiffBuffer) -> bool {
         // If the cache is not full, simply insert the value.
         if self.hdiff_buffers.len() != self.hdiff_buffers.cap().get() {
-            self.hdiff_buffers.put(state_root, (slot, buffer));
+            self.hdiff_buffers
+                .put(state_root, Arc::new(OnceCell::with_value((slot, buffer))));
```
dapplion: There's a race condition here

- thread 1: get or insert: insert empty one 
- thread 1: clone new once cell
- thread 1: start computing buffer
- thread 2: put buffer, replacing previous created once cell and wasting its compute
- thread 1: completes computing the buffer and the once cell is dropped

I don't feel it's a serious issue but noting it here

beacon_node/store/src/state_cache.rs
dapplion: I think it's problematic to populate the LRU cache with empty cells, it can cause it to evict valuable entries for no reason. I have this commit where we only insert if necessary without much more complexity b6bd9b384

beacon_node/store/src/state_cache.rs
michaelsproul: I guess technically not. The timer is meant to measure just the clone, which is why it's written like this. With your change, we'll also measure the function post-amble where all of the local variables get dropped. However this should be completely negligible in terms of time, so I think it's fine

beacon_node/store/src/state_cache.rs
michaelsproul: Even better, I think we can just remove the `OnceCell`. I did this in this commit: a3d53548e3e6adb1369345001426825fbfe6bfaf.

The reason being, the buffer cache is only meant to contain states prior to finalization, so we should not be priming it after loading a full state (which would be post finalization). In future we may alter this paradigm, but for now, just an `Arc` is enough.

michaelsproul: https://github.com/sigp/lighthouse/blob/a3d53548e3e6adb1369345001426825fbfe6bfaf/beacon_node/store/src/state_cache.rs#L58-L66

beacon_node/store/src/state_cache.rs
michaelsproul: Gone with removal of OnceCell, I think?

michaelsproul: Actually I think a similar race still exists, we could have two threads concurrently trying to compute and put the same hdiff buffer. However this is not a new issue in this PR, and IMO probably happens quite infrequently.

If we do some more measurements and find that this PR still reduces our state lock wait times and fork choice lock wait times, I think we should merge it. It's an incremental improvement.

michaelsproul:
The version of this branch with the Arc performs substantially worse, and I'm trying to understand why before I go back to the `OnceCell` approach to clean it up (I wouldn't want to break the fix).

The main differences in the `OnceCell` approach were:

1. We put the buffers for cached hot states into the hdiff cache after computing them (see [here](https://github.com/sigp/lighthouse/blob/b1816d074cba8c90dc5f937b8a9f8feaee619517/beacon_node/store/src/state_cache.rs#L133-L134)). In the `Arc` design & previously, only buffers for states from before finalization were allowed in the cache.
2. We allowed empty entries in the cache in the case of a cache miss (this should be detrimental, if anything).

One hypothesis is that the buffers for the hot states (added by (1)) are very useful, and this reduces IO write time and fork choice write lock hogging. However the metrics do not seem to bear this out? First, the graph showing the drop in fork choice write waits running OnceCell:

fork_choice_writes

During that flat period from around 2025-10-09 to 2025-10-15 we were running the OnceCell build. And it has a noticeable impact!

However during the same period, block import times are not really affected. If anything the max spike is higher, while the average spike is smaller:

import_delay

But maybe this is still good...

We also see a _big decrease_ in median Hdiff buffer load time:

load_hdiff_buffer

Other evidence is also consistent with the hot hdiff buffers being useful:

- We have a node with a larger hdiff buffer cache (the default is 1), which performs no better than its peers. This implies that more pre-finalization buffers are not really that useful.

So I think we want to loosen the requirement that only buffers for pre-finalization states are allowed, which probably means introducing the OnceCell.

I'm out of time for today but can come back to this next week. @dapplion if you have time to pick it up, you could start by reverting my three most recent commits (everything after ab4d1d39574ce7634cdc2105706dc680cff68baf).

dapplion:
Let's keep the OnceCell approach. The metrics are clear, using `max_over_time(beacon_block_delay_imported_time)` renders true spikes. I have added a commit to:
- Not leave empty cells in the map
- Also wait and get in the first branch of the if, in case some other thread is computing the value at that time

beacon_node/store/src/state_cache.rs
```diff
@@ -77,14 +86,89 @@ pub enum PutStateOutcome {
...
+    }
+
+    pub fn get_hdiff_buffer_by_state_root(&self, state_root: Hash256) -> Option<HDiffBuffer> {
+        let mut inner = self.inner.lock();
+
+        if let Some((_, buffer)) = inner
+            .hdiff_buffers
+            .hdiff_buffers
+            .get(&state_root)
+            .and_then(|cell| cell.get_or_try_init(|| Err(())).ok())
```
michaelsproul: Isn't this just the same as calling .get on the cell?

michaelsproul: Just noting that by calling .get before dropping inner we are potentially blocking while holding the mutex (if another thread is computing). However this probably isn't too much of an issue in practice

beacon_node/store/src/state_cache.rs
```diff
@@ -60,7 +68,8 @@ pub struct HotHDiffBufferCache {
     /// Cache of HDiffBuffers for states *prior* to the `finalized_state`.
     ///
     /// Maps state_root -> (slot, buffer).
-    hdiff_buffers: LruCache<Hash256, (Slot, HDiffBuffer)>,
+    // FIXME(sproul): put OnceCell around HDiffBuffer only?
+    hdiff_buffers: LruCache<Hash256, Arc<OnceCell<(Slot, HDiffBuffer)>>>,
```
michaelsproul: wdyt about moving the once cell just around the buffer? this would clean up some of the slot-based logic which I think is a bit dodgy now

beacon_node/store/src/state_cache.rs
```diff
@@ -77,14 +86,79 @@ pub enum PutStateOutcome {
...
+
+            let buffer_cell = inner
+                .hdiff_buffers
+                .hdiff_buffers
+                .get_or_insert(state_root, || Arc::new(OnceCell::new()))
+                .clone();
+            // Drop the lock before performing the conversion from state to buffer, as this
+            // conversion is somewhat expensive.
+            // Drop after setting the metric to minimize the time the OnceCell is in INCOMPLETE
+            // state
```
michaelsproul, line 133: Which metric is this referring to @dapplion?

michaelsproul:
Dropping from v8.0.0 as this only demonstrated a benefit on 1/5 Holesky nodes we tested on.

beacon_node/store/src/state_cache.rs
dapplion: `metrics::inc_counter_vec(&metrics::STORE_BEACON_HDIFF_BUFFER_CACHE_HIT, HOT_METRIC);`

michaelsproul: Ok, I was confused because that metric isn't relevant to the lock at all really. Like it's completely arbitrary whether we drop before or after setting it

dapplion: In line 128 we call `get_or_insert(state_root, || (state.slot(), Arc::new(OnceCell::new())))` and insert an empty OnceCell. The Cell has state `INCOMPLETE`. Then in line 137 we call `buffer_cell.get_or_init` setting the Cell to `RUNNING`. The sequence

- Drop inner lock
- Do some work
- Call `buffer_cell.get_or_init`

Increases the time another thread can get the cell from inner in the `INCOMPLETE` state. Moving the metric to *before* dropping the inner lock makes the "Do some work" window negligible. 

This is only relevant in my previous attempt where I used `get_or_init` in line 107 instead of `get`. Now it does not really matter.

PR #8177, opened 2025-10-09
Ensure status v2 returns the earliest available block when custody requirements are satisified for the da retention period

eserilev:
When `DataColumnCustodyInfo::earliest_available_slot` is less than or equal to the data availability boundary, we should return the earliest block slot when serving status-v2 request

Thanks @pawanjay176 for the help

We could alternatively update/prune `DataColumnCustodyInfo` during our regular blob/column pruning routine but that doesn't seem worth the extra effort.

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -7117,7 +7117,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         self.data_availability_checker.data_availability_boundary()
     }
 
-    /// The earliest epoch at which we require data column custody. 
+    /// The earliest epoch at which we require data column custody.
     /// It will just be the data availability boundary unless we are near the Fulu fork epoch.
     pub fn column_data_availability_boundary(&self) -> Option<Epoch> {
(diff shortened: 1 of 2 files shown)


jimmygchen (commented):
We only set the `earliest_available_data_column_slot` after a CGC increase here

https://github.com/sigp/lighthouse/blob/1dbc4f861b3f678516f6b3ba9cb448e3550b1b31/beacon_node/http_api/src/lib.rs#L3819-L3828

Once column backfill implementation is completed in #7907,  this field should get updated as backfill progresses, and eventually gets set to `None` when backfill completes:

https://github.com/sigp/lighthouse/blob/836c39efaaeeb29823d0b66bcd42adb1347ea3f1/beacon_node/store/src/metadata.rs#L208-L213

If this is not set (i.e. value == `None`), we return the `oldest_block_slot`, which I believe complies with the [latest spec clarification here](https://github.com/ethereum/consensus-specs/pull/4617/files)

https://github.com/sigp/lighthouse/blob/e5b4983d6baf85770fe4539a565d8a2dd462bc53/beacon_node/network/src/status.rs#L42-L47

I'm not sure if this is necessary, am I missing something?

PR #8181, opened 2025-10-09
Update `ForkName::latest_stable` to Fulu for tests

michaelsproul:
## Proposed Changes

Update `ForkName::latest_stable` to Fulu, reflecting our plan to stabilise Fulu in the immediate future!

This will lead to some more tests running with Fulu rather than Electra.


jimmygchen (approved):
👍

PR #8186, opened 2025-10-10
Refactor get_validator_blocks_v3 fallback

chong-he:
## Issue Addressed

#7727 introduced a bug in the logging, where as long as the node failed the SSZ `get_validator_blocks_v3` endpoint, it would log as `Beacon node does not support...`. However, the failure can be due to other reasons, such as a timed out error as found by @jimmygchen:

`WARN  Beacon node does not support SSZ in block production, falling back to JSON  slot: 5283379, error: HttpClient(url: https://ho-h-bn-cowl.spesi.io:15052/, kind: timeout, detail: operation timed out` 

## Proposed Changes

This PR made the error log more generic, so there is less confusion. 

Additionally, suggested by @michaelsproul, this PR refactors the `get_validator_blocks_v3` calls by trying all beacon nodes using the SSZ endpoint first, and if all beacon node fails the SSZ endpoint, only then fallback to JSON. 

It changes the logic from: 
"SSZ -> JSON for primary beacon node, followed by SSZ -> JSON for second beacon node and so on" to 
"SSZ for all beacon nodes -> JSON for all beacon nodes"

This has the advantage that if the primary beacon node is having issues and failed the SSZ, we avoid retrying the primary beacon node again on JSON (as it could be that the primary beacon node fail again); rather, we switch to the second beacon node.

## Additional Info

As the calling of `get_validator_blocks_v3` (and it's SSZ counterpart) is shifted to another function, the remaining part in the function `get_validator_block` is a bit of dangling in the air, so I moved them all to the now renamed `get_validator_block_and_publish_block` function and delete the original `get_validator_block` function.


validator_client/validator_services/src/block_service.rs
```diff
@@ -446,29 +445,75 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> BlockService<S, T> {
         //
         // Try the proposer nodes last, since it's likely that they don't have a
         // great view of attestations on the network.
-        let unsigned_block = proposer_fallback
```
macladson, line 456: This comment could probably be updated since it's a bit outdated now that we try SSZ on all responsive nodes first, and then do a second run with JSON

macladson (changes requested):
Looks good, just a small nit

It's a shame we can't detect the type of error first:
1. Connection refused (node offline)
2. SSZ not supported
Then we if get 2 we can immediately try JSON rather than moving onto the second BN.
But we would have to parse the error type from the API which probably isn't doable.

At this point, pretty much all BNs should be supporting SSZ by default so this should be fine

validator_client/validator_services/src/block_service.rs
chong-he: Thanks for the review. I updated the comment in [090e4c3](https://github.com/sigp/lighthouse/pull/8186/commits/090e4c35f27480b5a3626239450d434463252fdf)

Let me know if it's good

validator_client/validator_services/src/block_service.rs
```diff
@@ -444,31 +443,78 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> BlockService<S, T> {
 
         // Request block from first responsive beacon node.
         //
```
macladson:
```suggestion
```

validator_client/validator_services/src/block_service.rs
```diff
@@ -444,31 +443,78 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> BlockService<S, T> {
...
-        // Try the proposer nodes last, since it's likely that they don't have a
+        // Try to get_validator_block_v3 using SSZ first for all beacon nodes, if all failed then fallback to JSON
+        // the proposer nodes is always the last to try for both SSZ and JSON, since it's likely that they don't have a
         // great view of attestations on the network.
```
macladson:
```suggestion
        // Request an SSZ block from all beacon nodes in order, returning on the first successful response. 
        // If all nodes fail, run a second pass falling back to JSON.
        //
        // Proposer nodes will always be tried last during each pass since it's likely that they don't have a
        // great view of attestations on the network.
```

macladson (dismissed):
Just a slight edit for clarity

macladson:
Just needs a `cargo fmt` for some issue with whitespaces, otherwise looks pretty good to me!

chong-he:
> Just needs a `cargo fmt` for some issue with whitespaces, otherwise looks pretty good to me!

sorry I overlook the checks, fixed now

michaelsproul (approved):

michaelsproul:
@mergify requeue

michaelsproul:
@mergify requeue

PR #8187, opened 2025-10-10
Replace Debug formatting with Display for stable file naming in validator manager

sashaodessa:
## Issue Addressed

Fixes unstable file naming in validator manager tests that could lead to inconsistent behavior across different environments.

## Proposed Changes

- Replace `Debug` formatting (`{:?}`) with `Display` formatting (`.to_string()`) for generating file paths from `PublicKeyBytes`
- Updated 5 instances in `validator_manager/src/move_validators.rs` where file paths were constructed using `join(format!("{:?}", validating_pubkey))`

**Solution:** `PublicKeyBytes` implements `Display` trait which provides a stable, hex-encoded representation suitable for file naming.

### DIFF
--- a/validator_manager/src/move_validators.rs
+++ b/validator_manager/src/move_validators.rs
@@ -833,7 +833,7 @@ mod test {
                                 !src_vc
                                     .secrets_dir
                                     .path()
-                                    .join(format!("{:?}", initial_keystore.validating_pubkey))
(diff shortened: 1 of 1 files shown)


chong-he (commented):
These changes are in the test part of the `move_validators`, they are trivial and do not have benefits to improve the test. 

I suggest to close the PR

PR #8188, opened 2025-10-10
Persist only custody columns in db

### LINKED ISSUE #8129: Full node with no validators attached to store only custody columns and discard extra sampling columns
## Description

When subnet sampling was implemented in #6410, I didn't implement discarding the extra sampling columns (see [comment](https://github.com/sigp/lighthouse/pull/6410#issuecomment-2391356474)) for the reasons below
* It wasn't clear whether we were going to implement peer sampling in Fulu at the time, which would make the change redundant
* `CustodyContext` didn't exist and there wasn't an simple and clean way for the beacon chain to filter out sampling columns
https://github.com/sigp/lighthouse/blob/3de646c8b32b6da7d2ace48aab9ceb2e52bbe8a5/beacon_node/beacon_chain/src/validator_custody.rs#L145

So there was a bit of extra effort with some chance that it would be eventually reverted - and even with the change, the benefit for full nodes was quite minimal - they would still have to download & sample the 4 extra columns, and saving only 64kb for a 8-blob block.

However, this seems worth implementing now peer sampling has been completely removed from the spec, and also this is now straight forward to do because `BeaconChain` knows exact which columns to discard.

pawanjay176:
## Issue Addressed

Resolves #8129 

## Proposed Changes

Persist only the custody columns and not sampling columns like before.

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -3957,7 +3957,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         // See https://github.com/sigp/lighthouse/issues/2028
         let (_, signed_block, block_data) = signed_block.deconstruct();
 
-        match self.get_blobs_or_columns_store_op(block_root, block_data) {
+        match self.get_blobs_or_columns_store_op(block_root, signed_block.slot(), block_data) {
             Ok(Some(blobs_or_columns_store_op)) => {
                 ops.push(blobs_or_columns_store_op);
             }
@@ -7163,6 +7163,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
     pub(crate) fn get_blobs_or_columns_store_op(
         &self,
         block_root: Hash256,
+        block_slot: Slot,
         block_data: AvailableBlockData<T::EthSpec>,
     ) -> Result<Option<StoreOp<'_, T::EthSpec>>, String> {
         match block_data {
@@ -7175,7 +7176,15 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
                 );
                 Ok(Some(StoreOp::PutBlobs(block_root, blobs)))
             }
-            AvailableBlockData::DataColumns(data_columns) => {
+            AvailableBlockData::DataColumns(mut data_columns) => {
+                let columns_to_custody = self.custody_columns_for_epoch(Some(
+                    block_slot.epoch(T::EthSpec::slots_per_epoch()),
+                ));
+                // Supernodes need to persist all sampled custody columns
+                if columns_to_custody.len() != self.spec.number_of_custody_groups as usize {
+                    data_columns
+                        .retain(|data_column| columns_to_custody.contains(&data_column.index));
+                }
                 debug!(
                     %block_root,
                     count = data_columns.len(),
--- a/beacon_node/beacon_chain/tests/store_tests.rs
+++ b/beacon_node/beacon_chain/tests/store_tests.rs
@@ -4196,10 +4196,8 @@ async fn test_custody_column_filtering_regular_node() {
         .expect("should get stored column keys");
 
     // Verify only custody columns were stored
-    let stored_column_indices: std::collections::HashSet<_> =
-        stored_column_keys.into_iter().collect();
-    let expected_column_indices: std::collections::HashSet<_> =
-        custody_columns.iter().copied().collect();
+    let stored_column_indices: HashSet<_> = stored_column_keys.into_iter().collect();
+    let expected_column_indices: HashSet<_> = custody_columns.iter().copied().collect();
 
     assert_eq!(
         stored_column_indices, expected_column_indices,
@@ -4225,15 +4223,10 @@ async fn test_custody_column_filtering_supernode() {
     let (signed_block, all_data_columns) =
         beacon_chain::test_utils::generate_rand_block_and_data_columns::<E>(
             fork_name,
-            beacon_chain::test_utils::NumBlobs::Number(3), // Generate 3 blobs to produce data columns
+            beacon_chain::test_utils::NumBlobs::Number(1),
             &mut rng,
             &harness.spec,
         );
-
-    // Skip test if no data columns are generated
-    if all_data_columns.is_empty() {
-        return;
-    }
 
     let block_root = signed_block.canonical_root();
 
@@ -4258,21 +4251,12 @@ async fn test_custody_column_filtering_supernode() {
         .expect("should get stored column keys");
 
     // Verify ALL data columns were stored for supernodes
-    let stored_column_indices: std::collections::HashSet<_> =
-        stored_column_keys.into_iter().collect();
-    let all_column_indices: std::collections::HashSet<_> =
-        all_data_columns.iter().map(|dc| dc.index).collect();
+    let stored_column_indices: HashSet<_> = stored_column_keys.into_iter().collect();
+    let all_column_indices: HashSet<_> = all_data_columns.iter().map(|dc| dc.index).collect();
 
     assert_eq!(
         stored_column_indices, all_column_indices,
         "Supernode should store ALL data columns"
-    );
-
-    // Verify the count matches
-    assert_eq!(
-        stored_column_indices.len(),
-        all_data_columns.len(),
-        "Supernode should store the same number of columns as generated"
     );
 
     // Verify each stored column can be retrieved and matches original data
--- a/beacon_node/beacon_chain/tests/store_tests.rs
+++ b/beacon_node/beacon_chain/tests/store_tests.rs
@@ -4212,8 +4212,7 @@ async fn test_custody_column_filtering_supernode() {
         .await;
 
     // Supernodes are expected to store all data columns
-    let expected_custody_columns: HashSet<_> =
-        (0..E::number_of_columns() as u64).into_iter().collect();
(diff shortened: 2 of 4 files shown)


beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -7175,7 +7176,15 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
                 );
                 Ok(Some(StoreOp::PutBlobs(block_root, blobs)))
             }
-            AvailableBlockData::DataColumns(data_columns) => {
+            AvailableBlockData::DataColumns(mut data_columns) => {
+                let columns_to_custody = self.custody_columns_for_epoch(Some(
+                    block_slot.epoch(T::EthSpec::slots_per_epoch()),
+                ));
+                // Supernodes need to persist all sampled custody columns
```
dapplion, line 7183:
```suggestion
                // If `columns_to_custody` contains all possible custody groups, no need to filter
```
I would update in the comment the mention of supernode. In the future, nodes custodying >50% of columns and doing reconstruction may be supernodes?

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -4137,6 +4138,195 @@ async fn replay_from_split_state() {
...
+            fork_name,
+            beacon_chain::test_utils::NumBlobs::Number(3), // Generate 3 blobs to produce data columns
+            &mut rng,
+            &harness.spec,
+        );
+
+    // Skip test if no data columns are generated
+    if all_data_columns.is_empty() {
+        return;
+    }
```
dapplion: What's the reason of this skip?

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -4137,6 +4138,195 @@ async fn replay_from_split_state() {
...
+
+    // Generate a block with data columns using public test utils
+    let fork_name = harness.spec.fork_name_at_epoch(Epoch::new(0));
+    let mut rng = rand::rng();
+    let (signed_block, all_data_columns) =
+        beacon_chain::test_utils::generate_rand_block_and_data_columns::<E>(
+            fork_name,
+            beacon_chain::test_utils::NumBlobs::Number(3), // Generate 3 blobs to produce data columns
```
dapplion: One is enough, and will take less CPU

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -4137,6 +4138,195 @@ async fn replay_from_split_state() {
...
+    // Verify only custody columns were stored
+    let stored_column_indices: std::collections::HashSet<_> =
+        stored_column_keys.into_iter().collect();
+    let expected_column_indices: std::collections::HashSet<_> =
+        custody_columns.iter().copied().collect();
+
+    assert_eq!(
+        stored_column_indices, expected_column_indices,
+        "Regular node should only store custody columns"
+    );
```
dapplion, line 4189: This assert makes the asserts below redundant right? If stored_column_indices == expected_column_indices then stored_column_indices does not include what expected_column_indices does not include

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -4137,6 +4138,195 @@ async fn replay_from_split_state() {
...
+            .iter()
+            .find(|stored| stored.index == original_column.index);
+
+        assert!(
+            matching_stored.is_some(),
+            "Column {} should be present in stored data",
+            original_column.index
+        );
+
+        let stored_column = matching_stored.unwrap();
```
dapplion:
```suggestion
        let stored_column = stored_columns
            .iter()
            .find(|stored| stored.index == original_column.index)
            .unwrap_or_else(||
                panic!("Column {} should be present in stored data", original_column.index)
            );
```

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -4137,6 +4138,195 @@ async fn replay_from_split_state() {
...
+    let stored_columns = store
+        .get_data_columns(&block_root)
+        .expect("should get data columns")
+        .expect("data columns should exist");
+
+    assert_eq!(
+        stored_columns.len(),
+        all_data_columns.len(),
+        "All data columns should be retrievable"
+    );
```
dapplion, line 4227: This check is redundant with the check in the for loop below, and the assert below gives more info saying which column is missing instead of just logging "All data columns should be retrievable"

dapplion (commented):
Nice test! Just some minor comments

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: I think this is fine, supernode is explicitly defined here in the spec
https://github.com/ethereum/consensus-specs/blob/dev/specs/fulu/p2p-interface.md#supernodes

beacon_node/beacon_chain/tests/store_tests.rs
jimmygchen: I don't think this does anything given that we're using the `test_utils` to generate exactly 3 blobs, so there must be columns. We should remove this.

If I'm guessing right, the test might originally  be written using the randomly generated number of blobs from `exectuion_block_generator`-  we have this pattern in a lot of test code - i believe we should eliminate this pattern if we actually need to test blobs / data columns.

beacon_node/beacon_chain/tests/store_tests.rs
jimmygchen: I agree we could use less where we can. Just to note that this function doesn't actually generate blobs / proofs, it loads static generated ones, and just repeat them for the `num_blobs`, so the CPU overhead _should_ be minimal.

jimmygchen (commented):
Nice, the changes are simple enough i think we should include it in the rc.1 release once Lion's comments are addressed.

michaelsproul (approved):

beacon_node/beacon_chain/tests/store_tests.rs
jimmygchen: Addressed https://github.com/sigp/lighthouse/pull/8188/commits/84a4c152ce4d84e82b96c3ed8514aa62a67550c0

beacon_node/beacon_chain/tests/store_tests.rs
jimmygchen: Addressed https://github.com/sigp/lighthouse/pull/8188/commits/84a4c152ce4d84e82b96c3ed8514aa62a67550c0

beacon_node/beacon_chain/tests/store_tests.rs
jimmygchen: Addressed https://github.com/sigp/lighthouse/pull/8188/commits/84a4c152ce4d84e82b96c3ed8514aa62a67550c0

beacon_node/beacon_chain/tests/store_tests.rs
jimmygchen: Addressed https://github.com/sigp/lighthouse/pull/8188/commits/84a4c152ce4d84e82b96c3ed8514aa62a67550c0

beacon_node/beacon_chain/tests/store_tests.rs
jimmygchen: Addressed https://github.com/sigp/lighthouse/pull/8188/commits/84a4c152ce4d84e82b96c3ed8514aa62a67550c0

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -7175,7 +7176,15 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
                 );
                 Ok(Some(StoreOp::PutBlobs(block_root, blobs)))
             }
-            AvailableBlockData::DataColumns(data_columns) => {
+            AvailableBlockData::DataColumns(mut data_columns) => {
+                let columns_to_custody = self.custody_columns_for_epoch(Some(
```
jimmygchen, line 7180: The failing `weak_subjectivity_sync_easy` revealed an issue here
The `CustodyContext` is not yet initialised when building the chain, so this will panic on startup.

jimmygchen: Oh this is not true in the production code path

https://github.com/sigp/lighthouse/blob/78d330e4b7e2b76ab503cd88f4a365a6d7a0bcf0/beacon_node/beacon_chain/src/builder.rs#L585-L590

so i think we just need to fix the test

michaelsproul: Hmm, I thought those expects looked a bit sus. Maybe we could store all columns if the custody context is not initialized? It would only apply in this case on startup

Or do you think we can (easily) change the order of initialization?

jimmygchen: Yeah, this one is a bit tricky to avoid, we've evaluated some options here and it seems to be the simplest approach at the time:

https://github.com/sigp/lighthouse/pull/7792/

Fix for the test pushed here https://github.com/sigp/lighthouse/pull/8188/commits/cb74cce363ce37609077fecff6177c751d31e6d7

beacon_node/beacon_chain/tests/store_tests.rs
jimmygchen: Sorry you’re right, even if generation is cheap, if node have to process them we’ll end up using more CPUs.

created a PR here to bring it down
https://github.com/sigp/lighthouse/pull/8194

PR #8194, opened 2025-10-13
Reduce number of blobs used in tests to speed up CI

jimmygchen:
## Issue Addressed

`beacon-chain-tests` is now regularly taking 1h+ on CI since Fulu fork was added.

This PR attemtpts to reduce the test time by bringing down the number of blobs generated in tests - instead of generating 0..max_blobs, the generator now generates 0..1 blobs by default, and this can be modified by setting `harness.execution_block_generator.set_min_blob_count(n)`.

Note: The blobs are pre-generated and doesn't require too much CPU to generate however processing a larger number of them on the beacon chain does take a lot of time.

This PR also include a few other small improvements
- Our slowest test (`chain_segment_varying_chunk_size`) runs 3x faster in Fulu just by reusing chain segments
- Avoid re-running fork specific tests on all forks
- Fix a bunch of tests that depends on the harness's existing random blob generation, which is fragile

beacon chain test time on test machine is **~2x** faster:

### `unstable`

```
      Summary [ 751.586s] 291 tests run: 291 passed (13 slow), 0 skipped
```

### this branch

```
      Summary [ 373.792s] 291 tests run: 291 passed (2 slow), 0 skipped
```

The next set of tests to optimise is the ones that use [`get_chain_segment`](https://github.com/sigp/lighthouse/blob/77a9af96de0f693127055e381ece3e98dceea0a8/beacon_node/beacon_chain/tests/block_verification.rs#L45), as it by default build 320 blocks with supernode - an easy optimisation would be to build these blocks with cgc = 8 for tests that only require fullnodes.


jimmygchen:
Interestingly the broadcast validation test failure here was caused by the block not having a blob, and returning a different error message:
https://github.com/sigp/lighthouse/actions/runs/18466275201/job/52674350559?pr=8194

We could probably use `set_min_blob_count` to make the test deterministic, or assert for different errors depending on whether blob was included in the block.

i'll go with the former method for now as the tests here are already quite complicated.

beacon_node/beacon_chain/tests/block_verification.rs
```diff
@@ -297,19 +297,20 @@ async fn chain_segment_full_segment() {
 
 #[tokio::test]
 async fn chain_segment_varying_chunk_size() {
+    let (chain_segment, chain_segment_blobs) = get_chain_segment().await;
+    let blocks: Vec<RpcBlock<E>> = chain_segment_blocks(&chain_segment, &chain_segment_blobs)
+        .into_iter()
+        .collect();
```
jimmygchen, line 303: This was one of our slowest tests and takes 800s+ to run on CI.

`get_chain_segment` builds out a chain up to slow 320 and it's quite expensive. Previously we were doing this in a `for` loop. Moving this up and cloning the blocks gives a pretty good improvement.

On my laptop, this change reduced the test time (`FORK_NAME=fulu`) from ~3 mins down to 1 minute.

beacon_node/beacon_chain/src/test_utils.rs
```diff
@@ -3245,96 +3250,49 @@ pub enum NumBlobs {
     None,
 }
 
+macro_rules! add_blob_transactions {
+    ($message:expr, $payload_type:ty, $num_blobs:expr, $rng:expr, $fork_name:expr) => {{
+        let num_blobs = match $num_blobs {
+            NumBlobs::Random => $rng.random_range(1..=2),
```
michaelsproul: Might be nice to pull these into `const`s?

beacon_node/beacon_chain/tests/block_verification.rs
michaelsproul: We could probably drop a few of the loop iterations too. `1, 2, 31, 32, 33` might be sufficient.

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -3290,9 +3301,8 @@ async fn test_import_historical_data_columns_batch_mismatched_block_root() {
     for block in block_root_iter {
         let (block_root, _) = block.unwrap();
         let data_columns = harness.chain.store.get_data_columns(&block_root).unwrap();
-        assert!(data_columns.is_some());
 
-        for data_column in data_columns.unwrap() {
+        for data_column in data_columns.unwrap_or_default() {
```
michaelsproul, line 3305: I don't see why this is necessary. If data columns are None, won't the new assert below fail?

```rust
assert!(!data_columns_list.is_empty());
```

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -3374,15 +3389,16 @@ async fn test_import_historical_data_columns_batch_no_block_found() {
...
```
michaelsproul, line 3393: As above

michaelsproul (approved):
LGTM, just a few small Qs

beacon_node/beacon_chain/tests/store_tests.rs
jimmygchen: We may not have blobs in every block
The assertion below checks if we have at least some blobs in the block root range

beacon_node/beacon_chain/tests/block_verification.rs
jimmygchen: Updated, thanks!

beacon_node/beacon_chain/src/test_utils.rs
jimmygchen: Yep, updated.

For some reasons I thought I did 0..1, maybe i did change it to 1..=2 😅

beacon_node/beacon_chain/tests/store_tests.rs
michaelsproul: Ohhh yeah my bad, read this wrong

michaelsproul (approved):

PR #8196, opened 2025-10-13
Downgrade light client error logs

eserilev:
## Issue Addressed

Temporary stop gap for #7002

## Proposed Changes

Downgrade light client errors to debug

We eventually should fix our light client objects so they can consist of data across forks.


michaelsproul (approved):
LGTM, thanks!

PR #8199, opened 2025-10-14
Improve banning logic by grouping Ipv6 /56 prefixes

AgeManning:
As we now support IPv6 and it is becomming more prevelant, this PR improves the security posture of Lighthouse by grouping /56 prefix'd IP addresses when counting bans. 

What this means is that if a peer gets banned, they cannot just move to a new ipv6 address within their prefix and perform poorly again. We use our IP-colocation ban logic and classify all /56 prefix's as a single IP when counting bans. If the prefix accumulates enough bans we reject the entire prefix. 

NOTE: While I was testing I noted a lot of logging and delays in the tests. So while I was there I removed the logging and attempted to improve the test speed for the property based peer pruning test.


dknopik (approved):
LGTM. I undid the change to the property based test, as we seem to have migrated to `proptest!` in the meantime. I also reverted to the multi core tokio runtime as part of that.

pawanjay176:
@jxs would be good to have your review on this. Tagging because you are automatically added pretty much everywhere 😄

beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
```diff
@@ -29,6 +29,10 @@ const MAX_DC_PEERS: usize = 500;
 pub const MAX_BANNED_PEERS: usize = 1000;
 /// We ban an IP if there are more than `BANNED_PEERS_PER_IP_THRESHOLD` banned peers with this IP.
 const BANNED_PEERS_PER_IP_THRESHOLD: usize = 5;
+/// The prefix length for grouping IPv6 addresses when banning peers.
+/// Most ISPs allocate /56 prefixes to users, so we group banned peers under this prefix
+/// to prevent attackers from generating new IPv6 addresses within their allocation to avoid bans.
+const IPV6_BANNED_GROUPING_PREFIX: u8 = 56;
```
jxs:
```suggestion
```

beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
```diff
@@ -1419,13 +1423,41 @@ pub struct BannedPeersCount {
...
+            if FULL_SEGMENTS < 8 && REMAINING_BITS > 0 {
+                const MASK: u16 = !((1u16 << (16 - REMAINING_BITS)) - 1);
+                masked_segments[FULL_SEGMENTS] = segments[FULL_SEGMENTS] & MASK;
+            }
+
+            IpAddr::V6(std::net::Ipv6Addr::from(masked_segments))
+        }
+    }
+}
+
```
jxs:
```suggestion
/// Normalizes an IP address for grouped ban counting.
///
/// For [`IpAddr::V4`] the address is returned unchanged.
///
/// For [`IpAddr::V6`] the address is masked to the /56 prefix (typical ISP
/// allocation size for residential users), zeroing out the lower 72 host bits.
/// This groups all addresses from the same ISP-allocated /56 subnet under a
/// single key in the ban counter map, preventing attackers from evading bans
/// by cycling through addresses in their allocation.
fn normalize_ip_for_banning(ip: IpAddr) -> IpAddr {
    match ip {
        IpAddr::V4(_) => ip,
        IpAddr::V6(ipv6) => {
            const MASK: u128 = 0xFFFF_FFFF_FFFF_FF00_0000_0000_0000_0000;
            IpAddr::V6(std::net::Ipv6Addr::from(u128::from(ipv6) & MASK))
        }
    }
}
```

jxs (commented):
Thanks for the ping folks!
I think the current `normalize_ip_for_banning` is a little bit hard to follow, it manually computes full segments, remaining bits, and a partial mask.
`/56` is sort of well established so I don't think we need to assign it to a constant, I think it's easier to just hardcode the mask directly as a u128 constant.
This makes the function trivially simple: convert to u128, AND with the mask, convert back wdyt?

beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
dknopik: I have no particular preference between these two :)

beacon_node/lighthouse_network/src/peer_manager/peerdb.rs
```diff
@@ -1419,13 +1419,32 @@ pub struct BannedPeersCount {
...
+/// by cycling through addresses in their allocation.
+fn normalize_ip_for_banning(ip: IpAddr) -> IpAddr {
+    match ip {
+        IpAddr::V4(_) => ip,
+        IpAddr::V6(ipv6) => {
+            const MASK: u128 = 0xFFFF_FFFF_FFFF_FF00_0000_0000_0000_0000;
+            IpAddr::V6(std::net::Ipv6Addr::from(u128::from(ipv6) & MASK))
+        }
+    }
+}
```
jxs, line 1439:
```suggestion
}

```

jxs (approved):
LGTM

PR #8200, opened 2025-10-14
update libp2p dependency to upstream

jxs:
depend on git upstream as it's now up to date


michaelsproul:
We're about to do a v8.0.0 release, so I'll tag this for v8.1.0 so it can undergo more testing prior to release?

jxs:
fixed the conflicts, should be good to go

Cargo.toml
```diff
@@ -168,6 +166,20 @@ initialized_validators = { path = "validator_client/initialized_validators" }
 int_to_bytes = { path = "consensus/int_to_bytes" }
 itertools = "0.10"
 kzg = { path = "crypto/kzg" }
+libp2p = { git = "https://github.com/libp2p/rust-libp2p.git", default-features = false, features = [
```
ackintosh, line 169: Have you considered pinning this to a specific commit hash? That would make the dependency version explicit and help prevent unintended updates.

jxs: Hi Akihito, yeah, that's what was before.
The pinning is already achieved by `Cargo.lock` and we are following `master` it _shouldn't_ bring problems.
I.e if someone updates the libp2p dep nested on a PR that updates `Cargo.lock` it should be ok.
I feel that's preferable to hardcoding a commit hash that needs a PR every time we need a bump.
Do you prefer to be explicit on `Cargo.toml` as well?
cc @dknopik

ackintosh: I agree that your point makes sense, and I think relying on `Cargo.lock` is sufficient in this case.

That said, I believe specifying a tagged release (i.e., a version) would be ideal in the long run, so I just wanted to check whether this is intended as a temporary solution or a longer-term approach.

pawanjay176: I don't have a strong preference, but historically, we have explicitly pinned commits that we intend to put out on stable branch.

pawanjay176 (approved):
LGTM. would be good to test this with 8.1

PR #8201, opened 2025-10-14
Ignore extra columns in da cache

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Found this issue in sepolia. Note: the custody requirement for this node is 100. 
```
Oct 14 11:25:40.053 DEBUG Reconstructed columns                         count: 28, block_root: 0x4d7946dec0ab59f2afd46610d7c54af555cb4c2851d9eea7d83dd17cf6e96aae, slot: 8725628
Oct 14 11:25:45.568 WARN  Internal availability check failure           block_root: 0x4d7946dec0ab59f2afd46610d7c54af555cb4c2851d9eea7d83dd17cf6e96aae, error: Unexpected("too many columns got 128 expected 100")
```

So if any of the block components arrives late, then we reconstruct all 128 columns and try to add it to da cache and have more columns than needed for availability in the cache. 

There are 2 ways I can think of fixing this:
1. pass only the required columns to the da cache after reconstruction here https://github.com/sigp/lighthouse/blob/60df5f4ab609362711f4f518eb8f03df447bfedb/beacon_node/beacon_chain/src/data_availability_checker.rs#L647-L648
2. Ensure that we add only columns that we need to sample in the da cache. I think this is safer since we can add columns to the cache from multiple code paths and this fixes it at the source.

~~This PR implements (2).~~ Thought more about it, I think (1) is cleaner since we filter gossip and rpc columns also before calling `put_kzg_verified_data_columns`/

### DIFF
--- a/beacon_node/beacon_chain/src/data_availability_checker.rs
+++ b/beacon_node/beacon_chain/src/data_availability_checker.rs
@@ -1105,7 +1105,7 @@ mod test {
         // Set custody requirement to 65 columns (enough to trigger reconstruction)
         let epoch = Epoch::new(1);
         custody_context.register_validators(
-            vec![(0, 2048_000_000_000), (1, 32_000_000_000)], // 64 + 1
+            vec![(0, 2_048_000_000_000), (1, 32_000_000_000)], // 64 + 1
             Slot::new(0),
             &spec,
         );
--- a/beacon_node/beacon_chain/src/data_availability_checker.rs
+++ b/beacon_node/beacon_chain/src/data_availability_checker.rs
@@ -1091,6 +1091,101 @@ mod test {
         verification_result.expect_err("should have failed to verify blocks");
     }
 
+    #[test]
+    fn should_exclude_reconstructed_columns_not_required_for_sampling() {
+        // SETUP
+        let spec = Arc::new(ForkName::Fulu.make_genesis_spec(E::default_spec()));
+        let mut rng = StdRng::seed_from_u64(0xDEADBEEF0BAD5EEDu64);
+
+        let da_checker = new_da_checker(spec.clone());
+        let custody_context = &da_checker.custody_context;
+        let all_column_indices_ordered =
+            init_custody_context_with_ordered_columns(custody_context, &mut rng, &spec);
+
+        // Set custody requirement to 65 columns (enough to trigger reconstruction)
+        let epoch = Epoch::new(1);
+        custody_context.register_validators(
+            vec![(0, 2048_000_000_000), (1, 32_000_000_000)], // 64 + 1
+            Slot::new(0),
+            &spec,
+        );
+        let sampling_requirement = custody_context.num_of_data_columns_to_sample(epoch, &spec);
+        assert_eq!(
+            sampling_requirement, 65,
+            "sampling requirement should be 65"
+        );
+
+        let (block, data_columns) = generate_rand_block_and_data_columns::<E>(
+            ForkName::Fulu,
+            NumBlobs::Number(1),
+            &mut rng,
+            &spec,
+        );
+        let block_root = Hash256::random();
+        // Add the block to the DA checker
+        da_checker
+            .availability_cache
+            .put_pre_execution_block(block_root, Arc::new(block), BlockImportSource::Gossip)
+            .expect("should put block");
+
+        // Add 64 columns to the da checker (enough to be able to reconstruct)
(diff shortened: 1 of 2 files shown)


pawanjay176:
Thinking about this a bit more, seems like (2) is swallowing the error here.

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -1082,6 +1091,101 @@ mod test {
...
+        let (availability, reconstructed_columns) = match reconstruction_result {
+            DataColumnReconstructionResult::Success(result) => {
+                // dbg!(&result);
+                result
+            }
+            e => {
+                panic!("Expected successful reconstruction {:?}", e);
+            }
+        };
+
```
pawanjay176, line 1161: I also wanted to assert `matches!(availability, Availability::MissingComponents(_)))`
But that doesn't work because we are adding a pre execution block. I figured its okay because `reconstruction_result` wouldn't have been success otherwise.

jimmygchen: Just curious why wouldn't this work? A pre execution block should still result in `MissingComponents`

https://github.com/sigp/lighthouse/blob/9d0e44b2f7ac3dd33cf716fd2936cccc6cee9d43/beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs#L228-L231

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -644,8 +644,17 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
...
         );
 
+        let columns_to_sample = self
+            .custody_context()
+            .sampling_columns_for_epoch(slot.epoch(T::EthSpec::slots_per_epoch()), &self.spec);
+        let data_columns_to_import: Vec<_> = data_columns_to_publish
+            .iter()
+            .filter(|column| columns_to_sample.contains(&column.index()))
+            .cloned()
+            .collect();
```
jimmygchen, line 654: I made a spec PR here, as per our discussion this morning - I realise non supernondes would actually use more outbound bandwidth after reconstruction, i think we should just stick with publishing `data_columns_to_import`. It's a `SHOULD` on the current spec to publish all new columns.
https://github.com/ethereum/consensus-specs/pull/4657

beacon_node/beacon_chain/src/data_availability_checker.rs
jimmygchen: oh yeah `Availability` is probably not important for this test, it could be either `MissingComponents` or `Imported` depending on the execution status

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -1082,6 +1091,98 @@ mod test {
...
+
+        // Try reconstrucing
+        let reconstruction_result = da_checker
+            .reconstruct_data_columns(&block_root)
+            .expect("should reconstruct columns");
+
+        // Reconstruction should succeed
+        let (_availability, reconstructed_columns) = match reconstruction_result {
+            DataColumnReconstructionResult::Success(result) => {
+                // dbg!(&result);
```
jimmygchen: This line can be removed

jimmygchen:
> Thinking about this a bit more, seems like (2) is swallowing the error here.

Perhaps we can be a bit defensive here and perform a check, given the DA checker now knows about the sampling columns?

jimmygchen (commented):
Looks good! I've added some comments, let me know what you think.

beacon_node/beacon_chain/src/data_availability_checker.rs
pawanjay176: I meant it was returning MissingComponents when I wanted it to change after the reconstruction. But you're right we don't need it

beacon_node/beacon_chain/src/data_availability_checker.rs
pawanjay176: I agree. We can handle that in another PR after other clients go over it i guess.

jimmygchen (approved):
Looks good, thanks! 
Doesn't seem to impactful to do the extra check right now, as we already have checks in all the code paths calling DA checker. We're reworking the DA checker post fulu anyways, so it could be done in a separate PR or as part of the rework.

PR #8202, opened 2025-10-14
Remove `DataColumnInfo`

eserilev:
## Issue Addressed

#6572

## Proposed Changes

Remove data column info usage and add a non-reversible migration script.

Also fixes an issue mentioned in this comment regarding data column rpc requests:
https://github.com/sigp/lighthouse/issues/6572#issuecomment-3400076236

## Additional Info

The migration script doesn't necessarily need to be non-reversible. Downgrading could technically just be a no-op.

### DIFF
--- a/beacon_node/network/src/network_beacon_processor/rpc_methods.rs
+++ b/beacon_node/network/src/network_beacon_processor/rpc_methods.rs
@@ -1154,33 +1154,42 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
 
         let request_start_slot = Slot::from(req.start_slot);
 
-        let data_availability_boundary_slot = match self.chain.data_availability_boundary() {
-            Some(boundary) => boundary.start_slot(T::EthSpec::slots_per_epoch()),
-            None => {
-                debug!("Deneb fork is disabled");
-                return Err((RpcErrorResponse::InvalidRequest, "Deneb fork is disabled"));
-            }
-        };
+        let column_data_availability_boundary_slot =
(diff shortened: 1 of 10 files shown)


eserilev:
Write a test

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -1154,33 +1154,42 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
-                debug!("Deneb fork is disabled");
-                return Err((RpcErrorResponse::InvalidRequest, "Deneb fork is disabled"));
-            }
-        };
+        let column_data_availability_boundary_slot =
+            match self.chain.column_data_availability_boundary() {
+                Some(boundary) => boundary.start_slot(T::EthSpec::slots_per_epoch()),
+                None => {
+                    debug!("Deneb fork is disabled");
+                    return Err((RpcErrorResponse::InvalidRequest, "Deneb fork is disabled"));
```
michaelsproul, line 1162: It could be Fulu or Deneb that is disabled, but I think it's not wrong to say Fulu is disabled (because Deneb being disabled also implies Fulu being disabled).

```suggestion
                    debug!("Fulu fork is disabled");
                    return Err((RpcErrorResponse::InvalidRequest, "Fulu fork is disabled"));
```

michaelsproul:
Bumping this to v8.1.0 due to the schema change. We can do the RPC alone in its own PR

PR #8206, opened 2025-10-15
Add `eip_3076` crate

macladson:
## Issue Addressed

#7894

## Proposed Changes

Moves the `Interchange` format from `slashing_protection` and thus removes the dependency on `slashing_protection` from `eth2` which can now just depend on the slimmer `eip_3076` crate.

### DIFF
--- a/common/eip_3076/src/lib.rs
+++ b/common/eip_3076/src/lib.rs
@@ -190,7 +190,7 @@ mod tests {
                     signed_attestations: vec![SignedAttestation {
                         source_epoch: Epoch::new(0),
                         target_epoch: Epoch::new(5),
-                        signing_root: None,
+                        signing_root: Some(Hash256::from_low_u64_be(2)),
                     }],
                 },
                 InterchangeData {
--- a/common/eip_3076/src/lib.rs
+++ b/common/eip_3076/src/lib.rs
@@ -1,9 +1,14 @@
-use crate::InterchangeError;
 use serde::{Deserialize, Serialize};
 use std::cmp::max;
 use std::collections::{HashMap, HashSet};
+#[cfg(feature = "json")]
 use std::io;
 use types::{Epoch, Hash256, PublicKeyBytes, Slot};
+
+#[derive(Debug)]
+pub enum Error {
+    MaxInconsistent,
+}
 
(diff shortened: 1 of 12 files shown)


consensus/types/src/contribution_and_proof.rs
```diff
@@ -10,7 +10,6 @@ use test_random_derive::TestRandom;
 use tree_hash_derive::TreeHash;
 
 /// A Validators aggregate sync committee contribution and selection proof.
-
```
macladson, line 13: This extra line kept triggering Clippy. Not sure why this wasn't picked up before though

common/eip_3076/src/lib.rs
```diff
@@ -1,10 +1,15 @@
-use crate::InterchangeError;
 use serde::{Deserialize, Serialize};
 use std::cmp::max;
 use std::collections::{HashMap, HashSet};
+#[cfg(feature = "json")]
 use std::io;
 use types::{Epoch, Hash256, PublicKeyBytes, Slot};
```
macladson, line 6: These `types` imports add a lot of bloat to this crate. It doesn't really matter much right now since `eth2` needs `types` anyway but it might be worth considering some kind of `core` or `primitives` feature in `types` which only imports these basic types, especially if this crate gets used in  contexts where the rest of the types aren't already required.

michaelsproul (approved):
Perfect, I love this!

michaelsproul:
@macladson weirdly one of the interchange tests is failing with an out-of-bounds error, can you take a look? :pray:

https://github.com/sigp/lighthouse/actions/runs/18548309242/job/52870463882?pr=8211

macladson:
Whoops, I de-duplicated the `Interchange` construction right before I pushed and broke it :sweat_smile: 
Fixed now

macladson:
Hmm still getting failures 1/3 times.. Something about that test is non-deterministic

macladson:
It was from `HashMap` ordering. I've just gone ahead and removed those extra checks since minification behaviour is more thouroughly checked here anyway -> https://github.com/sigp/lighthouse/blob/unstable/validator_client/slashing_protection/src/interchange_test.rs#L247-L302

PR #8215, opened 2025-10-16
Revert incorrect fix made in #8179

jimmygchen:
## Issue Addressed

This PR reverts #8179.

It turns out that the fix was invalid because an unknown root is always not a finalized descendant:

https://github.com/sigp/lighthouse/blob/522bd9e9c6ac167f2231525e937c9ebbcb86cf6e/consensus/proto_array/src/proto_array.rs#L976-L979

so for any data columns with unknown parents, it will always penalise the gossip peer and disconnect it pretty quickly. On a small network, the node may lose all of its peers.

The impact is pretty obvious when the peer count is small and sync speed is slow, and is therefore easily reproducible by running a fresh supernode on devnet-3.

This isn't as obvious on a live testnet like holesky / sepolia, we haven't noticed this, probably due to its high peer count and sync speed - the nodes might be able to reach head quickly before losing too many peers.

## Proposed Changes

The previous behaviour isn't ideal but safe:  triggering unknown parent lookup and penalise the bad peer if it happens to be malicious or faulty. So for now it's safer to revert the change and plan for a proper fix after the v8 release.

### DIFF
--- a/beacon_node/beacon_chain/src/data_column_verification.rs
+++ b/beacon_node/beacon_chain/src/data_column_verification.rs
@@ -626,20 +626,21 @@ fn verify_parent_block_and_finalized_descendant<T: BeaconChainTypes>(
     chain: &BeaconChain<T>,
 ) -> Result<ProtoBlock, GossipDataColumnError> {
     let fork_choice = chain.canonical_head.fork_choice_read_lock();
-    let block_parent_root = data_column.block_parent_root();
-
-    // Do not process a column that does not descend from the finalized root.
-    if !fork_choice.is_finalized_checkpoint_or_descendant(block_parent_root) {
-        return Err(GossipDataColumnError::NotFinalizedDescendant { block_parent_root });
-    }
 
     // We have already verified that the column is past finalization, so we can
     // just check fork choice for the block's parent.
+    let block_parent_root = data_column.block_parent_root();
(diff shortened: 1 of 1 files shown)


beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -626,21 +626,22 @@ fn verify_parent_block_and_finalized_descendant<T: BeaconChainTypes>(
...
     let Some(parent_block) = fork_choice.get_block(&block_parent_root) else {
         return Err(GossipDataColumnError::ParentUnknown {
             parent_root: block_parent_root,
         });
     };
 
+    // Do not process a column that does not descend from the finalized root.
+    // We just loaded the parent_block, so we can be sure that it exists in fork choice.
+    if !fork_choice.is_finalized_checkpoint_or_descendant(block_parent_root) {
+        return Err(GossipDataColumnError::NotFinalizedDescendant { block_parent_root });
```
michaelsproul, line 642: I guess this is unreachable, but I'm ok with keeping the PR as a straight-up revert, and tracking the fix here:

- https://github.com/sigp/lighthouse/issues/8216

michaelsproul (approved):
LGTM, I think this is the best we can do on short notice.

Would appreciate another set of eyes on this, as I missed the issue with the original PR!

pawanjay176 (approved):

PR #8219, opened 2025-10-16
Combining (non)-distributed selection proofs

### LINKED ISSUE #7983: De-duplicate sync selection proof logic (distributed vs non-distributed)
We have added new logic for DVT in this PR:

- https://github.com/sigp/lighthouse/pull/7016

One of the goals was to keep the non-DVT code as similar as possible to its previous incarnation. However, this has lead to quite a bit of duplication. This issue is to track the de-duplication of this logic, by harmonising the distributed and non-distributed cases. E.g. by using `FuturesUnordered` for both.

nadtech-hub:
This one should close [#7983](https://github.com/sigp/lighthouse/issues/7983)


validator_client/validator_services/src/sync.rs
```diff
@@ -607,187 +606,81 @@ pub async fn fill_in_aggregation_proofs<S: ValidatorStore, T: SlotClock + 'stati
...
-                        warn!(
-                            validator_index,
-                            %slot,
-                            "error" = ?e,
-                            "Error determining is_aggregator"
-                        );
-                    }
-                }
+        for (validator_start_slot, duty) in pre_compute_duties {
+            if slot < *validator_start_slot {
```
nadtech-hub, line 626: doubting to keep this case and was wondering why it was only seen in distributed mode
@michaelsproul

validator_client/validator_services/src/sync.rs
```diff
@@ -607,187 +606,81 @@ pub async fn fill_in_aggregation_proofs<S: ValidatorStore, T: SlotClock + 'stati
...
+                continue;
             }
-        } else {
-            // For non-distributed mode
-            debug!(
-                period = sync_committee_period,
-                %current_slot,
-                %pre_compute_slot,
-                "Calculating sync selection proofs"
-            );
```
chong-he, line 697: This log is completely removed, but we would want this log in the normal/non-distributed case, together with the log about `Finished computing sync selection proofs`, so that we can know how long it took to compute the selection proofs. 

But for distributed case, this will not be accurate because it involves sending the request to the middleware and waiting to hear from them. What we could do is to add a condition that under distributed/parallel_sign case, we don't log this

validator_client/validator_services/src/sync.rs
```diff
@@ -607,187 +606,81 @@ pub async fn fill_in_aggregation_proofs<S: ValidatorStore, T: SlotClock + 'stati
...
                 }
             }
-
-            if num_validators_updated > 0 {
-                debug!(
-                    %slot,
-                    updated_validators = num_validators_updated,
-                    "Finished computing sync selection proofs"
-                );
-            }
```
chong-he, line 791: Refer to the comment about `calculating sync selection proofs`, the same applies here

validator_client/validator_services/src/sync.rs
```diff
@@ -607,187 +606,81 @@ pub async fn fill_in_aggregation_proofs<S: ValidatorStore, T: SlotClock + 'stati
...
 
-            // Add to global storage (we add regularly so the proofs can be used ASAP).
+        while let Some(result) = futures_unordered.next().await {
+            let Some((validator_index, proof_slot, subnet_id, proof)) = result else {
+                continue;
+            };
             let sync_map = duties_service.sync_duties.committees.read();
             let Some(committee_duties) = sync_map.get(&sync_committee_period) else {
-                debug!(period = sync_committee_period, "Missing sync duties");
+                debug!("period" = sync_committee_period, "Missing sync duties");
```
chong-he, line 662: I know the original code has the double quote but it's not necessary to have that
```suggestion
                debug!(period = sync_committee_period, "Missing sync duties");
```

validator_client/validator_services/src/sync.rs
```diff
@@ -607,187 +606,81 @@ pub async fn fill_in_aggregation_proofs<S: ValidatorStore, T: SlotClock + 'stati
...
+            match proof.is_aggregator::<S::E>() {
+                Ok(true) => {
+                    if let Some(Some(duty)) = validators.get(&validator_index) {
+                        debug!(
+                            validator_index,
+                            "slot" = %proof_slot,
+                            "subcommittee_index" = *subnet_id,
+                            // log full selection proof for debugging
+                            "full selection proof" = ?proof,
+                            "Validator is sync aggregator"
```
chong-he, line 678: In the interest of combining both normal and distributed cases, the "full selection proof" will make it confusing, I suggest we just stick to selection_proof, and keep the comment to indicate it is a full selection proof

Also we can remove the double quotes in the fields, I am not sure why I put them there in the past

```suggestion
                            validator_index,
                            slot = %proof_slot,
                            subcommittee_index = *subnet_id,
                            // In distributed mode, this is the full selection proof
                            selection_proof = ?proof,
                            "Validator is sync aggregator"
```

validator_client/validator_services/src/sync.rs
```diff
@@ -607,187 +606,81 @@ pub async fn fill_in_aggregation_proofs<S: ValidatorStore, T: SlotClock + 'stati
...
             };
+
             let validators = committee_duties.validators.read();
-            let num_validators_updated = validator_proofs.len();
 
-            for (validator_index, proofs) in validator_proofs {
-                if let Some(Some(duty)) = validators.get(&validator_index) {
-                    duty.aggregation_duties.proofs.write().extend(proofs);
-                } else {
-                    debug!(
```
chong-he, line 777: We can keep this part of the code until the end of this debug log

chong-he (commented):
Hi @nadtech-hub , can you update the target branch to `unstable`? 

Some review comments below, thanks

validator_client/validator_services/src/sync.rs
chong-he: It's only in the non-distributed mode now, I can't recall why I didn't have that in the distributed mode, but I think we can keep this

nadtech-hub:
I'll update tonight, sorry for delay

validator_client/validator_services/src/sync.rs
nadtech-hub: your comment is confusing, logging time in parallel mode is incorrect but the debug needs to be activated in parallel mode anyway

chong-he: sorry, I miss a "don't" in the comment, edited as above. What I mean is, under parallel_sign case, we don't log this; and under normal case, we log it (which is what it does currently)

chong-he:
> I'll update tonight, sorry for delay

Hi, would you be able to address the comments?

nadtech-hub:
> > I'll update tonight, sorry for delay
> 
> 
> 
> Hi, would you be able to address the comments? 

Hi, I'll do definitely. Apologize for such long delay

validator_client/validator_services/src/sync.rs
nadtech-hub: don't quite understand what you meant to restore. Specify lines

nadtech-hub: @chong-he please, clarify in more details

.github/workflows/nightly-tests.yml
```diff
@@ -6,6 +6,11 @@ on:
     # Run at 8:30 AM UTC every day
     - cron: '30 8 * * *'
   workflow_dispatch: # Allow manual triggering
+    inputs:
```
chong-he, line 9: There should be no changes in this file

chong-he (changes requested):
CI is failing, also there are some changes that shouldn't be there, could you fix this? Thanks

validator_client/validator_services/src/sync.rs
nadtech-hub: are all the attempts of updates need be counted or successful ones only?

PR #8223, opened 2025-10-16
Prepare `sensitive_url` for `crates.io`

macladson:
## Issue Addressed

Another good candidate for publishing separately from Lighthouse is `sensitive_url` as it's a general utility crate and not related to Ethereum. This PR prepares it to be spun out into its own crate.

## Proposed Changes

I've made the `full` field on `SensitiveUrl` private and instead provided an explicit getter called `.expose_full()`. It's a bit ugly for the diff but I prefer the explicit nature of the getter.
I've also added some extra tests and doc strings along with feature gating `Serialize` and `Deserialize` implementations behind the `serde` feature.

## Additional Info

I also removed `AsRef<str>` as this was previously used quite awkwardly and instead I've implemented a `.redacted()` getter which I think is cleaner.

### DIFF
--- a/validator_manager/src/list_validators.rs
+++ b/validator_manager/src/list_validators.rs
@@ -134,8 +134,7 @@ async fn run<E: EthSpec>(config: ListConfig) -> Result<Vec<SingleKeystoreRespons
     if let Some(ref beacon_url) = beacon_url {
         for validator in &validators_to_display {
             let beacon_node = BeaconNodeHttpClient::new(
-                SensitiveUrl::parse(beacon_url.as_ref())
-                    .map_err(|e| format!("Failed to parse beacon http server: {:?}", e))?,
+                beacon_url.clone(),
                 Timeouts::set_all(Duration::from_secs(12)),
             );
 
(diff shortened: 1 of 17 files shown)


validator_manager/src/list_validators.rs
```diff
@@ -134,8 +134,7 @@ async fn run<E: EthSpec>(config: ListConfig) -> Result<Vec<SingleKeystoreRespons
     if let Some(ref beacon_url) = beacon_url {
         for validator in &validators_to_display {
             let beacon_node = BeaconNodeHttpClient::new(
-                SensitiveUrl::parse(beacon_url.as_ref())
-                    .map_err(|e| format!("Failed to parse beacon http server: {:?}", e))?,
```
macladson, line 138: I'm not sure why we were re-parsing `beacon_url` since it's already a `SensitiveUrl` so I think we can just clone here instead.

michaelsproul: Agree yeah

michaelsproul (approved):
Nice! Keen!

michaelsproul (approved):

michaelsproul:
@mergify requeue

PR #8225, opened 2025-10-16
Fix boot node CLI flag naming inconsistency

sashaodessa:
## Description

Fixes CLI flag naming inconsistency in boot node configuration.

## Proposed Changes

- Rename `--enr-port` flag to `--enr-udp-port` in `boot_node/src/cli.rs` to match the argument identifier and maintain consistency with the adjacent `--enr-udp6-port` flag

## Additional Info

The mismatch between the argument identifier (`"enr-udp-port"`) and the CLI flag name (`--enr-port`) prevented users from correctly setting the ENR UDP port, as the flag would not be recognized. This fix ensures the boot node CLI is consistent and functional.

### DIFF
--- a/boot_node/src/cli.rs
+++ b/boot_node/src/cli.rs
@@ -76,7 +76,7 @@ pub fn cli_app() -> Command {
         )
         .arg(
             Arg::new("enr-udp-port")
-                .long("enr-port")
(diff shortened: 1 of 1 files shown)


chong-he:
Closing as low impact / trivial change

Issue #4525, opened 2023-07-21
Lighthouse `v4.3.0` doesn't send `Eth-Consensus-Version` header


xenowits:
## Description

The `/eth/v1/validator/blinded_blocks` [endpoint](https://ethereum.github.io/beacon-APIs/#/Validator/produceBlindedBlock) doesn't send the `Eth-Consensus-Version` header in the latest lighthouse release `v4.3.0`.

This breaks compatibility with [clients](https://github.com/attestantio/go-eth2-client/blob/master/http/http.go#L271-L275) that consume this API which errors when it can't find the `Eth-Consensus-Version` header in the response.

## Version

Lighthouse/v4.3.0-dfcb336/x86_64-linux

## Present Behaviour

Lighthouse BN **doesn't** return the `Eth-Consensus-Version` header in the response to API call to `/eth/v1/validator/blinded_blocks` endpoint.

See this screenshot:
<img width="1402" alt="Screenshot 2023-07-20 at 20 58 44" src="https://github.com/sigp/lighthouse/assets/43061995/ed040bee-86b8-49a6-8eda-39e71288f236">

```
curl --location 'http://lighthouse-bn-v4.3.0/eth/v1/validator/blinded_blocks/{slot}?randao_reveal=0xc00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000&skip_randao_verification='
```

## Expected Behaviour

Lighthouse BN should return the `Eth-Consensus-Version` header in the response to API call to `/eth/v1/validator/blinded_blocks` endpoint.

## Steps to resolve

None.

jimmygchen:
Thanks for raising this bug! It looks like we're also missing the `Eth-Consensus-Version` header in the `GET /eth/v1/validator/blocks` endpoint as well.

eserilev:
can i pick this one up?

Gua00va:
I would like to work on this issue.

jimmygchen:
Completed in #4528  🎉

PR #8227, opened 2025-10-16
Gloas containers additions

shane-moore:
## Changes
- renamed execution payload header to execution payload bid ([#4525](https://github.com/ethereum/consensus-specs/pull/4525))
- removed `builder_payment_threshold_numerator` and `builder_payment_threshold_denominator` from `chain_spec` `Config` since are constants per the [spec](https://ethereum.github.io/consensus-specs/specs/gloas/beacon-chain/#misc)
- `PayloadAttestationMessage.signature` type  updated to `Signature` since is an unaggregated signature
- add some envelope helpers, which are useful in later VC -> beacon node envelope requests PR's

@ethDreamer

### DIFF
--- a/consensus/types/src/execution_payload_envelope.rs
+++ b/consensus/types/src/execution_payload_envelope.rs
@@ -4,7 +4,6 @@ use beacon_block_body::KzgCommitments;
 use derivative::Derivative;
 use serde::de::{Deserializer, Error as _};
 use serde::{Deserialize, Serialize};
-use ssz::{Decode, DecodeError};
 use ssz_derive::{Decode, Encode};
 use superstruct::superstruct;
 use test_random_derive::TestRandom;
@@ -68,19 +67,6 @@ impl<'a, E: EthSpec> ExecutionPayloadEnvelopeRef<'a, E> {
         match self {
             Self::Gloas(envelope) => ExecutionPayloadRef::Gloas(&envelope.payload),
             Self::NextFork(envelope) => ExecutionPayloadRef::Gloas(&envelope.payload),
-        }
-    }
-}
-
-impl<E: EthSpec> ExecutionPayloadEnvelope<E> {
-    /// Custom SSZ decoder that takes a `ForkName` as context.
-    pub fn from_ssz_bytes_for_fork(bytes: &[u8], fork_name: ForkName) -> Result<Self, DecodeError> {
-        match fork_name {
-            ForkName::Gloas => ExecutionPayloadEnvelopeGloas::from_ssz_bytes(bytes)
-                .map(ExecutionPayloadEnvelope::Gloas),
-            _ => Err(DecodeError::BytesInvalid(format!(
-                "ExecutionPayloadEnvelope does not support fork {fork_name:?}"
-            ))),
         }
     }
 }
--- a/consensus/types/src/execution_payload_envelope.rs
+++ b/consensus/types/src/execution_payload_envelope.rs
@@ -4,6 +4,7 @@ use beacon_block_body::KzgCommitments;
 use derivative::Derivative;
 use serde::de::{Deserializer, Error as _};
 use serde::{Deserialize, Serialize};
+use ssz::{Decode, DecodeError};
 use ssz_derive::{Decode, Encode};
 use superstruct::superstruct;
 use test_random_derive::TestRandom;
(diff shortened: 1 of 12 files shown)


consensus/types/src/execution_payload_envelope.rs
```diff
@@ -70,6 +72,19 @@ impl<'a, E: EthSpec> ExecutionPayloadEnvelopeRef<'a, E> {
...
+        match fork_name {
+            ForkName::Gloas => ExecutionPayloadEnvelopeGloas::from_ssz_bytes(bytes)
+                .map(ExecutionPayloadEnvelope::Gloas),
+            _ => Err(DecodeError::BytesInvalid(format!(
+                "ExecutionPayloadEnvelope does not support fork {fork_name:?}"
+            ))),
+        }
+    }
+}
+
```
ethDreamer: This function isn't used so I would leave it out. We have other ways of doing these things if it is needed.

consensus/types/src/signed_execution_payload_envelope.rs
```diff
@@ -48,6 +48,26 @@ pub struct SignedExecutionPayloadEnvelope<E: EthSpec> {
 }
 
 impl<E: EthSpec> SignedExecutionPayloadEnvelope<E> {
+    /// Create a new `SignedExecutionPayloadEnvelope` from an `ExecutionPayloadEnvelope` and `Signature`.
+    pub fn from_envelope(envelope: ExecutionPayloadEnvelope<E>, signature: Signature) -> Self {
```
ethDreamer, line 52: This function is also not used so I would leave it out.

consensus/types/src/execution_payload_envelope.rs
shane-moore: yes, alternatively, we could match on `fork_name`, kind of like this:
```
SupportedProtocol::BlocksByRangeV2 => match fork_name {
            Some(ForkName::Altair) => Ok(Some(RpcSuccessResponse::BlocksByRange(Arc::new(
                SignedBeaconBlock::Altair(SignedBeaconBlockAltair::from_ssz_bytes(decoded_buffer)?),
            )))),

            Some(ForkName::Base) => Ok(Some(RpcSuccessResponse::BlocksByRange(Arc::new(
                SignedBeaconBlock::Base(SignedBeaconBlockBase::from_ssz_bytes(decoded_buffer)?),
            )))),
```

the idea was that it would be perhaps more extensible to use `from_ssz_bytes_for_fork` so you wouldn't have to add variants for new forks in various places across the repo. just add the fork variant once to `SignedExecutionPayloadEnvelope::from_ssz_bytes_for_fork` instead, so you can just do things like:
```
   Ok(Some(RpcSuccessResponse::ExecutionPayloadEnvelopesByRange(
                        Arc::new(SignedExecutionPayloadEnvelope::from_ssz_bytes_for_fork(
                            decoded_buffer,
                            fork_name,
                        )?),
                    )))
```

i used the `from_ssz_bytes_for_fork` in a couple p2p boilerplate PR's for `ExecutionPayloadEnvelopesByRange` and `ExecutionPayloadEnvelopesByRoot`
- https://github.com/shane-moore/lighthouse/blob/2abfbdab4814278a8ca64aa8119ffcb42ba0d25d/beacon_node/lighthouse_network/src/rpc/codec.rs#L741-L749
- https://github.com/shane-moore/lighthouse/blob/b2de0c5e01b2ff3ce50dfaee3450ca461d6c1fbb/beacon_node/lighthouse_network/src/rpc/codec.rs#L778-L783

but can easily update them to just use `match fork_name` approach instead. is preferable to kill `from_ssz_bytes_for_fork` and go this route?

consensus/types/src/signed_execution_payload_envelope.rs
shane-moore: i think this helper is just used later in this VC block production [PR](https://github.com/shane-moore/lighthouse/pull/15) here:
https://github.com/shane-moore/lighthouse/blob/4eac78c70d1b1cb046180d21b6f81ffa5d617d9a/validator_client/lighthouse_validator_store/src/lib.rs#L760-L784

perhaps I just add the `from_envelope` to that PR instead?

consensus/types/src/execution_payload_envelope.rs
ethDreamer: it's rare that we need a method like this because usually when we decode from the network we have context bytes to tell us what we decode and then we would do something like:
```rust
ExecutionPayloadEnvelope::Gloas(<_>::from_ssz_bytes(bytes))
```

when we *do* need a method for decoding by fork variants, we have traits for that. Since the future fork doesn't even exist yet, it's best to leave out for now.

consensus/types/src/signed_execution_payload_envelope.rs
ethDreamer: no that's fine we can leave this in then

consensus/types/src/execution_payload_envelope.rs
shane-moore: i feel you on that! thanks for the context about context bytes ha

just removed `SignedExecutionPayloadEnvelope::from_ssz_bytes_for_fork` in latest commit

Issue #8224, opened 2025-10-16
`getHeader` fails to deserialize json responses


jimmygchen:
## Description

The `BuilderHttpClient::get_builder_header` method fails to deserialize JSON responses from builder middleware like Vouch. The bug wasn't caught in testing because Lighthouse and `mev-boost` use SSZ by default, bypassing the broken JSON path.

Thanks for the report from a lighthouse-vouch user 🙏 

##  Version

`unstable`

##  Present Behaviour

When a builder returns spec-compliant JSON:

```json
{
    "version": "electra",
    "data": {
      "message": {...},
      "signature": "0x..."
    }
  }
```

Lighthouse fails with:

```
InvalidJson(Error("missing field `message`", line: 1, column: 2427))
```

The code at `beacon_node/builder_client/src/lib.rs:166-177` tries to deserialize `SignedBuilderBid` directly from the response body, but the body contains a `ForkVersionedResponse` wrapper. It looks for message at the top level instead of within the nested data field.

##  Expected Behaviour

Should successfully deserialize JSON responses with the `ForkVersionedResponse` wrapper format.

jimmygchen:
Fixed in #8228

PR #8228, opened 2025-10-16
Fix `get_header` JSON deserialization.

jimmygchen:
## Issue Addressed

#8224

## Proposed Changes

## Additional Info

### DIFF
--- a/beacon_node/builder_client/src/lib.rs
+++ b/beacon_node/builder_client/src/lib.rs
@@ -538,6 +538,13 @@ impl BuilderHttpClient {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use eth2::types::builder_bid::{BuilderBid, BuilderBidFulu};
+    use eth2::types::test_utils::{SeedableRng, TestRandom, XorShiftRng};
+    use eth2::types::{MainnetEthSpec, Signature};
+    use mockito::{Matcher, Server, ServerGuard};
+    use regex::Regex;
+
+    type E = MainnetEthSpec;
 
     #[test]
     fn test_headers_no_panic() {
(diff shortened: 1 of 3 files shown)


beacon_node/builder_client/src/lib.rs
```diff
@@ -543,9 +535,47 @@ impl BuilderHttpClient {
...
+/// Deserialize a JSON response containing a `ForkVersionedResponse` wrapper.
+///
+/// This handles the case where the response body contains:
+/// ```json
+/// {
+///   "version": "fulu",
+///   "data": { ... }
+/// }
+/// ```
+fn deserialize_json_fork_versioned_response<T>(
```
jimmygchen: Oh I just noticed I can just move this to implementation for `Deserialize`

michaelsproul: I think ForkVersionedResponse might already have an impl like this?

michaelsproul:
Nice catch!

beacon_node/builder_client/src/lib.rs
jimmygchen: oh yeah you're right let me fix it

beacon_node/builder_client/src/lib.rs
```diff
@@ -556,4 +555,144 @@ mod tests {
...
+        assert_eq!(response, mock_response_body);
+    }
+
+    fn mock_get_header_response(
+        server: &mut ServerGuard,
+        header_version_opt: Option<&str>,
+        content_type: ContentType,
+        response_body: ForkVersionedResponse<SignedBuilderBid<E>>,
+    ) {
+        let path_pattern = Regex::new(r"^/eth/v1/builder/header/\d+/.+/.+$").unwrap();
```
pawanjay176: Do we need the regex import at all here? can't we just use the regex string directly?

pawanjay176 (dismissed):
LGTM, I think we can skip the regex import in the test though

beacon_node/builder_client/src/lib.rs
jimmygchen: Good call, we don't need it. I've removed it.

michaelsproul (approved):
Looks good, let's merge

jimmygchen:
Tested this in our infra and seeing successful builder proposals on hoodi with json payloads. 🎉

PR #8230, opened 2025-10-17
Sync cleanups

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

1. In the batch retry logic, we were failing to set the batch state to `AwaitingDownload` before attempting a retry. This PR sets it to `AwaitingDownload` before the retry and sets it back to `Downloading` if the retry suceeded in sending out a request
2. Remove all peer scoring logic from retrying and rely on just de priorotizing the failed peer. I finally concede the point to @dapplion 😄 
3. Changes `block_components_by_range_request` to accept `block_peers` and `column_peers`. This is to ensure that we use the full synced peerset for requesting columns in order to avoid splitting the column peers among multiple head chains. During forward sync, we want the block peers to be the peers from the syncing chain and column peers to be all synced peers from the peerdb.
Also, fixes a typo and calls `attempt_send_awaiting_download_batches` from more places

### DIFF
--- a/beacon_node/network/src/sync/backfill_sync/mod.rs
+++ b/beacon_node/network/src/sync/backfill_sync/mod.rs
@@ -895,7 +895,7 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
                 request,
                 RangeRequestId::BackfillSync { batch_id },
                 &synced_peers,
-                &synced_peers, // column peers and block peers are same for backfill
+                &synced_peers, // All synced peers have imported up to the finalized slot so they must have their custody columns available
                 &failed_peers,
             ) {
                 Ok(request_id) => {
--- a/beacon_node/network/src/sync/backfill_sync/mod.rs
+++ b/beacon_node/network/src/sync/backfill_sync/mod.rs
@@ -313,7 +313,6 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
                     CouplingError::DataColumnPeerFailure {
                         error,
                         faulty_peers,
-                        action,
                         exceeded_retries,
                     } => {
                         debug!(?batch_id, error, "Block components coupling error");
@@ -325,11 +324,8 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
                             failed_columns.insert(*column);
                             failed_peers.insert(*peer);
                         }
-                        for peer in failed_peers.iter() {
-                            network.report_peer(*peer, *action, "failed to return columns");
-                        }
 
-                        // Only retry if peer failure **and** retries have been exceeded
+                        // Only retry if peer failure **and** retries haven't been exceeded
                         if !*exceeded_retries {
                             return self.retry_partial_batch(
                                 network,
--- a/beacon_node/network/src/sync/backfill_sync/mod.rs
+++ b/beacon_node/network/src/sync/backfill_sync/mod.rs
@@ -210,7 +210,7 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
                     .network_globals
                     .peers
                     .read()
-                    .synced_peers_for_epoch(self.to_be_downloaded, None)
+                    .synced_peers_for_epoch(self.to_be_downloaded)
                     .next()
                     .is_some()
                     // backfill can't progress if we do not have peers in the required subnets post peerdas.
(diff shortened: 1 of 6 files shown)


beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -899,6 +895,7 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
                 request,
                 RangeRequestId::BackfillSync { batch_id },
                 &synced_peers,
+                &synced_peers, // column peers and block peers are same for backfill
```
dapplion:
```suggestion
                &synced_peers, // All synced peers have imported up to the finalized slot so they must have their custody columns available
```

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -1013,7 +1015,8 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
                     chain_id: self.id,
                     batch_id,
                 },
-                &synced_peers,
+                &self.peers,
+                &synced_column_peers,
```
dapplion, line 1034:
```suggestion
                // Request blocks only from peers of this specific chain
                &self.peers,
                // Request column from all synced peers, even if they are not part of this chain. Since we request by root
                // this is safe. We request from peers outside the chain to reduce the chance of getting stuck in the case
                // this chain is useful but it has very few peers that agree on the exact same head root.
                &synced_column_peers,
```

dapplion (approved):
Looks good! Just two suggestions to document behaviour

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -883,12 +882,12 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
...
                         }
-                        for peer in failed_peers.iter() {
-                            network.report_peer(*peer, *action, "failed to return columns");
-                        }
                         // Retry the failed columns if the column requests haven't exceeded the
                         // max retries. Otherwise, remove treat it as a failed batch below.
                         if !*exceeded_retries {
+                            // Set the batch back to `AwaitingDownload` before retrying.
+                            // This is to ensure that the batch doesn't get stuck in `Downloading` state.
+                            batch.download_failed(None)?;
```
jimmygchen: Should we check the returned outcome here?
What happens if it returns `Ok(BatchState::Failed)`?

Looking at the code it looks like we may hit this?
https://github.com/sigp/lighthouse/blob/fb77ce9e192e6e478e56ce5423ffe0b43b3c4519/beacon_node/network/src/sync/range_sync/batch.rs#L344-L349

pawanjay176: Great catch. I realised I was also incrementing the batch failed count when we shouldn't be as we want to keep the batch retries separate from the column retries in the original retry PR.
Fixed in ac8adb4 . Please take a look

beacon_node/network/src/sync/range_sync/chain.rs
```diff
@@ -883,12 +882,22 @@ impl<T: BeaconChainTypes> SyncingChain<T> {
...
-                        }
                         // Retry the failed columns if the column requests haven't exceeded the
                         // max retries. Otherwise, remove treat it as a failed batch below.
                         if !*exceeded_retries {
+                            // Set the batch back to `AwaitingDownload` before retrying.
+                            // This is to ensure that the batch doesn't get stuck in `Downloading` state.
+                            //
+                            // DataColumn retries has a retry limit so calling `downloading_to_awaiting_download`
+                            // is safe.
+                            if let BatchOperationOutcome::Failed { blacklist } =
```
jimmygchen, line 893: `BatchOperationOutcome::Failed` is no longer reachable after changing the call  to `downloading_to_awaiting_download`.

Also i think the function name conveys clearly what it does, but not its intended usage, but I also struggle to come up with something better. 

`batch.download_failed_skip_attempt_count` seems a bit too wordy

jimmygchen: I think handling `BatchOperationOutcome::Failed` is a good idea though.

jimmygchen (approved):
Looks good to me too! Thanks for the cleanup

jimmygchen:
@mergify requeue

jimmygchen:
@mergify requeue

jimmygchen:
@mergify requeue

PR #8231, opened 2025-10-17
Feature gate test CLI flags

### LINKED ISSUE #6980: Remove or feature gate test CLI flags before PeerDAS release
A few test CLI flags are being added as part of testing PeerDAS, these are for devnet testing only, and should be removed from the `unstable` branch before the PeerDAS mainnet release once it's no longer needed.

This includes:
- [`--malicious-withhold-count`](https://github.com/sigp/lighthouse/blob/62a0f25f97249029c9b7efeb443dd8d085ecc5ed/beacon_node/src/cli.rs#L64)
- https://github.com/sigp/lighthouse/issues/6919
- https://github.com/sigp/lighthouse/issues/6973

If it's useful for future DAS testing, we could probably keep them with feature gates.

jimmygchen:
## Issue Addressed

Closes #6980

## Proposed Changes

I think these flags may be useful in future peerdas / das testing, and would be useful to keep. Hence I've gated them behind a `testing` feature flag.

### DIFF
--- a/lighthouse/Cargo.toml
+++ b/lighthouse/Cargo.toml
@@ -81,6 +81,7 @@ malloc_utils = { workspace = true, features = ["jemalloc"] }
 malloc_utils = { workspace = true, features = [] }
 
 [dev-dependencies]
+beacon_node = { workspace = true, features = ["testing"] }
 beacon_node_fallback = { workspace = true }
 beacon_processor = { workspace = true }
 eth2 = { workspace = true }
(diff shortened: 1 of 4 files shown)


lighthouse/Cargo.toml
```diff
@@ -81,6 +81,7 @@ malloc_utils = { workspace = true, features = ["jemalloc"] }
 malloc_utils = { workspace = true, features = [] }
 
 [dev-dependencies]
+beacon_node = { workspace = true, features = ["testing"] }
```
michaelsproul, line 84: I think this works OK because we're using Cargo's v2 resolver (https://doc.rust-lang.org/cargo/reference/resolver.html#feature-resolver-version-2). I guess we need this for the CLI tests involving these flags

michaelsproul (approved):

lighthouse/Cargo.toml
jimmygchen: Good to know, thanks! 
Yes this was added to get the tests working, without having to feature gate the tests

PR #8235, opened 2025-10-20
Allow get_beacon_proposer_index at next epoch

michaelsproul:
## Proposed Changes

Continued cleanup of proposer index logic, this PR allows `get_beacon_proposer_index` to be called at the next epoch (post-Fulu), bringing it in line with `get_beacon_proposer_indices` after the changes in:

- https://github.com/sigp/lighthouse/pull/8130

This is a safe change for Fulu due to the newly introduced lookahead.

## Additional Info

- [ ] Needs new/updated tests to cover this newly added case. I think it would be simplest to extend the tests added in #8130, so will wait for that to merge.

This change is non-essential for correctness, as `get_beacon_proposer_index` is currently only called for the current epoch (`state.slot()` or `block.slot()` after state advance), and in tests.

### DIFF
--- a/consensus/types/src/beacon_state.rs
+++ b/consensus/types/src/beacon_state.rs
@@ -1190,19 +1190,29 @@ impl<E: EthSpec> BeaconState<E> {
         // Proposer indices are only known for the current epoch, due to the dependence on the
         // effective balances of validators, which change at every epoch transition.
         let epoch = slot.epoch(E::slots_per_epoch());
-        // TODO(EIP-7917): Explore allowing this function to be called with a slot one epoch in the future.
-        if epoch != self.current_epoch() {
+        if epoch != self.current_epoch() && epoch != self.next_epoch()? {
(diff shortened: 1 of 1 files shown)


consensus/types/src/beacon_state.rs
```diff
@@ -1190,19 +1190,29 @@ impl<E: EthSpec> BeaconState<E> {
         // Proposer indices are only known for the current epoch, due to the dependence on the
         // effective balances of validators, which change at every epoch transition.
         let epoch = slot.epoch(E::slots_per_epoch());
-        // TODO(EIP-7917): Explore allowing this function to be called with a slot one epoch in the future.
-        if epoch != self.current_epoch() {
+        if epoch != self.current_epoch() && epoch != self.next_epoch()? {
```
dapplion, line 1193: Would be more intuitive to move this if check after the "Post-Fulu" comment

Ari4ka (approved):

PR #8238, opened 2025-10-20
Separate fullnode & supernode devnet checkpoint sync into two separate jobs

jimmygchen:
## Issue Addressed

The supernode checkpoint sync test was failing to backfill, it seems like the blob count on devnet-3 may be too high for a github hosted runner to run _both_ a supernode and a fullnode. 

Separating it into a separate test to see if it passes. Devnet-3 is probably going to be removed soon but I think it's worth try this out as we may want to test supernode sync on sepolia.

### DIFF
--- a/.github/workflows/local-testnet.yml
+++ b/.github/workflows/local-testnet.yml
@@ -179,7 +179,7 @@ jobs:
     continue-on-error: true
     strategy:
       matrix:
-        network: [sepolia, devnet]
+        network: [sepolia, devnet-fullnode, devnet-supernode]
     steps:
       - uses: actions/checkout@v5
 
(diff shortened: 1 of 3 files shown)


chong-he:
Maybe also need to modify the `script scripts/tests/checkpoint-sync.sh`? Because currently the script will start 2 nodes each (1 fullnode and 1 supernode). But after modification, there is only 1 node in each test. 

I see that the supernode has completed the backfill, but the `devnet-supernode` CI is failing because it still has fullnode there to check (but it shouldn't as there is only 1 node after this change)

jimmygchen:
oh yeah you're right - I'll close this one now since we no longer need devnet-3 tests. We can do the new tests in a separate PR.

Issue #8218, opened 2025-10-16
CLI flag for "semi-supernode"


michaelsproul:
## Description

A flag to custody half the columns, which allows for reconstruction, was discussed this morning.

We just need a name :rofl:

- `--semi-supernode`?
- `--halfnode`
- `--midnode`
- `--half-supernode`
- `--slimnode`

## Steps to resolve

1. Decide on flag name
2. Implement CLI cruft
3. Implement support in DA checker/CGC management. I guess a `--slimnode` with lots of validators still needs to custody its full requirements, so the slimnode 50% should just act as a starting point for custody.

michaelsproul:
This is potentially useful for L2s and their node providers.

mrabino1:
--semi-supernode is great !

antondlr:
`--subsupernode` ?

nflaig:
what about `--reconstruct-blobs` or a similar more intuitive flag?

dapplion:
Why not provide the freedom to set the min cgc of a node? It defaults to 4, but you can raise it to some number like 64.

So the semi supernode is `--min-cgc 64`, supernode is `--min-cgc 128` or its alias `--supernode`

jimmygchen:
> Why not provide the freedom to set the min cgc of a node? It defaults to 4, but you can raise it to some number like 64.
> 
> So the semi supernode is `--min-cgc 64`, supernode is `--min-cgc 128` or its alias `--supernode`

Any use case other than blob retrieval (64 & 128)? 

I think we could consider these in the future - I think there are many possible blob retrieval approaches (rpc, short lived gossip subscription), and `semi-supernode` approach is a starting point and a good fit for initial deployment i think (simple and straight froward)

jimmygchen:
Another possibility is `--blob-retrieval-method [reconstruct|rpc|supernode]`

jimmygchen:
Initial implementation with `--semi-supernode` completed in #8254. 
We'll consider other ideas in future releases.

PR #8241, opened 2025-10-20
Implement semi-supernode flag

pawanjay176:
## Issue Addressed

#8218 

## Proposed Changes

// TODO


beacon_node/beacon_chain/src/chain_config.rs
```diff
@@ -118,6 +119,10 @@ pub struct ChainConfig {
     pub invalid_block_roots: HashSet<Hash256>,
     /// Disable the getBlobs optimisation to fetch blobs from the EL mempool.
     pub disable_get_blobs: bool,
+    /// The node's custody type, determining how many data columns to custody and sample.
+    /// This is derived from CLI flags and not persisted.
+    #[serde(skip)]
```
michaelsproul, line 124: Most of these settings are derived from CLI flags? I don't think we need the `serde(skip)` here, if anything I think we might need it so we can write CLI tests in `lighthouse/tests/beacon_node.rs`.

michaelsproul: (the persistence of this config using serde is only for tests)

beacon_node/beacon_chain/src/custody_context.rs
```diff
@@ -164,12 +227,18 @@ impl<E: EthSpec> CustodyContext<E> {
 
     pub fn new_from_persisted_custody_context(
         ssz_context: CustodyContextSsz,
-        is_supernode: bool,
-    ) -> Self {
-        CustodyContext {
+        node_custody_type: NodeCustodyType,
+    ) -> Result<Self, String> {
+        if ssz_context.persisted_is_supernode && node_custody_type != NodeCustodyType::Supernode {
```
michaelsproul: So we might want a DB schema change to store the `NodeCustodyType`? Then we could enforce that the node can't change "down", i.e. supernode to semi-supernode/full, or semi-supernode to full?

Does anything break if we don't block the transition from semi-supernode to full? Because this check already covers both supernode cases.

pawanjay176: Sorry I initially added this check thinking we want to prevent changes to the `NodeCustodyType`. 
But it is technically possible to do, we just need to set the correct value of `earliest_available_slot`.

One use case is all the nodes on sepolia that had to run a supernode to reconstruct all the blobs could downgrade to a semi-supernode. 
We initially wanted to avoid switching of modes but its quite easy to handle in the rpc layer by just setting the right value of the `earliest_available_slot`. So I don't see any strong reason to not to allow this

beacon_node/src/cli.rs
```diff
@@ -59,6 +59,18 @@ pub fn cli_app() -> Command {
...
+            Arg::new("semi-supernode")
+                .long("semi-supernode")
+                .action(ArgAction::SetTrue)
+                .help_heading(FLAG_HEADER)
+                .conflicts_with("supernode")
+                .help("Run in minimal reconstruction mode. This node will subscribe to and custody \
+                          half of the data columns (enough for reconstruction), enabling efficient \
+                          data availability with lower bandwidth and storage requirements compared to \
+                          a supernode, while still supporting full blob reconstruction.")
+                .display_order(0)
```
jimmygchen, line 72: looks good, just need to update the cli help in lighthouse book

beacon_node/beacon_chain/src/custody_context.rs
```diff
@@ -249,8 +367,8 @@ impl<E: EthSpec> CustodyContext<E> {
                     "Custody group count updated"
                 );
                 return Some(CustodyCountChanged {
-                    new_custody_group_count: updated_cgc,
-                    sampling_count: self.num_of_custody_groups_to_sample(effective_epoch, spec),
+                    new_custody_group_count: updated_cgc as u64,
+                    sampling_count: self.sampling_count_at_epoch(None, spec) as u64,
```
jimmygchen, line 371: I noticed `num_of_custody_groups_to_sample` and `num_of_data_columns_to_sample` are now both mapped to `sampling_count_at_epoch` - I'm sure this is intentional, but I want to make sure the reason for the previous separation was clear - it was mainly to support subnet decoupling (which isn't my favourite feature) and unifying this to a single function is likely going to break that.

See:
- https://github.com/sigp/lighthouse/pull/7683

beacon_node/beacon_chain/src/schema_change/migration_schema_v26.rs
```diff
@@ -40,7 +40,12 @@ pub fn upgrade_to_v26<T: BeaconChainTypes>(
                 info!("Migrating `CustodyContext` to v26 schema");
                 let custody_context_v2 = CustodyContextSsz {
                     validator_custody_at_head: ssz_v24.validator_custody_at_head,
-                    persisted_is_supernode: ssz_v24.persisted_is_supernode,
+                    // TODO(pawan): fix
+                    persisted_node_custody_count: if ssz_v24.persisted_is_supernode {
```
jimmygchen, line 44: Note that we'll need a new schema migration file for this change

beacon_node/beacon_chain/src/custody_context.rs
```diff
@@ -179,7 +266,38 @@ impl<E: EthSpec> CustodyContext<E> {
...
+        // If the cgc has increased across restarts because of a change in `NodeCustodyType`,
+        // we need to update the `earliest_available_slot` to the current slot to be able
+        // to serve all columns we are advertising.
+        //
+        // If the cgc has decreased, then we have a bunch of old extra columns that we can serve,
+        // so there is no need to update the `earliest_available_slot`.
+        let update_earliest_available_slot = if cgc_current > cgc_persisted {
+            true
+        } else {
+            false
```
jimmygchen, line 298: Does this assume we have previously completed column backfill?

jimmygchen:
I've done a quick pass and I feel like this is quite a risky change before the release as there's a bit of refactor and it touches quite a lot of code related to custody / sampling and may be something we want to review and test throughly. I can continue the review tomorrow.

Is it possible to have `--semi-supernode` with minimal refactor and changes? I think the restart handling could be handled in a separate PR since we don't handle that between supernode and fullnode anyway and is something we may want to think about how to handle properly, e.g. prompt user about require resync before proceeding, prune extra columns etc

pawanjay176:
> I've done a quick pass and I feel like this is quite a risky change before the release as there's a bit of refactor 

I agree. Its a lot more change than I anticipated initially. I don't think we should rush it for the release either.

jimmygchen:
Thanks for this initial PR, I've taken some of the changes from this PR and incporated in #8254 which has been merged. Closing this one. 🎉

PR #8246, opened 2025-10-20
Use `millis_from_slot_start` when comparing against reconstruction deadline

jimmygchen:
## Issue Addressed

This recent PR below changes the max reconstruction delay to be a function of slot time. However it uses `seconds_from_slot_start` when comparing (and dropping `nano`), so it might delay reconstruction on networks where the slot time isn’t a multiple of 4, e.g. on gnosis this only happens at 2s instead of 1.25s.:
- https://github.com/sigp/lighthouse/pull/8067#discussion_r2443875068

## Proposed Changes

Use `millis_from_slot_start` when comparing against reconstruction deadline

Also added some tests for reconstruction delay.


eserilev (approved):
nice catch, lgtm!

michaelsproul (approved):
Nice catch, and tests look good :ok_hand:

PR #8247, opened 2025-10-20
Fix data column rpc request

eserilev:
## Issue Addressed
Fixes an issue mentioned in this comment regarding data column rpc requests:
https://github.com/sigp/lighthouse/issues/6572#issuecomment-3400076236

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -6913,6 +6913,44 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             );
     }
 
+    /// Get the earliest epoch in which the node has met its custody requirements.
+    /// A `None` response indicates that we've met our cutody requirements up to the
+    /// column data availability window
+    pub fn earliest_custodied_data_column_epoch(&self) -> Option<Epoch> {
+        self.store
+            .get_data_column_custody_info()
+            .unwrap_or(None)
+            .and_then(|info| info.earliest_data_column_slot)
+            .map(|slot| {
+                let mut epoch = slot.epoch(T::EthSpec::slots_per_epoch());
+                // If the earliest custodied slot isn't the first slot in the epoch
+                // The node has only met its custody requirements for the next epoch.
+                if slot > epoch.start_slot(T::EthSpec::slots_per_epoch()) {
+                    epoch += 1;
+                }
+                epoch
+            })
+    }
+
+    /// The data availability boundary for custodying columns. It will just be the
+    /// regular data availability boundary unless we are near the Fulu fork epoch.
+    pub fn column_data_availability_boundary(&self) -> Option<Epoch> {
+        match self.data_availability_boundary() {
+            Some(da_boundary_epoch) => {
+                if let Some(fulu_fork_epoch) = self.spec.fulu_fork_epoch {
+                    if da_boundary_epoch < fulu_fork_epoch {
+                        Some(fulu_fork_epoch)
+                    } else {
+                        Some(da_boundary_epoch)
+                    }
+                } else {
+                    None // Fulu hasn't been enabled
+                }
+            }
+            None => None, // Deneb hasn't been enabled
+        }
+    }
+
     /// This method serves to get a sense of the current chain health. It is used in block proposal
     /// to determine whether we should outsource payload production duties.
     ///
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -6935,7 +6935,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         self.store
             .put_data_column_custody_info(slot)
             .unwrap_or_else(
-                |e| tracing::error!(error = ?e, "Failed to update data column custody info"),
+                |e| error!(error = ?e, "Failed to update data column custody info"),
             );
     }
 
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -6934,9 +6934,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
     pub fn update_data_column_custody_info(&self, slot: Option<Slot>) {
         self.store
             .put_data_column_custody_info(slot)
-            .unwrap_or_else(
-                |e| error!(error = ?e, "Failed to update data column custody info"),
-            );
+            .unwrap_or_else(|e| error!(error = ?e, "Failed to update data column custody info"));
     }
 
     /// Get the earliest epoch in which the node has met its custody requirements.
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -6940,12 +6940,16 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
     }
 
     /// Get the earliest epoch in which the node has met its custody requirements.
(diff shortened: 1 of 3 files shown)


beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -4369,6 +4369,67 @@ async fn fulu_prune_data_columns_fork_boundary() {
...
+    let mut spec = ForkName::Electra.make_genesis_spec(E::default_spec());
+    let fulu_fork_epoch = Epoch::new(4);
+    spec.fulu_fork_epoch = Some(fulu_fork_epoch);
+    let db_path = tempdir().unwrap();
+    let store = get_store_generic(&db_path, StoreConfig::default(), spec);
+
+    if !store.get_chain_spec().is_peer_das_scheduled() {
+        // No-op if PeerDAS not scheduled.
+        panic!("PeerDAS not scheduled");
+        //return;
```
michaelsproul:
```suggestion
```

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -4369,6 +4369,67 @@ async fn fulu_prune_data_columns_fork_boundary() {
...
+    }
+
+    let harness = get_harness(store.clone(), LOW_VALIDATOR_COUNT);
+
+    // The column da boundary should be the fulu fork epoch
+    assert_eq!(
+        harness.chain.column_data_availability_boundary(),
+        Some(fulu_fork_epoch)
```
michaelsproul, line 4390: Just noting that this is in the future from the node's PoV. I think the impact of this is minimal, it maybe makes us attempt to respond to `DataColumnsByRange` requests from peers before the Fulu epoch is reached?

This doesn't seem to be a huge issue, as I think the block root iterator here will just be empty:

https://github.com/sigp/lighthouse/blob/358f5fc8dbdb947f61a6e6021c72bb3bc0ccd40a/beacon_node/network/src/network_beacon_processor/rpc_methods.rs#L1252-L1253

We have the same behaviour for blobs by range as well (when Deneb is scheduled but hasn't happened yet). I think this is OK, it's very cheap for us to respond to this request, so I think it doesn't pose a DoS risk

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -4369,6 +4369,67 @@ async fn fulu_prune_data_columns_fork_boundary() {
...
+async fn test_earliest_custodied_data_column_epoch() {
+    let spec = ForkName::Fulu.make_genesis_spec(E::default_spec());
+    let db_path = tempdir().unwrap();
+    let store = get_store_generic(&db_path, StoreConfig::default(), spec);
+    let custody_info_epoch = Epoch::new(4);
+
+    if !store.get_chain_spec().is_peer_das_scheduled() {
+        // No-op if PeerDAS not scheduled.
+        panic!("PeerDAS not scheduled");
+        //return;
```
michaelsproul:
```suggestion
```

michaelsproul (approved):
Looks good, just one small observation.

I might wait for a 2nd review before merging, seeing as this is potentially quite sensitive.

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -6939,6 +6939,44 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             );
     }
 
+    /// Get the earliest epoch in which the node has met its custody requirements.
+    /// A `None` response indicates that we've met our cutody requirements up to the
```
jimmygchen:
```suggestion
    /// A `None` response indicates that we've met our custody requirements up to the
```

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -6939,6 +6939,44 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+    /// column data availability window
+    pub fn earliest_custodied_data_column_epoch(&self) -> Option<Epoch> {
+        self.store
+            .get_data_column_custody_info()
+            .unwrap_or(None)
```
jimmygchen: should we at least log the store error here?

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -1204,33 +1204,42 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                        column_data_availability_boundary_slot
+                    } else {
+                        earliest_custodied_slot
+                    }
+                }
+                None => column_data_availability_boundary_slot,
+            };
 
-        if request_start_slot < oldest_data_column_slot {
+        if request_start_slot < earliest_custodied_data_column_slot {
```
jimmygchen, line 1231: i think `earliest_custodied_data_column_slot` gives you the slot that columns advertised at the _current_ cgc are **all** available right?

It may not necessary mean the node doesn't have a particular column. Would it be more suitable to use custdoy context to check for column availability here?

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -1204,33 +1204,42 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
             debug!(
                 %request_start_slot,
-                %oldest_data_column_slot,
-                %data_availability_boundary_slot,
+                %earliest_custodied_data_column_slot,
+                %column_data_availability_boundary_slot,
```
jimmygchen, line 1235: The log entry below needs to be updated i believe (older than `earliest_custodied_data_column_epoch`)

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
jimmygchen: ah if we're ok with not serving columns while backfilling columns that's ok too - since peers are unlikely to request from us given our `earliest_data_column_slot`.

beacon_node/beacon_chain/src/beacon_chain.rs
eserilev: [679899e](https://github.com/sigp/lighthouse/pull/8247/commits/679899eefdc417d6ae250084e84bfb6455c71975)

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
eserilev: I was under the assumption that a peer will be requesting columns from us based on our nodes status v2 response. during a cgc change our status v2 will only return the earliest slot in which we've fulfilled our custody requirements for the current cgc. So they shouldn't request anything before the `earliest_data_column_slot` after a cgc change.

On a slightly separate note, if we didn't have custody backfill I could be convinced to cross reference custody context to figure out if we could actually serve a column or not. But because of custody backfill I think we should be ok with just checking the `earliest_data_column_slot`. It keeps things simple here with the caveat that we're not serving columns that we may potentially have for some limited period of time.

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
eserilev: [9a4c63d](https://github.com/sigp/lighthouse/pull/8247/commits/9a4c63ddbbb3ea3187d923e04176c540e0ea4d1f)

beacon_node/beacon_chain/src/beacon_chain.rs
michaelsproul: nice catch Jimmy

michaelsproul:
Updates look good. Merging!

Issue #7440, opened 2025-05-12
Missing `version` in response of Beacon-API method `getPendingConsolidations`


Alleysira:
#### Overview

Hello developers.

I am conducting API testing on a Electra PoS devnet chain and noticed inconsistencies in the JSON-RPC responses among Prysm, Lighthouse(v7.0.1-e42406d/x86_64-linux), Nimbus, and Teku. 

In particular, the `/eth/v1/beacon/states/{state_id}/pending_consolidations` endpoint returns a truncated response from Lighthouse, which includes only the `data`,`execution_optimistic` and `finalized` field, whereas the `version` field as required in the spec at [`beacon-api`](https://ethereum.github.io/beacon-APIs/#/Beacon/getPendingConsolidations).

Thanks for your time.

#### Reproduce

1. Platform details (OS, architecture): Ubuntu 20.04, x86-64
2. Branch/commit used: `v7.0.1-e42406d/x86_64-linux` with geth/besu/nethermind/reth.
3. Commands being executed: I think the minimal repro is

```
curl -X GET 'http://localhost:4000/eth/v1/beacon/states/head/pending_consolidations' -H 'accept: application/json'
```

Results

```json
# lighthouse response
{"data": [], "execution_optimistic": false, "finalized": false}
# expected
{"data": [], "execution_optimistic": false, "finalized": false, "version": "electra"}
```

michaelsproul:
Thanks for the report!

We noticed this recently as well and I think it is fixed as part of this PR:

- https://github.com/sigp/lighthouse/pull/7372

The patch will be in v7.1.0

Alleysira:
> The patch will be in v7.1.0

Sorry for discussing in a closed issue. 

But I noticed that this is still not fixed in `Lighthouse/v8.0.0-rc.1-1fb94ce/x86_64-linux`. Other CL clients do not have this issue.

```
# lighthouse
{"data": [], "execution_optimistic": false, "finalized": true}
# prysm
{"data": [], "execution_optimistic": false, "finalized": false, "version": "electra"}
# teku
{"data": [], "execution_optimistic": false, "finalized": true, "version": "electra"}
# nimbus
{"data": [], "execution_optimistic": false, "finalized": true, "version": "electra"}
# lodestar
{"data": [], "execution_optimistic": false, "finalized": true, "version": "electra"}
# grandine
{"data": [], "execution_optimistic": false, "finalized": false, "version": "electra"}
```

[...]

PR #8251, opened 2025-10-21
Add `version` to the response of beacon API `getPendingConsolidations` 

chong-he:
## Issue Addressed

* #7440 

## Additional Info

Tested to confirm it works:

Before:
`{"execution_optimistic":true,"finalized":false,"data":[]}`

After:
`{"version":"electra","execution_optimistic":true,"finalized":false,"data":[]}`

### DIFF
--- a/beacon_node/http_api/src/lib.rs
+++ b/beacon_node/http_api/src/lib.rs
@@ -1236,8 +1236,8 @@ pub fn serve<T: BeaconChainTypes>(
             |state_id: StateId,
              task_spawner: TaskSpawner<T::EthSpec>,
              chain: Arc<BeaconChain<T>>| {
-                task_spawner.blocking_json_task(Priority::P1, move || {
-                    let (data, execution_optimistic, finalized) = state_id
+                task_spawner.blocking_response_task(Priority::P1, move || {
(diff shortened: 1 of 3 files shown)


michaelsproul (changes requested):
Do you mind updating the test for this endpoint too? Just to assert something about the returned `version`. Might need to update the HTTP client code in `common/eth2` as well

chong-he:
> Do you mind updating the test for this endpoint too? Just to assert something about the returned `version`. Might need to update the HTTP client code in `common/eth2` as well

Added a test in [32061c9](https://github.com/sigp/lighthouse/pull/8251/commits/32061c9baf60bcb8489ac31f0cb146a38259333e)

It seems that the `pending_deposits` and `pending_partial_withdrawals` endpoints do not return the `version` on the client side (server side is ok):
https://github.com/sigp/lighthouse/blob/e5b4983d6baf85770fe4539a565d8a2dd462bc53/common/eth2/src/lib.rs#L904-L918

It's returning `ExecutionOptimisticFinalizedResponse` instead of `ExecutionOptimisticFinalizedBeaconResponse` that contains the version info. Should we open an issue about this? (seeing that rc.2 is imminent and want to minimize changes)

michaelsproul (approved):

Issue #7650, opened 2025-06-26
Consider checking block signatures for column lookup


dapplion:
In this function https://github.com/sigp/lighthouse/blob/999b04517e35f8fce78525132d0b8e0a4816f31e/beacon_node/beacon_chain/src/data_availability_checker.rs#L228

We don't check the block signature. It's a bit sus but may not be an issue. Review if that opens DOS vectors, where an attacker triggers lookups with crappy columns

michaelsproul:
Fixed in:

- https://github.com/sigp/lighthouse/pull/8253

PR #8253, opened 2025-10-21
Improve block header signature handling

michaelsproul:
## Issue Addressed

Closes:

- https://github.com/sigp/lighthouse/issues/7650

## Proposed Changes

Reject blob and data column sidecars from RPC with invalid signatures.

## Additional Info

Two regression tests cover the simple cases. There are some more subtleties involving the events API, the slashing observation cache and the slasher which we _could_ test in future. But these are less impactful.

### DIFF
--- a/beacon_node/beacon_chain/tests/column_verification.rs
+++ b/beacon_node/beacon_chain/tests/column_verification.rs
@@ -1,7 +1,8 @@
 #![cfg(not(debug_assertions))]
 
 use beacon_chain::test_utils::{
-    AttestationStrategy, BeaconChainHarness, BlockStrategy, EphemeralHarnessType, test_spec,
+    AttestationStrategy, BeaconChainHarness, BlockStrategy, EphemeralHarnessType,
+    generate_data_column_sidecars_from_block, test_spec,
 };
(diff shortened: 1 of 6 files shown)


beacon_node/beacon_chain/tests/column_verification.rs
```diff
@@ -0,0 +1,116 @@
...
+            num_blocks,
+            BlockStrategy::OnCanonicalHead,
+            AttestationStrategy::AllValidators,
+        )
+        .await;
+
+    // Produce a block with blobs.
+    let head_state = harness.get_current_state();
+    let slot = head_state.slot() + 1;
+    let ((signed_block, opt_blobs), _) = harness.make_block(head_state, slot).await;
```
jimmygchen, line 75: I think we need to do `harness.execution_block_generator().set_min_blob_count()` to make sure there is a blob?

michaelsproul: Ooh yeah, can do that. It seems to pass as is, but that's safer

jimmygchen (approved):
Looks good to me!

PR #8254, opened 2025-10-21
Add `--semi-supernode` support

jimmygchen:
## Issue Addressed

Addresses #8218

A simplified version of #8241 for the initial release.

I've tried to minimise the logic change in this PR, although introducing the `NodeCustodyType` enum still result in quite a bit a of diff, but the actual logic change in `CustodyContext` is quite small.

The main changes are in the `CustdoyContext` struct 
* ~~combining `validator_custody_count` and `current_is_supernode` fields into a single `custody_group_count_at_head` field. We persist the cgc of the initial cli values into the `custody_group_count_at_head` field and only allow for increase (same behaviour as before).~~
* I noticed the above approach caused a backward compatibility issue, I've [made a fix](https://github.com/sigp/lighthouse/pull/8254/commits/15569bc085657b2c9a0933a20c8323a525efd30b) and changed the approach slightly (which was actually what I had originally in mind):
  * when initialising, only override the  `validator_custody_count` value if either flag `--supernode` or `--semi-supernode` is used; otherwise leave it as the existing default `0`. Most other logic remains unchanged.

All existing validator custody unit tests are still all passing, and I've added additional tests to cover semi-supernode, and restoring `CustodyContext` from disk.

Note: I've added a `WARN` if the user attempts to switch to a `--semi-supernode` or `--supernode` - this currently has no effect, but once @eserilev column backfill is merged, we should be able to support this quite easily.

Things to test
- [x] cgc in metadata / enr
- [x] cgc in metrics
- [x] subscribed subnets
- [x] getBlobs endpoint

### DIFF
--- a/beacon_node/beacon_chain/src/custody_context.rs
+++ b/beacon_node/beacon_chain/src/custody_context.rs
@@ -34,6 +34,11 @@ struct ValidatorRegistrations {
     /// So if epoch 10 and 11 has the same custody requirement, only 10 is stored.
     /// This map is never pruned, because currently we never decrease custody requirement, so this
     /// map size is contained at 128.
+    ///
+    /// If the node's is started with a cgc override (i.e. supernode/semi-supernode flag), the cgc
+    /// value is inserted into this map on initialisation with epoch set to 0. For a semi-supernode,
+    /// this means the custody requirement can still be increased if validator custody exceeds
+    /// 64 columns.
     epoch_validator_custody_requirements: BTreeMap<Epoch, u64>,
 }
 
--- a/beacon_node/beacon_chain/src/custody_context.rs
+++ b/beacon_node/beacon_chain/src/custody_context.rs
@@ -1,4 +1,5 @@
 use parking_lot::RwLock;
+use serde::{Deserialize, Serialize};
 use ssz_derive::{Decode, Encode};
 use std::marker::PhantomData;
 use std::sync::OnceLock;
@@ -6,6 +7,7 @@ use std::{
     collections::{BTreeMap, HashMap},
     sync::atomic::{AtomicU64, Ordering},
 };
+use tracing::warn;
 use types::data_column_custody_group::{CustodyIndex, compute_columns_for_custody_group};
 use types::{ChainSpec, ColumnIndex, Epoch, EthSpec, Slot};
 
@@ -36,6 +38,18 @@ struct ValidatorRegistrations {
 }
 
 impl ValidatorRegistrations {
+    /// Initialise the validator registration with some default custody requirements.
+    fn new(initial_custody_requirements: u64) -> Self {
+        let mut registrations = ValidatorRegistrations {
+            validators: Default::default(),
+            epoch_validator_custody_requirements: Default::default(),
+        };
+        registrations
+            .epoch_validator_custody_requirements
+            .insert(Epoch::new(0), initial_custody_requirements);
+        registrations
+    }
+
     /// Returns the validator custody requirement at the latest epoch.
     fn latest_validator_custody_requirement(&self) -> Option<u64> {
         self.epoch_validator_custody_requirements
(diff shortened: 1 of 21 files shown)


beacon_node/src/config.rs
```diff
@@ -108,6 +109,19 @@ pub fn get_config<E: EthSpec>(
 
     set_network_config(&mut client_config.network, cli_args, &data_dir_ref)?;
 
+    // Parse custody mode from CLI flags
+    let is_supernode = parse_flag(cli_args, "supernode");
+    let is_semi_supernode = parse_flag(cli_args, "semi-supernode");
+
+    client_config.chain.node_custody_type = if is_supernode {
+        client_config.network.subscribe_all_data_column_subnets = true;
```
jimmygchen, line 117: I've just noticed `subscribe_all_data_column_subnets` is redundant in network config, because `NetworkGlobals` already has the sampling columns when initialised, but I can do it in a follow up PR.

jimmygchen: Seems pretty straight forward to remove: https://github.com/jimmygchen/lighthouse/commit/e623eb652f84165926d07b33eddc42692c88bd4d

jimmygchen:
Some devnet-3 testing:

- All three types of nodes able to sync to head
- Sampling subnets correctly set
- Gossip topics correctly subscribed
- Able to fetch blobs from supernode and semi-supernode

beacon_node/beacon_chain/src/custody_context.rs
```diff
@@ -150,26 +205,45 @@ impl<E: EthSpec> CustodyContext<E> {
...
+    ///   changed the node's custody type via either the --supernode or --semi-supernode flags,
+    ///   and will require a resync until we implement column backfill for this scenario.
     pub fn new_from_persisted_custody_context(
         ssz_context: CustodyContextSsz,
-        is_supernode: bool,
+        node_custody_type: NodeCustodyType,
+        spec: &ChainSpec,
     ) -> Self {
+        let cgc_override = node_custody_type.get_custody_count_override(spec);
+        if let Some(cgc_from_cli) = cgc_override
```
pawanjay176, line 261: I wonder if we should just log a warn or explicitly refuse to start here and ask the user to resync if they want to change the mode.

beacon_node/beacon_chain/src/custody_context.rs
```diff
@@ -36,6 +38,23 @@ struct ValidatorRegistrations {
 }
 
 impl ValidatorRegistrations {
```
pawanjay176, line 47: Maybe we can document here that Epoch 0 is being used to store the cli parameter that the node was started with?

pawanjay176 (commented):
This is a neat idea. Its minimal changes so we can be confident that it doesn't mess with anything before the release. I have also tested this on devnet 3 and am convinved it works as advertised.
Just a few questions

beacon_node/beacon_chain/src/custody_context.rs
jimmygchen: Sounds good.
Reminded me that we need to inform the user that 
* CGC can't be reduced without a resync, i.e. once it's a supernode it can't be changed without a resync

beacon_node/beacon_chain/src/custody_context.rs
jimmygchen: As discussed earlier, we'll trigger backfill in this case for better UX. Mind if I do it in a separate PR?

beacon_node/beacon_chain/src/custody_context.rs
jimmygchen: https://github.com/sigp/lighthouse/pull/8254/commits/9a35a81c54d75a4f9c0179135e90f6b2ec0de23f

beacon_node/beacon_chain/src/custody_context.rs
pawanjay176: Sounds good.

pawanjay176 (approved):

jimmygchen:
ah I gotta pull unstable - the test are failing because of some new change from unstable.

michaelsproul:
Are we waiting for custody backfill to merge before merging this? Don't we want to add the feature to start custody backfill when the node changes to a semi-supernode/supernode?

eserilev:
if a node's cgc increases on start up, we just need two things for custody backfill to do its thing

- data column custody info needs to be updated to reflect the cgc change
- `CustodyContext::validator_registrations::epoch_validator_custody_requirements` needs to be updated to reflect the cgc change

jimmygchen:
@mergify requeue

jimmygchen:
> Are we waiting for custody backfill to merge before merging this? Don't we want to add the feature to start custody backfill when the node changes to a semi-supernode/supernode?

Oh yeah I've discussed with Pawan and going to make a follow up PR so it's bit easier to review
https://github.com/sigp/lighthouse/pull/8254#discussion_r2450049236

PR #8259, opened 2025-10-22
Remove redundant `subscribe_all_data_column_subnets` field from network

jimmygchen:
## Issue Addressed

Addresses this comment: https://github.com/sigp/lighthouse/pull/8254#discussion_r2447998786

We're currently using `subscribe_all_data_column_subnets` here to subscribe to all subnets
https://github.com/sigp/lighthouse/blob/522bd9e9c6ac167f2231525e937c9ebbcb86cf6e/beacon_node/lighthouse_network/src/types/topics.rs#L82-L92

But its unnecessary because the else path also works for supernode (uses `sampling_subnets` instead)

The big diffs will disappear once #8254 is merged.

### DIFF
--- a/beacon_node/lighthouse_network/src/discovery/enr.rs
+++ b/beacon_node/lighthouse_network/src/discovery/enr.rs
@@ -377,7 +377,8 @@ mod test {
     #[test]
     fn test_nfd_enr_encoding() {
         let spec = make_fulu_spec();
-        let enr = build_enr_with_config(NetworkConfig::default(), 4, &spec).0;
+        let enr =
+            build_enr_with_config(NetworkConfig::default(), spec.custody_requirement, &spec).0;
         assert_eq!(enr.next_fork_digest().unwrap(), TEST_NFD);
     }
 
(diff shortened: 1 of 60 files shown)


dapplion (approved):
LGTM!

@jimmygchen this commit fixes the failing test 4084d115f

beacon_node/lighthouse_network/src/discovery/enr.rs
```diff
@@ -386,56 +377,22 @@ mod test {
     #[test]
     fn test_nfd_enr_encoding() {
         let spec = make_fulu_spec();
-        let enr = build_enr_with_config(NetworkConfig::default(), None, &spec).0;
+        let enr = build_enr_with_config(NetworkConfig::default(), 4, &spec).0;
```
pawanjay176: maybe use spec.custody_requirement here instead of hardcoding?

pawanjay176 (approved):
LGTM, just one nit

beacon_node/lighthouse_network/src/discovery/enr.rs
jimmygchen: Added in https://github.com/sigp/lighthouse/pull/8259/commits/c19e963259ad7a64cb5e87beb76bd9259d8cdcbf

jimmygchen:
@dapplion thanks for the fix, I've pushed your changes.

PR #8263, opened 2025-10-22
Update spec tests to v1.6.0-beta.1

michaelsproul:
## Proposed Changes

Update the EF spec tests to v1.6.0-beta.1

There are a few new light client tests (which we pass), and some for progressive containers, which we haven't implemented (we ignore them).


jimmygchen (approved):
Nice!

PR #8265, opened 2025-10-22
Trigger backfill on startup if user switches to a supernode or semi-supernode

jimmygchen:
## Issue Addressed

This PR adds backfill functionality to nodes switching to become a supernode or semi-supernode. Please note that we currently only support a CGC increase, i.e. if the node's already custodying 67 columns, switching to semi-supernode (64) will have no effect.

## Proposed changes

From @eserilev 
> if a node's cgc increases on start up, we just need two things for custody backfill to do its thing
> 
> - data column custody info needs to be updated to reflect the cgc change
> - `CustodyContext::validator_registrations::epoch_validator_custody_requirements` needs to be updated to reflect the cgc change

- [x] Add tests
- [x] Test on devnet-3
  - [x] switch to supernode
  - [x] switch to semisupernode
- [x] Test on live testnets
- [x] Update docs (functions)

### DIFF
--- a/beacon_node/beacon_chain/src/builder.rs
+++ b/beacon_node/beacon_chain/src/builder.rs
@@ -931,18 +931,26 @@ where
 
         // Load the persisted custody context from the db and initialize
         // the context for this run
-        let custody_context = if let Some(custody) =
+        let (custody_context, cgc_changed_opt) = if let Some(custody) =
             load_custody_context::<E, THotStore, TColdStore>(store.clone())
         {
-            Arc::new(CustodyContext::new_from_persisted_custody_context(
+            let current_epoch = slot_clock
+                .now()
+                .ok_or("Unable to read slot clock")?
+                .epoch(E::slots_per_epoch());
+            CustodyContext::new_from_persisted_custody_context(
                 custody,
                 self.node_custody_type,
+                current_epoch,
                 &self.spec,
-            ))
+            )
         } else {
-            Arc::new(CustodyContext::new(self.node_custody_type, &self.spec))
+            (
+                CustodyContext::new(self.node_custody_type, &self.spec),
+                None,
+            )
         };
-        debug!(?custody_context, "Loading persisted custody context");
+        debug!(?custody_context, "Loaded persisted custody context");
 
         let beacon_chain = BeaconChain {
             spec: self.spec.clone(),
@@ -1019,7 +1027,7 @@ where
                     slot_clock,
                     self.kzg.clone(),
                     store,
-                    custody_context,
+                    Arc::new(custody_context),
                     self.spec,
                 )
                 .map_err(|e| format!("Error initializing DataAvailabilityChecker: {:?}", e))?,
@@ -1060,6 +1068,14 @@ where
                 "You must use the `--purge-db` flag to clear the database and restart sync. You may be on a hostile network."
             );
             return Err(format!("Weak subjectivity verification failed: {:?}", e));
+        }
+
+        if let Some(cgc_changed) = cgc_changed_opt {
+            // Update data column custody info if there's a CGC change from CLI flags.
+            // This will trigger column backfill.
+            let cgc_change_effective_slot =
+                cgc_changed.effective_epoch.start_slot(E::slots_per_epoch());
+            beacon_chain.update_data_column_custody_info(Some(cgc_change_effective_slot))
         }
 
         info!(
--- a/beacon_node/beacon_chain/src/builder.rs
+++ b/beacon_node/beacon_chain/src/builder.rs
@@ -1075,7 +1075,7 @@ where
             // This will trigger column backfill.
             let cgc_change_effective_slot =
                 cgc_changed.effective_epoch.start_slot(E::slots_per_epoch());
-            beacon_chain.update_data_column_custody_info(Some(cgc_change_effective_slot))
+            beacon_chain.update_data_column_custody_info(Some(cgc_change_effective_slot));
         }
 
         info!(
--- a/beacon_node/beacon_chain/src/builder.rs
+++ b/beacon_node/beacon_chain/src/builder.rs
@@ -934,14 +934,14 @@ where
         let (custody_context, cgc_changed_opt) = if let Some(custody) =
             load_custody_context::<E, THotStore, TColdStore>(store.clone())
         {
-            let current_epoch = slot_clock
-                .now()
-                .ok_or("Unable to read slot clock")?
+            let head_epoch = canonical_head
+                .cached_head()
+                .head_slot()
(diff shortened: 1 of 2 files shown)


beacon_node/beacon_chain/src/custody_context.rs
```diff
@@ -261,6 +264,11 @@ impl<E: EthSpec> CustodyContext<E> {
         if let Some(cgc_from_cli) = cgc_override
             && cgc_from_cli > ssz_context.validator_custody_at_head
         {
+            /*
+            TODO:
+            * when increased, add (next_epoch, 64) into the registration map AND update data column custody info with the next epoch to trigger backfill
+            * when reduced, log a warning, and preserve existing value?
```
jimmygchen: or re-initialise the map with just the new value. As there's an assumption that the CGC can only be increased

beacon_node/beacon_chain/src/custody_context.rs
```diff
@@ -247,39 +253,92 @@ impl<E: EthSpec> CustodyContext<E> {
...
+                });
+
+                debug!(
+                    info = "new CGC will be effective from the next epoch",
+                    ?node_custody_type,
+                    old_cgc = old_custody_group_count,
+                    new_cgc = validator_custody_at_head,
+                    effective_epoch = %effective_epoch,
+                    "Node custody type change caused a custody count increase",
+                );
```
jimmygchen, line 315: This section is the meat of the PR - if the CGC increases, we update it in a similar way to how we initialise a semi-supernode or supernode, except that we schedule the increase in the next epoch.

We return a `CustodyCountChanged` object, so we can update the `DataColumnCustodyInfo` in order to trigger column backfill:

https://github.com/jimmygchen/lighthouse/blob/8cedbd1d7f4f881195ac20dbde06ed08e1e2adca/beacon_node/beacon_chain/src/builder.rs#L1073-L1079

The rests are doc updates and new tests.

beacon_node/beacon_chain/src/builder.rs
```diff
@@ -931,18 +931,26 @@ where
 
         // Load the persisted custody context from the db and initialize
         // the context for this run
-        let custody_context = if let Some(custody) =
+        let (custody_context, cgc_changed_opt) = if let Some(custody) =
             load_custody_context::<E, THotStore, TColdStore>(store.clone())
         {
-            Arc::new(CustodyContext::new_from_persisted_custody_context(
+            let current_epoch = slot_clock
```
pawanjay176: I think this should be the `head_epoch`, not the `wall_slot_epoch` derived from the canonical_head?

I think we want the cgc change effect to happen from where we start syncing right? The wall clock epoch can be anywhere based on how much behind the node is

beacon_node/beacon_chain/src/custody_context.rs
```diff
@@ -247,39 +253,92 @@ impl<E: EthSpec> CustodyContext<E> {
...
     /// Restore the custody context from disk.
     ///
-    /// * If NodeCustodyType::custody_count < validator_custody_at_head, it means the attached
-    ///   validate stake has increased the node's CGC. We ignore the CLI input.
-    /// * If NodeCustodyType::custody_count > validator_custody_at_head, it means the user has
-    ///   changed the node's custody type via either the --supernode or --semi-supernode flags,
-    ///   and will require a resync until we implement column backfill for this scenario.
+    /// # Behavior
+    /// * If [`NodeCustodyType::get_custody_count_override`] < validator_custody_at_head, it means the attached
+    ///   validator stake has increased the node's CGC. We ignore the CLI input.
```
pawanjay176: This comment is a little unclear. Maybe something along the lines of
```
    ///   the node has a higher contribution of custody from the number of validators attached compared to the one derived from cli flags
```
Something a little less verbose than that though :P

beacon_node/beacon_chain/src/custody_context.rs
```diff
@@ -49,6 +49,10 @@ impl ValidatorRegistrations {
     ///
     /// If a `cgc_override` value is specified, the cgc value is inserted into the registration map
     /// and is equivalent to registering validator(s) with the same custody requirement.
+    ///
+    /// The node will backfill all the way back to either data_availability_boundary or fulu epoch,
+    /// and because this is a fresh node, setting the epoch to 0 is fine, as backfill will done via
```
eserilev:
```suggestion
    /// and because this is a fresh node, setting the epoch to 0 is fine, as backfill will be done via
```

beacon_node/beacon_chain/src/custody_context.rs
```diff
@@ -126,8 +129,11 @@ impl ValidatorRegistrations {
         }
     }
 
-    /// Updates the `epoch_validator_custody_requirements` map by pruning all values on/after `effective_epoch`
-    /// and updating the map to store the latest validator custody requirements for the `effective_epoch`.
+    /// Updates the `epoch -> cgc` map after backfill has been completed for
```
eserilev:
```suggestion
    /// Updates the `epoch -> cgc` map after custody backfill has been completed for
```

eserilev:
This all looks pretty good to me. I have a devnet3 node running right now and am about to finish backfilling. Will test this out shortly

eserilev (approved):
I got this working on a local devnet 3 node

The node was first started as a normal full node (cgc = 4). I then let it finish backfilling and restarted the node as a semi super node. It's now been doing custody backfill for the last hour (I still have like 4-5 hours to go). But seems like the triggering of custody backfill worked as intended. Nice work! 

I think Pawan is right that we could use the current head epoch instead of the current slot clock epoch, but seems like this would probably work either way.

The tests seem pretty good too, I'd like to piggy back off those tests in this [PR](https://github.com/sigp/lighthouse/pull/8267)

pawanjay176 (commented):
LGTM other than the calculation of current_epoch. 
The tests are great, I have also tested the same manually.

beacon_node/beacon_chain/src/builder.rs
jimmygchen: Good call! 
It'd also minimise the unnecessary CGC changes that may happen. I like the fact that it reduce the number of edge cases and is more predictable

beacon_node/beacon_chain/src/custody_context.rs
jimmygchen: Updated to
```rust
    /// * If [`NodeCustodyType::get_custody_count_override`] < validator_custody_at_head, it means
    ///   validators have increased the CGC beyond the derived CGC from cli flags. We ignore the CLI input.
```

eserilev (approved):
changes look good

pawanjay176 (approved):
LGTM

PR #8266, opened 2025-10-22
New design for blob/column pruning

michaelsproul:
## Issue Addressed

We are seeing some crazy IO utilisation on Holesky now that data columns have started to expire. Our previous approach of _iterating the entire blobs DB_ doesn't seem to be scaling.

## Proposed Changes

New blob pruning algorithm that uses a backwards block iterator from the epoch we want to prune, stopping early if an already-pruned slot is encountered.

## Additional Info

This is incompatible with intentionally gappy blob DBs, which we have thought about trying to support but have no near-term plans of implementing. We could probably make gappy DBs work by reverting to the more inefficient pruning.


jimmygchen (approved):
LGTM!

PR #8267, opened 2025-10-22
Fix CGC backfill race condition

eserilev:
## Issue Addressed

During custody backfill sync there could be an edge case where we update CGC at the same time where we are importing a batch of columns which may cause us to incorrectly overwrite values when calling `backfill_validator_custody_requirements`. To prevent this race condition, the expected cgc is now passed into this function and is used to check if the expected cgc == the current validator cgc. If the values arent equal, this probably indicates that a very recent CGC occurred so we do not prune/update values in the `epoch_validator_custody_requirements` map.

### DIFF
--- a/beacon_node/beacon_chain/tests/store_tests.rs
+++ b/beacon_node/beacon_chain/tests/store_tests.rs
@@ -3182,6 +3182,8 @@ async fn weak_subjectivity_sync_test(
     assert_eq!(store.get_anchor_info().state_upper_limit, Slot::new(0));
 }
 
+// This test prunes data columns from epoch 0 and then tries to re-import them via
+// the same code paths that custody backfill sync imports data columns
 #[tokio::test]
 async fn test_import_historical_data_columns_batch() {
     let spec = ForkName::Fulu.make_genesis_spec(E::default_spec());
@@ -3189,6 +3191,7 @@ async fn test_import_historical_data_columns_batch() {
     let store = get_store_generic(&db_path, StoreConfig::default(), spec);
     let start_slot = Epoch::new(0).start_slot(E::slots_per_epoch()) + 1;
     let end_slot = Epoch::new(0).end_slot(E::slots_per_epoch());
+    let cgc = 128;
 
     let harness = get_harness_import_all_data_columns(store.clone(), LOW_VALIDATOR_COUNT);
 
@@ -3208,6 +3211,7 @@ async fn test_import_historical_data_columns_batch() {
 
     let mut data_columns_list = vec![];
 
+    // Get all data columns for epoch 0
     for block in block_root_iter {
         let (block_root, _) = block.unwrap();
         let data_columns = harness.chain.store.get_data_columns(&block_root).unwrap();
@@ -3227,6 +3231,7 @@ async fn test_import_historical_data_columns_batch() {
 
     harness.advance_slot();
 
+    // Prune data columns
     harness
         .chain
         .store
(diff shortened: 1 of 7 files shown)


michaelsproul:
Updated the target branch to `release-v8.0`, the new branch for the release

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -3343,84 +3360,6 @@ async fn test_import_historical_data_columns_batch_mismatched_block_root() {
     ));
 }
 
-// This should verify that a data column sidecar associated to a block root that doesn't exist in the store cannot
-// be imported.
-#[tokio::test]
-async fn test_import_historical_data_columns_batch_no_block_found() {
```
eserilev, line 3349: this was a redundant test case

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -3182,13 +3182,16 @@ async fn weak_subjectivity_sync_test(
...
+// This test prunes data columns from epoch 0 and then tries to re-import them via
+// the same code paths that custody backfill sync imports data columns
 #[tokio::test]
 async fn test_import_historical_data_columns_batch() {
     let spec = ForkName::Fulu.make_genesis_spec(E::default_spec());
     let db_path = tempdir().unwrap();
     let store = get_store_generic(&db_path, StoreConfig::default(), spec);
     let start_slot = Epoch::new(0).start_slot(E::slots_per_epoch()) + 1;
     let end_slot = Epoch::new(0).end_slot(E::slots_per_epoch());
+    let cgc = 128;
```
jimmygchen, line 3194: nit: could use either `E::number_of_columns` or `harness.chain.data_availability_checker.custody_context().custody_group_count_at_head(&sepc)` instead.

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -3238,21 +3243,25 @@ async fn test_import_historical_data_columns_batch() {
...
-        .import_historical_data_column_batch(Epoch::new(0), data_columns_list)
+        .import_historical_data_column_batch(Epoch::new(0), data_columns_list, cgc)
         .unwrap();
+
     let block_root_iter = harness
         .chain
         .forwards_iter_block_roots_until(start_slot, end_slot)
         .unwrap();
 
+    // Assert that data columns now exist for epoch 0
```
jimmygchen, line 3264: Nice, these comments help! Thanks

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -3261,13 +3270,15 @@ async fn test_import_historical_data_columns_batch() {
...
 // This should verify that a data column sidecar containing mismatched block roots should fail to be imported.
+// This also covers any test cases related to data columns with incorrect/invalid/mismatched block roots.
 #[tokio::test]
 async fn test_import_historical_data_columns_batch_mismatched_block_root() {
     let spec = ForkName::Fulu.make_genesis_spec(E::default_spec());
     let db_path = tempdir().unwrap();
     let store = get_store_generic(&db_path, StoreConfig::default(), spec);
     let start_slot = Slot::new(1);
     let end_slot = Slot::new(E::slots_per_epoch() * 2 - 1);
+    let cgc = 128;
```
jimmygchen, line 3281: Same as above

jimmygchen (approved):
Nice, the logic looks correct to me. Just a small nit on hard coded values - i think the values is unlikely to change but would be good to keep hard coded values to a minimum for maintainability.

PR #8269, opened 2025-10-22
Only publish reconstructed columns that we need to sample

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

We were publishing columns all columns that we didn't already have in the da cache when reconstructing. This is unnecessary outbound bandwidth for the node that is supposed to sample fewer columns.
This PR changes the behaviour to publish only columns that we are supposed to sample in the topics that we are subscribed to.


jimmygchen (changes requested):
Looks good and a safe change to merge!

Just a failing test that needs update here:
https://github.com/sigp/lighthouse/actions/runs/18732633772/job/53432828777?pr=8269

jimmygchen (approved):

Issue #8248, opened 2025-10-20
Test improvement tracking issue


jimmygchen:
- [x] #8194
- [x] #8271 
- [x] #8319 
- [ ] Skip tests that aren't relevant for the fork when using `fork_from_env`, e.g. [prune_blobs_across_fork_boundary](https://github.com/sigp/lighthouse/blob/2f8587301d92c076c9811ac6fde3aaac2cf10418/beacon_node/beacon_chain/tests/store_tests.rs#L3854) shouldn't need to be ran for every fork, is only relevant between deneb - fulu
- [ ] Optimize slow tests - there are many that takes > 120s
- [ ] Fake Kzg Impl for some tests to speed things up

jimmygchen:
Re fake Kzg Impl, we can even turn it on in test + debug mode, that also avoids the stackoverflow issue with blobs while running tests in debug mode - and we still have the real verification covered with release tests on CI

PR #8271, opened 2025-10-23
Run CI tests only recent forks

pawanjay176:
## Issue Addressed

Partially addresses #8248

## Proposed Changes

Run the beacon chain, http and network tests only for recent forks instead of everything from phase 0.
Also added gloas also to the recent forks list. I thought that would be a good way to know if changes in the current fork affect future forks.

Not completely sure if we should run for future forks, but added it so that we can discuss here.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,10 +30,6 @@ TEST_FEATURES ?=
 # Cargo profile for regular builds.
 PROFILE ?= release
 
-# List of all hard forks. This list is used to set env variables for several tests so that
-# they run for different forks.
-FORKS=phase0 altair bellatrix capella deneb electra fulu gloas
-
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -35,7 +35,7 @@ PROFILE ?= release
 FORKS=phase0 altair bellatrix capella deneb electra fulu gloas
```
michaelsproul: May as well delete the `FORKS` variable, as it is now unused

michaelsproul (dismissed):

jimmygchen (approved):
Looks good and I've pushed @michaelsproul 's suggested changes.

jimmygchen:
PR to run all prior forks nightly here: https://github.com/sigp/lighthouse/pull/8319

PR #7923, opened 2025-08-22
Gloas(EIP-7732): Containers / Constants

ethDreamer:
## Issue Addressed

* #7850

This is the first round of the conga line! 🎉 

Just spec constants and container changes so far.

PR #8273, opened 2025-10-23
latest gloas-container fixes

shane-moore:
## Changes
After rebasing `unstable`, `gloas-containers` has some failing tests.  
https://github.com/sigp/lighthouse/pull/7923

The issue tends to be related to a test trying to produce a gloas block and then trying to access some state, such as `state.latest_execution_payload_header()`, which no longer exists post-gloas and throws an error. These types of errors are marked as TODO and will be handled in subsequent PR's related to block processing.

Note, I'm not sure about the [run-local-testnet ](https://github.com/sigp/lighthouse/actions/runs/18729898866/job/53424820223?pr=7923) failure. I suggest we review and merge this one and then see if the `run-local-testnet` failure is ephemeral or otherwise

@ethDreamer

### DIFF
--- a/beacon_node/network/src/network_beacon_processor/tests.rs
+++ b/beacon_node/network/src/network_beacon_processor/tests.rs
@@ -1705,8 +1705,9 @@ async fn test_blobs_by_range_spans_fulu_fork() {
     spec.fulu_fork_epoch = Some(Epoch::new(1));
     spec.gloas_fork_epoch = Some(Epoch::new(2));
 
+    // This test focuses on Electra→Fulu blob counts (epoch 0 to 1). Build 62 blocks since no need for Gloas activation at slot 64.
     let mut rig = TestRig::new_parametric(
-        64,
+        62,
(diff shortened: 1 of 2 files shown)


beacon_node/network/src/network_beacon_processor/tests.rs
```diff
@@ -1705,8 +1705,9 @@ async fn test_blobs_by_range_spans_fulu_fork() {
     spec.fulu_fork_epoch = Some(Epoch::new(1));
     spec.gloas_fork_epoch = Some(Epoch::new(2));
 
+    // This test focuses on Electra→Fulu blob counts (epoch 0 to 1). Build 62 blocks since no need for Gloas activation at slot 64.
```
ethDreamer, line 1708: You should add a
```rust
// TODO(gloas): switch back to 64 once blobs are built for gloas
```
comment here. We do want to re-enable this test later once blobs are built out.

Issue #8249, opened 2025-10-21
Custody info API


michaelsproul:
## Description

We're going to add an HTTP API for information about the custodied data columns:

- Custody group count (CGC)
- Custodied columns (list of indices)
- Earliest available data column slot (derived from DataColumnCustodyInfo, related: https://github.com/sigp/lighthouse/issues/6572)

I'm thinking the path for this API can be `/lighthouse/custody/info`. There's an argument it could be combined with `/lighthouse/database/info`, but some of this information is not strictly DB-related, so probably makes sense to have a separate API. Using a `/lighthouse` API means we have a lot of freedom to define and change the response format, so we can adjust to feedback from users/testing as well.

michaelsproul:
I think we can drop this from v8.0.0-rc.2 but still include it in v8.0.0. I didn't get to it today but it should be safe to add after the RC

PR #8276, opened 2025-10-23
Implement `/lighthouse/custody/info` API

michaelsproul:
## Issue Addressed

Closes:

- https://github.com/sigp/lighthouse/issues/8249

## Proposed Changes

New `/lighthouse/custody` API including:

- [x] Earliest custodied data column slot
- [x] Node CGC
- [x] Custodied columns

## Additional Info

Aiming for inclusion in v8.0.0 but not v8.0.0-rc.2

Needs:

- [x] Client code in `eth2`
- [x] Test

### DIFF
--- a/beacon_node/http_api/src/custody.rs
+++ b/beacon_node/http_api/src/custody.rs
@@ -1,21 +1,12 @@
 use beacon_chain::{BeaconChain, BeaconChainTypes};
-use serde::{Deserialize, Serialize};
+use eth2::lighthouse::CustodyInfo;
 use std::sync::Arc;
-use types::{EthSpec, Slot};
+use types::EthSpec;
 use warp_utils::reject::{custom_bad_request, custom_server_error};
-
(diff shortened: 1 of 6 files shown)


jimmygchen (approved):
Looks great! I like having the logic in a separate file from `lib.rs`.

I think it's worth adding this endpoint to the book?
https://github.com/sigp/lighthouse/blob/11bcccb353c073ab6d23690d0fd3a3b8940d8e95/book/src/api_lighthouse.md

chong-he:
> Looks great! I like having the logic in a separate file from `lib.rs`.
> 
> I think it's worth adding this endpoint to the book? https://github.com/sigp/lighthouse/blob/11bcccb353c073ab6d23690d0fd3a3b8940d8e95/book/src/api_lighthouse.md

I was thinking about this and maybe I can add this endpoint to the PR #8284 ? (as this PR is queued for merge now)

edit: nvm, 8284 also queued for merge now :sweat_smile:

Issue #7850, opened 2025-08-11
Implement Gloas Fork


ethDreamer:
This issue will track the tasks for implementing [Gloas fork](https://github.com/ethereum/consensus-specs/tree/master/specs/gloas). Expect many changes to this issue and the checklists as work evolves. Special thanks to @shane-moore who will be heavily assisting.

- [ ] [eip-7732](https://github.com/ethereum/consensus-specs/tree/master/specs/gloas)
  - [x] Add Gloas boiler plate (https://github.com/sigp/lighthouse/pull/7728)
  - [x] [Constants](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/beacon-chain.md#constants), [Preset](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/beacon-chain.md#preset) & [Configuration](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/beacon-chain.md#configuration) (https://github.com/sigp/lighthouse/pull/7923)
  - [ ] [Containers](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/beacon-chain.md#containers)
    - [x] [New containers](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/beacon-chain.md#new-containers)
      - [x] [`BuilderPendingPayment`](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/beacon-chain.md#builderpendingpayment) (https://github.com/sigp/lighthouse/pull/7923)
      - [x] [`BuilderPendingWithdrawal`](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/beacon-chain.md#builderpendingwithdrawal) (https://github.com/sigp/lighthouse/pull/7923)
      - [x] [`PayloadAttestationData`](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/beacon-chain.md#payloadattestationdata) (https://github.com/sigp/lighthouse/pull/7923)
      - [x] [`PayloadAttestation`](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/beacon-chain.md#payloadattestation) (https://github.com/sigp/lighthouse/pull/7923)
      - [x] [`PayloadAttestationMessage`](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/beacon-chain.md#payloadattestationmessage) (https://github.com/sigp/lighthouse/pull/7923)
[...]

PR #8279, opened 2025-10-23
Gloas(EIP-7732): Envelope Processing

ethDreamer:
## Issue Addressed

* #7850 

This is the second PR in the conga line! 🎉


beacon_node/store/src/partial_beacon_state.rs
michaelsproul, line 1: This file can be deleted. It is deleted on unstable by:

- https://github.com/sigp/lighthouse/pull/8591

michaelsproul:
TODO:

- [ ] Undo Michael's dodgy assumption around builder indices being validator indices (see 9e365e535ff3850714f29e1744b9eb78d9a7a3ab).
- [ ] Refactor withdrawals code to match spec (remove large blocks that are inlined)
- [ ] Introduce concept of `state.builders` (we don't have this in `BeaconState` yet)
- [ ] Possibly add another `ValidatorPubkeyCache` for the builder pubkeys

beacon_node/execution_layer/src/lib.rs
```diff
@@ -440,6 +548,10 @@ struct Inner<E: EthSpec> {
     /// This is used *only* in the informational sync status endpoint, so that a VC using this
     /// node can prefer another node with a healthier EL.
     last_new_payload_errored: RwLock<bool>,
+    /// Cache for deduplicating `notify_new_payload` requests.
+    ///
+    /// Handles both in-flight requests and recently completed requests.
+    new_payload_cache: Option<NewPayloadCache>,
```
dapplion, line 554: Why is this cache necessary?

michaelsproul: I'm down to get rid of it. Current plan for this PR is to just cherry-pick the beacon chain changes once the consensus code is stable.

Ari4ka (approved):

beacon_node/execution_layer/src/lib.rs
ethDreamer: @dapplion 

My understanding was that the [`DuplicateCache`](https://github.com/sigp/lighthouse/blob/stable/beacon_node/beacon_processor/src/lib.rs#L179) exists to prevent concurrent calls to [`process_block()`](https://github.com/sigp/lighthouse/blob/stable/beacon_node/beacon_chain/src/beacon_chain.rs#L3348) with the same block. This saves time ultimately by preventing concurrent calls of [`engine_NewPayload`](https://github.com/ethereum/execution-apis/blob/main/src/engine/cancun.md#engine_newpayloadv3) to the execution engine.

These concurrent calls happen because the head block can arrive multiple times through different sync channels. We expect this to happen **even more often** post-gloas because:
1. We expect to pump gas limits so the call will naturally take longer
2. We will call `engine_newPayload` immediately, but we won't `await` the `block_verification_handle` until after the block passes the DA cache

So the object we put into the DA cache will include a `block_verification_handle`, rather than being fully executed *before* being put into the cache as it is now. If we don't plan on doing a global block status table (which was discouraged for testing concerns), then putting the cache here seemed like the most reasonable place. All threads will have a handle to await, but only one call will be made to the EL.

michaelsproul:
Closing in favour of changes already merged:

- https://github.com/sigp/lighthouse/pull/8806

Let me know if you think we're missing anything from this branch. I don't think we are.

PR #8280, opened 2025-10-23
Create `axum_utils` and begin VC HTTP API migration 

macladson:
## Proposed Changes

- Creates `axum_utils` which contain common functions and types
- Begins migration of the VC HTTP API from warp to axum using `warpdrive`
- Migrates both metrics servers to `axum`


common/axum_utils/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+name = "axum_utils"
+version = "0.1.0"
+authors = ["Sigma Prime <contact@sigmaprime.io>"]
+edition = { workspace = true }
+
+[dependencies]
+axum = { workspace = true }
+axum-server = { version = "0.7", features = ["tls-rustls-no-provider"] }
+http = "1"
+lighthouse_version = { workspace = true }
```
dapplion: Can you remove this as dependency and make the consumer of this crate pass a version string? Would make this crate more reusable / agnostic and improve incremental compilation.

common/axum_utils/src/cors.rs
```diff
@@ -0,0 +1,229 @@
...
+        IpAddr::V6(addr) => format!("http://[{}]:{}", addr, port),
+    }
+}
+
+/// Build a CORS layer from an optional origins string and a default fallback
+///
+/// This is the main function for CLI usage:
+/// - If `allow_origin` is `Some`, parse it as comma-separated origins
+/// - If `allow_origin` is `None`, use the default IP and port
+pub fn build_cors_layer(
```
dapplion, line 159: Can you add a test for this function? Or from CLI arg to parsed CorsConfig

dapplion:
Does warpdrive mess with SSE in any way? Have you tested them?

macladson:
> Does warpdrive mess with SSE in any way? Have you tested them?

`warpdrive` has a basic example for SSE: https://github.com/macladson/warpdrive/blob/master/examples/sse.rs

Also if we find issues with it we can always migrate the SSE endpoints first

eserilev (commented):
For others reviewing:

- `warpdrive`: a nice lil crate Mac put together to convert req/resp from warp to axum. This allows us to use the existing warp server as a fallback service. When reviewing this PR, you should also review this new crate
- We should have buy-in from the team that this will be a "fast" migration from warp to axum in the validator client. We should not go multiple releases with this fallback mechanism in place in the validator client.
- We might just want to migrate the SSE endpoint in this PR and not deal with the overhead of supporting axum <-> warp SSE functionality
- We should run this on a Hoodi node and ensure were not seeing any degraded performance. Some metrics would be nice
- According to claude we may have missing test coverage for the following endpoints
`GET /lighthouse/ui/health`
`GET /lighthouse/ui/graffiti`
`GET /lighthouse/beacon/health`
`POST /lighthouse/beacon/update`
`GET /lighthouse/logs`
`DELETE /lighthouse/keystores`

We should add test coverage for those sometime during this migration

common/axum_utils/Cargo.toml
macladson: Good call. Done in [b4bacf8](https://github.com/sigp/lighthouse/pull/8280/commits/b4bacf864850d05ac8e4510315a91a5646fb0c08)

common/axum_utils/src/cors.rs
macladson: Added some in https://github.com/sigp/lighthouse/pull/8280/commits/0ae7f279b2baa93d94faa7a64d8a197f674314ba

Cargo.toml
```diff
@@ -243,6 +245,7 @@ sysinfo = "0.26"
 system_health = { path = "common/system_health" }
 task_executor = { path = "common/task_executor" }
 tempfile = "3"
+thiserror = "2"
```
pawanjay176, line 240: Is it necessary to have this? Dependencies wise, this doesn't seem like a heavy library, but i'm a bit concerned about fragmented error handling across the codebase.

common/axum_utils/src/cors.rs
```diff
@@ -0,0 +1,397 @@
...
+    let make_error = |reason: &str| CorsError::InvalidOrigin {
+        origin: s.to_string(),
+        reason: reason.to_string(),
+    };
+
+    if !s.contains("://") {
+        return Err(make_error("missing scheme (http:// or https://)"));
+    }
+
+    let mut parts = s.splitn(2, "://");
```
pawanjay176: i feel we can use `split_once` here instead for more concise code?

common/axum_utils/src/cors.rs
```diff
@@ -0,0 +1,397 @@
...
+    NoValidOrigins,
+}
+
+/// A validated CORS origin
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Origin {
+    /// Allow any origin (*)
+    Any,
+    /// A specific origin URL
+    Exact(http::HeaderValue),
```
pawanjay176, line 25: I feel we can have this variant accept a Vec of header values instead and that would make the parsing in `into_layer` simpler.

common/axum_utils/src/cors.rs
```diff
@@ -0,0 +1,397 @@
...
+        }
+
+        Ok(origins)
+    }
+
+    /// Convert this config into a Tower-compatible CorsLayer.
+    pub fn into_layer(self) -> Result<CorsLayer, CorsError> {
+        let origins = self.parse_origins()?;
+
+        let cors = match origins.as_slice() {
```
pawanjay176: I feel this can be simplified if we parse a list into the enum directly
```rust
pub fn into_layer(self) -> Result<CorsLayer, CorsError> {
        let cors = match self.parse_origins()? {
            Origin::Any => CorsLayer::new().allow_origin(tower_http::cors::Any),
            Origin::Exact(origins) => {
                let header_values: Vec<_> = origins.into_iter().map(|o| o.0).collect();
                CorsLayer::new().allow_origin(AllowOrigin::list(header_values))
            }
        };

        Ok(cors)
    }
```

common/axum_utils/src/server/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+    {
+        let tokio_listener = tokio::net::TcpListener::bind(self.address)
+            .await
+            .map_err(ServerError::ServerFailed)?;
+
+        let actual_addr = tokio_listener
+            .local_addr()
+            .map_err(ServerError::ServerFailed)?;
+
+        let std_listener = tokio_listener
```
pawanjay176: can't we just do `std::net::TcpListener::bind` directly. why are we binding tokio then converting to a std listener?

common/axum_utils/src/server/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+#[derive(Debug, Clone, Copy)]
+pub struct ServerInfo {
+    pub address: SocketAddr,
+    pub protocol: Protocol,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Protocol {
```
pawanjay176, line 98: Is this used just for logging? curious why we want to add the protocol here

validator_client/http_api/src/lib.rs
```diff
@@ -1397,20 +1398,40 @@ pub fn serve<T: 'static + SlotClock + Clone, E: EthSpec>(
...
+        .with(cors_builder.build())
+        .boxed();
 
-    let (listening_socket, server) = warp::serve(routes).try_bind_with_graceful_shutdown(
-        SocketAddr::new(config.listen_addr, config.listen_port),
-        async {
-            shutdown.await;
-        },
-    )?;
+    let axum_router = Router::new().fallback_service(WarpService::new(routes));
```
pawanjay176, line 1406: Just to confirm, we aren't running any routes on axum currently and everything will go through the fallback here right?

pawanjay176 (changes requested):
Looking good, just a few questions and nits

pawanjay176:
I have tested the CORS requests and the metrics servers and it works fine. Have you tested the https variants? Maybe we can deploy this to some spesi node and double check that the metrics api works as expected?

validator_client/http_api/src/lib.rs
macladson: That's right, there are no actual axum routes, it's all going through the compatibility layer, so all requests and responses ultimately go through axum, with all the heavy lifting happening in between in the warp routes.

common/axum_utils/src/server/mod.rs
macladson: That must've been leftover from when I was testing with the BN (since the new `serve_with_shutdown` builder pattern abstracts away the rustls setup). And I needed a new way to log the presence of TLS.

I think it would still be useful to keep around once we migrate the BN but you're right, for the VC it doesn't matter since we don't run that over TLS anyway

common/axum_utils/src/server/mod.rs
macladson: It was a while since I wrote this, but from my memory I had some issues with certain tests failing but I'm unable to reproduce them (even without doing `set_nonblocking(true)`). 

`from_tcp` converts the listener to a tokio listener internally anyway so I've set it to a std listener and we'll see if any issues show up during testing.

common/axum_utils/src/cors.rs
macladson: A lot of this cors code is taken from the `warp_utils` version (with some extra validation):
https://github.com/sigp/lighthouse/blob/unstable/common/warp_utils/src/cors.rs#L51

My thinking was that once the migration is complete we can just delete the `warp_utils` crate entirely (and deal with a bit of duplication until then). I'm happy with refactoring it though but that's why the code is written the way it is

Cargo.toml
macladson: `thiserror` is used in many of our dependencies already so there's no issue with dependency overhead. I added it in the hope that we can start using it more broadly in Lighthouse. It automatically implements `std::error::Error` on the enum along with associated error `source` tracking, and also makes it trivial to add `from` impls between nested error types. Basically it stops you being "lazy" when defining your error enums.

I wouldn't use it in all situations, especially if its just a flat enum and you can just use the default impl of `std::error::Error` but when you have nested error types in your variants (like we have here with std::io::Error), it makes it trivial to implement a complete `std::error::Error` impl while keeping it concise.

common/axum_utils/src/cors.rs
pawanjay176: Makes sense. I'm fine either way, your call on to do it now or later

Cargo.toml
pawanjay176: Makes sense

pawanjay176 (approved):
LGTM. I think we need to run this on some of our nodes and just verify that the metrics work the same before merging cc @antondlr

common/axum_utils/src/cors.rs
macladson: I've cleaned up the cors code a bit, but wasn't able to find a design I liked with a Vec of header values here. It's nice having individual cors origins and all the list parsing code just ended up getting moved around rather than eliminated, so I've left it like this for now

Ari4ka (approved):

Ari4ka (approved):

beacon_node/http_metrics/src/lib.rs
```diff
@@ -81,25 +94,12 @@ impl Default for Config {
...
     shutdown: impl Future<Output = ()> + Send + Sync + 'static,
 ) -> Result<(SocketAddr, impl Future<Output = ()>), Error> {
     let config = &ctx.config;
 
-    // Configure CORS.
-    let cors_builder = {
-        let builder = warp::cors()
-            .allow_method("GET")
-            .allow_headers(vec!["Content-Type"]);
-
```
macladson, line 95: I've added these back by adding them to the `CorsLayer`. Same with the VC metrics server

beacon_node/http_metrics/src/lib.rs
```diff
@@ -108,44 +108,59 @@ pub fn serve<T: BeaconChainTypes>(
...
-            shutdown.await;
-        },
-    )?;
+    let cors_layer = build_cors_layer(
+        config.allow_origin.as_deref(),
+        config.listen_addr,
+        config.listen_port,
+    )?
+    .allow_methods([Method::GET])
+    .allow_headers([header::CONTENT_TYPE]);
```
macladson, line 117: Here

validator_client/http_metrics/src/lib.rs
```diff
@@ -90,73 +103,74 @@ impl Default for Config {
...
+        }));
+
+    let cors_layer = build_cors_layer(
+        config.allow_origin.as_deref(),
+        config.listen_addr,
+        config.listen_port,
+    )?
+    .allow_methods([Method::GET])
+    .allow_headers([header::CONTENT_TYPE]);
+
```
macladson, line 135: And here

Ari4ka (approved):

macladson (commented):
I am marking this as `do-not-merge` while I consider some alternative designs

macladson:
I am closing this in favour of the `warpdrive`-less #9001

Ari4ka (approved):

PR #8281, opened 2025-10-23
Gloas modify process_withdrawals

shane-moore:
## Changes per [spec](https://ethereum.github.io/consensus-specs/specs/gloas/beacon-chain/#modified-process_withdrawals):
- `process_withdrawals` modified
  -  `get_expected_withdrawals` modified 
    -  `is_builder_payment_withdrawable` added

## How this works
gloas introduces the new `BeaconState.builder_pending_withdrawals`, which are pending payments from the in-protocol builder to the proposer that have a withdrawable epoch scheduled

`process_withdrawals` runs during block processing, the same as today.  However, in the nested `get_expected_withdrawals`, we now sweep through the list of `builder_pending_withdrawals` and convert the ones that have hit their withdrawable epoch to `Withdrawals`

Then, back in `process_withdrawals`, we will sweep these withdrawals and decrement the balance of the builder on the CL.  The payment to the fee recipient in the EL is made later via inclusion of `Withdrawals` in the envelope's `ExecutionPayload`.

@ethDreamer


ethDreamer (approved):
LGTM!

PR #8283, opened 2025-10-24
Gloas modify process_attestations

shane-moore:
## Changes per [spec](https://ethereum.github.io/consensus-specs/specs/gloas/beacon-chain/#modified-process_attestation)
- `process_attestations` modified
  - `is_attestation_same_slot` added
  -  `get_attestation_participation_flag_indices` modified

## How This Works
The fast path for converting a slot's `builder_pending_payment` to a `builder_pending_withdrawal` triggers when the execution payload envelope for that slot is received and verified.  In that case, we immediately queue the withdrawal corresponding to the CL block that carried the bid.

However, the envelope may never be released. As a fallback, we want to convert the `builder_pending_payment` if enough validators attested on time to the block of that slot.

### What this PR Changes
Therefore, the `builder_pending_payment` has a `weight` field. During block processing, this PR modifies `process_attestation` to updates this weight only for same slot attestations, adding the attester’s balance once per validator. (Note that same slot means that the `attestation.data.beacon_block_root` corresponds to the `state.get_block_root(data.slot)`).

### In a future PR
Later, during epoch processing, we'll check if the weight is above quorum and if so, convert the `builder_pending_payment` to a `builder_pending_withdrawal`, which will have a `withdrawable_epoch`.

@ethDreamer


consensus/state_processing/src/per_block_processing/process_attestations.rs
```diff
@@ -0,0 +1,322 @@
+use super::*;
```
eserilev: could we keep this stuff in `process_operations`? Makes the PR a bit easier to review if the diff here is minimized. The changes here are a bit sensitive and I don't want to miss anything

consensus/state_processing/src/common/get_attestation_participation.rs
```diff
@@ -32,9 +35,35 @@ pub fn get_attestation_participation_flag_indices<E: EthSpec>(
...
-    let is_matching_head =
+
+    let is_matching_blockroot =
         is_matching_target && data.beacon_block_root == *state.get_block_root(data.slot)?;
 
+    let is_matching_head = if state.fork_name_unchecked().gloas_enabled() {
+        let is_matching_payload = if is_attestation_same_slot(state, data)? {
+            // For same-slot attestations, data.index must be 0
+            if data.index != 0 {
+                // TODO(EIP7732): consider if we want to use a different error type, since this is more of an overloaded data index scenario as opposed to the InvalidCommitteeIndex previous error. It's more like `AttestationInvalid::BadOverloadedDataIndex`
```
eserilev: I think it'd be nice to introduce a new error variant

testing/ef_tests/src/cases/operations.rs
```diff
@@ -101,7 +102,7 @@ impl<E: EthSpec> Operation<E> for Attestation<E> {
         let mut ctxt = ConsensusContext::new(state.slot());
         if state.fork_name_unchecked().altair_enabled() {
             initialize_progressive_balances_cache(state, spec)?;
-            altair_deneb::process_attestation(
+            altair_gloas::altair::process_attestation(
```
eserilev: we'll write a test case for the gloas case in a subequent PR I assume?

eserilev (commented):
I haven't done a review against the spec yet, just some nits to hopefully make this PR a little bit easier to review

consensus/state_processing/src/per_block_processing/process_attestations.rs
shane-moore: @eserilev, I feel you on that! moved the logic over to `process_operations`

consensus/state_processing/src/common/get_attestation_participation.rs
shane-moore: for sure, I updated to the following but am happy to name it something else if another preference
```
  if data.index != 0 {
                return Err(Error::BadOverloadedDataIndex(data.index));
   }
```

testing/ef_tests/src/cases/operations.rs
shane-moore: yep! added a `TODO` to write some test cases for this.  we'll be able to write tests for gloas features once we can properly produce gloas blocks and payloads.  https://github.com/sigp/lighthouse/blob/72fec1dd226c3e0b862a0753e76375f082cc2c04/consensus/state_processing/src/per_block_processing/process_operations.rs#L214-L216 
probably within the next couple weeks I would think. At that point, I'll go back and create tests for this and my other recent PR's, which also have TODO's for testing

consensus/state_processing/src/common/is_attestation_same_slot.rs
```diff
@@ -0,0 +1,18 @@
+use safe_arith::SafeArith;
+use types::{AttestationData, BeaconState, BeaconStateError, EthSpec};
+
+/// Checks if the attestation was for the block proposed at the attestation slot.
+pub fn is_attestation_same_slot<E: EthSpec>(
```
ethDreamer: this seems like something that might just be a function on the `BeaconState` itself rather than in its own file?

shane-moore: @ethDreamer, makes sense to me. updated in https://github.com/sigp/lighthouse/pull/8283/commits/4d199a54b36bd33ae1961390afa7709cfa75a50a

consensus/state_processing/src/common/get_attestation_participation.rs
```diff
@@ -32,9 +33,34 @@ pub fn get_attestation_participation_flag_indices<E: EthSpec>(
     let is_matching_source = data.source == justified_checkpoint;
     let is_matching_target = is_matching_source
         && data.target.root == *state.get_block_root_at_epoch(data.target.epoch)?;
-    let is_matching_head =
+
+    let is_matching_blockroot =
```
ethDreamer: Honestly i feel this code would be much easier to check if you just assign a variable to `head_root_matches` like the spec does. The spec may over-assign variables but I don't think that matters terribly for performance but it's just much easier to compare to the spec then.

shane-moore: ah yah @ethDreamer , looks like `head_root_matches` came out of a more recent [PR](https://github.com/ethereum/consensus-specs/pull/4692) than the `v1.6.0-beta.1` release, which I thought was devnet-0 target. i asked in [epbs channel ](https://discord.com/channels/595666850260713488/874767108809031740/1442144157140389951) if that's still the target, and they seem like they're open to `devnet-0` being based on at least the latest release, `v1.6.1`.

i've pushed [commit](https://github.com/sigp/lighthouse/pull/8283/commits/c6973fa503003afe2ec156721a6ae7b850ac0ff4) to handle all the relevant changes to this function per latest spec.

Additionally, this week, I'll attempt to go through all the spec changes from `v1.6.0-beta.1` til recent as to handle anything we'll need for self-building

jimmygchen:
This has been implemented and merged, thanks @shane-moore !

Issue #8208, opened 2025-10-16
Attestation Processing Optimization


AgeManning:
# Overview 

There could be some optimizations made in BLS verification in our attestation processing. As this consumes the majority of our processing time, optimizations here could have significant impact on Lighthouse's performance. 

# Current Design 

As attestations come in we queue them in the beacon processor. The beacon processor greedily tries to process anything in its queue, but we order objects based on priority. For a fast processor, the queue's can be relatively empty. For attestations, if we can process them individually without the queue's filling up, we do this. If we check the queue and more than one attestation has queued up, we batch process it. 
See here: https://github.com/sigp/lighthouse/blob/stable/beacon_node/beacon_processor/src/lib.rs#L1072-L1080

And the actual processing here: https://github.com/sigp/lighthouse/blob/stable/beacon_node/beacon_processor/src/lib.rs#L1547-L1560

# Potential Optimisations

A fast host, may be able to process individual attestations all the time, but batch processing attestations will significantly reduce the total amount of processing. We start to see gains in slower machines that are forced into batch processing and potentially we might stand to gain adding artificial delays into the attestation queue to force more batch BLS verification. 

One strategy might be to check the time throughout the slot. If we have time before the attestation deadline, then we should group attestations more aggressively. As the time runs out we revert back towards single attestation processing. 

A caveat here is that the more we group attestations, the more we have "burst" like traffic on the network. Although we can verify batches faster, we may not be able to forward them in batches as fast on the network, so the grouping may need some fine tuning.

michaelsproul:
@hopinheimer Is going to start working on this with mentorship from me and @eserilev

eserilev:
@hopinheimer and I got on a call today and discussed this in a bit more detail. We've potentially found a way to implement this optimization while minimizing code diff. We actually already do something similar when delaying data column reconstruction.

There was this original optimization for column reconstruction
https://github.com/sigp/lighthouse/pull/7588

We then added a max delay to column reconstruction here
https://github.com/sigp/lighthouse/pull/7976

The usage of the `DelayQueue` and the max delay functionality is pretty much exactly what we'd need for this attestation processing optimization.

At a high level, we'd introduce a new `ReprocessingQueueMessage` variant, lets call it `DelayAttestation`.  When we call `send_unaggregated_attestation`, we'll send a `DelayAttestation` message directly to the reprocessing queue

In the reprocessing queue we'll introduce two new fields
[...]

PR #8285, opened 2025-10-24
Attestation processing optimization with batching

hopinheimer:
#8208 
Current the implementation follows a naive approach of batching every 50ms, but I do have an idea for greedy batcher which I'll be implementing in following commits.


beacon_node/beacon_processor/src/lib.rs
```diff
@@ -1304,7 +1305,22 @@ impl<E: EthSpec> BeaconProcessor<E> {
                                 }
                             }
                             _ if can_spawn => self.spawn_worker(work, created_timestamp, idle_tx),
-                            Work::GossipAttestation { .. } => attestation_queue.push(work),
+                            Work::GossipAttestation { .. } => {
+                                if let Ok(queued_batch_attestation) =
```
eserilev: instead of sending`GossipAttestation` to the beacon processor and then  `BatchedAttestation` to the reprocessing queue, we should just send `BatchedAttestation` directly to the reprocessing queue

eserilev: This would involve updating `fn send_unaggregated_attestation` in the `network_beacon_processor` to send `BatchedAttestation` directly to the reprocessing queue

hopinheimer: by sending to reprocessing_queue directly you mean not via either `Work::GossipAttestation` or `Work::Reprocess`? seems like it would require a lot of wiring to get it working( imo ) since `BeaconProcessorSend` scope doesn't include the reprocessing queue which is spawned at `beacon_processor` level. With that being said, not sending the `BatchAttestation` directly to the reprocessing queue -- attestations may not be batched immediately which slightly digresses from the expected behavior. love to hear your thoughts on it

beacon_node/beacon_processor/src/scheduler/work_reprocessing_queue.rs
```diff
@@ -784,6 +824,23 @@ impl<S: SlotClock> ReprocessQueue<S> {
...
+                let att_id = QueuedAttestationId::Batched(self.next_attestation);
+
+                let delay_key = self
+                    .batched_attestation_queue
+                    .insert(att_id, batch_processing_delay);
+
+                self.queued_batch_attestations
+                    .insert(self.next_attestation, (queued_batch_attestation, delay_key));
+
+                self.next_attestation += 1;
```
eserilev: I think this logic is a bit off. Each time we receive an attestation we are resetting the delay queue to 50ms. The delay queue will only trigger the attestations if we haven't received an attestation within 50ms OR `MAXIMUM_BATCHED_ATTESTATIONS`

What we want is to queue up as many attestations as possible and process them on 50ms intervals. I'm actually not 100% sure on the best way to implement this

One way could be to use the slot clock and do something like 

`batch_processing_delay = MAXIMUM_BATCHED_ATTESTATIONS -  self.slot_clock.millis_from_current_slot_start() % MAXIMUM_BATCHED_ATTESTATIONS`

Lets say we receive an attestation 4550 milliseconds into the slot. This means we'd set `batch_processing_delay` to 50ms. If we receive another attestation at 4555 milliseconds, we'd update `batch_processing_delay` to 45ms. If we then receive another attestation at 4600 milliseconds into the slot, we'd update `batch_processing_delay` to 50ms BUT the previous batch should have already ran (since the delay on `batched_attestation_queue` should have already reached 0)

hopinheimer: I had missed a major chunk of the logic missing in the previous commit.  Although, I have implemented the batching logic. I'm not sure if this is the right way to do it but here's my take on it. `12s` slot duration is divided into `50ms` batches and each batch is filled with attestations bounded by `MAXIMUM_BATCHED_ATTESTATIONS` but here's the catch though, say the limit `MAXIMUM_BATCHED_ATTESTATIONS` is hit for a given batch -- the next batch wouldn't start immediately. The next batch would have a delay `time_until_current_batch_deadline + 50ms`.

eserilev: I dont think we want to use `MAXIMUM_BATCHED_ATTESTATIONS` at all? This should be strictly time based right?

eserilev: we should just be processing whatever attestations we have any time we hit the 50ms deadline

eserilev: your code right now doesn't seem to process any attestations until we have 1024 of them?

beacon_node/beacon_processor/src/scheduler/work_reprocessing_queue.rs
```diff
@@ -784,6 +832,68 @@ impl<S: SlotClock> ReprocessQueue<S> {
...
+            InboundEvent::Msg(BatchedAttestation(queued_batch_attestation)) => {
+                let batch_processing_delay = QUEUED_BATCH_ATTESTATION_DELAY;
+
+                let mut time_to_next_batch = 0;
+
+                if let Some(batched_queue) = self
+                    .queued_batch_attestations
+                    .get_mut(&self.current_attestation_batch)
+                {
+                    if batched_queue.0.len() >= MAXIMUM_BATCHED_ATTESTATIONS {
```
eserilev: so we never process a batch of                                                                                                                                             attestations until the queue has at least `MAXIMUM_BATCHED_ATTESTTATIONS`? This could make us miss processing attestations. The batching here should be strictly time based

hopinheimer: No it actually does, any attestations after `MAXIMUM_BATCHED_ATTESTATIONS` is batched into the next batch. while the processing of the batch is not handled here but in `InboundEvent::ReadyAttestationBatch` here 
https://github.com/sigp/lighthouse/blob/1033f077cea8c26c4d03b25a0b122f3f37439054/beacon_node/beacon_processor/src/scheduler/work_reprocessing_queue.rs#L835

after the delay queue is consumed and the event is generated here
https://github.com/sigp/lighthouse/blob/1033f077cea8c26c4d03b25a0b122f3f37439054/beacon_node/beacon_processor/src/scheduler/work_reprocessing_queue.rs#L368

coming to my rational behind the bounded batches is my hypothesis that there might be some upper bound where batching becomes asymptotic(1024 is just an arbitrary number for now). again this is my hypothesis may be incorrect

beacon_node/beacon_processor/src/scheduler/work_reprocessing_queue.rs
```diff
@@ -951,6 +1072,17 @@ impl<S: SlotClock> ReprocessQueue<S> {
             &[ATTESTATIONS],
             self.attestations_delay_queue.len() as i64,
         );
+
+        if let Some(current_batch) = self
+            .queued_batch_attestations
+            .get(&self.current_attestation_batch)
```
michaelsproul: We should loop over all batches currently in `queued_batch_attestations`: there could be multiple batches because of the size limit on each batch

beacon_node/beacon_processor/src/lib.rs
```diff
@@ -1320,6 +1338,7 @@ impl<E: EthSpec> BeaconProcessor<E> {
                                     "Unsupported inbound event"
                                 )
                             }
+                            Work::DelayedAttestationBatch { .. } => {}
```
michaelsproul: I think we need a queue for delayed attestation batches here. This code will currently just drop them (if `can_spawn` is false)

michaelsproul: And with the queue, we need a corresponding `.pop()` in the big chain of `if let` statements above

hopinheimer: resolved this in [519805d](https://github.com/sigp/lighthouse/pull/8285/commits/519805d212454e5717658580e5381683a8872bad)

beacon_node/beacon_processor/src/scheduler/work_reprocessing_queue.rs
hopinheimer: [fbccd8c](https://github.com/sigp/lighthouse/pull/8285/commits/fbccd8c9b4fd13a02dea7a2530bb9405dd24a4e5) solved it here

beacon_node/beacon_processor/src/scheduler/work_reprocessing_queue.rs
```diff
@@ -852,6 +941,38 @@ impl<S: SlotClock> ReprocessQueue<S> {
...
+                let QueuedAttestationId::Batched(batch_id) = queued_id else {
+                    crit!("Invalid attestation Id batched for attestation");
+                    return;
+                };
+
+                if let Some(batch_attestation) = self.queued_batch_attestations.remove(&batch_id) {
+                    let mut attestations = GossipAttestationBatch::new();
+                    let mut iter = batch_attestation.0.into_iter();
+
+                    if let Some(first) = iter.next() {
```
dapplion, line 952: Is it easier to read to do `if !is_empty() { let first = batch_attestation.first().expect("not empty") }`?

beacon_node/beacon_processor/src/scheduler/work_reprocessing_queue.rs
```diff
@@ -48,6 +50,9 @@ pub const ADDITIONAL_QUEUED_BLOCK_DELAY: Duration = Duration::from_millis(5);
 /// For how long to queue aggregated and unaggregated attestations for re-processing.
 pub const QUEUED_ATTESTATION_DELAY: Duration = Duration::from_secs(12);
 
+/// Batched attestation delay.
+pub const QUEUED_BATCH_ATTESTATION_DELAY: Duration = Duration::from_millis(50);
```
dapplion: This must be a flag, and if set to 0 disable this mechanism. In case we observe propagation degradation have a way to turn it off. And upper bound the flag value to a reasonable value.

hopinheimer: I'll add the config flag for the same with an upper bound in the following commit thanks!

PR #8286, opened 2025-10-24
Gloas add process_payload_attestation

shane-moore:
## Changes per [spec](https://ethereum.github.io/consensus-specs/specs/gloas/beacon-chain/#payload-attestations)
- Added:
  - `process_payload_attestation`
    - `get_indexed_payload_attestation`
    -  `is_valid_indexed_payload_attestation`
  - `get_ptc`
    - `compute_balance_weighted_selection`
    - `compute_balance_weighted_acceptance`
- Modified:
  -  `compute_proposer_indices`
  -  `get_next_sync_committee_indices`
  -  `process_operations` 

## How This Works
CL blocks will now contain payload attestations that are collected from PTC members from the previous slot.  The payload attestations indicate whether the payload envelope and blobs arrived by the PTC deadline.  During block processing, we will now check over the payload attestations to ensure they're valid.  Hence, this PR introduces `process_payload_attestation` to do just that. 

Since a payload attestation to validate includes a `Bitvector` of size `PTC_SIZE`, we need to map each `1` bit to the ptc committee member from the previous slot in order to validate their signatures.  This is accomplished via introducing or modifying the helpers listed above.

@ethDreamer


consensus/types/src/beacon_state.rs
```diff
@@ -2733,6 +2754,109 @@ impl<E: EthSpec> BeaconState<E> {
...
+    /// Return whether to accept the selection of the validator `index`, with probability
+    /// proportional to its `effective_balance`, and randomness given by `seed` and `iteration`.
+    fn compute_balance_weighted_acceptance(
+        &self,
+        index: usize,
+        seed: &[u8],
+        iteration: usize,
+        spec: &ChainSpec,
+    ) -> Result<bool, Error> {
+        let effective_balance = self.get_effective_balance(index)?;
```
ethDreamer, line 2894: Since this is called in a loop, I wonder if it would be worth while to get the effective balance from the epoch cache rather than fetching it each time? To do that, you'd have to ensure the epoch cache is initialized. You could look into this or you could leave a TODO here for later.

consensus/types/src/beacon_state.rs
```diff
@@ -1078,7 +1080,15 @@ impl<E: EthSpec> BeaconState<E> {
                 let mut preimage = seed.to_vec();
                 preimage.append(&mut int_to_bytes8(slot.as_u64()));
                 let seed = hash(&preimage);
-                self.compute_proposer_index(indices, &seed, spec)
+
+                if self.fork_name_unchecked().gloas_enabled() {
```
ethDreamer: may be worth storing a bool for `gloas_enabled` outside this `map` and then referencing it here

consensus/types/src/beacon_state.rs
shane-moore: Analysis:
The epoch cache is a singleton per each `BeaconState` instance.  It gets initialized anytime we have a state transition as to ensure we're always getting fresh validator balances. A few places it's initialized:
- start of every per block processing
- start of every per epoch processing
- `build_all_caches` when loading state from disk
- during block production's `produce_partial_beacon_block`

`compute_balance_weighted_acceptance` is nested in `compute_balance_weighted_selection`, which will be called post-gloas by:
- `compute_proposer_indices` (11 call sites)
- `get_next_sync_committee_indices` (4 call sites)
- `get_ptc` (1 so far)

since `cargo nextest` run isn't applicable to gloas yet,  we can't simply try to use epoch cache to get validator balances during `compute_balance_weighted_acceptance` and see what fails, so I think a TODO is the right approach for now.  I'm open to other ideas as well.  From investigating a few call sites of  `compute_proposer_indices`, it seems there's a few places we'd need to initialize the epoch cache for this optimization to work.

add todo in https://github.com/sigp/lighthouse/pull/8286/commits/fab18e7de4cdc97d3e856fb9b81f46c7b6dc6fe3. lmk if is all good!

consensus/types/src/beacon_state.rs
shane-moore: good call, updated in https://github.com/sigp/lighthouse/pull/8286/commits/fab18e7de4cdc97d3e856fb9b81f46c7b6dc6fe3

consensus/types/src/beacon_state.rs
ethDreamer: looks good to me!

ethDreamer:
you may need to rebase this against the latest `gloas-envelope-processing` branch @shane-moore

shane-moore:
@ethDreamer, rebased! should be all set

PR #8288, opened 2025-10-25
Downgrade gossip logs set to INFO level

dapplion:
## Issue Addressed

Testing non-finality checkpoint synced these logs showed up in my INFO grep and were noisy. INFO should only include the notifier and exceptional events. I don't see why the user would care about this info.

## Proposed Changes

Downgrade to debug

### DIFF
--- a/beacon_node/network/src/network_beacon_processor/gossip_methods.rs
+++ b/beacon_node/network/src/network_beacon_processor/gossip_methods.rs
@@ -962,7 +962,7 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
 
         match &result {
             Ok(AvailabilityProcessingStatus::Imported(block_root)) => {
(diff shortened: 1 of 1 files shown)


dapplion:
TODO: Should also downgrade the one in from data columns.

TODO: This log has been logged for a very long time. Understand if users need / expect it.

jimmygchen (changes requested):
The downgrade for blob/column change makes sense to me. Do we want to keep the block at info level? it's been INFO since 2020, and may be useful to see new blocks coming through?

michaelsproul:
Yeah IMO `New block received` is fine and does not need removing

dapplion:
Set `New block received` back to INFO

pawanjay176 (approved):
Agree

jimmygchen (approved):

PR #8290, opened 2025-10-25
Prevent dropping large binary data to logs

dapplion:
## Issue Addressed

Testing non finalized checkpoint sync noticed this log that dumps blob data in Debug format to the logs.

## Proposed Changes

Log only block root and commitment of each blob


jimmygchen (approved):
Good catch!

PR #8295, opened 2025-10-27
chore: Add Dockerfile.dev for local development

kevaundray:
## Issue Addressed

Currently whenever we build the `Dockerfile` file for local development using kurtosis , it recompiles everything on my laptop, even if no changes are made.  This takes about 120 seconds on my laptop (might be faster on others).

## Proposed Changes

Conservatively, I created a new Dockerfile.dev, so that the original file is kept the same, even though its pretty similar.

This uses `--mount-type=cache` saving the target and registry folder across builds.

**Usage** 

```sh
docker build -f Dockerfile.dev -t lighthouse:dev .
```

## Additional Info

- If you immediately redo the above command, it should have no effect, ie it shouldn't rebuild
- Modifying the cli crate gives about a 6x speedup from scratch, and modifying consensus gives about a 2x speedup


kevaundray:
Note: the local_testnet kurtosis script has not been changed and still uses Dockerfile

jimmygchen:
Thanks @kevaundray, I haven't had a chance to try it out, but this is great!

Dockerfile.dev
```diff
@@ -0,0 +1,25 @@
...
+ARG PROFILE=release
+ARG CARGO_USE_GIT_CLI=true
+ENV FEATURES=$FEATURES
+ENV PROFILE=$PROFILE
+ENV CARGO_NET_GIT_FETCH_WITH_CLI=$CARGO_USE_GIT_CLI
+ENV CARGO_INCREMENTAL=1
+
+COPY . .
+# Persist the registry and target file across builds. See: https://docs.docker.com/build/cache/optimize/#use-cache-mounts
+RUN --mount=type=cache,target=/usr/local/cargo/registry \
```
jimmygchen, line 15: Nice I've used this a few times and it seems to have improved. (My machine is slow so I haven't really got enough iterations for measurements)

I noticed the link you provided also suggests mounting this directory, should we include it too?
`/usr/local/cargo/git/db`

jimmygchen: ah so this caches the bare git repos and commits etc, so likely not much time savings in terms of computation. I think we can experiment using this version - I'll make a PR to update the startup script to use this doclerfile.

jimmygchen (approved):
This is great, thanks for the PR @kevaundray !

PR #8296, opened 2025-10-27
Revise logging in BlobsByRoot requests

chong-he:
## Issue Addressed

#7756 introduces a logging issue, where the relevant log:
https://github.com/sigp/lighthouse/blob/da5b2317205efc72b98cdc922289acefe3f76a13/beacon_node/network/src/network_beacon_processor/rpc_methods.rs#L380-L385

obtains the `block_root` from `slots_by_block_root.keys()`. If the `block_root` is empty (block not found in the data availability checker), then the log will not show any block root:

`DEBUG BlobsByRoot outgoing response processed       peer_id: 16Uiu2HAmCBxs1ZFfsbAfhSA98rUUL8Q1egLPb6WpGdKZxX6HqQYX, block_root: [], returned: 4`

This PR revises to return the `block_root` in the request as a vector of block root

### DIFF
--- a/beacon_node/network/src/network_beacon_processor/rpc_methods.rs
+++ b/beacon_node/network/src/network_beacon_processor/rpc_methods.rs
@@ -19,7 +19,7 @@ use lighthouse_tracing::{
 };
 use methods::LightClientUpdatesByRangeRequest;
 use slot_clock::SlotClock;
-use std::collections::{HashMap, hash_map::Entry};
+use std::collections::{HashMap, HashSet, hash_map::Entry};
 use std::sync::Arc;
 use tokio_stream::StreamExt;
 use tracing::{Span, debug, error, field, instrument, warn};
@@ -293,7 +293,7 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
         inbound_request_id: InboundRequestId,
         request: BlobsByRootRequest,
     ) -> Result<(), (RpcErrorResponse, &'static str)> {
-        let requested_roots: Vec<Hash256> =
+        let requested_roots: HashSet<Hash256> =
             request.blob_ids.iter().map(|id| id.block_root).collect();
 
         let mut send_blob_count = 0;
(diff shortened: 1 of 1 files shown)


beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -293,6 +293,9 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
         inbound_request_id: InboundRequestId,
         request: BlobsByRootRequest,
     ) -> Result<(), (RpcErrorResponse, &'static str)> {
+        let requested_roots: Vec<Hash256> =
+            request.blob_ids.iter().map(|id| id.block_root).collect();
```
chong-he, line 297: A note that the behaviour before #7756 is to always return the first block_root

https://github.com/sigp/lighthouse/blob/e5b4983d6baf85770fe4539a565d8a2dd462bc53/beacon_node/network/src/network_beacon_processor/rpc_methods.rs#L296

A (possibly) desirable outcome from this is, we can deduce to know that if we are rejecting BlobsByRoot request after Fulu slots, for example, if `returned = 0`, then we can look for the block_root(s) and check if they are all Fulu, and if they are, then we know we don't return anything

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -293,6 +293,9 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
         inbound_request_id: InboundRequestId,
         request: BlobsByRootRequest,
     ) -> Result<(), (RpcErrorResponse, &'static str)> {
+        let requested_roots: Vec<Hash256> =
```
jimmygchen: I think we can use `HashSet` here to deduplicate block roots?

chong-he: Good idea, change in [b5d7afd](https://github.com/sigp/lighthouse/pull/8296/commits/b5d7afd6e13d9910981d8b6911857346fa5660e7)

jimmygchen (approved):
LGTM!

PR #8300, opened 2025-10-27
Replace tracing::debug! with debug! same for other levels

dapplion:
## Issue Addressed

Just visual clean-up, making logging statements look uniform. There's no reason to use `tracing::debug` instead of `debug`. If we ever need to migrate our logging lib in the future it would make things easier too.


michaelsproul:
Good call, just needs to pass CI

jimmygchen (approved):
LGTM. I've pushed the trivial import fixes.

jimmygchen:
Oops, my previous unstable merge commit was borked (not sure what happened there), I just did another push. Will wait for CI to complete.

jxs (dismissed):
no strong opinion here, but then let's also create a [`clippy.toml` ](https://doc.rust-lang.org/clippy/configuration.html) with [`disallowed_macros`](https://doc.rust-lang.org/clippy/lint_configuration.html#disallowed-macros)  so that this enforced for future PR's

dapplion:
@jxs I tried to do it but didnt found a way to do it that does not disallow the usage of 

use tracing::debug

debug!

jimmygchen:
> no strong opinion here, but then let's also create a [`clippy.toml` ](https://doc.rust-lang.org/clippy/configuration.html) with [`disallowed_macros`](https://doc.rust-lang.org/clippy/lint_configuration.html#disallowed-macros) so that this enforced for future PR's

Unless I'm misunderstanding the config - I don't think we want to disallow this macro - this PR is just a visual clean up to use imports instead of fully qualified macro path.  

I think this is good to merge now - I'm all in for adding rules into clippy config file if there's a suitable config to make our lives easier.

jxs:
> > no strong opinion here, but then let's also create a [`clippy.toml` ](https://doc.rust-lang.org/clippy/configuration.html) with [`disallowed_macros`](https://doc.rust-lang.org/clippy/lint_configuration.html#disallowed-macros) so that this enforced for future PR's
> 
> Unless I'm misunderstanding the config - I don't think we want to disallow this macro - this PR is just a visual clean up to use imports instead of fully qualified macro path.
> 
> I think this is good to merge now - I'm all in for adding rules into clippy config file if there's a suitable config to make our lives easier.

the disallowing was of the full path of the macro so that there would be consistence in their calling, thanks for trying it Lion!

PR #8301, opened 2025-10-27
Match BlockError into penalty in block gossip

dapplion:
## Issue Addressed

In process_gossip_unverified block we match the verification error and we should always:
- log something
- propagate_validation_result to libp2p

However, for some variants we don't do that. This PR refactors the match to always:
- Log something (at least one debug log per error)
- propagate_validation_result to libp2p
- Explicitly specify a penalty or set the penalty to None (force you to think about it)

If others thumbs up this style I can extend to the other gossip messages. This style makes gossip handling more consistent and less chance of forgetting to propagate results or penalize peers.

For some variants like `Slashable` we penalize the peer in our internal scoring but we don't penalize in the gossip layer. This is odd, anyone knows why?

### DIFF
--- a/beacon_node/beacon_chain/src/block_verification.rs
+++ b/beacon_node/beacon_chain/src/block_verification.rs
@@ -91,7 +91,7 @@ use std::fs;
 use std::io::Write;
 use std::sync::Arc;
 use store::{Error as DBError, KeyValueStore};
-use strum::AsRefStr;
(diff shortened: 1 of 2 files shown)


pawanjay176 (commented):
I like this pattern. Easier to keep track

dapplion:
- Superseded by https://github.com/sigp/lighthouse/pull/9801

PR #8302, opened 2025-10-27
Gloas modify get_pending_balance_to_withdraw

shane-moore:
## Changes per [spec](https://ethereum.github.io/consensus-specs/specs/gloas/beacon-chain/#modified-get_pending_balance_to_withdraw)
- modified `get_pending_balance_to_withdraw`

## How this works
There are a couple places in the codebase where we want to verify that a validator has no pending balances to withdraw before allowing them to initiate an exit or consolidation.  Pre-gloas, we were calling `get_pending_balance_to_withdraw` to check whether their `BeaconState.pending_partial_withdrawals` was empty and if so, let them proceed.  Since gloas introduces `BeaconState.builder_pending_payments` and `BeaconState.builder_pending_withdrawals`, we now have to account for these pending validator balances as well before letting them proceed with an exit or consolidation.

@ethDreamer
@eserilev

### DIFF
--- a/consensus/types/src/beacon_state.rs
+++ b/consensus/types/src/beacon_state.rs
@@ -2456,8 +2456,13 @@ impl<E: EthSpec> BeaconState<E> {
             .map_err(Into::into)
     }
 
+    // TODO(EIP-7732): The consensus spec PR for this change mentions that some EF tests will be needed but haven't been created yet.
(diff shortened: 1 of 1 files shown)


ethDreamer (approved):
LGTM!

PR #8313, opened 2025-10-28
gloas VC block production

shane-moore:
## Changes
- added VC side block production flow for gloas to support self-building for the devnet-0 release, per [spec](https://ethereum.github.io/consensus-specs/specs/gloas/validator/#block-proposal)

## How This Works
In Gloas, the `publish_block` flow in the VC is modified to add the get, sign, and publish of a payload envelope, which occus when the `BeaconBlock.body.signed_execution_payload_bid.message.builder_index` is the same as the proposer's validator index.

Additionally, when requesting a block from the BN, we will no longer receive blobs or kzg commitments.  Instead, payload envelopes and blobs will be released by the proposer in the self-building scenario or otherwise by the in-protocol builder.  There is no longer a concept of requesting unblinded or blinded blocks since all CL blocks just contain bids. 

Client-side endpoints were added to support requesting gloas block and additionally envelopes.  All the new and modified epbs beacon api endpoints are defined in [PR](https://github.com/ethereum/beacon-APIs/pull/552).

## Todo's (left comments in the code as well)
- the server side (BN) response to the api's modified and added in this PR will need to be handled since we're only touching the VC in this PR. This is contingent on the upcoming `envelope_verification.rs` flow since we request an envelope to sign and gossip during block production.  Additionally, the `produce_block` call to the BN will no longer respond with blobs for example, so we'll need to modify this endpoint.  Additionally, the BN endpoint that receives the signed block to broadcast flow will need to be modified as well.
- add tests for `produce_block_v4` and other new endpoints on the server side

@ethDreamer
@eserilev

### DIFF
--- a/common/eth2/src/lib.rs
+++ b/common/eth2/src/lib.rs
@@ -9,6 +9,8 @@
 
 pub mod beacon_response;
 pub mod error;
+use ssz::{Decode, Encode};
+
 #[cfg(feature = "lighthouse")]
 pub mod lighthouse;
 #[cfg(feature = "lighthouse")]
@@ -27,7 +29,9 @@ pub use reqwest::{StatusCode, Url};
 pub use sensitive_url::SensitiveUrl;
 
 use self::mixin::{RequestAccept, ResponseOptional};
+use self::types::ExecutionPayloadEnvelopeMetadata;
 use self::types::*;
+use ::types::{ExecutionPayloadEnvelope, ForkName};
 use bls::SignatureBytes;
 use context_deserialize::ContextDeserialize;
 use educe::Educe;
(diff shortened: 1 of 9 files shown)


jimmygchen:
@eserilev do we need this one?

eserilev:
@jimmygchen no we do not, we already have state-full gloas block production merged to unstable. We need stateless block production eventually. Shane has a spec PR for this (I think in the beacon api repo) but I don't think it's been merged yet

shane-moore:
> @jimmygchen no we do not, we already have state-full gloas block production merged to unstable. We need stateless block production eventually. Shane has a spec PR for this (I think in the beacon api repo) but I don't think it's been merged yet

yep, open beacon api pr https://github.com/ethereum/beacon-APIs/pull/580. it's pending a few discussions that will become a priority later, such as what will happen with `builder_boost_factor` for the block proposal endpoint considering the future from my understanding will be per-builder boost values set in the node config that we'll want to pass along to the BN. but yah, stay tuned for that

eserilev:
closing as we already have this in unstable. thanks Shane!

PR #8314, opened 2025-10-28
Cell Dissemination (Partial messages)

dknopik:
## Issue Addressed

- https://github.com/ethereum/consensus-specs/pull/4558
- https://eips.ethereum.org/EIPS/eip-8136

## Important locations

[‎consensus/types/src/data/partial_data_column_sidecar.rs](https://github.com/sigp/lighthouse/pull/8314/changes#diff-11a35759483d97e9a74bb936984d0fc554e9531482a79bf74119323ddd3d4d2f):

`PartialDataColumnSidecar`: the actual message. 
`PartialDataColumnSidecarRef`: the actual message, but using references: This allows partial.rs (see below) to create lightweight structs containing only some parts of the canonical partial data column. It encodes to the same messages as `PartialDataColumnSidecar`.
`PartialDataColumnHeader` : an optional header in the message, containing the commitments, inclusion proof and signed block header. This is sent on every first partial sent between two nodes. If it is NOT sent, Lighthouse will add it to the message if it was not 
`PartialDataColumn`: the sidecar, plus some information that we derive: the column index from the subnet, and the block hash from the group id.
`PartialDataColumnPartsMetadata`: The metadata used for the partial messages - see spec. Right now, we always request missing cells, so the request bitmap is effectively always all ones.

[beacon_node/lighthouse_network/src/types/partial.rs‎](https://github.com/sigp/lighthouse/pull/8314/changes#diff-8cf8bb42bfb4bcfb881a84ca0703cf0d3dd5b86111f3e178ade12dfcf9494a3b): The main interaction with libp2p

[‎beacon_node/beacon_chain/src/partial_data_column_assembler.rs](https://github.com/sigp/lighthouse/pull/8314/changes#diff-7bc36a775d1e20fdac5068278faa31e243b1b16469f8d9823a4e5c5c392dc591): Where the partials end up. As they complete they are moved to the overflow_lru_cache. On every update partials are republished, using the `PartialMergeResult` data.

## Ugly bits

- There are a lot of expects around the sidecar and the metadata as we can assert bounds. I think some of them can be removed if we add some convenience functions to `ssz_types` and/or `ethereum_ssz`.
- `PartialDataColumnHeader` is also bleeding into code paths not *directly* related to partial messages:  as `getBlobsV3` and `getBlobsV2` (if V3 is not available) are triggered from partial messages if they arrive before the block (which is not unlikely), `fetch_and_process_blobs_v2_or_v3` takes a header as argument instead of a full block, which is sufficient and can be derived from a block, but awkward if called from contexts unrelated to partial messages. In conjunction with ePBS implementation efforts, this should probably renamed, a common container created, or put behind a trait or something like that.

## Left to do

- Gloas impl (to be done in a future PR)
- [x] Unit tests
- [x] Wait for https://github.com/sigp/lighthouse/pull/8731
- [x] Wait for https://github.com/libp2p/rust-libp2p/pull/6275


dknopik:
Added some unit test with claude, let me know if it is a bit too much :)

consensus/types/src/data/partial_data_column_sidecar.rs
```diff
@@ -0,0 +1,480 @@
...
+#[cfg_attr(
+    feature = "arbitrary",
+    derive(arbitrary::Arbitrary),
+    arbitrary(bound = "E: EthSpec")
+)]
+#[derive(Debug, Clone, Encode, Decode, TreeHash, TestRandom, Educe)]
+#[educe(PartialEq, Eq, Hash(bound = "E: EthSpec"))]
+pub struct PartialDataColumnSidecar<E: EthSpec> {
+    pub cells_present_bitmap: CellBitmap<E>,
+    pub column: VariableList<Cell<E>, <E as EthSpec>::MaxBlobCommitmentsPerBlock>,
```
jimmygchen: nit:

```suggestion
    pub column: VariableList<Cell<E>, E::MaxBlobCommitmentsPerBlock>,
```

consensus/types/src/data/partial_data_column_sidecar.rs
```diff
@@ -0,0 +1,480 @@
...
+            cells_present_bitmap: BitList::with_capacity(block_blobs).unwrap(),
+            column: VariableList::new(vec![Default::default(); present_blobs]).unwrap(),
+            kzg_proofs: VariableList::new(vec![KzgProof::empty(); present_blobs]).unwrap(),
+            header: VariableList::new(vec![]).unwrap(), // header is not being sent on cell push
+        }
+        .as_ssz_bytes()
+        .len()
+    }
+
+    pub fn max_size(max_blobs_per_block: usize) -> usize {
```
jimmygchen: This is dead code, can we remove this?

If we keep this, I think we could use some documentation here - if im understanding this correctly, this type is essentially used as an union type - hence the max_size, is the full cell-only sidecar, right?

consensus/types/src/data/partial_data_column_sidecar.rs
```diff
@@ -0,0 +1,480 @@
...
+    pub cells_present_bitmap: CellBitmap<E>,
+    // It is fine to use `Vec` here as we never decode directly into this type, and only create
+    // this from the `PartialDataColumnSidecar` type above. This avoids a few ugly `expect` calls.
+    pub column: Vec<&'a Cell<E>>,
+    pub kzg_proofs: Vec<&'a KzgProof>,
+    pub header: Vec<&'a PartialDataColumnHeader<E>>,
+}
+
+impl<E: EthSpec> PartialDataColumnSidecar<E> {
+    pub fn min_size() -> usize {
```
jimmygchen: I can't find usage of this - is this required

consensus/types/src/data/partial_data_column_sidecar.rs
```diff
@@ -0,0 +1,480 @@
...
+            cells_present_bitmap: BitList::with_capacity(1).unwrap(),
+            column: VariableList::new(vec![Cell::<E>::default()]).unwrap(),
+            kzg_proofs: VariableList::new(vec![KzgProof::empty()]).unwrap(),
+            header: VariableList::new(vec![]).unwrap(),
+        }
+        .as_ssz_bytes()
+        .len()
+    }
+
+    pub fn size(present_blobs: usize, block_blobs: usize) -> usize {
```
jimmygchen: Same as above

beacon_node/beacon_chain/src/partial_data_column_assembler.rs
```diff
@@ -0,0 +1,548 @@
...
+
+                merged
+            } else {
+                added_cells += partial_column.sidecar.column.len();
+                // First time seeing this column index for this block
+                partial
+            };
+
+            // Check if merged column is now complete by trying to convert into full
+            let column = if let Some(full_column) = merged.try_clone_full(&header) {
```
jimmygchen, line 133: Just a note that `try_clone_full` is a deep clone and it's called inside the `self.assemblies` write lock. If lock contention becomes an issue, this may be the place to optimise.

beacon_node/beacon_chain/src/partial_data_column_assembler.rs
```diff
@@ -0,0 +1,548 @@
...
+    }
+
+    /// Get all current partials for a block for publishing after fetching local blobs.
+    /// To unlock future publishing, mark blobs as fetched locally.
+    /// We do this within one write lock to avoid useless double publishes.
+    pub fn get_partials_and_mark_as_local_fetched(
+        &self,
+        block_root: Hash256,
+        header: &Arc<PartialDataColumnHeader<E>>,
+    ) -> Option<Vec<KzgVerifiedCustodyPartialDataColumn<E>>> {
```
jimmygchen: I don't see where `None` is returned, do we need `Option` here?

consensus/types/src/data/partial_data_column_sidecar.rs
```diff
@@ -0,0 +1,480 @@
...
+        let mut new_proofs = Vec::new();
+        let mut column_idx = 0;
+
+        for (blob_idx, present) in self.cells_present_bitmap.iter().enumerate() {
+            if present {
+                if filter(blob_idx) {
+                    // Keep this cell
+                    let cell = self.column.get(column_idx)?;
+                    new_column.push(cell);
+                    let proof = self.kzg_proofs.get(column_idx)?;
```
jimmygchen: The early returns `?` here are problematic I think - that's a data integrity error and this should probably return an error and may need to penalise the peer for sending this? Otherwise, we're treating corrupted data the same as nothing passing the filter, and we may end up propagating these invalid messages.

consensus/types/src/data/partial_data_column_sidecar.rs
```diff
@@ -0,0 +1,480 @@
...
+                    new_proofs.push(proof);
+                } else {
+                    // Mark as not present
+                    new_bitmap
+                        .set(blob_idx, false)
+                        .expect("Within bounds due to clone above");
+                }
+                column_idx = column_idx
+                    .checked_add(1)
+                    .expect("Will not have more cells than 2^64 - 1");
```
jimmygchen: Panic here isn't possible right now (I think, but i haven't finished), but ideally we return an error and log a crit instead - even if a bug is introduced in the future in parital messages, the beacon node can still gracefully fallback to receiving full sidecars.

jimmygchen: Same applies to all the other `.expect`s in this file

jimmygchen (commented):
@dknopik I've started reviewing but haven't been able to get through this yet and I need to do some Gloas review now, so will come back to this shortly. 

I'll submit what I have so far, but I don't have the full picture yet, so some of my draft comments may not be correct

consensus/types/src/data/partial_data_column_sidecar.rs
dknopik: Removed - I believe I originally added this because I oriented myself along the code for full columns, but this ended up being unneeded.

beacon_node/beacon_chain/src/partial_data_column_assembler.rs
dknopik: Good catch, this was previously necessary - will directly return the Vec

consensus/types/src/data/partial_data_column_sidecar.rs
dknopik: Understandable. But if this is really triggered, we do not want a simple `UnexpectedOverflow` enum variant, because we really need to understand where exactly the overflow occurred.

I am actually thinking about adding `map` and `filter` functions  (or a combined  `filter_map`) to the ssz types, as encapsulating the fact that filtering values from a `VariableList` will not overflow it if capacity stays the same in there would allow far cleaner code in here.

consensus/types/src/data/partial_data_column_sidecar.rs
dknopik: Yep, agree.

Cargo.toml
```diff
@@ -277,6 +279,7 @@ debug = true
 [patch.crates-io]
 quick-protobuf = { git = "https://github.com/sigp/quick-protobuf.git", rev = "87c4ccb9bb2af494de375f5f6c62850badd26304" }
 
-[patch."https://github.com/libp2p/rust-libp2p.git"]
-libp2p = { git = "https://github.com/sigp/rust-libp2p.git", rev = "defcaf1a78cf5b70a723b3fee0e0be051c1dbd88" }
-libp2p-mplex = { git = "https://github.com/sigp/rust-libp2p.git", rev = "defcaf1a78cf5b70a723b3fee0e0be051c1dbd88" }
+# This is intentionally unset as we use jxs fork in this PR. TODO(dknopik): restore/remove
```
jimmygchen, line 281: Just adding a comment here as a reminder to update before we merge

consensus/types/src/data/partial_data_column_sidecar.rs
jimmygchen: Yeah, this is a good idea.

I agree we'd want some error logging if this happens, although there's a chance that the beacon node can recover and fallback to receiving full data columns?

As a temporary solution (if we don't get the SSZ changes in time) i think we should remove the `.expect`s. We don't want to set a precedent for their usage, and we intend to enforce this with lint.

consensus/types/src/data/partial_data_column_sidecar.rs
```diff
@@ -0,0 +1,446 @@
...
+    /// Creates a reference to this sidecar containing only the blob indices for which the passed
+    /// closure returns `true` and is present in `self`. Will return `None` if there is no overlap.
+    pub fn filter<F>(&self, filter: F) -> Option<PartialDataColumnSidecarRef<'_, E>>
+    where
+        F: Fn(usize) -> bool,
+    {
+        let mut new_bitmap = self.cells_present_bitmap.clone();
+        let mut new_column = Vec::new();
+        let mut new_proofs = Vec::new();
+        let mut column_idx = 0;
```
jimmygchen: The name `column_idx` is a bit confusing here, worth adding a comment noting this is not the `ColumnIndex` of the data column sidecar, or perhaps renaming to something like `column_vec_idx` to differntiate.

consensus/types/src/data/partial_data_column_sidecar.rs
```diff
@@ -0,0 +1,446 @@
...
+
+        Some(PartialDataColumnSidecarRef {
+            cells_present_bitmap: new_bitmap,
+            column: new_column,
+            kzg_proofs: new_proofs,
+            header: self.header.iter().collect(),
+        })
+    }
+
+    pub fn merge(&self, other: &Self) -> Option<Self> {
```
jimmygchen: Could use some doc here, not clear why this returns an `Option`

consensus/types/src/data/partial_data_column_sidecar.rs
```diff
@@ -0,0 +1,446 @@
...
+        let new_bitmap = self.cells_present_bitmap.union(&other.cells_present_bitmap);
```
jimmygchen, line 112: is it worth adding a length check here on both sidecars? i think it makes more sense if this function returns a `Result` - `Ok` if successfully merged; `Error` otherwise, e.g. header mismatch, length mismatch etc. This would also replace the `.expect()` with `?`.

I could be wrong on what's already been validated - either way i think some docs describing the assumptions would help.

consensus/types/src/data/partial_data_column_sidecar.rs
```diff
@@ -0,0 +1,446 @@
...
+            .zip(other.cells_present_bitmap.iter())
+        {
+            match presence_bits {
+                (false, false) => {}
+                (true, other) => {
+                    new_column
+                        .push(self.column.get(self_cell_idx)?.clone())
+                        .expect("Has same capacity");
+                    new_proofs
+                        .push(*self.kzg_proofs.get(self_cell_idx)?)
```
jimmygchen: if we use the iterator's `.next()` for these lists, then we don't need to maintain the indices and would make it a bit more readable IMO.

consensus/types/src/data/partial_data_column_sidecar.rs
```diff
@@ -0,0 +1,446 @@
...
+
+        Some(Self {
+            cells_present_bitmap: new_bitmap,
+            column: new_column,
+            kzg_proofs: new_proofs,
+            header: if !self.header.is_empty() {
+                self.header.clone()
+            } else {
+                other.header.clone()
+            },
```
jimmygchen: I guess an extra header equality check here is not required here, as the header is already validated during gossip verification?

consensus/types/src/data/partial_data_column_sidecar.rs
```diff
@@ -0,0 +1,446 @@
...
+
+    fn try_from(block: &SignedBeaconBlock<E, P>) -> Result<Self, Self::Error> {
+        Ok(Self {
+            kzg_commitments: block.message().body().blob_kzg_commitments()?.clone(),
+            signed_block_header: block.signed_block_header(),
+            kzg_commitments_inclusion_proof: block
+                .message()
+                .body()
+                .kzg_commitments_merkle_proof()?
+                .clone(),
```
jimmygchen: Unnecessary clone i think?

consensus/types/src/data/partial_data_column_sidecar.rs
```diff
@@ -0,0 +1,446 @@
...
+impl<E: EthSpec> PartialDataColumn<E> {
+    pub fn try_clone_full(
+        &self,
+        header: &PartialDataColumnHeader<E>,
+    ) -> Option<DataColumnSidecar<E>> {
+        if !self.sidecar.is_complete() {
+            return None;
+        }
+
+        Some(DataColumnSidecar::Fulu(DataColumnSidecarFulu {
```
jimmygchen, line 255: Reminder: will need to add gloas support

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -198,25 +202,80 @@ impl From<BeaconStateError> for GossipDataColumnError {
...
+    InconsistentPresentCount {
+        bitmap_popcount: usize,
+        cells_len: usize,
+        proofs_len: usize,
+    },
+    /// The partial message has a bitmap length that is inconsistent with the number of commitments.
+    ///
+    /// ## Peer scoring
+    /// The column sidecar is invalid and the peer is faulty
+    InconsistentMaxCount {
```
jimmygchen: This name is slightly confusing, what about
`InconsistentCommitmentsLength`?

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -263,7 +322,9 @@ impl<T: BeaconChainTypes, O: ObservationStrategy> GossipVerifiedDataColumn<T, O>
 
         if chain
             .data_availability_checker
-            .is_data_column_cached(&column_sidecar.block_root(), &column_sidecar)
+            .determine_missing_cells_full(&column_sidecar.block_root(), column_sidecar.as_ref())
```
jimmygchen, line 334: Why do we need to check the `parital_assembler`, wouldn't completed partials always be in the `availability_cache`? or is there a race?

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -475,10 +696,37 @@ pub fn verify_kzg_for_data_column<E: EthSpec>(
...
+    data_column: PartialDataColumn<E>,
+    unverified_cells: Vec<usize>,
+    header: &GossipVerifiedPartialDataColumnHeader<E>,
+    kzg: &Kzg,
+) -> Result<KzgVerifiedPartialDataColumn<E>, GossipPartialDataColumnError> {
+    let Some(filtered_column) = data_column
+        .sidecar
+        .filter(|idx| unverified_cells.contains(&idx))
+    else {
+        return Err(GossipDataColumnError::PriorKnownUnpublished.into());
```
jimmygchen, line 795: I don't think this variant is applicable in a kzg verification function?

This error variant means: we have the blobs from the EL but we haven't publish them; if we get it via gossip, we should accept and forward them.

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -518,13 +766,15 @@ pub fn validate_data_column_sidecar_for_gossip_fulu<T: BeaconChainTypes, O: Obse
...
-    if chain
+    // are made available through the `engine_getBlobs` method and/or partial messages have arrived.
+    // If it exists in the cache, we know it has already passed the gossip checks, even though this
+    // particular instance hasn't been seen / published on the gossip network yet (passed the
+    // `verify_is_unknown_sidecar` check above). In this case, we should accept it for gossip
+    // propagation.
+    if let Some(cells) = chain
         .data_availability_checker
-        .is_data_column_cached(&data_column.block_root(), &data_column)
+        .determine_missing_cells_full(&data_column.block_root(), data_column.as_ref())
```
jimmygchen, line 857: same as the other one - would the original `is_data_column_cached` be insufficient here?

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -562,6 +816,123 @@ pub fn validate_data_column_sidecar_for_gossip_fulu<T: BeaconChainTypes, O: Obse
...
+            Ok(verified) => verified,
+            Err(err) => {
+                return PartialColumnVerificationResult::Err(err);
+            }
+        }
+    } else {
+        // There is no header, so we check if we have a cached one to use
+        let Some(header) = chain
+            .data_availability_checker
+            .partial_assembler()
```
jimmygchen: we could just `?` here instead of using `.and_then()` so it doesn't incorrectly return the `MissingHeader` error if `partial_assembler` is `None`.

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -562,6 +816,123 @@ pub fn validate_data_column_sidecar_for_gossip_fulu<T: BeaconChainTypes, O: Obse
...
+            .and_then(|a| a.get_header(&column.block_root))
+            .map(GossipVerifiedPartialDataColumnHeader::new_from_cached)
+        else {
+            return PartialColumnVerificationResult::Err(
+                GossipPartialDataColumnError::MissingHeader,
+            );
+        };
+
+        // If there was no header, there must be at least one cell.
```
jimmygchen, line 938: I think you meant "if there is a header"?

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -562,6 +816,123 @@ pub fn validate_data_column_sidecar_for_gossip_fulu<T: BeaconChainTypes, O: Obse
...
+            },
+            header,
+        };
+    }
+
+    let Some(missing_cells) = chain
+        .data_availability_checker
+        .determine_missing_cells_partial(&block_root, &column)
+    else {
+        // TODO(dknopik): This shuld be a different error
```
jimmygchen, line 980: Should be addressed before merging

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -985,4 +1369,240 @@ mod test {
...
+            .data_availability_checker
+            .partial_assembler()
+            .unwrap()
+            .init(block_root, header.clone());
+
+        (harness, block_root, header)
+    }
+
+    #[tokio::test]
+    async fn partial_empty_message_without_cells_returns_error() {
```
jimmygchen: All tests in `beacon_chain` run per-fork via the `FORK_NAME` env var in CI. I think we want to make tests either:
1. generic to forks via `test_spec()` so they run on every fork; OR:
2. just test one specific fork (e.g. `return` if `fork_name_from_env() != Some(Fulu)`)

jimmygchen: minor note: `BeaconChainHarness` is pretty expensive to create, and it piles up pretty quickly if we use it to check all cases. If we need the `chain` to test the gossip checks, we could have them all in a single test as a  pragmatic solution, or do unit tests instead - but may need some refactor to make it testable.

consensus/types/src/data/partial_data_column_sidecar.rs
```diff
@@ -0,0 +1,446 @@
...
+    derive(arbitrary::Arbitrary),
+    arbitrary(bound = "E: EthSpec")
+)]
+#[derive(Debug, Clone, Encode, Decode, TreeHash, TestRandom, Educe)]
+#[educe(PartialEq, Eq, Hash(bound = "E: EthSpec"))]
+pub struct PartialDataColumnSidecar<E: EthSpec> {
+    pub cells_present_bitmap: CellBitmap<E>,
+    pub column: VariableList<Cell<E>, E::MaxBlobCommitmentsPerBlock>,
+    pub kzg_proofs: VariableList<KzgProof, E::MaxBlobCommitmentsPerBlock>,
+    pub header: VariableList<PartialDataColumnHeader<E>, U1>,
```
jimmygchen: `ssz_types` 0.14.1 is out, I guess you wanted to use the new `ListEncodedOption` type here?
https://github.com/sigp/ssz_types/pull/83

consensus/types/src/data/partial_data_column_sidecar.rs
dknopik: Yeah, or `cell_idx`. Good point!

consensus/types/src/data/partial_data_column_sidecar.rs
dknopik: Yeah. I guess validation is not enforced though, so it might make sense to move this to the `KzgVerifiedPartialDataColumn` instead?

beacon_node/beacon_chain/src/data_column_verification.rs
dknopik: Clarified in call, comment is correct

consensus/types/src/data/partial_data_column_sidecar.rs
```diff
@@ -0,0 +1,446 @@
...
+
+        for presence_bits in self
+            .cells_present_bitmap
+            .iter()
+            .zip(other.cells_present_bitmap.iter())
+        {
+            match presence_bits {
+                (false, false) => {}
+                (true, other) => {
+                    new_column
```
jimmygchen: Can we add a comment here explaining that both sides are KZG-verified so overlapping cells are guaranteed identical?

consensus/types/src/data/partial_data_column_sidecar.rs
jimmygchen: yeah make sense to do the verification there

beacon_node/lighthouse_network/src/types/partial.rs
```diff
@@ -0,0 +1,498 @@
...
+
+                let need = !peer_metadata
+                    .available
+                    .is_subset(&self.partial_column.sidecar.cells_present_bitmap);
+                let want = peer_metadata.request.difference(&peer_metadata.available);
+
+                let send = self
+                    .partial_column
+                    .sidecar
+                    .filter(|idx| want.get(idx).expect("Bound checked above"))
```
jimmygchen, line 212: This isn't reachable right now but we should avoid panics

beacon_node/beacon_chain/src/partial_data_column_assembler.rs
```diff
@@ -0,0 +1,548 @@
...
+}
+
+impl<E: EthSpec> PartialDataColumnAssembler<E> {
+    pub fn new(capacity: NonZeroUsize) -> Self {
+        Self {
+            assemblies: RwLock::new(LruCache::new(capacity)),
+        }
+    }
+
+    /// Returns true unless the header was already contained or the passed argument failed to convert to a header
```
jimmygchen, line 52: > the passed argument failed to convert to a header

I don't see this? is this outdated comment

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -466,6 +633,60 @@ impl<E: EthSpec> KzgVerifiedCustodyDataColumn<E> {
...
+
+    pub fn as_data_column(&self) -> &PartialDataColumn<E> {
+        &self.data
+    }
+
+    pub fn index(&self) -> ColumnIndex {
+        self.data.index
+    }
+
+    pub fn merge(&self, other: &Self) -> Option<Self> {
```
jimmygchen: Please add doc here

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -193,6 +201,19 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
         })
     }
 
+    /// Send a message on `message_tx` that  `peer_id` should be propagated on
```
jimmygchen, line 206: This comment is incorrect i think?

beacon_node/lighthouse_network/src/types/partial.rs
```diff
@@ -0,0 +1,502 @@
...
+impl<E: EthSpec> OutgoingPartialColumn<E> {
+    pub fn new(
+        partial_column: Arc<PartialDataColumn<E>>,
+        header: &PartialDataColumnHeader<E>,
+        header_sent_set: HeaderSentSet,
+    ) -> Self {
+        // For now, always request all cells
+        let mut request = partial_column.sidecar.cells_present_bitmap.clone();
+        for idx in 0..request.len() {
+            request
```
jimmygchen: request seems unused after this?

beacon_node/lighthouse_network/src/types/partial.rs
```diff
@@ -0,0 +1,502 @@
...
+                .set(idx, true)
+                .expect("Bound asserted via `len` above");
+        }
+        let metadata = PartialDataColumnPartsMetadata::<E> {
+            available: partial_column.sidecar.cells_present_bitmap.clone(),
+            request: partial_column.sidecar.cells_present_bitmap.clone(),
```
jimmygchen: I don't understand this, why is available and request set to the same value?

beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -886,6 +920,66 @@ impl<E: EthSpec> Network<E> {
...
+                            kind = %topic.kind(),
+                            "Could not publish partial message"
+                        );
+                    }
+                }
+
+                // add to metrics
+                if let Some(v) = metrics::get_int_gauge(
+                    &metrics::FAILED_PUBLISHES_PER_MAIN_TOPIC,
+                    &[&format!("{:?}", topic.kind())],
```
jimmygchen, line 975: might be worth tracking the partial publishing failures separately, perhaps extra metric so we know how many of the failures are partial / full if it occrus.

beacon_node/lighthouse_network/src/types/pubsub.rs
```diff
@@ -464,6 +464,35 @@ impl<E: EthSpec> PubsubMessage<E> {
     }
 }
 
+/// Decodes incoming partial data column sidecar from gossipsub partial protocol.
+/// Note: Currenly, data columns are the only supported partial messages. In future this could
```
jimmygchen, line 468: nit: typo
```suggestion
/// Note: Currently, data columns are the only supported partial messages. In future this could
```

beacon_node/lighthouse_network/src/types/pubsub.rs
```diff
@@ -464,6 +464,35 @@ impl<E: EthSpec> PubsubMessage<E> {
...
+            if group.first() != Some(&0) {
+                return Err(format!("Unknown data column format: {:?}", group.first()));
+            }
+            let block_root = Hash256::from_ssz_bytes(&group[1..])
+                .map_err(|e| format!("Error decoding group: {:?}", e))?;
+            let sidecar = PartialDataColumnSidecar::from_ssz_bytes(data)
+                .map_err(|e| format!("Error decoding sidecar: {:?}", e))?;
+            let data_column = PartialDataColumn {
+                block_root,
+                // Partial messages are spec'd under the assumption that there is one column per subnet.
```
jimmygchen, line 486: This is true for the current mainnet config, however PeerDAS has subnet decoupled from num of columns to allow network parameter tweaks, which means we supprot having having different values for:
- `DATA_COLUMN_SIDECAR_SUBNET_COUNT`
- `NUMBER_OF_COLUMNS`

So this isn't compatible with the partials spec.

beacon_node/src/config.rs
```diff
@@ -110,6 +110,11 @@ pub fn get_config<E: EthSpec>(
 
     set_network_config(&mut client_config.network, cli_args, &data_dir_ref)?;
 
+    if parse_flag(cli_args, "enable-partial-columns") {
+        client_config.network.enable_partial_columns = true;
+        client_config.chain.enable_partial_columns = true;
```
jimmygchen: Please add a cli test here https://github.com/sigp/lighthouse/blob/d7c78a7f89bd9f3a3330ef60d8a4e58959f73227/lighthouse/tests/beacon_node.rs#L2860

beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -918,6 +1012,16 @@ impl<E: EthSpec> Network<E> {
         );
     }
 
+    /// Informs the gossipsub about the failure of a partial message validation.
+    pub fn report_partial_message_validation_failure(
+        &mut self,
+        propagation_source: PeerId,
+        topic: GossipTopic,
+    ) {
+        self.gossipsub_mut()
```
jimmygchen, line 1021: Be good to add the `GOSSIP_UNACCEPTED_MESSAGES_PER_CLIENT` metric here i think

beacon_node/lighthouse_network/src/service/mod.rs
```diff
@@ -1290,6 +1394,53 @@ impl<E: EthSpec> Network<E> {
...
+            } => {
+                let topic = GossipTopic::decode(topic_hash.as_str())
+                    .inspect_err(|error| {
+                        debug!(
+                            topic = ?topic_hash,
+                            error,
+                            "Could not decode gossipsub partial message topic"
+                        );
+                    })
+                    .ok()?;
```
jimmygchen, line 1412: we shouldn't swallow decoding error here - should probably penalise peer

consensus/types/src/data/partial_data_column_sidecar.rs
dknopik: Done in https://github.com/sigp/lighthouse/pull/8314/commits/ca4b5f0003e3d4cff44eeaa46725caf55b9b537c

beacon_node/beacon_chain/src/data_column_verification.rs
dknopik: both done in https://github.com/sigp/lighthouse/pull/8314/commits/edfa71c6d308fc6c008ed0ed95e147866c6af5c4

beacon_node/lighthouse_network/src/types/pubsub.rs
```diff
@@ -464,6 +464,35 @@ impl<E: EthSpec> PubsubMessage<E> {
...
 
+/// Decodes incoming partial data column sidecar from gossipsub partial protocol.
+/// Note: Currenly, data columns are the only supported partial messages. In future this could
+/// return an enum.
+pub fn decode_partial<E: EthSpec>(
+    topic: &GossipTopic,
+    group: &[u8],
+    data: &[u8],
+) -> Result<PartialDataColumn<E>, String> {
+    match topic.kind() {
```
jimmygchen, line 475: not a concern for this PR but i'd imagine we need to decode with the `topic.fork_digest` to support Gloas

beacon_node/beacon_chain/src/data_column_verification.rs
dknopik: Done something similar in https://github.com/sigp/lighthouse/pull/8314/commits/40125cdf58826d002670b2edfc3202e9110b90e7 - I introduced an error for missing partial assembler.

beacon_node/lighthouse_network/Cargo.toml
```diff
@@ -21,11 +21,12 @@ ethereum_ssz_derive = { workspace = true }
 fixed_bytes = { workspace = true }
 fnv = { workspace = true }
 futures = { workspace = true }
+gossipsub = { workspace = true }
 hex = { workspace = true }
 if-addrs = "0.14"
 itertools = { workspace = true }
 libp2p = { workspace = true }
-libp2p-mplex = { git = "https://github.com/libp2p/rust-libp2p.git" }
+libp2p-mplex = { git = "https://github.com/jxs/rust-libp2p.git", branch = "gossipsub-partial-messages" }
```
jimmygchen, line 29: just a reminder comment to use canonical fork once merged

consensus/types/src/data/partial_data_column_sidecar.rs
dknopik: Done in https://github.com/sigp/lighthouse/pull/8314/commits/a0ffc72e20c2d1d2f91664864f36b80fd04d9be9

consensus/types/src/data/partial_data_column_sidecar.rs
dknopik: Done in https://github.com/sigp/lighthouse/pull/8314/commits/a0ffc72e20c2d1d2f91664864f36b80fd04d9be9

consensus/types/src/data/partial_data_column_sidecar.rs
dknopik: Done in https://github.com/sigp/lighthouse/pull/8314/commits/a0ffc72e20c2d1d2f91664864f36b80fd04d9be9

beacon_node/beacon_chain/src/data_column_verification.rs
dknopik: Done in https://github.com/sigp/lighthouse/pull/8314/commits/00e9c83ee94a83789c7fcef551e55deb93ef956a

consensus/types/src/data/partial_data_column_sidecar.rs
dknopik: Partially done in https://github.com/sigp/lighthouse/commit/a0ffc72e20c2d1d2f91664864f36b80fd04d9be9, 
does not return an Option anymore. Will also add doc comment

consensus/types/src/data/partial_data_column_sidecar.rs
dknopik: Done in https://github.com/sigp/lighthouse/pull/8314/commits/a0ffc72e20c2d1d2f91664864f36b80fd04d9be9

consensus/types/src/data/partial_data_column_sidecar.rs
dknopik: Done in https://github.com/sigp/lighthouse/pull/8314/commits/a0ffc72e20c2d1d2f91664864f36b80fd04d9be9

jimmygchen (changes requested):

consensus/types/src/data/partial_data_column_sidecar.rs
```diff
@@ -0,0 +1,486 @@
...
+                .body()
+                .kzg_commitments_merkle_proof()?,
+        })
+    }
+}
+
+#[derive(Debug, Clone, Encode, Decode, PartialEq, Eq)]
+pub struct PartialDataColumnPartsMetadata<E: EthSpec> {
+    pub available: CellBitmap<E>,
+    pub request: CellBitmap<E>,
```
jimmygchen: Doesnt' really impact SSZ encoding, but would be good to match the `requests` name in the spec .

beacon_node/beacon_chain/src/metrics.rs
```diff
@@ -1675,6 +1675,56 @@ pub static DATA_COLUMN_SIDECAR_GOSSIP_VERIFICATION_TIMES: LazyLock<Result<Histog
...
+        try_create_int_counter(
+            "beacon_partial_data_column_sidecar_header_processing_dupes_total",
+            "Number of partial data column sidecars verified for gossip (excluding dupes)",
+        )
+    });
+pub static PARTIAL_DATA_COLUMN_SIDECAR_HEADER_PROCESSING_SUCCESSES: LazyLock<Result<IntCounter>> =
+    LazyLock::new(|| {
+        try_create_int_counter(
+            "beacon_partial_data_column_sidecar_header_processing_successes_total",
+            "Number of partial data column sidecars verified for gossip (excluding dupes)",
```
jimmygchen, line 1696: help description is same as above metric and needs update here

beacon_node/beacon_chain/src/metrics.rs
```diff
@@ -1675,6 +1675,56 @@ pub static DATA_COLUMN_SIDECAR_GOSSIP_VERIFICATION_TIMES: LazyLock<Result<Histog
...
+        )
+    });
+pub static PARTIAL_DATA_COLUMN_SIDECAR_HEADER_GOSSIP_VERIFICATION_TIMES: LazyLock<
+    Result<Histogram>,
+> = LazyLock::new(|| {
+    try_create_histogram(
+        "beacon_partial_data_column_sidecar_header_gossip_verification_seconds",
+        "Full runtime of partial data column sidecars gossip verification",
```
jimmygchen, line 1704: same description as `PARTIAL_DATA_COLUMN_SIDECAR_GOSSIP_VERIFICATION_TIMES`, perhaps add `header` ?

consensus/types/src/data/partial_data_column_sidecar.rs
```diff
@@ -0,0 +1,486 @@
...
+                .map_err(|_| PartialDataColumnSidecarError::UnexpectedBounds)?,
+            header: if self.header.is_some() {
+                self.header.clone()
+            } else {
+                other.header.clone()
+            },
+        })
+    }
+
+    fn len(&self) -> Result<usize, PartialDataColumnSidecarError> {
```
jimmygchen, line 164: perhaps rename this to `verify_len`?

beacon_node/network/src/sync/manager.rs
```diff
@@ -141,6 +141,9 @@ pub enum SyncMessage<E: EthSpec> {
     /// A data column with an unknown parent has been received.
     UnknownParentDataColumn(PeerId, Arc<DataColumnSidecar<E>>),
 
+    /// A partial data column with an unknown parent has been received.
+    UnknownParentPartialDataColumn(PeerId, Hash256, Hash256, Slot),
```
jimmygchen, line 145: be worth documenting the two roots fields here for clarity

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3124,6 +3187,95 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+            &metrics::BEACON_PARTIAL_MESSAGE_CELLS_RECEIVED_TOTAL,
+            &[index_str.as_str()],
+            partial.sidecar.column.len() as u64,
+        );
+
+        // Check if we have custody of this column
+        let sampling_columns = self
+            .data_availability_checker
+            .custody_context()
+            .sampling_columns_for_epoch(slot.epoch(T::EthSpec::slots_per_epoch()), &self.spec);
```
jimmygchen: nit: could just use `self.sampling_columns_for_epoch`

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -338,6 +431,12 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
         &self,
         executed_block: AvailabilityPendingExecutedBlock<T::EthSpec>,
     ) -> Result<Availability<T::EthSpec>, AvailabilityCheckError> {
+        let block = executed_block.as_block();
+        if let Some(assembler) = &self.partial_assembler
+            && let Ok(header) = block.try_into()
+        {
+            assembler.init(block.canonical_root(), Arc::new(header));
```
jimmygchen: Can use block root here without recomputing `executed_block.import_data.block_root`

beacon_node/network/src/sync/manager.rs
```diff
@@ -903,6 +906,21 @@ impl<T: BeaconChainTypes> SyncManager<T> {
...
                 }
             }
+            SyncMessage::UnknownParentPartialDataColumn(peer_id, block_root, parent_root, slot) => {
+                debug!(%block_root, %parent_root, "Received unknown parent partial column message");
+                self.handle_unknown_parent(
+                    peer_id,
+                    block_root,
+                    parent_root,
+                    slot,
+                    BlockComponent::PartialDataColumn(DownloadResult {
```
jimmygchen, line 916: Do we need a new variant for partials? we handle them exactly the same way, and could just use `DataColumns`, but it does carry extra information if we do log this enum somewhere.

jimmygchen: it probably doesn't matter - historically we had the data column in the enum variant and we store them, but we no longer do that.

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -218,9 +218,15 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                 // Block is valid, we can now attempt fetching blobs from EL using version hashes
                 // derived from kzg commitments from the block, without having to wait for all blobs
                 // to be sent from the peers if we already have them.
-                let publish_blobs = false;
-                self.fetch_engine_blobs_and_publish(signed_beacon_block, block_root, publish_blobs)
+                if let Ok(header) = signed_beacon_block.as_ref().try_into() {
```
jimmygchen, line 221: just a note that the error path is not reachable today because `try_into` only returns error if the block is pre-deneb. I'd usually avoid ignoring error but in this case its probably fine.

beacon_node/network/src/network_beacon_processor/mod.rs
```diff
@@ -251,6 +251,32 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+        self: &Arc<Self>,
+        peer_id: PeerId,
+        column_sidecar: Box<PartialDataColumn<T::EthSpec>>,
+        seen_timestamp: Duration,
+        topic: GossipTopic,
+    ) -> Result<(), Error<T::EthSpec>> {
+        let processor = self.clone();
+        let process_fn = async move {
+            processor
+                .process_gossip_partial_data_column_sidecar(
```
jimmygchen, line 265: We discussed about this yesterday - but worth noting here that we are not passing subnet id to processing because the partial message design assumes column index == subnet id. This is true for mainnet config today, but isn't true for the peerdas design as it supports subnet coupling and these network parameters can be tweaked via configuration. 

we should add the number of subnet & number of column validation check -  probably to where we parse the cli flag?

beacon_node/lighthouse_network/src/types/partial.rs
dknopik: oof.

good catch. 🤦

jimmygchen:
I'm pretty close to finishing the review. 

I got claude to run a round of preliminary tests and looked at metrics, logs and traces on a local devnet - here's the report: https://hackmd.io/@jimmygchen/BJqjwMCnbx

TL;DR this seems to be working fine with unstable on a local network.

beacon_node/network/src/sync/manager.rs
dknopik: converted into struct variant instead

beacon_node/lighthouse_network/src/types/pubsub.rs
dknopik: I added a check on startup that these values are identical. Partial messages are not enabled otherwise.

beacon_node/http_api/src/publish_blocks.rs
```diff
@@ -514,15 +516,51 @@ fn publish_column_sidecars<T: BeaconChainTypes>(
...
-        .map(|data_col| {
-            let subnet = DataColumnSubnetId::from_column_index(*data_col.index(), &chain.spec);
-            PubsubMessage::DataColumnSidecar(Box::new((subnet, data_col)))
-        })
-        .collect::<Vec<_>>();
-    crate::utils::publish_pubsub_messages(sender_clone, pubsub_messages)
-        .map_err(|_| BlockError::BeaconChainError(Box::new(BeaconChainError::UnableToPublish)))
+    let mut full_messages = Vec::new();
+    let mut partial_columns = Vec::new();
+    let mut partial_header = None;
```
jimmygchen, line 521: We should get the partial header creation and publishing here if it's not enabled.

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -549,30 +923,45 @@ pub fn validate_data_column_sidecar_for_gossip_fulu<T: BeaconChainTypes, O: Obse
...
+    let Some(cells_to_kzg_verify) = chain
         .data_availability_checker
-        .is_data_column_cached(&data_column.block_root(), &data_column)
-    {
+        .missing_cells_for_column_sidecar(&data_column)
+        .map_err(|err| match err {
+            MissingCellsError::MismatchesCachedColumn => {
+                GossipDataColumnError::MismatchesCachedColumn
+            }
+            MissingCellsError::UnexpectedError(_) => todo!("handle unexpected error"),
```
jimmygchen, line 938: This is in production code path although unreachable unless partial flag is enabled. We should just return an error here.

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -595,6 +984,137 @@ pub fn validate_data_column_sidecar_for_gossip_fulu<T: BeaconChainTypes, O: Obse
...
+                header,
+            };
+        }
+        Err(MissingCellsError::MismatchesCachedColumn) => {
+            return PartialColumnVerificationResult::ErrWithValidHeader {
+                err: GossipDataColumnError::MismatchesCachedColumn.into(),
+                header,
+            };
+        }
+        Err(MissingCellsError::UnexpectedError(e)) => todo!("handle unexpected error {:?}", e),
```
jimmygchen, line 1079: same as above

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -262,22 +340,29 @@ impl<T: BeaconChainTypes, O: ObservationStrategy> GossipVerifiedDataColumn<T, O>
...
+                if O::observe() {
+                    observe_gossip_data_column(&column_sidecar, chain)?;
+                }
+                Err(GossipDataColumnError::PriorKnownUnpublished)
+            }
+            Err(MissingCellsError::MismatchesCachedColumn) => {
+                Err(GossipDataColumnError::MismatchesCachedColumn)
+            }
+            Err(MissingCellsError::UnexpectedError(_)) => {
+                todo!("handle unexpected error")
```
jimmygchen, line 363: return error

consensus/types/src/data/partial_data_column_sidecar.rs
```diff
@@ -0,0 +1,385 @@
...
+}
+
+impl<E: EthSpec> PartialDataColumnSidecar<E> {
+    pub fn is_complete(&self) -> bool {
+        self.cells_present_bitmap.num_set_bits() == self.cells_present_bitmap.len()
+    }
+
+    pub fn get(&self, idx: usize) -> Option<(&Cell<E>, &KzgProof)> {
+        self.column
+            .get(idx)
```
jimmygchen: `idx` is blob index, but `self.column` may not have all cells so the index may not match?

jimmygchen (commented):
Hi @dknopik this is looking pretty good now, I reckon we merge this and address any remaining issues separately - I'm going to do a quick final pass and fix CI issues and merge.

I think this comment is needs a fix if this is true:
> idx is blob index, but self.column may not have all cells so the index may not match?

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -262,22 +340,29 @@ impl<T: BeaconChainTypes, O: ObservationStrategy> GossipVerifiedDataColumn<T, O>
         // In this case, we should accept it for gossip propagation.
         verify_is_unknown_sidecar(chain, &column_sidecar)?;
 
-        if chain
+        match chain
             .data_availability_checker
-            .is_data_column_cached(&column_sidecar.block_root(), &column_sidecar)
+            .missing_cells_for_column_sidecar(&column_sidecar)
```
jimmygchen, line 345: I might be wrong but i think we should just treat the data column as not cached and proceed with publishing the column anyway.

I think this is probably worth iterating further - but for now, before partials is widely deployed on the network, we should prioritise having full columns published, even when there's an error in partials.

jimmygchen: I couldn't do the error handling cleanly without changing the function. Let's discuss this, but it shouldn't block the PR as it's partials only path.

jimmygchen (approved):
@dknopik thanks for your patience with this! 

~~I'm happy to merge this once this is updated to use the canonical fork:~~ (**EDIT:** The below has been addressed, I have delete the outdated TODO comment)
  - [`Cargo.toml:279`](https://github.com/sigp/lighthouse/pull/8314#discussion_r3019599356) - Reminder to update dependency before merge

Happy to merge this as is, and the remaining comments below to be addressed in a follow up PR:
* `todo!()`/`expect() in partial-enabled code paths:
  * [`data_column_verification.rs:938`](https://github.com/sigp/lighthouse/pull/8314#discussion_r3125304697)   
  * [`data_column_verification.rs:1079`](https://github.com/sigp/lighthouse/pull/8314#discussion_r3125305939)
  * [`data_column_verification.rs:363`](https://github.com/sigp/lighthouse/pull/8314#discussion_r3125307522) 
  * [`partial.rs:213`](https://github.com/sigp/lighthouse/pull/8314#discussion_r3071336258)
* `missing_cells_for_column_sidecar` handling
  * [`data_column_verification.rs:345`](https://github.com/sigp/lighthouse/pull/8314#discussion_r3131832010)

consensus/types/src/data/partial_data_column_sidecar.rs
jimmygchen: https://github.com/sigp/lighthouse/pull/8314/commits/0e075c542c46eb34c7bd018aa98454c4ef034941

PR #8315, opened 2025-10-28
fix: clarify `bb` vs `bl` variable names in BeaconProcessorQueue

kevaundray:
## Issue Addressed

since block and blob both start with `bl`, it was not clear how to differentiate between `blbroots_queue` and `bbroots_queue` 

After renaming, there also seems to be a discrepancy


beacon_node/beacon_processor/src/lib.rs
```diff
@@ -1435,10 +1435,10 @@ impl<E: EthSpec> BeaconProcessor<E> {
...
                         WorkType::ChainSegmentBackfill => backfill_chain_segment.len(),
                         WorkType::Status => status_queue.len(),
-                        WorkType::BlocksByRangeRequest => blbrange_queue.len(),
-                        WorkType::BlocksByRootsRequest => blbroots_queue.len(),
-                        WorkType::BlobsByRangeRequest => bbrange_queue.len(),
-                        WorkType::BlobsByRootsRequest => bbroots_queue.len(),
+                        WorkType::BlocksByRangeRequest => blob_brange_queue.len(),
+                        WorkType::BlocksByRootsRequest => blob_broots_queue.len(),
+                        WorkType::BlobsByRangeRequest => block_brange_queue.len(),
+                        WorkType::BlobsByRootsRequest => block_broots_queue.len(),
```
kevaundray: Discrepancy is here -- I believe they should be flipped

beacon_node/beacon_processor/src/lib.rs
```diff
@@ -123,10 +123,10 @@ pub struct BeaconProcessorQueueLengths {
...
     delayed_block_queue: usize,
     status_queue: usize,
-    bbrange_queue: usize,
-    bbroots_queue: usize,
-    blbroots_queue: usize,
-    blbrange_queue: usize,
+    block_brange_queue: usize,
+    block_broots_queue: usize,
+    blob_broots_queue: usize,
+    blob_brange_queue: usize,
```
kevaundray, line 129: This change was roughly:

```
bbrange => block_brange
blbrange => blob_brange
```

and the same was done for broots

beacon_node/beacon_processor/src/lib.rs
michaelsproul: Yeah! Good catch!

michaelsproul: @kevaundray do you mind pushing a new commit to fix this, then we can merge?

kevaundray: yep I can do that now!

michaelsproul (approved):

PR #8316, opened 2025-10-28
feat!: Optional execution proofs

kevaundray:
## Issue Addressed

This is a parallel PR to https://github.com/sigp/lighthouse/pull/7755 that includes more of the logic needed to for example sync.

## Proposed Changes

## Additional Info


beacon_node/beacon_chain/src/execution_proof_verification.rs
```diff
@@ -0,0 +1,625 @@
...
+    // 4. Check if the proof is already in the DA checker cache
+    // If it exists in the cache, we know it has already passed validation.
+    if chain
+        .data_availability_checker
+        .is_execution_proof_cached(&block_root, &execution_proof)
+    {
+        if O::observe() {
+            observe_gossip_execution_proof(&execution_proof, chain)?;
+        }
+        return Err(GossipExecutionProofError::PriorKnownUnpublished);
```
dapplion, line 188: Gossip has a de-duplication mechanism based on the message content. Is there a single possible `ExecutionProof` per subnet and slot that is valid? If yes, the observe cache feels unnecessary.

beacon_node/network/src/sync/manager.rs
```diff
@@ -132,6 +134,14 @@ pub enum SyncMessage<E: EthSpec> {
         seen_timestamp: Duration,
     },
 
+    /// An execution proof has been received from the RPC
+    RpcExecutionProof {
+        sync_request_id: SyncRequestId,
+        peer_id: PeerId,
+        execution_proof: Option<Arc<ExecutionProof>>,
+        seen_timestamp: Duration,
+    },
```
dapplion, line 143: I would not touch sync for now, get the core logic in first

beacon_node/beacon_chain/src/execution_proof_verification.rs
kevaundray: Yep -- there is only one subnet now but there are multiple proofs for the slot. So there is a single possible ExecutionProof for a (proof_id, slot_number, beacon_block_root) combination

kevaundray:
closing, spoke with team :)

PR #8317, opened 2025-10-29
Fix failing CI for `compile-with-beta-compiler`

hopinheimer:


consensus/types/src/light_client_bootstrap.rs
```diff
@@ -188,8 +188,12 @@ impl<E: EthSpec> LightClientBootstrap<E> {
         block: &SignedBlindedBeaconBlock<E>,
         chain_spec: &ChainSpec,
     ) -> Result<Self, Error> {
-        let mut header = beacon_state.latest_block_header().clone();
-        header.state_root = beacon_state.update_tree_hash_cache()?;
+        let header = beacon_state.update_tree_hash_cache()?;
+        beacon_state
+            .latest_block_header()
+            .clone()
+            .update_state_root(header);
```
michaelsproul: Although this silences the warning, I'm not sure it's right.

In the original code, the compiler is telling us that writing to `header.state_root` is completely pointless, because `header` is never used. That remains true with your refactor.

This looks like a bug in the original code to me. Either we need to use `header` in some way, or it is completely superfluous and can be deleted. I need to check the LC spec.

michaelsproul: The compiler warning on 1.92 beta is:

> warning: variable `header` is assigned to, but never used
   --> consensus/types/src/light_client_bootstrap.rs:191:13
    |
191 |         let mut header = beacon_state.latest_block_header().clone();
    |             ^^^^^^^^^^
    |
    = note: consider using `_header` instead
    = note: `#[warn(unused_variables)]` (part of `#[warn(unused)]`) on by default
> 
> warning: value assigned to `header` is never read
   --> consensus/types/src/light_client_bootstrap.rs:192:9
    |
192 |         header.state_root = beacon_state.update_tree_hash_cache()?;
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: maybe it is overwritten before being read?
    = note: `#[warn(unused_assignments)]` (part of `#[warn(unused)]`) on by default

hopinheimer: from a shallow overview to me it looked like it was updating the `state_root` in the `beacon_state.latest_block_header`

michaelsproul: yeah, that would make sense. The latest block header in the state doesn't contain the state root

but we probably don't want to mutate the original state, because that would make it invalid

michaelsproul: I think we can just delete the `header` variable completely. It looks like these two lines of code are old (2022/2023), and not necessary.

The spec describes initialising the `LightClientHeader` from the block, which doesn't require us to inject the state root: https://github.com/ethereum/consensus-specs/blob/master/specs/altair/light-client/full-node.md#block_to_light_client_header

And this agrees with how we create the header here:

https://github.com/sigp/lighthouse/blob/e5b4983d6baf85770fe4539a565d8a2dd462bc53/consensus/types/src/light_client_header.rs#L163-L172

hopinheimer: gotchya!

consensus/types/src/beacon_block_header.rs
```diff
@@ -62,6 +62,10 @@ impl BeaconBlockHeader {
             state_root: Default::default(),
         }
     }
+
+    pub fn update_state_root(&mut self, state_root: Hash256) {
```
michaelsproul: can revert this as well

hopinheimer: done

michaelsproul (approved):
LGTM!

We can merge this now and I'll run it by @eserilev once he's online

jimmygchen:
@mergifyio requeue

PR #8319, opened 2025-10-29
Add nightly tests workflow to test prior forks

jimmygchen:
## Issue Addressed

Once #8271 is merged, CI will only cover tests for `RECENT_FORKS` (prev, current, next)

To make sure functionalities aren't broken for prior forks, we run tests for these forks nightly. They can also be manually triggered.

Tested via manual trigger here: https://github.com/jimmygchen/lighthouse/actions/runs/18896690117

image

### DIFF
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -1,5 +1,5 @@
 # We only run tests on `RECENT_FORKS` on CI. To make sure we don't break prior forks, we run nightly tests to cover all prior forks.
-name: nightly-fork-tests
+name: nightly-tests
 
 on:
   schedule:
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly-tests.yml
```diff
@@ -0,0 +1,135 @@
+# We only run tests on `RECENT_FORKS` on CI. To make sure we don't break prior forks, we run nightly tests to cover all prior forks.
+name: nightly-tests
+
+on:
+  schedule:
+    # Run at 8:30 AM UTC every day
+    - cron: '30 8 * * *'
```
antondlr, line 7: is this just _a_ time slot or is there reasoning behind it?
I think aiming for it to be ready by AUS AM is maybe a good idea?
also, UTC start of workday there may be heavier-than-average demand but I have no evidence of that

antondlr (approved):
great!
I could not inspect the outcome on your fork though (no access) but it LGTM

.github/workflows/nightly-tests.yml
jimmygchen: yeah i looked at the timezones, I think it's more quiet on the lighthouse side , but maybe not on the anchor side?

jimmygchen: Thanks for the review! Let's try this out and we can move it if it doesn't work out

PR #8321, opened 2025-10-29
Fix flaky reconstruction test

jimmygchen:
## Issue Addressed

FIx flaky tests that depends on timing. Previously the test processes all 128 columns and expect reconstruction to happen after all columns are processed. There is a race here, and reconstruction could be triggered before all columns are processed.

I've updated the tests to process 64 columns, just enough for reconstruction and wait for 50ms for reconstruction to be triggered.

This PR requires the change made in https://github.com/sigp/lighthouse/pull/8194 for the test to pass consistently (blob count set to 1 for all blocks instead of random blob count between 0..max)


michaelsproul (approved):
LGTM, I want this change after [hitting the failure on one of my PRs](https://github.com/sigp/lighthouse/actions/runs/19219008589/job/54933238081?pr=8378).

GIMME!

Issue #7089, opened 2025-03-10
Checkpoint sync from non-finalized checkpoint


michaelsproul:
The Holesky non-finality incident has revealed the utility of being able to sync from a checkpoint that is not finalized.

We've added functionality to consider an arbitrary checkpoint finalized in Lighthouse's database in these PRs:

- https://github.com/sigp/lighthouse/pull/7059
- https://github.com/sigp/lighthouse/pull/7084

The problem is, the node still has to sync to the desired block before finalizing it.

To improve on this, we could use a similar approach to start from a `split` that is ahead of finalization. By updating _only the split_ and not other indicators of finalization, we can hopefully avoid the roadblocks which stymied our previous attempt at non-finalized checkpoint sync:

- https://github.com/sigp/lighthouse/pull/5128

michaelsproul:
Previous issue that might also contain relevant info:

- https://github.com/sigp/lighthouse/issues/5082

jimmygchen:
@michaelsproul @pawanjay176 does this need to be in `v8.0.0-rc.0`?

michaelsproul:
no

dapplion:
NOTE: test that a node after checkpoint sync into non-finalized checkpoint and before the first finalization can produce blocks that can become canonical.

PR #8324, opened 2025-10-29
Remove duplicate state in ProtoArray

dapplion:
## Issue Addressed

Part of a fork-choice tech debt clean-up https://github.com/sigp/lighthouse/issues/8325

https://github.com/sigp/lighthouse/issues/7089 (non-finalized checkpoint sync) changes the meaning of the checkpoints inside fork-choice. It turns out that we persist the justified and finalized checkpoints **twice** in fork-choice
1. Inside the fork-choice store
2. Inside the proto-array

There's no reason for 2. except for making the function signature of some methods smallers. It's not consistent with the rest of the crate, because in some functions we pass the external variable of time (current_slot) via args, but then read the finalized checkpoint from the internal state. Passing both variables as args makes fork-choice easier to reason about at the cost of a few extra lines.

## Proposed Changes

Remove the unnecessary state (`justified_checkpoint`, `finalized_checkpoint`) inside `ProtoArray`, to make it easier to reason about.


consensus/proto_array/src/ssz_container.rs
```diff
@@ -40,8 +42,8 @@ impl From<&ProtoArrayForkChoice> for SszContainer {
         Self {
             votes: from.votes.0.clone(),
             prune_threshold: proto_array.prune_threshold,
-            justified_checkpoint: proto_array.justified_checkpoint,
-            finalized_checkpoint: proto_array.finalized_checkpoint,
+            justified_checkpoint: <_>::default(),
+            finalized_checkpoint: <_>::default(),
```
michaelsproul: I think we might need to do the schema migration _now_ because of this. If someone runs this commit and stores `0x0` checkpoints on disk, then downgrades to a commit prior to this one, it will break.

michaelsproul (changes requested):
Needs a schema migration, see my comment.

consensus/proto_array/src/ssz_container.rs
dapplion: Pushed a commit to write the checkpoints to disk. Feels simpler than a migration. Is it best to lump a bunch of fork-choice improvements together and do a single migration?

michaelsproul (approved):

PR #8335, opened 2025-10-30
Proposer duties backwards compat

michaelsproul:
## Issue Addressed

The beacon API spec wasn't updated to use the Fulu definition of `dependent_root` for the proposer duties endpoint. No other client updated their logic, so to retain backwards compatibility the decision has been made to continue using the block root at the end of epoch `N - 1`, and introduce a new v2 endpoint down the track to use the correct dependent root.

Eth R&D discussion: https://discord.com/channels/595666850260713488/598292067260825641/1433036715848765562

## Proposed Changes

Change the behaviour of the v1 endpoint back to using the last slot of `N - 1` rather than the last slot of `N - 2`. This introduces the possibility of dependent root false positives (the root can change without changing the shuffling), but causes the least compatibility issues with other clients.

### DIFF
--- a/consensus/types/src/beacon_state.rs
+++ b/consensus/types/src/beacon_state.rs
@@ -911,6 +911,22 @@ impl<E: EthSpec> BeaconState<E> {
         }
     }
 
+    /// Returns the block root at the last slot of `epoch - 1`.
+    ///
+    /// This can be deleted after Glamsterdam and the removal of the v1 proposer duties endpoint.
+    pub fn legacy_proposer_shuffling_decision_root_at_epoch(
+        &self,
+        epoch: Epoch,
+        head_block_root: Hash256,
+    ) -> Result<Hash256, Error> {
+        let decision_slot = epoch.saturating_sub(1u64).end_slot(E::slots_per_epoch());
+        if self.slot() <= decision_slot {
+            Ok(head_block_root)
+        } else {
+            self.get_block_root(decision_slot).copied()
+        }
+    }
+
     /// Returns the block root which decided the proposer shuffling for the current epoch. This root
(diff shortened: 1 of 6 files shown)


consensus/types/src/beacon_state.rs
```diff
@@ -911,6 +911,22 @@ impl<E: EthSpec> BeaconState<E> {
...
 
+    /// Returns the block root at the last slot of `epoch - 1`.
+    ///
+    /// This can be deleted after Glamsterdam and the removal of the v1 proposer duties endpoint.
+    pub fn legacy_proposer_shuffling_decision_root_at_epoch(
+        &self,
+        epoch: Epoch,
+        head_block_root: Hash256,
+    ) -> Result<Hash256, Error> {
+        let decision_slot = epoch.saturating_sub(1u64).end_slot(E::slots_per_epoch());
```
jimmygchen, line 922: I think this is fine, but just want to note that the previous logic uses `min_seed_lookahead`, but no live network modifies this config

https://github.com/sigp/lighthouse/blob/d67ae921125dd22e15d91aa324a27725ac03df41/consensus/types/src/chain_spec.rs#L894-L900

jimmygchen (approved):
Looks good to me!

consensus/types/src/beacon_state.rs
michaelsproul: Oh yeah good point. Should be OK seeing as we plan to deprecate v1 proposer duties at Glamsterdam

- https://github.com/ethereum/beacon-APIs/pull/563

jimmygchen:
@mergify dequeue

jimmygchen:
@mergifyio requeue

Issue #8311, opened 2025-10-28
Reduce spurious recompilation due to `lighthouse_version`


michaelsproul:
## Description

@macladson and I tracked down the issues with `lighthouse_version` that cause spurious recompilations:

- https://github.com/m-ou-se/rust-git-version/issues/26

I think the solution is to abandon the `git-version` crate and the concept of "dirtyness" tracking, as IMO there is no way to do it correctly, and doing it badly (what we do currently) causes more recompilations for little benefit.

Most of the time we are running builds from specific commits (we should _always_ be doing this on our infra), so dirtyness tracking only affects local builds. The new behaviour will be to only trigger a recompilation of `lighthouse_version` if a commit is made.

PR #8336, opened 2025-10-30
Rework `lighthouse_version` to reduce spurious recompilation

macladson:
## Issue Addressed

#8311

## Proposed Changes

Removes the `git_version` crate from `lighthouse_version` and implements git `HEAD` tracking manually. 
This removes the (mostly) broken dirty tracking but prevents spurious recompilation of the `lighthouse_version` crate.

This also reworks the way crate versions are handled by utilizing workspace version inheritance and Cargo environment variables.
This means the _only_ place where Lighthouse's version is defined is in the top level `Cargo.toml` for the workspace. All relevant binaries then inherit this version. This largely makes the  `change_version.sh` script useless so I've removed it, although we could keep a version which just alters the workspace version (if we need to maintain compatibility with certain build/release tooling.

### When is a Rebuild Triggered?

1. When the build.rs file is changed.
2. When the HEAD commit changes (added, removed, rebased, etc)
3. When the branch changes (this includes changing to the current branch, and creating a detached HEAD)

Note that working/staged changes will not trigger a recompile of `lighthouse_version`.

## Additional Info

There might exist certain edge cases with packed git refs where a recompile might not get triggered after a git garbage collection cycle but I suspect this will be quite rare.


common/lighthouse_version/src/lib.rs
```diff
@@ -1,49 +1,25 @@
...
-        // NOTE: using --match instead of --exclude for compatibility with old Git
-        "--match=thiswillnevermatchlol"
-    ],
-    prefix = "",
-    suffix = "",
-    cargo_prefix = "",
-    cargo_suffix = "",
-    fallback = "00000000"
-);
+pub const COMMIT_PREFIX: &str = env!("GIT_COMMIT_PREFIX");
```
macladson, line 16: We could also consider moving to an 8 character commit hash. Currently we provide both a 7 character hash (for `--version`, etc) and an 8 character hash (for execution engine reporting). If we can use 8 character hashes for everything, that would reduce some complexity

scripts/change_version.sh
```diff
@@ -1,34 +0,0 @@
...
-}
-
-echo "Changing version from $FROM to $TO"
-
-update_cargo_toml ../account_manager/Cargo.toml
-update_cargo_toml ../beacon_node/Cargo.toml
-update_cargo_toml ../boot_node/Cargo.toml
-update_cargo_toml ../lcli/Cargo.toml
-update_cargo_toml ../lighthouse/Cargo.toml
-update_cargo_toml ../validator_client/Cargo.toml
```
macladson, line 31: I'm using workspace version inheritance for the binaries that were explicitly listed here, but we could also include the other ones (`database_manager`, `validator_manager`, `slasher`) to keep it consistent

pawanjay176 (approved):
This is amazing! I have been annoyed with this behaviour for years. Thanks for fixing it.
rust-analyzer runs so much quicker now that it doesn't recompile lighthouse_version each time.

michaelsproul (approved):
Looks great! So glad to have this fixed finally!

PR #8338, opened 2025-10-30
Gloas vc ptc duty

shane-moore:
## Changes
- added polling of ptc duties to the VC per [spec](https://ethereum.github.io/consensus-specs/specs/gloas/validator/#lookahead)

## How This Works
- The VC is currently checking whether one of its registered validators are in a voting committee, such as sync or attestation, or if they are the proposer for a slot.  With epbs, it's possible for a validator to also be in a ptc committee for at most one slot of the epoch.  More context on how validators are selected to be in a ptc was discussed on [discord](https://discord.com/channels/595666850260713488/874767108809031740/1430262787178758376).

## Implemented in the PR
`duties_service` will now spawn a background polling task for the current and next epoch to check whether any of the registered validators are in a ptc committee.  Since re-org is possible, we also poll every slot.  The beacon node querying is performed by new request endpoint `POST validator/duties/ptc/{epoch}` per [beacon api spec]( https://github.com/ethereum/beacon-APIs/pull/552).

To save on bandwidth similar to attestation committee polling, the ptc polling will make a tiny initial requests to learn the `dependent_root`, compare with the cached `dependent_root`, and only fetch full duties for the epoch if the root has changed (or if the root is unknown like for a new epoch).  Old duties are pruned based on `HISTORICAL_DUTIES_EPOCHS`. 

## Todo
- add tests noted in the diffs
- create beacon node (server side) response that will send the ptc duties to the VC.  Some of this work has already been done in other PR's such as the `get_ptc` helper, but the `warp` implementation still needs to be complete as to send the response to the VC

@ethDreamer
@eserilev


ethDreamer:
> Since re-org is possible, we also poll every slot.

I don't believe this is correct. Have a look at [`get_ptc()`](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/beacon-chain.md#new-get_ptc)
```python
def get_ptc(state: BeaconState, slot: Slot) -> Vector[ValidatorIndex, PTC_SIZE]:
    """
    Get the payload timeliness committee for the given ``slot``.
    """
    epoch = compute_epoch_at_slot(slot)
    seed = hash(get_seed(state, epoch, DOMAIN_PTC_ATTESTER) + uint_to_bytes(slot))
    indices: List[ValidatorIndex] = []
    # Concatenate all committees for this slot in order
    committees_per_slot = get_committee_count_per_slot(state, epoch)
    for i in range(committees_per_slot):
        committee = get_beacon_committee(state, slot, CommitteeIndex(i))
        indices.extend(committee)
    return compute_balance_weighted_selection(
        state, indices, seed, size=PTC_SIZE, shuffle_indices=False
    )
```
The committee selection appears constant for the same seed, and the seed is constant over `MIN_SEED_LOOKAHEAD = 1 epoch`

ethDreamer:
ohhh you just mean that you check the *cache* every slot to ensure we have the same dependent root.

validator_client/validator_services/src/duties_service.rs
```diff
@@ -276,6 +278,7 @@ type DependentRoot = Hash256;
 
 type AttesterMap = HashMap<PublicKeyBytes, HashMap<Epoch, (DependentRoot, DutyAndProof)>>;
 type ProposerMap = HashMap<Epoch, (DependentRoot, Vec<ProposerData>)>;
+type PtcMap = HashMap<Epoch, (DependentRoot, HashMap<PublicKeyBytes, PtcDuty>)>;
```
ethDreamer: If the beacon node only returns duties for pubkeys that this validator custodies, then this should probably just be a
```
HashMap<Epoch, (DependentRoot, Vec<PtcDuty>)>
```
as this Vec will almost always be quite small and iterating through it will be fast

eserilev (approved):
LGTM!

validator_client/validator_services/src/duties_service.rs
```diff
@@ -472,6 +478,24 @@ impl<S: ValidatorStore, T: SlotClock + 'static> DutiesService<S, T> {
             .count()
     }
 
+    /// Returns the total number of validators that have PTC duties in the given epoch.
+    pub fn ptc_count(&self, epoch: Epoch) -> usize {
+        // Only collect validators that are considered safe in terms of doppelganger protection.
```
pawanjay176: is this required? ptc votes aren't slashable right?

pawanjay176 (approved):
LGTM. Just a question

validator_client/validator_services/src/duties_service.rs
shane-moore: correct that ptc votes aren't slashable. will take a look at what i was thinking here tomorrow 🤙

eserilev: This was me, don't worry about it!

eserilev: removed doppleganger protection

eserilev:
@mergify requeue

PR #8339, opened 2025-10-30
Ensure custody backfill sync couples all responses before importing

eserilev:
## Issue Addressed

Custody backfill sync has a bug when we request columns from more than one peer per batch. The fix here ensures we wait for all requests to be completed before performing verification and importing the responses.

I've also added an endpoint `lighthouse/custody/backfill` that resets a nodes earliest available data column to the current epoch so that custody backfill can be triggered. This endpoint is needed to rescue any nodes that may have missing columns due to the custody backfill sync bug without requiring a full re-sync.

### DIFF
--- a/beacon_node/beacon_chain/src/custody_context.rs
+++ b/beacon_node/beacon_chain/src/custody_context.rs
@@ -1530,7 +1530,7 @@ mod tests {
         let custody_context = setup_custody_context(&spec, head_epoch, epoch_and_cgc_tuples);
 
         // Backfill from epoch 20 to 9
-        complete_backfill_for_epochs(&custody_context, Epoch::new(20), Epoch::new(9));
+        complete_backfill_for_epochs(&custody_context, Epoch::new(20), Epoch::new(9), final_cgc);
 
         // Reset validator custody requirements to the latest cgc requirements at `head_epoch` up to the boundary epoch
         custody_context.reset_validator_custody_requirements(head_epoch);
--- a/beacon_node/beacon_chain/src/custody_context.rs
+++ b/beacon_node/beacon_chain/src/custody_context.rs
@@ -151,7 +151,7 @@ impl ValidatorRegistrations {
         }
     }
 
-    /// Updates the `epoch -> cgc` map by prunning records in between `effective_epoch` and `boundary_epoch`
+    /// Updates the `epoch -> cgc` map by pruning records before `effective_epoch`
     /// while setting the `cgc` at `effective_epoch` to the latest validator custody requirement.
     ///
     /// This is used to restart custody backfill sync at `effective_epoch`
--- a/beacon_node/beacon_chain/src/custody_context.rs
+++ b/beacon_node/beacon_chain/src/custody_context.rs
@@ -134,8 +134,17 @@ impl ValidatorRegistrations {
     ///
     /// This is done by pruning all values on/after `effective_epoch` and updating the map to store
     /// the latest validator custody requirements for the `effective_epoch`.
-    pub fn backfill_validator_custody_requirements(&mut self, effective_epoch: Epoch) {
+    pub fn backfill_validator_custody_requirements(
+        &mut self,
+        effective_epoch: Epoch,
+        expected_cgc: u64,
+    ) {
         if let Some(latest_validator_custody) = self.latest_validator_custody_requirement() {
+            // If the expected cgc isn't equal to the latest validator custody a very recent cgc change may have occurred.
+            // We should not update the mapping.
+            if expected_cgc != latest_validator_custody {
+                return;
+            }
             // Delete records if
             // 1. The epoch is greater than or equal than `effective_epoch`
             // 2. the cgc requirements match the latest validator custody requirements
@@ -535,10 +544,14 @@ impl<E: EthSpec> CustodyContext<E> {
 
     /// The node has completed backfill for this epoch. Update the internal records so the function
     /// [`Self::custody_columns_for_epoch()`] returns up-to-date results.
-    pub fn update_and_backfill_custody_count_at_epoch(&self, effective_epoch: Epoch) {
+    pub fn update_and_backfill_custody_count_at_epoch(
+        &self,
+        effective_epoch: Epoch,
+        expected_cgc: u64,
+    ) {
         self.validator_registrations
             .write()
-            .backfill_validator_custody_requirements(effective_epoch);
+            .backfill_validator_custody_requirements(effective_epoch, expected_cgc);
     }
 
     /// The node is attempting to restart custody backfill. Update the internal records so that
@@ -630,11 +643,13 @@ mod tests {
         custody_context: &CustodyContext<E>,
         start_epoch: Epoch,
         end_epoch: Epoch,
+        expected_cgc: u64,
     ) {
         assert!(start_epoch >= end_epoch);
         // Call from end_epoch down to start_epoch (inclusive), simulating backfill
         for epoch in (end_epoch.as_u64()..=start_epoch.as_u64()).rev() {
-            custody_context.update_and_backfill_custody_count_at_epoch(Epoch::new(epoch));
+            custody_context
+                .update_and_backfill_custody_count_at_epoch(Epoch::new(epoch), expected_cgc);
         }
     }
 
@@ -1394,7 +1409,7 @@ mod tests {
         );
 
         // Backfill from epoch 20 down to 15 (simulating backfill)
-        complete_backfill_for_epochs(&custody_context, head_epoch, Epoch::new(15));
+        complete_backfill_for_epochs(&custody_context, head_epoch, Epoch::new(15), final_cgc);
 
         // After backfilling to epoch 15, it should use latest CGC (32)
         assert_eq!(
@@ -1432,9 +1447,61 @@ mod tests {
         let custody_context = setup_custody_context(&spec, head_epoch, epoch_and_cgc_tuples);
 
         // Backfill to epoch 15 (between the two CGC increases)
-        complete_backfill_for_epochs(&custody_context, Epoch::new(20), Epoch::new(15));
+        complete_backfill_for_epochs(&custody_context, Epoch::new(20), Epoch::new(15), final_cgc);
 
         // Verify epochs 15 - 20 return latest CGC (32)
+        for epoch in 15..=20 {
+            assert_eq!(
+                custody_context.custody_group_count_at_epoch(Epoch::new(epoch), &spec),
+                final_cgc,
+            );
+        }
+
+        // Verify epochs 10-14 still return mid_cgc (16)
+        for epoch in 10..14 {
+            assert_eq!(
+                custody_context.custody_group_count_at_epoch(Epoch::new(epoch), &spec),
+                mid_cgc,
+            );
+        }
+    }
+
+    #[test]
+    fn attempt_backfill_with_invalid_cgc() {
+        let spec = E::default_spec();
+        let initial_cgc = 8u64;
+        let mid_cgc = 16u64;
+        let final_cgc = 32u64;
+
+        // Setup: Node restart after multiple validator registrations causing CGC increases
+        let head_epoch = Epoch::new(20);
+        let epoch_and_cgc_tuples = vec![
+            (Epoch::new(0), initial_cgc),
+            (Epoch::new(10), mid_cgc),
+            (head_epoch, final_cgc),
+        ];
+        let custody_context = setup_custody_context(&spec, head_epoch, epoch_and_cgc_tuples);
+
+        // Backfill to epoch 15 (between the two CGC increases)
+        complete_backfill_for_epochs(&custody_context, Epoch::new(20), Epoch::new(15), final_cgc);
+
+        // Verify epochs 15 - 20 return latest CGC (32)
+        for epoch in 15..=20 {
+            assert_eq!(
+                custody_context.custody_group_count_at_epoch(Epoch::new(epoch), &spec),
+                final_cgc,
+            );
+        }
+
+        // Attempt backfill with an incorrect cgc value
+        complete_backfill_for_epochs(
+            &custody_context,
+            Epoch::new(20),
+            Epoch::new(15),
+            initial_cgc,
+        );
+
+        // Verify epochs 15 - 20 still return latest CGC (32)
         for epoch in 15..=20 {
             assert_eq!(
                 custody_context.custody_group_count_at_epoch(Epoch::new(epoch), &spec),
--- a/beacon_node/beacon_chain/src/custody_context.rs
+++ b/beacon_node/beacon_chain/src/custody_context.rs
@@ -120,9 +120,7 @@ impl ValidatorRegistrations {
             let effective_epoch =
                 (current_slot + effective_delay_slots).epoch(E::slots_per_epoch()) + 1;
             self.epoch_validator_custody_requirements
-                .entry(effective_epoch)
-                .and_modify(|old_custody| *old_custody = validator_custody_requirement)
-                .or_insert(validator_custody_requirement);
+                .insert(effective_epoch, validator_custody_requirement);
             Some((effective_epoch, validator_custody_requirement))
         } else {
             None
@@ -154,9 +152,7 @@ impl ValidatorRegistrations {
                 });
 
             self.epoch_validator_custody_requirements
-                .entry(effective_epoch)
-                .and_modify(|old_custody| *old_custody = latest_validator_custody)
-                .or_insert(latest_validator_custody);
+                .insert(effective_epoch, latest_validator_custody);
         }
     }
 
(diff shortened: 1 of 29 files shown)


beacon_node/http_api/tests/interactive_tests.rs
```diff
@@ -1050,22 +1053,87 @@ async fn proposer_duties_with_gossip_tolerance() {
...
 // Test that a request for next epoch proposer duties suceeds when the current slot clock is within
 // gossip clock disparity (500ms) of the new epoch.
 #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
 async fn lighthouse_custody_info() {
     let mut spec = test_spec::<E>();
 
-    // Skip pre-Fulu.
-    if !spec.is_fulu_scheduled() {
-        return;
-    }
```
eserilev, line 1062: we actually don't seem to have the `fork_from_env` feature available in the `http_api` tests so I dont think this test was actually doing anything in CI.

eserilev:
https://github.com/sigp/lighthouse/pull/8340 needs to be merged and cherry-picked into `release-v8.0.` for CI to pass

eserilev:
I'm testing the api changes locally, I'll mark this ready for review once I'm done with that

eserilev:
both the custody bakcfill sync bugdix and new api seem to be working just fine on a local sepolia node, tomorrow I'm going to deploy this to one of our internal hoodi nodes and try a custody backfill re-sync using the new api

beacon_node/http_api/tests/interactive_tests.rs
michaelsproul: We do:

https://github.com/sigp/lighthouse/blob/e5b4983d6baf85770fe4539a565d8a2dd462bc53/Makefile#L182-L183

michaelsproul: Turning it on in `beacon_chain` influences the behaviour of `test_spec::<E>()` everywhere

michaelsproul: This check is intended to avoid running the test too many times (it just runs for Fulu and later forks)

jimmygchen:
> both the custody bakcfill sync bugdix and new api seem to be working just fine on a local sepolia node, tomorrow I'm going to deploy this to one of our internal hoodi nodes and try a custody backfill re-sync using the new api

I'll do this now

book/src/api_lighthouse.md
```diff
@@ -447,6 +447,15 @@ indicating that all states with slots `>= 0` are available, i.e., full state his
 on the specific meanings of these fields see the docs on [Checkpoint
 Sync](./advanced_checkpoint_sync.md#how-to-run-an-archived-node).
 
+## `/lighthouse/custody/backfill`
+
+Starts a custody backfill sync from the current epoch with the nodes latest custody requirements. The sync won't begin immediately, it waits until the current epoch is finalized before triggering.
```
jimmygchen: Is this the same behaviour for regular custody backfill when CGC is increased?

eserilev: Sort of, the only difference is that we prune previous cgc records from that validator requirements mapping. After calling this endpoint the mapping will only contain one record , the latest cgc requirement at the current epoch

beacon_node/beacon_chain/src/custody_context.rs
```diff
@@ -150,6 +150,24 @@ impl ValidatorRegistrations {
                 .or_insert(latest_validator_custody);
         }
     }
+
+    /// Updates the `epoch -> cgc` map by pruning records before `effective_epoch`
+    /// while setting the `cgc` at `effective_epoch` to the latest validator custody requirement.
+    ///
+    /// This is used to restart custody backfill sync at `effective_epoch`
+    pub fn reset_validator_custody_requirements(&mut self, effective_epoch: Epoch) {
```
jimmygchen, line 163: What happens if the latest epoch in `epoch_validator_custody_requirements` is `effective_epoch + 1`?

I _think_ it keeps the `next_epoch` entry in the map, which is probably ok, so we may end up having something like this in the map initially:

```rust
  {
    (current_epoch, 128),
    (next_epoch, 128)
  }
```

and we also overwrite the `DataColumnCustodyInfo` from `next_epoch` to `current_epoch`, and use `current epoch + 1` as backfill start epoch, which also seems fine to me

jimmygchen: It looks like the `next_epoch` entry will be removed once backfill completes a batch, so we shouldn't end up having it permanently in the map

https://github.com/sigp/lighthouse/blob/56111a8224381d38d2437806f510ab79f4731f2a/beacon_node/beacon_chain/src/custody_context.rs#L139-L145

jimmygchen: It breaks the assumption here though

https://github.com/sigp/lighthouse/blob/56111a8224381d38d2437806f510ab79f4731f2a/beacon_node/beacon_chain/src/custody_context.rs#L33-L34

eserilev: Yeah there is that edge case you're right. But users that have a future epoch with a cgc change shouldn't be using this endpoint, as custody backfill will already be scheduled to start for them. This is really just an endpoint to fix nodes that were broken because of the custody backfill bug. Maybe we could add documentation to this endpoint mentioning that?

jimmygchen:
When custody backfill is manually triggered, it quickily fills up the beacon.log files with this log:
```
Nov 03 03:29:07.626 DEBUG Skipping data column import as identical data column exists  block_root: ...
```

I think we should remove this debug log, as it's a bit excessive when it logs 128 entries per slot on a supernode, and makes it difficult to debug if things go wrong.

beacon_node/beacon_chain/src/custody_context.rs
```diff
@@ -150,6 +150,24 @@ impl ValidatorRegistrations {
...
+        if let Some(latest_validator_custody_requirements) =
+            self.latest_validator_custody_requirement()
+        {
+            self.epoch_validator_custody_requirements
+                .retain(|&epoch, _| epoch >= effective_epoch);
+
+            self.epoch_validator_custody_requirements
+                .entry(effective_epoch)
```
jimmygchen: shouldn't this be `effective_epoch + 1`

We don't have the new CGC until backfill for  the epoch is completed right?

beacon_node/beacon_chain/src/custody_context.rs
jimmygchen: yeah adding doc sounds good, and maybe we could mentioned the known version that was broken (rc.2)

beacon_node/http_api/src/lib.rs
```diff
@@ -4604,6 +4604,36 @@ pub fn serve<T: BeaconChainTypes>(
...
+    // POST lighthouse/custody/backfill
+    let post_lighthouse_custody_backfill = warp::path("lighthouse")
+        .and(warp::path("custody"))
+        .and(warp::path("backfill"))
+        .and(warp::path::end())
+        .and(task_spawner_filter.clone())
+        .and(chain_filter.clone())
+        .then(
+            |task_spawner: TaskSpawner<T::EthSpec>, chain: Arc<BeaconChain<T>>| {
+                task_spawner.blocking_json_task(Priority::P0, move || {
```
jimmygchen: I don't think this matters if user's using this correctly and if this function is fast - in terms of priority this isn't higher priority than getting node in sync or performing validator duties - i think it may be worth downgrading? in the unlikely case that the `cached_head_read_lock` was held elsewhere, we could delay higher priority tasks.

beacon_node/http_api/src/lib.rs
```diff
@@ -4604,6 +4604,36 @@ pub fn serve<T: BeaconChainTypes>(
...
+                        .cached_head()
+                        .head_slot()
+                        .epoch(T::EthSpec::slots_per_epoch());
+                    let custody_context = chain.data_availability_checker.custody_context();
+                    // Reset validator custody requirements to `current_epoch` with the latest
+                    // cgc requiremnets.
+                    custody_context.reset_validator_custody_requirements(current_epoch);
+                    // Update `DataColumnCustodyInfo` to reflect the custody change.
+                    chain.update_data_column_custody_info(Some(
+                        current_epoch.start_slot(T::EthSpec::slots_per_epoch()),
```
jimmygchen: this should be `current_epoch + 1`, no?
because `earliest_data_column_slot` is the start slot from where we have all the required columns, we should reset it to a future slot - in case we have missing columns in the current epoch, we could be advertising incorrectly.

beacon_node/http_api/tests/interactive_tests.rs
```diff
@@ -1050,22 +1053,87 @@ async fn proposer_duties_with_gossip_tolerance() {
     );
 }
 
+// Test that a request to `lighthouse/custody/backfill` succeeds by verifying that `CustodyContext` and `DataColumnCustodyInfo`
+// have been updated with the correct values.
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn lighthouse_restart_custody_backfill() {
```
jimmygchen, line 1059: We should add the condition to skip pre fulu forks to avoid running this on prior forks.

beacon_node/http_api/tests/interactive_tests.rs
```diff
@@ -1050,22 +1053,87 @@ async fn proposer_duties_with_gossip_tolerance() {
...
+    let mut spec = test_spec::<E>();
+
+    spec.altair_fork_epoch = Some(Epoch::new(0));
+    spec.bellatrix_fork_epoch = Some(Epoch::new(0));
+    spec.capella_fork_epoch = Some(Epoch::new(0));
+    spec.deneb_fork_epoch = Some(Epoch::new(0));
+    spec.electra_fork_epoch = Some(Epoch::new(0));
+    spec.fulu_fork_epoch = Some(Epoch::new(0));
```
jimmygchen: This isn't necessary when using `test_spec` and `fork_from_env` (as michael mentioned, all tests in `http-api` runs with this)

beacon_node/http_api/tests/interactive_tests.rs
```diff
@@ -1050,22 +1053,87 @@ async fn proposer_duties_with_gossip_tolerance() {
...
+
+    let validator_count = 24;
+
+    let tester = InteractiveTester::<E>::new_supernode(Some(spec), validator_count).await;
```
jimmygchen, line 1069: Do we need supernode for this test?

This test takes 1 min to execute, which means every CI run could potentially be slowed down by 1 min. It seems a bit expensive for testing simple logic like this.

I think we need to think about strategies for these kind of tests, options include: 
* start a single `InteractiveTester` and run multiple http api tests
* have the api handling logic extracted into a function, so we can test it without an http server - we do lose the check for the API path though, but i'd argue the benefit of faster CI may outweigh its cost e.g.
  https://github.com/sigp/lighthouse/blob/e34c8816bad09053fcedab749468f5d607503e83/beacon_node/http_api/src/lib.rs#L4603

book/src/api_lighthouse.md
jimmygchen: There's a lint error in the book

```
src/api_lighthouse.md:458 MD012/no-multiple-blanks Multiple consecutive blank lines [Expected: 1; Actual: 2]
Exiting with errors. Run 'make mdlint' locally and commit the changes. Note that not all errors can be fixed automatically, if there are still errors after running 'make mdlint', look for the errors and fix manually.
```

beacon_node/beacon_chain/src/custody_context.rs
```diff
@@ -150,6 +150,24 @@ impl ValidatorRegistrations {
...
+                .and_modify(|old_custody| *old_custody = latest_validator_custody_requirements)
+                .or_insert(latest_validator_custody_requirements);
```
michaelsproul: Isn't this just `.insert`?

```suggestion
            self.epoch_validator_custody_requirements
                .insert(effective_epoch, latest_validator_custody_requirements)
```

beacon_node/beacon_chain/src/custody_context.rs
michaelsproul: I think you might be mixing this up with the `earliest_data_column_slot`? It seems that the `epoch` values in this map mean:

- for epochs `>= epoch`, the custody requirement is X.

So using `epoch` rather than `epoch + 1` seems right to me?

beacon_node/http_api/src/lib.rs
michaelsproul: Agree this should be `current_epoch` + 1.

jimmygchen:
@eserilev Thanks for the PR! I'm going to start addressing some changes so we can get it merged before end of day.

beacon_node/http_api/tests/interactive_tests.rs
jimmygchen: I'm adding this back.

beacon_node/beacon_chain/src/custody_context.rs
jimmygchen: Yeah, that's right.

It makes sense to use `epoch` in the context of the function, we should add the `+1` on the caller site though.
Because (under an edge case) we could be incorrectly advertising and requiring the incorrect CGC for the epoch that we may have partial columns.

Discussed with @michaelsproul as well - regardless this is a pretty low risk change and shouldn't break backfill.

beacon_node/http_api/tests/interactive_tests.rs
jimmygchen: This takes only 16s on my machine, I'm going to leave this for now and address the slow CI separately so we can release.

beacon_node/beacon_chain/src/custody_context.rs
jimmygchen: ah yeah you're right, i'll also update the identical code in the `backfill_validator_custody_requirements` function as  well

jimmygchen: Addressed in https://github.com/sigp/lighthouse/pull/8339/commits/580b584a8f0838906171a421c3b404ae8a905e07

beacon_node/http_api/tests/interactive_tests.rs
jimmygchen: Addressed in https://github.com/sigp/lighthouse/pull/8339/commits/580b584a8f0838906171a421c3b404ae8a905e07

beacon_node/http_api/tests/interactive_tests.rs
jimmygchen: Addressed in https://github.com/sigp/lighthouse/pull/8339/commits/580b584a8f0838906171a421c3b404ae8a905e07

beacon_node/http_api/src/lib.rs
jimmygchen: Addressed in https://github.com/sigp/lighthouse/pull/8339/commits/580b584a8f0838906171a421c3b404ae8a905e07

beacon_node/http_api/src/lib.rs
jimmygchen: Addressed in https://github.com/sigp/lighthouse/pull/8339/commits/580b584a8f0838906171a421c3b404ae8a905e07

beacon_node/beacon_chain/src/custody_context.rs
jimmygchen: Addressed in https://github.com/sigp/lighthouse/pull/8339/commits/580b584a8f0838906171a421c3b404ae8a905e07

book/src/api_lighthouse.md
jimmygchen: Addressed in https://github.com/sigp/lighthouse/pull/8339/commits/580b584a8f0838906171a421c3b404ae8a905e07

beacon_node/http_api/tests/interactive_tests.rs
jimmygchen: Addressed in https://github.com/sigp/lighthouse/pull/8339/commits/580b584a8f0838906171a421c3b404ae8a905e07

jimmygchen (approved):
Looks good from my side, be good to get another pair of eyes to review the changes I pushed

book/src/api_lighthouse.md
```diff
@@ -447,6 +447,16 @@ indicating that all states with slots `>= 0` are available, i.e., full state his
 on the specific meanings of these fields see the docs on [Checkpoint
 Sync](./advanced_checkpoint_sync.md#how-to-run-an-archived-node).
 
+## `/lighthouse/custody/backfill`
+
+Starts a custody backfill sync from the next epoch with the nodes latest custody requirements. The sync won't begin immediately, it waits until the next epoch is finalized before triggering.
```
michaelsproul:
```suggestion
Starts a custody backfill sync from the next epoch with the node's latest custody requirements. The sync won't begin immediately, it waits until the next epoch is finalized before triggering.
```

michaelsproul (approved):
LGTM, just one doc fix

book/src/api_lighthouse.md
jimmygchen: Thanks!

jimmygchen:
@mergifyio requeue

jimmygchen:
@mergify dequeue

jimmygchen:
@mergifyio requeue

Issue #7184, opened 2025-03-21
Abolish iterators that load states


dapplion:
Currently `RootsIterator` and friends load states to iterate unfinalized block and state roots. This has shown to be problematic for memory as states are very expensive. 

- https://github.com/sigp/lighthouse/pull/6750 allows us to use state summaries to iterate through ancestor states using the new `previous_state_root` field. With that we can deprecate the state roots iterator

However, we would still need a cheap way to iterate blocks. Loading blinded blocks is a bit more inefficient as we need to load a few tens of kilobytes of data per block. Assuming 30kB per block (4096*30kB = 120MB), that's still cheaper than the memory cost of a full-state which takes about ~1GB in memory and ~150MB of I/O.

To optimize that we could either use the fork-choice (if we implement a cheap clonable DAG based fork-choice) or have some cheap block summary like states.

CC @michaelsproul

dapplion:
Initial attempt specifically for state roots https://github.com/sigp/lighthouse/compare/tree-states-hot-rebase-oom...dapplion:lighthouse:abolish-iterators?expand=1

michaelsproul:
I think we could make iterators using the hot state summaries that cover all the cases we need, i.e. spanning unfinalized back to finalized, and block roots + state roots. We can rewrite the `Hybrid*` iterator to use your new iterator in place of `RootsIterator`.

We could also separately add an implementation of `root_at_slot` which allows a faster lookup of a single root at a particular slot, using the `diff_base_state_root` to jump further back (like we're using for forming hot state summaries).

michaelsproul:
As part of the rework of iterators we should remove the `no_state_root_iter` so we don't compute state roots here:

https://github.com/sigp/lighthouse/blob/9e77f82c5f924f47d045486366bdb2d0075d1e19/beacon_node/store/src/hot_cold_store.rs#L1821-L1827

PR #8353, opened 2025-11-03
Optimise `state_root_at_slot` for finalized slot

michaelsproul:
## Proposed Changes

This is an optimisation targeted at Fulu networks in non-finality.

While debugging on Holesky, we found that `state_root_at_slot` was being called from `prepare_beacon_proposer` a lot, for the finalized state:

https://github.com/sigp/lighthouse/blob/2c9b670f5d313450252c6cb40a5ee34802d54fef/beacon_node/http_api/src/lib.rs#L3860-L3861

This was causing `prepare_beacon_proposer` calls to take upwards of 5 seconds, sometimes 10 seconds, because it would trigger _multiple_ beacon state loads in order to iterate back to the finalized slot. Ideally, loading the finalized state should be quick because we keep it cached in the state cache (technically we keep the split state, but they usually coincide). Instead we are computing the finalized state root separately (slow), and then loading the state from the cache (fast).

Although it would be possible to make the API faster by removing the `state_root_at_slot` call, I believe it's simpler to change `state_root_at_slot` itself and remove the footgun. Devs rightly expect operations involving the finalized state to be fast.

## Additional Info

Speaking of footgun removal:

- https://github.com/sigp/lighthouse/issues/7184

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -883,6 +883,12 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             return Ok(None);
         }
 
+        // Fast-path for the split slot (which usually corresponds to the finalized slot).
+        let split = self.store.get_split_info();
+        if request_slot == split.slot {
+            return Ok(Some(split.state_root));
+        }
+
         // Try an optimized path of reading the root directly from the head state.
         let fast_lookup: Option<Hash256> = self.with_head(|head| {
             if head.beacon_block.slot() <= request_slot {
--- a/beacon_node/beacon_chain/tests/store_tests.rs
+++ b/beacon_node/beacon_chain/tests/store_tests.rs
@@ -3152,6 +3152,10 @@ async fn weak_subjectivity_sync_test(
             .get_state(&state_root, Some(slot), CACHE_STATE_IN_TESTS)
             .unwrap()
             .unwrap();
+        assert_eq!(
+            state_root,
+            beacon_chain.state_root_at_slot(slot).unwrap().unwrap()
+        );
         assert_eq!(state.slot(), slot);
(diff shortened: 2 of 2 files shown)


beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -883,6 +883,12 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             return Ok(None);
         }
 
+        // Fast-path for the split slot (which usually corresponds to the finalized slot).
+        let split = self.store.get_split_info();
```
michaelsproul, line 887: This introduces a new read of the `store.split` `RwLock` in this function, prior to reading the `canonical_head` `RwLock`.

The risk here is:

Some other thread is holding the split lock exclusively (i.e. the DB finalization migration is running), and this blocks our read. Counter point: this will happen anyway (and with other downsides) if we miss the `fast_lookup` on the head state and need to load states from disk.

Maybe we could sequence the split read _after_ the head read, as we mostly care about optimising the case where the `finalized_slot` is out of range of the head but known to the DB. Conversely, maybe using the split over the head is preferable, as there is more write-contention for the head (the head changes every slot whereas the split only changes at most once per epoch).

I suspect the effect sizes of both these phenomena are small. What we have now seems to be performing very well on Holesky, so I'm inclined to just leave it as-is pending further benchmarking.

dapplion: Why not use a try_read here to prevent the deadlock risk?

michaelsproul: There's no deadlock risk. I figure try_read isn't worth it because the alternative is doing state reconstruction. We may as well wait for the split lock, as it will be less work in total

dapplion (approved):

PR #8355, opened 2025-11-03
Gloas process execution payload bid

shane-moore:
## Changes per [spec](https://ethereum.github.io/consensus-specs/specs/gloas/beacon-chain/#new-process_execution_payload_bid):
Added
  - `process_execution_bid`
    - `verify_execution_payload_header_signature`
    - `has_builder_withdrawal_credential`
    - `is_builder_withdrawal_credential`
 
Modified
  - `has_compounding_withdrawal_credential`

## How this works
During block processing, we will no longer be verifying execution payloads but instead execution bids.  These bids were submitted by in-protocol builders, and the winning bid is included in the block by the proposer.

During verification, we perform many checks to ensure the builder who submitted the bid has enough funds to cover their bid and also that the bid is for the current slot and that the bid's parent CL and EL block hashes correspond to those from previous block known in the `BeaconState`.

Additionally, we will add the bid's value to the `state.builder_pending_payments` list.

@ethDreamer
@eserilev

### DIFF
--- a/consensus/state_processing/src/per_block_processing.rs
+++ b/consensus/state_processing/src/per_block_processing.rs
@@ -737,57 +737,61 @@ pub fn process_execution_payload_bid<E: EthSpec, Payload: AbstractExecPayload<E>
         }
     }
 
-    // Verify builder is active and not slashed
+    // Verify builder is active
     block_verify!(
         builder.is_active_at(state.current_epoch()),
         ExecutionPayloadBidInvalid::BuilderNotActive(builder_index).into()
     );
-    block_verify!(
-        !builder.slashed,
-        ExecutionPayloadBidInvalid::BuilderSlashed(builder_index).into()
-    );
 
-    // Check that the builder is active, non-slashed, and has funds to cover the bid
-    let pending_payments = state
-        .builder_pending_payments()?
-        .iter()
-        .filter_map(|payment| {
-            if payment.withdrawal.builder_index == builder_index {
-                Some(payment.withdrawal.amount)
-            } else {
-                None
-            }
-        })
-        .safe_sum()?;
+    // Only perform payment related checks if amount > 0
+    if amount > 0 {
+        // Verify builder is not slashed
+        block_verify!(
+            !builder.slashed,
+            ExecutionPayloadBidInvalid::BuilderSlashed(builder_index).into()
+        );
 
-    let pending_withdrawals = state
-        .builder_pending_withdrawals()?
-        .iter()
-        .filter_map(|withdrawal| {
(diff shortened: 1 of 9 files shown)


consensus/state_processing/src/per_block_processing.rs
```diff
@@ -692,3 +694,157 @@ pub fn get_expected_withdrawals<E: EthSpec>(
...
+    // Verify builder is active and not slashed
+    block_verify!(
+        builder.is_active_at(state.current_epoch()),
+        ExecutionPayloadBidInvalid::BuilderNotActive(builder_index).into()
+    );
+    block_verify!(
+        !builder.slashed,
+        ExecutionPayloadBidInvalid::BuilderSlashed(builder_index).into()
+    );
+
```
ethDreamer, line 749: hmmmm so the spec is written for clarity, not efficiency. Some of this work should be avoided if the amount is zero.

shane-moore: So true! def will keep that more top of mind when translating from spec to a performant code base. updated in commit https://github.com/sigp/lighthouse/pull/8355/commits/5ba3d543de09c2818d9acb55061b77cbbde862a5

consensus/state_processing/src/per_block_processing.rs
```diff
@@ -692,3 +694,161 @@ pub fn get_expected_withdrawals<E: EthSpec>(
...
+    // Verify builder is active
+    block_verify!(
+        builder.is_active_at(state.current_epoch()),
+        ExecutionPayloadBidInvalid::BuilderNotActive(builder_index).into()
+    );
+
+    // Only perform payment related checks if amount > 0
+    if amount > 0 {
+        // Verify builder is not slashed
+        block_verify!(
```
ethDreamer: The spec verifies that the builder is not slashed even on self-builds (aka verifies the proposer is not slashed). This needs to be outside the amount or this could lead to a consensus fault under the right conditions.

shane-moore: ah, I see! makes sense since slashed proposer could technically be selected to produce block but hasn't exited yet. and we'd want to have this `assert` as a guard

also probably important for invalidating blocks with `0` value bids submitted by slashed off-protocol builders. since we may go the route of allowing builder api respond with a `0` value bid and an  `off_protocol_bid_value`, then client can use that `off_protocol_bid_value` during bid selection

resolved in https://github.com/sigp/lighthouse/pull/8355/commits/7db4b92488cb13370e8c294aebf19d53a51040e8

ethDreamer (approved):
LGTM!

PR #8358, opened 2025-11-04
Update proposer-only section in the documentation

chong-he:
Delete the mention about reduce peer count to a safe minimal, as this is no longer true after Fusaka, thanks @jimmygchen for the flags and tips

### DIFF
--- a/book/src/advanced_blobs.md
+++ b/book/src/advanced_blobs.md
@@ -17,7 +17,7 @@ While both `--supernode` and `--semi-supernode` can serve blobs, a supernode wil
 
 Combining `--prune-blobs false` and `--supernode` (or `--semi-supernode`) implies that no data columns will be pruned, and the node will be able to serve blobs since using the flag.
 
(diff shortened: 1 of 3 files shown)


book/src/advanced_blobs.md
```diff
@@ -17,7 +17,7 @@ While both `--supernode` and `--semi-supernode` can serve blobs, a supernode wil
 
 Combining `--prune-blobs false` and `--supernode` (or `--semi-supernode`) implies that no data columns will be pruned, and the node will be able to serve blobs since using the flag.
 
-If you want historical blob data beyond the data availability period (18 days), you can backfill blobs or data columns with the experimental flag `--complete-blobs-backfill`. However, do note that this is an experimental feature and it may cause some issues, e.g., the node may block most of its peers.
+If you want historical blob data beyond the data availability period (18 days), you can backfill blobs or data columns with the experimental flag `--complete-blob-backfill`. However, do note that this is an experimental feature and it only works when the flag is present during a fresh checkpoint sync when the database is initialised. The flag will have no effect if the node is already running (with an existing database). During blobs backfill, the feature may cause some issues, e.g., the node may block most of its peers.
```
michaelsproul:
```suggestion
If you want historical blob data beyond the data availability period (18 days), you can backfill blobs or data columns with the experimental flag `--complete-blob-backfill`. However, do note that this is an experimental feature and it only works when the flag is present during a fresh checkpoint sync when the database is initialised. The flag will have no effect if the node is already running (with an existing database). During blob backfill, the feature may cause some issues, e.g., the node may block most of its peers.
```

michaelsproul (approved):
Looks good, let's merge!

jimmygchen:
@mergify requeue

PR #8364, opened 2025-11-04
Send idontwant on publish

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Send an idontwant after publishing on any topic. Normally, we would only expect idontwants to be useful for fanouts but I don't expect to see any significant improvement after merging https://github.com/sigp/lighthouse/pull/8269 . 

But regardless, I feel like sending idontwants on publish is the sensible default. I don't want to merge this until fulu is activated on mainnet and we can test it on a mainnet like network. I feel that testing this on devnets wouldn't be very representative.

### DIFF
--- a/beacon_node/lighthouse_network/src/config.rs
+++ b/beacon_node/lighthouse_network/src/config.rs
@@ -507,6 +507,7 @@ pub fn gossipsub_config(
         .duplicate_cache_time(duplicate_cache_time)
         .message_id_fn(gossip_message_id)
         .allow_self_origin(true)
+        .idontwant_on_publish(true)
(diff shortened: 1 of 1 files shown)


pawanjay176:
@jxs This fell off the radar. Thanks Daniel for reminding me. Would be good to merge this now imo. What do you think?

dknopik:
I am not too sure about this as-is.

Gossipsub uses the same byte threshold for this as on receive. The difference is that there is some latency between receive and forward due to verification. 

Here, the IDONTWANT is sent immediately before publish. I don't think that makes sense for 1000 byte messages. There is probably a better threshold, so we might need to enable setting two separate thresholds and run some experiments on hoodi to find a good value?

PR #8366, opened 2025-11-05
Fix unaggregated delay metric

hopinheimer:
## Proposed Changes

while working on this #7892 @michaelsproul pointed it might be a good metric to measure the delay from start of the slot instead of the current `slot_duration / 3`, since the attestations duties start before the `1/3rd` mark now with the change in the link PR.


michaelsproul (approved):
LGTM, thanks!

PR #8367, opened 2025-11-05
Downgrade and remove unnecessary logs

jimmygchen:
### Downgrade a non error to `Debug`

I noticed this error on one of our hoodi nodes:

```
Nov 04 05:13:38.892 ERROR Error during data column reconstruction       block_root: 0x4271b9efae7deccec3989bd2418e998b83ce8144210c2b17200abb62b7951190, error: DuplicateFullyImported(0x4271b9efae7deccec3989bd2418e998b83ce8144210c2b17200abb62b7951190)
```

This shouldn't be logged as an error and it's due to a normal race condition, and it doesn't impact the node negatively.

### Remove spammy logs

This logs is filling up the log files quite quickly and it is also something we'd expect during normal operation - getting columns via EL before gossip. We haven't found this debug log to be useful, so I propose we remove it to avoid spamming debug logs.

```
Received already available column sidecar. Ignoring the column sidecar
```

In the process of removing this, I noticed we aren't propagating the validation result, which I think we should so I've added this. The impact should be quite minimal - the message will stay in the gossip memcache for a bit longer but should be evicted in the next heartbeat.

### DIFF
--- a/beacon_node/network/src/network_beacon_processor/mod.rs
+++ b/beacon_node/network/src/network_beacon_processor/mod.rs
@@ -864,6 +864,9 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                     "Reconstruction not required for block"
                 );
             }
+            Err(BlockError::DuplicateFullyImported(_)) => {
+                debug!("Block has already been imported");
+            }
             Err(e) => {
                 error!(
                     %block_root,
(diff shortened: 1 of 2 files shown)


beacon_node/network/src/network_beacon_processor/mod.rs
```diff
@@ -866,6 +864,9 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                     "Reconstruction not required for block"
                 );
             }
+            Err(BlockError::DuplicateFullyImported(_)) => {
+                debug!("Block has already been imported");
```
michaelsproul: Maybe we should mention that this log is reconstruction related?

```suggestion
                debug!("Block already imported in parallel with reconstruction");
```

jimmygchen: Yep good idea!

michaelsproul (approved):

PR #8370, opened 2025-11-05
Download execution engine release binaries instead of building locally on CI

jimmygchen:
## Issue Addressed

The execution engine tests on CI have been failing intermittently and the flakiness have been a pain for CI stability
https://github.com/sigp/lighthouse/actions/runs/19085741237/job/54525434394

This PR attempts to improve this by downloading the release binaries instead of building them locally. This should also improve the time it takes to run this CI job.

If it fails to download binaries from the URL, it would fallback to the existing approach to build the binaries locally.

TODO:
- detect platform and download the correct file or just fallback to building locally if not on CI

### DIFF
--- a/testing/execution_engine_integration/src/build_utils.rs
+++ b/testing/execution_engine_integration/src/build_utils.rs
@@ -1,8 +1,20 @@
 use crate::SUPPRESS_LOGS;
+use serde_json::Value;
 use std::env;
 use std::fs;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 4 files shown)


jimmygchen:
This turned out to be quite a PITA, maybe kurtosis or docker is the way to go.

jimmygchen:
I'm closing this one - I think starting up with docker would be way easier if we want to maintain the current tests.
Also we need to update the fork to use latest. I'll make a new PR.

PR #8377, opened 2025-11-05
Remove `sensitive_url` and import from `crates.io`

macladson:
## Proposed Changes

Use the recently published `sensitive_url` and remove it from Lighthouse


michaelsproul (approved):
Looks good!

PR #8378, opened 2025-11-06
Optimise fork choice attestation dequeueing

michaelsproul:
## Proposed Changes

Optimise dequeuing of attestations in fork choice by avoiding reallocating the queue after every dequeue.

The included benchmarks shows that this takes ~35% off the runtime of `dequeue_attestations` over multiple slots:

```
dequeue_attestations/93750
                        time:   [10.603 ms 10.607 ms 10.613 ms]
                        change: [-34.468% -34.367% -34.274%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  4 (4.00%) high mild
  3 (3.00%) high severe
```

The baseline branch with just the benchmark added to `unstable` can be found here: https://github.com/michaelsproul/lighthouse/tree/dequeue-attestation-baseline-benchmark

## Additional Info

I played around with several `VecDeque`-based implementations. The one with `partition_point` and `rotate_left`/`split_off` is substantially faster than using `pop_front` or `drain`.

The **risk** of this PR is that the allocation for `queued_attestations` is _never_ shrunk/contracted, so in unusual circumstances it could grow very large. We could consider adding a hard cap on the number of items in the queue, as we've also had this queue grow in an unbounded way during times of memory corruption:


consensus/fork_choice/src/fork_choice.rs
```diff
@@ -1509,7 +1519,7 @@ where
     pub fn to_persisted(&self) -> PersistedForkChoice {
         PersistedForkChoice {
             proto_array: self.proto_array().as_ssz_container(),
-            queued_attestations: self.queued_attestations().to_vec(),
+            queued_attestations: self.queued_attestations().iter().cloned().collect(),
```
michaelsproul, line 1526: This is potentially slower than if we implemented `Encode` for `VecDeque`. The conversion from `VecDeque` to `Vec` is non-trivial in most cases where the `VecDeque` doesn't start at index 0:

https://doc.rust-lang.org/std/collections/struct.VecDeque.html#impl-From%3CVecDeque%3CT,+A%3E%3E-for-Vec%3CT,+A%3E

I'll check fork choice persistence times to make sure this isn't having a substantial impact.

consensus/fork_choice/src/fork_choice.rs
```diff
@@ -1469,7 +1479,7 @@ where
         let mut fork_choice = Self {
             fc_store,
             proto_array,
-            queued_attestations: persisted.queued_attestations,
+            queued_attestations: persisted.queued_attestations.into(),
```
michaelsproul, line 1484: This conversion `Vec => VecDeque` is trivial and cheap, but only happens once on startup anyway.

consensus/fork_choice/benches/benches.rs
```diff
@@ -0,0 +1,56 @@
+use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
+use fork_choice::{QueuedAttestation, dequeue_attestations};
+use std::collections::VecDeque;
+use types::{Epoch, Hash256, Slot};
+
+fn all_benches(c: &mut Criterion) {
+    let num_attestations = 1_500_000_usize / 16;
```
michaelsproul, line 7: This is ~2 slots worth of attestations on a 1.5M validator network

michaelsproul:
`dequeue_attestations` is no longer visible in the profiling flamegraph after this change :fire:

consensus/fork_choice/src/fork_choice.rs
```diff
@@ -253,25 +253,35 @@ impl<'a, E: EthSpec> From<IndexedAttestationRef<'a, E>> for QueuedAttestation {
...
+) -> VecDeque<QueuedAttestation> {
+    // Find the position of the first attestation to keep in the queue, or equivalently the number
+    // of attestations to pop from the front of the queue.
+    //
+    // We are safe to use `partition_point` as we know the queue is sorted by ascending slot.
+    // Benchmarks show that `partition_point` is substantially faster (-18%) than using `find`.
+    let to_pop = queued_attestations.partition_point(|a| a.slot < current_slot);
+
+    // Rotate the entries to remove into the *end* of the vec deque.
+    queued_attestations.rotate_left(to_pop);
```
dapplion, line 270: Why not change self.queued_attestations into a map of Slot -> Vec< QueuedAttestation> and when that slot arrives we just remove the entire vec? No need to rotate or split_off

michaelsproul: Good idea, I'll try it and see how the benchmark looks

michaelsproul: It will have the same problem of re-allocating memory for the `Vec`s though.

michaelsproul: It's 4x slower :skull:

```
dequeue_attestations/93750
                        time:   [41.906 ms 41.928 ms 41.953 ms]
Found 8 outliers among 100 measurements (8.00%)
  5 (5.00%) high mild
  3 (3.00%) high severe
```

Impl here: https://github.com/michaelsproul/lighthouse/commit/daf3617d35d0100642c19fbb13c94887e34af724.

One of the problems is that I didn't try optimising out the concatenation of `Vec`s. `dequeue_attestations` could definitely return like a `Vec<Vec<_>>` or something, but I didn't bother refactoring any further, because I think the lack of memory-reuse just makes it worse than the `VecDeque`.

michaelsproul:
Putting this back to `waiting-for-author` so we can address the unbounded memory growth. Going to try the hashmap approach with pre-allocated Vecs

PR #8385, opened 2025-11-09
fix(clap_utils): provide context-specific error messages for network config

DeVikingMark:
Handle None results from Eth2NetworkConfig::constant with context-aware error messages. 

When a user-specified network name is invalid, show the network name in the error. 

This makes debugging easier compared to the generic message that was used for all None cases.


common/clap_utils/src/lib.rs
```diff
@@ -20,18 +20,23 @@ pub const FLAG_HEADER: &str = "Flags";
 /// Try to parse the eth2 network config from the `network`, `testnet-dir` flags in that order.
 /// Returns the default hardcoded testnet if neither flags are set.
 pub fn get_eth2_network_config(cli_args: &ArgMatches) -> Result<Eth2NetworkConfig, String> {
-    let optional_network_config = if cli_args.contains_id("network") {
-        parse_hardcoded_network(cli_args, "network")?
+    let eth2_network_config = if cli_args.contains_id("network") {
+        let network_name = parse_required::<String>(cli_args, "network")?;
+        Eth2NetworkConfig::constant(network_name.as_str())?.ok_or_else(|| {
+            format!(
+                "Invalid network name: '{}'. Use --testnet-dir to specify a custom network.",
```
chong-he, line 27: Upon testing, this error is not reachable, because the clap would have already caught the incorrect network name used. For example, if we run: `lighthouse bn --network test`, the error is:

```
error: invalid value 'test' for '--network <network>'
  [possible values: mainnet, gnosis, chiado, sepolia, holesky, hoodi]
```

which I think is clear enough

chong-he (commented):
I think the changes introduced in this PR is not reachable, and it is in my opinion not worthwhile to make it reachable as the current error shown when an incorrect network name is specified is clear enough.

As such I suggest to close this PR

chong-he:
Closing as per the review comment above

Issue #8106, opened 2025-09-24
Ensure HTTPS links in Cargo.toml


michaelsproul:
## Description

We had an insecure HTTP dependency creep into our Cargo.toml, fixed by:

- https://github.com/sigp/lighthouse/pull/8094

It would be good to run a tool on CI to detect these, as suggested by @hopinheimer.

Gurkirat155:
@michaelsproul  can i take this issue

michaelsproul:
@Gurkirat155 sure! Thank you!

Gurkirat155:
I wasn’t able to finish this one in time, apologies for that. I see it’s already being resolved. I’ll pick up another issue soon.

chong-he:
Closed by #8464

PR #8386, opened 2025-11-09
feat: add CI check for HTTPS links in Cargo.toml

CreeptoGengar:
## Issue Addressed

#8106

## Proposed Changes

Added a CI check to detect insecure HTTP links in Cargo.toml files.

- Created `scripts/ci/check-https-links.sh` to scan all Cargo.toml files for HTTP links
- Integrated the check into the `check-code` job in CI workflow
- Prevented insecure HTTP dependencies from being merged

## Additional Info

_


chong-he:
Thanks for the PR. Could you please rebase to the `unstable` branch?

scripts/ci/check-https-links.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+# Check for insecure HTTP links in Cargo.toml files
```
chong-he, line 3: This file has permission issue when I tested locally. This is also caught in the CI: https://github.com/sigp/lighthouse/actions/runs/19207968175/job/55140383694?pr=8386

You need to make the file executable

scripts/ci/check-https-links.sh
```diff
@@ -0,0 +1,43 @@
...
+# This script ensures all git dependencies use HTTPS instead of HTTP
+#
+# This addresses the security concern raised in:
+# https://github.com/sigp/lighthouse/issues/8106
```
chong-he: This can be removed 
```suggestion
```

scripts/ci/check-https-links.sh
```diff
@@ -0,0 +1,43 @@
...
+        echo "$http_links"
+        found_http_links=true
+    fi
+done
+
+if [ "$found_http_links" = true ]; then
+    echo ""
+    echo "Please replace all HTTP links with HTTPS links in the Cargo.toml files above."
+    echo "For example, change: git = \"http://github.com/...\""
+    echo "                to: git = \"https://github.com/...\""
```
chong-he: This is a pedantic comment, so that the log printed is aligned
```suggestion
    echo "                 to: git = \"https://github.com/...\""
```

.github/workflows/test-suite.yml
```diff
@@ -403,6 +403,8 @@ jobs:
       run:  make arbitrary-fuzz
     - name: Run cargo audit
       run: make audit-CI
+    - name: Check for HTTPS links in Cargo.toml
+      run: ./scripts/ci/check-https-links.sh
```
chong-he: We can put this in `Makefile`, similar to other checks, for example, the mdlint check:

https://github.com/sigp/lighthouse/blob/e3ee7febce64c1b5a85c3ab0be0619571ee92d58/Makefile#L238-L241

and then replace with `make https-links` here. This is easier if we want to run locally, though this may not be run locally frequently

scripts/ci/check-https-links.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
```
chong-he: To be consistent with other script files, we can change to `#! /usr/bin/env bash`
```suggestion
#! /usr/bin/env bash

```

chong-he (commented):
Looks pretty good to me, tested and it's working locally. Left some comments, and also need to rebase to `unstable`

.github/workflows/test-suite.yml
```diff
@@ -337,6 +337,10 @@ jobs:
       run:  make arbitrary-fuzz
     - name: Run cargo audit
       run: make audit-CI
+    - name: Check for HTTPS links in Cargo.toml
+      run: |
+        chmod +x ./scripts/ci/check-https-links.sh
```
chong-he: I don't think we need to run chmod here
You can update the script file permission and push that file, that would be cleaner

chong-he (commented):
The script file is still not executable
Screenshot from 2025-11-15 08-35-07

You need to `chmod` and then push the file

PR #8390, opened 2025-11-10
Fix panic on a rare startup race condition

jimmygchen:
## Issue Addressed

A [discord user reported](https://discord.com/channels/605577013327167508/605577013331361793/1436490679256612884) a panic during startup and it wasn't recurring. This seems like a rare race condition that wasn't possible until custody backfill was introduced, and this queried custody context before it was initialised:

https://github.com/sigp/lighthouse/blob/51ad47f5dbdb5c6fa4064133f623b02c297369ef/beacon_node/network/src/sync/custody_backfill_sync/mod.rs#L262

custody context initialised here:
https://github.com/sigp/lighthouse/blob/43c5e924d74003d9eab0ca7fdf58f1c8b5e5db7a/beacon_node/client/src/builder.rs#L488

I've been thinking about fixing the race condition but thought this may be a simpler fix, however this end up being quite tedious and requiring a lot of changes.

I'm going to look into the alternatives of fixing the race conditions properly and compare the solutions.

NOTE: This currently targets `unstable`, but I can rebase this into the release branch if we want to include this in v8.0.1.

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -7137,25 +7137,30 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
 
     /// Compare columns custodied for `epoch` versus columns custodied for the head of the chain
     /// and return any column indices that are missing.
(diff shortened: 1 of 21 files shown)


jimmygchen:
Closing in favour of #8391

PR #8391, opened 2025-11-10
Fix custody context initialization race condition that caused panic

jimmygchen:
## Issue Addressed

Take 2 of #8390.

Fixes the race condition properly instead of propagating the error. I think this is a better alternative, and doesn't seem to look that bad.

## Proposed Changes

* Lift node id loading or generation from `NetworkService ` startup to the `ClientBuilder`, so that it can be used to compute custody columns for the beacon chain without waiting for Network bootstrap.

I've considered and implemented a few alternatives:
1. passing `node_id` to beacon chain builder and compute columns when creating `CustodyContext`. This approach isn't good for separation of concerns and isn't great for testability
2. passing `ordered_custody_groups` to beacon chain. `CustodyContext` only uses this to compute ordered custody columns, so we might as well lift this logic out, so we don't have to do error handling in `CustodyContext` construction. Less tests to update;.

### DIFF
--- a/beacon_node/beacon_chain/src/custody_context.rs
+++ b/beacon_node/beacon_chain/src/custody_context.rs
@@ -7,7 +7,6 @@ use std::{
     sync::atomic::{AtomicU64, Ordering},
 };
 use tracing::{debug, warn};
-use types::data_column_custody_group::{CustodyIndex, compute_columns_for_custody_group};
 use types::{ChainSpec, ColumnIndex, Epoch, EthSpec, Slot};
 
 /// A delay before making the CGC change effective to the data availability checker.
@@ -504,7 +503,7 @@ impl<E: EthSpec> CustodyContext<E> {
 
         // This is an unnecessary conversion for spec compliance, basically just multiplying by 1.
         let columns_per_custody_group = spec.data_columns_per_group::<E>() as usize;
-        let custody_column_count = columns_per_custody_group * (custody_group_count);
+        let custody_column_count = columns_per_custody_group * custody_group_count;
 
         &self.ordered_custody_column_indices[..custody_column_count]
     }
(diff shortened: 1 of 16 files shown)


consensus/types/src/data_column_custody_group.rs
```diff
@@ -76,6 +76,27 @@ pub fn get_custody_groups_ordered(
     Ok(custody_groups)
 }
 
+/// Returns a deterministically ordered list of custody columns assigned to a node,
+/// preserving the order in which they were computed during iteration.
+///
+/// # Arguments
+/// * `raw_node_id` - 32-byte node identifier
+/// * `spec` - Chain specification containing custody parameters
+pub fn compute_ordered_custody_column_indices<E: EthSpec>(
```
jimmygchen, line 85: Logic moved from `ClientBuilder`.

beacon_node/beacon_chain/src/custody_context.rs
```diff
@@ -623,16 +593,17 @@ mod tests {
...
 
-        let all_custody_groups_ordered = (0..spec.number_of_custody_groups).collect::<Vec<_>>();
-        custody_context
-            .init_ordered_data_columns_from_custody_groups(all_custody_groups_ordered, spec)
-            .expect("should initialise ordered data columns");
         custody_context
     }
 
+    fn generate_custody_column_indices() -> Vec<u64> {
+        (0..E::number_of_columns() as u64).collect::<Vec<_>>()
```
michaelsproul: All these tests use `0..number_of_columns`, which kind of implies the "ordering" is ordering by ascending column. But as I understand it (looking at `get_custody_groups_ordered`) the ordering is actually the ordering the columns are produced in from the `node_id`. Maybe we should use a "more arbitrary" ordering here, to make sure we're not accidentally depending on the ascending ordering?

michaelsproul (approved):
Looks a lot nicer without the `expects`, and the extra plumbing for initialisation doesn't seem too bad. Looks good!

beacon_node/beacon_chain/src/custody_context.rs
jimmygchen: Yeah good idea, I've pushed the fix

michaelsproul: Looks good!

michaelsproul (approved):

PR #8404, opened 2025-11-12
Delete the `block_cache`

michaelsproul:
## Proposed Changes

The `block_cache` was disabled by default for hindering performance. We have no plans to re-enable it, and likely a performant replacement would look somewhat different from what we have now.

This PR deletes the `block_cache` implementation entirely, achieving a slight simplification of our DB code, and also takes aim at the `do_atomically_with_block_and_blobs_cache` method which IMO was overly complicated for little gain.

In the process of implementing this change I also added an optimisation to write the blobs to the database in parallel with the fork choice write. This is safe and gains us a small amount on block import pre-Gloas. I would be happy to revert this change if we deem it not worth it post-Gloas.

## Additional Info

- [x] Deprecate CLI flag (with warning).
- [x] Consider re-adding cache for `DataColumnCustodyInfo`. No need: read is sub 1ms according to Claude's benchmark.
- [x] Rename `do_atomically_with_block_and_blobs_cache` to something more appropriate.
- [x] Work out whether we need a garbage collection/"hard pruning" for dangling blobs/columns.


beacon_node/store/src/hot_cold_store.rs
```diff
@@ -2482,23 +2239,12 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
     /// Fetch custody info from the cache.
     /// If custody info doesn't exist in the cache,
     /// try to fetch from the DB and prime the cache.
+    // FIXME(sproul): re-add cache for this?
```
michaelsproul: @eserilev Did you remember if caching the `DataColumnCustodyInfo` had a noticeable performance benefit? I think it probably helps a bit, but we also don't load it very often.

eserilev: @michaelsproul we don't load it very often. The only places it's used is when responding to `Status` and `DataColumnByRange` RPC requests and during custody backfill. During custody backfill and `DataColumnByRange` RPC requests any performance benefits from caching this value seems negligible to me, since we're already reading and/or writing from/to disk.

I don't know how often we respond to `Status` requests. But without caching we'd now have to read from disk when responding to this request. Doesn't seem like a big deal to me, but we could cache it in a similar way that we cache `AnchorInfo`, `BlobInfo`, etc.

michaelsproul: I'll try measuring it. I suspect LevelDB would cache it in memory for us anyway if we are reading frequently for status messages. Also wouldn't be too hard to cache as you said

PR #8410, opened 2025-11-13
fix: use second_payload for head in execution engine test

Bilogweb3:
Fix copy-paste bug where valid_payload was incorrectly used instead of second_payload when setting head block hash after creating and validating the second payload.

The test validates a second payload built on top of the first, but then mistakenly referenced the old valid_payload when updating the head. This matches the pattern used correctly in the EE B section.


dapplion:
Likely a farming PR

PR #8413, opened 2025-11-14
Do not require blobs from checkpoint servers from Fulu epochs.

jimmygchen:
## Issue Addressed

Addressed this comment here: https://github.com/sigp/lighthouse/issues/6837#issuecomment-3509209465

Lighthouse can only checkpoint sync from a server that can serve blob sidecars, which means they need to be at least custdoying 50% of columns (semi-supernodes)

This PR lifts this constraint, as blob sidecar endpoint is getting deprecated in Fulu, and we plan to fetch the checkpoint data columns from peers (#6837)

### DIFF
--- a/beacon_node/client/src/builder.rs
+++ b/beacon_node/client/src/builder.rs
@@ -345,7 +345,13 @@ where
                     .map_err(|e| format!("Unable to parse weak subj state SSZ: {:?}", e))?;
                 let anchor_block = SignedBeaconBlock::from_ssz_bytes(&anchor_block_bytes, &spec)
                     .map_err(|e| format!("Unable to parse weak subj block SSZ: {:?}", e))?;
-                let anchor_blobs = if anchor_block.message().body().has_blobs() {
+
+                // `BlobSidecar` is no longer used from Fulu onwards (superseded by `DataColumnSidecar`),
+                // which will be fetched via rpc instead (unimplemented).
+                let is_before_fulu = !spec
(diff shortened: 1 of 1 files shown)


chong-he (approved):
Looks good. 

From my understanding the `ClientGenesis::WeakSubjSszBytes` variant is when then data (block, state and blob) is provided, i.e., during manual checkpoint sync. I will put this small update in the PR about updating Lighthouse book

michaelsproul (approved):
Tested manually on Hoodi and it works

I have some hesitation about introducing a gap in the data column DB, but I think maybe it can be worked around using manual custody backfill sync anyway (the POST `/lighthouse/custody/backfill` API). It's just quite inefficient -- we have to wait for block backfill to complete, and then start custody backfill, just to download one slot's worth of missing data columns. I think this will also re-download all the existing columns too, but not sure.

jimmygchen:
> Tested manually on Hoodi and it works
> 
> I have some hesitation about introducing a gap in the data column DB, but I think maybe it can be worked around using manual custody backfill sync anyway (the POST `/lighthouse/custody/backfill` API). It's just quite inefficient -- we have to wait for block backfill to complete, and then start custody backfill, just to download one slot's worth of missing data columns. I think this will also re-download all the existing columns too, but not sure.

I think the current column backfill logic may need a bit of work to support this, it currently just backfill all the way to the start of DA window

PR #8415, opened 2025-11-14
Gloas ptc duties beacon node response

shane-moore:
## Changes
- BN will have to be able to respond to VC requests for ptc duties for current or next epoch, per [beacon api spec](https://github.com/ethereum/beacon-APIs/pulls)'s `POST /eth/v1/validator/duties/ptc/{epoch}`
- This PR uses vanilla `get_ptc` to calculate ptc members for a given slot, which is heavy computation when calculating ptc duties for all validators for an entire epoch. But this should be fine for `devnet-0`
- Future optimizations will be to add in my `ptc_cache` [PR](https://github.com/shane-moore/lighthouse/pull/10), which allows us cache ptc committees for current/previous/next epoch.  Additionally, we'll want out to build out a cache similar to `shuffling_cache` that allows us to cache instances of the `ptc_cache` depending upon the current fork choice head (shuffling Id) so that we don't have to recompute `ptc_cache` when fork choice switches heads and then back again, for example


dapplion (approved):
Looks good! The http_api tests are not running for gloas in this branch, but the new route is not used in production.

PR #8416, opened 2025-11-15
Consistent logic to select range sync start_slot

dapplion:
## Issue Addressed

From https://github.com/sigp/lighthouse/pull/8382 I noted that the logic in range sync to decide _where_ to start syncing from is scattered. When manual finalization is triggered, sync will fetch useless epochs.

## Proposed Changes

This is a somewhat sensitive change that I don't want to have buried in https://github.com/sigp/lighthouse/pull/8382. Since manual finalization is an avaialble feature in current stable I think it's okay to fix separately to https://github.com/sigp/lighthouse/pull/8382. 

This PR merges the logic to decide if a peer:
- is advanced or not
- should trigger head or finalized sync

In a single function. Plus, the variant `PeerSyncType::Advanced` includes which slot sync should start from.

### DIFF
--- a/beacon_node/network/src/sync/peer_sync_info.rs
+++ b/beacon_node/network/src/sync/peer_sync_info.rs
@@ -1,14 +1,16 @@
 use super::manager::SLOT_IMPORT_TOLERANCE;
+use crate::status::ToStatusMessage;
 use beacon_chain::{BeaconChain, BeaconChainTypes};
 use lighthouse_network::{SyncInfo, SyncStatus as PeerSyncStatus};
 use std::cmp::Ordering;
+use types::{Epoch, EthSpec, Hash256, Slot};
 
 /// The type of peer relative to our current state.
 pub enum PeerSyncType {
     /// The peer is on our chain and is fully synced with respect to our chain.
     FullySynced,
     /// The peer has a greater knowledge of the chain than us that warrants a full sync.
-    Advanced,
+    Advanced(PeerSyncTypeAdvanced),
     /// A peer is behind in the sync and not useful to us for downloading blocks.
     Behind,
 }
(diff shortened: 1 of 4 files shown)


beacon_node/network/src/sync/peer_sync_info.rs
```diff
@@ -18,13 +20,52 @@ impl PeerSyncType {
...
             PeerSyncType::Behind => PeerSyncStatus::Behind { info: info.clone() },
-            PeerSyncType::Advanced => PeerSyncStatus::Advanced { info: info.clone() },
+            PeerSyncType::Advanced(_) => PeerSyncStatus::Advanced { info: info.clone() },
+        }
+    }
+}
+
+pub enum PeerSyncTypeAdvanced {
+    Finalized {
+        target_slot: Slot,
```
michaelsproul, line 30: Needs comment

beacon_node/network/src/sync/peer_sync_info.rs
```diff
@@ -18,13 +20,52 @@ impl PeerSyncType {
...
+        target_root: Hash256,
+        start_epoch: Epoch,
```
michaelsproul, line 32: Should this be per-peer?

beacon_node/network/src/sync/peer_sync_info.rs
```diff
@@ -18,13 +20,52 @@ impl PeerSyncType {
...
+    },
+    Head {
+        target_slot: Slot,
+        target_root: Hash256,
+        start_epoch: Epoch,
+    },
+}
+
+#[derive(Clone)]
+pub(crate) struct LocalSyncInfo {
```
michaelsproul, line 42: Needs a comment

beacon_node/network/src/sync/peer_sync_info.rs
```diff
@@ -18,13 +20,52 @@ impl PeerSyncType {
...
+    pub finalized_epoch: Epoch,
+    pub local_irreversible_epoch: Epoch,
+}
+
+impl LocalSyncInfo {
+    pub fn new<T: BeaconChainTypes>(chain: &BeaconChain<T>) -> Self {
+        let status = chain.status_message();
+        // Max with the store in case the node has triggered manual finalization
+        let local_irreversible_epoch = std::cmp::max(
+            chain.head().finalized_checkpoint().epoch,
```
michaelsproul, line 53: May as well use `status.finalized_epoch()` here to avoid potential inconsistencies. It's also likely to be marginally quicker

beacon_node/network/src/sync/peer_sync_info.rs
```diff
@@ -63,24 +108,55 @@ pub fn remote_sync_type<T: BeaconChainTypes>(
...
-                || chain.block_is_known_to_fork_choice(&remote.head_root)
-            {
-                // This peer is near enough to us to be considered synced, or
-                // we have already synced up to this peer's head
+            if chain.block_is_known_to_fork_choice(&remote.head_root) {
+                // We have already synced up to this peer's head
                 PeerSyncType::FullySynced
             } else {
-                PeerSyncType::Advanced
+                let finality_advanced = remote.finalized_epoch > local.finalized_epoch + 1;
```
michaelsproul, line 127: A comment or a const for this `+ 1` would be good I think

michaelsproul:
Changing this to waiting-on-author, I think you said last week you were planning to make some more changes

PR #8417, opened 2025-11-15
Reimport the checkpoint sync block

dapplion:
## Issue Addressed

We want to not require checkpoint sync starts to include the required custody data columns, and instead fetch them from p2p.

Closes https://github.com/sigp/lighthouse/issues/6837

## Proposed Changes

The checkpoint sync slot can:
1. Be the first slot in the epoch, such that the epoch of the block == the start checkpoint epoch
2. Be in an epoch prior to the start checkpoint epoch

In both cases backfill sync already fetches that epoch worth of blocks with current code. This PR modifies the backfill import filter function to allow to re-importing the oldest block slot in the DB.

I feel this solution is sufficient unless I'm missing something. ~~I have not tested this yet!~~ Michael has tested this and it works.

### DIFF
--- a/beacon_node/beacon_chain/src/block_verification.rs
+++ b/beacon_node/beacon_chain/src/block_verification.rs
@@ -1685,20 +1685,14 @@ fn check_block_against_finalized_slot<T: BeaconChainTypes>(
     //
     // Fork choice has the most up-to-date view of finalization and there's no point importing a
     // block which conflicts with the fork-choice view of finalization.
-    let finalized_checkpoint = chain.canonical_head.cached_head().finalized_checkpoint();
-    let finalized_slot = finalized_checkpoint
+    let finalized_slot = chain
+        .canonical_head
+        .cached_head()
+        .finalized_checkpoint()
         .epoch
         .start_slot(T::EthSpec::slots_per_epoch());
 
-    let would_revert_finalized_slot = if block.slot() < finalized_slot {
-        true
-    } else if block.slot() == finalized_slot {
-        block_root != finalized_checkpoint.root
-    } else {
-        false
-    };
-    // Allow to re-import the finalized block if it's the same
-    if would_revert_finalized_slot {
+    if block.slot() <= finalized_slot {
         chain.pre_finalization_block_rejected(block_root);
         Err(BlockError::WouldRevertFinalizedSlot {
             block_slot: block.slot(),
--- a/beacon_node/beacon_chain/src/block_verification.rs
+++ b/beacon_node/beacon_chain/src/block_verification.rs
@@ -1685,14 +1685,20 @@ fn check_block_against_finalized_slot<T: BeaconChainTypes>(
     //
     // Fork choice has the most up-to-date view of finalization and there's no point importing a
     // block which conflicts with the fork-choice view of finalization.
-    let finalized_slot = chain
-        .canonical_head
-        .cached_head()
-        .finalized_checkpoint()
+    let finalized_checkpoint = chain.canonical_head.cached_head().finalized_checkpoint();
+    let finalized_slot = finalized_checkpoint
         .epoch
         .start_slot(T::EthSpec::slots_per_epoch());
 
-    if block.slot() <= finalized_slot {
+    let would_revert_finalized_slot = if block.slot() < finalized_slot {
+        true
+    } else if block.slot() == finalized_slot {
+        block_root != finalized_checkpoint.root
+    } else {
+        false
+    };
+    // Allow to re-import the finalized block if it's the same
+    if would_revert_finalized_slot {
         chain.pre_finalization_block_rejected(block_root);
         Err(BlockError::WouldRevertFinalizedSlot {
             block_slot: block.slot(),
--- a/beacon_node/beacon_chain/src/historical_blocks.rs
+++ b/beacon_node/beacon_chain/src/historical_blocks.rs
@@ -75,12 +75,13 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         let blob_info = self.store.get_blob_info();
         let data_column_info = self.store.get_data_column_info();
 
-        // Take all blocks with slots less than or equal to the oldest block slot.
+        // Take all blocks with slots less than the oldest block slot, or equal to the anchor slot.
         //
         // This allows for reimport of the blobs/columns for the finalized block after checkpoint
         // sync.
         let num_relevant = blocks.partition_point(|available_block| {
-            available_block.block().slot() <= anchor_info.oldest_block_slot
+            available_block.block().slot() < anchor_info.oldest_block_slot
+                || available_block.block().slot() == anchor_info.anchor_slot
         });
 
         let total_blocks = blocks.len();
@@ -114,7 +115,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         for available_block in blocks_to_import.into_iter().rev() {
             let (block_root, block, block_data) = available_block.deconstruct();
 
-            if block.slot() == anchor_info.oldest_block_slot {
+            if block.slot() == anchor_info.anchor_slot {
                 // When reimporting, verify that this is actually the same block (same block root).
                 let oldest_block_root = self
                     .block_root_at_slot(block.slot(), WhenSlotSkipped::None)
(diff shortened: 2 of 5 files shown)


michaelsproul:
Retargeting at release-v8.0 so we can include this in the v8.0.1 release and prevent gappy DBs

beacon_node/beacon_chain/src/historical_blocks.rs
```diff
@@ -198,7 +222,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             .ok_or(HistoricalBlockError::IndexOutOfBounds)?
             .iter()
             .map(|block| block.parent_root())
-            .chain(iter::once(anchor_info.oldest_block_parent));
+            .chain(iter::once(last_block_root));
```
michaelsproul, line 228: This is required for signature verification to pass when reimporting the anchor block. Without this, we try to use `oldest_block_parent` for `oldest_block` itself, which fails signature verification. Some of the weak subjectivity tests fail without this change.

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -1685,14 +1685,20 @@ fn check_block_against_finalized_slot<T: BeaconChainTypes>(
...
         .start_slot(T::EthSpec::slots_per_epoch());
 
-    if block.slot() <= finalized_slot {
+    let would_revert_finalized_slot = if block.slot() < finalized_slot {
+        true
+    } else if block.slot() == finalized_slot {
+        block_root != finalized_checkpoint.root
+    } else {
+        false
+    };
```
michaelsproul: I don't think this change actually does anything, because re-importing the finalized block in forwards sync will result in a `DuplicateFullyImported` status and no data being written.

Despite this, when syncing from a checkpoint block on an epoch boundary (not a skip), backfill has us covered:

> Nov 18 14:39:30.480 DEBUG Re-importing historic block                   block_root: 0xdbf8afdfbc3be1c714bcbc451f9569c14bb68e6bd6a85dea820bf6d2758f66cd, slot: 1768544
> Nov 18 14:39:30.481 DEBUG Writing data columns to store                 block_root: 0xdbf8afdfbc3be1c714bcbc451f9569c14bb68e6bd6a85dea820bf6d2758f66cd, count: 64

Tested on Hoodi. And I confirmed the columns for that block are available on the HTTP API.

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -770,8 +770,9 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
             );
         }
 
+        let imported_blocks = available_blocks.len();
```
michaelsproul: I guess we are making this change because we don't actually "import" all of the blocks in `blocks_to_import` now. But I don't actually think it's that misleading to say we do, so we could probably revert this?

Is that why you changed this @dapplion?

beacon_node/beacon_chain/src/block_verification.rs
michaelsproul: Reverted here:

- b46df5536b5f7f55fee5f510c0c7ef659711a031

beacon_node/network/src/network_beacon_processor/sync_methods.rs
michaelsproul: Reverted here:

- e1adef7efef751ed5286dbd59c9cabf72e4b273e

michaelsproul:
The only downside to this change that I can see is that we end up reimporting blocks which do not need importing:

> Nov 17 08:06:55.062 INFO  Lighthouse started                            version: "Lighthouse/v8.0.0-0341fd6"
Nov 18 14:36:32.830 INFO  Lighthouse started                            version: "Lighthouse/v8.0.0-226609c"
Nov 18 14:39:30.480 DEBUG Re-importing historic block                   block_root: 0xdbf8afdfbc3be1c714bcbc451f9569c14bb68e6bd6a85dea820bf6d2758f66cd, slot: 1768544
Nov 18 14:57:30.481 DEBUG Re-importing historic block                   block_root: 0xf1f00e894410b571727b00b79bce5b2c1c953773ff7ff66373cc764b7b1e1760, slot: 1760896
Nov 18 14:57:42.476 DEBUG Re-importing historic block                   block_root: 0xf1f00e894410b571727b00b79bce5b2c1c953773ff7ff66373cc764b7b1e1760, slot: 1760896
Nov 18 14:57:54.469 DEBUG Re-importing historic block                   block_root: 0xf1f00e894410b571727b00b79bce5b2c1c953773ff7ff66373cc764b7b1e1760, slot: 1760896
Nov 18 14:58:54.475 DEBUG Re-importing historic block                   block_root: 0x3e4bf54454eba9ee1c3210bb6266aad08277b1c95ce5b21d55fe19327ceddd51, slot: 1760576
Nov 18 15:25:05.419 INFO  Lighthouse started                            version: "Lighthouse/v8.0.0-e1adef7"
Nov 18 15:28:06.410 DEBUG Re-importing historic block                   block_root: 0xd828ed672bc8190c602562c376d0649e8c254d7eea786cb32cbdd51ac0332fb3, slot: 1768800
Nov 18 15:31:18.547 DEBUG Re-importing historic block                   block_root: 0x1c6c4e6c96f2e86206ce3ee955527163d2b567a4b092d955b82371a9a57feaa9, slot: 1767424

These logs are from two checkpoint syncs on Hoodi. We can see backfill reimporting the first anchor block at slot 1768544, and then continuing to re-import epoch boundary blocks as it backfills further. It even reimported slot 1760896 three times.

The second sync was from slot 1768800 which we see getting reimported, and then again backfill reimports another block 1767424.

I think this is probably fine for now, it might just have a small impact on disk I/O. It doesn't seem we are reimporting on every backfill batch, it must be something related to how batches are handled and what we get from peers.

This is not straight-forward to fix because we don't have easy access to the _slot_ of the anchor block. The `AnchorInfo` includes the `anchor_slot` which is the slot of the checkpoint _state_. I.e. in the case of skipped slots, it is ahead of the checkpoint block.

michaelsproul:
Confirmed working in the skipped slot case as well, just needed to fix an unrelated bug in unaligned checkpoint sync (commits: 173b71905e0e04b7beb8964296ca0bec0519da40 https://github.com/sigp/lighthouse/pull/8417/commits/ebf151f34328d1c22bc94cb835c69214de742792). IDK why this wasn't caught by the unaligned checkpoint sync test, I've opened an issue for that here: https://github.com/sigp/lighthouse/issues/8426.

Logs:

> Nov 18 15:55:48.713 INFO  Lighthouse started                            version: "Lighthouse/v8.0.0-173b719"
Nov 18 16:00:58.148 DEBUG Re-importing historic block                   block_root: 0xa8d0dc0b55435377bc8d2739dd7a7870811f8bce0435548f1c01fc8e8e1e8bfb, slot: 1768670

This was using `--checkpoint-state state_1768672.ssz --checkpoint-block block_1768670.ssz` on Hoodi.

michaelsproul (approved):
Happy to merge whenever

michaelsproul:
The unaligned checkpoint sync fix broke some tests, will fix.

beacon_node/beacon_chain/src/historical_blocks.rs
jimmygchen: ah yeah this makes sense

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -649,7 +649,7 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
                 .inspect(|cache| cache.lock().put_block(*block_root, full_block.clone()));
 
             DatabaseBlock::Full(full_block)
-        } else if !self.config.prune_payloads {
+        } else if !self.config.prune_payloads || *block_root == split.block_root {
```
michaelsproul, line 652: This is annoyingly subtle, we need the condition here so that we don't break in the case where the finalized payload is intentionally pruned (which is what happens when prune_payloads runs).

jimmygchen: Yeah I agree - can we add your comment in the code too? would be useful when we look at this again.

michaelsproul: Added a comment here:

- bfa481e304e2f05aa40fe0de177fbdeac4312d7b

michaelsproul:
Re-tested again on latest commit, for both aligned+unaligned cases. All working well.

beacon_node/store/src/hot_cold_store.rs
michaelsproul: Opened an issue here too to track simplifying this:

- https://github.com/sigp/lighthouse/issues/8431

jimmygchen (approved):
Looks good to me 🎉

michaelsproul:
@mergify requeue

PR #8419, opened 2025-11-17
Fix tracing span for execution payload verif

michaelsproul:
## Proposed Changes

Fix the span on execution payload verification (newPayload), by creating a new span rather than using the parent span. Using the parent span was incorrectly associating the time spent verifying the payload with `from_signature_verified_components`.


jimmygchen (approved):
LGTM!

michaelsproul:
@mergify queue

PR #8424, opened 2025-11-17
Convert RpcBlock to an enum that indicates availability

eserilev:
## RpcBlock

Convert `RpcBlock` into an enum with two variants: 

### `RpcBlock::FullyAvailable`

Indicates a fully available RPC block. An RpcBlock of this variant is guaranteed to have all blobs or the columns its required to custody. In the future we could move KZG verification as part of `FullyAvailable` construction validation logic so that we also have type safety guarntees that KZG verification for the block data is also valid

### `RpcBlock::BlockOnly`

Indicates a maybe available block. No blob/column data has been fetched and we don't check if the block is expected to have any blob data. This variant is mostly used in lookup sync.

When constructing an `RpcBlock` the arguments provided decide wether its a `FullyAvailable` or `BlocksOnly` variant

``` rust
pub fn new<T>(
        block_data: Option<AvailableBlockData<E>>,
        da_checker: Arc<DataAvailabilityChecker<T>>,
        spec: Arc<ChainSpec>,
    ) -> Result<Self, AvailabilityCheckError>
```

if `None` is passed in for `block_data` this indicates a `BlocksOnly` variant. If a `Some` value is passed in it will be treated as a `FullyAvailable` variant. See below for the various `block_data` variants:

### `AvailableBlockData`

``` rust
pub enum AvailableBlockData<E: EthSpec> {
    /// Block is pre-Deneb or has zero blobs
    NoData,
    /// Block is post-Deneb, pre-PeerDAS and has more than zero blobs
    Blobs(BlobSidecarList<E>),
    /// Block is post-PeerDAS and has more than zero blobs
    DataColumns(DataColumnSidecarList<E>),
}
```

When constructing the different variants of `AvailableBlockData` the following checks are made
- `AvailableBlockData::NoData`: no blobs are expected for the associated block
- `AvailableBlockData::Blobs`: all expected blobs are present and the block is post-Deneb and pre-PeerDAS
- `AvailableBlockData::DataColumns`: all expected data columns for the nodes custody requirements are present


eserilev:
need to fix a few tests but the general idea is here

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -759,14 +701,73 @@ pub enum AvailableBlockData<E: EthSpec> {
     DataColumns(DataColumnSidecarList<E>),
 }
 
+impl<E: EthSpec> AvailableBlockData<E> {
+    pub fn new(
```
ethDreamer: I think in this instance it would be cleaner to split this into two functions `new_with_blobs` and `new_with_columns` or something like that.

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -786,6 +787,69 @@ impl<E: EthSpec> AvailableBlock<E> {
...
+        block_data: AvailableBlockData<T::EthSpec>,
+        da_checker: Arc<DataAvailabilityChecker<T>>,
+        spec: Arc<ChainSpec>,
+    ) -> Result<Self, AvailabilityCheckError>
+    where
+        T: BeaconChainTypes<EthSpec = E>,
+    {
+        // check blob lengths match
+        // POSSIBLY - verify kzg - but probably not
+        // but this variant *should* ensure the data IS available
```
ethDreamer:
```suggestion
        // Ensure the block is actually available
```

beacon_node/beacon_chain/src/data_availability_checker.rs
eserilev: [8cadad9](https://github.com/sigp/lighthouse/pull/8424/commits/8cadad9cfe6e479657d6cb2b04ce49c6aec27233)

beacon_node/beacon_chain/src/data_availability_checker.rs
eserilev: [b5a534e](https://github.com/sigp/lighthouse/pull/8424/commits/b5a534e48461b9b5005bb252cdde42ffefc84221)

beacon_node/beacon_chain/src/data_availability_checker.rs
eserilev: and this [ed948e9](https://github.com/sigp/lighthouse/pull/8424/commits/ed948e96f836aab6f86b5681ac238835ec153853)

beacon_node/beacon_chain/src/block_verification_types.rs
```diff
@@ -1,174 +1,95 @@
...
-    }
-
-    /// Constructs a new `BlockAndBlobs` variant after making consistency
-    /// checks between the provided blocks and blobs. This struct makes no
-    /// guarantees about whether blobs should be present, only that they are
-    /// consistent with the block. An empty list passed in for `blobs` is
-    /// viewed the same as `None` passed in.
-    pub fn new(
-        block_root: Option<Hash256>,
+    pub fn new<T>(
```
ethDreamer, line 93: Add a comment above specifying when to use Some(block_data) or None

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -749,7 +691,7 @@ async fn availability_cache_maintenance_service<T: BeaconChainTypes>(
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub enum AvailableBlockData<E: EthSpec> {
```
ethDreamer, line 654: future refactor - move this to `block_verification_types`

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -759,31 +701,130 @@ pub enum AvailableBlockData<E: EthSpec> {
...
+
+    pub fn data_columns_len(&self) -> usize {
+        if let Some(data_columns) = self.data_columns() {
+            data_columns.len()
+        } else {
+            0
+        }
+    }
+}
+
```
ethDreamer, line 712: future refactor - move to block_verification_types

beacon_node/beacon_chain/tests/block_verification.rs
```diff
@@ -1868,3 +1971,198 @@ async fn import_execution_pending_block<T: BeaconChainTypes>(
         }
     }
 }
+
+// Test that `signature_verify_chain_segment` correctly handles a mix of `FullyAvailable`
+// and `BlockOnly` RpcBlocks. This situation should not happen in production
+// but this test may help catch potential future regressions.
+#[tokio::test]
```
ethDreamer, line 1980: flip this test

beacon_node/beacon_chain/tests/block_verification.rs
```diff
@@ -1868,3 +1971,198 @@ async fn import_execution_pending_block<T: BeaconChainTypes>(
...
+        chain_segment.len(),
+        "All blocks should be verified, not dropped"
+    );
+
+    // Verify block roots match
+    for (i, (expected_root, _)) in chain_segment.iter().enumerate() {
+        assert_eq!(verified[i].block_root(), *expected_root);
+    }
+}
+
```
ethDreamer, line 2016: test for successfully importing blocks without columns (that need them) when past the data availability boundary

eserilev: [534cfad](https://github.com/sigp/lighthouse/pull/8424/commits/534cfad3432d75fb2fb914de0d2fb71e27ff1adf)

beacon_node/beacon_chain/tests/block_verification.rs
eserilev: test is flipped

beacon_node/beacon_chain/src/data_availability_checker.rs
eserilev: added TODO

beacon_node/beacon_chain/src/data_availability_checker.rs
eserilev: added TODO

beacon_node/beacon_chain/src/block_verification_types.rs
eserilev: added comment

pawanjay176: sorry where was this added? can't find it

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -468,49 +438,21 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
...
-                    MaybeAvailableBlock::Available(AvailableBlock {
-                        block_root,
-                        block,
-                        blob_data: AvailableBlockData::DataColumns(
-                            data_columns.into_iter().map(|d| d.into_inner()).collect(),
-                        ),
-                        blobs_available_timestamp: None,
-                        spec: self.spec.clone(),
-                    })
+        for available_block in available_blocks {
```
pawanjay176, line 395: maybe we can move this to before the kzg verification to avoid expensive verification if the batch doesn't have required data? not super necessary though

beacon_node/network/src/network_beacon_processor/sync_methods.rs
```diff
@@ -722,18 +721,27 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
         downloaded_blocks: Vec<RpcBlock<T::EthSpec>>,
     ) -> (usize, Result<(), ChainSegmentFailed>) {
         let total_blocks = downloaded_blocks.len();
-        let available_blocks = match self
+        let mut available_blocks = vec![];
+
+        for downloaded_block in downloaded_blocks {
+            match downloaded_block {
+                RpcBlock::FullyAvailable(available_block) => available_blocks.push(available_block),
+                RpcBlock::BlockOnly { .. } => return (
```
pawanjay176, line 727: curious if we could just pass a vec of `AvailableBlock` to range/backfill syncs. Might be more work that we can do in another PR, but just wanted to check if there's a reason why we can't do it this way.

pawanjay176 (approved):
Really clean! Thanks for fixing this up.
Just minor nits and a question

beacon_node/network/src/network_beacon_processor/sync_methods.rs
eserilev: I started exploring that but the diff was too large to include in this PR. We can definitely explore this further in a future PR, I just added some notes about that to this issue here https://github.com/sigp/lighthouse/issues/8633

beacon_node/beacon_chain/src/data_availability_checker.rs
eserilev: [ab54242](https://github.com/sigp/lighthouse/pull/8424/commits/ab542426b35900d3aecea5d3b24fcb4411e2266c)

beacon_node/beacon_chain/src/block_verification_types.rs
eserilev: whoops comments are now added: [ab54242](https://github.com/sigp/lighthouse/pull/8424/commits/ab542426b35900d3aecea5d3b24fcb4411e2266c)

pawanjay176 (approved):

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -646,23 +646,41 @@ pub fn signature_verify_chain_segment<T: BeaconChainTypes>(
...
+            let consensus_context =
+                ConsensusContext::new(block.slot()).set_current_block_root(block_root);
+            match block {
+                RpcBlock::FullyAvailable(available_block) => SignatureVerifiedBlock {
+                    block: MaybeAvailableBlock::Available(available_block),
+                    block_root,
+                    parent: None,
+                    consensus_context,
+                },
+                RpcBlock::BlockOnly { block, block_root } => SignatureVerifiedBlock {
```
jimmygchen: Isn't this unreachable given that the above collect of `Result` would have returned an error if it sees any `BlockOnly` variants in the chain segment?

https://github.com/sigp/lighthouse/pull/8424/changes#diff-040e5811a1138cc99780af10517d4288761762d1bc610c719390db564eac524cR653-R660

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -759,31 +702,139 @@ pub enum AvailableBlockData<E: EthSpec> {
...
+                } else if blobs_required {
+                    return Err(AvailabilityCheckError::MissingBlobs);
+                }
+            }
+            AvailableBlockData::Blobs(blobs) => {
+                if !blobs_required {
+                    return Err(AvailabilityCheckError::InvalidAvailableBlockData);
+                }
+
+                if block.num_expected_blobs() != blobs.len() {
```
jimmygchen: We previously check that the commitments matches the block commitments,do we no longer do that?

eserilev: we validate commitments against the block during kzg verification (in `validate_data_columns` for example)

do you think we should be making these checks here instead? mark and I talked about maybe eventually moving kzg verification into here in a future refactor

beacon_node/beacon_chain/src/block_verification.rs
eserilev: yeah you're right, I cleaned this up a bit here: [a5cf26b](https://github.com/sigp/lighthouse/pull/8424/commits/a5cf26b1c8a92acad4f4f763d73f292b17cd1126)

beacon_node/beacon_chain/src/data_availability_checker.rs
pawanjay176: I think Jimmy's right.
If we don't check this, then we could potentially be matching a block with some random columns/blobs that don't actually belong to the block even though it kzg verifies correctly.

pawanjay176: actually turns out we effectively do this by verifying the inclusion proof here
https://github.com/sigp/lighthouse/blob/39c542a37b30284a88f8d631d3cbe0b08c21eb06/beacon_node/network/src/sync/network_context/requests/data_columns_by_range.rs#L34-L35

and then couple the blocks and columns by root later.

However, I think we should still have this check in the construction as the sync logic keeps changing all the time.

eserilev: ah i see what you guys mean

I've readded the commitment check here: 
ce653a2

beacon_node/beacon_chain/src/block_verification_types.rs
```diff
@@ -1,174 +1,117 @@
...
+///    or a post-Deneb block that doesn't require blobs/columns. Hence, it is fully self contained w.r.t
+///    verification. i.e. this block has all the required data to get verified and imported into fork choice.
 ///
-/// 2. `Block`: This can be a fully available pre-deneb block **or** a post-deneb block that may or may
-///    not require blobs to be considered fully available.
+/// 2. `BlockOnly`: This is a post-deneb block that may or may not require blobs to be considered fully available.
 ///
 /// Note: We make a distinction over blocks received over gossip because
 /// in a post-deneb world, the blobs corresponding to a given block that are received
 /// over rpc do not contain the proposer signature for dos resistance.
```
jimmygchen: Just noticed this comment is outdated and can be deleted i think?

beacon_node/beacon_chain/src/data_availability_checker/error.rs
```diff
@@ -22,6 +22,8 @@ pub enum Error {
     BlockReplayError(state_processing::BlockReplayError),
     RebuildingStateCaches(BeaconStateError),
     SlotClockError,
+    InvalidFork,
```
jimmygchen: I can't find where `InvalidFork` is used, is this needed?

eserilev: thanks for catching this, I removed it

beacon_node/beacon_chain/src/block_verification_types.rs
eserilev: deleted the comment

eserilev:
had a working sesh with @ethDreamer and @dapplion today and we think a better direction to go would be to completely separate the types between lookup sync and range sync. This PR gets us some of the way there, so one option is to merge this now and make the changes in a subsequent PR. Or just make the change in this PR. either way I'll start working on those changes Monday

eserilev:
I think this gets us to a nice intermediary step where we can eventually separate the lookup sync and range sync types completely. But this change makes gloas range sync a bit easier to implement so I think we should go ahead with merging this

beacon_node/beacon_chain/src/data_availability_checker.rs
jimmygchen: Seems a bit odd this is after blob kzg blob verification but before column verification. Would be more consistent if we do the check at the top, right? 

We can also refactor this to iterate once, so we don't check `data_columns_required_for_block` twice (i.e. fail fast) and should look a bit cleaner too, something like

```rust

let mut all_blobs = Vec::new();
let mut all_data_columns = Vec::new();

for available_block in available_blocks {
    let blobs_required = self.blobs_required_for_block(&available_block.block);
    let columns_required = self.data_columns_required_for_block(&available_block.block);

    match available_block.data() {
        AvailableBlockData::NoData if columns_required => {
            return Err(AvailabilityCheckError::MissingCustodyColumns);
        }
        AvailableBlockData::NoData if blobs_required => {
            return Err(AvailabilityCheckError::MissingBlobs);
        }
        AvailableBlockData::Blobs(blobs) if blobs_required => {
            all_blobs.extend(blobs.iter());
        }
        AvailableBlockData::DataColumns(cols) if columns_required => {
            all_data_columns.extend(cols.iter());
        }
        _ => { // internal error }
    }
}
```

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -1297,16 +1307,28 @@ impl<T: BeaconChainTypes> IntoExecutionPendingBlock<T> for RpcBlock<T::EthSpec>
...
-            .verify_kzg_for_rpc_block(self.clone())
-            .map_err(|e| {
-                BlockSlashInfo::SignatureNotChecked(
-                    self.signed_block_header(),
-                    BlockError::AvailabilityCheck(e),
-                )
-            })?;
-        SignatureVerifiedBlock::check_slashable(maybe_available, block_root, chain)?
+
+        let maybe_available_block = match &self {
```
jimmygchen, line 1311: nice, much cleaner

beacon_node/beacon_chain/src/block_verification_types.rs
```diff
@@ -1,204 +1,151 @@
...
-///    not require blobs to be considered fully available.
+/// 1. `FullyAvailable`: A fully available block. This can either be a pre-deneb block, a
+///    post-Deneb block with blobs, a post-Fulu block with the columns the node is required to custody,
+///    or a post-Deneb block that doesn't require blobs/columns. Hence, it is fully self contained w.r.t
+///    verification. i.e. this block has all the required data to get verified and imported into fork choice.
 ///
-/// Note: We make a distinction over blocks received over gossip because
-/// in a post-deneb world, the blobs corresponding to a given block that are received
-/// over rpc do not contain the proposer signature for dos resistance.
+/// 2. `BlockOnly`: This is a post-deneb block that may or may not require blobs to be considered fully available.
```
jimmygchen: This is slightly confusing for me - to clarify,  in the case of a post-deneb block that does not require blobs, it is `FullyAvailable` right?

And `BlockOnly` is only for post-deneb block that doesn't have the required blobs?

beacon_node/beacon_chain/src/data_availability_checker.rs
jimmygchen: oh we've already done the same check in `AvailableBlock::new`, so i'ts guranteed to have the required blobs/ data column, so we can remove this entire section right?

and the filter map logic could just be (extra filter removed)
```rust
let all_data_columns = available_blocks
            .iter()
            // this clone is cheap as it's cloning an Arc
            .filter_map(|available_block| available_block.blob_data.data_columns())
            .flatten()
            .collect::<Vec<_>>();
```

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -366,83 +366,54 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
...
+        match available_block.data() {
+            AvailableBlockData::NoData => {
+                if block_data_required {
+                    if available_block.block.fork_name_unchecked().fulu_enabled() {
+                        return Err(AvailabilityCheckError::MissingCustodyColumns);
+                    } else {
+                        return Err(AvailabilityCheckError::MissingBlobs);
+                    }
+                }
+            }
```
jimmygchen: This check is also redundant right? given that `AvailabilityBlock::new` covers it

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -366,83 +366,54 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
...
-        blocks: Vec<RpcBlock<T::EthSpec>>,
-    ) -> Result<Vec<MaybeAvailableBlock<T::EthSpec>>, AvailabilityCheckError> {
-        let mut results = Vec::with_capacity(blocks.len());
-        let all_blobs = blocks
+        available_blocks: &Vec<AvailableBlock<T::EthSpec>>,
+    ) -> Result<(), AvailabilityCheckError> {
+        let all_blobs = available_blocks
             .iter()
-            .filter(|block| self.blobs_required_for_block(block.as_block()))
+            .filter(|available_block| self.blobs_required_for_block(&available_block.block))
```
jimmygchen: redundant check i think

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -452,65 +423,36 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
                 .map_err(AvailabilityCheckError::InvalidBlobs)?;
         }
 
-        let all_data_columns = blocks
+        let all_data_columns = available_blocks
             .iter()
-            .filter(|block| self.data_columns_required_for_block(block.as_block()))
+            .filter(|available_block| self.data_columns_required_for_block(&available_block.block))
```
jimmygchen: redundant check i think

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -366,83 +366,54 @@ impl<T: BeaconChainTypes> DataAvailabilityChecker<T> {
...
-    /// This is more efficient than calling `verify_kzg_for_rpc_block` in a loop as it does
-    /// all kzg verification at once
+    /// Performs batch kzg verification for a vector of `AvailableBlocks`. This is more efficient than
+    /// calling `verify_kzg_for_available_block` in a loop.
     ///
     /// WARNING: This function assumes all required blobs are already present, it does NOT
     ///          check if there are any missing blobs.
     #[instrument(skip_all)]
-    pub fn verify_kzg_for_rpc_blocks(
+    pub fn batch_verify_kzg_for_available_blocks(
```
jimmygchen, line 388: The above WARNING can be removed as the function now requires `AvailableBlock`

beacon_node/beacon_chain/src/block_verification_types.rs
eserilev: Yes I've updated this comment to clear this up. 

`BlockOnly`: This is a post-deneb block that requires blobs to be considered fully available.`

beacon_node/beacon_chain/src/data_availability_checker.rs
eserilev: the refactor you suggested changes things here
[cdad1cd](https://github.com/sigp/lighthouse/pull/8424/commits/cdad1cd225d270b1b3dcedfbb606ede99e521792)

 I'm still making a sanity check, but its much clearer, and also indicates that its an unexpected error

beacon_node/beacon_chain/src/data_availability_checker.rs
eserilev: the refactor you suggested changes things here
[cdad1cd](https://github.com/sigp/lighthouse/pull/8424/commits/cdad1cd225d270b1b3dcedfbb606ede99e521792)

 I'm still making a sanity check, but its much clearer, and also indicates that its an unexpected error

beacon_node/beacon_chain/src/data_availability_checker.rs
eserilev: the refactor you suggested changes things here
[cdad1cd](https://github.com/sigp/lighthouse/pull/8424/commits/cdad1cd225d270b1b3dcedfbb606ede99e521792)

 I'm still making a sanity check, but its much clearer, and also indicates that its an unexpected error

beacon_node/beacon_chain/src/data_availability_checker.rs
eserilev: removed warning in two places (batch verify and verify kzg)

beacon_node/beacon_chain/src/data_availability_checker.rs
eserilev: we can do that, but I kinda like raising an `Unexpected` error message just for sanity. What do you think?

eserilev: the refactor is here
[cdad1cd](https://github.com/sigp/lighthouse/pull/8424/commits/cdad1cd225d270b1b3dcedfbb606ede99e521792)

beacon_node/beacon_chain/src/data_availability_checker.rs
eserilev: nvm i removed the redundant checks here 
b130672

beacon_node/beacon_chain/src/data_availability_checker.rs
eserilev: nvm i removed the redundant checks here 
b130672

beacon_node/beacon_chain/src/data_availability_checker.rs
eserilev: nvm i removed the redundant checks here 
b130672

beacon_node/beacon_chain/src/data_availability_checker.rs
```diff
@@ -759,31 +660,161 @@ pub enum AvailableBlockData<E: EthSpec> {
...
+                }
+            }
+            AvailableBlockData::DataColumns(data_columns) => {
+                if !columns_required {
+                    return Err(AvailabilityCheckError::InvalidAvailableBlockData);
+                }
+
+                let mut column_indices = da_checker
+                    .custody_context
+                    .custody_columns_for_epoch(Some(block.epoch()), &spec)
```
jimmygchen: i think we should use `sampling_columns_for_epoch` here instead? otherwise our full nodes won't be spec compliant (16 minimum samples)

beacon_node/network/src/sync/block_sidecar_coupling.rs
```diff
@@ -565,16 +598,26 @@ mod tests {
...
-        let spec = test_spec::<E>();
-        let expects_custody_columns = vec![1, 2, 3, 4];
+        let mut spec = test_spec::<E>();
+        spec.deneb_fork_epoch = Some(Epoch::new(0));
+        spec.fulu_fork_epoch = Some(Epoch::new(0));
+        let spec = Arc::new(spec);
+        let da_checker = Arc::new(test_da_checker(spec.clone()));
+        let expects_custody_columns = da_checker
+            .custody_context()
+            .custody_columns_for_epoch(None, &spec)
```
jimmygchen: tests here also need to make sure we use `sampling_columns_for_epoch`

beacon_node/beacon_chain/src/data_availability_checker.rs
jimmygchen: we should probably hold the sampling columns until they are avaialble and drop them during block import (i believe this is already happening?)

beacon_node/network/src/sync/block_sidecar_coupling.rs
eserilev: These tests are specifically for range sync. I don't think we download sampling columns during range sync

beacon_node/beacon_chain/src/data_availability_checker.rs
eserilev: As far as I can tell, we only download our custody requirement during range sync. I think `sampling_columns_for_epoch` would break range sync

During lookup we actually never construct an RpcBlock::Available variant. The blocks and blobs/columns are fetched and processed separately, whereas in range sync the components are coupled together before processing.
This is a big reason why we want to separate the types completely

beacon_node/network/src/sync/block_sidecar_coupling.rs
eserilev: oh looks like I'm wrong about that, it seems like we do download our sampling requirements as well

beacon_node/beacon_chain/src/data_availability_checker.rs
eserilev: nevermind! sampling columns for epoch is the right choice, it seems that we do fetch sampling columns both for range and backfill sync.

eserilev: and yes we do filter the sampling columns out before importing

eserilev:
revert sampling_columns_for_epoch optional epoch arg

eserilev:
this should be good to go

jimmygchen (approved):
Looks great now, thanks!

jimmygchen:
@mergify requeue

PR #8425, opened 2025-11-18
Compute missing_columns correctly

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

The difference is computed by taking the difference of expected with received. We were doing the inverse.

Thanks to Yassine for finding the issue.


michaelsproul:
@pawanjay176 can you rebase on `release-v8.0`, I think we can probably include this in v8.0.1

jimmygchen (approved):
I've rebased it on `release-v8.0`. 
Changes look good!

PR #8444, opened 2025-11-21
Fix: Recursively calculate directory size in `size_of_dir`

maradini77:
Fixed `size_of_dir_entry` to recursively calculate the size of directories. Previously, the function only returned the metadata size of directory entries (typically a few KB), completely ignoring files within subdirectories. This caused `size_of_dir` to return incorrect sizes for any directory tree with nested folders.

The fix checks if an entry is a directory and recursively calls `size_of_dir` to sum up all file sizes within the directory tree.


michaelsproul:
Please rebase your change on `unstable`.

michaelsproul (approved):
I was a little hesitant to merge this change, because:

1. We don't need it. The structure of the beacon chain database directory is completely flat (files inside a single directory).
2. I was concerned that this change could open the door to stack overflows or infinite loops.

Concern (2) is more important. Looking at the docs for [`DirEntry::metadata`](https://doc.rust-lang.org/std/fs/struct.DirEntry.html#method.metadata) and [`Metadata::is_dir`](https://doc.rust-lang.org/std/fs/struct.Metadata.html#method.is_dir) we see that it **would not** return `true` for a symlinked directory. This prevents the infinite loop attack.

The risk of a deeply nested directory creating a DoS via stack overflow (or very long run time) still exists, although it would require the datadir to be writable by an attacker, which is likely only possible if permissions are incorrectly set, _or_ there is a remote code execution vulnerability (which creates more serious problems beyond a simple DoS).

I'm leaning towards saying that this slightly increased risk is acceptable, given that it can be mitigated by setting file permissions correctly. This is despite the positive impact being low (no bug in LH's current behaviour). I think fixing the behaviour for the general case is still valuable, as it removes a footgun and saves us from potentially introducing a bug in the case where the assumption about a flat data directory is no longer true.

michaelsproul:
@maradini77 Do you have any thoughts on the above?

maradini77:
@michaelsproul Thanks for the careful review. Agreed: symlinks won’t cause an infinite loop because `is_dir` is false on them. I added recursion to avoid hidden bugs if the data directory ever stops being flat (e.g., future tools/plugins). To reduce DoS risk from deep nesting, I can switch to an iterative walk with a depth/entry cap and skip symlinks. Would you be comfortable if I add those protections?

michaelsproul:
@maradini77 Yeah let's make it iterative and capped, sounds good!

maradini77:
> @maradini77 Yeah let's make it iterative and capped, sounds good!

@michaelsproul updated

common/directory/src/lib.rs
```diff
@@ -64,22 +65,57 @@ pub fn parse_path_or_default_with_flag(
...
+                continue;
+            };
+            entries_seen += 1;
+
+            let Ok(file_type) = entry.file_type() else {
+                continue;
+            };
+            if file_type.is_symlink() {
+                continue;
+            }
```
michaelsproul: Couldn't we simplify this a little by continuing to use `entry.metadata()`? And just include a comment to say that `Metadata::is_dir` will return `false` for symlinks, and thus won't result in them being explored?

maradini77: @michaelsproul updated :)

michaelsproul: Nah I meant like delete this check entirely?

maradini77: @michaelsproul updated

maradini77:
@michaelsproul fmt fuxed

common/directory/src/lib.rs
macladson: ~~This actually isn't true. `Metadata::is_dir()` will return `true` for symlinks unless the metadata was obtained via `metadata_symlink()` (which we don't do): https://doc.rust-lang.org/std/fs/struct.Metadata.html#method.is_dir~~

~~You can verify this by checking `metadata.is_symlink()` inside the `is_dir()` check and it will come back true if your dir contains a symlink.
So we should swap this check to  `file_type()`  to avoid this and only compute the metadata if `file_type.is_file()`~~

common/directory/src/lib.rs
```diff
@@ -64,18 +65,51 @@ pub fn parse_path_or_default_with_flag(
...
+                continue;
+            };
+            // Note: Metadata::is_dir() is false for symlinks, so we won't recurse into them.
+            if metadata.is_dir() {
+                if depth < MAX_DEPTH {
+                    stack.push_back((entry.path(), depth + 1));
+                }
+            } else {
+                total_size = total_size.saturating_add(metadata.len());
+            }
```
macladson: As I mentioned in the other comment, we actually do need to do a `file_type()` check here. It should be very cheap so I don't think it is a performance concern. 

```suggestion
            let Ok(file_type) = entry.file_type() else {
                continue;
            };
            if file_type.is_dir() {
                if depth < MAX_DEPTH {
                    stack.push_back((entry.path(), depth + 1));
                }
            } else if file_type.is_file() {
                if let Ok(metadata) = entry.metadata() {
                    total_size = total_size.saturating_add(metadata.len());
                }
            }
```

macladson (changes requested):
Just noticed an inconsistency with the comments. Also, I think some tests would be useful to here to make sure that the behaviour we want is being enforced (particularly with regards to symlinks).

`tempfile` is a good crate to use for testing this and is already used elsewhere so it's already in our workspace dependencies. You can also create symlinks with the `symlink` function: https://doc.rust-lang.org/std/os/unix/fs/fn.symlink.html

common/directory/src/lib.rs
```diff
@@ -64,18 +65,52 @@ pub fn parse_path_or_default_with_flag(
...
+                continue;
+            };
+            if file_type.is_dir() {
+                if depth < MAX_DEPTH {
+                    stack.push_back((entry.path(), depth + 1));
+                }
+            } else if file_type.is_file() {
+                if let Ok(metadata) = entry.metadata() {
+                    total_size = total_size.saturating_add(metadata.len());
+                }
```
macladson: Sorry, the code I gave you didn't pass clippy whoops, this should fix it:
```suggestion
            } else if file_type.is_file()
                && let Ok(metadata) = entry.metadata()
            {
                total_size = total_size.saturating_add(metadata.len());
            }
```

common/directory/src/lib.rs
macladson: Sorry, I didn't notice we were using `DirEntry` and I was mistaking traversing symlinks with the function reporting the symlink path length (it is treated as if it is a regular file). The code I suggested will ignore symlinks entirely (not treat them as files). Arguably that behaviour is more correct though, but neither solution can lead to infinite recursion

macladson:
@maradini77 are you able to write some tests for this function?

Something like this:

```rust
#[cfg(test)]
mod tests {
    use super::*;
    use tempfile::TempDir;

    #[test]
    fn size_of_dir_empty() {
        let temp_dir = TempDir::new().unwrap();
        let size = size_of_dir(temp_dir.path());
        assert_eq!(size, 0);
    }

    // ... etc
}
```

PR #8445, opened 2025-11-22
Enable multi node attestation consensus for the VC

eserilev:
Allow for the VC to query all BN's at the beginning of each epoch and only enable attestation for that epoch if source and target checkpoint matches for a threshold of BN's

## TODO
- [x] make sure we are only comparing relevant fields in attestation data
- [x] As an optimization only check attestation data for the first slot in an epoch
- [x] Ensure we only fetch attestation data for the BN whose attestation data matches consensus
- [ ] Allow for consensus threshold to be set via CLI flag
- [ ] Write tests
- [ ] Test w/ single and multi bn setups w/ multiple consensus thresholds


eserilev:
@yorickdowne based on our convo in Buenos Aires, I thought you might find this draft PR interesting

galaxy-tbehrens:
Yes, I find it very interesting! Thank you! (This is the corpo account. I'm forwarding this to the Galaxy team as a potential option)

0xmrree:
Just curious are you gonna just pick one BN to get the ghost head (beacon_block_root)?

eserilev:
@0xmrree im actually doing most of my development on this branch at the moment
https://github.com/eserilev/lighthouse/pull/26

i've isolated the changes there to make things a bit easier to review

I'm not using the beacon block root at all. I'm just comparing the source and target checkpoints across all connected BN's and if a threshold agrees, I attest with one of the nodes that are in agreement. If a consensus threshold isnt reach, attestation is disabled for that epoch.

dapplion:
- Closing for now, the idea is solid but move to an issue

PR #8446, opened 2025-11-24
Optimise out block header calculation

michaelsproul:
## Issue Addressed

This is a `tracing`-driven optimisation. While investigating why Lighthouse is slow to send `newPayload`, I found a suspicious 13ms of computation on the hot path in `gossip_block_into_execution_pending_block_slashable`:

headercalc

Looking at the current implementation we can see that the _only_ thing that happens prior to calling into `from_gossip_verified_block` is the calculation of a `header`. We first call `SignatureVerifiedBlock::from_gossip_verified_block_check_slashable`:

https://github.com/sigp/lighthouse/blob/261322c3e3ee467c9454fa160a00866439cbc62f/beacon_node/beacon_chain/src/block_verification.rs#L1075-L1076

Which is where the `header` is calculated prior to calling `from_gossip_verified_block`:

https://github.com/sigp/lighthouse/blob/261322c3e3ee467c9454fa160a00866439cbc62f/beacon_node/beacon_chain/src/block_verification.rs#L1224-L1226

Notice that the `header` is _only_ used in the case of an error, yet we spend time computing it every time!

## Proposed Changes

This PR moves the calculation of the header (which involves hashing the whole beacon block, including the execution payload), into the error case. We take a cheap clone of the `Arc`'d beacon block on the hot path, and use this for calculating the header _only_ in the case an error actually occurs. This shaves 10-20ms off our pre-newPayload delays, and 10-20ms off every block processing :tada:

## Additional Info

I've also removed unnecessary header calculations from the RPC block codepath.

Here's a trace for a gossip block after this change. Note the header calculation right at the start of gossip verification, which encompasses some of the calculation of the block root:

Screenshot 2025-11-24 at 2 21 52 pm

I don't think we can/should remove this. Experimenting with a branch where we never compute the header (except for the slasher), shows marginal differences in timing. It is made more difficult by the variation in block root times being quite high across different blocks. Likely a targeted criterion bench would be better, but I'm happy to leave that for another PR.

### DIFF
--- a/beacon_node/beacon_chain/src/block_verification.rs
+++ b/beacon_node/beacon_chain/src/block_verification.rs
@@ -1221,9 +1221,13 @@ impl<T: BeaconChainTypes> SignatureVerifiedBlock<T> {
         from: GossipVerifiedBlock<T>,
         chain: &BeaconChain<T>,
     ) -> Result<Self, BlockSlashInfo<BlockError>> {
-        let header = from.block.signed_block_header();
-        Self::from_gossip_verified_block(from, chain)
-            .map_err(|e| BlockSlashInfo::from_early_error_block(header, e))
+        let block = from.block.clone();
+        Self::from_gossip_verified_block(from, chain).map_err(|e| {
+            // Lazily create the header from the block in case of error. Computing the header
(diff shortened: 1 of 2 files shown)


jimmygchen (approved):
Nice! Great find 🎉

PR #8451, opened 2025-11-24
Optimise pubkey cache initialisation during beacon node startup 

jimmygchen:
## Issue Addressed

Instrument beacon node startup and parallelise pubkey cache initialisation.

I instrumented beacon node startup and noticed that pubkey cache takes a long time to initialise, mostly due to decompressing all the validator pubkeys.

This PR uses rayon to parallelize the decompression on initial checkpoint sync. The pubkeys are stored uncompressed, so the decopression time is not a problem on subsequent restarts. On restarts, we still deserialize pubkeys, but the timing is quite minimal on Sepolia so I didn't investigate further.

`validator_pubkey_cache_new` timing on Sepolia:
* before: 109.64ms
* with parallelization: 21ms

on Hoodi:
* before: times out with Kurtosis after 120s
* with parallelization: 12.77s to import keys

**UPDATE**: downloading checkpoint state + genesis state takes about 2 minutes on my laptop, so it seems like the BN managed to start the http server just before timing out (after the optimisation).

image

### DIFF
--- a/beacon_node/client/src/builder.rs
+++ b/beacon_node/client/src/builder.rs
@@ -472,7 +472,6 @@ where
     }
 
     /// Starts the networking stack.
-    #[instrument(name = "start_networking_stack", skip_all)]
     pub async fn network(
         mut self,
         config: Arc<NetworkConfig>,
--- a/beacon_node/client/src/builder.rs
+++ b/beacon_node/client/src/builder.rs
@@ -42,7 +42,7 @@ use std::time::Duration;
 use std::time::{SystemTime, UNIX_EPOCH};
 use store::database::interface::BeaconNodeBackend;
 use timer::spawn_timer;
-use tracing::{debug, info, warn};
+use tracing::{debug, info, instrument, warn};
 use types::data_column_custody_group::compute_ordered_custody_column_indices;
 use types::{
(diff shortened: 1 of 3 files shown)


beacon_node/client/src/builder.rs
```diff
@@ -471,6 +472,7 @@ where
     }
 
     /// Starts the networking stack.
+    #[instrument(name = "start_networking_stack", skip_all)]
```
jimmygchen: I'm having trouble getting this span and the root span `start_beacon_node` exported to OTLP backend (tempo), and I couldn't figure out why. I don't see any errors in Tempo, so I suspect it's getting filtered / errorred while exporting.

jimmygchen: I suspect it's something to do with our span exporter setup - I'll try disabling workspace filter on the exporter and see if it fixes this.

dapplion (approved):
The logic in `import_new_pubkeys` is correct to me, pending your comments on instrumentation working right. Nice tests!

jimmygchen:
> The logic in `import_new_pubkeys` is correct to me, pending your comments on instrumentation working right. Nice tests!

Thanks! I've tried a few fixes but didn't work so I've removed those two spans for the time being.

Issue #8408, opened 2025-11-13
Run `cargo deny` or similar on CI


michaelsproul:
## Description

We're successfully getting our dependencies under control, but it would be nice to prevent regressions.

We could set up `cargo deny` on CI with a list of "banished" crates that we never want to see again, including:

- `ethers`
- `ethereum-types`
- `protobuf`

There are also other crates, which once we remove duplicate versions, we would like to try to prevent reintroducing duplicates of, e.g.

- `http`
- `reqwest`

Reducing duplicates of heavy crates (particularly ones with build scripts) is a good way to reduce compilation time, as well as attack surface.

chong-he:
`cargo deny` added in #8580

PR #8452, opened 2025-11-24
Consolidate reqwest versions

michaelsproul:
## Proposed Changes

Update `reqwest` to 0.12 so we only depend on a single version. This should slightly improve compile times and reduce binary bloat.

## Additional Info

For backwards-compatibility with `warp` there is a bit of converting between `StatusCode`s from `http 0.2.x` and `http 1.x`.

Related:

- https://github.com/sigp/lighthouse/issues/8408


macladson (approved):
Nice! Looks good. Although I'm surprised how few situations there are where we need to specifically insert a status code into a warp reply

Issue #5403, opened 2024-03-12
Mandatory fee recipient


michaelsproul:
## Description

We should consider making the `fee_recipient` mandatory, to avoid blocks going to junk addresses

See: https://twitter.com/Butta_eth/status/1767632279680024759

Lighthouse doesn't actually default to 0x00, but there have still been [some block rewards sent to our default 0x01](
https://etherscan.io/address/0x0000000000000000000000000000000000000001#mine), although not for >50 days at time of writing.

https://github.com/sigp/lighthouse/blob/2a3c709f8ceec4d47c1934446511660c429fe90a/beacon_node/execution_layer/src/lib.rs#L84-L85

## Steps to resolve

The hard part is deciding how and where to make it mandatory. There are many places the fee recipient can be set:

1. Fallback fee recipient in the beacon node. Should this be mandatory? Probably not?
2. Fallback fee recipient in the validator client. Probably the best candidate for being made mandatory.
3. Per-validator fee recipient from `validator_definitions.yaml`. We could make it mandatory that _some_ fee recipient is set for every validator, either all validators are configured with an individual fee recipient, or the fallback recipient is set. This could have better UX than making the fallback mandatory, as many users run validators on behalf of other entities and don't want to set a fallback address.

AbolareRoheemah:
Hi @michaelsproul can I be assigned this?

chong-he:
Completed in #8454 

Lighthouse VC now wouldn't start if there is no `suggested_fee_recipient` for all validators in `validator_definitions.yml` AND no `--suggested-fee-recipient` flag provided

PR #8454, opened 2025-11-24
added check for fee recipient per validator and added unit tests

AbolareRoheemah:
## Issue Addressed
Addresses #5403 

## Proposed Changes
 - Added `check_fee_recipient()` method to validate individual validators
 - Added `check_all_fee_recipients()` to validate all validators on startup
- Validator client now fails to start if any enabled validator lacks a fee recipient and no global flag is used.
 - Added Clear error messages to guide users on how to fix the issue
 - Added unit tests

## Additional Info
The images show attempts to run the VC: 1. without fee-recipient set for a validator and no global flag 2. Only global flag used. No individual fee-recipient set
Screenshot 2025-11-23 at 20 09 08
Screenshot 2025-11-23 at 20 10 21


common/account_utils/src/validator_definitions.rs
```diff
@@ -211,6 +211,26 @@ impl ValidatorDefinition {
...
+    pub fn check_fee_recipient(&self, global_fee_recipient: Option<Address>) -> Result<(), String> {
+        // Skip disabled validators. Also skip if validator has its own fee set, or the global flag is set
+        if !self.enabled || self.suggested_fee_recipient.is_some() || global_fee_recipient.is_some()
+        {
+            return Ok(());
+        }
+
+        // If nothing set
+        Err(format!(
+            "Validator {} is missing `suggested_fee_recipient`!\n\n\
```
eserilev: Can we update the error message to

```
Validator {} is missing a `suggested_fee_recipient`. Fix this by adding a `suggested_fee_recipient` in your `validator_definitions.yml` or by supplying a fallback fee recipient via the `--suggested-fee-recipient` flag.
```

We usually avoid `\n`. We can use `\` to continue the string literal across multiple lines

common/account_utils/src/validator_definitions.rs
```diff
@@ -211,6 +211,26 @@ impl ValidatorDefinition {
             },
         })
     }
+
+    pub fn check_fee_recipient(&self, global_fee_recipient: Option<Address>) -> Result<(), String> {
```
eserilev: I think for better UX, we should iterate across all fee recipients, build a list of ones missing fee recipients and then log an error to the user. This way they can see the full list of missing fee recipients all at once.

common/account_utils/src/validator_definitions.rs
```diff
@@ -409,6 +429,34 @@ impl ValidatorDefinitions {
...
+        if global_fee_recipient.is_some() {
+            let missing = self
+                .0
+                .iter()
+                .filter(|d| d.enabled && d.suggested_fee_recipient.is_none())
+                .count();
+            if missing > 0 {
+                tracing::warn!(
+                    "Global --suggested-fee-recipient is being used for {} validator(s). \
+                        Consider setting it in validator_definitions.yml for each one.",
```
eserilev:
```suggestion
                    "The fallback --suggested-fee-recipient is being used for {} validator(s). \
                        Consider setting the fee recipient for each validator individually via `validator_definitions.yml`.",
```

common/account_utils/src/validator_definitions.rs
```diff
@@ -409,6 +429,34 @@ impl ValidatorDefinitions {
...
+
+        // Friendly reminder for users using the global flag
+        if global_fee_recipient.is_some() {
+            let missing = self
+                .0
+                .iter()
+                .filter(|d| d.enabled && d.suggested_fee_recipient.is_none())
+                .count();
+            if missing > 0 {
+                tracing::warn!(
```
eserilev:
```suggestion
                tracing::info!(
```

eserilev (changes requested):
Thanks for this! The added tests are very much appreciated. I think we can make some slight tweaks here to improve the UX a bit.

AbolareRoheemah:
> Thanks for this! The added tests are very much appreciated. I think we can make some slight tweaks here to improve the UX a bit.

Hello @eserilev. Much thanks for the review and the kind words. 🙏

I thought to explain the changes I made with this last push. The `check_fee_reciepient` function now returns the public key of the validator with missing fee recipient. This is done to make collecting all missing validators easy.
Then `check_all_fee_recipients` now gathers all validators missing a fee recipient and reports them together in a single error message (comma-separated list).

I also modified existing tests to work with the new `check_fee_reciepient` return type and added a test (check_all_fee_recipients_reports_all_missing) that verifies multiple missing validators are all reported in a single error.

eserilev (approved):
LGTM!

eserilev:
Hi @AbolareRoheemah can you please run `cargo fmt --all` to fix the CI issues when you get the chance

common/account_utils/src/validator_definitions.rs
```diff
@@ -682,4 +739,250 @@ mod tests {
...
+            prefer_builder_proposals: None,
+            signing_definition: SigningDefinition::LocalKeystore {
+                voting_keystore_path: PathBuf::new(),
+                voting_keystore_password_path: None,
+                voting_keystore_password: None,
+            },
+        };
+
+        // Should return None since global fee recipient is set
+        let global_fee = Some(Address::from_str("0xa2e334e71511686bcfe38bb3ee1ad8f6babcc03d").unwrap());
```
chong-he: This is just a nit for better clarity 
```suggestion
        let global_fee_recipient = Some(Address::from_str("0xa2e334e71511686bcfe38bb3ee1ad8f6babcc03d").unwrap());
```

common/account_utils/src/validator_definitions.rs
```diff
@@ -682,4 +739,250 @@ mod tests {
...
+        let check_result = def.check_fee_recipient(global_fee);
+        assert!(check_result.is_none());
```
chong-he: For succinctness, we can combine this with the test `fee_recipient_check_fails_when_missing()` to avoid repeatedly defining the same fields. We can revise the test name a bit, then add the checks in the same tests

Edit: I believe we can even go "all out" for succinctness, to define a general `ValidatorDefinitions`, then for each test, update the necessary fields for each test purpose, and add an assert right below, i.e., instead of having multiple tests, we can have just 1 or 2 test functions, move the test name (and revise a bit) and put it as a comment right above and before the assert. This will be cleaner as well

common/account_utils/src/validator_definitions.rs
```diff
@@ -410,6 +420,52 @@ impl ValidatorDefinitions {
...
+        if !missing.is_empty() {
+            let pubkeys = missing
+                .iter()
+                .map(|pk| pk.to_string())
+                .collect::<Vec<_>>()
+                .join(", ");
+            
+            return Err(format!(
+                "The following validators are missing a `suggested_fee_recipient`: {}. \
+                 Fix this by adding a `suggested_fee_recipient` in your \
```
chong-he: I feel that `the validator_definitions.yml` reads better here, but feel free to take it or leave it
```suggestion
                 Fix this by adding a `suggested_fee_recipient` in the \
```

common/account_utils/src/validator_definitions.rs
```diff
@@ -410,6 +420,52 @@ impl ValidatorDefinitions {
...
+        if global_fee_recipient.is_some() {
+            let count = self
+                .0
+                .iter()
+                .filter(|d| d.enabled && d.suggested_fee_recipient.is_none())
+                .count();
+            if count > 0 {
+                info!(
+                    "The fallback --suggested-fee-recipient is being used for {} validator(s). \
+                     Consider setting the fee recipient for each validator individually via `validator_definitions.yml`.",
```
chong-he: This is about "reducing the severity" of the log, because if users only use the `suggested-fee-recipient` flag, this log will appear every time a VC starts, so it could be a bit of a warn thingy.

This is a my two cent, keep to hear from the author and other developers about this. (e.g., maybe even shift to debug log)?

```suggestion
                     You may alternatively set the fee recipient for each validator individually via `validator_definitions.yml`.",
```

chong-he (commented):
Looks pretty good to me. I have also tested manually and confirm that it's working. 

Some comments below.

AbolareRoheemah:
> Looks pretty good to me. I have also tested manually and confirm that it's working.
> 
> Some comments below.

Thank for your review @chong-he and sorry for the delayed response.

common/account_utils/src/validator_definitions.rs
AbolareRoheemah: Thanks. It does make things clearer. 👍

common/account_utils/src/validator_definitions.rs
AbolareRoheemah: Thanks for spotting this. I feel those two tests should be combined. 🤝

About having just 1 or 2 test functions, I actually tried compressing the tests into 3 functions, and while it makes things more compact, the functions appeared not so comprehensible due to how long they became. So I think I prefer to separate test concerns as much as possible.

common/account_utils/src/validator_definitions.rs
AbolareRoheemah: For this, I previously used a warn, and @eserilev suggested I change it to an info log.  I don't mind changing it to a debug or something.

AbolareRoheemah:
> Hi @AbolareRoheemah can you please run `cargo fmt --all` to fix the CI issues when you get the chance

Done! Sorry for the late response. 🙇‍♀️

common/account_utils/src/validator_definitions.rs
chong-he: Let's combine those 2 tests at least

To me, having a longer function is ok as long as the tests are clearly defined / commented above it. It is nicer to have fewer lines that can cover the tests (e.g., avoid defining the fields repeatedly) looks cleaner.

common/account_utils/src/validator_definitions.rs
chong-he: Let's wait to hear from Eitan and see what he says. 

I think even removing this log looks good to me (or move to debug), because the `suggested_fee_recipient` is something sensitive to users and having a new INFO log there could be a surprise and users may start to questions or panic if they should define in the yml file

common/account_utils/src/validator_definitions.rs
AbolareRoheemah: > For succinctness, we can combine this with the test `fee_recipient_check_fails_when_missing()` to avoid repeatedly defining the same fields. We can revise the test name a bit, then add the checks in the same tests

Yes. The two tests you highlighted have been combined in the update I pushed yesterday. 🫡

common/account_utils/src/validator_definitions.rs
eserilev: yeah debug or removing the log seems fine to me.

AbolareRoheemah: Noted

AbolareRoheemah: Done @eserilev @chong-he

chong-he:
> > Hi @AbolareRoheemah can you please run `cargo fmt --all` to fix the CI issues when you get the chance
> 
> Done! Sorry for the late response. 🙇‍♀️



> > Hi @AbolareRoheemah can you please run `cargo fmt --all` to fix the CI issues when you get the chance
> 
> Done! Sorry for the late response. 🙇‍♀️

Hi, you also need to change the `info!` to `debug!` if you want the log to display as `DEBUG`. Alternatively, for simplicity we can remove the log altogether and also reduce quite a few lines in the code

common/account_utils/src/validator_definitions.rs
AbolareRoheemah: Done @chong-he @eserilev

AbolareRoheemah:
> Hi, you also need to change the `info!` to `debug!` if you want the log to display as `DEBUG`.

I can't believe I changed the import from to debug and didn't change the actual implementation. 🤦‍♀️

> Alternatively, for simplicity we can remove the log altogether and also reduce quite a few lines in the code

For this, are you suggesting there's no need to inform users when the --suggested-fee-recipient fallback is being used? I would appreciate if you could clarify further. 🙏 @chong-he

chong-he:
> For this, are you suggesting there's no need to inform users when the --suggested-fee-recipient fallback is being used? I would appreciate if you could clarify further. 🙏 @chong-he

Yes, we can simplify to not logging anything, that can make it cleaner
but if you want to keep it I am good with it as well, we log as debug. I am fine either way

AbolareRoheemah:
> Yes, we can simplify to not logging anything, that can make it cleaner but if you want to keep it I am good with it as well, we log as debug. I am fine either way

Thanks for clarifying @chong-he. 

So I kept the debug log because I think it could possibly prevent accidental misconfiguration or missed rewards or just encourage or inform users of "best practices."

chong-he (approved):
Looks good now, thanks for the PR

jimmygchen:
@chong-he 

One thing worth considering: validators added at runtime via the HTTP API (POST `/lighthouse/validators/*` and the `web3signer/remotekeys` endpoints) don't go through this check. The `suggested_fee_recipient` field is optional in all those request bodies, so a validator can still be added without one and silently lose rewards to 0x00..01.

Not necessarily a blocker for this PR, but worth deciding whether runtime-added validators should get the same protection (either by validating in the API handlers, or noting it as a follow-up).

chong-he:
> @chong-he
> 
> One thing worth considering: validators added at runtime via the HTTP API (POST `/lighthouse/validators/*` and the `web3signer/remotekeys` endpoints) don't go through this check. The `suggested_fee_recipient` field is optional in all those request bodies, so a validator can still be added without one and silently lose rewards to 0x00..01.
> 
> Not necessarily a blocker for this PR, but worth deciding whether runtime-added validators should get the same protection (either by validating in the API handlers, or noting it as a follow-up).

Good point. 

If we validate in the API handlers, should it prevent importing the key if no `suggested_fee_recipient` is provided? My thinking is no because the KeyManager API or `/lighthouse/validators/keystore` has no requirement for this. Maybe add a warning during importing via HTTP API if it is not provided? Then, it will be checked upon a restart. What do you think?

PR #8455, opened 2025-11-24
Dockerfile with cargo artifacts caching

dapplion:
## Issue Addressed

When developing locally with kurtosis a typical dev workflow is:

loop:
- Build local lighthouse docker image
- Run kurtosis
- Observe bug
- Fix code

The docker build step would download and build all crates. Docker docs suggests an optimization to cache build artifacts, see https://docs.docker.com/build/cache/optimize/#use-cache-mounts

I have tested and it's like building Lighthouse outside of a docker environment 🤤  The docker build time after changing one line in the top beacon_node crate is 50 seconds on my local machine ❤️ 

The release path is un-affected. Do you have worries this can affect the output of the release binaries? This is too good of an improvement to keep it in a separate Dockerfile.


jimmygchen (approved):
This doesn't affect the release images as we use a separate `Dockerfile.cross` for release binaries.

I was thinking to do a bit more testing before merging them, but since you've also tested it, let's merge!

PR #8464, opened 2025-11-25
Add insecure-dep test task to Makefile and CI

uvizhe:
## Issue Addressed

#8106

## Proposed Changes

I added `insecure-deps` target to Makefile and a new step into `check-code` section of test-suite CI workflow that uses the former.

That bash multiliner is not ideal, I'd prefer a cargo plugin instead but none exists.

I also changed Cargo.toml to test that the new CI check works. Once we see a pipeline fails, I revert the change.

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -329,6 +329,8 @@ jobs:
       run: make cargo-fmt
     - name: Lint code for quality and style with Clippy
       run: make lint-full
+    - name: Check dependencies for usage of an unencrypted HTTP
+      run: make insecure-deps
     - name: Certify Cargo.lock freshness
       run: git diff --exit-code Cargo.lock
     - name: Typecheck benchmark code without running it
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -329,7 +329,7 @@ jobs:
       run: make cargo-fmt
     - name: Lint code for quality and style with Clippy
       run: make lint-full
-    - name: Check dependencies for usage of an unencrypted HTTP
+    - name: Check dependencies for unencrypted HTTP links
       run: make insecure-deps
     - name: Certify Cargo.lock freshness
       run: git diff --exit-code Cargo.lock
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -327,10 +327,10 @@ jobs:
           bins: cargo-audit,cargo-deny
     - name: Check formatting with cargo fmt
       run: make cargo-fmt
-    - name: Lint code for quality and style with Clippy
-      run: make lint-full
     - name: Check dependencies for unencrypted HTTP links
       run: make insecure-deps
+    - name: Lint code for quality and style with Clippy
+      run: make lint-full
     - name: Certify Cargo.lock freshness
       run: git diff --exit-code Cargo.lock
     - name: Typecheck benchmark code without running it
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -274,7 +274,7 @@ validator_test_rig = { path = "testing/validator_test_rig" }
 warp = { version = "0.3.7", default-features = false, features = ["tls"] }
 warp_utils = { path = "common/warp_utils" }
 workspace_members = { path = "common/workspace_members" }
-xdelta3 = { git = "http://github.com/sigp/xdelta3-rs", rev = "4db64086bb02e9febb584ba93b9d16bb2ae3825a" }
+xdelta3 = { git = "https://github.com/sigp/xdelta3-rs", rev = "4db64086bb02e9febb584ba93b9d16bb2ae3825a" }
 zeroize = { version = "1", features = ["zeroize_derive", "serde"] }
 zip = "0.6"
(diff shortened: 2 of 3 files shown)


uvizhe:
Oops, I did not notice there's already another PR on the same issue. Feel free to close this

Makefile
```diff
@@ -296,6 +296,15 @@ vendor:
 udeps:
 	cargo +$(PINNED_NIGHTLY) udeps --tests --all-targets --release --features "$(TEST_FEATURES)"
 
+# Checks dependencies for usage of an unencrypted HTTP
```
chong-he:
```suggestion
# Checks dependencies for unencrypted HTTP links
```

.github/workflows/test-suite.yml
```diff
@@ -329,6 +329,8 @@ jobs:
       run: make cargo-fmt
     - name: Lint code for quality and style with Clippy
       run: make lint-full
+    - name: Check dependencies for usage of an unencrypted HTTP
```
chong-he:
```suggestion
    - name: Check dependencies for unencrypted HTTP links
```

Cargo.toml
```diff
@@ -274,7 +274,7 @@ validator_test_rig = { path = "testing/validator_test_rig" }
 warp = { version = "0.3.7", default-features = false, features = ["tls"] }
 warp_utils = { path = "common/warp_utils" }
 workspace_members = { path = "common/workspace_members" }
-xdelta3 = { git = "https://github.com/sigp/xdelta3-rs", rev = "4db64086bb02e9febb584ba93b9d16bb2ae3825a" }
+xdelta3 = { git = "http://github.com/sigp/xdelta3-rs", rev = "4db64086bb02e9febb584ba93b9d16bb2ae3825a" }
```
chong-he: Why is this changed to http?

Makefile
```diff
@@ -296,6 +296,15 @@ vendor:
...
+# Tee preserves output. If there are matches, print a message and return 1
+insecure-deps:
+	bash -c \
+	  "find -name Cargo.toml \
+			| xargs grep -P \"git\s?=\s?[\\\"']http:\" \
+			| tee /dev/tty \
+			| [ \`wc -l\` == 0 ] || (echo \"Using plain HTTP in dependencies is forbidden\" && false)"
+
```
chong-he: We could add a message to display when no http link is found (the all good case)

chong-he (changes requested):

Cargo.toml
uvizhe: This is temporary, just to see the new check in action

uvizhe: I've just discovered that in the previous pipeline run the test fired due to a different and unforeseen reason (no /dev/tty). I need to redesign a test to account for this.
https://github.com/sigp/lighthouse/actions/runs/20455982263/job/58778996686

uvizhe: The third commit should resolve the issue. I think it's unnecessary to test it again with a broken dep.

Makefile
```diff
@@ -343,6 +343,15 @@ vendor:
 udeps:
 	cargo +$(PINNED_NIGHTLY) udeps --tests --all-targets --release --features "$(TEST_FEATURES)"
 
+# Checks dependencies for unencrypted HTTP links
+# Tee preserves output. If there are matches, print a message and return 1
```
chong-he: This line is not relevant now
```suggestion
```

Makefile
```diff
@@ -343,6 +343,15 @@ vendor:
...
+.ONESHELL:
+SHELL = /bin/bash
```
chong-he: These two lines seem not necessary, we can have the command directly (need to tweak a bit the command) to check for http links, and then can error or `exit 1` if found

Makefile
```diff
@@ -343,6 +343,12 @@ vendor:
 udeps:
 	cargo +$(PINNED_NIGHTLY) udeps --tests --all-targets --release --features "$(TEST_FEATURES)"
 
+# Checks dependencies for unencrypted HTTP links
+insecure-deps:
+	BAD_LINKS=`find -name Cargo.toml | xargs grep -P "git\s?=\s?[\"']http:"`; \
+	if [ "_$$BAD_LINKS" = "_" ]; then echo "All Git dependencies use secure HTTPS"; \
```
chong-he: I am not super familiar with shell but I think `if [ -z "$$BAD_LINKS" ];` achieves the same purpose and is cleaner

chong-he (commented):
We mostly use `exit 1` to indicate error in script files (e.g., in `./scripts/cli.sh`), but this isn't a script file and with only 3 lines it doesn't really need a separate script file. 

So using `false`  which achieves the same purposes is fine to me.

We could also add a @ in front of `BAD_LINKS` definition to suppress printing out command, but that's also not strictly necessary.

So once the below minor comment is fixed, it's good to go from my point of view

Makefile
uvizhe: Changed 'false' to 'exit 1' for clarity. Thanks for the review and suggestions

chong-he (approved):
Looks good to me, thanks for the PR!

dapplion (approved):

PR #8469, opened 2025-11-26
Delete attester cache

jimmygchen:
## Issue Addressed

Fixes attester cache write lock contention. Alternative to #8463.

## Description

Deletes the attester cache which was causing lock contention across multiple hot code paths:

1. **Attestation production** - Every request took a read lock to check the cache. Cache misses additionally held a [write lock](https://github.com/jimmygchen/lighthouse/blob/d6cec0ba50e60fbf5ff0f7fbde0352b07fbde399/beacon_node/beacon_chain/src/attester_cache.rs#L282-L286) while loading state from disk and [building committee caches](https://github.com/jimmygchen/lighthouse/blob/d6cec0ba50e60fbf5ff0f7fbde0352b07fbde399/beacon_node/beacon_chain/src/attester_cache.rs#L96-L97), blocking all other attestation requests
2. **Block processing** - Took write lock when caching block states, blocking all attestation production
3. **Finalization** - Took write lock when pruning, blocking all attestation production

We now use the state cache directly via `get_advanced_hot_state`, which builds all caches without lock contention.


beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -2005,34 +1993,27 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
-                // The suitable values were already cached. Return them.
-                cached_values
             } else {
-                debug!(
-                    ?beacon_block_root,
-                    %head_state_slot,
-                    %request_slot,
-                    "Attester cache miss"
-                );
+                let (advanced_state_root, mut state) = self
```
dapplion, line 2001: Can you add metrics for the number of times we hit this code path? Would be nice to have a histogram here or instrumentation of the time spent computing the hot state

dapplion:
Nice simplification!

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Good idea - I've instrumented the `produce_unaggregated_attestation` code path, and we already have a span for `get_advanced_hot_state`, so this should give us good visibility.

eserilev (approved):
This is awesome! LGTM

jimmygchen:
Resolved conflicts, will check the new metrics tomorrow before merging

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: I've checked the spans, most of it happens <1ms. 

There's only one instance where it took 88ms. Looking at the code, it's likely time taken to acquire fork choice read lock here:
https://github.com/sigp/lighthouse/blob/75824e6599fe301a45df688c91e739547f2bf84e/beacon_node/beacon_chain/src/beacon_chain.rs#L1974-L1977

jimmygchen:
I've compared the following metrics running this version against other hoodi nodes and see no noticeable differences on similar spec machines
* `beacon_attestation_production_seconds_bucket`
* `beacon_block_processing_seconds_bucket`

dapplion (approved):
Let's ship this!

PR #8471, opened 2025-11-26
Remove quickcheck in favour of proptest

michaelsproul:
## Proposed Changes

Consolidate our property-testing around `proptest`. This PR was written with Copilot and manually tweaked.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3064,16 +3064,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "env_logger"
-version = "0.8.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a19187fea3ac7e84da7dacf48de0c45d63c6a76f9490dae389aead16c243fce3"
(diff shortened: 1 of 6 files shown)


michaelsproul:
@copilot I'm getting errors in the merkle_proof tests when running locally, especially with the `PROPTEST_CASES=10000` env variable set:

> ---- tests::proptest_push_leaf_and_verify stdout ----

thread 'tests::proptest_push_leaf_and_verify' (32391330) panicked at consensus/merkle_proof/src/lib.rs:417:5:
Test aborted: Too many global rejects
	successes: 6231
	local rejects: 0
	global rejects: 1024
		1024 times at consensus/merkle_proof/src/lib.rs:441:13: int_leaves.len() <= 2usize.pow(depth as u32)

Could we generate the `int_leaves` at runtime based on the depth, such that the length is always `<= 2^depth`?

eserilev (approved):
LGTM
I reviewed with the help of Claude btw

PR #8488, opened 2025-11-28
Fix Docker build cache issue for deposit contract files

jimmygchen:
## Description

Fixes Docker build failures when using cache mounts. The build was failing with errors like:

```
error: couldn't read `common/deposit_contract/src/../contracts/v0.12.1_validator_registration.json`: No such file or directory (os error 2)
  --> common/deposit_contract/src/lib.rs:47:24
```

The deposit contract crate uses a build script (`build.rs`) to download contract ABI and bytecode files from GitHub, storing them in `common/deposit_contract/contracts/`. These files are then included at compile time via `include_bytes!()`.

The Docker cache mount for `/lighthouse/target` caused Cargo to skip re-running the build script on subsequent builds, but the generated contract files weren't persisted in the cache. This resulted in compilation failures because `include_bytes!()` couldn't find the required files.

Workaround right now is to run `docker builder prune -a -f`

The fix adds a cache mount for the contracts directory, ensuring generated files persist across builds.


michaelsproul:
I think the correct fix is for the build script to put the files in `target`. Build scripts aren't meant to write to the source directory

jimmygchen:
will go with #8518 instead which implements what @michaelsproul suggested above, it's a much better approach.

PR #8489, opened 2025-11-28
Update local testnet scripts for the fulu fork

jimmygchen:
## Issue Addressed

* Remove `fulu-devnet-3` testing on CI 
* Delete `scripts/local_testnet/network_params_das.yaml` and consolidate it into the main `network_params.yaml` file we use on CI
* Delete enclave before building image, so it doesn't cause slow image building.


eserilev (approved):
LGTM

PR #8496, opened 2025-11-28
fix(bls): fix is_infinity when aggregating onto empty AggregateSignature

### LINKED ISSUE #8491: BLS: Fix `is_infinity` flag when aggregating onto empty AggregateSignature
## Description

**Original PR:** This _potential_ bug was identified by @Galoretka but the PR went stale and was never reviewed / merged.
* #8013

I haven't confirmed the validity of the bug but thought it's worth investigating this further.

## Analysis by 🤖 (may not be accurate)

The `is_infinity` flag in `AggregateSignature` becomes inconsistent when aggregating an infinity signature onto an empty aggregate. This happens because the aggregation methods use conjunction (`&&`) to update the flag instead of checking if `self` is empty first.

**Current behavior:**
1. Start with `AggregateSignature::empty()` → `point: None`, `is_infinity: false`
2. Aggregate an infinity signature onto it
3. Result: `point` is now infinity, but `is_infinity` flag remains `false`

This breaks `eth_fast_aggregate_verify` for the edge case of empty pubkeys, causing the signature check at line 205 to fail when it should pass.

**Impact:** Low severity. Lighthouse doesn't trigger this code path in production because:
- Block production uses `SyncAggregate::new()` which starts from `infinity()`, not `empty()`
- Deserialization correctly sets the flag based on serialized bytes
- The bug only affects in-memory aggregation starting from `empty()`

This could affect external tools using Lighthouse's BLS library incorrectly or cause interoperability issues.

## Code References

The bug exists in two methods:

**`add_assign` method:**
https://github.com/sigp/lighthouse/blob/713e4779129bd999202b09c91e96d4deda16b6d3/crypto/bls/src/generic_aggregate_signature.rs#L125-L136

**`add_assign_aggregate` method:**
https://github.com/sigp/lighthouse/blob/713e4779129bd999202b09c91e96d4deda16b6d3/crypto/bls/src/generic_aggregate_signature.rs#L139-L150

## Steps to resolve

Update both `add_assign` and `add_assign_aggregate` methods to check if `self.point.is_none()` before updating `is_infinity`:

- When `self` is empty (`point.is_none()`), set `is_infinity = other.is_infinity`
- Otherwise, keep the existing conjunction behavior: `is_infinity = self.is_infinity && other.is_infinity`

This aligns with the documented behavior that `empty()` acts as "set to infinity before aggregation" and ensures the flag stays consistent with the underlying point.

**Testing:** Add a test case that:
1. Creates an `AggregateSignature::empty()`
2. Aggregates an infinity signature onto it
3. Verifies `is_infinity()` returns `true`
4. Verifies `eth_fast_aggregate_verify` with empty pubkeys returns `true`

## Additional Info

This is a correctness issue rather than a critical mainnet bug. Fixing it improves robustness for edge cases and external library usage.

figtracer:
## Issue
this bug happens because we are computing the flag `is_infinity` before branching, but the correct `is_infinity` value depends on the branch. 

let's say that `self` is empty and we add an infinity signature (so `other` is AggSig::infinity()):
1. `self.is_infinity` starts as `false`
2. `other.is_infinity` is `true`
3.  flag immediately updates to `false && true = false` incorrectly because we will then branch to `else` where we effectively add infinity + infinity; hence the flag should be `true && true = true`

to fix this:
1. we move the logical operator into the inner if statement: `self.is_infinity` flag is only set to `true` if both are points at infinity
2. we write `self.is_infinity = other.is_infinity`; if we branch to else, essentially we're adding infinity to whatever `other` is, so if `other` is infinity, `self.is_infinity = true`else `self.is_infinity = false`

## Test Results
before the fix:
```
---- blst::empty_aggregate_plus_infinity_should_be_infinity stdout ----

thread 'blst::empty_aggregate_plus_infinity_should_be_infinity' (454378) panicked at crypto/bls/tests/tests.rs:545:5:
is_infinity flag should be true after adding infinity to empty
```

after:
```
test blst::empty_aggregate_plus_infinity_should_be_infinity ... ok
(...)

test result: ok. 38 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.11s
```

fixes #8491


kirk-baird (approved):
Thank you for this fix @figtracer this is a great catch.

I agree that if the current aggregate is empty then we should set infinity to match the added point. (empty + infinity = infinity).

michaelsproul:
@mergify requeue

PR #8497, opened 2025-11-28
Gloas - add get_payload_attestation_endpoint

shane-moore:
## Changes
Add `GET /eth/v1/validator/payload_attestation_data/{slot}]` per beacon api [spec](https://github.com/ethereum/beacon-APIs/pull/552)

## TODO
- Implement actual payload_present check once  envelope tracking is all set
 - Implement actual `blob_data_available` check once da checker changes are all set
 - Enable tests once gloas block production is functional

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -2121,6 +2121,9 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
 
         let beacon_block_root = head.beacon_block_root;
 
+        // TODO(gloas) the spec only requires that the payload be seen. right
+        // now we are setting payload_present to `true` if the envelope was imported
+        // into fork choice which is a stricter requirement than the spec requires.
         let payload_present = self
             .canonical_head
             .fork_choice_read_lock()
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -2121,8 +2121,9 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
 
         let beacon_block_root = head.beacon_block_root;
 
-        // TODO(gloas) this isn't really what the `envelope_times_cache` is meant for.
-        // We should use a dedicated payload envelope cache instead (maybe the new Gloas DA cache?)
+        // TODO(gloas) do we want to use a dedicated envelope cache instead?
+        // Maybe the new gloas DA cache? (Or should the gloas DA cache use 
+        // the envelopes_times_cache internally?)
         let payload_present = self
             .envelope_times_cache
             .read()
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -2121,12 +2121,13 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
 
         let beacon_block_root = head.beacon_block_root;
 
-        // TODO(EIP-7732): Check if we've seen a SignedExecutionPayloadEnvelope
-        // referencing this block root. For now, default to false.
-        let payload_present = false;
+        let payload_present = self
+            .canonical_head
+            .fork_choice_read_lock()
+            .contains_payload(&beacon_block_root);
 
-        // TODO(EIP-7732): Check blob data availability. For now, default to false.
-        let blob_data_available = false;
+        // TODO(EIP-7732): Check blob data availability. For now, default to true.
+        let blob_data_available = true;
 
         Ok(PayloadAttestationData {
             beacon_block_root,
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -2122,7 +2122,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         let beacon_block_root = head.beacon_block_root;
 
         // TODO(gloas) do we want to use a dedicated envelope cache instead?
-        // Maybe the new gloas DA cache? (Or should the gloas DA cache use 
+        // Maybe the new gloas DA cache? (Or should the gloas DA cache use
         // the envelopes_times_cache internally?)
         let payload_present = self
             .envelope_times_cache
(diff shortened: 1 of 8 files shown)


eserilev (approved):
LGTM

I pushed up SSZ support and `contains_payload` logic

beacon_node/http_api/tests/tests.rs
```diff
@@ -8057,6 +8128,34 @@ async fn get_validator_attestation_data_with_skip_slots() {
         .await;
 }
 
+// TODO(EIP-7732): Remove `#[ignore]` once gloas block production is implemented
```
jimmygchen: block production is enabled, can we remove these ignores now?

beacon_node/http_api/tests/tests.rs
```diff
@@ -8057,6 +8128,34 @@ async fn get_validator_attestation_data_with_skip_slots() {
...
+#[ignore]
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn get_validator_payload_attestation_data() {
+    ApiTester::new()
```
jimmygchen, line 8115: All of these tests create a beacon chain harness before checking if gloas is enabled, and is quite wasteful, we should be able to check the fork name before creating the chain.

beacon_node/http_api/tests/tests.rs
```diff
@@ -8057,6 +8128,34 @@ async fn get_validator_attestation_data_with_skip_slots() {
...
+        .await;
+}
+
+// TODO(EIP-7732): Remove `#[ignore]` once gloas block production is implemented
+#[ignore]
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn get_validator_payload_attestation_data_ssz() {
+    ApiTester::new()
+        .await
+        .test_get_validator_payload_attestation_data_ssz()
```
jimmygchen: these integration tests can all share a single chain and server, i think we can test all payload attestation in one test, and move this call to `get_validator_payload_attestation_data` test

beacon_node/http_api/tests/tests.rs
eserilev: oh i think we need the gloas beacon chain harness

eserilev: ill reword the TODO

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -2097,6 +2097,49 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+        let head = self.head_snapshot();
+        if head.beacon_block.slot() != request_slot {
+            return Err(Error::NoBlockForSlot(request_slot));
+        }
+
+        let beacon_block_root = head.beacon_block_root;
+
+        // TODO(gloas) the spec only requires that the payload be seen. right
+        // now we are setting payload_present to `true` if the envelope was imported
+        // into fork choice which is a stricter requirement than the spec requires.
```
jimmygchen: Why are we doing this?

eserilev: we need some type of cach to track "seen envelopes". we dont have that functionality yet, so this is the next best thing

eserilev: maybe that can be part of the new da cache, not sure

jimmygchen: i think we just check for "seen" rather than validity?
https://github.com/ethereum/consensus-specs/blob/84af1af849c5f32c4909bc58c721e10998f17ac7/specs/gloas/validator.md?plain=1#L340-L343

eserilev: yeah, but we dont have a way to track "seen" yet w/o some type of cache

eserilev: could just default to true for now if you want. i thought it'd be better to check fork choice until we impl some type of seen envelope cache

jimmygchen: ah sorry i didn't see your comments when i posted mine. I was thinking if we could use `EnvelopeTimesCache`?  if the envelope for the root is observed over either gossip / RPC, its considered "seen" right? 

https://github.com/sigp/lighthouse/blob/7dab32dd16c997f592836f9728b2a2b4bf3ba756/beacon_node/beacon_chain/src/envelope_times_cache.rs#L87

jimmygchen: but yeah i think eventually we want to either use the real paylaod cache, or even have `EnvelopeTimesCache` in there to prevent inconsistencies

eserilev: oh i forgot about that cache. and i think im the one that added it lol. yeah I think that could work!

eserilev: Ok were using the `EnvelopeTimesCache` now. And updated the TODO to revisit this in the future

beacon_node/http_api/tests/tests.rs
eserilev: using a single harness now for both json and ssz

beacon_node/http_api/tests/tests.rs
eserilev: checking gloas before the harness is constructed

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: I don’t see the change yet, let me know when you’ve pushed it

consensus/fork_choice/src/fork_choice.rs
```diff
@@ -1484,6 +1484,12 @@ where
             && self.is_finalized_checkpoint_or_descendant(*block_root)
     }
 
+    /// Returns `true` if the payload is known **and** a descendant of the finalized root.
+    pub fn contains_payload(&self, block_root: &Hash256) -> bool {
```
jimmygchen: this is now dead code and can be removed?

jimmygchen: https://github.com/sigp/lighthouse/pull/8497/changes/532ddfb77c798e5799a4db4d7b62eaacf6c8f1db

jimmygchen (approved):
Looks good! Thanks

I've removed the unused code and I think this is good to go 🎉

PR #8498, opened 2025-11-28
fix(peerdb): use start_slot instead of end_slot for safer actions

0xMushow:
## Issue Addressed

None

## Proposed Changes

Discussed in private with @jimmygchen, Lighthouse's `earliest_available_slot` is guaranteed to always align with epoch boundaries, but as a safety implementation, we should use `start_slot` just in case other clients differ in their implementations.

At least we agreed it would be safer for `synced_peers_for_epoch`, I also made the change in `has_good_custody_range_sync_peer`, but this is to be reviewed please.

## Additional Info


jimmygchen (approved):
Yeah good catch, thanks!

PR #8500, opened 2025-11-28
Enforce stricter checks on certain constants

0xMushow:
## Issue Addressed

None

## Proposed Changes

All of these are performing a check, and adding a batch, or creating a new lookup, or a new query, etc..
Hence all of these limits would be off by one.

Example:

```rust
// BACKFILL_BATCH_BUFFER_SIZE = 5
  if self.batches.iter().filter(...).count() >= BACKFILL_BATCH_BUFFER_SIZE {
      return None;  // ← REJECT
  }
  // ... later adds batch via Entry::Vacant(entry).insert(...)
```

Without the `>` being changed to a `>=` , we would allow 6. The same idea applies to all changes proposed.

## Additional Info


jimmygchen (approved):
This is a valid cleanup, however note that changing > to >= reduces the effective runtime  limits by 1 (e.g., batch buffers go from 6 to 5, retries from 4 to 3). Impact should be  negligible but it is a subtle change in production behavior, not just a cleanup.

jimmygchen:
@mergify queue

PR #8505, opened 2025-12-01
Progressive data structures and tests EIP-7916, EIP-8016

michaelsproul:
## Proposed Changes

Run the EF tests for new types from [EIP-7916](https://eips.ethereum.org/EIPS/eip-7916) and [EIP-8016](https://eips.ethereum.org/EIPS/eip-8016):

- [x] `ProgressiveList`
- [x] `ProgressiveBitList`
- [x] Progressive containers
- [x] Compatible unions

The SSZ generic tests are passing if you run them without compiling the rest of Lighthouse:

```
cargo test -p ef_tests --release --features ef_tests -- ssz_generic
```

Everything else is probably broken.

## Additional Info

Depends on:

- https://github.com/sigp/milhouse/pull/82
- https://github.com/sigp/ethereum_ssz/pull/67
- https://github.com/sigp/tree_hash/pull/43

## Known Issues

- [ ] The introduction of `BitList<Progressive>::with_capacity` has created some type inference ambiguities causing `state_processing` not to compile. We should probably rename this method in `ethereum_ssz`.
- [x] Haven't merged `unstable` since the types refactor nor the gloas-containers merge.


michaelsproul:
I've made a bunch of progress on this, but it requires local checkouts of `ethereum_ssz` and `tree_hash` at the moment. These branches:

- https://github.com/sigp/ethereum_ssz/pull/67
- https://github.com/sigp/tree_hash/pull/43

Progressive bitlist tests are currently not passing, need to dig into why. The derive macro for progressive containers with 1 field works though :sweat_smile:

testing/ef_tests/src/cases/ssz_generic.rs
```diff
@@ -312,6 +362,111 @@ struct BitsStruct {
...
+
+#[derive(Debug, Clone, PartialEq, Decode, Encode, TreeHash, Deserialize)]
+#[tree_hash(
+    struct_behaviour = "progressive_container",
+    active_fields(0, 0, 0, 0, 1)
+)]
+#[context_deserialize(ForkName)]
+struct ProgressiveSingleListContainerTestStruct {
+    C: ProgressiveBitList,
+}
```
michaelsproul, line 446: Looking for feedback on the UX of this.

I quite like this UX for progressive containers because it closely follows the spec. We can just copy the `active_fields` vec straight into the struct attributes.

testing/ef_tests/src/cases/ssz_generic.rs
```diff
@@ -312,6 +362,111 @@ struct BitsStruct {
...
+    #[ssz(selector = "2")]
+    #[tree_hash(selector = "2")]
+    B1(ProgressiveSingleListContainerTestStruct),
+    #[ssz(selector = "3")]
+    #[tree_hash(selector = "3")]
+    C1(ProgressiveVarTestStruct),
+    #[ssz(selector = "4")]
+    #[tree_hash(selector = "4")]
+    A2(ProgressiveSingleFieldContainerTestStruct),
+}
```
michaelsproul: The syntax for compatible unions is a bit clunkier at the moment, requiring repetition of the selectors for the sake of `ssz` and `tree_hash`. We also repeat the `enum_behaviour` (this is nothing new, per se).

The reason for the duplication is that derive macros are expected to operate in separate namespaces. It is not necessarily the case that every type implementing `TreeHash` also implements `Encode`/`Decode`, or vice-versa.

We could possibly read _both_ attributes from one of the macros, and error in case of inconsistency?

macladson: Even if we kept the duplication we would still have to check for inconsistency right? Meaning that each macro would have to check for the presence of the other's attribute anyway. At that point it seems reasonable to infer the correct value in the case where one is left out

michaelsproul: Yeah good point, I'll try implementing that and seeing how it looks

michaelsproul: Ok yeah, not bad at all:

- https://github.com/sigp/tree_hash/pull/43/changes/4f7349913ba9ac8cce7112b2279364f8fcac1f5d

Zyra-V21:
Hi @michaelsproul (cc @macladson) !! I've been mapping SSZ progressive-type support across the Rust ecosystem for downstream EL adoption. `alloy` is currently missing `ProgressiveList<Withdrawal>`, which blocks EIP-6465 in Reth, and the same dependency holds for EIP-6404 / EIP-6466. The stack here (`ethereum_ssz#67`, `tree_hash#43`, `milhouse#82`, this PR) covers everything needed, and per your description the `ssz_generic` EF tests are already passing.

Reading the four PRs together, the visible blockers seem to be:

1. Mac's note on `tree_hash#43` about flipping the tree structure to the new spec , has the spec rev landed, and is the rework still pending?
2. The `BitList<Progressive>::with_capacity` type-inference issue you flagged here, plus the `unstable` rebase debt across the four PRs.
3. ~24% patch-coverage gap on `ethereum_ssz#67` (mostly `bitfield/progressive.rs`), and the compile-fail tests you were considering doc-test based.

If priorities pulled you elsewhere and you'd welcome help, I'd be glad to take on the rebase + spec-sync across the four PRs and close out those blockers. I'd do the work as draft PRs from a fork so you keep ownership of the design , happy to squash credit back to you and Mac on merge. Won't touch anything until you say go.

Two questions before I'd start:

1. Is the spec change Mac referenced documented anywhere I can read (consensus-specs commit, EIPs PR, or design discussion)?
2. Are there sigp-internal priorities (gloas rollout, Amsterdam timing) that should shape the order I tackle the four PRs?

Happy to move this to Discord if easier.

michaelsproul:
Hi @Zyra-V21 we've got it covered

michaelsproul:
Closing in favour of:

- https://github.com/sigp/lighthouse/pull/9450

PR #8508, opened 2025-12-01
Instrument attestation signing.

jimmygchen:
## Issue Addressed

We noticed attestation signing taking 2+ seconds on some of our hoodi nodes and the current traces doesn't provide enough details. This PR adds a few more spans to the `attestation_duty_cycle` code path in the VC.

Before: 
image

After:
image

### DIFF
--- a/validator_client/signing_method/src/lib.rs
+++ b/validator_client/signing_method/src/lib.rs
@@ -10,6 +10,7 @@ use reqwest::{Client, header::ACCEPT};
 use std::path::PathBuf;
 use std::sync::Arc;
 use task_executor::TaskExecutor;
+use tracing::instrument;
 use types::*;
 use url::Url;
 use web3signer::{ForkInfo, MessageType, SigningRequest, SigningResponse};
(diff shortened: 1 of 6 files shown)


validator_client/signing_method/src/lib.rs
```diff
@@ -179,13 +181,11 @@ impl SigningMethod {
                 // Spawn a blocking task to produce the signature. This avoids blocking the core
                 // tokio executor.
                 let signature = executor
-                    .spawn_blocking_handle(
-                        move || voting_keypair.sk.sign(signing_root),
-                        "local_keystore_signer",
-                    )
-                    .ok_or(Error::ShuttingDown)?
+                    .spawn_blocking_with_rayon_async(RayonPoolType::HighPriority, move || {
```
pawanjay176: Not sure why we are using rayon here? 🤔

jimmygchen: Oh just trying to isolate the possibility of spwan blocking handle causing the slowness. Going to deploy and see if there’s any differences in performance

jimmygchen: Switching to rayon seems to have improved the p50 numbers quite a bit:

image

with much higher CPU usage:
image

but p99 still looks pretty bad
image

jimmygchen: This trace is quite interesting, looks like `check_and_insert_attestation` is happening sequentially:

image

jimmygchen: We've identified the bottleneck, and will make the fix in a separate PR. I'll revert the rayon change here

michaelsproul (approved):

Issue #8509, opened 2025-12-01
Fix data_columns sorting when reconstructing blobs


0xMushow:
This [recent change](https://github.com/crate-crypto/rust-eth-kzg/pull/421) released in `v0.9.0` of the rust-eth-kzg library means that `data_columns` must be sorted.

Referencing to a similar issue: #7991 

The `reconstruct_blobs` function passes `cell_ids` to `recover_cells_and_compute_kzg_proofs`, which requires cell indices to be in ascending order per the spec. Unlike `reconstruct_data_columns`which already sorts:

https://github.com/sigp/lighthouse/blob/90dd5bb5d7b83aaf886bc94dcb41b9df43f3736a/beacon_node/beacon_chain/src/kzg_utils.rs#L390-L396

`reconstruct_blobs` was iterating over `data_columns` in whatever order they were passed, potentially violating the KZG recovery precondition.

https://github.com/sigp/lighthouse/blob/90dd5bb5d7b83aaf886bc94dcb41b9df43f3736a/beacon_node/beacon_chain/src/kzg_utils.rs#L334-L345

https://github.com/sigp/lighthouse/blob/90dd5bb5d7b83aaf886bc94dcb41b9df43f3736a/beacon_node/beacon_chain/src/kzg_utils.rs#L350

## Additional Info

According to the [specification](https://github.com/ethereum/consensus-specs/blob/master/specs/fulu/polynomial-commitments-sampling.md#recover_cells_and_kzg_proofs), indices passed to `recover_cells_and_compute_kzg_proofs` must be ordered.

```python
# Check that indices are in ascending order
assert cell_indices == sorted(cell_indices)
```

PR #8510, opened 2025-12-01
Fix data columns sorting when reconstructing blobs

0xMushow:
## Issue Addressed

Closes https://github.com/sigp/lighthouse/issues/8509


pawanjay176 (approved):
Nice catch!

PR #8513, opened 2025-12-02
Fix Makefile to avoid git describe error in CI

ackintosh:
## Issue Addressed

Fixes the error `fatal: No names found, cannot describe anything.` that occurs when running `make`
commands in CI (GitHub Actions).

https://github.com/sigp/lighthouse/actions/runs/19839541042/job/56844781126#step:5:13
> fatal: No names found, cannot describe anything.

## Proposed Changes

Changed the `GIT_TAG` variable assignment in the Makefile from immediate evaluation to lazy evaluation:

```diff
- GIT_TAG := $(shell git describe --tags --candidates 1)
+ GIT_TAG = $(shell git describe --tags --candidates 1)
```

This change ensures that git describe is only executed when `GIT_TAG` is actually used (in the `build-release-tarballs` target), rather than on every Makefile invocation.

## Additional Info

Root cause:
- GitHub Actions performs shallow clones by default without fetching git tags
- The previous implementation using `:=` (simple expansion) caused git describe to execute immediately when
 the Makefile was loaded, even for targets that don't use `GIT_TAG` (e.g., make install, make test)


eserilev (approved):
LGTM

PR #8516, opened 2025-12-02
Check slashability of attestations in batches to avoid sequential bottleneck

michaelsproul:
## Issue Addressed

Closes:

- https://github.com/sigp/lighthouse/issues/1914

## Proposed Changes

Sign attestations prior to checking them against the slashing protection DB. This allows us to avoid the sequential DB checks which are observed in traces here:

- https://github.com/sigp/lighthouse/pull/8508#discussion_r2576686107

## Additional Info

This PR builds on:

- https://github.com/sigp/lighthouse/pull/8508

This is a rework of Eitan's PR:

- https://github.com/sigp/lighthouse/pull/6219

I started by trying to resolve merge conflicts, but there were so many breakages I ended up redoing it. I also left out some of the other changes (like the `AttestationDataService`) as we are probably going to introduce a new version of that in the course of implementing the head monitor + consensus service, see:

- https://github.com/sigp/lighthouse/pull/8445


michaelsproul:
This isn't really cutting it in terms of performance, even with the change to prune more frequently.

I think we might need Rayon to cut the signing times:

- https://github.com/sigp/lighthouse/pull/8523

The previous branch with this + Rayon was working well.

michaelsproul:
Self-review complete. This is ready for review and hopefully merge :crossed_fingers:

validator_client/validator_services/src/attestation_service.rs
```diff
@@ -489,9 +472,9 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> AttestationService<S,
                     &[validator_metrics::ATTESTATIONS_HTTP_POST],
                 );
 
-                let single_attestations = attestations
+                let single_attestations = safe_attestations
                     .iter()
-                    .zip(validator_indices)
+                    .zip(validator_indices_ref)
```
jimmygchen: `validator_indices_ref` contains all the attestations to sign, while `safe_attestations` contains only the safe ones to publish (could be less if there is slashable ones), I think this may not work correctly if the vec sizes don't match?

jimmygchen: maybe the `sign_attestations` function can return a tuple of `attestations` and `validator_indices`? similar to the `join_all` return values before

validator_client/lighthouse_validator_store/src/lib.rs
```diff
@@ -747,96 +882,71 @@ impl<T: SlotClock + 'static, E: EthSpec> ValidatorStore for LighthouseValidatorS
...
-    async fn sign_attestation(
-        &self,
-        validator_pubkey: PublicKeyBytes,
-        validator_committee_position: usize,
-        attestation: &mut Attestation<E>,
-        current_epoch: Epoch,
-    ) -> Result<(), Error> {
-        // Make sure the target epoch is not higher than the current epoch to avoid potential attacks.
-        if attestation.data().target.epoch > current_epoch {
-            return Err(Error::GreaterThanCurrentEpoch {
```
jimmygchen, line 760: Looks like we no longer need this error variant and cna be deleted

validator_client/validator_services/src/attestation_service.rs
michaelsproul: Great catch. Ended up going for `Vec<(validator_index, attestation)>` as the return type in:

- 7099fdab1d16241807385921e1906da07501e74a

We can clean it up a bit more when we use `SingleAttestation` internally.

jimmygchen (approved):
Nice, I've done another round of review and this looks solid! Thanks

michaelsproul:
@mergify requeue

jimmygchen:
Seeing improved attestation task times but worse aggregate task times - i think you noticed and mentioned this last time,  just noting it down in case we want to investigate furhter next week @michaelsproul 

image

michaelsproul:
hmm I don't remember seeing this previously, might be worth a look

PR #8522, opened 2025-12-03
Try frequent pruning of slashing protection data

michaelsproul:
## Proposed Changes

Experiment to see if more frequent pruning keeps signing times flatter.


michaelsproul:
Rolled into:

PR #8523, opened 2025-12-03
Sign attestations in the VC using rayon

michaelsproul:
## Proposed Changes

Optimisation to sign attestations using rayon.

## Additional Info

Needs perf testing.


michaelsproul:
Merged into:

PR #8524, opened 2025-12-03
Reduce `eth2` dependency space 

macladson:
## Proposed Changes

Remove certain dependencies from `eth2`, and feature-gate others which are only used by certain endpoints.

| Removed | Optional | Dev only |
| -------- | -------- | -------- |
| `either` `enr` `libp2p-identity` `multiaddr` | `protoarray` `eth2_keystore` `eip_3076` `zeroize` `reqwest-eventsource` `futures` `futures-util` | `rand` `test_random_derive` |

This is done by adding an `events` feature which enables the events endpoint and its associated dependencies.
The `lighthouse` feature also enables its associated dependencies making them optional.

The networking-adjacent dependencies were removed by just having certain fields use a `String` instead of an explicit network type. This means the user should handle conversion at the call site instead. This is a bit spicy, but I believe `PeerId`, `Enr` and `Multiaddr` are easily converted to and from `String`s so I think it's fine and reduces our dependency space by a lot. The alternative is to feature gate these types behind a `network` feature instead.

## Additional Info

Here are some `cargo tree` stats before and after these changes:

```bash
# Before
➜ cargo tree -p eth2 --all-features -e normal | wc -l
906
➜ cargo tree -p eth2 --no-default-features -e normal | wc -l
906

# After
➜ cargo tree -p eth2 --all-features -e normal | wc -l
786
➜ cargo tree -p eth2 --no-default-features -e normal | wc -l
669
```


macladson:
We could also consider feature gating all SSZ related crates/dependencies if we can. The http client would then basically be JSON-only which is probably suitable for many use cases, and might even be required for WASM support

common/eth2/Cargo.toml
```diff
@@ -5,36 +5,33 @@ authors = ["Paul Hauner <paul@paulhauner.com>"]
 edition = { workspace = true }
 
 [features]
-default = ["lighthouse"]
-lighthouse = []
+default = []
+lighthouse = ["proto_array", "eth2_keystore", "eip_3076", "zeroize"]
```
michaelsproul, line 9: Another thing we could do is split the "standard" VC endpoints and Lighthouse VC endpoints. At the moment both are under the same feature as `lighthouse`. Only the standard _VC_ endpoints require `eip_3076`, for example.

common/eth2/src/types.rs
```diff
@@ -2208,7 +2210,8 @@ pub enum ContentType {
     Ssz,
 }
 
-#[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize, Encode, Decode, TestRandom)]
+#[cfg_attr(test, derive(TestRandom))]
```
michaelsproul, line 2213: Just noting that this attribute means we will not be able to use `TestRandom` on `BlobsBundle` in any crates that depend on `eth2`, because dependencies do not get compiled with `cfg(test)`:

- https://github.com/rust-lang/rust/issues/59168

This is fine at the moment, but might need to be adjusted in future. And in general I think we should be wary of putting any actual code behind `cfg(test)`. A separate `cfg(feature = "testing")` is the alternative.

common/eth2/src/types.rs
```diff
@@ -1,13 +1,13 @@
 //! This module exposes a superset of the `types` crate. It adds additional types that are only
 //! required for the HTTP API.
 
+pub use types::*;
```
michaelsproul, line 4: I like that the git blame for this line will now be your username :wink:

michaelsproul (approved):
Looks great! Definitely a step in the right direction!

Issue #4198, opened 2023-04-17
Fix and test fork revert logic


michaelsproul:
## Description

Lighthouse's mechanism for recovering from missed hard forks is currently broken. It is implemented in the `fork_revert` module, but hasn't worked since Altair.

Part of the reason for this is that it is hard to test comprehensively. The ideal test depends on access to _prior versions of Lighthouse_ which aren't available on CI (yet). The test looks something like this:

1. Run a testnet with two types of nodes:
   - Canonical chain: latest Lighthouse version, and fork epoch set for the latest hard fork (e.g. Capella).
   - Stale chain: previous Lighthouse version and _no fork epoch set_.
2. Wait until the testnet has advanced past the configured fork epoch. The canonical chain should continue (and finalize) with new blocks, while the stale chain _also_ continues. The nodes will likely disconnect from each other on P2P.
3. Shut down all the stale nodes and restart them with the latest version of Lighthouse & with the fork epoch configured. Ensure that they don't crash on startup and sync back up to the canonical chain.

There's a lot of local testnet infra described here which is currently not really up to scratch. We likely need changes like #3807 to land first so we can test these scenarios.

There's also a more minor guarantee that we _can_ test on CI without access to prior versions, which is that the current version of Lighthouse can revert a fork missed by the same version. We could likely test this as a beacon_chain test and it would get us ~50% of the way towards a more robust `fork_revert` module.

## TODO

- [ ] Fix fork_revert logic
- [ ] Basic test (single version)
- [ ] Comprehensive test (multiple versions)

danielrachi1:
I'd like to work on this issue. I have a few questions before I start:

Are there any existing tests for the fork revert logic? I didn't find anything under the beacon chain tests. I think I would need to implement at least one of the proposed tests before fixing the `fork_revert` module.

If I understood correctly, I could implement a test that checks if the current version of Lighthouse can revert a fork missed by the same version without having to wait for #3807 to be merged. So writing this (local and for CI) could be a good place to start. What do you think?

michaelsproul:
@danielrachi You're right that there aren't any existing tests for `fork_revert`. A good first step would be to write one, maybe in `beacon_chain/tests`. You're right that you won't need the local testnet stuff from #3807 in order to do it

danielrachi1:
I'm no longer working on this issue. If anyone else is interested, feel free to take it.

danielrachi1:
[...]

PR #8525, opened 2025-12-03
[False start] Fix and test fork revert logic

danielrachi1:
## Issue Addressed

#4198 

## Proposed Changes

- [ ] Add single version test on `testing/simulator/`
- [ ] Add multi version test on `scripts/tests/`
- [ ] Fix the fork revert logic on `beacon_node/beacon_chain/src/fork_revert.rs`

## Additional Info

wip

### DIFF
--- a/testing/node_test_rig/src/lib.rs
+++ b/testing/node_test_rig/src/lib.rs
@@ -33,9 +33,13 @@ const STARTUP_TIMEOUT: Duration = Duration::from_secs(60);
 /// is _local_ to this process).
 ///
 /// Intended for use in testing and simulation. Not for production.
+///
+/// Note: The datadir is persisted and will NOT be automatically cleaned up.
+/// Temporary directories accumulate in /tmp during development.
+/// Clean manually with: `rm -rf /tmp/lighthouse_node_test_rig*`
 pub struct LocalBeaconNode<E: EthSpec> {
     pub client: ProductionClient<E>,
-    pub datadir: TempDir,
+    pub datadir: PathBuf,
 }
 
 impl<E: EthSpec> LocalBeaconNode<E> {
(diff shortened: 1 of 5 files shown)


danielrachi1:
I've been working with the `simulator` package, but I don't think it's the right tool for this task.

The simulator is designed for testing network-level behaviors with multiple nodes. However, this fork revert test requires fine-grained control over individual node lifecycles. Specifically the ability to cleanly shut down nodes, preserve their databases, and restart them with different configurations.

The simulator's architecture makes this challenging because:
- Individual nodes are tightly coupled to their simulated networks
- The `LocalBeaconNode` lifecycle is difficult to control (nodes hold `Arc<BeaconChain>` references that may be retained by background tasks)
- Database connections aren't cleanly released when nodes are removed from the network
- Modifying node behavior leads to increasingly complex workarounds

### Current Status

On the last commit, running the simulation produces a database lock error:

```
thread 'main' panicked at testing/simulator/src/fork_revert_sim.rs:429:41:
called `Result::unwrap()` on an `Err` value: "Failed to restart stale node 0: Unable to open database:
LevelDbError(Error { message: \"IO error: lock /tmp/.../chain_db/LOCK: already held by process\" })"
```

This occurs because:

1. The database must be fully closed before restarting a node with the same datadir
2. Simply dropping `LocalBeaconNode` doesn't guarantee database closure due to `Arc<BeaconChain>` references held by background tasks
3. LevelDB's lock file remains held until all references are dropped

### Next Steps

I'm going to explore Kurtosis as an alternative, which may provide better control over individual node lifecycles. If that doesn't work, I'll either:
- Implement a dedicated test harness for fork revert testing, or
- Significantly refactor the simulator to support clean node shutdown and restart

Closing this PR as a false start. The architecture exploration and node restart infrastructure may still be useful for future work.

PR #8526, opened 2025-12-03
fix: populate participating_peers to enable correct peer penalization

sashass1315:
Backfill intended to penalize only peers that actually participated in serving batches, but `participating_peers` was never populated. This meant no peers were penalized on terminal faulty failure. This change records the block peer on successful batch downloads in on_block_response() and records the peer on error in inject_error() after registering a failed download. This restores the intended behavior without altering APIs or broader flow.


michaelsproul:
Thanks for the PR, please rebase your changes on `unstable` and we can assign someone to review

dapplion:
I think the concept of participating_peers in backfill sync doesn't make sense. If the hash chain is correct and the crypto passes validation that batch can be considered permanently ok. If backfill sync fails later for some reason, peers from previous batches don't need to be penalized. participating_peers existing in backfill sync feels like a copy-pasta from forward sync back in the day.

sashass1315:
> I think the concept of participating_peers in backfill sync doesn't make sense. If the hash chain is correct and the crypto passes validation that batch can be considered permanently ok. If backfill sync fails later for some reason, peers from previous batches don't need to be penalized. participating_peers existing in backfill sync feels like a copy-pasta from forward sync back in the day.

I’ve removed participating_peers from BackFillSync and updated on_batch_process_result so that now only penalize the peer that provided the faulty batch (similar to range sync)

beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -620,29 +609,21 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
                 imported_blocks,
                 penalty,
             } => {
+                // Penalize the peer that provided the faulty batch.
+                network.report_peer(*peer, *penalty, "backfill_faulty_batch");
```
dapplion, line 613: Note: we need to fix this for backfill sync and range sync to penalize the block peer or column(s) peer

dapplion (approved):
LGTM!

dapplion:
Thank you for the contribution, but we would not pursue this now and wait to a reflector of backfill

PR #8540, opened 2025-12-04
Remove `consensus/types` re-exports

macladson:
## Proposed Changes

There are certain crates which we re-export within `types` which creates a fragmented DevEx, where there are various ways to import the same crates. 

```rust
// consensus/types/src/lib.rs
pub use bls::{
    AggregatePublicKey, AggregateSignature, Error as BlsError, Keypair, PUBLIC_KEY_BYTES_LEN,
    PublicKey, PublicKeyBytes, SIGNATURE_BYTES_LEN, SecretKey, Signature, SignatureBytes,
    get_withdrawal_credentials,
};
pub use context_deserialize::{ContextDeserialize, context_deserialize};
pub use fixed_bytes::FixedBytesExtended;
pub use milhouse::{self, List, Vector};
pub use ssz_types::{BitList, BitVector, FixedVector, VariableList, typenum, typenum::Unsigned};
pub use superstruct::superstruct;
```

This PR removes these re-exports and makes it explicit that these types are imported from a non-`consensus/types` crate.

## Additional Information

This PR shouldn't lead to any actual changes in crates' dependencies or compilation, although they will now be explicitly listed in `Cargo.lock`.


pawanjay176 (approved):
LGTM. went through the changes and its as described.

PR #8536, opened 2025-12-04
Move validator http endpoints to a separate module

jimmygchen:
## Issue Addressed

Continuation of:
* #8529 

Moving `/validator` endpoints out of `http_api` to a separation module. This should improve code maintainability, incremental compilation time and rust analyzer performance.

This is a tedious but straight forward change, so we're going with a pair & insta-merge approach to avoid painful & slow async review. @michaelsproul and I paired on the first commit - I believe we are almost done, will pair with @pawanjay176 tomorrow to wrap it up and merge tomorrow. (cc @macladson )

PR #8543, opened 2025-12-05
Move beacon pool http api to its own separate module

jimmygchen:
## Issue Addressed

Continuation of:
* #8536

Moving `/beacon/pool` endpoints out of `http_api` to a separation module. This should improve code maintainability, incremental compilation time and rust analyzer performance.

This is a tedious but straight forward change, so we're going with a pair & insta-merge approach to avoid painful & slow async review


pawanjay176 (approved):

pawanjay176:
Forgot to reque this after mergify recovered. Queuing again now

PR #8550, opened 2025-12-07
Clarify `alloy` dependencies

macladson:
## Proposed Changes

Previously, we had a pinned version of `alloy` to fix some crate compatibility issues we encountered during the migration away from `ethers`. Now that the migration is complete we should remove the pin. This also updates alloy crates to their latest versions. 

## Additional Info

Also, I've moved all version definitions for alloy crates into the top level `Cargo.toml` as it clarifies which alloy crates we are using, even if some of them are only used in a single crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -142,9 +142,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-consensus"
-version = "1.0.42"
+version = "1.1.3"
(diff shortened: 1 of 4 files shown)


pawanjay176 (approved):
LGTM

eserilev (approved):
nice catch!

PR #8553, opened 2025-12-08
fix: remove dead restart_failed_sync from custody backfill

ANtutov:
Make custody backfill restart semantics match BackFillSync. Instead of immediately restarting after a failure, track failed attempts with a new failed_sync flag and only allow restart once a new fully-synced peer joins and sets restart_failed_sync. This prevents tight restart loops on the same bad peers and uses fully_synced_peer_joined from SyncManager to drive intentional retries.

### DIFF
--- a/beacon_node/network/src/sync/custody_backfill_sync/mod.rs
+++ b/beacon_node/network/src/sync/custody_backfill_sync/mod.rs
@@ -113,10 +113,6 @@ pub struct CustodyBackFillSync<T: BeaconChainTypes> {
     /// These are batches that we've skipped because we have no columns to fetch for the epoch.
     skipped_batches: HashSet<BatchId>,
 
-    /// When a custody backfill sync fails, we keep track of whether a new fully synced peer has joined.
-    /// This signifies that we are able to attempt to restart a failed chain.
-    restart_failed_sync: bool,
-
     /// Reference to the beacon chain to obtain initial starting points for custody backfill sync.
     beacon_chain: Arc<BeaconChain<T>>,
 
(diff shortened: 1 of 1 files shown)


beacon_node/network/src/sync/custody_backfill_sync/mod.rs
```diff
@@ -817,11 +810,6 @@ impl<T: BeaconChainTypes> CustodyBackFillSync<T> {
             return Ok(ProcessResult::Successful);
         }
 
-        // Check if we need to restart custody backfill sync due to a cgc change.
-        if self.restart_if_required() {
-            return Ok(ProcessResult::Successful);
-        }
```
eserilev, line 823: what is the reasoning for deleting this?

eserilev:
we probably want to keep `restart_failed_sync` and actually use it correctly, instead of deleting it

beacon_node/network/src/sync/manager.rs
```diff
@@ -575,7 +575,6 @@ impl<T: BeaconChainTypes> SyncManager<T> {
                 // inform the backfill sync that a new synced peer has joined us.
                 if new_state.is_synced() {
                     self.backfill_sync.fully_synced_peer_joined();
-                    self.custody_backfill_sync.fully_synced_peer_joined();
```
eserilev, line 578: i don't think we want to remove this, i think we want to actually get `restart_failed_sync` to work appropriately

ANtutov:
Also squashed tp reduce noise in commits

eserilev:
please read my comment above. We probably want to keep restard_failed_sync and fix its usage instead of deleting it

ANtutov:
> please read my comment above. We probably want to keep restard_failed_sync and fix its usage instead of deleting it

Done. I kept restart_failed_sync and wired it up like in BackFillSync, added a failed_sync flag so custody backfill only restarts after a failure once a new fully-synced peer joins.

eserilev:
have you been able to test these changes at all @ANtutov ?

PR #8558, opened 2025-12-09
Add `network` feature to `eth2`

macladson:
## Proposed Changes

This reverts some of the changes from #8524 by adding back the typed network endpoints with an optional `network` feature. Without the `network` feature, these endpoints (and associated dependencies) will not be built.

This means the `enr`, `multiaddr` and `libp2p-identity` dependencies have returned but are now optional


pawanjay176 (commented):
I'm not sure about this one. 

Do we have any downstream users of this crate that would need the networking types to be the right ones and not strings?

jxs (approved):
> I'm not sure about this one.
> 
> Do we have any downstream users of this crate that would need the networking types to be the right ones and not strings?

I was the one bringing this up in the previous PR to Mac. Since we’re using a typed language, it would be better to rely on the available types instead of Strings where possible right? 😅

pawanjay176 (commented):
Sorry, just realised that this would not matter for `beacon_node` build because we already compile the networking crates for it.

We need to gate the `get_node_identity` function too though, otherwise no feature compilation would fail with
```
error[E0412]: cannot find type `IdentityData` in this scope
    --> common/eth2/src/lib.rs:1939:69
     |
1939 |     pub async fn get_node_identity(&self) -> Result<GenericResponse<IdentityData>, Error> {
     |                                                                     ^^^^^^^^^^^^ not found in this scope
     |
note: found an item that was configured out
    --> common/eth2/src/types.rs:568:12
     |
 566 | #[cfg(feature = "network")]
     |       ------------------- the item is gated behind the `network` feature
```

common/eth2/src/lib.rs
```diff
@@ -1939,6 +1941,7 @@ impl BeaconNodeHttpClient {
     }
 
     /// `GET node/identity`
+    #[cfg(feature = "network")]
     pub async fn get_node_identity(&self) -> Result<GenericResponse<IdentityData>, Error> {
```
macladson, line 1945: Ah nice catch, fixed here

pawanjay176 (approved):

PR #8557, opened 2025-12-09
Fix testnet script

eserilev:
## Issue Addressed
 
 Fix an issue where a kurtosis testnet script was failing because no supernodes were provided

```
There was an error interpreting Starlark code 
Evaluation error: fail: Fulu fork is enabled (epoch: 0) but no supernodes are configured, no nodes have 128 or more validators, and perfect_peerdas_enabled is not enabled. Either configure a supernode, ensure at least one node has 128+ validators, or enable perfect_peerdas_enabled in network_params with 16 participants.
	at [github.com/ethpandaops/ethereum-package/main.star:83:57]: run
	at [github.com/ethpandaops/ethereum-package/src/package_io/input_parser.star:377:17]: input_parser
	at [0:0]: fail
```

PR #8551, opened 2025-12-08
fix: TOCTOU issue in network config

0xmrree:
Fixes TOCTOU issue in beacon node by replacing unsafe port-finding with OS-assigned ports (port 0), ensuring atomic port allocation without race conditions. (Basicly fixing a theoritcal race condition that would prevent the node to start up) Production beacon client is fully fixed with new compute_listen_ports() helper; integration tests retain deprecated approach temporarily pending maintainer input.

## Issue Addressed

Which issue # does this PR address? - https://github.com/sigp/lighthouse/issues/8490

## Changes Introduced

* Fixes TOCTOU vulnerability in beacon node network port allocation by using OS-assigned ports (port 0)
* Adds `compute_listen_ports()` helper function in `listen_addr.rs` to centralize port computation logic for TCP, discovery UDP, and QUIC UDP ports
* Implements overflow-safe port handling
* Adds unit tests covering zero-ports flag, explicit port overrides, and overflow scenarios
* Updates `beacon_node/src/config.rs` to use new helper for IPv4, IPv6, and dual-stack configurations
* Adds integration tests in `lighthouse/tests/beacon_node.rs` to verify `--zero-ports` flag behavior
* Provides safe `bind_*_any()` APIs for future migration of test code (currently unused)
* Retains one of the deprecated functions for `execution_engine_integration` tests pending maintainer guidance - the rest of the old methods were removed

## What's Fixed / Additional Info

* Production beacon node: fully protected against port allocation races
* Integration tests: still use deprecated approach (to be addressed separately) 

## Next steps
I think this PR should be merged as is, but we need to address the CL/EL integration test issues. Here I need help from maintainers. Basically, the EL/CL integration tests are still using the old approach that has the TOCTOU issue because after we start the EL process, we need to pass the ports chosen by the EL downstream, which is not easily possible without parsing stdout of the child process—a fragile approach I'm not sure you want to take. That being said, outside of trying to pass a file descriptor into the CLI of EL, this is the only approach I can think of. 

Let me know your thoughts. The relevant code is here: https://github.com/sigp/lighthouse/blob/stable/testing/execution_engine_integration/src/execution_engine.rs#L55 If you folks want me to go down the path of parsing the stdout of the EL CLI, I will make an additional PR after this is merged.

## Testing 

ran `cargo nextest run -p lighthouse --release` and passed all tests, see below
```
...
PASS [   0.014s] lighthouse::lighthouse_tests validator_manager::validator_exit_using_beacon_and_presign_flags
[...]

PR #8559, opened 2025-12-10
Do not request attestation data when attestation duty is empty

chong-he:
#8171 introduces a bug where the VC would call the BN for attestation_data even though there is no attestation duty. This PR fixes this by adding a check for empty duty. 

Also, I think the check in `sign_and_publish_attestations` is now redundant so I removed it.

Thanks @eserilev for the report!


eserilev (approved):
LGTM!

pawanjay176 (approved):
LGTM. 
@eserilev did this resolve the issue you were seeing?

0xmrree:
FYI the three tests are also failing for my PR: https://github.com/sigp/lighthouse/pull/8551

eserilev:
yep @pawanjay176 this should fix the issue where we are fetching attestation data for every slot regardless of validator duties

chong-he:
> FYI the three tests are also failing for my PR: #8551

This will be fixed by #8557

PR #8560, opened 2025-12-10
perf: remove unnecessary copy in AggregateSignature::serialize

Lil-Duckling-22:
### Motivation
The AggregateSignature::serialize() method in fake_crypto.rs was performing an unnecessary array copy, creating a new array and copying data when it could return self.0 directly. This is inconsistent with other serialize methods in the same file (PublicKey and Signature) and adds unnecessary overhead.

### Solution
Removed the redundant copy operation and return self.0 directly, matching the pattern used in other serialize methods. This improves consistency and eliminates unnecessary memory operations.


michaelsproul:
Please rebase your PR on `unstable` and then we can review

At a glance it looks ok, although it's low impact due to this being the `fake-crypto` module

Issue #8554, opened 2025-12-08
Set trace sampling as the default behavior


eserilev:
## Description

We received a report about high resource consumption with OpenTelemetry enabled. The user also enabled alloy traces + profiling which may have further harmed performance. Regardless of their specific set-up, we should default to trace sampling as it is effective at reducing resource consumption without losing out on observability

https://opentelemetry.io/docs/concepts/sampling/#why-sampling

The default tracing value should be around 1%, and should be configurable via a documented CLI flag

## The current work around

Using these env variables enables trace sampling at 1%

`OTEL_TRACES_SAMPLER=parentbased_traceidratio,OTEL_TRACES_SAMPLER_ARG=0.01`

<img width="826" height="378" alt="Image" src="https://github.com/user-attachments/assets/ad622af2-3c31-4827-9a48-24a61c4c98fb" />

Requiring users to supply these env variables is bad UX, especially since these are not documented within the Lighthouse book. Most users will probably want 1% trace sampling by default

## Error report

> we got a lot of errors and warnings about high load and missed many blocks in our ssv node cluster. It was not 100% clear if it was only because of the setting up of alloy with traces + profiles (my understanding is that it even takes more resources than traces). And of course we run a lot of different client combinations so it was most likely a combination of different problems.

> 2025-12-04 17:34:50.708
error
Dec 04 17:34:50.707 ERROR Block broadcast was delayed                   msg: "system may be overloaded, block may be orphaned", provenance: "builder", delay_ms: 3707, slot: 13170172, root: 0xa517287f765593a6bdde09c53d317540404547319419eeb9ecfd57517c2daf62

> 2025-12-04 15:42:39.330
error
Dec 04 15:42:39.330 ERROR Block was broadcast too late                  msg: "system may be overloaded, block likely to be orphaned", provenance: "builder", delay_ms: 4330, slot: 13169611, root: 0x0f9dfbbd83f3a864358cde181df9a6bcce0691dc9162f74c5a9901cab007fe59

> 2025-12-04 14:09:24.664
error
[...]

PR #8561, opened 2025-12-10
feat: adding `telemetry-trace-sample-rate` cli arg and set default to 1%

0xmrree:
Adding a new `telemetry-trace-sample-rate` cli arg for both VC and BN to adjust sampling rate for OpenTelemetry tracing spans. Default will now be 1%.

https://github.com/sigp/lighthouse/issues/8554

## Proposed Changes

Simply add the cli arg then set the sampler arg in the open telemetry object constructor via below variant  
https://docs.rs/opentelemetry_sdk/0.30.0/opentelemetry_sdk/trace/enum.Sampler.html#variant.TraceIdRatioBased

## Additional Info

Didn't add unit tests for this given low risk and low ROI to added meaningfully test in this case

## Testing

ran cargo nextest run -p lighthouse --release and all tests passed, see below
```
...
        PASS [   0.021s] lighthouse::lighthouse_tests validator_manager::validator_import_defaults
        PASS [   0.022s] lighthouse::lighthouse_tests validator_manager::validator_import_misc_flags
        PASS [   0.013s] lighthouse::lighthouse_tests validator_manager::validator_import_missing_both_file_flags
        PASS [   0.017s] lighthouse::lighthouse_tests validator_manager::validator_import_missing_token
        PASS [   0.014s] lighthouse::lighthouse_tests validator_manager::validator_import_using_both_file_flags
        PASS [   0.019s] lighthouse::lighthouse_tests validator_manager::validator_list_defaults
        PASS [   0.019s] lighthouse::lighthouse_tests validator_manager::validator_move_count
        PASS [   0.019s] lighthouse::lighthouse_tests validator_manager::validator_move_defaults
        PASS [   0.021s] lighthouse::lighthouse_tests validator_manager::validator_move_misc_flags_0
        PASS [   0.021s] lighthouse::lighthouse_tests validator_manager::validator_move_misc_flags_1
        PASS [   0.023s] lighthouse::lighthouse_tests validator_manager::validator_move_misc_flags_2
        PASS [  25.886s] lighthouse::lighthouse_tests beacon_node::validator_monitor_file_flag
        PASS [  26.837s] lighthouse::lighthouse_tests beacon_node::validator_monitor_metrics_threshold_custom
        PASS [  26.960s] lighthouse::lighthouse_tests beacon_node::validator_monitor_metrics_threshold_default
        PASS [  23.081s] lighthouse::lighthouse_tests beacon_node::validator_monitor_pubkeys_flag
        PASS [  47.206s] lighthouse::lighthouse_tests beacon_node::test_builder_disable_ssz_flag
        PASS [  20.507s] lighthouse::lighthouse_tests beacon_node::wss_checkpoint_flag
        PASS [  15.336s] lighthouse::lighthouse_tests beacon_node::zero_ports_flag
────────────
     Summary [ 696.662s] 310 tests run: 310 passed, 0 skipped
```

### DIFF
--- a/book/src/help_general.md
+++ b/book/src/help_general.md
@@ -83,6 +83,12 @@ Options:
           Override the OpenTelemetry service name. Defaults to 'lighthouse-bn'
           for beacon node, 'lighthouse-vc' for validator client, or 'lighthouse'
           for other subcommands.
+      --telemetry-trace-sample-rate <PERCENT>
+          OpenTelemetry trace sampling rate as a percentage (0-100). A value of
+          1 means 1% of traces are sampled. Lower values reduce resource
+          consumption. For more info see
+          https://opentelemetry.io/docs/concepts/sampling/#why-sampling
+          [default: 1]
   -V, --version
           Print version
 
--- a/lighthouse/src/main.rs
+++ b/lighthouse/src/main.rs
@@ -295,6 +295,23 @@ fn main() {
                 .display_order(0)
         )
         .arg(
+            Arg::new("telemetry-trace-sample-rate")
+                .long("telemetry-trace-sample-rate")
+                .value_name("PERCENT")
+                .help(
+                    "OpenTelemetry trace sampling rate as a percentage (0-100). \
+                    A value of 1 means 1% of traces are sampled. \
+                    Lower values reduce resource consumption. \
+                    For more info see https://opentelemetry.io/docs/concepts/sampling/#why-sampling"
+                )
+                .requires("telemetry-collector-url")
+                .value_parser(clap::value_parser!(u8).range(0..=100))
+                .default_value("1")
+                .action(ArgAction::Set)
+                .global(true)
+                .display_order(0)
+        )
+        .arg(
             Arg::new("datadir")
                 .long("datadir")
                 .short('d')
@@ -657,6 +674,7 @@ fn run<E: EthSpec>(
         .eth2_network_config(eth2_network_config)?
         .build()?;
 
+    let mut telemetry_sample_ratio = 0.01; // default to 1%
     if let Some(telemetry_collector_url) = matches.get_one::<String>("telemetry-collector-url") {
         let telemetry_layer = environment.runtime().block_on(async {
             let exporter = opentelemetry_otlp::SpanExporter::builder()
@@ -675,8 +693,19 @@ fn run<E: EthSpec>(
                     _ => "lighthouse".to_string(),
                 });
 
+            // Calculate sample percent as a ratio (percentage / 100)
+            telemetry_sample_ratio = (matches
+                .get_one::<u8>("telemetry-trace-sample-rate")
+                .copied()
+                .unwrap_or(1) as f64)
+                / 100.0;
+            let sampler = opentelemetry_sdk::trace::Sampler::ParentBased(Box::new(
+                opentelemetry_sdk::trace::Sampler::TraceIdRatioBased(telemetry_sample_ratio),
+            ));
+
             let provider = opentelemetry_sdk::trace::SdkTracerProvider::builder()
                 .with_batch_exporter(exporter)
+                .with_sampler(sampler)
                 .with_resource(
                     opentelemetry_sdk::Resource::builder()
                         .with_service_name(service_name)
@@ -824,6 +853,7 @@ fn run<E: EthSpec>(
             let executor = context.executor.clone();
             let mut config = beacon_node::get_config::<E>(matches, &context)?;
             config.logger_config = logger_config;
+            config.telemetry_sample_ratio = telemetry_sample_ratio;
             // Dump configs if `dump-config` or `dump-chain-config` flags are set
             clap_utils::check_dump_configs::<_, E>(matches, &config, &context.eth2_config.spec)?;
 
--- a/beacon_node/client/src/config.rs
+++ b/beacon_node/client/src/config.rs
@@ -79,6 +79,7 @@ pub struct Config {
     pub genesis_state_url: Option<String>,
     pub genesis_state_url_timeout: Duration,
     pub allow_insecure_genesis_sync: bool,
+    pub telemetry_sample_ratio: f64, // for testing only
 }
 
 impl Default for Config {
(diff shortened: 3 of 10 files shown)


lighthouse/src/main.rs
```diff
@@ -675,8 +691,28 @@ fn run<E: EthSpec>(
                     _ => "lighthouse".to_string(),
                 });
 
+            // Parse trace sampling rate, default to 1% (0.01)
+            let sample_rate = matches
+                .get_one::<f64>("telemetry-trace-sample-rate")
```
eserilev: i think we should keep this a usize and allow values between 0 - 100. we try to avoid using f64 in our codebase in general and i dont think we need more granularity here than full percentage points

lighthouse/src/main.rs
```diff
@@ -294,6 +294,22 @@ fn main() {
                 .global(true)
                 .display_order(0)
         )
+        .arg(
+            Arg::new("telemetry-trace-sample-rate")
```
eserilev, line 298: can you please write a test case for this flag. we have some example flag specific tests you can use as an example

lighthouse/src/main.rs
```diff
@@ -294,6 +294,22 @@ fn main() {
...
+                .value_name("RATE")
+                .help(
+                    "Trace sampling rate as a decimal between 0.0 and 1.0. \
+                    A value of 0.01 means 1% of traces are sampled. \
+                    Lower values reduce resource consumption. \
+                    Default: 0.01 (1%)"
+                )
+                .requires("telemetry-collector-url")
+                .value_parser(clap::value_parser!(f64))
+                .action(ArgAction::Set)
```
eserilev, line 310: lets set the default here explicitly via `.default_value("1")`

book/src/help_general.md
```diff
@@ -83,6 +83,9 @@ Options:
           Override the OpenTelemetry service name. Defaults to 'lighthouse-bn'
           for beacon node, 'lighthouse-vc' for validator client, or 'lighthouse'
           for other subcommands.
+      --telemetry-trace-sample-rate <RATE>
```
eserilev: in order to update this file you should actually just run `make cli-local`.

eserilev (approved):
Looks good, just a few minor things and it should be ready for another round of review. Thanks!

eserilev:
tagging as backwards incompatible because the default trace span sampling is reduced from 100% to 1%

lighthouse/src/main.rs
eserilev: I think we can also set a min and max value explicitly at the flag level

lighthouse/src/main.rs
0xmrree: Hmm ok i'm down to convert it to u8 but this needs to be convert to ratio in f64 when we set the variant.

see below 
```
TraceIdRatioBased([f64](https://doc.rust-lang.org/nightly/std/primitive.f64.html))

Sample a given fraction of traces. Fractions >= 1 will always sample. If the parent span is sampled, then it’s child spans will automatically be sampled. Fractions < 0 are treated as zero, but spans may still be sampled if their parent is. Note: If this is used then all Spans in a trace will become sampled assuming that the first span is sampled as it is based on the trace_id not the span_id
```

lighthouse/src/main.rs
0xmrree: sure and we can use the range of value parser for min and max

lighthouse/src/main.rs
0xmrree: gonna use u8 instead of usize given ValueParserFactory does not support usize for range, if its cool hehe

0xmrree:
@eserilev should be good to go for another review. Just used the config in the BN to keep the tests simple and following similar pattern as the other cli flag tests.

lighthouse/src/main.rs
```diff
@@ -675,8 +693,18 @@ fn run<E: EthSpec>(
                     _ => "lighthouse".to_string(),
                 });
 
+            // Calculate sample percent as a ratio (percentage / 100)
+            telemetry_sample_ratio = Some(matches
+                .get_one::<u8>("telemetry-trace-sample-rate")
+                .copied()
+                .unwrap_or(1) as f64 / 100.0);
+            let sampler = opentelemetry_sdk::trace::Sampler::ParentBased(Box::new(
+                opentelemetry_sdk::trace::Sampler::TraceIdRatioBased(telemetry_sample_ratio.unwrap()),
```
eserilev: EDIT: safe math is probably overkill here tbh, i think maybe just removing the Some wrapping and unwrapping is fine

I think we can remove the `Some` wrapping and unwrapping and also introduce safe math here 

maybe something like:

```suggestion
            // Calculate sample percent as a ratio (percentage / 100)
            telemetry_sample_ratio = matches
                .get_one::<u8>("telemetry-trace-sample-rate")
                .copied()
                .unwrap_or(1) as f64.safe_div(100.0).unwrap_or(0.01);
            let sampler = opentelemetry_sdk::trace::Sampler::ParentBased(Box::new(
                opentelemetry_sdk::trace::Sampler::TraceIdRatioBased(telemetry_sample_ratio),
```

you'll also need to add the safe math crate to `lighthouse/Cargo.toml`

eserilev (commented):
Really close! just one small thing and it should be good to go in my opinion

eserilev:
i think you might also need to run `cargo fmt --all` (see the CI failures for more details)

lighthouse/src/main.rs
0xmrree: i think safe_arith only implements traits for integers and not f64, probably because / 0 is handled by f64:inf so ya ill leave / 100 - i would also assume the sdk handles the f64:inf case but we are fine since we are hard coding the 100.0.

but ya I can remove the some on the ratio for sure

0xmrree:
@eserilev changed from Option<f64> to f64, removed unused module, formatted, and ran `cargo nextest run --release -p lighthouse telemetry_sample_rate` - should be good to go

eserilev (approved):
LGTM!

I believe the plan is to get [this](https://github.com/sigp/lighthouse/pull/8647) merged first

As it'll make trace sampling a bit more effective

Ari4ka (approved):

Ari4ka (approved):

Issue #8547, opened 2025-12-05
Duplicated crates: Tracking Issue


macladson:
## Description

This is a tracking issue for crates in Lighthouse where we have to compile multiple versions. I'll try to keep this updated as things change.

## Worst 10 duplicate crates by compile time

Using `cargo build --release --timings` I extracted the following data:

| Rank | Crate | Version | Compile Time | Wasted Time | Dependants |
|------|-------|---------|--------------|-------------|------------|
| **1** | **rustls** | v0.22.4 | 11.7s | **11.7s** | tokio-rustls |
| | | v0.23.35 | | | futures-rustls, hyper-rustls, libp2p-quic, libp2p-tls, quinn, quinn-proto, reqwest, tokio-rustls |
| **2** | **h2** | v0.3.27 | 7.6s | **7.6s** | hyper |
| | | v0.4.12 | | | hyper |
| **3** | **hyper** | v0.14.32 | 6.7s | **6.7s** | warp |
| | | v1.8.1 | | | beacon_node, hyper-rustls, hyper-timeout, hyper-tls, hyper-util, igd-next, mockito, reqwest, tonic, validator_client |
| **4** | **darling_core** | v0.13.4 | 2.0s | **4.0s** | darling, darling_macro |
| | | v0.20.11 | | | darling, darling_macro |
| | | v0.21.3 | | | darling, darling_macro |
| **5** | **syn** | v1.0.109 | 3.7s | **3.7s** | compare_fields_derive, context_deserialize_derive, darling_core, darling_macro, match-lookup, metastruct_macro, neli-proc-macros |
| | | v2.0.110 | | | alloy-rlp-derive, alloy-sol-macro, alloy-sol-macro-expander, alloy-sol-macro-input, alloy-tx-macros, asn1-rs-derive, asn1-rs-impl, async-stream-impl, async-trait, auto_impl, clap_derive, curve25519-dalek-derive, *(+36 more)* |
| **6** | **itertools** | v0.10.5 | 1.3s | **2.6s** | beacon_chain, beacon_node_fallback, beacon_processor, lighthouse_network, metastruct_macro, network, operation_pool, state_processing, store, types |
| | | v0.13.0 | | | ethereum_ssz, milhouse, superstruct |
| | | v0.14.0 | | | alloy-rpc-types-eth, compare_fields, eip4844, ekzg-single-open, prost-derive, ssz_types |
[...]

PR #8563, opened 2025-12-11
Update `syn` to `2.0.110`

macladson:
## Issue Addressed

#8547

## Proposed Changes

We are currently using an older version of `syn` in `test_random_derive`. Updating this removes one of the sources of `syn` `1.0.109` in our dependency tree.


michaelsproul (approved):

michaelsproul:
@mergify requeue

PR #8568, opened 2025-12-11
Remove nested spawns and error handling

### LINKED ISSUE #8520: Reduce number of `spawn`s in VC attestation service
## Description

While reviewing https://github.com/sigp/lighthouse/pull/8171, Jimmy noticed that the attestation service uses `spawn` a lot, which creates a requirement for lots of error handling. These spawns are somewhat unnecessary when tasks don't need to run in parallel (fetching the attestation data), and we could probably get away with _one_ spawn at the top-level, and making the main attestation service function `async`.

Even if we do want parallel execution (e.g. for aggregate production), we can accomplish this using `join_all`.

PoulavBhowmick03:
## Issue Addressed

Fixes #8520 

## Proposed Changes

Added a spawn at the top level, got rid of nested `spawn -> await -> spawn` and made the attestation service function async, and used `join_all` for parallel execution


validator_client/validator_services/src/attestation_service.rs
```diff
@@ -261,50 +256,21 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> AttestationService<S,
...
-                                attestation_data,
-                            ),
-                            "aggregate publish",
-                        );
-                    },
+        // For each committee index for this slot:
+        // Create and publish `SignedAggregateAndProof` for all aggregating validators.
+        let aggregate_futures = aggregate_duties_by_committee_index.into_iter().map(
+            |(committee_index, validator_duties)| {
+                self.clone().handle_aggregates(
```
dapplion, line 263: Since the perform attestation duties is spawn and its completition time doesn't matter you can move the 
`sleep_until(aggregate_production_instant).await;`
outside of before handle_aggregates call to await once

```
let _aggregates_timer = validator_metrics::start_timer_vec(
            &validator_metrics::ATTESTATION_SERVICE_TIMES,
            &[validator_metrics::AGGREGATES],
        );
```

inside `self.produce_and_publish_aggregates ` and then `handle_aggregates` can be dropped to simply code

PoulavBhowmick03: so i should drop the `handle_aggregates` completely, move the timer inside the `produce_and_publish_aggregates` fn and the `sleep_until()` before the previous `handle_aggregates` call, which will now be changed to `self.produce_and_publish_aggregates` directly?

dapplion: It's a suggestion, up to you

macladson (changes requested):
Unfortunately https://github.com/sigp/lighthouse/pull/7892 has changed some of the behaviour in this file so there are some conflicts. Are you able to resolve them? I'll give it a proper review once it's fixed

PoulavBhowmick03:
> Unfortunately #7892 has changed some of the behaviour in this file so there are some conflicts. Are you able to resolve them? I'll give it a proper review once it's fixed

sure, taking a look

Issue #8431, opened 2025-11-19
Investigate not pruning the split block's payload


michaelsproul:
## Description

We presently prune the execution payload of the `split` block itself, while retaining newer payloads. This creates some difficult-to-reason about edge cases, e.g. when we checkpoint sync we expect to be able to load the _full block_ for the head, which is equal to the split. We don't want to prune this payload, and need it to be available when we try to fetch it from the DB. See: https://github.com/sigp/lighthouse/pull/8417#discussion_r2536476202

To make this less confusing, we could avoid ever pruning the payload of the split block. Now that we have the `split.block_root` available, this should be straight-forward. We can modify `try_prune_execution_payloads` with the new behaviour and add new tests (or a codified invariant) to check that the split payload isn't pruned in both aligned and unaligned cases.

Andrurachi:
Hi @michaelsproul , can I work on this one as a follow up of the regression test for the related issue in PR #8458?

My plan would be to:
- Modify try_prune_execution_payloads to explicitly prevent pruning if the block matches the split.block_root.
- Verify the fix using the regression test pattern established in my open PR.

michaelsproul:
Hi @Andrurachi yeah that would be great!

Issue #8426, opened 2025-11-18
Improve tests for unaligned checkpoint sync


michaelsproul:
## Description

We had a latent bug in checkpoint sync from unaligned blocks (skipped slots at epoch boundaries), fixed here:

- https://github.com/sigp/lighthouse/pull/8417#issuecomment-3545062928

I don't know why the existing tests did not catch this:

- `weak_subjectivity_sync_unaligned_advanced_checkpoint`
- `weak_subjectivity_sync_unaligned_unadvanced_checkpoint`

To resolve this issue, we can try reverting the fix made in the linked PR, and writing a test that fails sans fix.

Andrurachi:
Hey! I’d like to take a shot at this if it’s open to outside contributors.
I can dig into the tests and make sure the regression is properly caught after locally reverting the fix.

Mind if I pick this up?

michaelsproul:
@Andrurachi yeah for sure, thanks!

michaelsproul:
Just to keep you posted, we changed the fix slightly to work around some awkward invariants that LH maintains:

- https://github.com/sigp/lighthouse/pull/8417#discussion_r2536476202

PR #8572, opened 2025-12-12
chore: avoid pruning split block execution payloads

Andrurachi:
## **Summary**

This PR ensures that the **split block’s execution payload is never pruned** during weak subjectivity checkpoint sync.
A guard is added to `try_prune_execution_payloads` so that pruning is skipped when `block_root == split_block_root`.

Two new regression tests verify this behavior across aligned and unaligned checkpoint scenarios.

### Code

* Add condition in `try_prune_execution_payloads` to skip pruning for the split block root.

### Tests

* Add helper **`verify_pruning_preserves_payload`** to set up aligned/unaligned checkpoint cases, enable `prune_payloads = true`, run checkpoint sync, and assert the split block payload is preserved.
* Add:

  * `pruning_preserves_payload_aligned`
  * `pruning_preserves_payload_unaligned`

## Relation to Other Work

The new test **`pruning_preserves_payload_unaligned`** covers the failure mode tested by this open PR #8458 and goes further. It verifies both that:

* the BeaconChain initializes successfully (same as the older test)
* the split block’s execution payload is actually present in the store after sync

Because of this, the earlier reproduction test becomes partially redundant, but it can still be useful because it explicitly demonstrates the original bug fixed in #8426 

## **Issue**

Fixes: #8431

### DIFF
--- a/beacon_node/beacon_chain/tests/store_tests.rs
+++ b/beacon_node/beacon_chain/tests/store_tests.rs
@@ -2967,6 +2967,152 @@ async fn reproduction_unaligned_checkpoint_sync_pruned_payload() {
     }
 }
 
+#[tokio::test]
+async fn pruning_preserves_payload_aligned() {
+    type E = MinimalEthSpec;
+    let num_initial_slots = E::slots_per_epoch() * 11;
+    let checkpoint_slot = Slot::new(E::slots_per_epoch() * 9); // Aligned Slot
+
+    let slots = (1..num_initial_slots).map(Slot::new).collect::<Vec<_>>();
+
+    verify_pruning_preserves_payload(checkpoint_slot, slots).await;
+}
+
+#[tokio::test]
+async fn pruning_preserves_payload_unaligned() {
+    type E = MinimalEthSpec;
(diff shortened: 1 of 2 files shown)


Andrurachi:
Hey @michaelsproul, I have a quick question on test coverage.

This PR covers the same regression as the recently merged `reproduction_unaligned_checkpoint_sync_pruned_payload` test (#8458), but with a stronger and more explicit assertion: it directly checks that the split block’s execution payload is preserved (rather than inferring correctness from successful startup and anchor payload availability).

Would you prefer removing the earlier reproduction test to avoid overlapping coverage, or keeping both? Happy to update the PR either way.

michaelsproul:
Hi @Andrurachi, the more tests the better I think.

Sorry for the very slow reply!

Andrurachi:
Hey @michaelsproul, rebased and conflicts resolved! 

I kept both the old reproduction test and the new explicit tests as requested, and updated the helpers to work with the latest APIs. Ready for review whenever you have time :)

dapplion:
Should this PR also cover the case during finality migration? Also the new tests can be folded into existing to save runtime

Issue #8457, opened 2025-11-25
Reduce validator client proposer duties polling with Fulu proposer lookahead


jimmygchen:
The validator client polls for proposer duties every slot, creating unnecessary load on beacon nodes. With Fulu activating in less than two weeks, we can use the new `proposer_lookahead` feature to reduce this.

## Problem

- The VC polls proposer duties [every single slot](https://github.com/sigp/lighthouse/blob/bdfade8e3de8e1d70776b52d348a54ca7cc248cb/validator_client/validator_services/src/duties_service.rs#L572-L595) (every 12 seconds on mainnet)
- Proposer duties don't change within an epoch (reorgs are already handled via the [`dependent_root` check](https://github.com/sigp/lighthouse/blob/bdfade8e3de8e1d70776b52d348a54ca7cc248cb/validator_client/validator_services/src/duties_service.rs#L1491))
- This means 31 out of 32 requests per epoch return identical data
- There's even a [comment acknowledging "0-epoch look-ahead"](https://github.com/sigp/lighthouse/blob/bdfade8e3de8e1d70776b52d348a54ca7cc248cb/validator_client/validator_services/src/duties_service.rs#L1433-L1437) as a limitation

## Proposed Solution

Fulu adds a [`proposer_lookahead` field](https://github.com/sigp/lighthouse/blob/bdfade8e3de8e1d70776b52d348a54ca7cc248cb/consensus/state_processing/src/upgrade/fulu.rs#L21-L37) to the beacon state with proposer indices for `min_seed_lookahead + 1` epochs ahead. We should:

- Poll only at epoch boundaries instead of every slot
- Keep the existing `dependent_root` mechanism for reorgs

## Testing

- Monitor VC logs to confirm polling happens only at epoch boundaries
- Check BN HTTP metrics to verify reduced request frequency
- Verify validators still propose correctly
- Test that reorgs trigger duty refreshes

michaelsproul:
> Use the proposer_lookahead field to cache assignments

I don't understand this point, the `proposer_lookahead` is on the beacon node/beacon state side. It is already being used by the proposer duties endpoint. I think the VC just needs to use the proposer duties endpoint, and can keep its same internal data structures. It can add in polling an epoch in advance post-Fulu.

jimmygchen:
> > Use the proposer_lookahead field to cache assignments
> 
[...]

PR #8574, opened 2025-12-13
Beacon proposer polling to epoch boundaries

NikhilSharmaWe:
## Issue Addressed

#8457

## Proposed Changes

The change modifies the duties service in [validator_client/validator_services/src/duties_service.rs](https://github.com/sigp/lighthouse/compare/unstable...NikhilSharmaWe:lighthouse:unstable?expand=1#diff-6a000d9457a59dd6746f9714121d4b086e63d492d31a821c263c829fb209aa0f) to use `duration_to_next_epoch()` instead of `duration_to_next_slot()`, while maintaining the same safety guarantees through the existing dependent_root check mechanism.


NikhilSharmaWe:
@jimmygchen  Could you please take a look at the changes and share your review?

PR #8585, opened 2025-12-15
fix: remove redundant rpc_custody_column_queue pop

Fibonacci747:
The work selection chain in BeaconProcessor::spawn_manager called rpc_custody_column_queue.pop() twice in a row. The second branch could never change the outcome (the first either returned Some and short-circuited, or both saw an empty queue), so it was effectively dead code. This change removes the redundant branch to make the queue prioritisation logic clearer without altering behaviour.


chong-he:
Closing as trivial and low impact

PR #8591, opened 2025-12-16
Delete `PartialBeaconState`

michaelsproul:
## Proposed Changes

While reviewing Gloas I noticed we were updating `PartialBeaconState`. This code isn't used since v7.1.0 introduced hdiffs, so we can delete it and stop maintaining it :tada:

Similarly the `chunked_vector`/`chunked_iter` code can also go!


pawanjay176 (approved):
My favourite kind of PR 🚀 
Looks like you have caught all references to the `PartialBeaconState`. LGTM

PR #8592, opened 2025-12-16
Tree-sync friendly lookup sync tests

dapplion:
## Issue Addressed

- Step 0 of the tree-sync roadmap https://github.com/sigp/lighthouse/issues/7678

Current lookup sync tests are written in an explicit way that assume how the internals of lookup sync work. For example the test would do:

- Emit unknown block parent message
- Expect block request for X
- Respond with successful block request
- Expect block processing request for X
- Response with successful processing request
- etc..

This is unnecessarily verbose. And it will requires a complete re-write when something changes in the internals of lookup sync (has happened a few times, mostly for deneb and fulu).

What we really want to assert is:

- WHEN: we receive an unknown block parent message
- THEN: Lookup sync can sync that block
- ASSERT: Without penalizing peers, without unnecessary retries 

## Proposed Changes

Keep all existing tests and add new cases but written in the new style described above. The logic to serve and respond to request is in this function `fn simulate` https://github.com/dapplion/lighthouse/blob/2288a3aeb11164bb1960dc803f41696c984c69ff/beacon_node/network/src/sync/tests/lookups.rs#L301 
- It controls peer behavior based on a `CompleteStrategy` where you can set for example "respond to BlocksByRoot requests with empty"
- It actually runs beacon processor messages running their clousures. Now sync tests actually import blocks, increasing the test coverage to the interaction of sync and the da_checker.
- To achieve the above the tests create real blocks with the test harness. To make the tests as fast as before, I disabled crypto with `TestConfig`

Along the way I found a couple bugs, which I documented on the diff.

## Review guide

Look at `lighthouse/beacon_node/network/src/sync/tests/lookups.rs` directly (no diff). 

Other changes are very minor and should not affect production paths


beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -743,6 +766,15 @@ impl<T: BeaconChainTypes> BlockLookups<T> {
...
         } else {
+            // A lookup may be in the following state:
+            // - Block awaiting processing from a different source
+            // - Blobs downloaded processed, and inserted into the da_checker
+            //
+            // At this point the block fails processing (e.g. execution engine offline) and it is
+            // removed from the da_checker. Note that ALL components are removed from the da_checker
+            // so when we re-download and process the block we get the error
+            // MissingComponentsAfterAllProcessed and get stuck.
+            lookup.reset_requests();
```
dapplion, line 782: Bug found on testing, lookups may be stuck given this sequence of events

beacon_node/network/src/sync/manager.rs
```diff
@@ -512,17 +514,18 @@ impl<T: BeaconChainTypes> SyncManager<T> {
...
         self.range_sync.peer_disconnect(&mut self.network, peer_id);
         let _ = self.backfill_sync.peer_disconnected(peer_id);
         self.block_lookups.peer_disconnected(peer_id);
 
+        // Inject a Disconnected error on all requests associated with the disconnected peer
+        // to retry all batches/lookups. Only after removing the peer from the data structures to
+        // sending retry requests to the disconnecting peer.
+        for sync_request_id in self.network.peer_disconnected(peer_id) {
+            self.inject_error(*peer_id, sync_request_id, RPCError::Disconnected);
+        }
```
dapplion, line 527: Minor bug, we need to remove the peer from the sync states (e.g. `self.block_lookups`) then inject the disconnect events. Otherwise we may send requests to peers that are already disconnected. I don't think there's risk of sync getting stuck if libp2p rejects sending messages to disconnected peers, but deserves a fix anyway.

beacon_node/network/src/sync/tests/lookups.rs
```diff
@@ -1158,1561 +1537,697 @@ fn stable_rng() {
...
     );
-    // Should not penalize peer, but network is not clear because of the blocks_by_range requests
-    rig.expect_no_penalty_for(peer_id);
-    rig.assert_ignored_chain(chain_hash);
+    assert_eq!(r.dropped_lookups(), 0, "no dropped lookups");
 }
 
 // Regression test for https://github.com/sigp/lighthouse/pull/7118
 // 8042 UPDATE: block was previously added to the failed_chains cache, now it's inserted into the
 // ignored chains cache. The regression test still applies as the chaild lookup is not created
```
jimmygchen: minor typo 

```suggestion
// ignored chains cache. The regression test still applies as the child lookup is not created
```

beacon_node/beacon_chain/src/chain_config.rs
```diff
@@ -162,6 +172,10 @@ impl Default for ChainConfig {
             invalid_block_roots: HashSet::new(),
             disable_get_blobs: false,
             node_custody_type: NodeCustodyType::Fullnode,
+            test_config: TestConfig {
+                disable_crypto: false,
+                disable_fetch_blobs: false,
```
jimmygchen: Does this currently do anything in the tests, other than skipping the attempt? 
The attempt will always fail because the mock-el does not support `getBlobs` right? 
I made an attempt to add it in #7986, but we didn't end up merging it because it didn't feel useful.

beacon_node/beacon_chain/src/test_utils.rs
```diff
@@ -783,7 +783,13 @@ where
     }
 
     pub fn get_full_block(&self, block_root: &Hash256) -> RpcBlock<E> {
-        let block = self.chain.get_blinded_block(block_root).unwrap().unwrap();
+        let block = self
+            .chain
+            .get_blinded_block(block_root)
+            .unwrap()
+            .unwrap_or_else(|| {
+                panic!("block root does not exist in external harness {block_root:?}")
```
jimmygchen: This isn't always "external" harness

```suggestion
                panic!("block root does not exist in harness {block_root:?}")
```

beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -117,15 +117,23 @@ pub struct BlockLookups<T: BeaconChainTypes> {
...
+    /// Used for testing assertions
+    metrics: BlockLookupsMetrics,
 }
 
 #[cfg(test)]
 use lighthouse_network::service::api_types::Id;
 
 #[cfg(test)]
+#[derive(Debug)]
 /// Tuple of `SingleLookupId`, requested block root, awaiting parent block root (if any),
```
jimmygchen: Please update doc - no longer a tuple and awaiting parent block removed.

beacon_node/network/Cargo.toml
```diff
@@ -58,6 +58,7 @@ gossipsub = { workspace = true }
 k256 = "0.13.4"
 kzg = { workspace = true }
 matches = "0.1.8"
+paste = "1.0.15"
```
jimmygchen:
```suggestion
paste = { workspace = true }
```

Makefile
```diff
@@ -36,8 +36,12 @@ PROFILE ?= release
 RECENT_FORKS_BEFORE_GLOAS=electra fulu
 
 # List of all recent hard forks. This list is used to set env variables for http_api tests
+# Include phase0 to test the code paths in sync that are pre blobs
```
jimmygchen, line 39: We already have `nightly-tests` that runs prior fork tests
https://github.com/sigp/lighthouse/pull/8319

But i just realised it hasn't been activated on the `sigp` fork because github only run scheduled workflows from the main branch (`stable`), we can either wait until the release or have a separate PR to `stable` to activate this.

jimmygchen: Made a PR to activate these nightly tests:
https://github.com/sigp/lighthouse/pull/8636

Makefile
```diff
@@ -226,10 +230,10 @@ test-op-pool-%:
 
 # Run the tests in the `network` crate for all known forks.
 # TODO(EIP-7732) Extend to support gloas by using RECENT_FORKS instead
-test-network: $(patsubst %,test-network-%,$(RECENT_FORKS_BEFORE_GLOAS))
+test-network: $(patsubst %,test-network-%,$(TEST_NETWORK_FORKS))
 
 test-network-%:
-	env FORK_NAME=$* cargo nextest run --release \
+	env FORK_NAME=$* cargo nextest run --no-fail-fast --release \
```
jimmygchen, line 239: Do you want to merge this, or just for your local testing? I think it's fine to not fail fast, as long as the job doens't take forever to run, e.g. beacon-chain-tests

beacon_node/network/src/sync/tests/mod.rs
```diff
@@ -58,17 +65,102 @@ struct TestRig {
...
+    /// Stores all `SyncMessage`s received from `sync_rx`
+    sync_rx_queue: Vec<SyncMessage<E>>,
     /// To send `SyncMessage`. For sending RPC responses or block processing results to sync.
     sync_manager: SyncManager<T>,
     /// To manipulate sync state and peer connection status
     network_globals: Arc<NetworkGlobals<E>>,
     /// Beacon chain harness
     harness: BeaconChainHarness<EphemeralHarnessType<E>>,
+    /// External beacon chain harness to produce blocks that are not imported
+    external_harness: BeaconChainHarness<EphemeralHarnessType<E>>,
```
jimmygchen: Any reason to have this as a field on `TestRig`? I see that it's only used in `build_chain`

beacon_node/network/src/sync/manager.rs
```diff
@@ -512,17 +514,18 @@ impl<T: BeaconChainTypes> SyncManager<T> {
...
-        }
-
         // Remove peer from all data structures
         self.range_sync.peer_disconnect(&mut self.network, peer_id);
         let _ = self.backfill_sync.peer_disconnected(peer_id);
         self.block_lookups.peer_disconnected(peer_id);
 
+        // Inject a Disconnected error on all requests associated with the disconnected peer
+        // to retry all batches/lookups. Only after removing the peer from the data structures to
+        // sending retry requests to the disconnecting peer.
```
jimmygchen: Missing word i think 

```suggestion
        // avoid sending retry requests to the disconnecting peer.
```

Makefile
dapplion: Could we keep this for network tests only? It's just one extra fork and makes it easy to debug and catch errors. For sync tests we should keep the forks that add new objects like run only 
- phase0, deneb, fulu

Makefile
dapplion: The --no-fail-fast gives you more information on CI on which set of tests failed. A single fork run is not that long so we don't save that much time. But the full report is useful

beacon_node/network/src/sync/tests/mod.rs
dapplion: Good find! Moved to build_chain

beacon_node/network/src/sync/block_lookups/mod.rs
dapplion: Done

beacon_node/beacon_chain/src/test_utils.rs
dapplion: Done

beacon_node/beacon_chain/src/chain_config.rs
dapplion: Removed `disable_fetch_blobs` as it does not materially increase speed (I think I added it to make the logs less messy) 


Also applied `disable_crypto` to block production in the TestRig

beacon_node/network/src/sync/tests/lookups.rs
dapplion: Done

beacon_node/network/src/sync/tests/lookups.rs
```diff
@@ -128,583 +242,1052 @@ impl TestRig {
...
 
-    fn test_setup_after_deneb_before_fulu() -> Option<Self> {
-        let r = Self::test_setup();
-        if r.after_deneb() && !r.fork_name.fulu_enabled() {
-            Some(r)
-        } else {
-            None
+    // Network / external peers simulated behaviour
+
+    async fn simulate(&mut self, complete_strategy: CompleteStrategy) {
```
jimmygchen: Would be good to add some brief docs here on what it does and how to use this.

The names `simulate` and `CompleteStrategy` are a bit generic. I thought about `NetworkResponseStrategy` or `PeerBehaviourStrategy` but realised the struct also has non peer behaviour related configuration, like `process_result_conditional` and `block_imported_while_processing`.  I can't think of a better name right now though.

beacon_node/network/src/sync/tests/lookups.rs
```diff
@@ -1,73 +1,177 @@
...
+        self.return_rpc_error = Some(error);
+        self
+    }
+
+    fn no_range_sync(mut self) -> Self {
+        self.skip_by_range_routes = true;
+        self
+    }
+
+    fn process_result<F>(mut self, f: F) -> Self
```
jimmygchen: I think we can improve this name to make it obvious this is implementing builder pattern, rather than actually processing results, could use the standard `with_` prefix.

beacon_node/network/src/sync/tests/lookups.rs
```diff
@@ -1,73 +1,177 @@
...
+    where
+        F: Fn() -> BlockProcessingResult + Send + Sync + 'static,
+    {
+        self.process_result_conditional = Some(Box::new(move |_| Some(f())));
+        self
+    }
+
+    fn block_imported_while_processing(mut self, block_root: Hash256) -> Self {
```
jimmygchen: same as above

beacon_node/beacon_chain/src/test_utils.rs
```diff
@@ -1410,9 +1418,12 @@ where
 
                             let mut agg_sig = AggregateSignature::infinity();
 
-                            agg_sig.add_assign(
-                                &self.validator_keypairs[*validator_index].sk.sign(message),
-                            );
+                            // If disable_crypto is true keep the attestation signature as infinity
+                            if self.chain.config.test_config.disable_crypto {
```
jimmygchen: I think we're missing a `!` here?

beacon_node/beacon_chain/src/test_utils.rs
```diff
@@ -1510,9 +1521,12 @@ where
...
```
jimmygchen: missing a `!` here?

beacon_node/network/src/sync/tests/mod.rs
```diff
@@ -68,7 +77,88 @@ struct TestRig {
...
+        }
+    }
+}
+
+#[derive(Debug)]
+struct SeenLookup {
+    /// Lookup's Id
+    id: Id,
+    block_root: Hash256,
+    max_seen_peers: HashSet<PeerId>,
```
jimmygchen: i think `max_` isn't necessary here, or does it mean something other than `seen peers`?

beacon_node/network/src/sync/tests/lookups.rs
```diff
@@ -1158,1561 +1532,697 @@ fn stable_rng() {
...
+        r.assert_successful_lookup_sync_parent_trigger()
+    } else {
+        r.assert_successful_lookup_sync();
+    }
+}
 
-    // Trigger the request
-    rig.trigger_unknown_block_from_attestation(block_hash, peer_id);
-    let id = rig.expect_block_lookup_request(block_hash);
+/// Assert that sync completes from a GossipUnknownParentBlob / UknownDataColumnParent
```
jimmygchen: typo 

```suggestion
/// Assert that sync completes from a GossipUnknownParentBlob / UnknownDataColumnParent
```

beacon_node/network/src/sync/tests/lookups.rs
```diff
@@ -1158,1561 +1532,697 @@ fn stable_rng() {
...
 }
 
 /// This is a regression test.
-#[test]
-fn test_same_chain_race_condition() {
-    let mut rig = TestRig::test_setup();
+/// Test added in https://github.com/sigp/lighthouse/commit/84c7d8cc7006a6f1f1bb5729ab222b9f85f72727
+/// TODO: This test was added on a very old version of lookup sync. It's unclear if the situation
+/// it wants to recreate is possible or problematic in current code. Skipping.
+#[ignore]
```
jimmygchen: Is this test failing?

beacon_node/network/src/sync/tests/lookups.rs
```diff
@@ -128,583 +242,1052 @@ impl TestRig {
...
+            .completed_lookups
+            - self.initial_block_lookups_metrics.completed_lookups
     }
 
-    fn parent_block_processed_imported(&mut self, chain_hash: Hash256) {
-        self.parent_block_processed(
-            chain_hash,
-            BlockProcessingResult::Ok(AvailabilityProcessingStatus::Imported(chain_hash)),
-        );
+    fn reset_metrics(&mut self) {
```
jimmygchen: `reset_metrics` seems a bit misleading here, perhaps something like `capture_metrics_baseline` is more suitable?

Makefile
jimmygchen: yeah thats reasonable, thanks

beacon_node/network/src/sync/tests/lookups.rs
```diff
@@ -128,583 +242,1052 @@ impl TestRig {
...
+        // Substract initial value to allow resetting metrics mid test
+        self.sync_manager.block_lookups().metrics().dropped_lookups
+            - self.initial_block_lookups_metrics.dropped_lookups
     }
 
-    fn parent_blob_processed(&mut self, chain_hash: Hash256, result: BlockProcessingResult) {
-        let id = self.find_single_lookup_for(self.find_oldest_parent_lookup(chain_hash));
-        self.single_blob_component_processed(id, result);
+    fn completed_lookups(&self) -> usize {
+        // Substract initial value to allow resetting metrics mid test
```
jimmygchen: typo 

```suggestion
        // Subtract initial value to allow resetting metrics mid test
```

beacon_node/network/src/sync/tests/lookups.rs
```diff
@@ -128,583 +242,1052 @@ impl TestRig {
...
+                self.network_rx_queue.push(ev);
+            }
+            while let Ok(ev) = self.beacon_processor_rx.try_recv() {
+                self.beacon_processor_rx_queue.push(ev);
+            }
+            while let Ok(ev) = self.sync_rx.try_recv() {
+                self.sync_rx_queue.push(ev);
+            }
+
+            // Choose at random which queue to process first
```
jimmygchen: I don't fully understand the purpose of deterministically choosing a random queue to process?

jimmygchen (changes requested):
Hey @dapplion 

I've done a round of review and have added a few comments. There's a logic error in the condition and the CI failures that needs to be fixed.

Overall I think the approach is good - offloading complexity to TestRig makes individual tests cleaner. I'm slightly concerned about the growing complexity of `TestRig` itself and how easy it would be to extend it in the future, but I guess this may be needed as sync itself is inheritlenty complex.

beacon_node/network/src/sync/tests/lookups.rs
dapplion: Maybe `SimulateConfig`?

dapplion:
@jimmygchen Thanks for the review! I've addressed all your feedback:

**Already fixed:**
- Documentation, naming improvements (`max_seen_peers` → `seen_peers`, `with_process_result`)
- Removed unused code (`disable_fetch_blobs`, `external_harness` field, random queue selection)
- Fixed panic message and typos

**Just fixed (8d29e73a9):**
- Renamed `import_block_before_process` → `with_import_block_before_process` for builder pattern consistency  
- Fixed incomplete test documentation

All 12 review threads resolved. The 2 remaining threads are my own bug notes.

beacon_node/genesis/src/interop.rs
```diff
@@ -124,7 +124,7 @@ impl<E: EthSpec> InteropGenesisBuilder<E> {
                     signature: Signature::empty().into(),
                 };
 
-                data.signature = data.create_signature(&keypair.sk, spec);
+                data.signature = SignatureBytes::empty();
```
jimmygchen: Double check if we need these changes in this file.
Should we feature gate this?

jimmygchen: Potentialy delete this function in a future PR?

beacon_node/network/src/sync/tests/lookups.rs
```diff
@@ -132,582 +236,1161 @@ impl TestRig {
...
+        assert!(self.created_lookups() > 0, "no created lookups");
+        assert_eq!(self.completed_lookups(), 0, "some completed lookups");
+        assert_eq!(
+            self.dropped_lookups(),
+            self.created_lookups(),
+            "not all dropped. Current lookups {:?}",
+            self.active_single_lookups(),
         );
+        self.expect_empty_network();
+        self.expect_no_active_lookups();
```
jimmygchen: Add documentation for naming conventions, e.g. 

* `.expect_` for mocking / setting up component behaviour
* `.assert_` for making assertion on behaviour

Makefile
```diff
@@ -36,8 +36,12 @@ PROFILE ?= release
 RECENT_FORKS_BEFORE_GLOAS=electra fulu
 
 # List of all recent hard forks. This list is used to set env variables for http_api tests
+# Include phase0 to test the code paths in sync that are pre blobs
 RECENT_FORKS=electra fulu gloas
 
+# For network tests include phase0 to cover genesis syncing (blocks without blobs or columns)
```
jimmygchen, line 42: is this necessary or can we just cover this in the nightly tests?

consensus/state_processing/src/per_block_processing/process_operations.rs
```diff
@@ -641,7 +649,7 @@ pub fn apply_deposit<E: EthSpec>(
     else {
         // The signature should be checked for new validators. Return early for a bad
         // signature.
-        if is_valid_deposit_signature(&deposit_data, spec).is_err() {
+        if verify_signatures.is_true() && is_valid_deposit_signature(&deposit_data, spec).is_err() {
```
jimmygchen: do we need to add the same check to all other places where `is_valid_deposit_signature` is called? or would it be cleaner to use feature so we don't have to pass `VerifySignatures` around and remember to add this check whereever we verify signature?
* https://github.com/sigp/lighthouse/blob/765dcae0f3522621a2b3d81f61209ffcc852230f/consensus/state_processing/src/per_block_processing/process_operations.rs#L889
* https://github.com/sigp/lighthouse/blob/605ef8e8e6cbb9bc974b767bb8a35a445d5a1936/consensus/state_processing/src/per_epoch_processing/single_pass.rs#L396

Makefile
jimmygchen: ah right, you responded above. No objection to this.

dapplion:
The `network-tests` CI failure is a pre-existing flaky test unrelated to this PR's changes:

```
data_column_reconstruction_at_deadline
  left: ["column_reconstruction", "gossip_data_column_sidecar", "worker_freed"]
 right: ["gossip_data_column_sidecar", "worker_freed", "nothing_to_do"]
```

The test has a latent race: on the last loop iteration, enqueuing the final gossip column triggers reconstruction nearly simultaneously. The `assert_event_journal_completes` in the loop expects to drain `gossip_data_column_sidecar` first, but `column_reconstruction` can fire before it. All merge queue runs on unstable pass — the race just surfaces under different CI timing/load. Worth hardening the event ordering assertion in a follow-up.

jimmygchen:
> The `network-tests` CI failure is a pre-existing flaky test unrelated to this PR's changes:
> 
> ```
> data_column_reconstruction_at_deadline
>   left: ["column_reconstruction", "gossip_data_column_sidecar", "worker_freed"]
>  right: ["gossip_data_column_sidecar", "worker_freed", "nothing_to_do"]
> ```
> 
> The test has a latent race: on the last loop iteration, enqueuing the final gossip column triggers reconstruction nearly simultaneously. The `assert_event_journal_completes` in the loop expects to drain `gossip_data_column_sidecar` first, but `column_reconstruction` can fire before it. All merge queue runs on unstable pass — the race just surfaces under different CI timing/load. Worth hardening the event ordering assertion in a follow-up.

are you @dapplion

beacon_node/network/src/sync/tests/lookups.rs
```diff
@@ -1161,1558 +1600,803 @@ fn stable_rng() {
...
+// `crypto_on` tests without the fake_crypto feature and make sure that processing fails, = to
+// assert that signatures and kzg proofs are checked
+#[tokio::test]
+async fn crypto_on_fail_with_invalid_block_signature() {
+    let mut r = TestRig::default();
+    r.build_chain(1).await;
+    r.corrupt_last_block_signature();
+    r.trigger_with_last_block();
+    r.simulate(SimulateConfig::happy_path()).await;
+    if cfg!(feature = "fake_crypto") {
```
jimmygchen, line 2323: Having this condition in the same test feels a bit confusing, as we're testing different outcomes and in the case of `fake_crypto` scenario, it doesn't match what the test name suggests,

I prefer something like:
```rust
  #[tokio::test]
  #[cfg(feature = "fake_crypto")]
  async fn block_signature_not_checked_with_fake_crypto() { ... }

  #[tokio::test]
  #[cfg(not(feature = "fake_crypto"))]
  async fn invalid_block_signature_fails_processing() { ... }
```

but yeah this is a minor style issue

jimmygchen (approved):
Hey @dapplion 

Nice work! This is a great improvement to the existing tests, which is becoming hard to maintain and will break once we change how sync works. 

It looks pretty solid now and we can keep iterating over the test interfaces as we add more tests.

Issue #8571, opened 2025-12-12
Missing values in `/eth/v1/config/spec`


barnabasbusa:
## Description
There are a few values that are missing from the output of `/eth/v1/config/spec` beacon api. 

## Version

latest unstable - 556e917

## Present Behaviour

Currently missing config values are:
`REORG_HEAD_WEIGHT_THRESHOLD` (20)
`REORG_PARENT_WEIGHT_THRESHOLD` (160)
`REORG_MAX_EPOCHS_SINCE_FINALIZATION` (2)
`EPOCHS_PER_SUBNET_SUBSCRIPTION` (256)
`ATTESTATION_SUBNET_COUNT` (64)
`ATTESTATION_SUBNET_EXTRA_BITS` (0)

Missing preset values:
`UPDATE_TIMEOUT` (8192) (diff value for minimal)

Missing domain types:
`DOMAIN_BLS_TO_EXECUTION_CHANGE`  (0x0a000000) 
## Expected Behaviour

All values should be returned with the correct values.

0xmrree:
@michaelsproul yo can i take this?

michaelsproul:
Yeah that'd be great! The reorg config might need relocating from
its current home

0xmrree:
@michaelsproul apologies for delay just starting on this now. I see that for those re org params they are currently be sourced from *cli args* then passed into fork choice algo. Should my pr include refactoring to remove these cli args and then hook up fork choice to ChainSpec which will source the reorg stuff from yaml net configs? - ill assume yes given your comment above

michaelsproul:
I would actually quite like to keep the CLI parameters if we can establish a straight-forward way to have them override the YAML values. Lighthouse pioneered the reorg feature before other clients, which is why we have flags for these parameters, that predate their inclusion in the YAML config.

The ideal behaviour would be:

1. Make these parameters mandatory (or provide defaults) in ChainSpec so they are always present 
2. If no CLI flag is set, use the value from the ChainSpec for configuration of the reorg feature (maybe we keep the fields in two places: ChainSpec and ChainConfig)
3. If a CLI flag is set, have it override the YAML config for the purposes of behaviour (how the node handles reorgs) and for API output

michaelsproul:
I'd be open to doing this change in two PRs. The non-reorg related flags could be in a small standalone PR, and the reorg flags could be in a follow up PR

0xmrree:
[...]

PR #8601, opened 2025-12-18
Update chainspec p2p and etc

0xmrree:
## Issue Addressed

[Missing values in /eth/v1/config/spec #8571
](https://github.com/sigp/lighthouse/issues/8571) - there will be follow up PR for the re org props

## Proposed Changes
1. As per above issue from EF dev ops, I added   
```
"EPOCHS_PER_SUBNET_SUBSCRIPTION": "256",
  "ATTESTATION_SUBNET_COUNT": "64",
  "ATTESTATION_SUBNET_EXTRA_BITS": "0",
  "UPDATE_TIMEOUT": "8192",
  "DOMAIN_BLS_TO_EXECUTION_CHANGE": "0x0a000000"
```
to `/eth/v1/config/spec`
2. Had to change the minimal config for UPDATE_TIMEOUT to get currents tests to pass. This is ok given UPDATE_TIMEOUT is not used in lighthouse as this config for light client spec from altair
3. ATTESTATION_SUBNET_PREFIX_BITS is now dynamically calculated and shimmed into the /eth/v1/config/spec output as advised by @michaelsproul 
## Testing

1. ran cargo nextest run -p types and passed all tests, see below

```
...
        PASS [   0.006s] types withdrawal::withdrawal_request::tests::test_ssz_round_trip
        PASS [   0.007s] types withdrawal::withdrawal_request::tests::test_tree_hash_root
        PASS [   0.010s] types::committee_cache default_values
        PASS [   5.431s] types::committee_cache initializes_with_the_right_epoch
        PASS [   5.489s] types::committee_cache fails_without_validators
        PASS [   5.534s] types::state cache_initialization
        PASS [   5.617s] types::state beacon_proposer_index
        PASS [   0.238s] types::state decode_base_and_altair
        PASS [   5.740s] types::committee_cache min_randao_epoch_correct
        PASS [   6.236s] types::committee_cache shuffles_for_the_right_epoch
        PASS [  19.782s] types::state committees::current_epoch_committee_consistency
        PASS [  19.816s] types::state committees::next_epoch_committee_consistency
        PASS [  19.540s] types::state committees::previous_epoch_committee_consistency
────────────
     Summary [  25.263s] 265 tests run: 265 passed, 0 skipped
```
2. Started beacon node via `./target/release/lighthouse bn --network sepolia ...` and ran 
```bash
curl http://localhost:5052/eth/v1/config/spec | jq '.data | {
  EPOCHS_PER_SUBNET_SUBSCRIPTION,
  ATTESTATION_SUBNET_COUNT,
  ATTESTATION_SUBNET_EXTRA_BITS,
  ATTESTATION_SUBNET_PREFIX_BITS,
  UPDATE_TIMEOUT,
  DOMAIN_BLS_TO_EXECUTION_CHANGE
}'
```
 which returned  
``` json
{
  "EPOCHS_PER_SUBNET_SUBSCRIPTION": "256",
  "ATTESTATION_SUBNET_COUNT": "64",
  "ATTESTATION_SUBNET_EXTRA_BITS": "0",
  "ATTESTATION_SUBNET_PREFIX_BITS": "6",
  "UPDATE_TIMEOUT": "8192",
  "DOMAIN_BLS_TO_EXECUTION_CHANGE": "0x0a000000"
}
```


michaelsproul:
You'll need to rebase on unstable to get CI to pass

michaelsproul:
I'll be OOO until Jan 12 but someone might be able to review

lighthouse/environment/tests/testnet_dir/config.yaml
```diff
@@ -76,6 +76,9 @@ CHURN_LIMIT_QUOTIENT: 65536
 # ---------------------------------------------------------------
 # 40%
 PROPOSER_SCORE_BOOST: 40
+REORG_HEAD_WEIGHT_THRESHOLD: 20
+REORG_PARENT_WEIGHT_THRESHOLD: 160
+REORG_MAX_EPOCHS_SINCE_FINALIZATION: 2
```
chong-he: This should be in another PR for the reorg-related config #8603 

```suggestion
```

consensus/types/src/core/chain_spec.rs
```diff
@@ -1849,9 +1855,13 @@ pub struct Config {
...
-    #[serde(default = "default_attestation_subnet_prefix_bits")]
-    #[serde(with = "serde_utils::quoted_u8")]
-    attestation_subnet_prefix_bits: u8,
+    #[serde(with = "serde_utils::quoted_u64")]
+    epochs_per_subnet_subscription: u64,
+    #[serde(with = "serde_utils::quoted_u64")]
+    attestation_subnet_count: u64,
+    #[serde(with = "serde_utils::quoted_u64")]
+    attestation_subnet_extra_bits: u64,
+    // attestation_subnet_prefix_bits is computed dynamically in ChainSpec, not serialized from config
```
chong-he: Maybe put the comment above the field?

consensus/types/src/core/chain_spec.rs
```diff
@@ -1966,8 +1976,27 @@ fn default_subnets_per_node() -> u8 {
...
+/// the most significant bit of n-1 (which is floor(log2(n-1))) and adding 1 to account for the 2^0 bit.
+/// Therefore, we get floor(log2(n-1)) + 1 = minimum number of bits needed for n values.
+/// This is the same as ceiling(log2(n)), but Rust does not have this function.
+fn compute_min_bits_for_n_values(n: u32) -> u32 {
+    
+    match n {
+        0 => 0,
+        n => (n - 1).ilog2() + 1, // ilog2 = floor(log2(n))
+    }
+}
```
chong-he: Not sure why we need another function for this? Is this because of the test? This could be put in the function `compute_attestation_subnet_prefix_bits` with `n` = the input argument `attestation_subnet_count` itself

Also wonder if we want to use `safe_add` here. From my impression Lighthouse code is quite strict in arithmetic calculations to avoid overflow/underflow. For example, I see another use of `ilog2()` with `safe_add()` here: 

https://github.com/sigp/lighthouse/blob/2ce6b51269708a1c28c69a3241028522ebc153df/consensus/types/src/core/eth_spec.rs#L337-L345

But this is a pretty simple addition, will see what the devs say about this.

common/eth2_network_config/built_in_network_configs/chiado/config.yaml
```diff
@@ -124,7 +124,7 @@ SUBNETS_PER_NODE: 2
 # 2**8 (= 64)
 ATTESTATION_SUBNET_COUNT: 64
 ATTESTATION_SUBNET_EXTRA_BITS: 0
-# ceillog2(ATTESTATION_SUBNET_COUNT) + ATTESTATION_SUBNET_EXTRA_BITS
+# computed at runtime: ceillog2(ATTESTATION_SUBNET_COUNT) + ATTESTATION_SUBNET_EXTRA_BITS
```
chong-he: Would prefer to keep the first comment unchanged, i.e., same as the spec: https://github.com/ethereum/consensus-specs/blob/b47fde2ab66d68dfc12010f99a2e50714c0d94f6/configs/mainnet.yaml#L181
and then maybe we can add a second line of comment mentioning that the `ATTESTATION_SUBNET_PREFIX_BITS` is not read from config.yaml, but computed at runtime

The same applies to others

consensus/types/src/core/chain_spec.rs
```diff
@@ -2942,6 +2995,12 @@ mod yaml_tests {
         MAX_PER_EPOCH_ACTIVATION_CHURN_LIMIT: 8
         CHURN_LIMIT_QUOTIENT: 65536
         PROPOSER_SCORE_BOOST: 40
+        REORG_HEAD_WEIGHT_THRESHOLD: 20
+        REORG_PARENT_WEIGHT_THRESHOLD: 160
+        REORG_MAX_EPOCHS_SINCE_FINALIZATION: 2
```
chong-he: Should this be in #8063?

consensus/types/src/core/preset.rs
```diff
@@ -145,6 +147,7 @@ impl AltairPreset {
             sync_committee_size: E::SyncCommitteeSize::to_u64(),
             epochs_per_sync_committee_period: spec.epochs_per_sync_committee_period,
             min_sync_committee_participants: spec.min_sync_committee_participants,
+            update_timeout: 8192, // light client constant (not used by Lighthouse)
```
chong-he: Can put the comment above the fied

consensus/types/presets/minimal/altair.yaml
```diff
@@ -23,4 +23,4 @@ EPOCHS_PER_SYNC_COMMITTEE_PERIOD: 8
 # 1
 MIN_SYNC_COMMITTEE_PARTICIPANTS: 1
 # SLOTS_PER_EPOCH * EPOCHS_PER_SYNC_COMMITTEE_PERIOD (= 8 * 8)
-UPDATE_TIMEOUT: 64
+UPDATE_TIMEOUT: 8192 # hard-coded to mainnet value for testing given this config is not used in Lighthouse
```
chong-he: The value is 64 for minimal presets: https://github.com/ethereum/consensus-specs/blob/b47fde2ab66d68dfc12010f99a2e50714c0d94f6/presets/minimal/altair.yaml#L23-L24

So the comment may not be necessary

consensus/types/presets/gnosis/altair.yaml
```diff
@@ -22,3 +22,5 @@ EPOCHS_PER_SYNC_COMMITTEE_PERIOD: 512
 # ---------------------------------------------------------------
 # 1
 MIN_SYNC_COMMITTEE_PARTICIPANTS: 1
+# SLOTS_PER_EPOCH * EPOCHS_PER_SYNC_COMMITTEE_PERIOD (= 16 * 512)
+UPDATE_TIMEOUT: 8192
```
chong-he: I may be missing something, but I can't find the `UPDATE_TIMEOUT` field in the gnosis spec: https://github.com/gnosischain/configs/tree/main/mainnet

chong-he (commented):
I have tested the beacon API output and the fields added in this PR are now displayed correctly as the output for mainnet. 

Some comments below. Some CIs are failing too, so would be great if you can fix them, thanks.

consensus/types/src/core/chain_spec.rs
0xmrree: there is no field, that comment is in place of the previously existing attestation_subnet_prefix_bits - just gonna remove given we comment the yaml

consensus/types/src/core/chain_spec.rs
0xmrree: 1. it's just for readability to have two functions, they are unit tested as one.. gonna keep as is if your cool with it
2. great catch about safe add - i will address

common/eth2_network_config/built_in_network_configs/chiado/config.yaml
0xmrree: agree

consensus/types/src/core/chain_spec.rs
0xmrree: yes, that was left over when i had everything in one pr

consensus/types/presets/minimal/altair.yaml
0xmrree: not sure why that makes comment not needed.

0xmrree: will move it above tho

chong-he: What I mean was the value should be 64 for the `minimal` presets (8192 is for mainnet presets). So we should keep the value unchanged

0xmrree: i see

consensus/types/src/core/chain_spec.rs
chong-he: So from my understanding `compute_min_bits_for_n_values` is used to provide the upper integer value of log2(n), i.e., ceillog2(n). 

This seems a bit of complicated, I think we could just use (like the example above): `n.next_power_of_two().ilog2()`, that will also give `ceillog2(n)`

(then we can also get rid of the addition/safe_add)

consensus/types/presets/minimal/altair.yaml
0xmrree: @chong-he also don't waste time reviewing yet still have not got around to fixing ci

consensus/types/presets/gnosis/altair.yaml
chong-he: I am not sure if I am following, I don't see how removing the `UPDATE_TIMEOUT` will cause an issue? It's not defined in the gnosis spec either 

I also see a new test for gnosis presets is added, it doesn't seem necessary

0xmrree: i realized i was wrong - all this stuff is solved with the serde defaults lol - next rev will have no config.yaml value changes

consensus/types/src/core/chain_spec.rs
```diff
@@ -1849,9 +1855,12 @@ pub struct Config {
     #[serde(default = "default_message_domain_valid_snappy")]
     #[serde(with = "serde_utils::bytes_4_hex")]
     message_domain_valid_snappy: [u8; 4],
-    #[serde(default = "default_attestation_subnet_prefix_bits")]
-    #[serde(with = "serde_utils::quoted_u8")]
-    attestation_subnet_prefix_bits: u8,
+    #[serde(with = "serde_utils::quoted_u64")]
+    epochs_per_subnet_subscription: u64,
```
chong-he, line 1870: CI is still failing, you need to do a `cargo fmt --all` and also add a default value here for the newly added fields

consensus/types/src/core/chain_spec.rs
0xmrree: next power of two function is good idea

0xmrree:
@chong-he  addressed the above feedback, ran cargo fmt --all, reran cargo nextest run -p types, and re tested api. the ci failures should be gone now that i have added default values for the added config fields. also the code is much simpler as i am not longer changing any values in the config anymore.

common/eth2_network_config/built_in_network_configs/chiado/config.yaml
chong-he: Erm, why is this back to before?

consensus/types/src/core/chain_spec.rs
```diff
@@ -1162,7 +1168,7 @@ impl ChainSpec {
             resp_timeout: default_resp_timeout(),
             message_domain_invalid_snappy: default_message_domain_invalid_snappy(),
             message_domain_valid_snappy: default_message_domain_valid_snappy(),
-            attestation_subnet_prefix_bits: default_attestation_subnet_prefix_bits(),
+            attestation_subnet_prefix_bits: compute_attestation_subnet_prefix_bits(64, 0),
```
chong-he: I wonder if it is better to put the `default_attestation_subnet_count()` and `default_attestation_subnet_extra_bits()` here when calling the function `compute_attestation_subnet_prefix_bits`?

consensus/types/src/core/chain_spec.rs
```diff
@@ -1426,6 +1433,7 @@ impl ChainSpec {
             inactivity_score_bias: 4,
             inactivity_score_recovery_rate: 16,
             min_sync_committee_participants: 1,
+            update_timeout: 8192,
```
chong-he, line 1439: This is for the `gnosis` config. I don't see it in the gnosis config file: https://github.com/gnosischain/configs/blob/main/mainnet/config.yaml so I think we don't have to put this field for the `gnosis` presets

consensus/types/src/core/chain_spec.rs
```diff
@@ -1528,7 +1538,7 @@ impl ChainSpec {
             message_domain_invalid_snappy: default_message_domain_invalid_snappy(),
             message_domain_valid_snappy: default_message_domain_valid_snappy(),
             max_request_blocks: default_max_request_blocks(),
-            attestation_subnet_prefix_bits: default_attestation_subnet_prefix_bits(),
+            attestation_subnet_prefix_bits: compute_attestation_subnet_prefix_bits(64, 0),
```
chong-he: Same comment as above

consensus/types/src/core/chain_spec.rs
```diff
@@ -1060,6 +1063,7 @@ impl ChainSpec {
             inactivity_score_bias: 4,
             inactivity_score_recovery_rate: 16,
             min_sync_committee_participants: 1,
+            update_timeout: 8192,
```
chong-he, line 1066: Actually I am not sure if we should put `update_timeout` here, because this belongs to the `presets` (which I have seen you put in `consensus/types/src/core/preset.rs`, which is the right place to put). 

I see other presets such as `INACTIVITY_PENALTY_QUOTIENT_ALTAIR` is being put here in the spec, but there are also quite some other presets are not here too. So I am unsure about this. We can leave it here now and wait for a developer's comment on this

chong-he (commented):
This is looking much better, just a bit of minor comments and something that I ain't sure about - will need a dev look into this

Thanks!

common/eth2_network_config/built_in_network_configs/chiado/config.yaml
0xmrree: its not, its 2nd line but it should just say computed at runtime tho

consensus/types/src/core/chain_spec.rs
0xmrree: My understanding is that those altair presets are in the chain spec but they don't come from them, instead there is just a test that reads the altair presets then asserts using from_chain_spec method in the altair preset class to make sure they align with. I followed same pattern as `inactivity_penalty_quotient_altair`. That being said that var is for light client anyway so it wont be used ...

common/eth2_network_config/built_in_network_configs/chiado/config.yaml
chong-he: oh yes, I was confused, sorry

consensus/types/src/core/chain_spec.rs
0xmrree: sure

consensus/types/src/core/chain_spec.rs
0xmrree: `inactivity_penalty_quotient_altair` is not in the gnosis config but yet you still have https://github.com/sigp/lighthouse/blob/unstable/consensus/types/src/core/chain_spec.rs#L1419. Am I missing something? If the attention was so strict to keep unused values out between two cases then there should of been separate structs IMO

0xmrree:
addressed feedback and ready for another review

consensus/types/src/core/chain_spec.rs
chong-he: That's right. I think we can keep it then.

consensus/types/src/core/chain_spec.rs
chong-he: yeah having a closer look, `chain_spec.rs` is the source of truth for the config values (including presets), and also see this in the comment:
https://github.com/sigp/lighthouse/blob/2ce6b51269708a1c28c69a3241028522ebc153df/consensus/types/src/core/chain_spec.rs#L46

so I think we are good

chong-he:
> addressed feedback and ready for another review

This now looks great to me, the beacon API query for `/eth/v1/config/spec` now contains the added 5 fields, thanks for the PR. 

As this involves spec change, we will wait for a final review from the team in case I miss anything.

0xmrree:
@chong-he if you have time it would be great if you could peek at https://github.com/sigp/lighthouse/pull/8613, should be really straight forward

michaelsproul (approved):
Looks good, thanks!

PR #8605, opened 2025-12-19
Remove duplicated `crypto` dependencies

macladson:
## Issue Addressed

#8547 

## Proposed Changes

This unifies the following `crypto` dependencies to a single version each:

- `sha2`
- `hmac`
- `pbkdf2`
- `aes` 
- `cipher` 
- `ctr` 
- `scrypt` 
- `digest` 

## Additional Info

This also turns off the default features of the `zip` crate as only `deflate` is needed.


crypto/eth2_keystore/src/keystore.rs
```diff
@@ -350,7 +351,7 @@ pub fn encrypt(
             // AES Encrypt
             let key = GenericArray::from_slice(&derived_key.as_bytes()[0..16]);
             let nonce = GenericArray::from_slice(params.iv.as_bytes());
-            let mut cipher = AesCtr::new(key, nonce);
+            let mut cipher = Ctr128BE::<Aes128>::new(key, nonce);
```
michaelsproul: How do we know that we were using the big-endian version previously? Is there a spec somewhere?

Cargo.lock
```diff
@@ -96,9 +83,9 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "831010a0f742e1209b3bcea8fab6a8e149051ba6099432c8cb2cc117dec3ead1"
 dependencies = [
  "aead",
- "aes 0.8.4",
- "cipher 0.4.4",
- "ctr 0.9.2",
+ "aes",
```
michaelsproul, line 86: Wanna add some of these to the `deny.toml`? Maybe `aes` and `sha2` at least

crypto/eth2_keystore/src/keystore.rs
macladson: The previous `Aes128Ctr` type used the old `Ctr64BE` type internally. See here:
 https://docs.rs/aes/0.7.5/src/aes/autodetect.rs.html#141

There's probably no reason to change the counter length here so I've updated the code to use `Ctr64BE` instead so it matches the old alias.

crypto/eth2_keystore/src/keystore.rs
```diff
@@ -444,13 +445,14 @@ fn derive_key(password: &[u8], kdf: &Kdf) -> Result<DerivedKey, Error> {
                 params.salt.as_bytes(),
                 params.c,
                 dk.as_mut_bytes(),
-            );
+            )
+            .expect("DKLEN is a valid length for PBKDF2 output");
```
pawanjay176: Why not propagate the error here instead of unwrapping in case the api changes in the future?

pawanjay176 (commented):
Looks good, just a question about the unwrap

crypto/eth2_keystore/src/keystore.rs
macladson: Oh yeah I was being lazy and never went back to fix this, good catch!
(For reference the [old API](https://docs.rs/pbkdf2/0.8.0/src/pbkdf2/lib.rs.html#113-123) had an internal `expect` originally)

crypto/eth2_keystore/src/keystore.rs
pawanjay176: I think the rfc linked in [eip 2335](https://eips.ethereum.org/EIPS/eip-2335#secret-decryption) mentions the endianness required

> The block counter is a 32-bit big-endian integer value

What we use seems to be not exactly in line with the RFC based on this comment here
https://github.com/sigp/lighthouse/blob/ced49dd265e01ecbf02b12073bbfde3873058abe/crypto/eth2_keystore/src/keystore.rs#L43-L57

We are using a 16 byte IV that is split into a 8 byte nonce + 8 byte counter that is stored in big endian format.

pawanjay176 (approved):
LGTM

crypto/eth2_keystore/src/keystore.rs
pawanjay176: There seems to be an existing PR to document this behaviour https://github.com/sigp/lighthouse/pull/8620 , so I'm merging this to avoid getting stale

Issue #7573, opened 2025-06-06
Flaky subnet subscription tests


jimmygchen:
We've been seeing flaky tests in subnet services for quite some time, seems like there are some race there, causing flakiness, We should fix this as it's happening quite frequently.

```
    thread 'subnet_service::tests::test::test_same_subnet_unsubscription' panicked at beacon_node/network/src/subnet_service/tests/mod.rs:358:13:
    assertion `left == right` failed
      left: [Unsubscribe(Attestation(SubnetId(3)))]
     right: []
    note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

https://github.com/sigp/lighthouse/actions/runs/15477117552/job/43575102120

DioBr4nd0:
hey @jimmygchen , Can you assign this to me?

figtracer:
i can take a look at this one if it's still a problem @jimmygchen 

first thing to do would be to try and recreate the error since the logs have expired

jimmygchen:
Hi @figtracer , that would be great! This is still a problem I believe.

figtracer:
still haven't been able to replicate the error, seems to be working fine

ackintosh:
@figtracer I was able to reproduce the error using the following script. You may need to let it run for a while to trigger the failure.

```bash
#!/bin/bash
set -e

for i in {1..1000}
do
  echo "Run #$i"

  cargo test --package network test_same_subnet_unsubscription -- --show-output

  echo "------------------------------"
done
```

figtracer:
> [@figtracer](https://github.com/figtracer) I was able to reproduce the error using the following script. You may need to let it run for a while to trigger the failure.
> 
> #!/bin/bash
> set -e
> 
> for i in {1..1000}
> do
>   echo "Run #$i"
> 
>   cargo test --package network test_same_subnet_unsubscription -- --show-output
> 
>   echo "------------------------------"
> done

will try this out and get back to you, i probably didn't run it enough

figtracer:
i got the error after ~300 runs! @ackintosh

i'm currently running it again with a small fix and the all the tests have passed for now (800 runs in)

ackintosh:
Thanks @figtracer! Just to add one data point: in my case, it sometimes took as many as ~1,500 runs to reproduce the error.

figtracer:
[...]

PR #8609, opened 2025-12-21
Fix flaky test_same_subnet_unsubscription()

figtracer:
## Issue Addressed
#7573 
```bash
thread 'subnet_service::tests::test::test_same_subnet_unsubscription' (1640572) panicked at beacon_node/network/src/subnet_service/tests/mod.rs:356:13:
  left: [Unsubscribe(Attestation(SubnetId(3)))]
 right: []
```

## Why
the race condition seems due to the usage of `is_subscribed()` instead of `is_subscribed_permanent()`, which was not following the comment's desired behaviour since `is_subscribed()` considers both permanent and temporary subscriptions.

```rust
// If we are permanently subscribed to this subnet, we won't see a subscribe message
```

so, previously, we could be taking the `if` branch for a temporary subscription
```rust
        if subnet_service.is_subscribed(&Subnet::Attestation(subnet_id1)) {
            // If we are permanently subscribed to this subnet, we won't see a subscribe message
            let _ = get_events_until_num_slots(&mut subnet_service, None, 1).await;
```
discarding `Subscribe` and potentially `Unsubscribe` events; thus, when reaching

```rust
        // Get event for 1 more slot duration, we should get the unsubscribe event now.
        let unsubscribe_event = get_events_until_num_slots(&mut subnet_service, None, 1).await;
```

the expected `Unsubscribe` could have been discarded already, resulting in [].

## Fix
refactored the test to use a more deterministic approach following the other tests and removed `is_subscribed()` since we take the same route by checking `is_subscribed_permanent()` first

## Results
```bash
Run #2000
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.77s
warning: the following packages contain code that will be rejected by a future version of Rust: num-bigint-dig v0.8.4
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 1`
     Running unittests src/lib.rs (target/debug/deps/network-ecf0aa6b04bd6e2d)

running 1 test
test subnet_service::tests::test::test_same_subnet_unsubscription ... ok

successes:

successes:
    subnet_service::tests::test::test_same_subnet_unsubscription

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 71 filtered out; finished in 6.11s
```

### DIFF
--- a/beacon_node/network/src/subnet_service/tests/mod.rs
+++ b/beacon_node/network/src/subnet_service/tests/mod.rs
@@ -338,7 +338,7 @@ mod test {
         // Unsubscription event should happen at slot 2 (since subnet id's are the same, unsubscription event should be at higher slot + 1)
         let expected = SubnetServiceMessage::Subscribe(Subnet::Attestation(subnet_id1));
 
-        if subnet_service.is_subscribed(&Subnet::Attestation(subnet_id1)) {
+        if subnet_service.is_subscribed_permanent(&Subnet::Attestation(subnet_id1)) {
             // If we are permanently subscribed to this subnet, we won't see a subscribe message
             let _ = get_events_until_num_slots(&mut subnet_service, None, 1).await;
         } else {
--- a/beacon_node/network/src/subnet_service/tests/mod.rs
+++ b/beacon_node/network/src/subnet_service/tests/mod.rs
@@ -356,6 +356,9 @@ mod test {
             assert_eq!([expected], unsubscribe_event[..]);
         }
 
+        // Ensure the subscription has been fully removed
+        let _ = get_events_until_num_slots(&mut subnet_service, None, 0).await;
+
         // Should  no longer be subscribed to any short lived subnets after unsubscription.
         assert_eq!(subnet_service.subscriptions().count(), 0);
     }
--- a/beacon_node/network/src/subnet_service/tests/mod.rs
+++ b/beacon_node/network/src/subnet_service/tests/mod.rs
@@ -335,31 +335,27 @@ mod test {
         // submit the subscriptions
         subnet_service.validator_subscriptions(vec![sub1, sub2].into_iter());
 
-        // Unsubscription event should happen at slot 2 (since subnet id's are the same, unsubscription event should be at higher slot + 1)
-        let expected = SubnetServiceMessage::Subscribe(Subnet::Attestation(subnet_id1));
+        let subnet = Subnet::Attestation(subnet_id1);
 
-        if subnet_service.is_subscribed_permanent(&Subnet::Attestation(subnet_id1)) {
-            // If we are permanently subscribed to this subnet, we won't see a subscribe message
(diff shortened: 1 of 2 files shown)


beacon_node/network/src/subnet_service/tests/mod.rs
```diff
@@ -335,28 +335,27 @@ mod test {
...
-        // Unsubscription event should happen at slot 2 (since subnet id's are the same, unsubscription event should be at higher slot + 1)
-        let expected = SubnetServiceMessage::Subscribe(Subnet::Attestation(subnet_id1));
+        let subnet = Subnet::Attestation(subnet_id1);
 
-        if subnet_service.is_subscribed(&Subnet::Attestation(subnet_id1)) {
-            // If we are permanently subscribed to this subnet, we won't see a subscribe message
-            let _ = get_events_until_num_slots(&mut subnet_service, None, 1).await;
+        // If permanently subscribed, no Subscribe/Unsubscribe events will be generated
+        if subnet_service.is_subscribed_permanent(&subnet) {
+            let _ = get_events_until_num_slots(&mut subnet_service, None, 3).await;
```
ackintosh, line 342: To make the test more robust, I think we should check that the event returned from `get_events_until_num_slots` is empty.

beacon_node/network/src/subnet_service/tests/mod.rs
```diff
@@ -335,28 +335,27 @@ mod test {
...
+                SubnetServiceMessage::Subscribe(subnet),
+                SubnetServiceMessage::Unsubscribe(subnet),
+            ];
+            let events = get_events_until_num_slots(
+                &mut subnet_service,
+                Some(2),
+                (MainnetEthSpec::slots_per_epoch()) as u32,
+            )
+            .await;
+            assert_eq!(events, expected);
```
ackintosh, line 355: The original intent here (before this fix) was to check which event we get *at each slot*.  
So I think it would be preferable to check the events slot by slot, for example like this:

```rust
let subscribe_event = get_events_until_num_slots(&mut subnet_service, Some(1), 1).await;
assert_eq!(subscribe_event, expected);

let unsubscribe_event = get_events_until_num_slots(&mut subnet_service, Some(1), 1).await;
assert_eq!(unsubscribe_event, expected);
```

ackintosh (commented):
Thanks, @figtracer ! I've left a few comments.

ackintosh:
@figtracer I understand you might be busy, so I can open another PR that includes the changes I mentioned and switches the base branch from `stable` to `unstable`. It would be based on this PR, so all credit would go to you.

Please let me know if you're planning to make further updates to this PR. Thanks!

figtracer:
> @figtracer I understand you might be busy, so I can open another PR that includes the changes I mentioned and switches the base branch from `stable` to `unstable`. It would be based on this PR, so all credit would go to you.
> 
> Please let me know if you're planning to make further updates to this PR. Thanks!

sorry for dropping this! got caught up in other work

PR #8610, opened 2025-12-22
perf(bls): remove redundant clone in aggregate verify methods

avorylli:
Remove unnecessary clone() before to_signature() in fast_aggregate_verify andaggregate_verify. 
The method works on a reference, so cloning is redundant.

Aligns with other usages in the file and avoids unnecessary allocations.


chong-he:
Closing as trivial / low impact

PR #8611, opened 2025-12-22
fix: rename txn to cursor in Cursor::delete_while match arms

mattvaux:
Noticed the delete_while method was using `txn` in the match arms while all other Cursor methods use `cursor`.


chong-he:
Closing as trivial / low impact

PR #8603, opened 2025-12-19
Reorg Config Spec Modifications

0xmrree:
## Issue Addressed
Need to source` REORG_HEAD_WEIGHT_THRESHOLD`, `REORG_PARENT_WEIGHT_THRESHOLD`, `REORG_MAX_EPOCHS_SINCE_FINALIZATION` from config.yaml and include in spec API

Which issue # does this PR address?
https://github.com/sigp/lighthouse/issues/8571
This is 2/2 PR to address the above issue, the other is https://github.com/sigp/lighthouse/pull/8601

## Proposed Changes

as advised by feedback 
1. If cli props `proposer-reorg-threshold`, `proposer-reorg-parent-threshold`, or `proposer-reorg-epochs-since-finalization` are provided override the config from the config.yaml
2.  Add `REORG_HEAD_WEIGHT_THRESHOLD`, `REORG_PARENT_WEIGHT_THRESHOLD`, `REORG_MAX_EPOCHS_SINCE_FINALIZATION` to /eth/v1/config/spec

## Additional Info

1. I considered possible removing the cli prop `disable-proposer-reorgs` but after going back to phase0 spec I can see that its language implies that calling get_proposer_head is optional thus I'm leaving it as is. 
2. @michaelsproul suggested I make the re org props in chainspec not optional but I left it as is so I did not have to change the get_proposer_head logic in the consensus code. 

## Testing

in addition to unit tests, curled the spec api and got 
```
{
...
  "REORG_HEAD_WEIGHT_THRESHOLD": "20",
  "REORG_PARENT_WEIGHT_THRESHOLD": "160",
  "REORG_MAX_EPOCHS_SINCE_FINALIZATION": "2"
...
}
```

PR #8613, opened 2025-12-29
Add rest of missing specs to spec api

0xmrree:
## Issue Addressed

[8571](https://github.com/sigp/lighthouse/issues/8571)

## Proposed Changes

Did a quick inventory and found that the following was in the spec config at https://github.com/ethereum/consensus-specs/blob/master/configs/mainnet.yaml but not in lighthouse's `/eth/v1/config/spec`. This is of course not including the configs added via: [8603](https://github.com/sigp/lighthouse/pull/8603) and [8601](https://github.com/sigp/lighthouse/pull/8601)

 SLOT_DURATION_MS
 PROPOSER_REORG_CUTOFF_BPS
 ATTESTATION_DUE_BPS
 AGGREGATE_DUE_BPS
 CONTRIBUTION_DUE_BPS
 SYNC_MESSAGE_DUE_BPS

## Additional Info
Surprisingly, even tho the above configs in the ChainSpec struct, none of them are actually being used. For now adding them to spec API would be good to keep things consistent across clients I would assume. Also these values from the chain spec are not being read from the config - I did not make it so, such that I don't have to change the gnosis config.yaml 

## Testing
unit tests and ran  
```
curl http://localhost:5052/eth/v1/config/spec | jq '.data | {
 SLOT_DURATION_MS,
 PROPOSER_REORG_CUTOFF_BPS,
 ATTESTATION_DUE_BPS,
 AGGREGATE_DUE_BPS,
 CONTRIBUTION_DUE_BPS,
 SYNC_MESSAGE_DUE_BPS
}'
```


0xmrree:
cc: @eth2353

chong-he:
The added fields are pretty new to the spec and I think they are being implemented in #7944 ?

0xmrree:
@chong-he https://github.com/sigp/lighthouse/pull/7944/ adds `slot_duration_ms` to ChainSpec from config.yaml but does not do the same for the bps props. @eserilev should I do a new rev for this PR to remove slot_duration and then hook up the bps props from the config.yaml to ChainSpec? - or just cancel this PR.

0xmrree:
closing as stale

PR #8614, opened 2025-12-29
perf: remove allocations from merkle tree proof verification logic

kevaundray:
## Issue Addressed

This PR does two small things:

- Removes the allocations that were happening on each loop
- Makes it more explicit that the bit in the index is only being used to specify the order of the inputs for the hash function

## Additional Info

Benchmarked on Apple M4 Max and there is a 33% performance increase, which concretely amounted to ~1.2 microseconds savings for depth=32.
Not a lot, so I think the main performance benefit would be the reduction in _possible_ allocations that would have happened during the for loop.

### DIFF
--- a/consensus/merkle_proof/src/lib.rs
+++ b/consensus/merkle_proof/src/lib.rs
@@ -1,4 +1,4 @@
-use ethereum_hashing::{ZERO_HASHES, hash, hash32_concat};
+use ethereum_hashing::{ZERO_HASHES, hash32_concat};
 use safe_arith::ArithError;
 use std::sync::LazyLock;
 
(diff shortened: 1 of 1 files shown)


consensus/merkle_proof/src/lib.rs
```diff
@@ -382,20 +382,19 @@ pub fn verify_merkle_proof(
 pub fn merkle_root_from_branch(leaf: H256, branch: &[H256], depth: usize, index: usize) -> H256 {
     assert_eq!(branch.len(), depth, "proof length should equal depth");
 
-    let mut merkle_root = leaf.as_slice().to_vec();
+    let mut merkle_root = leaf.0;
```
kevaundray, line 385: Output of the hash function is a H256/[u8;32] so we can have merkle_root be a [u8;32] and skip the heap allocations on each loop

consensus/merkle_proof/src/lib.rs
```diff
@@ -382,20 +382,19 @@ pub fn verify_merkle_proof(
 pub fn merkle_root_from_branch(leaf: H256, branch: &[H256], depth: usize, index: usize) -> H256 {
     assert_eq!(branch.len(), depth, "proof length should equal depth");
```
kevaundray, line 383: I skipped it, but its possible to add another invariant here where the number of bits in the `index` should be more than the depth.

This should never trigger unless we are for example on a 16 bit architecture and the tree depth is something like 32/64. Though I imagine that others things assume usize is u32/u64, so they would break well before we reach this line

michaelsproul (approved):
Looks great, nice improvement thanks!

jimmygchen (approved):

PR #8623, opened 2026-01-05
Revised log when all validators have exited

chong-he:
Lighthouse VC logs `Awaiting activation` as long as no validators are active. This includes the case of validators pending activation or all validators have exited the beacon chain. This log is confusing in the latter case when all validators have exited (reported a few times on Discord).

This PR checks the validator status when no validators are active, and logs a different one.

Not sure if there is a cleaner way to do this? I am open to suggestions

Edit: resolve to tweaking the log, see comment below

### DIFF
--- a/validator_client/validator_services/src/notifier_service.rs
+++ b/validator_client/validator_services/src/notifier_service.rs
@@ -146,7 +146,7 @@ pub async fn notify<S: ValidatorStore, T: SlotClock + 'static>(
                 validators = total_validators,
                 %epoch,
                 %slot,
-                "Awaiting activation"
+                "All validators inactive"
             );
         }
     } else {
--- a/validator_client/validator_services/src/notifier_service.rs
+++ b/validator_client/validator_services/src/notifier_service.rs
@@ -165,9 +165,7 @@ pub async fn notify<S: ValidatorStore, T: SlotClock + 'static>(
 
                 match validator_data {
                     Ok(Some(response)) => validator_status.push(response.data.status),
-                    Ok(None) => {
-                        error!("Validator not found in beacon chain")
-                    }
+                    Ok(None) => {}
                     Err(e) => {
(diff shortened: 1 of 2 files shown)


validator_client/validator_services/src/notifier_service.rs
```diff
@@ -142,12 +145,52 @@ pub async fn notify<S: ValidatorStore, T: SlotClock + 'static>(
...
-            );
+            let pubkeys: Vec<PublicKeyBytes> = duties_service
+                .validator_store
+                .voting_pubkeys(DoppelgangerStatus::ignored);
+
+            let mut validator_status = Vec::new();
+            for pubkey in pubkeys {
+                let validator_data = duties_service
+                    .beacon_nodes
+                    .first_success(|beacon_node| async move {
```
michaelsproul: It's not ideal to make another batch of HTTP requests here, which could take a long time.

Instead I think we should track which validators have exited in the `DutiesService`.

michaelsproul (changes requested):
This covers the case where all validators are exited, but it might be nice to also cover the case where some are exited. I wonder if we could easily add a method to `DutiesService` called `exited_count`, which we could use for choosing between `Awaiting activation` and `All validators have exited`, _and_ for providing additional info on `Some validators active`/`All validators active`.

validator_client/validator_services/src/notifier_service.rs
chong-he: As per our discussion, the current `DutiesService` calls the endpoint `get_beacon_states_validator_id` only when the VC starts (or has un-cached validator index). 

To get updated info about whether a validator has exited, it requires calling endpoint, which burdens the BN for a small UX improvement, and it is also tricky to determine when to call the endpoint. So we resolve to tweaking the log to cover the case where validators are not activated yet and have all exited

michaelsproul (approved):

Issue #5019, opened 2023-12-19
Validator client HTTP API wouldn't start without a beacon node


chong-he:
## Description

Starting `lighthouse vc –http` alone wouldn’t start the validator client HTTP API. HTTP API queries will get connection refused. 

Once `lighthouse bn –http` is started, only then the validator client logs `INFO HTTP API started` and HTTP API queries works. If the bn is stopped afterwards, the validator client HTTP API calls still work.

From @michaelsproul, the reason being that the validator client waits to fetch genesis from the BN on start-up here: https://github.com/sigp/lighthouse/blob/441fc1691b69f9edc4bbdc6665f3efab16265c9b/validator_client/src/lib.rs#L384-L388

## Version

Lighthouse v4.5.0-441fc16

## Present Behaviour

Starting `lighthouse vc –http` alone wouldn’t start the validator client HTTP API

## Expected Behaviour

Starting `lighthouse vc –http` alone should start the validator client HTTP API

## Steps to resolve

From @michaelsproul, a possible solution is to refactor that so that the validator client just gets those values (or from the hardcoded genesis state in the network config)

michaelsproul:
Related:

- https://github.com/sigp/lighthouse/pull/4714

I don't think that PR caused a regression, but it didn't fully solve the issue (in the offline BN case)

Nanle100:
is this issue still open?
I'd love to walk on it

chong-he:
> is this issue still open? I'd love to walk on it

Yes it's still open. 

The reason that VC waits for the beacon HTTP is to get the genesis and chain info so that VC can function. So the solution is to how to let the VC figuring out how to get those info without the beacon node

Nanle100:
Okay!
let me try to work this issues

jimmychu0807:
Hi @chong-he I am new to lighthouse dev and it seems this issue has not progressed much. May I pick up on this?

Could you also provide some guidance while I am on the way of solving this issue? Thanks.

chong-he:
> Hi [@chong-he](https://github.com/chong-he) I am new to lighthouse dev and it seems this issue has not progressed much. May I pick up on this?
> 
> Could you also provide some guidance while I am on the way of solving this issue? Thanks.

[...]

PR #8638, opened 2026-01-08
chore(validator_client): Read genesis time and genesis validators root from eth2_network_config

jimmychu0807:
## Issue Addressed

#5019 

## Proposed Changes

If there is a known eth2_network_config, we read the genesis time and validators root from the config.

## Other

@chong-he I made the code update, pretty much following [your comment](https://github.com/sigp/lighthouse/issues/5019#issuecomment-3708653511). Please take a look. Also, let me know if there is any protocol that I should follow when making PR to lighthouse repo. Thanks.

### DIFF
--- a/common/eth2_network_config/src/lib.rs
+++ b/common/eth2_network_config/src/lib.rs
@@ -135,10 +135,9 @@ impl Eth2NetworkConfig {
 
     /// The `genesis_time` of the genesis time.
     pub fn genesis_time<E: EthSpec>(&self) -> Result<Option<u64>, String> {
-        Ok(
-            self.get_genesis_state_from_bytes::<E>()
-                .map(|state| Some(state.genesis_time()))?
-        )
+        Ok(self
+            .get_genesis_state_from_bytes::<E>()
+            .map(|state| Some(state.genesis_time()))?)
     }
 
     /// The `genesis_validators_root` of the genesis state.
--- a/common/eth2_network_config/src/lib.rs
+++ b/common/eth2_network_config/src/lib.rs
@@ -135,9 +135,8 @@ impl Eth2NetworkConfig {
 
     /// The `genesis_time` of the genesis time.
     pub fn genesis_time<E: EthSpec>(&self) -> Result<Option<u64>, String> {
-        Ok(self
-            .get_genesis_state_from_bytes::<E>()
-            .map(|state| Some(state.genesis_time()))?)
+        self.get_genesis_state_from_bytes::<E>()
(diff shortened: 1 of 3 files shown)


chong-he:
Hey @jimmychu0807 , I see your message on Discord but thought I will reply here as it is PR-related. 

The CLA doesn't look like it is signed, can you do that again? It would be great if you can fix the CI failing too, thanks

jimmychu0807:
@chong-he cargo fmt'ed

jimmychu0807:
@chong-he I have run most of the code in test-suite.yml/check-code step by hand and they pass now. Hopefully now it will pass all the CI test.

In future, maybe I should PR to my own repo first to let the CI run.

common/eth2_network_config/src/lib.rs
```diff
@@ -133,6 +133,12 @@ impl Eth2NetworkConfig {
         self.genesis_state_source != GenesisStateSource::Unknown
     }
 
+    /// The `genesis_time` of the genesis time.
+    pub fn genesis_time<E: EthSpec>(&self) -> Result<Option<u64>, String> {
+        self.get_genesis_state_from_bytes::<E>()
```
michaelsproul: I wonder if we should hardcode the `genesis_time` into the `Url` variant of `GenesisStateSource`, like we do for the genesis validators root. Then this change will work for networks like Hoodi where we do not bundle the full genesis state.

michaelsproul: I think this would be a relatively straight-forward change. Let me know if you need any tips @jimmychu0807

chong-he:
Tested this with mainnet with just starting the VC, seeing `Failed to resolve pubkey to index` errors (the validators don't exist on the chain), probably that's because the VC still tries to call the BN to check for the validators. I think that's fine given that the error is expected when the VC is not connected to BN? (other errors are present too but also due to no BN present)

common/eth2_network_config/src/lib.rs
chong-he: tested with Hoodi, currently VC can't start on Hoodi with error: `Genesis state bytes missing from Eth2NetworkConfig`

jimmychu0807: I see, so some genesis states do not have genesis time!

Will also test on Hoodi next.

jimmychu0807:
@michaelsproul @chong-he code updated to have testnet genesis times hard-coded in eth2_config `GenesisStateSource::Url` type.

michaelsproul (approved):

PR #8641, opened 2026-01-08
Get blobs v2 metrics

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Add standardized metrics for getBlobsV2 from https://github.com/ethereum/beacon-metrics/pull/14. 

## Additional Info

I didn't add the metrics in the engine api `execution_layer_request_times` metric because the naming would be at odds with the names for the standardized metrics.

Also, did not add for getBlobsV1 as we are no longer using it post fulu. We should probably get rid of all deprecated engine api endpoints at some point.

### DIFF
--- a/beacon_node/beacon_chain/src/fetch_blobs/mod.rs
+++ b/beacon_node/beacon_chain/src/fetch_blobs/mod.rs
@@ -250,7 +250,8 @@ async fn fetch_and_process_blobs_v2<T: BeaconChainTypes>(
 
     // Track request count and duration for standardized metrics
     inc_counter(&metrics::BEACON_ENGINE_GET_BLOBS_V2_REQUESTS_TOTAL);
-    let _timer = metrics::start_timer(&metrics::BEACON_ENGINE_GET_BLOBS_V2_REQUEST_DURATION_SECONDS);
+    let _timer =
+        metrics::start_timer(&metrics::BEACON_ENGINE_GET_BLOBS_V2_REQUEST_DURATION_SECONDS);
 
     let response = chain_adapter
         .get_blobs_v2(versioned_hashes)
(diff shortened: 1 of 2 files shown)


eserilev (approved):

beacon_node/beacon_chain/src/fetch_blobs/mod.rs
```diff
@@ -247,13 +247,22 @@ async fn fetch_and_process_blobs_v2<T: BeaconChainTypes>(
 
     metrics::observe(&metrics::BLOBS_FROM_EL_EXPECTED, num_expected_blobs as f64);
     debug!(num_expected_blobs, "Fetching blobs from the EL");
+
+    // Track request count and duration for standardized metrics
+    inc_counter(&metrics::BEACON_ENGINE_GET_BLOBS_V2_REQUESTS_TOTAL);
+    let _timer =
+        metrics::start_timer(&metrics::BEACON_ENGINE_GET_BLOBS_V2_REQUEST_DURATION_SECONDS);
```
jimmygchen, line 254: I think we need to drop this timer after we get a response back below, otherwise it records the duration until the function completes, which would include computing columns and processing them.

beacon_node/beacon_chain/src/metrics.rs
```diff
@@ -1690,6 +1690,30 @@ pub static BLOBS_FROM_EL_RECEIVED: LazyLock<Result<Histogram>> = LazyLock::new(|
     )
 });
 
+pub static BEACON_ENGINE_GET_BLOBS_V2_REQUESTS_TOTAL: LazyLock<Result<IntCounter>> =
```
jimmygchen, line 1696: I think it may be worth adding a comment to tag these as standard metrics from the `beacon-metrics` repo, so we don't rename them without realising they are standardised metrics.

pawanjay176: Good call. added in [87dcecb](https://github.com/sigp/lighthouse/pull/8641/commits/87dcecb401df30bf35029b5342a018ddcdecf2f8)

beacon_node/beacon_chain/src/fetch_blobs/mod.rs
pawanjay176: Damn totally missed that. Thanks for catching it. I need to review claude's code more thoroughly before committing 😅  Fixed in 26a30a6

jimmygchen (approved):
LGTM!

PR #8643, opened 2026-01-11
Cleanup `consensus/types`  re-exports

macladson:
## Proposed Changes

Removes some of the temporary re-exports in `consensus/types`. 

I am doing this in multiple parts to keep each diff small.


pawanjay176 (approved):
LGTM

PR #8644, opened 2026-01-11
Remove `workspace_members` proc macro

macladson:
## Issue Addressed

I noticed that the `logging` crate was often being spuriously recompiled due to the `workspace_crates` macro running even when no changes needed to be made.

## Proposed Changes

This removes the proc macro and instead uses a build script to construct the list of workspace members and save them to a file in `logging`'s `OUT_DIR`. This build script should only be rerun when actually necessary which should reduce spurious recompiles.

## Additional Info

The build script will be rerun when:
1. This build.rs file is changed.
2. The workspace Cargo.toml changes.
3. Build dependencies change.


macladson:
Closing this for now since the problem seems intermittent and I wasn't able to reproduce it on other machines

PR #8646, opened 2026-01-12
Remove `http-api` tests from nightly and run tests on `unstable` by default

jimmygchen:
`http-api` tests are failing for earlier forks, but I think we might be able to just remove them for  earlier forks in the nightly tests as they're no longer relevant.

https://github.com/sigp/lighthouse/actions/runs/20869883982/job/59969302898

I've also updated the workflow to run on `unstable` by default, and allow for override when the workflow is manually triggered.

This PR points to the `stable` branch, as GitHub actions only schedule workflows from the `stable` branch.

### DIFF
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -6,6 +6,11 @@ on:
     # Run at 8:30 AM UTC every day
     - cron: '30 8 * * *'
   workflow_dispatch: # Allow manual triggering
+    inputs:
+      branch:
+        description: 'Branch to test'
+        required: false
+        default: 'unstable'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


jimmygchen:
Manually triggered a workflow from my fork:
https://github.com/jimmygchen/lighthouse/actions/runs/20905910142

michaelsproul:
I guess the CI failures are just because of compiler updates

jimmygchen:
Yep that's right. The manual workflow linked above passed. 

I'm not sure why it hasn't triggered on this fork yet, and I don't see anything wrong with the [`schedule`](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#schedule) syntax used. 

Might try to do a manual run after merging to see if it fixes this.

michaelsproul (approved):

PR #8647, opened 2026-01-12
Implement custom OpenTelemetry sampler to filter uninstrumented traces

jimmygchen:
## Description

Adds a `PrefixBasedSampler` to filter out traces that don't originate from known instrumented code paths. This reduces noise from uninstrumented code paths when exporting to OTLP backends.

Root spans use the `lh_` prefix to identify Lighthouse instrumented entry points. The sampler is used with OpenTelemetry's `ParentBased` sampler:
- Root spans are sampled only if their name starts with `lh_`
- Child spans automatically inherit their parent's sampling decision

This enables effective trace sampling for https://github.com/sigp/lighthouse/issues/8554 - without filtering, we get spans from uninstrumented code paths (e.g. `fork_choice_write_lock` only traces), making low sample rates ineffective at capturing meaningful instrumented traces.

## Additional Info

The `lh_` prefix approach eliminates the need to maintain an allowlist of span names - new instrumented spans just need the prefix to be exported. The prefix is kept short to minimize storage overhead in tracing backends.


beacon_node/lighthouse_tracing/src/lib.rs
```diff
@@ -81,3 +85,48 @@ pub const LH_BN_ROOT_SPAN_NAMES: &[&str] = &[
...
+    fn should_sample(
+        &self,
+        _parent_context: Option<&Context>,
+        _trace_id: opentelemetry::trace::TraceId,
+        name: &str,
+        _span_kind: &SpanKind,
+        _attributes: &[opentelemetry::KeyValue],
+        _links: &[Link],
+    ) -> SamplingResult {
+        if self.allowed_names.contains(&name) {
```
eserilev: A small optimization could be to make allowed_names a set instead of a list. Right now the list of allowed spans is relatively small so it might not matter much until the list gets bigger

beacon_node/lighthouse_tracing/src/lib.rs
```diff
@@ -81,3 +85,48 @@ pub const LH_BN_ROOT_SPAN_NAMES: &[&str] = &[
...
+/// This sampler is designed to be used with `Sampler::ParentBased`, which will:
+/// - Use this sampler's decision for root spans (no parent)
+/// - Automatically inherit the parent's sampling decision for child spans
+///
+/// This ensures that only traces starting from known instrumented code paths are exported,
+/// reducing noise from uninstrumented library code.
+#[derive(Debug, Clone)]
+pub struct AllowedRootSpanSampler {
+    allowed_names: &'static [&'static str],
+}
```
eserilev: Maybe a few unit tests here could be nice?

eserilev (commented):
just a few small things on my end

beacon_node/lighthouse_tracing/src/lib.rs
jimmygchen: I just had a thought: perhaps we could do prefix instead, so we don't have to maintain this list at all, because longer term this could be quite bad devex - say I add a trace, but forgot to add to this new trace to the allowed list, then i build and deploy BUT couldn't find the trace and had to debug - this wastes some dev cycles.

I'm thinking to add a `lh_` prefix (keeping it short so that it doesn't take up a lot of backend storage), what do you think?

beacon_node/lighthouse_tracing/src/lib.rs
jimmygchen: Yes, added!

beacon_node/lighthouse_tracing/src/lib.rs
jimmygchen: I've added this - i think this is a better longer term solution than maintaining a list. Let me know your thoughts!

eserilev: I agree, this seems much better

jimmygchen: Looking much better without the noise \o/

image

eserilev (approved):
LGTM!

One more thought I had was to maybe add a custom proc macro so that we don't have to do

``` rust
#[instrument(name = "lh_produce_unaggregated_attestation", skip_all, fields(%request_slot, %request_index), level = "debug")]
fn produce_unaggregated_attestation()
```

and instead do something like this

``` rust
#[lh_instrument(skip_all, fields(%request_slot, %request_index), level = "debug")]
fn produce_unaggregated_attestation()
```

where the proc macro automatically appends `lh_` to the function name

jimmygchen:
> LGTM!
> 
> One more thought I had was to maybe add a custom proc macro so that we don't have to do
> 
> ```rust
> #[instrument(name = "lh_produce_unaggregated_attestation", skip_all, fields(%request_slot, %request_index), level = "debug")]
> fn produce_unaggregated_attestation()
> ```
> 
> and instead do something like this
> 
> ```rust
> #[lh_instrument(skip_all, fields(%request_slot, %request_index), level = "debug")]
> fn produce_unaggregated_attestation()
> ```
> 
> where the proc macro automatically appends `lh_` to the function name

Thanks, I see the convenience with the macro, but IMO the gain is quite trivial to justify adding custom macro for this - it may not be immediately obvious what this macro does without looking at the implementation and might add confusion vs when to use `#[instrument]` macro - which is still what we will use most of the time when creating spans. I have a slight preference to be explicit here for readability, but I'm open if you and others think it's useful to add.

eserilev:
sorry thought I already left a comment. yes I agree with you, the proc macro on second thought seems pretty useless. LGTM!

eserilev (approved):

PR #8651, opened 2026-01-13
Update buckets for metric

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

The `beacon_data_column_sidecar_computation_seconds` used to record the full kzg proof generation times before we changed getBlobsV2 to just return the full proofs + cells. This metric should be taking way less time than 100ms which was the minimum bucket previously. 

Update the metric to use the default buckets for better granularity.


jimmygchen (approved):
Makes sense! thanks

Issue #8652, opened 2026-01-13
Expose a minimal set of core types in `consensus/types`


macladson:
## Description

We should move to provide a minimal set of core types available in `consensus/types`. These types should be available without needing to compile the rest of the modules in `consensus/types`. The `core` module is already set up to provide this, we just need to remove any incompatible dependencies and implement the appropriate features/feature gates.

Currently there are a few minor blockers to this:

- `core/chain_spec.rs` contains dependencies on `fork`,  `data`, `state` and `execution`.
- `core/eth_spec.rs` depends on `state`.
- `core/signing_data.rs` depends on `fork`.
- `test_utils`

I think these are all able to be simply resolved by either moving required types into `core` (such as `ExecutionBlockHash` which will eliminate the dependency on `execution`, or moving certain state related functions to `beacon_state` instead, which might even be a more natural place for them to exist.

I think `fork` can be reasonably added as a dep onto `core` as it is minimal (does not contain any other crate deps) and is very intertwined with the spec types which we would want to provide anyway.

`test_utils` we could conceivably gate behind a `testing` feature but I haven't given a lot of thought to what the implications for that would be.

PR #8653, opened 2026-01-13
Remove `state` dependency from `core` module in `consensus/types`

macladson:
## Issue Addressed

#8652 

## Proposed Changes

- This removes instances of `BeaconStateError` from `eth_spec.rs`, and replaces them directly with `ArithError` which can be trivially converted back to `BeaconStateError` at the call site.
- Also moves the state related methods on `ChainSpec` to be methods on `BeaconState` instead. I think this might be a more natural place for them to exist anyway.


pawanjay176 (approved):
I don't know if there was a good reason to have it in the chain_spec but this looks much cleaner and more natural to me.

michaelsproul (approved):
LGTM, nice improvement!

PR #8659, opened 2026-01-14
Use `if-addrs` instead of `local_ip_address`

macladson:
## Proposed Changes

Swaps out the `local_ip_address` dependency for `if-addrs`. The reason for this is that is that `local_ip_address` is a relatively heavy dependency (depends on `neli`) compared to `if-addrs` and we only use it to check the presence of an IPv6 interface. This is an experiment to see if we can use the more lightweight `if-addrs` instead.

## Additional Info

Note that the older `if-addrs` version `0.10` exists in our `Cargo.lock` as an optional dependency of `if-watch` but we do not actually compile it


ackintosh (approved):

macladson:
After talking to @jxs, I'm going to try to see if we can use `if-watch` directly here instead

macladson:
@jxs what do you think about merging this one? I've pretty much given up on #8662 but feel free to play around with it if you think it's worth it

jxs (approved):
LGTM, thanks Mac!

PR #8662, opened 2026-01-15
Use `if-watch` instead of `local_ip_address`

macladson:
## Proposed Changes

This is an alternative to #8659 which just uses `if-watch` which we already depend on transitively through `libp2p`.

### DIFF
--- a/beacon_node/lighthouse_network/src/config.rs
+++ b/beacon_node/lighthouse_network/src/config.rs
@@ -270,11 +270,8 @@ impl Config {
         let mut ctx = std::task::Context::from_waker(futures::task::noop_waker_ref());
 
         // Ensure all addresses are added.
-        loop {
-            match watcher.poll_if_event(&mut ctx) {
-                std::task::Poll::Ready(Ok(_)) => continue,
-                _ => break,
-            }
+        while let std::task::Poll::Ready(Ok(_)) = watcher.poll_if_event(&mut ctx) {
+            continue;
         }
 
         watcher
--- a/beacon_node/lighthouse_network/src/config.rs
+++ b/beacon_node/lighthouse_network/src/config.rs
@@ -5,8 +5,8 @@ use crate::{Enr, PeerIdSerialized};
 use directory::{
     DEFAULT_BEACON_NODE_DIR, DEFAULT_HARDCODED_NETWORK, DEFAULT_NETWORK_DIR, DEFAULT_ROOT_DIR,
 };
+use if_watch::{IpNet, tokio::IfWatcher};
 use libp2p::Multiaddr;
-use local_ip_address::local_ipv6;
 use network_utils::listen_addr::{ListenAddr, ListenAddress};
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
@@ -262,13 +262,13 @@ impl Config {
     /// A helper function to check if the local host has a globally routeable IPv6 address. If so,
     /// returns true.
     pub fn is_ipv6_supported() -> bool {
-        // If IPv6 is supported
-        let Ok(std::net::IpAddr::V6(local_ip)) = local_ipv6() else {
+        let Ok(watcher) = IfWatcher::new() else {
             return false;
         };
 
-        // If its globally routable, return true
-        is_global_ipv6(&local_ip)
+        watcher
+            .iter()
+            .any(|addr| matches!(addr, IpNet::V6(ip) if is_global_ipv6(&ip.addr())))
     }
 
     pub fn listen_addrs(&self) -> &ListenAddress {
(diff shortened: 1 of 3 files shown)


macladson (commented):
~Actually I think this doesn't work.~
~I think this uses `if-addrs` under the hood anyway:~

~https://github.com/libp2p/if-watch/blob/master/src/fallback.rs~

beacon_node/lighthouse_network/src/config.rs
```diff
@@ -262,13 +262,13 @@ impl Config {
     /// A helper function to check if the local host has a globally routeable IPv6 address. If so,
     /// returns true.
     pub fn is_ipv6_supported() -> bool {
-        // If IPv6 is supported
-        let Ok(std::net::IpAddr::V6(local_ip)) = local_ipv6() else {
+        let Ok(watcher) = IfWatcher::new() else {
```
macladson: I think this is initialized empty and so we would have to force a `resync` somehow

ackintosh (commented):
Thanks @macladson! It seems preferable to use if-addrs directly, since that's what if-watch uses under the hood.

macladson:
> Thanks @macladson! It seems preferable to use if-addrs directly, since that's what if-watch uses under the hood.

I actually misunderstood the feature paths, it only uses `if-addrs` for Windows, so this might still be worth doing if we can fix the PR

ackintosh:
Let me clarify: `if-watch` uses `if-addrs` on [Windows](https://github.com/libp2p/if-watch/blob/9b87056c658279fa505f92d05a24f97e76514d1b/src/win.rs#L68) and [macOS](https://github.com/libp2p/if-watch/blob/31c4780d803431914aab95ec9139aa8e5795f80c/src/apple.rs#L58), but not on Linux. On Linux, [`if-watch` uses `rtnetlink` instead](https://github.com/libp2p/if-watch/blob/943a412adb0009371356dcbb491d573f98e36902/src/linux.rs#L58), despite `if-addrs` supporting Linux.

I guess the reason is that [`if-watch` wants to receive *events* for network interface up/down](https://github.com/libp2p/if-watch/blob/943a412adb0009371356dcbb491d573f98e36902/src/linux.rs#L107-L108),  whereas `if-addrs` only provides a snapshot.

In our case, we simply want to get a list of network interfaces, so using `if-addrs` directly seems straightforward.

jxs:
> Let me clarify: `if-watch` uses `if-addrs` on [Windows](https://github.com/libp2p/if-watch/blob/9b87056c658279fa505f92d05a24f97e76514d1b/src/win.rs#L68) and [macOS](https://github.com/libp2p/if-watch/blob/31c4780d803431914aab95ec9139aa8e5795f80c/src/apple.rs#L58), but not on Linux. On Linux, [`if-watch` uses `rtnetlink` instead](https://github.com/libp2p/if-watch/blob/943a412adb0009371356dcbb491d573f98e36902/src/linux.rs#L58), despite `if-addrs` supporting Linux.
> 
> I guess the reason is that [`if-watch` wants to receive _events_ for network interface up/down](https://github.com/libp2p/if-watch/blob/943a412adb0009371356dcbb491d573f98e36902/src/linux.rs#L107-L108), whereas `if-addrs` only provides a snapshot.
> 
> In our case, we simply want to get a list of network interfaces, so using `if-addrs` directly seems straightforward.

still, if we can avoid another dependency and just use `if-watch` as it's transitive would be nice, is it possible?

beacon_node/lighthouse_network/src/config.rs
```diff
@@ -262,13 +262,24 @@ impl Config {
...
         };
 
-        // If its globally routable, return true
-        is_global_ipv6(&local_ip)
+        // Hack to poll for addresses.
+        let mut ctx = std::task::Context::from_waker(futures::task::noop_waker_ref());
+
+        // Ensure all addresses are added.
+        loop {
+            match watcher.poll_if_event(&mut ctx) {
```
macladson: So the code was panicking since `tokio::IfWatcher` requires a runtime. Adding this either to the function itself or piping in an existing runtime from elsewhere feels way overkill to me.

I have switched to `smol::IfWatcher` which doesn't need an existing runtime, and I think it technically works now. The downside is, the `smol` feature added a bunch of extra deps so it totally defeats the purpose of this change.

So with that in mind, I am heavily leaning in favour of the original PR with `if-addrs` (https://github.com/sigp/lighthouse/pull/8659).

Unless anyone has a cleaner solution that I missed?

ackintosh: One idea I had is to add a function to `if-watch`, such as `list_interfaces()`, that returns the network interfaces available at the current time, internally backed by `if-addrs`.

This would avoid adding an extra dependency to Lighthouse while still allowing us to obtain the interface list.

macladson:
Closing in favour of #8659

PR #8666, opened 2026-01-15
Remove `execution` dependency from `core` module in `consensus/types`

macladson:
## Issue Addressed

#8652 

## Proposed Changes

This moves the `ExecutionBlockHash` from the `execution` module to the `core` module. This allows `core` to not depend on the `execution` module, and the `ExecutionBlockHash` is a pretty core  part of our types so I think it makes sense.

## Additional Information

Ideally we would pull this type from `alloy`, but I think with our specific requirements, particularly around `tree_hash` and `ssz`, that might not be viable.


pawanjay176 (approved):

macladson:
@mergifyio requeue

PR #8665, opened 2026-01-15
More `consensus/types` re-export cleanup

macladson:
## Proposed Changes
Remove more of the temporary re-exports from `consensus/types`

## Additional Info

Previous PR: https://github.com/sigp/lighthouse/pull/8643

PR #8672, opened 2026-01-16
Remove remaining facade module re-exports from `consensus/types`

macladson:
## Proposed Changes

Removes the remaining facade re-exports from `consensus/types`. 
I have left `graffiti` as I think it has some utility so am leaning towards keeping it in the final API design.

## Additional Info

Previous PRs:
- #8643 
- #8665


pawanjay176 (approved):
Again only mechanical changes that followed from removing the re-exports from the types crate. 
LGTM

PR #8677, opened 2026-01-19
Gloas consensus v1.7.0-alpha.1

michaelsproul:
## Proposed Changes

This PR breaks out the consensus changes from:

- https://github.com/sigp/lighthouse/pull/8279

And merges in:

- https://github.com/sigp/lighthouse/pull/8283

The goal is to get up to date with the latest spec ([v1.7.0-alpha.1](https://github.com/ethereum/consensus-specs/releases/tag/v1.7.0-alpha.1)) and get the EF tests passing. We do not need the `BeaconChain` envelope processing changes for this, so they will stay in #8279.


consensus/state_processing/src/per_block_processing.rs
```diff
@@ -631,72 +696,163 @@ pub fn get_expected_withdrawals<E: EthSpec>(
...
+    // Verify the bid signature
+    let signed_bid = block.body().signed_execution_payload_bid()?;
+
+    let bid = &signed_bid.message;
+    let amount = bid.value;
+    let builder_index = bid.builder_index;
+    let builder = state.get_validator(builder_index as usize)?;
+
+    // For self-builds, amount must be zero regardless of withdrawal credential prefix
+    if builder_index == block.proposer_index() {
```
shane-moore, line 860: this self-building check has been updated to `if builder_index == BUILDER_INDEX_SELF_BUILD`
https://ethereum.github.io/consensus-specs/specs/gloas/beacon-chain/#new-process_execution_payload_bid

additionally, looks like a few more mods have happened to this function in the spec as well since we've made the in-protocol builder a new staked actor

all the beacon chain spec page's implementations we've made supported up til `v1.6.1`, so probably good idea to go through the spec PR's from then til [v1.7.0-alpha.1](https://github.com/ethereum/consensus-specs/releases/tag/v1.7.0-alpha.1) as to be sure we're compliant

michaelsproul: yeah we are working through this now, planning to go through every function

michaelsproul:
Closing. I'm still grabbing stuff from this branch, but most of it is merged already, and this doesn't need to be a PR.

PR #8678, opened 2026-01-19
Remove `syn` version `1`

macladson:
## Issue Addressed
#8547 

## Proposed Changes

This updates a few of our crates to remove the older `syn 1` crate.

This updates:
- `criterion` -> `0.8`
- `itertools` -> `0.14`

And also certain `sigp` crates:
- `xdelta3` -> [`fe39066`](https://github.com/sigp/xdelta3-rs/commit/fe3906605c87b6c0515bd7c8fc671f47875e3ccc)
- `superstruct` -> `0.10.1`
- `ethereum_ssz` -> `0.10.1`
- `tree_hash` -> `0.12.1`
- `metastruct` -> `0.1.4`
- `context_deserialize` -> `0.2.1`
- `compare_fields` -> `0.1.1`

## Additional Info

`syn 1` still exists in the lockfile as a dependency of `ark-ff` which is an optional dep of alloy (which we don't use),  but we no longer compile it. This is enforced by the new entry in `deny.toml`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -112,6 +112,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "alloca"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e5a7d05ea6aea7e9e64d25b9156ba2fee3fdd659e34e41063cd2fc7cd020d7f4"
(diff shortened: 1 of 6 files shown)


pawanjay176 (approved):
NIce!

PR #8682, opened 2026-01-20
Add Gloas data column support 

eserilev:
data columns: https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/p2p-interface.md#modified-datacolumnsidecar

There are some TODOs related to gloas data column gossip verification that should be handled in a separate PR


michaelsproul:
Thanks! 🙌 

Looks like this would have been a lot of work :pray:

beacon_node/beacon_chain/src/observed_data_sidecars.rs
```diff
@@ -73,8 +63,8 @@ impl<E: EthSpec> ObservableDataSidecar for DataColumnSidecar<E> {
 /// like checking the proposer signature.
 pub struct ObservedDataSidecars<T: ObservableDataSidecar> {
     finalized_slot: Slot,
-    /// Stores all received data indices for a given `(ValidatorIndex, Slot)` tuple.
-    items: HashMap<ProposalKey, HashSet<u64>>,
+    /// Stores all received data indices for a given `slot`.
+    items: HashMap<Slot, HashSet<u64>>,
```
eserilev: Doesn't seem to be a reason why we use `ProposalKey` instead of `Slot`. post-gloas data columns no longer have a `signed_beacon_block_header` field which means we cant easily access the `proposer_index`

jimmygchen: This is for DOS protection - if we use slot, an attacker could produce a column sidecar that passes gossip and prevent other column sidecars from getting processed.

Checking the spec, it looks like we need to use `beacon_block_root` and `index`
https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/p2p-interface.md#data_column_sidecar_subnet_id

eserilev: ah okay they key is now block root and slot 
42c71c5

beacon_node/http_api/src/block_id.rs
```diff
@@ -462,17 +464,18 @@ impl BlockId {
...
                         Ok(Some(data_column)) => Some(Ok(data_column)),
                         Ok(None) => None,
                         Err(e) => Some(Err(warp_utils::reject::unhandled_error(e))),
-                    },
-                )
+                    }
+                })
                 .collect::<Result<Vec<_>, _>>()?;
 
             reconstruct_blobs(&chain.kzg, data_columns, blob_indices, block, &chain.spec).map_err(
```
jimmygchen, line 481: `reconstruct_blobs()` will return error in gloas, because `signed_block_header` and `kzg_commitments_inclusion_proof`  no longer exists in the Gloas `DataColumnSidecar` - may be worth adding a TODO here so we don't miss this?

https://github.com/eserilev/lighthouse/blob/f25308822943798a42fff3af8c9a38d04375db3c/beacon_node/beacon_chain/src/kzg_utils.rs#L460-L470

eserilev: added [d7b692a](https://github.com/sigp/lighthouse/pull/8682/commits/d7b692a9fcda37286800c6c84aeecef2ea3f2227)

consensus/types/src/data/data_column_sidecar.rs
```diff
@@ -38,15 +39,49 @@ pub struct DataColumnsByRootIdentifier<E: EthSpec> {
 
 pub type DataColumnSidecarList<E> = Vec<Arc<DataColumnSidecar<E>>>;
 
+#[derive(Debug, PartialEq, Clone)]
+pub enum Error {
+    IncorrectStateVariant,
```
jimmygchen: could just add this to `DataColumnSidecarError`?

consensus/types/src/data/data_column_sidecar.rs
```diff
@@ -55,20 +90,58 @@ pub struct DataColumnSidecar<E: EthSpec> {
...
+            DataColumnSidecar::Gloas(column) => column.beacon_block_root,
+        }
+    }
+
+    /// Custom SSZ decoder that takes a `ForkName` as context.
+    pub fn from_ssz_bytes_for_fork(
+        bytes: &[u8],
+        fork_name: ForkName,
+    ) -> Result<Self, ssz::DecodeError> {
+        if fork_name.gloas_enabled() {
```
jimmygchen: would it be safer to use `match` here, in the unlikely case if a new variant is added we don't convert into the wrong type here.

testing/ef_tests/tests/tests.rs
```diff
@@ -659,9 +660,9 @@ mod ssz_static {
 
     #[test]
     fn data_column_sidecar() {
-        SszStaticHandler::<DataColumnSidecar<MinimalEthSpec>, MinimalEthSpec>::fulu_and_later()
+        SszStaticHandler::<DataColumnSidecarFulu<MinimalEthSpec>, MinimalEthSpec>::fulu_only()
             .run();
-        SszStaticHandler::<DataColumnSidecar<MainnetEthSpec>, MainnetEthSpec>::fulu_and_later()
+        SszStaticHandler::<DataColumnSidecarFulu<MainnetEthSpec>, MainnetEthSpec>::fulu_only()
```
jimmygchen, line 665: should we add Gloas too?

beacon_node/beacon_chain/src/fetch_blobs/fetch_blobs_beacon_adapter.rs
```diff
@@ -67,27 +67,25 @@ impl<T: BeaconChainTypes> FetchBlobsBeaconAdapter<T> {
             .map_err(FetchEngineBlobError::RequestFailed)
     }
 
-    pub(crate) fn blobs_known_for_proposal(
+    pub(crate) fn blobs_known_for_observation_key(
```
jimmygchen, line 70: i think we only want to use this new tuple from gloas?

beacon_node/beacon_chain/src/fetch_blobs/fetch_blobs_beacon_adapter.rs
```diff
@@ -67,27 +67,25 @@ impl<T: BeaconChainTypes> FetchBlobsBeaconAdapter<T> {
...
         self.chain
             .observed_blob_sidecars
             .read()
-            .known_for_proposal(&proposer_key)
+            .known_for_observation_key(&observation_key)
             .cloned()
     }
 
-    pub(crate) fn data_column_known_for_proposal(
+    pub(crate) fn data_column_known_for_observation_key(
```
jimmygchen, line 81: same as above - i think fulu rules remains unchanged

beacon_node/beacon_chain/src/fetch_blobs/mod.rs
```diff
@@ -391,9 +391,12 @@ async fn compute_custody_columns_to_import<T: BeaconChainTypes>(
                     .map_err(FetchEngineBlobError::DataColumnSidecarError)?;
 
                 // Only consider columns that are not already observed on gossip.
-                if let Some(observed_columns) = chain_adapter_cloned.data_column_known_for_proposal(
-                    ProposalKey::new(block.message().proposer_index(), block.slot()),
-                ) {
+                if let Some(observed_columns) = chain_adapter_cloned
```
jimmygchen: we may need separate logic for fulu and gloas here

beacon_node/network/src/sync/block_lookups/mod.rs
```diff
@@ -88,7 +88,11 @@ impl<E: EthSpec> BlockComponent<E> {
         match self {
             BlockComponent::Block(block) => block.value.parent_root(),
             BlockComponent::Blob(blob) => blob.value.block_parent_root(),
-            BlockComponent::DataColumn(column) => column.value.block_parent_root(),
+            BlockComponent::DataColumn(column) => match column.value.as_ref() {
+                DataColumnSidecar::Fulu(column) => column.block_parent_root(),
+                // TODO(gloas) we don't have a parent root post gloas, not sure what to do here
+                DataColumnSidecar::Gloas(column) => column.beacon_block_root,
```
jimmygchen, line 94: this might actually be correct for lookup but need to confirm - the parent is technically the beacon block root - if we don't know the beacon block root, we'd have to look it up.

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3543,8 +3562,11 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         publish_fn: impl FnOnce() -> Result<(), BlockError>,
     ) -> Result<AvailabilityProcessingStatus, BlockError> {
         if let Some(slasher) = self.slasher.as_ref() {
-            for data_colum in &data_columns {
-                slasher.accept_block_header(data_colum.signed_block_header());
+            for data_column in &data_columns {
+                // TODO(gloas) no block header post-gloas, what should we do here
+                if let DataColumnSidecar::Fulu(c) = data_column.as_data_column() {
```
jimmygchen, line 3571: Looks like this can be skipped and superceded by a different gossip check in gloas:

https://github.com/ethereum/consensus-specs/blob/81458afc6aad6985c533785c8d2860d87a993241/specs/gloas/p2p-interface.md?plain=1#L385

beacon_node/beacon_chain/src/observed_data_sidecars.rs
```diff
@@ -99,9 +113,9 @@ impl<T: ObservableDataSidecar> ObservedDataSidecars<T> {
 
         let data_indices = self
             .items
-            .entry(ProposalKey {
+            .entry(ObservationKey {
```
jimmygchen: `ObservationKey` may have to be an enum now to support Gloas, and we could pass `ObservationKey` and leave it to the caller to construct

jimmygchen (changes requested):
Nice work! 
I haven't finished but I've added comments I have so far - we mainly need to make sure it doesn't impact / change Fulu code path before we merge this, so I think we'll need to retain the `ObservedDataSidecars` behaviour for Fulu.

eserilev:
I know a new observation key technically affects fulu code paths but it seems like unnecessary overhead to maintain separate code paths for this section of the code. Is there a reason why block root + slot won't work the same as proposer index + slot?

beacon_node/network/src/sync/block_lookups/mod.rs
eserilev: right, yeah it might be correct. Pawan said he's looking into this as he's working on lookup sync

eserilev:
re: observation key
After talking with jimmy, we're going to keep the fulu code path the same as it currently is

consensus/types/src/data/data_column_sidecar.rs
eserilev: 22abdf0

testing/ef_tests/tests/tests.rs
eserilev: 622f67f

beacon_node/beacon_chain/src/beacon_chain.rs
eserilev: i've updated the TODO comment. We can implement the gloas specific gossip checks in a separate PR

beacon_node/beacon_chain/src/observed_data_sidecars.rs
eserilev: ObservationKey is now an enum and supports both the existing pre gloas functionality and the new post gloas functionality

consensus/types/src/data/data_column_sidecar.rs
```diff
@@ -38,15 +39,44 @@ pub struct DataColumnsByRootIdentifier<E: EthSpec> {
...
 #[cfg_attr(
     feature = "arbitrary",
     derive(arbitrary::Arbitrary),
     arbitrary(bound = "E: EthSpec")
 )]
-#[derive(Debug, Clone, Serialize, Deserialize, Encode, Decode, TreeHash, TestRandom, Educe)]
-#[serde(bound = "E: EthSpec")]
-#[educe(PartialEq, Eq, Hash(bound(E: EthSpec)))]
-#[context_deserialize(ForkName)]
+#[derive(Debug, Clone, Serialize, TreeHash, Encode, Decode, Educe, Deserialize)]
```
jimmygchen: I think we can't `Decode` without fork here and should probably remove this?

consensus/types/src/data/data_column_sidecar.rs
```diff
@@ -38,15 +39,44 @@ pub struct DataColumnsByRootIdentifier<E: EthSpec> {
...
+#[educe(PartialEq, Hash(bound(E: EthSpec)))]
+#[serde(untagged)]
+#[tree_hash(enum_behaviour = "transparent")]
+#[ssz(enum_behaviour = "transparent")]
+#[serde(bound = "E: EthSpec", deny_unknown_fields)]
```
jimmygchen: nit: can  be merged into one line with `serde(untagged)` above

beacon_node/beacon_chain/src/observed_data_sidecars.rs
```diff
@@ -50,19 +63,66 @@ impl<E: EthSpec> ObservableDataSidecar for DataColumnSidecar<E> {
...
+
+        if spec.fork_name_at_slot::<E>(slot).gloas_enabled() {
+            Ok(Self::new_block_root_key(sidecar.beacon_block_root(), slot))
+        } else {
+            Self::new_proposer_key(sidecar.proposer_index(), slot)
+        }
+    }
+
+    pub fn new_proposer_key(
+        proposer_index: Option<ValidatorIndex>,
```
jimmygchen: A bit odd to use `Option` here given this is mandatory

beacon_node/beacon_chain/src/observed_data_sidecars.rs
```diff
@@ -201,7 +259,7 @@ mod tests {
     #[test]
     fn pruning() {
         let spec = Arc::new(test_spec::<E>());
-        let mut cache = ObservedDataSidecars::<BlobSidecar<E>>::new(spec);
+        let mut cache = ObservedDataSidecars::<BlobSidecar<E>, E>::new(spec.clone());
```
jimmygchen: I've just realised we've got no coverage for `DataColumnSidecar`, and we're testing Gloas + BlobSidecar, which isn't a real code path that we'd hit. Seeing this is existing code and this PR is a blocker for other gloas, happy to address this in a separate PR but I thought it's worth mentioning it here.

beacon_node/beacon_chain/src/observed_data_sidecars.rs
```diff
@@ -97,35 +157,34 @@ impl<T: ObservableDataSidecar> ObservedDataSidecars<T> {
...
-        let data_indices = self
-            .items
-            .entry(ProposalKey {
-                slot: data_sidecar.slot(),
-                proposer: data_sidecar.block_proposer_index(),
-            })
-            .or_insert_with(|| {
-                HashSet::with_capacity(T::max_num_of_items(&self.spec, data_sidecar.slot()))
-            });
+        let observation_key = ObservationKey::new::<T, E>(data_sidecar, &self.spec)?;
```
jimmygchen, line 169: stale function comment above

beacon_node/beacon_chain/src/fetch_blobs/mod.rs
```diff
@@ -391,9 +393,21 @@ async fn compute_custody_columns_to_import<T: BeaconChainTypes>(
                     .map_err(FetchEngineBlobError::DataColumnSidecarError)?;
 
                 // Only consider columns that are not already observed on gossip.
-                if let Some(observed_columns) = chain_adapter_cloned.data_column_known_for_proposal(
-                    ProposalKey::new(block.message().proposer_index(), block.slot()),
-                ) {
+                let observation_key = if spec
+                    .fork_name_at_slot::<T::EthSpec>(block.slot())
+                    .gloas_enabled()
```
jimmygchen: I think it would be a bit more consistent to extract this to a `ObservationKey::from_block` function, given we have the same  fork checking logic in `new()`

beacon_node/beacon_chain/src/test_utils.rs
```diff
@@ -3363,39 +3368,76 @@ pub fn generate_data_column_sidecars_from_block<E: EthSpec>(
...
+                let kzg_cell = cell_bytes.try_into().unwrap();
+                let kzg_proof = kzg_proofs.into_iter().next().unwrap();
+                (kzg_cell, kzg_proof)
+            })
+            .collect::<(Vec<_>, Vec<_>)>();
+
+        vec![(cells.try_into().unwrap(), proofs.try_into().unwrap()); kzg_commitments.len()]
+    };
+
+    if block.fork_name_unchecked().gloas_enabled() {
```
jimmygchen: nit: we could just move this to the if/else branch above?

beacon_node/beacon_chain/src/observed_data_sidecars.rs
jimmygchen: Oh i think we should at least add a test case for the duplicate block root scenario in Gloas

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3222,7 +3235,14 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         // Reject RPC columns referencing unknown parents. Otherwise we allow potentially invalid data
         // into the da_checker, where invalid = descendant of invalid blocks.
         // Note: custody_columns should have at least one item and all items have the same parent root.
-        if let Some(parent_root) = custody_columns.iter().map(|c| c.block_parent_root()).next()
+        // TODO(gloas) ensure this check is no longer relevant post gloas
```
jimmygchen, line 3241: I think we may need a different code path for gloas - this is normally unreachable in gloas because of the above fork choice check - we may have the gloas block in fork choice right?

beacon_node/beacon_chain/src/fetch_blobs/mod.rs
```diff
@@ -391,9 +393,21 @@ async fn compute_custody_columns_to_import<T: BeaconChainTypes>(
...
+                {
+                    Some(ObservationKey::new_block_root_key(block_root, block.slot()))
+                } else {
+                    ObservationKey::new_proposer_key(
+                        Some(block.message().proposer_index()),
+                        block.slot(),
+                    )
+                    .ok()
```
jimmygchen: if we can make `propsoer_index` mandatory then things simplify a lot here i think

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -124,7 +137,6 @@ pub enum GossipDataColumnError {
     ///
     /// The peer isn't faulty, but we do not forward it over gossip.
     PriorKnown {
```
jimmygchen, line 138: I think we may want to separate the Fulu and Gloas errors here so we don't lose info

Perhaps something like `PriorKnownProposer` and `PriorKnownBlockRoot`?

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -209,17 +224,37 @@ impl<T: BeaconChainTypes, O: ObservationStrategy> GossipVerifiedDataColumn<T, O>
...
+            }
+            // TODO(gloas) support gloas data column variant
+            DataColumnSidecar::Gloas(_) => Err(GossipDataColumnError::InvalidVariant),
+        }
+    }
+
+    /// Construct a `GossipVerifiedBlob` that is assumed to be valid.
+    ///
+    /// This should ONLY be used for testing.
+    pub fn __assumed_valid(column: Arc<DataColumnSidecar<T::EthSpec>>) -> Self {
```
jimmygchen: Can't find usage of this, can be deleted?
we also have `__new_for_testing`

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -478,11 +509,15 @@ where
 }
 
 #[instrument(skip_all, level = "debug")]
-pub fn validate_data_column_sidecar_for_gossip<T: BeaconChainTypes, O: ObservationStrategy>(
+pub fn validate_data_column_sidecar_for_gossip_fulu<T: BeaconChainTypes, O: ObservationStrategy>(
     data_column: Arc<DataColumnSidecar<T::EthSpec>>,
     subnet: DataColumnSubnetId,
     chain: &BeaconChain<T>,
 ) -> Result<GossipVerifiedDataColumn<T, O>, GossipDataColumnError> {
+    let DataColumnSidecar::Fulu(data_column_fulu) = data_column.as_ref() else {
```
jimmygchen, line 510: would be able to get rid of this error if we just accept `DataColumnSidecarFulu` instead

beacon_node/beacon_chain/src/data_column_verification.rs
```diff
@@ -827,19 +870,19 @@ mod test {
...
             )
         };
-        empty_data_column_sidecars_fails_validation(&harness, &verify_fn).await;
+        empty_data_column_sidecars_fails_validation_fulu(&harness, &verify_fn).await;
         data_column_sidecar_commitments_exceed_max_blobs_per_block(&harness, &verify_fn).await;
     }
 
     #[tokio::test]
-    async fn test_new_for_block_publishing() {
+    async fn test_new_for_block_publishing_fulu() {
```
jimmygchen, line 883: can some of these tests be generic over fulu/gloas, or do we need ot make a copy of the same tests?

beacon_node/beacon_chain/src/kzg_utils.rs
```diff
@@ -460,7 +564,7 @@ mod test {
         let spec = ForkName::Fulu.make_genesis_spec(E::default_spec());
```
jimmygchen, line 565: we'll want a test for Gloas too. If the tests can be made generic, we could use `test_spec()` to ceate the spec and reuse the tests - but this may not be possible

beacon_node/lighthouse_network/tests/rpc_tests.rs
```diff
@@ -963,7 +963,8 @@ fn test_tcp_columns_by_root_chunked_rpc() {
     let messages_to_send = 32 * num_of_columns;
 
     let spec = Arc::new(spec_with_all_forks_enabled());
-    let current_fork_name = ForkName::Fulu;
+    let slot = 320u64.into();
```
jimmygchen: what is this slot? looking at `spec_with_all_forks_enabled` this would always be a gloas slot right, so the Fulu code path below isn't reachable? would be good to cover both fulu and gloas, perhaps we could extract the test logic to function ?

beacon_node/lighthouse_network/tests/rpc_tests.rs
```diff
@@ -1129,7 +1144,8 @@ fn test_tcp_columns_by_range_chunked_rpc() {
     let messages_to_send = 32;
 
     let spec = Arc::new(spec_with_all_forks_enabled());
-    let current_fork_name = ForkName::Fulu;
+    let slot: Slot = 320u64.into();
```
jimmygchen: same as above

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -1471,14 +1471,22 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
                     StoreOp::PutDataColumns(block_root, data_columns) => {
-                        let indices = data_columns.iter().map(|c| c.index).collect();
-                        StoreOp::DeleteDataColumns(*block_root, indices)
+                        let indices = data_columns.iter().map(|c| *c.index()).collect();
+                        if let Some(column) = data_columns.first() {
+                            let slot = column.slot();
+                            let fork_name = self.spec.fork_name_at_slot::<E>(slot);
+                            StoreOp::DeleteDataColumns(*block_root, indices, fork_name)
+                        } else {
+                            return Err(Error::DBError {
```
jimmygchen: i dont know if this is reachable but could be a potential regression here if it is reacahble - perhaps we want to retain existing behaviour, i.e. if there's no data column, do nothing - unless we're absolutely sure this isn't reachable under normal scenarios in Fulu.

jimmygchen (changes requested):
I've finished the first round of review, nice work! 
Let me know if you have any questions and if you have any thoughts on the comments.

beacon_node/beacon_chain/src/fetch_blobs/mod.rs
eserilev: added from_block function

beacon_node/beacon_chain/src/observed_data_sidecars.rs
eserilev: updated comment and left a TODO

beacon_node/beacon_chain/src/observed_data_sidecars.rs
eserilev: I left a TODO, i think this whole section of tests needs to be updated

beacon_node/beacon_chain/src/fetch_blobs/mod.rs
eserilev: proposer_index is now mandatory

beacon_node/beacon_chain/src/data_column_verification.rs
eserilev: 80a992b 

PriorKnown now accepts an ObservationKey. based on the ObservationKey variant it should be clear if its a Fulu or Gloas error

beacon_node/beacon_chain/src/data_column_verification.rs
eserilev: yeah but I think ill then need to clone the Fulu variant to reconstruct the parent `DataColumnSidecar` when i pass it to functions that share verification logic between fulu and gloas 

Like these that expect a `&DataColumnSidecar`
``` rust
    verify_data_column_sidecar(&data_column, &chain.spec)?;
    verify_index_matches_subnet(&data_column, subnet, &chain.spec)?;
    verify_sidecar_not_from_future_slot(chain, column_slot)?;
    verify_slot_greater_than_latest_finalized_slot(chain, column_slot)?;
    verify_is_unknown_sidecar(chain, &data_column)?;
```
I guess we could also pass in both the `Arc<DataColumnSidecar>` and `&DataColumnSideCarFulu` as arguments but that feels a bit akward as well

Either way I think we should avoid the clone

beacon_node/beacon_chain/src/kzg_utils.rs
eserilev: I added a TODO

beacon_node/beacon_chain/src/data_column_verification.rs
eserilev: probably could be, I added a TODO. until verification logic is implemented for gloas data columns, these tests wont be useful for Gloas

beacon_node/lighthouse_network/tests/rpc_tests.rs
eserilev: 4cd5a72

beacon_node/lighthouse_network/tests/rpc_tests.rs
eserilev: 4cd5a72

beacon_node/store/src/hot_cold_store.rs
eserilev: I've slightly updated the error handling here 
4cd5a72 

I don't think this is reachable, but if it is we are handling the error in a similar way to other rollback operations. I think we'd want to error here if this was somehow reachable

beacon_node/beacon_chain/src/data_column_verification.rs
eserilev: deleted

beacon_node/beacon_chain/src/beacon_chain.rs
eserilev: We should have the gloas block in fork choice by the time we are processing columns over rpc. I added a TODO about potentially separating this to a different code path

beacon_node/beacon_chain/src/test_utils.rs
eserilev: done

beacon_node/beacon_chain/src/observed_data_sidecars.rs
eserilev: this no longer uses Option

consensus/types/src/data/data_column_sidecar.rs
eserilev: removed

beacon_node/beacon_chain/src/observed_data_sidecars.rs
eserilev: e9682f1

eserilev: tests have been updated

eserilev:
This should be good to go

jimmygchen (approved):
LGTM, thanks @eserilev and @pawanjay176! let's merge

I'll make a separate PR to enable beaocn chain tests for gloas so they get run on CI.

jimmygchen:
@mergify requeue

jimmygchen:
> LGTM, thanks @eserilev and @pawanjay176! let's merge
> 
> I'll make a separate PR to enable beaocn chain tests for gloas so they get run on CI.

ah just checked the tests and looks like it needs block production to work before we can enable it - i'll add this to the planning issue.

PR #8683, opened 2026-01-20
Remove libp2p multiaddress 

jxs:
After a user reported on Discord reported having problems starting Lighthouse with discovery disabled and providing `--boot-nodes`
I noticed that when a bootnode is provided as ENR lighthouse connects to it directly using libp2p, when a bootnode is a multiaddress it needs to be a UDP multiaddress so that Lighthouse uses Discovery to fetch that node's ENR.
But then later when bootstrapping libp2p, Lighthouse tries to dial those multiaddresses directly from libp2p if they are TCP, thing is, they won't be, as that was not allowed when the config was parsed.
Lighthouse also supports `libp2p-nodes` config flag that dials the addresses provided in the same way as when providing a valid `ENR`.

This PR basically removes the `--libp2p-nodes` flag and uniforms on `boot-nodes` flag allowing users to pass TCP multiaddresses with the `--boot-nodes` flag

### DIFF
--- a/beacon_node/src/cli.rs
+++ b/beacon_node/src/cli.rs
@@ -364,7 +364,7 @@ pub fn cli_app() -> Command {
                 .long("libp2p-addresses")
                 .value_name("MULTIADDR")
                 .help("One or more comma-delimited multiaddrs to manually connect to a libp2p peer \
-                       without an ENR. DEPRECATED. The --libp2p-nodes flag is deprecated and replaced by --boot-nodes")
+                       without an ENR. DEPRECATED. The --libp2p-addresses flag is deprecated and replaced by --boot-nodes")
                 .action(ArgAction::Set)
                 .display_order(0)
         )
--- a/beacon_node/src/cli.rs
+++ b/beacon_node/src/cli.rs
@@ -359,15 +359,6 @@ pub fn cli_app() -> Command {
                 This disables this feature, fixing the ENR's IP/PORT to those specified on boot.")
                 .display_order(0)
         )
-        .arg(
-            Arg::new("libp2p-addresses")
-                .long("libp2p-addresses")
-                .value_name("MULTIADDR")
-                .help("One or more comma-delimited multiaddrs to manually connect to a libp2p peer \
-                       without an ENR.")
-                .action(ArgAction::Set)
-                .display_order(0)
-        )
         // NOTE: This is hide because it is primarily a developer feature for testnets and
         // debugging. We remove it from the list to avoid clutter.
         .arg(
--- a/beacon_node/src/cli.rs
+++ b/beacon_node/src/cli.rs
@@ -359,6 +359,15 @@ pub fn cli_app() -> Command {
                 This disables this feature, fixing the ENR's IP/PORT to those specified on boot.")
                 .display_order(0)
         )
+        .arg(
+            Arg::new("libp2p-addresses")
+                .long("libp2p-addresses")
+                .value_name("MULTIADDR")
+                .help("One or more comma-delimited multiaddrs to manually connect to a libp2p peer \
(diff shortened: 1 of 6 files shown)


ackintosh (approved):
Thanks @jxs for the cleanup! ✨ Looks good to me.

One small thought: how about having a deprecation period for the  `--libp2p-nodes` flag to preserve CLI compatibility? We usually do this for CLI changes -- keeping the old flag as an alias, emitting a warning when used ( e.g. the `--logfile` deprecation in https://github.com/sigp/lighthouse/pull/7723).

pawanjay176 (approved):
This looks clean. I had no idea we accepted multiaddrs in `boot-nodes` as well.
Agree with @ackintosh about the deprecation instead of delete. I think some existing infra might depend on the flag.

ackintosh:
@jxs @pawanjay176 Also, I've filed a PR to improve the UX for the case where a user specifies a UDP multiaddr with `--boot-nodes` together with `--disable-discovery`. It would be great if you could take a look when you have some time.

https://github.com/jxs/lighthouse/pull/6

jxs (commented):
addressed Akihito's suggestion and re-introduced the flag with deprecation notice, PTAL

beacon_node/src/config.rs
```diff
@@ -1210,7 +1204,9 @@ pub fn set_network_config(
         config.boot_nodes_multiaddr = multiaddrs;
     }
 
+    // DEPRECATED: can be removed in v8.2.0./v9.0.0
     if let Some(libp2p_addresses_str) = cli_args.get_one::<String>("libp2p-addresses") {
+        warn!("The --libp2p-nodes flag is deprecated and replaced by --boot-nodes");
```
ackintosh:
```suggestion
        warn!("The --libp2p-addresses flag is deprecated and replaced by --boot-nodes");
```

beacon_node/src/cli.rs
```diff
@@ -364,7 +364,7 @@ pub fn cli_app() -> Command {
                 .long("libp2p-addresses")
                 .value_name("MULTIADDR")
                 .help("One or more comma-delimited multiaddrs to manually connect to a libp2p peer \
-                       without an ENR.")
+                       without an ENR. DEPRECATED. The --libp2p-nodes flag is deprecated and replaced by --boot-nodes")
```
ackintosh:
```suggestion
                       without an ENR. DEPRECATED. The --libp2p-addresses flag is deprecated and replaced by --boot-nodes")
```

book/src/help_bn.md
```diff
@@ -225,7 +225,8 @@ Options:
           be careful to avoid filling up their disks.
       --libp2p-addresses <MULTIADDR>
           One or more comma-delimited multiaddrs to manually connect to a libp2p
-          peer without an ENR.
+          peer without an ENR. DEPRECATED. The --libp2p-nodes flag is deprecated
```
ackintosh:
```suggestion
          peer without an ENR. DEPRECATED. The --libp2p-addresses flag is deprecated
```

ackintosh (commented):
Thanks @jxs!

ackintosh (approved):
LGTM!

michaelsproul:
@mergify dequeue

michaelsproul:
Just quickly dequeueing so I can merge another PR which bypasses mergify:

- https://github.com/sigp/lighthouse/pull/8778

Looks like CI for the book might also be failing?

jxs:
> Just quickly dequeueing so I can merge another PR which bypasses mergify:
> 
> * [Merge v8.1.0 (stable) into unstable #8778](https://github.com/sigp/lighthouse/pull/8778)
> 
> Looks like CI for the book might also be failing?

yup, thanks Michael. Addressed

michaelsproul:
@mergify queue

PR #8684, opened 2026-01-20
subscribe to attest and sync duties polling completion

shane-moore:
## Issue Addressed
Allow downstream consumers, i.e. anchor, to subscribe to attestation and sync duties BN polling completion events

## Proposed Changes
Add watch channels to `DutiesService` that signal when attestation and sync committee duty polling completes. Signals fire after duties are stored in the map but before background selection proof pre-computation.

## Additional Info
Useful for DVT middleware that needs to know when duties are available to begin its own processing

### DIFF
--- a/validator_client/validator_services/src/duties_service.rs
+++ b/validator_client/validator_services/src/duties_service.rs
@@ -28,7 +28,10 @@ use std::sync::Arc;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::time::Duration;
 use task_executor::TaskExecutor;
-use tokio::{sync::mpsc::Sender, time::sleep};
+use tokio::{
+    sync::{mpsc::Sender, watch},
+    time::sleep,
+};
 use tracing::{debug, error, info, warn};
 use types::{ChainSpec, Epoch, EthSpec, Hash256, SelectionProof, Slot};
 use validator_metrics::{ATTESTATION_DUTY, get_int_gauge, set_int_gauge};
@@ -373,6 +376,9 @@ impl<S, T> DutiesServiceBuilder<S, T> {
     }
 
     pub fn build(self) -> Result<DutiesService<S, T>, String> {
+        let (attesters_poll_tx, _) = watch::channel(Slot::default());
+        let (sync_poll_tx, _) = watch::channel(Slot::default());
+
         Ok(DutiesService {
             attesters: Default::default(),
             proposers: Default::default(),
@@ -394,6 +400,8 @@ impl<S, T> DutiesServiceBuilder<S, T> {
             enable_high_validator_count_metrics: self.enable_high_validator_count_metrics,
             selection_proof_config: self.attestation_selection_proof_config,
             disable_attesting: self.disable_attesting,
+            attesters_poll_tx,
+            sync_poll_tx,
         })
     }
 }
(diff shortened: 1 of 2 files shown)


validator_client/validator_services/src/duties_service.rs
```diff
@@ -974,12 +996,19 @@ async fn poll_beacon_attesters_for_epoch<S: ValidatorStore + 'static, T: SlotClo
...
+        .unwrap_or_default();
+
     // No need to bother the BN if we don't have any validators.
     if local_indices.is_empty() {
         debug!(
             %epoch,
             "No validators, not downloading duties"
         );
+        // Signal that sync committee duty polling is complete for this slot.
+        duties_service.attesters_poll_tx.send_replace(current_slot);
```
michaelsproul, line 1011: Shouldn't this be keyed by `epoch`? I'm thinking of a case where the attestation duties update is slow to run, so a slot that does not match `epoch` is sent. E.g. current slot is 65, but `epoch` is still at 0.

This also doesn't seem to handle the fact this function is called twice with different epoch values:

https://github.com/sigp/lighthouse/blob/1476c20cfc9a5272fe05d81b5f9bec4dd79fc23a/validator_client/validator_services/src/duties_service.rs#L816-L823

https://github.com/sigp/lighthouse/blob/1476c20cfc9a5272fe05d81b5f9bec4dd79fc23a/validator_client/validator_services/src/duties_service.rs#L841-L845

i.e. right now we could send the same `slot` (the current slot) for both of these updates.

Maybe we should communicate an increasing epoch to downstream consumers? And allow updates with equal epochs to cover reorgs?

validator_client/validator_services/src/duties_service.rs
```diff
@@ -1020,8 +1049,10 @@ async fn poll_beacon_attesters_for_epoch<S: ValidatorStore + 'static, T: SlotClo
             .collect::<Vec<_>>()
     };
 
+    // No validators have conflicting (epoch, dependent_root) values or missing duties for the epoch.
     if validators_to_update.is_empty() {
-        // No validators have conflicting (epoch, dependent_root) values or missing duties for the epoch.
+        // Signal that sync committee duty polling is complete for this slot.
+        duties_service.attesters_poll_tx.send_replace(current_slot);
```
michaelsproul, line 1055: Do we need to fire an event here? This branch covers the case where nothing has changed, so presumably we don't need to communicate anything to downstream, do we?

validator_client/validator_services/src/duties_service.rs
shane-moore: Closing this PR  cuz after further investigation and kurtosis testing, we've determined anchor doesn't actually need this.                                                                                    
                                                                                                                                                                                        
 Context: In anchor, we cache total attestation and sync duties for each slot at slot start by reading from `duties_service.attesters(slot)` and `sync_duties.get_duties_for_slot()`.  We then use this total when `produce_selection_proofs` (attestation and sync) are called in anchor. The total needs to be correct, and we initially thought we needed to wait for `DutiesService` to finish polling the beacon node before reading because I was worried about node restarts or re-orgs invalidating the attesters and sync duty caches, hence this PR

but after further investigation, it seems that the `fill_in_selection_proofs` call waits til slot N + 1 before calling `produce_selection_proofs`, so it should be safe to read the total `duties_service.attesters(slot)`  at the beginning of slot N + 1, without waiting for an event
https://github.com/sigp/lighthouse/blob/39c542a37b30284a88f8d631d3cbe0b08c21eb06/validator_client/validator_services/src/duties_service.rs#L1297-L1302

And same reasoning for sync:
https://github.com/sigp/lighthouse/blob/39c542a37b30284a88f8d631d3cbe0b08c21eb06/validator_client/validator_services/src/sync.rs#L604-L609                                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                  
Basically, in both cases, `fill_in_selection_proofs` sleeps until the next slot before calling produce_selection_proofs, so the cache write happens in slot N and reads happen in slot N+1 so seems like no race condition                                                 
                                                                                                                                                                                        
Appreciate your review points, both were valid concerns (slot vs epoch keying, firing events on no-op branches), but since we're removing the wait logic entirely, no need        
                                                                                                                                                                                        
Thanks for the review!

PR #8692, opened 2026-01-22
Implement Gloas withdrawals and refactor

michaelsproul:
This PR implements the Gloas logic for withdrawals and refactors the Capella/Electra logic to match the refactored spec. There should be no consensus changes for the versions prior to Gloas (just refactoring).


michaelsproul:
This is passing the Gloas spec tests, but there's still some more polish I'd like to apply before merging.

consensus/state_processing/src/per_block_processing.rs
```diff
@@ -513,190 +520,3 @@ pub fn compute_timestamp_at_slot<E: EthSpec>(
...
-}
-
-/// Apply withdrawals to the state.
-/// TODO(EIP-7732): abstract this out and create gloas variant
-pub fn process_withdrawals<E: EthSpec, Payload: AbstractExecPayload<E>>(
-    state: &mut BeaconState<E>,
-    payload: Payload::Ref<'_>,
-    spec: &ChainSpec,
-) -> Result<(), BlockProcessingError> {
-    if state.fork_name_unchecked().capella_enabled() {
```
michaelsproul, line 645: I've moved this check up to `process_block`/`per_block_processing` so that this function doesn't just no-op in case Capella is disabled. We should not be calling `process_withdrawals` prior to Capella, that is a serious error.

michaelsproul:
Ready for review!

consensus/state_processing/src/per_block_processing/withdrawals.rs
```diff
@@ -0,0 +1,520 @@
...
+pub fn get_expected_withdrawals<E: EthSpec>(
+    state: &BeaconState<E>,
+    spec: &ChainSpec,
+) -> Result<
+    (
+        Withdrawals<E>,
+        Option<usize>,
+        Option<usize>,
+        Option<u64>,
+        u64,
```
michaelsproul: One change we should probably make is adding an `ExpectedWithdrawals` type with named fields so that these return values don't just look completely arbitrary.

consensus/state_processing/src/per_block_processing/withdrawals.rs
```diff
@@ -0,0 +1,520 @@
...
+    state: &BeaconState<E>,
+    withdrawal_index: &mut u64,
+    withdrawals: &mut Vec<Withdrawal>,
+) -> Result<Option<usize>, BlockProcessingError> {
+    let Ok(builder_pending_withdrawals) = state.builder_pending_withdrawals() else {
+        // Pre-Gloas, nothing to do.
+        return Ok(None);
+    };
+
+    let withdrawals_limit = E::max_withdrawals_per_payload();
```
eserilev, line 88: I believe this should be `E::max_withdrawals_per_payload() - 1`
https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/beacon-chain.md#new-get_builder_withdrawals

consensus/state_processing/src/per_block_processing/withdrawals.rs
```diff
@@ -0,0 +1,520 @@
...
+
+    let mut processed_count = 0;
+    for withdrawal in builder_pending_withdrawals {
+        let has_reached_limit = withdrawals.len() == withdrawals_limit;
```
eserilev, line 102: I think since `withdrawals.len()` monotonically increases by 1 this doesn't matter but the spec does have this as:
`withdrawals.len() >= withdrawals_limit`

consensus/state_processing/src/per_block_processing/withdrawals.rs
michaelsproul: Holy shit, good catch. Bit worrying that isn't covered by spec tests.

michaelsproul: Oh, actually this must have changed since v1.7.0-alpha.1:

https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.1/specs/gloas/beacon-chain.md#new-get_builder_withdrawals

michaelsproul: I think we have to leave this until we get a new spec release, but I will add a TODO.

consensus/state_processing/src/per_block_processing/withdrawals.rs
michaelsproul: This is also a change since v1.7.0-alpha.1

https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.1/specs/gloas/beacon-chain.md#withdrawals

consensus/state_processing/src/per_block_processing/withdrawals.rs
eserilev: ah yeah sounds good. Sooo much spec churn 😅

consensus/state_processing/src/per_block_processing/withdrawals.rs
```diff
@@ -0,0 +1,528 @@
...
+        }))
+    } else {
+        Ok(ExpectedWithdrawals::Capella(ExpectedWithdrawalsCapella {
+            withdrawals,
+            processed_sweep_withdrawals_count,
+        }))
+    }
+}
+
+pub fn get_builder_withdrawals<E: EthSpec>(
```
eserilev, line 77: nit: Might be nice to add a link to the spec as a doc comment
https://ethereum.github.io/consensus-specs/specs/gloas/beacon-chain/#new-get_builder_withdrawals

consensus/state_processing/src/per_block_processing/withdrawals.rs
```diff
@@ -0,0 +1,528 @@
...
+            address: withdrawal.fee_recipient,
+            amount: withdrawal.amount,
+        });
+        withdrawal_index.safe_add_assign(1)?;
+        processed_count.safe_add_assign(1)?;
+    }
+    Ok(Some(processed_count))
+}
+
+pub fn get_pending_partial_withdrawals<E: EthSpec>(
```
eserilev, line 122: nit: might be nice to add a link to the spec as a doc comment
https://ethereum.github.io/consensus-specs/specs/electra/beacon-chain/#new-get_pending_partial_withdrawals

consensus/state_processing/src/per_block_processing/withdrawals.rs
eserilev: justin traglia mentioned we could pin to a specific nightly spec test release version if we wanted to make this change now. not sure how I feel about that, but just wanted to mention that its an option

consensus/state_processing/src/per_block_processing/withdrawals.rs
```diff
@@ -0,0 +1,528 @@
...
+        // Pre-Electra nothing to do.
+        return Ok(None);
+    };
+    let epoch = state.current_epoch();
+
+    let withdrawals_limit = std::cmp::min(
+        withdrawals
+            .len()
+            .safe_add(spec.max_pending_partials_per_withdrawals_sweep as usize)?,
+        E::max_withdrawals_per_payload(),
```
eserilev: I think this should be `E::max_withdrawals_per_payload() - 1`
https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.1/specs/electra/beacon-chain.md#new-get_pending_partial_withdrawals

This edge case seems to be missing a spec test?

eserilev: also kinda crazy that they're going back and changing electra like this...

consensus/state_processing/src/per_block_processing/withdrawals.rs
```diff
@@ -0,0 +1,528 @@
...
+        let next_index = state
+            .next_withdrawal_builder_index()?
+            .safe_add(processed_builders_sweep_count)?;
+        let next_builder_index = next_index.safe_rem(state.builders()?.len() as u64)?;
+        *state.next_withdrawal_builder_index_mut()? = next_builder_index;
+    }
+    Ok(())
+}
+
+fn update_next_withdrawal_validator_index<E: EthSpec>(
```
eserilev, line 404: nit: doc comment link to spec
https://ethereum.github.io/consensus-specs/specs/capella/beacon-chain/#new-update_next_withdrawal_validator_index

consensus/state_processing/src/per_block_processing/withdrawals.rs
```diff
@@ -0,0 +1,528 @@
...
+    state: &mut BeaconState<E>,
+    processed_partial_withdrawals_count: u64,
+) -> Result<(), BlockProcessingError> {
+    state
+        .pending_partial_withdrawals_mut()?
+        .pop_front(processed_partial_withdrawals_count as usize)?;
+    Ok(())
+}
+
+fn update_next_withdrawal_builder_index<E: EthSpec>(
```
eserilev, line 389: nit: doc comment link to spec:
https://ethereum.github.io/consensus-specs/specs/gloas/beacon-chain/#new-update_next_withdrawal_builder_index

consensus/state_processing/src/per_block_processing/withdrawals.rs
```diff
@@ -0,0 +1,528 @@
...
+    state: &mut BeaconState<E>,
+    processed_builder_withdrawals_count: u64,
+) -> Result<(), BlockProcessingError> {
+    state
+        .builder_pending_withdrawals_mut()?
+        .pop_front(processed_builder_withdrawals_count as usize)?;
+    Ok(())
+}
+
+fn update_pending_partial_withdrawals<E: EthSpec>(
```
eserilev, line 379: nit: doc comment link to spec
https://ethereum.github.io/consensus-specs/specs/electra/beacon-chain/#new-update_pending_partial_withdrawals

consensus/state_processing/src/per_block_processing/withdrawals.rs
```diff
@@ -0,0 +1,528 @@
...
+
+fn update_payload_expected_withdrawals<E: EthSpec>(
+    state: &mut BeaconState<E>,
+    withdrawals: &Withdrawals<E>,
+) -> Result<(), BlockProcessingError> {
+    *state.payload_expected_withdrawals_mut()? = List::new(withdrawals.to_vec())?;
+    Ok(())
+}
+
+fn update_builder_pending_withdrawals<E: EthSpec>(
```
eserilev, line 369: nit: doc comment link to spec
https://ethereum.github.io/consensus-specs/specs/gloas/beacon-chain/#new-update_builder_pending_withdrawals

consensus/state_processing/src/per_block_processing/withdrawals.rs
```diff
@@ -0,0 +1,528 @@
...
+    withdrawals: &Withdrawals<E>,
+) -> Result<(), BlockProcessingError> {
+    // Update the next withdrawal index if this block contained withdrawals
+    if let Some(latest_withdrawal) = withdrawals.last() {
+        *state.next_withdrawal_index_mut()? = latest_withdrawal.index.safe_add(1)?;
+    }
+    Ok(())
+}
+
+fn update_payload_expected_withdrawals<E: EthSpec>(
```
eserilev, line 361: nit: doc comment link to spec
https://ethereum.github.io/consensus-specs/specs/gloas/beacon-chain/#new-update_payload_expected_withdrawals

consensus/state_processing/src/per_block_processing/withdrawals.rs
```diff
@@ -0,0 +1,528 @@
...
+) -> bool {
+    let has_sufficient_effective_balance =
+        validator.effective_balance >= spec.min_activation_balance;
+    let has_excess_balance = balance > spec.min_activation_balance;
+    validator.exit_epoch == spec.far_future_epoch
+        && has_sufficient_effective_balance
+        && has_excess_balance
+}
+
+fn update_next_withdrawal_index<E: EthSpec>(
```
eserilev, line 350: nit: doc link to spec
https://ethereum.github.io/consensus-specs/specs/capella/beacon-chain/#new-update_next_withdrawal_index

consensus/state_processing/src/per_block_processing/withdrawals.rs
```diff
@@ -0,0 +1,528 @@
...
+        .filter(|withdrawal| withdrawal.validator_index == validator_index)
+        .map(|withdrawal| withdrawal.amount)
+        .safe_sum()?;
+    state
+        .get_balance(validator_index as usize)?
+        .safe_sub(withdrawn)
+        .map_err(Into::into)
+}
+
+fn is_eligible_for_partial_withdrawals(
```
eserilev, line 337: nit: doc comment link to spec
https://ethereum.github.io/consensus-specs/specs/electra/beacon-chain/#new-is_eligible_for_partial_withdrawals

consensus/state_processing/src/per_block_processing/withdrawals.rs
```diff
@@ -0,0 +1,528 @@
...
+        validator_index = validator_index
+            .safe_add(1)?
+            .safe_rem(state.validators().len() as u64)?;
+        processed_count.safe_add_assign(1)?;
+    }
+
+    Ok(processed_count)
+}
+
+pub fn get_balance_after_withdrawals<E: EthSpec>(
```
eserilev, line 321: nit: doc comment link to spec 
https://ethereum.github.io/consensus-specs/specs/capella/beacon-chain/#new-get_balance_after_withdrawals

consensus/state_processing/src/per_block_processing/withdrawals.rs
```diff
@@ -0,0 +1,528 @@
...
+        let next_validator_index = state
+            .next_withdrawal_validator_index()?
+            .safe_add(spec.max_validators_per_withdrawals_sweep)?
+            .safe_rem(state.validators().len() as u64)?;
+        *state.next_withdrawal_validator_index_mut()? = next_validator_index;
+    }
+    Ok(())
+}
+
+pub fn apply_withdrawals<E: EthSpec>(
```
eserilev, line 431: nit: Doc comment link to spec
https://ethereum.github.io/consensus-specs/specs/gloas/beacon-chain/#modified-apply_withdrawals

consensus/state_processing/src/per_block_processing/withdrawals.rs
```diff
@@ -0,0 +1,528 @@
...
+        }
+    }
+    Ok(())
+}
+
+pub mod capella_electra {
+    use super::*;
+
+    /// Apply withdrawals to the state.
+    pub fn process_withdrawals<E: EthSpec, Payload: AbstractExecPayload<E>>(
```
eserilev, line 461: nit: doc comment link to spec
https://ethereum.github.io/consensus-specs/specs/electra/beacon-chain/#modified-process_withdrawals

consensus/state_processing/src/per_block_processing/withdrawals.rs
```diff
@@ -0,0 +1,528 @@
...
+
+        Ok(())
+    }
+}
+
+pub mod gloas {
+    use super::*;
+
+    /// Apply withdrawals to the state.
+    pub fn process_withdrawals<E: EthSpec>(
```
eserilev, line 501: nit: doc comment link to spec
https://ethereum.github.io/consensus-specs/specs/gloas/beacon-chain/#modified-process_withdrawals

eserilev (commented):
Feel free to take it or leave it with the nits

consensus/state_processing/src/per_block_processing/withdrawals.rs
michaelsproul: I don't want to do that because the devnet will presumably target this spec, or a new release which we can update to when it comes out

michaelsproul:
Gonna avoid linking to the specs because I find these links always go stale and become useless. I think we could say "equivalent to `X` in the spec" on each function, but that should already kind of be implied.

consensus/state_processing/src/per_block_processing/withdrawals.rs
michaelsproul: I think this might be the Gnosis fix, which is not reachable on minimal or mainnet specs. I'll dig in.

michaelsproul: I've made all our limits consistent with v1.7.0-alpha.1, which meant adding this change, and reverting one of the others:

- 1f4a43e564122dd17ec94b0ec52aee8f6850cafc

My understanding remains that these limit differences aren't relevant on mainnet where `MAX_PENDING_PARTIALS_PER_WITHDRAWALS_SWEEP` (=8) is less than the per-payload limit (`MAX_WITHDRAWALS_PER_PAYLOAD`=16). However these _were_ previously equal on Gnosis (both 8, now 6 and 8).

I've also asked @jtraglia in a DM about how we plan to test these changes, given they are intended to be unreachable.

michaelsproul: There are more tests on the way in the next spec release. I am fine to merge this as-is and we can revisit once tests come out for alpha.2

michaelsproul:
Ready for re-review. There are quite a lot of TODOs, but these are mostly reminders for us to make changes at the next spec release.

consensus/state_processing/src/per_block_processing/withdrawals.rs
jtraglia: > also kinda crazy that they're going back and changing electra like this...

Yes, this refactor was long overdue 😅 with Gloas changes, these functions became even more complex.

> I think this might be the Gnosis fix, which is not reachable on minimal or mainnet specs.

The gnosis issue was considered during the refactor. I wanted to ensure those two limits could equal each other, which wasn't possible before.

> I am fine to merge this as-is and we can revisit once tests come out for alpha.2

Hmm the current limit (for this function, in this PR) doesn't look right to me. Using this limit in `get_pending_partial_withdrawals` would cause a consensus issue. The limit in this function should contain the `- 1` in alpha.1 specs. So yeah, we definitely were missing a test for this then. See: https://ethspec.tools/#specs/v1.7.0-alpha.1/functions-get_pending_partial_withdrawals

Note that the limit issue fix was made in [this PR](https://github.com/ethereum/consensus-specs/pull/4832). It fixed the limits in `get_builder_withdrawals` and `get_builders_sweep_withdrawals`. The issue _should not_ affect pre-Gloas code; that's why we didn't create another release.

michaelsproul: We've updated this branch now to match v1.7.0-alpha.1 everywhere:

- 1f4a43e564122dd17ec94b0ec52aee8f6850cafc

eserilev (approved):
nice, LGTM!

Do we wait until after the 8.1 release to merge this?

eserilev:
Sounds like alpha.2 is being released early next week. Might make sense to wait on that and make some of the TODO changes before merging

michaelsproul:
I wouldn't mind just merging this now seeing as it's ready, and then a small PR for alpha.2

jimmygchen:
@mergify requeue

PR #8694, opened 2026-01-22
Remove `data` dependency from from `core` module in `consensus/types

macladson:
## Issue Addressed

#8652

## Proposed Changes

Implements "simplified" versions of `max_blobs_by_root_request` and `max_data_columns_by_root_request` which do not depend on type information from the `data` module. I've also added tests which test the original implementation against the simplified one to ensure they don't deviate.

Also moves `all_data_column_sidecar_subnets` from a method on `ChainSpec` to a function which just takes `ChainSpec` as an argument.

## Additional Info

This marks the final module blocking the `core` feature set (other than `test_utils`). I will add a `full` feature in a future PR which will allow compilation of `consensus/types` with just the `core` and `fork` modules.

### DIFF
--- a/consensus/types/src/core/chain_spec.rs
+++ b/consensus/types/src/core/chain_spec.rs
@@ -8,7 +8,7 @@ use safe_arith::{ArithError, SafeArith};
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use serde_utils::quoted_u64::MaybeQuoted;
 use ssz::Encode;
-use ssz_types::{RuntimeVariableList, VariableList};
+use ssz_types::RuntimeVariableList;
 use tree_hash::TreeHash;
 
 use crate::{
@@ -16,7 +16,6 @@ use crate::{
         APPLICATION_DOMAIN_BUILDER, Address, ApplicationDomain, EnrForkId, Epoch, EthSpec,
         EthSpecId, ExecutionBlockHash, Hash256, MainnetEthSpec, Slot, Uint256,
     },
-    data::{BlobIdentifier, DataColumnSubnetId, DataColumnsByRootIdentifier},
     fork::{Fork, ForkData, ForkName},
 };
 
@@ -821,10 +820,6 @@ impl ChainSpec {
             ),
             _ => Some(std::cmp::max(fork_epoch, blob_retention_epoch)),
         }
-    }
-
-    pub fn all_data_column_sidecar_subnets(&self) -> impl Iterator<Item = DataColumnSubnetId> {
-        (0..self.data_column_sidecar_subnet_count).map(DataColumnSubnetId::new)
     }
 
     /// Worst-case compressed length for a given payload of size n when using snappy.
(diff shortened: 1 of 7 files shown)


consensus/types/src/core/chain_spec.rs
```diff
@@ -2110,37 +2105,22 @@ fn max_blocks_by_root_request_common(max_request_blocks: u64) -> usize {
...
-    RuntimeVariableList::<BlobIdentifier>::new(
-        vec![empty_blob_identifier; max_request_blob_sidecars],
-        max_request_blob_sidecars,
-    )
-    .expect("creating a RuntimeVariableList of size `max_request_blob_sidecars` should succeed")
-    .as_ssz_bytes()
-    .len()
+pub(crate) fn max_blobs_by_root_request_common(max_request_blob_sidecars: u64) -> usize {
+    (max_request_blob_sidecars as usize)
+        .safe_mul(40)
```
pawanjay176: can we add a comment here for how we got 40? maybe extract that into a variable like `let blob_identifier_ssz_size = 40`?

pawanjay176: also mention that the `List` type doesn't add any byte ovehead

consensus/types/src/core/chain_spec.rs
```diff
@@ -2110,37 +2105,22 @@ fn max_blocks_by_root_request_common(max_request_blocks: u64) -> usize {
...
-    )
-    .expect("creating a RuntimeVariableList of size `max_request_blocks` should succeed")
-    .as_ssz_bytes()
-    .len()
+pub(crate) fn max_data_columns_by_root_request_common<E: EthSpec>(
+    max_request_blocks: u64,
+) -> usize {
+    let bytes_per_element = 8_usize
+        .safe_mul(E::number_of_columns())
+        .and_then(|b| b.safe_add(40))
```
pawanjay176: same as above

consensus/types/src/data/data_column_sidecar.rs
```diff
@@ -170,3 +170,43 @@ impl From<SszError> for DataColumnSidecarError {
...
+            max_request_blocks,
+        )
+        .expect("creating a RuntimeVariableList of size `max_request_blocks` should succeed")
+        .as_ssz_bytes()
+        .len()
+    }
+
+    #[test]
+    fn max_data_columns_by_root_request_matches_simplified() {
+        for n in [0, 1, 2, 8, 16, 32, 64, 128, 256, 512, 1024] {
```
pawanjay176, line 204: maybe we can make sure it also works for `default_max_request_blocks` too since that's what's actually gets used?

pawanjay176 (commented):
Looks good. Just a few nits about docs and a suggestion for a test

consensus/types/src/data/data_column_sidecar.rs
macladson: Isn't the value that actually gets used `128`?

pawanjay176: Ohh yeah true. I don't think that changes. This is good then

pawanjay176 (approved):

PR #8695, opened 2026-01-22
Disable `legacy-arith` by default in `consensus/types`

macladson:
## Issue Addressed
Currently, `consensus/types` cannot build with `no-default-features` since we use "legacy" standard arithmetic operations.

## Proposed Changes

- Remove the offending arithmetic to fix compilation.
- Rename `legacy-arith` to `saturating-arith` and disable it by default.

## Additional Info

We could consider removing support for standard arithmetic entirely, although I think having an opt-in feature which makes the dangers of enabling it clear is an okay solution for now, and won't break any external users (and our own `state_processing` crate).

Incidentally, the ergonomics around using explicit methods such as `saturating_add` with integer types (which is available even without `saturating-arith`) is a bit awkward, since it requires casting to u64 first.


pawanjay176 (approved):
This LGTM, but it would be good if someone else can also take a look

consensus/state_processing/Cargo.toml
```diff
@@ -40,7 +39,7 @@ test_random_derive = { path = "../../common/test_random_derive" }
 tracing = { workspace = true }
 tree_hash = { workspace = true }
 typenum = { workspace = true }
-types = { workspace = true }
+types = { workspace = true, features = ["saturating-arith"] }
```
michaelsproul: Not necessary

consensus/types/src/state/committee_cache.rs
```diff
@@ -79,7 +79,13 @@ impl CommitteeCache {
             .saturating_sub(spec.min_seed_lookahead)
             .saturating_sub(1u64);
 
-        if reqd_randao_epoch < state.min_randao_epoch() || epoch > state.current_epoch() + 1 {
+        if reqd_randao_epoch < state.min_randao_epoch()
```
michaelsproul, line 80: Should also remove the `allow` lint in this file

macladson: There are still some more `#[allow(clippy::arithmetic_side_effects)]` in `beacon_state.rs` and in `validator.rs`. Those seem a bit more involved to remove so I have left them for now

pawanjay176: Looks like this has been resolved except the ones in validator.rs?

macladson: There is still the SSZ one in `beacon_state.rs`:

https://github.com/sigp/lighthouse/blob/8d72cc34ebe3c94c0856a0c9534943179e11b39f/consensus/types/src/state/beacon_state.rs#L3173-L3174

pawanjay176 (approved):
LGTM. I'm fine with merging without handling the more involved allow lints see https://github.com/sigp/lighthouse/pull/8695#discussion_r2791754928.
Will wait for ✅  from @michaelsproul

pawanjay176:
Gonna go ahead and merge this one to avoid it going stale

Issue #8633, opened 2026-01-07
Move some DA checker functionality to `CustodyContext`


eserilev:
## Description

`DataAvailabilityChecker`: a cache to hold fully valid data that can't be imported to fork choice yet

`CustodyContext`: Contains all the information the node requires to calculate the  number of  blobs/columns to be custodied when checking for DA.

We have the following functions implemented for `DataAvailabilityChecker` that should be moved to `CustodyContext` 

- `blobs_required_for_epoch`
- `data_columns_required_for_block`
- `data_availability_boundary`
- `is_deneb` (maybe we probably remove this completely, I don't think it's used anywhere)

This will help keep a better separation of concerns between the `DataAvailabilityChecker` and `CustodyContext`  types

## Additional Notes

As part of this refactor:
- [x] `DataAvailabilityChecker` no longer needs a reference to slot clock, slot clock should move to `CustodyContext`. https://github.com/sigp/lighthouse/pull/8696
- [x] `RpcBlock::new` should accept a `CustodyContext` as an argument instead of a `DataAvailabilityChecker` https://github.com/sigp/lighthouse/pull/8696
- [ ] Move `AvailableBlock` and `AvailableBlockData` from `data_availability_checker.rs` to `block_verification_types.rs`
- [ ] Delete `enum DataSidecars` and replace its usage with `AvailableBlockData` 
- [ ] Explore passing `RpcBlock::FullyAvailable` to range and backfill sync
- [ ] Ensure all `TODO(#8633)` references in the codebase are resolved

eserilev:
Completed in #9515

PR #8696, opened 2026-01-22
Move DA boundary checks to CustodyContext

eserilev:
Partially addresses #8633

This PR moves data availability boundary and blob/column requirement checks from DataAvailabilityChecker to       
CustodyContext, consolidating custody-related logic in one place.

We need this because were creating a da checker v2 for gloas, and duplicating this logic across both DA checkers would be wasteful

Note that this PR is currently built on top of https://github.com/sigp/lighthouse/pull/8424 so this is blocked until #8424 can be merged


eserilev:
This should be ready for a review. If its too much code churn we don't need to do this. But it would be nice considering were going to have a duplicate da checker for gloas

pawanjay176 (commented):
> moves data availability boundary and blob/column requirement checks from DataAvailabilityChecker to
CustodyContext, consolidating custody-related logic in one place.

I'm not very convinced of this premise. 
I don't think the da boundary is a custody related logic. Its the same regardless of the node's custody requirements. 
The `data_columns_required_for_epoch` function doesn't use any custody related parameters either.

eserilev:
I'll close this, I think theres other ways to prevent code duplication. Thanks for taking a look!

Issue #8381, opened 2025-11-06
Update VC to submit beacon committee selections once per epoch


KaloyanTanev:
## Description

Something we've missed during the beacon committee selection submissions is that they are being sent once per slot, instead of once per epoch, at the beginning of the epoch. It generally works if only Lighthouse VC is used in a DV cluster or another VC that has the same issue (Lodestar and Nimbus are such).

BN Spec - https://ethereum.github.io/beacon-APIs/#/Validator/submitBeaconCommitteeSelections

## Version

`unstable`

## Present Behaviour

Submitting beacon committee selection once per slot, for the current slot.

## Expected Behaviour

Submitting beacon committee selection once per epoch, for the current and next epoch.

## Steps to resolve

Run a DV cluster with 1/2 Lighthouse VCs and 1/2 Teku VCs.

chong-he:
I read the information in the Obol issue and this is my understanding:

- Teku queries all 64 validators at the start of the epoch
- Lighthouse queries ~2 validators each slot in the epoch
- Only first slot reaches threshold and can aggregate
- By the time it reaches the second slot and onwards, Teku requests have timeout and didn't reach threshold, thus aggregated attestations fail

Is that right? 
And this does not affect the `sync_committee_selections` endpoint (which also queries every slot)

KaloyanTanev:
Yap, that's mostly correct. One difference though is that the threshold is reached for each slot, at the start of the slot (when Lighthouse submits its partial selection proof). However, an actual aggregation is never produced, because Teku VC is still hanging to receive its 64 full selection proofs. Keep in mind the Teku's request (in the example) consists of 64 partial selection proofs, so it's expecting a response of 64 full selection proofs. We cannot return only a portion of them. Meaning that when Lighthouse receives the full selection proof for the slot and then submits its partial signatures, Teku is still hanging and not submitting its.

Last I checked, `sync_committee_selections` are submitted by all clients per slot.

michaelsproul:
Addressed by:

- https://github.com/sigp/lighthouse/pull/8699

PR #8699, opened 2026-01-27
Call beacon_committee_selections only once per epoch

chong-he:
## Issue Addressed

* #8381 

## Proposed Changes

Add another if in the if/else branch for selections_endpoint case, and leaving the current code for non-selections_endpoint case mostly untouched. Now the if/else branch in `fill_in_selection_proofs` has 3 branches: 
- 1 for DVT with selections_endpoint (added in this PR)
- 1 for DVT without selections_endpoint (Anchor) (untouched)
- Non-DVT (untouched)

Tested and it's working, also thanks very much to @KaloyanTanev for helping to test and confirmed that the calling has reduced and aggregated attestations are still working

### DIFF
--- a/validator_client/validator_services/src/duties_service.rs
+++ b/validator_client/validator_services/src/duties_service.rs
@@ -1389,7 +1389,7 @@ async fn fill_in_selection_proofs<S: ValidatorStore + 'static, T: SlotClock + 's
                             .is_aggregator(duty.committee_length as usize, &duties_service.spec)
                             .map_err(Error::<S::Error>::InvalidModulo)
                         {
-                            // Aggregator, store the selection_proof in result_hashmap
+                            // Aggregator, return the result
                             Ok(true) => Ok((duty, Some(selection_proof))),
                             // Not an aggregator, do nothing and continue
                             Ok(false) => continue,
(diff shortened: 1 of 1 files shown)


KaloyanTanev (approved):

shane-moore:
lgtm!

validator_client/validator_services/src/duties_service.rs
```diff
@@ -221,6 +165,69 @@ async fn make_selection_proof<S: ValidatorStore + 'static, T: SlotClock>(
...
+            }
+        })
+        .collect::<FuturesUnordered<_>>()
+        .try_collect::<Vec<_>>()
+        .await?;
+
+    let epoch = duties
+        .first()
+        .map(|attester_data| attester_data.slot.epoch(S::E::slots_per_epoch()))
+        .unwrap_or_default();
```
michaelsproul, line 198: This should probably be an error (we should always have at least one duty).

michaelsproul (approved):
Looks good to me, I'm confident this doesn't affect the non-DVT codepath, and the logic for selections (Obol) looks correct.

validator_client/validator_services/src/duties_service.rs
michaelsproul: Happy to leave as-is though.

PR #8700, opened 2026-01-27
Gloas gossip boilerplate

eserilev:
## Issue Addressed

All the required boilerplate for gloas gossip. We'll include the gossip message processing logic in a separate PR

### DIFF
--- a/beacon_node/beacon_processor/src/lib.rs
+++ b/beacon_node/beacon_processor/src/lib.rs
@@ -937,7 +937,7 @@ impl<E: EthSpec> BeaconProcessor<E> {
                         } else if let Some(item) = work_queues.unknown_block_attestation_queue.pop()
                         {
                             Some(item)
-                        // Check execution payload bids. Most proposers will equest bids directly from builders
+                        // Check execution payload bids. Most proposers will request bids directly from builders
                         // instead of receiving them over gossip.
                         } else if let Some(item) =
                             work_queues.gossip_execution_payload_bid_queue.pop()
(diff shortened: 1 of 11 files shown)


beacon_node/beacon_processor/src/lib.rs
```diff
@@ -914,6 +935,17 @@ impl<E: EthSpec> BeaconProcessor<E> {
                         } else if let Some(item) = work_queues.unknown_block_aggregate_queue.pop() {
                             Some(item)
                         } else if let Some(item) = work_queues.unknown_block_attestation_queue.pop()
+                        {
+                            Some(item)
+                        // Check execution payload bids. Most proposers will equest bids directly from builders
```
jimmygchen:
```suggestion
                        // Check execution payload bids. Most proposers will request bids directly from builders
```

jimmygchen (approved):
LGTM!

beacon_node/network/src/network_beacon_processor/mod.rs
```diff
@@ -423,6 +423,92 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+        let process_fn = move || {
+            processor.process_gossip_payload_attestation(
+                message_id,
+                peer_id,
+                *payload_attestation_message,
+            )
+        };
+
+        self.try_send(BeaconWorkEvent {
+            drop_during_sync: false,
```
jimmygchen: we may want to drop this during sync?

jimmygchen: I've flip this bool in https://github.com/sigp/lighthouse/pull/8700/commits/9a489274ded94c9d820452091beb550d82da396b

I believe this is correct, but let me know otherwise!

jimmygchen: All changes in this PR are additions and unreachable in Fulu, so I think we're good to merge!

jimmygchen:
@mergify requeue

beacon_node/network/src/network_beacon_processor/mod.rs
jimmygchen: Chat with Eitan and also dropping `GossipExecutionPayloadBid` event during sync

Issue #7011, opened 2025-02-19
Should lighthouse sync blocks outside da window?


dapplion:
The Deneb p2p makes it impossible for nodes to download blobs outside the data availability (da) window (approx 18 days from the clock slot).

During forward sync, Lighthouse currently imports these blocks outside the data availability (DA) retention window regardless of whether they were available.

Consider the following chain. Should a Lighthouse full node attest to this head?

<img width="723" alt="Image" src="https://github.com/user-attachments/assets/97c82bc0-4fb3-4873-82cd-bc769fbd34cd" />

And this one?

<img width="723" alt="Image" src="https://github.com/user-attachments/assets/7e470567-646f-4b78-859d-f4e788629f04" />

IMO it should not. A full node must independently check that blocks are valid and available, without relying on any majority. A finalized un-available block is still an un-available block that must be rejected.

Therefore LH should reject blocks from outside the DA window that it can't independently verify to be available. If a node runner has a node that has been stopped for more than 18 days, it should do _one_ of the following:
- Checkpoint sync into a finalized checkpoint inside the retention window
- Run the node with a flag like `--unsafe-da-sync`
- Manually tell the node that some block hash and its ancestors are available

pawanjay176:
Sorry finally got around to this. It makes sense to me now.

I agree that we should refuse to forward sync in this scenario and be explicit. On starup, we could check if `wall_clock_slot-head_slot >=da window` and exit with log a message asking the user to checkpoint sync or run with `--unsafe-da-sync`.

dapplion:
> I agree that we should refuse to forward sync in this scenario and be explicit. On starup, we could check if wall_clock_slot-head_slot >=da window and exit with log a message asking the user to checkpoint sync or run with --unsafe-da-sync.

Yeah that would be nice for UX. The check should be in sync too. When validating blocks we should only allow blocks to not have data if that flag is present

eserilev:
Talked to Lion about this the other day. I don't think sync needs to be aware of this flag because sync code paths will never be reached if the nodes head slot is outside the da window (lighthouse will just shut down and return an error message). If unsafe forward sync is enabled via this new flag, sync will just act like it does today

PR #8706, opened 2026-01-27
feat: new cli flag to enable unsafe sync

PoulavBhowmick03:
## Issue Addressed

Addresses #7011 

## Proposed Changes

1. Introduced a new CLI flag: `--enable-unsafe-sync` , disabled by default
2. DA window check in `builder.rs` on start
    - After loading the head snapshot, compares head epoch against the DA boundary
    - If head is outside the DA window, logs error and exits
    - If `--enable-unsafe-sync` is true, continues
  3. New config field: `enable_unsafe_sync`


beacon_node/src/cli.rs
```diff
@@ -1245,6 +1245,17 @@ pub fn cli_app() -> Command {
                 .help_heading(FLAG_HEADER)
                 .display_order(0)
         )
+        .arg(
+            Arg::new("enable-unsafe-sync")
```
eserilev, line 1249: Could we add a test for this new flag? There are example tests for other flags you can pretty much copy paste

beacon_node/beacon_chain/src/builder.rs
```diff
@@ -847,6 +847,43 @@ where
...
+                    if head_epoch < da_boundary {
+                        crit!(
+                            head_epoch = %head_epoch,
+                            da_boundary = %da_boundary,
+                            head_slot = %head_snapshot.beacon_block.slot(),
+                            "Head is outside the Data Availability window!"
+                        );
+                        crit!(
+                            "Your node's head is from before the DA window (~18 days). \
+                             Lighthouse cannot independently verify blob availability for historical blocks. \
```
eserilev:
```suggestion
                            "Your node's head is outside the DA window (~18 days). \
                             Lighthouse cannot independently verify blob availability for historical blocks. \
```

beacon_node/beacon_chain/src/builder.rs
```diff
@@ -847,6 +847,43 @@ where
...
+                             Options: \
+                             1. Use checkpoint sync: --checkpoint-sync-url <URL> \
+                             2. Use --enable-unsafe-sync to bypass this check"
+                        );
+                        return Err(format!(
+                            "Head epoch {} is before DA boundary epoch {}. \
+                             Use --enable-unsafe-sync to bypass.",
```
eserilev:
```suggestion
                            "Head epoch {} is before DA boundary epoch {}.
```

beacon_node/src/cli.rs
PoulavBhowmick03: sure! doing so

PR #8708, opened 2026-01-28
Disable `sqlite` by default

macladson:
## Issue Addressed

Compiling types with the `sqlite` feature more than doubles the total compile time. Most downstream users don't need `sqlite` compatibility for `Slot` and `Epoch` types.

## Proposed Changes

Disable the `sqlite` feature by default.

### DIFF
--- a/consensus/types/Cargo.toml
+++ b/consensus/types/Cargo.toml
@@ -8,7 +8,7 @@ authors = [
 edition = { workspace = true }
 
 [features]
-default = ["sqlite", "legacy-arith"]
+default = ["legacy-arith"]
 # Allow saturating arithmetic on slots and epochs. Enabled by default, but deprecated.
 legacy-arith = []
 sqlite = ["dep:rusqlite"]
--- a/validator_client/slashing_protection/Cargo.toml
+++ b/validator_client/slashing_protection/Cargo.toml
@@ -23,7 +23,7 @@ serde = { workspace = true }
 serde_json = { workspace = true }
 tempfile = { workspace = true }
 tracing = { workspace = true }
-types = { workspace = true }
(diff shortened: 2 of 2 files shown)


eserilev (approved):
nice, lgtm!

Cargo.toml
```diff
@@ -272,7 +272,7 @@ tracing_samplers = { path = "common/tracing_samplers" }
 tree_hash = "0.12.0"
 tree_hash_derive = "0.12.0"
 typenum = "1"
-types = { path = "consensus/types" }
+types = { path = "consensus/types", features = ["sqlite"] }
```
pawanjay176: Can we move the feature to the `slashing_protection` crate where this is used? 
If we enable this feature only for the `slashing_protection` crate, we will save compiling costs on beacon chain tests and other tests that don't require slashing protection right?

pawanjay176 (approved):
This is great! Just a nit
afaiu, these impls are only used in slashing_protection crate? Just verifying that we can't completely move this to that crate because we would have to wrap them up in newtypes?

Cargo.toml
michaelsproul: Yeah good shout, I was thinking it wouldn't make any difference because of Cargo's feature unification, but it probably isn't unifying the `slashing_protection` crate when compiling tests (would be worth checking)

macladson: Nice idea, I like this much better!

And yes, we need to implement traits from `rusqlite` on `Slot` and `Epoch` so for them to exist in `slashing_protection` we would have to wrap them in newtypes which I suspect isn't very feasible.

pawanjay176 (approved):
Nice!

PR #8715, opened 2026-01-28
Add critical instructions and hooks for Claude Code

dapplion:
## Summary

- Add CRITICAL section at top of CLAUDE.md to ensure Claude always runs formatting and linting after code changes
- Add `.claude/settings.json` with post-tool-call hook to remind about formatting and linting after Edit/Write operations

This helps ensure Claude Code consistently follows the project's code quality requirements.

## Test plan

- [x] Verify CLAUDE.md changes are visible
- [x] Verify hook configuration is valid JSON

🤖 Generated with [Claude Code](https://claude.com/claude-code)

(Lion here)

Claude recognizes that it ignored Claude.md most of the time, and recommends this edits to actually lint and format when doing changes.


eserilev (approved):

.claude/settings.json
```diff
@@ -0,0 +1,10 @@
+{
+  "hooks": {
+    "post-tool-call": [
+      {
+        "matcher": "Edit|Write",
+        "command": "echo '\n[Reminder] Run: cargo fmt --all && make lint-fix && cargo check'"
```
jimmygchen: I can't find `post-tool-call` on claude documentation, is this documented somewhere else?

https://code.claude.com/docs/en/hooks#common-json-fields

michaelsproul:
I thought the plan was to add a commit hook that checks `cargo fmt`, so this is enforced by Git and not vibes

.claude/settings.json
dapplion: Hmm should be post-tool-use?

dapplion:
> I thought the plan was to add a commit hook that checks cargo fmt, so this is enforced by Git and not vibes

Do we want to have git hooks that apply to everyone?

.claude/settings.json
```diff
@@ -0,0 +1,15 @@
...
+        "matcher": "Edit|Write",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "echo '\n[Reminder] Run: cargo fmt --all && make lint-fix && cargo check'"
+          }
+        ]
+      }
+    ]
+  }
```
dapplion, line 14: Updated this format until my local claude CLI accepted the format.

Cargo.toml
```diff
@@ -120,6 +120,7 @@ bytes = "1.11.1"
 # Turn off c-kzg's default features which include `blst/portable`. We can turn on blst's portable
 # feature ourselves when desired.
 c-kzg = { version = "2.1", default-features = false }
+cargo-husky = { version = "1", default-features = false, features = ["precommit-hook", "run-cargo-fmt"] }
```
jimmygchen: hmm I'm not sure if it's worth introducing a dependency for this. It's also unmaintained, but I see benefits over native git hooks - without devs having to manually set it up.

Now i feel like it might be better to just include it in the AI file you originally suggested.

With `cargo-husky` requiring at least a `cargo` task to trigger, i think there might be a chance that this never gets run, e.g. AI-generated PRs etc.

.claude/settings.json
```diff
@@ -0,0 +1,15 @@
+{
+  "hooks": {
+    "PostToolUse": [
+      {
+        "matcher": "Edit|Write",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "echo '\n[Reminder] Run: cargo fmt --all && make lint-fix && cargo check'"
```
jimmygchen: I don't think we need cargo check if we already run clippy

Cargo.toml
dapplion: I can add a make command to add it your local dir, and then we expect team members to add it

pawanjay176 (commented):
@dapplion I'm wondering why we can't have our own local versions of the settings.json. Why does this need to be committed for everyone?

jking-aus:
> @dapplion I'm wondering why we can't have our own local versions of the settings.json. Why does this need to be committed for everyone?

settings.json at project level is good for sharing hooks and tooling across the team. Your .claude/settings.local.json is not committed and overrides the project level settings

https://code.claude.com/docs/en/settings

.claude/settings.json
jimmygchen: I'm removing the `cargo check` command and merging this - please correct me if i wrong

jimmygchen (approved):
LGTM, thanks!

PR #8716, opened 2026-01-29
Gloas data availability checker

eserilev:
TODO
Add more details

image


beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3292,33 +3315,45 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+                        // We use metric here because logging this would be *very* noisy.
+                        metrics::inc_counter_vec(
+                            &metrics::KZG_DATA_COLUMN_RECONSTRUCTION_INCOMPLETE_TOTAL,
+                            &[reason],
+                        );
+                        Ok(None)
+                    }
+                }
             }
+            ReconstructionOutcome::Payload(_data_column_reconstruction_result) => todo!(),
```
eserilev: return an error and add a `TODO`

Ari4ka (approved):

Ari4ka (approved):

beacon_node/lighthouse_network/src/rpc/methods.rs
```diff
@@ -528,16 +529,21 @@ impl BlobsByRootRequest {
 pub struct DataColumnsByRootRequest<E: EthSpec> {
     /// The list of beacon block roots and column indices being requested.
     pub data_column_ids: RuntimeVariableList<DataColumnsByRootIdentifier<E>>,
+    pub fork_name: ForkName,
```
pawanjay176, line 566: Maybe add a comment here saying that this isn't part of the consensus object? 
Mainly to avoid serializing the fork_name when sending it over the wire.

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -1053,6 +1053,14 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
                         &metrics::BEACON_BLOB_DELAY_FULL_VERIFICATION,
                         processing_start_time.elapsed().as_millis() as i64,
                     );
+
+                    // If a block is in the da_checker, sync maybe awaiting for an event when block is finally
+                    // imported. A block can become imported both after processing a block or data column. If a
```
pawanjay176:
```suggestion
                    // imported. A block can become imported both after processing a block or data column. If 
```

pawanjay176: Any specific reason why we moved it from below to here other than this being cleaner?

beacon_node/beacon_chain/src/data_availability_router.rs
```diff
@@ -0,0 +1,379 @@
...
+    }
+}
+
+/// Trait for data availability operations on availability checkers.
+///
+/// Both `DataAvailabilityChecker` (v1) and `DataAvailabilityChecker` (v2) implement
+/// this trait. The associated types differ:
+/// - V1: Returns `Availability<E>` containing `AvailableExecutedBlock<E>`
+/// - V2: Returns `Availability<E>` containing `(Hash256, DataColumnSidecarList<E>)` (block root + columns)
+pub trait AvailabilityCache<T: BeaconChainTypes>: Send + Sync {
```
pawanjay176: Given that the da checker holds both v1 and v2 variants and we are explicitly dispatching to v1 and v2, the trait doesn't seem necessary to me.

Right now, it just seems to be documenting shared behaviour between both variants. Do you think its required?

eserilev: Yeah the trait was made to just be explicit about what functionality is shared between v1 and v2. Might be a lot of overhead just for documentation. I don't mind deleting it

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
eserilev: yeah just to clean it up a bit

Ari4ka (approved):

Ari4ka (approved):

jimmygchen:
Superseded by #9209.

PR #8717, opened 2026-01-29
Payload envelope db operations

eserilev:
## Issue Addressed

Adds support for payload envelopes in the db. This is the minimum we'll need to store and fetch payloads.

### DIFF
--- a/beacon_node/store/src/hot_cold_store.rs
+++ b/beacon_node/store/src/hot_cold_store.rs
@@ -93,6 +93,7 @@ struct BlockCache<E: EthSpec> {
     block_cache: LruCache<Hash256, SignedBeaconBlock<E>>,
     blob_cache: LruCache<Hash256, BlobSidecarList<E>>,
     data_column_cache: LruCache<Hash256, HashMap<ColumnIndex, Arc<DataColumnSidecar<E>>>>,
+    payload_envelope_cache: LruCache<Hash256, SignedExecutionPayloadEnvelope<E>>,
     data_column_custody_info_cache: Option<DataColumnCustodyInfo>,
 }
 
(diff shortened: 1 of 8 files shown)


Ari4ka (approved):

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -93,6 +93,7 @@ struct BlockCache<E: EthSpec> {
     block_cache: LruCache<Hash256, SignedBeaconBlock<E>>,
     blob_cache: LruCache<Hash256, BlobSidecarList<E>>,
     data_column_cache: LruCache<Hash256, HashMap<ColumnIndex, Arc<DataColumnSidecar<E>>>>,
+    payload_envelope_cache: LruCache<Hash256, SignedExecutionPayloadEnvelope<E>>,
```
michaelsproul: I think we can just remove this, and then merge.

Block cache is going in the bin eventually:

- https://github.com/sigp/lighthouse/pull/8404

eserilev: ive removed it here: 
7367189

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -745,6 +745,32 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
+        match self
+            .hot_db
+            .get_bytes(SignedExecutionPayloadEnvelope::<E>::db_column(), key)?
+        {
+            Some(bytes) => {
+                let envelope = SignedExecutionPayloadEnvelope::from_ssz_bytes(&bytes)?;
+                Ok(Some(envelope))
+            }
+            None => Ok(None),
+        }
```
michaelsproul, line 763: This func could probably just be `self.hot_db.get_item()` but seeing as this DB schema is likely temporary we can leave as is.

michaelsproul (approved):
LGTM for first devnet!

PR #8718, opened 2026-01-29
Emit `NewHead` SSE event earlier in block import

jimmygchen:
## Description

Moves the NewHead SSE event emission from `after_new_head()` to `import_block()`, the earliest point after we've computed new head. This will enable validator clients to attest as soon as a new head is computed (https://github.com/sigp/lighthouse/pull/7892), rather than having to wait until block and fork choice is persisted to disk.

Previously, the head event was emitted after several slow operations including database reads, state cache building, and `persist_fork_choice()` (on epoch transitions, which takes 500ms - 1s on mainnet). This delayed event delivery to subscribed clients.

The new location emits the event as soon as fork choice confirms the block is the new head, before any of these slower operations. This mirrors the existing `early_attester_cache` pattern which already operates on the same assumption. 

Note: 
* `Head` events are only emitted for recent, non-optimistic blocks (same conditions as `early_attester_cache`), i.e. events are not emitted during sync. This is because `fork_choice.get_head` is not computed during sync / optimistic imports. This shouldn't cause any issues, but I've mark this PR as a breaking change.

### DIFF
--- a/beacon_node/beacon_chain/src/beacon_chain.rs
+++ b/beacon_node/beacon_chain/src/beacon_chain.rs
@@ -3862,6 +3862,19 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
                     &self.spec,
                 )
                 .map_err(|e| BlockError::BeaconChainError(Box::new(e.into())))?;
+
+            // Register a server-sent-event for a new block.
+            if let Some(event_handler) = self
+                .event_handler
+                .as_ref()
+                .filter(|handler| handler.has_block_subscribers())
+            {
+                event_handler.register(EventKind::Block(SseBlock {
+                    slot: block.slot(),
+                    block: block_root,
+                    execution_optimistic: payload_verification_status.is_optimistic(),
+                }));
+            }
         }
 
         // If the block is recent enough and it was not optimistically imported, check to see if it
@@ -4071,7 +4084,6 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             block,
             block_root,
             block_time_imported,
-            payload_verification_status,
             current_slot,
         );
 
@@ -4359,7 +4371,6 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
         block: BeaconBlockRef<T::EthSpec>,
         block_root: Hash256,
         block_time_imported: Duration,
-        payload_verification_status: PayloadVerificationStatus,
         current_slot: Slot,
     ) {
         // Only present some metrics for blocks from the previous epoch or later.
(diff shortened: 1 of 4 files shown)


jimmygchen:
Assertoor test is failing because it tries to query for the block after a new block event, but the HTTP returns 404, since it hasn't been written to the database.

Options:
* emit block and head after block is writen to the DB. Downside is we cant use events to attest early
* emit block and head early, and serve http from cache. This is a **breaking** change, because we only add attestable blocks to the cache, so we would only emit recent blocks - currently we emit all imported blocks. 
* emit head first and leave block unchanged. head event will come before block event, which may look confusing

(2) might be best from users point of view, but this adds breaking change not directly related to this PR.

For now we'll go with (3) and leave block event unchanged, given the head event change is also a slightly breaking change.

beacon_node/beacon_chain/tests/events.rs
```diff
@@ -213,3 +201,45 @@ async fn data_column_sidecar_event_on_process_rpc_columns() {
...
+    let block_root = signed_block.canonical_root();
+    let state_root = signed_block.message().state_root();
+
+    harness
+        .process_block(target_slot, block_root, (signed_block, blobs))
+        .await
+        .unwrap();
+
+    // Verify the head event was emitted with correct data
+    let head_event = head_event_receiver.try_recv().unwrap();
```
michaelsproul, line 235: Don't see the value in `try_recv` here (over `recv`) but may as well leave it, as it's consistent with other tests.

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -3937,6 +3905,51 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
+
+                        // Register a server-sent-event for a new head.
+                        if let Some(event_handler) = self
+                            .event_handler
+                            .as_ref()
+                            .filter(|handler| handler.has_head_subscribers())
+                        {
+                            let head_slot = state.slot();
+                            let state_root = block.state_root();
+                            let old_head_slot = self.canonical_head.cached_head().head_slot();
```
michaelsproul: I worry we might be introducing a deadlock option here:

https://github.com/sigp/lighthouse/blob/010519d9b59b4452c6db95771265a6e5dff90069/beacon_node/beacon_chain/src/beacon_chain.rs#L3838-L3840

We are holding fork choice and locking the cached head. This would deadlock if there's any other location where we lock in the other order (`cached_head; fork_choice`).

I don't know if there's actually a place where we do that, but to mitigate this risk we could read the cached head earlier in this method prior to taking the fork choice lock at all?

michaelsproul (commented):
LGTM on the whole, just needs the fork choice lock tweak mentioned, I think.

beacon_node/beacon_chain/src/beacon_chain.rs
jimmygchen: Nice catch! 
I've made the suggested change.

michaelsproul (approved):

PR #8719, opened 2026-01-29
Gloas state cache

dapplion:
## Issue Addressed

Allow the in-memory state cache to store both post-block states (payload status = empty) and post-payload states (payload status = full)

## Proposed Changes

Add a new level to the state cache

```diff
- /// Map from block_root -> slot -> state_root.
+ /// Map from block_root -> slot -> payload_status -> state_root.
```

And force consumers to decide which payload_status they perfer when querying by block and slot

```rs
/// Filter for querying states by payload status.
pub enum PayloadStatusFilter {
    /// Only return the state where the execution payload was included.
    Full,
    /// Only return the state where the execution payload was not included.
    Empty,
    /// Return any state, preferring full to empty.
    Any,
}
```

The storage of HDiff buffers requires fixes, to be addressed in another PR


beacon_node/store/src/hot_cold_store.rs
```diff
@@ -3915,6 +3966,7 @@ pub struct HotStateSummary {
     pub slot: Slot,
     pub latest_block_root: Hash256,
     pub latest_block_slot: Slot,
+    pub latest_block_is_full: bool,
```
michaelsproul, line 3969: I am not fully convinced this is necessary/desirable. I came looking for this PR while working on:

- https://github.com/sigp/lighthouse/issues/8869

I was hoping to just grab the simple `state_cache` fix. I think the HDiff stuff could work without this field, because the block's state and payload's state have different `previous_state_root`s.

(partly I'm hesitant about this field because it necessitates a DB migration/etc in `unstable`, which is a bunch of extra complexity)

dapplion: How do you know if you need to process the last payload to replay the state?

michaelsproul: I'm working on this now, will link you

michaelsproul: Something like this

```rust
    /// Compute the `StatePayloadStatus` for a stored state based on its summary.
    ///
    /// In future this might become a field of the summary, but this would require a whole DB
    /// migration. For now we use an extra read from the DB to determine it.
    fn get_hot_state_summary_payload_status(
        &self,
        summary: &HotStateSummary,
    ) -> Result<StatePayloadStatus, Error> {
        // Treat pre-Gloas states as `Pending`.
        if !self
            .spec
            .fork_name_at_slot::<E>(summary.slot)
            .gloas_enabled()
        {
            return Ok(StatePayloadStatus::Pending);
        }

        // Treat genesis state as `Pending` (`BeaconBlock` state).
        let previous_state_root = summary.previous_state_root;
        if previous_state_root.is_zero() {
            return Ok(StatePayloadStatus::Pending);
        }

        // Load the hot state summary for the previous state. If it has the same slot as this
        // summary then we know this summary is for a `Full` block (payload state).
        // NOTE: We treat any and all skipped-slot states as `Pending` by this definition, which is
        // perhaps a bit strange (they could have a payload most-recently applied).
        let previous_state_summary = self
            .load_hot_state_summary(&previous_state_root)?
            .ok_or(Error::MissingHotStateSummary(previous_state_root))?;

        if previous_state_summary.slot == summary.slot {
            Ok(StatePayloadStatus::Full)
        } else {
            Ok(StatePayloadStatus::Pending)
        }
    }
```

It feels pretty hacky though, I think long-term we would want a migration. I'm just hesitant to do that right now when we can just do 1 extra read instead.

michaelsproul: Need the same thing for cold states too. I'm going to look at wiring these up now.

michaelsproul:
Closing in favour of changes already merged in:

- https://github.com/sigp/lighthouse/pull/8886

Thanks for scoping this out though!

PR #4801, opened 2023-10-03
Prevent Overflow LRU Cache from Exploding

ethDreamer:
## Issue Addressed

* #4757

In a nutshell, the problem with the [`OverflowLRUCache`](https://github.com/sigp/lighthouse/blob/c7ddf1f0b1cfc19a03fcadc7d5ab13d6d595c69d/beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs#L354) before this PR is that it [caches](https://github.com/sigp/lighthouse/blob/c7ddf1f0b1cfc19a03fcadc7d5ab13d6d595c69d/beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs#L56) an [`AvailabilityPendingExecutedBlock`](https://github.com/sigp/lighthouse/blob/c7ddf1f0b1cfc19a03fcadc7d5ab13d6d595c69d/beacon_node/beacon_chain/src/block_verification_types.rs#L255) for each entry, which [contains the entire `BeaconState`](https://github.com/sigp/lighthouse/blob/c7ddf1f0b1cfc19a03fcadc7d5ab13d6d595c69d/beacon_node/beacon_chain/src/block_verification_types.rs#L294). If we end up with multiple forks and many unavailable blocks, this cache can have many entries, causing it to consume large amounts of memory.

I've addressed this by creating a small `StateLRUCache` which will accept the `AvailabilityPendingExecutedBlock`, move the `BeaconState` to a very limited [LRU cache](https://github.com/ethDreamer/lighthouse/blob/e8ef884c544fe51a3d8ddad94df7e4ea43addaf6/beacon_node/beacon_chain/src/data_availability_checker/state_lru_cache.rs#L55) and return a [`DietAvailablityPendingExecutedBlock`](https://github.com/ethDreamer/lighthouse/blob/e8ef884c544fe51a3d8ddad94df7e4ea43addaf6/beacon_node/beacon_chain/src/data_availability_checker/state_lru_cache.rs#L19) which contains all the same data except the `BeaconState` has been replaced with a root.

If many unavailable blocks are stored at the same time in the `DataAvailabiltyCache`, the excess states will be dropped. If those states are needed later they are recovered by loading the parent `BeaconState` from disk and replaying the block.

PR #8724, opened 2026-01-30
Remove state lru cache

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

In https://github.com/sigp/lighthouse/pull/4801 , we added a state lru cache to avoid having too many states in memory which was a concern with 200mb+ states pre tree-states. 
With https://github.com/sigp/lighthouse/pull/5891 , we made the overflow cache a simpler in memory lru cache that can only hold 32 pending states at the most and doesn't flush anything to disk. As noted in #5891, we can always fetch older blocks which never became available over rpc if they become available later.

Since we merged tree states, I don't think the state lru cache is relevant anymore. Instead of having the `DietAvailabilityPendingExecutedBlock` that stores only the state root, we can just store the full state in the `AvailabilityPendingExecutedBlock`. 
Given entries in the cache can span max 1 epoch (cache size is 32), the underlying `BeaconState` objects in the cache share most of their memory. The state_lru_cache is one level of indirection that doesn't give us any benefit. 
Please check me on this cc @dapplion


beacon_node/beacon_chain/src/data_availability_checker/overflow_lru_cache.rs
```diff
@@ -307,15 +286,13 @@ impl<E: EthSpec> PendingComponents<E> {
...
         };
 
-        let AvailabilityPendingExecutedBlock {
-            block,
-            import_data,
-            payload_verification_outcome,
-        } = recover(*block.clone(), &self.span)?;
+        let block_arc = block.block.clone();
+        let import_data = block.import_data.clone();
+        let payload_verification_outcome = block.payload_verification_outcome.clone();
```
eserilev: nit: I think you can keep the same pattern here (just without the recover)

``` rust
let AvailabilityPendingExecutedBlock {
   block,
   import_data,
   payload_verification_outcome
} = block.as_ref().clone()
```

eserilev (approved):
LGTM!

Ari4ka (approved):

jimmygchen (approved):
This is great, nice simplification 👍

PR #8726, opened 2026-02-02
fix: set slot_number to latest_block_header.slot in Gloas upgrade

qu0b:
## Summary

- During the Fulu→Gloas state upgrade, `upgrade_to_gloas()` on `ExecutionPayloadHeaderFulu` hardcoded `slot_number: 0` in the new Gloas header
- This causes beacon state root divergence between Lighthouse and Lodestar at the Gloas fork boundary (slot 8 with minimal preset)
- Fix: pass `pre.latest_block_header.slot` to `upgrade_to_gloas()` so the upgraded header carries forward the correct slot number

## Test plan

- [x] Built with `FEATURES="spec-minimal"` — compiles cleanly
- [x] Tested locally with Kurtosis (LH supernode + Geth, 2x LS + NM): all 3 nodes in consensus through 7+ epochs, epoch 4 finalized, same head root throughout
- [x] Confirmed this resolves the LS + NM beacon state root divergence at the Gloas fork boundary (previously failed every time)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/consensus/state_processing/src/upgrade/gloas.rs
+++ b/consensus/state_processing/src/upgrade/gloas.rs
@@ -64,7 +64,9 @@ pub fn upgrade_state_to_gloas<E: EthSpec>(
         current_sync_committee: pre.current_sync_committee.clone(),
         next_sync_committee: pre.next_sync_committee.clone(),
         // Execution - upgrade header from Fulu to Gloas
-        latest_execution_payload_header: pre.latest_execution_payload_header.upgrade_to_gloas(),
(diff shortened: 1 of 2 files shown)


michaelsproul:
We haven't finished implementing Gloas at all, so nothing is expected to work rn.

michaelsproul:
Oh, this is for BAL. Sure, could probably merge this.

qu0b:
please merge

pawanjay176 (approved):

PR #8731, opened 2026-02-02
Never use MIN_EPOCHS_FOR_DATA_COLUMN_SIDECARS_REQUESTS for networks that started with Fulu enabled

dknopik:
## Issue Addressed

Lighthouse uses `MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS` for blob **and column retention** instead of `MIN_EPOCHS_FOR_DATA_COLUMN_SIDECARS_REQUESTS` if Fulu activated less than `MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS` epochs ago - also if Fulu activated at genesis. This causes unexpected behaviour, as there are no blob sidecars to be stored or requested in such networks.

## Proposed Changes

~~Add a special case to avoid that logic in post-Fulu genesis networks (`fulu_fork_epoch == 0`)~~

If the blob retention period would start in the fulu fork epoch, use the `min_epochs_for_data_column_sidecars_requests`, as there are no blobs to retain in Fulu.


consensus/types/src/core/chain_spec.rs
```diff
@@ -821,14 +821,19 @@ impl ChainSpec {
...
-            ),
+            Some(fulu_fork_epoch)
+                if blob_retention_epoch > fulu_fork_epoch || fulu_fork_epoch == Epoch::new(0) =>
+            {
+                Some(
+                    current_epoch.saturating_sub(self.min_epochs_for_data_column_sidecars_requests),
+                )
+            }
             _ => Some(std::cmp::max(fork_epoch, blob_retention_epoch)),
         }
```
dapplion: Is it clearer to use if else?

```rust
    pub fn min_epoch_data_availability_boundary(&self, current_epoch: Epoch) -> Option<Epoch> {
        let blob_retention_epoch =
            current_epoch.saturating_sub(self.min_epochs_for_blob_sidecars_requests);
        if let Some(fulu_fork_epoch) = self.fulu_fork_epoch
            && blob_retention_epoch >= fulu_fork_epoch
        {
            Some(current_epoch.saturating_sub(self.min_epochs_for_data_column_sidecars_requests))
        } else if let Some(deneb_fork_epoch) = self.deneb_fork_epoch {
            Some(std::cmp::max(deneb_fork_epoch, blob_retention_epoch))
        } else {
            None
        }
    }
```

dknopik: Yeah good call. I also thought about `blob_retention_epoch >= fulu_fork_epoch` instead of the special case for `fulu_fork_epoch == Epoch::new(0)`, but was not entirely sure it's fine.

consensus/types/src/core/chain_spec.rs
michaelsproul, line 3386: Can we also add a new test to cover the Fulu-from-genesis case?

dknopik: done

jimmygchen (approved):
Looks good!

dknopik:
@mergifyio queue

michaelsproul:
@mergify requeue

dknopik:
@mergifyio queue

PR #8739, opened 2026-02-03
Remove `kzg_commitments` from `DataColumnSidecarGloas`

jimmygchen:
## Description

Aligns `DataColumnSidecarGloas` with the Gloas spec (EIP-7732) by removing the `kzg_commitments` field. In Gloas, commitments come from `block.body.signed_execution_payload_bid.message.blob_kzg_commitments` instead of being embedded in the sidecar.


michaelsproul (approved):
LGTM! Nice catch

michaelsproul:
@mergify queue

PR #8754, opened 2026-02-04
Gloas local block building MVP

eserilev:
## Issue Addressed

The flow for local block building is
1. Create execution payload and bid
2. Construct beacon block
3. Sign beacon block and publish
4. Sign execution payload and publish

This PR adds the beacon block v4 flow , GET payload envelope and POST payload envelope (local block building only). The spec for these endpoints can be found here:  https://github.com/ethereum/beacon-APIs/pull/552  and is subject to change.

We needed a way to store the unsigned execution payload envelope associated to the execution payload bid that was included in the block. I introduced a new cache that stores these unsigned execution payload envelopes. the GET payload envelope queries this cache directly so that a proposer, after publishing a block, can fetch the payload envelope + sign and publish it.

I kept payload signing and publishing within the validators block service to keep things simple for now. The idea was to build out a block production MVP for devnet 0, try not to affect any non gloas code paths and build things out in such a way that it will be easy to deprecate pre-gloas code paths later on (for example block production v2 and v3).

We will eventually need to track which beacon node was queried for the block so that we can later query it for the payload. But thats not needed for the devnet.


validator_client/validator_services/src/block_service.rs
```diff
@@ -538,14 +611,103 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> BlockService<S, T> {
...
+    }
+
+    /// Fetch, sign, and publish the execution payload envelope for Gloas.
+    /// This should be called after the block has been published.
+    ///
+    /// TODO(gloas): For multi-BN setups, we need to track which beacon node produced the block
+    /// and fetch/publish the envelope from that same node. The envelope is cached per-BN,
+    /// so fetching from a different BN than the one that built the block will fail.
+    /// See: https://github.com/sigp/lighthouse/pull/8313
+    #[instrument(skip_all, fields(%slot, ?validator_pubkey))]
```
jimmygchen: These fields don't seem necessary to record, as they're already recorded on the caller site

validator_client/validator_services/src/block_service.rs
```diff
@@ -538,14 +611,103 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> BlockService<S, T> {
...
+        slot: Slot,
+        validator_pubkey: &PublicKeyBytes,
+    ) -> Result<(), BlockError> {
+        info!(slot = slot.as_u64(), "Fetching execution payload envelope");
+
+        // Fetch the envelope from the beacon node (builder_index=0 for local building)
+        let envelope = proposer_fallback
+            .request_proposers_last(|beacon_node| async move {
+                beacon_node
+                    .get_validator_execution_payload_envelope::<S::E>(slot, 0)
```
jimmygchen: Please declare a constant for local building

validator_client/validator_services/src/block_service.rs
```diff
@@ -538,14 +611,103 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> BlockService<S, T> {
...
+        // Publish the signed envelope
+        proposer_fallback
+            .request_proposers_first(|beacon_node| {
+                let signed_envelope = signed_envelope.clone();
+                async move {
+                    beacon_node
+                        .post_beacon_execution_payload_envelope(&signed_envelope)
+                        .await
+                        .map_err(|e| {
+                            BlockError::Recoverable(format!(
```
jimmygchen, line 710: This should probably be irrecoverable once we implement sticky block production - once it fails on the specified BN, retrying will not work

validator_client/signing_method/src/web3signer.rs
```diff
@@ -19,6 +19,7 @@ pub enum MessageType {
     SyncCommitteeSelectionProof,
     SyncCommitteeContributionAndProof,
     ValidatorRegistration,
+    ExecutionPayloadEnvelope,
```
jimmygchen, line 23: This is most likely what they're going to use - but i think we should add a TODO here since i don't see this in their code / docs yet.

validator_client/lighthouse_validator_store/src/lib.rs
```diff
@@ -1242,4 +1242,33 @@ impl<T: SlotClock + 'static, E: EthSpec> ValidatorStore for LighthouseValidatorS
...
     }
+
+    /// Sign an `ExecutionPayloadEnvelope` for Gloas (local building).
+    /// The proposer acts as the builder and signs with the BeaconBuilder domain.
+    async fn sign_execution_payload_envelope(
+        &self,
+        validator_pubkey: PublicKeyBytes,
+        envelope: ExecutionPayloadEnvelope<E>,
+    ) -> Result<SignedExecutionPayloadEnvelope<E>, Error> {
+        let domain_hash = self.spec.get_builder_domain();
```
jimmygchen: Is this the right domain to use?

jimmygchen: https://github.com/ethereum/consensus-specs/blob/2938e1ad74cea54f1a24508a85704d5bd87837ad/specs/gloas/beacon-chain.md?plain=1#L128

beacon_node/http_api/src/validator/execution_payload_bid.rs
```diff
@@ -0,0 +1,52 @@
...
+        .then(
+            |slot: Slot,
+             _accept_header: Option<Accept>,
+             not_synced_filter: Result<(), Rejection>,
+             task_spawner: TaskSpawner<T::EthSpec>,
+             _chain: Arc<BeaconChain<T>>| {
+                task_spawner.spawn_async_with_rejection(Priority::P0, async move {
+                    debug!(
+                        ?slot,
+                        "Execution paylaod bid production request from HTTP API"
```
jimmygchen: typo:

```suggestion
                        "Execution payload bid production request from HTTP API"
```

beacon_node/beacon_chain/src/block_production/gloas.rs
```diff
@@ -0,0 +1,749 @@
...
+            .as_gloas()
+            .map_err(|_| BlockProductionError::GloasNotImplemented)?
+            .to_owned();
+
+        let bid = ExecutionPayloadBid::<T::EthSpec> {
+            parent_block_hash: state.latest_block_hash()?.to_owned(),
+            parent_block_root: state.get_latest_block_root(state_root),
+            block_hash: execution_payload.block_hash(),
+            prev_randao: execution_payload.prev_randao(),
+            fee_recipient: Address::ZERO,
```
jimmygchen, line 712: `SignedProposerPreferences` is not implemented but we should probably add a TODO here

beacon_node/beacon_chain/src/block_production/gloas.rs
```diff
@@ -0,0 +1,749 @@
...
+            gas_limit: execution_payload.gas_limit(),
```
jimmygchen: same as above

beacon_node/beacon_chain/src/block_production/gloas.rs
```diff
@@ -0,0 +1,749 @@
...
+                        )
+                    } else {
+                        return Err(BlockProductionError::MissingKzgCommitment(
+                            "No KZG commitments from the payload".to_owned(),
+                        ));
+                    }
+                }
+                // TODO(gloas) we should never receive a blinded response.
+                // Should return some type of `Unexpected` error variant as this should never happen
+                // in the V4 block production flow
```
jimmygchen: Yeah agree, I think we could just do that?

beacon_node/beacon_chain/src/block_production/gloas.rs
```diff
@@ -0,0 +1,749 @@
...
+            .map_err(BlockProductionError::TokioJoin)??;
+
+        // Part 2/3 (async)
+        //
+        // Produce the execution payload bid.
+        // TODO(gloas) this is strictly for building local bids
+        // We'll need to build out trustless/trusted bid paths.
+        let (execution_payload_bid, state, payload_data) = self
+            .clone()
+            .produce_execution_payload_bid(state, state_root_opt, produce_at_slot, 0, u64::MAX)
```
jimmygchen: Be good to use named constants for `bid_value` and `BuilderIndex`, even if this is incomplete

beacon_node/beacon_chain/src/pending_payload_envelopes.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    /// Check if an envelope exists for the given slot.
+    pub fn contains(&self, slot: Slot) -> bool {
+        self.envelopes.contains_key(&slot)
+    }
+
+    /// Prune envelopes older than `current_slot - max_slot_age`.
+    ///
+    /// This removes stale envelopes from blocks that were never published.
+    pub fn prune(&mut self, current_slot: Slot) {
```
jimmygchen, line 65: Should add a TODO to implement pruning - right now this cache is never pruned

validator_client/lighthouse_validator_store/src/lib.rs
jimmygchen: ah it should just be `spec.domain_beacon_proposer`

consensus/types/src/execution/execution_payload_envelope.rs
```diff
@@ -28,6 +28,49 @@ impl<E: EthSpec> SignedRoot for ExecutionPayloadEnvelope<E> {}
...
     use super::*;
     use crate::MainnetEthSpec;
+    use crate::test_utils::TestRandom;
+    use rand::SeedableRng;
+    use rand_xorshift::XorShiftRng;
 
     ssz_and_tree_hash_tests!(ExecutionPayloadEnvelope<MainnetEthSpec>);
+
+    #[test]
+    fn signing_root_is_deterministic() {
```
jimmygchen: not sure if these tests are necessary?

beacon_node/beacon_chain/src/block_production/gloas.rs
```diff
@@ -0,0 +1,749 @@
...
+            parent_block_hash: state.latest_block_hash()?.to_owned(),
+            parent_block_root: state.get_latest_block_root(state_root),
+            block_hash: execution_payload.block_hash(),
+            prev_randao: execution_payload.prev_randao(),
+            fee_recipient: Address::ZERO,
+            gas_limit: execution_payload.gas_limit(),
+            builder_index,
+            slot: produce_at_slot,
+            value: bid_value,
+            execution_payment: 0,
```
jimmygchen: Could you add a comment explaining why this is set to 0?

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -4601,7 +4605,7 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
 
     /// Load a beacon state from the database for block production. This is a long-running process
     /// that should not be performed in an `async` context.
-    fn load_state_for_block_production(
+    pub fn load_state_for_block_production(
```
jimmygchen: we could potentially just move this to the `block_production` mod and make it private, or make it `pub(crate)` instead of exposing this as `pub`

beacon_node/beacon_chain/src/block_production/gloas.rs
```diff
@@ -0,0 +1,749 @@
...
+            slot,
+            randao_reveal,
+            graffiti_settings,
+            verification,
+        )
+        .await
+    }
+
+    // TODO(gloas) need to implement builder boost factor logic
+    pub async fn produce_block_on_state_gloas(
```
jimmygchen, line 121: I think it's worth instrumenting this

beacon_node/beacon_chain/src/block_production/gloas.rs
```diff
@@ -0,0 +1,749 @@
...
+                return Err(BlockProductionError::InvalidBlockVariant(
+                    "Cannot construct a block pre-Gloas".to_owned(),
+                ));
+            }
+            BeaconState::Electra(_) => {
+                return Err(BlockProductionError::InvalidBlockVariant(
+                    "Cannot construct a block pre-Gloas".to_owned(),
+                ));
+            }
+            BeaconState::Fulu(_) => {
```
dapplion: You can collapse all of these into _ at the end

beacon_node/execution_layer/src/lib.rs
```diff
@@ -334,6 +334,14 @@ impl<E: EthSpec, Payload: AbstractExecPayload<E>> BlockProposalContents<E, Paylo
             Self::PayloadAndBlobs { block_value, .. } => block_value,
         }
     }
+    pub fn blob_kzg_commitments(&self) -> Option<&KzgCommitments<E>> {
```
jimmygchen: Is this function used? I can't find any usage

validator_client/validator_services/src/block_service.rs
eserilev: I think we should be able to `POST` the signed envelope to any BN, we just need to `GET` the unsigned envelope from the BN that produced the block. I thought the Irrecoverable/Recoverable error stuff was in the context of slashings? I don't think there's any slashing risk to retrying the payload envelope logic here

beacon_node/beacon_chain/src/beacon_chain.rs
eserilev: [fea43fb](https://github.com/sigp/lighthouse/pull/8754/commits/fea43fb0c860ed2c32fff75ad6bd6f000839f1c2)

beacon_node/beacon_chain/src/block_production/gloas.rs
eserilev: instrumentation added

beacon_node/execution_layer/src/lib.rs
eserilev: deleted

beacon_node/beacon_chain/src/block_production/gloas.rs
eserilev: I added a constant that should make it clear, plus added a TODO

consensus/types/src/execution/execution_payload_envelope.rs
eserilev: yeah my bad, didnt do a self review of this yet and some AI slop snuck in

beacon_node/beacon_chain/src/pending_payload_envelopes.rs
eserilev: added a TODO

beacon_node/beacon_chain/src/block_production/gloas.rs
eserilev: using constants here now

beacon_node/beacon_chain/src/block_production/gloas.rs
eserilev: added `Unexpected` error variant

beacon_node/beacon_chain/src/block_production/gloas.rs
eserilev: we have a TODO regarding `SignedProposerPreferences` above this function definition

validator_client/lighthouse_validator_store/src/lib.rs
eserilev: fixed, thanks!

validator_client/signing_method/src/web3signer.rs
eserilev: added a TODO

validator_client/validator_services/src/block_service.rs
eserilev: added constant

validator_client/validator_services/src/block_service.rs
eserilev: removed fields

validator_client/validator_services/src/block_service.rs
eserilev: [aa07952](https://github.com/sigp/lighthouse/pull/8754/commits/aa0795238ee04c7be4cb8e5c54de4498e52b2e7a) updated a relevant TODO here

consensus/types/src/execution/execution_payload_envelope.rs
eserilev: this has been deleted

beacon_node/beacon_chain/src/block_production/gloas.rs
```diff
@@ -0,0 +1,779 @@
...
+            )
+            .map_err(|_| BlockProductionError::GloasNotImplemented)?;
+
+            signed_envelope.message.state_root = state.update_tree_hash_cache()?;
+
+            // Cache the envelope for later retrieval by the validator for signing and publishing.
+            let envelope_slot = payload_data.slot;
+            self.pending_payload_envelopes
+                .write()
+                .insert(envelope_slot, signed_envelope.message);
```
pawanjay176, line 601: Curious why we decided to cache by root instead of slot. Is it because the beacon api used slot?

beacon_node/beacon_chain/src/errors.rs
```diff
@@ -310,6 +310,7 @@ pub enum BlockProductionError {
     MissingSyncAggregate,
     MissingExecutionPayload,
     MissingKzgCommitment(String),
+    MissingStateRoot,
```
pawanjay176: Unused

beacon_node/beacon_chain/src/errors.rs
```diff
@@ -320,6 +321,7 @@ pub enum BlockProductionError {
     SszTypesError(ssz_types::Error),
     // TODO(gloas): Remove this once Gloas is implemented
     GloasNotImplemented,
+    Unexpected(String),
```
pawanjay176: Maybe we can accept a string with `GloasNotImplemented` variant and move the `Unexpected` use to `GloasNotImplemented` to ensure we remember to handle it properly later

beacon_node/http_api/src/beacon/execution_payload_envelope.rs
```diff
@@ -0,0 +1,129 @@
...
+) -> Result<Response, Rejection> {
+    let slot = envelope.message.slot;
+    let beacon_block_root = envelope.message.beacon_block_root;
+
+    // Basic validation: check that the slot is reasonable
+    let current_slot = chain.slot().map_err(|_| {
+        warp_utils::reject::custom_server_error("Unable to get current slot".into())
+    })?;
+
+    // Don't accept envelopes too far in the future
```
pawanjay176: Is this part of the specification somewhere? Seems like a weird condition to add. Don't think we have anything similar for block publishing?

pawanjay176: I would assume we eventually have something like `BroadcastValidation::Gossip` validation logic like we have for blocks + blobs once the api gets more refined.

beacon_node/beacon_chain/src/execution_payload.rs
```diff
@@ -367,9 +367,18 @@ pub fn get_execution_payload<T: BeaconChainTypes>(
     let timestamp =
         compute_timestamp_at_slot(state, state.slot(), spec).map_err(BeaconStateError::from)?;
     let random = *state.get_randao_mix(current_epoch)?;
-    let latest_execution_payload_header = state.latest_execution_payload_header()?;
-    let latest_execution_payload_header_block_hash = latest_execution_payload_header.block_hash();
-    let latest_execution_payload_header_gas_limit = latest_execution_payload_header.gas_limit();
+    // In GLOAS (ePBS), the execution payload header is replaced by
+    // `latest_block_hash` and `latest_execution_payload_bid`.
+    let (latest_execution_payload_header_block_hash, latest_execution_payload_header_gas_limit) =
+        if state.fork_name_unchecked() == ForkName::Gloas {
```
pawanjay176: Nice catch. small nit: maybe use `gloas_enabled` instead

beacon_node/http_api/tests/tests.rs
```diff
@@ -3736,6 +3737,241 @@ impl ApiTester {
...
+
+            self.client
+                .post_beacon_blocks_v2(&signed_block_request, None)
+                .await
+                .unwrap();
+
+            assert_eq!(self.chain.head_beacon_block(), Arc::new(signed_block));
+
+            // Sign and publish the execution payload envelope
+            let domain = self.chain.spec.get_builder_domain();
```
pawanjay176: This seems is `ApplicationDomain::Builder` as opposed to `Domain::BeaconBuilder`

beacon_node/http_api/src/produce_block.rs
```diff
@@ -87,6 +130,39 @@ pub async fn produce_block_v3<T: BeaconChainTypes>(
...
+        Some(api_types::Accept::Ssz) => Response::builder()
+            .status(200)
+            .body(block.as_ssz_bytes().into())
+            .map(|res: Response<Body>| add_ssz_content_type_header(res))
+            .map(|res: Response<Body>| add_consensus_version_header(res, fork_name))
+            .map(|res| add_consensus_block_value_header(res, consensus_block_value_wei))
+            .map_err(|e| -> warp::Rejection {
+                warp_utils::reject::custom_server_error(format!("failed to create response: {}", e))
+            }),
+        _ => Ok(warp::reply::json(&beacon_response(
```
pawanjay176: shouldn't this use non-default `ProduceBlockV4Metadata`?

pawanjay176 (commented):
Looking great. just a few nits.
I like how the gloas logic is nicely separate

pawanjay176:
2 more issues i found while testing:
1. Calling getPayload in gloas panics here https://github.com/eserilev/lighthouse/blob/d1401452193bcb3aa8b4b9ac74df88f0d7aa8a91/consensus/types/src/execution/dumb_macros.rs#L32 when we try to convert the getPayload response to a full payload. Not sure if we can easily convert a gloas execution payload response to a full payload here
2. publishing a gloas signed beacon block as ssz would fail here https://github.com/eserilev/lighthouse/blob/d1401452193bcb3aa8b4b9ac74df88f0d7aa8a91/common/eth2/src/types.rs#L1987-L2006 . We can just special case gloas there to just decode as `SignedBeaconBlock`

eserilev:
thanks @pawanjay176 

regarding (1) I've separated the payload construction path now as well. We will be able to deprecate the old path post-Gloas. The new gloas path is also much simpler because we dont care about full/blinded and local vs builder payloads

[91e42ce](https://github.com/sigp/lighthouse/pull/8754/commits/91e42ceb9ec27c61dd087b88a5c388f968dfea26)

and for (2) 022ee4e

validator_client/validator_services/src/block_service.rs
```diff
@@ -538,14 +612,108 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> BlockService<S, T> {
 
         self_ref
             .sign_and_publish_block(
-                proposer_fallback,
+                proposer_fallback.clone(),
```
jimmygchen: Doesn't seem like a clone is necessary here, could just change the function to accept a `&ProposerFallback`

validator_client/validator_services/src/block_service.rs
```diff
@@ -538,14 +612,108 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> BlockService<S, T> {
...
+        &self,
+        proposer_fallback: ProposerFallback<T>,
+        slot: Slot,
+        validator_pubkey: &PublicKeyBytes,
+    ) -> Result<(), BlockError> {
+        info!(slot = slot.as_u64(), "Fetching execution payload envelope");
+
+        // Fetch the envelope from the beacon node. Use builder_index=BUILDER_INDEX_SELF_BUILD for local building.
+        let envelope = proposer_fallback
+            .request_proposers_last(|beacon_node| async move {
```
jimmygchen: should we just use `.request()` until we support multi BN setup?
This currently fetches the envelope using `request_proposers_last` and publish with `request_proposers_first`, so this will always fail when the `--proposer-only` flag is used

validator_client/lighthouse_validator_store/src/lib.rs
```diff
@@ -1242,4 +1242,37 @@ impl<T: SlotClock + 'static, E: EthSpec> ValidatorStore for LighthouseValidatorS
...
+
+        // Execution payload envelope signing is not slashable, bypass doppelganger protection.
+        let signing_method = self.doppelganger_bypassed_signing_method(validator_pubkey)?;
+
+        let signature = signing_method
+            .get_signature_from_root::<E, FullPayload<E>>(
+                SignableMessage::ExecutionPayloadEnvelope(&envelope),
+                signing_root,
+                &self.task_executor,
+                None,
```
jimmygchen: we're likely going to need `ForkInfo` here, given it's required in most web3signer message types

beacon_node/beacon_chain/src/block_production/gloas.rs
```diff
@@ -0,0 +1,902 @@
...
+            attestations,
+            deposits,
+            voluntary_exits,
+            sync_aggregate,
+            payload_attestations,
+            bls_to_execution_changes,
+        } = partial_beacon_block;
+
+        let beacon_block = match &state {
+            BeaconState::Base(_) => {
```
jimmygchen: nit: could collapse these prior fork arms into one

beacon_node/http_api/src/lib.rs
```diff
@@ -2466,6 +2486,14 @@ pub fn serve<T: BeaconChainTypes>(
         task_spawner_filter.clone(),
     );
 
+    // GET validator/execution_payload_envelope/{slot}/{builder_index}
+    let get_validator_execution_payload_envelope = get_validator_execution_payload_envelope(
+        eth_v1.clone().clone(),
```
jimmygchen: Redundant double clones. Looks like i started this 😅

beacon_node/beacon_chain/src/block_production/gloas.rs
```diff
@@ -0,0 +1,902 @@
...
+                signature: Signature::empty(),
+            };
+
+            // TODO(gloas) add better error variant
+            process_execution_payload_envelope(
+                &mut state,
+                None,
+                &signed_envelope,
+                VerifySignatures::False,
+                VerifyStateRoot::False,
```
jimmygchen, line 584: I see that state root is calculated below, but i think a brief inline comment explaining why both are false would help future readers.

consensus/types/src/core/consts.rs
```diff
@@ -28,6 +28,8 @@ pub mod deneb {
 pub mod gloas {
     pub const BUILDER_INDEX_SELF_BUILD: u64 = u64::MAX;
     pub const BUILDER_INDEX_FLAG: u64 = 1 << 40;
+    pub const BID_VALUE_SELF_BUILD: u64 = 0;
+    pub const EXECUTION_PAYMENT_TRUSTLESS_BUILD: u64 = 0;
```
jimmygchen: I think this file was originally intended for spec constants
https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/beacon-chain.md#constants

perhaps we could keep those custom constants where it's being used (i.e. `block_production`)?

beacon_node/beacon_chain/src/block_production/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+use types::{BeaconState, Hash256, Slot};
+
+use crate::{
+    BeaconChain, BeaconChainTypes, BlockProductionError, StateSkipConfig,
+    fork_choice_signal::ForkChoiceWaitResult, metrics,
+};
+
+mod gloas;
+
+impl<T: BeaconChainTypes> BeaconChain<T> {
```
jimmygchen, line 15: Note: all 3 functions moved here without modification 👍

common/eth2/src/lib.rs
```diff
@@ -2408,6 +2409,252 @@ impl BeaconNodeHttpClient {
...
+            path.query_pairs_mut()
+                .append_pair("builder_boost_factor", &builder_booster_factor.to_string());
+        }
+
+        if let Some(GraffitiPolicy::AppendClientVersions) = graffiti_policy {
+            path.query_pairs_mut()
+                .append_pair("graffiti_policy", "AppendClientVersions");
+        }
+
+        Ok(path)
```
jimmygchen, line 2451: This PR currently implements only the stateful operation path. Just a note that `include_payload` header hasn't been implemented yet and the default at the beacon-API spec level is `true` (stateless operation) - so just a heads up if we run into issues with pairing with other clients:

https://github.com/ethereum/beacon-APIs/blob/25de51f8e0b13f7873b22c6dabacd44a9a816316/apis/validator/block.v4.yaml#L64

common/eth2/src/lib.rs
```diff
@@ -2408,6 +2409,252 @@ impl BeaconNodeHttpClient {
...
+            .get_bytes_opt_accept_header(path, Accept::Ssz, self.timeouts.get_validator_block)
+            .await?;
+
+        let response_bytes = opt_response.ok_or(Error::StatusCode(StatusCode::NOT_FOUND))?;
+
+        ExecutionPayloadEnvelope::from_ssz_bytes(&response_bytes).map_err(Error::InvalidSsz)
+    }
+
+    /// `POST v1/beacon/execution_payload_envelope`
+    pub async fn post_beacon_execution_payload_envelope<E: EthSpec>(
```
jimmygchen, line 2634: I think we also need an SSZ variant of this

validator_client/validator_services/src/block_service.rs
```diff
@@ -538,14 +612,108 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> BlockService<S, T> {
...
+                beacon_node
+                    .get_validator_execution_payload_envelope::<S::E>(
```
jimmygchen: should we default to SSZ for both:
* `get_validator_execution_payload_envelope` and
* `post_beacon_execution_payload_envelope`

beacon_node/http_api/src/produce_block.rs
eserilev: [b4a8e9a](https://github.com/sigp/lighthouse/pull/8754/commits/b4a8e9ab07aa9904faaa865c3a662566492f81a0)

common/eth2/src/types.rs
```diff
@@ -1751,6 +1751,20 @@ pub struct ProduceBlockV3Metadata {
     pub consensus_block_value: Uint256,
 }
 
+/// Metadata about a `ProduceBlockV3Response` which is returned in the body & headers.
```
jimmygchen: outdated comment

beacon_node/http_api/tests/tests.rs
eserilev: [12bbc5b](https://github.com/sigp/lighthouse/pull/8754/commits/12bbc5bcfe542a98078417549c7bcc5459de6dc2)

beacon_node/beacon_chain/src/execution_payload.rs
eserilev: removed this bit of code after separating gloas get payload to its own code path

beacon_node/http_api/tests/tests.rs
```diff
@@ -3726,6 +3727,251 @@ impl ApiTester {
...
+            };
+
+            self.client
+                .post_beacon_execution_payload_envelope(&signed_envelope)
+                .await
+                .unwrap();
+
+            self.chain.slot_clock.set_slot(slot.as_u64() + 1);
+        }
+
```
jimmygchen, line 3889: Maybe we can extract common code from these two `block_production_v4` functions to reduce duplicates?

beacon_node/beacon_chain/src/errors.rs
eserilev: [21738c0](https://github.com/sigp/lighthouse/pull/8754/commits/21738c04c3fe11655685630bb7bfe0fe879989cf)

beacon_node/http_api/src/beacon/execution_payload_envelope.rs
```diff
@@ -0,0 +1,129 @@
...
+
+// POST beacon/execution_payload_envelope
+pub(crate) fn post_beacon_execution_payload_envelope<T: BeaconChainTypes>(
+    eth_v1: EthV1Filter,
+    task_spawner_filter: TaskSpawnerFilter<T>,
+    chain_filter: ChainFilter<T>,
+    network_tx_filter: NetworkTxFilter<T>,
+) -> ResponseFilter {
+    eth_v1
+        .clone()
```
jimmygchen: this clone can be removed

beacon_node/beacon_chain/src/block_production/gloas.rs
eserilev: you mean cache by slot instead of by root? yeah solely because of the beacon api. I added a TODO so we can revist this in the future

beacon_node/http_api/src/beacon/execution_payload_envelope.rs
eserilev: I deleted this for now and updated the TODO

beacon_node/http_api/src/beacon/execution_payload_envelope.rs
```diff
@@ -0,0 +1,110 @@
...
+        builder_index = envelope.message.builder_index,
+        "Publishing signed execution payload envelope to network"
+    );
+
+    // Publish to the network
+    crate::utils::publish_pubsub_message(
+        network_tx,
+        PubsubMessage::ExecutionPayload(Box::new(envelope)),
+    )
+    .map_err(|_| {
```
jimmygchen, line 108: yes, we should perform gossip validation before publishing

validator_client/validator_services/src/block_service.rs
eserilev: [621bd9f](https://github.com/sigp/lighthouse/pull/8754/commits/621bd9f71e3ee960c4736c65a686e788f2ddf2be)

beacon_node/http_api/src/beacon/execution_payload_envelope.rs
```diff
@@ -0,0 +1,110 @@
...
+    task_spawner_filter: TaskSpawnerFilter<T>,
+    chain_filter: ChainFilter<T>,
+    network_tx_filter: NetworkTxFilter<T>,
+) -> ResponseFilter {
+    eth_v1
+        .clone()
+        .and(warp::path("beacon"))
+        .and(warp::path("execution_payload_envelope"))
+        .and(warp::path::end())
+        .and(warp::body::json())
```
jimmygchen, line 63: There's also a proposed mandatory consensus version header in the beacon-API spec PR

https://github.com/ethereum/beacon-APIs/blob/25de51f8e0b13f7873b22c6dabacd44a9a816316/apis/beacon/execution_payload/envelope_post.yaml#L23-L24

I'm not sure if it's required?

beacon_node/http_api/src/validator/execution_payload_envelope.rs
```diff
@@ -0,0 +1,105 @@
...
+                            .header("Content-Type", "application/octet-stream")
+                            .header("Eth-Consensus-Version", fork_name.to_string())
+                            .body(envelope.as_ssz_bytes().into())
+                            .map_err(|e| {
+                                warp_utils::reject::custom_server_error(format!(
+                                    "Failed to build SSZ response: {e}"
+                                ))
+                            }),
+                        _ => {
+                            let json_response = GenericResponse { data: envelope };
```
jimmygchen: Should this be `ForkVersionedResponse`?

validator_client/lighthouse_validator_store/src/lib.rs
eserilev: 7226bba

validator_client/validator_services/src/block_service.rs
eserilev: `request()` only returns a `Result<(),...>`. I'll just use `request_proposers_first` for now across both api requests

eserilev: 5442281

beacon_node/http_api/src/beacon/execution_payload_envelope.rs
eserilev: yeah probably, the impl here was based on an older PR that was closed last week. I thnk we'll need a separate issue to revisit these endpoints because right now the spec is very fluid

common/eth2/src/lib.rs
eserilev: same response as below: https://github.com/sigp/lighthouse/pull/8754#discussion_r2810998051

beacon_node/http_api/src/beacon/execution_payload_envelope.rs
eserilev: https://github.com/sigp/lighthouse/issues/8828

beacon_node/http_api/src/validator/execution_payload_envelope.rs
eserilev: 489ef00

beacon_node/http_api/src/beacon/execution_payload_envelope.rs
eserilev: removed

common/eth2/src/lib.rs
eserilev: added ssz variant

validator_client/validator_services/src/block_service.rs
eserilev: yes: [857d6ad](https://github.com/sigp/lighthouse/pull/8754/commits/857d6ad3c4094f7ac73249c6ec0a75d6c062bc33)

common/eth2/src/types.rs
eserilev: updated comment

beacon_node/http_api/tests/tests.rs
eserilev: extracted some of the comon code

beacon_node/http_api/src/beacon/execution_payload_envelope.rs
eserilev: will handle it in a the payload processing branch once this is merged

beacon_node/beacon_chain/src/block_production/gloas.rs
eserilev: collapsed

beacon_node/http_api/src/lib.rs
eserilev: removed redundant clones

beacon_node/beacon_chain/src/block_production/gloas.rs
eserilev: added comment

consensus/types/src/core/consts.rs
eserilev: moved to block_production module

common/eth2/src/lib.rs
```diff
@@ -2408,6 +2409,275 @@ impl BeaconNodeHttpClient {
...
+        path.path_segments_mut()
+            .map_err(|()| Error::InvalidUrl(self.server.clone()))?
+            .push("beacon")
+            .push("execution_payload_envelope");
+
+        self.post_generic_with_consensus_version_and_ssz_body(
+            path,
+            envelope.as_ssz_bytes(),
+            Some(self.timeouts.proposal),
+            ForkName::Gloas,
```
jimmygchen: This is a potential footgun - execution envelope will eventually be a superstruct, i think a safer way to do this is to implement a `fork_name` function on the `ExecutionPayloadEnvelope`

common/eth2/src/types.rs
jimmygchen: I don't see the update,  I'll push an update for this

beacon_node/http_api/src/beacon/execution_payload_envelope.rs
```diff
@@ -0,0 +1,109 @@
...
+pub async fn publish_execution_payload_envelope<T: BeaconChainTypes>(
+    envelope: SignedExecutionPayloadEnvelope<T::EthSpec>,
+    _chain: Arc<BeaconChain<T>>,
+    network_tx: &UnboundedSender<NetworkMessage<T::EthSpec>>,
+) -> Result<Response, Rejection> {
+    let slot = envelope.message.slot;
+    let beacon_block_root = envelope.message.beacon_block_root;
+
+    // TODO(gloas): We should probably add validation here i.e. BroadcastValidation::Gossip
+    info!(
```
jimmygchen, line 96: Be good to add a gloas fork check so there's no way to trigger this code path on a mainnet node, even though this should have no impact on the network

beacon_node/http_api/tests/tests.rs
jimmygchen: I think this function name is a bit misleading now - I've updated it to `get_builder_application_domain`

common/eth2/src/lib.rs
jimmygchen: https://github.com/sigp/lighthouse/pull/8754/commits/8bbcdf4513e5092b15300c167c73f5ee0ba1d6ea

beacon_node/http_api/src/beacon/execution_payload_envelope.rs
jimmygchen: https://github.com/sigp/lighthouse/pull/8754/commits/8bbcdf4513e5092b15300c167c73f5ee0ba1d6ea

jimmygchen (approved):
I've reviewed this and I think it's in a good shape now 🎉

PR #8762, opened 2026-02-06
fix: map IgnoredRpcBlock metric to correct queue

VolodymyrBg:
## What was wrong

`IgnoredRpcBlock` work items are enqueued into `rpc_block_queue` (grouped with `RpcBlock`), but the queue length metric was incorrectly reading from `rpc_blob_queue` (grouped with `RpcBlobs`). This caused the
`BEACON_PROCESSOR_QUEUE_LENGTH` metric for `IgnoredRpcBlock` to report the length of an unrelated queue.

## What this PR does

Moves `WorkType::IgnoredRpcBlock` from the `RpcBlobs` metric branch to the `RpcBlock` branch, matching the enqueue logic.


michaelsproul:
What's with the unrelated changes in `beacon_node/network/src/sync/tests/lookups.rs`? Can you fix this please @VolodymyrBg?

Issue #8756, opened 2026-02-04
Allow LH to build without OpenSSL


michaelsproul:
It would be nice to be able to build Lighthouse without OpenSSL.

There are probably a few places it gets pulled in:

- reqwest?
- warp?

jimmygchen:
Completed in #8768

PR #8768, opened 2026-02-07
Remove dependency on OpenSSL

macladson:
## Issue Addressed

https://github.com/sigp/lighthouse/issues/8756

## Proposed Changes

Only the Web3Signer actually needs OpenSSL in order to parse PKCS12 certificates. This updates the function to instead manually parse the cert (using the `p12-keystore` crate) and converts it to a `PEM` certificate (using the `pem` crate) which can be directly converted to a `reqwest::tls::Identity` as this can be done directly in `rustls`.

## Additional Info

There may be some situations where a .p12 certificate could be strangely formatted (for example, if not generated by openssl) and break this function where it would succeed before. I'm not sure how to quantify the risks here but we could consider adding more test cases.


testing/web3signer_tests/tls/generate.sh
```diff
@@ -1,12 +1,5 @@
 #!/bin/bash
 
-# The lighthouse/key_legacy.p12 file is generated specifically for macOS because the default `openssl pkcs12` encoding
-# algorithm in OpenSSL v3 is not compatible with the PKCS algorithm used by the Apple Security Framework. The client
-# side (using the reqwest crate) relies on the Apple Security Framework to parse PKCS files.
-# We don't need to generate web3signer/key_legacy.p12 because the compatibility issue doesn't occur on the web3signer
-# side. It seems that web3signer (Java) uses its own implementation to parse PKCS files.
-# See https://github.com/sigp/lighthouse/issues/6442#issuecomment-2469252651
```
macladson, line 8: I believe now that we are using pure Rust, we no longer rely on the Apple Security Framework and so can support OpenSSL v3 even on MacOS

ackintosh (approved):
Looks good to me!

michaelsproul:
Can we add openssl to the list of banned crates here?

https://github.com/sigp/lighthouse/blob/edba56b9a654cea555cb0db2e8d0712525686973/deny.toml#L7-L19

pawanjay176 (approved):
LGTM. The web3 signer tests pass so seem like this is good to go.

dapplion (approved):
LGTM

Issue #6484, opened 2024-10-14
De-duplicate pubkey caches across `BeaconState`/`BeaconChain`


michaelsproul:
## Description

Presently Lighthouse has two notions of a _pubkey cache_. There's the "global" cache attached to the `BeaconChain`, and "local" caches attached to each `BeaconState`.

Since in-memory `tree-states`, the local caches are stored using a persistent data structure, meaning that many beacon states can share most of the pubkey cache without duplication. The idea behind this issue is to extend this _structural sharing_ to the global public key cache.

## Implementation

One way to implement the cache would be for **new** beacon state caches to be initialised from the global cache, by cloning and making the necessary mutations. The persistent data structures (from `rpds`) then ensure that memory is shared

Another approach would be to have a global cache that is aware of changes to pubkeys over time, and can respond with results for different `epoch` values. The beacon states could contain an `Arc<RwLock<..>>` reference to this cache, and make queries. This approach might be more compatible with _validator index reuse_ should it be implemented in future, as the global cache would never "forget" about old overwritten validators. This could be advantageous when reloading old states from disk, as the cache would already contain the relevant information and would not need to be rebuilt.

dapplion:
> This approach might be more compatible with validator index reuse

I don't think we should consider index reuse for now. If it ever comes it will be in a long while. Beam chain might happen in the middle

michaelsproul:
Two separate cleanups:

- Remove `pubkey_bytes` field from global cache. It is rarely used and quite unnecessary because we can always load pubkey bytes from the head state, or load a PublicKey and quickly compress it.
- De-dupe the global and state caches. Lion and I discussed this offline and settled on an approach where we remove the cache from the `BeaconState`, and have a persistent cache that is cloned before block processing (keeps block processing lock free).

michaelsproul:
Lodestar hit an issue with Electra deposits which is worth keeping in mind when we refactor the cache:

- https://github.com/ChainSafe/lodestar/pull/7284#issuecomment-2547305198

The main thing to avoid is checking the global cache for the existence of a key that was added on a side-chain at a later epoch.

michaelsproul:
Takes ~800ms to build this cache on a state cache miss, so worth eliminating that to reduce the impact of state cache misses.

michaelsproul:
If we _don't_ abolish the local pubkey caches, there's a micro-optimisation here that we could try:

- https://github.com/sigp/lighthouse/pull/7849#discussion_r2266650349

PR #8781, opened 2026-02-10
Gloas envelope consensus and more operations tests

michaelsproul:
## Proposed Changes

- Implement new `process_execution_payload` (as `process_execution_payload_envelope`).
- Implement new processing for deposit requests, including logic for adding new builders to the registry with index reuse.
- Enable a bunch more operations EF tests (most of them except bid processing/payload attestations/etc which we don't have code for yet).

## Additional Info

- There is a rather important performance consideration in `process_deposit_request_post_gloas` which I will open a separate issue for, as it conflicts with: https://github.com/sigp/lighthouse/issues/6484.


consensus/state_processing/src/per_block_processing/process_operations.rs
```diff
@@ -763,6 +794,111 @@ pub fn process_deposit_requests<E: EthSpec>(
...
+    // [New in Gloas:EIP7732]
+    // Regardless of the withdrawal credentials prefix, if a builder/validator
+    // already exists with this pubkey, apply the deposit to their balance
+    // TODO(gloas): this could be more efficient in the builder case
+    let builder_index = state
+        .builders()?
+        .iter()
+        .enumerate()
+        .find(|(_, builder)| builder.pubkey == deposit_request.pubkey)
+        .map(|(i, _)| i as u64);
```
dapplion, line 824: Should we add state.get_builder_index ? We should add a cache for builder pubkeys -> index. Need to make them a persistent data structure as they can change based on deposit ordering

consensus/state_processing/src/envelope_processing.rs
```diff
@@ -0,0 +1,276 @@
...
+        .safe_add(state.slot().as_u64().safe_rem(E::slots_per_epoch())?)?
+        as usize;
+    let payment_mut = state
+        .builder_pending_payments_mut()?
+        .get_mut(payment_index)
+        .ok_or(EnvelopeProcessingError::BuilderPaymentIndexOutOfBounds(
+            payment_index,
+        ))?;
+
+    // Re-order the blanking out of the pending payment to avoid a double-lookup.
```
dapplion:
```suggestion
    // TODO(gloas): Re-order the blanking out of the pending payment to avoid a double-lookup.
```

consensus/state_processing/src/envelope_processing.rs
```diff
@@ -0,0 +1,276 @@
...
+    process_withdrawal_requests,
+};
+use safe_arith::{ArithError, SafeArith};
+use tree_hash::TreeHash;
+use types::{
+    BeaconState, BeaconStateError, BuilderIndex, BuilderPendingPayment, ChainSpec, EthSpec,
+    ExecutionBlockHash, Hash256, SignedExecutionPayloadEnvelope, Slot,
+};
+
+macro_rules! envelope_verify {
```
eserilev, line 15: nit: another option could be to rename the existing `block_verify` macro to `verify_condition` (or something like that) and use it here instead of defining a similar macro

consensus/types/src/state/beacon_state.rs
```diff
@@ -1966,6 +1981,60 @@ impl<E: EthSpec> BeaconState<E> {
...
+            builders.push(builder)?;
+        } else {
+            *builders
+                .get_mut(builder_index as usize)
+                .ok_or(BeaconStateError::UnknownBuilder(builder_index))? = builder;
+        }
+        Ok(builder_index)
+    }
+
+    pub fn get_index_for_new_builder(&self) -> Result<BuilderIndex, BeaconStateError> {
```
dapplion, line 2032:
```suggestion
    // TODO(gloas): Optimize this function if we see a lot of registered builders on-chain
    pub fn get_index_for_new_builder(&self) -> Result<BuilderIndex, BeaconStateError> {
```

dapplion (approved):
Looks good to me, and spec is correct. Some minor comments about optimizing but we can go live on gloas without them and do it later if builder adoption explodes

consensus/state_processing/src/per_block_processing/process_operations.rs
michaelsproul: Yeah, this is what I wanted to open up discussion about. It seems a bit annoying to do this, because it's yet-another-cache, and potentially duplicated against a global builder pubkey cache (if we add one in future).

I'll open an issue

consensus/state_processing/src/envelope_processing.rs
michaelsproul: Not a TODO, this is a statement of intent. I'll clarify that in the comment.

consensus/state_processing/src/per_block_processing/process_operations.rs
michaelsproul: - https://github.com/sigp/lighthouse/issues/8783

consensus/types/src/state/beacon_state.rs
michaelsproul: I'll add this, but I would be quite hesitant to add any kind of cache here as it would be both slot and balance dependent. Could be fiddly.

eserilev (approved):
matches the spec, except the new payload TODO. nice work!

consensus/state_processing/src/envelope_processing.rs
michaelsproul: yeah there was a TODO about removing this, but it felt pretty harmless to keep this macro given how short it is

michaelsproul:
Just pushed two new commits:

- More operations tests: 64a261b00817b095bb689c654eae37253d4b0710
- Addressing Lion's feedback: 2e13d7ff2f0161c8928e7d2217399b5a79d67e12

eserilev (approved):

PR #8481, opened 2025-11-27
Add User-Agent header to validator client HTTP requests

jimmygchen:
## Issue Addressed

Closes #7963

## Proposed Changes

Adds a `User-Agent` HTTP header to validator client requests to beacon nodes. The header format is `Lighthouse/v<version>-<commit>` (e.g., `Lighthouse/v8.0.0-67da032`).

This helps distributed validator middleware operators identify what their peers are running, similar to how Prysm already implements this.

PR #8786, opened 2026-02-10
add user-agent header to all http clients

### LINKED ISSUE #7963: Add `User-Agent` HTTP header for validator client requests

dapplion:
## Issue Addressed

Closes #7963

## Proposed Changes

Adds a `User-Agent` HTTP header to all outgoing HTTP requests. The header format is `Lighthouse/v<version>-<commit>` (e.g., `Lighthouse/v8.0.0-67da032`).

This helps distributed validator middleware operators identify what their peers are running.

**Clients updated:**
- `BeaconNodeHttpClient` (BN API client)
- `ValidatorClientHttpClient` (VC API client)
- `HttpJsonRpc` (execution layer engine API)
- `BuilderHttpClient` (already had user-agent)
- `MonitoringHttpClient` (metrics endpoint)
- Web3Signer client
- `ProductionValidatorClient` beacon node builder

**Key design decision:** the `user_agent` is passed as a `&str` constructor parameter rather than having library crates depend on `lighthouse_version` directly. This avoids cargo cache invalidation cascading through the dependency tree on every git change, since `lighthouse_version` uses `git_version!()`.

Only top-level/binary crates reference `lighthouse_version::VERSION`. Library crates (`eth2`, `initialized_validators`, `beacon_node_fallback`) receive it as a parameter.

## Related PRs

- #8481 — initial implementation for `BeaconNodeHttpClient` only
- #7990 — broader implementation with tests (stalled)

## Test plan

- [x] `cargo check` passes
- [x] `cargo fmt --all` clean
- [x] `make lint-fix` passes


chong-he:
If I understand correctly, after this change, when the VC calls the HTTP requests to the BN, it should include its version. 

I tested a running BN and VC and run: `sudo tcpdump -A -i lo port 5052`, I see the request sent by VC does not contain the `user-agent`, only has `accept` and `host`:

```
22:04:47.429301 IP localhost.54174 > localhost.5052: Flags [.], ack 32, win 510, options [nop,nop,TS val 3648558767 ecr 3648558767], length 0
E..4..@.@.D>..............A.!..c.....(.....
.x...x..
22:04:51.803986 IP localhost.54184 > localhost.5052: Flags [P.], seq 3425:3497, ack 105371, win 881, options [nop,nop,TS val 3648563141 ecr 3648558342], length 72
E..|..@.@.^..............v4..R.....q.p.....
.x...x..GET /eth/v1/node/version HTTP/1.1
accept: */*
host: localhost:5052


22:04:51.804203 IP localhost.5052 > localhost.54184: Flags [P.], seq 105371:105588, ack 3497, win 512, options [nop,nop,TS val 3648563142 ecr 3648563141], length 217
E.... @.@................R...v4H...........
.x...x..HTTP/1.1 200 OK
content-type: application/json
server: Lighthouse/v8.1.0-39edcf2/x86_64-linux
content-length: 61
date: Fri, 13 Feb 2026 14:04:51 GMT

{"data":{"version":"Lighthouse/v8.1.0-39edcf2/x86_64-linux"}}
```

common/eth2/src/lib.rs
```diff
@@ -176,6 +176,14 @@ impl BeaconNodeHttpClient {
...
     }
 
+    pub fn with_user_agent(mut self, user_agent: &str) -> Self {
+        self.client = reqwest::Client::builder()
+            .user_agent(user_agent)
+            .build()
+            .unwrap_or_else(|_| reqwest::Client::new());
+        self
+    }
+
```
macladson: Just noting that there is a footgun here where if you construct a custom `reqwest::Client` and call `from_components` to build the `BeaconNodeHttpClient`, _then_ call `with_user_agent`, then the previous configuration gets silently overwritten. Currently we only use it after `new` which doesn't have that problem.

I've been thinking for a while that `BeaconNodeHttpClient` (and `ValidatorClientHttpClient`) might need a builder pattern which wraps the `reqwest::ClientBuilder`, but generally we don't really need complicated setups so for now, I think constructing it `from_components` is fine.

So with that said, since this interface is public and is designed for external use, I would favour removing this function and just calling the `reqwest::ClientBuilder`'s `user_agent` method directly and using `from_components` instead.

common/eth2/src/lighthouse_vc/http_client.rs
```diff
@@ -66,6 +66,14 @@ impl ValidatorClientHttpClient {
...
     }
 
+    pub fn with_user_agent(mut self, user_agent: &str) -> Self {
+        self.client = reqwest::Client::builder()
+            .user_agent(user_agent)
+            .build()
+            .unwrap_or_else(|_| reqwest::Client::new());
+        self
+    }
+
```
macladson: Same applies here

macladson (changes requested):
I believe the reason it isn't working is because we never actually set the user agent when constructing the `BeaconNodeHttpClient` instance that is used by the VC on startup. We need to add a
```rust
.user_agent(&config.user_agent)
```
here:


https://github.com/sigp/lighthouse/blob/edba56b9a654cea555cb0db2e8d0712525686973/validator_client/src/lib.rs#L287-L291

Also, I noticed that while the PR mentions updating `HttpJsonRpc` and `MonitoringHttpClient` (and web3signer) but I don't see any diffs for those. Are those still required?

KaloyanTanev:
Hey @dapplion, is there an update on this one?

chong-he:
Just did some minor fix. I believe @dapplion has addressed the comments. I tested with BN + VC and ca see that the `user-agent` requests now has Lighthouse version:

```
accept: */*
user-agent: Lighthouse/v8.1.3-c6bfbc9
host: localhost:5052
```

This PR is ready for another review.

validator_client/src/config.rs
```diff
@@ -92,6 +92,9 @@ pub struct Config {
     #[serde(flatten)]
     pub initialized_validators: InitializedValidatorsConfig,
     pub disable_attesting: bool,
+    /// User-Agent string for HTTP clients (set by the top-level binary).
+    #[serde(skip)]
```
michaelsproul, line 98: For testing we might need to take off `serde(skip)`.

beacon_node/client/src/config.rs
```diff
@@ -79,6 +79,9 @@ pub struct Config {
     pub genesis_state_url: Option<String>,
     pub genesis_state_url_timeout: Duration,
     pub allow_insecure_genesis_sync: bool,
+    /// User-Agent string for HTTP clients (set by the top-level binary).
+    #[serde(skip)]
+    pub user_agent: String,
```
michaelsproul, line 84: For testing:

- One BN test
- One VC test

Check that all "downstream" configs with `user_agent` are initialised when the top-level config is initialised.

chong-he: Added some tests in [add test](https://github.com/sigp/lighthouse/pull/8786/commits/e45204ecb944b3e12d8664adedc6460c6bd9cfc3)

KaloyanTanev:
Hey guys, is there any chance we can get this merged and tagged before gloas? We'd love to keep get data of node operators which are on pre-gloas versions of their VCs and raise warnings to update.

macladson:
Hi @KaloyanTanev, most of the team is pretty flat out with Gloas at the moment but I'll try to review this by the end of the week and try to get this over the line

PR #8793, opened 2026-02-10
Correctly compute process times during `ProcessHealth::observe`

macladson:
## Issue Addressed
I believe I found a bug where during computation of `pid_process_seconds_total` we add `children_system` twice.

## Proposed Changes

I assume that it was originally intended to add `children_system` and `children_user` once each


dapplion (approved):
Nice catch! Looks good to me

PR #8795, opened 2026-02-10
rename --reconstruct-historic-states to --archive

dapplion:
## Summary

- Rename the `--reconstruct-historic-states` CLI flag to `--archive`
- Keep `--reconstruct-historic-states` as a hidden alias for backwards compatibility
- Rename the internal config field `reconstruct_historic_states` to `archive`
- Update all documentation references in the book
- Update and add CLI tests (including alias test)

## Test plan

- [x] `cargo check` passes
- [x] `cargo fmt` clean
- [x] `make lint-full` passes (ran via pre-push hook)
- [x] CLI tests pass in release mode: `archive_flag`, `archive_flag_alias`, `no_archive_flag`, `genesis_backfill_with_archive_flag`

@dapplion 
_spec'ed by me, written by opus 4.6, manually reviewed by me_

### DIFF
--- a/beacon_node/src/cli.rs
+++ b/beacon_node/src/cli.rs
@@ -1246,8 +1246,9 @@ pub fn cli_app() -> Command {
                 .display_order(0)
         )
         .arg(
-            Arg::new("reconstruct-historic-states")
-                .long("reconstruct-historic-states")
+            Arg::new("archive")
+                .long("archive")
+                .alias("reconstruct-historic-states")
                 .help("After a checkpoint sync, reconstruct historic states in the database. This requires syncing all the way back to genesis.")
                 .action(ArgAction::SetTrue)
                 .help_heading(FLAG_HEADER)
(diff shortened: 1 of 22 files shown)


beacon_node/src/cli.rs
```diff
@@ -1246,8 +1246,9 @@ pub fn cli_app() -> Command {
                 .display_order(0)
         )
         .arg(
-            Arg::new("reconstruct-historic-states")
-                .long("reconstruct-historic-states")
+            Arg::new("archive")
+                .long("archive")
+                .alias("reconstruct-historic-states")
```
michaelsproul, line 1251: Technically this description is wrong, the flag is now also relevant for genesis sync. Maybe we could state something general like "store all states" and elaborate that for checkpoint sync it means the states will be reconstructed after backfill completes

michaelsproul (changes requested):
Looks good. The only change I can think of that's missing is a deprecation warning for the old flag. Unless we don't want to deprecate it yet?

dapplion:
Updated the help text to be more general: "Store all beacon states in the database. When checkpoint syncing, states are reconstructed after backfill completes."

Re deprecation: we'd prefer to keep `--reconstruct-historic-states` as a permanent alias rather than deprecate it, so no warning added.

michaelsproul (approved):
Looks good now, thanks!

I'll label this as backwards-incompat (even though it isn't technically) so it gets flagged for a mention in the release notes.

PR #8801, opened 2026-02-11
Gloas payload bid consensus

michaelsproul:
## Proposed Changes

- [x] Consensus changes for execution payload bids
- [x] EF tests for bids (and `block_header` -- no changes required).

### DIFF
--- a/consensus/state_processing/src/per_block_processing.rs
+++ b/consensus/state_processing/src/per_block_processing.rs
@@ -1,12 +1,17 @@
 use crate::consensus_context::ConsensusContext;
-use errors::{BlockOperationError, BlockProcessingError, HeaderInvalid};
+use errors::{
+    BlockOperationError, BlockProcessingError, ExecutionPayloadBidInvalid, HeaderInvalid,
+};
 use rayon::prelude::*;
 use safe_arith::{ArithError, SafeArith};
-use signature_sets::{block_proposal_signature_set, get_pubkey_from_state, randao_signature_set};
+use signature_sets::{
+    block_proposal_signature_set, execution_payload_bid_signature_set,
+    get_builder_pubkey_from_state, get_pubkey_from_state, randao_signature_set,
+};
 use std::borrow::Cow;
 use tree_hash::TreeHash;
 use typenum::Unsigned;
-use types::*;
+use types::{consts::gloas::BUILDER_INDEX_SELF_BUILD, *};
 
 pub use self::verify_attester_slashing::{
     get_slashable_indices, get_slashable_indices_modular, verify_attester_slashing,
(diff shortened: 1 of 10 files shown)


consensus/state_processing/src/per_block_processing/signature_sets.rs
```diff
@@ -30,6 +31,9 @@ pub enum Error {
     ValidatorUnknown(u64),
     /// Attempted to find the public key of a validator that does not exist. You cannot distinguish
```
pawanjay176:
```suggestion
    /// Attempted to find the public key of a builder that does not exist. You cannot distinguish
```

consensus/state_processing/src/per_block_processing/signature_sets.rs
```diff
@@ -71,6 +75,25 @@ where
         .map(Cow::Owned)
 }
 
+/// Helper function to get a validator public key from a `state`.
```
eserilev:
```suggestion
/// Helper function to get a builder public key from a `state`.
```

consensus/state_processing/src/per_block_processing/errors.rs
```diff
@@ -452,6 +461,38 @@ pub enum ExitInvalid {
...
+    BadSignature,
+    /// The builder is not active.
+    BuilderNotActive(u64),
+    /// The builder has insufficient balance to cover the bid
+    InsufficientBalance {
+        builder_index: u64,
+        builder_balance: u64,
+        bid_value: u64,
+    },
+    /// Bid slot doesn't match state slot
```
pawanjay176: Isn't this for the block slot?

pawanjay176 (approved):
LGTM, just nits

eserilev (approved):
LGTM!

consensus/state_processing/src/per_block_processing/signature_sets.rs
eserilev: oh whoops didnt see pawans review lol

consensus/state_processing/src/per_block_processing.rs
```diff
@@ -522,3 +527,171 @@ pub fn compute_timestamp_at_slot<E: EthSpec>(
...
+        }
+        .into()
+    );
+
+    block_verify!(
+        bid.parent_block_root == block.parent_root(),
+        ExecutionPayloadBidInvalid::ParentBlockRootMismatch {
+            block_parent_root: block.parent_root(),
+            bid_parent_root: bid.parent_block_root,
+        }
```
dapplion: 🤖 Duplicate `parent_block_root` check — this block is identical to the one right above it (lines 642-649). Looks like a copy-paste artifact.

dapplion (commented):
Spec comparison review — compared against `consensus-specs/specs/gloas/beacon-chain.md` and [potuz's annotated GLOAS spec](https://www.potuz.net/posts/gloas-annotated-1/).

consensus/state_processing/src/per_block_processing/signature_sets.rs
```diff
@@ -30,6 +31,9 @@ pub enum Error {
     ValidatorUnknown(u64),
     /// Attempted to find the public key of a validator that does not exist. You cannot distinguish
     /// between an error and an invalid block in this case.
+    BuilderUnknown(BuilderIndex),
+    /// Attempted to find the public key of a validator that does not exist. You cannot distinguish
```
dapplion, line 35: 🤖 The doc comment above `BuilderUnknown` (line 32) still says "validator" — should say "builder". Same issue pawan flagged on the function doc, but this one in the `Error` enum wasn't mentioned.

consensus/state_processing/src/per_block_processing.rs
```diff
@@ -522,3 +527,171 @@ pub fn compute_timestamp_at_slot<E: EthSpec>(
...
+
+        let payment_index = E::SlotsPerEpoch::to_usize()
+            .safe_add(bid.slot.as_usize().safe_rem(E::SlotsPerEpoch::to_usize())?)?;
+
+        *state
+            .builder_pending_payments_mut()?
+            .get_mut(payment_index)
+            .ok_or(BlockProcessingError::BeaconStateError(
+                BeaconStateError::InvalidBuilderPendingPaymentsIndex(payment_index),
+            ))? = pending_payment;
```
dapplion, line 681: **Update**: I compared against [the stale local spec](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/beacon-chain.md) initially. The latest upstream spec no longer has `pending_balance_removals` — it matches Lighthouse's approach exactly:

```python
def can_builder_cover_bid(
    state: BeaconState, builder_index: BuilderIndex, bid_amount: Gwei
) -> bool:
    builder_balance = state.builders[builder_index].balance
    pending_withdrawals_amount = get_pending_balance_to_withdraw_for_builder(state, builder_index)
    min_balance = MIN_DEPOSIT_AMOUNT + pending_withdrawals_amount
    if builder_balance < min_balance:
        return False
    return builder_balance - min_balance >= bid_amount
```

So this implementation is correct and aligned with the latest spec. No action needed here.

consensus/state_processing/src/per_block_processing.rs
```diff
@@ -522,3 +527,171 @@ pub fn compute_timestamp_at_slot<E: EthSpec>(
...
+
+    // Verify commitments are under limit
+    let max_blobs_per_block = spec.max_blobs_per_block(state.current_epoch()) as usize;
+    block_verify!(
+        bid.blob_kzg_commitments.len() <= max_blobs_per_block,
+        ExecutionPayloadBidInvalid::ExcessBlobCommitments {
+            max: max_blobs_per_block,
+            bid: bid.blob_kzg_commitments.len(),
+        }
+        .into()
```
dapplion, line 619: **Update**: I was comparing against a stale local spec. The [latest upstream spec](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/beacon-chain.md) now uses `blob_kzg_commitments: List[KZGCommitment, MAX_BLOB_COMMITMENTS_PER_BLOCK]` (full list, not a root) and includes this same length check:

```python
assert (
    len(bid.blob_kzg_commitments)
    <= get_blob_parameters(get_current_epoch(state)).max_blobs_per_block
)
```

So this is spec-aligned. No action needed.

eserilev: im not sure I follow this comment, the bid in the spec has the `blob_kzg_commitments`, not the root.

consensus/state_processing/src/per_block_processing.rs
eserilev: you might be looking at an old spec? i dont see a mention of `PendingBalanceRemoval` in the current spec

consensus/state_processing/src/per_block_processing/signature_sets.rs
eserilev: nope this comment is for the error `ValidatorPubkeyUnknown` variant. It should still say `validator`

consensus/state_processing/src/per_block_processing.rs
eserilev: I also dont see a mention of `balance_proof` in the current spec

consensus/state_processing/src/per_block_processing.rs
michaelsproul: Ok one good catch by the robot, will fix.

consensus/state_processing/src/per_block_processing.rs
dapplion: Right, potuz spec is outdated already...

consensus/state_processing/src/per_block_processing/signature_sets.rs
michaelsproul: Addressed in fd6839f42534b2761d7412e1b3ec41ad4914977f

consensus/state_processing/src/per_block_processing/errors.rs
michaelsproul: Addressed in fd6839f42534b2761d7412e1b3ec41ad4914977f

consensus/state_processing/src/per_block_processing.rs
michaelsproul: Addressed in fd6839f42534b2761d7412e1b3ec41ad4914977f

michaelsproul:
All comments reviewed, thanks gang

consensus/types/src/builder/builder.rs
```diff
@@ -24,3 +24,12 @@ pub struct Builder {
...
 }
+
+impl Builder {
+    /// Check if a builder is active in a state with `finalized_epoch`.
+    ///
+    /// This implements `is_active_builder` from the spec.
+    pub fn is_active_at_finalized_epoch(&self, finalized_epoch: Epoch, spec: &ChainSpec) -> bool {
+        self.deposit_epoch < finalized_epoch && self.withdrawable_epoch == spec.far_future_epoch
+    }
+}
```
dapplion, line 35: Nit: the spec calls this [`is_active_builder`](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/beacon-chain.md#new-is_active_builder). Would it be worth renaming to `is_active_builder` for easier cross-referencing? The method-on-Builder approach is fine to avoid a redundant list lookup.

michaelsproul: I thought about this, but I figure including builder in the name is redundant when it's a method on builder, and putting `finalized_epoch` in the name makes it clear at the call site that you must pass the finalized epoch and not e.g. the current epoch

consensus/types/src/builder/builder.rs
```diff
@@ -24,3 +24,12 @@ pub struct Builder {
     pub deposit_epoch: Epoch,
     pub withdrawable_epoch: Epoch,
 }
+
+impl Builder {
+    /// Check if a builder is active in a state with `finalized_epoch`.
+    ///
+    /// This implements `is_active_builder` from the spec.
+    pub fn is_active_at_finalized_epoch(&self, finalized_epoch: Epoch, spec: &ChainSpec) -> bool {
```
dapplion, line 32: Should this be named `is_active_builder` to match spec naming?

michaelsproul: Farming comment credits? Isn't this the same as your other comment? https://github.com/sigp/lighthouse/pull/8801#discussion_r2796127964

eserilev (approved):

PR #8804, opened 2026-02-12
fix: auto-populate ENR UDP port from discovery listen port

selfuryon:
When `--discovery-port` is set without `--enr-udp-port`, the UDP port was missing from the ENR. TCP and QUIC ports already had fallback logic to use the listen port, but UDP did not. This adds the same .`or_else()` fallback for udp4/udp6 using disc_port from listen addresses.


pawanjay176 (approved):
LGTM. Thanks for the PR.
Paging @jxs for checking if we are missing anything

jxs (approved):
thanks for this, and thanks for the ping Pawan! LGTM

PR #8806, opened 2026-02-12
Gloas payload envelope processing

eserilev:
Implement payload envelope gossip verification and processing


beacon_node/beacon_chain/src/execution_payload.rs
```diff
@@ -152,10 +160,11 @@ async fn notify_new_payload<T: BeaconChainTypes>(
...
                     ?execution_block_hash,
-                    root = ?block.tree_hash_root(),
-                    graffiti = block.body().graffiti().as_utf8_lossy(),
-                    proposer_index = block.proposer_index(),
-                    slot = %block.slot(),
+                    // TODO(gloas) are these other logs important?
+                    root = ?beacon_block_root,
+                    // graffiti = block.body().graffiti().as_utf8_lossy(),
+                    // proposer_index = block.proposer_index(),
+                    // slot = %block.slot(),
```
eserilev: we wont have this info post gloas, I'd like to delete these fields if thats okay

beacon_node/beacon_chain/src/execution_payload.rs
```diff
@@ -197,10 +206,11 @@ async fn notify_new_payload<T: BeaconChainTypes>(
...
```
eserilev: would like to delete these as well

eserilev:
This one should be ready for a first pass

eserilev:
I tried looking at removing the beacon chain dependency from the gossip path. It didn't seem to make unit testing that much easier, the dependency on `CachedHead` adds complexity. I think we can take a look at making these codepaths more unit-testable later on down the line

pawanjay176:
I'll take a look today

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -1151,6 +1154,23 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
 
+    /// Returns the full block at the given root, if it's available in the database.
+    ///
+    /// Should always return a full block for pre-merge and post-gloas blocks.
+    pub fn get_full_block(
+        &self,
+        block_root: &Hash256,
+    ) -> Result<Option<SignedBeaconBlock<T::EthSpec>>, Error> {
+        match self.store.try_get_full_block(block_root)? {
+            Some(DatabaseBlock::Full(block)) => Ok(Some(block)),
```
pawanjay176: We could potentially look up a block status table here to figure out if we have processed the block and it turned out to be invalid

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
```diff
@@ -0,0 +1,318 @@
...
+    pub fn new(
+        signed_envelope: Arc<SignedExecutionPayloadEnvelope<T::EthSpec>>,
+        chain: &BeaconChain<T>,
+    ) -> Result<Self, EnvelopeError> {
+        let envelope = &signed_envelope.message;
+        let payload = &envelope.payload;
+        let beacon_block_root = envelope.beacon_block_root;
+
+        // Check that we've seen the beacon block for this envelope and that it passes validation.
+        // TODO(EIP-7732): We might need some type of status table in order to differentiate between:
```
pawanjay176, line 96: This is a great point in favour of having some sort of a block status table because invalid blocks aren't kept in fork choice. 
So for a payload referencing an invalid block, we have no way of knowing if we have already processed and rejected the block without keeping that info around.

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
```diff
@@ -0,0 +1,318 @@
...
+        //
+        // Presently these two cases are conflated.
+        let fork_choice_read_lock = chain.canonical_head.fork_choice_read_lock();
+        let Some(proto_block) = fork_choice_read_lock.get_block(&beacon_block_root) else {
+            return Err(EnvelopeError::BlockRootUnknown {
+                block_root: beacon_block_root,
+            });
+        };
+
+        let latest_finalized_slot = fork_choice_read_lock
```
pawanjay176: Should we be using the `canonical_head` for this like we do for blocks? Also, we should probably get this post making sure the block is available in fork choice

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
```diff
@@ -0,0 +1,318 @@
...
+        if payload.block_hash != execution_bid.block_hash {
+            return Err(EnvelopeError::BlockHashMismatch {
+                committed_bid: execution_bid.block_hash,
+                envelope: payload.block_hash,
+            });
+        }
+
+        // Verify the envelope signature.
+        //
+        // For self-build envelopes, we can use the proposer cache for the fork and the
```
pawanjay176:
```suggestion
        // For self-built envelopes, we can use the proposer cache for the fork and the
```

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
```diff
@@ -0,0 +1,318 @@
...
+        let beacon_block_root = envelope.beacon_block_root;
+
+        // Check that we've seen the beacon block for this envelope and that it passes validation.
+        // TODO(EIP-7732): We might need some type of status table in order to differentiate between:
+        //
+        // 1. Blocks we haven't seen (IGNORE), and
+        // 2. Blocks we've seen that are invalid (REJECT).
+        //
+        // Presently these two cases are conflated.
+        let fork_choice_read_lock = chain.canonical_head.fork_choice_read_lock();
```
pawanjay176: If we have a block_processing_table, we could have a `Processed(Bid, bool)` state that is only entered post adding to fork choice. That way, we could potentially need only a single call to make sure the block is valid and to do all consequent checks with the bid

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
```diff
@@ -0,0 +1,318 @@
...
+        // Verify the envelope signature.
+        //
+        // For self-build envelopes, we can use the proposer cache for the fork and the
+        // validator pubkey cache for the proposer's pubkey, avoiding a state load from disk.
+        // For external builder envelopes, we must load the state to access the builder registry.
+        let builder_index = envelope.builder_index;
+        let block_slot = envelope.slot;
+        let block_epoch = block_slot.epoch(T::EthSpec::slots_per_epoch());
+        let proposer_shuffling_decision_block =
+            proto_block.proposer_shuffling_root_for_child_block(block_epoch, &chain.spec);
```
pawanjay176: I'm not sure this is correct. 

We need the shuffling for proto_block, not proto_block's child block afaiu.

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
```diff
@@ -0,0 +1,318 @@
...
+                    let snapshot = load_snapshot(envelope_ref, chain)?;
+                    opt_snapshot = Some(Box::new(snapshot.clone()));
+                    Ok((snapshot.state_root, snapshot.pre_state))
+                },
+            )?;
+            let fork = proposer.fork;
+
+            let pubkey_cache = chain.validator_pubkey_cache.read();
+            let pubkey = pubkey_cache
+                .get(block.message().proposer_index() as usize)
```
pawanjay176, line 188: If we are using the proposer_index from the block, shouldn't we also check if it matches what we computed in `proposer` above?

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
```diff
@@ -0,0 +1,318 @@
...
+            let proposer = chain.with_proposer_cache::<_, EnvelopeError>(
+                proposer_shuffling_decision_block,
+                block_epoch,
+                |proposers| proposers.get_slot::<T::EthSpec>(block_slot),
+                || {
+                    debug!(
+                        %beacon_block_root,
+                        "Proposer shuffling cache miss for envelope verification"
+                    );
+                    let snapshot = load_snapshot(envelope_ref, chain)?;
```
pawanjay176: just noting that `load_snapshot` again acquires the fork_choice lock which it uses to basically get what we have as `proto_block` in the current scope. Maybe we can pass an optional `proto_block` to the function as well?

pawanjay176: actually this is just a read_lock so maybe not that big a deal

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
```diff
@@ -0,0 +1,318 @@
...
+                    opt_snapshot = Some(Box::new(snapshot.clone()));
+                    Ok((snapshot.state_root, snapshot.pre_state))
+                },
+            )?;
+            let fork = proposer.fork;
```
pawanjay176, line 177: Just wondering if the fork used here v/s in the non self build case below would be identical. We had issues in the past wrt the state not being advanced to the new fork and that causing signature verification errors.

I think this is safe, but would be good if @michaelsproul can confirm

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
pawanjay176, line 1: maybe we can name this file envelope_verification given that we do more than gossip verification here?

beacon_node/beacon_chain/src/payload_envelope_verification/payload_notifier.rs
```diff
@@ -0,0 +1,93 @@
...
+        block: Arc<SignedBeaconBlock<T::EthSpec>>,
+        notify_execution_layer: NotifyExecutionLayer,
+    ) -> Result<Self, EnvelopeError> {
+        let payload_verification_status = {
+            let payload_message = &envelope.message;
+
+            match notify_execution_layer {
+                NotifyExecutionLayer::No if chain.config.optimistic_finalized_sync => {
+                    let new_payload_request = Self::build_new_payload_request(&envelope, &block)?;
+                    if let Err(e) = new_payload_request.perform_optimistic_sync_verifications() {
```
pawanjay176, line 36: I think we should add a todo here to reassess if optimistic syncing works similarly to pre-gloas.

beacon_node/beacon_chain/src/execution_payload.rs
pawanjay176: slot might be useful too, otherwise i agree

pawanjay176 (commented):
Some early comments, still going through post gossip verification

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
```diff
@@ -0,0 +1,390 @@
...
+};
+
+impl<T: BeaconChainTypes> BeaconChain<T> {
+    /// Returns `Ok(block_root)` if the given `unverified_envelope` was successfully verified and
+    /// imported into the chain.
+    ///
+    /// Items that implement `IntoExecutionPendingEnvelope` include:
+    ///
+    /// - `GossipVerifiedEnvelope`
+    ///
```
pawanjay176: Add a todo here that implements this for a `SignedExecutionEnvelope` when received over rpc?

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
```diff
@@ -0,0 +1,390 @@
...
+    #[instrument(skip_all)]
+    fn import_execution_payload_envelope(
+        &self,
+        signed_envelope: AvailableEnvelope<T::EthSpec>,
+        block_root: Hash256,
+        state: BeaconState<T::EthSpec>,
+        _payload_verification_status: PayloadVerificationStatus,
+        parent_block: Arc<SignedBeaconBlock<T::EthSpec>>,
+    ) -> Result<Hash256, EnvelopeError> {
+        // ----------------------------- ENVELOPE NOT YET ATTESTABLE ----------------------------------
```
pawanjay176: What do you mean by `ATTESTABLE` here? We will attest to the ptc only based on if we have seen the envelope or not

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
```diff
@@ -0,0 +1,318 @@
...
+        self,
+        chain: &Arc<BeaconChain<T>>,
+        notify_execution_layer: NotifyExecutionLayer,
+    ) -> Result<ExecutionPendingEnvelope<T::EthSpec>, EnvelopeError> {
+        let signed_envelope = self.signed_envelope;
+        let envelope = &signed_envelope.message;
+        let payload = &envelope.payload;
+
+        // Verify the execution payload is valid
+        let payload_notifier = PayloadNotifier::new(
```
pawanjay176: We need to probably add some mechanism to ensure that we don't call new_payload multiple times similar to the duplicate_cache or a block table cache. Can add a todo here maybe?

ethDreamer: I had added a [`NewPayloadCache`](https://github.com/sigp/lighthouse/commit/f3b79839a198223c0f0dc25ce2148e1f0d3f32da) previously as a way to prevent this. But the [pr that holds it](https://github.com/sigp/lighthouse/pull/8279) has been pulled apart and merged separately. We could add it here.

beacon_node/beacon_chain/src/execution_payload.rs
eserilev: [6e89ba6](https://github.com/sigp/lighthouse/pull/8806/commits/6e89ba63be49ff34e4ab70a9ba82b2aee4841f98)

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
eserilev: good point! 
147f2e2

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
eserilev: I added this to the TODO comment

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
eserilev: oh nice catch! yes I think this might be incorrect

eserilev: I think after Fulu when `child_block_epoch == block_epoch`, we return `self.current_epoch_shuffling_id.shuffling_decision_block` which is the shuffling decision block for proto_block's own epoch. But still this is semantically wrong

eserilev: maybe we should add a method on `Block` that computes the proposer shuffling root for the block's current slot? That way we are semantically correct and we dont need to touch the fork choice lock at all. What do you think?

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
eserilev: fc7d6c9

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
eserilev: That's a good idea. I introduced a `load_snapshot_from_state_root` that is used when we already have the `proto_block`. I think that makes things a bit cleaer that an optional `proto_block`


30241f5

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
eserilev: I created a new file `execution_pending_envelope.rs` that contains all the `IntoExecutionPending` logic. What do you think about that?

eserilev: 2093dc1

beacon_node/beacon_chain/src/payload_envelope_verification/payload_notifier.rs
eserilev: added TODO

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
eserilev: so its only if we've seen the envelope? And not actually if we've imported it?

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
eserilev: I've already flagged this in a TODO in `process_gossip_execution_payload_envelope`

added a bit more detail to the TODO so that its clear

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
eserilev: added a TODO

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
```diff
@@ -0,0 +1,385 @@
...
+                    self.import_available_execution_payload_envelope(Box::new(envelope))
+                        .await
+                }
+                ExecutedEnvelope::AvailabilityPending() => Err(EnvelopeError::InternalError(
+                    "Pending payload envelope not yet implemented".to_owned(),
+                )),
+            }
+        };
+
+        // Verify and import the block.
```
dapplion: Review comments, still have block namings

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
```diff
@@ -0,0 +1,385 @@
...
+                        crit!(
+                            error = ?e,
+                            "Envelope processing error"
+                        );
+                    }
+                };
+                Err(EnvelopeError::BeaconChainError(e))
+            }
+            // The block failed verification.
+            Err(other) => {
```
dapplion, line 140: This fallback match is sus, since creation of this function there are many other internal errors that are not `Err(EnvelopeError::BeaconChainError(e))`. Why log _some_ internal errors as crits and others as warns?

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
```diff
@@ -0,0 +1,385 @@
...
+            Err(e) => {
+                error!(
+                    msg = "Restoring fork choice from disk",
+                    error = &e,
+                    ?block_root,
+                    "Failed to store data columns into the database"
+                );
+                // TODO(gloas) implement failed write handling to fork choice
+                // let _ = self.handle_import_block_db_write_error(fork_choice);
+                return Err(EnvelopeError::InternalError(e));
```
dapplion: get_blobs_or_columns_store_op is not fallible anymore, you can delete this error handling

beacon_node/beacon_chain/src/beacon_proposer_cache.rs
```diff
@@ -164,6 +165,82 @@ impl BeaconProposerCache {
...
+            epoch = %proposal_epoch,
+            "Priming proposer shuffling cache"
+        );
+
+        Ok::<_, Err>(EpochBlockProposers::new(proposal_epoch, fork, proposers))
+    })?;
+
+    // Run the accessor function on the computed epoch proposers.
+    accessor(epoch_block_proposers).map_err(Into::into)
+}
```
dapplion, line 242: Is this code unchanged? Why the move?

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
dapplion: According to the spec `[REJECT] block passes validation.` so that means full consensus validation right?

In that case we can just use the reprocessing queue keyed by block_root. Is that enough? We need the post state of the block import to import the payload anyway, and to acquire a write lock on fork-choice so the two operations are strictly sequential

beacon_node/beacon_chain/src/beacon_proposer_cache.rs
eserilev: its to remove the dependency on beacon chain. The idea was to have a `GossipVerifiedContext` that only has the fields we need to verify envelopes. Would allow us to write much lighter weight test than using the full beacon chain harness. Not sure how i feel about these changes yet, but thats the general idea

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
eserilev: `EnvelopeError` isn't necessarily an internal error while a `BeaconChainError` is

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
eserilev: oh, is there a reason why we still have it returning a `Result` then?

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
eserilev: fixed, thanks

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
eserilev: i've deleted these comments, they're copy pasta artifacts

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
dapplion: I can't find any reason

dapplion (commented):
Deleted

beacon_node/beacon_chain/src/payload_envelope_verification/execution_pending_envelope.rs
```diff
@@ -0,0 +1,109 @@
...
+            &mut state,
+            Some(snapshot.state_root),
+            &signed_envelope,
+            // verify signature already done for GossipVerifiedEnvelope
+            VerifySignatures::False,
+            VerifyStateRoot::True,
+            &chain.spec,
+        )?;
+
+        Ok(ExecutionPendingEnvelope {
```
dapplion, line 93: `MaybeAvailableEnvelope::AvailabilityPending` is always created here, but `import.rs:101-103` returns an error for that variant. So no envelope can be successfully imported through this pipeline right now. Is this intentional WIP, or should this produce `Available` when all data columns are present?

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
```diff
@@ -0,0 +1,384 @@
...
+
+        // We're declaring the envelope "imported" at this point, since fork choice and the DB know
+        // about it.
+        let envelope_time_imported = timestamp_now();
+
+        // TODO(gloas) depending on what happens with light clients
+        // we might need to do some light client related computations here
+
+        metrics::stop_timer(db_write_timer);
+        metrics::inc_counter(&metrics::ENVELOPE_PROCESSING_SUCCESSES);
```
dapplion: `ENVELOPE_PROCESSING_SUCCESSES` is already incremented at line 118 in `process_execution_payload_envelope`. This second increment would double-count on a successful import.

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
```diff
@@ -0,0 +1,446 @@
...
+            .finalized_checkpoint()
+            .epoch
+            .start_slot(T::EthSpec::slots_per_epoch());
+
+        // TODO(EIP-7732): check that we haven't seen another valid `SignedExecutionPayloadEnvelope`
+        //                 for this block root from this builder - envelope status table check
+        let block = match ctx.store.try_get_full_block(&beacon_block_root)? {
+            Some(DatabaseBlock::Full(block)) => Arc::new(block),
+            Some(DatabaseBlock::Blinded(_)) | None => {
+                return Err(EnvelopeError::from(BeaconChainError::MissingBeaconBlock(
```
dapplion, line 126: This relies in the invariant that 
```
block -> fork_choice
ensures
block -> db
```

dapplion: Why is the blinded block not sufficient?

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
dapplion: why not match 

Err(EnvelopeError::BeaconChainError(BeaconChainError::TokioJoin(e))) => {
       debug
Err(other) =>
      warn

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
```diff
@@ -0,0 +1,384 @@
...
+                msg = "Restoring fork choice from disk",
+                error = ?e,
+                "Database write failed!"
+            );
+            return Err(e.into());
+            // TODO(gloas) handle db write failure
+            // return Err(self
+            //    .handle_import_block_db_write_error(fork_choice)
+            //    .err()
+            //    .unwrap_or(e.into()));
```
dapplion, line 300: This can be un-commented now

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
```diff
@@ -0,0 +1,384 @@
...
+        }
+
+        drop(db_span);
+
+        // TODO(gloas) drop fork choice lock
+        // The fork choice write-lock is dropped *after* the on-disk database has been updated.
+        // This prevents inconsistency between the two at the expense of concurrency.
+        // drop(fork_choice);
```
dapplion, line 308: Why is this commented

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
```diff
@@ -0,0 +1,384 @@
...
+        // TODO(gloas) emit SSE event if the payload became the new head payload
+        drop(post_exec_timer);
+
+        // It is important NOT to return errors here before the database commit, because the envelope
+        // has already been added to fork choice and the database would be left in an inconsistent
+        // state if we returned early without committing. In other words, an error here would
+        // corrupt the node's database permanently.
+
+        // Store the envelope and its state, and execute the confirmation batch for the intermediate
+        // states, which will delete their temporary flags.
```
dapplion: This comment doesn't really apply anymore, we deprecated temporary flags

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
```diff
@@ -0,0 +1,384 @@
...
+            return Err(EnvelopeError::BlockRootUnknown { block_root });
+        }
+
+        // TODO(gloas) when the code below is implemented we can delete this drop
+        drop(fork_choice_reader);
+
+        // TODO(gloas) no fork choice logic yet
+        // Take an exclusive write-lock on fork choice. It's very important to prevent deadlocks by
+        // avoiding taking other locks whilst holding this lock.
+        // let fork_choice = parking_lot::RwLockUpgradableReadGuard::upgrade(fork_choice_reader);
```
dapplion, line 248: You can add a function in fork_choice that just logs to have this code uncommented

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
```diff
@@ -0,0 +1,384 @@
...
+            }
+        }
+    }
+
+    /// Accepts a fully-verified payload envelope and awaits on its payload verification handle to
+    /// get a fully `ExecutedEnvelope`.
+    ///
+    /// An error is returned if the verification handle couldn't be awaited.
+    #[instrument(skip_all, level = "debug")]
+    pub async fn into_executed_payload_envelope(
```
dapplion: Does this need to be pub?

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -3248,25 +3251,178 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                    root = ?verified_envelope.signed_envelope.beacon_block_root(),
+                    "New envelope received"
+                );
+
+                self.propagate_validation_result(message_id, peer_id, MessageAcceptance::Accept);
+
+                verified_envelope
+            }
+            // TODO(gloas) penalize peers accordingly
+            Err(_) => return None,
```
jimmygchen, line 3336: May be worth adding this before cross client interop to help us catch bugs.

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -3248,25 +3251,178 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                trace!(
+                    %slot,
+                    %block_root,
+                    "Processed envelope, waiting for other components"
+                )
+            }
+
+            Err(_) => {
+                // TODO(gloas) implement peer penalties
+                warn!("process_gossip_verified_execution_payload_envelope_failed")
```
jimmygchen: be worth logging the errors at least, and consider handling this prior to interop

beacon_node/beacon_chain/src/metrics.rs
```diff
@@ -21,6 +21,44 @@ pub const VALIDATOR_MONITOR_ATTESTATION_SIMULATOR_SOURCE_ATTESTER_HIT_TOTAL: &st
 pub const VALIDATOR_MONITOR_ATTESTATION_SIMULATOR_SOURCE_ATTESTER_MISS_TOTAL: &str =
     "validator_monitor_attestation_simulator_source_attester_miss_total";
 
+/*
+* Execution Payload Envelope Procsesing
```
jimmygchen:
```suggestion
* Execution Payload Envelope Processing
```

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -3248,25 +3251,178 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                let _process_fn = Box::pin(async move {
+                    inner_self
+                        .process_gossip_verified_execution_payload_envelope(
+                            peer_id,
+                            verified_envelope,
+                        )
+                        .await;
+                });
+
+                // TODO(gloas) send to reprocess queue
```
jimmygchen, line 3365: Raised https://github.com/sigp/lighthouse/issues/8922

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
eserilev: theres no blinded block post gloas

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
eserilev: Going to handle this in a separate PR

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
eserilev: Going to handle this in a separate PR

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
eserilev: no longer pub

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
eserilev: so we dont forget to drop the fork choice lock here

beacon_node/beacon_chain/src/payload_envelope_verification/execution_pending_envelope.rs
eserilev: this is intentional WIP

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
eserilev: removed double counting metrics

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
eserilev: removed

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
eserilev: logging the error

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
eserilev: ok sure, pushed up this change

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
eserilev: added TODO and we'll chat with Michael

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -3248,25 +3251,181 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
         }
     }
 
-    pub async fn process_gossip_execution_payload(
+    pub async fn process_gossip_execution_payload_envelope(
```
jimmygchen, line 3262: Add span here

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
```diff
@@ -0,0 +1,446 @@
...
+    /// gossip network. The envelope is not imported into the chain, it is just partially verified.
+    ///
+    /// The returned `GossipVerifiedEnvelope` should be provided to `Self::process_execution_payload_envelope` immediately
+    /// after it is returned, unless some other circumstance decides it should not be imported at
+    /// all.
+    ///
+    /// ## Errors
+    ///
+    /// Returns an `Err` if the given envelope was invalid, or an error was encountered during verification.
+    pub async fn verify_envelope_for_gossip(
```
jimmygchen, line 251: Span may be useful here

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
```diff
@@ -0,0 +1,446 @@
...
+                    block: block.message().proposer_index(),
+                    local_shuffling: expected_proposer as u64,
+                });
+            }
+
+            let pubkey_cache = ctx.validator_pubkey_cache.read();
+            let pubkey = pubkey_cache
+                .get(block.message().proposer_index() as usize)
+                .ok_or_else(|| EnvelopeError::UnknownValidator {
+                    builder_index: block.message().proposer_index(),
```
jimmygchen: should this be `proposer_index`?

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
```diff
@@ -3248,25 +3251,181 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                );
+
+                // TODO(gloas) do we need to recompute head?
+                // should canonical_head return the block and the payload now?
+                self.chain.recompute_head_at_current_slot().await;
+
+                // TODO(gloas) metrics
+            }
+            Ok(AvailabilityProcessingStatus::MissingComponents(slot, block_root)) => {
+                trace!(
```
jimmygchen: could make this debug

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
```diff
@@ -0,0 +1,355 @@
...
+                        .await
+                }
+                ExecutedEnvelope::AvailabilityPending() => Err(EnvelopeError::InternalError(
+                    "Pending payload envelope not yet implemented".to_owned(),
+                )),
+            }
+        };
+
+        // Verify and import the payload envelope.
+        match import_envelope.await {
```
jimmygchen, line 108: We have another match here on status here:
https://github.com/sigp/lighthouse/blob/65b94da1e6d02179ca7113382044877d03639935/beacon_node/network/src/network_beacon_processor/gossip_methods.rs#L3398

do we need both? or can we just do the logging/metric in one place

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
```diff
@@ -0,0 +1,355 @@
...
+        _payload_verification_status: PayloadVerificationStatus,
+        parent_block: Arc<SignedBeaconBlock<T::EthSpec>>,
+    ) -> Result<Hash256, EnvelopeError> {
+        // Everything in this initial section is on the hot path for processing the envelope.
+
+        let post_exec_timer =
+            metrics::start_timer(&metrics::ENVELOPE_PROCESSING_POST_EXEC_PROCESSING);
+
+        // Check the payloads parent block against weak subjectivity checkpoint.
+        self.check_block_against_weak_subjectivity_checkpoint(
```
jimmygchen: do we need this?

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
```diff
@@ -0,0 +1,355 @@
...
+            block_root,
+            &state,
+        )?;
+
+        // Take an upgradable read lock on fork choice so we can check if this block has already
+        // been imported. We don't want to repeat work importing a block that is already imported.
+        let fork_choice_reader = self.canonical_head.fork_choice_upgradable_read_lock();
+        if !fork_choice_reader.contains_block(&block_root) {
+            return Err(EnvelopeError::BlockRootUnknown { block_root });
+        }
```
jimmygchen, line 236: This should not be reachable for a gossip verified envelope

jimmygchen: We could potentially do a defensive check here to see if the payload is already in fork choice

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
```diff
@@ -0,0 +1,355 @@
...
+        block_root: Hash256,
+        envelope_slot: Slot,
+        envelope_time_imported: Duration,
+    ) {
+        let envelope_delay_total =
+            get_slot_delay_ms(envelope_time_imported, envelope_slot, &self.slot_clock);
+
+        // Do not write to the cache for envelopes older than 2 epochs, this helps reduce writes
+        // to the cache during sync.
+        if envelope_delay_total < self.slot_clock.slot_duration().saturating_mul(64) {
```
jimmygchen: maybe extract 64 to a constant and have some docs (2 epochs)

beacon_node/beacon_chain/src/payload_envelope_verification/mod.rs
```diff
@@ -0,0 +1,313 @@
...
+    SlotMismatch { block: Slot, envelope: Slot },
+    /// The validator index is unknown
+    UnknownValidator { builder_index: u64 },
+    /// The block hash doesn't match the committed bid
+    BlockHashMismatch {
+        committed_bid: ExecutionBlockHash,
+        envelope: ExecutionBlockHash,
+    },
+    /// The block's proposer_index does not match the locally computed proposer
+    IncorrectBlockProposer { block: u64, local_shuffling: u64 },
```
jimmygchen: rename `block` to `proposer_index`

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -6466,62 +6455,14 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
...
-            );
-
-            Ok::<_, E>(EpochBlockProposers::new(proposal_epoch, fork, proposers))
-        })?;
-
-        // Run the accessor function on the computed epoch proposers.
-        accessor(epoch_block_proposers).map_err(Into::into)
+        crate::beacon_proposer_cache::with_proposer_cache(
+            &self.beacon_proposer_cache,
+            &self.spec,
```
jimmygchen: move spec to the last param

beacon_node/beacon_chain/src/metrics.rs
```diff
@@ -21,6 +21,44 @@ pub const VALIDATOR_MONITOR_ATTESTATION_SIMULATOR_SOURCE_ATTESTER_HIT_TOTAL: &st
...
+        "Full runtime of payload envelope processing",
+    )
+});
+pub static ENVELOPE_PROCESSING_DB_WRITE: LazyLock<Result<Histogram>> = LazyLock::new(|| {
+    try_create_histogram(
+        "payload_envelope_processing_db_write_seconds",
+        "Time spent writing a newly processed payload envelope and state to DB",
+    )
+});
+pub static ENVELOPE_PROCESSING_POST_EXEC_PROCESSING: LazyLock<Result<Histogram>> =
```
jimmygchen: we probably dont need this for payload as we probably aren't doing much before making the payload attestable in `import`

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
eserilev: [b760791](https://github.com/sigp/lighthouse/pull/8806/commits/b760791c307ddadba6d5022fd7fe3bc209651494)

eserilev: [d3a07e0](https://github.com/sigp/lighthouse/pull/8806/commits/d3a07e00f098443d56ae743a1f6269f71106723c)

beacon_node/beacon_chain/src/beacon_chain.rs
eserilev: moved to last param

beacon_node/beacon_chain/src/metrics.rs
eserilev: removed

beacon_node/beacon_chain/src/payload_envelope_verification/mod.rs
eserilev: updated

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
eserilev: done

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
eserilev: added TODO

beacon_node/beacon_chain/src/payload_envelope_verification/import.rs
eserilev: removed

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
eserilev: removed the duplicate log

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
eserilev: updated to proposer index

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
eserilev: added span to `process_gossip_execution_payload_envelope`

beacon_node/network/src/network_beacon_processor/gossip_methods.rs
eserilev: done

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
```diff
@@ -0,0 +1,447 @@
...
+
+        // Verify the envelope signature.
+        //
+        // For self-built envelopes, we can use the proposer cache for the fork and the
+        // validator pubkey cache for the proposer's pubkey, avoiding a state load from disk.
+        // For external builder envelopes, we must load the state to access the builder registry.
+        let builder_index = envelope.builder_index;
+        let block_slot = envelope.slot;
+        let envelope_epoch = block_slot.epoch(T::EthSpec::slots_per_epoch());
+        // TODO(gloas) verify correctness of this, or it might be useful to introduce a clearer function here.
```
jimmygchen: Had a brief conversation with michael today, and he thinks we'd have to load the parent block instead - using `proto_block.parent_root`, which is an option: `None` if the block is the finalized block, and it's a reachable scenario that needs to be handled.

jimmygchen: We also talked about tightening up the terminology:
* Parent block: is the block of the previous non-skip slot
* Payload block: is the block of the payload, same slot

eserilev: thanks, ill get to those changes tomorrow

and yeah the terminology makes sense to me.

eserilev: After talking to jimmy, we realized that in this code path the block is already guaranteed to have been imported into fork choice. that means the `proto_block.current_epoch_shuffling_id` already carries the correct `shuffling_decision_block` for this epoch

eserilev: [8230f42](https://github.com/sigp/lighthouse/pull/8806/commits/8230f424ea3b7a62a2224f160b96cc8c30caa0ec)

beacon_node/beacon_chain/src/execution_payload.rs
```diff
@@ -109,12 +109,19 @@ impl<T: BeaconChainTypes> PayloadNotifier<T> {
         if let Some(precomputed_status) = self.payload_verification_status {
             Ok(precomputed_status)
         } else {
-            notify_new_payload(&self.chain, self.block.message()).await
+            notify_new_payload(
+                &self.chain,
+                self.block.message().tree_hash_root(),
```
michaelsproul: This is a performance regression for the pre-Gloas codepath. I think we may as well just revert the refactoring changes in this file.

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
michaelsproul: How does the reprocessing queue help? Do you mean we send the envelope to the reprocess queue, trigger a single block lookup for its block, process the block, see that it's invalid, and _then_ send a reject for the envelope? I don't think we have the machinery for this? If the block is invalid and we've rejected it, currently we have no memory of it, right?

I think this is a good case for the block status table tbh, it could be much simpler to reason about: "have we seen this block? yep, and it was invalid -> reject".

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
```diff
@@ -0,0 +1,449 @@
...
+
+        let latest_finalized_slot = ctx
+            .canonical_head
+            .cached_head()
+            .finalized_checkpoint()
+            .epoch
+            .start_slot(T::EthSpec::slots_per_epoch());
+
+        // TODO(EIP-7732): check that we haven't seen another valid `SignedExecutionPayloadEnvelope`
+        //                 for this block root from this builder - envelope status table check
```
michaelsproul, line 122: Could be one of the `Observed*` caches, or yeah, I guess we could do this in block status.

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
```diff
@@ -0,0 +1,449 @@
...
+        };
+
+        drop(fork_choice_read_lock);
+
+        let latest_finalized_slot = ctx
+            .canonical_head
+            .cached_head()
+            .finalized_checkpoint()
+            .epoch
+            .start_slot(T::EthSpec::slots_per_epoch());
```
michaelsproul, line 119: Could consider reading this from fork choice to reduce the dependency surface further (would just need fork choice, not all of `CanonicalHead`). Problem is that `fork_choice` is private within canonical head.

beacon_node/beacon_chain/src/execution_payload.rs
jimmygchen: https://github.com/sigp/lighthouse/pull/8806/commits/7029aa18e002e4ed49b5e487d93ef4e1bf3ff6a8

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
michaelsproul: Yep, checked that the proposer cache uses `let fork = spec.fork_at_epoch(proposal_epoch);`. The proposer cache is now "safe by design"

beacon_node/beacon_chain/src/payload_envelope_verification/execution_pending_envelope.rs
```diff
@@ -0,0 +1,114 @@
...
+        notify_execution_layer: NotifyExecutionLayer,
+    ) -> Result<ExecutionPendingEnvelope<T::EthSpec>, EnvelopeError> {
+        let signed_envelope = self.signed_envelope;
+        let envelope = &signed_envelope.message;
+        let payload = &envelope.payload;
+
+        // TODO(gloas)
+        // We might want to make some cursory checks to see if it's worth
+        // validating the payload against the execution engine
+        // e.g `check_block_relevancy(block.as_block(), block_root, chain)?;`
```
jimmygchen: I think the fork choice check may be sufficient

beacon_node/beacon_chain/src/payload_envelope_verification/mod.rs
```diff
@@ -0,0 +1,285 @@
+//! The incremental processing steps (e.g., signatures verified but not the state transition) is
+//! represented as a sequence of wrapper-types around the block. There is a linear progression of
```
michaelsproul:
```suggestion
//! represented as a sequence of wrapper-types around the envelope. There is a linear progression of
```

beacon_node/beacon_chain/src/payload_envelope_verification/payload_notifier.rs
michaelsproul: It does, we just need to make sure the block hash calculations are working.

michaelsproul (approved):
Looks good! Keen to get this in

jimmygchen (approved):
LGTM!

beacon_node/beacon_chain/src/payload_envelope_verification/gossip_verified_envelope.rs
jimmygchen: https://github.com/sigp/lighthouse/issues/8948

eserilev:
Thank you so much for all the help ❤️

PR #8815, opened 2026-02-13
Implement proposer lookahead endpoint

### LINKED ISSUE #8809: Implement `/eth/v1/beacon/states/{state_id}/proposer_lookahead` endpoint
The new endpoint `/eth/v1/beacon/states/{state_id}/proposer_lookahead` is currently missing in Lighthouse. This is introduced in Fulu upgrade: https://github.com/ethereum/beacon-APIs/pull/565

romeoscript:
## Issue Addressed

closes #8809

## Proposed Changes

Implement GET /eth/v1/beacon/states/{state_id}/proposer_lookahead ([beacon-APIs#565](https://github.com/ethereum/beacon-APIs/pull/565)). Returns the proposer lookahead from Fulu+ states; 400 for pre-Fulu. Includes integration test.

## Additional Info
Follows the existing pending_consolidations endpoint pattern.

### DIFF
--- a/beacon_node/http_api/src/beacon/states.rs
+++ b/beacon_node/http_api/src/beacon/states.rs
@@ -8,7 +8,7 @@ use crate::version::{
 };
 use beacon_chain::{BeaconChain, BeaconChainError, BeaconChainTypes, WhenSlotSkipped};
 use eth2::types::{
-    ValidatorBalancesRequestBody, ValidatorId, ValidatorIdentitiesRequestBody,
+    self as api_types, ValidatorBalancesRequestBody, ValidatorId, ValidatorIdentitiesRequestBody,
     ValidatorsRequestBody,
 };
 use ssz::Encode;
@@ -171,12 +171,12 @@ pub fn get_beacon_state_proposer_lookahead<T: BeaconChainTypes>(
         .clone()
         .and(warp::path("proposer_lookahead"))
         .and(warp::path::end())
-        .and(warp::header::optional::<eth2::types::Accept>("accept"))
+        .and(warp::header::optional::<api_types::Accept>("accept"))
         .then(
             |state_id: StateId,
              task_spawner: TaskSpawner<T::EthSpec>,
              chain: Arc<BeaconChain<T>>,
-             accept_header: Option<eth2::types::Accept>| {
+             accept_header: Option<api_types::Accept>| {
                 task_spawner.blocking_response_task(Priority::P1, move || {
                     let (data, execution_optimistic, finalized, fork_name) = state_id
                         .map_state_and_execution_optimistic_and_finalized(
@@ -199,7 +199,7 @@ pub fn get_beacon_state_proposer_lookahead<T: BeaconChainTypes>(
                         )?;
 
                     match accept_header {
-                        Some(eth2::types::Accept::Ssz) => Response::builder()
+                        Some(api_types::Accept::Ssz) => Response::builder()
                             .status(200)
                             .body(data.as_ssz_bytes().into())
                             .map(|res: Response<Body>| add_ssz_content_type_header(res))
--- a/beacon_node/http_api/src/beacon/states.rs
+++ b/beacon_node/http_api/src/beacon/states.rs
@@ -71,67 +71,6 @@ pub fn get_beacon_state_pending_consolidations<T: BeaconChainTypes>(
         .boxed()
 }
 
-// GET beacon/states/{state_id}/proposer_lookahead
-pub fn get_beacon_state_proposer_lookahead<T: BeaconChainTypes>(
-    beacon_states_path: BeaconStatesPath<T>,
-) -> ResponseFilter {
-    beacon_states_path
-        .clone()
-        .and(warp::path("proposer_lookahead"))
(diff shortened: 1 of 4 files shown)


beacon_node/http_api/src/beacon/states.rs
```diff
@@ -69,6 +69,52 @@ pub fn get_beacon_state_pending_consolidations<T: BeaconChainTypes>(
...
+                            },
+                        )?;
+
+                    execution_optimistic_finalized_beacon_response(
+                        ResponseIncludesVersion::Yes(fork_name),
+                        execution_optimistic,
+                        finalized,
+                        data,
+                    )
+                    .map(|res| warp::reply::json(&res).into_response())
```
jimmygchen, line 112: This endpoint also need to support SSZ responses

https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Beacon/getProposerLookahead

e.g:
* https://github.com/sigp/lighthouse/blob/fc552a6c34231400f81e24aca56e4d9c4f45714a/beacon_node/http_api/src/lib.rs#L1551-L1552

I noticed that a few of the existing endpoints (including `pending_consolidations`) aren't fully spec compliant either, we could fix those in a separate PR.

jimmygchen (dismissed):
Looks good! Just a small comment on missing SSZ support.

beacon_node/http_api/src/beacon/states.rs
romeoscript: i've added the SSZ response , also about the `pending_consolidations` can you create an Issue for it ? so i'd work on it

beacon_node/http_api/src/beacon/states.rs
```diff
@@ -69,6 +72,66 @@ pub fn get_beacon_state_pending_consolidations<T: BeaconChainTypes>(
...
+             chain: Arc<BeaconChain<T>>,
+             accept_header: Option<eth2::types::Accept>| {
+                task_spawner.blocking_response_task(Priority::P1, move || {
+                    let (data, execution_optimistic, finalized, fork_name) = state_id
+                        .map_state_and_execution_optimistic_and_finalized(
+                            &chain,
+                            |state, execution_optimistic, finalized| {
+                                let Ok(lookahead) = state.proposer_lookahead() else {
+                                    return Err(warp_utils::reject::custom_bad_request(
+                                        "Proposer lookahead not available".to_string(),
```
dapplion: state.proposer_lookahead only returns error when the state is pre-Fulu, the error can say that to be more informative

common/eth2/src/lib.rs
```diff
@@ -898,6 +898,27 @@ impl BeaconNodeHttpClient {
...
+            .map_err(|()| Error::InvalidUrl(self.server.clone()))?
+            .push("beacon")
+            .push("states")
+            .push(&state_id.to_string())
+            .push("proposer_lookahead");
+
+        self.get_fork_contextual(path, |fork| fork)
+            .await
+            .map(|opt| opt.map(BeaconResponse::ForkVersioned))
+    }
```
dapplion, line 920: Can you add a test too for the SSZ variant?

romeoscript: yh , done @dapplion

beacon_node/http_api/src/beacon/states.rs
```diff
@@ -68,6 +71,67 @@ pub fn get_beacon_state_pending_consolidations<T: BeaconChainTypes>(
...
+                        )
+                        .map(|res| warp::reply::json(&res).into_response()),
+                    }
+                    .map(|resp| add_consensus_version_header(resp, fork_name))
+                })
+            },
+        )
+        .boxed()
+}
+
```
chong-he, line 119: This is a nit pick. I just thought it would be nice to not have this code in between the `pending_*` endpoints. 

I thought it would be cleaner to have the 3 `pending_* (consolidations, deposits, partial_withdrawals) ` together as they are introduced at the same time pretty much, so it would be nice to move this code below them.

chong-he (commented):
Looks pretty good to me. Just one minor comment below

I would say we can combine the test `test_beacon_states_proposer_lookahead_ssz` into the JSON one, but I see that there isn't much boilerplate to reduce, so don't worry about it.

chong-he (approved):
Looks good to me. Thanks for the PR

chong-he:
Simplifying the import of `api_types` in [a78aa7a](https://github.com/sigp/lighthouse/pull/8815/commits/a78aa7ac96745a90035647d05877e8fabebbe510) so that it's consistent with the current code base, and also got this idea when reviewing #8907. This will be good to reduce the merge conflict later on as well

beacon_node/http_api/src/beacon/states.rs
chong-he: Resolving this as the SSZ response has been added, I tested it as well and can see the endpoint's response

beacon_node/http_api/src/beacon/states.rs
chong-he: I believe the current error `Proposer lookahead is not available for pre-Fulu states` is clear enough

PR #8819, opened 2026-02-14
Fix eth2 compilation by feature gating `SseEventSource`

macladson:
## Issue Addressed

`eth2` is currently unable to be built without the `events` feature.

## Proposed Changes

Feature gates the `SseEventSource` match arm in the `status` function.

## Additional Info

We could consider adding some build checks to CI for `eth2` and `types` and each feature combination to avoid this in the future.

### DIFF
--- a/common/eth2/src/error.rs
+++ b/common/eth2/src/error.rs
@@ -102,6 +102,7 @@ impl Error {
                     None
                 }
             }
+            #[cfg(feature = "events")]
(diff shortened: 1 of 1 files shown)


michaelsproul (approved):
Looks good, confirmed locally that this compiles without the feature flag. 

But I agree we need to test these less-used features on CI if we want to support them. Doing this without bloating CI runtimes proportional to the number of features might prove tricky.

PR #8827, opened 2026-02-16
Gloas payload attestation consensus

michaelsproul:
## Proposed Changes

- Implement `process_payload_attestation`
- Implement EF tests for payload attestations (allows simplification of handler now that we support all `operations` tests).
- Update the `BlockSignatureVerifier` to signature-verify payload attestations

## Additional Info

There are some messy commits leftover from merging this branch, can be ignored:

- https://github.com/sigp/lighthouse/pull/8824


consensus/state_processing/src/per_block_processing/is_valid_indexed_payload_attestation.rs
```diff
@@ -0,0 +1,32 @@
...
+pub fn is_valid_indexed_payload_attestation<E: EthSpec>(
+    state: &BeaconState<E>,
+    indexed_payload_attestation: &IndexedPayloadAttestation<E>,
+    verify_signatures: VerifySignatures,
+    spec: &ChainSpec,
+) -> Result<(), BlockOperationError<Invalid>> {
+    // Verify indices are non-empty and sorted (duplicates allowed)
+    let indices = &indexed_payload_attestation.attesting_indices;
+    verify!(!indices.is_empty(), Invalid::IndicesEmpty);
+    verify!(indices.is_sorted(), Invalid::BadValidatorIndicesOrdering);
```
michaelsproul, line 15: Using `is_sorted` rather than the handwritten thing we use for attestations. The only loss is that we no longer get the index in the error. But that's fairly niche for the amount of extra complexity, so IMO this is better.

We can do the same for regular attestations in a different PR.

jimmygchen (approved):
Nice, very clean and matches the spec 👍

PR #8829, opened 2026-02-16
Remove `reqwest` re-exports from `eth2`

macladson:
## Proposed Changes

Remove `reqwest` from being re-exported within `eth2`

## Additional Info

This was largely unnecessary and just bloats the export space for downstream users.


common/eth2/src/lib.rs
```diff
@@ -22,8 +22,6 @@ pub use beacon_response::{
 };
 
 pub use self::error::{Error, ok_or_error, success_or_error};
-pub use reqwest;
-pub use reqwest::{StatusCode, Url};
 pub use sensitive_url::SensitiveUrl;
```
macladson, line 25: I am leaving the `SensitiveUrl` re-export as this is a tiny crate with only one public type and is necessary to build a `BeaconNodeHttpClient`

jxs (approved):
LGTM Mac

PR #8838, opened 2026-02-17
Head monitor: per-stream recovery and restart on candidate update

### LINKED ISSUE #8741: Head monitor service resilience improvements
## Description

Follow-up improvements for the beacon head monitor service introduced in #7892. The current implementation is a solid MVP, but several enhancements were identified during review that would improve resilience and robustness.

### 1. Per-stream error handling and retry

Currently, if a beacon node is restarted after the VC starts, the event stream for that node is never re-established until the VC itself restarts. The service only restarts when *all* streams fail.

The current error handling returns from the entire function on any stream error:

https://github.com/sigp/lighthouse/blob/0e9dab3878c8f594ac4e3b0094b71514424b3e54/validator_client/beacon_node_fallback/src/beacon_head_monitor.rs#L173-L178

A more resilient approach would handle errors per-stream independently and retry failed streams, so a flaky BN doesn't disrupt monitoring of healthy BNs.

### 2. Handle dynamic beacon node list updates

When `update_candidates_list` is called via API with a new BN list, stale SSE connections are maintained and the index mapping can become outdated. For example, if connected to nodes `[A, B, C]` and the list changes to `[B, C, D]`, head events from `B` would still report index 1 (its original position) even though `B` is now at index 0 in the candidate list.

Options discussed:
- Trigger service restart when candidate list changes
- Key cache on endpoint/ID rather than index

The cache purge mechanism exists but only runs on service restart:

https://github.com/sigp/lighthouse/blob/0e9dab3878c8f594ac4e3b0094b71514424b3e54/validator_client/beacon_node_fallback/src/beacon_head_monitor.rs#L97-L98

### 3. Unit tests

Add unit tests for:
- `BeaconHeadCache` methods
- `first_success_or_fallback` function (similar to existing `first_success_should_try_nodes_in_order` test)

## Additional Info

- Discussion: #7892
- Related to parallel BN handling: #7908

jimmygchen:
## Description

Adds per-stream recovery to the head monitor. Previously, when a beacon node's SSE stream disconnected, `select_all` would silently drop it with no way to reconnect — that BN was lost for the lifetime of the monitor. Now disconnected streams are tracked and retried once per slot, giving up only after 5 consecutive cycles with all BNs unreachable.

Also adds a `Notify`-based restart when `update_candidates_list` is called, so stale SSE connections are replaced with ones targeting the updated candidate list.

```
Before:  BN1 ───────────────────►       After:  BN1 ───────────────────►
         BN2 ─x (lost forever)                  BN2 ─x (retry next slot)
         BN3 ───────────────────►               BN3 ───────────────────►
```

New metrics: `vc_head_monitor_stream_disconnections_total`, `vc_head_monitor_stream_reconnections_total`, `vc_head_monitor_restarts_total`.

Closes #8741

## Additional Info

Cached heads are removed on disconnect to keep `is_latest` accurate. The `SelectAll` poll is gated on `!is_empty()` to avoid busy-looping when all streams have ended.

### DIFF
--- a/validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
+++ b/validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
@@ -9,8 +9,8 @@ use tokio::sync::RwLock;
 use tracing::{debug, info, warn};
 use types::EthSpec;
 use validator_metrics::{
-    HEAD_MONITOR_RESTARTS, HEAD_MONITOR_STREAM_DISCONNECTIONS,
-    HEAD_MONITOR_STREAM_RECONNECTIONS, inc_counter,
+    HEAD_MONITOR_RESTARTS, HEAD_MONITOR_STREAM_DISCONNECTIONS, HEAD_MONITOR_STREAM_RECONNECTIONS,
+    inc_counter,
 };
 
 type CacheHashMap = HashMap<usize, SseHead>;
--- a/validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
+++ b/validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
@@ -1,8 +1,9 @@
-use crate::BeaconNodeFallback;
+use crate::{BeaconNodeFallback, CandidateBeaconNode};
 use eth2::types::{EventKind, EventTopic, Hash256, SseHead};
 use futures::StreamExt;
+use futures::stream::BoxStream;
 use slot_clock::SlotClock;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use tokio::sync::RwLock;
 use tracing::{debug, info, warn};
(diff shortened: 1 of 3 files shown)


validator_client/beacon_node_fallback/src/lib.rs
```diff
@@ -544,6 +557,10 @@ impl<T: SlotClock> BeaconNodeFallback<T> {
             cache.purge_cache().await;
         }
```
macladson, line 558: We can probably remove this `purge_cache` since we are purging already during the notify.

validator_client/beacon_node_fallback/src/beacon_head_monitor.rs
```diff
@@ -90,112 +129,177 @@ pub async fn poll_head_event_from_beacon_nodes<E: EthSpec, T: SlotClock + 'stati
...
+                                }
+
+                                consecutive_all_failed = 0;
+                            }
+                            Ok(event) => {
+                                warn!(event_kind = event.topic_name(), candidate_index, "Unexpected event from BN");
+                            }
+                            Err(e) => {
+                                warn!(error = ?e, node_index = candidate_index, "Head event stream error");
+                            }
```
macladson, line 229: Both of these cases would be pretty unusual to get, do you think it might be worth restarting the stream in these cases? 

Although in a case where an SSE endpoint is sending malformed responses, it's not obvious if we should keep retrying the stream vs spamming warn logs.

The benefit of retrying is that we would only ever get 1 warn log per slot, rather than getting 1 warn per event.

validator_client/beacon_node_fallback/src/lib.rs
jimmygchen: Thanks for the review! My bad I haven't actually self review this, will do some time this week. But i think `update_candidates_list` can be called when the head monitor isn't running? In that case, only the purge in `update_candidates_list` would fire. Not sure though, I can confirm a bit later.

macladson: Oh sorry, I was a bit too eager :grin:

I believe if the head monitor isn't running, the head cache would be `None` anyway so the purge wouldn't fire in that case

jimmygchen:
I've just skimmed through this and I think this implementation is overcomplicated. Will make a new PR.

PR #8843, opened 2026-02-18
Fix duplicate data columns in DataColumnsByRange responses

### LINKED ISSUE #8842: Regression in `DataColumnsByRange` responses (duplicates included)
## Description

This line is wrong:

https://github.com/sigp/lighthouse/blob/2f7a1f3ae8b450f8042001a67ae88dd675251e3c/beacon_node/network/src/network_beacon_processor/rpc_methods.rs#L979-L980

Regression was introduced in:

- #8682

The method previously returned `Vec<Hash256>` which de-duplicated correctly. But now returns `Vec<(Hash256, Slot)>` which does not de-duplicate blocks correctly because the slots are unique. This results in duplicate columns being served to peers, and down-scoring by those peers (`DuplicatedData` errors).

## Version

This is a bug in v8.1.0 only.

jimmygchen:
## Description

Fix a regression introduced in #8682 where `get_block_roots_from_store` returns duplicate block roots for skip slots, causing duplicate data columns to be served in `DataColumnsByRange` responses. Peers receiving these duplicates downscore the node with `DuplicatedData` errors.

The return type was changed from `Vec<Hash256>` to `Vec<(Hash256, Slot)>`, but `.unique()` on tuples no longer deduplicates correctly because skip slots share the same block root but have different slots. Fix by using `.unique_by(|(root, _)| *root)`.

Closes #8842

### DIFF
--- a/beacon_node/network/src/network_beacon_processor/rpc_methods.rs
+++ b/beacon_node/network/src/network_beacon_processor/rpc_methods.rs
@@ -977,7 +977,10 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
         };
 
         // remove all skip slots i.e. duplicated roots
-        Ok(block_roots.into_iter().unique().collect::<Vec<_>>())
+        Ok(block_roots
+            .into_iter()
+            .unique_by(|(root, _)| *root)
+            .collect::<Vec<_>>())
     }
 
     /// Handle a `BlobsByRange` request from the peer.
--- a/beacon_node/network/src/network_beacon_processor/tests.rs
+++ b/beacon_node/network/src/network_beacon_processor/tests.rs
@@ -120,6 +120,42 @@ impl TestRig {
         .await
     }
 
+    pub async fn new_with_skip_slots(
+        chain_length: u64,
+        skip_slots: &HashSet<u64>,
+    ) -> Self {
+        let mut spec = test_spec::<E>();
+        spec.shard_committee_period = 2;
+        let spec = Arc::new(spec);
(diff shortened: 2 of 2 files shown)


beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -977,7 +977,10 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
         };
 
         // remove all skip slots i.e. duplicated roots
-        Ok(block_roots.into_iter().unique().collect::<Vec<_>>())
+        Ok(block_roots
+            .into_iter()
+            .unique_by(|(root, _)| *root)
+            .collect::<Vec<_>>())
```
michaelsproul, line 983: Can we add a test for this? From memory there are some tests in this file.

jimmygchen: yes absolutely! working on a test now

jimmygchen: https://github.com/sigp/lighthouse/pull/8843/commits/84a77be4460f1dfd4c0add78a6788e9a26e5c5e3

michaelsproul:
@jimmygchen Just needs a cargo fmt

jimmygchen:
Thanks @michaelsproul!

We really should merge this lol
* https://github.com/sigp/lighthouse/pull/8715

michaelsproul (approved):
Looks good. Test logic makes sense (request 1 column over a range of slots -> block roots should be unique).

PR #8846, opened 2026-02-18
Validator client head monitor timeout fix

michaelsproul:
## Issue Addressed

Fix a bug in v8.1.0 whereby the VC times out continuously with:

> Feb 18 02:03:48.030 WARN  Head service failed retrying starting next slot  error: "Head monitoring stream error, node: 0, error: SseClient(Transport(reqwest::Error { kind: Decode, source: reqwest::Error { kind: Body, source: TimedOut } }))"

## Proposed Changes

- Remove the existing timeout for the events API by using `Duration::MAX`. This is necessary as the client is configured with a default timeout. This is the only way to override/remove it.
- DO NOT add a `read_timeout` (yet), as this would need to be configured on a per-client basis. We do not want to create a new Client for every call as the early commits on this branch were doing, as this would bypass the TLS cert config, and is also wasteful.

## Additional Info

Awaiting confirmation that this fix works after manual testing. Will update.

### DIFF
--- a/common/eth2/src/lib.rs
+++ b/common/eth2/src/lib.rs
@@ -2808,7 +2808,7 @@ impl BeaconNodeHttpClient {
         let mut es = self
             .client
             .get(path)
-            .timeout(self.timeouts.events)
+            .read_timeout(self.timeouts.events)
             .eventsource()
             .map_err(Error::SseEventSource)?;
         // If we don't await `Event::Open` here, then the consumer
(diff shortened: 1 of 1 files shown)


michaelsproul:
Can confirm that manual testing on a VC shows that the timeout is gone (tested commit cc1d64d333b337e438a5b28aa80854c74a747af7).

jimmygchen (approved):
Nice!

michaelsproul:
Can also confirm that stopping one of the attached VCs results in:

>  Feb 18 04:22:09.149 WARN  Head service failed retrying starting next slot  error: "Head monitoring stream error, node: 0, error: SseClient(StreamEnded)"

Ari4ka (approved):

0xsarawut (commented):
https://keybase.io/sigp/pgp_keys.asc?fingerprint=15e66d941f697e28f49381f426416dc3f30674b0

PR #8847, opened 2026-02-18
Add sync batch state metrics

jimmygchen:
## Description

This PR contains metrics that compliments the fixes made in #8039.

Add `sync_batches` metric to track the number of batches in each state across all sync types. Labels are `sync_type` (`range_finalized`/`range_head`/`backfill`/`custody_backfill`) and `state` (`awaiting_download`, `downloading`, `awaiting_processing`, `processing`, `awaiting_validation`).

This gives production visibility into the batch buffer depth across range sync, backfill sync, and custody backfill sync. Reported every 5 seconds alongside existing network request metrics.

### DIFF
--- a/beacon_node/network/src/metrics.rs
+++ b/beacon_node/network/src/metrics.rs
@@ -462,6 +462,13 @@ pub static SYNCING_CHAIN_BATCH_AWAITING_PROCESSING: LazyLock<Result<Histogram>> 
             ]),
         )
     });
+pub static SYNCING_CHAIN_BATCHES: LazyLock<Result<IntGaugeVec>> = LazyLock::new(|| {
+    try_create_int_gauge_vec(
+        "sync_batches",
+        "Number of batches in sync chains by sync type and state",
+        &["sync_type", "state"],
+    )
+});
 pub static SYNC_SINGLE_BLOCK_LOOKUPS: LazyLock<Result<IntGauge>> = LazyLock::new(|| {
     try_create_int_gauge(
         "sync_single_block_lookups",
--- a/beacon_node/network/src/sync/backfill_sync/mod.rs
+++ b/beacon_node/network/src/sync/backfill_sync/mod.rs
@@ -8,6 +8,7 @@
 //! If a batch fails, the backfill sync cannot progress. In this scenario, we mark the backfill
 //! sync as failed, log an error and attempt to retry once a new peer joins the node.
 
+use crate::metrics;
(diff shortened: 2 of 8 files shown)


jimmygchen (commented):
States that drop to zero won't get their gauges reset — `set_gauge_vec` is only called for states present in `totals`, so a previous non-zero value lingers until something re-enters that state. Same issue at the chain_type level if all finalized (or head) chains get removed.

Could initialize all label combos to 0 before setting actuals, e.g.:

```rust
for chain_type_str in ["Finalized", "Head"] {
    for state in ["awaiting_download", "downloading", "awaiting_processing", "processing", "awaiting_validation"] {
        metrics::set_gauge_vec(&metrics::SYNCING_CHAIN_BATCHES, &[chain_type_str, state], 0);
    }
}
```

then do the current iteration to overwrite with real counts.

jimmygchen (commented):
States that drop to zero won't get their gauges reset — `set_gauge_vec` is only called for states present in `totals`, so a previous non-zero value lingers until something re-enters that state. Same issue at the chain_type level if all finalized (or head) chains get removed.

Could initialize all label combos to 0 before setting actuals, e.g.:

```rust
for chain_type_str in ["Finalized", "Head"] {
    for state in ["awaiting_download", "downloading", "awaiting_processing", "processing", "awaiting_validation"] {
        metrics::set_gauge_vec(&metrics::SYNCING_CHAIN_BATCHES, &[chain_type_str, state], 0);
    }
}
```

then do the current iteration to overwrite with real counts.

dapplion:
Suggestion: replace `&'static str` labels and `gauge.reset()` with a `BatchMetricsState` enum.

- Adds a `BatchMetricsState` enum with strum derives (`EnumIter`, `IntoStaticStr`) so we iterate all variants and set missing ones to `0` explicitly, removing the need for `gauge.reset()`.
- Adds a `Failed` variant covering both `Poisoned` and `Failed`, so `metrics_state()` returns the enum directly (no `Option`).

See commit: https://github.com/dapplion/lighthouse/commit/cf9af4dda

jimmygchen:
Thanks for the suggestion and commit\! Cherry-picked your enum approach and simplified a bit further on top — trimmed the derives down to just what's needed (`Debug, Clone, Copy, PartialEq, Eq, EnumIter, IntoStaticStr`) and replaced the BTreeMap accumulation with direct filter+count per state. Added you as co-author.

dapplion (approved):
LGTM!

PR #8761, opened 2026-02-05
Remove merge transition code

dapplion:
## Summary

Remove merge transition code that is no longer needed. The merge (Ethereum's transition from proof-of-work to proof-of-stake) occurred in September 2022. All related transition logic, readiness checks, and terminal difficulty validation can now be safely removed.

This PR represents the squashed and rebased version of the work at commit `ca6a1b66791e29b9ab47b6faa67424ede0281a06`.

## Changes

This PR removes approximately 2,100 net lines of code across 39 files, including:

- **Terminal difficulty detection**: Removed `get_terminal_pow_block_hash()`, `get_pow_block_at_total_difficulty()`, and related terminal block hash validation logic from execution layer
- **Bellatrix readiness checks**: Simplified `bellatrix_readiness.rs` from merge transition readiness validation to just genesis execution payload consistency checks  
- **Optimistic transition block verification**: Deleted entire `otb_verification_service.rs` (369 lines) and related optimistic import logic specific to the merge transition
- **Merge-specific notifications**: Removed terminal block found logging, merge readiness warnings, and transition status messages from notifier
- **Operation pool**: Cleaned up merge-specific attestation and block proposal logic that handled transition edge cases
- **Test infrastructure**: Removed merge transition test cases from `bellatrix.rs` (212 lines), `capella.rs` (156 lines), and other integration tests

## Files Changed

Key modifications:
- `beacon_node/beacon_chain/src/otb_verification_service.rs` - **deleted**
- `beacon_node/beacon_chain/src/bellatrix_readiness.rs` - simplified to genesis checks only
- `beacon_node/execution_layer/src/lib.rs` - removed terminal difficulty methods
- `beacon_node/client/src/notifier.rs` - removed merge readiness logging
- `beacon_node/operation_pool/src/lib.rs` - cleaned up transition-specific logic
- Various test files - removed obsolete merge transition tests

## Testing

- Existing test suite passes
- Removed tests were specific to merge transition scenarios that are no longer relevant
- Core functionality (block processing, attestations, execution layer integration) unchanged

PR #8856, opened 2026-02-18
Skip payload_invalidation tests prior to Bellatrix

michaelsproul:
## Issue Addressed

Fix the failure of the beacon-chain tests for phase0/altair, which now only runs nightly.

## Proposed Changes

Just skip the payload invalidation tests, they don't make any sense prior to Bellatrix anyway.

## Additional Info

This PR is redundant (and the nightly tests too) once we merge:

- https://github.com/sigp/lighthouse/pull/8761


dapplion (approved):
LGTM

PR #8859, opened 2026-02-18
Process head_chains in descending order of number of peers

pawanjay176:
## Issue Addressed

N/A

## Proposed Changes

Another find by @gitToki. Sort the preferred_ids in descending order as originally intended from the comment in the function.

### DIFF
--- a/beacon_node/network/src/sync/range_sync/chain_collection.rs
+++ b/beacon_node/network/src/sync/range_sync/chain_collection.rs
@@ -351,7 +351,8 @@ impl<T: BeaconChainTypes> ChainCollection<T> {
             .iter()
             .map(|(id, chain)| (chain.available_peers(), !chain.is_syncing(), *id))
             .collect::<Vec<_>>();
-        preferred_ids.sort_unstable();
+        // Sort in descending order
+        preferred_ids.sort_unstable_by(|a,b| b.cmp(a));
 
         let mut syncing_chains = SmallVec::<[Id; PARALLEL_HEAD_CHAINS]>::new();
         for (_, _, id) in preferred_ids {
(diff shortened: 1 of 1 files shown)


beacon_node/network/src/sync/range_sync/chain_collection.rs
```diff
@@ -351,7 +351,8 @@ impl<T: BeaconChainTypes> ChainCollection<T> {
             .iter()
             .map(|(id, chain)| (chain.available_peers(), !chain.is_syncing(), *id))
             .collect::<Vec<_>>();
-        preferred_ids.sort_unstable();
+        // Sort in descending order
+        preferred_ids.sort_unstable_by(|a, b| b.cmp(a));
```
michaelsproul, line 355: I'm satisfied that this works for the `!chain.is_syncing()` part too, as the not-syncing chains will have `true`, which sorts higher than `false` (syncing).

michaelsproul (approved):

beacon_node/network/src/sync/range_sync/chain_collection.rs
michaelsproul, line 359: Spicy little `expect` here 😱

michaelsproul: Probably fine, could change it to a `crit!` in some other PR

beacon_node/network/src/sync/range_sync/chain_collection.rs
pawanjay176: Yeah I agree. Based on the comment, I think the intention was always to construct `preferred_ids` such that the descending sort works, but we sorted ascending by mistake.

dapplion (approved):

PR #8860, opened 2026-02-18
Breakup RPCBlock into LookupBlock & RangeSyncBlock

ethDreamer:
This should make sync and the DA checker much more clear.


beacon_node/beacon_chain/src/test_utils.rs
```diff
@@ -2460,19 +2473,33 @@ where
...
+                .process_block(
+                    block_root,
+                    LookupBlock::new(block),
+                    NotifyExecutionLayer::Yes,
+                    BlockImportSource::Lookup,
+                    || Ok(()),
+                )
+                .await?
+                .try_into()
+                .expect("block blobs are available")
```
eserilev: nit: becaeuse this is a lookup block we dont expect blobs to be available right? the `expect` seems to be a bit misleading i think

beacon_node/beacon_chain/src/test_utils.rs
```diff
@@ -2460,40 +2473,54 @@ where
...
```
eserilev, line 2677: the expect message is a bit misleading since we are not expecting a block w/ blobs

beacon_node/beacon_chain/src/test_utils.rs
```diff
@@ -2427,20 +2427,33 @@ where
...
```
eserilev, line 2618: the expect message is a bit misleading since we are not expecting a block w/ blobs

beacon_node/beacon_chain/src/test_utils.rs
```diff
@@ -2525,62 +2552,45 @@ where
...
+                    block_data,
                     &self.chain.data_availability_checker,
                     self.chain.spec.clone(),
                 )?
             } else {
-                RpcBlock::new(
+                RangeSyncBlock::new(
                     block,
-                    None,
+                    AvailableBlockData::NoData,
```
eserilev, line 2848: I believe this should be a `LookupSync` block, since we were previously passing `None` as `BlockData` in `RpcBlock::new`

beacon_node/beacon_chain/src/test_utils.rs
```diff
@@ -2592,27 +2602,18 @@ where
...
-            }
+                AvailableBlockData::NoData
+            };
+
+            RangeSyncBlock::new(
+                block,
+                block_data,
+                &self.chain.data_availability_checker,
+                self.chain.spec.clone(),
+            )?
```
eserilev, line 2871: I dont think this keeps the previous test logic intact. I think that passing `None` to `BlockData` during `RpcBlock` construction should be equivalent to constructing a `LookupSync` block

eserilev (commented):
This looks great. Really makes things much clearer imo

I just had a few questions about a few test functions

beacon_node/beacon_chain/src/test_utils.rs
ethDreamer: The distinction between `RangeSyncBlock` and `LookupBlock` was handled by the `is_available` boolean (`true = RangeSyncBlock`, `false = LookupBlock`). I removed `is_available` as an input to this function so that it only handles the case where `is_available` is true (thus always returning a `RangeSyncBlock`). The caller of the function directly handles this by locally constructing a `LookupBlock` if `is_availlable` is `false`.

beacon_node/beacon_chain/src/test_utils.rs
ethDreamer: this only occurs when `is_available` is `false`, but `is_availble` is no longer passed to this function and the caller handles this case directly.

beacon_node/beacon_chain/src/test_utils.rs
ethDreamer: The error is correct because of the [`impl TryInto<SignedBeaconBlockHash> for AvailabilityProcessingStatus`](https://github.com/sigp/lighthouse/blob/v8.1.0/beacon_node/beacon_chain/src/beacon_chain.rs#L188). The error is triggered when `process_block` returns `AvailabiltyProcessingStatus::MissingComponents(..)`. But it could still be available if your test pre-loaded the components in the DA cache. Or you could be expecting it to fail. Either way, the behavior is the same as before.

beacon_node/beacon_chain/src/test_utils.rs
ethDreamer: See the comment on the other instance of this. This error message is accurate because it's triggered when `process_block()` returns `AvailabiltyProcessingStatus::MissingComponents(..)`. If you pre-load the DA cache with the components in this case, this test will still pass. Or the caller is expecting it to fail.

beacon_node/beacon_chain/src/test_utils.rs
eserilev: ohh gotcha, i missed that

eserilev (approved):
thanks for answering my questions, LGTM!

beacon_node/beacon_chain/src/block_verification_types.rs
```diff
@@ -13,76 +13,63 @@ use types::{
...
 
-/// A block that has been received over RPC. It has 2 internal variants:
-///
-/// 1. `FullyAvailable`: A fully available block. This can either be a pre-deneb block, a
-///    post-Deneb block with blobs, a post-Fulu block with the columns the node is required to custody,
-///    or a post-Deneb block that doesn't require blobs/columns. Hence, it is fully self contained w.r.t
-///    verification. i.e. this block has all the required data to get verified and imported into fork choice.
-///
-/// 2. `BlockOnly`: This is a post-deneb block that requires blobs to be considered fully available.
+pub struct LookupBlock<E: EthSpec> {
```
pawanjay176, line 19: Can we add docs to it specifying that its only a wrapper over a signed beacon block? Its useful when reviewing sync code

beacon_node/beacon_chain/src/block_verification_types.rs
```diff
@@ -13,76 +13,63 @@ use types::{
...
+    pub fn block_root(&self) -> Hash256 {
+        self.block_root
+    }
+
+    pub fn block_cloned(&self) -> Arc<SignedBeaconBlock<E>> {
+        self.block.clone()
+    }
+}
+
+/// A fully available block that has been constructed by range sync.
```
pawanjay176, line 43: A little more context that says it also should contain all blob data would be helpful

beacon_node/beacon_chain/src/test_utils.rs
```diff
@@ -2735,23 +2759,23 @@ where
         state_root
     }
 
-    /// Builds an `Rpc` block from a `SignedBeaconBlock` and blobs or data columns retrieved from
+    /// Builds an `RangeSyncBlock` from a `SignedBeaconBlock` and blobs or data columns retrieved from
```
pawanjay176:
```suggestion
    /// Builds a `RangeSyncBlock` from a `SignedBeaconBlock` and blobs or data columns retrieved from
```

beacon_node/beacon_chain/tests/column_verification.rs
```diff
@@ -80,16 +80,13 @@ async fn rpc_columns_with_invalid_header_signature() {
...
-        .unwrap();
     let availability = harness
         .chain
         .process_block(
             block_root,
-            rpc_block,
+            LookupBlock::new(signed_block.clone()),
             NotifyExecutionLayer::Yes,
-            BlockImportSource::RangeSync,
+            BlockImportSource::Lookup,
```
pawanjay176, line 89: was the previous code wrong here to classify the import source as range sync? Does look like it but just confirming

pawanjay176 (approved):
This is great. I just have minor nits. 
I initially thought that `RangeSyncBlock` should be just `SyncBlock` because block_lookup uses it too. But I just learnt that block lookups processes blocks and columns separately even for parent lookups. 

Would like to get a ✅ from @dapplion as well to make sure this is in line with his plans for gloas lookups as well

beacon_node/beacon_chain/tests/column_verification.rs
ethDreamer: Yeah this test is about verifying that columns with invalid header signatures are properly rejected during lookup operations, not during range sync operations. So the previous classification was wrong. One of the benefits of making it explicit.

beacon_node/beacon_chain/src/block_verification_types.rs
```diff
@@ -38,6 +41,10 @@ impl<E: EthSpec> LookupBlock<E> {
 }
 
 /// A fully available block that has been constructed by range sync.
+/// The block contains all the data required to import into fork choice.
+/// This includes any and all blobs/columns required, including zero if
+/// none are required. This could happend if the block is pre-deneb or
```
pawanjay176: This can happen

ethDreamer: damnit.. fixed

pawanjay176 (approved):
LGTM. I know that Lion is on board with the idea so gonna go ahead and merge so that we can build other sync PRs on this

PR #8864, opened 2026-02-19
Add --jwt-secret-path to lcli mock-el

jimmygchen:
## Description

Allow `lcli mock-el` to read a pre-shared JWT secret from a file via `--jwt-secret-path`, instead of always using the hardcoded `DEFAULT_JWT_SECRET`. This enables external tools (e.g. kurtosis ethereum-package) to provide their own JWT secret.

The two flags are mutually exclusive:

| Flag | Behavior |
|---|---|
| `--jwt-output-path` | Current behavior: use default secret, write to file |
| `--jwt-secret-path` | Read secret from file, don't write anywhere |

The new flag accepts a hex-encoded secret file (with or without `0x` prefix).

## Additional Info

No breaking changes — existing usage with only `--jwt-output-path` is unchanged.

### DIFF
--- a/lcli/src/mock_el.rs
+++ b/lcli/src/mock_el.rs
@@ -13,7 +13,8 @@ use std::sync::Arc;
 use types::*;
 
 pub fn run<E: EthSpec>(mut env: Environment<E>, matches: &ArgMatches) -> Result<(), String> {
-    let jwt_path: PathBuf = parse_required(matches, "jwt-output-path")?;
+    let jwt_output_path: Option<PathBuf> = parse_optional(matches, "jwt-output-path")?;
+    let jwt_secret_path: Option<PathBuf> = parse_optional(matches, "jwt-secret-path")?;
     let listen_addr: Ipv4Addr = parse_required(matches, "listen-address")?;
     let listen_port: u16 = parse_required(matches, "listen-port")?;
     let all_payloads_valid: bool = parse_required(matches, "all-payloads-valid")?;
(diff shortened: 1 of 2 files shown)


macladson (approved):
LGTM!

Just wondering if `jwt-secret-path` and `jwt-output-path` should be mutually exclusive? Seems strange to write a output file when you are already providing an input file. Might be slightly tidier UX?

pawanjay176:
Agree with Mac. I think the 2 flags should conflict. I don't see why we would want to read from a file and write to a different file either.

lcli/src/mock_el.rs
```diff
@@ -25,8 +26,23 @@ pub fn run<E: EthSpec>(mut env: Environment<E>, matches: &ArgMatches) -> Result<
...
-    std::fs::write(jwt_path, hex::encode(DEFAULT_JWT_SECRET)).unwrap();
+
+    let jwt_key = if let Some(secret_path) = jwt_secret_path {
+        let hex_str = std::fs::read_to_string(&secret_path)
+            .map_err(|e| format!("Failed to read JWT secret file: {}", e))?;
+        let secret_bytes = hex::decode(strip_prefix(hex_str.trim()))
+            .map_err(|e| format!("Invalid hex in JWT secret file: {}", e))?;
+        JwtKey::from_slice(&secret_bytes)
+            .map_err(|e| format!("Invalid JWT secret length (expected 32 bytes): {}", e))?
+    } else {
```
dapplion: Is it more clear to do here `else if let Some(jwt_path) = jwt_output_path {`
and then error in the final else?

jimmygchen: Done in e23b8d1.

macladson (approved):
Perfect!

PR #8865, opened 2026-02-19
Remove `psutil` and compute metrics from `procfs`

macladson:
## Issue Addressed

We can remove `psutil` as a dependency as it is infrequently maintained and unnecessary. We can simply compute the same information using a combination of `procfs` and `nix` (which is already a transitive dep)

## Proposed Changes

Compute health metrics using `procfs` instead of `psutil` and remove `psutil` as a dependency.


common/health_metrics/Cargo.toml
```diff
@@ -8,5 +8,8 @@ eth2 = { workspace = true, features = ["lighthouse"] }
 metrics = { workspace = true }
 
 [target.'cfg(target_os = "linux")'.dependencies]
+nix = { version = "0.26", default-features = false, features = ["fs"] }
```
macladson, line 11: We deliberately use an outdated version of `nix` to avoid a duplicate dependency. Once our dependencies update their version of `nix`, we can bump this too

common/health_metrics/src/observe.rs
```diff
@@ -96,32 +158,419 @@ impl Observe for ProcessHealth {
...
+        ours.abs_diff(theirs) <= tolerance_bytes
+    }
+
+    /// Compare the new procfs-based implementation against psutil to verify
+    /// we get equivalent values from the same underlying /proc data source.
+    ///
+    /// Note: Some values (memory, I/O counters) can change between when we read
+    /// our values and when psutil reads its values, so we allow small tolerances.
+    #[test]
+    fn compare_system_health_with_psutil() {
```
macladson: I got Claude to generate some comparison tests which verify that our `procfs`-based implementation matches the previous `psutil` version (within certain tolerances). This will hopefully mean our users don't see system metrics vary wildly when they update.

I just wanted these to ensure they can pass CI but we can remove them once we are sure the new implementation works fine. ~We can also keep them around, either is fine with me.~

macladson: Actually once we are confident this works, we should remove them. Because of the usage of tolerances, they could be flaky

common/health_metrics/Cargo.toml
```diff
@@ -8,5 +8,8 @@ eth2 = { workspace = true, features = ["lighthouse"] }
...
 procfs = { version = "0.18", default-features = false }
+
+[target.'cfg(target_os = "linux")'.dev-dependencies]
 psutil = "3.3.0"
```
dapplion: But this PR is not removing psutil?

macladson: Once we remove the consistency tests we can remove it. Also, even if we kept the tests it is now just a dev-dependency so it will never be compiled in any release build

macladson (commented):
Now that the tests verified that the returned data is within tolerance for both `procfs` and `psutil`, I have removed the tests

PR #8866, opened 2026-02-19
Bump sqlite deps to remove `hashlink 0.8`

macladson:
## Issue Addressed

#8547 

## Proposed Changes

Bump the following crates to remove `hashlink 0.8`:
- `rusqlite`
- `r2d2-sqlite`
- `yaml-rust2`

## Additional Info

Once we get new releases from `libp2p` and `discv5` we will unify on `hashlink 0.11`, which will also remove `hashlink` `0.9` and `0.10`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3427,9 +3427,9 @@ dependencies = [
 
 [[package]]
 name = "fallible-iterator"
-version = "0.2.0"
+version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


pawanjay176 (approved):
Looks good. Does libp2p/discv5 upstream already have the required hashlink dependency or do we need to update them as well?

macladson:
> Looks good. Does libp2p/discv5 upstream already have the required hashlink dependency or do we need to update them as well?

They have both been bumped on `main` already, so we are just awaiting a release

PR #8867, opened 2026-02-19
Feature-gate all uses of `arbitrary`

macladson:
## Proposed Changes

Feature gate all uses of `arbitrary` so it is not compiled during release builds.


pawanjay176 (approved):
Nice!

PR #8870, opened 2026-02-20
Add getBlobsV1 and getBlobsV2 support to mock EL server

jimmygchen:
## Description

Revive #7986 which was approved but closed after `network-tests` flaked repeatedly in the merge queue.

Adds `engine_getBlobsV1` and `engine_getBlobsV2` handlers to the mock EL server. When blobs have been generated during payload building (integration tests), they're returned by matching versioned hashes from the existing `blobs_bundles`. When no blobs exist (lcli paired with a real CL), empty results are returned — preventing the CL from logging "method not supported" errors.

Changes from #7986:
- `.to_vec().into()` → `.to_vec().try_into().ok()?` for V2 cell proofs (`VariableList` no longer implements `From<Vec<_>>`)
- Spec-compliant return types: V1 returns `Vec<Option<BlobAndProofV1>>` (per-element nullable, positionally matching the request), V2 returns `Option<Vec<BlobAndProofV2>>` (all-or-nothing, `null` if any blob is missing)

### DIFF
--- a/beacon_node/execution_layer/src/test_utils/execution_block_generator.rs
+++ b/beacon_node/execution_layer/src/test_utils/execution_block_generator.rs
@@ -484,34 +484,36 @@ impl<E: EthSpec> ExecutionBlockGenerator<E> {
 
     /// Look up a blob and proof by versioned hash across all stored bundles.
     pub fn get_blob_and_proof(&self, versioned_hash: &Hash256) -> Option<BlobAndProof<E>> {
-        self.blobs_bundles.values().find_map(|blobs_bundle| {
-            let (blob_idx, _) =
-                blobs_bundle
-                    .commitments
-                    .iter()
-                    .enumerate()
-                    .find(|(_, commitment)| {
-                        &kzg_commitment_to_versioned_hash(commitment) == versioned_hash
-                    })?;
-            let is_fulu_bundle = blobs_bundle.blobs.len() < blobs_bundle.proofs.len();
-            let blob = blobs_bundle.blobs.get(blob_idx)?.clone();
-            if is_fulu_bundle {
-                let start = blob_idx * E::cells_per_ext_blob();
-                let end = start + E::cells_per_ext_blob();
-                let proofs = blobs_bundle
-                    .proofs
-                    .get(start..end)?
-                    .to_vec()
-                    .try_into()
-                    .ok()?;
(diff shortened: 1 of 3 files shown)


beacon_node/execution_layer/src/test_utils/handle_rpc.rs
```diff
@@ -468,6 +468,39 @@ pub async fn handle_rpc<E: EthSpec>(
...
+                            return Err((
+                                format!("{} called after Fulu fork!", method),
+                                FORK_REQUEST_MISMATCH_ERROR_CODE,
+                            ));
+                        }
+                    }
+                }
+            }
+
+            Ok(serde_json::to_value(blobs_and_proofs).unwrap())
```
dapplion: Should this return `Vec<Option<...>>`? To set the missing blobs to null

jimmygchen: Good catch — fixed in the latest push. V1 now returns `Vec<Option<BlobAndProofV1>>` with nulls at positions of missing hashes, and V2 returns `Option<Vec<BlobAndProofV2>>` (null if any blob is missing), per the execution-apis spec.

beacon_node/execution_layer/src/test_utils/execution_block_generator.rs
```diff
@@ -480,6 +482,38 @@ impl<E: EthSpec> ExecutionBlockGenerator<E> {
...
+        self.blobs_bundles.values().find_map(|blobs_bundle| {
+            let (blob_idx, _) =
+                blobs_bundle
+                    .commitments
+                    .iter()
+                    .enumerate()
+                    .find(|(_, commitment)| {
+                        &kzg_commitment_to_versioned_hash(commitment) == versioned_hash
+                    })?;
+            let is_fulu_bundle = blobs_bundle.blobs.len() < blobs_bundle.proofs.len();
```
pawanjay176: can we not decide this based on the slot? Seems like a non-obvious condition. 
Maybe add a comment explaining this?

jimmygchen: Good catch, added in https://github.com/sigp/lighthouse/pull/8870/commits/22bd6a0dd2de2b5bc3edae0a282d82e88a68ac2d

pawanjay176 (approved):

PR #8871, opened 2026-02-20
Remove unreachable void pattern for ConnectionLimits

### LINKED ISSUE #6356: Remove unreachable void patterns once MSRV >= 1.82

jimmygchen:
## Description

Remove the `#[allow(unreachable_patterns)]` and unreachable `BehaviourEvent::ConnectionLimits` match arm. The MSRV is well above 1.82, so the compiler natively recognises that `ConnectionLimits` events are uninhabited (`Infallible`).

Closes #6356

### DIFF
--- a/beacon_node/lighthouse_network/src/service/mod.rs
+++ b/beacon_node/lighthouse_network/src/service/mod.rs
@@ -1861,8 +1861,6 @@ impl<E: EthSpec> Network<E> {
                     self.inject_upnp_event(e);
                     None
                 }
-                #[allow(unreachable_patterns)]
-                BehaviourEvent::ConnectionLimits(le) => libp2p::core::util::unreachable(le),
             },
             SwarmEvent::ConnectionEstablished { .. } => None,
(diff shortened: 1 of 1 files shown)


michaelsproul (approved):
Nice :fire:

PR #8874, opened 2026-02-20
Bump `uuid` to remove duplicate

macladson:
## Issue Addressed

#8547 

## Proposed Changes

Bump the version of `uuid` in our Cargo.toml to version `1` which removes `uuid 0.8` and unifies it across the workspace to version `1.19.0`.

## Additional Info

While this version of `uuid` isn't the latest, newer versions of `uuid` have bumped `getrandom` which would duplicate it for our workspace.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -18,6 +18,7 @@ deny = [
     { crate = "pbkdf2", deny-multiple-versions = true, reason = "takes a long time to compile" },
     { crate = "scrypt", deny-multiple-versions = true, reason = "takes a long time to compile" },
     { crate = "syn", deny-multiple-versions = true, reason = "takes a long time to compile" },
+    { crate = "uuid", deny-multiple-versions = true, reason = "dependency hygiene" },
 ]
 
 [sources]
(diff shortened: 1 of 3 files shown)


pawanjay176 (approved):

michaelsproul:
Could deny multiple versions in the `deny.toml`

macladson:
> Could deny multiple versions in the `deny.toml`

Ah yeah, I wasn't sure how to categorize it since it doesn't fit our usual "takes a long time to compile" criteria, but I think we can add a general "dependency hygiene" category for cases like this. Its also nice that it implies it is reasonable to remove the deny if there's a compelling reason.

deny.toml
```diff
@@ -18,6 +18,7 @@ deny = [
     { crate = "pbkdf2", deny-multiple-versions = true, reason = "takes a long time to compile" },
     { crate = "scrypt", deny-multiple-versions = true, reason = "takes a long time to compile" },
     { crate = "syn", deny-multiple-versions = true, reason = "takes a long time to compile" },
+    { crate = "uuid", deny-multiple-versions = true, reason = "dependency hygiene" },
```
macladson, line 21: Done here

michaelsproul:
Yeah I think that's good. Now cargo dent can flag the regression for us and we'll at least be made to consciously accept uuid duplication

michaelsproul (approved):

PR #8876, opened 2026-02-20
fix: repair Gloas block replay crash with IncorrectStateVariant

### LINKED ISSUE #8869: Update block replayer for Gloas
During local devnet-0 testing we came across the following error when querying gloas states 

```
{\"code\":500,\"message\":\"UNHANDLED_ERROR: DBError(LoadingHotStateError(\\\"get state\\\", 
0x1f4dea650b6e38cfe0bea431fa978c9643ec88ca3c6521444fdecff833c05383, '
BlockReplayError(BlockProcessing(ExecutionPayloadBidInvalid { reason: ParentBlockHashMismatch { state_block_hash: 
0x616de917c182c8d648ce460f88d25b1b9335d45d2b95ab70261c9968d31ad52c, bid_parent_hash: 
0x58ded490393a053c1b9388bb5d30de0845ebc5500d902fe5aeae562515589bd8 } }))))\",\"stacktraces\":[]}
```

In `BlockReplayer::apply_blocks` we call `per_block_processing` but never process the associated execution payload envelopes. 

For example

We replay the block at slot N. The bid is validated but `state.latest_block_hash` isn't updated. The execution payload envelope at slot N should have updated `state.latest_block_hash`. At slot N+1 we attempt to validate the bid, but `state.latest_block_hash` was never updated. So we get the above failure message.

I believe the correct solution is to call `process_execution_payload` during block replay. We could update `apply_blocks` to accept a `Vec` of blocks and payload envelopes. 

A potential easier, yet hacky approach could be to update `latest_block_hash` from the bid directly. This is simpler but skips other state changes from envelope processing (deposit requests, withdrawal requests, consolidation requests etc.). I'm not sure if the tradeoffs here are acceptable.

bit2swaz:
## Issue Addressed
Fixes #8869

## Proposed Changes
Gloas (EIP-7732) decouples the execution payload from the beacon block, the block body carries only a `signed_execution_payload_bid`, while the actual payload arrives separately as a `SignedExecutionPayloadEnvelope`. Two code paths in `per_block_processing` were calling pre-Gloas accessors that don't exist on Gloas block bodies, causing block replay to crash with `IncorrectStateVariant` and the HTTP API to return `500 UNHANDLED_ERROR` for any finalized
Gloas state.

**Bug 1: execution payload path** (`per_block_processing.rs`)
`is_execution_enabled()` returns `true` for Gloas (Capella is always enabled, making `is_merge_transition_complete` return true). This caused `body.execution_payload()?` to be called on a Gloas body, which returns `Err(IncorrectStateVariant)`.
Fix: guard the `process_withdrawals` / `process_execution_payload` block with `&& !gloas_enabled()`.

**Bug 2: Electra operations path** (`process_operations.rs`)
`electra_enabled()` is also `true` for Gloas, causing `block_body.execution_requests()?` to be called. That field only exists on Electra/Fulu bodies, so Gloas returned `Err(IncorrectStateVariant)`.
Fix: guard the execution-requests block with `&& !gloas_enabled()`.

**Envelope processing architecture**
Following the EIP-7732 spec design, envelopes are applied between blocks using a look-ahead:
after applying block N, if block N+1's `bid.parent_block_hash` matches `state.latest_execution_payload_bid.block_hash`, the envelope for block N is applied before continuing. This is wired through:
- `process_execution_payload_envelope()`: new function that validates the hash chain and updates `state.latest_block_hash`.
- `BlockReplayer`: new `payload_envelopes` map and look-ahead loop in `apply_blocks`.
- `HotColdDB`: new `DBColumn::ExecPayloadEnvelope`, `put/get_execution_payload_envelope` methods, and `replay_blocks` wired to load and pass envelopes for any Gloas blocks in the replay range.

**Regression test**
`gloas_block_replay_no_incorrect_state_variant`:  builds a minimal Gloas genesis, constructs a
valid blinded Gloas block at slot 1, and asserts `BlockReplayer::apply_blocks` succeeds.

## Additional Info
- The envelope processing in `process_execution_payload_envelope` is intentionally minimal for now (updates latest_block_hash`, validates parent hash chain). Withdrawals, execution requests, and other envelope contents are left as TODOs, consistent with the broader in-progress EIP-7732 implementation in this codebase.
- `put_execution_payload_envelope` is added but not yet called at block import time; that wiring belongs in the beacon chain's block processing pipeline and can be done as a follow-up once the full Gloas block import path is built out.
- All 39 existing `state_processing` unit tests pass. `cargo clippy -D warnings` and `cargo fmt` are clean on the changed packages.

### DIFF
--- a/beacon_node/store/src/hot_cold_store.rs
+++ b/beacon_node/store/src/hot_cold_store.rs
@@ -780,6 +780,42 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold> 
             .map(|payload| payload.is_some())
     }
 
+    /// Store a Gloas (EIP-7732) execution payload envelope in the hot database.
+    ///
+    /// The key is the execution `block_hash` committed to by the builder in the corresponding
+    /// `signed_execution_payload_bid`.  This allows efficient lookup during block replay using
+    /// only the information present in each block's bid header.
+    pub fn put_execution_payload_envelope(
+        &self,
+        exec_block_hash: &ExecutionBlockHash,
+        envelope: &SignedExecutionPayloadEnvelope<E>,
+        ops: &mut Vec<KeyValueStoreOp>,
+    ) {
+        ops.push(KeyValueStoreOp::PutKeyValue(
+            DBColumn::ExecPayloadEnvelope,
+            exec_block_hash.as_ssz_bytes(),
+            envelope.as_ssz_bytes(),
+        ));
+    }
+
+    /// Load a Gloas (EIP-7732) execution payload envelope by the execution block hash.
+    ///
+    /// Returns `None` when no envelope has been stored for the given hash (e.g. the builder did
+    /// not deliver the payload for that slot, or envelope storage is not yet implemented for the
+    /// current database schema).
+    pub fn get_execution_payload_envelope(
+        &self,
+        exec_block_hash: &ExecutionBlockHash,
+    ) -> Result<Option<SignedExecutionPayloadEnvelope<E>>, Error> {
+        let key = exec_block_hash.as_ssz_bytes();
+        match self.hot_db.get_bytes(DBColumn::ExecPayloadEnvelope, &key)? {
+            Some(bytes) => Ok(Some(SignedExecutionPayloadEnvelope::from_ssz_bytes(
+                &bytes,
+            )?)),
+            None => Ok(None),
+        }
+    }
+
     /// Get the sync committee branch for the given block root
     /// Note: we only persist sync committee branches for checkpoint slots
     pub fn get_sync_committee_branch(
@@ -2474,6 +2510,28 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold> 
 
         if let Some(pre_slot_hook) = pre_slot_hook {
             block_replayer = block_replayer.pre_slot_hook(pre_slot_hook);
+        }
+
+        // For Gloas (EIP-7732) replay: attempt to load the execution payload envelope for each
+        // block from the hot DB, keyed by the execution block hash that the builder committed to.
+        // If any block is a Gloas block, we build the envelope map and wire it into BlockReplayer
+        // so the look-ahead logic can apply envelopes between consecutive blocks.
+        let any_gloas = blocks
+            .iter()
+            .any(|b| b.fork_name(&self.spec).is_ok_and(|f| f.gloas_enabled()));
+        if any_gloas {
+            let mut envelopes = HashMap::new();
+            for block in &blocks {
+                if let Ok(ForkName::Gloas) = block.fork_name(&self.spec)
+                    && let BeaconBlockBodyRef::Gloas(body) = block.message().body()
+                {
+                    let bid_hash = body.signed_execution_payload_bid.message.block_hash;
+                    if let Some(envelope) = self.get_execution_payload_envelope(&bid_hash)? {
+                        envelopes.insert(bid_hash, envelope.message);
+                    }
+                }
+            }
+            block_replayer = block_replayer.payload_envelopes(envelopes);
         }
 
         block_replayer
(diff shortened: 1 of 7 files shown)


beacon_node/store/src/lib.rs
```diff
@@ -310,6 +310,12 @@ pub enum DBColumn {
     /// Execution payloads for blocks more recent than the finalized checkpoint.
     #[strum(serialize = "exp")]
     ExecPayload,
+    /// Execution payload envelopes for Gloas (EIP-7732) blocks in the hot database.
+    ///
+    /// - Key: 32-byte execution `block_hash` from the block's `signed_execution_payload_bid`.
+    /// - Value: SSZ-encoded `SignedExecutionPayloadEnvelope`.
+    #[strum(serialize = "ppe")]
+    ExecPayloadEnvelope,
```
eserilev, line 318: We already have a `PayloadEnvelope` column defined

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -780,6 +780,42 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
             .map(|payload| payload.is_some())
     }
 
+    /// Store a Gloas (EIP-7732) execution payload envelope in the hot database.
+    ///
+    /// The key is the execution `block_hash` committed to by the builder in the corresponding
+    /// `signed_execution_payload_bid`.  This allows efficient lookup during block replay using
+    /// only the information present in each block's bid header.
+    pub fn put_execution_payload_envelope(
```
eserilev, line 788: we already have a function `payload_envelope_as_kv_store_ops` that does this

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -780,6 +780,42 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
+            envelope.as_ssz_bytes(),
+        ));
+    }
+
+    /// Load a Gloas (EIP-7732) execution payload envelope by the execution block hash.
+    ///
+    /// Returns `None` when no envelope has been stored for the given hash (e.g. the builder did
+    /// not deliver the payload for that slot, or envelope storage is not yet implemented for the
+    /// current database schema).
+    pub fn get_execution_payload_envelope(
```
eserilev, line 806: we already have a function `get_payload_envelope` that does this. Note that we fetch execution payload envelopes by block root, not by execution block hash

eserilev (changes requested):
Looks like this PR duplicates some DB functionality that already exists from this PR
https://github.com/sigp/lighthouse/pull/8717/changes#top

We should just use the existing db operations. Also note that the payload envelopes are keyed by beacon block root, not execution block hash

eserilev:
I'm sorry, I'm going to go ahead and close this PR. This reads as fully LLM-generated which isn't something were interested in accepting. I'm not anti-LLM but we could produce this same output ourselves in a matter of minutes. Contributing should be a learning experience and I'm happy to support external contributors who want to dig in and genuinely engage with the code.

bit2swaz:
> I'm sorry, I'm going to go ahead and close this PR. This reads as fully LLM-generated which isn't something were interested in accepting. I'm not anti-LLM but we could produce this same output ourselves in a matter of minutes. Contributing should be a learning experience and I'm happy to support external contributors who want to dig in and genuinely engage with the code.

ah, fair point. i leaned way too heavily on an llm to try and blast through the boilerplate this weekend and completely missed the broader context of #8717 and the correct db keys. lesson learned.

i genuinely want to dig into the codebase and learn the EIP-7732 architecture properly. if youre still open to it, id love to scrap this branch, review the db operations you linked in #8717, and take a manual, ground up pass at wiring the look ahead loop the right way. if not, i completely understand keeping it closed.

sorry for any inconveniences i may have caused

eserilev:
@bit2swaz hey no worries, yes please feel free to continue working on this. Were going to update our contribution guidelines regarding AI usage, I think the guidelines here might give a general idea of what we think good contributions could look like https://github.com/sigp/lighthouse/pull/8879

PR #8878, opened 2026-02-20
Implement gloas block gossip verification changes

eserilev:
https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/p2p-interface.md#beacon_block


beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -1038,8 +1067,15 @@ impl<T: BeaconChainTypes> GossipVerifiedBlock<T> {
             });
         }
 
-        // Validate the block's execution_payload (if any).
-        validate_execution_payload_for_gossip(&parent_block, block.message(), chain)?;
+        // [New in Gloas]: Skip payload validation checks. The payload now arrives separetely
```
michaelsproul:
```suggestion
        // [New in Gloas]: Skip payload validation checks. The payload now arrives separately
```

michaelsproul (approved):
LGTM!

PR #8880, opened 2026-02-20
Refactor/stream vc vote publishing

shane-moore:
## Proposed Changes

Changes four `ValidatorStore` batch signing methods to return `impl Stream` instead of `Future`. Services consume the stream and publish each batch as it arrives.  No behavioral change for lh since `LighthouseValidatorStore` wraps everything in `stream::once`

Also replaces anonymous tuples in method signatures with named structs
  
## Motivation
Anchor needs to yield signing results per SSV committee. Each committee runs consensus independently, so streams let faster committees publish without waiting on slower ones.


validator_client/validator_store/src/lib.rs
```diff
@@ -156,21 +174,23 @@ pub trait ValidatorStore: Send + Sync {
...
+    /// Sign a batch of sync committee messages and return results as a stream of batches.
+    fn sign_sync_committee_signatures(
+        self: &Arc<Self>,
+        messages: Vec<SyncMessageToSign>,
+    ) -> impl Stream<Item = Result<Vec<SyncCommitteeMessage>, Error<Self::Error>>> + Send;
+
+    /// Sign a batch of sync committee contributions and return results as a stream of batches.
+    fn sign_sync_committee_contributions(
+        self: &Arc<Self>,
+        contributions: Vec<ContributionToSign<Self::E>>,
```
shane-moore, line 191: @dknopik @michaelsproul, made some helper structs like `ContributionToSign` to replace inline tuple structs like
```
(
            u64,
            PublicKeyBytes,
            SyncCommitteeContribution<Self::E>,
            SyncSelectionProof,
        )
```
wdyt?

validator_client/validator_services/src/attestation_service.rs
```diff
@@ -560,96 +560,114 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> AttestationService<S,
...
                 }
-            })
-            .collect::<Vec<_>>();
-        let single_attestations = &single_attestations;
-        let validator_indices = single_attestations
-            .iter()
-            .map(|att| att.attester_index)
-            .collect::<Vec<_>>();
-        let published_count = single_attestations.len();
+                Err(e) => {
```
shane-moore, line 656: I was originally doing `return Err('...')` in this variant after the log. I think an issue with that is that if we hit an error with one batch, then we'd early return before the rest could be processed

also, propagating the error up just logs an error, which we're doing with the `crit!` anyways, so seems fine to allow the stream to complete and hit the `Ok()`, since the logging happens regardless

open to thoughts from @dknopik and otherwise

chong-he (approved):
I see that the signing of attestations and sync committee messages are now wrapped in `stream::once`. 

This looks great when I go through the changes, mostly refactoring of the code

validator_client/validator_store/src/lib.rs
```diff
@@ -106,40 +139,25 @@ pub trait ValidatorStore: Send + Sync {
...
-    ///
-    /// * Vec of (validator_index, pubkey, validator_committee_index, attestation).
+    /// Returns a stream of batches of successfully signed attestations. Each batch contains
+    /// attestations that passed slashing protection, along with the validator index of the signer.
+    /// Eventually this will be replaced by `SingleAttestation` use.
     ///
     /// Output:
     ///
     /// * Vec of (validator_index, signed_attestation).
+    ///
```
macladson: Can remove this dangling comment

validator_client/validator_services/src/sync_committee_service.rs
```diff
@@ -247,78 +247,56 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> SyncCommitteeService<S
...
+                            count = committee_signatures.len()
+                        ))
+                        .await
+                        .map_err(|e| {
+                            error!(
+                                %slot,
+                                error = %e,
+                                "Unable to publish sync committee messages"
+                            );
+                        })?;
```
macladson: I think this `?` will cause the while loop to exit prematurely on the first failure. Rather than a `map_err(..)?` we should do a match like we do in `sign_and_publish_attestations`.

This is assuming we want to actually continue on any failure (per your comment)

validator_client/validator_services/src/sync_committee_service.rs
```diff
@@ -389,77 +367,60 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> SyncCommitteeService<S
...
+                            count = signed_contributions.len()
+                        ))
+                        .await
+                        .map_err(|e| {
+                            error!(
+                                %slot,
+                                error = %e,
+                                "Unable to publish signed contributions and proofs"
+                            );
+                        })?;
```
macladson: Same applies here

validator_client/validator_services/src/sync_committee_service.rs
```diff
@@ -247,78 +247,56 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> SyncCommitteeService<S
...
+            .collect();
+
+        let signature_stream = self
+            .validator_store
+            .sign_sync_committee_signatures(messages_to_sign);
+        tokio::pin!(signature_stream);
+
+        while let Some(result) = signature_stream.next().await {
+            match result {
+                Ok(committee_signatures) => {
```
macladson: This is probably an edge case but for consistency we could check if `committee_signatures` is empty like we do for attestations. Same for contributions

macladson (commented):
Looks good! This is a cool idea and I'm a fan of the helper structs. Just had a few comments

validator_client/validator_store/src/lib.rs
shane-moore: for sure! removed in https://github.com/sigp/lighthouse/pull/8880/changes/66ff63a8d8d9fd75c5a842477f7789b93a996ee0

validator_client/validator_services/src/sync_committee_service.rs
shane-moore: really appreciate this catch! we do want to continue on failures. updated in https://github.com/sigp/lighthouse/pull/8880/changes/66ff63a8d8d9fd75c5a842477f7789b93a996ee0

validator_client/validator_services/src/sync_committee_service.rs
shane-moore: also fixed in https://github.com/sigp/lighthouse/pull/8880/changes/66ff63a8d8d9fd75c5a842477f7789b93a996ee0

validator_client/validator_services/src/sync_committee_service.rs
shane-moore: agree with this. no need to publish if empty, updated in https://github.com/sigp/lighthouse/pull/8880/changes/66ff63a8d8d9fd75c5a842477f7789b93a996ee0

validator_client/validator_services/src/attestation_service.rs
```diff
@@ -560,96 +560,108 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> AttestationService<S,
...
-            ),
-            Err(e) => error!(
-                error = %e,
-                committee_index = attestation_data.index,
-                slot = slot.as_u64(),
-                "type" = "unaggregated",
-                "Unable to publish attestations"
-            ),
+        if !published_any {
+            warn!("No attestations were published");
```
macladson: This is the only one which has `published_any` and the associated warn. Do we want to extend this idea to the aggregates/sync_committees paths too?

macladson (approved):
This is looking good now! Just had one question but it's not a blocker

validator_client/validator_services/src/attestation_service.rs
shane-moore: I think attestation is slightly special since that path can end up empty after slashing protection / same data filtering, whereas aggregates + sync mostly just inherit already logged signer failures. I did rename `published_any` to `received_non_empty_batch` though since the old name was misleading

macladson: Ah yeah, makes sense :+1:

PR #8881, opened 2026-02-21
Fork aware `Attestation` ssz decoding

hopinheimer:
## Issue Addressed

@pk910 mentioned that we are not decoding based on the current fork and defaulting to latest fork. This is a nice to have change for testing purposes


consensus/types/src/attestation/attestation.rs
```diff
@@ -261,6 +261,17 @@ impl<E: EthSpec> Attestation<E> {
             Self::Electra(attn) => attn.get_aggregation_bits(),
         }
     }
+
+    pub fn from_ssz_bytes_for_fork(
```
michaelsproul, line 265: To fix the bug that was reported we probably also need to use this func in the network layer.

consensus/types/src/attestation/attestation.rs
```diff
@@ -582,6 +593,12 @@ impl<'de, E: EthSpec> ContextDeserialize<'de, ForkName> for Attestation<E> {
...
 }
 
+impl<E: EthSpec> ForkVersionDecode for Attestation<E> {
+    fn from_ssz_bytes_by_fork(bytes: &[u8], fork_name: ForkName) -> Result<Self, ssz::DecodeError> {
+        Self::from_ssz_bytes_for_fork(bytes, fork_name)
+    }
+}
+
 /*
 impl<'de, E: EthSpec> ContextDeserialize<'de, ForkName> for Vec<Attestation<E>> {
```
michaelsproul, line 603: This commented out code could be revived or deleted as part of this PR too.

michaelsproul (changes requested):

eserilev:
I think we can probably close this right? We've deprecated pretty much all use of the pre electra attestation variant

PR #8879, opened 2026-02-20
Update contribution guidlines regarding LLM usage

eserilev:
inspired by https://www.eff.org/about/opportunities/volunteer/coding-with-eff

PR #8884, opened 2026-02-23
feat(malloc_utils): add mimalloc and tcmalloc allocator feature flags

### LINKED ISSUE #8840: Test new allocators (mimalloc, tcmalloc)
## Description

Jemalloc [has been deprecated](https://jasone.github.io/2025/06/12/jemalloc-postmortem/). We could continue using it despite the deprecation, or start testing alternative allocators. I think the two main candidates are `mimalloc` and `tcmalloc`. We may as well test both.

To test we should run some local benchmarks (like `lcli transition-blocks`), and run some nodes. It would probably be trivial to add feature flags for mimalloc and/or tcmalloc.

0xMars42:
Closes #8840

Adds `mimalloc` and `tcmalloc` feature flags to `malloc_utils`, `lighthouse`, and `lcli` crates.

## Usage

```sh
cargo build --release -p lighthouse --features mimalloc
cargo build --release -p lighthouse --features tcmalloc
```

mimalloc/tcmalloc override jemalloc via cfg guards (same pattern as `sysmalloc`), so `--no-default-features` is not needed despite jemalloc being hardcoded in the binary deps.

Both features are unix-only enabling them on Windows produces a compile_error. Enabling both simultaneously also produces a compile_error.

## Benchmarks

`lcli transition-blocks` on a recent mainnet finalized block (10 runs, `--exclude-cache-builds --exclude-post-block-thc`):

| Allocator | Median | Min | Max |
|-----------|--------|-----|-----|
| jemalloc | 1001ms | 982ms | 1051ms |
| mimalloc | 1002ms | 970ms | 1171ms |
| tcmalloc | 1066ms | 1041ms | 1191ms |

jemalloc and mimalloc are effectively identical on block processing. tcmalloc is ~6% slower. These numbers only cover `transition-blocks` though live node testing with sustained load and varying allocation patterns could show different results.

## Changes

- `common/malloc_utils/Cargo.toml` add mimalloc + tcmalloc optional deps and features
- `common/malloc_utils/src/lib.rs` conditional compilation for new allocators, compile_error guards for invalid combinations
- `common/malloc_utils/src/mimalloc_alloc.rs` new `#[global_allocator]` module
- `common/malloc_utils/src/tcmalloc_alloc.rs` new `#[global_allocator]` module
- `lighthouse/Cargo.toml` expose mimalloc/tcmalloc features
- `lcli/Cargo.toml` expose mimalloc/tcmalloc features

Allocator metrics for mimalloc/tcmalloc are stubbed out can be filled in as a follow-up once we decide which allocator to invest in.


michaelsproul:
Please familiarise yourself with our policy regarding AI contributions (currently in review but likely to be merged very soon):

- https://github.com/sigp/lighthouse/pull/8879

lcli/Cargo.toml
```diff
@@ -11,6 +11,10 @@ normal = ["malloc_utils"]
 [features]
 portable = ["bls/supranational-portable"]
 fake_crypto = ['bls/fake_crypto']
+# Use mimalloc as the global allocator (alternative to jemalloc).
+mimalloc = ["malloc_utils/mimalloc"]
+# Use tcmalloc (gperftools) as the global allocator (alternative to jemalloc).
+tcmalloc = ["malloc_utils/tcmalloc"]
```
michaelsproul: Do you mind updating this to `tcmalloc-better` and re-running the `lcli` benchmark?

https://crates.io/crates/tcmalloc-better

From my understanding, the `tcmalloc` crate is using a very old version of tcmalloc called `gperftools`. I found more context here: https://google.github.io/tcmalloc/gperftools.html

michaelsproul (changes requested):
Looks like a good starting point.

I think there are probably a few more considerations to help us choose between `tcmalloc` and `mimalloc`, like:

- Node performance: we should deploy some long-running nodes with both allocators and observe their performance
- Debug tooling: which allocator has better tools for debugging, e.g. live allocation counts, heap introspection, etc.
- Compile time: does one of the two compile substantially faster?

lcli/Cargo.toml
0xMars42: Thanks, I didn't see that crate

michaelsproul:
Interesting that modern tcmalloc doesn't support macOS. That might be a reason to go deeper on mimalloc.

It could also be that the macOS system allocator is fine for Lighthouse. It's mostly the glibc allocator we want to avoid.

0xMars42:
> Interesting that modern tcmalloc doesn't support macOS. That might be a reason to go deeper on mimalloc.
> 
> It could also be that the macOS system allocator is fine for Lighthouse. It's mostly the glibc allocator we want to avoid.

Thanks for the great feedback! Don't hesitate if you'd like me to adjust anything else.

jxs:
Hi, and thanks for this!
The benchmarks provided lack depth thou, I think ideally we'd also want to compare RSS and time spent during allocator operations like this user provided [here](https://news.ycombinator.com/item?id=47403847).

Meanwhile Meta just recently published https://engineering.fb.com/2026/03/02/data-infrastructure/investing-in-infrastructure-metas-renewed-commitment-to-jemalloc/.
So we should probably wait some more months to see how `jemalloc` evolves since it has been un-deprecated

0xMars42:
Thanks for the pointers @jxs! I went through the [HN benchmarks from skavi](https://news.ycombinator.com/item?id=47403847) in detail and ran the requested measurements on Lighthouse workloads.

## Benchmark Setup

- **CPU:** AMD Ryzen 7 7800X3D (8C/16T)
- **RAM:** 32 GB
- **Kernel:** 6.19.7 (Fedora 42)
- **Block:** mainnet slot 13989601 (Electra, 136 KB)
- **Pre-state:** finalized state at slot 13989600 (308 MB)
- **Allocator versions:** jemalloc 5.3.0 (tikv-jemallocator 0.6), mimalloc 0.1.48 (libmimalloc-sys 0.1.44), tcmalloc-better 0.1.19

## Wall Time & RSS

### `lcli transition-blocks` (10 internal runs × 3 invocations)

| Allocator | Median | Min | Max | Peak RSS |
|-----------|--------|-----|-----|----------|
| jemalloc | 20.06s | 20.03s | 20.12s | 3,020 MB |
| mimalloc | 20.02s | 19.98s | 20.14s | 3,016 MB |
| tcmalloc | 20.04s | 19.92s | 20.06s | 3,018 MB |

### `lcli skip-slots` (32 slots / 1 epoch, 3 runs)

| Allocator | Wall Time | Peak RSS |
|-----------|-----------|----------|
| jemalloc | 7.17s | 2,130 MB |
| mimalloc | 7.16s | 2,130 MB |
| tcmalloc | 7.19s | 2,130 MB |

### `lcli state-root` (10 runs)

| Allocator | Wall Time | Peak RSS |
|-----------|-----------|----------|
| jemalloc | 1.29s | 1,254 MB |
| mimalloc | 1.29s | 1,254 MB |
| tcmalloc | 1.28s | 1,254 MB |

## Allocator Time (perf profiling)

To match skavi's methodology, I profiled the time actually spent in allocator operations using `perf record` with debug symbols:

For `transition-blocks` (jemalloc build), **total allocator overhead is ~0.67% of CPU time** (~130ms out of 20s). The breakdown:

```
0.17%  malloc_consolidate
0.12%  _int_free_chunk
0.11%  _int_malloc
0.11%  cfree
0.05%  malloc
0.05%  unlink_chunk
0.01%  __rdl_alloc (Rust global allocator)
```

The dominant hotspot is BLS cryptography (`blst` library), which accounts for ~76% of CPU (`__mulx_mont_384` alone is 59%). Allocation is simply not a bottleneck for Lighthouse's current workloads.

This explains why all three allocators perform identically there's so little allocation pressure that the allocator choice is immaterial. This is fundamentally different from skavi's async tokio apps where allocation overhead was 1-10% of total runtime.

## Build Times (clean allocator crate + lcli relink)

| Allocator | Build Time |
|-----------|------------|
| jemalloc | 32.0s |
| mimalloc | 18.1s |
| tcmalloc | 18.2s |

## Conclusion

For `lcli` batch workloads, there is no measurable performance difference between allocators Lighthouse is compute-bound (BLS), not allocation-bound.

That said, a **long-running beacon node** under sustained load (HTTP API, gossip, syncing) may have different allocation patterns particularly around memory fragmentation over time. I don't currently have a long-running setup to test this.

Given Meta's [renewed investment in jemalloc](https://engineering.fb.com/2026/03/02/data-infrastructure/investing-in-infrastructure-metas-renewed-commitment-to-jemalloc/) and the fact that these benchmarks show no compelling reason to switch for batch workloads, it might make sense to keep this PR as **optional allocator support** for operators who want to experiment, rather than advocating for a default change. Happy to adjust scope.

PR #8886, opened 2026-02-23
Update database and block replayer to handle payload envelopes

michaelsproul:
## Issue Addressed

Closes:

- https://github.com/sigp/lighthouse/issues/8869

## Proposed Changes

- Update `BlockReplayer` to support replay of execution payload envelopes.
- Update `HotColdDB` to load payload envelopes and feed them to the `BlockReplayer` for both hot + cold states. However the cold DB code is not fully working yet (see: https://github.com/sigp/lighthouse/issues/8958).
- Add `StatePayloadStatus` to allow callers to specify whether they want a state with a payload applied, or not.
- Fix the state cache to key by `StatePayloadStatus`.
- Lots of fixes to block production and block processing regarding state management.
- Initial test harness support for producing+processing Gloas blocks+envelopes
- A few new tests to cover Gloas DB operations

## Additional Info

Based on:

- https://github.com/sigp/lighthouse/pull/8806


michaelsproul:
I've got claude extending the EF tests to test database codepaths. Will post that PR separately and use it to test some more.

Still a few pieces missing too (state cache, cold DB, etc).

beacon_node/beacon_chain/src/block_verification.rs
```diff
@@ -1991,9 +1997,30 @@ fn load_parent<T: BeaconChainTypes, B: AsBlock<T::EthSpec>>(
         // Retrieve any state that is advanced through to at most `block.slot()`: this is
         // particularly important if `block` descends from the finalized/split block, but at a slot
         // prior to the finalized slot (which is invalid and inaccessible in our DB schema).
+        //
+        // Post-Gloas we must also fetch a state with the correct payload status. If the current
+        // block builds upon the payload of its parent block, then we know the parent block is FULL
+        // and we need to load the full state.
+        let (payload_status, parent_state_root) =
+            if block.as_block().fork_name_unchecked().gloas_enabled() {
+                let parent_bid_block_hash = parent_block.payload_bid_block_hash()?;
```
eserilev: Testing on kurtosis and I think this might cause a bug at the fork transition since there wont be a bid for the parent block at the fork

michaelsproul: Oh yeah shit I think you're right. Did you patch this on the ePBS-devnet branch?

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -4692,7 +4715,12 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
                     .ok_or(Error::MissingBeaconBlock(parent_block_root))?;
                 let (state_root, state) = self
                     .store
-                    .get_advanced_hot_state(parent_block_root, proposal_slot, block.state_root())?
+                    .get_advanced_hot_state(
+                        parent_block_root,
+                        StatePayloadStatus::Pending,
+                        proposal_slot,
+                        block.state_root(),
+                    )?
```
ethDreamer, line 4684: Oooof I don't think this one is simple anymore..

So this particular version of `get_expected_withdrawals()` is called only by [prepare_beacon_proposer()](https://github.com/sigp/lighthouse/blob/v8.1.1/beacon_node/beacon_chain/src/beacon_chain.rs#L6145). This function is meant to grab the state which the next proposer will build on and use it to calculate the `expected_withdrawals` for the `payload_attributes`.

But the correct state could be *either* the `block_state` or the `payload_state` depending on the timeliness & validity of the payload. And unfortunately, the `expected_withdrawals` changes depending on which one it is. Looking at [`process_execution_payload()`](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/beacon-chain.md#new-process_execution_payload) we create a new `payload_state`:
```python
def process_execution_payload(
    state: BeaconState,
    signed_envelope: SignedExecutionPayloadEnvelope,
    execution_engine: ExecutionEngine,
    verify: bool = True,
) -> None:
    ...
    if amount > 0:
        state.builder_pending_withdrawals.append(payment.withdrawal)
```
but then the next time we call [`process_block()`](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/beacon-chain.md#block-processing) we see:
```python
def process_block(state: BeaconState, block: BeaconBlock) -> None:
    ...
    process_withdrawals(state) # <-- could be either a block_state or payload_state
...
def process_withdrawals(
    state: BeaconState,
) -> None:
    ...
    expected = get_expected_withdrawals(state)
...
def get_expected_withdrawals(state: BeaconState) -> ExpectedWithdrawals:
    withdrawal_index = state.next_withdrawal_index
    withdrawals: List[Withdrawal] = []

    # [New in Gloas:EIP7732]
    # Get builder withdrawals
    builder_withdrawals, withdrawal_index, processed_builder_withdrawals_count = (
        get_builder_withdrawals(state, withdrawal_index, withdrawals)
    )
```
and [`get_builder_withdrawals()`](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/beacon-chain.md#new-get_builder_withdrawals) drains from `state.builder_pending_withdrawals` which was just appended to during `process_execution_payload()` if the payload was timely and valid:
```python
def get_builder_withdrawals(
    state: BeaconState,
    withdrawal_index: WithdrawalIndex,
    prior_withdrawals: Sequence[Withdrawal],
) -> Tuple[Sequence[Withdrawal], WithdrawalIndex, uint64]:
    withdrawals_limit = MAX_WITHDRAWALS_PER_PAYLOAD - 1
    assert len(prior_withdrawals) <= withdrawals_limit

    processed_count: uint64 = 0
    withdrawals: List[Withdrawal] = []
    for withdrawal in state.builder_pending_withdrawals:
        ...
        withdrawals.append(
            Withdrawal(
                index=withdrawal_index,
                validator_index=convert_builder_index_to_validator_index(builder_index),
                address=withdrawal.fee_recipient,
                amount=withdrawal.amount,
            )
        )
    ...
    return withdrawals, withdrawal_index, processed_count
```

Not sure what the best way is to proceed here beyond having *two* potential payload attributes to use.. the tough part is that we're calculating this in advance by design..

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -2023,7 +2041,12 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             } else {
                 let (advanced_state_root, mut state) = self
                     .store
-                    .get_advanced_hot_state(beacon_block_root, request_slot, beacon_state_root)?
+                    .get_advanced_hot_state(
+                        beacon_block_root,
+                        StatePayloadStatus::Pending,
+                        request_slot,
+                        beacon_state_root,
+                    )?
```
ethDreamer, line 2043: This should be fine as there should be no difference between a `block_state` and `payload_state` in terms of the shuffling.

beacon_node/beacon_chain/src/beacon_chain.rs
```diff
@@ -6730,7 +6711,12 @@ impl<T: BeaconChainTypes> BeaconChain<T> {
             } else {
                 let (state_root, state) = self
                     .store
-                    .get_advanced_hot_state(head_block_root, target_slot, head_block.state_root)?
+                    .get_advanced_hot_state(
+                        head_block_root,
+                        StatePayloadStatus::Pending,
+                        target_slot,
+                        head_block.state_root,
+                    )?
```
ethDreamer, line 6625: This should be fine as there should be no difference between a `block_state` and `payload_state` in terms of the shuffling.

beacon_node/beacon_chain/src/block_verification.rs
ethDreamer: If we are going to load the envelope here, do we want to put it into the `PreProcessingSnapshot`? This function is the only place a `PreProcessingSnapShot` is constructed, and you mentioned it would be useful to have in another comment (which you appear to have deleted now).

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -2510,7 +2663,58 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
+        let envelopes = self.load_payload_envelopes_for_blocks(
+            &blocks,
+            end_block_root,
+            desired_payload_status,
+        )?;
+
+        Ok((blocks, envelopes))
+    }
+
+    pub fn load_payload_envelopes_for_blocks(
```
ethDreamer, line 2674: infinitely cleaner than when i tried to implement this lol

beacon_node/beacon_chain/src/beacon_chain.rs
michaelsproul: Ah great catch. That is really gross

beacon_node/http_api/src/block_rewards.rs
```diff
@@ -138,11 +144,12 @@ pub fn compute_block_rewards<T: BeaconChainTypes>(
                     ))
                 })?;
 
+            // TODO(gloas): handle payloads?
```
ethDreamer: Strangely.. this function is only used by an undocumented api endpoint (the `GET` version is documented, the `POST` version which uses this is not). I guess with the post version, you pass `beacon_blocks` manually.. idk why anyone would want this instead of just specifying a start and an end slot like the other API method.. but making this work would require you to manually load the envelopes and apply them.. but still use the same blocks that were passed in.. which seems dumb..

michaelsproul: ohh this was for blockprint, so it could compute rewards for fictitious blocks

we can probably just delete this endpoint now that blockprint is ded

beacon_node/beacon_chain/src/beacon_chain.rs
michaelsproul: Yes, I believe so. Lion has a doc on this that I think we should consider summarising and including in the repo: https://hackmd.io/@dapplion/gloas_dependant_root

beacon_node/beacon_chain/src/beacon_chain.rs
pawanjay176: This is a great catch! It kinda relates to this issue in terms of timing https://github.com/sigp/lighthouse/issues/8912

We have 3 cases when this function gets called:
1. payload has arrived and completed `process_payload_envelope`
2. payload has arrived but not completed `process_payload_envelope` so the head state hasn't updated to the post envelope state
3. payload hasn't arrived

For (2) and (3), once the payload completes processing, we should probably treat it similar to a re-org and just re-fcu with the updated params. It would also make sense to slightly delay the fcu call by tweaking `prepare_payload_lookahead ` to that effect.

michaelsproul: Opened an issue for this, as I think it would bloat this PR too much to add it here:

- https://github.com/sigp/lighthouse/issues/8957

I'll also add a TODO in the code.

beacon_node/beacon_chain/src/block_verification.rs
michaelsproul: Sounds sensible to me. I've opened an issue for this rather than growing this PR further:

- https://github.com/sigp/lighthouse/issues/8961

beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -5427,6 +5437,427 @@ fn check_finalization(harness: &TestHarness, expected_slot: u64) {
     );
 }
 
+// ===================== Gloas Store Tests =====================
```
michaelsproul, line 5483: Note for reviewers: all of these new tests pass (running with `FORK_NAME=gloas`), but none of the other beacon chain tests are expected to pass yet. Need some more glue and plumbing in the `BeaconChainHarness` (in a follow-up PR). Might also need fork choice.

beacon_node/http_api/src/attestation_performance.rs
```diff
@@ -205,8 +205,9 @@ pub fn get_attestation_performance<T: BeaconChainTypes>(
             })
             .collect::<Result<Vec<_>, _>>()?;
 
+        // TODO(gloas): add payloads
```
michaelsproul, line 208: @macladson is gonna raise a PR to delete these `/lighthouse/analysis` APIs :relaxed:

beacon_node/store/src/reconstruct.rs
```diff
@@ -67,6 +67,7 @@ where
 
         state.build_caches(&self.spec)?;
 
+        // TODO(gloas): handle payload envelope replay
```
michaelsproul, line 70: Issue for this here:

- https://github.com/sigp/lighthouse/issues/8962

michaelsproul:
Self-review complete. This is ready for a review (and merge :crossed_fingers: )

michaelsproul:
Failing cargo audit, need:

- https://github.com/sigp/lighthouse/pull/8886

beacon_node/beacon_chain/src/block_verification.rs
michaelsproul: Bug fixed here:

https://github.com/sigp/lighthouse/pull/8886/commits/2b3157ea38dac441809962baa556d79e20e3ac9a

consensus/types/src/block/signed_beacon_block.rs
```diff
@@ -365,6 +366,32 @@ impl<E: EthSpec, Payload: AbstractExecPayload<E>> SignedBeaconBlock<E, Payload>
...
+
+    /// Convenience accessor for the block's bid's `block_hash`.
+    ///
+    /// This method returns an error prior to Gloas.
+    pub fn payload_bid_block_hash(&self) -> Result<ExecutionBlockHash, BeaconStateError> {
+        self.message()
+            .body()
+            .signed_execution_payload_bid()
+            .map(|bid| bid.message.block_hash)
+    }
```
eserilev, line 378: Nit: Pre-Gloas, block hash is available via `body.execution_payload.block_hash`. Could rename the function to `execution_block_hash` and return block hash post-belatrix -> gloas

consensus/state_processing/src/block_replayer.rs
```diff
@@ -208,23 +246,96 @@ where
...
             }
 
             while self.state.slot() < block.slot() {
-                let state_root = self.get_state_root(&blocks, i)?;
+                let mut state_root = self.get_state_root(&blocks, i)?;
+
+                // Apply the payload for the *previous* block if the bid in the current block
+                // indicates that the parent is full (and it hasn't already been applied).
+                state_root = if block.fork_name_unchecked().gloas_enabled()
+                    && self.state.slot() == self.state.latest_block_header().slot
```
eserilev, line 315: for others reviewing, this invariant helped clear things up in my head

`state.slot() >= state.latest_block_header().slot` is always true

consensus/state_processing/src/block_replayer.rs
```diff
@@ -268,9 +379,24 @@ where
             }
         }
 
+        // Apply the last payload if desired.
+        let mut opt_state_root = if let StatePayloadStatus::Full = self.desired_state_payload_status
+            && let Some(last_block) = blocks.last()
+        {
+            let envelope = next_envelope_at_slot(self.state.slot())?;
+            Some(self.apply_payload_envelope(&envelope, last_block.state_root())?)
```
eserilev, line 387: Noting that if the desired payload status is `Full` but there is no payload, we error out. Should the replayer return the pending state if the payload doesn't exist? Are there situations where we might just want the latest valid state i.e. `Full` or `Pending` if no envelope

consensus/state_processing/src/state_advance.rs
```diff
@@ -77,6 +77,11 @@ pub fn partial_state_advance<E: EthSpec>(
     // (all-zeros) state root.
     let mut initial_state_root = Some(if state.slot() > state.latest_block_header().slot {
         state_root_opt.unwrap_or_else(Hash256::zero)
+    } else if state.slot() == state.latest_block_header().slot
+        && !state.latest_block_header().state_root.is_zero()
+    {
+        // Post-Gloas Full block case.
```
eserilev:
```suggestion
        // Post-Gloas Full state case.
```

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -1846,6 +1857,55 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
+
+        // Treat genesis state as `Pending` (`BeaconBlock` state).
+        let previous_state_root = summary.previous_state_root;
+        if previous_state_root.is_zero() {
+            return Ok(StatePayloadStatus::Pending);
+        }
+
+        // Load the hot state summary for the previous state.
+        //
+        // If it has the same slot as this summary then we know this summary is for a `Full` block
```
eserilev:
```suggestion
        // If it has the same slot as this summary then we know this summary is for a `Full` state
```

eserilev (commented):
I'm about halfway through this review. Everything looks really good, just one question. Will continue the review tomorrow

And fwiw this PR has already been battle tested on devnet-0 and is probably the biggest reason we've been able to survive

beacon_node/store/src/hot_cold_store.rs
```diff
@@ -2520,17 +2655,71 @@ impl<E: EthSpec, Hot: ItemStore<E>, Cold: ItemStore<E>> HotColdDB<E, Hot, Cold>
...
+        }
+
+        // Load the payload for the last block if desired.
+        // TODO(gloas): check that we don't load a duplicate in the case of a skipped slot
+        if let StatePayloadStatus::Full = desired_payload_status {
+            let envelope = self.get_payload_envelope(&end_block_root)?.ok_or(
+                HotColdDBError::MissingExecutionPayloadEnvelope(end_block_root),
+            )?;
+            envelopes.push(envelope);
+        }
```
eserilev, line 2705: Similar question to https://github.com/sigp/lighthouse/pull/8886/changes#r2910430601

Do we want the option of fetching the latest envelopes regardless of wether the last one is empty or full?

eserilev (approved):
LGTM!
Just had a question or two

consensus/state_processing/src/block_replayer.rs
michaelsproul: I think that's probably not useful. The caller should know which state they need and request accordingly, I think it's a bit dodgy to do anything best-effort.

If the caller doesn't require the payload, then they should just request `Pending`.

consensus/types/src/block/signed_beacon_block.rs
michaelsproul: I'm not sure if these sorts of universal functions are more confusing or less. I think I'm erring on the side of isolating Gloas functionality from pre-Gloas functionality, so that it's clear where a function is intended to be used, and we don't end up doing weird stuff trying to shoehorn old code into new norms.

Another similar example is `is_parent_block_full`. I think that should probably return an error on all non-Gloas blocks, because the concept isn't well-defined pre-Gloas.

beacon_node/store/src/hot_cold_store.rs
michaelsproul: I think for now (similar to the other API) we should do as the user instructs. If we need the flexibility in future we can add it.

michaelsproul:
I think all comments addressed. I'll merge once Lion's PR is in unstable. Probably don't need to fix the invariant failures yet even, that could be follow-up.

- https://github.com/sigp/lighthouse/pull/8952

eserilev (approved):

michaelsproul:
@mergify dequeue

michaelsproul:
Not sure why the simulator is suddenly failing

michaelsproul:
@mergify requeue

PR #8890, opened 2026-02-24
Ensure consistent cleanup of range sync request tracking

jimmygchen:
## Issue Addressed

Ensure `SyncNetworkContext` request-tracking maps are cleaned up on all exit paths.

`SyncNetworkContext` tracks active range sync requests in two maps: `components_by_range_requests` and `custody_backfill_data_column_batch_requests`. Several code paths left stale entries in these maps:

#### `components_by_range_requests`:
- `retry_columns_by_range`: when peer selection or request sending fails mid-retry, the entry created for the original request was left behind.
- Chain removal: when a range sync chain is removed (peer disconnect, chain failure), its associated entries were not cleaned up.

#### `custody_backfill_data_column_batch_requests`:
- `custody_backfill_data_columns_response`: when the aggregated response completed with a `DataColumnPeerFailure` (missing or bad columns), the entry was retained. The caller creates a new entry for the retry, so the old one should always be removed on completion.


beacon_node/network/src/sync/network_context.rs
```diff
@@ -533,6 +584,14 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
         Ok(())
     }
 
+    /// Remove all `components_by_range_requests` entries associated with the given range sync chain.
+    /// Defense-in-depth cleanup when a chain is removed.
+    pub fn remove_components_by_range_for_chain(&mut self, chain_id: Id) {
+        self.components_by_range_requests.retain(|key, _| {
+            !matches!(key.requester, RangeRequestId::RangeSync { chain_id: id, .. } if id == chain_id)
```
pawanjay176: should we do something similar for backfill?
In fail_sync, we clear all batches. Should we also remove all backfill components from `components_by_range_requests` at that point?

beacon_node/network/src/sync/tests/range.rs
```diff
@@ -631,3 +641,140 @@ fn finalized_sync_not_enough_custody_peers_on_start() {
     let last_epoch = advanced_epochs + EXTRA_SYNCED_EPOCHS;
     r.complete_and_process_range_sync_until(last_epoch, filter());
 }
+
+/// Test that components_by_range_requests entries are cleaned up when a chain is removed.
+#[test]
+fn chain_removal_cleans_up_components_by_range() {
+    let mut rig = TestRig::test_setup();
+    // Only run post-PeerDAS (components_by_range only exists with data columns)
+    if !rig.fork_name.fulu_enabled() {
```
pawanjay176: This test doesn't really need data columns though. we can run pre fulu too.

beacon_node/network/src/sync/tests/range.rs
```diff
@@ -631,3 +641,140 @@ fn finalized_sync_not_enough_custody_peers_on_start() {
...
+}
+
+/// Test that retry_columns_by_range cleans up the components_by_range entry when
+/// no custody peer is available for the retry.
+#[test]
+fn retry_columns_by_range_cleans_up_on_no_peers() {
+    use lighthouse_network::rpc::BlocksByRangeRequest;
+
+    let mut rig = TestRig::test_setup();
+    // Only run post-PeerDAS
```
pawanjay176: same as above

pawanjay176 (commented):
Looks good mostly, just curious what you think about the backfill case.

beacon_node/network/src/sync/network_context.rs
```diff
@@ -251,6 +251,47 @@ pub enum RangeBlockComponent<E: EthSpec> {
...
+    pub(crate) fn active_request_counts(&self) -> ActiveRequestCounts {
+        ActiveRequestCounts {
+            components_by_range: self.components_by_range_requests.len(),
+            custody_backfill_batches: self.custody_backfill_data_column_batch_requests.len(),
+        }
+    }
+
+    /// Inserts a test entry into `custody_backfill_data_column_batch_requests` with a single
+    /// sub-request that expects the given custody columns. Used for testing cleanup behavior.
+    pub(crate) fn insert_test_custody_backfill_entry(
```
dapplion: Is this stub necessary? I wonder if you can cause the system to add to `custody_backfill_data_column_batch_requests` with regular sync messages

beacon_node/network/src/sync/network_context.rs
```diff
@@ -494,7 +535,12 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
                 active_request_count_by_peer,
                 peers_to_deprioritize,
             )
-            .map_err(|e| format!("{:?}", e))?;
+            .map_err(|e| {
+                // Clean up the components_by_range_requests entry before returning error
+                self.components_by_range_requests
+                    .retain(|key, _| key.id != id);
```
dapplion: Damn is hard to reason about

beacon_node/network/src/sync/network_context.rs
```diff
@@ -251,6 +251,47 @@ pub enum RangeBlockComponent<E: EthSpec> {
...
 }
 
+/// Counts of active meta-requests, for verifying resource cleanup in tests.
+#[cfg(test)]
+pub(crate) struct ActiveRequestCounts {
+    pub components_by_range: usize,
+    pub custody_backfill_batches: usize,
+}
+
+#[cfg(test)]
```
dapplion: For clarify can we put the cfg(test) on each function to make it extra clear to mantainers that this specific function is for tests? We have 3 impl blocks in this 1 file

dapplion: This compacted diff threw me off, the .map_err happens in an impl that's not cfg(test)

Image

beacon_node/network/src/sync/range_sync/range.rs
```diff
@@ -337,6 +337,8 @@ where
             debug!(id = chain.id(), ?sync_type, reason = ?remove_reason, op, "Chain removed");
         }
 
+        network.remove_components_by_range_for_chain(chain.id());
```
dapplion: Why are components_by_range requests stale in the first place? They should either:
- drop if 0 no peers, or timeout if allowed to have 0 peers
- fail after N retries

In no case they should linger there forever

jimmygchen: Good point, i'll remove `remove_components_by_range_for_chain`

beacon_node/network/src/sync/network_context.rs
jimmygchen: This is an impl with `TestBeaconChainType<E>` as generic parameter, hence the entire block is `cfg(test)`, because the type `TestBeaconChainType` is also `cfg(test)`.

I'll come back to this later once we get a working fix. I'm not convinced this PR fixed the issue.

beacon_node/network/src/sync/network_context.rs
jimmygchen: removed this completely as discussed

beacon_node/network/src/sync/tests/range.rs
jimmygchen: Removed `chain_removal_cleans_up_components_by_range`

beacon_node/network/src/sync/tests/range.rs
jimmygchen: This one calls `retry_columns_by_range` so only PeerDAS i think?

beacon_node/network/src/sync/network_context.rs
jimmygchen: Removed the stub and the test that used it. Replaced with [`components_by_range_cleaned_up_after_retry_cycle`](https://github.com/sigp/lighthouse/pull/8890/commits/4c8abf789c41410b9e528bdb15f2c6085dd5883f) which drives the full retry loop through events only and no internal state mutation.

beacon_node/network/src/sync/range_sync/range.rs
jimmygchen: Just to add, in the normal flow they shouldn't (the two fixes in this PR) and this was just being defensive but hides the problem.  With the fixes in place, this is unnecessary, so I've removed this.

beacon_node/network/src/sync/network_context.rs
jimmygchen: lol ignore my comment above, I also got confused by the collapsed view too. The test functions have been moved to the `impl<E: EthSpec> SyncNetWorkContext<TestBeaconChainHarnessType<E>` block

jimmygchen:
@dapplion I've pushed a fixed for a 3rd issue I found. I know the fix but I have not yet have the chance to self-review the actual implementation, so I've marked this as `work-in-progress` but would be useful if you can look at it.

I'm also in the process of adding a test.

beacon_node/network/src/sync/tests/range.rs
```diff
@@ -631,3 +679,268 @@ fn finalized_sync_not_enough_custody_peers_on_start() {
     let last_epoch = advanced_epochs + EXTRA_SYNCED_EPOCHS;
     r.complete_and_process_range_sync_until(last_epoch, filter());
 }
+
+/// Test that retry_columns_by_range cleans up the components_by_range entry when
+/// no custody peer is available for the retry.
+#[test]
+fn retry_columns_by_range_cleans_up_on_no_peers() {
+    use lighthouse_network::rpc::BlocksByRangeRequest;
```
dapplion, line 770: Why does Claude do this random imports inside functions?

beacon_node/network/src/sync/tests/range.rs
```diff
@@ -631,3 +679,268 @@ fn finalized_sync_not_enough_custody_peers_on_start() {
...
+        peer_id: block_peer_0,
+        beacon_block: Some(block.into()),
+        seen_timestamp: D,
+    });
+    rig.send_sync_message(SyncMessage::RpcBlock {
+        sync_request_id: SyncRequestId::BlocksByRange(blocks_req_id_0),
+        peer_id: block_peer_0,
+        beacon_block: None,
+        seen_timestamp: D,
+    });
```
dapplion: @jimmygchen and you complained about the verbosity of my tests? 😆

beacon_node/network/src/sync/backfill_sync/mod.rs
```diff
@@ -435,14 +435,22 @@ impl<T: BeaconChainTypes> BackFillSync<T> {
...
         if matches!(error, BackFillError::Paused) {
             return Ok(());
         }
 
         // Set the state
         self.set_state(BackFillState::Failed);
+        // Clean up any orphaned components_by_range entries for backfill.
+        network.remove_components_by_range_requests(|r| {
+            matches!(r, RangeRequestId::BackfillSync { .. })
+        });
```
dapplion, line 454: I still don't get why this clean up is necessary

dapplion: Network requests are finite by definition as libp2p enforces timeouts + we (should) have a finite amount of retries. So eventually all requests either succeed or fail. If a request completes (ok/nok) for a backfill run of forward sync chain that no longer exists then fine. But if `components_by_range_requests` never expire that points to an issue in the design of those

pawanjay176: Yeah I agree. This is hiding something in backfill that we should address

beacon_node/network/src/sync/network_context.rs
```diff
@@ -1821,6 +1848,22 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
         ] {
             metrics::set_gauge_vec(&metrics::SYNC_ACTIVE_NETWORK_REQUESTS, &[id], count as i64);
         }
+
+        // Detect stale components_by_range entries (older than 60s)
```
jimmygchen, line 1847: This is not useful and can be removed

beacon_node/network/src/sync/tests/range.rs
jimmygchen: you got me 🤣

beacon_node/network/src/sync/network_context.rs
jimmygchen: right, so it looks like there's nothing that needs address here?

beacon_node/network/src/sync/tests/range.rs
jimmygchen: I've actually told claude to NOT do it and fix them, but looks like it either missed this one, or added a new test that followed the old pattern 💀

beacon_node/network/src/sync/backfill_sync/mod.rs
jimmygchen: I think this is the sequence:

Consider two in-flight batches A & B:
1. Batch A fails due to a peer failure and exceeds retry, `fail_sync` clears all the batches (including B)
  * https://github.com/sigp/lighthouse/blob/6166ad2eb298a17e97c9f05638267175057c495e/beacon_node/network/src/sync/backfill_sync/mod.rs#L366-L369
  * https://github.com/sigp/lighthouse/blob/6166ad2eb298a17e97c9f05638267175057c495e/beacon_node/network/src/sync/backfill_sync/mod.rs#L443-L444
2. Batch B's responses arrives, hasn't exceeded retry, so it's entry was kept to be retried later here:
  * https://github.com/sigp/lighthouse/blob/f4a6b8d9b97bfcbf901423c109901561d3e7e928/beacon_node/network/src/sync/network_context.rs#L790-L797
3. It then reach [`inject_error`](https://github.com/sigp/lighthouse/blob/f4a6b8d9b97bfcbf901423c109901561d3e7e928/beacon_node/network/src/sync/manager.rs#L1370-L1377), which is suppose to perform retry, however the batch no longer exists, so it never performs the retry and the request stays in the map. https://github.com/sigp/lighthouse/blob/6166ad2eb298a17e97c9f05638267175057c495e/beacon_node/network/src/sync/backfill_sync/mod.rs#L373-L374

So I think the correct fix here is to clear the requests when the batches are cleared in step 1, which is exactly what this fix does.

beacon_node/network/src/sync/tests/range.rs
jimmygchen: I'll manually trim this down now

beacon_node/network/src/sync/network_context.rs
```diff
@@ -487,19 +507,24 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
+        let columns_by_range_peers_to_request = match self.select_columns_by_range_peers_to_request(
+            failed_columns,
+            peers,
+            active_request_count_by_peer,
+            peers_to_deprioritize,
+        ) {
+            Ok(peers) => peers,
+            Err(e) => {
+                let id = ComponentsByRangeRequestId { id, requester };
+                self.components_by_range_requests.remove(&id);
```
jimmygchen, line 514: Not sure if this is correct? if we can't find peers, we shouldn't just remove the `components_by_range` request.

beacon_node/network/src/sync/network_context.rs
```diff
@@ -518,11 +543,17 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
                 )
             })
-            .collect::<Result<Vec<_>, _>>()
-            .map_err(|e| format!("{:?}", e))?;
+            .collect();
+
+        let data_column_requests = match data_column_requests {
+            Ok(reqs) => reqs,
+            Err(e) => {
+                self.components_by_range_requests.remove(&id);
```
jimmygchen, line 546: Not sure if this is correct, if we fail to send a single column by range request - we probably shouldnt' remove the `components_by_range_request`? otherwise it won't be able to retry.

beacon_node/network/src/sync/network_context.rs
```diff
@@ -1787,11 +1816,9 @@ impl<T: BeaconChainTypes> SyncNetworkContext<T> {
...
 
         if let Some(data_column_result) = entry.get_mut().responses() {
-            if data_column_result.is_ok() {
-                // remove the entry only if it coupled successfully with
-                // no errors
-                entry.remove();
-            }
+            // Request is complete — always remove the entry. On error, the caller
+            // will create a new entry for the retry.
+            entry.remove();
```
jimmygchen, line 1816: Does this mean if we previously fail with a coupling error, the request stays in `custody_backfill_data_column_batch_requests` forever?
and when retried a new one gets created?

beacon_node/network/src/sync/network_context.rs
jimmygchen: 🤖 Tracing through the code: when `retry_columns_by_range` fails here, `retry_partial_batch` swallows the error and returns `Ok(KeepChain)`. The batch stays in `AwaitingDownload`. Later, `attempt_send_awaiting_download_batches` picks it up and calls `send_batch` → `block_components_by_range_request`, which creates a **new** entry with a new ID. So the old entry is orphaned — nobody calls `retry_columns_by_range` with the old ID again.

That said, removing the entry means `send_batch` re-downloads blocks that were already successfully received. Ideally the system would keep the entry and route the `AwaitingDownload` batch back through `retry_columns_by_range` (column-only retry) when peers become available, rather than through `send_batch` (full batch retry). But `attempt_send_awaiting_download_batches` doesn't currently distinguish "needs column-only retry" from "needs full batch download."

So the removal is correct given the current flow, but it points to a gap — there's no path to resume a column-only retry after a transient peer shortage.

beacon_node/network/src/sync/network_context.rs
jimmygchen: 🤖 Same situation as above. When this fails, `reinsert_failed_column_requests` is never called, so the entry's internal `requests` map still has the old (completed) sub-request IDs, not the new retry ones. Any responses from successfully-sent requests would hit `add_custody_columns` → `requests.get_mut(&req_id)` → `"unknown data columns by range req_id"` → error removal at line 774. And if the first send fails (`.collect()` short-circuits), no responses arrive at all — permanent leak.

The batch then falls through to `AwaitingDownload` → `send_batch` which creates a new entry, same as the peer selection case.

So keeping the entry doesn't help with the current flow — the entry can't be reused because the retry sub-requests were never registered in it. The deeper fix would be the same: handle partial send failures more granularly (register successful sends, track failed columns for later retry) so the entry can be reused.

beacon_node/network/src/sync/network_context.rs
jimmygchen: 🤖 Yes. Tracing the flow: when `responses()` returns `Err(DataColumnPeerFailure{exceeded_retries: false})`, the entry is kept (`is_ok()` is false → no removal). The error propagates to `custody_backfill_sync::on_data_column_response` → `download_failed` → `Continue` → `send_batch` → `custody_backfill_data_columns_batch_request`, which creates a **new** entry with a new `CustodyBackFillBatchRequestId`. The old entry is never referenced again.

So always-remove here is correct — the old entry is dead once the retry creates a fresh one.

jimmygchen:
Pushing this out of v8.1.1 as this still need further refinement and testing, and only affects some edge cases. 
Worth considering alternative fix from @dapplion  https://github.com/dapplion/lighthouse/pull/67

jimmygchen:
This may not be the solution we want but this hasn't been addressed, will reopen for tracking

jimmygchen:
Partially superseded by #9496. The remaining backfill fix to be done in a separate PR as the approach is outdated.

PR #8891, opened 2026-02-24
Delete dysfunctional fork_revert feature

michaelsproul:
## Issue Addressed

I found myself having to update this code for Gloas, and figured we may as well delete it seeing as it doesn't work.

See:

- https://github.com/sigp/lighthouse/issues/4198

## Proposed Changes

Delete all `fork_revert` logic and the accompanying test.

## Additional Info

This PR was written entirely by Claude, but I have reviewed the changes.


beacon_node/beacon_chain/tests/store_tests.rs
```diff
@@ -3924,188 +3924,6 @@ async fn finalizes_after_resuming_from_db() {
     );
 }
 
-#[allow(clippy::large_stack_frames)]
-#[tokio::test]
-async fn revert_minority_fork_on_resume() {
```
jimmygchen, line 3929: Stale comment in beacon_state.rs:60

The comment on DEFAULT_PRE_ELECTRA_WS_PERIOD references the deleted test by name:

```rust
  // The default pre-electra WS value is set to 256 to allow for `basic-sim``, `fallback-sim``
  // test case `revert_minority_fork_on_resume` to pass.
```

beacon_node/beacon_chain/src/fork_revert.rs
```diff
@@ -1,204 +0,0 @@
...
-            CORRUPT_DB_MESSAGE
-        ));
-    }
-
-    warn!(
-        target_fork = %current_fork,
-        %fork_epoch,
-        "Reverting invalid head block"
-    );
-    let block_iter = ParentRootBlockIterator::fork_tolerant(&store, head_block_root);
```
jimmygchen, line 48: `ParentRootBlockIterator::fork_tolerant` is now unused, can we delete that too?

https://github.com/sigp/lighthouse/blob/f3fd1f210b2f4ed7d208f81f9a09e1edced3bb3d/beacon_node/store/src/iter.rs#L269

jimmygchen (approved):
LGTM!

beacon_node/beacon_chain/src/fork_revert.rs
michaelsproul: Deleted by Claude and checked by me:

- b29c6c0e48b466e3fc07697491288961f0a38992

beacon_node/beacon_chain/tests/store_tests.rs
michaelsproul: Updated in:

- b29c6c0e48b466e3fc07697491288961f0a38992

PR #8895, opened 2026-02-24
Add debug spans to DB write paths

jimmygchen:
## Description

When backfill rate limiting is disabled, `persist_blocks_and_blobs` has been observed taking 27s. We suspect backfill is hogging LevelDB write locks on `hot_db`/`blobs_db`, but there's no breakdown within the span to confirm.

Adds `debug_span!` to both the head import (`do_atomically_with_block_and_blobs_cache`) and backfill (`import_historical_block_batch`) DB write paths so we can see which DB write is the bottleneck.


michaelsproul (approved):
Looks good.

Reminds me I should polish up:

- https://github.com/sigp/lighthouse/pull/8404

PR #8896, opened 2026-02-24
Gloas serve envelope rpc

eserilev:
## Issue Addressed

Serves envelope by range and by root requests. Added PayloadEnvelopeStreamer so that we dont need to alter upstream code when we introduce blinded payload envelopes.


beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -254,6 +255,113 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                requested = requested_envelopes,
+                returned = %send_envelope_count,
+                "ExecutionPayloadEnvelopes outgoing response processed"
+            );
+        };
+
+        let requested_envelopes = request.beacon_block_roots.len();
+        let mut envelope_stream = match self
+            .chain
+            .get_payload_envelopes_checking_caches(request.beacon_block_roots.to_vec())
```
dapplion: Why is it necessary to have this stream gadget instead of a simple for loop?

```rs
for block_root in block_roots {
  if Some(payload) = chain.get_payload_checking_caches(block_root) {
    send_to_peer(payload)
  }
}
```

jimmygchen (changes requested):
The new protocols need to be added to this function:
https://github.com/sigp/lighthouse/blob/ffc2b97699dff59c91dc07375e090e3473508240/beacon_node/lighthouse_network/src/rpc/protocol.rs#L389

Added some tests to catch this kind of bug:
* https://github.com/sigp/lighthouse/pull/8920

beacon_node/lighthouse_network/src/rpc/codec.rs
```diff
@@ -650,6 +668,16 @@ fn handle_rpc_response<E: EthSpec>(
...
+        SupportedProtocol::PayloadEnvelopesByRangeV1 => {
+            Ok(Some(RpcSuccessResponse::PayloadEnvelopesByRange(Arc::new(
+                SignedExecutionPayloadEnvelope::from_ssz_bytes(decoded_buffer)?,
+            ))))
+        }
+        SupportedProtocol::PayloadEnvelopesByRootV1 => {
+            Ok(Some(RpcSuccessResponse::PayloadEnvelopesByRoot(Arc::new(
+                SignedExecutionPayloadEnvelope::from_ssz_bytes(decoded_buffer)?,
+            ))))
+        }
```
jimmygchen: should probably check the `ForkName` given this protocol has context bytes?

beacon_node/lighthouse_network/src/rpc/methods.rs
```diff
@@ -505,6 +516,31 @@ impl BlocksByRootRequest {
...
+}
+
+impl PayloadEnvelopesByRootRequest {
+    pub fn new(
+        beacon_block_roots: Vec<Hash256>,
+        fork_context: &ForkContext,
+    ) -> Result<Self, String> {
+        let max_requests_envelopes = fork_context
+            .spec
+            .max_request_blocks(fork_context.current_fork_name());
```
jimmygchen: This should probably be `MAX_REQUEST_PAYLOADS`?

beacon_node/lighthouse_network/src/rpc/protocol.rs
```diff
@@ -549,6 +568,12 @@ impl ProtocolId {
             Protocol::Goodbye => RpcLimits::new(0, 0), // Goodbye request has no response
             Protocol::BlocksByRange => rpc_block_limits_by_fork(fork_context.current_fork_name()),
             Protocol::BlocksByRoot => rpc_block_limits_by_fork(fork_context.current_fork_name()),
+            Protocol::PayloadEnvelopesByRange => {
+                rpc_block_limits_by_fork(fork_context.current_fork_name())
+            }
+            Protocol::PayloadEnvelopesByRoot => {
+                rpc_block_limits_by_fork(fork_context.current_fork_name())
```
jimmygchen: we may need a `rpc_payload_envelope_limits` function and update the `rpc_block_limits_by_fork` to include Gloas block size limit.

beacon_node/lighthouse_network/src/rpc/protocol.rs
```diff
@@ -511,6 +525,11 @@ impl ProtocolId {
                 <OldBlocksByRangeRequestV2 as Encode>::ssz_fixed_len(),
             ),
             Protocol::BlocksByRoot => RpcLimits::new(0, spec.max_blocks_by_root_request),
+            Protocol::PayloadEnvelopesByRange => RpcLimits::new(
+                <PayloadEnvelopesByRangeRequest as Encode>::ssz_fixed_len(),
+                <PayloadEnvelopesByRangeRequest as Encode>::ssz_fixed_len(),
+            ),
+            Protocol::PayloadEnvelopesByRoot => RpcLimits::new(0, spec.max_blocks_by_root_request),
```
jimmygchen: we may want to consider adding a separate config for this.

beacon_node/network/src/network_beacon_processor/tests.rs
```diff
@@ -2102,3 +2129,226 @@ async fn test_data_columns_by_range_no_duplicates_with_skip_slots() {
...
+    let mut actual_count = 0;
+    while let Some(next) = rig.network_rx.recv().await {
+        if let NetworkMessage::SendResponse {
+            peer_id: _,
+            response: Response::PayloadEnvelopesByRange(envelope),
+            inbound_request_id: _,
+        } = next
+        {
+            if envelope.is_some() {
+                actual_count += 1;
```
jimmygchen: It may be worth verifying the roots too if it's not difficult?

beacon_node/beacon_chain/src/execution_payload_envelope_streamer.rs
```diff
@@ -0,0 +1,138 @@
...
+        for beacon_block_root in beacon_block_roots {
+            let cached_envelope = self.check_payload_envelope_cache(beacon_block_root);
+
+            let envelope_result = if cached_envelope.is_some() {
+                Ok(cached_envelope)
+            } else {
+                // TODO(gloas) we'll want to use the execution layer directly to call
+                //  the engine api method eth_getBlockByHash()
+                self.store
+                    .get_payload_envelope(&beacon_block_root)
```
jimmygchen: synchronous db read inside async task here, use a blocking task (`spawn_blocking`)

beacon_node/beacon_chain/src/execution_payload_envelope_streamer.rs
```diff
@@ -0,0 +1,138 @@
...
+    fn check_payload_envelope_cache(
+        &self,
+        _beacon_block_root: Hash256,
+    ) -> Option<Arc<SignedExecutionPayloadEnvelope<T::EthSpec>>> {
+        // if self.check_caches == CheckCaches::Yes
+        None
+    }
+
+    // used when the execution engine doesn't support the payload bodies methods
+    async fn stream_payload_envelopes_fallback(
```
ethDreamer: I would just rename this to `stream_payload_envelopes` as there will be no fallback since all ELs will implement `engine_getPayloadBodiesBy*` methods

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
eserilev: We're going to use `getPayloadBodiesByRange` which allows us to fetch a batch of payloads in one engine api request. If we did the simple for loop we'd have to make an engine api request for each payload

beacon_node/beacon_chain/src/execution_payload_envelope_streamer.rs
eserilev: yeah sounds good, i removed the engine capabilities check

beacon_node/lighthouse_network/src/rpc/methods.rs
eserilev: added MAX_REQUEST_PAYLOADS

beacon_node/beacon_chain/src/execution_payload_envelope_streamer.rs
eserilev: added spawn_blocking. I had to introduce a new span_blocking_and_await fn inside task executor. its the same `spawn_blocking` wrapper we have in beacon chain. I did this because I didnt want envelope streamer to rely on beacon chain.

beacon_node/network/src/network_beacon_processor/tests.rs
eserilev: We are now verifying roots

beacon_node/lighthouse_network/src/rpc/protocol.rs
eserilev: introduced separate config

beacon_node/lighthouse_network/src/rpc/protocol.rs
eserilev: updated rpc_block_limits_by_fork and introduced rpc_payload_envelope_limits

beacon_node/lighthouse_network/src/rpc/codec.rs
eserilev: checking fork name

beacon_node/beacon_chain/src/execution_payload_envelope_streamer.rs
ethDreamer: My point was that you should rename this function to `stream_payload_envelopes` and not `stream_payload_envelopes_fallback` because the `fallback` is what you call when the EL didn't support `engine_getPayloadBodyiesBy*` methods.

eserilev: yeah I did, but I also removed the engine capabilities check. Since the fallback mechanism was used when the EL didn't support the payload bodies by request

beacon_node/lighthouse_network/src/rpc/codec.rs
```diff
@@ -548,6 +553,19 @@ fn handle_rpc_request<E: EthSpec>(
...
+        SupportedProtocol::PayloadEnvelopesByRangeV1 => {
+            Ok(Some(RequestType::PayloadEnvelopesByRange(
+                PayloadEnvelopesByRangeRequest::from_ssz_bytes(decoded_buffer)?,
+            )))
+        }
+        SupportedProtocol::PayloadEnvelopesByRootV1 => Ok(Some(
+            RequestType::PayloadEnvelopesByRoot(PayloadEnvelopesByRootRequest {
+                beacon_block_roots: RuntimeVariableList::from_ssz_bytes(
+                    decoded_buffer,
+                    spec.max_request_blocks(current_fork),
```
jimmygchen: This should be `spec.max_request_payloads()` to be consistent with the other fixes

jimmygchen (changes requested):
The new protocols need count validation here too, otherwise a peer can make us iterate the full chain instead of the expected max 128 slots:

https://github.com/sigp/lighthouse/blob/e1d3dcc8dc19f137756f75c42c592964a07adfae/beacon_node/lighthouse_network/src/rpc/handler.rs#L957

beacon_node/beacon_chain/src/execution_payload_envelope_streamer.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    async fn stream_payload_envelopes(
+        self: Arc<Self>,
+        beacon_block_roots: Vec<Hash256>,
+        sender: UnboundedSender<(Hash256, Arc<PayloadEnvelopeResult<T::EthSpec>>)>,
+    ) {
+        let results = match self.load_envelopes(&beacon_block_roots).await {
+            Ok(results) => results,
+            Err(e) => {
+                send_errors(beacon_block_roots, sender, e).await;
```
jimmygchen: may be worth logging some errors/warn here?

consensus/types/src/block/beacon_block.rs
```diff
@@ -724,6 +727,135 @@ impl<E: EthSpec, Payload: AbstractExecPayload<E>> EmptyBlock for BeaconBlockGloa
...
+        }
+        for _ in 0..E::MaxPayloadAttestations::to_usize() {
+            block
+                .body
+                .payload_attestations
+                .push(payload_attestation.clone())
+                .unwrap();
+        }
+
+        block
```
jimmygchen: Also need to fill this variable list:
```rust
block.body.signed_execution_payload_bid.message.blob_kzg_commitments
```

eserilev: ah nice catch 
fbc5f0d

beacon_node/beacon_chain/src/execution_payload_envelope_streamer.rs
eserilev: added warn log

eserilev:
seems we were missing count validation for both envelopes and data columns [5b4e3ad](https://github.com/sigp/lighthouse/pull/8896/commits/5b4e3add49e9b9b0b567c2acd5748b2881aeb125)

jimmygchen (dismissed):
This one still needs to be addressed:

https://github.com/sigp/lighthouse/pull/8896/changes#r2895335958

jimmygchen:
> seems we were missing count validation for both envelopes and data columns [5b4e3ad](https://github.com/sigp/lighthouse/pull/8896/commits/5b4e3add49e9b9b0b567c2acd5748b2881aeb125)

ah we have the check here for data columns here, but what you have done is probably better as we avoid wasting a beacon processor slot:
https://github.com/sigp/lighthouse/blob/691c8cf8e69d4d40c4969c2bc493dc1eed9af99f/beacon_node/network/src/network_beacon_processor/rpc_methods.rs#L1195-L1201

beacon_node/beacon_chain/src/execution_payload_envelope_streamer.rs
ethDreamer: That was very strange.. the github interface wasn't showing me the update.. even after refreshing. Had to click back to the discussion and then back again to see it. Whatever I see it now.

eserilev:
There's potentially a bug on devnet 0 where we're serving non canonical envelope data. I'm investigating

beacon_node/lighthouse_network/src/rpc/protocol.rs
eserilev: We still need to set  `rpc_block_limits_by_fork` max to be bellatrix_max for historical sync compatibility

consensus/types/src/execution/execution_payload_envelope.rs
```diff
@@ -8,7 +8,9 @@ use ssz_derive::{Decode, Encode};
 use test_random_derive::TestRandom;
 use tree_hash_derive::TreeHash;
 
-#[derive(Debug, Clone, Serialize, Encode, Decode, Deserialize, TestRandom, TreeHash, Educe)]
+#[derive(
```
pawanjay176: Can we create an explicit `empty()` method for the default case?

consensus/types/src/execution/signed_execution_payload_envelope.rs
```diff
@@ -22,6 +23,36 @@ pub struct SignedExecutionPayloadEnvelope<E: EthSpec> {
...
+            message: ExecutionPayloadEnvelope::default(),
+            signature: Signature::empty(),
+        }
+        .as_ssz_bytes()
+        .len()
+    }
+
+    /// Returns the maximum SSZ-encoded size.
+    #[allow(clippy::arithmetic_side_effects)]
+    pub fn max_size() -> usize {
```
pawanjay176, line 38: might be better to have a max_size on the `ExecutionPayloadEnvelope` as well and use it here instead so we don't forget to update it if the spec changes

common/task_executor/src/lib.rs
```diff
@@ -12,6 +12,26 @@ use crate::rayon_pool_provider::RayonPoolProvider;
 pub use crate::rayon_pool_provider::RayonPoolType;
 pub use tokio::task::JoinHandle;
 
+/// Error type for spawning a blocking task and awaiting its result.
+#[derive(Debug)]
+pub enum SpawnBlockingError {
```
pawanjay176: I'm not a fan of doing this without refactoring the `spawn_blocking_handle` function in beacon_chain.rs to also use this. 
We'll just end up having multiple functions that do the same thing for no reason.

beacon_node/beacon_chain/src/execution_payload_envelope_streamer.rs
```diff
@@ -0,0 +1,638 @@
...
+#[derive(Debug)]
+pub enum Error {
+    ChildBlockNotFound,
+}
+
+pub struct PayloadEnvelopeStreamer<T: BeaconChainTypes> {
+    // TODO(gloas) remove expect when execution layer field
+    // is no longer dead.
+    #[expect(dead_code)]
+    execution_layer: ExecutionLayer<T::EthSpec>,
```
pawanjay176: Why don't we start with using the beacon chain here instead and do beacon_chain.execution_layer when we need to access it? that way we get to keep pretty similar logic to the beaocn block streamer.
We need the beacon_chain to access the db anyway

beacon_node/beacon_chain/src/execution_payload_envelope_streamer.rs
```diff
@@ -0,0 +1,638 @@
...
+
+    fn load_envelope(
+        self: &Arc<Self>,
+        beacon_block_root: &Hash256,
+    ) -> Result<Option<Arc<SignedExecutionPayloadEnvelope<T::EthSpec>>>, BeaconChainError> {
+        if let Some(cached_envelope) = self.check_payload_envelope_cache(beacon_block_root) {
+            Ok(Some(cached_envelope))
+        } else {
+            // TODO(gloas) we'll want to use the execution layer directly to call
+            //  the engine api method eth_getBlockByHash()
```
pawanjay176: do you mean getPayloadBodiesByRange or some equivalent thing that returns in the right format for the CL?

eserilev:
This is ready for a review. We're skipping envelope verification here, since we dont have fork choice yet.

dapplion:
Suggestion: avoid wrapping `canonical_head` in `Arc` — `BeaconChain` is already behind an `Arc` everywhere, so the streamer can hold `Arc<BeaconChain<T>>` instead.

Implement `CanonicalHeadReader` on `BeaconChain<T>` (delegating to `self.canonical_head`) and pass `self.clone()` to the streamer. This avoids the structural change to the `canonical_head` field.

See: https://github.com/dapplion/lighthouse/commit/3e6afae31

dapplion:
Also, `BeaconBlockGloas::full()` is dead code — it's defined but never called (`SignedExecutionPayloadEnvelope::max_size()` computes the size arithmetically instead). Beyond being unused, the impl is fragile: it starts from `empty(spec)` then mutates fields with push loops, so adding a new field compiles fine but silently leaves it empty. A direct struct construction would catch missing fields at compile time.

Removed it along with 9 unused imports it pulled in: https://github.com/dapplion/lighthouse/commit/ed8c5495f

eserilev:
@dapplion your changes make sense, i went ahead and merged them. thanks!

beacon_node/beacon_chain/src/execution_payload_envelope_streamer.rs
pawanjay176: If testing is the concern, we can take a similar approach to the fetch_blobs adapter. 

Here, it seems like we need ExecutionLayer, CanonicalHead and the store all of which are present in the BeaconChain

beacon_node/beacon_chain/src/execution_payload_envelope_streamer.rs
eserilev: yes, i've updated the comment

beacon_node/beacon_chain/src/execution_payload_envelope_streamer.rs
eserilev: ive went with the adapter pattern

common/task_executor/src/lib.rs
eserilev: this is no longer needed due to the adapter pattern change and has been removed

consensus/types/src/execution/signed_execution_payload_envelope.rs
eserilev: 54dfa0c

consensus/types/src/execution/execution_payload_envelope.rs
eserilev: 54dfa0c

eserilev:
went ahead with the adapter pattern, i think its the cleaner approach. thanks pawan for suggesting it. theres a bit of code churn from those changes, sorry about that... this should be ready for another round of reviews

dapplion:
🤖 Note on memory amplification (inherited pattern, not new to this PR):

`load_envelopes()` materializes all results into a `Vec` before streaming, and `launch_stream()` uses an `unbounded_channel()`. With `max_request_payloads = 128` and practical envelope sizes ~2MB (gas-limited), one request buffers ~256MB. Per-peer limits allow 4 concurrent requests (2 by-range + 2 by-root via `MAX_CONCURRENT_REQUESTS=2`), so ~1GB per peer. With 220 max peers, theoretical worst case is ~220GB.

(lion) However in practice the Beacon processor has limited worker threads, so not all 1024 queued items run simultaneously, on high CPU machine it may be a slight concern.

The `BlockStreamer` has the exact same pattern and has been running in production, so this isn't a regression — just worth noting as the envelope payloads can be larger than blocks. Rate limiting (128 items/10s per peer) and peer scoring provide some mitigation.

beacon_node/beacon_chain/src/payload_envelope_streamer/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+    request_source: EnvelopeRequestSource,
+}
+
+// TODO(gloas) eventually we'll need to expand this to support loading blinded payload envelopes from the db
+// and fetching the execution payload from the EL. See BlockStreamer impl as an example
+impl<T: BeaconChainTypes> PayloadEnvelopeStreamer<T> {
+    pub(crate) fn new(
+        adapter: EnvelopeStreamerBeaconAdapter<T>,
+        request_source: EnvelopeRequestSource,
+    ) -> Arc<Self> {
```
pawanjay176, line 46: Any reason why we are returning an Arc here?

beacon_node/beacon_chain/src/payload_envelope_streamer/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+        executor.spawn(
+            self.stream_payload_envelopes(block_roots, envelope_tx),
+            "get_payload_envelopes_sender",
+        );
+        UnboundedReceiverStream::new(envelope_rx)
+    }
+}
+
+/// Create a `PayloadEnvelopeStreamer` from a `BeaconChain` and launch a stream.
+#[cfg(not(test))]
```
pawanjay176, line 197: can this just be removed?

beacon_node/beacon_chain/src/payload_envelope_streamer/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+                            // Ensure that the envelopes we're serving match our view of the canonical chain.
+
+                            // When loading envelopes before the split slot, there is no need to check.
+                            // Non-canonical payload envelopes will have already been pruned.
+                            if split_slot > envelope.slot() {
+                                results.push((*root, Ok(Some(envelope))));
+                                continue;
+                            }
+
+                            match streamer.adapter.block_has_canonical_payload(root) {
```
pawanjay176, line 127: Just to clarify, this is for when a peer requests something by range which hasn't been finalized. In that case, even if the payload doesn't have weight, we would still have it in the hot db and end up returning it? I'm not completely clear as to why byroot requests should return the non-canonical payload in that case. Can we document the behaviour for the function so that its clear to the callers as well?

beacon_node/beacon_chain/src/payload_envelope_streamer/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+        };
+
+        for (root, result) in results {
+            if sender.send((root, Arc::new(result))).is_err() {
+                break;
+            }
+        }
+    }
+
+    pub fn launch_stream(
```
pawanjay176, line 178: this doesn't need to be pub i think?

beacon_node/beacon_chain/src/payload_envelope_streamer/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+
+async fn send_errors<E: EthSpec>(
+    block_roots: &[Hash256],
+    sender: UnboundedSender<(Hash256, Arc<PayloadEnvelopeResult<E>>)>,
+    beacon_chain_error: BeaconChainError,
+) {
+    let result = Arc::new(Err(beacon_chain_error));
+    for beacon_block_root in block_roots {
+        if sender.send((*beacon_block_root, result.clone())).is_err() {
+            break;
```
pawanjay176, line 216: Should probably log an error here to indicate the issue if the channel ever closes unexpectedly?

beacon_node/lighthouse_network/src/rpc/rate_limiter.rs
```diff
@@ -109,7 +109,11 @@ pub struct RPCRateLimiter {
     blbrange_rl: Limiter<PeerId>,
     /// BlobsByRoot rate limiter.
     blbroot_rl: Limiter<PeerId>,
-    /// DataColumnssByRoot rate limiter.
+    /// PayloadEnvelopesByRange rate limiter.
+    perange_rl: Limiter<PeerId>,
```
pawanjay176: envrange_rl maybe?

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -983,6 +1083,176 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+                        envelopes_sent += 1;
+                        self.send_network_message(NetworkMessage::SendResponse {
+                            peer_id,
+                            inbound_request_id,
+                            response: Response::PayloadEnvelopesByRange(Some(envelope.clone())),
+                        });
+                    }
+                }
+                Ok(None) => {
+                    trace!(
```
pawanjay176, line 1216: maybe this should be a debug. this shouldn't happen too often and if it does happen then it may be a bug on our side?

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
```diff
@@ -983,6 +1083,176 @@ impl<T: BeaconChainTypes> NetworkBeaconProcessor<T> {
...
+        let req_start_slot = req.start_slot;
+        let req_count = req.count;
+
+        debug!(
+            %peer_id,
+            count = req_count,
+            start_slot = %req_start_slot,
+            "Received ExecutionPayloadEnvelopesByRange Request"
+        );
+
```
pawanjay176, line 1129: maybe we should check if the slot is post gloas before looking into the db?

pawanjay176 (commented):
This is looking good. Keen to merge it soon. A few nits

beacon_node/network/src/router.rs
```diff
@@ -309,6 +327,11 @@ impl<T: BeaconChainTypes> Router<T> {
             Response::DataColumnsByRange(data_column) => {
                 self.on_data_columns_by_range_response(peer_id, app_request_id, data_column);
             }
+            // TODO(EIP-7732): implement outgoing payload envelopes by range and root
+            // responses once sync manager requests them.
+            Response::PayloadEnvelopesByRoot(_) | Response::PayloadEnvelopesByRange(_) => {
+                unreachable!()
```
pawanjay176: can we just log here instead of adding an unreachable? might be some time before we fix this

beacon_node/beacon_chain/src/payload_envelope_streamer/mod.rs
eserilev: byroots must serve envelopes even if they're non-canonical or it wouldn't be possible to sync other branches. I believe its the same for blocks and columns by root

for non-finalized data, byrange will check that the envelope is canonical before serving it. 

I added another another comment under the by roots logic here

beacon_node/beacon_chain/src/payload_envelope_streamer/mod.rs
eserilev: it does need to be pub, we call it outside of the struct impl

beacon_node/beacon_chain/src/payload_envelope_streamer/mod.rs
eserilev: this is the entry point for calling the envelope streamer from the beacon chain. It helps keep `EnvelopeStreamerBeaconAdapter` encapsulated within the `payload_envelope_streamer` module

beacon_node/beacon_chain/src/payload_envelope_streamer/mod.rs
eserilev: The only pub function in `EnvelopeStreamer` requires an `Arc<Self>`. This also follows the block streamer pattern

beacon_node/network/src/router.rs
eserilev: added a log

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
eserilev: added a check that the start slot is post gloas

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
eserilev: i dont think this would be a bug on our side, this could just be that theres a missed envelope for a slot (maybe the builder withheld, it was propagated late, re-orged for w/e reason etc). Any non-canonical envelopes would be pruned post finalization, so I'm not sure how often it would happen but this wouldn't be a bug. 

I was actually debating not logging anything (thats what we do for `data_columns_by_range`).

beacon_node/lighthouse_network/src/rpc/rate_limiter.rs
eserilev: updated names

beacon_node/beacon_chain/src/payload_envelope_streamer/mod.rs
eserilev: added an error log

beacon_node/network/src/network_beacon_processor/rpc_methods.rs
pawanjay176: Ohh that's a good point. I think not logging anything should be good too then

beacon_node/beacon_chain/src/payload_envelope_streamer/mod.rs
pawanjay176: But we are still within the same module though right. Its fine though. I thought it might just be potentially confusing if the caller should call `launch_stream` or `launch_payload_envelope_stream`

pawanjay176 (approved):
I think this looks good now. It is mostly separate from the current logic and the rpc methods won't be enabled until gloas is enabled so this is safe on mainnet too. I'm happy to merge this and iterate on this in later

pawanjay176:
@mergify requeue

eserilev:
@mergify requeue

jimmygchen:
@mergify dequeue

jimmygchen:
@mergify requeue

PR #8898, opened 2026-02-24
Add ai agent safety tooling

dapplion:
Claude recommended this strategies to enforce some of our agent rules programatically. Looks interesting, opening as work-in-progress to discuss it

- stop hook: blocks agent if cargo check fails
- consensus safety hook: warns on .unwrap(), unsafe arithmetic in consensus/
- /test skill: maps changed files to correct test commands
- /safety-check skill: diff-based safety analysis
- /devnet skill: local testnet with assertoor checks
- test-impact mapping in DEVELOPMENT.md
- crate-level CLAUDE.md for beacon_chain, state_processing, store
- enhanced pre-commit hook with consensus safety checks

## Issue Addressed

## Proposed Changes

## Additional Info

### DIFF
--- a/.claude/hooks/consensus-safety-check.sh
+++ b/.claude/hooks/consensus-safety-check.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+# Warn agent about unsafe patterns in critical code paths.
+# Runs after Edit/Write. Non-blocking (warns, doesn't block).
+
+set -uo pipefail
+
+INPUT=$(cat)
+FILE_PATH=$(echo "$INPUT" | jq -r '.tool_input.file_path // empty')
+
+# Only check .rs files in critical paths
+if [[ -z "$FILE_PATH" ]]; then exit 0; fi
+if [[ "$FILE_PATH" != *.rs ]]; then exit 0; fi
+if [[ "$FILE_PATH" != *"consensus/"* ]] && [[ "$FILE_PATH" != *"beacon_chain/"* ]]; then exit 0; fi
+if [[ "$FILE_PATH" == *"consensus/types/"* ]]; then exit 0; fi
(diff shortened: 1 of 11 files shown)


.claude/hooks/consensus-safety-check.sh
michaelsproul, line 1: This is unnecessary when we check this more comprehensively with clippy here:

https://github.com/sigp/lighthouse/blob/edba56b9a654cea555cb0db2e8d0712525686973/consensus/state_processing/src/lib.rs#L1-L13

.githooks/pre-commit
```diff
@@ -1,5 +1,38 @@
...
+        # Check for .expect( in added lines
+        if git diff --cached -- "$file" | grep '^\+' | grep -v '^\+\+\+' | grep -q '\.expect('; then
+            echo "ERROR: .expect() in consensus file: $file"
+            echo "  Use ? operator or .ok_or() instead."
+            FOUND_ISSUES=true
+        fi
+    done
+    if [ "$FOUND_ISSUES" = true ]; then
+        echo ""
+        echo "Consensus crate safety violations found. Fix before committing."
```
michaelsproul, line 35: This could just be `make lint`

michaelsproul:
My feeling on these sorts of LLM-enforced rules is that they're a bit fake. I think relying on the LLM to police itself seems doomed to fail if the LLM is intentionally sneaky or forgetful.

My preference is for externally-enforced guard-rails: Clippy lints, git commit hooks, strict limits on LLM permissions, sandboxing, etc. Anything "inside" the LLM seems suspect to me, but willing to be proved wrong.

pawanjay176:
> My preference is for externally-enforced guard-rails

Strongly agree!

michaelsproul:
Closing this as lacking consensus

I've been very happy with `nono` as a Claude sandbox btw: https://github.com/always-further/nono

PR #8901, opened 2026-02-25
Make ExecutionHash not option in ForkChoiceParams

dapplion:
## Issue Addressed

Builds on https://github.com/sigp/lighthouse/pull/8761. Now that merge transition code is being removed, execution block hashes no longer need to be Option types throughout the fork choice and beacon chain.

Two key changes:
1. Blocks before the merge are just marked as valid. This is only relevant for genesis sync. Setting the those block statuses to valid instead of irrelevant is a rename simplification for a rarely used path
2. If the finalized or justified checkpoint point to an unknown block root in fork-choice **that becomes a hard error**. Before in unstable, this failure was confusing / hidden since it just returned None = as blocks before the merge

---

- This PR helps make things simpler for https://github.com/sigp/lighthouse/pull/8382. In that PR it is possible for the justified and finalized checkpoints to point to unknown blocks, so we want to make that failure case clear and explicit

### DIFF
--- a/beacon_node/beacon_chain/src/canonical_head.rs
+++ b/beacon_node/beacon_chain/src/canonical_head.rs
@@ -109,13 +109,12 @@ pub struct CachedHead<E: EthSpec> {
     finalized_checkpoint: Checkpoint,
     /// The payload status of the head block, as determined by fork choice.
     head_payload_status: proto_array::PayloadStatus,
-    /// The `execution_payload.block_hash` of the block at the head of the chain. Set to `None`
-    /// before Bellatrix.
-    head_hash: Option<ExecutionBlockHash>,
-    /// The `execution_payload.block_hash` of the justified block. Set to `None` before Bellatrix.
-    justified_hash: Option<ExecutionBlockHash>,
-    /// The `execution_payload.block_hash` of the finalized block. Set to `None` before Bellatrix.
-    finalized_hash: Option<ExecutionBlockHash>,
+    /// The `execution_payload.block_hash` of the block at the head of the chain.
+    head_hash: ExecutionBlockHash,
+    /// The `execution_payload.block_hash` of the justified block.
+    justified_hash: ExecutionBlockHash,
+    /// The `execution_payload.block_hash` of the finalized block.
+    finalized_hash: ExecutionBlockHash,
 }
 
 impl<E: EthSpec> CachedHead<E> {
@@ -267,25 +266,25 @@ impl<T: BeaconChainTypes> CanonicalHead<T> {
     pub fn new(
         fork_choice: BeaconForkChoice<T>,
         snapshot: Arc<BeaconSnapshot<T::EthSpec>>,
+        head_block_root: Hash256,
         head_payload_status: proto_array::PayloadStatus,
-    ) -> Self {
-        let fork_choice_view = fork_choice.cached_fork_choice_view();
-        let forkchoice_update_params = fork_choice.get_forkchoice_update_parameters();
+    ) -> Result<Self, Error> {
+        let params = forkchoice_update_parameters::<T>(&fork_choice, head_block_root)?;
         let cached_head = CachedHead {
             snapshot,
-            justified_checkpoint: fork_choice_view.justified_checkpoint,
-            finalized_checkpoint: fork_choice_view.finalized_checkpoint,
+            justified_checkpoint: fork_choice.justified_checkpoint(),
+            finalized_checkpoint: fork_choice.finalized_checkpoint(),
             head_payload_status,
-            head_hash: forkchoice_update_params.head_hash,
-            justified_hash: forkchoice_update_params.justified_hash,
-            finalized_hash: forkchoice_update_params.finalized_hash,
+            head_hash: params.head_hash,
+            justified_hash: params.justified_hash,
+            finalized_hash: params.finalized_hash,
         };
 
-        Self {
+        Ok(Self {
             fork_choice: CanonicalHeadRwLock::new(fork_choice),
             cached_head: CanonicalHeadRwLock::new(cached_head),
             recompute_head_lock: Mutex::new(()),
-        }
+        })
     }
 
     /// Load a persisted version of `BeaconForkChoice` from the `store` and restore `self` to that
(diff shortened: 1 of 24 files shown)


dapplion (commented):
Deleted

consensus/proto_array/src/proto_array_fork_choice.rs
```diff
@@ -39,94 +39,39 @@ pub enum ExecutionStatus {
...
     /// An EL has not yet verified the execution payload.
     Optimistic(ExecutionBlockHash),
-    /// The block is either prior to the merge fork, or after the merge fork but before the terminal
-    /// PoW block has been found.
-    ///
-    /// # Note:
-    ///
-    /// This `bool` only exists to satisfy our SSZ implementation which requires all variants
-    /// to have a value. It can be set to anything.
-    Irrelevant(bool),
```
dapplion, line 100: @michaelsproul we know that no blocks in fork-choice have this status. If we remove this variant knowing that it is un-used, we are fine right?

dapplion:
Let's hold up to have a design of fork-choice Gloas

beacon_node/beacon_chain/src/canonical_head.rs
```diff
@@ -109,13 +109,12 @@ pub struct CachedHead<E: EthSpec> {
...
-    /// The `execution_payload.block_hash` of the justified block. Set to `None` before Bellatrix.
-    justified_hash: Option<ExecutionBlockHash>,
-    /// The `execution_payload.block_hash` of the finalized block. Set to `None` before Bellatrix.
-    finalized_hash: Option<ExecutionBlockHash>,
+    /// The `execution_payload.block_hash` of the block at the head of the chain.
+    head_hash: ExecutionBlockHash,
+    /// The `execution_payload.block_hash` of the justified block.
+    justified_hash: ExecutionBlockHash,
+    /// The `execution_payload.block_hash` of the finalized block.
+    finalized_hash: ExecutionBlockHash,
```
dapplion, line 117: This is the key change: when the head changes we MUST know the execution hashes of its head / just / fin blocks

beacon_node/beacon_chain/src/canonical_head.rs
```diff
@@ -1516,3 +1517,36 @@ fn observe_head_block_delays<E: EthSpec, S: SlotClock>(
...
+        justified_hash: fork_choice
+            .get_block_execution_block_hash(&fork_choice.justified_checkpoint().root)
+            .ok_or(Error::BlockMissingFromForkChoice(
+                fork_choice.justified_checkpoint().root,
+            ))?,
+        finalized_hash: fork_choice
+            .get_block_execution_block_hash(&fork_choice.finalized_checkpoint().root)
+            .ok_or(Error::BlockMissingFromForkChoice(
+                fork_choice.finalized_checkpoint().root,
+            ))?,
```
dapplion, line 1549: We expect fork-choice to contain nodes for the head + justified block + finalized block. This invariant is always true and pruning removes nodes _prior_ to finalized and justified is always a descendant of finalized. When we initialize fork-choice there's always a node for the justified and finalized checkpoint

beacon_node/beacon_chain/src/execution_payload.rs
```diff
@@ -57,8 +56,19 @@ impl<T: BeaconChainTypes> PayloadNotifier<T> {
     ) -> Result<Self, BlockError> {
         let payload_verification_status = if block.fork_name_unchecked().gloas_enabled() {
             // Gloas blocks don't contain an execution payload.
-            Some(PayloadVerificationStatus::Irrelevant)
-        } else if is_execution_enabled(state, block.message().body()) {
+            Some(PayloadVerificationStatus::Verified)
```
dapplion, line 60: We coerce Irrelevant status to Verified to not have to deal with Option return types. Semantically it's the same.

consensus/fork_choice/src/fork_choice.rs
```diff
@@ -365,8 +357,6 @@ pub struct ForkChoice<T, E> {
     proto_array: ProtoArrayForkChoice,
     /// Attestations that arrived at the current slot and must be queued for later processing.
     queued_attestations: Vec<QueuedAttestation>,
-    /// Stores a cache of the values required to be sent to the execution layer.
-    forkchoice_update_parameters: ForkchoiceUpdateParameters,
```
dapplion, line 369: Caching this here is unnecessary. The values are already cached in canonical head. Removing them from here simplifies reasoning about them and moves the error condition (justified / finalized nodes not present) outside of the fork-choice code

beacon_node/beacon_chain/src/canonical_head.rs
```diff
@@ -306,17 +304,14 @@ impl<T: BeaconChainTypes> CanonicalHead<T> {
         let mut fork_choice =
             <BeaconChain<T>>::load_fork_choice(store.clone(), reset_payload_statuses, spec)?
                 .ok_or(Error::MissingPersistedForkChoice)?;
-        let current_slot_for_head = fork_choice.fc_store().get_current_slot();
-        let (_, head_payload_status) = fork_choice.get_head(current_slot_for_head, spec)?;
-        let fork_choice_view = fork_choice.cached_fork_choice_view();
```
dapplion, line 311: No cached view, we recompute the head on init

consensus/fork_choice/src/fork_choice.rs
```diff
@@ -415,17 +405,21 @@ where
...
                 )
             } else if let Ok(execution_payload) = anchor_block.message().execution_payload() {
                 // Pre-Gloas forks: do not set payload hashes, they are only used post-Gloas.
                 if execution_payload.is_default_with_empty_roots() {
-                    (ExecutionStatus::irrelevant(), None, None)
+                    (
+                        ExecutionStatus::Valid(ExecutionBlockHash::zero()),
+                        None,
+                        None,
+                    )
```
dapplion, line 422: Fork-choice does not process pre-merge blocks, and this change is cosmetic. If we accept having Valid(0) for this case, the simplicity of not having to think about Option in downstream code is worth it.

consensus/proto_array/src/proto_array_fork_choice.rs
michaelsproul: I think we should update the v29 schema migration to do Irrelevant -> Valid(0x00). That would require keeping this variant around a little longer too.

consensus/proto_array/src/proto_array_fork_choice.rs
```diff
@@ -1008,9 +966,16 @@ impl ProtoArrayForkChoice {
             next_epoch_shuffling_id: block.next_epoch_shuffling_id().clone(),
             justified_checkpoint: *block.justified_checkpoint(),
             finalized_checkpoint: *block.finalized_checkpoint(),
-            execution_status: block
-                .execution_status()
-                .unwrap_or_else(|_| ExecutionStatus::irrelevant()),
+            // V29 (Gloas) nodes don't have execution_status. Use
+            // execution_payload_block_hash which is always present on V29
+            // (the zero fallback is cosmetic and unreachable).
+            execution_status: block.execution_status().unwrap_or_else(|_| {
```
michaelsproul, line 972: Could maybe use `None` for Gloas instead, to prevent misuse like:

- get_head -> block root B
- get_block(B)
- look at block_B.execution_status
- assume block_B.execution_status is valid (e.g. send to EL in fcU), even though we may not have seen this block

Or another option could be to check `payload_received` and set `Valid` if it is true, or `Optimistic` otherwise. Although this is also a conflation.

PR #8904, opened 2026-02-26
Add HTTP API spec test to Lighthouse

chong-he:
Add HTTP spec tests to Lighthouse. This is done by pulling the spec from the beacon API repository: https://github.com/ethereum/beacon-APIs/tree/master, then using [oas3](https://docs.rs/oas3/latest/oas3/) crate to parse the spec. We then make HTTP requests to Lighthouse for the beacon node endpoints, and compare Lighthouse response to the spec in terms of return fields, return type and regex pattern check for String type.

## Additional Info
- ~~Currently blocked, waiting for implementation of the `proposer_lookahead` endpoint~~ 
- ~~Currently only include `GET` endpoint,~~ ~~and some "live data" aren't checked yet (e.g, voluntary exit, pending deposits etc)~~ 
- All GET and POST responses (for those POST endpoints that have a response body) are checked
- All POST request body are checked
- Live data are generated in the test (e.g., voluntary exit, pending deposits) so that the GET endpoints can have a response to be checked
- Done some test to modify the output and the spec test catches and difference between Lighthouse and the spec, and failed.

Thank you to @michaelsproul and @macladson who have been very helpful.


beacon_node/http_api/tests/spec_tests.rs
```diff
@@ -0,0 +1,502 @@
...
+
+        // complement the check above to check the total number of required fields are the same
+        // result may contain more fields in some endpoints (e.g., extra_data in /eth/v1/debug/fork_choice)
+        // missing proposer_lookahead in the response for /eth/v2/debug/beacon/states/{state_id}
+        if endpoint == "/eth/v2/debug/beacon/states/{state_id}"
+            || endpoint == "/eth/v1/debug/fork_choice"
+            || endpoint == "/eth/v1/node/identity"
+        {
+        } else {
+            assert_eq!(
```
michaelsproul: As discussed, I think this check is too strict.

Instead we could check that all the result fields are one of the fields from the schema (i.e. no extra fields that aren't either required or optional). This will probably only fail for the state endpoint until we update to the version that includes `BeaconStateFulu`.

chong-he: `assert_eq` strict check is removed in [a9cdc74](https://github.com/sigp/lighthouse/pull/8904/commits/a9cdc744de3fb8726a0ce3a91a97b8818803e9a5)

beacon_node/http_api/Cargo.toml
```diff
@@ -54,7 +53,8 @@ warp_utils = { workspace = true }
 
 [dev-dependencies]
 genesis = { workspace = true }
-proto_array = { workspace = true }
```
chong-he, line 57: This, and the `either` and `lru` unused dependencies is the tips given by Rust-Rover IDE, I try removing them and it compiles, so I think we are not using them

chong-he:
Added POST endpoints to the test and it catches a spec diff with the beacon-APIs:
`Error: "Endpoint /eth/v1/beacon/rewards/attestations/{epoch} missing required field `finalized`"`

~~I tested and this endpoint indeed doesn't have the `finalized` field in the response:~~
```
{
  "execution_optimistic": true,
  "data": {
    "ideal_rewards": [
      {
        "effective_balance": "1000000000",
        "head": "68",
        "target": "132",
        "source": "71",
        "inactivity": "0"
      },
```
~~Will work on a fix for the endpoint.~~

~~Marking the PR as blocked now.~~

Edit: this is fixed in #8967

beacon_node/http_api/Cargo.toml
macladson: Nice! `proto_array` actually is used but is already a regular dep, so there's no need to duplicate it here. This is good :+1:

beacon_node/http_api/tests/spec_tests.rs
```diff
@@ -0,0 +1,947 @@
...
+        listening_socket,
+        external_peer_id,
+        network_rx,
+        ..
+    } = create_api_server(harness.chain.clone(), &harness.runtime).await;
+
+    harness.runtime.task_executor.spawn(server, "api_server");
+
+    let port = listening_socket.port();
+    let client = BeaconNodeHttpClient::new(
```
macladson: I think we can probably skip this and just use raw `reqwest`. We deliberately don't use any of the typed client methods and just call `get` and `post_with_response` directly anyway. Using raw `reqwest` would remove the need to make `get` and `post_with_response` `pub` as well which is nice.

common/eth2/src/lib.rs
```diff
@@ -201,7 +201,7 @@ impl BeaconNodeHttpClient {
     }
 
     /// Perform a HTTP GET request.
-    async fn get<T: DeserializeOwned, U: IntoUrl>(&self, url: U) -> Result<T, Error> {
+    pub async fn get<T: DeserializeOwned, U: IntoUrl>(&self, url: U) -> Result<T, Error> {
```
macladson: No need to make this `pub`. See my above comment

common/eth2/src/lib.rs
```diff
@@ -356,7 +356,7 @@ impl BeaconNodeHttpClient {
     }
 
     /// Perform a HTTP POST request, returning a JSON response.
-    async fn post_with_response<T: Serialize, U: IntoUrl, R: DeserializeOwned>(
+    pub async fn post_with_response<T: Serialize, U: IntoUrl, R: DeserializeOwned>(
```
macladson: Same here

beacon_node/http_api/tests/spec_tests.rs
```diff
@@ -0,0 +1,947 @@
...
+        blinded_block,
+    }
+}
+
+// Extract the full ObjectSchema for each endpoint, the ObjectSchema contains all info that we need for the checks
+async fn extract_all_endpoints() -> ObjectSchemaByEndpoint {
+    // Obtain the complete Beacon APIs yaml file using the latest release version (not the dev versino)
+    // TODO: switch to latest release before the Gloas upgrade
+    let yaml = reqwest::get(
+        "https://github.com/ethereum/beacon-APIs/releases/download/v4.0.0/beacon-node-oapi.yaml",
```
macladson: Is there a way we can automatically pull the latest release rather than hard coding a version? This seems like something that we could easily forget about.

I'm also wondering if we should follow the `ef-spec` pattern and download the spec separately to the test with a makefile. That might be overkill though since this is just a single file.

macladson (commented):
Nice this is pretty exciting! Just a few comments

beacon_node/http_api/tests/spec_tests.rs
chong-he: This is a great feedback, thank you. 

I simplify to use raw `reqwest` here: [d1d4919](https://github.com/sigp/lighthouse/pull/8904/commits/d1d49190b0695ad36a471776a25505661858609f)

common/eth2/src/lib.rs
chong-he: As above

common/eth2/src/lib.rs
chong-he: As above

beacon_node/http_api/tests/spec_tests.rs
chong-he: I change to always use the latest release link:

https://github.com/ethereum/beacon-APIs/releases/latest/download/beacon-node-oapi.yaml

Seems to work, now that upstream repository has a pre-release `v5.0.0-alpha.1` but the link is still on v4.0.0 as shown in the yaml file:
`version: v4.0.0 - Ethereum Proof-of-Stake Consensus Specification v1.5.0`
which is good

I also deleted the TODO since the latest download looks like a solution for a need to manually update.

chong-he: The only issue I can think of is that once the latest version releases and Lighthouse still hasn't implemented some endpoints, the spec test will probably fail and blocking the CI. 

Currently, we have some endpoints waiting to be implemented: [github.com/sigp/lighthouse/issues/9045](https://github.com/sigp/lighthouse/issues/9045)
but most of them are very new except the `/v2/node/version`. But also maybe because they are so new that the latest version may not be released anytime soon so we have time to implement the new endpoints.

macladson: Yeah, thinking about it some more, since it is just a single file, I actually think we should check-in the `beacon-node-oapi.yaml` file itself (into git) instead of downloading it every test run. Then, as the file is updated upstream with new specs, we can ensure we pass the tests before updating it so we don't block CI. And we won't run the risk of hitting any Github rate limits or timeouts

chong-he: That's a good idea. I have made the changes in [25edf13](https://github.com/sigp/lighthouse/pull/8904/commits/25edf13875cf6c534f2eb55e4e3ede9855208bb2). This also makes the function to extract the endpoints become non-async which is nice.

I put the file `beacon-node-oapi.yaml` in the same directory as `spec_test.rs`. I am not sure if there is a better place to put this file? 

Edit: one "issue" is that we add 70k lines of code to the code base, is this all good?

chong-he: Added the logic to either use Github token to download the file (CI) or to download locally (when test locally).

PR #8905, opened 2026-02-26
perf(vc): skip redundant proposer duty downloads within epoch

### LINKED ISSUE #8457: Reduce validator client proposer duties polling with Fulu proposer lookahead

0xMars42:
## Issue Addressed

Closes #8457

## Proposed Changes

Skip the proposer duties HTTP download when duties for the current epoch
are already cached. The per-slot block service notification is preserved.

Reduces proposer duty requests from 32 to 1 per epoch.

### DIFF
--- a/validator_client/validator_services/src/duties_service.rs
+++ b/validator_client/validator_services/src/duties_service.rs
@@ -1513,9 +1513,10 @@ async fn fill_in_selection_proofs<S: ValidatorStore + 'static, T: SlotClock + 's
 /// through the slow path every time. I.e., the proposal will only happen after we've been able to
 /// download and process the duties from the BN.
 ///
-/// To reduce BN load, the HTTP download is skipped when duties for the current epoch are already
-/// cached. This reduces polling from 32 HTTP requests/epoch to 1, since proposer shuffling does
-/// not change within an epoch.
+/// To reduce redundant work, the response processing is skipped when the `dependent_root` from
+/// the beacon node matches the cached value. Proposer shuffling does not change within an epoch
+/// unless a re-org alters the `dependent_root`, so this avoids re-filtering and re-inserting
+/// identical duties on every slot while still detecting re-orgs.
 async fn poll_beacon_proposers<S: ValidatorStore, T: SlotClock + 'static>(
     duties_service: &DutiesService<S, T>,
     block_service_tx: &mut Sender<BlockServiceNotification>,
@@ -1542,18 +1543,6 @@ async fn poll_beacon_proposers<S: ValidatorStore, T: SlotClock + 'static>(
         duties_service.validator_store.as_ref(),
     )
     .await;
-
-    // Skip the HTTP download if we already have proposer duties for the current epoch.
-    // Proposer shuffling is determined at the epoch boundary and does not change within
-    // an epoch, so downloading again would return identical data. The cache notification
-    // above still fires every slot to ensure the block service produces blocks on time.
-    if duties_service.proposers.read().contains_key(&current_epoch) {
-        duties_service
-            .proposers
-            .write()
-            .retain(|&epoch, _| epoch + HISTORICAL_DUTIES_EPOCHS >= current_epoch);
-        return Ok(());
-    }
 
     // Collect *all* pubkeys, even those undergoing doppelganger protection.
     //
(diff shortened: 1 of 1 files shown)


validator_client/validator_services/src/duties_service.rs
```diff
@@ -1540,6 +1543,18 @@ async fn poll_beacon_proposers<S: ValidatorStore, T: SlotClock + 'static>(
...
+    // Proposer shuffling is determined at the epoch boundary and does not change within
+    // an epoch, so downloading again would return identical data. The cache notification
+    // above still fires every slot to ensure the block service produces blocks on time.
+    if duties_service.proposers.read().contains_key(&current_epoch) {
+        duties_service
+            .proposers
+            .write()
+            .retain(|&epoch, _| epoch + HISTORICAL_DUTIES_EPOCHS >= current_epoch);
+        return Ok(());
+    }
```
chong-he: An early return here will skip the `dependent_root` update and check that comes afterwards:

https://github.com/sigp/lighthouse/blob/3deab9b0410233c1d57bddfaa9903cc6fbdaa958/validator_client/validator_services/src/duties_service.rs#L1583-L1595

So if the `dependent_root` changes, the VC will not be aware

chong-he (commented):
I am not convinced that this is a required change, as it neglects the updated `dependent_root` (if there is one) that comes afterwards. Edit: maybe I miss something, so let's wait for a dev review on this

Edit2: going through older PR, this is a bit related to the issue #8457 and a PR for the issue #8574

0xMars42:
Thanks for the review @chong-he, good catch on the `dependent_root` you're right that the early return would miss reorgs.
What about adding a lightweight reorg check before skipping? Instead of calling `get_validator_duties_proposer` every slot, we could call `get_beacon_blocks_root(BlockId::Slot(epoch_start - 1))` to verify the `dependent_root` hasn't changed. If it matches the cached one skip, if not full re-fetch.
That way we keep the perf gain (1 full call + 31 lightweight root checks vs 32 full calls per epoch) while still catching reorgs. WDYT?

chong-he:
> Thanks for the review @chong-he, good catch on the `dependent_root` you're right that the early return would miss reorgs. What about adding a lightweight reorg check before skipping? Instead of calling `get_validator_duties_proposer` every slot, we could call `get_beacon_blocks_root(BlockId::Slot(epoch_start - 1))` to verify the `dependent_root` hasn't changed. If it matches the cached one skip, if not full re-fetch. That way we keep the perf gain (1 full call + 31 lightweight root checks vs 32 full calls per epoch) while still catching reorgs. WDYT?

It still a call to the beacon node though.

This PR seems to overlap with #8574 which also solves for the same issue. From #8574, we see that it modifies the call to per epoch and uses `sleep(duration).await` to wait till the next epoch. So the call to the function `poll_beacon_proposers` (which is the function being modified in this PR) is not called every slot. On this point, I would say the solution in #8574  is cleaner.

chong-he:
Thanks for the PR, after reviewing, I find that the PR #8574 is a cleaner solution. It is also opened earlier than this. So I am closing this in favour of #8574

